PR #1, opened 2021-04-12
Httpx networks

dalf:
## What does this PR do?

* outgoing.networks:
   * can contains network definition
   * propertiers: enable_http, verify, http2, max_connections, max_keepalive_connections,
     keepalive_expiry, local_addresses, support_ipv4, support_ipv6, proxies, max_redirects, retries
   * retries: 0 by default, number of times searx retries to send the HTTP request (using different IP & proxy each time)
   * local_addresses can be "192.168.0.1/24" (it supports IPv6)
   * support_ipv4 & support_ipv6: both True by default
     see searx/searx#1034
* each engine can define a "network" section:
   * either a full network description
   * either reference an existing network

* all HTTP requests of engine use the same HTTP configuration (it was not the case before, see proxy configuration in master)

## See

https://github.com/metasearch-lab/searx/wiki/Milestones#milestone-13---network


return42:
FYI: merged into https://github.com/return42/searx-next/commit/03f30f3244

PR #12, opened 2021-04-22
[mod] update /stats

dalf:
## What does this PR do?

image

## Related issues

https://github.com/searxng/searxng/wiki/Milestones#milestone-14---better-statistics-about-the-engines


return42:
I really like it ..  thanks a lot .. works perfect .. simple theme needs to be added.

I pushed c207031 on top of your branch .. includes some nits I had .. if you do not like c207031, just drop it and force-push.

return42:
@dalf if you think this DRAFT is ready to review, leave a note.

dalf:
Except the Makefile conflict, it is ready.

return42:
LGTM except the `@table-bg` in the dark theme

image

I tried to set `@table-bg` in the `searx/static/themes/oscar/src/less/logicodev-dark/oscar.less` but this won't help, since the wrong table background comes from the `searx/static/themes/oscar/src/less/bootstrap/bootstrap.less` which is used in the dark and bright theme.  But I'm not an expert in bootstrap & less.

PR #22, opened 2021-04-24
[docker] multiarch build

dalf:
Add a new Makefile target `make docker.buildx` to build and push a multiarch docker image (it can't be only build).

use buildx with the --cache-from and --cache-to options to cache the layers
only the last built is cached

---

About buildx:
* https://docs.docker.com/buildx/working-with-buildx/
* https://github.com/docker/buildx/#binary-release


return42:
Looks very interesting to me .. I hope I have the time to test it these days .. my docker skill is worse :-)

dalf:
As suggested https://github.com/searx/searx-docker/issues/27 , most probably it requires:
* a base image with all the dependencies (and built). This image is built only when the dependencies change.
* an second image using the base image, with only the python code.

Or even a dedicated ARM host.

dalf:
[EDIT] `--cache-from` and `--cache-to` options make it simple.

----

Since the job takes one hour:
* it is possible to create `searxng-base`  image with all the dependencies (distro & pip)
* the image is pulled : if it fails it is built and pushed.
* then `searxng` image uses this `searxng-base`.

So docker step takes one hour only when the dependencies change.

dalf:
ping @unixfox

unixfox (approved):

return42:
@unixfox thanks a lot for your review!.. @dalf I would like to test this PR also .. but ATM I do not have the time for .. it will take some days before I can start with.  Since @unifox already approved this PR .. feel free to merge (I can do my tests afterwards).

PR #28, opened 2021-04-26
[mod] processors: show identical error messages on /search and /stats

dalf:
## What does this PR do?

In `searx.search.processors.abstract.EngineProcessor`, change the signature of `handle_exception`: 
`def handle_exception(self, result_container, exception_or_message, suspend=False):`

it was previously:
`def handle_exception(self, result_container, reason, exception, suspend=False, display_exception=True):`

The exception class name is recorded instead of the `reason`.

The exception class name is translated to an user message in `webapp.py`, see `exception_classname_to_text`.

Note: offline engine exception are displayed as "unexpected crash" instead of `str(exception)`, but the error log in `/stats` displays the exception class name.

The error logs reported `/search` in the same way than in `/stats`

* search for `!stackoverflow test`, `!seznam test` and some other broken engines.

### DIFF
--- a/searx/search/processors/abstract.py
+++ b/searx/search/processors/abstract.py
@@ -59,22 +59,28 @@ class EngineProcessor(ABC):
         key = id(key) if key else self.engine_name
         self.suspended_status = SUSPENDED_STATUS.setdefault(key, SuspendedStatus())
 
-    def handle_exception(self, result_container, reason, exception, suspend=False, display_exception=True):
+    def handle_exception(self, result_container, exception_or_message, suspend=False):
         # update result_container
-        error_message = str(exception) if display_exception and exception else None
-        result_container.add_unresponsive_engine(self.engine_name, reason, error_message)
+        if isinstance(exception_or_message, BaseException):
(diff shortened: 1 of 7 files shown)


searx/search/processors/abstract.py
```diff
@@ -91,7 +97,7 @@ def _extend_container_basic(self, result_container, start_time, search_results):
     def extend_container(self, result_container, start_time, search_results):
         if getattr(threading.current_thread(), '_timeout', False):
             # the main thread is not waiting anymore
-            self.handle_exception(result_container, 'Timeout', None)
+            self.handle_exception(result_container, 'timeout', None)
```
dalf, line 107: `'timeout'` matches the key `timeout` of `exception_classname_to_text` (webapp.py)

return42 (approved):
LGTM!

Hint: I rebased your branch mod-processors-error-message and placed 924f9afea37 on top ... remove this commt if you think its not the right place.  For me, 924f9afea37 was a *product* of my review, so I added this commit on top  .. hope this is OK for you.

PR #32, opened 2021-04-27
[fix] searx.network: fix rare cases where LOOP is None

dalf:
## What does this PR do?

* searx.network.client.LOOP is initialized in a thread
* searx.network.__init__ imports LOOP which may happen
  before the thread has initialized LOOP

This commit adds a new function "searx.network.client.get_loop()"
to fix this issue

With uwsgi, some worker may be broken when LOOP is None

Not easily reproducible.

### DIFF
--- a/searx/network/client.py
+++ b/searx/network/client.py
@@ -120,7 +120,6 @@ class AsyncHTTPTransportFixed(httpx.AsyncHTTPTransport):
 
 
 def get_transport_for_socks_proxy(verify, http2, local_address, proxy_url, limit, retries):
-    global LOOP, TRANSPORT_KWARGS
     # support socks5h (requests compatibility):
     # https://requests.readthedocs.io/en/master/user/advanced/#socks
     # socks5://   hostname is resolved on client side
@@ -136,7 +135,7 @@ def get_transport_for_socks_proxy(verify, http2, local_address, proxy_url, limit
     return AsyncProxyTransportFixed(proxy_type=proxy_type, proxy_host=proxy_host, proxy_port=proxy_port,
                                     username=proxy_username, password=proxy_password,
                                     rdns=rdns,
-                                    loop=LOOP,
+                                    loop=get_loop(),
                                     verify=verify,
                                     http2=http2,
                                     local_address=local_address,
@@ -190,6 +189,11 @@ def new_client(enable_http, verify, enable_http2,
 
     transport = get_transport(verify, enable_http2, local_address, None, limit, retries)
     return httpx.AsyncClient(transport=transport, mounts=mounts, max_redirects=max_redirects)
+
+
+def get_loop():
+    global LOOP
+    return LOOP
 
 
 def init():
--- a/searx/network/__init__.py
+++ b/searx/network/__init__.py
@@ -9,7 +9,7 @@ import httpx
 import h2.exceptions
 
 from .network import get_network, initialize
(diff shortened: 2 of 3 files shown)


searx/network/client.py
```diff
@@ -136,7 +135,7 @@ def get_transport_for_socks_proxy(verify, http2, local_address, proxy_url, limit
     return AsyncProxyTransportFixed(proxy_type=proxy_type, proxy_host=proxy_host, proxy_port=proxy_port,
                                     username=proxy_username, password=proxy_password,
                                     rdns=rdns,
-                                    loop=LOOP,
+                                    loop=get_loop(),
```
return42, line 138: I rebased and pushed force ... BTW I added this replacement for the global LOOP variable .. I guess you missed it.

searx/network/client.py
```diff
@@ -192,6 +191,11 @@ def new_client(enable_http, verify, enable_http2,
     return httpx.AsyncClient(transport=transport, mounts=mounts, max_redirects=max_redirects)
 
 
+def get_loop():
+    global LOOP
```
dalf, line 195: There is no need `global LOOP` since LOOP is not assigned (?)

searx/network/client.py
dalf: Can you add another commit rather than push force, so I can easy see what has been changed ?

return42: > Can you add another commit

sure .. sorry.

dalf: NP

searx/network/client.py
return42: > There is no need global LOOP since LOOP is not assigned (?)

To access a global variable inside a function there is no need to use global keyword in general .. but I think it is good practice to use global keyword to explicit define a variable as _global_ instead of being [implicit](https://docs.python.org/3/faq/programming.html#what-are-the-rules-for-local-and-global-variables-in-python).

BTW: LOOP is assigned by `None` in line 25  / otherwise you would get an `NameError: name 'LOOP' is not defined`:

https://github.com/searxng/searxng/blob/283ae7bfad05fedbf6c7a7e7d7addfc3ae7ed9c7/searx/network/client.py#L25

TLDR: compared to C++ in real object oriented languages types are also objects, here is the id of the `None` type object:

    >>> id(None)
    9480720

dalf: >  I think it is good practice to use global keyword to explicit define a variable as global instead of being implicit. 

Ok.
But new code will have a lot of `global` since there are a lot of global variables.

return42: [On one hand, requiring global for assigned variables provides a bar against unintended side-effects. On the other hand, if global was required for all global references, you’d be using global all the time. You’d have to declare as global every reference to a built-in function or to a component of an imported module. This clutter would defeat the usefulness of the global declaration for identifying side-effects.](https://docs.python.org/3/faq/programming.html#what-are-the-rules-for-local-and-global-variables-in-python)

> But new code will have a lot of global since there are a lot of global variables.

Hm ... if it is to much, then we can switch to use implicit globals .. lets see ..I can't judge which is better for searx / I do use globals very rare in my projects.

return42:
If you think this PR is ready to merge .. please merge.  I can't give a _approved_ from github-GUI / I guess it is because I'm in the list of _committers_ of this PR.

PR #36, opened 2021-04-27
[version] update searxng version of branch master to 1.1.0-pre

return42:
## What does this PR do?

update searxng version of branch master to 1.1.0-pre

The version number of the branch master must to be detached from the version of the last release.

### DIFF
--- a/searx/version.py
+++ b/searx/version.py
@@ -1,26 +1,16 @@
-# -*- coding: utf-8 -*-
-'''
-searx is free software: you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
 
-searx is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Affero General Public License for more details.
+"""Version of searx
 
-You should have received a copy of the GNU Affero General Public License
-along with searx. If not, see < http://www.gnu.org/licenses/ >.
+"""
 
-(C) 2013- by Adam Tauber, <asciimoo@gmail.com>
-'''
-
-# version of searx
 VERSION_MAJOR = 1
-VERSION_MINOR = 0
-VERSION_BUILD = 0
+VERSION_MINOR = 1
+VERSION_BUILD = "0-pre"
 
-VERSION_STRING = "{0}.{1}.{2}".format(VERSION_MAJOR,
-                                      VERSION_MINOR,
-                                      VERSION_BUILD)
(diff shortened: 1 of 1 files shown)


dalf:
Should we keep `major.minor.patch` ?
When do we remove `-pre` ?

What about `year.month.day` ? 

I would like to add the last git commit with something like this:
```python
import pathlib

def get_git_revision(base_path):
    git_dir = pathlib.Path(base_path) / '.git'
    with (git_dir / 'HEAD').open('r') as head:
        ref = head.readline().split(' ')[-1].strip()

    with (git_dir / ref).open('r') as git_hash:
        return git_hash.readline().strip()

print(get_git_revision('.'))
```

[EDIT] more complete versions:
* https://gist.github.com/pkienzle/5e13ec07077d32985fa48ebe43486832
* https://github.com/spapas/python-git-info
* https://github.com/pypa/setuptools_scm

return42:
> When do we remove -pre ?

With the next release (major or minor).  After the (tagged) release commit, we have to increment minor-number again and add a  `-pre` ... or should we better use [`-dev`](https://www.python.org/dev/peps/pep-0440/#developmental-releases)?

> Should we keep major.minor.patch ?
> What about year.month.day ?

I do not have a preference to change the given  numbering, but in my projects I also use a _year month day_ numbering.  If you see some benefit in switching the numbering, lets do it.

Take into account, there is no way back since (e.g) `2021.x.y` in sense of semantic versioning comes  _after_ `1.x.y`. 

> I would like to add the last git commit with something like this:

I think this is not a very good idea .. this binds installation of a searxng version to an SCM system (git).  I think we should not establish such dependencies.

dalf:
What trigger a new version / what is the target of a new version ?

> I think this is not a very good idea .. this binds installation of a searxng version to an SCM system (git). I think we should not establish such dependencies.

When the user press the button "Submit a new issue on Github" in `/stats`, the git commit is useful to know which version of the engine is used (we can even make a link to the source on github). The version won't give enough information since it can cover different version of the code.

return42:
> When the user press the button "Submit a new issue on Github" in

I hope we will not receive to many duplicate issues .:) ... lets see.

> What trigger a new version / what is the target of a new version ?

Good question ..here are my 2cent 

... searx deploys packages on PyPi for this a version numbering is needed.  I think it is not a good idea to deploy searxng at PyPi.  I would like to ask "do we need versioning?" .. I don't think so, each commit is a new version I think.  By example: if a central engine like google is broken and a fix is available this fix needs to deploy on each instance as soon as possible.  Version numbers do not count here, I think.

I think I need to get away from the idea that searxng is a versioned program. 

With this in mind I see that your approach (using commit hash) seems to be the better one after all.

If we do so, if each commit is a version we can only deploy from the repository.  

... which means we bind deployment onto git (and a git hoster)
... this is a heavyweight consequence, I think

Side note: that was also a reason why I asked to place the issue form in `searx/templates/__common__` so we can easily  replace it by a template for a different issue system (from a different git hoster).

If I sum up my thoughts I would say that we can go this way but then, we should clearly exclude other pathes of  deployment (e.g. PyPi).

PR #38, opened 2021-04-28
[filtron.sh] make filtron rule file configurable

return42:
To select a different file with filtron rules, set environment

    FILTRON_RULES_TEMPLATE

the default is

    utils/templates/etc/filtron/rules.json

The installation is done by the new function install_rules() which offers a
multiple choice in case of collisions (known from searx.sh install setup).

update: a local test is not easy, you can do it in a LXC but I tested it on my instance.

### DIFF
--- a/utils/filtron.sh
+++ b/utils/filtron.sh
@@ -321,7 +321,7 @@ install_rules() {
     choose_one action "What should happen to the rules file? " \
            "keep configuration unchanged" \
            "use origin rules" \
-           "start interactiv shell"
+           "start interactive shell"
     case $action in
         "keep configuration unchanged")
             info_msg "leave rules file unchanged"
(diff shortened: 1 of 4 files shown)


utils/filtron.sh
```diff
@@ -303,6 +302,47 @@ go get -v -u github.com/asciimoo/filtron
...
+    case $action in
+        "keep configuration unchanged")
+            info_msg "leave rules file unchanged"
+            ;;
+        "use origin rules")
+            backup_file "${FILTRON_RULES}"
+            info_msg "install origin rules"
+            cp "${FILTRON_RULES_TEMPLATE}" "${FILTRON_RULES}"
+            ;;
+        "start interactiv shell")
```
dalf: small typo: `interactive`

return42: Oops .. fixed in 666bf1d

PR #40, opened 2021-04-28
[mod] engine errors: link to the stats to create an github issue

dalf:
#### Results

The engine name in error boxes are links to `/stats?engine=...` :
image

image

image

#### Worst case scenario:
image

#### In the preferences:
image

* The reliability on the right is a link too.
* The tooltip on the engine name contains a link to `/stats?engine=...`

* `/stats?engine=...` displays the detailed error logs.
* It create an incentive to create a bug report.

Note: when there are some results, the error box can be big if a lot of engines crash, but it is more readable.

With the simple and oscar theme:
* `!stackoverflow test`
* `!seznam !stackoverflow test`

### DIFF
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -12,6 +12,11 @@
     {%- if stats[search_engine.name]['result_count'] -%}
         <p>{{ _('Number of results') }}: {{ stats[search_engine.name]['result_count'] }} ( {{ _('Avg.') }} )</p>{{- "" -}}
     {%- endif -%}
+    {%- if reliabilities[search_engine.name].errors -%}
+        <a href="{{ url_for('stats', engine=search_engine.name|e) }}" title="{{ _('View error logs and submit a bug report') }}">
+          {{ _('View error logs and submit a bug report') }}
+        </a>
+    {%- endif -%}
 </div>
 {%- endif -%}
 {%- endmacro %}
@@ -420,22 +425,22 @@
             </div>
         </div>
 
-	<p class="text-muted">
-	  {{ _('These settings are stored in your cookies, this allows us not to store this data about you.') }}
+        <p class="text-muted">
+          {{ _('These settings are stored in your cookies, this allows us not to store this data about you.') }}
           {{ _("These cookies serve your sole convenience, we don't use these cookies to track you.") }}
         </p>
 
         <p>
-	  {{ _('Search URL of the currently saved preferences') }}
-	  <small class="text-muted">({{ _('Note: specifying custom settings in the search URL can reduce privacy by leaking data to the clicked result sites.') }})</small>:
+          {{ _('Search URL of the currently saved preferences') }}
+          <small class="text-muted">({{ _('Note: specifying custom settings in the search URL can reduce privacy by leaking data to the clicked result sites.') }})</small>:
         </p>
 
-	<div class="tab-pane">
+        <div class="tab-pane">
           <input readonly="" class="form-control select-all-on-click cursor-text" type="url" value="{{ url_for('index', _external=True) }}?preferences={{ preferences_url_params|e }}{% raw %}&amp;q=%s{% endraw %}">
           <input type="submit" class="btn btn-primary" value="{{ _('save') }}" />
         <a href="{{ url_for('index') }}"><div class="btn btn-default">{{ _('back') }}</div></a>
         <a href="{{ url_for('clear_cookies') }}"><div class="btn btn-default">{{ _('Reset defaults') }}</div></a>
-	</div>
+        </div>
     </form>
 </div>
 {% endblock %}
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -50,9 +50,11 @@
 {% endif %}
 {% if checker_result or errors %}
 <td class="{{ css_align_class }} {{ label }}">{{- "" -}}
-    <span aria-labelledby="{{engine_name}}_reliablity">
-        {%- if reliabilities[engine_name].checker %}{{ icon('exclamation-sign', 'The checker fails on the some tests') }}{% endif %} {{ r -}}
-    </span>{{- "" -}}
+    <a href="{{ url_for('stats', engine=engine_name|e) }}">{{- "" -}}
+        <span aria-labelledby="{{engine_name}}_reliablity">
+            {%- if reliabilities[engine_name].checker %}{{ icon('exclamation-sign', 'The checker fails on the some tests') }}{% endif %} {{ r -}}</a>
+        </span>{{- "" -}}
+    </a>{{- "" -}}
     <div class="engine-tooltip text-left" role="tooltip" id="{{engine_name}}_reliablity">
         {%- if checker_result -%}
         <p>{{ _("Failed checker test(s): ") }} {{ ', '.join(checker_result) }}</p>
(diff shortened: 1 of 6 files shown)


return42:
update: in 0bd35f7 the link is in the about-toolbox.

Does it make sense to have a link to the stats/engine page in the preferences also ... I mean to have a link in the in cell  "Engine name" ..

image

searx/templates/oscar/preferences.html
```diff
@@ -50,9 +55,11 @@ <h5><a href="{{about.website}}" rel="noreferrer">{{about.website}}</a></h5>
 {% endif %}
 {% if checker_result or errors %}
 <td class="{{ css_align_class }} {{ label }}">{{- "" -}}
-    <span aria-labelledby="{{engine_name}}_reliablity">
-        {%- if reliabilities[engine_name].checker %}{{ icon('exclamation-sign', 'The checker fails on the some tests') }}{% endif %} {{ r -}}
-    </span>{{- "" -}}
+    <a href="{{ url_for('stats', engine=engine_name|e) }}">{{- "" -}}
+        <span aria-labelledby="{{engine_name}}_reliablity">
+            {%- if reliabilities[engine_name].checker %}{{ icon('exclamation-sign', 'The checker fails on the some tests') }}{% endif %} {{ r -}}</a>
```
return42, line 60: Should the condition for the exclamation-sign based on _checker_ or on _errors_?

    {%- if reliabilities[engine_name].checker %}

versus ..

    {%- if reliabilities[engine_name].errors %}

Compare: In 0bd35f799 I used `reliabilities[<engine.name>].errors`

searx/templates/oscar/preferences.html
```diff
@@ -418,22 +425,22 @@ <h3>{{ _('Cookies') }}</h3>
...
         </p>
 
-	<div class="tab-pane">
+        <div class="tab-pane">
           <input readonly="" class="form-control select-all-on-click cursor-text" type="url" value="{{ url_for('index', _external=True) }}?preferences={{ preferences_url_params|e }}{% raw %}&amp;q=%s{% endraw %}">
           <input type="submit" class="btn btn-primary" value="{{ _('save') }}" />
         <a href="{{ url_for('index') }}"><div class="btn btn-default">{{ _('back') }}</div></a>
         <a href="{{ url_for('clear_cookies') }}"><div class="btn btn-default">{{ _('Reset defaults') }}</div></a>
-	</div>
+        </div>
```
return42, line 443: This is an _untabify_ from 0bd35f7 .. I forgot to mentioned this in the commit message / sorry.

return42:
@dalf I finished my review and added a patch on top .. drop it if you think it is not what you want.

dalf:
Alternative ideas: 

* the link can be on the error instead of the engine, so it is clear that link is about the error (actually stats and errors):
  stackoverflow [(CAPTCHA)](https://example.com/stats?engine=stackoverflow)
* the reliability values can always have a link (even at 100%).

return42:
> Alternative ideas:

Sorry I  am slow .. I did not understand what you mean.   In commit  0bd35f7 I implemented "View error logs and submit a bug report" in the toolbox, based on condition `reliabilities[<engine.name>].errors`.  

image

What me drives was the `0` and the missing `!` on the right side, which I thought is not very intuitive.  In https://github.com/searxng/searxng/pull/40#pullrequestreview-647258005 I suggested to use condition  `{%- if reliabilities[engine_name].errors %}` to get the `!` on the right side also 

ATM `{%- if reliabilities[engine_name].checker %}`is only given, if checker has at least runs one time .. but the error exists, even if the checker is not configured (will never be started).

searx/templates/oscar/preferences.html
dalf: Following what I have understood of the ARIA recommendation: yes, if should be `{%- if reliabilities[engine_name].errors %}`

---
The issue is how to display the checker result without displaying a spreadsheet.

Here some context:

`r` ( `= reliabilities.get(engine_name, {}).get('reliablity')` ) is the percentage of requests that have not raised an exception ; but even the requests don't raise exceptions, the checker can see the engine as broken: 
* [supported language](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/search/processors/online.py#L193-L201), [time range](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/search/processors/online.py#L185-L191), [safe search](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/search/processors/online.py#L203-L208) results are [reported in the `/preferences` page](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/templates/oscar/macros.html#L138-L139). It is "strange" to report 100% reliability and at the same time to report some broken features.
* if the `simple` test fails,  `r` is set 0:
https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/webapp.py#L943-L946
* if [paging](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/search/processors/online.py#L174-L183) or [rosebud](https://github.com/searxng/searxng/blob/fe3831db1aa73ebda6eb12fa3a321002c176ccc1/searx/settings.yml#L123-L131) tests fail, there are reports other than:
   * this exclamation mark next to the `r` value.
   * the tooltip on the `r` value.

So if the exclamation mark is displayed with this condition `{%- if reliabilities[engine_name].errors %}`, how to display that the `paging` test fail ?

dalf:
I did not write the context about "alternative idea"

>  In commit 0bd35f7 I implemented "View error logs and submit a bug report" in the toolbox, based on condition reliabilities[<engine.name>].errors.

👍 LGTM.

The alternative idea is about 0b84a515bea503b01cfd601ad09df735cbadf4ff  : 
image

searx/templates/oscar/preferences.html
dalf: What about
```jinja2
{%- if reliabilities[engine_name].errors or reliabilities[engine_name].checker %}{{ icon('exclamation-sign', 'The checker fails on the some tests') }}{% endif %} {{ r -}}</a>
```

?

return42: > What about

LGTM 👍🏻

return42: > of the ARIA recommendation: 

shame on me, did not know about ARIA .. I think I have to read some paper / now I see the meaning of `aria-labelledby`  🤦🏻

> So if the exclamation mark is displayed with this condition {%- if reliabilities[engine_name].errors %}, how to display that the paging test fail ?

If I have understand everything, I think your suggestion in https://github.com/searxng/searxng/pull/40#discussion_r622788834 is the best solution.  May be instead of the text 

    'The checker fails on the some tests'

we should choose a text which is more general / e.g.: _The engine fails for some reason._

dalf: I'm going to open another for this (and the typo `{{engine_name}}_reliablity` --> `{{engine_name}}_relialibility`

return42: > I'm going to open another for this

fine .. feel free to merge this PR, if you think it is OK / for me it is OK.

return42:
Thank you .. especial for updating the initial comment .. I like the "worst case scenario" :-)

dalf:
Idea: when there are a lot of errors, only the first errors can be display, then a button at the bottom of the box can expand the list (similar to the infobox).

return42:
> Idea: when there are a lot of errors

Is this really a (common) use case? .. I don't think so .. If such a lot of engines are listed, the instance is worse maintained and I think it is good for the users to recognize it immediately.

IMO the feature is perfect as is .. (except what we discussed in https://github.com/searxng/searxng/pull/40#pullrequestreview-647258005)

PR #52, opened 2021-05-03
Fix Qwant's fetch_languages function

MarcAbonce:
Cherry-picked from searx/searx/pull/2799

Fixes https://github.com/searxng/searxng/runs/2481629616?check_suite_focus=true


return42 (approved):
Thanks a lot for maintaining this branch!

PR #58, opened 2021-05-05
[mod] multithreading only in searx.search.* packages

dalf:
## What does this PR do?

This PR prepares the new architecture change, everything about multithreading belong to the searx.search.* packages

previously the call to the "init" function of the engines was done in searx.engines:
* the network was not set (the HTTP request sent without using the defined proxy / configuration).
* it requires to monkey patch the code to avoid HTTP requests during the tests

* keep future async and await in searx.search (not in searx.engines).
* respect network configuration even at the initialization step.

* `make run`
* start the checker (embedded and an external tool)

### DIFF
--- a/searx/search/__init__.py
+++ b/searx/search/__init__.py
@@ -29,9 +29,11 @@ from searx.results import ResultContainer
 from searx import logger
 from searx.plugins import plugins
 from searx.search.models import EngineRef, SearchQuery
-from searx.search.processors import processors, initialize as initialize_processors
-from searx.search.checker import initialize as initialize_checker
+from searx.engines import load_engines
+from searx.network import initialize as initialize_network
 from searx.metrics import initialize as initialize_metrics, counter_inc, histogram_observe_time
+from searx.search.processors import PROCESSORS, initialize as initialize_processors
+from searx.search.checker import initialize as initialize_checker
 
 
 logger = logger.getChild('search')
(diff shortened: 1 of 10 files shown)


searx/search/__init__.py
```diff
@@ -29,9 +29,11 @@
 from searx import logger
 from searx.plugins import plugins
 from searx.search.models import EngineRef, SearchQuery
-from searx.search.processors import processors, initialize as initialize_processors
-from searx.search.checker import initialize as initialize_checker
+from searx.engines import load_engines
+from searx.network import initialize as initialize_network
 from searx.metrics import initialize as initialize_metrics, counter_inc, histogram_observe_time
+from searx.search.processors import PROCESSORS, initialize as initialize_processors
```
dalf, line 35: `processors` is renamed `PROCESSORS`, so the `searx.search.processors` module is not hidden.

searx/search/processors/online.py
```diff
@@ -139,15 +148,15 @@ def search(self, query, params, result_container, start_time, timeout_limit):
             self.handle_exception(result_container, e, suspend=True)
             logger.error("engine {0} : HTTP requests timeout"
                          "(search duration : {1} s, timeout: {2} s) : {3}"
-                         .format(self.engine_name, time() - start_time,
+                         .format(self.engine_name, default_timer() - start_time,
```
dalf, line 151: `start_time` is set using `default_timer()` : https://github.com/searxng/searxng/blob/d36adfa59f242a8775ad74245c696d62b7727a36/searx/search/__init__.py#L188

return42 (approved):
Thanks a lot .. I made some tests with builtin checker and checker from command line.  I do not have any issues with.

On top I placed fa0d05c3 with some pylint stuff .. I think it is a good chance to lint touched files, because when a file is touched by a PR this file is also tested.

If you think this is not needed, just drop fa0d05c3.

PR #60, opened 2021-05-06
[mod] Server-Timing header: add render field

dalf:
## What does this PR do?

Add a "render" field to the Server-Timing header
This field is the time spent in the searx.webapp.render function (Jinja2 rendering)

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing

Sometimes there is about 300ms delay between the slowest engine and the total time on my instance:

image

Locally the "Timings" tab will be empty: it is shown [only with HTTPS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing#performanceservertiming_interface).

See the "Headers" tab:
image

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -480,8 +480,11 @@ def render(template_name, override_theme=None, **kwargs):
         for css in plugin.css_dependencies:
             kwargs['styles'].add(css)
 
-    return render_template(
+    start_time = default_timer()
+    result = render_template(
         '{}/{}'.format(kwargs['theme'], template_name), **kwargs)
+    request.render_time += default_timer() - start_time
+    return result
 
 
 def _get_ordered_categories():
@@ -498,6 +501,7 @@ def _get_ordered_categories():
 @app.before_request
 def pre_request():
     request.start_time = default_timer()
+    request.render_time = 0
     request.timings = []
     request.errors = []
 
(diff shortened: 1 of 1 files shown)


return42:
I merged this PR on top of https://darmarit.org/searx/ and made some tests (you can use this link for your own tests).

> Sometimes there is about 300ms delay between the slowest engine and the total time on my instance

In this PR you added the rendering time .. what is your experience, does the delay depends on the rendering time?

I tried oscar & simple theme and I clicked through all categories to see how long it takes to render the different templates.  In my tests (on my server) rendering took from ~25ms up to ~100ms.

I do not have a clue where  the "delay" does come from, rendering seems not the cause.

Most often I can observe long delays in category "files", here is an example from

https://darmarit.org/searx/search?q=maceo%20parker&categories=files&language=en-US

image

I guess the "delay" is caused by something different we do not measure.

OT: does it make sense to make 'Server-Timing' header configurable (somewhere in the `server:` settings)?

dalf:
>  what is your experience, does the delay depends on the rendering time?

The choice of measuring the rendering time is a guess, the difference between the slowest engine and the total time may come from something else (plugins ?). 

>  In my tests (on my server) rendering took from ~25ms up to ~100ms.

Sometimes on my instance it takes more than 200ms.

`strace -e openat,open,stat,read -f -p ...` doesn't show anything special (templates are read once, some stats on .py files nothing more).

> I guess the "delay" is caused by something different we do not measure.

time out are not "displayed" in the Server-Timing header.

And the 4 seconds come from the apk mirror engine timeout.

return42 (approved):
LGTM / Thanks! .. on the long run we should consider to make Server-Timing Header configurable, I think.

PR #66, opened 2021-05-09
review of docker entry point & docker documentation

return42:
**WIP** .. no need to review right now .. 

------

## What does this PR do?

- review of docker-entrypoint.sh
- revision of the docker documentation

additional hints from @dalf:

* about filtron: https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/dockerfiles/uwsgi.ini#L35-L37
* the static files are served by uwsgi
    * uwsgi configuration: https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/dockerfiles/uwsgi.ini#L39-L44 (not that ETag header is not supported)
    * they are compressed when the image is built: https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/Dockerfile#L66-L68
    * also see https://github.com/searx/searx/pull/2242

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -1,62 +1,66 @@
 #!/bin/sh
 
 help() {
-	printf "Command line:\n\n"
-	printf "  -h  Display this help\n"
-	printf "  -d  Dry run to update the configuration files.\n"
-	printf "  -f  Always update on the configuration files (existing files are renamed with the .old suffix)\n"
-	printf "      Without this option, the new configuration files are copied with the .new suffix\n"
-	printf "\nEnvironment variables:\n\n"
-	printf "  INSTANCE_NAME settings.yml : general.instance_name\n"
-	printf "  AUTOCOMPLETE  settings.yml : search.autocomplete\n"
-	printf "  BASE_URL      settings.yml : server.base_url\n"
-	printf "  MORTY_URL     settings.yml : result_proxy.url\n"
-	printf "  MORTY_KEY     settings.yml : result_proxy.key\n"
-	printf "  BIND_ADDRESS  uwsgi bind to the specified TCP socket using HTTP protocol. Default value: \"${DEFAULT_BIND_ADDRESS}\"\n"
-	printf "\nVolume:\n\n"
-	printf "  /etc/searx    the docker entry point copies settings.yml and uwsgi.ini in this directory (see the -f command line option)\n"
-	echo
-	exit 0
+    cat <<EOF
+Command line:
+  -h  Display this help
+  -d  Dry run to update the configuration files.
+  -f  Always update on the configuration files (existing files are renamed with
+      the .old suffix).  Without this option, the new configuration files are
+      copied with the .new suffix
+Environment variables:
+  INSTANCE_NAME settings.yml : general.instance_name
+  AUTOCOMPLETE  settings.yml : search.autocomplete
+  BASE_URL      settings.yml : server.base_url
+  MORTY_URL     settings.yml : result_proxy.url
+  MORTY_KEY     settings.yml : result_proxy.key
+  BIND_ADDRESS  uwsgi bind to the specified TCP socket using HTTP protocol.
+                Default value: ${DEFAULT_BIND_ADDRESS}
+Volume:
+  /etc/searx    the docker entry point copies settings.yml and uwsgi.ini in
+                this directory (see the -f command line option)"
+
+EOF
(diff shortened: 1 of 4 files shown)


dockerfiles/docker-entrypoint.sh
```diff
@@ -96,50 +103,51 @@ update_conf() {
...
 fi
 
-#
 touch /var/run/uwsgi-logrotate
 chown -R searx:searx /var/log/uwsgi /var/run/uwsgi-logrotate
 unset MORTY_KEY
 
 # Start uwsgi
-printf 'Listen on %s\n' "${BIND_ADDRESS}"
+echo 'Listen on %s' "${BIND_ADDRESS}"
```
dalf: it should be `echo 'Listen on' "${BIND_ADDRESS}"` without `%s` ?
or keep `printf` as above (line 124, 127, ....) ?

dockerfiles/docker-entrypoint.sh
```diff
@@ -1,62 +1,67 @@
...
 FORCE_CONF_UPDATE=0
 DRY_RUN=0
+
 while getopts "fdh" option
 do
     case $option in
-	f)
-	    FORCE_CONF_UPDATE=1
+
+	f) FORCE_CONF_UPDATE=1 ;;
```
dalf: Is it possible to use only space and not tab in the whole file ?

dockerfiles/docker-entrypoint.sh
```diff
@@ -1,62 +1,67 @@
...
-# display the help message without the version
-if [ "$1" = "help" ]; then
-	help
-fi
-
-# Version
-export SEARX_VERSION=$(su searx -c 'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' 2>/dev/null)
-printf 'searx version %s\n\n' "${SEARX_VERSION}"
+BIND_ADDRESS="${BIND_ADDRESS:-${DEFAULT_BIND_ADDRESS}}"
+export BIND_ADDRESS
```
dalf: For my own knowledge why it can't be `export BIND_ADDRESS="${BIND_ADDRESS:-${DEFAULT_BIND_ADDRESS}}"` in one line ?

dockerfiles/docker-entrypoint.sh
```diff
@@ -1,62 +1,67 @@
...
 done
 
+get_searx_version(){
+    su searx -c \
+       'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' \
+       2>/dev/null
+}
+
+SEARX_VERSION="$(get_searx_version)"
+export SEARX_VERSION
```
dalf, line 58: Why not `export SEARX_VERSION="$(get_searx_version)"` ?

dockerfiles/docker-entrypoint.sh
return42: > or keep printf as above (line 124, 127, ....) ?

I tend to used echo because I think it is more readable, but this is my very personal opinion. More experienced users say it is better to use printf ([Basically, it's a portability (and reliability) issue](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819)).

on second thought, I would say it is better to stay with printf.

dockerfiles/docker-entrypoint.sh
return42: > Is it possible to use only space and not tab in the whole file ?

Yes, I forgot to _untabify_.

dockerfiles/docker-entrypoint.sh
return42: It can, my fail.

dockerfiles/docker-entrypoint.sh
return42: The return value of the export will always _true_ and the return value from the `get_searx_version` will be ignored, see  [SC2155: Declare and assign separately to avoid masking return values.](https://github.com/koalaman/shellcheck/wiki/SC2155)

return42:
I merged this PR in the [master branch of my fork](https://github.com/return42/searxng/commits/master).  The modified documentation is online available at: https://return42.github.io/searxng/admin/installation-docker.html

dalf:
I've tried the different options, all seems to work.

return42:
> I've tried the different options, all seems to work.

Thanks for review, merge and all the hints you gave! .. for me it was a good exercise in using docker :)

PR #74, opened 2021-05-14
[fix] webapp.py: pylint: disable=assigning-non-slot for request.*

dalf:
## What does this PR do?

Fix tests

Fix `make test`
Note: actually, the code should use flask.g.* instead of request.*

## Note

Why does the issue pop now ???

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -483,7 +483,8 @@ def render(template_name, override_theme=None, **kwargs):
     start_time = default_timer()
     result = render_template(
         '{}/{}'.format(kwargs['theme'], template_name), **kwargs)
-    request.render_time += default_timer() - start_time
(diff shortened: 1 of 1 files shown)


return42:
> Why does the issue pop now ???

werkzeug has been [upgraded to 2.0.0 these days](https://pypi.org/project/Werkzeug/#history) and we do not pin sub-requirements. We have pinned ..

https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/requirements.txt#L4

and in this version, Flask requires ` "Werkzeug>=0.15"`

https://github.com/pallets/flask/blob/93dd1709d05a1cf0e886df6223377bdab3b077fb/setup.py#L55-L56

return42 (approved):

PR #80, opened 2021-05-15
[enh] add offline engine for sqlite database

return42:
To test & demonstrate this implementation download:

  https://liste.mediathekview.de/filmliste-v2.db.bz2

and unpack into searx/data/filmliste-v2.db, in your settings.yml define a sqlite engine named "demo"::

    - name : demo
      engine : sqlite
      shortcut: demo
      categories: general
      result_template: default.html
      database : searx/data/filmliste-v2.db
      query_str :  >-
        SELECT title || ' (' || time(duration, 'unixepoch') || ')' AS title,
               COALESCE( NULLIF(url_video_hd,''), NULLIF(url_video_sd,''), url_video) AS url,
               description AS content
          FROM film
         WHERE title LIKE :wildcard OR description LIKE :wildcard
         ORDER BY duration DESC
      disabled : False

Query to test: `!demo concert`

This is a rewrite of the implementation from commit [1]

[1] https://github.com/searx/searx/commit/8e90a214ce5cf60e0640e2ba812fa432d7bdef2f

Suggested-by: @virtadpt https://github.com/searx/searx/issues/2808

### DIFF
--- a/searx/engines/sqlite.py
+++ b/searx/engines/sqlite.py
@@ -1,0 +1,78 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=missing-function-docstring
+
+"""SQLite database (Offline)
+
+"""
+
+import sqlite3
+
+from searx import logger
+
+logger = logger.getChild('SQLite engine')
+
+engine_type = 'offline'
+database = ""
+query_str = ""
+limit = 10
+paging = True
+result_template = 'key-value.html'
+
+class SQLiteDB:
+    """
+    Implements a `Context Manager`_ for a SQLite.
+
+    usage::
+
+        with SQLiteDB('test.db') as cur:
+            print(cur.execute('select sqlite_version();').fetchall()[0][0])
+
+    .. _Context Manager: https://docs.python.org/3/library/stdtypes.html#context-manager-types
+    """
+
+    def __init__(self, db):
+        self.database = db
+        self.connect = None
+
+    def __enter__(self):
+        self.connect = sqlite3.connect(self.database)
+        self.connect.row_factory = sqlite3.Row
+        return self.connect.cursor()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is None:
+            self.connect.commit()
+        self.connect.close()
+
+def init(engine_settings):
+    if 'query_str' not in engine_settings:
+        raise ValueError('query_str cannot be empty')
+
+    if not engine_settings['query_str'].lower().startswith('select '):
+        raise ValueError('only SELECT query is supported')
+
+def search(query, params):
+    global database, query_str, result_template  # pylint: disable=global-statement
+    results = []
+
+    query_params = {
+        'query': query,
+        'wildcard':  r'%' + query.replace(' ',r'%') + r'%',
+        'limit': limit,
+        'offset': (params['pageno'] - 1) * limit
+    }
+    query_to_run = query_str + ' LIMIT :limit OFFSET :offset'
+
+    with SQLiteDB(database) as cur:
+
+        cur.execute(query_to_run, query_params)
+        col_names = [cn[0] for cn in cur.description]
+
+        for row in cur.fetchall():
+            item = dict( zip(col_names, map(str, row)) )
+            item['template'] = result_template
+            logger.debug("append result --> %s", item)
+            results.append(item)
+
+    return results
--- a/searx/engines/sqlite.py
+++ b/searx/engines/sqlite.py
@@ -6,9 +6,8 @@
 
 """
 
-import sqlite3
-
 from searx import logger
+from searx.utils import SQLiteCursor
 
 logger = logger.getChild('SQLite engine')
 
@@ -18,32 +17,6 @@ query_str = ""
 limit = 10
 paging = True
 result_template = 'key-value.html'
-
-class SQLiteDB:
-    """
-    Implements a `Context Manager`_ for a SQLite.
-
-    usage::
-
-        with SQLiteDB('test.db') as cur:
-            print(cur.execute('select sqlite_version();').fetchall()[0][0])
-
-    .. _Context Manager: https://docs.python.org/3/library/stdtypes.html#context-manager-types
-    """
-
-    def __init__(self, db):
-        self.database = db
-        self.connect = None
-
-    def __enter__(self):
-        self.connect = sqlite3.connect(self.database)
-        self.connect.row_factory = sqlite3.Row
-        return self.connect.cursor()
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        if exc_type is None:
-            self.connect.commit()
-        self.connect.close()
 
 def init(engine_settings):
     if 'query_str' not in engine_settings:
@@ -64,7 +37,7 @@ def search(query, params):
     }
     query_to_run = query_str + ' LIMIT :limit OFFSET :offset'
 
-    with SQLiteDB(database) as cur:
+    with SQLiteCursor(database) as cur:
 
         cur.execute(query_to_run, query_params)
         col_names = [cn[0] for cn in cur.description]
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1003,6 +1003,26 @@ engines:
     timeout : 3.0
     disabled : True
 
+  # For this demo of the sqlite engine download:
+  #   https://liste.mediathekview.de/filmliste-v2.db.bz2
+  # and unpack into searx/data/filmliste-v2.db
+  # Query to test: "!demo concert"
+  #
+  # - name : demo
+  #   engine : sqlite
+  #   shortcut: demo
+  #   categories: general
+  #   result_template: default.html
+  #   database : searx/data/filmliste-v2.db
+  #   query_str :  >-
+  #     SELECT title || ' (' || time(duration, 'unixepoch') || ')' AS title,
+  #            COALESCE( NULLIF(url_video_hd,''), NULLIF(url_video_sd,''), url_video) AS url,
+  #            description AS content
+  #       FROM film
+  #      WHERE title LIKE :wildcard OR description LIKE :wildcard
+  #      ORDER BY duration DESC
+  #   disabled : False
+
   - name : torrentz
     engine : torrentz
     shortcut : tor
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -9,6 +9,7 @@
 import sys
 import re
 import importlib
+import sqlite3
 
 from numbers import Number
 from os.path import splitext, join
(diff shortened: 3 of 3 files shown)


searx/engines/sqlite.py
```diff
@@ -0,0 +1,49 @@
...
+def search(query, params):
+    query_params = {'query': query}
+    query_to_run = query_str + ' LIMIT {0} OFFSET {1}'.format(limit, (params['pageno'] - 1) * limit)
+
+    connection = sqlite3.connect(database)
+    cur = connection.cursor()
+    cur.execute(query_to_run, query_params)
+    results = _fetch_results(cur)
+    cur.close()
+    connection.close()
```
dalf: To avoid memory / connection / cursor leak (??) :
```python
connection = sqlite3.connect(database)
try:
  cur = connection.cursor()
  try:
    cur.execute(query_to_run, query_params)
    return _fetch_results(cur)
  finally:
    cur.close()
finally:
  connection.close()
```

(not tested)

searx/engines/sqlite.py
```diff
@@ -0,0 +1,49 @@
...
+def init(engine_settings):
+    if 'query_str' not in engine_settings:
+        raise ValueError('query_str cannot be empty')
+
+    if not engine_settings['query_str'].lower().startswith('select '):
+        raise ValueError('only SELECT query is supported')
+
+def search(query, params):
+    query_params = {'query': query}
+    query_to_run = query_str + ' LIMIT {0} OFFSET {1}'.format(limit, (params['pageno'] - 1) * limit)
```
dalf: To avoid any `.format` (??):
```python
query_params = {'query': query, 'limit': limit, 'offset': (params['pageno'] - 1) * limit}
query_to_run = query_str + ' LIMIT :limit OFFSET :offset'
```

(not tested)

searx/engines/sqlite.py
return42: > To avoid memory / connection / cursor leak (??) :

I will implement a [context manager](https://docs.python.org/3/library/stdtypes.html#context-manager-types).

searx/engines/sqlite.py
return42: > To avoid any .format (??)

LGTM, thanks! / BTW: using _execute parameters_ `cur.execute(..., query_params)` is more safe from SQL injection!

searx/engines/sqlite.py
dalf: I haven't seen a context manager for sqlite except for SQL transaction:
https://docs.python.org/3/library/sqlite3.html#using-the-connection-as-a-context-manager

return42: > I haven't seen a context manager for sqlite

I have already implemented one .. wait until I have rebased this PR ...

return42:
@dalf I rewrite the entire engine and added an example.

I guess it is not related to this PR, but I get an `AttributeError` from threading for offline engines (should I add an issue for?):

```
  File "searx/search/processors/offline.py", line 24, in search
    self.extend_container(result_container, start_time, search_results)
  File "searx/search/processors/abstract.py", line 126, in extend_container
    self._extend_container_basic(result_container, start_time, search_results)
  File "searx/search/processors/abstract.py", line 111, in _extend_container_basic
    page_load_time = get_time_for_thread()
  File "searx/network/__init__.py", line 47, in get_time_for_thread
    return THREADLOCAL.total_time
AttributeError: '_thread._local' object has no attribute 'total_time'
```

update: has been fixed in #85

searx/engines/sqlite.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logger.getChild('SQLite engine')
+
+engine_type = 'offline'
+database = ""
+query_str = ""
+limit = 10
+paging = True
+result_template = 'key-value.html'
+
+class SQLiteDB:
```
dalf: IMO the function version is more readable:
```python
from contextlib import contextmanager

@contextmanager
def sqlite_cursor(database_url):
	try:
		connect = sqlite3.connect(database_url)
		connect.row_factory = sqlite3.Row
		cursor = self.connect.cursor()
		try:
			yield cursor
		finally:
			cursor.close()
	finally:
		if connect:
			connect.close()
```

searx/engines/sqlite.py
```diff
@@ -0,0 +1,78 @@
...
+        self.connect = None
+
+    def __enter__(self):
+        self.connect = sqlite3.connect(self.database)
+        self.connect.row_factory = sqlite3.Row
+        return self.connect.cursor()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is None:
+            self.connect.commit()
```
dalf: There is no need to commit for a SELECT : https://docs.python.org/3/library/sqlite3.html#sqlite3-controlling-transactions

searx/engines/sqlite.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(self, db):
+        self.database = db
+        self.connect = None
+
+    def __enter__(self):
+        self.connect = sqlite3.connect(self.database)
+        self.connect.row_factory = sqlite3.Row
+        return self.connect.cursor()
```
dalf: The cursor is not closed.

searx/engines/sqlite.py
return42: Not in this use case, but in more general use cases a cursor can execute update statements. I placed the class in the utils section and renamed it:

https://github.com/searxng/searxng/pull/80/files#diff-8d38f1d52f68ec93ec84ebb815fbbb6240fd70d8076dc11ee55021ad58c5658eR617

searx/engines/sqlite.py
return42: Normally if the db connection is closed the cursor lands in garbage collector without any cost.  But you are right; explicit closing is better. 

https://github.com/searxng/searxng/pull/80/files#diff-8d38f1d52f68ec93ec84ebb815fbbb6240fd70d8076dc11ee55021ad58c5658eR647

searx/engines/sqlite.py
return42: Hm if you insist we can use a decorator .. I'm more used to use classes.  And I'm not a  fan boy "exceptional programming" .. try blocks are expensive and often miss used I think. Compare, less indentations no try block needed:

```python
class SQLiteCursor:
    def __init__(self, db):
        self.database = db
        self.connect = None
        self.cursor = None
    def __enter__(self):
        self.connect = sqlite3.connect(self.database)
        self.connect.row_factory = sqlite3.Row
        self.cursor = self.connect.cursor()
        return self.cursor
    def __exit__(self, exc_type, exc_val, exc_tb):
        if exc_type is None:
            self.connect.commit()
        self.cursor.close()
        self.connect.close()
```

return42:
@dalf I rebased and placed two more commits on top .. before we finally merge, I will clean up the history.

searx/engines/sqlite.py
dalf: Scenario:
* some queries crash (whatever the reason)
* cursor and connect are not closed right away but with GC.

The try...finally blocks are here to make the connection whatever happens.

If the cursor is closed with the database it can be:
```python
@contextmanager
def sqlite_cursor(database_url):
	try:
		connect = sqlite3.connect(database_url)
		connect.row_factory = sqlite3.Row
		yield self.connect.cursor()
	finally:
		if connect:
			connect.close()
```

But everything can be fits into the `search` function.

virtadpt: > ```python
>     def __exit__(self, exc_type, exc_val, exc_tb):
>         if exc_type is None:
>             self.connect.commit()
>         self.cursor.close()
>         self.connect.close()
> ```

Searches are read-only operations.  I don't think a SQL COMMIT is required here.

return42: > Searches are read-only operations. I don't think a SQL COMMIT is required here.

see my https://github.com/searxng/searxng/pull/80#discussion_r636230676

searx/engines/sqlite.py
virtadpt: Searx is a (meta) search engine.  Is there a use case for using it to do anything other than query databases?

searx/engines/sqlite.py
return42: > Scenario:
>    some queries crash (whatever the reason)

May I do not understand all .. AFAIK for this we use the "with" statement.  Can't speak for the decorator solution, but the `__exit__` method of the class is called,  no matter what was the reason for leaving the "with"" block.

> But everything can be fits into the search function.

Hm .. I do not see the benefit or I misunderstood you.

To get things done: do you like to patch on top what ever solution you prefer .. I can reorder the history afterwards.

searx/engines/sqlite.py
return42: Not at the moment but may be we wish to store some data one day.  This is just a class (context manager) generalized to use fore multi usage, placed in module `utils.py`

return42:
Sorry for the rebase mess, I pushed my local branch to the wrong remote branch / I screwed this PR back to a24b10c

return42:
@dalf if you do not have time right now to [improve the cursor context](https://github.com/searxng/searxng/pull/80#discussion_r636327695) I would to suggest to merge this PR / nothing needs to be perfect from its first commit.  

If you do not have any considerable doubt, lets merge .. we can improve it later.

dalf:
without e0d64a0e2c9bbf2a1530ad9516c4a77e8e26d947 and 10ea2eef3affd6e801db070c62430a5afabe1053 and using [contextlib](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager):

```python
# lint: pylint
# pylint: disable=missing-function-docstring

"""SQLite database (Offline)

"""

import sqlite3
import contextlib

from searx import logger


logger = logger.getChild('SQLite engine')

engine_type = 'offline'
database = ""
query_str = ""
limit = 10
paging = True
result_template = 'key-value.html'


def init(engine_settings):
    if 'query_str' not in engine_settings:
        raise ValueError('query_str cannot be empty')

    if not engine_settings['query_str'].lower().startswith('select '):
        raise ValueError('only SELECT query is supported')


@contextlib.contextmanager
def sqlite_cursor():
    """
    see:
    * https://docs.python.org/3/library/sqlite3.html#sqlite3.connect
    * https://www.sqlite.org/uri.html
    * https://docs.python.org/3/library/contextlib.html#contextlib.closing

    read-only so there is an error instead of empty file on the file system,
    and then an error on the query execution.
    """
    global database  # pylint: disable=global-statement
    uri = 'file:' + database + '?mode=ro'
    with contextlib.closing(sqlite3.connect(uri, uri=True)) as connect:
        connect.row_factory = sqlite3.Row
        with contextlib.closing(connect.cursor()) as cursor:
            yield cursor


def search(query, params):
    global query_str, result_template  # pylint: disable=global-statement
    results = []

    query_params = {
        'query': query,
        'wildcard':  r'%' + query.replace(' ',r'%') + r'%',
        'limit': limit,
        'offset': (params['pageno'] - 1) * limit
    }
    query_to_run = query_str + ' LIMIT :limit OFFSET :offset'

    with sqlite_cursor() as cur:

        cur.execute(query_to_run, query_params)
        col_names = [cn[0] for cn in cur.description]

        for row in cur.fetchall():
            item = dict( zip(col_names, map(str, row)) )
            item['template'] = result_template
            logger.debug("append result --> %s", item)
            results.append(item)

    return results
```

---

The cursor is explicitly closed, if we rely on connection to do that, the code is even more simple / straightforward :

```python
@contextlib.contextmanager
def sqlite_cursor():
    """
    see:
    * https://docs.python.org/3/library/sqlite3.html#sqlite3.connect
    * https://www.sqlite.org/uri.html
    * https://docs.python.org/3/library/contextlib.html#contextlib.closing

    read-only so there is an error instead of empty file on the file system.
    """
    uri = 'file:' + database + '?mode=ro'
    with contextlib.closing(sqlite3.connect(uri, uri=True)) as connect:
        connect.row_factory = sqlite3.Row
        yield contextlib.closing(connect.cursor()) 
```

see [contextlib.closing](https://github.com/python/cpython/blob/bee66d3cb98e740f9d8057eb7f503122052ca5d8/Lib/contextlib.py#L281) : if there is an exception, the connection is closed.

For now, there is no need for more. If we need sqlite access somewhere else (searx.data.* ?), we can eventually move `sqlite_cursor` to `searx.utils`, but once we are sure to use it in another location than `searx.engines.sqlite`.

---

https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager can be in another PR.

PR #84, opened 2021-05-18
[fix] unit test: don't load /etc/searx/settings.yml

### LINKED ISSUE #79: unit test reads settings from /etc/searx
The unit test should not read settings from a `settings.yml` file located in `/etc/searx`

dalf:
## What does this PR do?

Add a new environment variable SEARX_DISABLE_ETC_SETTINGS
to disable loading of /etc/searx/settings.yml

unit tests:
* set SEARX_DISABLE_ETC_SETTINGS to 1
* remove SEARX_SETTINGS_PATH if it exists

Reproducible tests.

```sh
source ./local/py3/bin/activate
strace -f -e trace=file python -m nose2 -s tests/unit 2>&1 | grep "/etc/searx"
```

`/etc/searx/settings.yml`  should not appeared.

## Related issues

Closes: #79

### DIFF
--- a/searx/settings_loader.py
+++ b/searx/settings_loader.py
@@ -40,6 +40,9 @@ def get_user_settings_path():
         # enviroment variable SEARX_SETTINGS_PATH
         return check_settings_yml(environ['SEARX_SETTINGS_PATH'])
 
+    if environ.get('SEARX_DISABLE_ETC_SETTINGS', '').lower() in ('1', 'true'):
+        return None
+
     # if not, get it from searx code base or last solution from /etc/searx
     return check_settings_yml('/etc/searx/settings.yml')
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,2 +1,5 @@
 import os
+
 os.environ['SEARX_DEBUG'] = '1'
+os.environ['SEARX_DISABLE_ETC_SETTINGS'] = '1'
+os.environ.pop('SEARX_SETTINGS_PATH', None)


return42:
OT here, but the main problem I have with the settings is: Settings are loaded by a side effect of "import searx".

https://github.com/searxng/searxng/blob/d255e5637b45a05d7a3b01d261affdcfc120b653/searx/__init__.py#L28

This makes it hard to control (and understand) **when** the setting file is loaded and we need environment variables to control which file is the correct settings file at import time.  Normally I expect that the application loads its settings when the application is started / I mean: load settings is not a task of an import statement, it is a task of the initialization procedure of the application.

As I said often at the searx project: IMO it is a design failure when an import statement load files like settings and does the main application setup ... but here in this PR, it might be OT.

ATM I'm in a review of this PR .. Thereby I noticed the loading of the setting file at import time in a negative way and I just wanted to leave the notice above, without having a solution at hand. But I would say it is worth to have this in mind, may be we can change this in the future.

dalf:
I'm aware settings.yml load at import time issue, but when I read:

> The unit test should not read settings from a settings.yml file located in /etc/searx

I thought in addition / in consequence the tests can be broken when there is a `/etc/settings.yml`, so I straced `make test`

return42:
>  so I straced `make test`

Yes this is one of the drawbacks which shows me that we have a design issue.

I think for the #79 a solution like this draft is perfect .. (give me some more time to review/think about) .. on the long term we should consider to fix the architecture issue.  Not now, but when we do some design decisions we should also think about this settings loaded while importing a module (in that case _entering a python package_).

Another drawback of loading settings while import is, that test tools can't catch (mask) exception thrown while import searx to prepare the test ..

https://github.com/searxng/searxng/blob/94e6ed6110d140d4d1006e286a05d178090d9d78/tests/unit/test_answerers.py#L5-L6

If the settings file from SEARX_SETTINGS_PATH  is not readable, an IOError is raised before the test is started and the message from unit test is a bit missleading ...

```
======================================================================
ERROR: tests.unit.test_answerers (nose2.loader.ModuleImportFailure)
ImportError: Failed to import test module: tests.unit.test_answerers
  File "searx/local/py3/lib/python3.8/site-packages/nose2/plugins/loader/discovery.py", line 201, in _find_tests_in_file
    module = util.module_from_name(module_name)
  File "searx/local/py3/lib/python3.8/site-packages/nose2/util.py", line 77, in module_from_name
    __import__(name)
  File "share/searx/tests/unit/test_answerers.py", line 5, in <module>
    from searx.answerers import answerers
  File "searx/searx/__init__.py", line 28, in <module>
    settings, settings_load_message = searx.settings_loader.load_settings()
AttributeError: partially initialized module 'searx' has no attribute 'settings_loader' (most likely due to a circular import)
```

return42:
Patch LGTM .. fixes #79 .. lets merge.

PR #86, opened 2021-05-19
[mod] remove gc.collect() after each user request

dalf:
## What does this PR do?

When searx runs using the docker image, the render time is often high (about 200ms).
When the explicit call to the garbage collector is removed, the render time is often bellow 50ms.

I suspect these causes:
* httpx and its dependencies creates a lot of small objects which change what is kept in the memory or not after a full garbage collector.
* musl from Alpine, since the memory is managed in a different way.

200ms to render a Jinja2 template is high.

Run the docker image and check the render time (in the Server-Timing header).

Alternative: switch or at least check with a debian docker image ( [whoogle uses](https://github.com/benbusby/whoogle-search/blob/2bd17bef47af11569d2f524d26f3e14dd6308374/Dockerfile#L1) 3.8-slim )

## Related issues
gc.collect was added in commit ffbee4bb8273e
Timings see:https://github.com/searxng/searxng/pull/60

### DIFF
--- a/searx/search/__init__.py
+++ b/searx/search/__init__.py
@@ -1,19 +1,6 @@
-'''
-searx is free software: you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
-
-searx is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Affero General Public License for more details.
-
-You should have received a copy of the GNU Affero General Public License
-along with searx. If not, see < http://www.gnu.org/licenses/ >.
-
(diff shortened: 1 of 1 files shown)


return42:
FYI: I rebased your branch remove-call-to-gc on top of branch master.

BTW: I removed::

    from _thread import start_new_thread

I placed pylint patch (6f1446d5) on top, drop it if you have a doubt about.

return42:
> Alternative: switch or at least check with a debian docker image ( [whoogle uses](https://github.com/benbusby/whoogle-search/blob/2bd17bef47af11569d2f524d26f3e14dd6308374/Dockerfile#L1) 3.8-slim )

I made some local tests (not in docker) .. I observed rendering times from 10-60ms with gc and 10-15ms without gc. Commit ffbee4b talks about "reduce memory footprint" ..  so I guess merging this PR is a "performance vs memory"  decision which is independent of the docker image question.

I would vote for "performance"  .. IMO lets merge this PR.

Issue #69, opened 2021-05-11
engine: openstreetmap.py


dalf:
https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/searx/engines/openstreetmap.py#L25-L26
could be
`https://nominatim.openstreetmap.org/search.php?{query}&polygon_geojson=1&format=jsonv2&addressdetails=1&extratags=1&dedupe=1`

see : https://nominatim.org/release-docs/latest/api/Search/

Note:
* It seems [`class`](https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/searx/engines/openstreetmap.py#L76-L79) has been replaced by `category`.
* I don't see `address29` in the results
* each result contains an `extratags` entry which provides:
  * `wikidata link`
  * `wikipedia link`
  * `opening_hours`

-->

* It is possible to remove [call to https://overpass-api.de from the browser](https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/searx/static/themes/oscar/src/js/leaflet_map.js#L19-L100) ([also](https://github.com/searxng/searxng/blob/9a383d15334eb58e273f394f315cf46398e597bb/searx/templates/oscar/result_templates/map.html#L42-L49) )
* Since nominatim is fast it is possible to send an additional [query to get the images from wikidata](https://w.wiki/3Jxs)

-->

In less than 800ms per request, usually faster, the results can contain:
* the wikipedia & wikidata links
* opening hours (see https://wiki.openstreetmap.org/wiki/Key:opening_hours#Python )
* a thumbnail

PR #90, opened 2021-05-21
[enh] improve openstreetmap engine

dalf:
## What does this PR do?

implements ideas described in https://github.com/searxng/searxng/issues/69

* the master branch allow to make direct HTTP request from the browser to overpass API : it breaks the privacy idea.
* include wikidata data: an image, website & wikipedia links.
* the overpass data are now fetched directly using the nominatim API.

This PR is a draft / a demo  : 
* [x] there are some `style=...` in `map.html`
* [x] the simple theme is not updated. 
* [x] localization is not implemented.
* [x] the HTML layout can be improved.
* [x] the javascript to call the overpass API is not removed.
* [x] `searx/data/osm_tags.json` should be named  `searx/data/osm.json` and can be smaller.

See https://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases

Some example:
* `!osm museum in berlin`
* `!osm ATMs in bruxelles`
* `!osm paris`
* `!osm cinémas in lyon`


searx/engines/wikidata.py
```diff
@@ -64,6 +65,7 @@
         mwapi:language "%LANGUAGE%".
         ?item wikibase:apiOutputItem mwapi:item.
   }
+  MINUS { ?item wdt:P31 wd:Q21281405 . }
```
dalf: Not related to this PR....
(see line 55)

return42:
Wow .. very impressive POC .. thanks!

For me it is hard to recognize all the interrelationships. I think it would be good to have some comments or doc-strings in the final version that describe the relations.

searx/engines/openstreetmap.py
```diff
@@ -23,24 +29,176 @@
...
 
-    params['url'] = base_url + search_string.format(query=query)
+    params['url'] = base_url + search_string.format(query=urlencode({'q': query}))
     params['route'] = route_re.match(query)
+    params['headers']['User-Agent'] = searx_useragent()
 
     return params
 
 
+def fetch_wikidata(nominatim_json, user_langage):
```
return42, line 203: I think functions like this needs to be documented .. for a newbee it will take hours and hours to understand whats going on here .. most often I gave up after 15min. 

The criticism is not related to this PR it is more general, I often stuck because no function in searx is documented. I don't think every little thing needs to be documented, but things that are not obvious from the implementation should be documented. 

What I mean: in the future we should more document .. I am also addressing myself :)

searx/engines/openstreetmap.py
```diff
@@ -23,24 +29,176 @@
...
+                r['wikidata'] = {
+                    'image': get_external_url('wikimedia_image', result['image']['value']),
+                    'image_sign': get_external_url('wikimedia_image', result.get('sign', {}).get('value')),
+                    'image_symbol': get_external_url('wikimedia_image', result.get('symbol', {}).get('value')),
+                }
+                r['wikipedia_label'] = result.get('wikipediaLabel', {}).get('value')\
+                    or result.get('itemLabel', {}).get('value')
+
+
+def set_value(d, k, o, *args):
```
return42: If I am not wrong, this function is not used.

dalf:
I know the step is huge about wikidata, and even more with the interconnection with OSM, wikipedia etc...

There are a lot of trial and error in the current code because data are never perfectly normalized.
For example lines from 270 to 276

https://github.com/searxng/searxng/blob/9c026cfea9e37c20eefa550e0b6248f13a3fe4f0/searx/engines/openstreetmap.py#L270-L276

the code tries to get an image from:

* wikidata
* ['extratags']['image']
* ['extratags']['wikimedia_commons']

I haven't found documentation about the last two, perhaps there is, I just discovered them making queries.

### OSM

  * If you look for a node detail (for example [Berlin](https://www.openstreetmap.org/node/240109189)), you see that there is wikidata and wikipedia links (at the bottom). Of course it is not always defined.

### Nominatim

  * it is the search engine used by OSM
  * the API: https://nominatim.org/release-docs/latest/api/Search/
  * out of scope of this PR: It can be installed locally: https://github.com/osm-search/Nominatim
  * there are some special words: https://wiki.openstreetmap.org/wiki/Nominatim/Special_Phrases

### overpass
  * see https://overpass-turbo.eu/ (not exactly the same URL than in searx(ng) code, but basically it allows to search in OSM like in a database).
  * it is just a way to query OSM.
  * currently use in the master branch with direct call 
     * see https://github.com/searxng/searxng/blob/2398e9a1fea6c5c8c2822766d5b0ba38031aec2e/searx/static/themes/oscar/src/js/leaflet_map.js#L20
     * it provides different information including the wikidata identifier ( For example Q2 for earth : https://www.wikidata.org/wiki/Q2 ).
  * the interesting data fetched from overpass can now be fetched directly from nominatim with `extratags=1` , that how openstreetmap.py gets all the wikidata items from line 131 to line 138.

  https://github.com/searxng/searxng/blob/9c026cfea9e37c20eefa550e0b6248f13a3fe4f0/searx/engines/openstreetmap.py#L131-L138
   
  So after line 138, `wikidata_ids` contains a list of wikidata id. Only one request will be sent to wikidata to fetch the images from all these ids. The query defined by `wikidata_image_sparql `.


### wikidata

  * each wikipedia has a wikidata page (on the left side, there is a link under the "Tool" section)
  * some (most of ) wikidata pages don't have a wikipedia page
  * wikidata stores tripple, the most common use case is (object, property, value).
  * It is possible to query wikidata using SPARQL. 
      * See https://query.wikidata.org (especially the example queries, the help and the tools).
      * See https://www.mediawiki.org/wiki/Wikidata_Query_Service/User_Manual
  * For example : 
     * the get an image from the Earth, https://www.wikidata.org/wiki/Q2 : you can click on "Image" and see it goes to https://www.wikidata.org/wiki/Property:P18
     * a query to get the same information: https://w.wiki/3NCG
     * it is possible to retrieve also more an image and the population: https://w.wiki/3NCH
        * you can see that the query return only one result for the population despite the fact that https://www.wikidata.org/wiki/Q2 display a list. It is possible to retrieve the whole list with a more complicated query. 
     * the same query without copy / paste of `Q2`: https://w.wiki/3NCg
     * with `?item` in the `select` : https://w.wiki/3NCN
     * but what the label for `?item` would nice: https://w.wiki/3NCn
     * now if we remove `values ?item { wd:Q2 }` : https://w.wiki/3NCU
     * a different thing: get the English wikipedia page for a wikidata item: https://w.wiki/3NCo
  * So you can get the query `wikidata_image_sparql` and try it in https://query.wikidata.org (just replace %QUERY% by one or more wikidata ids)

searx/engines/openstreetmap.py
dalf: * Lines 131 to 138: 
  * set `wikidata_id` to the wikidata ids of `['extratags']['wikidata']` or `['extratags']['wikidata link']` (I have seen both...)
  * set `wd_to_results` to mapping from a wikidata id to a list of result (the actual result in `nominatim_json`).
* Lines 141 to 143: send the query using `wikidata_id`.
* Line 144: for each result of the SPARQL request (more or less for each wikidata id (*) ).
* line 146: update the results of `nominatim_json` using `wd_to_results` = for each result of `nominatim_json` which reference `wikidata_id`, set the `wikidata` entry.

(*) the result may contain more than one row per wikidata id (same issue than `SELECT * FROM a INNER JOIN b ON b.parent_id = a.id`, if there are two images, there are two rows for the same item). It is silently ignored in the code.

return42:
>   * So you can get the query `wikidata_image_sparql` and try it in https://query.wikidata.org (just replace %QUERY% by one or more wikidata ids)

Does it make sense to assemble queries like:

https://github.com/searxng/searxng/blob/4fc347a7dbbd92b37d5da50f15e387c364d06459/searx/engines/openstreetmap.py#L38-L56

in one module? .. e.g. in the https://github.com/searx/searx/blob/master/searx/engines/wikidata.py ore in a new module searx.wikidata

dalf:
> Does it make sense to assemble queries like in one module?

The HTTP requests are spread in different locations using a HTTP client.
I see SPARQL request in the same way: the SPARQL requests are sent to wikidata.org but for different purposes / intents: an additional module will create an cognitive redirection from the call to the query, and I don't know if there is a benefit to put all the queries in one place? (more than `git grep send_wikidata_query`). (??)

return42:
https://github.com/searxng/searxng/pull/113 has been merged, do you like to rebase this draft?

dalf:
> #113 has been merged, do you like to rebase this draft?

work in progress, it is more than a rebase

searx/static/themes/oscar/js/searx.min.js
```diff
@@ -1,4 +1,4 @@
-/*! oscar/searx.min.js | 01-06-2021 | https://github.com/searxng/searxng  */
+/*! oscar/searx.min.js | 02-06-2021 | https://github.com/searxng/searxng  */
```
return42: bikeshedding: this patch can be moved to the commit which includes the build products [static].

dalf: Done.

searx/data/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 
 __init__ = ['ENGINES_LANGUGAGES', 'CURRENCIES', 'USER_AGENTS', 'EXTERNAL_URLS', 'WIKIDATA_UNITS', 'EXTERNAL_BANGS',
-            'bangs_loader', 'ahmia_blacklist_loader']
+            'OSM_KEYS_TAGS', 'bangs_loader', 'ahmia_blacklist_loader']
```
return42: @dalf do you know where this `__init__` variable is used?  I have problems to find the usage.

dalf: It should `__all__ ` not `__init__` !

dalf: updated

return42:
@dalf: is it OK when I rebase this branch?

return42:
@dalf sorry for my rebase mess / I had to move the build of the static & data files to the end of the patch series.  As you can see in the commit log, I fixed some minor issues .. hope this is OK for you, otherwise ask me to screw back.

return42 (approved):

dalf:
Why rebuild / refetch the data ?

dalf:
Thank you for the fixes and updates.

One note: 
* the commit c75425655fdadf9554b97ae0309a6181acd34ce3 already contains searx/data/osm_keys_tags.json
* so I wouldn't call `make data.all` in d42a663

return42:
> so I wouldn't call make data.all in d42a663

I don't know if it was a good idea, but I thought why not rebuild all static files in one commit, even if some of the files are not a part of this PR and the [searx/data/osm_keys_tags.json](https://github.com/searxng/searxng/pull/90/commits/d42a6633bbbf64170221e6db60b7c1616cfeda6a#diff-4c08d39e572ad20eb6393e43df96e85bde66678073082c3bf2e4ab8f02b86678) was also updated.

If you want, I can reduce d42a663 to themes build (,/searx/static).

dalf:
> If you want, I can reduce d42a663 to themes build (,/searx/static).

+1

return42:
> If you want, I can reduce d42a663 to themes build (,/searx/static).

FYI : done.

PR #92, opened 2021-05-23
Xpath enhencements

return42:
## What does this PR do?

XPath enhancements, read along commit messages.

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -37,7 +37,14 @@ Search URL of the engine, replacements are:
 ``{pageno}``:
   Page number if engine supports pagging :py:obj:`paging`
 
+``{lang}``:
+  ISO 639-1 language code (en, de, fr ..)
 """
+
+lang_all='en'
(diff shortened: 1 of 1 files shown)


dalf:
Should the documentation:
* includes `.. automodule::` for the different engines ?
* and a link from the [engine names](https://searxng.github.io/searxng/admin/engines.html) to the module documentation.

Should the tooltip in the `/preferences` page include a link to source code &| the automodule documentation ? (especially if we know the git commit).

return42:
> includes .. automodule:: for the different engines ?

I don't think so, xpath is a generic engine which needs to be documented for the developers & admins.. others (dedicated) engines most often do not need a documentation.

If we see the need for documenting a engine, we can add documentation .. but we should not try to document a engine which do not need any special documentation.

> Should the tooltip in the /preferences page include a link to source code &| the automodule documentation ? (especially if we know the git commit).

Lets wait with .. ATM I do not see the benefit in .. BTW we have a link to the DOCS_URL

https://github.com/searxng/searxng/blob/8283ce3d79b67bad0820dad7e02ab8140ec44898/docs/conf.py#L45-L49

PR #93, opened 2021-05-24
Some minor Genius improvements

return42:
## What does this PR do?

Some minor Genius improvements, read along commit messages.

### DIFF
--- a/searx/engines/genius.py
+++ b/searx/engines/genius.py
@@ -1,11 +1,16 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Genius
+# lint: pylint
+# pylint: disable=invalid-name, missing-function-docstring
+"""Genius
+
 """
 
 from json import loads
 from urllib.parse import urlencode
 from datetime import datetime
+
+from searx import logger
+logger = logger.getChild('genius engine')
 
 # about
 about = {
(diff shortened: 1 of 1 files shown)


searx/engines/genius.py
```diff
@@ -27,49 +32,54 @@
...
+    )
     return params
 
 
 def parse_lyric(hit):
     try:
         content = hit['highlights'][0]['value']
-    except:
+    except Exception as e:  # pylint: disable=broad-except
+        logger.error(e, exc_info=True)
```
dalf, line 48: Is the exception an error that should be fixed ?
I feel the error logging here is only useful for debugging purpose.

This alternate version doesn't raise an exception:
```python
highlights = hit.get('highlights')
content = highlights[0].get('value', '') if highlights else ''
```

With logging, it would be only be `logger.debug` not `logger.error` :
```python
highlights = hit.get('highlights')
content = highlights[0].get('value', '') if highlights else ''
if content == '':
  logger.debug('content is empty')
```

searx/engines/genius.py
```diff
@@ -27,49 +32,54 @@
...
         content = ''
     timestamp = hit['result']['lyrics_updated_at']
-    result = {'url': hit['result']['url'],
-              'title': hit['result']['full_title'],
-              'content': content,
-              'thumbnail': hit['result']['song_art_image_thumbnail_url'],
-              'template': 'videos.html'}
```
dalf, line 47: Just note, not a request for change: technically this is right to use `default.html` instead of `videos.html`, but the oscar layout is different (it doesn't change anything for the simple theme).

searx/engines/genius.py
return42: This commit is about lint (see my https://github.com/searxng/searxng/issues/102#issuecomment-849640425) .. I don't want to change it functional.

searx/engines/genius.py
return42: IMO both are missing a result template for "music" hits.

PR #96, opened 2021-05-25
lint & fix unsplash engine

return42:
## What does this PR do?

[pylint] searx/engines/unsplash.py, add logger & norm indentation

- fix messages from pylint
- add logger and log request URL
- normalized various indentation

[fix] unsplash engine - 'searx:result: invalid title:'

- Use result 'alt_description' as title, if not given use
  default title 'unknown'.
- Use result 'description' from unsplash as 'content'

Fix error::

    DEBUG:searx:result: invalid title: {..., 'title': None, 'content': '', 'engine': 'unsplash'}

    make run

search e.g. `!unsplash dali`, there should be no more `DEBUG:searx:result: invalid title:` .. and you will see some more results than before .. (those with title 'unknown').

### DIFF
--- a/searx/engines/unsplash.py
+++ b/searx/engines/unsplash.py
@@ -62,7 +62,8 @@ def response(resp):
                 'url': clean_url(result['links']['html']),
                 'thumbnail_src': clean_url(result['urls']['thumb']),
                 'img_src': clean_url(result['urls']['raw']),
-                'title': result['description'],
-                'content': ''
+                'title': result.get('alt_description') or 'unknown',
+                'content': result.get('description') or ''
             })
+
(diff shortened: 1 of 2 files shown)


return42:
About commit : `[mod] unsplash engine - activated by default`

For me unsplash seems a reliable image source .. I activated unsplash by default in commit 97d173e.

This decision is arguable, what do other think?

dalf:
> This decision is arguable, what do other think?

I don't know.

A completely different approach can be this:
* all engines are disabled by default.
* during setup when the secret is set, the admin can choose the engines enabled by default.

But that's another issue & PR.

PR #97, opened 2021-05-25
[docs] reorder blog articles

return42:
For details read along the [commit messages](https://github.com/searxng/searxng/pull/97/commits)

Hint: there a two patches with a fix in this PR:

- https://github.com/searxng/searxng/pull/97/commits/97d801e8a097c33a384d0b44cdbbc61a985373a6
- https://github.com/searxng/searxng/pull/97/commits/f122cb0e27dbb07cb3b5926ae7910cdc65c8fcbb


docs/admin/engines/command-line-engines.rst
```diff
@@ -59,7 +62,3 @@ This development was sponsored by `Search and Discovery Fund`_ of `NLnet Foundat
 
 .. _Search and Discovery Fund: https://nlnet.nl/discovery
 .. _NLnet Foundation: https://nlnet.nl/
-
```
dalf: "Next steps" section can be removed.

docs/admin/engines/private-engines.rst
```diff
@@ -55,11 +53,5 @@ and come up with new solutions which fit your use case.
 Acknowledgement
 ===============
 
-This development was sponsored by `Search and Discovery Fund`_ of `NLnet Foundation`_ .
```
dalf: "Next steps" section can be removed.

return42: Thanks .. done .. I will push  it later .. there is a lot of other doc-stuff I am working on ATM ..

return42:
FYI: online HTML representation of this PR is available from https://return42.github.io/searxng/

searx/engines/demo_online.py
```diff
@@ -0,0 +1,92 @@
...
+# if there is a need for globals, use a leading underline
+_my_online_engine = None
+
+def init(engine_settings):
+    """Initialization of the (online) engine.  If no initialization is needed, drop
+    this init function.
+
+    """
+    global _my_online_engine  # pylint: disable=global-statement
+    _my_online_engine = engine_settings.get('name')
```
dalf, line 52: I understand it makes a demo of the `init` function, but the parameter is not used.

Out of topic for this PR, `engine_settings` values are [copied as attributes into the module](https://github.com/searxng/searxng/blob/5f76238d5cd6f9c850d5237fe639ba2900fd3e6c/searx/engines/__init__.py#L117). So `engine_settings.get('name')` could be `name` (with `global _my_online_engine, name` above).

And `engine_settings` can be:
```python
{
    k: getattr(engine, k) 
    for k in dir(engine)
    if not k.startswith('__') and not callable(getattr(engine, k))
}
```
instead of [get_engine_from_settings](https://github.com/searxng/searxng/blob/5f76238d5cd6f9c850d5237fe639ba2900fd3e6c/searx/utils.py#L488-L500)

docs/dev/engine_overview.rst
```diff
@@ -1,310 +1,317 @@
...
-A few of the options have default values in the engine, but are often
-overwritten by the settings.  If ``None`` is assigned to an option in the engine
-file, it has to be redefined in the settings, otherwise searx will not start
-with that engine.
+.. sidebar:: engine's global names
 
-The naming of overrides is arbitrary.  But the recommended overrides are the
-following:
+   Global names with a leading underline are *private to the engine* and will
+   not be overwritten.
```
dalf, line 91: Technically, names starting with `_` can be overwritten but they can be `None`.

dalf (approved):
LGTM

docs/dev/engine_overview.rst
return42: Ooops  .. I forgot to reflect your hint into the documentation.  What Do you think, is it better to fix the documentation or is it better to fix the code that overrides _private members_?

dalf: I think it is better to fix the documentation.

searx/engines/demo_online.py
return42: Hm, I guess I have a problem to understand why we have this argument `engine_settings` .. does it make sense to drop this argument? .. for me it was a bit confusing that we update module's name space with the values from YAML (`engine_data`):

https://github.com/searxng/searxng/blob/5f76238d5cd6f9c850d5237fe639ba2900fd3e6c/searx/engines/__init__.py#L117-L125

and later, when the processor is initialized ...

https://github.com/searxng/searxng/blob/5c5db719d2039d34aa1426bf7eb3e57055d4a54a/searx/search/processors/abstract.py#L70-L72

we call a function that gather `engine_data` from the YAML settings.

In sense of _single point of definition_ I guess it is better to say, that everything is in the namespace of engine's modul and this can be configured by (is updated from) the YAML file.

But may be I'm totally wrong, since I do not have a complete overview.

PR #99, opened 2021-05-26
[enh] add settings option to enable/disable search formats

### LINKED ISSUE #95: Add ability to disable RSS/JSON/CSV
Bots are the only users of this feature on a public instance, and they abuse it too much that the engines rate limit pretty quickly the IP address of the instance.

I would like to disable the results in RSS/JSON/CSV format.

Using filtron in order to block the use of the RSS/JSON/CSV.

**Additional context**

return42:
## What does this PR do?

- eff337a4 [enh] add settings option to enable/disable search formats

## How to test local

In settings.yml edit

    search:
        ...
        formats: [html]

The denied formats should no longer available in the link box:

image

Request a format: ( http://127.0.0.1:8888/search?q=foo&format=csv ) should result in a 403

image

## Related issues

Closes #95

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -505,13 +505,14 @@ NOT_EXISTS = object()
 """Singleton used by :py:obj:`get_value` if a key does not exists."""
 
 
-def get_value(dictionary, keyword, *keys, default=NOT_EXISTS):
+def get_value(dictionary, *keys, default=NOT_EXISTS):
     """Return the value from a *deep* mapping type (e.g. the ``settings`` object
     from yaml).  If the path to the *key* does not exists a :py:obj:`NOT_EXISTS`
     is returned (non ``KeyError`` exception is raised).
 
     .. code: python
 
+       >>> from searx import settings
        >>> from searx.utils import get_value, NOT_EXISTS
        >>> get_value(settings, 'checker', 'additional_tests', 'rosebud', 'result_container')
        ['not_empty', ['one_title_contains', 'citizen kane']]
@@ -519,7 +520,7 @@ def get_value(dictionary, keyword, *keys, default=NOT_EXISTS):
        >>> get_value(settings, 'search', 'xxx') is NOT_EXISTS
        True
        >>> get_value(settings, 'search', 'formats')
-       ['csv', 'json', 'rss']
+       ['html', 'csv', 'json', 'rss']
 
     The list returned from the ``search.format`` key is not a mapping type, you
     can't traverse along non-mapping types.  If you try it, you will get a
@@ -529,7 +530,7 @@ def get_value(dictionary, keyword, *keys, default=NOT_EXISTS):
 
        >>> get_value(settings, 'search', 'format', 'csv') is NOT_EXISTS
        True
-       >>> get_value(settings, 'search', 'formats')[0]
+       >>> get_value(settings, 'search', 'formats')[1]
        'csv'
 
     For convenience you can replace :py:ref:`NOT_EXISTS` by a default value of
@@ -541,20 +542,15 @@ def get_value(dictionary, keyword, *keys, default=NOT_EXISTS):
            print("csv format is denied")
 
     """
-    if not isinstance(dictionary, Mapping):
-        raise TypeError("expected mapping type, got %s" % type(dictionary))
 
-    ret_val = dictionary.get(keyword, default)
-
-    if ret_val is default:
-        return ret_val
-
-    if len(keys):
-        if not isinstance(ret_val, Mapping):
-            ret_val = default
-        else:
-            ret_val = get_value(ret_val, *keys, default=default)
-    return ret_val
+    obj = dictionary
+    for k in keys:
+        if not isinstance(obj, Mapping):
+            raise TypeError("expected mapping type, got %s" % type(obj))
+        obj = obj.get(k, default)
+        if obj is default:
+            return obj
+    return obj
 
 
 def get_xpath(xpath_spec):
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -8,6 +8,7 @@ from os.path import splitext, join
 from random import choice
 from html.parser import HTMLParser
 from urllib.parse import urljoin, urlparse
+from collections.abc import Mapping
 
 from lxml import html
 from lxml.etree import ElementBase, XPath, XPathError, XPathSyntaxError, _ElementStringResult, _ElementUnicodeResult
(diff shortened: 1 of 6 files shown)


dalf:
Is it possible to have 192a6f4fa7cd61489f5e50fcc1d0e2b255c84478 / e9a38ff5e83715bb471972d0731922bde1306cd8 and e9a38ff5e83715bb471972d0731922bde1306cd8 in another PR ?

---

About e9a38ff5e83715bb471972d0731922bde1306cd8

Instead of `access_denied.search/format` , what do you think  of `search.formats` with a default value of `[csv, json, rss]` (since `html` is always included) ?

return42:
> Is it possible to have 192a6f4 / e9a38ff and e9a38ff in another PR ?

if you want ..  and we review this PR first.

>  what do you think of search.formats with a default value of [csv, json, rss] (since html is always included) ?

I can change to `search.formats`.

dalf:
About the `get_value` function: should the PR 2501 from searx should be merged?

return42:
>  should the PR 2501 from searx should be merged?

Hm 2501 LGTM .. but can't rate it in deep .. to get #95 done I vote to merge 2501 in another PR.

dalf:
>  I vote to merge 2501 in another PR.

Yes sure.

return42:
> what do you think of search.formats with a default value of [csv, json, rss] (since html is always included) ?

All formats can be (de)selected / including html:

https://github.com/searxng/searxng/pull/99/commits/23f6c4cc98a5b4c4accd2fbdcede1c397ff5f982#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR21

**HINT** If admins maintain their own settings.yml file, they need to add these formats to the search section:

     search:
          ...
          formats: [html, csv, json, rss]

Otherwise they will get only "403 Forbidden".  The robot test is one example, where I needed to add this line:

https://github.com/searxng/searxng/pull/99/commits/23f6c4cc98a5b4c4accd2fbdcede1c397ff5f982#diff-ceeab343071eef75af2f6bf9d5fa7a3f73802313f03b519291e75ba55f1c91d2

searx/webapp.py
```diff
@@ -686,6 +691,9 @@ def search():
     if output_format not in ['html', 'csv', 'json', 'rss']:
         output_format = 'html'
 
+    if output_format not in get_value(settings, 'search', 'formats', default=[]):
```
dalf: `default=[]` --> `default=['html', 'csv', 'json', 'rss']` otherwise it will break existing setup (without `search.formats`).

Looking at the code above:

```python
OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']

...
    allowed_formats = get_value(settings, 'search', 'formats', default=OUTPUT_FORMATS)
    output_format = request.form.get('format')
    if output_format not in OUTPUT_FORMATS:
        # undefined or unknown format: fallback to the first allowed format
        # default setting: 'html' 
        output_format  = allowed_formats[0]
    elif output_format not in allowed_formats:
        # the format is valid but not allowed: error 403
        flask.abort(403)
```

---

The fallback part keeps the API compatibility (`if output_format not in OUTPUT_FORMATS`)

searx/utils.py
```diff
@@ -500,6 +501,62 @@ def get_engine_from_settings(name):
...
+
+    if ret_val is default:
+        return ret_val
+
+    if len(keys):
+        if not isinstance(ret_val, Mapping):
+            ret_val = default
+        else:
+            ret_val = get_value(ret_val, *keys, default=default)
+    return ret_val
```
dalf: A version without recursive call:

```python
def get_value(dictionary, *keys, default=NOT_EXISTS):
    obj = dictionary
    for k in keys:
        if not isinstance(obj, Mapping):
            raise TypeError("expected mapping type, got %s" % type(obj))
        obj = obj.get(k, default)
        if obj is default:
            return obj
    return obj
```

return42: > otherwise it will break existing setup (without search.formats).

good idea to set the default including all formats .. but should we set the default in the code above or is it better placed nearby loading the settings?

Next thought: is it good to have defaults in settings.yml and other defauts in the code (e.g. nearby loading settings).

What I want is a single point of definitions .. ATM this single point is the settings,yml .. should we really add a new place where defaults could be located? (hope I was clear)

searx/webapp.py
return42: > otherwise it will break existing setup (without search.formats).

good idea to set the default including all formats .. but should we set the default in the code above or is it better placed nearby loading the settings?

Next thought: is it good to have defaults in settings.yml and other defauts in the code (e.g. nearby loading settings).

What I want is a single point of definitions .. ATM this single point is the settings,yml .. should we really add a new place where defaults could be located? (hope I was clear)

searx/utils.py
return42: > A version without recursive call:

Thanks, good suggestion / implemented in 61c3cf46

searx/webapp.py
dalf: > placed nearby loading the settings

I don't understand the location.

A simple solution for now: a the top of webapp.py (?)

---

With the searx PR 2501:
* the default value is in `searx/settings_defaults.py`
* `allowed_formats = get_value(settings, 'search', 'formats', default=OUTPUT_FORMATS)` --> `allowed_formats = settings['search']['formats']` (we are sure there is a value)

As discuss in this PR, a Settings class would be better, then the Preferences can use a Settings instance.

return42: > I don't understand the location.

I mean a file like ..

> With the searx PR 2501:
>    the default value is in searx/settings_defaults.py

I think a solution like PR2501 will help ... 

> As discuss in this PR, a Settings class would be better, then the Preferences can use a Settings instance.

PR2501 also implements static_path handling .. for me it is a bit hard to merge all this into tthe current searxng without to fail ... If you have time, do you like to merge this PR?  If so, I would vote to implement a class.

return42:
Implemented `OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']` in https://github.com/searxng/searxng/pull/99/files#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R167

searx/webapp.py
```diff
@@ -683,9 +693,12 @@ def search():
 
     # output_format
     output_format = request.form.get('format', 'html')
-    if output_format not in ['html', 'csv', 'json', 'rss']:
+    if output_format not in OUTPUT_FORMATS:
         output_format = 'html'
 
+    if output_format not in get_value(settings, 'search', 'formats', default=OUTPUT_FORMATS):
+        flask.abort(403)
```
dalf, line 700: Looking at the documentation: https://flask.palletsprojects.com/en/2.0.x/api/#flask.abort

There is no need for an exception, just `return 'Access denied', 403` is enough.

return42: Hm, there is a difference .. 

AFAIK:  just a `return 'Access denied', 403` will be handled by the flask app (Flask generates a 403 response with the given string).   `flask.abort(403)` will raise an exception and the 403 page comes from the HTTP server.

If I was unclear, try both versions in a local instance (make run).

The question for me is: what do we want, should flask build the response or should the 403 be propagated over the http server? I decided to give the 403 back to the http server.

dalf: > the 403 page comes from the HTTP server.

So either Werkzeug or uwsgi ?

* https://uwsgi-docs.readthedocs.io/en/latest/Changelog-1.9.21.html#error-pages

>  I decided to give the 403 back to the http server.

why ?

what about: 

https://github.com/searxng/searxng/blob/b48b4c93d586f52418295a15a1c93934f007c41f/searx/webapp.py#L1106-L1112

?

return42: > So either Werkzeug or uwsgi ?

AFAIK: Whatever (error page) is configured in the http server. Does not depend on the technology which is used to bind the Flask-App.

>  why?

It was just my gut feeling: I thought is a **Forbidden** (403)  which is better passed upwards.

The places you quote are  **Bad Request**"s (400) .. means: _"the request could not be understood by the server due to malformed syntax."_ which is IMO different to "**Forbidden**".

I do not have much experiences and I don't want to judge it .. if you prefer 403 handled by the Flask-App it is OK for me .. if so, the  _"why?"_ left on my side :)

return42: I don't know if it helps, but Flask's documentation says: [Redirects and Errors: ... to abort a request early with an error code, use the abort() function:](https://flask.palletsprojects.com/en/2.0.x/quickstart/#redirects-and-errors)

dalf: > AFAIK: Whatever (error page) is configured in the http server. Does not depend on the technology which is used to bind the Flask-App.

I wish to reduce the number of configuration locations.

But it returns an HTTP 403 whatever the solution.

PR #100, opened 2021-05-26
[pylint] webapp.py 

return42:
## What does this PR do?

only lint no functional change

1. cc7fcf0d * [pylint] webapp.py - fix all messages from pylint
2. ae253622 [pylint] webapp.py - fix all wrong-import-order messages from pylint

removed `__main__`:

3. 85364eda [fix] webapp.py - running without installation is not supported

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -2,7 +2,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
 # pylint: disable=missing-function-docstring
+"""WebbApp
 
+"""
 import hashlib
 import hmac
 import json
@@ -122,13 +124,14 @@ else:
     old_thread_init = threading.Thread.__init__
 
     def new_thread_init(self, *args, **kwargs):
+        # pylint: disable=protected-access, disable=c-extension-no-member
         old_thread_init(self, *args, **kwargs)
         setproctitle.setthreadtitle(self._name)
     threading.Thread.__init__ = new_thread_init
 
 if sys.version_info[0] < 3:
     print('\033[1;31m Python2 is no longer supported\033[0m')
-    exit(1)
+    sys.exit(1)
 
 logger = logger.getChild('webapp')
 
@@ -138,7 +141,7 @@ WSGIRequestHandler.protocol_version = "HTTP/{}".format(settings['server'].get('h
 # check secret_key
 if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
     logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
-    exit(1)
+    sys.exit(1)
 
 # about static
 static_path = get_resources_directory(searx_dir, 'static', settings['ui']['static_path'])
@@ -157,6 +160,14 @@ for indice, theme in enumerate(themes):
     theme_img_path = os.path.join(static_path, 'themes', theme, 'img', 'icons')
     for (dirpath, dirnames, filenames) in os.walk(theme_img_path):
         global_favicons[indice].extend(filenames)
+
+STATS_SORT_PARAMETERS = {
+    'name': (False, 'name', ''),
+    'score': (True, 'score', 0),
+    'result_count': (True, 'result_count', 0),
+    'time': (False, 'total', 0),
+    'reliability': (False, 'reliability', 100),
+}
 
 # Flask app
 app = Flask(
@@ -252,8 +263,8 @@ def _get_translations():
 flask_babel.get_translations = _get_translations
 
 
-def _get_browser_or_settings_language(request, lang_list):
-    for lang in request.headers.get("Accept-Language", "en").split(","):
+def _get_browser_or_settings_language(req, lang_list):
+    for lang in req.headers.get("Accept-Language", "en").split(","):
         if ';' in lang:
             lang = lang.split(';')[0]
         if '-' in lang:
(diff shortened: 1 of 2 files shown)


searx/webapp.py
```diff
@@ -622,12 +666,14 @@ def index():
     )
 
 
-@app.route('/search', methods=['GET', 'POST'])
-def search():
+@app.route('/search', endpoint='search', methods=['GET', 'POST'])
+def route_search():
```
dalf: I guess the purpose is to avoid name conflict?
What about moving functions that are not routes in some other modules like in https://github.com/searx/searx/pull/2313 ?

searx/webapp.py
```diff
@@ -659,8 +704,9 @@ def search():
     except SearxParameterException as e:
         logger.exception('search error: SearxParameterException')
         return index_error(output_format, e.message), 400
-    except Exception as e:
-        logger.exception('search error')
+    except Exception as e:  # pylint: disable=broad-except
+        logger.exception(e, exc_info=True)
+        logger.debug('search error')
```
dalf: Is `logger.exception("search error")` not enough?

searx/webapp.py
```diff
@@ -1113,18 +1174,18 @@ def stats():
     engine_stats = get_engines_stats(filtered_engines)
     engine_reliabilities = get_reliabilities(filtered_engines, checker_results)
 
-    SORT_PARAMETERS = {
+    sort_parameters = {
```
dalf: Before this PR, this variable should have been declared at the top of the file as "constant".

searx/webapp.py
return42: For me? .. yes :-) .. I will remove the logger.debug message.

searx/webapp.py
return42: good point, I will move it to the top of the file

searx/webapp.py
return42: > I guess the purpose is to avoid name conflict?

yes

> What about moving functions that are not routes in some other modules

I think we should progress modularization more and more in the future, but here it would not help for names like

    @app.route('/preferences', endpoint='preferences', methods=['GET', 'POST'])
    def preferences():

searx/webapp.py
return42: fixed here https://github.com/searxng/searxng/commit/e295bbf0679fe1def6983be969b735ef20d26c19#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R164-R171

searx/webapp.py
return42: fixed here:
- https://github.com/searxng/searxng/commit/e295bbf0679fe1def6983be969b735ef20d26c19#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R320-R322
- https://github.com/searxng/searxng/commit/e295bbf0679fe1def6983be969b735ef20d26c19#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07L309
- https://github.com/searxng/searxng/commit/e295bbf0679fe1def6983be969b735ef20d26c19#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R713-R714

searx/webapp.py
dalf: I prefer to keep the straightforward names:
* The full name is `searx.webapp.preferences`.
* we can rename this to `searx.routes.preferences`, as it is done in [whoogle](https://github.com/benbusby/whoogle-search/blob/main/app/routes.py), so this is explicite --> webapp.py would be the launcher.
* all Flask application I've seen use `endpoint` only as a last resort.

return42: > I prefer to keep the straightforward names:
> * The full name is `searx.webapp.preferences`.

The endpoint name is the name of the function `preferences`.

> * we can rename this to `searx.routes.preferences`, as it is done in [whoogle](https://github.com/benbusby/whoogle-search/blob/main/app/routes.py), so this is explicite --> webapp.py would be the launcher.

Yes, on a long term we should modularise into a module named `routes` .. but the endpoint name will be the name of the function.

> * all Flask application I've seen use `endpoint` only as a last resort.

They have clean namespaces and modularisation

This PR is dedicated to linit and modularisation is not in the scope .. we can decide this later.  I add a `# pylint: disable=redefined-outer-name` to the lines causing name clashes.

dalf: > but the endpoint name will be the name of the function.

does it cause any sort of problem?

return42: I am not familiar in deep but for flask itself endpoints are substantial for the routing.

- see section  [Endpoints and URLs](https://flask.palletsprojects.com/en/2.0.x/tutorial/views/#endpoints-and-urls)
- https://flask.palletsprojects.com/en/2.0.x/api/#flask.Flask.add_url_rule

We use it here (calls are coming from the templates):

https://github.com/searxng/searxng/blob/b1da97d0922020dabcdc971d16e3e7a95be58c1b/searx/webapp.py#L346-L353

And we use this:

https://github.com/searxng/searxng/blob/b1da97d0922020dabcdc971d16e3e7a95be58c1b/searx/webapp.py#L473

to set the `<body class="{{ endpoint }}_endpoint" >` in the templates.

return42:
I solved your requests:

- https://github.com/searxng/searxng/pull/100#pullrequestreview-669474568
- https://github.com/searxng/searxng/pull/100#pullrequestreview-669475561

And  I add a `# pylint: disable=redefined-outer-name` to the lines causing name clashes / [removed the endpoint](https://github.com/searxng/searxng/pull/100#discussion_r640504624).

Can we merge this PR?

PR #107, opened 2021-05-28
[enh] add engine Wikimini (fr.wikimini.org)

return42:
Online encyclopedia for children (only language fr), merged from [1]

[1] https://github.com/searx/searx/pull/2819

I can't judge (only fr) whether we really need these engine.

@dalf: if you think we don't need it, just close this PR.


dalf:
>  if you think we don't need it, just close this PR.

why not

PR #109, opened 2021-05-28
Merge MySQL and PostgreSQL from searx

return42:
## What does this PR do?

Merge MySQL and PostgreSQL from searx / for details read along the commit messages.

A HTML preview of the documentation is available from: https://return42.github.io/searxng/blog/sql-engines.html

### DIFF
--- a/searx/engines/mysql_server.py
+++ b/searx/engines/mysql_server.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=missing-function-docstring
+"""MySQL database (offline)
+
+"""
+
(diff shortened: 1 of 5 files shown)


searx/engines/mysql_server.py
```diff
@@ -0,0 +1,57 @@
...
+auth_plugin = 'caching_sha2_password'
+host = "127.0.0.1"
+database = ""
+username = ""
+password = ""
+query_str = ""
+limit = 10
+paging = True
+result_template = 'key-value.html'
+_connection = None
```
dalf, line 22: This will raise an exception when the engine is loaded:
https://github.com/searxng/searxng/blob/da111f7fec1c377696ed33a3c97cb5665bb01ab9/searx/engines/__init__.py#L102-L104

PR #110, opened 2021-05-28
[mod] move all default settings into searx.settings_defaults

dalf:
## What does this PR do?

see searx PR 2501 and... 2313 (startlette).

* searx loads `searx/settings.yml` into the `settings` variable.
* then, the default values are defined usually using this pattern similar to `settings[....].get('aVariable', default)` 
* it fails for `settings['ui']['static_path']`, see https://github.com/searx/searx/issues/2252#issuecomment-764755567

This PR adds a new function `searx.settings_set_defaults.settings_set_defaults` to set all the default values in one place.

Then instead of 
```
static_path = get_resources_directory(searx_dir, 'static', settings['ui']['static_path'])
```

It is possible to write ```settings['ui']['static_path']```

See https://github.com/searx/searx/issues/2252#issuecomment-764755567 (more precisely the `static_path` definition)

This PR is request for comment / for archive : most probably a better solution is to create some classes in similar way to the `*Settings` classes.

* to be written

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -1,0 +1,194 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+import typing
+import numbers
+import errno
+import os
+import logging
+from os.path import dirname, abspath
+
+from searx.languages import language_codes as languages
+
+searx_dir = abspath(dirname(__file__))
+
+logger = logging.getLogger('searx')
+OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']
+LANGUAGE_CODES = ('', 'all') + tuple(l[0] for l in languages)
+OSCAR_STYLE = ('logicodev', 'logicodev-dark', 'pointhi')
+CATEGORY_ORDER = [
+    'general',
+    'images',
+    'videos',
+    'news',
+    'map',
+    'music',
+    'it',
+    'science',
+    'files',
+    'social medias',
+]
+STR_TO_BOOL = {
+    '0': False,
+    'false': False,
+    'off': False,
+    '1': True,
+    'true': True,
+    'on': True,
+}
+_UNDEFINED = object()
+
+
+class SettingsValue:
+    """Check and update a setting value
+    """
+
+    def __init__(self,
+                 type_definition: typing.Union[None, typing.Any, typing.Tuple[typing.Any]]=None,
+                 default: typing.Any=None,
+                 environ_name: str=None):
+        self.type_definition = type_definition \
+                               if type_definition is None or isinstance(type_definition, tuple) \
+                               else (type_definition,)
+        self.default = default
+        self.environ_name = environ_name
+
+    @property
+    def type_definition_repr(self):
+        types_str = [t.__name__ if isinstance(t, type) else repr(t)
+                         for t in self.type_definition]
+        return ', '.join(types_str)
+
+    def check_type_definition(self, value: typing.Any) -> None:
+        if value in self.type_definition:
+            return
+        type_list = tuple(t for t in self.type_definition if isinstance(t, type))
+        if not isinstance(value, type_list):
+            raise ValueError('The value has to be one of these types/values: {}'\
+                             .format(self.type_definition_repr))
+
+    def __call__(self, value: typing.Any) -> typing.Any:
+        if value == _UNDEFINED:
+            value = self.default
+        # override existing value with environ
+        if self.environ_name and self.environ_name in os.environ:
+            value = os.environ[self.environ_name]
+            if self.type_definition == (bool,):
+                value = STR_TO_BOOL[value.lower()]
+        #
+        self.check_type_definition(value)
+        return value
+
+
+class SettingsDirectoryValue(SettingsValue):
+    """Check and update a setting value that is a directory path
+    """
+
+    def check_type_definition(self, value: typing.Any) -> typing.Any:
+        super().check_type_definition(value)
+        if not os.path.isdir(value):
+            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), value)
+
+    def __call__(self, value: typing.Any) -> typing.Any:
+        if value == '':
+            value = self.default
+        return super().__call__(value)
+
+
+def apply_schema(settings, schema, path_list):
+    error = False
+    for key, value in schema.items():
+        if isinstance(value, SettingsValue):
+            try:
+                settings[key] = value(settings.get(key, _UNDEFINED))
+            except Exception as e:  # pylint: disable=broad-except
+                # don't stop now: check other values
+                logger.error('%s: %s', '.'.join([*path_list, key]), e)
+                error = True
+        elif isinstance(value, dict):
+            error = error or apply_schema(settings.setdefault(key, {}), schema[key], [*path_list, key])
+        else:
+            settings.setdefault(key, value)
+    if len(path_list) == 0 and error:
+        raise ValueError('Invalid settings.yml')
+    return error
+
+
+SCHEMA = {
+    'general': {
+        'debug': SettingsValue(bool, False, 'SEARX_DEBUG'),
+        'instance_name': SettingsValue(str, 'searxng'),
+        'contact_url': SettingsValue((None, False, str), None),
+    },
+    'brand': {
+    },
+    'search': {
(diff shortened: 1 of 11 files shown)


searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+
+def default(value):
+    def apply(existing, _path):
+        return value if not existing else existing
+    return apply
+
+
+def is_instance(types):  # pylint: disable=unused-argument
+    def apply(existing, path):
+        nonlocal types
```
return42: Does it make sense 

- to name the inner function `def _inner(..):`
- to name argument `types` --> `type_list`
- add a doc-string that this function returns a _closure_ , what are the arguments of the _closures_ and for what the _closures_  is used ... 

Most often it is very helpful to have simple code examples in the doc-string (e.g. like [here](https://github.com/searxng/searxng/pull/110/files#diff-8d38f1d52f68ec93ec84ebb815fbbb6240fd70d8076dc11ee55021ad58c5658eL517-L523))

// not only here .. lets add small doc-strings to all _closures_.

searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+            sys.exit(1)
+        return existing
+    return apply
+
+
+def override_by_environ(environ_name):
+    def apply(existing, _path):
+        if environ_name in os.environ:
+            return os.environ[environ_name]
+        return existing
```
return42: can be simplified to:
```python
    return os.environ.get(environ_name, existing)
```

searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+
+searx_dir = abspath(dirname(__file__))
+
+logger = logging.getLogger('searx')
+OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']
+
+
+def default(value):
+    def apply(existing, _path):
+        return value if not existing else existing
```
return42: Can be simplified to:

```python
return existing or value
```

Why does the caller of the _closure_ calls the _closure_ with unneeded arguments?

Does it make sense to catch all unneeded arguments by `def inner(existing, *_args, **_kwargs)`

// do you know, why we do not need `nonlocal value` here?

searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def is_instance(types):  # pylint: disable=unused-argument
+    def apply(existing, path):
+        nonlocal types
+        if None in types and existing is None:
+            return existing
+        types = tuple(filter(lambda t: t is not None, types))
+        if not isinstance(existing, types):
+            types_str = list(map(lambda t: t.__name__ if t is not None else 'None', types))
```
return42: I would suggest, that a more simple (and robust) form should be good enough for the log message:

```python
    types_str_list = [repr(t) for t in type_list]
```

searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+                return False
+            logger.warning('%s : ignore value from %s environment variable', '.'.join(path), environ_name)
+        return existing
+    return apply
+
+
+def get_resources_directory(searx_directory, subdirectory):
+    resources_directory = os.path.join(searx_directory, subdirectory)
+    if not os.path.isdir(resources_directory):
+        raise Exception(resources_directory + " is not a directory")
```
return42: [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) fits better.

searx/settings_defaults.py
```diff
@@ -0,0 +1,172 @@
...
+    return resources_directory
+
+
+def apply_schema(settings, schema, path):
+    for key, value in schema.items():
```
return42, line 109: Argument `path` should be named `path_list`

return42 (changes requested):
I really like the concept of having a _schema_  

What bothers me a bit is the fact that we have the settings implementations spread over different files (settings_defaults, setting_loader). I would like to prefer a class like the following, which implements the schema and the loader:

```python
class Settings:

    schema = {
        ...
    }

    def __init_(self):
        self.setings = None

    def load_settings(self, ...):
        # here comes the implementation from searx.settings_loader.load_settings
        ...

    def apply_schema(self, path_list):
        # here self.schema & self.settings can be used
```

searx/settings_defaults.py
dalf: > Why does the caller of the closure calls the closure with unneeded arguments?

To be able to log a warning:
https://github.com/searxng/searxng/blob/cc05240031cc3ab3cfb67173376d96c39e55252e/searx/settings_defaults.py#L52

> Does it make sense to catch all unneeded arguments by def inner(existing, *_args, **_kwargs)

The returned closures are called by this line, and only by this line:
https://github.com/searxng/searxng/blob/cc05240031cc3ab3cfb67173376d96c39e55252e/searx/settings_defaults.py#L74 

----

The idea of `apply_schema`:
* for each `key`, `value` of the `schema` do `settings.setdefault(key, value)` with recursive calls if `value` is a `dict`.
* since there is a recursive call, `settings` here might actually be `settings['ui']['theme_args']` (for example).
* but there are two exceptions:
   * if `value` is `callable`or `tuple` (both won't be found in `settings`)
   * if `value` is a `tuple` the code expects that all items of tuple are `callable`
   * for each `callable` named `f` do `settings[key]=f(settings[key], path)`.

----

Perhaps a class like with a constructor like this:
`SettingsValue(type_list=[str], environ_name='SEARX_BIND_ADDRESS', default='127.0.0.1')`
is better (all arguments are optional).

return42: > The returned closures are called by this line, and only by this line:

Ah, this was the point I missed / thanks. Another point I missed: `settings` is the dictionary from `searx.settings_loader.load_settings()` .. I think it would be more _readable_ when we have a class which implements all the _settings_ stuff.

> Perhaps a class like with a constructor like this:

See my example, not a `SettingsValue` class, I mean a `Settings` class which implements the whole setting stuff (this from `settings_loader` and this from `settings_defaults`).

With the hints you gave above I can try to implement such a class in a separated branch (the branch will start on top of your's). But I can't predict when I have time for.

searx/settings_defaults.py
```diff
@@ -0,0 +1,194 @@
...
+        if isinstance(value, SettingsValue):
+            try:
+                settings[key] = value(settings.get(key, _UNDEFINED))
+            except Exception as e:  # pylint: disable=broad-except
+                # don't stop now: check other values
+                logger.error('%s: %s', '.'.join([*path_list, key]), e)
+                error = True
+        elif isinstance(value, dict):
+            error = error or apply_schema(settings.setdefault(key, {}), schema[key], [*path_list, key])
```
dalf, line 118: Raise an `ValueError` exception at the end, so `apply_schema` can scan all the values.

searx/settings_defaults.py
return42: FYI: started to implement a class `Settings` .. it s far from finish but might be illustrate what I mean, see my [mod-default-settings](https://github.com/return42/searxng/commits/mod-default-settings) branch.

dalf: Few comments on the spot / idea:
* What do you think of a module `searx.settings` so the loader part is in different file ? Purpose: avoid a huge file doing everything.
* `settings = Settings(...)`  then `settings['ui']['static_path'] should work as before, otherwise it will requires too much changes everywhere.
* but  for example `get_themes`, `get_static_files` and `get_result_templates` (from searx.webutils) can move into this class in a way another.

( IMO, If it includes the change from this PR, it should be another PR )

return42: > What do you think  ...

Yes, this is what I also thought, at least _more or less_  because I haven't (can't) yet thought all to its end.

> IMO, If it includes the change from this PR, it should be another PR 

Yes, lets first finish this PR .. afterwards we can discuss how to assemble all or only a part in a new settings.py file (module).

searx/settings_defaults.py
dalf: see 
```python
    @property
    def type_definition_repr(self):
        types_str = [t.__name__ if isinstance(t, type) else repr(t)
                         for t in self.type_definition]
        return ', '.join(types_str)
```

so instead of ```<class '__main__.A'>```, the function returns ```A```. It is just a helper.

searx/settings_defaults.py
dalf: see ```SettingsValue``` class.

return42 (approved):
From my review I pushed 495d86b on top, as usual :) ..  take along with this PR or drop what is not needed.

PR #81, opened 2021-05-16
 [fix] make /stats more CSP compliant

return42:
## What does this PR do?

Make /stats more CSP compliant, for details read commit messages.

## Why is this change important?

Reported-by: #57

## How to test this PR locally?

Test locally (see below), before we merge I will test this PR with CSP _activated_ on my [public instance](https://darmarit.org/searx/stats).

Aktivate checker ..

```diff
modified   searx/settings.yml
@@ -106,7 +106,7 @@ outgoing: # communication with search engines
 
 checker:
     # disable checker when in debug mode
-    off_when_debug: True
+    off_when_debug: False
 
     # scheduling: interval or int
     # use "scheduling: False" to disable scheduling
@@ -114,9 +114,9 @@ checker:
     # * uncomment "scheduling" section
     # * add "cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1" to your uwsgi.ini
 
-    # scheduling:
-    #    start_after: [300, 1800]  # delay to start the first run of the checker
-    #    every: [86400, 90000]  # how often the checker runs
+    scheduling:
+        start_after: [10, 20]  # delay to start the first run of the checker
+        every: [1000, 2000]  # how often the checker runs
 
     # additional tests: only for the YAML anchors (see the engines section)
```

Start instance ..

    make run 

Doe some queries, wait until checker has finished .. and after visit /stats page

PR #113, opened 2021-05-29
[mod] remove overpass API call

dalf:
## What does this PR do?

* prepare the code for the PR #90

* pros: remove a direct call to overpass from the browser.
* cons: the photon results won't contains the same information:
  * it is possible get the wikidata id from OSM, and it is fast:
       * for example: https://api.openstreetmap.org/api/0.6/nodes.json?nodes=240109189,204648
       * https://wiki.openstreetmap.org/wiki/API_v0.6#JSON_Format
       * then it is possible to get the images, etc...
       * that's mean 3 fast requests (photon, osm, wikidata).

* `!map ushuai` (or whatever location)
* check using the oscar and simple themes.


return42:
Suggestion: to simplify the review, can we separate the patch of the static files from the patches that change the sources ... compare #81 where only the last patch commits files changed by `make themes.all` ...

1. cdfb4b7f [fix] make /stats more CSP compliant - bar graph
2. 0a6c4884 [fix] make /stats more CSP compliant - simple theme
3. 34859d0e [fix] make /stats more CSP compliant - oscar theme
4. 3e9ad7ae [fix] make /stats more CSP compliant - github issue form
5. 0507e185 [fix] bar graph and rename CSS class engine-scores -> engine-score
6. 73d73dcf [static] make /stats more CSP compliant- make themes.all

return42:
@dalf is it OK if I rebase this branch and split 27a9c95 into two commits?

dalf:
Thank you.

Side note: I've started to see if it is possible to compile the theme using the github action: https://github.com/dalf/searxng/blob/17edb2613d8aaee83e346d33b219e72f53890741/.github/workflows/themes.yml

PR #116, opened 2021-05-29
[mod] searx.engines.__init__: refactoring

dalf:
## What does this PR do?

See comments


searx/engines/__init__.py
```diff
@@ -67,61 +53,57 @@ def load_engine(engine_data):
...
             sys.exit(1)
 
     # assign supported languages from json file
-    if engine_data['name'] in ENGINES_LANGUAGES:
-        setattr(engine, 'supported_languages', ENGINES_LANGUAGES[engine_data['name']])
+    if engine_name in ENGINES_LANGUAGES:
+        setattr(engine, 'supported_languages', ENGINES_LANGUAGES[engine_name])
 
     # find custom aliases for non standard language codes
-    if hasattr(engine, 'supported_languages'):
```
dalf, line 112: The condition is removed because it is always True: `engine_default_args` contains `supported_languages`, so the ` for arg_name, arg_value in engine_default_args.items():...` set the attribute `supported_languages`.

searx/engines/__init__.py
```diff
@@ -67,61 +53,57 @@ def load_engine(engine_data):
...
+    # check if engine is inactive
+    if getattr(engine, 'inactive', False) is True:
+        return None
 
     # checking required variables
     for engine_attr in dir(engine):
         if engine_attr.startswith('_'):
             continue
-        if engine_attr == 'inactive' and getattr(engine, engine_attr) is True:
-            return None
```
dalf, line 101: Moved above.

searx/engines/__init__.py
```diff
@@ -67,61 +53,57 @@ def load_engine(engine_data):
...
                 engine.categories = list(map(str.strip, param_value.split(',')))
-        else:
+        elif param_name != 'engine':
             setattr(engine, param_name, param_value)
 
+    # set default attributes
     for arg_name, arg_value in engine_default_args.items():
         if not hasattr(engine, arg_name):
-            setattr(engine, arg_name, arg_value)
+            setattr(engine, arg_name, copy.deepcopy(arg_value))
```
dalf, line 130: deepcopy because `language_aliases` is going to be modified (should be an actual comment)

searx/engines/__init__.py
```diff
@@ -146,24 +128,27 @@ def load_engine(engine_data):
...
-
     engine_shortcuts[engine.shortcut] = engine.name
-
-    return engine
+    for category_name in engine.categories:
+        categories.setdefault(category_name, []).append(engine)
 
 
 def load_engines(engine_list):
-    global engines, engine_shortcuts
```
dalf, line 162: The references are kept, so there is no write access, so there is no need for `global`.

searx/engines/__init__.py
```diff
@@ -45,6 +28,7 @@
 engine_default_args = {'paging': False,
                        'categories': ['general'],
```
return42: just a nit: while refactoring, lets fix indentation here  

    engine_default_args = {
        'paging': False,
        ....

searx/engines/__init__.py
```diff
@@ -57,6 +41,8 @@
 
 
 def load_engine(engine_data):
+    # pylint: disable=too-many-branches, too-many-statements, protected-access
```
return42: Do you like to add a doc-string that documents 

- engine attributes from YAML `settings:engines/<engine>` argument `engine_data`
- set default attributes (from `engine_default_args`)
- check _inactive_  is True or False
- what does _"checking required variables"_ means

searx/engines/__init__.py
```diff
@@ -67,61 +53,57 @@ def load_engine(engine_data):
...
         logger.exception('Cannot load engine "{}"'.format(engine_module))
         return None
 
+    # set engine attributes from engine_data
     for param_name, param_value in engine_data.items():
-        if param_name == 'engine':
-            pass
-        elif param_name == 'categories':
+        if param_name == 'categories':
             if param_value == 'none':
```
return42: Do we support category "none" .. I haven't found any usage of a string "none" in settings.yml nor in one of the `engines/*.py`

return42:
I think setting the headers to ... ja-JP

https://github.com/searxng/searxng/blob/737f5f05d2bfdd2344578ba81e8832b9b752ee78/searx/engines/__init__.py#L131-L134

is questionable.  It comes from 9b6ffed06 [PR-searx 2602](https://github.com/searx/searx/pull/2602) / there is a detailed description I do not have fully understand right now, but since I have updated my instance, sometimes (rare) I observe japan results in my german queries .

dalf:
> I think setting the headers to ... ja-JP  is questionable.

? 
This change was made to fix searx/data/engines_languages.json updates nothing more and looking at it works: https://github.com/searx/searx/commit/d9dc3376d0b559fc23d8eb23adbbee7a24629753#diff-aea534b290d3bf70e79b052cdcf0d1b5df7cbbe92184e40bbe5b373d5699f188

return42:
> This change was made to fix searx/data/engines_languages.json updates nothing more an

Oops .. my fail .. I have sometims (rare) japan results in my german queries, I thought it could be related to this implementation, but now I see it is totaly unrelated .. forget my comment.

dalf:
Currently, `searx.engines.load_engine()` exits the program when:
* there is a missing attribute on an engine
* there is an underscore in the name of the engine

--- 
looking at 
https://github.com/searxng/searxng/blob/856729226d5822a3285483689f9f8ba5c2bafc07/searx/engines/sqlite.py#L25-L30

it can be rewritten:
```python
query_str = None  # instead of query_str = ""

def init(engine_settings):
    if not query_str.lower().startswith('select '):  # here `searx.engines.load_engine()` will copy the value from settings.yml to the variable
        raise ValueError('only SELECT query is supported')
```

`ValueError` won't stop the engine.

---

--> I suggest `searx.engines.load_engine` return `None` instead of `sys.exit(1)`.

return42 (approved):
@dalf I pushed three commits on top of your branch / except 1c13beee no functional change

If you think this commits have a value just merge them along this PR into master, otherwise drop what is not needed.

searx/engines/__init__.py
```diff
@@ -1,169 +1,220 @@
...
+        return False
+
+    return True
+
+
+def register_engine(engine):
+    engines[engine.name] = engine
     if engine.shortcut in engine_shortcuts:
         logger.error('Engine config error: ambigious shortcut: {0}'.format(engine.shortcut))
         sys.exit(1)
```
return42, line 207: Should we remove this sys.exit() also?

dalf: I'm not sure what should be done, so we can keep sys.exit(1).

But the name has the same issue with the name:
```python
    if engine.name in engines:
        logger.error('Engine config error: ambigious name: {0}'.format(engine.name))
        sys.exit(1)

    if engine.shortcut in engine_shortcuts:
        logger.error('Engine config error: ambigious shortcut: {0}'.format(engine.shortcut))
        sys.exit(1)
```

return42: Thanks! / I will rewrite my last commit to check the name also.

return42 (approved):
Thanks! .. lets merge ..

searx/engines/__init__.py
```diff
@@ -1,169 +1,225 @@
...
 
+    # load_module
     engine_module = engine_data['engine']
-
     try:
-        engine = load_module(engine_module + '.py', engine_dir)
+        engine = load_module(engine_module + '.py', ENGINE_DIR)
     except (SyntaxError, KeyboardInterrupt, SystemExit, SystemError, ImportError, RuntimeError):
         logger.exception('Fatal exception in engine "{}"'.format(engine_module))
         sys.exit(1)
```
dalf, line 99: Still one `sys.exit(1)` here otherwise it is possible to have a syntax error in the engine code (it had happened before).

PR #124, opened 2021-06-01
merge redis offline engine from searx 

return42:
## What does this PR do?

merge redis offline engine from searx / for details read along the commit messages.

### DIFF
--- a/searx/engines/redis_server.py
+++ b/searx/engines/redis_server.py
@@ -20,11 +20,10 @@ paging = False
 result_template = 'key-value.html'
 exact_match_only = True
 
-redis_client = None
+_redis_client = None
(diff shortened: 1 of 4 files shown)


searx/engines/redis_server.py
```diff
@@ -0,0 +1,75 @@
...
+port = 6379
+password = ''
+db = 0
+
+# engine specific variables
+paging = False
+result_template = 'key-value.html'
+exact_match_only = True
+
+redis_client = None
```
dalf: Should be `_redis_client` to avoid an error when the engine is loaded.

return42: Oops .. thanks! .. fixed in 39c18274

PR #128, opened 2021-06-03
[enh] flask debug mode: reload the app when searx/settings.yml changes

dalf:
## What does this PR do?

In debug mode, reload the application when searx/settings.yml changes.

Useful when trying to develop an engine or trying different settings (with / without proxies, etc...)

* `make run`
* edit `searx/settings.yml`
* check the application is reloaded

It is possible to check the change doesn't affect the producution mode:
* edit `searx/settings.yml` to define a secret key
* `. ./local/py3/bin/activate`
* `python -m searx.webapp`
* change `searx/settings.yml`: nothing happens.


return42 (approved):
Works perfect .. missed that feature since a long time :)  .. thanks!

PR #132, opened 2021-06-04
[mod] "make docs" doesn't initialize the engines

dalf:
## What does this PR do?

avoid external HTTP requests with "make docs"


return42 (approved):
Perfect / thanks!

PR #133, opened 2021-06-04
[enh] add .editorconfig

return42:
## What does this PR do?

add `.editorconfig`

EditorConfig [1] helps maintain consistent coding styles for multiple developers
working on the same project across various editors and IDEs.

The EditorConfig specification [2] support is pre installed in common IDEs [3]
and plugins for many others are available [4].

[1] https://editorconfig.org
[2] https://editorconfig-specification.readthedocs.io/
[3] https://editorconfig.org/#pre-installed
[4] https://editorconfig.org/#download

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,40 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+[*.py]
+max_line_length = 119
+
+[*.html]
+indent_size = 4
+
+[*.json]
+indent_size = 4
+insert_final_newline = ignore
+
+# Minified JavaScript files shouldn't be changed
+[**.min.js]
+indent_style = ignore
+insert_final_newline = ignore
+
+# Makefiles always use tabs for indentation
+[Makefile]
+indent_style = tab
+
+# Batch files use tabs for indentation
+[*.bat]
+indent_style = tab
+
+[docs/**.rst]
+max_line_length = 79
+
+[*.yml]
+indent_size = 2


return42:
@dalf what do you think about this PR adding a `.editorconfig`?

dalf:
In the projects I've seen or I've been working for the editor / IDE configuration is outside the source code repository.
In some projects, a documentation describes how to setup an IDE, but nothing mandatory as long the CI gives a green check (more or less).

The reason is each setup is most probably different for each individual: even if the source code is formatted in the same way.

When to use EditorConfig since it doesn't cover the whole setup ? And as soon there is another formatter, it requires to deal with two configurations.

Moreover, it requires to deal with a specific version of EditorConfig:
* About the VSCodes extension : https://github.com/editorconfig/editorconfig-vscode/issues
   * max_line_length doesn't seems to be supported: issue 272
   * high CPU load: issue 314 (not confirmed)
   * Looking for new maintainer, issue 263 (opened in 2019)
* About Emacs extension: https://github.com/editorconfig/editorconfig-emacs/issues 
   *  Editorconfig causes 'Revert buffer from file ...: issue 246
   *  Should we disable editorconfig if .dir-locals is present? issue 148  --> this is specific to this extension. An IDE configuration in another extension may behave differently.

Looking at these issues, I've difficulties to advice EditorConfig: an extension for one IDE may have issues we don't know, or simply behave slightly differently from the other extensions.

As command line formatter, why not something use something more powerful like autopep8, black for example?
 
So my take here is not to merge this PR.

return42:
> The reason is each setup is most probably different for each individual: even if the source code is formatted in the same way.
> Moreover, it requires to deal with a specific version of EditorConfig: .... Looking at these issues

Personally I do not need the `.editorconfig`, for me this file is more like a (point of) definitions.   I do not want to force to use any EditConfig tools.  But when someone ask me "what are the common indentation style" I would route him to the [.editorconfig](https://github.com/searxng/searxng/pull/133/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR5-R11)

other examples for _unspoken definitions_:

- https://github.com/searxng/searxng/pull/133/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR36-R37
- https://github.com/searxng/searxng/pull/133/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR13-R14

IMO there is no need to use IDE tools at all, but a minimal definition about how files committed to SearXNG should be formatted is helpful I think.

> As command line formatter, why not something use something more powerful like autopep8, black for example?

I changed my mind about this, in the past I was against such "automation" tools but in the meantime I realized that such tools can be very helpful. We can discuss this aspect in another PR.

dalf:
One idea:
* create a new documentation page: code style, so we can list the different tools: pylint, yamllint, shellcheck .
* in this page, add a reference to EditorConfig with the content of `.editorconfig`.
* add a disclaimer about the fact we don't use this tool.

PR #136, opened 2021-06-08
/preferences: use simple layout for the oscar theme (Issue 115)

### LINKED ISSUE #115: /preferences: use simple layout for the oscar theme
Suggestion: change how the options are displayed in the /preferences page in the oscar theme :
* General
  * Default categories
  * Search languages
  * Autocomplete
  * SafeSearch
  * Open Access DOI rewrite (plugin)
  * Open Access DOI resolver
* Engines
* User Interface
  * Interface language
  * Themes
  * Results on new tabs
  * Infinite scroll (plugin)
  * Search on category select (plugin)
  * Hotkeys (plugins)
* Privacy
  * HTTP method
  * Image proxy
  * Tracker URL remover (plugin)
* Cookies
* Query syntax (new tab, a mix between documentation and auto documentation from the current settings)
  * Documentation about the query syntax (bang, external bangs, ...)
  * Answers (list of configured answers)
  * Special engines: currency, translations (list the engines that ).
  * External bangs

dalf:
#### General tab

image

#### User Interface tab

image

#### Privacy tab

image

#### Special queries tab

image

#### Cookies tab

image

From the usability point of view, for an user who is not the admin of the instance, the plugin tab should not exist.

For plugin with `plugin_category="query"`, there are two new values 
  * `query_keywords` 
  * `query_examples` 

* `make run`
* check the preferences are saved with the selected choices.

## Related issues

Close #115

## Notes

* the engine currency, mymemory translated and dictzone should also be listed in the "Special Queries" tab.
* the "On / Off" select should be replace by a toggle (for example "Show advanced settings")

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -70,6 +70,8 @@ class PluginStore():
                 if not hasattr(plugin, plugin_attr) or not isinstance(getattr(plugin, plugin_attr), plugin_attr_type):
                     setattr(plugin, plugin_attr, plugin_attr_type())
             plugin.id = plugin.name.replace(' ', '_')
+            if not hasattr(plugin, 'preference_section'):
+                plugin.preference_section = 'general'
             self.plugins.append(plugin)
 
     def call(self, ordered_plugin_list, plugin_type, request, *args, **kwargs):
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -72,6 +72,11 @@ class PluginStore():
             plugin.id = plugin.name.replace(' ', '_')
             if not hasattr(plugin, 'preference_section'):
                 plugin.preference_section = 'general'
+            if plugin.preference_section == 'query':
+                for plugin_attr in ('query_keywords', 'query_examples'):
+                    if not hasattr(plugin, plugin_attr):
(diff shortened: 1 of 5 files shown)


return42:
I tested this PR with an external plugin:

    ./manage pyenv.cmd pip install git+https://github.com/return42/tgwf-searx-plugins

```diff
@@ -137,8 +137,8 @@ outgoing:
 # External plugin configuration, for more details see
 #   https://searxng.github.io/searxng/dev/plugins.html
 #
-# plugins:
-#   - plugin1
+plugins:
+  - only_show_green_results
 #   - plugin2
 #   - ...
```

and it works like a charm :-) 

grafik

I really like this new design, it is much more straight forward.

> an alternative: General, User Interface, Privacy, Engines, Special Queries, Cookies

This alternative  (having "User Interface" at second tab) is more intuitive I think.

> preference_examples may be not the right name (perhaps more preference_syntax).

Why not _example_, the column is also named "Examples" .. and BTW "example"  is more flexible than a name like "syntax" which is more restrictive in a future usage.

Lets make a PR from this draft!

return42 (approved):
Very cool .. thanks a lot!

FYI: I rebased your branch and delete a leading whitespace in `searx/templates/oscar/preferences.html` (f83b6427)

PR #147, opened 2021-06-15
Implement a script to document external bangs

return42:
## What does this PR do?

Implement a script to document external bangs

External Bangs are undocumented

    make docs

~or visit https://return42.github.io/searxng/admin/external_bang.html~


dalf:
it is nice because it is without javascript and there is search feature

About the table: the bang should be sorted by (rank desc, bang)
So there is no need to display the Rank column.

But:
* it is huge: 160 files.
* Moreover if the external bang documentation is merged with the current documentation, the search result can confusing.
For example: https://return42.github.io/searxng/search.html?q=network

Idea:
* a github workflow is [triggered when external_bangs.json](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) is modified on master
 * it creates and builds the .rst files
 * it pushes them on another side project searxng.github.io/externalbangs for example.
 * the documentation on searxng.github.io/searxng can reference this


So we don't have to care about the external bang documentation: it can updated automatically.

return42:
> it is nice because it is without javascript and there is search feature

The search feature is javascript :-)

> it is huge: 160 files.

Yes, IMO it is to big .. I just pushed this PR to have a discussion about and to see what are your idea:

> it pushes them on another side project searxng.github.io/externalbangs for example.

Why not moving the whole database of external bangs to this repository from where instances can load the JSON file and user can visit a HTML page with the documentation?

Should I give it a try and create such a reposetory?

dalf:
> The search feature is javascript :-)

I wrote javascript, but was thinking "static web site".

> Why not moving the whole database of external bangs to this repository

So
* this dependencies would contain:
   * data/external_bangs.json
   * external_bang.py
   * update_external_bangs.py
* it would be published on pypi
* it would be a new dependency added to `requirements.txt` on this repository.
?

👍 

For reference:
* https://duckduckgo.com/bang?
* https://mosermichael.github.io/duckduckbang/html/main.html ( https://github.com/mosermichael/duckduckbang )

return42:
FYI: I started https://github.com/return42/bangbang / but ATM it is only the boilerplate code which might be rebased often the next time .. I will invite you, when I think it is good enough to have a look.

PR #150, opened 2021-06-16
simple theme: move the source files to searx/statics/themes/simple/src

dalf:
## What does this PR do?

See commit messages.
Directory structure of searx/static/themes/simple :
* src : contains source code <--- so the source is in one location.
* css / fonts / js: contains build files (include css/images )
* img : contains image for this theme

The dependencies 
* has been updated (build
* autocomplete-js and leaflet are declared as packages.json depdenencies.

`devDependencies` of `packages.json` includes all required dependencies, so  
https://github.com/searxng/searxng/blob/127552f67bfa8d19f43cd8d77a256e525633c5a7/manage#L16
is not required.

## Why is this change important?

* Same clean up than the oscar theme
* moving the source code to another direction will be more clear ( https://github.com/searxng/searxng/issues/141 ).

## How to test this PR locally?

1. clean browser cache
1. using the simple theme:
  * check autocompletion:
    * type text, press enter
    * type text, choose an autocompletion, press enter
  * check the images rendering
  * check "show map" of the OSM / photon results.
  * check various icons are displayed correctly (generated by webfont plugin).

## Related issues

Related to https://github.com/searxng/searxng/issues/141

Issue #141, opened 2021-06-10
Static files: move source to a dedicated directory.


dalf:
Since it is a topic in itself, here a new issue :

From https://github.com/searxng/searxng/issues/137#issuecomment-857755567
> In first order we should separate source code from the ./searx/static folder. 

I would see:
* src/static
* src/searx
* src/searx/static <-- built files

But a lot of stuff expect searx source code in the `searx` directory.

So If the source folder for the static files is `static/src` or just `static`.
```
static (or static/src)
	themes
		oscar
			img           # searx/static/themes/oscar/css/images AND searx/static/themes/oscar/css/img
			js            # searx/static/themes/oscar/src/js
			less          # searx/static/themes/oscar/src/less
			package.json
			gruntfile.js
			README.rst
		simple
			js            # searx/static/themes/simple/js/searx_head AND searx/static/themes/simple/js/searx_src
			less          # searx/static/themes/simple/less
			...           # incomplete because a similar PR to https://github.com/searx/searx/pull/2346 must be done
			gruntfile.js
			package.json
```

return42:
> I would see:
> -  ...
> - src/searx/static <-- built files
> 
> But a lot of stuff expect searx source code in the searx directory.

For the first step we should leave the static build products where they are and move only the source files of the build products to another location

> So If the source folder for the static files is static/src or just static.

Hm, why not `./src/static`? .. and build products will be generated at `searx/static` (as is)

BTW: what about the files in `./searx/data`, I mean the sources that generate these files? Should we move `searx_extra/update` also to `src/data`? At the end we have something like:

```
.
├── doc
~
├── searx
│   ├── data
│   ├── static
│   │   ├── plugins
│   │   └── themes
~
└── src
    ├── data
    └── static
```

> I would see:
>
>    src/..

If there comes the day we think it is a good idea, we can also move `./searx` to `./src/searx`

dalf:
> Hm, why not  `./src/static` ?

👍 

There are files that are not compiled:
* searx/static/themes/oscar/fonts
* searx/static/themes/oscar/img
* searx/static/themes/oscar/css/images

I'm not sure what to do for these files.

return42:
Leave them where they are .. in a second step, we can have a look what files can be generated ... E.g. 

- `searx/searx/static/themes/simple/img/favicon.png` can be genrated from  `searx/static/themes/simple/img/searx_logo.svg` and
- fonts can be updated from their origin
- ..

return42:
[...]

PR #151, opened 2021-06-16
make themes: 2 node_modules directories instead of 3

dalf:
## Initial notes
* To merge after #150
* Currently look at the last two commits
* Only https://github.com/searxng/searxng/pull/150/commits/9108c90f0e9d17a8a14b502be3d33d129c93d294 is required from #150

## What does this PR do?

```make themes``` creates three node_modules directory:
* `.`
* `./searx/statics/themes/oscar`
* `./searx/statics/themes/simple`

The first one (at the root) contains dependencies described here:
https://github.com/searxng/searxng/blob/fdc367d58b3be8a54c3aaac3f2e873429c7ab040/manage#L16

This is not desirable for different reasons:
* `npm audit` doesn't work
* dependabot can't update the [dependencies](https://github.com/dependabot)
* `grunt` has to be called in a not standard way.

A second commit changes `node.clean` : it calls `npm run clean` for each themes.

simpler build phases

* `make themes`
* `make themes.ocar`
* `make node.clean`

## Related issues

Related to:
* #150 
* #137
* https://github.com/searxng/searxng/wiki/Milestones#milestone-19---data-upgrade--build-process

### CODE REVIEW

File: manage [hunk 1]
```diff
@@ -333,6 +333,7 @@ node.clean() {
        ./searx/static/themes/oscar/node_modules \
        ./searx/static/themes/simple/package-lock.json \
        ./searx/static/themes/simple/node_modules
+       ./searx/static/themes/simple/ion.less \
```
return42: This is a (backslash) typo, but the `node.clean` command is fixed in 14ba56f

return42:
@dalf I review this PR #151 instead of #150 .. I have pushed a commit on top but my review is not yet finished.

I have a dump question, why do we move to 

    searx/static/themes/simple/src/

I thought in https://github.com/searxng/searxng/issues/141#issuecomment-860091107 we have agreed to move the source to:

    src/static/themes/simple/

And to place (or leave) only the _binaries_ in  `searx/static/themes/simple`

Or did I misunderstood you (vice versa / you me)?

dalf:
`static/themes/simple/gruntfile.js` is wired to have everything in the same directory, so I've started with a small step.

utils/searx.sh
```diff
@@ -88,7 +95,8 @@ BUILD_PACKAGES_centos="\
 firefox graphviz graphviz-gd ImageMagick librsvg2-tools
 texlive-xetex-bin texlive-collection-fontsrecommended
 texlive-collection-latex dejavu-sans-fonts dejavu-serif-fonts
-dejavu-sans-mono-fonts fontforge ttfautohint"
+dejavu-sans-mono-fonts fontforge ttfautohint
+"
```
dalf: No npm for this one ?

return42: These are the yum packages for cenos-7 (we have support only for centos-7). Modern versions of centos, fedora and RHEL are using dnf package manager. Automation of the yum package installation is more complicate and centos-7 was old from the beginning when we added it.  I vote to drop these old distribution.

Comiting the newline was a fail of mine / I will remove the newline. 

**update** In this context, please have a look at my commit ae677cb6 / is this commit OK for you?

utils/searx.sh
```diff
@@ -88,7 +88,7 @@ BUILD_PACKAGES_centos="\
 firefox graphviz graphviz-gd ImageMagick librsvg2-tools
 texlive-xetex-bin texlive-collection-fontsrecommended
 texlive-collection-latex dejavu-sans-fonts dejavu-serif-fonts
-dejavu-sans-mono-fonts"
+dejavu-sans-mono-fonts fontforge ttfautohint"
```
return42: I haven't tested centos-7 (yum package manager), I don't know if this works .. see https://github.com/searxng/searxng/pull/151#discussion_r653031106

Lets ignore CentOS-7 as developer (build) platform.

**update** In this context, please have a look at my commit ae677cb6 / is this commit OK for you?

searx/static/themes/simple/package.json
```diff
@@ -20,8 +20,8 @@
     "leaflet": "^1.7.1"
   },
   "scripts": {
-    "build": "npm install && grunt",
-    "start": "grunt watch",
-    "test": "grunt"
+    "all": "npm install && grunt",
```
return42, line 23: For my understanding: is `all` the default _target_ when calling `npm install`?

return42 (approved):
I finished my review and placed ae677cb6 on top of this branch.

FY: I won't review #150 since it is already a part of this PR #151.

searx/static/themes/simple/package.json
dalf: See https://docs.npmjs.com/cli/v7/commands/npm-run-script
There is no default command, but there are some shortcuts:
* `npm start` ==> `npm run start`
* same for `restart`, `stop`, `test`.

In package.json of this PR:
* `npm run all` : download dependencies, build the theme
* `npm run build`: build the theme
* `npm run watch`: same spirit than `make docs-live` but for this theme instead of the documentation : any change to the source files recompile the theme ( https://github.com/gruntjs/grunt-contrib-watch )

Issue #137, opened 2021-06-09
Static files: build process


dalf:
Result of `make themes` is a mess to deal with: it creates conflicts on all pull requests, if the master branch modified the built files. So two or more pull requests about the themes will create git conflicts.

A solution:
* commit only the source code.
* ask the administrators to build the themes, but it requires a lot of more dependencies

Another solution can a submodule with the built files, but I expect difficulties to synchronize everything especially when there are different branches.

Here a proposal based on the principale used in the previous pull requests: 
* the last commit in a branch contains the built files
* add some scripts to manage this commit

Note: "source files" and "build files" refer to the static files only.

#### `static`

build static files (may be rename "themes")

* call searx_extra/update/update_pygments.py
* call "themes"

#### `static.clean`

It is actually "git restore", but as if the built files are not commited, so "git clean" (and in the same way than "make clean")

* git restore [built files]

#### `static.git.drop.commit`

* drop the last commit if it is made by "static.git.commit"

#### `static.git.branch.check`

Check the current branch:

* exit code 0 if there is neither commit with a built file, neither a commit with modified source file.
* exit code 1 if one of the commits of branch contains a mix of static built file and source file.
* exit code 2 if there is modified source file.
* exit code 3 if there are some uncommited built file.

#### `static.git.commit`

build & commit static files (update commit or new commit)

* call "static.git.branch.check"
* if exit code is 0 or 1: stop
* if exit code is 3:
    * call "static.clean"
* if exit code is 2 or 3:
	* call "static.git.commit.drop"
	* call "static"
	* commit built files (only built files nothing else)

#### `static.git.restore`

after "git add", make sure there is not a mix a built & source files

* check if the staged files contains a mixed of source files and built files
* if yes
	* git restore --staged [built files]
	* display a warning to the user to call "make static.git.commit"

#### `make static.git.resolve.conflict`

* it should be the only files to conflict (no other files), otherwise stop the hook.
* git checkout --theirs [built files] (keep our built files in the branch, it will dropped anyway).
* git add [built files]
* git rebase --continue

					
## Git hooks

Enabled by default, can be disabled globally with a configuration file.

				
#### `post-rebase hook`

if some built files conflict:
* call "make static.git.resolve.conflict"
* ask the user to call "make static.commit" (that will drop the built commit).

#### `pre-commit hook`

* call "static.git.restore"
* display a warning to call "make static.commit"

#### With git hooks

* Commit change in the source files (static files):
```sh
git add .  # all files including built files
[...]

PR #157, opened 2021-06-18
static build: move generated files to searx/static/themes/*/src/generated

dalf:
## What does this PR do?

* normalize.css becomes a packages.json dependency
* make themes.all update pygments*.less, these files are created "generated" directories (oscar & simple).
* grunt-webfont creates ion.less in the "generated" directory: 
   * the fonts are commited
   * `npm run webfont` build the fonts, it requires `fontforge` and `ttfautohint`.

* update of the pygment version is easier: just rebuild the themes.
* all external dependencies are in packages.json that "npm audit" can scan.

* `!scc searx` with the different themes.
* check the icons of the simple theme.

## Related issues

* https://github.com/searxng/searxng/issues/137
* https://github.com/searxng/searxng/wiki/Milestones#milestone-19---data-upgrade--build-process


searx/static/themes/simple/src/less/style.less
```diff
@@ -9,7 +9,7 @@
 @import "../../__common__/less/new_issue.less";
 @import "../../__common__/less/stats.less";
 
-@import "normalize.less";
+@import (less) "../../node_modules/normalize.css/normalize.css";
```
return42: Just for my knowledge; why do we import a CSS file interpreted as [(less)](https://lesscss.org/features/#import-atrules-feature-less). Why not [(css)](https://lesscss.org/features/#import-atrules-feature-css) or [(inline) to include the file in the output so that all CSS will be in one file.](https://lesscss.org/features/#import-atrules-feature-inline)

searx/static/themes/oscar/generated/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+*/
+!.gitignore
```
return42: Do we need these`*/generated/` folders?  Why not leave them where they are  (in `*/src/`)? And if we do not want to versioning them any longer, we can remove them from the repository and add a `.gitignore` (to the `*/src/` folders).

With my comment from above in mind, I have doubts if we should really remove these less files. For me these files are source code (even if we create them with a script).  Compare with the files in [docs/build-templates](https://github.com/searxng/searxng/tree/master/docs/build-templates)  which will be update by [docs.prebuild](
https://github.com/searxng/searxng/blob/master/manage#L149-L160) and `docs.prebuild` is called (e.g.) by [`docs.html`](https://github.com/searxng/searxng/blob/master/utils/lib.sh#L719-L722).

manage
```diff
@@ -303,6 +303,9 @@ node.env() {
 
     (   set -e
 
+        build_msg DATA "searx_extra/update/update_pygments.py"
+        pyenv.cmd python searx_extra/update/update_pygments.py
+
```
return42: Hm, are the pygments CSS files a part of the _build environment_ or should we better treat them as _build products_?
May be my mindset is not clear here, but I would say that files we generate are _build products_ and should be build in the build process and not here when we are in the _process that settles up the environment_ .. I am wrong? It would help me if you give me some arguments why it is not wrong to build files in this phase.

searx/static/themes/simple/.gitignore
```diff
@@ -1,2 +1 @@
 /node_modules
-ion.less
```
return42, line 2: See my comment about `searx/static/themes/oscar/generated/.gitignore`.  I know, we removed this less file in 7137d2893 but now I want to say, that this less file is _source code_ to me. May be I have a wrong view on this aspect / do you have an argument that could convince me?

dalf: `ion.less` is the result `grunt run build`, if this file is committed, each `Static build` commit will include `ion.less`.
So `src/less` will have a mix of source code and generated files.

searx/static/themes/simple/src/less/style.less
dalf: * `(css)` doesn't work because: _This means the import statement will be left as it is._ , it should be inlined...
* `(inline)` should work (I'm missed this option).

searx/static/themes/simple/.gitignore
return42: > `ion.less` is the result `grunt run build`, if this file is committed, each `Static build` commit will include `ion.less`.

Yes, but this file won't change / by what (line) is this file created, I had a look the gruntfile but I can't find a passage where _generated_ folder is used (in "webfont).

> So `src/less` will have a mix of source code and generated files.

See my doc-example in https://github.com/searxng/searxng/pull/157#discussion_r654796752 ..  the files in docs/build-templates will also _committed each 'Doc build'_  but I decided to add them to the repository since they are needed to understand the reST document.

dalf: > I can't find a passage where generated folder is used (in "webfont).

* https://github.com/searxng/searxng/blob/eb6832eb95f064a9450f5658f3cee52d374c6a31/searx/static/themes/simple/gruntfile.js#L119-L120
* https://github.com/searxng/searxng/blob/eb6832eb95f064a9450f5658f3cee52d374c6a31/searx/static/themes/simple/src/less/style.less#L25

What if the generated files:
* are commited in `src/generated` ?
* commited with the `Static build` commit ?

return42: Sorry that I hang at this point but before answering your question I will understand it all and I have two more questions:

* https://github.com/searxng/searxng/blob/eb6832eb95f064a9450f5658f3cee52d374c6a31/searx/static/themes/simple/gruntfile.js#L119-L120

If `destLess` is `/searx/static/themes/simple` whhy is the less file landing in `searx/static/themes/simple/generated`?

**update:** I was looking at the wrong branch (sorry), here is what I was looking for: https://github.com/searxng/searxng/pull/157/files#diff-db917d52d6475d53bdaf81c0648e8c1979e6e34b743d5182b0bf8e1ca61411ce

----

Do you know  the difference of [grunt-webfont](https://www.npmjs.com/package/grunt-webfont) and [grunt-webfonts](https://www.npmjs.com/package/grunt-webfonts) .. if I am not wrong https://github.com/sapegin/grunt-webfont is _archieved_ (last release is  years old) and https://github.com/L2jLiga/grunt-webfonts is moore up-to-date / but I also see a critical issue https://github.com/L2jLiga/grunt-webfonts/issues/20

return42: > What if the generated files:
> * are commited in `src/generated` ?
> * commited with the `Static build` commit ?

`src/generated`  is a good place but I wont recommend it in the `[static] make themes` commit.

1. the aim is to split the static commit from the src commit
2. these (less) files will only change when I change something in src or when I change the build rules (or I am wrong?)

Beside your _commit question_, my grunt-webfont(s) question resists, If you could have a look at it would be great / thanks!

dalf: grunt-webfont is (one of) the last peace that need an update.
There is also https://github.com/Frontify/grunt-webfont

return42: > grunt-webfont is (one of) the last peace that need an update.

OK, I will wait about your next commits  .. ATM I am a bit confused: 

1. https://github.com/Frontify/grunt-webfont is a fork of
2. https://github.com/jleider/grunt-webfont which is a fork of
3. https://github.com/sapegin/grunt-webfont

Only the last one is available from npm https://www.npmjs.com/package/grunt-webfont 

.. and then there is another package with a trailing 's' at the end ..

https://www.npmjs.com/package/grunt-webfonts from https://github.com/L2jLiga/grunt-webfonts

.. and none of them seems maintained regularly?

return42: Another problem I have (on ubu20044):

```
> ttf2woff2@2.0.3 install /x/share/searx/searx/static/themes/simple/node_modules/ttf2woff2
> (node-gyp rebuild > builderror.log) || (exit 0)

  File "/usr/share/nodejs/node-gyp/gyp/gyp_main.py", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3254, in <module>
    def _initialize_master_working_set():
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3237, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3266, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 584, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 901, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 787, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'gyp==0.1' distribution was not found and is required by the application
```

To decouple from OS it might be a solution  to run in pyenv / in a first try it worked for me)

```diff
diff --git a/manage b/manage
index 61376514..cbf81779 100755
--- a/manage
+++ b/manage
@@ -470,13 +470,13 @@ themes.all() {
 
 themes.oscar() {
     build_msg GRUNT "theme: oscar"
-    npm --prefix searx/static/themes/oscar run build
+    pyenv.cmd npm --prefix searx/static/themes/oscar run build
     dump_return $?
 }
 
 themes.simple() {
     build_msg GRUNT "theme: simple"
-    npm --prefix searx/static/themes/simple run build
+    pyenv.cmd npm --prefix searx/static/themes/simple run build
     dump_return $?
 }
```

But I am not very happy with this solution .. to much tools involved here: start a python process to start a grunt tporocess to start  a node implementation, which starts a python application which starts binaries tools from the OS ... pew.

return42: > grunt-webfont is (one of) the last peace that need an update.

Do we really need to automate the build of the font and css @font-face?  Do you expect any changes in the "ion" font?  If not, I think it is better to create the font once and add the font once to the repository in the "static/" folder.

return42:
BTW:  these two do no longer exist / while moving to /src can you have a eye on this file ..

https://github.com/searxng/searxng/blob/0e3a87b73acecb085acf0ed6f00c1434e32227bd/.gitattributes#L1-L2

manage
dalf: When do you suggest to update the pygments.less files ?
I guess when the pygments pypi package is updated ?
How not to forget about that ?
Because this is exactly the current situation: pygments.less files are out of sync with the pypi package.

Are git conflicts on these files a problem ? Nope. 
In case of conflicts: forget about the existing files, generated them, done.

This lead me to think that they are like the "node_modules" directories: a first step to build the theme sources, but instead of a HTTP download, it is a call to Python script.

return42: > I guess when the pygments pypi package is updated ?
> Because this is exactly the current situation: pygments.less files are out of sync with the pypi package.

OK, I understand; it is hard to automate the build of the less files when the python package is updated.

> This lead me to think that they are like the "node_modules" directories: 

IMO this is conceptual the wrong conclusion.  My questions are: 

- Do we update other /static files (which needs to be commited) when we build the target `node.env`? .. ASFAIK the answer would be 'no'.
- Do we nee node.env to create the LESS files? .. no, the less files are only related to the python packages, not related to node.

 > When do you suggest to update the pygments.less files ?

When the themes are build, when we build the targets `themes.oscar` and `themes.simple`.  Or with other words: the less files are not a part of the environment, the less files are a part of the themes.  To be precise: the less files are a prerequisite of the themes we commit into the repo somewhere under /src. 

- Since LESS files are only related to python, not to node, we need a target `pygments.less` to update these files.
-  `pygments.less` should be a prerequisite  of the `themes.*` targets

I already prepared a patch for this  .. should I push my patch on top of your branch?

**update** https://github.com/return42/searxng/commit/93e365f2

searx/static/themes/simple/.gitignore
dalf: webfont provides [two engines](https://github.com/Frontify/grunt-webfont#available-engines):
* fontforge
* node

I've tried the node engine: the built font doesn't include magnet.svg, I've tried to modified the .svg file in different ways without fixing the issue. So it is no go.

The commit cfe850aade7b2d7b050c6683a080b77cc25d1fed includes the build fonts.

`npm run webfont` can build the fonts.

return42: > `npm run webfont` can build the fonts.

Yes, but with the errors I posted in https://github.com/searxng/searxng/pull/157#discussion_r654836723 (at leats on my desktops)

IMO all the (grunt, node) webfont solutions have a lack of care .. and the mix of used technologies is scary to me.

Except the Patch in my  https://github.com/searxng/searxng/pull/157#discussion_r654836723 I do not have a better solution at hand right now.  So I would suggest to leave it as-is for the first.  On the long term I guess I would have look for a solution implemented in a small  python or shell script. 

grunt-webfont implements not much more than a call of the  https://fontforge.org/docs/scripting/python.html API .. unfortunately I have no time for this at the moment :-o

return42:
The files

- searx/static/themes/oscar/package-lock.json
- searx/static/themes/simple/package-lock.json

needs to be added to the corresponding `.gitignore` files since these files will be deleted when `make node.clean` is called.

Theses files has been added in 10ee43f _Static build_

dalf:
About packages-lock.json: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

return42:
If we add them to the repository we must not delete the files whenever a _clean_ target is called!

**But I vote to add them to the ignore list ...**

We do not have this typical development & build cycles where a lock file is needed to re-build the _product_.  We commit the product to the /static folder, so there is no need  to re-build it in the exact same way. 

With other words: we _freeze_ the product so there is no need to _freeze_ the _"how to build the product"_ also.

In our development & build cycles these lock files will be overwritten every time one calls the `node.env` target.  What means in our cycles these lock files are a part of the environment and can be deleted when the environment is cleaned up. 

This is in contrast to to a typical client development, where the build is not committed and is re-created by each team member .. for what it is needed to have the details in the lock file.

searx/static/themes/simple/.gitignore
return42: >> Do we really need to automate the build of the font and css @font-face?

> `npm run webfont` can build the fonts.

I think your solution in dca3bcca9 is the best compromise .. thanks!!

manage
return42: I think your d20f6a1f is the best solution .. there was a small typo I fixed in 1559ed38

return42:
> https://github.com/searxng/searxng/blob/0e3a87b73acecb085acf0ed6f00c1434e32227bd/.gitattributes#L1-L2

I fixed this in c1f72b71  ... I don't know if we really need this `.gitattributes` ... ATM the file is incomplete

return42 (approved):

Issue #159, opened 2021-06-20
It's possible to get Google results even when getting blocked by Google Recaptcha.


unixfox:
On the mobile UI of Google Search, the button `More results` is not affected by Google rate limiting and I can still do requests while I'm actively blocked by the original Google search. You can even fetch the results from the first page by modifying the request.

The results are given in strange raw data, but you can extract the needed HTML code, which seems to be **the same** code as the one from Google search if you fake the user agent to a desktop browser.

RAW response data: https://gist.github.com/unixfox/0cb7eebd3add42bfcbf42ea29a063b89#file-raw-txt

HTML manually parsed from the RAW response data: https://gist.github.com/unixfox/0cb7eebd3add42bfcbf42ea29a063b89#file-beautifier-html

Here is how to do requests:
- URL: `https://www.google.com/search?vet=12ahUKEwjE4O6xoajxAhWL_KQKHVCLBKoQxK8CegQIAhAG..i&ved=2ahUKEwjE4O6xoajxAhWL_KQKHVCLBKoQqq4CegQIAhAI&yv=3&q=test&prmd=vmin&ei=c0fQYITbBIv5kwXQlpLQCg&start=0&sa=N&asearch=arc&async=arc_id:srp_510,ffilt:all,ve_name:MoreResultsContainer,next_id:srp_5,use_ac:true,_id:arc-srp_510,_pms:qs,_fmt:pc`
- Headers:
	- user-agent: `Mozilla/5.0 (X11; Linux x86_64; rv:91.0) Gecko/20100101 Firefox/91.0`
	- accept: `*/*`
	- sec-fetch-site: `same-origin`
	- sec-fetch-mode: `cors`
	- sec-fetch-dest: `empty`
	- referer: `https://www.google.com/`
	- accept-encoding: `gzip, deflate`
	- accept-language: `en-US,en;q=0.9`

The query parameters are very similar to the ones from the original Google search. Not all query parameters are required, and some can be omitted. That's also the case of the headers.

@return42 Any idea how we could implement that in Searx?

unixfox:
@return42 I just updated my comment with a simpler way of extracting the data. Please have a look at it again.

return42:
Thanks!! .. I will have a look, but give me some time :-)

----

update: related [comment](https://github.com/searxng/searxng/issues/4182#issuecomment-2612347570)

unixfox:
Well I tried myself and this just need a couple of modified lines, and it works! Here is the patch:
```patch
[...]

PR #160, opened 2021-06-21
[experimental] google: use the mobile UI

dalf:
## What does this PR do?

see #159

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -34,6 +34,7 @@ categories = ['general']
 paging = True
 time_range_support = True
 safesearch = True
+use_mobile_ui = False
 supported_languages_url = 'https://www.google.com/preferences?#languages'
 
 # based on https://en.wikipedia.org/wiki/List_of_Google_domains and tests
@@ -266,6 +267,13 @@ def request(query, params):
         params, supported_languages, language_aliases, True
     )
 
+    additional_parameters = {}
+    if use_mobile_ui:
+        additional_parameters = {
+            'asearch': "arc",
+            'async': 'arc_id:srp_510,ffilt:all,ve_name:MoreResultsContainer,next_id:srp_5,use_ac:true,_id:arc-srp_510,_pms:qs,_fmt:pc'  # pylint: disable=line-too-long
+        }
+
     # https://www.google.de/search?q=corona&hl=de&lr=lang_de&start=0&tbs=qdr%3Ad&safe=medium
     query_url = 'https://' + lang_info['subdomain'] + '/search' + "?" + urlencode({
         'q': query,
@@ -273,6 +281,7 @@ def request(query, params):
         'ie': "utf8",
         'oe': "utf8",
         'start': offset,
+        **additional_parameters,
     })
 
     if params['time_range'] in time_range_dict:
@@ -282,9 +291,12 @@ def request(query, params):
     params['url'] = query_url
 
     params['headers'].update(lang_info['headers'])
-    params['headers']['Accept'] = (
-        'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8'
-    )
+    if use_mobile_ui:
+        params['headers']['Accept'] = '*/*'
+    else:
+        params['headers']['Accept'] = (
+            'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8'
+        )
 
     return params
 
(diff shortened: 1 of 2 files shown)


unixfox:
One minor request change is to stop searching for number_of_results when use_mobile_ui is enabled because the div with the ID `result-stats` doesn't exist.

dalf:
https://github.com/searxng/searxng/blob/95e634ad1394b6fe6ba1482229adfe68a655edee/searx/engines/google.py#L323-L324

?

unixfox:
> https://github.com/searxng/searxng/blob/95e634ad1394b6fe6ba1482229adfe68a655edee/searx/engines/google.py#L323-L324
> 
> ?

Oops never mind.

return42:
For me this patch seems to work.  I can't validate if it also works when google blocks me, because google does not block me :-)

Does anyone have an idea how I can enforce to be blocked by google?

In the following I will leave some comments from my _reverse engineering_  ... (similar to https://github.com/searxng/searxng/issues/159#issue-925654312)

The content type is plain text:

grafik

The response text is idiosyncratic ... starting with `)]}'`

grafik

in the further course one finds fragments of HTML, one JS `<script>` tag  and one `<style>` CSS tag.

What me wonders: the response is not (valid) XML but it seems that lxml's parser (`lxml.html.fromstring(resp.text)`)could handle such a idiosyncratic data structure (read ["Parsing HTML fragments" and "Really broken pages"](https://lxml.de/lxmlhtml.html#parsing-html)

https://github.com/searxng/searxng/blob/0e3a87b73acecb085acf0ed6f00c1434e32227bd/searx/engines/google.py#L299-L303
 
I tested with `!google foo` and got a lot of answers, much more than from the traditional google and most of them are questionable (in the following screenshot I delimited the answers with `//`) 

grafik

------

Randomly I checked _paging_ and different languages (which means different domains `google.com`, `google.de`, `google.es` ...)

Over all I would say that this solution seems to work and that the lxml parser could handle this smashed plain text file (response).

I vote to give it a try in a production environment .. lets merge.

unixfox:
> For me this patch seems to work. I can't validate if it also works when google blocks me, because google does not block me :-)
> 
> Does anyone have an idea how I can enforce to be blocked by google?
> 

A simple stress tool like [hey](https://github.com/rakyll/hey) will get you blocked after just 5 minutes. Here is the command that I used:
```
hey -c 100 -n 20000 https://www.google.com/search?q=test
```

> The response text is idiosyncratic ... starting with `)]}'`
> 
> grafik

According to a friend, it seems like this is protobuf data serialized to JavaScript, more about that here: https://github.com/marin-m/pbtk/issues/15#issuecomment-733184394

> What me wonders: the response is not (valid) XML but it seems that lxml's parser (`lxml.html.fromstring(resp.text)`)could handle such a idiosyncratic data structure (read ["Parsing HTML fragments" and "Really broken pages"](https://lxml.de/lxmlhtml.html#parsing-html)

We could extract the correct HTML code, that's what I wanted to do at first, but lxml parsed the content without any issues, so I gave up the idea.

> I tested with `!google foo` and got a lot of answers, much more than from the traditional google and most of them are questionable (in the following screenshot I delimited the answers with `//`)

It seems like I'm getting more Google answers, maybe it is not parsing the correct one... I don't know.

> I vote to give it a try in a production environment .. lets merge.

By the way, it's already in test on https://searx.be!

But before merging it, do you think we should reduce the amount of parameters given in the `async` query parameter? I don't really understand the purpose of all the values in the query parameter, maybe some parameters could be used by Google in order to track Searx? So some parameters aren't really needed to pass and we could remove them?

dalf:
Note: if settings.yml doesn't include `use_mobile_ui: true`, the engine doesn't use the mobile UI.

unixfox:
> Note: if settings.yml doesn't include `use_mobile_ui: true`, the engine doesn't use the mobile UI.

use_mobile_ui seems to be to true in the default settings.yml though: https://github.com/searxng/searxng/blob/master/searx/settings.yml#L586

return42:
> But before merging it, do you think we should reduce the amount of parameters given in the async query parameter? I

Here is what I tried and what works for me:

```python
    additional_parameters = {}
    if use_mobile_ui:
        additional_parameters = {
            'async': 'use_ac:true,_fmt:pc'
        }
```

unixfox:
Should we also document this enhancement in the settings.yml?

return42:
> Should we also document this enhancement 

If we have consolidated the development, I can add some doc-strings which could be shown here https://searxng.github.io/searxng/src/index.html

return42:
Oops I see this PR is merged .. I will implement one more PR which reduces the parameters and add the documentation.

PR #165, opened 2021-06-21
improve & document google engine

return42:
## What does this PR do?

- google engine: reduce mobile UI parameters to what is needed
- google answers: normalize space of the answers.
- add documentation from the sources of the google engines / here is a preview https://return42.github.io/searxng/src/searx.engines.google.html

About details, read commit messages.

Start instance `make run` and do some google queries.

## Related issues

- #159

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -311,9 +311,10 @@ def response(resp):
     dom = html.fromstring(resp.text)
 
     # results --> answer
-    answer = eval_xpath(dom, '//div[contains(@class, "LGOjhe")]//text()')
-    if answer:
-        results.append({'answer': ' '.join(answer)})
+    answer_list = eval_xpath(dom, '//div[contains(@class, "LGOjhe")]')
+    if answer_list:
+        answer_list = [_.xpath("normalize-space()") for _ in answer_list]
+        results.append({'answer': ' '.join(answer_list)})
     else:
         logger.debug("did not find 'answer'")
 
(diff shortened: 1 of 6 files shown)


searx/engines/google.py
```diff
@@ -312,9 +328,10 @@ def response(resp):
     dom = html.fromstring(resp.text)
 
     # results --> answer
-    answer = eval_xpath(dom, '//div[contains(@class, "LGOjhe")]//text()')
-    if answer:
-        results.append({'answer': ' '.join(answer)})
+    answer_list = eval_xpath(dom, '//div[contains(@class, "LGOjhe")]')
+    if answer_list:
+        answer_list = [_.xpath("normalize-space()") for _ in answer_list]
```
dalf, line 333: is `normalize-space()`  the same thing that `.strip()` in Python ?

return42: No not really the same.  Hard to describe with my bad english, I will try it:

The old selector was `'//div[contains(@class, "LGOjhe")]//text()'` which returns text of each node and sub-node and sub-sub-node .... with you get a lot of text nodes, some of them are empty some of the contain some text. If you concatenate these text nodes you will get more spaces as you want and you concatenate things which are in different `LGOjhe` tags,

The new selector is `'//div[contains(@class, "LGOjhe")]'` and returns a list of `LGOjhe` tags.  When you call `normalize-space()` on each of these nodes you get the _normalized text_ for each node.

[The normalize-space function strips leading and trailing white-space from a string, replaces sequences of whitespace characters by a single space, and returns the resulting string.](https://docs.w3cub.com/xslt_xpath/xpath/functions/normalize-space)

Hard to explain, I guess my explanation was not good.

Issue #162, opened 2021-06-21
Remove themes


dalf:
Currently there are two themes: oscar and simple.

It is possible to choose at runtime which themes to use.

This issue suggests:
* to remove this choice to keep only one theme.
* keep the settings `static_path` and `templates_path`, so it is possible to another runtime at setup time, not runtime.

Why:
* the plugins and themes conflict each other:
   * looking at the source code of plugins https://github.com/searxng/searxng/tree/master/searx/static/plugins/js
   * everything is wired for the oscar: the plugins except jquery + the css classes from oscar
* it duplicates the maintenance: for example https://github.com/searxng/searxng/pull/136 despite its title, it is not exactly a copy of the simple layout to oscar (the "Special Queries" is new and should be back ported).
* we can focus one theme and make it right.

[EDIT] Remove the themes but the styles: light, dark (and we add "auto" using the browser settings).

return42:
What theme would you prefere to drop?

OT: the word _theme_ is IMO a bit misleading, if we have only one theme we should better choose the word _client_ I think.

dalf:
I would the simple theme without convection,  I think it can't be done now.

oscar
  * pros:
     * supports ARIA recommendations
     * RTL languages.
     * supports infinite scroll
     * image details
  * cons: 
     * bloated.
     * use bootstrap 3: deprecated.
     * the keyboard navigation UI is broken visually.

simple:
  * pros:
     * light
     * better mobile UI
     * keyboard 
  * cons:
     * see issue 1025 of searx.
     * the pros of oscar
  * note: see https://github.com/paulgoio/searx

Bootstrap 5 is lighter and without jquery, but it requires to migrate to the version 4 then to the version 5... wouldn't be easier to just upgrade the simple theme?

return42:
I would also vote to drop oscar since it is really a mess.

What might the users say? @unifox do you have a feeling about?

_From developers scope reducing the themes to one and continue with the "simple"-theme would be a great relief._

B0pol:
I prefer oscar theme

return42:
>  I prefer oscar theme

Since the title of this PR "to remove" I have to ask; you prefer to stay with the oscar the and are not interested in other themes .. right?

If so .. what drawbacks do you see with the 'simple' theme?

B0pol:
yes

> what drawbacks do you see with the 'simple' theme?

functional:
- no dark theme
- no image details

personal opinion:
- looks uglier, especially on desktop.
- not centered

[...]

PR #168, opened 2021-06-22
Lint the less files of the simple theme

return42:
## What does this PR do?

Lint the less files of the simple theme

About details read along the commit messages.

This PR is limited to the simple theme (I hope we can drop the oscar theme, see #162)

### DIFF
--- a/searx/static/themes/simple/package.json
+++ b/searx/static/themes/simple/package.json
@@ -9,11 +9,14 @@
     "grunt-contrib-less": "~3.0.0",
     "grunt-contrib-uglify": "~5.0.1",
     "grunt-contrib-watch": "~1.1.0",
+    "grunt-stylelint": "^0.16.0",
     "grunt-webfont": "^1.7.2",
     "ionicons-npm": "^2.0.1",
     "jslint": "^0.12.1",
     "less": "^4.1.1",
-    "less-plugin-clean-css": "^1.5.1"
+    "less-plugin-clean-css": "^1.5.1",
+    "stylelint": "^13.13.1",
+    "stylelint-config-standard": "^22.0.0"
   },
   "dependencies": {
     "autocomplete-js": "2.7.1",
@@ -25,6 +28,8 @@
     "build": "grunt",
     "watch": "grunt watch",
     "webfont": "grunt webfont",
-    "clean": "rm -Rf node_modules package-lock.json ion.less"
+    "clean": "rm -Rf node_modules package-lock.json ion.less",
+    "stylelint": "grunt stylelint",
+    "stylelint-fix": "grunt stylelint --fix"
   }
 }
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -4,6 +4,8 @@
 * To convert "style.less" to "style.css" run: $make styles
 */
 
+// stylelint-disable no-descending-specificity
+
 @stacked-bar-chart: rgb(0, 0, 0);
 
 @import "../../__common__/less/new_issue.less";
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -16,13 +16,6 @@
 @import "toolkit.less";
 @import "autocomplete.less";
 
-// ion-icon
-.ion-icon-big {
-  .ion-icon;
-
-  font-size: 149%;
-}
-
 // for index.html template
 @import "index.less";
 
@@ -31,6 +24,13 @@
 
 // Search-Field
 @import "search.less";
+
+// ion-icon
+.ion-icon-big {
+  .ion-icon;
+
+  font-size: 149%;
+}
 
 // Main LESS-Code
 html {
@@ -52,13 +52,6 @@ main {
 main {
   width: 100%;
   // position: absolute;
-}
-
-footer {
-  // position:absolute;
-  bottom: 0;
-  width: 100%;
-  height: 50px;
 }
 
 #main_preferences,
@@ -711,7 +704,7 @@ article.result-images[data-vim-selected]::before {
 
   #linkto_preferences {
     display: none;
-    postion: fixed !important;
+    position: fixed !important;
     top: 100px;
     right: 0;
   }
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -8,24 +8,18 @@
 
 @import "../../__common__/less/new_issue.less";
 @import "../../__common__/less/stats.less";
-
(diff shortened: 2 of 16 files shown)


manage
```diff
@@ -488,8 +488,12 @@ themes.oscar() {
 }
 
 themes.simple() {
-    build_msg GRUNT "theme: simple"
-    npm --prefix searx/static/themes/simple run build
+    (   set -e
+        build_msg LINT "theme: simple"
+        npm --prefix searx/static/themes/simple run stylelint
+        build_msg GRUNT "theme: simple"
+        npm --prefix searx/static/themes/simple run build
```
dalf: `grunt build` already calls `jshint`: 
https://github.com/searxng/searxng/blob/056550d523ddf4e06821957e13623fc322cf4675/searx/static/themes/simple/gruntfile.js#L197

so I would add the call `stylelint` from `gruntfile.js` using [grunt-stylelint](https://www.npmjs.com/package/grunt-stylelint):
```js
grunt.registerTask('default', ['jshint', 'stylelint', 'copy', 'concat', 'uglify', 'less:development', 'less:production']); 
```

searx/static/themes/simple/package.json
```diff
@@ -25,6 +27,7 @@
     "build": "grunt",
     "watch": "grunt watch",
     "webfont": "grunt webfont",
-    "clean": "rm -Rf node_modules package-lock.json ion.less"
+    "clean": "rm -Rf node_modules package-lock.json ion.less",
+    "stylelint": "stylelint -f unix 'src/less/**/*.less'"
```
dalf: Based on the comment on `manage` :
```js
  "stylelint": "grunt stylelint",
  "stylelint-fix": "grunt --fix stylelint"
```


https://www.npmjs.com/package/grunt-stylelint#fix

searx/static/themes/simple/src/less/style.less
```diff
@@ -704,7 +704,7 @@ article.result-images[data-vim-selected]::before {
 
   #linkto_preferences {
     display: none;
-    postion: fixed !important;
+    postion: fixed !important; // stylelint-disable property-no-unknown
```
dalf: There is a typo `postion` --> `position` !

dalf (commented):
Expect the 3 comments LGTM.

I've compiled the theme and look at the result: everything is okay.

manage
return42: I have seen that there is a grunt-stylelint, but it uses an old ["stylelint": "13.8.0"](https://github.com/wikimedia/grunt-stylelint/commit/5c07ec3767145aeafae043db8d5d3f95f055aec3) while I decided to use the latest versions of stylelint & stylelint-config-standard in our [package.json](https://github.com/searxng/searxng/pull/168/commits/e22dbdda53445008c6b089365c3ba152413a76bc#diff-0f087f6a54013c2b6e21074952c187930cd193591ec780748f18423271435f90).

Do you see a disadvantage in this? ..

Foregive me if I am naive and not used to work with grunt: My impression is that the grunt packages are partly outdated and not updated regularly, so I asking myself why not directly use the npm packages (npm run configuration).  What I mean: we have these shell scripts to build our product, do we realy need additionally grunt packages which wrap old libraries? .. is my thought wrong, what do you think?

searx/static/themes/simple/src/less/style.less
return42: LOL :-) .. here you can see that I am a dyslexic .. I rebased and dropped this commit 50495757, the fix is now in 5b7e7aa.

manage
dalf: The dependency is ["stylelint": "^13.8.0"](https://github.com/wikimedia/grunt-stylelint/blob/9e56310596ec39c587f5ef20361a73dd0be9fb40/package.json#L26)

See https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

`npm install grunt-stylelint stylelint --save-dev` (from https://www.npmjs.com/package/grunt-stylelint), installs the last version:
```js
{
    "grunt-stylelint": "^0.16.0",
    "stylelint": "^13.13.1",
    "stylelint-config-standard": "^22.0.0"
}
```

return42: > "stylelint": "^13.8.0" .. installs the last version

Argh, you are right .. I have overseen the `^` and thought they are using old versions .. sorry, I am unexercised with npm .. forget my comment, I will patch to use grunt.

searx/static/themes/simple/src/less/style.less
dalf: Actually I compiled the theme, start the devtools, saw the property was not recognized, and then I realized the problem was not the value but the property name...

manage
return42: Sorry for another dump question, just for my knowledge of npm;

> npm install grunt-stylelint stylelint --save-dev (from https://www.npmjs.com/package/grunt-stylelint), installs the last version:

What version of stylelint would be used, when I install only::

    npm install grunt-stylelint

When I look at the `package-lock.json` file I can't find an entry for stylelint / I also looked at the node_modules folder, there is no stylelint package.  This is what me wonders, I thought  `package-lock.json` describes the full dependencies.

return42: Now I added a stylelint task::

```js
    stylelint: {
      options: {
        formatter: 'unix',
      },
      src: [
        'src/less/**/*.less',
      ]
    },
```

searx/static/themes/simple/package.json
return42: > `"stylelint-fix": "grunt --fix stylelint"`

It seems `"stylelint-fix": "grunt stylelint --fix"`did also work .. I haven't found a documentation which describes what arguments from the command line are passed through .. what I asking myself is what about common arguments like `--help` or `-v`, these are grunt options and stylint has also the same options.  I looked at:
 
- https://gruntjs.com/using-the-cli and
- https://gruntjs.com/api/grunt.option

return42:
@dalf: I rebased, to get the line back ..

-  dropped this commit 5049575, the fix is now in https://github.com/searxng/searxng/commit/0b7d03c6#diff-88d28efcc4bd7593a1a47d69dfcb5f50c147b93fceb99c58d7ee3b731e01a1f9R707
- a3f58a3 is new since your last review
- dfc5c3bc has been modified to fit your request about [grunt](https://github.com/searxng/searxng/pull/168#discussion_r656784723)

The following patches has not been changed (except the SHA commit caused by the rebase)

- [fix] simple theme: use stylint to fix common lint errors
- [stylelint] simple theme: fix some errors reported by stylelint
- [stylelint] disable role 'no-descending-specificity'

> I've compiled the theme and look at the result: everything is okay.

I did not committed the "Static build" .. I will test this with your  #174 :-)

dalf (approved):

PR #169, opened 2021-06-22
[fix] tidy up ignore lists .gitignore & .dockerignore

return42:
## What does this PR do?

tidy up ignore lists `.gitignore` & `.dockerignore`

@dalf: could you have a look at the `.dockerignore` list please / we have the new ignore files at:

- https://github.com/searxng/searxng/blob/master/searx/static/themes/oscar/.gitignore and
- https://github.com/searxng/searxng/blob/master/searx/static/themes/simple/.gitignore

Is there anything we have to consider?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,9 +4,6 @@
 */*/*/*~
 */*/*/*/*~
 
-#
-local/
-
 # Git
 .git
 .gitignore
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -48,9 +32,13 @@ node_modules/
 
 .tx/
 
-#
+# to sync with .gitignore
+geckodriver.log
+.coverage
+coverage/
+cache/
```
dalf, line 39: I can't find a reference to this directory

return42: geckodriver.log is created by the test.robot, coverage file & folder are created by target test.coverage and folder cache is used by ...

https://github.com/searxng/searxng/blob/98d1661f723289504cf814bca23a34c04b94c789/utils/lib.sh#L29-L31

dalf (approved):

PR #171, opened 2021-06-22
Build static cleanup

dalf:
## What does this PR do?

See commit messages.


return42 (approved):
LGTM .. what do you think, should we `-diff` some files from the data folder?

- https://github.com/searxng/searxng/blob/master/searx/data/ahmia_blacklist.txt
- https://github.com/searxng/searxng/blob/master/searx/data/currencies.json
- ... ?

dalf:
currencies.json is fetched from wikidata which may contains vandalized data.
a quick overview of what have changed is interesting to check that. 

about ahmia_blacklist.txt: it is on purpose not readable, but at least we can check there are only additions.

and most importantly we can check the update*.py scripts still  fetch the data.

PR #173, opened 2021-06-23
[mod] utils/lxc.sh: detect conflict of docker & LXC in the iptables

return42:
## What does this PR do?

Docker is blocking network of existing LXC containers / there is a conflict in
the iptables setup of Docker & LXC.  With this patch:

- utils/lxc.sh checks internet connectivity (instead of silently hang)
- Chapter "Internet Connectivity & Docker" describes the problem and made a
  suggestion for a solution a solution

### DIFF
--- a/docs/dev/lxcdev.rst
+++ b/docs/dev/lxcdev.rst
@@ -52,7 +52,12 @@ software:
 and the script :ref:`lxc.sh`, with we can scale our installation, maintenance or
 even development tasks over a stack of isolated containers / what we call the:
 
-  **searxNG LXC suite**
+  **SearxNG LXC suite**
+
+.. hint::
+
+   If you see any problems with the internet connectivity of your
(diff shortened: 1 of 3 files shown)


dalf:
ping @unixfox

unixfox:
I don't really use LXD that much, so I can't really review sorry.

return42:
> I don't really use LXD that much, so I can't really review sorry.

I checked LXD .. it is more about side effects in docker. I do not expect side effects in docker, but it might be good to check this setting of iptables on your HOST:

      sudo -H iptables -F FORWARD
      sudo -H iptables -P FORWARD ACCEPT

What does ..

     sudo -H iptables-save | grep FORWARD

print .. has the change any negative side effect to docker containers.  Just check if your docker container works as usual / there is no need to reebot the HOST, if you reboot the HOST the settings from above are gone.

PR #174, opened 2021-06-23
[mod] add make targets to manage the build files in the /searx/static directory

dalf:
## What does this PR do?

The idea is to avoid conflict when there are differents branches with changes are made on the static files.
A solution is to ask the administrators to build the files from the sources, but it requires to install the npm dependencies.
So the solution in this commit keep the sources and the build files in the same git repository.

In one branch, the modification of the source (*.less, *.js) are commited without the built files.
The built files are commited in a uniq commit, with a commit message "Static build"
In case of merge or rebase, this commit can be dropped.

New make targets:
* `static.build.commit.drop`:
    * drop the last "Static build".
    * The command checks that there are only build files in the commit.
* `static.build.commit` :
    * call "make static.build.commit.drop"
    * call "make themes.all"
    * commit the files
* `static.git.restore.staged` :
    * `git restore --staged <build files>`
* `static.git.restore` :
    * `git restore <build files>`

#### Use cases

* Typical
  * `git add .` # the build files are included
  * `make static.git.restore` # restore them.
  * `git commit`  # commit only the source
  * `make static.build.commit`  # build & commit the built files from the sources

* Rebase with change of the static files in the current branch and in the master branch:
  * `make static.build.commit.drop`
  * `git pull --rebase origin/master`
  * `make static.build.commit`

* change some *.less files
* call `make themes`
* call `static.git.restore`
* check only the build files has been restored.
* call `make static.build.commit`
* check the build commit  is here
* call `make static.build.commit.drop`
* check the build commit has been removed
* add a dummy commit
* call `make static.build.commit.drop`
* check the last has not been dropped

## Related issues

* https://github.com/searxng/searxng/issues/137
* The last step for https://github.com/searxng/searxng/wiki/Milestones#milestone-19---data-upgrade--build-process


utils/manage_static.sh
```diff
@@ -0,0 +1,151 @@
...
+    searx/static/themes/oscar/src/generated/pygments-logicodev.less
+    searx/static/themes/oscar/src/generated/pygments-pointhi.less
+    searx/static/themes/simple/css
+    searx/static/themes/simple/js
+    searx/static/themes/simple/src/generated/pygments.less
+)
+
+CURRENT_BRANCH="$(git branch --show-current)"
+STAGED_FILES=$(git diff --name-only --cached)
+
```
return42: I see drawbacks with this globlas:

1. git is called when the script is loaded
2. the values are initialized when the script is loaded, but the script itself changes the origin / I mean after a static.git.restore the values in this globals are old.

The last one is not a problem in this small script, but when we consider to integrate this script as a library there might be use cases where we e.g. call static.build.commit.drop twice in the same shell process.

I will commit some patches on top of yours ... coming soon.

utils/manage_static.sh
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env bash
+# -*- coding: utf-8; mode: sh indent-tabs-mode: nil -*-
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+BUILD_COMMIT_MESSAGE="Static build"
```
return42: Is it OK when we use 

    [build] /static

may be later we have also a 

    [build] /data

dalf: I'm not sure to understand. 
Whatever the name, we just need one.

return42: > Whatever the name, we just need one.

OK, then lets use `[build] /static`

> I'm not sure to understand.

This PR is for _manage_static_ files we generate in the static folder ... may be later, we add a _manage_data_ script which then can use `[build] /data` .. so we have commit messages with canonical names.

utils/manage_static.sh
```diff
@@ -0,0 +1,151 @@
...
+static.git.restore() {
+    static.git.restore.staged
+
+    NOT_COMMITED_FILES="$(git diff --name-only)"
+    for i in ${BUILT_PATHS[*]}; do
+        NOT_COMMITED_FILES_FOR_I=$(echo "${NOT_COMMITED_FILES}" | grep "^${i}")
+        if [ -n "${NOT_COMMITED_FILES_FOR_I}" ]; then
+            git restore ${NOT_COMMITED_FILES_FOR_I}
+        fi
+    done
```
return42: From the comment I read:

    # "git restore" of the built files.

I'm not sure I understand this fully, but isn't a `git restore ${BUILT_PATHS[*]}` not the same?

If so, don't fix it while I'm in the review.

return42:
I think you mixed the usage of `[*]` with `[@]` when iterating over an array, here is a good description what the difference is: [How to iterate over a Bash Array? (loop)](https://www.shell-tips.com/bash/arrays/#how-to-iterate-over-a-bash-array-loop) 

I will fix it / don't change while I'm in review.

return42:
@dalf FYI:: I am just review this PR .. sorry for the delay but I have to do some work in my _real live_ job in parallel.

return42 (approved):
I rebased and placed several commits on top / read along commit messages. To summarize: 

The commands:

- static.build.restore
- static.build.commit
- static.build.drop

are now available from `./manage`.  The implementation is located at `utils/lib_static.sh` and imported by `./manage`.

To avoid name conflicts some names has been *normalized*.

The script has been added to the `test.shell` target to lint.

A documentation has been added.

The banner from the grunt tasks has been removed, because they are responsible for unwanted diffs.

Tha last commit adds the build product made by `static.build.commit`.

PR #179, opened 2021-06-26
[enh] add Brave-Search engine (XPath)

return42:
## What does this PR do?

Add https://search.brave.com which was requested by @kaonashi696 [1].
This patch was suggested by @eagle-dogtooth [2].

[1] https://github.com/searx/searx/issues/2857
[2] https://github.com/searx/searx/issues/2857#issuecomment-867199241

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1503,6 +1503,16 @@ engines:
       require_api_key: false
       results: html
 
+  - name: brave
+    shortcut: brave
+    engine: xpath
+    paging: false
+    search_url: https://search.brave.com/search?q={query}
+    url_xpath: //div[@class="snippet fdb"]/a/@href
+    title_xpath: //span[@class="snippet-title"]
+    content_xpath: //div[@class="snippet-content"]
+    categories: general
+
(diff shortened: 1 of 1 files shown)


dalf:
Something like
```yaml
    results_xpath: //div[@class="snippet"]
    url_xpath: //div[@class="snippet fdb"]/a/@href
    title_xpath: //span[@class="snippet-title"]
    content_xpath: //div[@class="snippet-content"]
    categories: general
```
should be better to keep title, description and url "synchronized".

nonetrix:
Hmm seems really easy to add engines might look into adding one🤔 

Also, if an engine has a Rest API in XML or JSON could that be used instead?

return42:
Brave is a XPath engine, [The XPath engine is a generic engine with which it is possible to configure engines in the settings.](https://searxng.github.io/searxng/admin/engines/searx.engines.xpath.html), more elaborate engines are implemented in a [python module](https://github.com/searxng/searxng/tree/master/searx/engines) / see [Demo Online Engine](https://searxng.github.io/searxng/src/searx.engines.demo_online.html)

Issue #166, opened 2021-06-21
UnicodeEncodeError for preferences url


rany2:
**Version of Searx**
```
$ git log -1
commit f3e56836d6481f9807ac232b71c584c0535de783 (HEAD -> master, origin/master, origin/HEAD)
Merge: f4da4baf 0ef6aa51
Author: Alexandre Flament <alex@al-f.net>
Date:   Mon Jun 21 20:12:07 2021 +0200

    Merge pull request #165 from return42/patch-google
    
    improve & document google engine
``` 

**How did you install Searx?**
Executing the python file without docker.

**What happened?**
I tried to use the preferences url but it didn't work.

**How To Reproduce**
[...]

PR #180, opened 2021-06-26
[fix] decoding of saved preferences in the URL

return42:
## What does this PR do?

[fix] decoding of saved preferences in the URL

To compress saved preferences in the URL was introduced in 5f758b2d3 and
slightly fixed in 8f4401462.  But the main fail was not fixed; The decompress
function returns a binary string and this binary should first be decoded to a
string before it is passed to urllib.parse_qs.

BTW: revert the hot-fix from 5973491

Related-to: https://github.com/searxng/searxng/issues/166


dalf (approved):
Thank you!

PR #204, opened 2021-07-13
[fix] Qwant engine - implement API v3 and add engine "qwant videos"

return42:
## What does this PR do?

Implement Qwant API v3, the old API (https://api.qwant.com/api) seems no longer work.

I guess the offset (aka paging) needs more tests.


return42:
language support needs to be reviewed ... WIP ...

searx/engines/qwant.py
```diff
@@ -25,98 +49,142 @@
...
+        language = match_language(
+            params['language'],
+            # pylint: disable=undefined-variable
+            supported_languages,
+            language_aliases,
+        )
         params['url'] += '&locale=' + language.replace('-', '_').lower()
 
-    params['headers']['User-Agent'] = 'Mozilla/5.0 (X11; Linux x86_64; rv:69.0) Gecko/20100101 Firefox/69.0'
+    params['headers']['User-Agent'] = gen_useragent()
```
dalf: This is automatically set by https://github.com/searxng/searxng/blob/ff180cf2dbab35d93b013b68afc19b795e7b9b71/searx/search/processors/online.py#L61

return42: Thanks for the hint, I fixed it and rebased. I also added the missing data.language for the new engine "qwant videos".

In the second patch I _enabled_ the Qwant engines .. is this OK .. meanings?

PR is now ready to review

PR #213, opened 2021-07-18
[mod] drop searx.brand namespace

return42:
## What does this PR do?

Remove `searx.brand` namespace. The `searx.brand` namespace was an interim solution which has been added in commit
9e53470b4 (see commit message there).

For details read along the commit messages of this [PR](https://github.com/searxng/searxng/pull/213/commits)


return42:
@dalf can you please confirm that docker.build has now issues.

searx/templates/simple/base.html
```diff
@@ -51,10 +51,10 @@
   <footer>
     <p>
     {{ _('Powered by') }} <a href="{{ url_for('about') }}">searxng</a> - {{ searx_version }} — {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
-        <a href="{{ brand.GIT_URL }}">{{ _('Source code') }}</a> |
-        <a href="{{ brand.ISSUE_URL }}">{{ _('Issue tracker') }}</a> |
-        <a href="{{ brand.PUBLIC_INSTANCES }}">{{ _('Public instances') }}</a>{% if brand.CONTACT_URL %} |
-        <a href="{{ brand.CONTACT_URL }}">{{ _('Contact instance maintainer') }}</a>{% endif %}
+        <a href="{{ get_setting('brand.git_url', '') }}">{{ _('Source code') }}</a> |
```
dalf: The default parameters `''` are already set in settings_default.py

utils/build_env.py
```diff
@@ -16,18 +16,26 @@
...
-    ('TWITTER_URL'            , brand.TWITTER_URL),
+    ('SEARX_URL'              , _env('server.base_url','')),
+    ('GIT_URL'                , _env('brand.git_url', '')),
+    ('GIT_BRANCH'             , _env('brand.git_branch', '')),
+    ('ISSUE_URL'              , _env('brand.issue_url', '')),
+    ('DOCS_URL'               , _env('brand.docs_url', '')),
+    ('PUBLIC_INSTANCES'       , _env('brand.public_instances', '')),
+    ('CONTACT_URL'            , _env('general.contact_url', '')),
+    ('WIKI_URL'               , _env('brand.wiki_url', '')),
+    ('TWITTER_URL'            , _env('brand.twitter_url', '')),
```
dalf: This code can be removed

searx/__init__.py
```diff
@@ -32,52 +32,24 @@
...
-
-    @property
-    def TWITTER_URL(self):
-        return self.get_val('brand', 'twitter_url')
-
-
-brand = _brand_namespace()
+_unset = object()
+
+def get_setting(name, default=_unset):
```
dalf, line 37: Is this function useful?
`settings_default.py`:
* makes sure the value exist.
* sets a default if the value is not set in `settings.yml`.

For example, `get_settings('brand.git_url')` can be replaced by `settings['brand']['git_url']` without issue.

dalf (changes requested):
Thank you.

See my comment about the `get_settings` function.

utils/build_env.py
return42: Why do you think this can be removed?  .. the bash scripts do need these environments in the `utils/brand.env`.

searx/__init__.py
return42: This function exists for convenience. Your example is good for values having a default value (in the SCHEMA or in the `settings.yml` file). But what about settings which do not exists by default, those will raise an `KeyError`. With this function you have a default option.

Further I think addressing an option by a doted name is more readable:

    by doted name        : 'brand.git_url'
    by dict's __getitem__: ['brand']['git_url'] 

My be this is only my very personal view on this topic, but I think it is much more comfortable to have a function in templates (or in the `docs/conf.py` or `utils/build_env.py` or ..) which returns a default whenever an option is unset instead of juggling in the template to avoid an `KeyError` exception.

searx/templates/simple/base.html
return42: This is correct, the intention here was to make the templates more robust when - for what ever reason - one of theses options is not set. 

Before this PR there has been no defaults in the SCHEMA for the [`brand.*` options](https://github.com/searxng/searxng/pull/213/files#diff-6a82f4161887c10ef8e86cf63056ae5832896914bd0240628ec076cd075c4fff)

I am unclear if this is really needed .. If you think having another default here is to much, I can remove the empty strings.

utils/build_env.py
dalf: This code --> just the the TWITTER_URL line

return42: Twitter-URL is removed (entirely) in https://github.com/searxng/searxng/pull/213/commits/8edf829e of this PR.

searx/__init__.py
dalf: TLDR: The only thing that I would remove if the default value in most (all?) `get_settings` calls : it spreads the default values everywhere.

---

Mandatory values according the AGPL license (the app should stop if there are not defined):
* `git_url`
* `git_branch`

--> no default values. An exception is excepted if the value is undefined.
Side note: reading #36 , perhaps in another PR, the code should get these values from the current git repository (or be overridden by settings.yml value ?). The discussion about this should go into the #36 .

Optional (?) values:
* `issue_url`
* `new_issue_url`
* `docs_url`
* `public_instances`
* `wiki_url`

--> if the value is not specified, the Jinja templates & the documentation should hide the links instead of rendering empty links.

So in theory the schema should be:
```python
    'brand': {
        'git_url': SettingsValue(str),
        'git_branch': SettingsValue(str),
        'issue_url': SettingsValue(str, None),
        'new_issue_url': SettingsValue(str, None),
        'docs_url': SettingsValue(str, None),
        'public_instances': SettingsValue(str, None),
        'wiki_url': SettingsValue(str, None),
    },
```

But in the master branch, the code returns an empty string for undefined values... so actually it doesn't change on this aspect.

---

An alternative solution to `get_settings`:

```python
from typing import Any
from searx import settings


_unset = object()


class Settings:

    __slots__ = 'settings',

    def __init__(self, settings):
        self.settings = settings

    def __getitem__(self, item):
        return self.settings[item]

    def getvalue(self, name: str, default=_unset) -> Any:
        if name in ('settings',):
            return super().__getattribute__(name)
        value = self.settings
        for a in name.split('.'):
            if isinstance(value, dict):
                value = value.get(a, _unset)
            else:
                value = _unset

            if value is _unset:
                if default is _unset:
                    raise KeyError(name)
                value = default
                break

        return value

    def __getattribute__(self, name: str) -> Any:
        """Optional hack: allow to write settings.brand.git_url 
        in a similar way to threading.local
        no default value.

        Might be useful later if the settings.yml becomes a class similar to the Preferences class.
        """
        if name in ('settings', '__slots__', 'getvalue'):
            return super().__getattribute__(name)
        value = self.settings[name]
        if isinstance(value, dict):
            # it would be better to create the Settings instance once, instead at each call.            
            return Settings(value)
        return value

s = Settings(settings)
print(s['brand']['git_url'])  # s is subscriptabled as it was the case before.

print(s.getvalue('brand.git_url'))  # getsettings becomes a method. To avoid confusion it is named getvalue not get.
print(s.getvalue('brand.issue_url', 'https://github.com'))  # with a default value

print(s.brand.git_url)  # s has attribute (optional hack)
print(s.engines[0]['engine'])  # can't write s.egines[0].engine
```

return42:
@dalf thanks a lot for your review and your remarks .. I rebased the patches you mentioned: 

>  it spreads the default values everywhere.

Ah, OK I understand .. I fixed it in 3f638ed1, if you are looking for `get_setting` usages, you will see that there are no more defaults in the function call.

> Mandatory values according the AGPL license (the app should stop if there are not defined):

good point I haven't thought about  .. but whatever we do, we will have the defaults in:

https://github.com/searxng/searxng/blob/17e028e57928f7925a2d1ee1177d28ff33ff8487/searx/settings.yml#L6-L13

> So in theory the schema should be:

beside what I said above, it make sense to remove the default (empty string) in the SCHEMA ..

See 2a2b6c4d / with None default, I needed to add the values to the settings_robot.yml also.

>  if the value is not specified, the Jinja templates & the documentation should hide the links instead of rendering empty links.

IMO this is not in the scope of this PR .. lets do it after we have merged this PR.

> An alternative solution to get_settings

Very interesting approach you suggested .. I hesitate to implement, since this affects all implementations I have done in this PR and in PR #216 .. so I confirm with your remark ..

> Might be useful later if the settings.yml becomes a class similar to the Preferences class.

PR #216, opened 2021-07-21
normalize environment of installation tasks (shell) with YAML settings

return42:
## What does this PR do?

This PR continues the work on _normalization_ of all settings into the `settings.yml` file.  While PR #213 has its focus on removing the need of a `searx.brand` namespace (which is nothing more than a indirection to YAML settings), this PR has it focus on the shell scripts (tasks running outside of an installed instance).

In commit 94851790 we have centralized all SearXNG setups in the settings.yml file:

    94851790 [mod] move brand options from Makefile to settings.yml

This step has not yet been completed for the installation procedures!  Since all SearXNG setups are done in the `settings.yml` these environment variables needs to be removed from the `./conf.sh` file.  Scripts and other tasks running outside of
an instance got the needed values from the `./utils/brand.env` file.

We have the `make buildenv` step which creates the `./utils/brand.sh` but this environment is not harmonized with the environments one might set in his `.config.sh` file.

The true is, since 94851790 these shell scripts are more or less broken and do fail if you change some of the YAML settings (but not the corresponding values in the `.config.sh`).

With this PR all SearXNG instance settings are removed from the `./config.sh` and the  `./utils/brand.env` has been cleaned up to what is needed by the installation tasks.

To harmonize shell script's environment wit the YAML settings, a complete revise of the installation steps and a documentation was needed. To understand all details please read along the commit messages starting from commit  (the commits before are from PR #213 ).

    [mod] load .config.sh from installation (utils/lib_install.sh)

I tested fully installations (including filtron & morty & apache with apache & nginx HTTP-server) in LXC and I have done a fully installation at https://darmarit.org/searx/ 

To test the installation scripts and the YAML Settings in a real use case, a _brand_ is needed in which some customizing of the installation options are changed (in the `settings.yml`). The last patch (1dc4347e) was added only for testing (do not merge this patch):

    [brand] !!DO NOT MERGE!! this patch uses branch fix-searx.sh @return42

This patch changes the buildenv to::

    GIT_URL='https://github.com/return42/searxng'
    GIT_BRANCH='fix-searx.sh'
    SEARX_PORT='7777'
    SEARX_BIND_ADDRESS='127.0.0.12'

To test installation procedure, clone feature branch (fix-searx.sh)::

    $ cd ~/Downloads
    $ git clone --branch fix-searx.sh https://github.com/return42/searxng searxng
    $ cd searxng

    $ sudo ./utils/searx.sh install all
    ...
    $ sudo ./utils/filtron.sh install all
    $ sudo ./utils/filtron.sh nginx install
    ...
    $ sudo ./utils/morty.sh install all
    $ sudo ./utils/morty.sh nginx install

If you do not want to install on your desktop use a LXC container,  here is what I have done to prepare a Container:

    sudo -H ./utils/lxc.sh remove searx-archlinux
    sudo -H FORCE_TIMEOUT=5 ./utils/lxc.sh build searx-archlinux
    sudo -H ./utils/lxc.sh cmd searx-archlinux bash

The last commands opens a bash in the container,  in this bash execute the commands from above:

    ./utils/searx.sh install all
    ./utils/filtron.sh install all
    ./utils/filtron.sh nginx install
    ./utils/morty.sh install all
    ./utils/morty.sh nginx install


return42:
FYI: a preview oft the modified documentation is available here ..

- [make buildenv](https://return42.github.io/searxng/dev/makefile.html#make-buildenv)
- [setting: brand](https://return42.github.io/searxng/admin/engines/settings.html#brand)
- [setting: server](https://return42.github.io/searxng/admin/engines/settings.html#server)
- see tab [Use default settings](https://return42.github.io/searxng/admin/installation-searx.html#configuration)

return42:
FYI: rebased this PR on current master, where PR #213 is already merged.

utils/searx.sh
```diff
@@ -447,6 +467,39 @@ verify_continue_install(){
...
+for name in "${1}".split('.'):
+   current += name
+   memo = memo.get(name, unset)
+   if memo is unset:
+       sys.stderr.write("error: setting '%s' does not exists\n" % current)
+       sys.exit(42)
+   else:
+       current += "."
+print(memo)
+sys.exit(0)
```
dalf, line 490: Why not use the `get_settings` function?

dalf:
Disclaimers:
* I understand the installation scripts only on the surface, so I may miss different aspects.
* On the long run, I really wish to have a different repository for installation scripts, here I feel it goes in the opposite direction.

IMO, commits f61c918dd4ea211f32874f1072ecfca0f39fcca4 and 8f03d2b091570a07618a0c4f5ce644bd98f28f01 mixes the install and runtime phases:
* the install configuration = the configuration to install searx and optionnaly filtron and morty (set up the services, etc...)
* the runtime configuration = settings.yml (may be updated by the install phase)

With this commit, the brand.env file contains values about the runtime configuration, not about the brand, and it is commited.
Moreover, it creates an issue that is fixed by 8f03d2b091570a07618a0c4f5ce644bd98f28f01

If I understand correctly, in the commit 0a6e6444aa4ffad46ec64ce92cabecce5d5bb3ee, filtron.sh uses server.bind_address & server.port values.
Why not keep them in .config.sh as it is the case now ?
* in the install phase, the filtron.sh can rely .config.sh not on settings.yml (perhaps, searx.sh can modify the deployed settings.yml)
* once installed, I'm not sure what should be done / what is done:
   * either it can be read settings.yml : there is no need to store and commit a file for that. A shell script can call a Python script to read these values (actually implemented in 0a6e6444aa4ffad46ec64ce92cabecce5d5bb3ee)
   * either it can continue to rely on .config.sh

Question: is the settings.yml uses during the install phase different from the one that will be use at runtime ( utils/templates/etc/searx/settings.yml ) ? If yes, it would make sense to keep all install variables in .config.sh

Perhaps I miss something.

return42:
> On the long run, I really wish to have a different repository for installation scripts,

We talked about, I have it in mind .. may be first I will remove LXC and then installation procedures but I can't promise you when I have time for .. but it will not be forgotten.

> IMO, commits f61c918 and 8f03d2b mixes the install and runtime phases:

Yes, I do not have a complete clear view where the cutting line is, I see two scenarios:

1. install from a branch / in this branch the settings.yml includes _"runtime configuration"_ and _"brand configuration"_ and `utils/brand.env` exports the needed values oft the _brand_ and the _runtime_ to the shell scripts.
2. install branch `master` directly from a clone of [SearXNG](https://github.com/searxng/searxng) / in this scenario first settings.yml is modified and  `utils/brand.env` is updated by a `make buildenv` before installation can start.

In both scenarios the shell scripts got the values from the `etc/settings.yml` by sourcing the `utils/brand.env` environment. What me sometimes confuse: after installation you have `etc/settings.yml` and `utils/brand.env` local in the clone from which you installed and both files also exist in the installation you have done before. And not to forget there is the real runtime environment of the instance which build from `/etc/searx/settings.yml`.

> Why not keep them in .config.sh as it is the case now ?

May be [I starred to long at this goat](https://en.wikipedia.org/wiki/The_Men_Who_Stare_at_Goats_(film)) and lost something .. I can give it a try but first I need a distance. This PR pushes some things to rights and fixes real bugs.  If there is no severe issue in the installation procedure and it is OK for you I would merge this PR and later with some distance I can think again new about `.conf.sh`

utils/searx.sh
return42: Thanks for pointing out, there was also a bug / I fixed the commit in a rebase / more details in commit message (57b8f340a66), see remark about `utils/lib_install.sh`

Issue #222, opened 2021-07-23
Bug: qwant engine


return42:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

current master

**How did you install SearXNG?**

Installation scripts 

**What happened?**

Various exceptions in the qwant engine

**How To Reproduce**

https://darmarit.org/searx/stats?engine=qwant

**Technical report**

Error
  * Error: ValueError
  * Percentage: 0
  * Parameters: `()`
  * File name: `searx/engines/qwant.py:159`
  * Function: `response`
  * Code: `pub_date = datetime.fromtimestamp(item['date'], None)`

Error
  * Error: TypeError
  * Percentage: 0
  * Parameters: `('an integer is required (got type NoneType)',)`
  * File name: `searx/engines/qwant.py:196`
  * Function: `response`
  * Code: `pub_date = datetime.fromtimestamp(item['date'])`

Error
  * Error: httpx.ConnectTimeout
  * Percentage: 0
  * Parameters: `(None, None, 'api.qwant.com')`
  * File name: `searx/search/processors/online.py:99`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Error
  * Error: TypeError
  * Percentage: 0
  * Parameters: `('unsupported type for timedelta seconds component: NoneType',)`
  * File name: `searx/engines/qwant.py:195`
  * Function: `response`
  * Code: `length = timedelta(seconds=item['duration'])`

Error
  * Error: searx.exceptions.SearxEngineAPIException
  * Percentage: 0
  * Parameters: `('API error::locale must be a string,locale must be one of the following values: en_gb, en_ie, en_us, en_ca, en_in, en_my, en_au, en_nz, cy_gb, gd_gb, de_de, de_ch, de_at, fr_fr, br_fr, fr_be, fr_ch, fr_ca, fr_ad, fc_ca, ec_ca, co_fr, es_es, es_ar, es_cl, es_co, es_mx, es_pe, es_ad, ca_es, ca_ad, ca_fr, eu_es, eu_fr, it_it, it_ch, pt_br, pt_pt, pt_ad, nl_be, nl_nl, pl_pl, zh_hk, zh_cn, fi_fi, bg_bg, et_ee, hu_hu, da_dk, nb_no, sv_se, ko_kr, th_th, cs_cz, ro_ro, el_gr',)`
  * File name: `searx/engines/qwant.py:114`
  * Function: `response`
[...]

PR #223, opened 2021-07-24
[fix] qwant engine - hardening the engine

return42:
## What does this PR do?

Hardening qwant engine by handle missing values more permissive, details see commit messages.

be more reliable

The exception from lang=all can be tested by: `!want :all foo`  / Other Errors are hard to reproduce but it is obviously that the patched code is more permissive. 

Most of the errors reported in #222 are hard to reproduce, except the  `!want :all foo`

## Related issues

https://github.com/searxng/searxng/issues/222

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -156,7 +156,10 @@ def response(resp):
                 })
 
             elif mainline_type == 'news':
-                pub_date = datetime.fromtimestamp(item['date'], None)
+
+                pub_date = item['date']
+                if pub_date is not None:
+                    pub_date = datetime.fromtimestamp(pub_date)
                 news_media = item.get('media', [])
                 img_src = None
                 if news_media:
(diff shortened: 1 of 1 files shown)


searx/engines/qwant.py
```diff
@@ -144,8 +146,8 @@ def response(resp):
         mainline_items = row.get('items', [])
         for item in mainline_items:
 
-            title = item['title']
-            res_url = item['url']
+            title = item.get('title', None)
+            res_url = item.get('url', None)
```
dalf, line 150: minor comment: it can be `res_url = item.get('url')` (without `None` since this is the default value)
See: https://docs.python.org/3/library/stdtypes.html#dict.get

return42: > without None since this is the default value

Thanks, I will take this in mind for the future.

PR #226, opened 2021-07-25
[mod] hardening SearXNG instances by default (formats)

return42:
## What does this PR do?

Deny formats has been implemented in 6ed4616d.

To harden SearXNG instances by default, other formats than HTML should be
denied.  Most of JSON, RSS and CSV requests are bots [1]::

    Bots are the only users of this feature on a public instance, and they abuse
    it too much that the engines rate limit pretty quickly the IP address of the
    instance.

[1] https://github.com/searxng/searxng/issues/95

hardening SearXNG instances by default (formats)

Can't be tested local, but it is obviously that disabling those formats only used by bots is hardening the instance.

## Hint ##

Flask app in unit tests now uses the `unit/settings/test_settings.yml`  (e02b5469)

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -27,7 +27,9 @@ search:
   # max ban time in seconds after engine errors
   max_ban_time_on_fail: 120
   # remove format to deny access, use lower case.
-  formats: [html, csv, json, rss]
+  # formats: [html, csv, json, rss]
+  formats:
+    - html
 
 server:
   port: 8888
--- a/tests/unit/__init__.py
+++ b/tests/unit/__init__.py
@@ -1,0 +1,6 @@
+import os
+from os.path import dirname, sep, abspath
+
+# In unit tests the user settings from unit/settings/test_settings.yml are used.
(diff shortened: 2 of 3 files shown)


dalf:
* in the case of a public instance: there is filtron
* in the case of a private instance: I guess the bots won't know the instance?

but I don't have a clear opinion of this.

return42:
> in the case of a public instance: there is filtron

Filtron does not detect all bots .. An example from practice: Since I have denied cvs, json and rss access on my public instance at darmarit.org  I am not blocked any more by any engine .. and I had also before a strong configured filtron in front.

I think @unixfox (how opened https://github.com/searxng/searxng/issues/95) has the similar experiences.

> in the case of a private instance: I guess the bots won't know the instance?

I think so, on the opposite; users running a local instance most often do not use these formats. If they need e.g. csv it is not hard to activate.

In practice most SearXNG hoster will install an instance and do not take care about the denied/allowed formats, most often they do need only HTML and open the door for bots since theses formats are enabled by default.

PR #229, opened 2021-07-27
version based on git

dalf:
## What does this PR do?

This commit removes:
* the need to update the brand for GIT_URL and GIT_BRANCH.
* these settings:
https://github.com/searxng/searxng/blob/ed0db4d61a2e0dae4953eb813604e2d8834868bb/searx/settings.yml#L13-L16

these values are read from the git repository: it requires the directory to be git repository.

It is possible to call `python -m searx.version freeze` to freeze the current version (it creates the `searx/version_frozen.py` file)
It is useful when the code is install outside git (distro package, docker, etc...)

The code is based on [setuptool_scm](https://github.com/pypa/setuptools_scm/blob/main/src/setuptools_scm/git.py): it calls the `git` command line to retrieve the values. See `searx/version.py`.

* see comments of https://github.com/searxng/searxng/pull/36
* most of the fork don't update the brand ( for example: https://salsa.debian.org/debian/searx/-/blob/master/searx/brand.py https://gitlab.e.foundation/e/cloud/my-spot/-/blob/master/utils/brand.env )
* `make docker.build` : this PR fixes a possible inconsistency in the master branch:
https://github.com/searxng/searxng/blob/ed0db4d61a2e0dae4953eb813604e2d8834868bb/manage#L274-L278
  * GIT_URL is from the brand.
  * SEARX_GIT_VERSION and VERSION_GITCOMMIT are from the local git repository.
  * in a fork, if the brand is not up to date (the user has not changed the git_url) these values won't match.
* in another PR : the bug report form will contains the git URL and commit automatically.

With the change, the footer is always up to date without additional configuration:
image

* `1.0.0-590-f2979c72` comes from `git describe`
* `-dirty` shows that there are some uncommited changes.

* `make run`: check the version at the bottom of the page is the right one.
* create a tag locally, check the version is still right.
* call `python -m searx.version freeze`
   * add some change
   * check that version is same as before the changes
* `make docker`:  
   * check the version at the bottom of the page is the right one.
* `make docs`
* use `utils/searx.sh` to install the app
  * check the version
  * check the `utils/searx.sh update` works

## Related issues

Closes: https://github.com/searxng/searxng/pull/36

### DIFF
--- a/searx/version.py
+++ b/searx/version.py
@@ -1,26 +1,142 @@
-# -*- coding: utf-8 -*-
-'''
-searx is free software: you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=missing-function-docstring,missing-module-docstring,missing-class-docstring
 
-searx is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Affero General Public License for more details.
+import re
+import os
+import shlex
+import subprocess
+import logging
 
-You should have received a copy of the GNU Affero General Public License
-along with searx. If not, see < http://www.gnu.org/licenses/ >.
+# fallback values
+# if there is searx.version_frozen module, and it is not possible to get the git tag
+VERSION_STRING = "1.0.0"
+VERSION_TAG = "1.0.0"
+GIT_URL = "unknow"
+GIT_BRANCH = "unknow"
 
-(C) 2013- by Adam Tauber, <asciimoo@gmail.com>
-'''
+logger = logging.getLogger("searx")
 
-# version of searx
-VERSION_MAJOR = 1
-VERSION_MINOR = 0
-VERSION_BUILD = 0
+SUBPROCESS_RUN_ENV = {
+    "PATH": os.environ["PATH"],
+    "LC_ALL": "C",
+    "LANGUAGE": "",
+}
 
-VERSION_STRING = "{0}.{1}.{2}".format(VERSION_MAJOR,
-                                      VERSION_MINOR,
-                                      VERSION_BUILD)
+
+def subprocess_run(args, **kwargs):
+    """Call :py:func:`subprocess.run` and return (striped) stdout.  If returncode is
+    non-zero, raise a :py:func:`subprocess.CalledProcessError`.
+    """
+    if not isinstance(args, (list, tuple)):
+        args = shlex.split(args)
+
+    kwargs["env"] = kwargs.get("env", SUBPROCESS_RUN_ENV)
+    kwargs["encoding"] = kwargs.get("encoding", "utf-8")
+    kwargs["stdout"] = subprocess.PIPE
+    kwargs["stderr"] = subprocess.PIPE
+    # raise CalledProcessError if returncode is non-zero
+    kwargs["check"] = True
+    proc = subprocess.run(args, **kwargs)  # pylint: disable=subprocess-run-check
+    return proc.stdout.strip()
+
+
+def get_git_url_and_branch():
+    try:
+        ref = subprocess_run("git rev-parse --abbrev-ref @{upstream}")
+    except subprocess.CalledProcessError:
+        ref = subprocess_run("git rev-parse --abbrev-ref master@{upstream}")
+    origin, git_branch = ref.split("/", 1)
+    git_url = subprocess_run(["git", "remote", "get-url", origin])
+
+    # get https:// url from git@ url
+    if git_url.startswith("git@"):
+        git_url = git_url.replace(":", "/", 2).replace("git@", "https://", 1)
+    if git_url.endswith(".git"):
+        git_url = git_url.replace(".git", "", 1)
+
+    return git_url, git_branch
+
+
+def get_git_version():
+    try:
+        tag = subprocess_run("git describe HEAD")
+        # a. HEAD is on tag name, example: tag = "v1.0.1"
+        # b. HEAD is not a tag name, example "<tag>-<distance>-g<commit>"
+        tag_version, tag_distance, tag_commit = (tag.split("-") + ["", ""])[:3]
+        if re.match(r"v[0-9]+\.[0-9]+\.[0-9]+", tag_version):
+            # tag_version "v1.0.0" becomes "1.0.0" (without the v)
+            # other patterns are kept untouched
+            tag_version = tag_version[1:]
+        # remove "g" prefix from tag_commit
+        if tag_commit and tag_commit[0] == "g":
+            tag_commit = tag_commit[1:]
+        # set git_version to "1.0.0-590-0686e274" or '1.0.0'
+        git_version = "-".join(filter(bool, [tag_version, tag_distance, tag_commit]))
+    except subprocess.CalledProcessError:
+        # fall back to "YYYY.MM.DD.Hash" if there is no tag at all
+        git_version = subprocess_run(r"git show -s --format='%as-%h'")
+        # PEP 440: replace - with .
+        tag_version = git_version = git_version.replace("-", ".")
+
+    # add "-dirty" suffix if there are uncommited changes except searx/settings.yml
+    try:
+        subprocess_run(
+            "git diff --quiet -- . ':!searx/settings.yml' ':!utils/brand.env'"
+        )
+    except subprocess.CalledProcessError as e:
+        if e.returncode == 1:
+            git_version += "-dirty"
+        else:
+            logger.warning(
+                '"%s" returns an unexpected return code %i', e.returncode, e.cmd
+            )
+    return git_version, tag_version
+
+
+try:
+    from searx.version_frozen import VERSION_STRING, VERSION_TAG, GIT_URL, GIT_BRANCH
+except ImportError:
+    try:
+        try:
+            VERSION_STRING, VERSION_TAG = get_git_version()
+        except subprocess.CalledProcessError as ex:
+            logger.error("Error while getting the version: %s", ex.stderr)
+        try:
+            GIT_URL, GIT_BRANCH = get_git_url_and_branch()
+        except subprocess.CalledProcessError as ex:
+            logger.error("Error while getting the git URL & branch: %s", ex.stderr)
+    except FileNotFoundError as ex:
+        logger.error("%s is not found, fallback to the default version", ex.filename)
+
+
+logger.info("version: %s", VERSION_STRING)
+
+if __name__ == "__main__":
+    import sys
+
+    if len(sys.argv) >= 2 and sys.argv[1] == "freeze":
+        # freeze the version (to create an archive outside a git repository)
+        python_code = f"""# SPDX-License-Identifier: AGPL-3.0-or-later
+# this file is generated automatically by searx/version.py
+
+VERSION_STRING = "{VERSION_STRING}"
+VERSION_TAG = "{VERSION_TAG}"
+GIT_URL = "{GIT_URL}"
+GIT_BRANCH = "{GIT_BRANCH}"
+"""
+        with open(
+            os.path.join(os.path.dirname(__file__), "version_frozen.py"), "w"
+        ) as f:
+            f.write(python_code)
+            print(f"{f.name} created")
+    else:
+        # output shell code to set the variables
+        # usage: eval "$(python -m searx.version)"
+        shell_code = f"""
+VERSION_STRING="{VERSION_STRING}"
+VERSION_TAG="{VERSION_TAG}"
+GIT_URL="{GIT_URL}"
+GIT_BRANCH="{GIT_BRANCH}"
+"""
+        print(shell_code)
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,7 @@
 from setuptools import setup
 from setuptools import find_packages
 
-import os
-import sys
-
-from searx.version import VERSION_STRING
+from searx.version import VERSION_TAG, GIT_URL
 from searx import get_setting
 
 with open('README.rst', encoding='utf-8') as f:
@@ -21,12 +18,12 @@ with open('requirements-dev.txt') as f:
 
 setup(
     name='searx',
-    version=VERSION_STRING,
+    version=VERSION_TAG,
     description="A privacy-respecting, hackable metasearch engine",
     long_description=long_description,
     url=get_setting('brand.docs_url'),
     project_urls={
-        "Code": get_setting('brand.git_url'),
+        "Code": GIT_URL,
         "Issue tracker": get_setting('brand.issue_url')
     },
     classifiers=[
--- a/utils/build_env.py
+++ b/utils/build_env.py
@@ -7,15 +7,12 @@ import sys
 import os
 from os.path import realpath, dirname, join, sep, abspath
 
-from searx.version import GIT_URL, GIT_BRANCH
-
 repo_root = realpath(dirname(realpath(__file__)) + sep + '..')
 sys.path.insert(0, repo_root)
 
-# Under the assumption that a brand is always a fork assure that the settings
-# file from reposetorie's working tree is used to generate the build_env, not
-# from /etc/searx/settings.yml.
-os.environ['SEARX_SETTINGS_PATH'] = abspath(dirname(__file__) + sep + 'settings.yml')
+# Assure that the settings file from reposetorie's working tree is used to
+# generate the build_env, not from /etc/searx/settings.yml.
+os.environ['SEARX_SETTINGS_PATH'] = join(repo_root, 'etc', 'settings.yml')
 
 def _env(*arg, **kwargs):
     val = get_setting(*arg, **kwargs)
@@ -49,9 +46,10 @@ for name, option in name_val:
     if not os.environ.get(name, _unset) is _unset:
         del os.environ[name]
 
-# After the variables are unset in the environ, we can import settings
-# (get_setting) from searx module.
+# After the variables are unset in the environ, we can import from the searx
+# package (what will read the values from the settings.yml).
 
+from searx.version import GIT_URL, GIT_BRANCH
 from searx import get_setting
 
 print('build %s (settings from: %s)' % (brand_env, os.environ['SEARX_SETTINGS_PATH']))
--- a/utils/build_env.py
+++ b/utils/build_env.py
@@ -7,6 +7,8 @@ import sys
 import os
 from os.path import realpath, dirname, join, sep, abspath
 
+from searx.version import GIT_URL, GIT_BRANCH
+
 repo_root = realpath(dirname(realpath(__file__)) + sep + '..')
 sys.path.insert(0, repo_root)
 
@@ -29,9 +31,6 @@ def _env(*arg, **kwargs):
 # - ./manage function buildenv.unset_env()
 
 name_val = [
-
-    ('GIT_URL'                , 'brand.git_url'),
-    ('GIT_BRANCH'             , 'brand.git_branch'),
 
     ('SEARX_URL'              , 'server.base_url'),
     ('SEARX_PORT'             , 'server.port'),
@@ -57,8 +56,9 @@ from searx import get_setting
 
 print('build %s (settings from: %s)' % (brand_env, os.environ['SEARX_SETTINGS_PATH']))
 sys.path.insert(0, repo_root)
-from searx import settings
 
 with open(repo_root + sep + brand_env, 'w', encoding='utf-8') as f:
     for name, option in name_val:
         print("export %s='%s'" % (name, _env(option)), file=f)
+    print(f"export GIT_URL='{GIT_URL}'", file=f)
+    print(f"export GIT_BRANCH='{GIT_BRANCH}'", file=f)
--- a/utils/searx.sh
+++ b/utils/searx.sh
@@ -213,20 +213,26 @@ main() {
                 *) usage "$_usage"; exit 42;;
             esac ;;
         install)
-            rst_title "SearXNG (install)" part
             sudo_or_exit
             case $2 in
-                all) install_all ;;
+                all)
+                    rst_title "SearXNG (install)" part
+                    install_all
+                    ;;
                 user)
+                    rst_title "SearXNG (install user)"
                     verify_continue_install
                     assert_user
                     ;;
                 pyenv)
+                    rst_title "SearXNG (install pyenv)"
                     verify_continue_install
                     create_pyenv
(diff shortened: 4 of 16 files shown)


searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+
+
+def do_ex_safe(cmd, cwd="."):
+    out, err, exit_code = do_ex(cmd, cwd)
+    if exit_code != 0:
+        raise ShellException(cmd, exit_code, out, err)
+    return out
+
+
+def get_public_git_url(git_url):
```
dalf: Here there is some tricks (?) which works for github and gitlab.

utils/searx.sh
```diff
@@ -445,16 +442,6 @@ prompt_installation_status(){
...
-            instance_setting="$(prompt_installation_setting brand.git_url)"
-            if ! [ "$instance_setting" = "$remote_url" ]; then
-                warn_msg "instance's brand.git_url: '${instance_setting}'" \
-                         "differs from local clone's remote URL: ${remote_url}"
-            fi
-            instance_setting="$(prompt_installation_setting brand.git_branch)"
-            if ! [ "$instance_setting" = "$branch" ]; then
-                warn_msg "instance brand.git_branch: ${instance_setting}" \
-                         "differs from local clone's branch: ${branch}"
-            fi
```
return42: This part was intended to check if an existing installation is from the same GIT_URL like the _clone_ from which you called the installation procedure.

By example: say you have an existing installation from branch https://github.com/return42/searxng/tree/darmarit.org and at some day you clone from https://github.com/searxng/searxng/tree/master (into your Download folder or where ever) and then you run the installation scripts (or other maintenance tasks) from this fresh clone in your Download folder.

If there comes the day, we have isolated installation procedures from SearXNG repository, we can drop this check.

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=str(cwd),
+        env=dict(
+                no_git_env(os.environ),
+                # os.environ,
+                # try to disable i18n
+                LC_ALL="C",
+                LANGUAGE="",
```
return42: nit: indent by 4 spaces.

return42:
Just for my understanding, where do you see the cutting line of installation procedure and the repository which will be installed?

On the long run, we want to isolate all the installation scripts (e.g. ./utils/searx.sh) from the SearXNG repository.

Lets imagine we have a command line named _install_searxng_ and it's usage is like:

    install_searxng <git-url> <git-branch-name>

With I can suppose two **installation types**:

1. Those directly install the `master` branch from `https://github.com/searxng`
2. Those installing a branch from another repository / e.g. like I do, I install branch `darmarit.org` from my fork `https://github.com/return42/searxng`

Now my question is: where does GIT_URL and GIT_BRANCH belongs to?  Should these values in the repository or should this values stored somewhere in the installation?

If I understand this Draft correct, these values are stored in a new file `./searx/version_frozen.py`.  Is this file a part of the repository (will it be committed to the fork) or not .. and what does it mean .. what about this file when updating an existing SearxNG installation?

Having both **installation types** (from above) and an update process in mind, I always revolve around this point and can't get any further.

Can you describe what scenarios you have in mind and what solution you suppose.

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+    if isinstance(str_or_bytes, str):
+        return str_or_bytes.strip()
+    return str_or_bytes.decode("utf-8", "surrogateescape").strip()
+
+
+def do_ex(cmd, cwd="."):
+    if os.name == "posix" and not isinstance(cmd, (list, tuple)):
+        cmd = shlex.split(cmd)
+
+    p = _popen_pipes(cmd, cwd)
```
return42: IMO we should not implement a function (_popen_pipes) that does nothing more than just one function call / we should place the subprocess.Popen constructor right here.  Having the constructor call in a function, breaks the _with construct_ and is hard to read.

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+    # pre-commit hooks
+    # In git 1.9.1 (maybe others), git exports GIT_DIR and GIT_INDEX_FILE
+    # while running pre-commit hooks in submodules.
+    # GIT_DIR: Causes git clone to clone wrong thing
+    # GIT_INDEX_FILE: Causes 'error invalid object ...' during commit
+    return {
+        k: v
+        for k, v in env.items()
+        if not k.startswith("GIT_")
+        or k in ("GIT_EXEC_PATH", "GIT_SSH", "GIT_SSH_COMMAND")
```
return42: Do I miss something, what is the difference of `startswith("GIT_")` and `in ("GIT_EXEC_PATH", "GIT_SSH", "GIT_SSH_COMMAND")`

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+
+def _popen_pipes(cmd, cwd):
+    # pylint: disable=consider-using-with
+    return subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        cwd=str(cwd),
+        env=dict(
+                no_git_env(os.environ),
```
return42: Except the PATH variable, is there any need for the _no GIT env_?

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+        ref = do_ex_safe('git rev-parse --abbrev-ref @{upstream}')
+    except ShellException:
+        ref = do_ex_safe('git rev-parse --abbrev-ref master@{upstream}')
+    origin, git_branch = ref.split('/', 1)
+    git_url = do_ex_safe(f'git remote get-url {origin}')
+    return get_public_git_url(git_url), git_branch
+
+
+def get_git_version():
+    git_version = do_ex_safe(r'git describe --match "v[0-9]*\.[0-9]*\.[0-9]*" HEAD')
```
return42: I think we should not use strings to start a command line tool (needs be executed in a shell / or like you do, using `shlex.split(cmd)` ).  For the sake of simplicity I implemented the [fspath.callEXE](https://return42.github.io/fspath/fspath-api/fspath.fspath.html#fspath.fspath.callEXE). 
```python
>>> import fspath
>>> fspath.callEXE('git', 'describe')
('v1.0.0-590-g0686e274\n', '', 0)
```

If you do not want to use fspath (which is ATM only in requirements-dev.txt) you can use:

```python
>>> import subprocess
>>> proc = subprocess.run(['git', 'describe'], stdout=subprocess.PIPE)
>>> proc.stdout.decode()
'v1.0.0-590-g0686e274\n'

```

dalf:
> Should these values in the repository or should this values stored somewhere in the installation?

That is the point of this PR : `searx/version_frozen.py` is not commited (neither in this git repository nor the forks).
The git repository already has this information and when searxng starts , it calls the `git` command to get the GIT_URL, GIT_BRANCH, VERSION_STRING.

Actually, there are two modes:
* running searxng from a git repository (this repository or another one): version_frozen.py is not required. `version.py` will get the GIT_URL, GIT_BRANCH, VERSION_STRING from the git repository (using the `git` command). In this mode, the searxng installation is updated using `git pull` : searxng will read the `VERSION_STRING` value when it starts.
* running searxng outside a git repository ( debian, arch, yunohost packages ):
   * the files have to come from somewhere and they comes from a git repository at one point or another. So it is possible to call `python -m searx.version freeze` to create `searx/version_frozen.py`.  After that the version is hard coded, the `git` command line is not called anymore.
   * then it is possible to copy / archive the files .

If you run searxng from your fork, `GIT_URL` will be your fork URL.
Technically, `version.py` will:
* get the upstream reference of the current branch `git rev-parse --abbrev-ref @{upstream}` (if the current branch doesn't an upstream, it will check the master branch).
* get the remote URL: `git remote get-url {origin}`

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+    git_version_s = git_version.split('-')
+    git_version_s[0] = git_version_s[0][1:]
+    if len(git_version_s) == 3:
+        git_version_s[2] = git_version_s[2][1:]
```
return42: Juggling with the index in the list is hard to read I recommend to name the values:

```python
tag = proc.stdout.decode().strip()
tag_version, tag_distance, tag_commit = tag.split('-')
```

setup.py
```diff
@@ -41,7 +38,7 @@
     author='Adam Tauber',
     author_email='asciimoo@gmail.com',
     license='GNU Affero General Public License',
-    packages=find_packages(exclude=["tests*", "searx_extra"]),
+    packages=find_packages(exclude=["tests*", "searx_extra", "searx.static.themes.simple.node_modules", "searx.static.themes.oscar.node_modules"]),
```
return42: Is this a part of this PR or is this a bugfix out of the scope of this PR?

return42 (commented):
@dalf thanks for your explanation / sounds good to me .. but to be true I need some practice to see if there is a trap I don't see right now. Anyway, lets go into this direction ... I made some remarks in my review .. I think the version.py can be improved.

If you want, I can try to improve the implementation .. but I cant promise you when.

setup.py
dalf: Nope, this is a leftover from a test.

searx/version.py
dalf: `not k.startswith("GIT_") or k in ...` : 
* it accepts `GIT_EXEC_PATH`, `GIT_SSH`, `GIT_SSH_COMMAND` keys
* and all other keys except the keys that start with `GIT_`.

searx/version.py
dalf: > I think we should not use strings to start a command line tool (needs be executed in a shell / or like you do, using shlex.split(cmd) ).

I don't know understand why `shlex.split(cmd)` is a problem ( it is a copy / paste from setup_scm ).
For me, 
`'git describe --match "v[0-9]*\.[0-9]*\.[0-9]*" HEAD'` 
is more readable than
`['git', 'describe', '--match', 'v[0-9]*\\.[0-9]*\\.[0-9]*', 'HEAD']`

return42: Not really a problem but I would prefer a notation which do not need any kind of *syntax parser".  If you prefer to use one string instead of a argument list you can set `shell=True` in the `subprocess.*` functions (no need for shlex).  Nevertheless ..
[Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names)](https://docs.python.org/3/library/subprocess.html#frequently-used-arguments)

searx/version.py
return42: 🤦🏻 .. sorry.

searx/version.py
dalf: * `shell=True` is a different thing: it calls a shell to actually call the command.
* in `version.py`, all commands are hard coded except one.

return42: > `shell=True` is a different thing: it calls a shell to actually call the command.

Maybe I am used to it differently: for me, when I see a command line in one string I suppose this line will be executed through the shell (including wildcards and other shell builtins) and when I see a list of arguments I suppose that a "program" _is executed_ directly in a subprocess. 

> in `version.py`, all commands are hard coded except one.

I do not understand what you mean by this sentence  .. what means "hard coded" here and "except one" .. all commands in version.py are _hard coded_ (in the meaning of: a constant string is used)

searx/version.py
```diff
@@ -1,26 +1,150 @@
...
+# some code is from https://github.com/pypa/setuptools_scm/blob/b0b46fd723f6649a7ba4895b005bc89db1529e22/src/setuptools_scm/utils.py  # pylint: disable=line-too-long
+# ( MIT License )
+
+import os
+import shlex
+import subprocess
+
+
+class ShellException(Exception):
+
```
return42: Do we need a exception architecture .. why not using the return code of the subprocess directly?

May be it is all better shown by example: I hope I have time this evening .. then I will implement a alternative just to show how I would implement the functionality.

dalf: Look at the `get_git_url_and_branch` function: it raises an exception if any of the git calls fail.

Then
```python
    try:
        GIT_URL, GIT_BRANCH = get_git_url_and_branch()
    except ShellException as e:
        ...
```
is straightforward

searx/version.py
dalf: Done:
https://github.com/searxng/searxng/blob/e5d1cd3c9deccfd2099e48baed855dcc00113ed0/searx/version.py#L93-L106

Note: If the last commit is a tag `git describe` will return only the tag without distance or commit hash.

setup.py
dalf: Fixed

searx/version.py
dalf: See:
https://github.com/searxng/searxng/blob/e5d1cd3c9deccfd2099e48baed855dcc00113ed0/searx/version.py#L37-L41

searx/version.py
dalf: Done.
Note: in this scenario ( communicate), it seems _with_ advantage is to deal with Ctrl-C:
https://github.com/python/cpython/blob/12fc0d28fcdeca32314d0755d3b30c7d96907440/Lib/subprocess.py#L1014-L1043

searx/version.py
dalf: Done (using black on the whole file).

searx/version.py
return42: > Note: in this scenario ( communicate), it seems with advantage is to deal with Ctrl-C:

Thanks for the link, but .. Nope, I think the "with" context is to close the stdout & stderr.  The try/except block is around closing stdin .. but its also the first time I looked at this :-)

searx/version.py
```diff
@@ -1,26 +1,173 @@
...
+def ensure_stripped_str(str_or_bytes):
+    if isinstance(str_or_bytes, str):
+        return str_or_bytes.strip()
+    return str_or_bytes.decode("utf-8", "surrogateescape").strip()
+
+
+def call_subprocess(cmd, cwd="."):
+    if os.name == "posix" and not isinstance(cmd, (list, tuple)):
+        cmd = shlex.split(cmd)
```
return42: sorry if I'm being annoying :-)  .. What use case (OS) do you have in mind for _non posix_ .. and what happens with the sub process call when ` os.name != "posix"`?

searx/version.py
dalf: Is it acceptable to have to string in this particular use case?

> all commands in version.py are hard coded 

yes except:
https://github.com/searxng/searxng/blob/e5d1cd3c9deccfd2099e48baed855dcc00113ed0/searx/version.py#L79

The shell is not involved only `git`, and `origin` comes from a previous git output.

searx/version.py
dalf: My take was setup_scm has most probably dealt with various platforms.

But, more in detail, from https://docs.python.org/3/library/subprocess.html#subprocess.Popen :

> On Windows, if args is a sequence, it will be converted to a string in a manner described in Converting an argument sequence to a string on Windows. This is because the underlying CreateProcess() operates on strings.

And according to https://stackoverflow.com/questions/1854/python-what-os-am-i-running-on `os.name != 'posix'` only on Windows (and Jython which is out of scope and perhaps some other implementation I'm not aware of, yes it should be `os.name != 'nt'` according to what I understand even in setup_scm).

But anyway `searx/version.py` works on Windows even with the commit 7710c9422fceda1d45cbdda634b80b89ceee2ac6

searx/version.py
```diff
@@ -1,26 +1,178 @@
...
+
+logger.info("version: %s", VERSION_STRING)
+
+if __name__ == "__main__":
+    import sys
+
+    if len(sys.argv) >= 2 and sys.argv[1] == "freeze":
+        # freeze the version (to create an archive outside a git repository)
+        # Looks at the current index and checks to see if merges or updates are needed by checking stat() information.
+        call_subprocess_safe("git update-index -q --refresh")
```
return42: For my understanding, for what is this command needed?

Update: it seems that the -q option suppress the return code ... 

```
$ git update-index --refresh
searx/version.py: needs update
error(1)
$ git update-index -q --refresh
(0)
```

utils/searx.sh
dalf: I have updated this section:
https://github.com/searxng/searxng/blob/7710c9422fceda1d45cbdda634b80b89ceee2ac6/utils/searx.sh#L444-L458

searx/version.py
dalf: See commit 2d60d80fe7cf04d50933c6e9469b1f625111309d

This is copy / paste from the existing code (from fbe40001d35ea5cf74c92f35d76c642a480a5c9f ):
https://github.com/searxng/searxng/blob/feef577f996b00c4b4b610b6e3a17e2183d20927/manage#L235-L240

There is no reason to use the low level and "git diff --quiet" makes less syscalls (so most probably faster).

return42:
FYI: alternative [version.py](https://github.com/return42/searxng/blob/version-from-git/searx/version.py
) implementation (including last patches from here) https://github.com/return42/searxng/commits/version-from-git

utils/build_env.py
```diff
@@ -57,8 +56,9 @@ def _env(*arg, **kwargs):
 
 print('build %s (settings from: %s)' % (brand_env, os.environ['SEARX_SETTINGS_PATH']))
 sys.path.insert(0, repo_root)
-from searx import settings
```
return42, line 60: This is the fragile part in the old implementation, read the comments in the lines before.  The problem is, that the settings.yml is imported as soon as you import any searx modules.  If the buildenv should work, you first need to remove any environment (see code above) before importing searx, otherwise the environment will overwrite the values you have modified in the settings.yml.

I fixed it in my branch https://github.com/return42/searxng/commit/6a2d4915a5a8698083d9b21cf6d89c95d2b3caf7

utils/searx.sh
return42: I'm testing the installation procedures .. and I will fix it, give me some time (this solution do not work for me).

searx/version.py
dalf: Updated

return42:
> Actually, there are two modes:
> * running searxng from a git repository (this repository or another one): version_frozen.py is not required. `version.py` will get the GIT_URL, GIT_BRANCH, VERSION_STRING from the git repository (using the `git` command). In this mode, the searxng installation is updated using `git pull` : searxng will read the `VERSION_STRING` value when it starts.
> * running searxng outside a git repository ( debian, arch, yunohost packages ):
>   * the files have to come from somewhere and they comes from a git repository at one point or another. So it is possible to call `python -m searx.version freeze` to create `searx/version_frozen.py`.  After that the version is hard coded, the `git` command line is not called anymore.
>   * then it is possible to copy / archive the files .

Hm, now I see what I missed .. this is run-time, but I asked for the installation-time:

> With I can suppose two **installation types**:
>
> 1. Those directly install the `master` branch from `https://github.com/searxng`

What I have tried / with the fixes in [my branch](https://github.com/return42/searxng/commits/version-from-git)

1.1. clone `https://github.com/searxng/searxng`
1.2. edit `etc/settings.yml` (e.g. edit server.port from 8888 to 7777)
1.3. make buildenv
1.4. install complete suite::

     ./utils/searx.sh install all
     ./utils/filtron.sh install all
     ./utils/filtron.sh nginx install
     ./utils/morty.sh install all

> 2. Those installing a branch from another repository / e.g. like I do, I install branch `darmarit.org` from my fork `https://github.com/return42/searxng`

2.1. clone (e.g.) `https://github.com/return42/searxng`
2.2. checkout branch (e.g.) version-from-git
2.3. install complete suite::

     ./utils/searx.sh install all
     ./utils/filtron.sh install all
     ./utils/filtron.sh nginx install
     ./utils/morty.sh install all

-----

How to proceed? I would suggest to reset your branch to the state of mine.  Otherwise I have to repeat all tests in your branch (I do not have more time today).

My branch is based on your e5d1cd3c and includes your patch 2d60d80fe7 (and 7710c9422 is not needed in my version.py).

What I mean, everything from your branch (except 93e3e3f2102fd) is in my fully tested branch.  I prefer my branch since 93e3e3f2102fd in your branch means that I have to repeat all my tests.

Is it OK for you to reset your branch with mine?

Afterwards we can squash some commits / I think.

dalf:
~I would like to understand what are the fixes that you have implemented: I don't understand what is not working?~

I would like to know how to test? Just the command you have provided?

(I'm installing a VM to run the scripts)

return42:
>  would like to understand what are the fixes that you have implemented: I don't understand what is not working?

my branch starts at your commit e5d1cd3 and places some patches on top. In detail:

- https://github.com/return42/searxng/commit/5d7757d62bbd797d77ea3df95830011f2695119a is what you have merged (by hand) into your 93e3e3f  / this is the part which is untested if we continue with your branch
- https://github.com/return42/searxng/commit/784a213c0a6802ed4f817f2f04c84c9491314790 is the same what you have done in 2d60d80 (but in the version.py from mine)
- https://github.com/return42/searxng/commit/6a2d4915a5a8698083d9b21cf6d89c95d2b3caf7  is the same what you have merged in 6bb6f79

Until here both branches are more or less identical, while 93e3e3f is untested https://github.com/return42/searxng/commit/5d7757d62bbd797d77ea3df95830011f2695119a is tested.

The next commits are not in your branch, they are only in my branch, fixing some issues I had at installation time ...

- https://github.com/return42/searxng/commit/a1d86da711a67bb88f6ccd8ae77d23ad208f33f5 fixes a bug I had in my tests
- https://github.com/return42/searxng/commit/054d9ddd84af0435097621c7662c84ea60e191a0 improves some messages in the installation procedures.

Sorry if this is a mess of two branches but we developed parallel in two branches.  You merged parts from mine and I merged parts from yours :-)

return42:
If you think PR is ready to review, remove the "Draft" tag .. then I will run my tests this evening.

return42:
> I would like to know how to test? Just the command you have provided?

Yes. Frst I create a LXC:

    sudo -H ./utils/lxc.sh remove searx-archlinux
    sudo -H FORCE_TIMEOUT=5 ./utils/lxc.sh build searx-archlinux

Then I jump in the container ..

    sudo -H ./utils/lxc.sh cmd searx-archlinux bash
 
in this bash in the container I run the installation procedures shown in https://github.com/searxng/searxng/pull/229#issuecomment-889775570

I have tested only in archlinux but others should be the same.

return42 (approved):
Works flawless / lets merge.

BTW: Thank you for your patience with me :-)

dalf:
@paulgoio, since this PR has merged, it seems you have to copy all the files from this repository to [update your fork](https://github.com/paulgoio/searx/commit/3e3049f2944e667c1cbf1e08a33d9d93d74c23aa)?
If you have an issue feel free to create open one.

HLFH:
@dalf 

You said:
> It is possible to call python -m searxng.version freeze to freeze the current version (it creates the searxng/version_frozen.py file). It is useful when the code is installed outside git (distro package, Docker, AUR etc...).

Can the command python -m searxng.version freeze be in the [documentation](https://searxng.github.io/searxng/)?
Otherwise, without this relevant info, people stay with an about page having broken links for *SearXNG sources*.

Issue #195, opened 2021-07-03
Move from transifex to Weblate


return42:
Over the years, transifex has become proprietary crap.

We should move to a [self-hosted](https://weblate.org/de/download/) weblate [1] instance I think.

Related-to: https://github.com/searx/searx/issues/268#issuecomment-769630687
Discussed-at: https://github.com/searx/searx/discussions/2552

[1] https://github.com/WeblateOrg/weblate

dalf:
The [hardware requirement](https://docs.weblate.org/en/latest/admin/install/venv-debian.html) is:
* 2 GB of RAM (4GB recommended)
* 2 CPU cores 
* 1 GB of storage space

Do you have any suggestion where to host this service?

return42:
> The hardware requirement is:

I think the the hardware requirement is not my problem since we will host only one project with a handful of contributors ("concurrent users", see C&P below).

> Do you have any suggestion where to host this service?

The work I see is maintaining the service, managing accounts and having a backup strategy, a vacation replacement of the admin .. etc ... it is the organizational tasks that still make me hesitate.

If it make sense I can start a _"pilot"_ on my darmarit.org server .. later if we think it is a value solution we can consider to rent some hardware and own a domain name that fits (I can pay from my company).  My Limitation: at the moment I have a lot to do for my customers and do not know yet when I will have time for a _"pilot"_  ... suggestions?

----

![grafik](https://user-images.githubusercontent.com/554536/126658210-f836a2a6-7b20-47e9-a151-307bb40fd5e8.png)

comradekingu:
@bourrel Use https://hosted.weblate.org/hosting/ or https://weblate.bubu1.eu/
Add the user "kingu" as admin.

return42:
@comradekingu not sure I understand you, do you know a weblate hoster open for FLOSS projects like SearXNG?

comradekingu:
@return42 Both of those are exactly that.

return42:
@dalf what do you think about creating a project at 

- https://hosted.weblate.org/hosting/ 
- or https://weblate.bubu1.eu/ 

I do not have any preferences / both support registration via github account.  May be `hosted.weblate.org` has a wider user base. The  `weblate.bubu1.eu` service is hosted by Marcus Hoffmann, see https://bubu1.eu/services/

comradekingu:
I'd say Bubu. Guy is doing a lot to help with decentralization of various ecosystems.
Two good options.

dalf:
[...]

PR #230, opened 2021-07-27
Remove zh_Hant_TW translation

dalf:
## What does this PR do?

See https://github.com/searxng/searxng/issues/195#issuecomment-887518808


return42 (approved):
Can't find any issue / seems to work flawless.

PR #236, opened 2021-07-31
manage script: remove unused reference to utils/brand.env and .config.sh

dalf:
## What does this PR do?

`./utils/brand.env` defines 
* SEARX_URL
* SEARX_PORT
* SEARX_BIND_ADDRESS
* GIT_URL
* GIT_BRANCH

`.config.sh` defines some variables for the scripts in the `utils` directory.

None of these values are referenced by:
* `./utils/makefile.include`
* `./manage`
* `./lib/lib.sh`
* `./utils/lib_static.sh`

In this PR, `manage` includes neither `./utils/brand.env` nor `./config.sh`.

Did I miss something?

Decouple installation ( ./utils/*.sh) from the repository (./manage and Makefile)

try the different `Makefile` targets.

### DIFF
--- a/manage
+++ b/manage
@@ -5,9 +5,6 @@
 
 # shellcheck source=utils/lib.sh
 source "$(dirname "${BASH_SOURCE[0]}")/utils/lib.sh"
-# shellcheck source=utils/brand.env
-source "${REPO_ROOT}/utils/brand.env"
-source_dot_config
 
 # shellcheck source=utils/lib_static.sh
 source "$(dirname "${BASH_SOURCE[0]}")/utils/lib_static.sh"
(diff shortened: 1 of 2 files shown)


utils/lib.sh
```diff
@@ -37,17 +37,6 @@ if [[ -z ${DIFF_CMD} ]]; then
...
-
-source_dot_config() {
-    if [[ ! -e "${DOT_CONFIG}" ]]; then
-        err_msg "configuration does not exists at: ${DOT_CONFIG}"
-        return 42
-    fi
-    # shellcheck disable=SC1090
-    source "${DOT_CONFIG}"
-}
-
```
return42: You are correct, we do not need this here.  Can we still leave this function in here in SearXNG? As I wrote in the other post, I also use this library in other projects.  I had also written that I want to modularize the library.  It is very costly for me if I would have to do this modularization once here in SearXNG and then again in the other projects.  I want to use the same files in all projects.  Later, when the modularization is done, we will no longer need certain modules here in SearXNG and then these things would finally disappear from SearXNG.  My time is limited and I would like to avoid to do the same task twice .. it would be a great help to me if this is OK for you .. give me some time, I have an eye on.

return42:
@dalf: this PR looks good to me, I can't see any issue. See my https://github.com/searxng/searxng/pull/236#discussion_r680366112 : it would be helpful for me, if we can leave the `source_dot_config()` function in the `lib.sh` even if we do not use it.

While I reviewed this PR I fixed a comment in 7efaf3a and I removed the brand.git_url (& branch) from the settings documentation (we missed to removed it in PR #229).  The latter patch is not really in the scope of this PR if you do not want this here, drop c8d0f41 and I will send another PR for this.

dalf:
I will remove e5b9a7311143456b7f3a26de585e61faf0be5e7c but can you create another PR for 
7efaf3a  and 
c8d0f41 ?

PR #238, opened 2021-07-31
Automatic update of searx/translations/messages.pot and messages.po files.

dalf:
## What does this PR do?

See commit messages:
* remove .tx directory (transifex)
* babel.cfg: ignore node_modules directories
* update searx_extra/update/update_translations.sh
   * the script expects searx/translations/messages.pot to be commited
   * it calls "pybabel extract", and if there is a meaningful change,
   * it calls "pybabel update"
* github workflow:
   * after each on the master branch of searxng (not elsewhere), a call `searx_extra/update/update_translations.sh` check if there is some changes on `searx/translations/messages.pot`
   * if yes it create a PR with a commit similar to 596577e907607b582709aa7afd51a0cc9df9ca9f so weblate can get the last messages.
* update the messages.po files (it haven't been done since a while).

The workflow is disabled on this branch, how it is on the master branch:
image

see https://github.com/dalf/searxng/actions/runs/1085458894

If this github workflow doesn't create issue later the PR can be replaced by a commit on the master branch.

## Why is this change important?

Related to #195 / integration of weblate.

## How to test this PR locally?

* call  `searx_extra/update/update_translation.sh`
* call again `searx_extra/update/update_translation.sh`: see there is no change (no call to `pybabel update`)
* change `searx/translations/messages.pot`
* call again `searx_extra/update/update_translation.sh`: check that the script calls `pybabel update`

The github workflow can only be tested on the master branch, add a commit to replace `searxng`:
https://github.com/searxng/searxng/blob/53156cd4c66c5fe0a6059515f5df7b7bbe02d35a/.github/workflows/integration.yml#L96

## Related issues

Related to #195

PR #246, opened 2021-08-03
Translation: weblate integration

### LINKED ISSUE #195: Move from transifex to Weblate

dalf:
## What does this PR do?

* after each commit on the master branch:
  check if searx/translations/messages.pot must be updated, if yes:
  * merge commits from weblate
  * build & commit the messages.mo files
  * update & commit messages.pot
  * during these steps, the weblate translations are locked

* every Friday, the translations from weblate are committed into the master branch

Avoid out of sync translations:
* with #238 , the messages are pushed to weblate
* with this PR, the translations are pulled from weblate.

WARNING: with this PR, weblate don't have the right to write on the repository, but the commits made by weblate are merged without notice. It is not possible to create a pull request: it will create a conflict on the weblate repository later. 
I'm really "meh" about this.

A better solution is a dedicated project searxng/translations:
* messages.pot update are pushed to this project
* weblate commits are pushed directly into searxng/translations
* a scheduled workflow create a PR into this repository with the updates.

N/A.
It has to be on a fork ( .github/workflows/*.yml: change `searxng` to fork name). 

## Related issues

Close https://github.com/searxng/searxng/issues/195

### DIFF
--- a/.github/workflows/translations-update.yml
+++ b/.github/workflows/translations-update.yml
@@ -1,0 +1,34 @@
+name: "Update translations (weblate)"
+on:
+  schedule:
+    - cron: "05 07 * * 5"
+  workflow_dispatch:
+
+jobs:
+  babel:
+    name: Weblate
+    runs-on: ubuntu-20.04
+    if: ${{ github.repository_owner == 'searxng' && github.ref == 'refs/heads/master' }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+        token: ${{ secrets.WEBLATE_GITHUB_TOKEN }}
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+        architecture: 'x64'
+    - name: weblate & git setup
+      env:
+        WEBLATE_CONFIG: ${{ secrets.WEBLATE_CONFIG }}
+      run: |
+        mkdir -p ~/.config
+        echo "${WEBLATE_CONFIG}" > ~/.config/weblate
+        git config --global user.email "alex.searxng@al-f.net"
+        git config --global user.name "searxng-bot"
+    - name: Merge and push transation updates
+      id: update
+      run: make V=1 weblate.merge.push.updates
+  
--- a/.github/workflows/translations-update.yml
+++ b/.github/workflows/translations-update.yml
@@ -8,7 +8,7 @@ jobs:
   babel:
     name: Weblate
     runs-on: ubuntu-20.04
-    if: ${{ github.repository_owner == 'searxng' && github.ref == 'refs/heads/master' }}
+    if: ${{ github.repository_owner == 'dalf' && github.ref == 'refs/heads/master' }}
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


manage
```diff
@@ -126,19 +126,55 @@ babel.sha256sum() {
...
+        git commit -m "[translations] update messages.pot and messages.po files"
+        git push
     )
-    dump_return $?
+    exitcode=$?
+    (   # make sure to always unlock weblate
+        set -e
+        pyenv.cmd wlc unlock
+    )
+    dump_return $exitcode
```
return42: I am not familiar with the weblate workflows. Can you add comments to the `ci.babel.update()` and `weblate.merge.push.update()` functions  .. it might be good to describe what these function do in a 1., 2., 3. ... order.

manage
```diff
@@ -126,19 +126,55 @@ babel.sha256sum() {
...
+}
+
 ci.babel.update() {
     local sha_before
```
return42: add `exitcode` (used in line 172) to the local names.

manage
```diff
@@ -126,19 +126,55 @@ babel.sha256sum() {
     grep "msgid" "searx/translations/messages.pot" | sort | sha256sum | cut -f1 -d ' '
 }
 
+weblate.merge.push.updates() {
+    (   set -e
```
return42, line 127: add `local exitcode` (used in line 142)

.github/workflows/translations-update.yml
```diff
@@ -0,0 +1,34 @@
...
+        WEBLATE_CONFIG: ${{ secrets.WEBLATE_CONFIG }}
+      run: |
+        mkdir -p ~/.config
+        echo "${WEBLATE_CONFIG}" > ~/.config/weblate
+        git config --global user.email "alex.searxng@al-f.net"
+        git config --global user.name "searxng-bot"
+    - name: Merge and push transation updates
+      id: update
+      run: make V=1 weblate.merge.push.updates
+
```
return42: nit: strip whitespaces from the end of the file

return42:
@dalf thanks for one step further in the weblate workflows. I am not yet familiar with the weblate workflows / what do you think, does it make sense to replace the old transifex https://searxng.github.io/searxng/dev/translation.html by a description of the translations workflows. At least commenting the [shell functions](https://github.com/searxng/searxng/pull/246#discussion_r681840554) is needed, I think.

dalf:
Yes it makes sense to update https://searxng.github.io/searxng/dev/translation.html

but the more I think about it, the more I wish to move the translations updates to another branch or repository because this PR merges commits from weblate directly into the master branch:
* weblate updates this new repository.
* messages.pot update are pushed to this new repository (from this repository).
* every Friday a PR is created into this repository with the updates ( messages.po, messages.mo are built at this moment). It will squash all the changes from weblate but at least the master branch is shielded by the reviews of the PRs.

return42:
A branch and a PR would be better (like we have for the [update data PR](https://github.com/searxng/searxng/runs/3212384297?check_suite_focus=true#step:7:1)) .. is it a lot more work to rewrite this PR to use a branch for merges from weblate?

manage
dalf: Sum up:
* weblate keeps a clone of the master branch of searxng/searxng.
* it reads searx/translations/messages.pot to get the text to translate
* it reads searx/translations/*/LC_MESSAGES/messages.po to get the existing translations
* it updates & commit searx/translations/*/LC_MESSAGES/messages.po  to persist new translations

[EDIT] one workflow allows weblate to commit directly on the master branch.

So if the CI here changes messages.pot or messages.po, weblate must be locked = no change = no new commit on the weblate clone.

If messages.pot can't be updated from the code directly: weblate will have its own version and it will requires a conflict resolution.

So if messages.pot has some changes, the workflow is 
* lock weblate
* stash to remove messages.pot from the CI clone.
* pull weblate changes (weblate.merge.push.updates function)
  * merge the commits from weblate (only the .po files)
  * build the .mo files, commit
  * push the commits to the master branch
* stash pop to get back messages.pot 
* update the message.po from messages.pot (actually I'm not sure this is necessary: perhaps weblate take care of that)
* commit (messages.pot & message.po files)
* push the commit
* weblate unlock (like a try.... finally : it must be call to allow changes on weblate again).
* after unlock, weblate may add new commits (if the translators make some changes).

For reference: https://docs.weblate.org/en/latest/workflows.html

dalf:
I don't think so, but I'm new to weblate, so I discover at the same time.

For reference: https://docs.weblate.org/en/latest/faq.html#how-can-i-limit-weblate-access-to-only-translations-without-exposing-source-code-to-it

return42:
>  For reference: https://docs.weblate.org/en/latest/faq.html#how-can-i-limit-weblate-access-to-only-translations-without-exposing-source-code-to-it

I would not vote for a git-submodule .. for me it is OK but have seen many newbees fighting with git-submodules and pulls

dalf:
> I would not vote for a git-submodule

That's not my intention.

dalf:
See #250

PR #248, opened 2021-08-03
[mod] "npm run webfont" require fontforge & ttfautohint packages

return42:
## What does this PR do?

Add documentation & buildhost packages needed by:

    "npm run webfont" require fontforge & ttfautohint packages

See commit dca3bcca9::

     [mod] simple theme: include fonts

     "npm run webfont" to build the fonts directory.
     It requires fontforge and ttfautohint distro packages

     partial revert of commit 7137d28

I don't know why, but the same commit dca3bcca9 [removes](https://github.com/searxng/searxng/commit/dca3bcca9e1917dd3874ba9e40e1055e32caba34#diff-ffd563c55bf191e4f0f3bdc162add389733738c2e88c5a3e9c22d54585f7be69) the fontforge &
ttfautohint packages which has been added in 7137d28.  I assume it was an
mistake of the *partial revert of commit 7137d28*.

    make clean node.env

## Author's checklist

Lets have a look if github integration ...

https://github.com/searxng/searxng/blob/34ffd08187d49b2e5de387f63662c18fda800064/.github/workflows/integration.yml#L51-L60

.. continues to have an [issue with gyp](https://github.com/searxng/searxng/runs/3219005909?check_suite_focus=true#step:4:48)

With commit e09fd43 d60f957  [issue with gyp still exists](https://github.com/searxng/searxng/pull/248/checks?check_run_id=3231685008#step:4:418)

## Related issues

Related-to:  https://github.com/searxng/searxng/pull/157#discussion_r654836723

### DIFF
--- a/docs/dev/quickstart.rst
+++ b/docs/dev/quickstart.rst
@@ -5,6 +5,8 @@ Development Quickstart
 ======================
 
 .. _npm: https://www.npmjs.com/
+.. _fontforge: https://github.com/fontforge/fontforge
+.. _ttfautohint: https://www.freetype.org/ttfautohint/doc/ttfautohint.html
 
 Searx loves developers, just clone and start hacking.  All the rest is done for
 you simply by using :ref:`make <makefile>`.
@@ -31,7 +33,12 @@ If you implement themes, you will need to compile styles and JavaScript before
 
    make themes.all
 
-Don't forget to install npm_ first.
+Don't forget to install npm_, ttfautohint_ and fontforge_ first.  To install
+all system requirements of a :ref:`buildhosts` use::
+
+  sudo -H ./utils/searx.sh install buildhost
+
+Otherwise, install only what you need at least:
 
 .. tabs::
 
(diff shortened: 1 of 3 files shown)


return42:
grunt-webfont  is still problematic /  [issue with gyp still exists](https://github.com/searxng/searxng/pull/248/checks?check_run_id=3231685008#step:4:418)

grunt-webfont implements not much more than a call of the https://fontforge.org/docs/scripting/python.html API.

May be it is better to  call this API directly from our virtualenv instead of .. [start a python process to start a grunt process to start a node implementation, which starts a python application which starts binaries tools from the OS](https://github.com/searxng/searxng/pull/157#discussion_r654836723)

Further see: 

- https://github.com/searxng/searxng/pull/157#discussion_r655234695
- https://github.com/searxng/searxng/pull/157#discussion_r654828362

dalf:
I end up to think that a webfont is not the solution because the tools to build it are not updated / maintained.

An alternative solution can be css with embedded svg.

PR #253, opened 2021-08-07
[fix] settings.yml - replace http:// URLs by https://

return42:
## What does this PR do?

[fix] settings.yml - replace http:// URLs by https://


dalf:
From https://github.com/searx/searx/pull/2922

PR #254, opened 2021-08-08
[experimental] replace httpx by aiohttp

dalf:
## What does this PR do?

Replace httpx by aiohttp

Partial change:
* aiohttp doesn't support HTTP digest require by one engine
* test not modified ( aiounittest doesn't create a asyncio task which breaks an aiohttp requirement)

Experimental change

`make run` / `python -m searx.webapp`


dalf:
The current image proxy seems reliable and fast enough even on public instances, so I close this PR.

PR #256, opened 2021-08-14
[experimental] Starlette

dalf:
## What does this PR do?

Replace Uwsgi+Flask by Starlette+Uvicorn ( + Gunicorn in production mode see `searx/run.py`).
It seems the image proxy is fast enough to drop Morty.

https://github.com/searxng/searxng/wiki/Milestones#milestone-12---async

```
./local/py3/bin/activate
SEARX_DEBUG=1 python -m searx.webapp
```


dalf:
It seems to work, but according to https://github.com/encode/starlette/issues/279#issuecomment-505068010 starlette requires to drop Babel. I think [Quart](https://pgjones.gitlab.io/quart/) is a better solution if we want to go into the async direction.

PR #257, opened 2021-08-16
Mod image proxy

dalf:
## What does this PR do?

Image proxy:
* make sure the httpx connection is closed even if there is an error.
* use HTTP/1 instead of HTTP/2 : lower CPU load.
* don't send If-Modified-Since and If-None-Match for privacy reason.

See above: privacy, more reliable.

search for `!images anything`

### DIFF
--- a/searx/network/__init__.py
+++ b/searx/network/__init__.py
@@ -5,6 +5,7 @@
 import asyncio
 import threading
 import concurrent.futures
+from types import MethodType
 from timeit import default_timer
 
 import httpx
@@ -161,6 +162,7 @@ def patch(url, data=None, **kwargs):
 def delete(url, **kwargs):
     return request('delete', url, **kwargs)
 
+
 async def stream_chunk_to_queue(network, queue, method, url, **kwargs):
     try:
         async with network.stream(method, url, **kwargs) as response:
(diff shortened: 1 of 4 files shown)


return42:
I assume this is not about morty, right?  If I am not wrong we need to test image_proxy from

https://github.com/searxng/searxng/blob/9667142d5c45ae52f19b361e69a61900e09fe005/searx/webapp.py#L1082-L1083

and to test, we need to set `image_proxy = true` in the ...

https://github.com/searxng/searxng/blob/9667142d5c45ae52f19b361e69a61900e09fe005/searx/settings.yml#L44

dalf:
[EDIT] Yes just set `image_proxy: true` in settings.yml

---

To use the `image_proxy` endpoint:
* comment out https://github.com/searxng/searxng/blob/9667142d5c45ae52f19b361e69a61900e09fe005/searx/settings.yml#L93-L97
* enable the image proxy in the preferences

See:
https://github.com/searxng/searxng/blob/9667142d5c45ae52f19b361e69a61900e09fe005/searx/webapp.py#L406-L412

return42:
> [EDIT] Yes just set image_proxy: true in settings.yml

sorry for my late response / this was what I had tested .. thanks for the merge!

PR #258, opened 2021-08-16
[mod] plugins: minor change

dalf:
## What does this PR do?

required attributes: display a different message when the attribute has the wrong type

Tiny change, make #256 minor simple.

## Related issues

Related #256

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -40,7 +40,8 @@ required_attrs = (('name', str),
                   ('default_on', bool))
 
 optional_attrs = (('js_dependencies', tuple),
-                  ('css_dependencies', tuple))
+                  ('css_dependencies', tuple),
+                  ('preference_section', str))
 
 
 class Plugin():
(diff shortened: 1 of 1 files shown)


searx/plugins/__init__.py
```diff
@@ -63,9 +64,17 @@ def register(self, *plugins, external=False):
...
-                if not hasattr(plugin, plugin_attr) or not isinstance(getattr(plugin, plugin_attr), plugin_attr_type):
+                if not hasattr(plugin, plugin_attr):
                     logger.critical('missing attribute "{0}", cannot load plugin: {1}'.format(plugin_attr, plugin))
                     exit(3)
+                attr = getattr(plugin, plugin_attr)
+                if not isinstance(attr, plugin_attr_type):
+                    type_attr = str(type(attr))
+                    logger.critical(
+                        'attribute "{0}" is of type {2}, must be of type {3}, cannot load plugin: {1}'
+                        .format(plugin_attr, plugin, type_attr, plugin_attr_type)
```
return42, line 75: Nit: I think string replacement by log-function arguments [`*args`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) consumes less resources when logging is disabled (since the formatted string it is _late evaluated_), while the format operation of the string is always executed, even if logging is disabled.

return42 (approved):
Beside a _nit_ this PR LGTM.

searx/plugins/__init__.py
dalf: Yes, clearly.
`sys.exit()`  just after "hides / solves" the issue.

PR #259, opened 2021-08-16
[fix] searx.search.checker.get_result() always return a dict

dalf:
## What does this PR do?

searx.search.checker.get_result() always return a dict `{ 'status': '....' ... }`
So `checker_results['status'] == 'ok'` is enough to check the checker result.
See https://github.com/searxng/searxng/blob/9667142d5c45ae52f19b361e69a61900e09fe005/searx/webapp.py#L981

`/preferences` endpoint: avoid crash in some rare cases.


return42 (approved):
LGTM

PR #260, opened 2021-08-17
[mod] searx.webapp.get_locale: read locale from the preferences

dalf:
## What does this PR do?

pre_request already set the locale.
simplify `searx.webapp.get_locale` function.

Read the preferences in one place: pre_request function.

* clear the cookie: locale and language must follow the browser preferences.
* save the preferences with a different locale: check the locale has changed.
* make a request with `?locale=oc`: check the locale has changed to `oc`.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -223,21 +223,28 @@ exception_classname_to_text = {
     'lxml.etree.ParserError': parsing_error_text,
 }
 
-_flask_babel_get_translations = flask_babel.get_translations
-
 
 # monkey patch for flask_babel.get_translations
+_flask_babel_get_translations = flask_babel.get_translations
 def _get_translations():
     if has_request_context() and request.form.get('use-translation') == 'oc':
         babel_ext = flask_babel.current_app.extensions['babel']
         return Translations.load(next(babel_ext.translation_directories), 'oc')
     return _flask_babel_get_translations()
-
-
 flask_babel.get_translations = _get_translations
 
 
-def _get_browser_or_settings_language(req, lang_list):
+@babel.localeselector
+def get_locale():
+    locale = request.preferences.get_value('locale') if has_request_context() else 'en'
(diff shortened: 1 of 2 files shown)


searx/webapp.py
```diff
@@ -247,37 +254,7 @@ def _get_browser_or_settings_language(req, lang_list):
         locale = match_language(lang, lang_list, fallback=None)
         if locale is not None:
             return locale
-    return settings['search']['default_lang'] or 'en'
-
-
-@babel.localeselector
-def get_locale():
-    if 'locale' in request.form\
```
dalf, line 255: See pre_request function:
https://github.com/searxng/searxng/blob/a925254c3d19c6a3badf43469816fc1b497e2990/searx/webapp.py#L530-L542

If `locale in request.form` then the value has been parsed, and `request.preferences.get_value('locale')` contains the same value.

searx/webapp.py
```diff
@@ -247,37 +254,7 @@ def _get_browser_or_settings_language(req, lang_list):
...
-       and request.form['locale'] in LOCALE_NAMES:
-        # use locale from the form
-        locale = request.form['locale']
-        locale_source = 'form'
-    elif request.preferences.get_value('locale') != '':
```
dalf, line 260: See pre_request:
https://github.com/searxng/searxng/blob/a925254c3d19c6a3badf43469816fc1b497e2990/searx/webapp.py#L547-L548

If `not request.preferences.get_value('locale')` then the value comes from `get_locale()` which return `_get_browser_or_settings_language(request, UI_LOCALE_CODES)` in this case.

return42 (approved):
Thanks for simplifying code and the detail explanations.

> make a request with ?locale=oc: check the locale has changed to oc

My browser locale is 'de' and I tested with `?locale=en` .. works like a charm .. perfect!

PR #261, opened 2021-08-17
[upd] upgrade httpx 0.19.0

dalf:
## What does this PR do?

adjust searx.network module to the new internal API
see https://github.com/encode/httpx/pull/1522

There is a new dependency: [anyio](https://pypi.org/project/anyio/).

See https://github.com/encode/httpcore/pull/345

Upgrade to the last version of httpx which fixes some bugs.

* AnyIO 3.3.1 fixes issue https://github.com/encode/httpcore/issues/382
* remove `except  httpcore.CloseError`: closed by https://github.com/encode/httpcore/pull/310

Up to date HTTP client.

## Incoming change from httpx 1.0.0

* httpx: https://github.com/encode/httpx/pull/1840
* httpx: https://github.com/encode/httpx/pull/1808

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,10 +7,10 @@ lxml==4.6.3
 pygments==2.10.0
 python-dateutil==2.8.2
 pyyaml==5.4.1
-httpx[http2]==0.17.1
+httpx[http2]==0.19.0
 Brotli==1.0.9
 uvloop==0.16.0; python_version >= '3.7'
 uvloop==0.14.0; python_version < '3.7'
-httpx-socks[asyncio]==0.3.1
+httpx-socks[asyncio]==0.4.1
 langdetect==1.0.9
 setproctitle==1.2.2
--- a/searx/metrics/error_recorder.py
+++ b/searx/metrics/error_recorder.py
@@ -74,9 +74,11 @@ def get_request_exception_messages(exc: HTTPError)\
     status_code = None
     reason = None
     hostname = None
-    if hasattr(exc, 'request') and exc.request is not None:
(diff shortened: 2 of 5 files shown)


return42:
@dalf what do you think, should we priorize this PR (upgrading httpx  and wait until the issue of anyio is fixed) over  PR https://github.com/searxng/searxng/pull/254 ?

dalf:
`aiohttp` doesn't support HTTP Digest Authentification (even with a third party), it is a problem for the yacy engine: https://github.com/searxng/searxng/blob/33ea1d59096c08cc5b5045f975525d311d1960d1/searx/engines/yacy.py#L10

But aiohttp:
* may allow to drop the morty dependency in a private instance  (with or without combine with starlette, I'm not sure).
* is (often|always) faster than httpx: 
  * https://github.com/encode/httpx/issues/838#issuecomment-598919546
  * https://github.com/dalf/pyhttp-benchmark

About httpx and HTTP/2:
* It may actually slow down the response time compare to httpx + HTTP/1.1
* When the requests are sent through a proxy using HTTP/2, searxng uses a lot of more memory (is it a memory leak?) and the load average increase too.
* I _feel_ the reason is the complexity of the HTTP/2 protocol + Python speed (and perhaps async).

---

About this PR: I'm favor too way a little more:
* the version 1.0.0 will be released in a weeks.
* it seems there is no issue with the current version.
* in the meantime test the aiohttp branch.

---

About the choice between httpx or aiohttp: 
* keep httpx for now
* later, why not both ? we already have C extensions, and aiohttp has wheels: https://pypi.org/project/aiohttp/#files

dalf:
tested on my instance without issue.

PR #275, opened 2021-08-28
[enh] Add Packagist search engine

koelle25:
image

## What does this PR do?

This adds [Packagist](https://packagist.org) as a search engine. I based this engine on the existing Docker Hub engine, the engine's favicon was downloaded from the Packagist website with `wget` and converted to a PNG with ImageMagick, and the shortcut was plucked from [DDG's bangs database](https://duckduckgo.com/bang?q=packagist). It supports the parsing of URLs, titles, content, and published dates of Composer packages.

As a regular user of the Packagist repository myself, I'm sure many developers would find the direct ability to search for Composer packages very convenient.

This PR is fully compatible with `make test`. To see it in action:

- `make run`
- query `!pack polyfill-php80`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -938,6 +938,26 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name: packagist
+    engine: json_engine
+    paging: true
+    search_url: https://packagist.org/search.json?q={query}&page={pageno}
+    results_query: results
+    url_query: url
+    title_query: name
+    content_query: description
+    categories: it
+    disabled: true
+    timeout: 5.0
+    shortcut: pack
+    about:
+      website: https://packagist.org
+      wikidata_id: Q108311377
+      official_api_documentation: https://packagist.org/apidoc
+      use_official_api: true
+      require_api_key: false
+      results: JSON
+
   - name: pdbe
     engine: pdbe
     shortcut: pdb


searx/engines/packagist.py
```diff
@@ -0,0 +1,78 @@
...
+    for item in body.get("results"):
+        result = {}
+
+        url = item.get("url")
+
+        result["url"] = url
+        result["title"] = item.get("name")
+        result["content"] = item.get("description")
+
+        result.update(metadata(url))
```
return42: @dalf what do you think, from a http process model view, is it OK to send additional HTTP requests here in the response function?  If not, what might be an alternative.

searx/engines/packagist.py
```diff
@@ -0,0 +1,78 @@
...
+paging = True
+
+base_url = "https://packagist.org/"
+search_url = base_url + "search.json?{query}&per_page=25"
+
+def metadata(package_url):
+    result = {}
+    body = loads(requests.get(package_url + ".json").text)
+
+    package = body.get("package")
```
return42: @koelle25 what do you think, does it make sense to collect some more additional data (and ad it to the `result["content"]`) .. e.g. `repository`.

grafik

return42 (commented):
@koelle25 thanks a lot for your contribution / BTW excellent work, I would say.

HINT: I rebased your [branch add-packagist](https://github.com/koelle25/searxng/commits/add-packagist)  on top of master.

@dalf & @koelle25  could you have a look at my comments / thanks!

searx/engines/packagist.py
dalf: It is not a problem, and it is already the case for different engines:
* openstreetmap (call to wikidata = another HTTPS request)
* seznam: additional HTTPS request in the `request` function
* some others.

The only issue here is the number of requests: one per result.

All requests are sequential, so it can be slow and timeout.

searx/engines/packagist.py
```diff
@@ -0,0 +1,78 @@
...
+# pylint: disable=missing-function-docstring
+"""Packagist (IT)
+
+"""
+
+from json import loads
+from urllib.parse import urlencode
+from dateutil import parser
+
+import requests
```
dalf: Can you replace by `import searx.network` ? searx(ng) use httpx and not requests.

And then line 38, replace 
`requests.get(package_url + ".json")`
by
`searx.network.get(package_url + ".json")`

So it uses the same HTTP connection.

Even more be simplified:
```python
body = searx.network.get(package_url + ".json").json()
```

So there is no need to import `from json import loads`.

return42: > Can you replace by

See e3c1286 

> So there is no need to import `from json import loads`.

Nit: still used in the response function  / `body = loads(resp.text)`

searx/engines/packagist.py
return42: > All requests are sequential, so it can be slow and timeout.

From chapter ["Best practices when using the Packagist.org API"](https://packagist.org/apidoc):

    If you send requests in parallel, be a good citizen and do a maximum of 10 concurrent requests,
    or up to 20 if you only fetch static files. 

@dalf: do you have a suggestion to parallelize HTTP requests (with searx.network from e3c1286)?

dalf: > do you have a suggestion to parallelize HTTP requests (with searx.network from e3c1286)?

Switch to async, only for the engines processing, so it is possible to send multiple requests and then wait the result concurrently.

Without async: 10 new threads, but 5 requests to this engines = 55 threads (one per requests, one per result for each requests). It is not workable.

koelle25:
Well, I just found an easier way to implement this functionality :sweat_smile: by using the `engine: json_engine` (just like it is done with _npm_).

Should I close this PR and open another one? Or should I push the changes here?

dalf:
You can push the changes here.

return42:
@koelle25 I squashed your branch and merged it on master / thanks!

PR #289, opened 2021-09-04
[fix] remove minimum length of content for XPath engine

return42:
## What does this PR do?

Instead of raising an exception and therefore hiding all results of the engine, 
remove minimum length of content for XPath engine

It make sense to remove that requirement in order to allow the implementation of
search engines that do not always have a description.  In fact some search
engines that in 99% of the case have a description like Brave Search or Mojeek
crash completely if they for some reason included a result with no description.

Suggested-by: 0xhtml in https://github.com/searx/searx/discussions/2933

To test this patch try Mojeek:

    !mjk xyz

before and after the patch.

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -194,7 +194,7 @@ def response(resp):
 
             url = extract_url(eval_xpath_list(result, url_xpath, min_len=1), search_url)
             title = extract_text(eval_xpath_list(result, title_xpath, min_len=1))
-            content = extract_text(eval_xpath_list(result, content_xpath, min_len=1))
+            content = extract_text(eval_xpath_list(result, content_xpath))
             tmp_result = {'url': url, 'title': title, 'content': content}
 
             # add thumbnail if available


dalf:
I agree, but it is not enough to make it work:
https://github.com/searxng/searxng/blob/39bbd15b39a7d7da49148bbb8079f974045e982d/searx/results.py#L190-L192

Should be something like this (not tested):
```python
 elif 'content' in result:
     result['content'] = result['content'] or ''
     if not isinstance(result['content'], str):
         logger.debug('result: invalid content: %s', str(result)) 
         error_msgs.add('invalid content')
```

return42:
> I agree, but it is not enough to make it work

It is enough to avoid an exception in the engine / logging errors in the ResultContainer (when the content is invalid) is still valid I think .. or I'm wrong?

return42:
@dalf you gave my last comment  a 👍🏻 , but review request is still pending / should I merge or do you still review?

dalf:
Thank you.
Merge, we can change results.py later.

Issue #57, opened 2021-05-05
Make HTML templates more CSP compliant


return42:
**What happened?**

We have a lot of not CSP compliant `style='...'` attributes (and element) in our HTML templates

This single style element is not CSP compliant:

https://github.com/searxng/searxng/blob/d36adfa59f242a8775ad74245c696d62b7727a36/searx/templates/oscar/base.html#L39-L44

But there is also a lot of `style='...'` attributes usage ...

**update:**: with patches from this PR https://github.com/searxng/searxng/pull/81 applied, the following remain ...
```
./oscar/preferences.html:25:<td class="{{ label }}" style="padding: 2px">{{- "" -}}
./oscar/preferences.html:29:        <span style="width: calc(max(2px, 100%*{{ (stats[engine_name].time / max_rate95)|round(3) }}))" class="stacked-bar-chart-median"></span>{{- "" -}}
./oscar/preferences.html:30:        <span style="width: calc(100%*{{ ((stats[engine_name].rate80 - stats[engine_name].time) / max_rate95)|round(3) }})" class="stacked-bar-chart-rate80"></span>{{- "" -}}
./oscar/preferences.html:31:        <span style="width: calc(100%*{{ ((stats[engine_name].rate95 - stats[engine_name].rate80) / max_rate95)|round(3) }})" class="stacked-bar-chart-rate95"></span>{{- "" -}}
./oscar/preferences.html:284:                                    <th scope="col" style="width: 10rem">{{ _("Selected language") }}</th>
./oscar/preferences.html:285:                                    <th scope="col" style="width: 10rem">{{ _("SafeSearch") }}</th>
./oscar/preferences.html:286:                                    <th scope="col" style="width: 10rem">{{ _("Time range") }}</th>
./oscar/preferences.html:288:                                    <th scope="col" class="text-right"  style="width: 7rem">{{ _("Max time") }}</th>
./oscar/preferences.html:289:                                    <th scope="col" class="text-right" style="width: 7rem">{{ _("Reliablity") }}</th>
./oscar/result_templates/default.html:19:        <img src="{{ image_proxify(result.img_src or result.thumbnail) }}" title="{{ result.title|striptags }}" style="width: auto; max-height: 60px; min-height: 60px;" class="col-xs-2 col-sm-4 col-md-4 result-content">
./oscar/result_templates/map.html:64:        <div style="height:300px; width:100%; margin: 10px 0;" id="osm-map-{{ index }}"></div>
[...]

PR #292, opened 2021-09-05
[fix] make HTML templates more CSP compliant

return42:
## What does this PR do?

Make HTML templates more CSP compliant, for details read commit messages.

Reported-by: #57

1. open preference page 
2. do a map query and open the box showing the map


searx/templates/oscar/preferences.html
```diff
@@ -27,9 +27,9 @@ <h5><a href="{{about.website}}" rel="noreferrer">{{about.website}}</a></h5>
     <span class="stacked-bar-chart-value">{{- stats[engine_name].time -}}</span>{{- "" -}}
     <span class="stacked-bar-chart" aria-labelledby="{{engine_name}}_chart" aria-hidden="true">
         {%- if max_rate95 is not none and max_rate95 > 0 -%}
-        <span style="width: calc(max(2px, 100%*{{ (stats[engine_name].time / max_rate95)|round(3) }}))" class="stacked-bar-chart-median"></span>{{- "" -}}
-        <span style="width: calc(100%*{{ ((stats[engine_name].rate80 - stats[engine_name].time) / max_rate95)|round(3) }})" class="stacked-bar-chart-rate80"></span>{{- "" -}}
-        <span style="width: calc(100%*{{ ((stats[engine_name].rate95 - stats[engine_name].rate80) / max_rate95)|round(3) }})" class="stacked-bar-chart-rate95"></span>{{- "" -}}
+        <div class="bar-chart-serie1 bar{{ (100 * (stats[engine_name].time / max_rate95))|round }}"></div>{{- "" -}}
```
dalf: Why use `bar-chart-serie1` instead of `stacked-bar-chart-median` ?

In `/stats`, `bar-chart-serie1` is the HTTP time.
Here it is the median time.

Same question for the following lines:
* `bar-chart-serie2` / `stacked-bar-chart-rate80`
* `bar-chart-serie3` / `stacked-bar-chart-rate95`

In this PR:
image

Using `stacked-bar-chart-*`:
image

UI choices:
* no color because the engine tab already contains a lot of data.
* highlight the median time (only lines for quartile 80 / 95)

dalf:
About the commit 9e30330 : why not use `make static.build.commit` ?

searx/templates/oscar/preferences.html
return42: > Why use bar-chart-serie1 instead of stacked-bar-chart-median ?

My fail, no reason for .. fixed it to use stacked-bar-*

return42:
> why not use make static.build.commit ?

old habits .. I am forgetful :)  .. fixed and used  `static.build.commit`

dalf:
No problem :-)

(The names `static.build.commit` and `themes.all` seems not related, not very helpful).

Tested: it works as expected.
Thank you.

return42:
> The names `static.build.commit` and `themes.all` seems not related, not very helpful

Hm, I assume it depends on the point of view; for me `themes.all` is just one step in the `static.build.commit`.  Both have their values; 

- `themes.all` is typically used in development circles
- `static.build.commit` is more a CI function which _build_ the `/static` folder and creates commits. One step (and ATM the only step) of the _build_ process is `themes.all`, may be there comes the day we need some more steps to build the `/static` folder.

dalf:
> static.build.commit is more a CI function

The github workflows don't call `static.build.commit`.

return42:
Sorry, I was misleading; I mean it is _like_ a CI function:

If we know a suitable trigger we can add it to the github integration workflow (may be with some modifications) like we have it done with the translations:

https://github.com/searxng/searxng/blob/a26e4eff10b8ed41d7012390d4efc575b9da8eb0/.github/workflows/integration.yml#L93-L94

[For each commit on the master branch of SearXNG origin the GitHub job babel / Update translations branch checks for updated translations.](https://searxng.github.io/searxng/dev/translation.html)

PR #293, opened 2021-09-06
[fix] translation: lock weblate only when there is an actual change of messages.pot

### LINKED ISSUE #290: github job "Update translations branch " locks and unlock weblate, even if there are no changes detected
The github job ..

https://github.com/searxng/searxng/blob/33ea1d59096c08cc5b5045f975525d311d1960d1/.github/workflows/integration.yml#L93-L94

locks and unlock weblate, even if there are no meaningful changes see ..
1. `wlc lock` https://github.com/searxng/searxng/runs/3513545417#step:6:12
2. `BABEL     no changes detected, exiting` https://github.com/searxng/searxng/runs/3513545417#step:6:236
3. `wlc unlock` https://github.com/searxng/searxng/runs/3513545417#step:6:240

The `wlc lock` comes from here ..

https://github.com/searxng/searxng/blob/33ea1d59096c08cc5b5045f975525d311d1960d1/manage#L244

and the `wlc unock` from here ..

https://github.com/searxng/searxng/blob/33ea1d59096c08cc5b5045f975525d311d1960d1/manage#L300

The lock and unlock on weblate could causes a cascade of actions on weblate (e.g. release some mails and may be other hooks)

dalf:
## What does this PR do?

After each commit on the master branch, there is a call to `weblate.push.translations`:
https://github.com/searxng/searxng/blob/a26e4eff10b8ed41d7012390d4efc575b9da8eb0/.github/workflows/integration.yml#L96
https://github.com/searxng/searxng/blob/a26e4eff10b8ed41d7012390d4efc575b9da8eb0/.github/workflows/integration.yml#L130

`weblate.push.translations` locks weblate each time.

This PR locks weblate only when there is an actual change of messages.pot.

See #290 : keep weblate history clean, avoid triggers on lock.

N/A

## Related issues

Close #290


manage
```diff
@@ -261,6 +259,12 @@ weblate.push.translations() {
             return 0
```
return42: If I'm not wrong, this return ends the subshell started in line 242 by `(` .  The shell will continue where the subshell ends, see closing  `)` below .. and `pyenv.cmd wlc unlock` will be executed .. or I'm wrong?

**update:** compare with https://github.com/searxng/searxng/issues/290#issuecomment-912990501 where I moved the block with the _return_ out of the subprocess.

return42:
Does https://github.com/searxng/searxng/blob/master/docs/dev/translation.svg needs an update?

return42 (commented):
See my comments.

manage
dalf: How to keep `set -e` for the block before `pyenv.cmd wlc lock` ?

dalf:
> Does https://github.com/searxng/searxng/blob/master/docs/dev/translation.svg needs an update?

The check is not shown, and it is not accurate.
It should be better with this:

image

The creation of the worktree / git pull are not shown (it is already the case).

return42:
> It should be better with this:

Do you like to update the svg in the repository along with this PR?

manage
return42: Try this patch ... **update** see https://github.com/searxng/searxng/pull/379

```diff
diff --git a/manage b/manage
index 6f6f1a4f..488a93cc 100755
--- a/manage
+++ b/manage
@@ -240,9 +240,6 @@ weblate.push.translations() {
     local messages_pot diff_messages_pot last_commit_hash last_commit_detail \
           last_commit_message exitcode
     (   set -e
-        # lock change on weblate
-        pyenv.cmd wlc lock
-
         # get translations branch in git worktree (TRANSLATIONS_WORKTREE)
         weblate.translations.worktree
 
@@ -258,8 +255,18 @@ weblate.push.translations() {
                             git diff -- "searx/translations/messages.pot")
         if ! echo "$diff_messages_pot" | grep -qE "[\+\-](msgid|msgstr)"; then
             build_msg BABEL 'no changes detected, exiting'
-            return 0
+            return 1
         fi
+        return 0
+    )
+    exitcode=$?
+    if [ "$exitcode" ]; then # FIX see #379 :: if [ "$exitcode" -gt 0 ]; then
+       return 0
+    fi
+
+    (   set -e
+        # lock change on weblate
+        pyenv.cmd wlc lock
 
         # save messages.pot in the translations branch for later
         pushd "${TRANSLATIONS_WORKTREE}"
```

return42 (approved):
Cant test it, but LGTM / Thanks!

PR #296, opened 2021-09-06
one logger per engine

return42:
## What does this PR do?

Implements _one logger per engine_ requested in https://github.com/searxng/searxng/issues/98 

## How to test this PR locally?

    make run

Run some queries and have a look at the debug messages..

`!google_videos foo` (python module)

    DEBUG:searx.engines.google videos:HTTP header Accept-Language --> de-DE,de;q=0.8,,en;q=0.6,*;q=0.5

`!brave foo` (XPath engine)

    DEBUG:searx.engines.brave:found 20 results

## Related issues

Rleated: https://github.com/searxng/searxng/issues/98

PR #297, opened 2021-09-06
debug mode: more readable logging

dalf:
## What does this PR do?

* In development add a new dependency: [coloredlogs](https://pypi.org/project/coloredlogs/) (declared in `requirements-dev.txt` but optional in the code, see `searx/__init__.py`)
* Logs about an engine always use the engine logger ( #296 ), include log from `searx.search.processors` and `searx.metrics`.
* HTTP request are logged with the network name: a wikipedia HTTP request is logged with `searx.network.wikipedia` instead of `httpx._client`.

Make the debug log more readable:

image

A lot of information are logged in debug mode, which are not easy to read.

But it makes messages more difficult to find in the code:
image

But a `git grep` is easy:
```
$ git grep "engine time"
searx/search/__init__.py:        # max of all selected engine timeout
searx/search/__init__.py:                    PROCESSORS[th._engine_name].logger.error('engine timeout')
```

* `make run`
* production mode:
   * change the secret key
   * start in production mode
* check the log with uwsgi
* check the log with docker

Draft because 
* HTTP redirection are not logged (fixed in https://github.com/encode/httpx/pull/1806 --> we need to upgrade httpx to a new version).
* HTTP timeout doesn't log the URL (it was not the case before ?).
* it may requires some adjustment (or close the PR).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ run:  install
 	sleep 2 ; \
 	xdg-open http://127.0.0.1:8888/ ; \
 	) &
-	SEARX_DEBUG=1 ./manage pyenv.cmd python ./searx/webapp.py
+	SEARX_DEBUG=1 ./manage pyenv.cmd python -m searx.webapp
 
 PHONY += install uninstall
 install uninstall:
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -17,3 +17,4 @@ linuxdoc==20210324
 aiounittest==1.4.0
 yamllint==1.26.3
 wlc==1.12
+coloredlogs==15.0.1
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -2,12 +2,22 @@
 # lint: pylint
 # pylint: disable=missing-module-docstring
 
+import sys
+import os
 from os.path import dirname, abspath
+
 import logging
 
 import searx.unixthreadname
 import searx.settings_loader
 from searx.settings_defaults import settings_set_defaults
+
+
+# Debug
+LOG_FORMAT_DEBUG = '%(levelname)-7s %(name)-30.30s: %(message)s'
+
+# Production
+LOG_FORMAT_PROD = '%(asctime)-15s %(levelname)s:%(name)s: %(message)s'
 
 searx_dir = abspath(dirname(__file__))
 searx_parent_dir = abspath(dirname(dirname(__file__)))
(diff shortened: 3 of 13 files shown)


dalf:
@return42 @unixfox what are you opinions on this type of log?

return42:
It seems `coloredlogs` is active in _dumb_ terminals ...

grafik

In the lib.sh we use ..

https://github.com/searxng/searxng/blob/9ef7f38e4f20bad5a76867aa3e5049a92bc58e59/utils/lib.sh#L112-L114

to detect _pipes_ and _dumb_ terminals ...

Is there something similar we can do when initialize `coloredlogs`?

dalf:
I can't find something about this, but this function should be enough:

```python
import os
import sys

def is_color_terminal():
    if os.getenv('TERM', None) in ('dumb', 'unknown'):
        return False
    return sys.stdout.isatty()
```

dalf:
Updated: 
* `make run | cat` works without color.
* `TERM=dumb make run` works too.

unixfox:
Looks cool, I like it.

return42 (approved):
Tested / much better to read logs, like it .. Thanks!

return42 (approved):
@dalf you force pushed but did not ask for a review .. can we merge this PR anyway?

For me this PR works and I like to use this logging in my development :)

dalf:
About push force fix this method:
https://github.com/searxng/searxng/blob/6abc37efc75a472b977306e5352c10358f3a1d08/searx/network/network.py#L137-L147

Before the push force, the embedded image proxy didn't work because `log_response` was expecting a value for `response.elapsed`. Since the response is streamed, there is no value.

The new value takes of that.

---

About merging: everything works except one thing: HTTP redirect are not logged anymore.
See httpx PR : https://github.com/encode/httpx/pull/1806
We need to upgrade to httpx 0.19.0 or later --> https://github.com/searxng/searxng/pull/261

If the missing log are not an issue (it is temporary), we can merge.

return42:
> If the missing log are not an issue (it is temporary), we can merge.

Thanks for the info, then I think it is better you decide if we should merge :)

dalf:
Ok I'm going to merge.

---

httpx uses [rich](https://rich.readthedocs.io/en/stable/logging.html) was has nice traceback logger, but it requires more configuration to fit into SearXNG.

But it is a dev dependency we can change later.

Issue #284, opened 2021-09-03
Backport of searx/plugins/hostname_replace.py


dalf:
See 
* https://github.com/searx/searx/blob/master/searx/plugins/hostname_replace.py
* https://github.com/searx/searx/pull/2724

A quick win: use `false` to remove the host name.
```
hostname_replace:
  www.libhunt.com: false
```

return42:
I'm not happy with https://github.com/searx/searx/pull/2724 there is also https://github.com/searx/searx/pull/2638 but  I would suggest to implement something complete new like I suggested in https://github.com/searx/searx/pull/2724#issuecomment-813393855

dalf:
About the syntax: I prefer to have rewrite rules per URL component (host name, query, path) : it avoids the issue describe just after.

> A list of expressions can be configured in settings.yml at engine level

What is the use case ? About the give example (rewrite the result of the youtube engine), it won't work if an another engine returns a youtube links.

return42:
> About the syntax: I prefer to have rewrite rules per URL component (host name, query, path) 

Good point!

> it won't work if an another engine returns a youtube links.

Hm .. can't remember, now where you point it out I can't see any proper use case. May be I have not thought to the end while I have written this comment 🤦🏻 / lets forget rewrite on engine level.

dalf:
Possible syntax to combine both:

```yaml
rewrite_result:
    hostname:
      - '^([a-z]{2,3})?.m.wikipedia.org$': '\1.wikipedia.org'
      - '(.*\.)?youtube\.com$': 'invidiou.site'
      - '(.*\.)?youtu\.be$': 'invidiou.site'
      - '(.*\.)?youtube-noocookie\.com$': 'invidiou.site'
      - '(.*\.)?libhunt.com': false  # remove the result
    url:  # dangerous settings
      - '^https://twitter\.com/([^\\?]*)?\?lang=.*$': 'https://nitter.net/\1'
```

return42:
I assume we use attribute names from [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) .. right? Should we have the other attributes also configurable? .. is what I would suggest.

dalf:
The other fields:
- `scheme` : ? not sure, this is related https everywhere (BTW: https://www.eff.org/deeplinks/2021/04/https-everywhere-now-uses-duckduckgos-smarter-encryption )
[...]

PR #302, opened 2021-09-07
[mod] plugin: call on_result for each result of each engines.

dalf:
## What does this PR do?

Currently, searx.search.Search calls on_result once the engine results have been merged
(ResultContainer.order_results).

on_result plugins can rewrite the results: once the URL(s) are modified, even they can be merged,
it won't be the case since ResultContainer.order_results has already be called.

This commit call on_result for each result of each engines.
In addition the on_result function can return False to remove the result.

Note: the on_result function now run on the engine thread instead of the Flask thread.

With an implementation of https://github.com/searxng/searxng/issues/284 some identical results won't be merged.
It is already the case in the science category when the "Open Access DOI rewrite" plugin is enabled.

---

Alternative to filter out some result:
https://github.com/searxng/searxng/blob/a3789b3bb44fd929a730518da678b7754a2c18cb/searx/plugins/ahmia_filter.py#L23-L33

With this PR: https://github.com/dalf/searxng/blob/2b96ace61118d0a94eb74f6d078728a0c73276de/searx/plugins/ahmia_filter.py : 
```python
def on_result(request, search, result):
    if not result.get('is_onion') or not result.get('parsed_url'):
        return True
    result_hash = md5(result['parsed_url'].hostname.encode()).hexdigest()
    return result_hash not in get_ahmia_blacklist()
```

### DIFF
--- a/docs/dev/plugins.rst
+++ b/docs/dev/plugins.rst
@@ -26,8 +26,8 @@ Example plugin
    # attach callback to the post search hook
    #  request: flask request object
    #  ctx: the whole local context of the post search hook
-   def post_search(request, ctx):
-       ctx['search'].suggestions.add('example')
+   def post_search(request, search):
+       search.result_container.suggestions.add('example')
        return True
 
 External plugins
@@ -50,20 +50,52 @@ Plugin entry points
 
 Entry points (hooks) define when a plugin runs. Right now only three hooks are
 implemented. So feel free to implement a hook if it fits the behaviour of your
-plugin.
+plugin. A plugin doesn't need to implement all the hooks.
 
-Pre search hook
----------------
 
-Runs BEFORE the search request. Function to implement: ``pre_search``
+.. py:function:: pre_search(request, search) -> bool
 
-Post search hook
-----------------
+   Runs BEFORE the search request.
 
-Runs AFTER the search request. Function to implement: ``post_search``
+   `search.result_container` can be changed.
 
-Result hook
------------
+   Return a boolean:
 
-Runs when a new result is added to the result list. Function to implement:
-``on_result``
+   * True to continue the search
+   * False to stop the search
+
+   :param flask.request request:
+   :param searx.search.SearchWithPlugins search:
+   :return: False to stop the search
+   :rtype: bool
+
+
+.. py:function:: post_search(request, search) -> None
+
+   Runs AFTER the search request.
+
+   :param flask.request request: Flask request.
+   :param searx.search.SearchWithPlugins search: Context.
+
+
+.. py:function:: on_result(request, search, result) -> bool
+
+   Runs for each result of each engine.
+
+   `result` can be changed.
+
+   If `result["url"]` is defined, then `result["parsed_url"] = urlparse(result['url'])`
+
+   .. warning::
+      `result["url"]` can be changed, but `result["parsed_url"]` must be updated too.
+
+   Return a boolean:
+
+   * True to keep the result
+   * False to remove the result
+
+   :param flask.request request:
+   :param searx.search.SearchWithPlugins search:
+   :param typing.Dict result: Result, see - :ref:`engine results`
+   :return: True to keep the result
+   :rtype: bool
--- a/searx/results.py
+++ b/searx/results.py
@@ -145,7 +145,7 @@ class ResultContainer:
     """docstring for ResultContainer"""
 
     __slots__ = '_merged_results', 'infoboxes', 'suggestions', 'answers', 'corrections', '_number_of_results',\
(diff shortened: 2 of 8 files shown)


return42:
I think the prototype of the `on_result` function should be described (what are the in and outs, what can be done with the arguments `(request, search, result)` / e.g developers might expect a documentation here in the [Result Hook chapter](https://github.com/searxng/searxng/blob/master/docs/dev/plugins.rst#result-hook)

searx/results.py
```diff
@@ -145,9 +145,9 @@ class ResultContainer:
     """docstring for ResultContainer"""
 
     __slots__ = '_merged_results', 'infoboxes', 'suggestions', 'answers', 'corrections', '_number_of_results',\
-                '_ordered', 'paging', 'unresponsive_engines', 'timings', 'redirect_url', 'engine_data'
+                '_closed', 'paging', 'unresponsive_engines', 'timings', 'redirect_url', 'engine_data', 'on_result'
 
-    def __init__(self):
+    def __init__(self, on_result=None):
```
return42: Sorry, I do not understand when is the argument on_result not None?  See my next question, it might be related

searx/results.py
```diff
@@ -156,43 +156,48 @@ def __init__(self):
...
+
         standard_result_count = 0
         error_msgs = set()
         for result in list(results):
             result['engine'] = engine_name
-            if 'suggestion' in result:
+            if 'suggestion' in result and self.on_result(result):
                 self.suggestions.add(result['suggestion'])
-            elif 'answer' in result:
+            elif 'answer' in result and self.on_result(result):
```
return42, line 176: Another point I do not understand, why `self.on_result(result)` is called, I would expect a call with the arguments needed by the plugin's `on_result(request, search, result)`.  May be I mix up things do not understand the concept in deep / sorry.

return42 (commented):
See my questions ..

searx/results.py
dalf: `git grep "ResultContainer"`: 
* the checker creates ResultContainer without usage of the plugins
* `tests/unit/test_results.py` should be updated.

return42: > git grep "ResultContainer

That is what I have done, but I find only constructors without arguments `ResultContainer()`

searx/results.py
dalf: https://github.com/dalf/searxng/blob/2b96ace61118d0a94eb74f6d078728a0c73276de/searx/search/__init__.py#L179-L193

* `ResultContainer.on_result = SearchWithPlugins_on_result(self, result)`
* `SearchWithPlugins_on_result(self, result)` calls `plugins.call(self.ordered_plugin_list, 'on_result', self.request, self, result)`

The idea is to keep everything about the plugins in one place: the SearchWithPlugins class.

Perhaps it should be done differently?

searx/results.py
dalf: True...

https://github.com/dalf/searxng/blob/2b96ace61118d0a94eb74f6d078728a0c73276de/searx/search/__init__.py#L182 : 
* `self.result_container.on_result = self._on_result`

The parameter `on_result` is not given because the `Search` constructor creates the `ResultContainer`.

In the current design the `on_result` parameter can be removed.

dalf:
I've updated the documentation:
image

docs/dev/plugins.rst
```diff
@@ -50,20 +50,52 @@ Plugin entry points
...
+   * True to continue the search
+   * False to stop the search
+
+   :param flask.request request:
+   :param searx.search.SearchWithPlugins search:
+   :return: False to stop the search
+   :rtype: bool
+
+
+.. py:function:: post_search(request, search, result) -> None
```
return42: post_search does not have a result argument.

return42 (changes requested):
I tested this PR with an external plugin:

    ./manage pyenv.cmd pip install git+https://github.com/return42/tgwf-searx-plugins

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index e5eb9dd6..1f7f35b6 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -140,8 +140,8 @@ outgoing:
 # External plugin configuration, for more details see
 #   https://searxng.github.io/searxng/dev/plugins.html
 #
-# plugins:
-#   - plugin1
+plugins:
+  - only_show_green_results
 #   - plugin2
 #   - ...
```

The plugin uses `list(search.result_container._merged_results)` ATM it seems downward compatible ...

https://github.com/return42/tgwf-searx-plugins/blob/646dae0dc2551e83ce6bc32a461eacc6f2e5df09/only_show_green_results.py#L103-L111

docs/dev/plugins.rst
dalf: Updated

return42 (approved):

PR #305, opened 2021-09-09
[fix] logger per engine: make .logger is always initialized

dalf:
## What does this PR do?
the openstreetmap engine imports code from the wikidata engine.
before this commit, specific code make sure to copy the logger variable to the wikidata engine.

with this commit `searx.engines.load_engine` makes sure the `logger` is initialized.
The implementation scans `sys.modules` for module name starting with `searx.engines`.

Without this PR, `import searx.engines.an_engine` may not work:
* `searx.engines.an_engine.logger` won't be initialized
* as soon as code in `searx.engines.an_engine` makes a reference to `searx.engines.an_engine.logger` the engine crash.

With this PR, `searx.engines.load_engine` makes sure the .logger is initialized.

See the result of `git grep "from searx.engines"`

---

About this line
https://github.com/searxng/searxng/blob/1973e4ecf691f173299fb11abc61750009bdef84/searx/engines/__init__.py#L95

For example the azlyrics engine, loads the `xpath` module.
The name in `sys.module` is `xpath` without the prefix `searx.engines`.

So, this line
https://github.com/searxng/searxng/blob/1973e4ecf691f173299fb11abc61750009bdef84/searx/engines/ebay.py#L7
imports a new copy of the `xpath` module.

That's why even if the logger for the wikidata engine is initialized (`engine.logger = ...`), when the OSM engine imports code from the wikidata engine, the logger is not initialized without special care.

* `!osm somewhere`

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -110,8 +110,24 @@ def load_engine(engine_data):
     if is_missing_required_attributes(engine):
         return None
 
-    engine.logger = logger.getChild(engine_name)
+    set_loggers(engine, engine_name)
+
     return engine
+
+
+def set_loggers(engine, engine_name):
+    # set the logger for engine
+    engine.logger = logger.getChild(engine_name)
+    # the engine may have load some other engines
+    # may sure the logger is initialized
+    for module_name, module in sys.modules.items():
+        if (
+            module_name.startswith("searx.engines")
+            and module_name != "searx.engines.__init__"
+            and not hasattr(module, "logger")
+        ):
+            module_engine_name = module_name.split(".")[-1]
+            module.logger = logger.getChild(module_engine_name)
 
 
 def update_engine_attributes(engine, engine_data):
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -438,8 +438,3 @@ def get_key_label(key_name, lang):
         if labels is None:
             return None
     return get_label(labels, lang)
-
-
-def init(_):
-    import searx.engines.wikidata  # pylint: disable=import-outside-toplevel
-    searx.engines.wikidata.logger = logger


return42:
> The name in sys.module is xpath without the prefix searx.engines

If so, I think the problem is not the initialization of the logger, the problem is that we have two copies of the engine module in the namespace of the runtime.

If I'm not wrong we have to fix ..

https://github.com/searxng/searxng/blob/1973e4ecf691f173299fb11abc61750009bdef84/searx/utils.py#L428-L438

The problem is, that  (e.g.) xpath is registered by `sys.modules['xpath'] = module` while the unique module name should be `sys.modules['searx.engines.xpath'] = module`

dalf:
> The problem is, that (e.g.) xpath is registered by sys.modules['xpath'] = module while the unique module name should be sys.modules['searx.engines.xpath'] = module

I've made a first implementation that does this, but I'm not sure this is what we want.
For example, searxng loads the module `searx.engines.xpath` many times: once per engine is settings.yml that uses the xpath engine. To sum up:

* The modules in `searx.engines.*` are like classes.
* The values of `searx.engines.engines` are instances.
* `searx.engines.load_engine` is the generic constructor.

From this, when ` from searx.engines.xpath import extract_text ` what should do?
* Import a random(*) instance ( `sys.modules['searx.engines.xpath'] = module` ). Random = the last engine declared in settings.yml
* Import the class (known state).

The later one works. So I think we can remove `sys.modules[modname] = module` from `load_module`.

return42:
Sorry, I wasn't aware about `searx.engines.engines` .. and now I understand the problem :-)

> To sum up:
> * The modules in `searx.engines.*` are like classes.
> * The values of `searx.engines.engines` are instances.

Hm, from my point both are _module namespaces_ ....

`from searx.engines import xpath` imports module xpath and register this _module namespaces_ in the _global namespace of the loaded modules_

```
>>> from searx.engines import xpath
>>> xpath
<module 'searx.engines.xpath' from '/800GBPCIex4/share/searx/searx/engines/xpath.py'>
>>> sys.modules['searx.engines.xpath']
<module 'searx.engines.xpath' from '...../searx/searx/engines/xpath.py'>
```

while `load_module()` just returns a _module namespaces_

```
>>> load_engines(settings['engines'])['azlyrics']
<module 'xpath' from '/800GBPCIex4/share/searx/searx/engines/xpath.py'>
```

> So I think we can remove `sys.modules[modname] = module` from `load_module`.

Yes we should not register in `sys.modules` what is loaded by `load_module()`. Or with other words: only imported _module namespaces_   should be registered  in `sys.modules`, not those are loaded.

But the https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly is just mystic for me :) .. I do not know what the methods from `import.utils` do _behind the scene_.

dalf:
About `load_module`, see issue #312

searx/engines/__init__.py
```diff
@@ -110,10 +110,24 @@ def load_engine(engine_data):
...
+    # set the logger for engine
+    engine.logger = logger.getChild(engine_name)
+    # the engine may have load some other engines
+    # may sure the logger is initialized
+    for module_name, module in sys.modules.items():
+        if module_name.startswith('searx.engines') \
+            and module_name != 'searx.engines.__init__' \
+            and not hasattr(module, 'logger'):
+            module_engine_name = module_name.split('.')[-1]
+            module.logger = logger.getChild(module_engine_name)
```
return42: I recommend to use parentheses ..

```diff
@@ -121,9 +121,10 @@ def set_loggers(engine, engine_name):
     # the engine may have load some other engines
     # may sure the logger is initialized
     for module_name, module in sys.modules.items():
-        if module_name.startswith('searx.engines') \
-            and module_name != 'searx.engines.__init__' \
-            and not hasattr(module, 'logger'):
+        if (module_name.startswith('searx.engines')
+            and module_name != 'searx.engines.__init__'
+            and not hasattr(module, 'logger')
+        ):
```

return42 (changes requested):
I tried to test before and after this patch has applied ...

> Without this PR, `import searx.engines.an_engine` may not work:
> * `searx.engines.an_engine.logger` won't be initialized
> * as soon as code in `searx.engines.an_engine` makes a reference to `searx.engines.an_engine.logger` the engine crash.

.. sounds logical for me, but I can't reproduce a crash (before this patch is applied). I used `!osm paris` to test.

dalf:
The OSM engine has a workaround that is removed by this PR: 
https://github.com/searxng/searxng/blob/c3473480f4255a81d83ac48201e0702778398d23/searx/engines/openstreetmap.py#L443-L445

If you remove these lines, `!osm paris` doesn't work.

searx/engines/__init__.py
dalf: updated

return42 (approved):

PR #306, opened 2021-09-09
[doc] searxng brand

dalf:
## What does this PR do?

image

Update copyright notice

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -9,9 +9,9 @@ from searx.version import VERSION_STRING, GIT_URL, GIT_BRANCH
 
 # Project --------------------------------------------------------------
 
-project = u'searx'
-copyright = u'2015-2020, Adam Tauber, Noémi Ványi'
-author = u'Adam Tauber'
+project = 'SearXNG'
+copyright = '2021 SearXNG team, 2015-2021 Adam Tauber, Noémi Ványi'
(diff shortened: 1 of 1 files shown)


return42:
How did you made the ["searXNG" logo](https://github.com/searxng/searxng/blob/a447b256156f621cf4d525f71f71fc001c17f98f/docs/static/img/searxng.png)?

- shouldn't it be "SearXNG"?
- The SearXNG logo should be stored in its origin format (e.g. SVG)
- I think the typography is improvable, the proportions of the letters are not right and there are hard breaks like this one
grafik

If you want, I can patch on top of your branch what I think improves the typography. Unfortunately this might take some time since I do not have much time at the moment.

dalf:
> How did you made the "searXNG" logo?

Inkscape + a lot of search on Internet, since I don't know how to use it :-)

I can restore the logo, so
* we can merge this PR (if there are no other issues about it)
* take the time to create a logo

return42:
> I can restore the logo, so ... we can merge this PR (if there are no other issues about it)

Yes, lets do it that way.

dalf:
Done

dalf:
... and of course when I rewrote https://github.com/searxng/searxng/blob/13a4d20e8dc3fbad8f4094b318a96153d563823e/docs/conf.py#L135
I forgot to replace `Searx`  by `SearXNG`...

PR #311, opened 2021-09-10
[fix] searx/results.py: strip result['content'] only if it exists

dalf:
## What does this PR do?

fix PR #302

otherwise an exception is raised, and the results of the engine are ignored.

code before PR #302 
https://github.com/searxng/searxng/blob/ec834935381e5cfb4ebc3820ed2f668d0fc7855d/searx/results.py#L239-L241


return42 (approved):
:-)

PR #314, opened 2021-09-12
Engine stats link

mrpaulblack:
## What does this PR do?

This PR adds a new Engine stats link to the footer and makes the engine stats headline clickable (when viewing stats by one specific engine).

## Why is this PR important?

This makes makes it easier for people to find the stats page (which can right now only be accessed by knowing the /stats path or when an engine had an error through the info boxes; and even then it does not show the full stats page).

```make run```


return42 (approved):
Thanks! .. a link I missed since a long time :)

PR #318, opened 2021-09-12
[brand] docs - normalize project name to SearXNG

return42:
## What does this PR do?

In the documentation, normalize project name to SearXNG.

### DIFF
--- a/docs/admin/engines/nosql-engines.rst
+++ b/docs/admin/engines/nosql-engines.rst
@@ -42,7 +42,7 @@ Extra Dependencies
 ------------------
 
 For using :ref:`engine redis_server` or :ref:`engine mongodb` you need to
-install additional packages in Python's Virtual Environment of your SearxNG
+install additional packages in Python's Virtual Environment of your SearXNG
 instance.  To switch into the environment (:ref:`searx-src`) you can use
 :ref:`searx.sh`::
 
--- a/docs/admin/engines/sql-engines.rst
+++ b/docs/admin/engines/sql-engines.rst
@@ -10,7 +10,7 @@ SQL Engines
    - `PostgreSQL <https://www.postgresql.org>`_
    - `MySQL <https://www.mysql.com>`_
 
-With the *SQL engines* you can bind SQL databases into SearxNG.  The following
+With the *SQL engines* you can bind SQL databases into SearXNG.  The following
 Relational Database Management System (RDBMS) are supported:
 
 - :ref:`engine sqlite`
@@ -98,7 +98,7 @@ Extra Dependencies
 ------------------
 
 For using :ref:`engine postgresql` or :ref:`engine mysql_server` you need to
-install additional packages in Python's Virtual Environment of your SearxNG
+install additional packages in Python's Virtual Environment of your SearXNG
 instance.  To switch into the environment (:ref:`searx-src`) you can use
 :ref:`searx.sh`::
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,5 @@
 ==================
-Welcome to searxng
+Welcome to SearXNG
 ==================
 
     *Search without being tracked.*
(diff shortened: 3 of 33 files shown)


docs/admin/engines/nosql-engines.rst
```diff
@@ -42,7 +42,7 @@ Extra Dependencies
 ------------------
 
 For using :ref:`engine redis_server` or :ref:`engine mongodb` you need to
-install additional packages in Python's Virtual Environment of your SearxNG
+install additional packages in Python's Virtual Environment of your SearXNG
 instance.  To switch into the environment (:ref:`searx-src`) you can use
```
dalf, line 46: `searx-src` --> `searxng-src`.

docs/index.rst
```diff
@@ -1,19 +1,15 @@
 ==================
-Welcome to searxng
+Welcome to SearXNG
 ==================
 
     *Search without being tracked.*
 
-.. hint::
-
-   This is not searx, but searxng.
```
dalf, line 9: I would keep this notice especially since the logo is not different.

dalf:
https://github.com/searxng/searxng/blob/823a2fc7926441ba4375d74340a7ab08ac85ef9d/docs/conf.py#L135

`Searx Documentation` --> `SearXNG Documentation` (this title is referenced by search engines).

docs/admin/engines/sql-engines.rst
```diff
@@ -98,7 +98,7 @@ Extra Dependencies
 ------------------
 
 For using :ref:`engine postgresql` or :ref:`engine mysql_server` you need to
-install additional packages in Python's Virtual Environment of your SearxNG
+install additional packages in Python's Virtual Environment of your SearXNG
 instance.  To switch into the environment (:ref:`searx-src`) you can use
```
dalf, line 102: `searx-src` --> `searxng-src`

dalf:
expect few tiny things, LGTM.

docs/admin/engines/nosql-engines.rst
return42: This is a reference to ..

https://github.com/searxng/searxng/blob/823a2fc7926441ba4375d74340a7ab08ac85ef9d/docs/admin/installation-searx.rst#L37-L40

IMO it is better we do not change anchors, otherwise we have to change all references (and [intersphinx](https://certik.github.io/sphinx/ext/intersphinx.html) links from the outside would be broken).

*update*: The name `searx-src` comes from `/usr/local/searx/searx-src`

docs/admin/engines/sql-engines.rst
return42: See https://github.com/searxng/searxng/pull/318#discussion_r707076457

return42:
> expect few tiny things, LGTM.

Fixed your remarks / For a simple review I pushed a "WIP" commit on top .. before we merge I will squash the commits.

docs/admin/engines/nosql-engines.rst
dalf: If I understand correctly: 
* It is an internal reference
* It is an "issue" if some external pages make a reference to this anchor (the HTML anchor).

Am I missing something?

return42: > It is an internal reference

the _landing point_ is `.. _searx-src:` (see above)

> It is an "issue" if some external pages make a reference to this anchor (the HTML anchor).

Not the HTML anchor.  It is a bit OT since I suppose nobody from outside has linked this _landing point_, but to understand I will give a short introduce to what [intersphinx](https://certik.github.io/sphinx/ext/intersphinx.html) is ...

With intersphinx in your project you generate a `objects.inv` file which is also distributed, e.g here is our https://searxng.github.io/searxng/objects.inv

In this file there is a mapping from the name of the _landing point_ to the final URL.

Other projects can refer this file, like we do with some other (external) projects ...

https://github.com/searxng/searxng/blob/823a2fc7926441ba4375d74340a7ab08ac85ef9d/docs/conf.py#L92-L99

The object.inv files are loaded at build time of the (HTML) documentation ..

```
make docs
...
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://flask.palletsprojects.com/objects.inv...
loading intersphinx inventory from https://jinja.palletsprojects.com/objects.inv...
loading intersphinx inventory from https://return42.github.io/linuxdoc/objects.inv...
loading intersphinx inventory from https://www.sphinx-doc.org/en/master/objects.inv...
```

With the help of this mappings, sphinx can create HTML links to HTML anchors in external external projects.

In our documentation you can find some examples how external pages can be referenced ...

https://searxng.github.io/searxng/dev/reST.html#smart-refs

grafik

hope I could clarify ..

dalf: Yes it clarify, thank you.

I still think we can be break the interlink: I'm nearly sure that no other external documentation makes a link to this reference.
But that's not important.

return42: > I'm nearly sure that no other external documentation makes a link to this reference.

:-) .. d'accord

> I still think we can be break the interlink

Yes we could, but in this case I won't, because it refers to https://searxng.github.io/searxng/admin/installation-searx.html#searx-src where the installation of the `/usr/local/searx/searx-src` is described.

I think we will never change this folder like we will never change the name `searx` of the python package.

PR #319, opened 2021-09-14
plugins: refactor initialization

### LINKED ISSUE #315: Don't use flask_babel at import time
At import time the language domain is unknown, using `flask_babel` to translate values at import time makes no sense . . see discussion at https://github.com/searxng/searxng/pull/308#discussion_r706385553

This issue is related to a common design mistake: [Don’t run code at import time](https://www.benkuhn.net/importtime). Here the problem is, that some modules try to initialize their values which are depend on an application context. Such values should be initialized within the application (means: by the application).

dalf:
## What does this PR do?

Add a new function "init" call when the app starts.
The function can:
* return False to disable the plugin (case of ahmia_filter plugin).
* modify the Flask app.

See https://github.com/dalf/searxng/blob/mod-plugins/searx/plugins/ahmia_filter.py

Reference to the plugins are by Python module names instead of plugin names (related to https://github.com/searxng/searxng/issues/315)

The function `searx.plugins.load_plugin` looks like `searx.engines.load_engines`.

BREAKING CHANGE: the user preferences (cookies) about the plugins will be lost.

Embedded plugins are automatically detected.

With this change a plugin can change the web application:
* add / remove / replace some HTTP endpoints
* add middleware, etc...

Draft:
* ~if would be nice if the `init` could replace the application: `app = Flask_Babel(app)`.~
* ~perhaps implement a temporary workaround to avoid a breaking change about the cookies.~ (it just requires to go the preferences and make some changes and save)

* `make run`
* check it works with an external plugin including static resources

## Related issues

Close #315

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -234,9 +234,9 @@ def plugin_module_names():
     yield_plugins = set()
 
     # embedded plugins
-    for module_name in iter_modules(path=[dirname(__file__)]):
-        yield (module_name, False)
-        yield_plugins.add(module_name)
+    for module in iter_modules(path=[dirname(__file__)]):
+        yield (__name__ + "." + module.name, False)
+        yield_plugins.add(module.name)
     # external plugins
     for module_name in settings['plugins']:
         if module_name not in yield_plugins:
@@ -246,6 +246,6 @@ def plugin_module_names():
 
 def initialize(app):
     for module_name, external in plugin_module_names():
-        plugin = load_and_initialize_plugin(__name__ + "." + module_name.name, external, (app, settings))
+        plugin = load_and_initialize_plugin(module_name, external, (app, settings))
         if plugin:
             plugins.register(plugin)
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -59,7 +59,6 @@ def sync_resource(base_path, resource_path, name, target_dir, plugin_dir):
 
 
 def prepare_package_resources(plugin, plugin_module_name):
-    # pylint: disable=consider-using-generator
     plugin_base_path = dirname(abspath(plugin.__file__))
 
     plugin_dir = plugin_module_name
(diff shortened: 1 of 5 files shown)


searx/webapp.py
```diff
@@ -1351,6 +1332,27 @@ def page_not_found(_e):
...
+#  __name__ != "__main__" if searx.webapp is imported (make test, make docs, uwsgi...)
+# see run() at the end of this file : searx_debug activates the reload feature.
+werkzeug_reloader = flask_run_development or (searx_debug and __name__ == "__main__")
+
+# initialize the engines except on the first run of the werkzeug server.
+if (not werkzeug_reloader
+    or (werkzeug_reloader
+        and os.environ.get("WERKZEUG_RUN_MAIN") == "true") ):
+    plugin_initialize(app)
+    search_initialize(enable_checker=True)
```
dalf, line 1353: move the initialization at the end of the file, so the plugins can overwrite things.

return42:
FYI I needed to fix initialize() aa5a5147b2  .. while I on the way I used a generator 5731b6b .. if you think generator fails for any reason, please drop 5731b6b.

I haven't finished my tests today, I will continue the next days ..

searx/plugins/__init__.py
```diff
@@ -130,74 +46,200 @@ def sync_resource(base_path, resource_path, name, target_dir, plugin_dir):
...
             utime(resource_path, ns=(dep_stat.st_atime_ns, dep_stat.st_mtime_ns))
-        except:
-            logger.critical('failed to copy plugin resource {0} for plugin {1}'.format(file_name, name))
-            exit(3)
+        except IOError:
+            logger.critical(
+                "failed to copy plugin resource {0} for plugin {1}".format(
+                    file_name, name
+                )
+            )
```
return42, line 54: Nit .. do we need such a lot of new-lines?  I often use ..

```python
            logger.critical(
                "failed to copy plugin resource {0} for plugin {1}".format(
                    file_name, name ))
```
.. even though I know this is not up to standard / but I think it is more readable.

return42 (approved):
FYI: I tested with:

- https://github.com/asciimoo/searx_external_plugin_example
- https://github.com/return42/tgwf-searx-plugins

Patch aa5a5147b is needed and 5731b6b70 is optional.

PR #327, opened 2021-09-17
[mod] utils/lib.sh: add pyenv.activate

### LINKED ISSUE #310: inconsistent ouput of `$(pyenv.cmd wlc lock-status)`
See https://github.com/searxng/searxng/runs/3564689542?check_suite_focus=true#step:6:367

The output of `$(pyenv.cmd wlc lock-status)` is 
```
PYENV     [check] import yaml --> OK
PYENV     OK
locked: True
PYENV     [check] import yaml --> OK
PYENV     OK
```

So this check fail:
https://github.com/searxng/searxng/blob/75daef8986142bff65ef67aff4f278e41a4d0854/manage#L165-L167

It can be reproduce locally:
* `VERBOSE=1 ./manage pyenv.cmd wlc lock-status`
* `VERBOSE=0 ./manage pyenv.cmd wlc lock-status`

---

Possible fixes:
* `$(VERBOSE=0 pyenv.cmd wlc lock-status)`
* Enable the virtualenv since everything runs inside `( set -e .... )`:
https://github.com/searxng/searxng/blob/75daef8986142bff65ef67aff4f278e41a4d0854/manage#L392-L393
-->
```sh
    (   set -e
        # shellcheck source=/dev/null 
        source "${PY_ENV_BIN}/activate" 
        if [ "$(wlc lock-status)" != "locked: True" ]; then
            die 1 "weblate must be locked, currently: $(wlc lock-status)"
        fi
        # weblate: commit pending changes
        wlc pull
        wlc commit
...
```

dalf:
## What does this PR do?

* initialize virtualenv (pyenv.install)
* activate virtualenv

Use it in `./manage` in some locations.

See #310 : the pattern `$(pyenv.cmd command)` has a inconsistent output.
`pyenv.activate` allows to first install & activate the virtualenv
then `$(pyenv.cmd command)` can be replaced by `$(command)`

* `make docker`
* `make gecko.driver`
* ( test weblate targets are difficult to test )

## Related issues

close #310

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -725,6 +725,14 @@ pyenv.cmd() {
     )
 }
 
+
+pyenv.activate() {
+    pyenv.install
+    # shellcheck source=/dev/null
+    source "${PY_ENV_BIN}/activate"
+}
+
+
 # Sphinx doc
 # ----------
 
--- a/manage
+++ b/manage
@@ -322,22 +322,26 @@ weblate.push.translations() {
 }
 
 data.all() {
-    data.languages
-    data.useragents
-    data.osm_keys_tags
-    build_msg DATA "update searx/data/ahmia_blacklist.txt"
-    pyenv.cmd python searx_extra/update/update_ahmia_blacklist.py
-    build_msg DATA "update searx/data/wikidata_units.json"
-    pyenv.cmd python searx_extra/update/update_wikidata_units.py
-    build_msg DATA "update searx/data/currencies.json"
-    pyenv.cmd python searx_extra/update/update_currencies.py
+    (   set -e
+        pyenv.activate
+        data.languages
+        data.useragents
+        data.osm_keys_tags
+        build_msg DATA "update searx/data/ahmia_blacklist.txt"
+        python searx_extra/update/update_ahmia_blacklist.py
+        build_msg DATA "update searx/data/wikidata_units.json"
+        python searx_extra/update/update_wikidata_units.py
+        build_msg DATA "update searx/data/currencies.json"
+        python searx_extra/update/update_currencies.py
+    )
 }
 
 
 data.languages() {
     (   set -e
+        pyenv.activate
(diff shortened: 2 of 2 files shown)


return42:
Hm, do I miss something? .. I can't find function definition of `pyenv.activate` / the one you suggested in https://github.com/searxng/searxng/issues/310#issuecomment-917969345

dalf:
Added (I've made a first version where I've changed most of the `pyenv.cmd` calls, and then forgot to bring back the `utils/lib.sh` patch)

utils/lib.sh
```diff
@@ -725,6 +725,16 @@ pyenv.cmd() {
     )
 }
 
+
+pyenv.activate() {
+	pyenv.install
+    # shellcheck source=/dev/null
+	source "${PY_ENV_BIN}/activate"
```
return42: nit: mixed space and tabs

utils/lib.sh
```diff
@@ -725,6 +725,16 @@ pyenv.cmd() {
...
+    # make sure the exit code is 0, this function can be called with set -e
+	[ "$VERBOSE" = "1" ] && set -x || :
```
return42: Since this is not a subshell, do not activate `set -x` in a function.  The caller should decide if he wants to log `set -x`

IMO the following should be enough ...

```sh
pyenv.activate() {
    pyenv.install
    # shellcheck source=/dev/null
    source "${PY_ENV_BIN}/activate"
}
```

return42 (changes requested):
LGTM accept my two comments.

utils/lib.sh
dalf: I haven't fixed locations outside the pyenv.activation scope.

return42 (approved):

return42:
@dalf I pushed 7a3a1cd on top of your branch. If you think it does not fit into this PR just drop the commit.

All in all I think we can merge this PR.

PR #336, opened 2021-09-20
[theme] remove categories, language and time range from index.html

mrpaulblack:
## What does this PR do?

This PR removes the categories selection, time range as well as the language selection from the initial landing page (index.html) of the simple theme; the oscar theme already has a advanced button that hides that stuff anyway.

This unclutters the landing page and makes the usage more straightforward... (especially the category selection on the landing page seems kinda out of place !?). The time range selection, language selection and categories are still available from the result page.

```make run```

## Author's checklist

image


MrPaulBlack:
I was also thinking about adding the searXNG slogan underneath the search input field; Something like this maybe?
image

return42:
just my 5cent .. what about a "Advanced Settings" option like we have in the oscar theme?

grafik

mrpaulblack:
Yeah that could be an option, but I believe that the landing page should be simple as possible. User that still want the language and time range selector can still do so on the result page (after doing a search request). The advanced settings options would still in the end enable the same layout as before. So then you would need to think how the site looks not only on mobile, tablet and desktop, but also with the menu closed and opened. This change just makes it more simple and straightforward to maintain and simpler for the end user...

dalf:
I'm going to merge, and see how it goes, but I'm puzzled at same time: 
* I wish a way to search images directly: there is `!images ...` but it is not "compatible" with the category selection buttons.
* some other search engines have this option:
  * there is images.google.com to go directly the image search. 
  * bing has "Image Videos..." at the top of its page.
  * yandex has the same Image, Videos, etc... at the top of the input field.
  * but duckduckgo, qwant many others just have a search input text.

mrpaulblack:
image

Maybe something like this as a compromise? By default searches the general category, but the user could also press the search images button to go directly to the image category... (This does not overwhelm with to many options on the landing page, while still giving the user an option for an image search right away; more advanced users can use the !category query for the other categories.)

PR #337, opened 2021-09-22
[fix] fix searx.js null pointer because of missing category div

mrpaulblack:
## What does this PR do?

I didn't catch that my pr #336 actually throws an error in the browser console while loading the landing page (The search itslf still works tho). This PR fixes that by making sure the category div exists, before setting the category in searx.js.

Fixes error in browser console when accessing the landing page.

```make themes.all```

```make run```

## Author's checklist

On a side note the error is the following:

```
Uncaught TypeError: n.querySelector(...) is null
    <anonymous> searx.js:786
    ready searx.js:69
    <anonymous> searx.js:743
    <anonymous> searx.js:808
searx.js:786:8
```


dalf:
Thank you, I haven't seen it (may be we should add a check of the console in [browser tests](https://github.com/searxng/searxng/tree/master/tests/robot))

May I ask you to split the commit into two:
* one with only the change in searx/static/themes/simple/src/js/main/searx_search.js 
* `make static.build.commit`: it will build and commit the themes.
?

The idea is: 
* the source code in one commit: it is easier to follow the changes.
* the commit made by `make static.build.commit` can be throw away in case of git conflict.

Additional commands:
* `make static.build.drop` drops the last commit if it has been built by `make static.build.commit`.
* `make static.build.restore` apply `git restore` to the built files.

So the workflow is:
1. build the file while coding (`make themes` or `npm run build` in `searx/static/themes/simple`)
2. `make static.build.restore` to remove built files
3. `git add`  & `git commit` as usual (without the built files).
5. `make static.build.commit`

With this workaround, we can the source and built files in the same repository, while being able to deal with pull requests.

I hope it is not too cumbersome.

mrpaulblack:
@dalf done. And to be honest, I don't get why the build files would even be tracked in this repository (Other than for the searx.space checks). The workflow could be that they are just built by the github CI, when building the docker container maybe?

dalf:
There are different ways to install the app:
* Docker: Dockerfile can build the themes as you said.
* git (git clone or using the scripts in the utils directory): if the built files are not in the repository, it requires the npm dependencies on the server. An external repository with the built files breaks the existing installations.

PR #340, opened 2021-09-24
Simple theme preferences refactor

mrpaulblack:
## What does this PR do?

This PR changes up the simple theme preferences to match the oscar theme. It adds the special queries tab, reordered the tabs to match oscar and cleans up a few misplaced white spaces.

The main motivation was to add the special queries tab for feature parity between simple and oscar and to streamline the preferences.

```make run```

## Author's checklist

image

### CODE REVIEW

File: searx/templates/simple/preferences.html [hunk 1]
```diff
@@ -213,55 +280,51 @@ <h2>{{ _('Preferences') }}</h2>
...
+        <th>{{ _('Description') }}</th>
+        <th>{{ _('Examples') }}</th>
+      </tr>
+      <td></td>
+      <th scope="colgroup" colspan="4">{{ _('This is the list of searx\'s instant answering modules.') }}</th>
+      {% for answerer in answerers %}
+      <tr>
+        <td></td>
+        <td>{{ answerer.keywords|join(', ') }}</td>
+        <td><div class="checkbox">{{ answerer.info.name }}</div></td>
```
mrpaulblack: I just used the checkbox class to make the special queries table look a bit better (with the margin); This seems a bit hacky tho!? Should I rather add an extra class to the css theme for this maybe?

dalf:
Thank you !

What do you think of this layout ? (based on your PR with some CSS adjustement)
image

```css
/* specific to "specialquery" tab */
#tab-contentquery table td {
	text-align: left;
	height: 3rem;
}
```
(the HTML doesn't use the `checkbox` CSS class).

And two generic changes that can be part of another PR: 
```css
/* max width for the preferences, can be applied to some other pages like stats and about */
#main_preferences {
    margin: 3em auto;
    width: 90rem;
}

/* adjust the checkbox with the above changes, 
    see the engine tab
*/
.checkbox-onoff {
	margin: 8px 1rem;
}
```

searx/templates/simple/preferences.html
mrpaulblack: changed in 7cb5099

mrpaulblack:
@dalf Thanks I added your CSS for the special queries tab; tho I used !important since the text-align gets otherwise overritten by the `text-align: center` in `#main_preferences table td` selector. On another note `letter-spacing: -0.1px;` should be removed from toolkit.less line 305 in another PR, since this is the reason why the width of a selected tab in preferences changes compared to a non selected one... (so the tab bar in the preferences moves around when items get clicked)

searx/templates/simple/preferences.html
dalf: I'm not sure if this is intended, but there is still the `checkbox` class.
image

Without the labels don't overlap.

return42:
In my Firefox I have some overwrites ... in the name column

grafik

but these are not related to this PR, I have them also in other tabs ..

grafik

grafik

I have this "overwrites" since the beginning .. and only in Firefox .. I hadn't had the time to check by what this are caused ... sorry.

searx/templates/simple/preferences.html
mrpaulblack: Oh sorry missed that; Fixed in 3017227

dalf:
LGTM 
image

image

@return42  is it a browser cache issue?

return42:
> @return42 is it a browser cache issue?

may be .. I'm not at the PC from where I posted the screen shoots in the comment above, but I tested this PR with Firefox on my laptop and everything looks fine ... lets ignore my comment from above.

PR #341, opened 2021-09-24
[enh] verify that Tor proxy works every time searx starts

dalf:
## What does this PR do?

based on @MarcAbonce commit on searx (PR 2981)

When `using_tor_proxy` is True, check that each new `httpx.AsyncClient`  are actually connected through Tor

When the server starts, all networks are checked.

Tricky part, it is possible to specify multiple proxies even with Tor.

For example:
```yaml
using_tor_proxy: true
proxies:
  "all://":
    - socks5h://127.0.0.1:9050
    - socks5h://127.0.0.1:9051
```

In this example:
* when the server starts, only the first proxy `socks5h://127.0.0.1:9050` is checked.
* `socks5h://127.0.0.1:9051` is going to be checked on the first use.

It shouldn't be an issue since this configuration most probably doesn't make sense with Tor. The only drawback is a slower response time on the first use of the second proxy.

Ensure that the instance maintainers are configuring their Tor proxies correctly.

Nearly the same protocol than the PR 2981 on searx:
```
Default scenario:

    Run searx with default settings.
    No verification should be made. Everything should work as usual.

Correctly configured Tor:

    Add a proxy such as all:// : socks5h://127.0.0.1:9050 and run Tor in that port.
    Set the using_tor_proxy setting as true.
    Run the server.
    **Logs should show requests to https://check.torproject.org/api/ip**

Incorrectly configured Tor:

    Set the using_tor_proxy setting as true.
    Keep the proxies section empty or proxy to Tor using socks5 instead of socks5h.
    Run the server.
    Logs should show an explanatory error and the server should stop running.
```

## Related issues

* Potential enhancement of this PR: https://github.com/benbusby/whoogle-search/pull/137 uses [stem](https://pypi.org/project/stem/) to reacquire a new Tor profile in case of error.
* PR 2981 on searx, ~most probably this PR can be merged on searx.~

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -11,7 +11,7 @@ from itertools import cycle
 import httpx
 
 from searx import logger, searx_debug
-from .client import new_client, get_loop
+from .client import new_client, get_loop, AsyncHTTPTransportNoHttp
 
 
 logger = logger.getChild('network')
@@ -42,9 +42,11 @@ class Network:
     __slots__ = (
         'enable_http', 'verify', 'enable_http2',
         'max_connections', 'max_keepalive_connections', 'keepalive_expiry',
-        'local_addresses', 'proxies', 'max_redirects', 'retries', 'retry_on_http_error',
+        'local_addresses', 'proxies', 'using_tor_proxy', 'max_redirects', 'retries', 'retry_on_http_error',
         '_local_addresses_cycle', '_proxies_cycle', '_clients', '_logger'
     )
+
+    _TOR_CHECK_RESULT = {}
 
     def __init__(
             # pylint: disable=too-many-arguments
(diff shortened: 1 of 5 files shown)


searx/network/network.py
```diff
@@ -144,15 +148,35 @@ async def log_response(self, response: httpx.Response):
...
 
-    def get_client(self, verify=None, max_redirects=None):
+    @staticmethod
+    async def check_tor_proxy(client: httpx.AsyncClient, proxies) -> bool:
+        if proxies in Network._TOR_CHECK_RESULT:
+            return Network._TOR_CHECK_RESULT[proxies]
+
+        result = True
+        # ignore client._transport because it is not used with all://
+        for transport in client._mounts.values():  # pylint: disable=protected-access
```
dalf, line 158: check all mounted transports except the default one: 
* the default transport is not used with the configuration `all://: socks5h://127.0.0.1:9050`.
* there is a loophole with `https://somewhere: socks5h://127.0.0.1:9050` and if the engine(s) send(s) requests to other domains.

[EDIT] see #342

MarcAbonce:
Thank you! Sorry I didn't comment on time but yeah, this PR works better.

PR #343, opened 2021-09-24
[fix] checker: fix memory usage

dalf:
## What does this PR do?

* download images using the "image_proxy" network (HTTP/1 instead of HTTP/2)
* don't cache data: URL (reduce memory usage)

Drastically reduce the memory usage when the checker is enabled.
 
* enable the checker in settings.yml
* `kill -USR1 <pid>`
* check the memory usage
* `kill -USR1 <pid>` again
* check the memory usage : it shouldn't be similar to the previous value

### DIFF
--- a/searx/search/checker/impl.py
+++ b/searx/search/checker/impl.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+import gc
 import typing
 import types
 import functools
@@ -14,6 +15,7 @@ from langdetect.lang_detect_exception import LangDetectException
 import httpx
 
 from searx import network, logger
+from searx.utils import gen_useragent
 from searx.results import ResultContainer
 from searx.search.models import SearchQuery, EngineRef
 from searx.search.processors import EngineProcessor
@@ -58,7 +60,47 @@ def _is_url(url):
 
 
 @functools.lru_cache(maxsize=8192)
-def _is_url_image(image_url):
+def _download_and_check_if_image(image_url: str) -> bool:
+    """Download an URL and check if the Content-Type starts with "image/"
+    This function should not be called directly: use _is_url_image
+    otherwise the cache of functools.lru_cache contains data: URL which might be huge.
+    """
+    retry = 2
+
+    while retry > 0:
+        a = time()
+        try:
+            # use "image_proxy" (avoid HTTP/2)
+            network.set_context_network_name('image_proxy')
+            stream = network.stream('GET', image_url, timeout=10.0, allow_redirects=True, headers={
+                'User-Agent': gen_useragent(),
+                'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
+                'Accept-Language': 'en-US;q=0.5,en;q=0.3',
+                'Accept-Encoding': 'gzip, deflate, br',
+                'DNT': '1',
+                'Connection': 'keep-alive',
+                'Upgrade-Insecure-Requests': '1',
+                'Sec-GPC': '1',
+                'Cache-Control': 'max-age=0'
+            })
+            r = next(stream)
+            r.close()
+            is_image = r.headers["content-type"].startswith('image/')
+            del r
+            del stream
+            return is_image
+        except httpx.TimeoutException:
+            logger.error('Timeout for %s: %i', image_url, int(time() - a))
+            retry -= 1
+        except httpx.HTTPError:
+            logger.exception('Exception for %s', image_url)
+            return False
+    return False
+
+
+def _is_url_image(image_url) -> bool:
+    """Normalize image_url
+    """
     if not isinstance(image_url, str):
         return False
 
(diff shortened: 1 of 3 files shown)


return42:
I am not sure if I have done everything correctly, here is what I have tested ..

I don't know if it was needed, but I also activated `image_proxy` ...

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 3fefb62f..65d2b11b 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -41,7 +41,7 @@ server:
 
   secret_key: "ultrasecretkey"  # change this!
   # Proxying image results through searx
-  image_proxy: false
+  image_proxy: true
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
   # POST queries are more secure as they don't show up in history but may cause
@@ -164,7 +164,7 @@ outgoing:
 
 checker:
   # disable checker when in debug mode
-  off_when_debug: true
+  off_when_debug: false
 
   # use "scheduling: false" to disable scheduling
   # scheduling: interval or int
```

My tests are done like this ...

**Before this patch:**

In the first terminal typed ...

```
$ make run
...
INFO    searx                         : Send SIGUSR1 signal to pid 27597 to start the checker
...
```

In a second terminal typed ...

```
$ kill -USR1 27597
$ ps -o vsz=,rss= 27597
1159728 148760
```

After checker ends ...

```
$ ps -o vsz=,rss= 27597
1288144 334796
```

**With patch applied:**


```
$ make run
...
INFO    searx                         : Send SIGUSR1 signal to pid 29357 to start the checker
...
```

In a second terminal typed ...

```
$ kill -USR1 29357
$ ps -o vsz=,rss= 29357
1278360 214800
```

After checker ends ...

```
$ ps -o vsz=,rss= 29357
1351072 331200
```
**Conclusion ...**

```
before patch --> VSZ: 1.288.144 kB // RSS: 334.796 kB
after  patch --> VSZ: 1.351.072 kB // RSS: 331.200 kB
```

Resident (RSS ) seems 1% better .. but I'm not a *memory expert* and I don't know if I have measured the right things here.

A last word about repeating checker runs .. If I repeat `kill -USR1 29357` a second time, 

```
$ ps -o vsz=,rss= 29357
1369312 369332
```

Resident seems to grow over the time .. does it mean we have a memory leak?

dalf:
Side note: with the last commits, this PR fixes the checker and the image proxy.

In additional to the original changes:
* the checker download only the first 64kb of the images using `searx.network.stream`.
* in the master branch, `searx.network.network` returns a `httpx.Reponse` with a modified `close` method. In this PR, a memory leak has been fixed when in the `close` method.
* at the end of the tests for a processor, the URL cache is cleared, and there is a call to the garbage collector.

The RSS value is about 320 000 KB after the checker: when called multiple times, the RSS value can increase multiple times in a row (5 times), and then come back to ~ 320 000.

searx/search/checker/impl.py
```diff
@@ -88,15 +83,37 @@ def _is_url_image(image_url):
                 'Sec-GPC': '1',
                 'Cache-Control': 'max-age=0'
             })
-            if r.headers["content-type"].startswith('image/'):
-                return True
-            return False
+            r = next(stream)
+            r.close()
+            is_image = r.headers["content-type"].startswith('image/')
```
return42: One issue I had while testing was a "HTTP/1.1 503 Service Unavailable" which was not robust handled by `_download_and_check_if_image(..)` function

On a 503 response there is no `r.headers["content-type"]` ... the checker stops at this point, may be we can handle 503 (and other HTTP errors) more robust at this place


```
DEBUG   searx.network.image_proxy     : HTTP Request: GET https://tse2.mm.bing.net/th?id=OIP.Cwz2_RWJmKgzb_8YY391wQHaD2&pid=15.1 "HTTP/1.1 200 OK" (image/jpeg)
DEBUG   searx.network.image_proxy     : HTTP Request: GET https://tse3.mm.bing.net/th?id=OIP.PRLNrsmo2YIPb-0Lr0tGCAHaEK&pid=15.1 "HTTP/1.1 503 Service Unavailable"
ERROR   searx                         : Error while running the checker
  File "searx/searx/search/checker/background.py", line 64, in run
    checker.run()
  File "searx/searx/search/checker/impl.py", line 433, in run
    self.run_test(test_name)
  File "searx/searx/search/checker/impl.py", line 419, in run_test
    rct_list = [self.get_result_container_tests(test_name, search_query) for search_query in search_query_list]
  File "searx/searx/search/checker/impl.py", line 419, in <listcomp>
    rct_list = [self.get_result_container_tests(test_name, search_query) for search_query in search_query_list]
  File "searx/searx/search/checker/impl.py", line 413, in get_result_container_tests
    result_container_check.check_basic()
  File "searx/searx/search/checker/impl.py", line 270, in check_basic
    self._check_results(results)
  File "searx/searx/search/checker/impl.py", line 245, in _check_results
    self._check_result(result)
  File "searx/searx/search/checker/impl.py", line 236, in _check_result
    if not _is_url_image(thumbnail_src):
  File "searx/searx/search/checker/impl.py", line 116, in _is_url_image
    return _download_and_check_if_image(image_url)
  File "searx/searx/search/checker/impl.py", line 88, in _download_and_check_if_image
    is_image = r.headers["content-type"].startswith('image/')
  File "searx/local/py3/lib/python3.8/site-packages/httpx/_models.py", line 990, in __getitem__
    raise KeyError(key)
KeyError: 'content-type'
```

return42 (changes requested):
@dalf I tested with 0d7ec508 ... I can confirm that the memory usage stops somewhere ~320kB / thanks for fixing this.

There is only one point I would suggest, see my comment below ...

searx/search/checker/impl.py
dalf: It should be fixed with (I'm going to update the PR):
```python
            # use "image_proxy" (avoid HTTP/2)
            network.set_context_network_name('image_proxy')
            stream = network.stream('GET', image_url, timeout=10.0, allow_redirects=True, headers={
                'User-Agent': gen_useragent(),
                'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
                'Accept-Language': 'en-US;q=0.5,en;q=0.3',
                'Accept-Encoding': 'gzip, deflate, br',
                'DNT': '1',
                'Connection': 'keep-alive',
                'Upgrade-Insecure-Requests': '1',
                'Sec-GPC': '1',
                'Cache-Control': 'max-age=0'
            })
            r = next(stream)
            r.close()
            if r.status_code == 200:
                is_image = r.headers.get('content-type', '').startswith('image/')
            else:
                is_image = False
            del r
            del stream
            return is_image
```

PR #345, opened 2021-09-24
simple theme css refactor

mrpaulblack:
## What does this PR do?

Some simple theme improvements:

* make selected tabs in preferences keep the same width, compared to not being selected (they don't move the tab bar around anymore) -> removing bolt font and change in letter spacing.

* remove various `padding`, `margin` and `width` elements from sidebar and results -> on desktop: 10em margin on the left; on tablet -> 0.5em margin; on phone no margin at all; also there is a 2em margin between content and footer as well as content and the header search bar

* introduce grid layout to the result page to make layout changes between PC, Tablet and Smartphone smoother (based on dalf@0f60925) -> image page has a 0.5em margin on all screen sizes

* sticky footer at the bottom of the page (also from dalf@0f60925)

* remove various elements from toolkit to streamline url copy elements and engine errors in new grid layout

* add max width of 85em to stats, pref and about page; make margin around .checkbox-onoff elements bigger

* add a margin to index on landing page to center the search more (on desktop 16em, on tablet 6em and on mobile 0 to not get hidden by keyboard)

* hide the backtotop element on tablet and mobile

## Why is this change important?

Make the simple theme more polished...

## How to test this PR locally?

```make themes.all```

```make run```

## Author's checklist

Desktop

image

Tablet

image

Phone

image

PR #350, opened 2021-09-27
[fix] engine description position in simple theme

mrpaulblack:
## What does this PR do?

https://github.com/searxng/searxng/pull/345 changed the preferences to have a max width of 85em and to be positioned in the middle of the screen. This broke the engine description modal, since its position is hardcoded. This PR changes this by letting the modal calculate its position based on the screen width...

Makes the simple theme more polished looking.

```make run```

## Author's checklist

On desktop:
image

On tablet:
image


return42 (approved):
This patch itself LGTM and I think it is ready to merge / Thanks!

Might be OT here, but I see problems with modal dialogues on touch devices, but I am too inexperienced to give the right answers ..

1. First I miss a link to to the settings options on mobile (small) devices.
2. How (when) is a modal shown (their is no "mouse over" event I think).
3. Positioning of modal dialogues do not fit to small devices (see screenshot below) / what might be an conceptual knowledge gap of mine and is related to 2. 

grafik

dalf:
Thank you!

> How (when) is a modal shown (their is no "mouse over" event I think).

On my mobile, the tooltip is displayed when I touch the engine name.

> Positioning of modal dialogues do not fit to small devices (see screenshot below) 

@return42, I'm not sure how to do that on mobile. Can you create a new issue about this subject?

PR #353, opened 2021-09-28
[mod] engines - add Stack Exchange API v2.3

return42:
## What does this PR do?

engines - add Stack Exchange API v2.3

With the Stack Exchange API we can configure several engines from the https://stackexchange.com/sites

In the second patch the broken (CAPTCHA) stackoverflow engine is replaced by this API. 

I also added patches with configurations for askubuntu.com and superuser.com


dalf:
If I search for `!stackoverflow python reverse string`, the HTML titles contain some HTML entities:
* `Python 3.8 (Luhn&#39;s Algorithm Alternative): How do I convert a list into a string?`
* `Traceback &lt;module&gt; from extractdocx import * &lt;module&gt; from docx import opendocx, getdocumenttext from exceptions import PendingDeprecationWarning`

return42:
@dalf: fixed decoding HTML entities in ecb3912b

dalf (approved):
LGTM. Thank you.

---

I think at one point we need to think about the fields per result template (and may be add some more templates).

In this engine there is:
* the rating
* the author
* the published date (?)

Some other engines serialize some field too (especially, in the video category)

return42:
> I think at one point we need to think about the fields per result template (and may be add some more templates).

I confirm, a stackexchange template might be a good alternative .. beside this, it is hard for me to guess which fields are supported in what template .. I opened #357 .. you know at best what fields we have in the simple theme, if you have time, do you like to add a documentation?

PR #354, opened 2021-09-28
[mod] /image_proxy: fix memory leak

dalf:
## What does this PR do?

`/image_proxy`: 
* make sure to close the httpx response
* use [`direct_passthrough`](https://werkzeug.palletsprojects.com/en/2.0.x/wrappers/#werkzeug.wrappers.Response) for a slight improvement of the performance.

No memory leak.

* `make run`
* enable the image proxy
* watch the RSS (`watch ps -o vsz=,rss= <pid>`)

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1089,12 +1089,11 @@ def image_proxy():
             'DNT': '1',
         }
         set_context_network_name('image_proxy')
-        stream = http_stream(
+        resp, stream = http_stream(
             method = 'GET',
             url = url,
             headers = request_headers
         )
-        resp = next(stream)
         content_length = resp.headers.get('Content-Length')
         if (content_length
             and content_length.isdigit()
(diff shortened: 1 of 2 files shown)


searx/webapp.py
```diff
@@ -1124,22 +1123,29 @@ def image_proxy():
...
-            total_length = 0
-            for chunk in stream:
-                total_length += len(chunk)
-                if total_length > maximum_size:
-                    break
-                yield chunk
-
-        return Response(forward_chunk(), mimetype=resp.headers['Content-Type'], headers=headers)
+        response = Response(
+            stream,
```
dalf, line 1141: Important note: the size limit of 5MB is removed.

return42: Is it? .. in line [1080](https://github.com/searxng/searxng/pull/354/commits/a9c3c88cc0edf4dc3183f2188c518ec4a5b37055#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R1080) there is a `maximum_size` which is checked in line 1100  `if ... and int(content_length) > maximum_size`

return42 (approved):
I tested before/after this patch applied. It is very hard for me to judge which implementations consumes less memory since memory is also consumed by other parts (I assume the stats do consume some memory also).

My conclusion is very subjective, but I would prefer this patch over the existing implementation.

dalf:
Yes... it would require a dedicated benchmark.

I've watched the RSS value using `watch ps -o vsz=,rss= <pid>`, according to what I've seen:
* it increases when there are engine requests
* it is stable when there are a long list of request through the image_proxy <-- without this PR, the RSS increase here too.

The "key" part is https://github.com/searxng/searxng/blob/a9c3c88cc0edf4dc3183f2188c518ec4a5b37055/searx/webapp.py#L1145 combine with #343 it makes sure that the httpx.Response is properly closed.

Also in `searx.network.stream`, there is [one less "yield" layer](https://github.com/searxng/searxng/blob/c23aa5760cb42003c1372ccdef1611695504eb9e/searx/network/__init__.py#L241) since it [returns the generator directly](https://github.com/searxng/searxng/blob/a9c3c88cc0edf4dc3183f2188c518ec4a5b37055/searx/network/__init__.py#L239).

I'm going to merge.

ping @MrPaulBlack : it may change the memory usage in production.

mrpaulblack:
RAM usage seems to went down a tiny bit on my production server after this merge actually (The dip in the middle is when I updated my instance; also keep in mind that I am also running other services on that VPS so that usage is not only from searxng; the yellow line is the used RAM):

image

dalf:
May be not easy to say, but do you know if the used RAM increase over the days if you don't restart searxng?

mrpaulblack:
Before this PR I think it did, yes:

image

(I need to wait a few days to see if this PR fixed the issue tho...)

mrpaulblack:
After leaving the searx installation running for 2 days with the fix from this PR:
image


(I think the memory leak issue is fixed; at least I cannot see any steady increase in RAM usage over time...)

dalf:
Thank you !
( I have created https://github.com/searxng/searxng-docker/issues/3 )

PR #355, opened 2021-09-28
[simple theme] dark mode and some css cleanup

mrpaulblack:
## What does this PR do?

This PR is adding a dark theme to the simple theme by:

* replacing less vars by CSS vars (this also makes it possible for people to easily change the color theme in the compiled searx.css, since the preprocessor does not touch these css vars...)

* dropping some of the vars (vars like `@color-result-link-font` seems a little to verbose!?)

* change every hardcoded color in the theme to be a css var in defenitions.less

* add a dark class with a dark color theme

* update result design on tablet and mobile

* Make UI elements bigger: more padding on search form and buttons

* Minor changes and removed unused CSS (See commit messages)

Another PR to eventually have feature parity between simple and oscar theme... Also make the simple theme more polished...

```make themes.simple```

```make run```

## Author's checklist

Dark Theme:

image
image
image

Light Theme:
image
image
image

## Related issues

Partially solves: #347 and #176

### CODE REVIEW

**Comment by dalf** [comment]
I'm not sure if it should be a new issue:
here a demo implementation about the dark theme https://gist.github.com/dalf/0f983799c20b75b441f8c089372f0cfe (only the wires between CSS, JS and cookies, not the actual changes).
* the idea is have a new user preferences to choose between "automatic", "light", "dark" theme (only for the simple theme), more or less equivalent of "oscar-style" --> the user can choose the variant in the preferences.
* on `/` and `/search` there is a button on the top right corner to switch between "light" and "dark".
   * this button is displayed only when JS is activated. 
   * when clicked, this button makes a background call to `/preferences` to update the cookies (so only python code sets the cookies)
* the CSS variables are on `html`, not `::root` so the Jinja2 template defines the variant --> it works without Javascript (except the top right button).

If this suggestion is okay / makes sense:
* I can create another PR with the changes about webapp.py + 00_searx_toolkit.js
* or you can copy / paste / modify the code from the gist.

File: searx/templates/simple/macros.html [hunk 1]
```diff
@@ -39,9 +42,11 @@ <h3>{{ result_link(result.url, result.title|safe) }}</h3>
 {%- endmacro -%}
 
 <!-- Draw result sub footer -->
-{%- macro result_sub_footer(result, proxify) -%}
```
mrpaulblack, line 42: ### TODO

add proxify again since the cache and proxy result button moved back here...

mrpaulblack:
@dalf Sorry for the late reply... Can you please test my progress with the simple theme? I am not a UI dev and I just tested different colors until the updated light theme and the dark theme looked good to me (heavily opinionated tho). Right now I basically just cleaned up the stylesheets and removed unused stuff as well as reworking the CSS vars...

The next step is to implement the options for the user to choose between light, dark and auto... (I dont really have time for that right now tho, since university is going to start for me on Monday again... But I wanted to at least finish the dark theme until then...)


[EDIT] I also played around with transitions. But in my opinion they only look good, if every viewable element uses that transition, which works on the startpage and about page, but on the result view there are just too many elements... So instead of putting the transition only on the html and important elements, I would rather leave them out completely... I think it looks kinda odd when the background has a transition on theme switch, but the rest of the page transitions instantly... What do you think about that? (But I would still use transition on focus and hover elements in the future tho...)

dalf:
Thank you for this amazing work.

I'm going to merge your PR.

There are few things here and there, but it can be discuss in other pull requests / issues. ( for example 09324f34925676a50fd6be6406e09b4805df4f04 )

PR #361, opened 2021-09-30
SearXNG theme oscar

dalf:
## What does this PR do?

SearXNG brand: Oscar theme

Related to https://github.com/searxng/searxng/pull/358

## Why is this change important?

See https://github.com/searxng/searxng/pull/358

PR #358, opened 2021-09-29
Rename all from searx to searxng

HLFH:
## What does this PR do?

Rename all from searx to searxng (files, folder, content)

This change allows searxng to be a complete fork with its own name everywhere.

## Why is this change important?
This is important as searxng is now too different from searx with 600+ more commits. searxng is now a successful fork.

I was doing the AUR package for searxng on ArchLinux, and so before packaging searxng, it is best to rename everything to searxng.

PR #371, opened 2021-10-01
Searxng theme simple

dalf:
## What does this PR do?

SearXNG brand: 
* Simple theme
* remove hack from https://github.com/searxng/searxng/pull/361
  * https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/static/themes/__common__/js/image_layout.js#L161
  * https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/static/themes/oscar/src/js/01_init.js#L27

Related to #358


searx/static/themes/simple/src/js/main/searx_keyboard.js
```diff
@@ -1,12 +1,12 @@
-/*global searx*/
```
return42, line 1: Nit: all files we touch should get a SPDX-License-Identifier, if there is no Copyright a simple tag should be enough:

    /* SPDX-License-Identifier: AGPL-3.0-or-later */

searx/static/themes/simple/src/js/main/searx_mapresult.js
```diff
@@ -1,28 +1,11 @@
-/**
```
return42, line 1: see https://github.com/searxng/searxng/pull/371#discussion_r720340276

searx/static/themes/simple/src/js/main/searx_keyboard.js
dalf: updated

return42 (approved):
I tested this PR and did not find any issue / approved!

There is only one thing I also wrote in my comments: While we are in the bulk-renaming I think we should add SPDX tags to each file we touch, even if there is no licence remark nor a copyright exists.

PR #375, opened 2021-10-02
SearXNG: searx_extra

dalf:
## What does this PR do?

SearXNG: searx_extra directory


return42 (approved):
I pushed b791359 on top, could you have a look at .. if you want, squash with 1bb82a6b5  .. or drop the patch if you think it is to much / thanks!

searxng_extra/standalone_searx.py
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env python
-"""Script to run searx from terminal.
+# lint: pylint
+
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# (C) Copyright Contributors to the SearXNG project.
+# (C) Copyright Contributors to the searx project (2014 - 2021)
```
dalf, line 6: I have changed this line: I'm not the original author of this file.
See https://github.com/searx/searx/pull/1591

dalf:
Thank you for the review & the additional commit.

Issue #368, opened 2021-10-01
Image category: layout using flexbox instead of javascript


dalf:
The image sizes are changed to fill the screen width:

Oscar:
![image](https://user-images.githubusercontent.com/1594191/135585174-3ba68eb4-5ef4-4659-94fb-3e7035895db9.png)

Simple theme:
![image](https://user-images.githubusercontent.com/1594191/135585247-2e4c8328-f6a0-4937-b295-cc23a77d37c1.png)

The current implementation use javascript: https://css-tricks.com/adaptive-photo-layout-with-flexbox/

It would be interesting to use CSS:
* to reduce the number of javascript files
* not to use the `style` attributes

Possible implementation: https://css-tricks.com/adaptive-photo-layout-with-flexbox/

The implementation can focus on the simple theme.

dalf:
from https://github.com/searxng/searxng/issues/523#issuecomment-974117015
ping @mrpaulblack 

After a quick test, it works, but it requires to create HTML element to group the images when there are mixed result (from the general category and image category).

(I don't have a preference between flex or grid).

mrpaulblack:
Just as a heads up I strongly prefer grid layout...

return42:
@mrpaulblack how would you rate it, is this PR a requirement for #665?

mrpaulblack:
Yeah its part of the simple theme board, I am going to start working on the project in early January and tackle a lot of these issues... But please just hold on for a while, thanks :+1: 

[EDIT] and yeah I forgot this when I approved your PR earlier today :/

PR #376, opened 2021-10-03
Simple theme: image detail

dalf:
## What does this PR do?

One screenshot (the detail are opened on the right, to avoid to hide the thumbnails, they are grouped on the left side of the screen. When the detail is closed, the thumbnails use the whole screen width):
image

TODO (a lot of issues here and there):
* [x] back to top when the detail is hidden when there are only image results.
* [x] ~image selection: sometimes, the blue border doesn't follow the image border (smaller at the bottom on the image)~ related to #368
* [x] detail text: hard coded black & white.
* [x] when the detail is opened, the page offset must be updated to make sure the current selected image is on the screen (the layout reflow may hide the current selected image).
* [x] it is possible to `j` and `k` to go on the next and the previous image, perhaps add some button on screen (?)
* [x] close the detail with the `Esc` key ?
* [x] display a spinner while the JS is loading the full size image.
* [x] define CSS variables from the detail
* [x] the keys `j` and `k` update the detail
* [x] don't hide the search fields at the top of the page (?)
* [x] if the title or description are mangled, display the full text when mouse is over the text
* [x] center the arrows in the next and previous buttons
* [ ] make the touch zone of the next and previous buttons bigger to be usable on a touch screen
* [ ] be able to close the detail modal, if a user presses the currently selected image on the result page (again)

Note: the implementation uses Javascript.
Perhaps it is possible not to use Javascript with anchor ( `#image1` ) and [`:target`](https://developer.mozilla.org/en-US/docs/Web/CSS/:target) CSS selector. It requires to replace [`data-vim-selected`](https://github.com/searxng/searxng/blob/e39a03cc61e9792afb34084fb4d9973a61deecea/searx/static/themes/simple/src/less/style.less#L127-L129) attribute selector with anchors.

Expected behavior / Improve UX.

`make themes` to build the static files

Then on a computer:
* `!general !images something`
* `!images something`

On tablet and smartphone, the detail can be full screen (for now I've not tried).

Important:
* the full image size is proxied in the detail. It change the workload on the instance.
* some full image size URL use the `http` protocol: they are blocked by the proxy
* some full image weight more than 5MB: they are blocked too (only when [the `Content-Length` is specified](https://github.com/searxng/searxng/blob/e39a03cc61e9792afb34084fb4d9973a61deecea/searx/webapp.py#L1097-L1101)). The template falls back to the image thumbnail (see https://github.com/searxng/searxng/blob/4599b6cc16dbbaca1949093c512d88aa7e021f2f/searx/static/themes/simple/src/less/style.less#L191-L206 )

## Related issues

* map results and embedded players (videos, music) can use the same widget (instead of inline widgets).


return42:
I played with this PR .. here are some thoughts I had ,,

On small devices a scroll bar (up/down) is needed ..

grafik

The distance between image an description seems depend on the window size ..

grafik

When i resizer (smaller) the browser window, the distance grows ..

grafik

The "image modal" overwrites the menue in the top row (navigation bar on top).

grafik

dalf:
I've pushed force and update

> The distance between image an description seems depend on the window size ..

The idea is to have the text at a fixed location at the bottom of the screen, so it is possible to browse the different images to look for the authors (for example) easily. The location shouldn't not change as it was before.

There are still issues:
* ~the close button doesn't work on tablet and phone screen~
* with the keyboard (j, k keys), the thumbnails are not show

searx/templates/simple/result_templates/images.html
```diff
@@ -1,4 +1,26 @@
...
+<article class="result result-images {% if result['category'] %}category-{{ result['category'] }}{% endif %}">{{- "" -}}
+        <a {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">{{- "" -}}
+                <img class="image_thumbnail" src="{% if result.thumbnail_src %}{{ image_proxify(result.thumbnail_src) }}{% else %}{{ image_proxify(result.img_src) }}{% endif %}" loading="lazy" title="{{ result.title|striptags }}" alt="{{ result.title|striptags }}" />{{- "" -}}
+                <span class="title">{{ result.title|striptags }}</span>{{- "" -}}
+        </a>{{- "" -}}
+        <div class="detail">{{- "" -}}
+                <a class="result-detail-close" href="#">{{ icon('close') }}</a>{{- "" -}}
+                <a class="result-images-source" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">
+                        {%- if result.thumbnail_src -%}
+                                <img src="" data-src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}">
```
dalf, line 12: The thumbnail url is empty to reduce the page size.

The thumbnail is already copied twice (`<article...><a href="..."> To avoid a third copy, the `src` is initialized by javascript.

Note: without javascript, the link opens the full size image instead of the detail.

---

`data-src` contains the full size URL.
Previously it was two images, the full image on top of the thumbnail, but it requires `position: absolute`, and I haven't found a way to fit the image into the dedicated space without JS.

So JS loads the full size image in the background and switch the image when it is ready (a spinner would be helpful for the user). See searx/static/themes/simple/src/js/main/searx_results.js

dalf:
Current screenshot:

image

mrpaulblack:
I tried out your PR and this is really cool! The only small thing I think would make this even better is to be able to close the detail modal, if a user presses the currently selected image on the result page (again)...

[EDIT] The contrast between links and the background is not high enough with the light theme as well as the boarder having way to much contrast (The best solution would probably be to either make new vars for the colors in the detail modal and make them the same in the light and dark theme or to hardcode them in the `detail.less`, because they wont change anyway)
image

(also images do not load when using the arrows on the top right in the details modal; I need to switch back to the already downloaded image to make it work;)

return42:
Can I (we) rebase this PR?

> The idea is to have the text at a fixed location at the bottom of the screen, so it is possible to browse the different images to look for the authors (for example) easily. The location shouldn't not change as it was before.

OK, I understand the intention .. but on mobile, the UX is a bit strange and tiny to me .. may be it is easier to give up the intention?

grafik

On Desktop, the "image modal" overwrites the menue in the top row (navigation bar on top).

grafik

When I use the  Forward button (in the top right corner), sometime the image is not shown.  In the developer view I can see, that the image request has been succeed but I have to go backward & forward to get the image shown (I tested this in FFox & Chrome).    

grafik

dalf:
I have updated the TODO list in PR comment.

> OK, I understand the intention .. but on mobile, the UX is a bit strange and tiny to me .. may be it is easier to give up the intention?

The image is expected at the center of the screen, so what is the image is center vertically ? (even if I have issue to do that without only CSS)

There is also an issue when the phone screen is horizontal (the labels are too big and the image too small).

> When I use the Forward button (in the top right corner

Yes, it uses the same code than the `j` and `k` shortcuts.

searx/static/themes/simple/src/js/main/searx_results.js
```diff
@@ -31,17 +31,74 @@
...
+      scrollTop = document.documentElement.scrollTop || document.body.scrollTop,
+      results = d.getElementById('results');
       if (e !== null) {
-        if (scrollTop >= 200) {
-          e.style.opacity = 1;
+        if (scrollTop >= 100) {
+          results.classList.add('scrolling');
         } else {
-          e.style.opacity = 0;
+          results.classList.remove('scrolling');
```
dalf, line 98: If the viewport is not at the top of the document, `#results` has the CSS class `scrolling`.

This remove a style attribute usage for the `#backToTop` element.

dalf:
#### Desktop
image

When the user scrolls, the detail takes the full height. 

Note: when the user clicks on different images, the Javascript makes sure the selection is visible and may be scroll the window. It shouldn't do that with the detail opened, this is UX bug.

#### Tablet
image

#### Phone
image


It starts to be complete, what is missing:
* it is not easy to close the detail with a second click on image (`searx_keyboard.js` is actually about the current selection, `searx_results.js` is also about the current selection). (workaround: the Esc key)
* the phone layout needs some care, but I wish to do that in another PR: it works and it is improvement of the master branch. What comes in my mind: 
  * The image is now centered vertically, is it really ugly with the labels at the bottom ? the labels needs something like a border or a line at the top.
  * Most probably the next / previous button need to be on the right and left sides ?
  * the `<meta name="viewport"` tag needs `initial-scale=1` (in [base.html](https://github.com/searxng/searxng/blob/2624034cd6f417b52ed3f9f3f5226720fa0b53a1/searx/templates/simple/base.html#L10)):
    * horizontal layout: the image is really small, and the labels are huge.
    * vertical layout: The labels are too small.

mrpaulblack:
Yeah tested your PR again and it works fine...
Actually your phone layout works fine as long as touch input is disabled... (I am having the same scaling issues with the search bar on mobile, but only when touch input is enabled; so its probably something with that).

So yeah works great on my PC, apart from the phone scaling and the issues you already mentioned...

Phone without touch input:
image

Phone with touch input:
image

[EDIT] also I think its better to have the back and forth buttons both on the right because its less mouse travel between them that way...

Alos some vars are only in the light theme, but should also be in the dark theme IMO

searx/static/themes/simple/src/less/definitions.less
```diff
@@ -156,6 +164,9 @@ html {
     --color-result-engines-font: #777;
     --color-result-search-url-border: #333;
     --color-result-search-url-font: #fff;
+    --color-result-detail-font: #fff;
+    --color-result-detail-label-font: lightgray;
+    --color-result-detail-background: #000;
```
mrpaulblack, line 170: There are some vars missing in the dark theme that are part of the light theme. I mean you could only add them to the light them, because they stay the same, but I think its better to add them to both themes.

dalf: Done in commit fd374d6322c6f0919832b6465bc311dac68bb753

dalf:
according to https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#viewport_basics , the typical value for viewport is 
```html
<meta name="viewport" content="width=device-width, initial-scale=1">
```

With this value, the phone layout is bigger but also overflow. I think this is the way to go to fix the small font issue on phone.

PR #380, opened 2021-10-04
[fix] interface language zh_CN and zh_TW don't work

dalf:
## What does this PR do?

before this commit, when the user picks "Chinese, China" or "Chinese, Taiwan",
the locale fallbacks to English.

This commit fixes this issue (the translations branch must be updated)

## Why is this change important?

Bug fix

## Related issues

close #339

PR #382, opened 2021-10-05
Fix match_language issue to make zh-TW match to zh-Hant-TW

MarcAbonce:
## What does this PR do?

This fixes the issue in #380 where setting `'Accept-Language': 'zh-tw;q=0.8` fails to pick `zh_Hant_TW` as the locale.

The issue is that PyBabel separates locales with underscores but we use hyphens everywhere else.

I haven't tested this thoroughly yet. It looks like it shouldn't break anything else but I don't know anymore...

1. Set `'Accept-Language': 'zh-tw;q=0.8`
2. Go to `/preferences`.
3. Verify that both search language and UI locale are in `zh-TW` or `zh_Hant_TW`.
4. Check with other locales to make sure I didn't break something else.

## Related issues
Part of #380

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -380,9 +380,12 @@ def _match_language(lang_code, lang_list=[], custom_aliases={}):  # pylint: disa
         return lang_code
 
     # try to get the most likely country for this language
-    subtags = get_global('likely_subtags').get(lang_code)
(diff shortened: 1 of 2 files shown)


return42:
FYI this patch has been merged into #380 ..

PR #383, opened 2021-10-05
SearXNG: docker

dalf:
## What does this PR do?

SearXNG brand: the docker image
* installs the file in `usr/local/searxng` (inside the image)
* uses `searxng` user
* uses the volume `/etc/searxng`. 

searx compatibility:
* If `/etc/searx/settings.yml` exists, it is copied to `/etc/searxng`
* If `/etc/searx/uwsgi.ini` exists, is it ignored (the content requires some updates)
* If `/etc/searx/settings.yml` or `/etc/searx/uwsgi.ini` exist:
   * a warning is logged
   * the file `/etc/searx/deprecated_volume_read_me.txt` is created, and notice the administrator.

The documentation is updated accordingly.

See https://github.com/searxng/searxng/pull/358

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,25 @@
 FROM alpine:3.14
-ENTRYPOINT ["/sbin/tini","--","/usr/local/searx/dockerfiles/docker-entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
 EXPOSE 8080
 VOLUME /etc/searx
+VOLUME /etc/searxng
 VOLUME /var/log/uwsgi
 
-ARG SEARX_GID=977
-ARG SEARX_UID=977
+ARG SEARXNG_GID=977
+ARG SEARXNG_UID=977
 
-RUN addgroup -g ${SEARX_GID} searx && \
-    adduser -u ${SEARX_UID} -D -h /usr/local/searx -s /bin/sh -G searx searx
+RUN addgroup -g ${SEARXNG_GID} searxng && \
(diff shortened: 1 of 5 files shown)


unixfox:
I haven't had the time to test this PR but I guess the major test is to install a working Searx container then migrate to a searxng container and see if it still works.

return42:
> * If `/etc/searx/settings.yml` exists, it is copied to `/etc/searxng`

Related issue: https://github.com/searxng/searxng/pull/401

@dalf should we rebase this PR first .. to get #401 ?

dalf:
> @dalf should we rebase this PR first .. to get #401 ?

In the docker image, SearXNG uses `/etc/searxng/settings.yml` and falls back to `/etc/searx/settings.yml`.

But before the application starts, if `/etc/searxng/settings.yml` doesn't exist then it is copied from `/usr/local/searxng/searx/settings.yml` ( see `dockerfiles/docker-entrypoint.sh` ) So the fall back to `/etc/searx/settings.yml` never happens.

That's why, in this PR `/etc/searx/settings.yml` is copied to `/etc/searxng/settings.yml`, in addition `/etc/searx/deprecated_volume_read_me.txt` contains a warning.

The test is:
* install https://github.com/searx/searx-docker
* ./start.sh
* ./stop.sh
* change `./searx/settings.yml` ( the instance name for example ).
* change the docker image from `searx/searx` to `searxng/searxng`.
* ./start.sh
* check the instance name is still changed.
* check there is `./searx/deprecated_volume_read_me.txt`.

dalf:
It works on a new VM as expected with only one issue: in settings.yml, `brand.new_issue_url` is required, so SearXNG doesn't start. Once `brand.new_issue_url` is set, everything is okay.

Issue #312, opened 2021-09-10
searx.utils.load_engine: remove  sys.modules[modname] = module


dalf:
See this example:
```python
from searx.engines import load_engine
from searx.utils import get_engine_from_settings

azlyrics = load_engine(get_engine_from_settings('azlyrics'))
archive_is = load_engine(get_engine_from_settings('archive is'))
searx = load_engine(get_engine_from_settings('searx'))
print(azlyrics, azlyrics.name)
print(archive_is, archive_is.name)
print(searx)

import xpath
import searx_engine

print(xpath, xpath.name)
print(searx_engine, searx_engine.name)
```

```
<module 'xpath' from '/home/alexandre/code/searx/searx/engines/xpath.py'> azlyrics
<module 'xpath' from '/home/alexandre/code/searx/searx/engines/xpath.py'> archive is
<module 'searx_engine' from '/home/alexandre/code/searx/searx/engines/searx_engine.py'>
<module 'xpath' from '/home/alexandre/code/searx/searx/engines/xpath.py'> archive is
<module 'searx_engine' from '/home/alexandre/code/searx/searx/engines/searx_engine.py'> searx
```

It works, but if an engine is named after an existing module, the former overrides the later.

So in
https://github.com/searxng/searxng/blob/76e0f6807ccddb12ca9efe3b2343735d6e1777d4/searx/utils.py#L428-L438

these lines should be removed:
```
    if modname in sys.modules:
        del sys.modules[modname]

    sys.modules[modname] = module
```

`importlib.util.spec_from_file_location` still requires a module name.
I'm not sure if this is important or not, and if yes, what naming convention to choose.

PR #386, opened 2021-10-06
[mod] set 'engine.supported_languages' from the origin python module

### LINKED ISSUE #384: Bug: google engine with specified "language: it_IT"
Branch: master
Version: 1.0.0-885-feb2d4dd

I used searxng-docker.

**What happened?**

I tried to set up the Google engine with `language: it_IT` but it crashes.
**How To Reproduce**

Add this engine:
```
-  name: google italian
   engine: google
   shortcut: goit
   language: it_IT
   use_mobile_ui: false
```
and try to perform a search.

**Expected behavior**

Results from the google engine with a specified language.

**Additional context**
I followed the [guide in the wiki](https://searxng.github.io/searxng/user/search_syntax.html#multilingual-search) to add multilingual search to the letter but it does not work.

**Technical report**

Error
  * Error: AttributeError
  * Percentage: 100
  * Parameters: `()`
  * File name: `searx/engines/google.py:232`
  * Function: `get_lang_info`
  * Code: `ret_val['params']['hl'] = lang_list.get(lang_country, language)`

return42:
## What does this PR do?

Set `engine.supported_languages` from the origin python module

The key of the dictionary 'searx.data.ENGINES_LANGUAGES' is the *engine name*
configured in settings.xml.  When multiple engines are configured to use the
same origin engine (e.g. `engine: google`)::

    - name: google
      engine: google
      use_mobile_ui: false
      ...

    - name: google italian
      engine: google
      use_mobile_ui: false
      language: it
      ...

    - name: google mobile ui
      engine: google
      shortcut: gomui
      use_mobile_ui: true

There exists no entry for ENGINES_LANGUAGES[engine.name] (e.g. `name: google
mobile ui` or `name: google italian`).  This issue can be solved by recreate the
ENGINES_LANGUAGES::

    make data.languages

But this is nothing an SearXNG admin would like to do when just configuring
additional engines, since this just doubles entries in ENGINES_LANGUAGES and
BTW: `make data.languages` has various external requirements which might be not
installed or not available, on a production host.

With this patch, if engine.name fails, ENGINES_LANGUAGES[engine.engine] is used
to get the engine.supported_languages (e.g. `google` for the engine named
`google mobile`).

For an engine, when there is `language: ...` in the YAML settings, the engine
supports only one language, in this case engine.supported_languages should
contains this value defined in settings.yml (e.g. `it` for the engine named
`google italian`).

Activate the engines shown above in the settings.yml, call `make run` and do some queries e.g.  `!goit foo`, `!goit foo :en`, `!gomui foo :en`, `!gomui foo :it` and `foo`

## Related issues

Closes: https://github.com/searxng/searxng/issues/384

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -137,7 +137,7 @@ def update_engine_attributes(engine, engine_data):
             if isinstance(param_value, str):
                 param_value = list(map(str.strip, param_value.split(',')))
             engine.categories = param_value
-        elif param_name != 'engine':
+        else:
             setattr(engine, param_name, param_value)
 
     # set default attributes
@@ -147,10 +147,34 @@ def update_engine_attributes(engine, engine_data):
 
 
 def set_language_attributes(engine):
-    # pylint: disable=protected-access
     # assign supported languages from json file
     if engine.name in ENGINES_LANGUAGES:
         engine.supported_languages = ENGINES_LANGUAGES[engine.name]
+
+    elif engine.engine in ENGINES_LANGUAGES:
+        # The key of the dictionary ENGINES_LANGUAGES is the *engine name*
+        # configured in settings.xml.  When multiple engines are configured in
+        # settings.yml to use the same origin engine (python module) these
+        # additional engines can use the languages from the origin engine.
+        # For this use the configured ``engine: ...`` from settings.yml
+        engine.supported_languages = ENGINES_LANGUAGES[engine.engine]
+
+    if hasattr(engine, 'language'):
+        # For an engine, when there is `language: ...` in the YAML settings, the
+        # engine supports only one language, in this case
+        # engine.supported_languages should contains this value defined in
+        # settings.yml
+        if engine.language not in engine.supported_languages:
+            raise ValueError(
+                "settings.yml - engine: '%s' / language: '%s' not supported" % (
+                    engine.name, engine.language ))
+
+        if isinstance(engine.supported_languages, dict):
+            engine.supported_languages = {
+                engine.language : engine.supported_languages[engine.language]
+            }
+        else:
+            engine.supported_languages = [engine.language]
 
     # find custom aliases for non standard language codes
     for engine_lang in engine.supported_languages:
@@ -172,6 +196,7 @@ def set_language_attributes(engine):
             'Accept-Language': 'ja-JP,ja;q=0.8,en-US;q=0.5,en;q=0.3',  # bing needs a non-English language
         }
         engine.fetch_supported_languages = (
+            # pylint: disable=protected-access
             lambda: engine._fetch_supported_languages(
                 get(engine.supported_languages_url, headers=headers))
         )
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -601,6 +601,17 @@ engines:
     # additional_tests:
     #   android: *test_android
 
+  # - name: google italian
+  #   engine: google
+  #   shortcut: goit
+  #   use_mobile_ui: false
+  #   language: it
+
+  # - name: google mobile ui
+  #   engine: google
+  #   shortcut: gomui
+  #   use_mobile_ui: true
+
   - name: google images
     engine: google_images
     shortcut: goi


searx/engines/__init__.py
```diff
@@ -149,8 +149,13 @@ def update_engine_attributes(engine, engine_data):
 def set_language_attributes(engine):
     # pylint: disable=protected-access
     # assign supported languages from json file
+
     if engine.name in ENGINES_LANGUAGES:
         engine.supported_languages = ENGINES_LANGUAGES[engine.name]
+    else:
+        # search for the configured engine from YAML setting: ``engine: ...``
+        if engine.engine in ENGINES_LANGUAGES:
```
dalf: Nitpicking, lines 155 to 157 can be written:
```python
elif engine.engine in ENGINES_LANGUAGES:
```

return42: OK, I will fix .. another question, instead of `engine.engine` we can also use `engine.__name__` what do you think is better?

dalf: `engine.__name__` is related to https://github.com/searxng/searxng/issues/312

dalf:
This PR close #384 

----

Note : even with this configuration
```yaml
-  name: google italian
   engine: google
   shortcut: goit
   language: it_IT
   use_mobile_ui: false
```

`_fetch_supported_languages` returns all the languages :
https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/engines/google.py#L389-L401

-->

`_is_selected_language_supported` is going to return `True` for many languages (this function is only used to display a check mark in the `/preferences`, column `Supports selected language`) 

https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/webapp.py#L1055-L1065

---

Looking at 
https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/search/processors/abstract.py#L152-L156

An alternative to this PR is
```python
if hasattr(self.engine, 'language') and self.engine.language: 
  engine.supported_languages = [ engine.language ]
elif engine.name in ENGINES_LANGUAGES:
  engine.supported_languages = ENGINES_LANGUAGES[engine.name]
```

searx/engines/__init__.py
return42: > `engine.__name__` is related to https://github.com/searxng/searxng/issues/312

not really .. #312 is talking about registering modules in sys.modules, when loaded by `searx.utils.load_module(filename, module_dir)` .. the `engine.__name__` is set by `importlib` anyway .. and is not related to sys.modules.

return42:
> An alternative to this PR is

I guess I understand only the half of your last comment ... the problem of #384 is, that I need to run `make data.languages` when I configure an alternative engine to an _origin_ engine (like shown in the example).

The `make data.languages` uses the ` _fetch_supported_languages(resp)` from the google module, when it creates entries for the `google italian` engine (this worked out of the box).

But the question is, is it really useful to call `make data.languages`  every time I create such an engine like `google italian`?
My Answer is "no" since we have the needed values already stored in `ENGINES_LANGUAGES`, only the `engine.name` is the wrong key .. the correct key for such engines is `engine.__name__`, the name of the python module (what I called the _origin_ engine).  BTW: `make data.languages` has various external requirements which might be not installed or not available, on a production host.

I think the _get and set_ of `ENGINES_LANGUAGES` is not related to the processor (`searxng/searx/search/processors/abstract.py`) and the fix should not placed there 

... or do I miss something?

dalf:
For an engine, when there is `language: it_IT` in `settings.yml`, whatever the language the user asks for, the outgoing request is going to be the value defined in `settings.yml` (abstract.py, line 152 to 156) : the user choice is overwritten.

In other words, the engine supports (from the user point of view) only one language --> I think in this case `engine.supported_languages` should contains this value defined in `settings.yml`

return42:
> In other words, the engine supports (from the user point of view) only one language --> I think in this case `engine.supported_languages` should contains this value defined in `settings.yml`

Thanks for clarify! This brings me to another thought ..

What you say is correct for this example (when language is defined in the settings.yml engine) what about an engine like:

```yaml

  - name: google
    engine: google
    shortcut: go
    # see https://searxng.github.io/searxng/src/searx.engines.google.html#module-searx.engines.google
    use_mobile_ui: false

  - name: google mobile ui
    engine: google
    shortcut: gomui
    use_mobile_ui: true
```
I think we should cover both cases ...

dalf:
I agree.

```python
if hasattr(self.engine, 'language') and self.engine.language: 
  engine.supported_languages = [ engine.language ]
elif engine.name in ENGINES_LANGUAGES:
  engine.supported_languages = ENGINES_LANGUAGES[engine.name]
elif engine.engine in ENGINES_LANGUAGES:
  engine.supported_languages = ENGINES_LANGUAGES[engine.engine]
```

It should not create side effect for engines like `xpath` and `json_engine` since `ENGINES_LANGUAGES["xpath"]` and `ENGINES_LANGUAGES["json_engine"]` are not defined.

searx/engines/__init__.py
dalf:
```python
import searx
import searx.engines.google
print(searx.engines.google.__name___)
# searx.engines.google
print(searx.engines.engines['google'].__name__)
# google
```

Is it expected (?)
I _guess_ the reason of the second output is (?):
https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/utils.py#L434
where `modname == 'google'`

return42: > I guess the reason of the second output is (?)

I think so, importlib sets the `__name__` when the module is loaded  ... don't ask me any details, the "importlib" is mystic to me ... but you can repeat your example with this patch applied and the result is the same ... this was also what I have tested before I posted https://github.com/searxng/searxng/pull/386#discussion_r723089965

```diff
diff --git a/searx/utils.py b/searx/utils.py
index 22824d829..c9ec0bc8d 100644
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -427,13 +427,10 @@ def match_language(locale_code, lang_list=[], custom_aliases={}, fallback='en-US
 
 def load_module(filename, module_dir):
     modname = splitext(filename)[0]
-    if modname in sys.modules:
-        del sys.modules[modname]
     filepath = join(module_dir, filename)
     # and https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
     spec = importlib.util.spec_from_file_location(modname, filepath)
     module = importlib.util.module_from_spec(spec)
-    sys.modules[modname] = module
     spec.loader.exec_module(module)
     return module
 ```

> Is it expected (?)

I would expect, that the importlib stets the `__name__` whenever a module is loaded. This is at least what Python expect:

[ Within a module, the module’s name (as a string) is available as the value of the global variable `__name__`](https://docs.python.org/3/tutorial/modules.html)

~BTW: do you like to send a PR that applies the patch above (copied from your https://github.com/searxng/searxng/issues/312)~ .. see #390

**Comment by return42** [comment]
I tried first what you suggested in https://github.com/searxng/searxng/pull/386#issuecomment-935970710 but it feels not good to me, to place the initialisation of `engine.supported_languages` in the processor .. IMO this should be a part of the engine initialization.  This is what I implemented in 4aae7e537 .. what do you think about?

File: searx/engines/__init__.py [hunk 2]
```diff
@@ -146,11 +146,29 @@ def update_engine_attributes(engine, engine_data):
...
+        # configured in settings.xml.  When multiple engines are configured in
+        # settings.yml to use the same origin engine (python module) these
+        # additional engines can use the languages from the origin engine.
+        # For this use the configured ``engine: ...`` from settings.yml
+        engine.supported_languages = ENGINES_LANGUAGES[engine.engine]
+
+    # For an engine, when there is `language: ...` in the YAML settings, the
+    # engine supports only one language, in this case engine.supported_languages
+    # should contains this value defined in settings.yml
+    language = engine_data.get('language', None)
```
dalf: `engine_data.get('language')` since `None` is implicit.

searx/engines/__init__.py
```diff
@@ -146,11 +146,29 @@ def update_engine_attributes(engine, engine_data):
...
+    if language:
+        if isinstance(engine.supported_languages, dict):
+            engine.supported_languages = {language : engine.supported_languages[language]}
```
dalf: This code crashes when `engine.supported_languages[language]` doesn't exist and `engine.supported_languages` is dictionary.

So it crashes with (see [engine_languages.json](https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/data/engines_languages.json#L25446-L25451) ) :
```yaml
- engine: google
   language: zh-Hans-CN
   ...
```

But not with (see [engine_languages.json](https://github.com/searxng/searxng/blob/feb2d4dda575eb1501fb5a8427de364bed7bc179/searx/data/engines_languages.json#L237-L240) ):
```yaml
- engine: bing
   language: zh-Hans-CN
   ...
```

I suggest to replace line 168 by:
```python
if isinstance(engine.supported_languages, dict) and language in engine.supported_languages:
```

searx/engines/__init__.py
```diff
@@ -146,11 +146,29 @@ def update_engine_attributes(engine, engine_data):
...
+
     if engine.name in ENGINES_LANGUAGES:
         engine.supported_languages = ENGINES_LANGUAGES[engine.name]
+    elif engine.engine in ENGINES_LANGUAGES:
+        # The key of the dictionary ENGINES_LANGUAGES is the *engine name*
+        # configured in settings.xml.  When multiple engines are configured in
+        # settings.yml to use the same origin engine (python module) these
+        # additional engines can use the languages from the origin engine.
+        # For this use the configured ``engine: ...`` from settings.yml
+        engine.supported_languages = ENGINES_LANGUAGES[engine.engine]
```
dalf, line 160: Nitpicking / optional / not important, it can be :
```python
engine.supported_languages = ENGINES_LANGUAGES[engine_data["engine"]]
```

so the change on line 140 is not necessary.

dalf (changes requested):
I've added some comments.

---

Yes  `searx.engines.__init__` is the right location to initialize `engine.supported_languages` (Sorry, my comment was not clear).

searx/engines/__init__.py
return42: > This code crashes ..

I know, IMO it is a miss configuration when the configured language is not in the supported languages .. I also would like to add an exception when language is not in a list type:
```python
else:
    if language not in engine.supported_languages:
        raise KeyError(language)
    engine.supported_languages = [language] 
```

I looked  for an Exception-Type that fits better, but haven't found one .. may be should use SearxException?

https://github.com/searxng/searxng/blob/adeb084cf4ffc59f1e49b47d049e7541be6dd5dd/searx/exceptions.py#L19-L20

searx/engines/__init__.py
return42: I can fix it like you suggested .. but when I implemented 140, I looked back in the commit history and did not found a reason why `elif param_name != 'engine'` is better than a simple `else` .. IMO the `elif` is a relic from the old days which is not needed anymore.

searx/engines/__init__.py
dalf: `SearxSettingsException` ?

https://github.com/searxng/searxng/blob/master/searx/settings_loader.py#L22-L29

return42: This is what I thought first, but this exception type is only for loading the YAML fie and needs a `file_name` argument

    SearxSettingsException(e, file_name) f

searx/engines/__init__.py
dalf: I think the reason is redundancy `engine.__name__ == engine.engine`.

But as described, nitpicking.

searx/engines/__init__.py
dalf: ...

https://github.com/searxng/searxng/blob/adeb084cf4ffc59f1e49b47d049e7541be6dd5dd/searx/settings_defaults.py#L116-L132

`settings_defaults`:
* log details using `logger`
* raises a `ValueError` exception.

return42:
@dalf I _resolved_ your remarks .. the code now looks much simpler than before, see 19e41c137 / thanks!

dalf (approved):
Thank you!

PR #389, opened 2021-10-06
[nix] searxng recipe for installation via nix

efim:
This PR adds recipe for installing searxng via nix package manager, so that admins that use searx service on their nix managed servers could easily change it to searxng

## What does this PR do?

Adds:
- searxng package : "searxng" python description that's created with [mach-nix](https://github.com/DavHau/mach-nix), i.e extracts dependencies from `requirements.txt` of the project
- overlay for easier installation into nix configuration
- development shell with make & python with dependencies for local execution of built artifact

Searx is packaged for Nix, and can be installed & configured easily, [here sample](https://github.com/efim/nix-searx-container-example/blob/master/flake.nix)
This change would allow admins to easily use SearXNG with minimal modification of the configurations on their side.

If Nix is installed (available for Linux & MacOS), then
- modify `./searx/settings.yml` secret value to non-default
- `nix build` would build the project, and symlink it to `./result`
- `nix-shell` would create subshell with env var SEARX_SETTINGS_PATH pointing to `./searx/settings.yml` and add `searx-run` to PATH
- `searx-run` would launch built project

## Author's checklist

Using `mach-nix` to create recipe for SearXNG is different from the way Searx is packaged currently.
Mach nix takes exact versions of python dependencies from the `requirements.txt`, while [Searx definition](https://github.com/NixOS/nixpkgs/blob/nixos-21.05/pkgs/servers/web-apps/searx/default.nix#L71) takes versions of python libraries that are available in "nix store" and not necessarily match requirements (as I understand)

Also using mach-nix allows for easier maintaining of the nix recipe.
- `nix flake lock --update-input nixpkgs ` would pin "nixpkgs" to latest version
- any change to "requirements.txt" would automatically get picked up by the `nix build`

Downside: I spent quite a while trying to figure out how to add plugins to SearXNG that is packaged this way and wasn't successful, yet hopeful that will be able to figure it out in the future.
For now [nix packaged external Searx plugin](https://github.com/return42/tgwf-searx-plugins/pull/1) would only work with Searx

I suspect that writing more extensive documentation on SearXNG installation with nix would be useful,
but I created this PR with code first, to get the feedback and comments

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,99 @@
+{
+  description = "SearXNG";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs";
+    mach-nix-src.url = "github:DavHau/mach-nix";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  inputs.flake-compat = {
+    url = "github:edolstra/flake-compat";
+    flake = false;
+  };
+
+  outputs = { self, nixpkgs, flake-utils, mach-nix-src, ... }:
+    let
+      # TODO : for some reason `nix flake show` results in error:
+      # ```error: a 'aarch64-darwin' with features {} is required to build '/nix/store/s79ixgs3xy250x1b9vyy90g7dzd6bsh4-conda-channels.json.drv',
+      #   but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}```
+      # Some internal part of mach-nix attempts to build derivation into target system and cross-build fails
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+      #
+      # Helper function to generate an attrset '{ x86_64-linux = f "x86_64-linux"; ... }'.
+      forAllSystems = f:
+        nixpkgs.lib.genAttrs supportedSystems (system: f system);
+
+      python = "python38";
+      pypiDataRev = "18250538b9eeb9484c4e3bdf135bc1f3a2ccd949";
+      pypiDataSha256 = "01hdjy5rpcjm92fyp8n6lci773rgbsa81h2n08r49zjhk9michrb";
+      requirements = builtins.readFile ./requirements.txt;
+
+      nixpkgsFor = forAllSystems (
+        system:
+          import nixpkgs {
+            inherit system;
+            overlays = [ self.overlay ];
+          }
+      );
+    in
+      {
+
+        overlay = final: prev:
+          let
+            mach-nix = import mach-nix-src {
+              pkgs = prev;
+              inherit python pypiDataRev pypiDataSha256;
+            };
+          in
+            {
+              searxng = (
+                mach-nix.buildPythonPackage {
+                  inherit requirements;
+                  pname = "SearXNG";
+                  version = "1.0";
+                  src = ./.;
+                }
+              );
+              dev-python =
+                  mach-nix.mkPython {
+                    inherit python;
+                    requirements = requirements;
+                  };
+            };
+        packages = forAllSystems (
+          system:
+            {
+              inherit (nixpkgsFor.${system}) searxng;
+            }
+        );
+
+        # The default package for 'nix build'. This makes sense if the
+        # flake provides only one package or there is a clear "main"
+        # package.
+        # Returns Searx with plugin enabled
+        defaultPackage = forAllSystems (system: self.packages.${system}.searxng);
+
+        # To use as `nix develop`
+        # will provide shell with
+        # - make utility for `make run` and `make test`
+        # - python with dependencies, to run searx built with nix
+        # - env var pointing to default repository searx config
+        devShell = forAllSystems (
+          system: nixpkgsFor.${system}.mkShell
+            rec {
+              buildInputs = with nixpkgsFor.${system}; [
+                dev-python
+                gnumake
+              ];
+              settingsFile = builtins.readFile ./searx/settings.yml;
+              SEARX_SETTINGS_PATH = "${settingsFile}";
+            }
+        );
+
+        nixosModules.searxng = { pkgs, ... }:
+          {
+            nixpkgs.overlays = [ self.overlay ];
+          };
+      };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -3,16 +3,20 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs";
-    mach-nix-src.url = "github:DavHau/mach-nix";
     flake-utils.url = "github:numtide/flake-utils";
+    pypi-deps-db = {
+      url = "github:DavHau/pypi-deps-db";
+      flake = false;
+    };
+    mach-nix-src.url = "github:DavHau/mach-nix";
+    mach-nix-src.inputs.pypi-deps-db.follows = "pypi-deps-db";
+    flake-compat = {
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -0,0 +1,99 @@
...
+        # - env var pointing to default repository searx config
+        devShell = forAllSystems (
+          system: nixpkgsFor.${system}.mkShell
+            rec {
+              buildInputs = with nixpkgsFor.${system}; [
+                dev-python
+                gnumake
+              ];
+              settingsFile = builtins.readFile ./searx/settings.yml;
+              SEARX_SETTINGS_PATH = "${settingsFile}";
```
dalf, line 100: `SEARX_SETTINGS_PATH` works, but the new value is `SEARXNG_SETTINGS_PATH`

dalf (commented):
> modify ./searx/settings.yml secret value to non-default

SearXNG looks for `settings.yml` in:
* `SEARXNG_SETTINGS_PATH` environment variable.
* `/etc/searxng/settings.yml` (actually not, but it should).
* `/etc/searx/settings.yml`
* `settings.yml` next to `webapp.py`

Perhaps, before the last step, it should try on `~/.config/searxng/settings.yml`?

> Mach nix takes exact versions of python dependencies from the requirements.txt

This is the thing to do: SearXNG is tightly couple to `httpx[http2]` and `httpx-socks[asyncio]`

> Downside: I spent quite a while trying to figure out how to add plugins to SearXNG that is packaged this way

What are the difficulties? May be we can help.
If the copy of the static files is a problem:
* define a deployement directory
* set settings.yml

The thing that would be interresting perhaps in a second step if how to deploy the application with uwsgi.

flake.nix
efim: Hello!

This line `settingsFile = builtins.readFile ./searx/settings.yml;` was intended by me for usage in development shell, 
so that locally build SearXNG could be launched easily after the repository checkout with only
1. Set new secret in ./searx/config.yml
2.
 ```
$ nix build
$ nix-shell
$ ./searx-run
```

And if a developer already has `SEARXNG_SETTINGS_PATH` or `SEARX_SETTINGS_PATH` set in their environment, then they can do
```
$ nix buld
# ./result/bin/searx-run
```
Since right now only reason `searx-run` is not successful is error because of default secret

It was not my intention to create a package of SearXNG with preattached configuration.

Usually admins that would install Searx with Nix would do so by providing configuration in the nix file, by setting some of the provided options ( [link with option definitions](https://github.com/NixOS/nixpkgs/blob/nixos-21.05/nixos/modules/services/networking/searx.nix) ,[ link with example configuration that installs searx & configures it](https://github.com/efim/nix-searx-container-example/blob/master/non-flake-config.nix#L49) )

And there is also some kind of configuration for optionally using uswgi in Searx module for NixOS, I suspect that it is being maintained.
As I understand (and tested a little bit) reusing same Searx service configuration with "pkg" modified to SearXNG seems to work well right now

**Comment by dalf** [comment]
I'm sorry for the delay.

I don't know nix enough, may be this error is not a issue:
* I have merge your PR in a local branch
* edited `searx/settings.yml`.
* `nix build`
* `nix-shell`
* `searx-run` : everything works fine.
* few days later: merge your PR based on another local branch based on the lastest master branch.
* `nix build` fails with this error:

```
[alexandre@alexandre-vm:~/code/searx] efim-master+* ± nix build
builder for '/nix/store/dmphxlx6nwhqijyy96s9x6hyvbwyx5ql-mach_nix_file.drv' failed with exit code 1; last 10 log lines:
  Mach-nix version: master
  Python: 3.8.11
  Cause: Requirements conflict: flask ((('==', '2.0.2'),),)
  The requirements which caused the error:
    flask ((('==', '2.0.2'),),)
  
  The given requirements might contain package versions which are not yet part of the dependency DB
  currently used. The DB can be updated by specifying 'pypiDataRev' when importing mach-nix.
  For examples see: https://github.com/DavHau/mach-nix/blob/master/examples.md
  
[0 built (1 failed)]
error: build of '/nix/store/dmphxlx6nwhqijyy96s9x6hyvbwyx5ql-mach_nix_file.drv' failed
(use '--show-trace' to show detailed location information)
```

**Comment by efim** [comment]
> I'm sorry for the delay.
> 
> I don't know nix enough, may be this error is not a issue:
> 
>     * I have merge your PR in a local branch
> 
>     * edited `searx/settings.yml`.
> 
>     * `nix build`
> 
>     * `nix-shell`
> 
>     * `searx-run` : everything works fine.
> 
>     * few days later: merge your PR based on another local branch based on the lastest master branch.
> 
>     * `nix build` fails with this error:
> 
> 
> ```
> [alexandre@alexandre-vm:~/code/searx] efim-master+* ± nix build
> builder for '/nix/store/dmphxlx6nwhqijyy96s9x6hyvbwyx5ql-mach_nix_file.drv' failed with exit code 1; last 10 log lines:
>   Mach-nix version: master
>   Python: 3.8.11
>   Cause: Requirements conflict: flask ((('==', '2.0.2'),),)
>   The requirements which caused the error:
>     flask ((('==', '2.0.2'),),)
>   
>   The given requirements might contain package versions which are not yet part of the dependency DB
>   currently used. The DB can be updated by specifying 'pypiDataRev' when importing mach-nix.
>   For examples see: https://github.com/DavHau/mach-nix/blob/master/examples.md
>   
> [0 built (1 failed)]
> error: build of '/nix/store/dmphxlx6nwhqijyy96s9x6hyvbwyx5ql-mach_nix_file.drv' failed
> (use '--show-trace' to show detailed location information)
> ```

Thank you for finding this out during the review!

That is an actual problematic situation, for which I was able to find a suitable (I think) solution

With the update from the latest commit process of reacting to such situation is to execute
`nix flake lock --update-input pypi-deps-db`

In the next paragraph I'll describe source of the situation and reasons why it occurs:
Flake is designed specifically to allow fixing of all build inputs, so that result of the build is the same

Mach-nix allows for one higher step of dependency declaration - each revision of "flake.nix" takes versions of inputs from "flake.lock", but also from "requirements.txt" of same revision.
So when dependencies in the "requirements.txt" are changed and committed - that means that whole "flake" dependencies are pinned to new versions.
Hovewer when only "requirements.txt" are changed - other inputs, such as database that's being used for nix recipes of the python dependencies is not updated, and never versions of dependencies might not be found in that (pinned to specific revision) database.

So sometimes updating "pypi-deps-db" is a useful thing.
Previously I've specified revision of that repository manually, since "mach-nix-src" doesn't always internally targets latest database revision (which is updated automatically twice a day) and without overriding of the version of the database sometimes dependencies cannot be resolved.

With current change it should be really easy to manage update of the database, `nix flake lock` command should make all necessary changes, if anytime in the future updated python dependencies would require "pypi-db" update

@dalf could you please check again, that your branch with this nix flake pulled in works either as is, or produces same error with that error being fixable by described `nix flake lock` command?

**Comment by dalf** [comment]
I have issues to get `nix flake` command.

I've followed https://nixos.wiki/wiki/Flakes for Non-NixOS : 
```
$ nix-env -iA nixpkgs.nixUnstable
replacing old 'nix-2.3.16'
installing 'nix-2.3pre6895_84de821'
$ nano ~/.config/nix/nix.conf  # to add experimental-features = nix-command flakes
$ nix flake
error: 'flake' is not a recognised command
```

dalf:
I close this PR for now mainly because nix flakes are still experimental (also I was not able to compile to build the package).

PR #390, opened 2021-10-06
[fix] don't mix loaded modules with imported modules (sys.modules)

### LINKED ISSUE #312: searx.utils.load_engine: remove  sys.modules[modname] = module

return42:
## What does this PR do?

Don't mix loaded modules with imported modules (sys.modules)

The utils.load_module() function is used to load a python file (aka module) and
return the module's namespace.  SearXNG uses this function to load *engines and
answerers* from arbitrary locations with arbitrary modifications.  These are not
real python modules and it is not intended to mix this *engines and answerers*
with the python modules registered in sys.modules.

## Related issues

Suggested-by: @dalf in https://github.com/searxng/searxng/issues/312
Closes: https://github.com/searxng/searxng/issues/312

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import sys
 import re
 import importlib
 
@@ -427,13 +426,10 @@ def match_language(locale_code, lang_list=[], custom_aliases={}, fallback='en-US
 
 def load_module(filename, module_dir):
     modname = splitext(filename)[0]
-    if modname in sys.modules:
-        del sys.modules[modname]
     filepath = join(module_dir, filename)
     # and https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
     spec = importlib.util.spec_from_file_location(modname, filepath)
     module = importlib.util.module_from_spec(spec)
-    sys.modules[modname] = module
     spec.loader.exec_module(module)
     return module
 


dalf:
Thank you !

I have postpone because of
* `spec = importlib.util.spec_from_file_location(modname, filepath)` : `modname` should be like `searxng.dynloading.something`.
* answers use this function, but they can be loaded in the same way than plugins using `importlib.import_module` (see also https://github.com/searxng/searxng/pull/319 )

But it closes the issue anyway.

return42:
> `spec = importlib.util.spec_from_file_location(modname, filepath)` : `modname` should be like `searxng.dynloading.something`.

Is it really needed? .. what will be the benefit? .. and why is `searxng.dynloading.something` better than `searxng.engines.something`?

I would not recommend to set the `__name__` to any special before it is clear what the benefit is .. may be I missed it ...

PR #391, opened 2021-10-07
feat: Add torznab engine

### LINKED ISSUE #385: Jackett engine request
**Working URL to the engine**
Jackett is a self hosted application.

Jackett is the standard the facto to aggregate torrent search. Searx already sypports some torrent websites but using jackett it would open the possibility to search a lot of indexers, expecially private ones.

**Features of this engine**

Jackett is open source and self hostable.

Files

paolobasso99:
## What does this PR do?
Add a torznb engine.

SearXNG already supports some torrent trackers.
Adding support for the torznab apis opens the possibility to use [Jackett](https://github.com/Jackett/Jackett) and [Prowlarr](https://github.com/Prowlarr/Prowlarr) to query a lot of public and private trackers.

In order to test this PR locally it is needed to:

- install [Jackett](https://github.com/Jackett/Jackett)
- add a torrent tracker
- add the engine to `settings.yml` as:
  ```
  - name: jackett
    engine: torznab
    shortcut: jkt
    base_url: http://localhost:9117/api/v2.0/indexers/all/results/torznab
    enable_http: true # if using localhost
    api_key: xxxxxxxxxxxxxxxxx
  ```
- test the engine performing a file search

## Related issues
Closes #385 (created by me)

### DIFF
--- a/searx/engines/torznab.py
+++ b/searx/engines/torznab.py
@@ -35,10 +35,11 @@ api_key = ''
 # https://newznab.readthedocs.io/en/latest/misc/api/#predefined-categories
 torznab_categories = []
 
+def init(engine_settings=None): # pylint: disable=unused-argument
+    if len(base_url) < 1:
+        raise ValueError('missing torznab base_url')
 
 def request(query, params):
-    if len(base_url) < 1:
-        raise SearxEngineAPIException('missing torznab base_url')
 
     search_url = base_url + '?t=search&q={search_query}'
     if len(api_key) > 0:
@@ -47,13 +48,12 @@ def request(query, params):
         search_url += '&cat={torznab_categories}'
 
     params['url'] = search_url.format(
-        search_query=quote(query),
-        api_key=api_key,
-        torznab_categories=",".join(torznab_categories)
+        search_query = quote(query),
+        api_key = api_key,
+        torznab_categories = ",".join([str(x) for x in torznab_categories])
     )
 
     return params
-
 
 def response(resp):
     results = []
--- a/searx/engines/torznab.py
+++ b/searx/engines/torznab.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
 """Torznab WebAPI
 
 A engine that implements the `torznab WebAPI`_.
@@ -7,9 +8,9 @@ A engine that implements the `torznab WebAPI`_.
 
 """
 
+from datetime import datetime
 from urllib.parse import quote
 from lxml import etree
-from datetime import datetime
 
 from searx.exceptions import SearxEngineAPIException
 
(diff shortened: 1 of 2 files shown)


searx/engines/torznab.py
```diff
@@ -0,0 +1,133 @@
...
+
+        result["publishedDate"] = None
+        try:
+            result["publishedDate"] = datetime.strptime(
+                get_property(item, 'pubDate'), '%a, %d %b %Y %H:%M:%S %z')
+        except (ValueError, TypeError) as e:
+            pass
+
+        result["seed"] = get_torznab_attr(item, 'seeders')
+        result["leech"] = get_torznab_attr(item, 'leechers')
```
return42, line 114: I dont know why, but I never see 'leechers' in the XML, even if there are some .. e.g. this is the result from my jackett instance ..

grafik

And this is the same result (except leechers) in SearXNG ..

grafik

I also checked the XML response on various queries but never see a `<torznab:attr name="leechers" value="..." />` tag.  May be there is something wrong in my Jackett instance?

searx/settings.yml
```diff
@@ -1643,6 +1643,19 @@ engines:
...
+# opens the possibility to query a lot of public and private indexers directly from searXNG.
+#  - name: torznab
+#    engine: torznab
+#    shortcut: trz
+#    base_url: http://localhost:9117/api/v2.0/indexers/all/results/torznab
+#    enable_http: true # if using localhost
+#    api_key: xxxxxxxxxxxxxxx
+#    torznab_categories: # optional
+#     - 2000
+#     - 5000
```
return42: Do you know where we can get a full list of supported categories (numbers and meaning)?

paolobasso99: Yes
The categories supported by Jackett are: https://github.com/Jackett/Jackett/wiki/Jackett-Categories

return42 (changes requested):
FYI: on top of your branch I pushed two commits with marginal fix / are these changes OK for you?

Beside the comment about _leechers_ I can't find any issue more.  Did _leechers_ work in your tests?

Thanks a lot for this PR!

paolobasso99:
While the `leechers` attribute exists I think that Jackett just compute the number of leechers as `peers - seeds`.
I added a commit to reflect this behaviour when the attribute is missing.

return42:
> I added a commit to reflect this behaviour when the attribute is missing.

Cewl .. thanks / that was what I also thought .. I will continue with my tests.  May be I squash some commits before merge.

Error handling works !

grafik

return42 (approved):

return42:
@paolobasso99 this was your first PR to SearXNG, excellent work I would say .. thanks a lot!!!

paolobasso99:
Thanks!
I am using SearXNG as my main search engine now so if I will need some other engine/feature I will try to add it myself.

return42:
> so if I will need some other engine/feature I will try to add it myself.

that's good .. we need contributors .. feel welcome :)

OT: you can now drop your [torznab_engine branch](https://github.com/paolobasso99/searxng/tree/torznab_engine)

sbstratos79:
I have opened 2 issues about this engine. #1124 & #1125 
#1124 is especially important if somebody decides to use this to serve results from a private tracker as it could potentially lead to the instance owner's private tracker's account getting banned.

PR #401, opened 2021-10-08
SearXNG: load /etc/searxng/settings.yml

dalf:
## What does this PR do?

Try to load `/etc/searxng/settings.yml`
If it is not possible, try to load `/etc/searx/settings.yml`

There is a second commit to reference /etc/searxng/settings.yml (after #383 , I _feel_ that the point where the solution is a bing bang like #358 )

SearXNG brand, see #358


docs/admin/engines/settings.rst
```diff
@@ -4,7 +4,7 @@
 ``settings.yml``
 ================
 
-This page describe the options possibilities of the :origin:`searx/settings.yml`
+This page describe the options possibilities of the :origin:`searxng/settings.yml`
```
return42: This file has not been moved to `<root>/searxng` right now .. we should rename this line not before the "big bang" is done.

docs/admin/engines/settings.rst
```diff
@@ -25,10 +25,10 @@ settings.yml location
 The initial ``settings.yml`` we be load from these locations:
 
 1. the full path specified in the ``SEARXNG_SETTINGS_PATH`` environment variable.
-2. ``/etc/searx/settings.yml``
+2. ``/etc/searxng/settings.yml``
 
 If these files don't exist (or are empty or can't be read), SearXNG uses the
-:origin:`searx/settings.yml` file.  Read :ref:`settings use_default_settings` to
+:origin:`searxng/settings.yml` file.  Read :ref:`settings use_default_settings` to
```
return42: Same here and on other places ..

manage
```diff
@@ -107,9 +107,9 @@ buildenv() {
     # settings file from repository's working tree are used by default
     SEARXNG_SETTINGS_PATH="${REPO_ROOT}/searx/settings.yml"
 
-    if [ -r '/etc/searx/settings.yml' ]; then
-        if ask_yn "should settings read from: /etc/searx/settings.yml"; then
-            SEARXNG_SETTINGS_PATH='/etc/searx/settings.yml'
+    if [ -r '/etc/searxng/settings.yml' ]; then
```
dalf, line 114: Temporary: it should also try on `/etc/searx/settings.yml` too.

return42:
@dalf to test the shell scripts, I need to push on top of your branch .. I will inform you when I have finished my tests.

dalf:
About the last commit: there is already a. PR about Docker. This PR declares two volumes: /etc/searxng and /etc/searx so the transition can done easily even with automatic updates (like watchtower).

return42:
Ooops .. now I see #383, I will remove 0b86ff297 ...

manage
return42: I added an error (deprecated) message.

return42 (approved):
I added 3385627 and tested the installation scripts from ./utils .. for me it works / approved!

return42:
> after #383 , I feel that the point where the solution is a bing bang like #358 )

I am going to merge this PR  / to see what I have to do in the ./utils scripts before we start the "big bang"

PR #403, opened 2021-10-09
Zlibrary engine support

paolobasso99:
## What does this PR do?

Add support for a zlibrary engine.
I used a python file insead of the xpath engine in order to display a more descriptive content with authors, publisher, year, file type, dimension, language and ratings.

According to Wikipedia:

`As of 15 September 2021, Z-Library states that it provides access to more than 8,533,000 books and 84,837,000 articles.[1] According to Z-Library, it is "the world's largest ebook library" for eBooks`

So it would be awesome to query this large library directly from SearXNG.

Pull and search for a book using `!zlib`.

## Additional context
Zlibrary supports filtering by language but a list of supported languages is found only nested inside a dificult to scrape html script tag in a search result page. For the moment I did not bother to add language support but ideas are welcome.

### DIFF
--- a/searx/engines/zlibrary.py
+++ b/searx/engines/zlibrary.py
@@ -1,0 +1,98 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Z-Library
+
+Z-Library uses regional domains (see https://z-lib.org). Known ``base_url:``
+
+- base_url: https://b-ok.cc
+- base_url: https://de1lib.org
+- base_url: https://booksc.eu does not have cover preview
+- base_url: https://booksc.org does not have cover preview
+
+"""
+
+from urllib.parse import quote
+from lxml import html
+
+from searx.utils import extract_text, eval_xpath
+from searx.network import get as http_get
+
+# about
+about = {
+    "website": "https://z-lib.org",
+    "wikidata_id": "Q104863992",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+categories = ['files']
+paging = True
+base_url = ''
+
+def init(engine_settings=None):
+    global base_url # pylint: disable=global-statement
+
+    if "base_url" not in engine_settings:
+        resp = http_get('https://z-lib.org', timeout=5.0)
+        if resp.ok:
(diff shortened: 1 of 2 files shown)


searx/engines/zlibrary.py
```diff
@@ -0,0 +1,70 @@
...
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+categories = ['files']
+paging = True
+# z-library uses regional domains. https://z-lib.org
+base_url = 'https://b-ok.cc'
```
dalf: Should the engine loads `https://z-lib.org/` on [initialization](https://github.com/searxng/searxng/blob/7f72844ca679d64e56412c4b86f17d1b3b526ae0/searx/engines/mysql_server.py#L23-L38) and then use the URL from the `Books` link?

searx/engines/zlibrary.py
```diff
@@ -0,0 +1,70 @@
...
+                item, './/div[@title="Publisher"]')),
+            file_type=extract_text(eval_xpath(
+                item, './/div[contains(@class, "property__file")]//div[contains(@class, "property_value")]')),
+            language=extract_text(eval_xpath(
+                item, './/div[contains(@class, "property_language")]//div[contains(@class, "property_value")]')),
+            book_rating=extract_text(eval_xpath(
+                item, './/span[contains(@class, "book-rating-interest-score")]')),
+            book_quality=extract_text(eval_xpath(
+                item, './/span[contains(@class, "book-rating-quality-score")]')),
+        )
```
dalf: The cover can be added with something like:
`result["img_src"] = extract_text(eval_xpath(item, './/img[contains(@class, "cover")]/@src'))`

paolobasso99: Added a commit (673ef3854889219a421074943407eee0c780ef5b) to include the thumbnail.

searx/engines/zlibrary.py
paolobasso99: See 4e819263b3f27cdb2b779f23f1ee4dab9e676fe9

return42:
@paolobasso99  thanks for this additional engine! .. I will squash & rebase your branch / BTW: I will change the MS-Win CR-LF to unix LF endings.

PR #407, opened 2021-10-14
[fix] language param for qwant

mrpaulblack:
## What does this PR do?

Qwant returns low quality results in the wrong language, when the language is set to en-US (the request is done as `en_us`). This PR fixes this issue by returning the language with an uppercase country code like this: `en_US`. (If found this out after trying around with https://www.qwant.com/?locale=en_US&q=c%2B%2B+arrays&t=web; so on the Qwant site itself its the en_US syntax; so its probably the same for the API?)

Higher quality results with Qwant when using en-US language.

```make run```

## Author's checklist

Before this PR:
image

With this PR applied:
image

## Related issues

Related to fe67f1478f2f6b0939f378e7a1c65b7303f35763

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -85,14 +85,14 @@ def request(query, params):
 
     # add language tag
     if params['language'] == 'all':
-        params['url'] += '&locale=en_us'
+        params['url'] += '&locale=en_US'
     else:
         language = match_language(
             params['language'],
             supported_languages,
             language_aliases,
         )
-        params['url'] += '&locale=' + language.replace('-', '_').lower()
+        params['url'] += '&locale=' + language.replace('-', '_')
 
     params['raise_for_httperror'] = False
     return params


return42:
>  Higher quality results with Qwant when using en-US language.

Not only when using en_US .. seems that qwant-web do not like the _lowercase country codes_ in the URL (I tested with `de_de` / `de_DE`) .. while the response is always lowercase (`de_de`).

grafik

I also tested, if the effect the same on qwant-news, qwant-images, qwant-videos (see config below). But when my analysis is correct, these engines do not care about lower- or uppercase.

Seems qwant-web is different and has some quirks other do not have.

https://github.com/searxng/searxng/blob/ee5e9f9e1d488d8d24086b8111ac6ecff4abd77a/searx/settings.yml#L1072-L1099

return42 (approved):
Thanks!!

PR #419, opened 2021-10-18
[simple theme] less vars for different device sizes

mrpaulblack:
## What does this PR do?

Replace the hardcoded device size in the media queries of the simple theme with less vars defined in definition.less

Makes the queries easier to understand and now the values can be changes easily in definition.less.

```make run```

## Related issues

#418

### CODE REVIEW

File: searx/static/themes/simple/src/less/autocomplete.less [hunk 1]
```diff
@@ -60,7 +60,7 @@
   }
 }
 
-@media screen and (max-width: @results-width) {
+@media screen and (max-width: @phone) {
```
dalf, line 63: Not issue: it changes the value from 45rem to 50em.

dalf:
Thank you!

searx/static/themes/simple/src/less/autocomplete.less
mrpaulblack: Yes I am aware; I tested this an it makes more sense to make the autocomplete change when going to phone size instead of the 45em result width...

PR #423, opened 2021-10-21
[enh] reduce the available language list

### LINKED ISSUE #406: Remove unused languages from list
How can I remove the ton of languages from the list?
I only use 2.

If that is not possible, can we please get a setting to define `available_languages` in yaml?

Thanks,

dalf:
## What does this PR do?

In settings.yml:
```yml
search:
  languages:
   - es
   - en-GB
```

Reduce the language list:
image

image

Note: `all` is always included.

A locale instance, it can be easier to have a small set of language choices.

* in `settings.yml`, uncomment the `search.languages` section
* `make run`
* check the language list
* select the default language in the preference, save the preferences, check the right value is saved.
* select another language, save the preferences, check the right value is saved.

## Author's checklist

Note: in the master branch, I don't know how the `all` choice get accepted by `searx.preferences.SearchLanguageSetting` (I had to add `['all', '']` instead of `['']` in `searx.preferences.Preferences.__init__`this PR can 

Alternative: the API can accept all the language, only the UI shows a reduced list.

## Related issues

close #406

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -20,6 +20,13 @@ search:
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: ""
+  # Available languages
+  # languages:
+  #  - all
+  #  - es
+  #  - de
+  #  - it-IT
+  #  - en-GB
   # ban time in seconds after engine errors
   ban_time_on_fail: 5
   # max ban time in seconds after engine errors
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -17,7 +17,7 @@ searx_dir = abspath(dirname(__file__))
 
 logger = logging.getLogger('searx')
 OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']
-LANGUAGE_CODES = ('', 'all') + tuple(l[0] for l in languages)
+LANGUAGE_CODES = ['all'] + list(l[0] for l in languages)
 OSCAR_STYLE = ('logicodev', 'logicodev-dark', 'pointhi')
 CATEGORY_ORDER = [
     'general',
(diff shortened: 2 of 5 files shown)


return42:
> Note: in the master branch, I don't know how the `all` choice get accepted by `searx.preferences.SearchLanguageSetting` (I had to add `['all', '']` instead of `['']` in `searx.preferences.Preferences.__init__`

I think this line is what you missed ..

https://github.com/searxng/searxng/blob/cec9ded2e664da501955396a2d6c438ae8c117f3/searx/preferences.py#L20

I'm a bit confused, it seems we have defined it twice ...

https://github.com/searxng/searxng/blob/cec9ded2e664da501955396a2d6c438ae8c117f3/searx/settings_defaults.py#L20

searx/settings.yml
```diff
@@ -20,6 +20,10 @@ search:
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: ""
+  # Available languages in the UI
+  #languages:
+  #  - es
+  #  - en-GB
```
return42: Search-language and UI-language / not only UI

I suggest a comment like this .. (instead of `en-GB` lets use `en`)

```
  # Available languages
  # languages:
  #   - de
  #   - en
  #   - es
  #   - fr
  #   - it-IT
```

return42 (approved):
Works .. Thanks!!

searx/settings.yml
dalf: fixed in the last push-forced commit

dalf:
> I'm a bit confused, it seems we have defined it twice ...

I forgot this issue: I will update the PR.

searx/settings_defaults.py
```diff
@@ -98,6 +99,18 @@ def __call__(self, value: typing.Any) -> typing.Any:
...
         return value
 
+
+class SettingSublistValue(SettingsValue):
+    """Check the value is a sublist of type definition.
+    """
+
+    def check_type_definition(self, value: typing.Any) -> typing.Any:
+        if not isinstance(value, Collection):
+            raise ValueError('The value has to a list')
```
return42, line 108: Nit: "The value has to be a list type'

return42 (approved):
Really useful feature, thanks a lot! .. When I tested this PR one thought comes in mind; does it make sense if we set the `search.languages` in settings,yml according to the list we have translations for .. just one thought I had .. meanings?

searx/settings_defaults.py
dalf: fixed

dalf: > does it make sense if we set the search.languages in settings,yml according to the list we have translations

* settings.yml is updated by the user
* the language list is automatically updated

It won't mix up well over time: some user won't update the language list and will have issues. Exception: #432

PR #424, opened 2021-10-21
[mod] settings_defaults.py: default values for the brand section

dalf:
## What does this PR do?

Currently `brand.new_issue_url` must be a `str` but there is no default value:
https://github.com/searxng/searxng/blob/cec9ded2e664da501955396a2d6c438ae8c117f3/searx/settings_defaults.py#L143

This PR provides default values for the `brand` section.

Default values from the `brand` section makes the migration from searx easier.

## Author's checklist

A possible extension: make external links optional (link to searx.space, issues, etc...), so there are not displayed in the UI.

## Related issues

Related to https://github.com/searxng/searxng/pull/383#issuecomment-948453913

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -139,11 +139,11 @@ SCHEMA = {
         'contact_url': SettingsValue((None, False, str), None),
     },
     'brand': {
-        'issue_url': SettingsValue(str, None),
-        'new_issue_url': SettingsValue(str, None),
-        'docs_url': SettingsValue(str, None),
-        'public_instances': SettingsValue(str, None),
-        'wiki_url': SettingsValue(str, None),
+        'issue_url': SettingsValue(str, 'https://github.com/searxng/searxng/issues'),
+        'new_issue_url': SettingsValue(str, 'https://github.com/searxng/searxng/issues/new'),
+        'docs_url': SettingsValue(str, 'https://searxng.github.io/searxng'),
+        'public_instances': SettingsValue(str, 'https://searx.space'),
+        'wiki_url': SettingsValue(str, 'https://github.com/searxng/searxng/wiki'),
     },
     'search': {
         'safe_search': SettingsValue((0,1,2), 0),


searx/settings_defaults.py
```diff
@@ -139,11 +139,11 @@ def apply_schema(settings, schema, path_list):
...
-        'issue_url': SettingsValue(str, None),
-        'new_issue_url': SettingsValue(str, None),
-        'docs_url': SettingsValue(str, None),
-        'public_instances': SettingsValue(str, None),
-        'wiki_url': SettingsValue(str, None),
+        'issue_url': SettingsValue(str, 'https://github.com/searxng/searxng/issues'),
+        'new_issue_url': SettingsValue(str, 'https://github.com/searxng/searxng/issues/new'),
+        'docs_url': SettingsValue(str, 'https://searxng.github.io/searxng'),
+        'public_instances': SettingsValue(str, 'https://searx.space'),
+        'wiki_url': SettingsValue(str, 'https://github.com/searxng/searxng/wiki'),
```
return42, line 146: > Currently `brand.new_issue_url` must be a `str` but there is no default value

A "default" is in ...

https://github.com/searxng/searxng/blob/84447b87e118d914a39b6d5d5e4d4227aaade09f/searx/settings.yml#L6-L11

But I think it is not wrong to have an default in `settings_defaults.py` (this is also what we have done for other mandatory values).

return42 (approved):

searx/settings_defaults.py
dalf: The purpose is make the upgrade from searx easy: `brand.new_issue_url` is the only value that stops SearXNG to start from the searx version of `settings.yml`.

return42: Ah, you mean moving from an existing searx (https://github.com/searx/searx) installation to the SearXNG fork .. right?

dalf: yes. I took:
* start searx/searx-docker (but same without docker)
* change searx/searx to searxng/searxng
* start again 

without this PR, SearXNG doesn't start because of `brand.new_issue_url` missing value.

PR #426, opened 2021-10-21
[mod] simple theme: autocompletion doesn't use the style attribute.

### LINKED ISSUE #352: autocomplete-js used by simple theme is not CSP compliant
The [autocomplete-js](http://autocomplete-js.com/) lib used by the simple theme is not CSP compliant [(and old / last commit from 2018)](https://github.com/autocompletejs/autocomplete.js/commits/2.7).

The CSP issue is, that this lib set the style attributes by `setAttribute("style", ...)` like here:

https://github.com/searxng/searxng/blob/ca67f1555aa3c7dbb8dbb7ace7d4a0be9b65c717/searx/static/themes/simple/js/searx.js#L1419

This invokes the HTML parser and CSP is triggered / see [How to let script to use setAttribute 'style' without breaking CSP](https://stackoverflow.com/a/57633533) and [best practice](https://joji.me/en-us/blog/html-attribute-vs-dom-property/#md-best-practice)

dalf:
## What does this PR do?

Simple theme: change the autocompletion css and js not to use the style attribute.

The implementation has different issues:
* the right border can have a small gap.
* when the mouse is over the magnifier on the right, the 

image

image

image

* remove `style` attributes.
* if the two above issues are fixed, cleaner UI.

* `make themes`
* `make run`

## Related issues

close #352

### DIFF
--- a/searx/static/themes/simple/src/less/autocomplete.less
+++ b/searx/static/themes/simple/src/less/autocomplete.less
@@ -2,10 +2,13 @@
 
 .autocomplete {
   position: absolute;
+  width: 44.7rem;
+  margin-top: 2.4rem;
   max-height: 0;
   overflow-y: hidden;
   text-align: left;
-  .rounded-corners;
+  border-bottom-left-radius: 10px;
+  border-bottom-right-radius: 10px;
 
   &:active,
   &:focus,
@@ -24,7 +27,7 @@
 
     > li {
       cursor: pointer;
-      padding: 8px 0 8px 8px;
+      padding: 8px 0 8px 0.75rem;
 
       &.active,
       &:active,
@@ -50,9 +53,15 @@
     background-color: var(--color-autocomplete-background);
     color: var(--color-autocomplete-font);
     border: 1px solid var(--color-autocomplete-border);
+    border-top: none;
     max-height: 500px;
     overflow-y: auto;
-    z-index: 100;
+    z-index: 100000000;
+
+    li:first-child {
+      margin-top: 0.475rem;
+      border-top: 1px solid var(--color-autocomplete-border);
+    }
 
     &:empty {
       display: none;
@@ -62,7 +71,11 @@
 
 @media screen and (max-width: @phone) {
   .autocomplete {
+    position: absolute;
+    width: calc(100% - 0.5rem);
+    top: 2.9rem;
     bottom: 0;
+    margin-top: 0;
   }
 
   .autocomplete > ul > li {
--- a/searx/static/themes/simple/src/less/autocomplete.less
+++ b/searx/static/themes/simple/src/less/autocomplete.less
@@ -7,8 +7,7 @@
   max-height: 0;
   overflow-y: hidden;
   text-align: left;
-  border-bottom-left-radius: 10px;
-  border-bottom-right-radius: 10px;
+  .rounded-bottom-corners;
 
   &:active,
   &:focus,
--- a/searx/static/themes/simple/src/js/main/search.js
+++ b/searx/static/themes/simple/src/js/main/search.js
@@ -67,13 +67,10 @@
           },
           MinChars: 4,
           Delay: 300,
+          _Position:function() {
+            this.DOMResults.setAttribute("class", "autocomplete");
+          },
         }, "#" + qinput_id);
-
-        // hack, see : https://github.com/autocompletejs/autocomplete.js/issues/37
-        w.addEventListener('resize', function() {
-          var event = new CustomEvent("position");
-          qinput.dispatchEvent(event);
-        });
       }
 
       qinput.addEventListener('focus', placeCursorAtEndOnce, false);
--- a/searx/static/themes/simple/src/less/mixins.less
+++ b/searx/static/themes/simple/src/less/mixins.less
@@ -18,6 +18,11 @@
   border-radius: @radius;
 }
 
+.rounded-bottom-corners (@radius: 10px) {
+  border-bottom-left-radius: @radius;
+  border-bottom-right-radius: @radius;
+}
+
 .rounded-corners-tiny (@radius: 5px) {
   border-radius: @radius;
 }


searx/static/themes/simple/src/less/autocomplete.less
```diff
@@ -2,10 +2,13 @@
 
 .autocomplete {
   position: absolute;
+  width: 44.7rem;
```
dalf, line 5: It should be 44,6rem = 40rem (#q) + 1.8rem (#clear_search) + 2.8rem (#send_search)
In Firefox, it looks better with 44.7rem, but not with Chrome ==> it should be implemented in a different way.

searx/static/themes/simple/src/less/autocomplete.less
```diff
@@ -50,9 +53,15 @@
...
     border: 1px solid var(--color-autocomplete-border);
+    border-top: none;
     max-height: 500px;
     overflow-y: auto;
-    z-index: 100;
+    z-index: 100000000;
+
+    li:first-child {
+      margin-top: 0.475rem;
+      border-top: 1px solid var(--color-autocomplete-border);
```
dalf, line 62: trick to hide the bottom round corners of #q , but there is a glitch with #send_search:hover

searx/static/themes/simple/src/js/main/searx_search.js
```diff
@@ -67,13 +67,10 @@
           },
           MinChars: 4,
           Delay: 300,
+          _Position:function() {
+            this.DOMResults.setAttribute("class", "autocomplete");
```
dalf: override https://github.com/autocompletejs/autocomplete.js/blob/87069524f3b95e68f1b54d8976868e0eac1b2c83/src/autocomplete.ts#L381-L384 

```typescript
        _Position:function(): void {
            this.DOMResults.setAttribute("class", "autocomplete");
            this.DOMResults.setAttribute("style", "top:" + (this.Input.offsetTop + this.Input.offsetHeight) + "px;left:" + this.Input.offsetLeft + "px;width:" + this.Input.clientWidth + "px;");
        },
```

searx/static/themes/simple/src/less/autocomplete.less
```diff
@@ -2,10 +2,13 @@
...
 .autocomplete {
   position: absolute;
+  width: 44.7rem;
+  margin-top: 2.4rem;
   max-height: 0;
   overflow-y: hidden;
   text-align: left;
-  .rounded-corners;
+  border-bottom-left-radius: 10px;
+  border-bottom-right-radius: 10px;
```
mrpaulblack: I think it would be cleaner to add this as a mixin: ```.rounded-bottom-corners;``` and do this in `mixins.less`:
```less
.rounded-bottom-corners (@radius: 10px) {
  -webkit-border-bottom-left-radius: @radius;
  -webkit-border-bottom-right-radius: @radius;
  -moz-border-bottom-left-radius: @radius;
  -moz-border-bottom-right-radius: @radius;
  border-bottom-left-radius: @radius;
  border-bottom-right-radius: @radius;
}
```
(I am not sure if the border-bottom property actually exists for webkit and moz tho)

return42:
> Simple theme: change the autocompletion css and js not to use the style attribute.

FYI I rebased your PR and placed a settings.yml on top, to test autocompletion with CSP Header

grafik

searx/static/themes/simple/src/less/autocomplete.less
return42: FYI `-webkit-border...-radius` (28a597e43e) and ` -moz-border...-radius` (8935b1d) are no longer needed.

> I think it would be cleaner to add this as a mixin: ```.rounded-bottom-corners;``` and do this in `mixins.less`

Thanks for the hint, I will implement this today .. I still working on this PR .. coming soon ... see 2981102cb8

searx/static/themes/simple/src/less/autocomplete.less
return42: Anther issue: the 40rem ...

https://github.com/searxng/searxng/blob/33e2599b8217e263c744fc14f2c89118ef363866/searx/static/themes/simple/src/less/search.less#L133

on phone ..

https://github.com/searxng/searxng/blob/33e2599b8217e263c744fc14f2c89118ef363866/searx/static/themes/simple/src/less/search.less#L219

is changed to 100% ..

https://github.com/searxng/searxng/blob/33e2599b8217e263c744fc14f2c89118ef363866/searx/static/themes/simple/src/less/search.less#L242-L243

mrpaulblack: I think on phones the google approach is best... Let the autocomple take up the entire screen and move the search input form to the top... This way it does not matter where the search form is normally on the page and we can show the maximum amount of autocomplete suggestions to the user.
image

dalf:
* autocomplete-js seems abandoned and there are some unresolved bugs.
* it allows some customizations at different levels in less than 700 lines.
* the customization that are required by searxng are going to copy / paste a lot of code from autocomplete-js with changes here and there (for example the CSS would be easier if the autocomplete is before `#q`, so we can write `.autocomplete.open + #q { change border }`. This is not doable for now without a huge copy paste. Without CSS, it requires to JS to change the CSS classes of #q, same pitfall, it requires to copy / paste from autocomplete-js).
* there are some light redundancy between autocomplete-js and the simple theme: 
   * send HTTP request: searx/static/themes/simple/src/js/main/00_toolkit.js and https://github.com/autocompletejs/autocomplete.js/blob/87069524f3b95e68f1b54d8976868e0eac1b2c83/src/autocomplete.ts#L652-L667
   * deal with the keyboard: searx/static/themes/simple/src/js/main/keyboard.js and https://github.com/autocompletejs/autocomplete.js/blob/87069524f3b95e68f1b54d8976868e0eac1b2c83/src/autocomplete.ts#L130-L236

The solution I see is to integrate our own autocomplete code into the simple theme:
* allow [searxng.http](https://github.com/searxng/searxng/blob/e4a2d354aa4c0c97a096de47a8bcf6bb22e4d153/searx/static/themes/simple/src/js/main/00_toolkit.js#L62) to send HTTP headers
* be able to switch between between different keyboard handlers (#q focused or not, perhaps it can be useful for the image details too for example, or with the map)
* then integrate autocomplete-js: I think we can keep the javascript version (= no typescript).

This would allow:
* to have a full screen autocomplete on mobile without hack.
* on mobile, to have the arrow on the right to replace the query instead of sending it:
  image
* other customizations

return42:
> The solution I see is to integrate our own autocomplete code into the simple theme:

OK, that's fine .. but first lets fix the issue (it is a show stopper) / I send 

- #582

dalf:
I close this PR:
* #352 has been closed by #582
* various bug should be fixed with a custom autocomplete-js implementation : https://github.com/searxng/searxng/pull/426#issuecomment-986203210

PR #436, opened 2021-10-26
[mod] update_languages: min_engines_per_country=7

### LINKED ISSUE #432: Missing language: pt-BR

1.0.0-950-dfdf71bf
(docker)

Docker

**What happened?**
Language `pt-BR` is missing.
https://github.com/searxng/searxng/blob/master/searx/languages.py

It was present in a previous version.

Google supports this language (and some other engines) but it is not showing up on the list.

dalf:
## What does this PR do?

a (language,country) tuple is included if 7 engines have it, was 10 before.

In addition to `pt-BR`, the change includes also some other languages

## Related issues

close #432


return42 (approved):
Thanks!

Just a nit: Does it make sense to split this one commit into two? .. first with the patch in `update_languages.py`, second which calls `make data.language`?

dalf:
> Just a nit: Does it make sense to split this one commit into two? .. first with the patch in update_languages.py, second which calls make data.language?

If we had some tool to rollback similar to `make static.build.commit` may be, it is not the case ; so I can't see a scenario where we need two commits.

return42:
> so I can't see a scenario where we need two commits.

just for readability of the patches I thought ... but as I said; it is a _nit_ / lets merge.

PR #440, opened 2021-10-26
[simple theme] cleanup result article url

mrpaulblack:
## What does this PR do?

Make URL in articles clickable and streamline the heading on desktop, tablet and phone layout. Add a 8px top and bottom margin to the h3 heading of the article.

Make the result template cleaner and the theme more coherent on all different screen sizes.

```make run```

## Author's checklist

Desktop:
image

Mobile:
image


return42 (approved):
Nice UX .. thanks! .. 

FYI: I needed to rebase the first commit and rebuild /static path. BTW I fixed the commit message slightly (added an empty line between first line and the description block)

dalf:
This is personal taste: on some results, I find it is difficult to know where the results are.
( Also related to the engine list and the link to the cached version, out of topic: the "cached" link could be a triple vertical dot menu on the right side of the URL. When clicked it can show the "cached" and "proxified" links --> that's another issue)

Example (for the last results, I have to "process" the page not to link _Use speedtest..._ and _https://fast.com_. Quote arround process, because when I read it, it is clear, but at first sight it is not as clear as before):
image

With 6px instead of 8px, I don't have this feeling:
image

return42:
> for the last results, I have to "process" the page not to link Use speedtest... and https://fast.com. Quote arround process, because when I read it, it is clear, but at first sight it is not as clear as before

Sorry my bad english .. I do not understand what you mean ..

> With 6px instead of 8px, I don't have this feeling:

And I do not see the difference of 6px vs. 8px / but, may I need some glasses :)

mrpaulblack:
Yeah I agree the results without a description look kinda ugly tbh. So a fix could be to show like a default description. Something like: `This page did not provide a description.` just to make search results more coherent. (This would probably also help a let in the It category where like every 2nd result does not have a description)

But yeah looking at the screenshots 6px does indeed look better than 8px. It makes it more clear that the `https://fast.com` url is part of `Internet Speed Test | Fast.com` and not of the above `Speedtest by Ookla...` article...

PR #441, opened 2021-10-26
[RFC] experimental: favicon in url of article in result

mrpaulblack:
## What does this PR do?

Add favicons to URLs similar to what duckduckgo does. This PR is basically: https://github.com/searx/searx/pull/2951.  Basically this assumes the favicons are at URL/favicon.ico and they go through the image proxy, if enabled. This adds a lot of requests to every search request, since the images are not getting cached?. If the image proxy is disabled, the user connects to all of the servers in a search request, which is def. not good for privacy (tho its the same with the image category so idk).

I would like some feedback if the added polish of this PR outweighs the drawbacks of a lot more requests to the search instance on every search...

Make the simple theme more polished / Feature parity with other search providers.

(This PR builds on top of  #440)

```make themes.simple```

```make run```

## Author's checklist

Desktop:
image

(Tablet is pretty much the same as desktop)

Phone:
image

### CODE REVIEW

**Comment by dalf** [comment]
I'm meh about this PR because it will ping a lot of domains, even with the image proxy.

> the images are not getting cached

yes, the image are not cached for privacy reason (see https://github.com/searxng/searxng/commit/43fcaa642a63d75096b33d44ce7f7c0de1bce614 , only these headers are sent to the browser `Content-Type`, `Content-Encoding`, `Content-Length`, `Length`.

---

It is a different project, but looking at:
* https://opendata.stackexchange.com/questions/14007/list-of-top-10k-websites-and-their-favicons
* https://smalldata.dev/posts/favicon-mosaic/
* https://gist.github.com/chilts/7229605

It is possible to make a side local server to host the favicons. 1 million websites x 1 Ko (avg size of a favicon) = 1Go. It is a lot of through the network, but locally it is not too much.

File: searx/templates/simple/macros.html [hunk 1]
```diff
@@ -12,6 +12,10 @@
     <img width="14" height="14" class="favicon" src="{{ url_for('static', filename='themes/simple/img/icons/' + favicon + '.png') }}" alt="{{ favicon }}" />
 {%- endmacro %}
 
+{% macro result_favicon_url(domain) -%}
+    https://www.google.com/s2/favicons?domain={{domain}}
```
return42, line 16: This request will lead in tracking the SearXNG users by google.

I think we need such a kind of service in SearXNG itself, a database that maps from domain to favicon.  To cite @dalf: 

[It is possible to make a side local server to host the favicons. 1 million websites x 1 Ko (avg size of a favicon) = 1Go. It is a lot of through the network, but locally it is not too much.](https://github.com/searxng/searxng/pull/441#issuecomment-953156937)

dalf:
I close this PR: it is a great additional from UX point of view, but not for the privacy.

return42:
Even this PR is closed I want to add one more thought: Whatever favicon we might have one day, we should consider that some URLs are modified by [`hostname_replace`](https://github.com/searxng/searxng/discussions/911).

BoKKeR:
Could an external favicon solution be considered? Favicons could be scraped and distributed separately. Instead of each instance having to fetch the favicons alone. We could fetch the top domains, update them periodically and save them for local distribution to be used. The favicons dont change often and they are rather small at 32x32pixels. Even saving 10m favicons wouldnt consume much memory or space.

unixfox:
How is that worth it when Google can regurgitate links for websites all around the world? People aren't going on just a subset of 10000 websites. The web is way bigger than that.

PR #449, opened 2021-10-28
Simple theme: minor adjustements

dalf:
## What does this PR do?

* index page: margin top is 24% of the view port.
  avoid to scroll a small screen,
  center the content in the middle of the screen

* link to preferences at the same height same the input fields

* increase the category tab heighs

* increase the margin bottom of the query field

* reduce the margin to 0.4rem ( = 6px) around the h3 elements in the results.

Better UI (?)

* index page: resize vertically the windows, check 

* link to the preferences:
  * index page, check the position of the preferences icon
  * search something: the icon should be at the same place (still under the mouse cursor)
  * resize the window vertically: check the icon is at the same height same the input fields (#q, etc...)

* category tab: matter of taste...
  * search for images
  * display the detail of an image, check the detail doesn't over the search fields:
image

* check everything again with a phone and a tablet window size.


searx/static/themes/simple/src/less/search.less
```diff
@@ -30,6 +30,7 @@
 
 #search_wrapper {
   padding: 0;
+  margin-bottom: 0.4rem;
```
mrpaulblack: I think it would be better to increase the `colomn-gap` in `#search` to `0.4rem` which has basically the same effect...

searx/static/themes/simple/src/less/search.less
```diff
@@ -244,7 +244,7 @@
 
   label {
     cursor: pointer;
-    padding: 4px 10px;
+    padding: 0.3rem 0.75rem 0.5rem 0.75rem;
```
mrpaulblack, line 246: Should also be changes for `@phone` media query? (right now has a `padding: 10px;`)

searx/static/themes/simple/src/less/style.less
```diff
@@ -157,7 +157,7 @@ article.result-images[data-vim-selected]::before {
   h3 {
     font-size: 1.1em;
     word-wrap: break-word;
-    margin: 8px 0 8px 0 !important;
+    margin: 0.4rem 0 0.4rem 0 !important;
     padding: 0;
     margin-bottom: 0;
```
mrpaulblack: Remove the `margin-bottom` and the `!important` on the margin. This is a mistake that I overlooked in my PR tbh; See: https://github.com/MrPaulBlack/searxng/commit/92913ddac7402163f27c3f4fbc49c2fd860b8c44

mrpaulblack (commented):
Suggestions:

searx/static/themes/simple/src/less/search.less
dalf: The column-gap is already set to 1rem : https://github.com/searxng/searxng/blob/8985d3e6ef3c1371b21bee69074f0e38acae9967/searx/static/themes/simple/src/less/search.less#L11

its value increases only the gap between the logo and ( #q , #categories_container )

searx/static/themes/simple/src/less/search.less
dalf: Yes 

0.3rem 0.75rem 0.5rem 0.75rem --> 4.8px 12px 8px 12px

image

Some space between the categories and the languages / time selects can be visually better, but I'm not sure how to do that.

searx/static/themes/simple/src/less/search.less
mrpaulblack: oh mb I meant `row-gap` in `#search`

dalf: face palm...

Done in commit bd81999fb699d9e61a6fa4168590f123cdb51ede

searx/static/themes/simple/src/less/style.less
dalf: Done in commit bd81999

searx/static/themes/simple/src/less/search.less
mrpaulblack: You could change the `row-gap` from `0` to `1em` in `#search` for `@phone` layout in Line 175 of `search.less`... (To add space between the categories and laguage / time selector)

dalf: row-gap won't work because the `select` elements and `#q` are in the same grid area ( `search` ).
With some margins it works, see e2d312f6ae69d8bf7a36d4809b3f8544a92bb6b2

mrpaulblack: I tested the current version and looks good to me. (sry; I am going to read the code next time a bit more in depth before opening like a million review comments...)

PR #451, opened 2021-10-28
Simple theme: SearXNG logo

dalf:
## What does this PR do?

image

image

image

Font: [Libre Franklin](https://www.wikidata.org/wiki/Q102310413)

⚠️ The favicon is unchanged.

Start with a (temporary) logo for the simple theme.
It can be changed later, but at least it shows the right name.


dalf:
ping @MrPaulBlack

mrpaulblack:
Yeah I think this looks a lot better than the old SearX logo for the simple theme. The only thing I would change is change the `min-height` in `.index .title` in `index.less` Line 12 to `4rem`, so the image has some space to the edges of the screen and (heavily opinionated) I like it more being a little smaller... Apart from that LGTM :+1: 

With min-height 4rem:
image

image

PR #452, opened 2021-10-28
healthcheck endpoint from https://github.com/searx/searx/pull/2992

mrpaulblack:
## What does this PR do?

This PR adds a `/healthz` endpoint which does nothing more than returning plaintext `OK` response. (We could build on this by implementing healthcheck in the `Dockerfile` of this repo or in the `docker-compose.yml` of searxng/searxng-docker)

It is useful for service discovery tools (consul/kubernetes etc.), to quickly say, that service is alive and running.

```make run```

```curl https://127.0.0.1:8080/healtz -i```

## Related issues

From https://github.com/searx/searx/pull/2992

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -618,6 +618,11 @@ def index():
     )
 
 
+@app.route('/healthz', methods=['GET'])
+def health():
+    return Response('OK', mimetype='text/plain')
+
+
 @app.route('/search', methods=['GET', 'POST'])
 def search():
     """Search query in q and return results.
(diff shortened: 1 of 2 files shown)


return42:
I have seen the PR from searx origin at that time but hesitated to merge, since I asked myself: Is the response from `/config` not enough?

mrpaulblack:
Well yes but the response from `/config` is quite large tbh. So using the an extra `/healthz` endpoint is a lot faster and more efficient if a maintainer uses the endpoint in k8s or docker and ping it like every minute....

return42 (approved):
FYI: to have the `Author*:` fields from the origin in the commit:

grafik

I cherry piked from the origin commit https://github.com/ajgon/searx/commit/54a2cd040e8411afeabf4feefa29f51d77b19469 and force pushed on your PR branch.

I hope this is OK for you?

mrpaulblack:
Yes, thank you. I tried cherry picking, but it did not work for me...

PR #460, opened 2021-10-30
Simple theme: various fixes

dalf:
## What does this PR do?

Description of the different commits.
Use the confused reaction to ask for multiple PR instead this one including various topics.

### [fix] simple theme: cookie table doesn't overflow the expected width 

In the master branch, the cookie table overflows outside the `#main_preferences` element.
This commit fixes this issue:

image

### [fix] simple theme: /preferences: dark theme version for the response time bar charts

Dark theme version: the small bar charts showing the response times are displayed the `--color-base-font` color.
image

* [fade](https://lesscss.org/functions/#color-operations-fade) from the less doesn't work on CSS variable.
* the way to defined the opacity in CSS with color is `rgba(r, g, b, a)`
* the workaround is to defined an additional CSS variable `--color-base-font-rgb` that has the exact same value than `--color-base-font` but written with R, G, B values.

### [fix] simple theme: /preferences: dark theme version for select widgets 

Implemented with two different SVG because background image can't use the CSS variables.
The SVG files are stored in `searx/static/themes/simple/src/svg`, then inline into the CSS using [data-uri](https://lesscss.org/functions/#misc-functions-data-uri)

image

image

###  [mod] simple theme: use :root instead of html to define CSS variables 

It doesn't do anything, except it will easier to use inline SVG later.

### [mod] simple theme: infobox: center the image
image

* `make run`


searx/static/themes/simple/src/less/toolkit.less
```diff
@@ -339,6 +339,12 @@ select {
       border-bottom: 1px solid var(--color-search-border);
     }
   }
+
+  @media (prefers-color-scheme: dark) {
+    select {
+      background-image: data-uri('image/svg+xml;charset=UTF-8', '../svg/select-dark.svg');
```
mrpaulblack: This is going to break, once the dark mode toggle gets implemented (or et least it would be a copy of the changes in definitions.less)... So I think it would be better to set the path as a var in `definitions.less`.. What do you think? 
[EDIT] apart from that LGTM :+1: 

(Also after reading your commit I am not sure if it is even possible to use cs vars here...?)

dalf: `data-uri` is a less function that read and encode the file, so the CSS doesn't reference external files:
https://github.com/searxng/searxng/blob/4dd35f6d71d231334263365610775323166e9c88/searx/static/themes/simple/css/searxng.css#L1705-L1709

So it is not possible to use a CSS variable, but a LESS variable is possible.

mrpaulblack: Yeah that is a great idea, thank you! I just think its better to keep the dark/light and color specific stuff in `definitions.less`. So adding a less var with the path to the select depending on dark/light would be the better option I believe... (If that works tho; cannot test it right now sry)

[EDIT] so the idea is less generates the prefer dark theme media query automatically, since the are two different values in the svg path less variable, but idk if that actually works tho

dalf: > ] so the idea is less generates the prefer dark theme media query automatically, since the are two different values in the svg path less variable, but idk if that actually works tho

I'm not sure how to do that.

Looking at https://lesscss.org/features/#variables-feature-variable-interpolation it is possible to have

```less
@{dark-selector} {
   ....
}
```

but `dark-selector: @media (prefers-color-scheme: dark), html.dark-theme` doesn't work.

The only way I see is two less variables and copy / paste the definitions.
This is what postcss does: https://equinusocio.dev/blog/making-dark-theme-switcher-with-postcss/ (see the "Where is the magic?" section).

A workaround to avoid to rewrite everything ( = to keep less), we can use either a mixin or even a dedicated file (`dark.less`).

---

[EDIT] Continuation of this PR in https://github.com/dalf/searxng/tree/simple-dark-mode-fixes

dalf: I'm going to merge this PR, we can adjust the dark mode definition later.

PR #462, opened 2021-10-31
[RFC] Simple theme: fix & improve the dark mode

dalf:
## What does this PR do?

Update the dark mode of the simple theme

### Back to top button

Was blue / violet before:

image

### Slightly lighter border 

image

### /preferences

* darker mouse over selection
* checkbox: darker background (to enable / disable engine for example)
* lighter line below the tabs
* button: dark font color instead of white color (fix a contrast issue)
* select: lighter underline. Fix an issue with Chrom* browser: the open selection was not readable (white font with a white background).

image

image

image

image

`make run`

## Author's checklist

* May the lines are too light compare to the background ?
* There is too much contrast in the cookie tab


return42:
Dump question, how can I switch to the dark mode? .. I activated a dark-theme from the FFox themes but this does not change SearXNG simple theme into dark mode. Also tried [
image
](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Examine_and_edit_CSS#view_media_rules_for_prefers-color-scheme) in FFox developer suite, but it does not change the rendering.

mrpaulblack:
So yeah for me setting 
image
 works just fine in Firefox Dev Tools for me.. You could modify `definitions.less` and just add the dark mode the the `:root` as a workaroud, but idk why it would work on my pc and not on yours (FYI: I am using latest firefox flatpak on linux...)

Oh also: I think the css selector for the dark mode does not care about the firefox theme; It honors the theme of the desktop environment: so changing to dark mode in windows 10 settings or in kde preferences for example should trigger the dark mode for searxng

mrpaulblack:
@dalf Yeah I am not a big fan of making the border-bottom of the tr in the tables and the hover darker tbh. I would rather make them slightly lighter as they where before or something like: `#2d2d2d`. (I think in a dark theme only darker shadows look good and everything else should be lighter for highlights tbh); everything else LGTM :+1:

dalf:
The darker border-bottom was clearly an issue, the darker hover was more a try / a test.
A lighter hover is what I see everywhere else.

I've pushed force the update commits using `#2d2d2d`

return42:
> but idk why it would work on my pc and not on yours

Argh .. sorry, my fail .. I have a very restricted [/etc/firefox/syspref.js](https://github.com/arkenfox/user.js) .. if I disable the restrictions, dark/light mode works in my FFox as expected  / thanks!

return42:
EDIT: this PR has been comited while I was testing it .. this are my tests:


> * darker mouse over selection

this is your screenshot ..

image

this is what I get with FFox & Chrome  (I assume this is realted to https://github.com/searxng/searxng/pull/462#issuecomment-956527198 .. right?

grafik

> * checkbox: darker background (to enable / disable engine for example)

Works / and is much better than before !

> * lighter line below the tabs

Works

> * button: dark font color instead of white color (fix a contrast issue)

Works / and is much better than before !

PR #430, opened 2021-10-22
[simple theme] add logo to search form on result page

mrpaulblack:
## What does this PR do?

Add a logo to the left side of the search form with a link back to the index page. Reworked parts of the search as grid layout to achieve this and added the logo to the jinja2 template. (Also removed various paddings and margins from the elements in the grid layout since the grid layout is taking care of spacing.)

On mobile the categories have more padding to make them easier clickable. Also the issue in dalf comment is happening at screen widths smaller than 360px and only when touch input is present. This was happening before this PR at a `2rem` smaller screen width, but I think its still fine.

## Why is this change important?

Make the simple theme more polished and feature parity with other search engines.

## How to test this PR locally?

```make themes.simple```

```make run```

## Author's checklist

Desktop:
image

Tablet:
image

Phone:
image

## Related issues

Closes #427

PR #378, opened 2021-10-03
[logo] SearXNG logo & trademark (WIP)

return42:
## What does this PR do?

SearXNG logo & trademark (WIP)

## Why is this change important?

SearXNG need its own logo.

## How to test this PR locally?

A first draft, not yet ready to test!

Have a look at:  https://github.com/return42/searxng/tree/searxng-logo/src/brand

I am not very creative  .. comments are welcome ... personally I prefer ...

grafik

----

Related-to: https://github.com/searxng/searxng/issues/370

PR #473, opened 2021-11-03
[simple theme] SearXNG wordmark & favicon

return42:
## What does this PR do?

[simple theme] SearXNG wordmark & favicon

["the favicon is not really suited for dark mode (basically we need different icons for dark and for light mode eventually)"](https://github.com/searxng/searxng/pull/430#issuecomment-951967239)

## Author's checklist

Favicon (see tab) and wordmark (left to the search form)  has been changed ..

grafik

before this PR:

image

## Related issues

https://github.com/searxng/searxng/pull/378


searx/templates/simple/search.html
```diff
@@ -1,6 +1,6 @@
 <form id="search" method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
   <a id="search_logo" href="{{ url_for('index') }}">
-    <img class="search_logo_img" src="{{ url_for('static', filename='img/favicon.png') }}">
+    <img class="search_logo_img" src="{{ url_for('static', filename='img/searxng-wordmark.svg') }}">
```
dalf: To avoid an additional HTTP request, it can be replaced by an optimized inline version of the SVG file (429 bytes) :
```svg
<svg viewBox="0 0 92 92" version="1.1"><g transform="translate(-40.92 -17.42)"><circle cx="75.92" cy="53.9" r="30" fill="none" stroke="#3050ff" stroke-width="10"/><path d="M67.51 37.92a18 18 0 0 1 21.06 3.3 18 18 0 0 1 3.13 21.09" fill="none" stroke="#3050ff" stroke-width="5"/><rect width="18.85" height="39.96" x="3.71" y="122.09" ry="0" transform="rotate(-46.23)" fill="#3050ff"/></g></svg>
```

This version can be an external file that is included using the Jinja2 syntax.

The version above was made using https://jakearchibald.github.io/svgomg/ with these options:
* set precision to 2
* check "prefer viewbox to width/height" (at the bottom of the options)
* Manually remove `xmlns="http://www.w3.org/2000/svg"`.

For reference:
* https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_In_HTML_Introduction
* how the GitHub logo in the top left corner in embedded.

---

[EDIT] Also, the PNG file size can be optimized and reduced to 1.2KB.

return42: > This version can be an external file that is included using the Jinja2 syntax.

I assume you are talking about something like this ... ?
```python

from flask import Markup
---
    return render(
        'results.html',
        wordmark_svg = Markup(open('.....img/searxng-wordmark.svg').read)
```
> To avoid an additional HTTP request

Yes, we can avoid one request, but on the opposite (if I'm not wrong) we lost flexibility with the `static/themes/simple` folder 

https://github.com/searxng/searxng/blob/580a55fe866b82e93dec6bc4e07fa8a0f2056475/searx/settings.yml#L64-L66

dalf: What about 

webapp.py:
```python
wordmark_contents = {}
for indice, theme in enumerate(themes):
    ...
    workmark_file = os.path.join(settings['ui']['static_path'], 'theme', theme, 'img/searxng-wordmark.svg')
    if os.path.isfile(workmark_file):
        with os.open(workmark_file, 'rb') as f:
            wordmark_contents[theme] = f.read()
    ...


def render(...):
  ...
  kwargs['wordmark_content'] = wordmark_contents.get(kwargs['theme'])
  ...
```

search.html:
```jinja2
{{ wordmark_content }}
```

?

[EDIT] Too complicated.

Perhaps this:
* add a new function : `content_for`, same as `url_for` but actually returns the content.
* `kwargs['content_for'] = content_for` 

--> `{{ content_for('img/searxng-wordmark.svg') }}`

searx/templates/simple/search.html
```diff
@@ -1,6 +1,8 @@
 <form id="search" method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
   <a id="search_logo" href="{{ url_for('index') }}">
-    <img class="search_logo_img" src="{{ url_for('static', filename='img/searxng-wordmark.svg') }}">
+    <div class="search_logo_img">
+      {{ insert_xml('img/searxng-wordmark.svg') }}
+    </div>
```
return42: this leads into  ..

grafik

 and I do not know how I can set a fixed size for the `<div>` container or how the can set a fixed box-size for the  _inline SVG_ ..

I'm not an CSS expert and I haven't understand the CSS box model really.

searx/templates/simple/search.html
return42: @dalf I have tried what you suggested, see dd4da6f / but now I have a problem to size the inline SVG, see 
https://github.com/searxng/searxng/pull/473/commits/dd4da6f465349488ef101d76773d630dffea0d24#r743818450

dalf:
I've pushed another simpler solution (it requires to build the themes): 
* `{% include '__common__/logo.svg' without context %}`  (`without context` so the result is cached).
* `__common__/logo.svg` is the minified version (perhaps it should be `simple/logo.svg` ? )

So, there is no need for a special function that read the .svg at each request.

mrpaulblack:
I suggest the following (adding the `margin: 0.5rem 0 auto 0;` to the svg selector):

```less
#search_logo {
  grid-area: logo;
  display: flex;
  align-items: center;
  justify-content: center;

  svg {
    flex: 1;
    width: 30px;
    height: 30px;
    margin: 0.5rem 0 auto 0;
  }
}
```

(it makes sure that the logo is always displayed next to the search input field)
On tablet it looks like that with the fix:
image

return42:
@dalf @MrPaulBlack: thanks a lot for your helping comments, I placed what you suggested in commit:

- fc00cb1 [simple theme] replace old searx logo by searxng-wordmark.min.svg

Adittionally ``make static.build.commit`` now builds additional files:

- the minified SVG is generated by [html-minifier](https://github.com/kangax/html-minifier) in commit dc02b17
- simple/img/favicon.png is generated by `convert_if_newer()` in commit 4a01683

In 21e690c3 I optimized the origin `searxng-wordmark.svg` and a792d5fc builds minified `searxng-wordmark.min.svg` and converts `favicon.png`  [(see svg + png files from build)](https://github.com/searxng/searxng/pull/473/commits/a792d5fc2323c6ba9e5b077e18eb1eb2518a632d?authenticity_token=9h%2BlDdrXPX2Bn8zwPl1G15iQS14GhNXbvANclBesLLuounheiTzFkW2Xc7kfa5jnqKYBtlZgvTOhEmf4fSQOJg%3D%3D&file-filters%5B%5D=.png&file-filters%5B%5D=.svg).

mrpaulblack (commented):
:-1: The favicon.png does not have a alpha channel...
image

Is this intended? (Or to rephrase: is it possible to generate favicon.png with an alpha channel from the svg?)

Everything else LGTM:
:+1:  the `make themes.all` and `make themes.simple` exec the htmlmin step for simple properly
:+1: watermark displays correctly on desktop, tablet and mobile
:+1: the jinja2 template inlines the svg correctly into the html

searx/static/themes/simple/package.json
```diff
@@ -8,6 +8,8 @@
     "grunt-contrib-jshint": "~3.1.1",
     "grunt-contrib-less": "~3.0.0",
     "grunt-contrib-uglify": "~5.0.1",
+    "grunt-xmlmin": "~0.1.8",
+    "grunt-contrib-htmlmin": "~3.1.0",
```
dalf, line 12: [grunt-image](https://www.npmjs.com/package/grunt-image) uses [svgo](https://www.npmjs.com/package/svgo). The min.svg file size can be reduced by ~ half (it removes the sodipodi elements for example).

manage
```diff
@@ -692,11 +692,44 @@ themes.oscar() {
 }
 
 themes.simple() {
-    build_msg GRUNT "theme: simple"
-    npm --prefix searx/static/themes/simple run build
+    local src="searx/static/themes/simple/src"
+    local static="searx/static/themes/simple"
+    (   set -e
+        convert_if_newer "$src/svg/searxng-wordmark.svg" "$static/img/favicon.png" -background none -resize 64x64
```
return42: > :-1: The favicon.png does not have a alpha channel... image Is this intended? (Or to rephrase: is it possible to generate favicon.png with an alpha channel from the svg?)

Oops .. it is fixed right here by adding [`-background none`](https://www.imagemagick.org/script/command-line-options.php#background) to the convert command.

dalf: I wish to keep all the build steps inside gruntfiles.js

In the simple theme directory, `npm run watch` allows to rebuild the theme a files change. If we mix bash and node it starts to be complicated to do that.

There are some grunt modules to convert to svg to png, but I haven't tried them.

Otherwise it is possible to call a command line from grunt: https://www.npmjs.com/package/grunt-exec

But, the logo won't change as often the other files, so IMO, it is not an issue if there is automatic build of the favicon.

return42: > I wish to keep all the build steps inside gruntfiles.js

Hm, .. I understand to want to have all theme build-tasks at one place (grunt), but the drawback I see is; with [grunt-exec](https://www.npmjs.com/package/grunt-exec) we can't express a conditional _convert if origin is newer_. IMO calling shell tasks from grunt increase complexity unneeded.

searx/static/themes/simple/package.json
return42: OK I understand .. can we optimize this later?  First I think we should clarify which version of svgo we should use. ATM we have an issue reported when `make node.clean node.env` is build:

    npm WARN deprecated svgo@0.6.6: This SVGO version is no longer supported. Upgrade to v2.x.x.

This issue is caused by grunt-webfont and might be also related to https://github.com/searxng/searxng/pull/488#issuecomment-963606592 / Hint: https://github.com/sapegin/grunt-webfont is not maintained since years, see archived repository at https://github.com/sapegin/grunt-webfont and we have other issues with, see  https://github.com/searxng/searxng/pull/248#issuecomment-894672328

return42:
@dalf @MrPaulBlack I rebased this PR and gave some feedback to your requests. What do you think, is this PR ready to merge or is there anything left to do?

mrpaulblack:
LGTM :+1:

dalf:
The favicon.png does not have an alpha channel:
https://github.com/searxng/searxng/blob/7ea3faf033b94de82eefd1a92cebfe9241735945/searx/static/themes/simple/img/favicon.png

---
With  convert "ImageMagick 6.9.10-23 Q16 x86_64 20190101 https://imagemagick.org" (Ubuntu 20.04), the rendered image is broken.

manage
```diff
@@ -692,11 +692,45 @@ themes.oscar() {
 }
 
 themes.simple() {
-    build_msg GRUNT "theme: simple"
-    npm --prefix searx/static/themes/simple run build
+    local src="searx/static/themes/simple/src"
+    local static="searx/static/themes/simple"
+    (   set -e
+        convert_if_newer "$src/svg/searxng-wordmark.svg" "$static/img/favicon.png" \
+                         -transparent white -resize 64x64
```
return42, line 699: > The favicon.png does not have an alpha channel:

@dalf thanks a lot for your attention! ... I'm very sorry for the mess .. it is not ImageMagick what is broken .. it is me 🤦🏻 

Its now fixed ...

grafik

mrpaulblack: LGTM :+1: 
image

Issue #463, opened 2021-10-31
Which default theme ?


dalf:
Oscar is the current default theme.

In my opinion, the simple themes look nicer, but it has some issues compare to the Oscar theme:
* it doesn't support RTL languages (doable with some LESS mixins).
* it doesn't follow the ARIA recommendations.
* it doesn't support the infinite scroll and some other features.

At some point, I would like to the simple theme the default one.

mrpaulblack:
In my opinion the simple theme can be set as default when these are implemented / fixed (highly opinionated):


- [x] the dark mode  gets exposed in settings and as a toggle in the navbar (I started on that one; well or at least tried: https://github.com/MrPaulBlack/searxng/tree/dark-theme-pref-toggle)
- [x] make the preferences work on mobile (so fix for: #356 and #351 as well as https://github.com/searxng/searxng/issues/176#issuecomment-927746318)
- [x] make the detail modal display correctly on mobile (#445)
- [x] (maybe?) make the opening pages on a new tab plugin work (I was looking at: https://github.com/searxng/searxng/blob/e4025cd1644e749168d40f63cf6720e7764b6a7f/searx/templates/simple/macros.html#L16 and I have honestly no idea, why it would not work, since its a 1to1 copy from oscar)
- [x] solve CSP error with autocomplete in simple theme (the current lib adds style attributes to html; see #426 to solve #352)

(the most important thing [in my opinion] is to make the theme work as well on mobile as it does on desktop and tablet and to make it clear that it does indeed support a dark mode now; since its not clear how to switch between light and dark version rn)

mrpaulblack:
@dalf I am not really sure how to implement the dark theme toggle and setting entry, because I am not well versed with the js files of the simple theme and the webadapter.py.... So if you want to take the dark mode toggle (and auto theme as well as pref. entry), than I would like to rework preferences to work on mobile... (Going to start work on this next week)...

dalf:
@mrpaulblack no problem, I will do it :)

return42:
While we are here.. I would like to insist on PR https://github.com/searxng/searxng/pull/426 which solves issue https://github.com/searxng/searxng/issues/352 .. what is a show stopper for me :-o

[EDIT] see screen shot in https://github.com/searxng/searxng/issues/352#issuecomment-930880903 .. or test simple theme on https://darmarit.org/searx/

mrpaulblack:
[...]

PR #479, opened 2021-11-05
[fix] simple theme: fix open in a new tab preference

dalf:
## What does this PR do?

Simple theme: "Open in a new tab" was not working:
* macros in `searx/templates/simple/macros.html` try to access `results_on_new_tab` .
* since the macros were imported without context, `results_on_new_tab`  was not initialized: https://jinja.palletsprojects.com/en/3.0.x/templates/#import-context-behavior

With this PR, Jinja2 won't cache the HTML results from the macro calls. I don't think this is a new issue: most of the macro content relies on the current result, so the cache is useless.

Fix a long standing bug.

* set "Open in a new tab" in preferences.
* for each category (General, Images, Videos, ...), make a query
* check the results open in a new tab

* disable "Open in a new tab" preferences.
* check for each category, the results open in the current tab.

## Related issues

Related to https://github.com/searxng/searxng/issues/463#issuecomment-955787945

### DIFF
--- a/searx/templates/simple/result_templates/code.html
+++ b/searx/templates/simple/result_templates/code.html
@@ -1,4 +1,4 @@
-{% from 'simple/macros.html' import result_header, result_sub_header, result_sub_footer, result_footer, result_footer_rtl %}
+{% from 'simple/macros.html' import result_header, result_sub_header, result_sub_footer, result_footer, result_footer_rtl with context %}
 
 {{ result_header(result, favicons, image_proxify) -}}
 {{- result_sub_header(result) -}}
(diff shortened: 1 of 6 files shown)


mrpaulblack:
I tested every category on desktop, tablet and phone and everything LGTM, except:

* links do not work on the image detail modal at all anymore (this can be tested with reddit engine on social and the image engines on image category)

(also: The image detail modal scales correctly on phone layout with touch enabled when opening it from the social category)

(also: also: Is the code result template even used at all?)

dalf:
Thank you for the quick feedback.

> links do not work on the image detail modal at all anymore (this can be tested with reddit engine on social and the image engines on image category)

This bug exists in the master branch too (and paulgo.io). I would like to fix this issue in another PR.

> The image detail modal scales correctly on phone layout with touch enabled when opening it from the social category

That's weird, because the last changes don't seem related to this.
Anyway the labels are too small or too big in the horizontal position.

> Is the code result template even used at all?

For one engine: [searchcode_code](https://github.com/searxng/searxng/blob/master/searx/engines/searchcode_code.py)

PR #481, opened 2021-11-05
Simple themes: fixes in keyboard.js

dalf:
## What does this PR do?

see commit messages:
* disable the hotkeys when they are not enabled in the preferences
* hotkeys help dialog:
  * dark mode: #555 border (same as infoboxes and other borders)
  * remove a call to console.log
  * center the dialog without using the style attribute.
* the URL in the image detail works as expected (not related to the hotkeys, but the change happens in keyboard.js ...)

Various bug fixes.


mrpaulblack:
everything LGTM:

* disabling vim hotkeys in pref. disables them
* there is not console log and style attribute from the dialog-modal
* opening the url (with and without the open on new tab plugin) in the detail modal works now

Issue #434, opened 2021-10-25
Searx POST method


simpvn:
**1.0.0**

**How did you install SearXNG?**
I'm using public instance searx.be

**What happened?**
POST method does not work on Brave browser as search query is shown on search bar and history

**How To Reproduce**
Enable POST method and search anything

**Expected behavior**
Search query is hidden on search bar and history

**Additional context**
Startpage does not have the same problem

<img width="781" alt="Screen Shot 2021-10-25 at 12 33 35" src="https://user-images.githubusercontent.com/78357927/138640109-f31ea140-4197-4ca6-a678-96b368b2724e.png">
<img width="1178" alt="Screen Shot 2021-10-25 at 12 33 54" src="https://user-images.githubusercontent.com/78357927/138639953-a21ebc16-ffb7-410a-92ac-1038a2f537cb.png">
79b-4ab1-bfdd-a5c5d3671768.png">
<img width="669" alt="Screen Shot 2021-10-25 at 12 34 25" src="https://user-images.githubusercontent.com/78357927/138639962-c67be16e-2498-48f8-a9fe-55621a413a77.png">
<img width="947" alt="Screen Shot 2021-10-25 at 12 34 39" src="https://user-images.githubusercontent.com/78357927/138639966-f43f0cae-eb69-4ed0-aea1-41bc8fc33411.png">

unixfox:
Which method is set in the section "privacy" from the preferences page?

Did you try to search again? It should use POST after that.

simpvn:
> Which method is set in the section "privacy" from the preferences page?
> 
> Did you try to search again? It should use POST after that.

POST does not work
Like the pics above, chossing POST doesn't hide anything

dalf:
* With Brave `Version 1.31.87 Chromium: 95.0.4638.54 (Build officiel) (64 bits)`, no extension
* searx.be, if I go to the preferences, "Privacy" tab, I select "POST", save
* After that, if I enter a query on searx.be and then press enter (or click on magnifier on the right of the input field on the web page), the HTTP request is using the POST method:

![image](https://user-images.githubusercontent.com/1594191/139209018-96668911-b631-46be-bf52-dfa4469cc2f4.png)

Do you have any installed extensions ?

unixfox:
> > Which method is set in the section "privacy" from the preferences page?
> > Did you try to search again? It should use POST after that.
> 
> POST does not work Like the pics above, chossing POST doesn't hide anything

[...]

PR #485, opened 2021-11-06
[mod] new preference: query_in_title

dalf:
## What does this PR do?

* disable by default
* settings.yml: ui.query_in_title
* in /preferences: privacy tab

when enabled, the result page's title contains the user query.

previously:
* oscar theme: the query was always included
* simple theme: the query was included with the GET method

image

image

Some browser records the page history using the HTML title pages.

See https://github.com/searxng/searxng/issues/434#issuecomment-962023693

With the simple and oscar themes:
* make a query : check there is NO title
* change the preferences: check there is a title
* reset the preferences
* in settings.yml: set `ui.query_in_title: true`
* check there is a title
* in settings.yml: set `lock: - query_in_title`
* `/preferences`: check the option has disappeared.

## Related issues

Related to #434

### DIFF
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -280,6 +280,17 @@
                         {{ preferences_item_footer(image_proxy_info, image_proxy_label, rtl) }}
                         {% endif %}
 
+                        {% if 'query_in_title' not in locked_preferences %}
+                        {% set query_in_title_label = _('Query in the page\'s title') %}
+                        {% set query_in_title_info = _('When enabled, the result page\'s title contains your query. Your browser can record this title') %}
+                        {{ preferences_item_header(query_in_title_info, query_in_title_label, rtl, 'query_in_title') }}
+                            <select class="form-control {{ custom_select_class(rtl) }}" name="query_in_title" id="query_in_title">
+                                <option value="1" {% if query_in_title  %}selected="selected"{% endif %}>{{ _('Enabled') }}</option>
+                                <option value="" {% if not query_in_title %}selected="selected"{% endif %}>{{ _('Disabled')}}</option>
+                            </select>
+                        {{ preferences_item_footer(query_in_title_info, query_in_title_label, rtl) }}
+                        {% endif %}
+
                         {{ plugin_of_category('privacy' )}}
                     </div>
                 </fieldset>
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -237,6 +237,18 @@
       <div class="description">{{ _('Proxying image results through SearXNG') }}</div>
     </fieldset>
     {% endif %}
+    {% if 'query_in_title' not in locked_preferences %}
+    <fieldset>
+      <legend>{{ _('Query in the page\'s title') }}</legend>
+      <p class="value">
+        <select name='query_in_title'>
+          <option value="1" {% if query_in_title %}selected="selected"{% endif %}>{{ _('Enabled') }}</option>
+          <option value="" {% if not query_in_title %}selected="selected"{% endif %}>{{ _('Disabled') }}</option>
+        </select>
+      </p>
+      <div class="description">{{ _('When enabled, the result page\'s title contains your query. Your browser can record this title.') }}</div>
+    </fieldset>
+    {% endif %}
     {{ plugin_preferences('privacy') }}
   {{ tab_footer() }}
 
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -405,6 +405,17 @@ class Preferences:
                     'on': True,
                 }
             ),
+            'query_in_title': MapSetting(
+                settings['ui']['query_in_title'],
+                is_locked('query_in_title'),
+                map={
+                    '': settings['ui']['query_in_title'],
+                    '0': False,
+                    '1': True,
+                    'True': True,
+                    'False': False
+                }
+            ),
         }
 
         self.engines = EnginesSetting('engines', choices=engines)
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -66,6 +66,9 @@ ui:
   static_path: ""
   # Custom templates path - leave it blank if you didn't change
   templates_path: ""
+  # query_in_title: When true, the result page's titles contains the query
+  # it decreases the privacy, since the browser can records the page titles.
+  query_in_title: false
   # ui theme
   default_theme: oscar
   # Default interface locale - leave blank to detect from browser information or
@@ -91,6 +94,7 @@ ui:
 #     - language
 #     - autocomplete
 #     - method
+#     - query_in_title
 
 # searx supports result proxification using an external service:
 # https://github.com/asciimoo/morty uncomment below section if you have running
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -186,6 +186,7 @@ SCHEMA = {
         },
         'results_on_new_tab': SettingsValue(bool, False),
         'advanced_search': SettingsValue(bool, False),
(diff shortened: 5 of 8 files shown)


mrpaulblack (changes requested):
I suggest changing the line in results.html in both oscar and simple to (This way there is a space between the query and the `-` like so: `time - SearXNG`; right now the title looks like this in firefox: `time- SearXNG`):

```html
{% block title %}{% if query_in_title %}{{ q|e }} - {% endif %}{% endblock %}
````

* :+1: enabeling the setting in pref. in oscar and simple works
* :+1: disabeling the setting in pref. in oscar and simple works
* :+1: reseting the settings work
* :-1: enabeling query in title does NOT work through settings.yml (`query_in_title: true` in ui section; this is the case for both themes)
* :+1: locking the setting in settings.yml works for both themes

**Comment by dalf** [comment]
> 👎 enabeling query in title does NOT work through settings.yml (query_in_title: true in ui section; this is the case for both themes)

~Fixed.~ (the GitHub workflow fails)
Also, the values is stored in `query_in_title` (underscore) in settings, preferences and Jinja context (no mismatch with `query-in-title`).

@unixfox : the default value is off. So the title will disappear from page's title when the instances are upgraded unless the user explicitly set the value to on.

Possible better implementation: 
* set the default value to "on" for FF browser, "off" otherwise.
* in `settings.yml`: `query_in_title: auto` (`false`, `true` are still possible, `auto` is additional value)
* the function `searx.webapp.pre_request`  set the value according to the user agent (something similar is already for the HTTP method).

**Comment by dalf** [comment]
`make test.robot` fails because of this line:
https://github.com/searxng/searxng/blob/49879a4a63652d638bce21d8cdf1da3d65415d3a/searx/templates/oscar/preferences.html#L289

More precisely, removing `{% if not query_in_title %}selected="selected"{% endif %}` from this line fixes this error:
```
Error occured: Traceback (most recent call last):
  File "/home/alexandre/code/searx/tests/robot/__main__.py", line 69, in main
    run_robot_tests([getattr(test_webapp, x) for x in dir(test_webapp) if x.startswith('test_')])
  File "/home/alexandre/code/searx/tests/robot/__main__.py", line 62, in run_robot_tests
    test(browser)
  File "/home/alexandre/code/searx/tests/robot/test_webapp.py", line 55, in test_preferences_engine_select
    assert browser.find_by_xpath('//input[@id="engine_general_dummy__general"]').first.checked
AssertionError
```

I can't figure out why.

Looking at the HTML page:
```
export SEARXNG_SETTINGS_PATH=$(pwd)/tests/robot/settings_robot.yml
make run
xdg-open http://localhost:11111
```
I don't see any issue.

**Comment by mrpaulblack** [comment]
Also: Setting anything it the settings on the user device through the /preferences endpoint does not work anymore. It seems the POST request when pressing the save btn does not work anymore:

image

Also the debug log:
```log
ERROR   searx.webapp                  : Invalid choice: 
  File "/home/local/code/external/searxng/searx/webapp.py", line 519, in pre_request
    preferences.parse_dict(request.form)
  File "/home/local/code/external/searxng/searx/preferences.py", line 460, in parse_dict
    self.key_value_settings[user_setting_name].parse(user_setting)
  File "/home/local/code/external/searxng/searx/preferences.py", line 206, in parse
    raise ValidationException('Invalid choice: {0}'.format(data))
searx.preferences.ValidationException: Invalid choice: 
DEBUG   searx.webapp                  : http%3A//127.0.0.1%3A8888/preferences uses locale `en`
INFO    werkzeug                      : 127.0.0.1 - - [07/Nov/2021 11:01:05] "POST /preferences HTTP/1.1" 302 -
DEBUG   searx.webapp                  : http%3A//127.0.0.1%3A8888/ uses locale `en`
```

dalf:
@MrPaulBlack the issue you describe is exactly why the tests fail : 
* the preferences are [saved](https://github.com/searxng/searxng/blob/90acb342d32b1ce890402acbc24cc9a5793fddc6/tests/robot/test_webapp.py#L46) : it fails silently.
* then code checks the preferences are [changed](https://github.com/searxng/searxng/blob/90acb342d32b1ce890402acbc24cc9a5793fddc6/tests/robot/test_webapp.py#L55)

I've added
https://github.com/searxng/searxng/blob/80fb77476fad4b229418c530f3ffda67f357a15a/searx/preferences.py#L409
(`image_proxy` preference definition has the "workaround")

So now it should work.

---

Perhaps with Firefox, the default value should be `true` and `false` in other cases.

mrpaulblack:
FYI: Firefox does actually log the page title in history

image

I think this setting should be off by default...

dalf:
> FYI: Firefox does actually log the page title in history

even wit the POST method ?

It is off by default, see searx/settings_defaults.py

mrpaulblack:
Woow you are right; It does not log page title when using POST method with query in page title enabled mb...

dalf:
By default, for a Firefox user, this setting can be enabled when the method is POST, and disabled in all other cases.
It is doable but.... that's a mess from the user point of view: why suddenly the title has disappear? 

So as you said, it should be disabled by default.

There are an inconsistent settings:
* in the /preferences, it is in the privacy section.
* in `settings.yml`, the preferences in under the `ui` section (there is no privacy section).

mrpaulblack (approved):
I think the inconsistency in settings.yml is fine; since its the only privacy setting from the theme in settings.yml and still kinda fits into that category (if there are more privacy items we should probably think about adding the category to settings.yml...). I also retested and everything works now :+1: (settings.yml change in simple and oscar and also changing in /pref. in simple and oscar).... The only thing I would change is to add a space after the query and before the `-`:

image

(But dont know how and its not that important if not tbh)

searx/templates/oscar/preferences.html
```diff
@@ -280,6 +280,17 @@ <h3>{{ _('Privacy') }}</h3>
                         {{ preferences_item_footer(image_proxy_info, image_proxy_label, rtl) }}
                         {% endif %}
 
+                        {% if 'query_in_title' not in locked_preferences %}
+                        {% set query_in_title_label = _('Query in the page\'s title') %}
+                        {% set query_in_title_info = _('When enabled, the result page\'s title contains your query. Your browser can record this title') %}
```
return42, line 285: While I was in translation I recognized that we have this sentence twice, here in oscar (missing the final dot) and in simple (with a dot at the end).

searx/templates/simple/preferences.html
```diff
@@ -237,6 +237,18 @@ <h2>{{ _('Preferences') }}</h2>
...
+    {% if 'query_in_title' not in locked_preferences %}
+    <fieldset>
+      <legend>{{ _('Query in the page\'s title') }}</legend>
+      <p class="value">
+        <select name='query_in_title'>
+          <option value="1" {% if query_in_title %}selected="selected"{% endif %}>{{ _('Enabled') }}</option>
+          <option value="" {% if not query_in_title %}selected="selected"{% endif %}>{{ _('Disabled') }}</option>
+        </select>
+      </p>
+      <div class="description">{{ _('When enabled, the result page\'s title contains your query. Your browser can record this title.') }}</div>
```
return42, line 249: Here is a final dot.

searx/templates/oscar/preferences.html
return42: fixed in 3f3097c7

searx/templates/simple/preferences.html
return42: fixed in 3f3097c7

unixfox:
image

There is a space missing before the `-`, not sure where to fix that 🤔

return42:
> There is a space missing 

Discussion moved to https://github.com/searxng/searxng/pull/560

Issue #176, opened 2021-06-25
Simple theme tasks


dalf:
From:
* issue 1025 in searx
* https://github.com/searxng/searxng/issues/162

### General : UI
- [x] replace LESS variables by CSS variables, so the user can choose the line height, dark / light easily (?)
- [x] footer at the botton of the page: see https://github.com/dalf/searxng/commit/df0914e3b5298585015467abb0f9edd01e48dfea
- [x] center or move the results to the right.
- [x] redo color theme and just overall bigger UI elements; round corners and so on (right now its a little blunt)
- [x] add dark and black theme

see https://joe-crick.github.io/pure-css-components/

### General : UX
- [x] infinite_scroll plugin settings
- [x] opening pages in a new tab settings
- [x] do the RTL style 
- [x] engine crash warning : add a close button
- [ ] infoboxes : some properties value should be a link to another search ( ```Programming language :Python```  --> be able to click on ```Python``` to search that on searx)

###  Search
- [x] update HTML input attributes: see https://github.com/benbusby/whoogle-search/blob/main/app/templates/search.html
- [x] add safe search right to Language combobox and Time combobox
- [x] apply #1226 to Simple theme ( https://github.com/searx/searx/pull/1226 )

### On mobile
- [x] do not set the focus on the search input text : the keyboard hides the results.
- [x] make the search input smaller when without focus to give space for the logo and the preferences button. With focus, make it full width wide as now.
- [ ] embedded videos on mobile : too large, outside the screen
- [ ] selected category on mobile : scroll to the first selected category.

### Image category
- [x] give the user the choice to view the page a image is from instead of the actual image source

### Reduce the network usage
- [ ] create a small bandwidth version using javascript (use JS to add languages choices, cached links, categories)

return42:
### On mobile
- [x] add a link to the preference menu
- [x] add link to _about_ section

mrpaulblack:
I dont understand this one:

> add link to about section (On mobile)

There is also the about section link in the footer on mobile? @return42

return42:
I miss a clear link to the [about page](https://darmarit.org/searx/about)

ATM we have a link _more or less hidden_ in the footer under the word "searxng"

![grafik](https://user-images.githubusercontent.com/554536/148656192-5c58c34e-e027-41cf-9d81-a16cf5bcb711.png)

I think we need the "about" link at a more visible place.

---

The about page has some hints which might have a value for the users (e.g. browser integration, see below) 


[...]

PR #486, opened 2021-11-06
Simple theme: mobile layout: allow access to the preferences

dalf:
## What does this PR do?

Use [:placeholder-shown](https://caniuse.com/css-placeholder-shown) CSS selector which is supported since ~ 2017. If the browser doesn't support this selector, the preferences button remains hidden.

Front page:
image

With a query:
image

Without a query:
image

Fix a UX bug

## Related issues

related #176 and #463


mrpaulblack (approved):
LGTM :+1: 

Just FYI: This makes the issue with scaling worse on mobile since the `#q` element of the search form does have some sort of min width (and the margin-right makes it worse):

image

dalf:
It seems I miss something, I can't reproduce the issue :

image

---

With Chrome, I see a slight change of the width which change the whole page a little.

mrpaulblack:
Try to make the width of the page smaller than 360px; The problem is that the `#q` input has some sort of min width. Meaning  when you make the width of the entire page smaller at some point the text input field does not get smaller. There are two things that can happen now:

* when touch input is disabled the page is going to have a horizontal scroll bar
* when touch input is enabled: the entire view-port gets smaller since there is now horizontal scroll bar allowed (I guess?)

The solution: Find a way to make the `#q` element of the search input smaller when the width of the entire page decreases (This issue already existed before this PR, but since the margin-right got added it got worse, since the effect starts earlier now); but tbh this can also be done in another PR, since the issue is also in the master branch and the phone also needs to be really small for this to happen...

[EDIT] maybe this only happens with firefox and not with cromium based browsers?

dalf:
With Chrome, iPhone 5, DPI 2.0 :
image

Issue #445, opened 2021-10-28
simple theme: image detail: phone layout


dalf:
The image detail on phone has different issues:
* [ ] vertically, the labels at the bottom are too small and / or too far away from the image.
* [ ] horizontally, the labels are too big and the image is too small.

It may requires to change the viewport:
* [current value](https://github.com/searxng/searxng/blob/master/searx/templates/simple/base.html) 
* recommended value: https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#viewport_basics

---

For reference #376

dalf:
Technical note: https://github.com/searxng/searxng/blob/master/searx/static/themes/__common__/js/image_layout.js seems to get the wrong screen width on a smartphone. So behind the opened detail, the images spread a wider width than the screen. Consequence: the image detail adjust its width to this erroneous width.

PR #491, opened 2021-11-11
[fix] mobile result scaling in simple theme

mrpaulblack:
## What does this PR do?

Fixes the issue where result page would scale weird on small width screens, because of some min width on the `input#q` element on the search form. This is achieved by setting the width of that element to 100% (basically 100% of the 1fr of the grid element so it scales correctly now, when the width decreases). Also this PR hides the watermark on device width smaller 20rem (ultra-small-phone) to make space for the search input field.
]The changes viewport basically sets the initial scale to 1 (so on high dpi screens the browser can zoom in by default. We should not change this since the browser provider probably knows best which zoom level to use  by default). The max-scale and user-scale got dropped  so the user can zoom in as much as they like which should help with accessibility...

[EDIT] thanks to https://github.com/ZetaTom for the fix! :D

Make the simple theme A LOT more polished on mobile.

`make run`

* result page for each category should scale correctly on mobile now
* there is still the issue with the image detail modal scaling incorrectly (https://github.com/searxng/searxng/issues/445). It seems that the image is way to big for the side in portrait with some sort of min-width now. But since it was broken before, just differently, I think its fine to merge this anyway and fix the image in another PR

## Author's checklist

Small Phone:
image

Ultra Small Phone:
image


dalf (approved):
👍 LGTM

The image detail is broken with:
`<meta name="viewport" content="width=device-width, initial-scale=1">` 

But this is the right direction / the recommended value for smartphones.

PR #496, opened 2021-11-12
Simple theme: add "Engine tokens" field.

dalf:
## What does this PR do?

Add the missing ["Engine tokens"](https://searxng.github.io/searxng/admin/engines/private-engines.html) input field:

image

image

When selected:

image

image

* Check the tokens are saved
* Check the UI with different screen widths.

## Author's checklist

This part is hacky:
```css
#main_preferences .value select {
  width: 14rem;
}
#main_preferences .value input[type="text"] {
  width: 13.25rem;
```

I'm not sure how to fix that: `display: flex` on parent & `flex: 1` don't fix the width issue (some `select` element are too large)

## Related issues

* In the "User interface" tab, "Show advanced settings" is missing too. When enabled, the language / time selection / categories are shown in the front page (the user doesn't need to click a button to show these fields). The simple theme doesn't have that.
* In the "User interface" tab, "Style" is missing too (let the user choose the Oscar style).


mrpaulblack:
> In the "User interface" tab, "Show advanced settings" is missing too. When enabled, the language / time selection / categories are shown in the front page (the user doesn't need to click a button to show these fields). The simple theme doesn't have that.

I do not think that we need that; see: https://github.com/searxng/searxng/pull/336#issuecomment-923893527

> In the "User interface" tab, "Style" is missing too (let the user choose the Oscar style).

Is going to be implement with the dark, light, auto switch...

[EDIT] also the UI scales fine on desktop, and tablet. On mobile it works like the rest of the settings and needs to be reworked as a whole in another PR I think...

dalf:
Sorry, my comment was not clear: the purpose is to make a "diff" with the Oscar theme: except these two settings already discussed, with this PR, all the settings are shown in the Simple theme.

About the phone layout, as you said, it needs another PR.

mrpaulblack (changes requested):
I added a 4px radius in a commit (same as the search url in sidebar on result page) to engine token input in pref. Apart from that UI, setting a string LGTM :+1:

return42:
FYI: I'm in a review .. I rebased locally .. please do not change anything while I'm in review ..

If it is OK I will push a cleaned up rebase with patch from @MrPaulBlack  included ... 

coming soon ...

return42 (approved):
Perfect! .. lets merge ..

mrpaulblack (approved):

PR #499, opened 2021-11-13
[fix] startpage autocompletion

dalf:
## What does this PR do?

Fix startpage autocompletion.

Bug fix.


return42 (approved):
Works like a charm :) / thanks!

PR #500, opened 2021-11-13
Various marginal fixes and improvements

return42:
## What does this PR do?

Various marginal fixes and improvements / read along the commit messages

hint: ~the first commit disables  "Ahmia blacklist" plugin by default~ / all other patches do not change any functional


searx/plugins/ahmia_filter.py
```diff
@@ -7,7 +7,7 @@
 
 name = "Ahmia blacklist"
 description = "Filter out onion results that appear in Ahmia's blacklist. (See https://ahmia.fi/blacklist)"
-default_on = True
```
dalf: Despite this value, it should be disabled by default: see the `init` function.
The plugin is actually enabled only when `settings['outgoing']['using_tor_proxy'] is True`

return42: Thanks! .. I dropped 906690c and fixed comment about in settings.yml (6debda24 line 163)

searx/templates/simple/icons.html
```diff
@@ -20,13 +20,13 @@
...
-} 
+}
 -%}
-      
+
 {% macro icon(action, alt) -%}
   {{ icons[action] | replace("ionicon", "ion-icon") | safe }}
 {%- endmacro %}
-      
+
```
dalf, line 29: I missed that point: the fix about icons.html should take place here:
https://github.com/searxng/searxng/blob/26a6cbadef821c71aed9e5a2c82cb65eecf0586b/searx/static/themes/simple/gruntfile.js#L197-L215

return42: Yes, my fail 🤦🏻 .. I already send https://github.com/searxng/searxng/pull/501 to fix it in the right place

PR #501, opened 2021-11-13
[fix] delete trailing whitespace

return42:
## What does this PR do?

[fix] delete trailing whitespace


dalf (approved):
👍 

---

In theory, a change of gruntfile.js requires a call to `make static.build.commit`, but #500 already includes the changes.

PR #509, opened 2021-11-16
[engines] move library of genesis from general to files category

mrpaulblack:
## What does this PR do?

Move library of genesis from general to files category (maybe science would be better?).

I think it fits better in files and also: https://github.com/paulgoio/searxng/issues/13

```make run```

## Author's checklist

(Also it says on engines page that the engine does not support HTTPS; but in settings.yml the search url is HTTPS !?)

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -806,7 +806,7 @@ engines:
     url_xpath: //a[contains(@href,"bookfi.net/md5")]/@href
     title_xpath: //a[contains(@href,"book/")]/text()[1]
     content_xpath: //td/a[1][contains(@href,"=author")]/text()
-    categories: general
+    categories: files
     timeout: 7.0
     disabled: true
     shortcut: lg


return42:
> and also: paulgoio/searxng#13

FYI: library genesis is disabled by default

> maybe science would be better?

It is a _search engine for scientific articles and books_ / I would vote for _science_ category.

> Also it says on engines page that the engine does not support HTTPS; but in settings.yml the search url is HTTPS

You can remove the `enable_http: true` from the settings / but I miss some [documentation](https://searxng.github.io/searxng/admin/engines/settings.html) about options `enable_http` and `enable_https2`  / @dalf knows more about network layer, may be he can add some doc about.

dalf:
By default any HTTP connection is refused.
That can be tricky: the URL can be https://... it actually redirects to http://.... 

It was the case of an engine last year, may be libgen: the URL has been updated, but not the `enable_http`.
As @return42 , if the engine works without `enable_http: True`, then let's remove it.

---

Side note: I think the categorization needs some cleanup:
* gpodder, does it really fit into the general category ?
* mymemory translated , etymonline , duden, dictzone are dictionnaries (some times with specific not documentated syntax)
* erowid, azlyrics don't fit into the general category.
etc...

There is no perfect default solution, but I think the current situation can be improved.

return42:
>  if the engine works without enable_http: True, then let's remove it.

confirmed / it works without this setting.

return42:
>  if the engine works without enable_http: True, then let's remove it.

confirmed / it works without this setting.

return42:
Hm, this PR misses to remove ..

> >  if the engine works without enable_http: True, then let's remove it.

> confirmed / it works without this setting.

I just confirmed that it works without this setting / not that this setting has been removed in this PR .. sorry if it was misleading.

Opened: https://github.com/searxng/searxng/pull/520

mrpaulblack:
yeah I wanted to rework all the category placements of the engines... And also remove the http line; Should have marked this as draft sorry..

PR #510, opened 2021-11-17
[simple theme] rework select and impove url

### LINKED ISSUE #372: Simple theme: no javascript: searchbox: labels overlap
When javascript is disabled, the labels at the end of the searchbox overlaps:
image

https://github.com/searxng/searxng/blob/e39a03cc61e9792afb34084fb4d9973a61deecea/searx/templates/simple/search.html#L5-L6

The icon should be displayed and include labels (for terminals).

### LINKED ISSUE #504: [Simple Theme]: Suggestions to an even cleaner results page.

When we look at other search engine's results page, we can see that almost all of them replace the `/` in the result URL with `>`. This tends to look cleaner than having simply the URL in the result. I particularly like Brave's approach which shows just the domain name in bold darker letters (benefits somewhat to prevent phishing) and then the paths in lighter color,  the font size of the results is larger and more pleasing to the eye. Furthermore, as we can see in the second Brave screenshot below, if the path is greater than the size of the box then its simply hidden. 

So, if we move the URL below the title we can follow Brave's approach with the domain name being highlighted in bold and the forward slash `/` being replaced with a right chevron `>` and make the overall font-size of the results slightly larger. If we wish to keep the URL exactly where it is now (above the result title) then DDG's approach may look better with the `http://domain.com > path > path2`. Also, I believe someone proposed adding the website icons next to the results, and in this case, we may add them to the left of the URL like DDG does. 

I have not considered any at the moment. 

**Additional context**

DDG:
image

Brave Search: 
image

Brave Search with Overflow hidden:
image

SearXNG:
image

Even though this is mostly just eye candy stuff it may help the search engine look and feel similar to others. Thank you.

### What do you think?

mrpaulblack:
## What does this PR do?

This PR improves the simple theme:

* reworked select boxes in preferences
* moved search filter down underneath categories on result page
* added safe search as a filter; gets not saved to pref. (which happened on accident; but I actually think its the better option to not set the cookie every time someone just wants to change safe search once)
* make the article URL a flexbox consisting of scheme and domain and the second part is the path (the path gets cut of at the end of the line)
* replace / with > in the URL and drop params
* make URL 1rem and the h3 in article 1.2em
* fix the overlapping buttons when js is disabled in the search input
* padding around image previews changes to make videos, music and news category look better (maps looks a bit silly, but the whole category is kinda broken anyway)

Polish up the simple theme and make the URL and h3 heading easier to read as well as cleanup the select in pref. and search header.

```make run```

* make sure to test each cetegory with phone width and desktop width and also check pref. in dark and light theme

## Author's checklist

Desktop result page:
image

Mobile result page:
image

Desktop pref. general:
image

Desktop index with js disabled:
image

Mobile result with js disabled:
image

## Related issues

Closes #504, #372

Adding safesearch to search form #176

### CODE REVIEW

**Comment by mrpaulblack** [comment]
I have still two small issues:
* when them screen width gets small the domain part of a  url does not get cut off and there is horizontal scrolling (I dont know how to only cut it, if the path is already completely cut off)
* the path gets cut off at the right side; I would prefer it to cut of on the left...

(The first one needs to be fixed before merge, but do not know a good option how to do)

**Comment by return42** [comment]
> moved search filter down underneath categories on result page

Really like this / one question, can we size the place for the filters, to inhibit overwrites in other languages like DE ...

grafik

**Comment by mrpaulblack** [comment]
Ok so the fix would be to not set the dropdown as a background iamge but inline it as an svg (also makes css a lot simpler)... going to do this tomorrow....

**Comment by return42** [comment]
One nit I have / if you compare the margins top & button of the categories with the margins of the filters you see that these are differs .. for esthetic reason I would suppose to align them .. it is the same with the font color, the contrast of filters is higher. I would also adjust both with each other  .. but this is only my very personal preference. 

grafik

**Comment by return42** [comment]
Sorry for being pedantic :-) ...

> - make the article URL a flexbox consisting of scheme and domain and the second part is the path (the path gets cut of at the end of the line)
> - replace / with > in the URL and drop params

I like the new `>` ...

grafik

but I miss the `.mp4` at the end .. do you remember in https://github.com/searxng/searxng/issues/504#issuecomment-968831376 you suggested:

> IMO it is still pretty easy to get this information from the url with the > notation tbh:
> image

**Comment by dalf** [comment]
I like a lot of changes of this PR.

The location of the 3 filters is weird for me, but this is common place in many search engines, so okay (IMO, the current location is okay on a full screen, but meh on smaller screen, and the addition of the "safe" filter make the line too wide anyway).

* Note about the select element: ddg uses `select::after` and a custom font.
* About the safe filter: a label before the value would be helpful: the "None" text is puzzling without context.
* I would put the safe filter at the end.
* on mobile: would a horizontal scroll would be better? (with the 3 filters, it can be another PR).
* about URL, I like the fact because it highlights the domain name, but the URL from youtube looks all the same: 
image
   * Would it be help to have a Python function to rewrite the URL and then use it in the template?
   * I don't know how to fix this without restoring the parameters or making an endless list of specific rules. 
* Out of topic: In the no js version, the labels should be shorter ( "clear search"  -> "clear", "send search" -> "search", but visually it becomes "clear search", it can be fixed with a border-right: 1px on the clear button).

**Comment by mrpaulblack** [comment]
Yes that is one of the problems I am still having...

> the path gets cut off at the right side; I would prefer it to cut of on the left...

While writing my comment I found the solution: just use `direction: rtl;` on the path.... This way the left side gets cut of... :+1:

**Comment by mrpaulblack** [comment]
> Note about the select element: ddg uses select::after and a custom font.

ooofff

> About the safe filter: a label before the value would be helpful: the "None" text is puzzling without context.

yes agreed

> I would put the safe filter at the end.

also yes

> on mobile: would a horizontal scroll would be better? (with the 3 filters, it can be another PR).

good idea; actually I want to rewrite the result filter placement based on return42s comment anyway; I will probably just use grid layout which is easy to change on mobile (once it works on desktop)

> about URL, I like the fact because it highlights the domain name, but the URL from youtube looks all the same

Well I could just include the params in the path with jinja2 (at least I think so...) for every url, but tbh is it usefull to display the param anyway (people can right click on the url)

> Out of topic: In the no js version, the labels should be shorter ( "clear search" -> "clear", "send search" -> "search", but visually it becomes "clear search", it can be fixed with a border-right: 1px on the clear button).

yes

**Comment by return42** [comment]
> > Note about the select element: ddg uses select::after and a custom font.

> ooofff

Personally I think we do not need a special font and for me the select boxes are good as they are .. but this is something we can discuss in another PR or issue .. even if I had some nits mentioned above I think we should not overload this PR ..

**Comment by dalf** [comment]
> Personally I think we do not need a special font 

We just removed it, so no :)
The usage of `select::after` caught my attention, may be not relevant here.

**Comment by dalf** [comment]
Another method for the dropdown (on GH, see the dropdown icon):
```
{
    display: inline-block;
    width: 0;
    height: 0;
    vertical-align: middle;
    content: "";
    border-style: solid;
    border-width: 4px 4px 0px 4px;
    border-right-color: transparent;
    border-bottom-color: transparent;
    border-left-color: transparent;
}
```

[EDIT] Actually more or less out of topic.

**Comment by dalf** [comment]
About the ellipsis of URL:
<details>
<summary>A patch, no ellipsis but the path shrink first and then the host...</summary>

```patch
diff --git a/searx/static/themes/simple/src/less/style.less b/searx/static/themes/simple/src/less/style.less
index e16b53da..c869da7b 100644
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -223,8 +223,35 @@ article[data-vim-selected].category-social {
 
   .url_wrapper {
     display: flex;
+    flex-wrap: nowrap;
+    overflow: hidden;
+    flex-direction: row;
     margin: 0;
     padding: 0;
+
+    .url_o1 {
+      white-space: nowrap;
+      flex-shrink: 1;
+    }
+
+    .url_o1::after {
+      content: " ";
+      width: 1ch;
+      display: inline-block;
+    }
+
+    .url_o2 {
+      overflow: hidden;
+      white-space: nowrap;
+      flex-basis: content;
+      flex-grow: 0;
+      flex-shrink: 1;
+      text-align: right;
+
+      .url_i2 {
+        float: right;
+      }
+    }
   }
 
   .url {
diff --git a/searx/templates/simple/macros.html b/searx/templates/simple/macros.html
index 113dd181..b06fc37d 100644
--- a/searx/templates/simple/macros.html
+++ b/searx/templates/simple/macros.html
@@ -18,11 +18,11 @@
 <!-- Draw result header -->
 {% macro result_header(result, favicons, image_proxify) -%}
 <article class="result {% if result['template'] %}result-{{ result.template|replace('.html', '') }}{% else %}result-default{% endif %} {% if result['category'] %}category-{{ result['category'] }}{% endif %}{% for e in result.engines %} {{ e }}{% endfor %}">
-  {% if result.parsed_url.path | last == "/" %}
-  <p class="url_wrapper" aria-hidden="true">{{ result_link(result.url, result.parsed_url.scheme + "://" + result.parsed_url.netloc, 'url') }}{{ result_link(result.url, result.parsed_url.path | reverse | replace("/", "" ,1) | reverse | replace("/", " › "), 'url url_path') }}</p>
-  {%- else -%}
-  <p class="url_wrapper" aria-hidden="true">{{ result_link(result.url, result.parsed_url.scheme + "://" + result.parsed_url.netloc, 'url') }}{{ result_link(result.url, result.parsed_url.path.replace("/", " › "), 'url url_path') }}</p>
-  {% endif -%}
+  {{- result_open_link(url, "url_wrapper") -}}
+  {%- for part in get_pretty_url(result.parsed_url) -%}
+  <span class="url_o{{loop.index}}"><span class="url_i{{loop.index}}">{{- part -}}</span></span>
+  {%- endfor %}
+  {{- result_close_link() -}}
   {%- if result.img_src %}{{ result_open_link(result.url) }}<img class="image" src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}" title="{{ result.title|striptags }}" class="image" />{{ result_close_link() }}{% endif -%}
   {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}" alt="{{ result.title|striptags }}"/>{{ result_close_link() }}{% endif -%}
index 4dd7ec72..6039be54 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -419,6 +419,16 @@ def _get_enable_categories(all_categories):
     return [x for x in all_categories if x in enabled_categories]
 
 
+def get_pretty_url(parsed_url):
+    path = parsed_url.path
+    path = path[:-1] if len(path) > 0 and path[-1] == '/' else path
+    path = path.replace("/", " › ")
+    return [
+        parsed_url.scheme + "://" + parsed_url.netloc,
+        path
+    ]
+
+
 def render(template_name, override_theme=None, **kwargs):
     # values from the HTTP requests
     kwargs['endpoint'] = 'results' if 'q' in kwargs else request.endpoint
@@ -457,6 +467,7 @@ def render(template_name, override_theme=None, **kwargs):
     kwargs['searx_version'] = VERSION_STRING
     kwargs['searx_git_url'] = GIT_URL
     kwargs['get_setting'] = get_setting
+    kwargs['get_pretty_url'] = get_pretty_url
 
     # helpers to create links to other pages
     kwargs['url_for'] = url_for_theme  # override url_for function in templates
```
</details>

There is a way to make the ellipsis work putting the path and then the host in HTML (with a mix of rtl / ltr), but that's ... not the way to fix this issue IMO.

---

From 2013: 
* https://www.w3.org/wiki/Text-overflow_middle_cropping
* https://bugzilla.mozilla.org/show_bug.cgi?id=740910

mrpaulblack:
Woow thanks, but I found a better selution which only takes two lines: Put a `flex: auto auto;` on the domain and a `flex: 1 1;` on the path... I am going to upload a WIP commit in the next few minutes; the only thing that is missing now is that the url path is cutting of on the right and not the left but I am inclined to merge this PR in its current state and fix up the issue later tbh...

mrpaulblack:
oh wait your patch also fixes the path cutting of on the left thanks!...

mrpaulblack:
sry for the commit spam; ready for review now;

Should work now:
* 3 search filters on every result page (the text does not overlap the drop down arrow)
* safe search filter is labeled with SafeSearch
* design of select, text input and categories now match on pref. page
* url in result article now cuts of path on the left and on small screens the domain, after the path gets cut of
* when js is disabled: there is a clear and a search btn
* when js is disabled: the category tooltip is now displayed above the categories on result page
* URL and h3 in results are now bigger (1rem and 1.2em)

return42:
While I am just right now in the review, one note I want to leave here ..

> safe search filter is labeled with SafeSearch

I think, we should fix the German translation ...

grafik

https://weblate.bubu1.eu/translate/searxng/searxng/en/?checksum=56b1b2e873251718&sort_by=-priority,position#translations

@mrpaulblack you are native DE speaker, what do you think, should I fix the translation that way:

    Sichere Suche (SafeSearch) --> Sichere Suche

mrpaulblack:
Yeah I think that would be way better :+1:

dalf:
> url in result article now cuts of path on the left and on small screens the domain, after the path gets cut of

After a second thought, without an ellipsis, I think the cut in the middle is confusing

mrpaulblack:
can we do this in another PR? Since there is a space between domain and path I think its fine for now...

return42 (approved):
> when js is disabled: there is a clear and a search btn

grafik

- pressing the clean button does not have any effect (tested: chrome & ffox).
- "Click on the magnifier ..." is wrong here, since there is a "search" button.  IMO we can drop this "Click on the magnifier ..." completely, the UI is intuitive enough I think.

> when js is disabled: the category tooltip is now displayed above the categories on result page

I am confused, do we have a tooltip on the result page? I assume you mean "Click on the magnifier to perform search" / see my comment above, IMO we can drop this.

Over all I think we should merge this PR, it is a huge step in the right direction / polish can be done afterwards.

Thanks a lot!

PR #512, opened 2021-11-17
[fix] Qwant search query string

ZetaTom:
## What does this PR do?
Search string: "!qwant time"
Resulting request URL: https://api.qwant.com/v3/search/web?q=q=time&count=10&offset=0&device=desktop&safesearch=1&locale=en_US
Notice the double `q=`

Resulting request URL after fix: https://api.qwant.com/v3/search/web?q=time&count=10&offset=0&device=desktop&safesearch=1&locale=en_US

Improve qwant search result quality.
Currently the search string is prefixed with `q=` resulting in unwanted search results.

`make run`

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -59,7 +59,7 @@ category_to_keyword = {
 }
 
 # search-url
-url = 'https://api.qwant.com/v3/search/{keyword}?q={query}&count={count}&offset={offset}'
+url = 'https://api.qwant.com/v3/search/{keyword}?{query}&count={count}&offset={offset}'
 
 def request(query, params):
     """Qwant search request"""


mrpaulblack (approved):
We tested with general, images, videos and news category. They now return the same results as with `qwant.com`. Also looking at other engines, they use the same syntax. (So I think it is just a typo tbh)

return42:
@ZetaTom thanks for fixing the issue.

> So I think it is just a typo 

Yes, I can confirm, It is a typo from mine in ef6e1bd6 / sorry :)

The required `q` is already in the urlencode ..

https://github.com/searxng/searxng/blob/ae2f6c63f4fbdddee9a89a453ddcd7439a387f13/searx/engines/qwant.py#L79-L81

PR #446, opened 2021-10-28
[brand] SearXNG environment variables & /utils scripts

return42:
## What does this PR do?

Renames various environment variables from `SEARX_...` to `SEARXNG_..` 

For more details have a look at the commit messages

## How to test this PR locally?

I tested the scripts in LXC and at https://darmarit.org/searx/search

## Related issues

Related: 

- https://github.com/searxng/searxng/pull/358
- https://github.com/searxng/searxng/issues/425

PR #515, opened 2021-11-18
 Various modification of the documentation and the theme sphinx-theme

return42:
## What does this PR do?

Various modification of the documentation and the theme sphinx-theme.

For any details, read along the commit messages.

    make docs.clean docs.live

You should not see any error massage in the build log of the documentation.

Have a look at the produce HTML .. the logo has been changed ...

grafik

and there is a "source" link in the sidebar.

grafik


dalf:
LGTM with two notes:
* I would move the favicon to a `brand` directory, so the documentation, the themes (oscar and simple) can use it. Same for the "SearXNG" image. Since it requires to rebuild the simple theme, it can be done in a second PR.
* I don't know how to solve this issue:  this PR changes the documentation in a way that someone who has followed the current version and then tries read it again, won't find the various files (since there are renamed). I mean a copy / paste of `cat /etc/apache2/sites-available/searxng.conf` won't work, since the previous was `cat /etc/apache2/sites-available/searx.conf`.

return42:
> I would move the favicon to a brand directory, 

👍🏻 

> I don't know how to solve this issue:  ...won't find the various files (since there are renamed). 

That's right / I also do not know a solution for .. while this PR is only about the documentation, the files has been renamed in PR https://github.com/searxng/searxng/pull/446 / we discussed this problem in this PR https://github.com/searxng/searxng/pull/446#issuecomment-954599668 and we merged https://github.com/searxng/searxng/pull/456 to have a script that checks the installation.

If we think there is a need for, we can extend [searxng_check.py](https://github.com/searxng/searxng/blob/master/utils/searxng_check.py).

[EDIT] if I have a doubt about my instance (darmarit.org), I simple remove SearXNG, morty & filtron and do a complete new installation (thanks to the scripts it is just a 5 minutes task).

PR #518, opened 2021-11-19
[fix] simple theme: fix image detail layout on mobile

### LINKED ISSUE #445: simple theme: image detail: phone layout

dalf:
## What does this PR do?

See commit messages.

An screenshot to visualize the fixed issue. The images when the detail is opened (but hidden using the devtools):
image

we see that images overflow on the right (we can actually scroll).

---

About the second commit. On the master branch, the image detail slightly overlaps the search fields:
image

Fix UI / UX issue.

* `make run`
* using the devtools, set the device to mobile
* search for something in the image category
* click on an image

## Author's checklist

I've tried
```less
#main_results #results.image-detail-open.only_template_images {
  width: 59.25rem !important;
}

@media screen and (max-width: @tablet) {
  #main_results #results.image-detail-open.only_template_images {
    width: 98% !important;
  }
}

but it doesn't work
```

the BackToTop button keeps the hard coded, but it is hidden on mobile...
```
#main_results #results.only_template_images.image-detail-open #backToTop {
  left: 56.75rem !important;
  right: inherit;
}
```

## Related issues

Close https://github.com/searxng/searxng/issues/445


mrpaulblack (approved):
LGTM I tested with ff dev tools and both mobile fix works (image sometimes not visible on landscape, but has nothing to do with this PR) and also the height of the image modal no matches the search form header... :+1: 

[EDIT] also thank you for the fix; one step closer to making simple theme the default :+1:

return42:
> The images when the detail is opened (but hidden using the devtools):

confirmed!

> On the master branch, the image detail slightly overlaps the search fields:

confirmed!

PR #521, opened 2021-11-19
Simple theme: allow the user to choose auto, light, dark style

dalf:
## What does this PR do?
Simple theme: this PR allows the user choose the theme style:
* auto: it follow the browser preferences (same as now)
* light: force the light style
* dark: force the dark style

image

----
Implementation:
* in definition.less, the values of the dark style are defined in a LESS mixins, which to define them twice (once for `:root.theme-dark`, once for the media query).
* `settings.yml`: `ui.theme_args.simple_style` (underscore)
* new cookie and value in preferences.py: `simple-style` (dash like `oscar-style`)
* the choose auto, light, dark can be translated
* the html element contains the user choice. For example : `<html class="no-js theme-dark" lang="fr-FR" >`. 
```
<html class="no-js theme-dark" lang="en-EN" >
```
* `searx/static/themes/simple/src/js/head/00_init.js` is changed not to erase this new CSS class.
* the CSS about the `select` element also make a detection about style.

One step further to change the default theme.

matrix tests:
* in the SearXNG preferences: light, dark, auto
* in the browser preferences: light, dark

Each time:
* check if the SearXNG preferences are followed
* check the icon of the `select` HTML element: 

dark: image
light: image

## Author's checklist

* The new cookie is `simple-style`, may be it should be `simple_style` ? (the only exception to use a dash is `oscar-style`).
* In a second step, it is possible to let the user change the style directly from the `/` and `/search`

## Related issues

Related to https://github.com/searxng/searxng/issues/463#issuecomment-955787945


mrpaulblack (approved):
LGTM :+1:  (but I would prefer it to be `simpl_style` tbh)

* when simple style is auto the system pref. gets selceted
* when simple style is light the light theme bets displayed regardless of device pref.
* when simple style is dark the dark style gets displayed regardless of device pref.

Also looks like the favicon is broken?:
image

dalf:
> but I would prefer it to be simpl_style tbh

I will change it

> Also looks like the favicon is broken?

Related to the second part of this comment: https://github.com/searxng/searxng/pull/473#issuecomment-967231150

dalf:
Updated to use `simple_style` everywhere instead of a mix with `simple-style` in some places.

/!\ there is no static build commit since it breaks the favicon

mrpaulblack:
LGTM; I also pushed a static build; It worked fro me... (I am running latest manjaro -> ImageMagick is version 7.1.0.10-1; maybe your version is broken or smt....)

return42 (approved):
Cool .. works perfect .. thanks to @mrpaulblack building the favicon, which is also perfect.

In my Chrome the "auto" option do not work .. but this seems an Chrome issue, when I navigate to github content, the system preferences are also ignored in Chrome...

This brings my to an other question, if some browser do have an issue with "auto" it would be nice to have dark/light toggle button directly on the result (and intro) page close to the "hamburger" menu.

dalf:
> it would be nice to have dark/light toggle button directly on the result (and intro) page close to the "hamburger" menu.

yes that's the plan. It requires javascript to send an AJAX request to save the preferences.

return42:
> save the preferences.

hm, really save preferences? .. There are user having all kind of cookies disabled.  I don't think this setting should be "saved" .. an request param could be a solution(?).

mrpaulblack:
> hm, really save preferences? .. There are user having all kind of cookies disabled. I don't think this setting should be "saved" .. an request param could be a solution(?).

Well in the search post request (or get if chromium) the theme is already sent as an hidden input; maybe there could be another one for the style. The problem is tho that this would only be sent when a search request is made e.g. not on /about /stats and index page an so on... The only way I see this work is saving the cookie or by setting a param as you suggested and setting it with every click on a link (seems a little overly complicated tho); I think best option is a cookie.

return42:
An alternative could be, if this "toggle button" only works when a request is send .. IMO this is the main use-case of this toggle button .. if the user wants more, he/she should allow a cookie .. or just use a browser that supports "auto" ..

What I mean .. I just want to switch temporary between dark/light in the result page .. like you have done it for the result filters in https://github.com/searxng/searxng/pull/510 .. this would be an convenient option I think.

PR #522, opened 2021-11-19
[fix] brand - partial revert of e82e37ce6

return42:
## What does this PR do?

In commit e82e37ce6 accidentally the brand file was touched / partial revert::

    export GIT_URL='https://github.com/searxng/searxng'
    export GIT_BRANCH='master'

Related: https://github.com/searxng/searxng/pull/515


dalf:
Since it is built each time from settings.yml, does this file need to be commit?

PR #524, opened 2021-11-19
[drop] engine digg - https://digg.com/api is no longer available

return42:
## What does this PR do?

[drop] engine digg - https://digg.com/api is no longer available

REST / JSON - API does no longer exists

---

[EDIT] Since digg no longer works, we do nat have a active engine in the social-media category.  Enable reddit by default to have at least one engine back in this category .. see second commit.


mrpaulblack (approved):
LGTM :+1: 

Yeah reddit is also way faster than the 10secound timeout on my instance and I agree with enabling it by default

PR #527, opened 2021-11-20
[fix] simple theme: fix modal dialog

dalf:
## What does this PR do?

Currently, the help of keyboard shortcuts is displayed incorrectly:
image

This PR fixes this issue using `display:block`.

In addition, the dialog appears with a slight animation

Fix UI

* `make themes`
* `make run`
* enable vim shortcuts in the preferences
* press key `h`


return42:
ready to review?

dalf:
Yes (the static build commit is missing)

return42:
LGTM .. one nit I have .. if mobile user accidentality types `h` the modal pops up, but there  is no chance to close the modal dialog .. due to centric alignment, the _close window `X`_ is not shown in the view port.

grafik

mrpaulblack:
I added a small commit to move the keyframe animations into a new file: animations.less I think it would be good to move all animations and all transitions into this file to keep them all together; Fell free to force push over my commit it is just an example...

return42:
>  one nit I have .. if mobile user accidentality types h the modal pops up, but there is no chance to close the modal dialog

Ooops, I was wrong .. 

grafik

FYI: I rebased and added the `[build] /static` on top ..

Thanks for this PR!

dalf:
> I added a small commit to move the keyframe animations into a new file: animations.less I think it would be good to move all animations and all transitions into this file to keep them all together; Fell free to force push over my commit it is just an example...

Note: there is another animation here:
https://github.com/searxng/searxng/blob/0e27578e3528f6dc95a0ae8801e1972b3fdc1766/searx/static/themes/simple/src/less/search.less#L247-L250

PR #528, opened 2021-11-20
[brand] initial - folder src/brand for searxng-wordmark

return42:
## What does this PR do?

Folder src/brand is intended to place logo, wordmark and more in.  This commit moves the origin (source) of searxng-wordmark.svg from the simple theme into this folder.

About the new created src/ folder:

On the long term we can move all the files from `static/themes/<name>/src/*` into a `src/themes/<name>/*` folder.

The simple theme is the wrong place for brand files / these files are used not only by the simple theme.

    make themes.simple


mrpaulblack (approved):
LGTM;

I really like this idea. Moving all the build files for the themes into the `/src` folder will really make it more easy for new devs to start working on the themes IMO...

return42:
> I really like this idea. Moving all the build files for the themes into the `/src` 

I discussed this with @dalf (https://github.com/searxng/searxng/issues/141#issuecomment-890481065): beside it is more clear to new devs, the source do not need to be public in the `/static` / For example:  https://darmarit.org/searx/static/themes/simple/gruntfile.js

PR #535, opened 2021-11-21
[simple template] return a default string if there is no description

### LINKED ISSUE #458: simple theme: UI issue with results without description
_Originally posted by @MrPaulBlack in https://github.com/searxng/searxng/issues/457#issuecomment-954903607_ :

I think we should add a default description if the returned result from the engine does not provide any; Right now results look really dull without a description and a simple `The website did not provide a description.` could make the elements of the simple theme a lot more streamlined looking...

Current layout:
image

(And yeah you are right, once images are using grid layout it should be pretty easy to fix the padding. Just a selector if they are inside the default result container and making the left margin or gap bigger...)

mrpaulblack:
## What does this PR do?

Return a default string if an article does not have any description. This applies to default and videos results (general, videos, news, music, it, sciences and social media category). I am not sure about the string that should be returned tho...

Make the articles on the result page of the simple theme look more uniform...

```make run```

* This can be best tested in the it category.

## Author's checklist

Desktop IT category:
image

## Related issues

Closes #458

### CODE REVIEW

**Comment by return42** [comment]
FYI I rebased and added a commit to format the default text in italic ..

grafik

@dalf @mrpaulblack : in commit edd0c104a3a I added some indents for better readability / what do you think, are there any reasons not to indent?

**Review by return42 — APPROVED** [review]

**Comment by mrpaulblack** [comment]
Cool thnak you; could you add the italic font selector to mixins.less instead of style.less please?

**Comment by return42** [comment]
> could you add the italic font selector to mixins.less

clear! / done .. 

Before merge, lets wait, what @dalf says about my indentation in jinja2 files ..

> in commit edd0c104a3a I added some indents for better readability / what do you think, are there any reasons not to indent?

... I don't know if there are any drawbacks, when we add indentations.

File: searx/templates/simple/result_templates/videos.html [hunk 1]
```diff
@@ -5,7 +5,15 @@
...
-{% if result.content %}<p class="content">{{ result.content|safe }}</p>{% endif %}
+<p class="content">
+  {%- if result.content %}
+    {{ result.content|safe }}
+  {%- else %}
+    <div class='italic_font'>
+      {{ _('This site did not provide any description.')|safe }}
+    </div>
+  {% endif -%}
+</p>
```
dalf, line 17: Until now most (all? at least some?) of the CSS classes are named after their meaning.

If we continue to follow this spirit, I suggest this:
```jinja2
{%- if result.content -%}
  <p class="content">{{ result.content|safe }}</p>
{% - else - %}
  <p class="content empty_content">{{ _('This site did not provide any description.')|safe }}</p>
{%- endif -%}
```

If someone wants an preference to hide the empty content, it can be implemented only with CSS.

Yes, it is also possible to have `.content .italic_font` except when we will have another usage of the `italic_font`.

mrpaulblack: So if I get this right: The class name should be specific to the html element it got put on to and not to the actual CSS attributes... I think this is the better option too. @return42 do you want to do the changes? I can do them otherwise this evening....

dalf: > The class name should be specific to the html element it got put on to and not to the actual CSS attributes... 

As an alternative, instead of `empty_content` it can be `empty_element`, so we can flag other HTML elements with this CSS class ; and it is still about the meaning.

return42: I can do it .. if we rename it, is `mixins.less` any longer the right place? ... I don't think so (there are no class names that are specific to the html element it got put on).

Is `style.less`a better place?

return42: > If someone wants an preference to hide the empty content, it can be implemented only with CSS.

done!  .. hope `style.less` is the right place .. otherwise change it up to your meanings.

mrpaulblack:
Thanks LGTM :+1:

PR #537, opened 2021-11-23
[simple theme] remove '-moz-border-radius' property

return42:
## What does this PR do?

[simple theme] remove '-moz-border-radius' & '-webkit-border-radius' property

Support for the prefixed version (-moz-border-radius) was removed in Gecko
13.0 (Firefox 13.0 / Thunderbird 13.0 / SeaMonkey 2.10) [1].

[1] https://developer.mozilla.org/pt-BR/docs/Web/CSS/border-radius

Remove warning spam from the console ..

grafik

Open SearXNG site in FFox browsers


mrpaulblack (approved):
LGTM :+1:
IMO I dont think we need to leave this in for such old browser version tbh; so agreed with the PR

dalf:
What about `-webkit-border-radius` ? Do we remove it ?

mrpaulblack:
`Safari ignores border-radius on <select> elements unless -webkit-appearance is overridden to an appropriate value.`

So I would keep the webkit vendor prefix since there is a 5px border radius on the select elements...

dalf:
Okay, let's keep the webkit vendor prefix ( https://caniuse.com/?search=css3%20border-radius doesn't make a reference to this. )

--- 

Using CSS grids, the simple theme requires Safari 10.1 on OSX and Safari 10.3 on iOS: https://caniuse.com/?search=grid

return42:
Hm, I am a little bit confused ..

> Safari ignores border-radius on `<select>` elements unless -webkit-appearance is overridden to an appropriate value.

... who says?

> Okay, let's keep the webkit vendor prefix ( https://caniuse.com/?search=css3%20border-radius doesn't make a reference to this. )

.. when I look at https://caniuse.com/?search=css3%20border-radius I can see `border-radius` is supported since 2010 (macOS & iOS) 

.. or I am wrong? / If not I would like to remove `-webkit-border-radius` like @dalf suggested in the first place.

return42:
FWIW: I removed `-webkit-border-radius` as @dalf suggested .. maybe there is someone who can test this PR in Safari ...

mrpaulblack:
> ... who says?

Check the notes on safari in the compat. table: https://developer.mozilla.org/pt-BR/docs/Web/CSS/border-radius

The tradeoff if we remove it would be that the border radius in pref on the select boxes wont work on safari; but cannot test this and dont know if its worth keeping it just for that tbh...

dalf:
* The MDN documentation is not clear:
   * it seems it's about the Chrome version 1 to 4
   * the English version doesn't have this information
* `<select style="border-radius: 1rem">...`  works on Chrome : https://jsfiddle.net/dalf73/nk6x9y1a/1/

return42:
> Check the notes on safari in the compat. 

Ooops . I have overlooked this / thanks.

> it seems it's about the Chrome version 1 to 4

Chrome? .. @mrpaulblack mentioned Safari

MDN is talking about Safari 5.0

grafik

Safari 6.0 is on the market since 2012, I think we can drop the vendor prefix in this case.

PR #539, opened 2021-11-23
fix PR #535

dalf:
## What does this PR do?

Use the "empty_element" CSS class.


return42 (approved):
Aargh, my fail / thanks for the fix!

mrpaulblack:
Oh and I did not catch it before merge either sorry!

PR #543, opened 2021-11-24
[fix] templates: remove unneeded escape \' of single quotation mark

return42:
## What does this PR do?

[fix] templates: remove unneeded escape \' of single quotation mark

BTW: remove a leading dot in the simple theme [1].

Strings like:

    'Query in the page\'s title'

are hard to read / remove escape sequence by using double quotation marks for strings:

    "Query in the page's title"

## Related issues

[1] https://github.com/searxng/searxng/pull/485/files/80fb77476fad4b229418c530f3ffda67f357a15a#r756112716


dalf (approved):
As far I understand the purpose is also to remove the additional dot ( https://github.com/searxng/searxng/pull/485#pullrequestreview-814956363 ).

PR #545, opened 2021-11-25
[fix] simple-theme templates: polish HTML <img> elements

### LINKED ISSUE #523: [simple theme] news: image, alternative text overwrites box intended for the image

current master

**How To Reproduce**

In category news search for items which have a long alternative text / and the image is not available ..

**Expected behavior**

Text should not overwrite ..

**Screenshots & Logs**

grafik

return42:
## What does this PR do?

[fix] simple-theme templates: polish HTML  elements

- `` [1] element does not have an end tag, no need for a leading `/>`
- add global attribute title[2] to `result_templates/images.html` (result-images-source)
- in jinja macro 'macro result_header' remove duplicate of class="image"
- in jinja macro 'macro result_header' remove `alt` attribute (fix [3])::

      alt="{{ result.title|striptags }}"

  the result.title is already shown in::

[1] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img
[2] https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-title
[3] https://github.com/searxng/searxng/issues/523

## Related issues

Closes: https://github.com/searxng/searxng/issues/523

### DIFF
--- a/searx/templates/simple/infobox.html
+++ b/searx/templates/simple/infobox.html
@@ -1,6 +1,6 @@
 <aside class="infobox">
   <h2><bdi>{{ infobox.infobox }}</bdi></h2>
-  {% if infobox.img_src %}<img src="{{ image_proxify(infobox.img_src) }}" title="{{ infobox.infobox|striptags }}" alt="{{ infobox.infobox|striptags }}" />{% endif %}
+  {% if infobox.img_src %}<img src="{{ image_proxify(infobox.img_src) }}" title="{{ infobox.infobox|striptags }}" alt="{{ infobox.infobox|striptags }}">{%- endif -%}
   <p><bdi>{{ infobox.content | safe }}</bdi></p>
   {% if infobox.attributes %}
   <div class="attributes">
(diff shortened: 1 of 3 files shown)


mrpaulblack (approved):
LGTM; I think this
image
 Is way better than the overlapping text before :+1:

return42:
> Is way better than the overlapping text

yes, it is a little bit better than before but there is space left to improve https://github.com/searxng/searxng/issues/523#issuecomment-974117015

dalf:
I'm sorry for the late comment.

in images.html, is the title attribute necessary?

It shows duplicate information on the screen:
image

* the tooltip managed by the browser
* the existing tooltip.

I prefer to keep the existing tooltip: it can contain HTML elements, and can be placed in the location we want.

return42:
Sorry, my fail .. I haven't checked truly, I thought the rendered tooltip is taken from the `title=` attribute / I can fix this, but first I have a question:

> I prefer to keep the existing tooltip: it can contain HTML elements, and can be placed in the location we want.

Can you point me to the place where I can see the _tooltip_ implementation in simple theme /  is it in the template or in the JS part?

dalf:
> Can you point me to the place where I can see the tooltip implementation in simple theme / is it in the template or in the JS part?

https://github.com/searxng/searxng/blob/f0db33e14b5b443eedc25f5aeba6cd330476c851/searx/static/themes/simple/src/less/style.less#L354-L364

In the devtool (Firefox, Chrome, ...) when you select a HTML element in the "Inspector" tab, you can see the reference to the source code:
image

Here this is the CSS class when the element is hidden. You can right click on the element, "Change pseudo class", "hover" --> you will be able to see the element as if the mouse is hover it.

return42:
Thanks for explanation .. 

> It shows duplicate information on the screen

I guess I have found the cause for .. if I'm not wrong, it is not caused by this PR, even if I revert 8fbffb1409 I see the tooltip from the browser. 

I will send a PR that fix this issue.

return42:
>> It shows duplicate information on the screen

> I will send a PR that fix this issue.

@dalf: discussion moved to PR #556.

PR #551, opened 2021-11-26
[fix] searx.results: fix pylint issue "useless-with-lock"

dalf:
## What does this PR do?

Replace "with threading.Lock():" by "with self._lock"

In the master branch:
* https://github.com/searxng/searxng/blob/da1d502cf7096c0c2f43b6b88dc413f91a43ebba/searx/results.py#L272-L273
* https://github.com/searxng/searxng/blob/da1d502cf7096c0c2f43b6b88dc413f91a43ebba/searx/results.py#L317-L318

The code creates a new RLock, acquires the lock, release it, delete it.

This PR creates one RLock per ResultContainer instance.

Make the code thread safe.

A test requires to creates to call `ResultContainer.extend` from two different threads to make sure one thread waits the other.
Even it may occur some times, that's difficult to do with user queries.

## Related issues

Related to https://github.com/searxng/searxng/pull/549

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -145,7 +145,8 @@ class ResultContainer:
     """docstring for ResultContainer"""
 
     __slots__ = '_merged_results', 'infoboxes', 'suggestions', 'answers', 'corrections', '_number_of_results',\
-                '_closed', 'paging', 'unresponsive_engines', 'timings', 'redirect_url', 'engine_data', 'on_result'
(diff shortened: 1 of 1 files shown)


return42 (approved):
Can't test threading but I did not see any regression in my local tests / lets merge.

mrpaulblack (approved):
Yeah it also worked on my system... I also think we should merge... If it does not work, it is going to show the latest on integration testing in my fork where i have 16 threads by default... (if I understand this PR correctly)

PR #553, opened 2021-11-26
[fix] oscar theme: use SearXNG logo and wordmark

return42:
## What does this PR do?

[fix] oscar theme: use SearXNG logo and wordmark

This is SearXNG / not searx

Have a look at oscar theme

## Author's checklist

The days of the oscar themes are counted / I just copied the files, no need for a build chain, like we have implemented for the simple theme.

----

grafik


mrpaulblack (approved):
I did some small fixups (feel free to force push if you dont like): center the search bar more on index and also use the SearXNG logo for ponthi since the watermark does not make sense here;

Everything else LGTM :+1:

return42:
> center the search bar more on index and also use the SearXNG 

good point / thanks a lot!

PR #554, opened 2021-11-26
[fix] google-videos engine: ignore news articles

return42:
## What does this PR do?

[fix] google-videos engine: ignore news articles

In the video search, google also sometimes includes news.  E.g. in the DE language when you search for `!gov paris`, google adds an article from a german newspaper (FAZ), I assume these are sponsored links (not tagged advertisement?)

Those links do not have an image / this patch ignores *video links* without an image ID.

grafik

    !gov paris

---

OT here / but while war are a google engines ..

@unixfox you pinged me about _"Another Google search parameter to set country"_ issue 3073 at searx .. I don't know if we really need this parameter / if you think there is a need for, please open an issue here at SearXNG / I don't want to spread a discussions about a SearXNG topic in other repositories / hope that is OK for you / thanks!

### DIFF
--- a/searx/engines/google_videos.py
+++ b/searx/engines/google_videos.py
@@ -154,25 +154,23 @@ def response(resp):
     # parse results
     for result in eval_xpath_list(dom, '//div[contains(@class, "g ")]'):
 
-        # google *sections*
+        # ignore google *sections*
         if extract_text(eval_xpath(result, g_section_with_header)):
             logger.debug("ingoring <g-section-with-header>")
             continue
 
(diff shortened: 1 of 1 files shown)


unixfox:
Not sure why you don't want to discuss on the searx repo, it's about a general discussion with the google engine... but I can copy the issue to the searxng repo.
Or we can talk about it on matrix/element.

return42:
FYI: I'm going to merge this PR, it fixes an issue in the google-videos engine in the master branch.

PR #562, opened 2021-11-29
[simple] ImageLayout.watch: `img_load_error.svg` if img load fails

return42:
## What does this PR do?

[simple] ImageLayout.watch: `img_load_error.svg` if img load fails

**before**

grafik

**after**

grafik

Use simple theme and search in category `!images` for images that can't be loaded.

## Some words about class ImageLayout

The https://github.com/searxng/searxng/blob/master/searx/static/themes/simple/js/searxng.js is build by a grunt task ..

https://github.com/searxng/searxng/blob/d0e21a01b4e14df905f25445a3f6bcbec481671c/searx/static/themes/simple/gruntfile.js#L91-L93

The `/__common__/js/*.js` concats also https://github.com/searxng/searxng/blob/master/searx/static/themes/__common__/js/image_layout.js where a modified copy of the of the "Google-image-layout" (`ImageLayout`) is implemented [1][2][3].

[1] https://ptgamr.github.io/2014-09-12-google-image-layout/
[2] https://ptgamr.github.io/google-image-layout/src/google-image-layout.js
[3] https://github.com/ptgamr/google-image-layout/tree/master

----


mrpaulblack:
After just briefly checking out your PR I think the SVG for a missing image should not be part of the brand folder but rather part of the simple theme src folder (e.g. some sub directory)... Also thank you for the PRs to fix up all the issues with the simple theme :slightly_smiling_face: :+1:

searx/static/themes/__common__/js/image_layout.js
```diff
@@ -136,6 +147,11 @@
     var results_nodes = d.querySelectorAll(this.results_selector);
     var results_length = results_nodes.length;
 
+    function img_load_error(event) {
+      // console.log("ERROR can't load: " + event.originalTarget.src);
+      event.originalTarget.src = window.searxng.static_path + 'img/img_load_error.svg';
```
dalf: Nit: `window.searxng.static_path` --> `w.searxng.static_path` (`w = window`, see at the end of the file)

---

In the Oscar theme, the image won't be found.

A solution: create a not generate d file in `img/img_load_error.svg` in the Oscar theme.

Another solution is to make the image optional with an additional parameter `errorImgUrl`:

```js
function ImageLayout(container_selector, results_selector, img_selector, verticalMargin, horizontalMargin, maxHeight, errorImgUrl) {
    this.errorImgUrl = errorImgUrl

    ...
    function img_load_error(event) {
      // console.log("ERROR can't load: " + event.originalTarget.src);
      event.originalTarget.src = this.errorImgUrl
    }

    ...

    if (this.errorImgUrl) {
        img.addEventListener('error', img_load_error, {once: true});
    }
}
```

then replace
https://github.com/searxng/searxng/blob/0311eba538df589bc11a4d76d3c94b104c5b764d/searx/static/themes/simple/src/js/main/results.js#L6
by
```js
searxng.image_thumbnail_layout = new searxng.ImageLayout('#urls', '#urls .result-images', 'img.image_thumbnail', 14, 6, 200, searxng.static_path + 'img/img_load_error.svg');
```

return42: > Another solution is to make the image optional with an additional parameter errorImgUrl:

I picked up this your suggestion, but instead of adding (more) arguments to the ImageLayout constructor, I added the `window.searxng.theme` namespace which can be passed through to the JS implementations / 7dc9cc9

IMO we should have one place (_single point of definition_) where we set all the theme values (parameters) for the JS client side.  If you think it is a good idea to have `window.searxng.theme` we can move `verticalMargin`, `horizontalMargin`, `maxHeight` and may be other paramaters as well to this namespace.

dalf: * When #368 will be implemented, `searx/static/themes/__common__/js/image_layout.js` is going to be removed. The margins, widths will be CSS variables. There is one exception to this: the error handling. It can be implemented as a function in `searx/static/themes/simple/src/js/main/results.js`.
* `searx/static/themes/simple/src/js/head/00_init.js` exists only for one reason : it changes some CSS classes. If this change is made at the end of the HTML file, for a very short time, the HTML page is displayed without these changes, which is uncomfortable.
* --> Why not for a central place, but I wish to avoid the `searx/static/themes/simple/src/js/head/00_init.js` location: the browser loads this file in the `<head>` section = the larger the file, the slower the response time. One variable won't change the response time, but if we move all the theme constant to this place, I don't know how long the list will be. 

Dumb idea: why not CSS variables and then read them from Javascript? We could change the image according to the theme, media queries etc for free. See https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#values_in_javascript ?

[EDIT] or we can postpone this decision later and add one more argument to ImageLayout so we can focus on the `img_load_error.svg` topic.

PR #565, opened 2021-11-30
[mod] simple theme: move initilization of document.searxng to searxng.min.js

dalf:
## What does this PR do?

`searxng.head.min.js` focuses on browser features detection nothing more (to add / remove some CSS classes).
The parsing and evaluation of `searxng.head.min.js` blocks the browser rendering (since it is at the top of the page, the page is blank).
Ideally, we should find a way to remove this file.

* We can add as many parameters in `searxng.min.js`.
* If some CSS trick are found, we can remove `searxng.head.min.js`.

For example: 
* `<noscript><link rel="stylesheet" href="searxng.nojs.min.css"></noscript>`
* About touch screen: https://ferie.medium.com/detect-a-touch-device-with-only-css-9f8e30fa1134

* `make run`
* check the UI in a desktop browser.
* check the UI in a mobile browser with a  touch screen.

## Related issues

Related to https://github.com/searxng/searxng/pull/562


searx/templates/simple/base.html
```diff
@@ -19,14 +19,7 @@
   {% endif %}
   {% block styles %}{% endblock %}
   <!--[if gte IE 9]>-->
```
return42: I assume we can drop the IE9 quirks `<!--[if gte IE 9]>-->`  .. / Microsoft has discouraged the use of Internet Explorer as a default browser.

return42:
> The parsing and evaluation of `searxng.head.min.js` blocks the browser rendering (since it is at the top of the page, the page is blank).

Maybe a dump question; just for my knowledge, how can I reproduce it / how can I measure the _blank time_?

dalf:
To test I use:
* mobile rendering in a browser with an HTTPS + http2 server in front of the Flask application.
* The network tab to throttle the connection to "Slow 3G"
* Lighthouse from Chrome (in the devtools)

> Maybe a dump question; just for my knowledge, how can I reproduce it / how can I measure the blank time?

I'm sorry, "Blank time" is misleading. 
What happens on mobile:
* the browser loads the CSS
* then the HTML page and render the page according to  the CSS.

If the `js` and `touch` CSS classes are added at the end of the page, the browser will have to redraw the whole page.
This was the case when `searx.head.min.js` was created (2 years ago, https://github.com/searx/searx/pull/1647 ), it seems different now because I don't see this side effect anymore at with Chrome with a "slow 3G" connection.

But Lighthouse reports a [Cumulative Layout Shift](https://web.dev/cls/) of 0 in the master branch.  When `<script src="{{ url_for('static', filename='js/searxng.head.min.js') }}"></script>` is moved just above `<script src="{{ url_for('static', filename='js/searxng.min.js') }}"...`, the Cumulative Layout Shift is between 200 and 500ms.

I think `searx.head.min.js` still make sense for this reason (and because I'm not sure how a browser avoid a reflow when the CSS classes are changed at the end of the HTML page).

return42:
@dalf I rebased on current master, I added a commit on top and I build the static files

Thanks a lot for the explanation about debugging Cumulative Layout Shift.

I made some test (throttled to GPRS) .. I got best experience when (eecc36aa):

a. `searxng.head.min.js` is requested (loaded) as soon as possible
b. `searxng.min.js` is requested (loaded) right after CSS
c. other stuff like favicon should be requested (loaded) at least.

My experience: The order in the HTML determines the order of the requests, but the load time might be different (depends on the size of the file). Since `searxng.head.min.js` is very small, it is fully loaded before CSS is loaded --> add & remove CSS classes will be done before CSS is fully loaded .. what gives me the best UI experience.

return42 (approved):
From my side this PR is ready to merge

dalf:
I'm going to close this PR: I can't reproduce the number you get, or when I can the other metrics get worse.
The initial change https://github.com/searxng/searxng/pull/565/commits/059f4157faaf4796964a1c71556f948fdfe3a698 is not that important.
We can just keep in mind that this file exists only because we need to detect touch screen and JS, nothing more.

PR #538, opened 2021-11-23
SearXNG JavaScript Style Guide

return42:
## What does this PR do?

Define a SearXNG JavaScript Style Guide

The SearXNG JavaScript Style Guide is a set of eslint rules [1] based on the
eslint:recommended [2] rules.

    "extends": "eslint:recommended"

The additional rules adopted from the JavaScript Standard Style [3][4].

[1] https://eslint.org/docs/rules/
[2] https://github.com/kunalgolani/eslint-config/tree/master/packages/recommended
[3] https://standardjs.com/rules.html#javascript-standard-style
[4] https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json

The first three commits do fix small issues in preparation

## Why is this change important?

consistent code quality

PR #566, opened 2021-11-30
[mod] simple theme: smaller build

dalf:
## What does this PR do?

remove:
* searx/static/themes/simple/js/searxng.head.js
* searx/static/themes/simple/js/searxng.js
* searx/static/themes/simple/css/searxng-rtl.css
* searx/static/themes/simple/css/searxng.css

These files are never used.
About the js files: the .map files references the sources instead of the concated version.

Remove some useless files from the built files.

Draft because:
* #538 goes first.
* the same thing can done for the Oscar theme.

* `make run`, check Javascript is working (clear a search, image detail, map)

## Author's checklist

Except `searx/static/themes/simple/css/images`, all files are now built from the sources.

## Related issues

Related to #141


return42:
> These files are never used.

I pushed 82125862 to remove this files from the static folder.

> the .map files references the sources instead of the concated version.

this works, as long as we have the `src` folder underneath the /static path

    "sources":["../src/js/main/00_toolkit.js","../src/js/main/keyboard.js", ...

To get #141 done, we need to copy these files to `searx/static/themes/simple/js/`


> Draft because:
> * #538 goes first.

I don't think so, cleaning up the source files is IMO not related to a StyleGuide 

> * the same thing can done for the Oscar theme.

No need for, don't lets waste time with this theme.  We should make simple theme the default theme, then we can remove oscar theme and then we can assemble a build chain in `/src` that is *straight forward* and do no longer mess by the oscar theme.

With other words: oscar is in the way to have a clean build chain for client side stuff / making simple theme to the default (and removing oscar) should be prioritized.

searx/static/themes/simple/gruntfile.js
```diff
@@ -110,20 +93,15 @@ module.exports = function(grunt) {
       dist: {
         files: {
           'js/searxng.head.min.js': ['js/searxng.head.js'],
```
return42: `js/searxng.head.js` will no longer be created / compare task ` 'js/searxng.head.js': ['src/js/head/*.js'],` which has been removed in line 90 (task ` concat:`) / fixed in 82125862

utils/lib_static.sh
```diff
@@ -4,16 +4,16 @@
...
+    'searx/static/themes/oscar/css/*.css'
+    'searx/static/themes/oscar/js/*.js'
+    'searx/static/themes/oscar/src/generated/pygments-logicodev.less'
+    'searx/static/themes/oscar/src/generated/pygments-pointhi.less'
+    'searx/static/themes/simple/css/*.css'
+    'searx/static/themes/simple/js/*.js'
+    'searx/static/themes/simple/src/generated/pygments.less'
+    'searx/static/themes/simple/img'
+    'searx/templates/__common__/searxng-wordmark.min.svg'
+    'searx/templates/simple/icons.html'
```
return42: In commit 4704efa3 and commit ab454742 I removed the .map files from the repository.

The files are created by `themes.all` but marked as ignored in the .gitignore files in the css & js folders.

With we have the .map for locally debug purpose, but do not deploy them under the /static URL.  BTW: this minimize the commits from `make static.build.commit`.

return42 (changes requested):
Hi @dalf, this PR animated me to add various improvements on top of branch.  Please read through the commit messages and the comments from mine. Adopt what you think is good and drop what you don't think is good.

utils/lib_static.sh
dalf: How the browser will find the .map files?

return42: If you run (e.g.) `make themes.all`  the .map files are where they are .. nothing changed. But there is no need to commit the .map files .. hope I was clear / otherwise ask.

dalf: I wish to keep the .map files in the build because the day there is an error on an instance (not on localhost), it is easy to debug.

return42: > Also, the AGPL3 asks to provide the source code.

The source code is public here at github / no need to serve sources on every instance / BTW python sources are not served online in the instance.

> I wish to keep the .map files in the build because the day there is an error on an instance (not on localhost), it is easy to debug.

Really, did you ever had the need for? / I'm not familiar with, is it common to serve debug artefacts on a production server? 

If you want the map files back, we can leave it as is, but what me rumbles is the git-commit noise on the static folder caused by .map files. 

E.g have a look at the history of https://github.com/searxng/searxng/commits/master/searx/static/themes/oscar/css/logicodev.min.css.map

This .map file is changed every time `make static.build.commit` is called, even if nothing has changed in the _logicodev style_ since month // the symbols in this file are written in an unsorted manner, what changes the file, each time the file is newly generated by the grunt task.

dalf: > The source code is public here at github / no need to serve sources on every instance

Few years ago, some public instances added some tracking. That's why there is an HTML grade on searx.space. The .map files allows to evaluate if a change is harmful or not for the privacy. It doesn't provide a clear answer, but it's help. In addition if the .map files are removed, may be there is a catch.

Look at searx.space, search for `Cjs` on the HTML grade filter, some instances don't provide the source code (= the hashes of the downloaded files don't match the hahes of the git repository).

It is matter of transparency IMO.

> Really, did you ever had the need for? / I'm not familiar with, is it common to serve debug artefacts on a production server?

To understand the CSP issues. Example on your instance to make sure what is going on about the autocompletion & the CSP headers.

---

The topic of this PR is to remove 4 unused files (2 js, 2 css), nothing more.

searx/static/themes/simple/gruntfile.js
dalf: Thank you

utils/lib_static.sh
return42: > It is matter of transparency IMO.

OK I understand

> The topic of this PR is to remove 4 unused files (2 js, 2 css), nothing more.

OK, will remove from the branch, what is not a part of this topic .. done.

return42:
@dalf is it OK, when we merge this PR? .. I have some patches in the loop where the merge of this PR is needed for.

PR #573, opened 2021-12-01
[fix] lazy loading of <img> tags

return42:
## What does this PR do?

[fix] lazy loading of  tags

Images should include dimension attributes.  Without `width` and `height` specified, image dimensions are 0×0 pixels at first.  ...  In this case the browser determines that all of them are visible to the user and decides to load everything [1].

In CSS the `width` is set to a value and the `height` is _unset_ to scale the image proportional in both dimensions.

[1] https://web.dev/browser-level-image-lazy-loading/#images-should-include-dimension-attributes
[2] https://caniuse.com/loading-lazy-attr

save bandwith

In the browser have a look at the network activity while scrolling image results / like it is shown in this video https://storage.googleapis.com/web-dev-assets/native-lazy-loading/lazyload.webm


searx/templates/simple/result_templates/images.html
```diff
@@ -11,7 +11,7 @@
                         {%- if result.thumbnail_src -%}
                                 <img src="" data-src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}">
                         {%- else -%}
-                                <img src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}" loading="lazy">
+                                <img src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}" loading="lazy" width="200" height="200">
```
dalf, line 14: The detail is hidden by default, so the width & height can be skipped here.

dalf (commented):
As far I understand the width & height are here to take some space the time the image loaded, so other images are bellow or on the side instead of all images stacked together ; then these values are overridden by the CSS. 

It works well.

return42:
> As far I understand the width & height are here to take some space the time the image loaded, 

Not the time the image is loaded.  .. See my initial comment of this PR

_"""Without `width` and `height` specified, image dimensions are 0×0 pixels at first.  ...  In this case the browser determines that all of them are visible to the user and decides to load everything"""_

For lazy loading, the Browser has to calculate which images are fit into the current view-port. Since the browser can't now the sizes of an image before the browser loads the image, the browser assume the size of each image is 0x0 .. when the size of each image is zero, the browser calculates, that all images are fit into the view-port.

To give the browser a size before the image is loaded, we need to set `width` and `height` in the HTML  tag.

The CSS "unset" of the height comes into account, when the (then loaded) image is placed in the view-port. 

Very complicated, my english is to bad for a better description, read https://web.dev/browser-level-image-lazy-loading/#images-should-include-dimension-attributes it is a very good article.

To sum up: if you set `loading="lazy"` you have to set an initial `width` and `height` / otherwise all images are loaded immediately and not lazy.

searx/templates/simple/result_templates/images.html
return42: Haven't tested the behavior of initial hidden images & lazy loading, but I recommend to set an initial `width` and `height` whenever `loading="lazy"` is set .. see https://github.com/searxng/searxng/pull/573#issuecomment-985721745

return42:
@dalf I rebased to solve the conflicts and I left two comments .. If you think PR is ready to merge, please merge .. otherwise ask for modifications / thanks!

searx/templates/simple/result_templates/images.html
dalf: After some digging:
* even without `loading="lazy"`, the browser won't the image (at least FF and Chrome).
* the block https://github.com/searxng/searxng/blob/5307fd841d11d0bead4ef07147b46e584b3b1fe4/searx/templates/simple/result_templates/images.html#L10-L16 can be replaced by
```jinja2
                <a class="result-images-source" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">
                        <img src="data:," data-src="{{ image_proxify(result.img_src) }}" alt="{{ result.title|striptags }}">
                </a>{{- "" -}}
```

Note: the empty src is replaced by `data:,` to avoid additional HTTP request (it is safety net here: it actually doesn't happen because [the src attribute is set before the image is shown](https://github.com/searxng/searxng/blob/5307fd841d11d0bead4ef07147b46e584b3b1fe4/searx/static/themes/simple/src/js/main/results.js#L57-L65)):
* https://humanwhocodes.com/blog/2009/11/30/empty-image-src-can-destroy-your-site/
* https://stackoverflow.com/questions/5775469/whats-the-valid-way-to-include-an-image-with-no-src

---

The change can be either another PR, or in this one because it all about the lazy loading after all.

searx/templates/simple/result_templates/images.html
```diff
@@ -1,6 +1,6 @@
 <article class="result result-images {% if result['category'] %}category-{{ result['category'] }}{% endif %}">{{- "" -}}
         <a {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">{{- "" -}}
-                <img class="image_thumbnail" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} src="{% if result.thumbnail_src %}{{ image_proxify(result.thumbnail_src) }}{% else %}{{ image_proxify(result.img_src) }}{% endif %}" loading="lazy" alt="{{ result.title|striptags }}">{{- "" -}}
+                <img class="image_thumbnail" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} src="{% if result.thumbnail_src %}{{ image_proxify(result.thumbnail_src) }}{% else %}{{ image_proxify(result.img_src) }}{% endif %}" alt="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{- "" -}}
```
dalf, line 3: AFIU, the `height="200"` matches the 200 here:
https://github.com/searxng/searxng/blob/5307fd841d11d0bead4ef07147b46e584b3b1fe4/searx/static/themes/simple/src/js/main/results.js#L6

searx/templates/simple/macros.html
```diff
@@ -23,8 +23,8 @@
   <span class="url_o{{loop.index}}"><span class="url_i{{loop.index}}">{{- part -}}</span></span>
   {%- endfor %}
   {{- result_close_link() -}}
-  {%- if result.img_src %}{{ result_open_link(result.url) }}<img class="image" src="{{ image_proxify(result.img_src) }}" title="{{ result.title|striptags }}">{{ result_close_link() }}{% endif -%}
-  {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}">{{ result_close_link() }}{% endif -%}
+  {%- if result.img_src %}{{ result_open_link(result.url) }}<img class="image" src="{{ image_proxify(result.img_src) }}" title="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{ result_close_link() }}{% endif -%}
```
dalf, line 26: The actual values are `width="101" height="101"` (with CSS).

image

Suggestion: set the width & height to 101 to avoid reflow?

--
Sorry the screenshot not is English of the [Page Inspector](https://developer.mozilla.org/en-US/docs/Tools)

searx/templates/simple/macros.html
```diff
@@ -23,8 +23,8 @@
...
+  {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{ result_close_link() }}{% endif -%}
```
dalf, line 27: Same comment here, except the height is variable.
A compromise: `width="288" height="180"`

searx/templates/simple/result_templates/images.html
return42: That is correct. Since width is not know until image is loaded, I chose 200x200.

dalf: See my comments below: we can avoid random value since the CSS will set explicitly the width in rem value:
https://github.com/searxng/searxng/blob/9fc27088ef9cbbbc09b07a1b60cc78b469b1d0f3/searx/static/themes/simple/src/less/style.less#L265-L280

It is not always true for `.thumbnail`, for tablet & phone the width is changed:
https://github.com/searxng/searxng/blob/9fc27088ef9cbbbc09b07a1b60cc78b469b1d0f3/searx/static/themes/simple/src/less/style.less#L677-L679

but it will will the right width at in some cases (and, _not related to this PR_, don't look the web pages without CSS).

Otherwise, LGTM.

searx/templates/simple/macros.html
return42: > The actual values are `width="101" height="101"` (with CSS).
> Suggestion: set the width & height to 101 to avoid reflow?

100 or 100,001 or ... it is all wrong (reflow?) since `rem` is related to the font size (em) of the browser and the font size is different on all devices.
 
IMO setting img sizes in em is the issue https://github.com/searxng/searxng/pull/573#discussion_r767320596

Should we set a img size in points / like we have done in the images template?

searx/templates/simple/macros.html
return42: see my https://github.com/searxng/searxng/pull/573#discussion_r767429453

searx/templates/simple/result_templates/images.html
return42: > * even without `loading="lazy"`, the browser won't the image (at least FF and Chrome).
>
> * the block ... can be replaced by
> ...
> The change can be either another PR

I have a different opinion for two reasons:

1. I think the if clause in the template is more robust and better to read & understand
2. IMO it is not an issue to set lazy=loading (including with & height) in the template, no matter what CSS will hide or not.

return42:
@dalf I rebased this PR to solve conflicts. I left some comments about your remarks, but I did not changed anything right now.

I have some different opinions about your remarks / may be I am wrong.

If you see issues in this PR, please push your patches on top of this branch, otherwise I would suggest to merge this PR since it fixes a long standing issue (even if the image handling might not be perfect in one or the other manner).

mrpaulblack:
@return42 lazy loading the image category does not work for me. All the  tags with the image_thumbnail class get loaded even tho they are way outside my viewport with phone layout... It works great with all the other categories tho. For example video works correctly...

Maybe because the height is not unset for images in the image category? Also is the style attribute that got added in the jinja2 template to set initial width and height going to break CSP, since inline CSS is normally not allowed...?

I am also ok with setting the value without `em`; I am pretty sure it is common to use px values to set the width and height of a (non vector) image on websites... And IMO it works good in this PR :+1:

return42:
> Also is the style attribute that got added in the jinja2 template to set initial width and height going to break CSP, since inline CSS is normally not allowed...?

I can't reproduce a CSP issue, I tested with:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 305b254c..cc1bc304 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -16,7 +16,7 @@ search:
   # Existing autocomplete backends: "dbpedia", "duckduckgo", "google",
   # "startpage", "swisscows", "qwant", "wikipedia" - leave blank to turn it off
   # by default.
-  autocomplete: ""
+  autocomplete: "google"
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: ""
@@ -48,7 +48,7 @@ server:
 
   secret_key: "ultrasecretkey"  # change this!
   # Proxying image results through searx
-  image_proxy: false
+  image_proxy: true
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
   # POST queries are more secure as they don't show up in history but may cause
@@ -60,6 +60,7 @@ server:
     X-Download-Options: noopen
     X-Robots-Tag: noindex, nofollow
     Referrer-Policy: no-referrer
+    Content-Security-Policy: "default-src 'self'; img-src 'self' data: ;"
 
 ui:
   # Custom static path - leave it blank if you didn't change
@@ -70,7 +71,7 @@ ui:
   # it decreases the privacy, since the browser can records the page titles.
   query_in_title: false
   # ui theme
-  default_theme: oscar
+  default_theme: simple
   # Default interface locale - leave blank to detect from browser information or
   # use codes from the 'locales' config section
   default_locale: ""
```

AFAIK `width` & `hight` attributes of the `` are not CSS / they are properties of the img tag, which can be overwritten by CSS `width` & `hight`.

return42:
> @return42 lazy loading the image category does not work for me. All the tags with the image_thumbnail class get loaded ...
> Maybe because the height is not unset for images in the image category?

Hm, why do you think `height` is not set for `image_thumbnail`? `width` & `hight` are set here (scroll to the right side):

https://github.com/searxng/searxng/pull/573/commits/3598560f24e1f16d32b2b2b0c1a6c7f732e62891#diff-a081f1bb46dcae5f720b215dae26480ab77b3eddb7d0d0ec00976d8f587c0443

> even tho they are way outside my viewport with phone layout..

About *viewport & lazy loading* : The browser loads some more images than those are in the viewport, but the browser does not load all images.  If you initial open a image search, there are some images loaded, but not all .. scroll below in the images results and you will see that more and more images are requested while you are scrolling.

Don't be irritated by the fact that the browser first loads more images than only those that fit into the viewport .. scroll down and you will see that lazy-load comes into account.

Can you please validate? .. and give a short feedback .. may be I have misunderstood you.

return42:
@mrpaulblack I rebased this PR to solve conflicts. I left some comments about your remarks, but I did not changed anything right now.

I have some different opinions about your remarks / may be I am wrong.

If you see issues in this PR, please push your patches on top of this branch, otherwise I would suggest to merge this PR since it fixes a long standing issue (even if the image handling might not be perfect in one or the other manner).

mrpaulblack:
Ok yeah I misunderstood the CSP thing since its not a style attribute, but rather attributes of the img tag :+1: Also height is set correctly for all images... Lazy load also works correctly for me in image category idk probably tested it wrong before @dalf I am also ok with the 200px height and width initially, since it would be a pain to set the correct size in jinja template depending on use case and I think that approximate is ok. Feel free to merge, if you agree :+1: LGTM

PR #584, opened 2021-12-05
[mod] simple-theme: remove vendor CSS

return42:
## What does this PR do?

[mod] simple-theme: remove vendor CSS

**flexbox**: Browser support for flexbox is excellent, and the majority of browsers do not need a prefix at this point. Safari was the last of the major browsers to   remove prefixes, with the release of Safari 9 in 2015. [1]

**user-select**: Vendor prefix of user-select is not needed, see 'Browser compatibility' [2].

[1] https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Backwards_Compatibility_of_Flexbox#status_in_browsers
[2] https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility


mrpaulblack:
tbh they should not be mixins if its only one generic line of css code IMO. Can you remove the mixins and replace every mixin call in the less files with just the specific line of css? For example replace `.flexbox;` with `display: flex;` and so on...

Apart from that removing the vendor prefixes LGTM :+1:

return42:
@mrpaulblack sorry for long delay ..

> tbh they should not be mixins if its only one generic line of css code I

that true .. I removed the mixins / good to know: most of them hadn't been used :)

Hint: additional I pushed `[mod] simple theme: remove vendor prefix of width: max-content` on this branch.

return42:
FYI: https://github.com/searxng/searxng/pull/584#commits-pushed-459f33a nothing changed / just rebased to solve conflicts

dalf:
Thank you!

PR #583, opened 2021-12-05
[fix] simple theme: remove width  45% from language and time filters

return42:
## What does this PR do?

[fix] simple theme: remove width  45% from language and time filters

All three filters (`language`, `time_range` and `safesearch`) are rendered in one line.  A size of 45% for `language` and `time_range` left only 10% for the `safesearch` filter.  Solution: drop with from `language` and `time_range`.

## How to test this PR locally?

On a small device / **before**:

grafik

and **after** PR applied:

grafik

## Author's checklist

See also https://github.com/searxng/searxng/pull/585

PR #585, opened 2021-12-05
simple theme: normalize search header & categories

return42:
## What does this PR do?

Improve search header & categories on smaller devices ..

**before**

grafik

**after**

grafik

In the result page scale page size from desktop size down to very (very) small device and the layout should flow without any hickup.

## Author's checklist

In combination with https://github.com/searxng/searxng/pull/583 the layout on smaller devices has an better flow.


mrpaulblack:
The categories work correctly on mobile when enabling touch input and reloading the page... (The category icons are missing on purpose on phone since they take up to much space IMO)... The watermark gets already removed on phones smaller width 20rem and should still be there on bigger width devices IMO... The burger menu on the right gets only displayed when the search form `#q` input filed is empty...

I think its good to make the categories look better on small screen devices without touch input support, but I am not sure if its a good idea to change the other stuff that we just recently fixed in other PRs around tbh...

[EDIT] also removing the vendor prefixes is great!

return42:
@mrpaulblack thanks for your review and your helping comments / sorry if I removed to much.  I think I have to look deeper, this was my first try. I will continue my work in a week or two .. in the meantime I converted this PR draft.

return42:
I close this PR / normalizing was I dumb idea of mine, sorry for the noise. Instead I:

- fixed category layout in https://github.com/searxng/searxng/pull/603 and ..
- removed the vendor prefixes in https://github.com/searxng/searxng/pull/584

Issue #589, opened 2021-12-07
Support httpx 0.21.1 and httpcore 0.14.3 


HLFH:
httpcore 0.14 got a [redesign](https://github.com/encode/httpcore/pull/420).
See [this as well](https://github.com/encode/httpx/blob/716749e3fd3c99cb47cc20b96def280dcb3c8d16/CHANGELOG.md#0210-15th-november-2021).

After updating them, I am currently getting the issue:
```
déc. 07 15:37:46 arch-server uwsgi[338276]: AttributeError: module 'httpcore._utils' has no attribute 'URL'
déc. 07 15:37:46 arch-server uwsgi[338276]: unable to load app 0 (mountpoint='') (callable not found or import error)
```

In https://github.com/searxng/searxng/blob/master/searx/network/client.py, this would not work:
```
async def close_connections_for_url(
    connection_pool: httpcore.AsyncConnectionPool, url: httpcore._utils.URL
):
```

dalf:
According to the last commits I've seen on httpx, `close_connections_for_url` can be removed (connections are now closed even in case of error).

I think AsyncProxyTransportFixed and AsyncHTTPTransportFixed can be more simple (untested):

```python
class AsyncProxyTransportFixed(AsyncProxyTransport):
    async def handle_async_request(
        self, method, url, headers=None, stream=None, extensions=None
    ):
		try:
			return await super().handle_async_request(
				method, url, headers=headers, stream=stream, extensions=extensions
			)
		except (ProxyConnectionError, ProxyTimeoutError, ProxyError) as e:
			raise httpx.ProxyError from e
		except OSError as e:
			# socket.gaierror when DNS resolution fails
			raise httpx.ConnectError from e


class AsyncHTTPTransportFixed(httpx.AsyncHTTPTransport):
    async def handle_async_request(
        self, method, url, headers=None, stream=None, extensions=None
    ):
		try:
			return await super().handle_async_request(
				method, url, headers=headers, stream=stream, extensions=extensions
			)
		except OSError as e:
			# socket.gaierror when DNS resolution fails
			raise httpx.ConnectError from e
```

---

I will give a try by the end of the week.

HLFH:
Closing in favour of https://github.com/searxng/searxng/pull/596.

PR #596, opened 2021-12-11
Upgrade httpx

dalf:
## What does this PR do?

Upgrade to httpx 0.21.2

Change in searx.network.client:
* remove `close_connections_for_url` : httpcore closes the connections now.
* keep `AsyncProxyTransportFixed` to map the exception from httpx_socks to httpx exception WITH the "request" object.
* remove `AsyncHTTPTransportFixed`: the retry mecanism has moved to searx.network.network to decrease the number of access to the private API of httpx / httpcore.

`httpx.RemoteProtocolError` happens when a server closes an HTTP/2 connection. In the master branch, `AsyncHTTPTransportFixed` closes all connections. In this PR, the `httpx.AsyncClient` is closed. It makes no difference except when different proxies are specified, for example:

```yaml
proxies:
  https://bing.com/: socks5h://127.0.0.1:1337
  all://: socks5h://127.0.0.1:1338
```

With this configuration, and with an `httpx.RemoteProtocolError` exception coming from the proxy on the port 1337:
* in the master branch: connections using the proxy on the port 1337 are closed. Connections using the proxy on the port 1338 are untouched.
* in this PR: connections using the proxy on ports 1337 and 1338 are closed.

@unixfox @tiekoetter @paulgio I don't know you use this configuration?

An alternative configuration not impacted by this PR:
```
outgoing:
	proxies:
		all://: socks://127.0.0.1:1337	

engines:
  - name: bing
    engine: bing
    shortcut: bi
	proxies:
		all://: socks://127.0.0.1:1338
```

---

~This is WIP, todo list~:
  * [x] use a released httpcore version
  * [x] check for memory leak over a week
  * [x] check for error log over a week:
     * in the master branch, there is a close connections + retry mechanism in a case of specific error.
       this a workaround to fix some of the httpcore issues
     * the latest httpcore commit seems to have fix most of them
     * the change in this commit remove the "close connections + retry"
     * an error log over a long time can confirm that httpcore has been fixed

* Remove a bunch of hack to fix an old httpcore version
* With this change, httpx can be upgraded easily (= just change requirements.txt)

A lot of changes have been made to httpx, httpcore and anyio since the last upgrade, a full check is require over a long time.

what can be done locally:
* check the error for various HTTP error (domain not found)
* check that HTTP proxy works
* check that SOCKS proxy works
* check that HTTP2 connection works
* check this scenario: 
    * make a HTTP2 request
    * wait for the server to disconnect
    * maka another HTTP2 request

## Related issues

Related to https://github.com/searxng/searxng/issues/589

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,9 +7,9 @@ lxml==4.7.1
 pygments==2.11.0
 python-dateutil==2.8.2
 pyyaml==6.0
-httpx[http2]==0.19.0
+httpx[http2]==0.21.2
 Brotli==1.0.9
 uvloop==0.16.0
-httpx-socks[asyncio]==0.4.1
+httpx-socks[asyncio]==0.7.2
 langdetect==1.0.9
 setproctitle==1.2.2
--- a/searx/network/client.py
+++ b/searx/network/client.py
@@ -6,8 +6,6 @@ import asyncio
 import logging
 import threading
 
-import anyio
-import httpcore
 import httpx
 from httpx_socks import AsyncProxyTransport
 from python_socks import parse_proxy_url, ProxyConnectionError, ProxyTimeoutError, ProxyError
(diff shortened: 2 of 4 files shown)


mrpaulblack:
I tested this on my instance over a week:
image

Memory Increased over the span of 3 days by 13megabytes which is IMO margin of error.

Also with this PR memory usage dropped a little and the performance is the same or even slightly better in general IMO

dalf:
The only missing part is a new release of httpcore, after that the PR is ready to merge.

unixfox:
I don't use proxies anymore, but I do I've plans to add some in the future but not for the default engines, for some engines that block public instances like qwant.

mrpaulblack:
@unixfox I have actually qwant on my instance enabled, because they never IP block it tbh (and on top of this qwant is really fast; around 400ms per request)...

HLFH:
Besides the missing httpcore release, the branches upgrade-httpx and/or test-uwsgi are ready to merge.

PR #609, opened 2021-12-17
[fix] engine bing-news: replace the http:// by https://

return42:
## What does this PR do?

[fix] engine bing-news: replace the http:// by https://

BTW: add bing_news to the pylint process

## Why is this change important?

Images from bing-news are no longer loaded from http://

## How to test this PR locally?

    !bin london

PR #610, opened 2021-12-17
[fix] simple theme: hide the image if img load fails

return42:
## What does this PR do?

~[fix] simple theme: show `img_load_error.svg` if img load fails~

[fix] simple theme: hide the image if img load fails

Add event listener to query selector::

    '#urls img.image'

see https://github.com/searxng/searxng/pull/610#issuecomment-997640132


dalf:
Would it better to just hide the image?

return42:
> Would it better to just hide the image?

I don't know what is better? ... since there is an image URL, I think there should be _placeholder_ images, but this is my very personal view.

BTW: the screenshot from above is not representative, normally you will see very rare the placeholder image.  The screenshots shows, that the bing-news engine has some issues .. we fixed in https://github.com/searxng/searxng/pull/609

dalf:
> I don't know what is better? 

From the user point of view, I think it is better to hide the image: `img_load_error.svg` is helplful in the image category because it still allows to select the image. IMO, in the news category, the fact there is a missing image won't help to choose the links.

From a developer point of view, the place holder is signal that may be the engine needs to be updated (at least give a look). The browser console should show the same information too, but it requires some additional steps.

return42:
>  but it requires some additional steps.

what do you expect from this PR .. should I disable the image if it can't be loaded .. does [Cumulative Layout Shift (CLS)](https://web.dev/cls/) need any attention here?

return42:
ping @dalf

dalf:
> what do you expect from this PR

Hide the images that can't be loaded.

> does Cumulative Layout Shift (CLS) need any attention here?

as you said, this doesn't happen often.

return42:
> Hide the images that can't be loaded.

OK .. done!

dalf:
Thank you

PR #611, opened 2021-12-18
[fix] bing engine: fix paging support, show inital page.

return42:
## What does this PR do?
    
    [fix] bing engine: fix paging support, show inital page.
    - Split search-term in one for initial query and one for following queries.
    - Set some headers in HTTP requests, bing needs for paging support.
    - IMO `//div[@class="sa_cc"]` does no longer match in a bing response.

    [pylint] Bing (Web) engine

Follow up queries for the pages needed to be fixed / results from initial bing response are not shown.

    !bing london

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Bing (Web)
+# lint: pylint
+"""Bing (Web)
 """
 
 import re
@@ -8,7 +8,6 @@ from urllib.parse import urlencode
 from lxml import html
 from searx.utils import eval_xpath, extract_text, match_language
 
-# about
 about = {
     "website": 'https://www.bing.com',
     "wikidata_id": 'Q182496',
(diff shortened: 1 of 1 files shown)


searx/engines/bing.py
```diff
@@ -21,78 +22,113 @@
...
 
     search_path = search_string.format(
-        query=urlencode({'q': query}),
-        offset=offset)
+        query = urlencode({'q': query}),
+        offset = offset)
 
-    params['url'] = base_url + search_path
+    if offset > 1:
+        referer = base_url + inital_query.format(query = urlencode({'q': query}))
```
dalf, line 69: AFIU, it works only for the second page.

dalf (approved):
LGTM, thank you for the update.

Side note: I realize that an engine can't tell there is no more page.

searx/engines/bing.py
return42: right! But please be aware ... 

1. the template for the request URL is selected in lines 49-50.
2. here the "referer" URL is build and added to the HTPP headers

Bing wants to have the "reverer" URL in the HTTP header of the second and following pages.

The behaviour of the initial bing response is a little bit quirky .. may be there is space left to improve, but with this implementation I got the (4 or 5) results from the very first page of a bing query.  I missed this entries in the old implementation.

PR #483, opened 2021-11-05
[fix] Dockerfile image

dalf:
## What does this PR do?

ignore existing wheels, rebuild all of them
fix an issue the lxml wheel for the muslc

The build time may increase.

## Why is this change important?

Fix #482

## How to test this PR locally?

* `make docker`
* check the build image name, may be this is not `searxng/searxng` (for me `dalf/searxng`).
* `docker run --rm              -d -p ${PORT}:8080              -e "BASE_URL=http://localhost:$PORT/"              -e "INSTANCE_NAME=my-instance"              searxng/searxng` 

## Author's checklist

Alternative (?): use the Debian slim image.

## Related issues

Close #482

Issue #482, opened 2021-11-05
docker image broken


dalf:
The last docker image doesn't work:
* the web page shows `Internal Server Error`
* `/var/log/uwsgi/uwsgi.ini` :
```
Traceback (most recent call last):
  File "/usr/local/searxng/searx/webapp.py", line 61, in <module>
    from searx.engines import (
  File "/usr/local/searxng/searx/engines/__init__.py", line 22, in <module>
    from searx.utils import load_module, match_language, gen_useragent
  File "/usr/local/searxng/searx/utils.py", line 11, in <module>
    from lxml import html
  File "/usr/lib/python3.9/site-packages/lxml/html/__init__.py", line 53, in <module>
    from .. import etree
ImportError: cannot import name 'etree' from 'lxml' (/usr/lib/python3.9/site-packages/lxml/__init__.py)
unable to load app 0 (mountpoint='') (callable not found or import error)
```

How to test:
* https://searxng.github.io/searxng/admin/installation-docker.html#searxng-searxng
* (or https://github.com/searxng/searxng-docker )

I've tried:
1. to remove the latest-build-cache tag from https://hub.docker.com/r/searxng/searxng/tags
2. trigger again https://github.com/searxng/searxng/actions/runs/1426096185

Same result.
Perhaps it is related to https://github.com/searxng/searxng/pull/477

ACodingGenie:
Can confirm. Same issue here for a few hours :'(

mrpaulblack:
yeah same for my latest staging: https://staging.paulgo.io (pipeline: https://paulgo.dev/paulgoio/searxng/-/jobs/744); I am building without cache in my gitlab ci/cd so its def. not a problem with github pipeline...

dalf:
Fixed, but it takes 3 hours to build the docker image instead of 1 hour (for amd64, arm64 and arm/v7):
https://github.com/searxng/searxng/blob/1ccf296a558260e46f5b7a294807bd74285e70fc/manage#L436

It happens on when the requirements*.txt files change.

The limit on GitHub is 6 hours: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

PR #612, opened 2021-12-18
[mod] Dockerfile: use binary from pypi

dalf:
## What does this PR do?

lxml doesn't raise any exception

When a dependency change, the GitHub workflow runs for 3 hours. The limit is 6 hours.
This PR restore the usage of the wheels, and the build time should come back to about 1 hour.

* `make docker`
* run the docker image

## Related issues

* #482
* #483

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,7 @@ RUN apk upgrade --no-cache \
     uwsgi-python3 \
     brotli \
  && pip3 install --upgrade pip wheel setuptools \
- && pip3 install --no-cache  --no-binary :all: -r requirements.txt \
+ && pip3 install --no-cache -r requirements.txt \
  && apk del build-dependencies \
  && rm -rf /root/.cache
 


Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN apk upgrade --no-cache \
     uwsgi-python3 \
     brotli \
  && pip3 install --upgrade pip wheel setuptools \
- && pip3 install --no-cache  --no-binary :all: -r requirements.txt \
+ && pip3 install --no-cache -r requirements.txt \
```
return42, line 51: I think we can remove the `--no-cache`, BTW here is what we do in the SearXNG installation script ... additionally we install the `pyymal` module and instead of installing `requirements.txt` we install the SearXNG repository into SEARX_PYENV (`${SERVICE_HOME}/searx-pyenv`) ..

**for installation**

https://github.com/searxng/searxng/blob/4da1e0026c930686e8c5b0545b2c97f8b834289b/utils/searx.sh#L722-L729

**for update**

https://github.com/searxng/searxng/blob/4da1e0026c930686e8c5b0545b2c97f8b834289b/utils/searx.sh#L410-L414

return42 (commented):
I can't test it, but it LGTM .. I left a comment in the source code ...

Dockerfile
dalf: AFIK `--no-cache` will cache the dependencies in `~/.cache/pip`, but the docker image will never this directory again (since the dependencies are not updated in one docker image)

return42 (approved):

PR #614, opened 2021-12-18
[pylint] Bing (Videos) engine

return42:
## What does this PR do?

- Add Bing (Videos) engine to pylint process.
- Fix remarks from pylint and remove obsolete try/except block

Code Review / no functional change!

    make test.pylint

### DIFF
--- a/searx/engines/bing_videos.py
+++ b/searx/engines/bing_videos.py
@@ -1,11 +1,13 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Bing (Videos)
+# lint: pylint
+"""Bing (Videos)
+
 """
 
 from json import loads
-from lxml import html
 from urllib.parse import urlencode
+
+from lxml import html
 
 from searx.utils import match_language
 from searx.engines.bing import language_aliases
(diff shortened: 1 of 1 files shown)


searx/engines/bing_videos.py
```diff
@@ -28,74 +32,87 @@
...
-    params['cookies']['SRCHHPGUSR'] = \
+    params['cookies']['SRCHHPGUSR'] = (
         'ADLT=' + safesearch_types.get(params['safesearch'], 'DEMOTE')
+    )
 
     # language cookie
-    language = match_language(params['language'], supported_languages, language_aliases).lower()
+    language = match_language(
+        params['language'], supported_languages, language_aliases
+    ).lower()
```
dalf: I don't understand why the function is split into 3 lines.

return42: Just my subjective feeling of _readability_ / don't let us look to close, instead lets enforce https://github.com/searxng/searxng/pull/619 where we can define our Code-Style commitment.

return42:
@dalf I rebased this PR (to get code style from black / https://github.com/searxng/searxng/pull/614#discussion_r772193521).

searx/engines/bing_videos.py
```diff
@@ -82,22 +84,18 @@ def response(resp):
     dom = html.fromstring(resp.text)
 
     for result in dom.xpath('//div[@class="dg_u"]'):
-        try:
```
dalf, line 85: it is more than pylint, but why not.

Note that should be an issue: create a context manage named "no_exception":
```
for result in dom.xpath('//div[@class="dg_u"]'):
    with no_exception():
        ...
```

-->

exception are hidden EXCEPT in debug mode

dalf (approved):

searx/engines/bing_videos.py
return42: > it is more than pylint, but why not.

more or less, it is pylint ...  [W0702 (bare-except)](http://pylint-messages.wikidot.com/messages:w0702)

> exception are hidden EXCEPT in debug mode

The problem of this try block is to catch all exception classes, what is an antipattern ...

>  create a context manage named "no_exception":

no please not .. it is a really bad practice to catch all issues / issues in the engines are handled by the caller of the engine.

I tested the bing engines and I can't find an exception, so for what is a try/except block needed? ..

PR #617, opened 2021-12-18
[fix] theme: build `img_load_error.svg` as SVG instead of PNG

mrpaulblack:
## What does this PR do?

Fix build so exported `img_load_error.svg` file is actually a svg in the build and not a png with .svg ending.

Fix bug in Simple theme. (See current prod: http://searx.be/static/themes/simple/img/img_load_error.svg)

`make run`

got to `http://127.0.0.1:8888/static/themes/simple/img/img_load_error.svg` to check that an svg gets actually returned and test that it gets rendered correctly in image category when an image is missing.

## Author's checklist

/


return42 (approved):
Oops, was my fail .. thanks for your fix!

PR #618, opened 2021-12-19
[enh] limiter plugin

dalf:
## What does this PR do?

The limiter of this PR can replace filtron:
* it rates limite the number of request per IP and per (IP, User-Agent)
* it blocks some bots

It is based on the same rules than https://github.com/searxng/searxng-docker

Redis stores the aggregated values.
However, data stored in Redis never contains the IP addresses, only HMAC using the secret_key

Simplify the stack.
With this PR, it is possible to remove Morty & Filtron.

* start redis on the default port (localhost:6379)
* `make run`
* check everything works as expected
* in settings.yml, enable the limiter `server.limiter: true`
* `curl -vvvv http://localhost:8888/search?q=test` should be blocked
* check the image proxy is not blocked

## Author's checklist

The fact that the rules are Python code can make customization harder.

### DIFF
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -1,0 +1,129 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pyright: basic
+"""Some bot protection / rate limitation
+
+Enable the plugin in ``settings.yml``:
+
+- ``server.limiter: true``
+- ``redis.url: ...`` check the value, see :ref:`settings redis`
+"""
+
+import hmac
+import re
+from flask import request
+
+from searx.shared import redisdb
+
+name = "Request limiter"
+description = "Limit the number of request"
+default_on = False
+preference_section = 'service'
+
+
+re_bot = re.compile(
+    r'('
+    + r'[Cc][Uu][Rr][Ll]|[wW]get|Scrapy|splash|JavaFX|FeedFetcher|python-requests|Go-http-client|Java|Jakarta|okhttp'
+    + r'|HttpClient|Jersey|Python|libwww-perl|Ruby|SynHttpClient|UniversalFeedParser|Googlebot|GoogleImageProxy'
+    + r'|bingbot|Baiduspider|yacybot|YandexMobileBot|YandexBot|Yahoo! Slurp|MJ12bot|AhrefsBot|archive.org_bot|msnbot'
+    + r'|MJ12bot|SeznamBot|linkdexbot|Netvibes|SMTBot|zgrab|James BOT|Sogou|Abonti|Pixray|Spinn3r|SemrushBot|Exabot'
+    + r'|ZmEu|BLEXBot|bitlybot'
+    + r')'
+)
+
+
+def is_accepted_request(inc_get_counter) -> bool:
+    # pylint: disable=too-many-return-statements
+    user_agent = request.headers.get('User-Agent', '')
+    x_forwarded_for = request.headers.get('X-Forwarded-For', '')
+
+    if request.path == '/image_proxy':
+        if re_bot.match(user_agent):
+            return False
+        return True
+
+    c = inc_get_counter(interval=25, keys=[b'IP limit, all paths', x_forwarded_for])
+    if c > 30:
+        return False
+
+    c = inc_get_counter(interval=60, keys=[b'useragent limit, all paths', x_forwarded_for, user_agent])
+    if c > 30:
+        return False
+
+    if request.path in ('/', '/search') and ('q' in request.args or 'q' in request.form):
+        if re_bot.match(user_agent):
+            return False
+
+        if 'Accept-Language' not in request.headers:
+            return False
+
+        if request.headers.get('Connection') == 'close':
+            return False
+
+        accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
+        if 'gzip' not in accept_encoding_list or 'deflate' not in accept_encoding_list:
+            return False
+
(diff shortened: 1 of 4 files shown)


return42:
FYI: PR was 56 commits behind / I rebased from f62213b to 098c675, no function change.

return42:
I will left some unsorted thoughts I have ....

Should we add this redis-feature as a plugin or as a internal SearXNG storage system?  I would prefer a _storage_ component in SearXNG which can be used by the _limiter_ and by new upcoming features.

The redis component in SearXNG should be open for various connectors, in this PR it is only IP (host & port) / I think we should use [redis-py](https://pypi.org/project/redis/) with we can use common [connector markups](https://github.com/redis/redis-py#connections) including [unix sockets](https://wiki.archlinux.org/title/Redis#Listen_on_socket).

Since I like to add a redis component in SearXNG, I would also like to implement a `./utils/redis.sh` script to install and setup a redis server.  The installation methods should be similar to what we implemented in the `./utils/morty.sh` and `./utils/filtron.sh` scripts (I am just looking forward for).

searx/plugins/limiter.py
```diff
@@ -0,0 +1,139 @@
...
+
+def is_accepted_request(get_aggregation_count):
+    # pylint: disable=too-many-return-statements
+    user_agent = request.headers.get('User-Agent')
+    x_forwarded_for = request.headers.get('X-Forwarded-For')
+
+    if request.path == '/image_proxy':
+        if re_bot.match(user_agent):
+            return False
+        return True
```
return42: > It is based on the same rules than https://github.com/searxng/searxng-docker

IMO we need a solution that is configurable, e.g. like filtron .. or little bit less generic ..

dalf: > IMO we need a solution that is configurable

I agree, that's the issue of this PR
But what is the configuration is a python script?
I mean allow to replace
`limiter: true` the settings.yml, 
by `limiter: the_path_of_a_python_script.py`

This Python script has to define one function `is_accepted_request` and receive a parameter the `get_aggregation_count` function.

In settings.yml:
* `limiter: false`: the limiter is disabled.
* `limiter: true`: the default configuration (the one recommended for public instances. New compare to filtron: we can upgrade the configuration).
* `limiter: /opt/searxng/my_limit_conf.py`: a custom configuration that replace the existing one. The default configuration can be used as a template.

return42:
@dalf I need some time to do my own research about redis (I want to try redis as a core component of SearXNG) .. is it OK when we toggle this PR into DRAFT in the meantime?

dalf:
Can we start simple with only a Redis connection object initialized in a different file than limiter.py? For example, it can support TCP or Unix socket. So we can merge this PR: if it works on a public instance, we can remove filtron and morty: it simplifies a lot of things. (note: morty is still usefull as a HTML proxy).

And in a second time, we can focus on Redis.

return42:
> Can we start simple with only a Redis connection object initialized in a different file than limiter.py?

Yes  .. but I need this week ...

>  For example, it can support TCP or Unix socket. 

Lets use [redis-py](https://pypi.org/project/redis/) 

> if it works on a public instance, we can remove filtron 

This needs a minimal installation of a redis instance / I am working on a script that is good enough for "single" SearXNG instance (simple & isolated redis service dedicated to SearXNG and connected via Unix socket) / see WIP https://github.com/return42/searxng/commits/lib_redis

dalf:
I wish to keep redis optional.
Currently, the plugin doesn't import redis except when enabled.

> Lets use redis-py

The plugin depends on redis-py.

> This needs a minimal installation of a redis instance

The docker image requires only a change in [docker-compose.yml](https://github.com/searxng/searxng-docker/blob/master/docker-compose.yaml). My intent is to try with docker on a public instance to validate the solution. Perhaps the test can be done in similar way to the upgrade_httpx PR, without a merge into the master branch.

return42:
> I wish to keep redis optional.

sorry if I was misleading / "core component" yes, but only mandatory for some features

> The plugin depends on redis-py.

Oops, sorry I had something old in mind.

> Perhaps the test can be done in similar way to the upgrade_httpx PR, without a merge into the master branch.

Yes, why not .. I will prepare my script for minimal redis (dedicated to SearXNG), with I can test it on my public instance (non docker).

return42:
> Can we start simple with only a Redis connection object initialized in a different file than limiter.py?

@dalf I implemented stuff to handle server & client .. could you have a look at / what do you think about:

- https://github.com/searxng/searxng/pull/686

mrpaulblack:
@return42 are you testing this PR on your instance?; Otherwise I can test this with my fork for a while...

dalf:
@mrpaulblack I've rebased / updated the PR with the master branch.

return42:
@dalf sorry, I have overseen this PR .. is it OK to move from DRAFT state to "ready for review"?

dalf:
> is it OK to move from DRAFT state to "ready for review"?

It haven't been tested in a real instance / in production.
I'm not sure about the redisdb usage.
The configuration is hard coded.

Except these points, it is ready.

return42:
> @return42 are you testing this PR on your instance?; Otherwise I can test this with my fork for a while..
> It haven't been tested in a real instance / in production.

This PR is an addition that does not conflict with an exiting instance and (for me) it is easier to test it in production when this PR is merged.  With this PR merged I can toggle between "filtron" and "limiter".

return42 (approved):
@dalf I pushed 279bfb67 on top with some suggestions from mine. You can squash in your commit what you like, others just drop.

dalf:
> You can squash in your commit what you like, others just drop.

Thank you, I've squashed the two commits.

PR #622, opened 2021-12-20
[docs] Switch from searx to SearXNG

### LINKED ISSUE #450: Document how to switch from searx to searxng

I want to install searxng (on debian). I already have a running instance of searx (not searxng).
The documentation says how to install from scratch, or how to upgrade searxng. Assuming a simple recipe is possible, a few lines explaining how to convert an existing searx instance to use searxng instead would be reassuring and helpful.

A step by step guide in the documentation of how to change a running searx installation into a running searxng installation

I could just try what might work, but that risks leaving me with a broken / botched installation.

**Additional context**

I guess there are plenty of others with the same request.

return42:
## What does this PR do?

Document a switch from searx to SearXNG.

    make docs

~See preview at: https://return42.github.io/searxng/admin/installation-switch2ng.html~

## Related issues

Closes: https://github.com/searxng/searxng/issues/450

### DIFF
--- a/docs/admin/index.rst
+++ b/docs/admin/index.rst
@@ -12,6 +12,7 @@ Administrator documentation
    installation-nginx
    installation-apache
    installation-docker
+   installation-switch2ng
    update-searxng
    engines/index
    api
--- a/docs/admin/installation-switch2ng.rst
+++ b/docs/admin/installation-switch2ng.rst
@@ -1,0 +1,75 @@
+.. _installation switch2ng:
+
+============================
+Switch from searx to SearXNG
+============================
+
+.. sidebar:: info
+
+   - :pull:`456`
+   - `A comment about rolling release <https://github.com/searxng/searxng/pull/446#issuecomment-954730358>`_
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+If you have a searx installation on your sever and want to switch to SearXNG,
+you need to uninstall searx first.  If you have an old searx docker installation
+replace your docker image / see :ref:`installation docker`.
+
+If your searx instance was installed *"Step by step"* or by the *"Installation
+scripts"*, you need to undo the installation procedure completely.  If you have
+morty & filtron installed, it is recommended to uninstall these services also.
+In case of scripts, to uninstall use the scripts from the origin you installed
+searx from.
+
+If you have removed the old searx installation, clone from SearXNG and and start
+with your installation procedure (e.g. :ref:`installation scripts`):
+
+.. code:: bash
+
+   $ cd ~/Downloads
+   $ git clone https://github.com/searxng/searxng.git searxng
+   $ cd searxng
+   $ ...
+
+``.config.sh``
+==============
+
+Please take into account; SearXNG has normalized ``.config.sh`` with
+``settings.yml`` and some of the environment settings has been removed from or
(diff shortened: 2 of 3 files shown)


ralphrmartin:
Thanks! Here are my observations
(1) It would help to explain explicitly how to uninstall searx - your "In case of scripts, to uninstall use the scripts from the origin you installed searx from." would be more helpfully be replaced by "Run the uninstall script found at ..."
(2) Most users would like to keep their settings, and the notes do not explicitly explain how to do this. While some notes are given under ".config.sh", they are of the forms of diffs rather than instructions.
What I had in mind is more a step-by-step set of instructions, rather than these "hints" - in other words a sequence of explicit steps for the user to carry out. E.g. move the file ... to new location ..., and replace all instances of "searx" by "searxng", You must also make the following further changes ...

return42:
> (1) It would help to explain explicitly how to uninstall searx 

I understand your concern, but I do not want to describe how an external software should be uninstalled / IMO this is a task of the software owner.

> (2) Most users would like to keep their settings, and the notes do not explicitly explain how to do this.

There is not much to say, all settings are done in the settings.yml / there was a conflict of .config.sh and settings.yml in the past, this was an issue of .config.sh and these values has been removed from .config.sh.
 
For more details, users can ask here on github / do you have a special problem in mind?

ralphrmartin:
(1) OK, but a link to the searx page(s) with the uninstall instructions would still be helpful
(2) What is not clear is whether using the old searx settings.yml in searxng will work, or whether some differences exist which need to be taken care of. So, a statement like this would help:
Most settings in /etc/searx/settings.yml are also applicable to /etc/searxng/settings.yml, but the following differences may need adjustment
(a) searx settings not used in searxng....
(b) searx settings with different options / meanings in searxng...
(c) new searxng settings not in searx...

Hope these suggestions and comments are useful.

return42:
> Hope these suggestions and comments are useful.

Yes thanks! .. tbh I think this will result in a never ending list / with each commit in SearXNG the delta of these two projects increase.  Personally I don't want to wast time with a look "backward", especial since it is not clear what "backward" means / from what searx fork a admin comes from / there are several searx forks e.g. [my-spot](https://gitlab.e.foundation/e/cloud/my-spot) to name just one.

IMO it is very simple / if someone wants to replace a software, the old software needs to be uninstalled according to the deinstallation procedure of the old software .. and the new software needs to be installed .. and at least the documentation of the basic settings should be read [settings.yml](https://searxng.github.io/searxng/admin/engines/settings.html)

SearXNG is a rolling release, to catch rough issues we implemented a check-procedure:

    $ sudo -H ./utils/searx.sh install check

with each significant change, the check procedure will be supplemented with further tests & checks.

Combined with the documentation (in general and from this PR) the admin should be able to solve issues.  Everything more can be discussed in the community at [#searxng:matrix.org](https://matrix.to/#/#searxng:matrix.org) or here in an issue at github.

@dalf what do you think, do you miss anything .. do we need to document more here ?

mrpaulblack (approved):
@return42 I think this PR is ready to merge tbh. It is not really feasible for us to write a migration, since we would need to update this for each breaking commit (since we are rolling release this would be even harder).

Letting the user know to uninstall SearX completely and install SearXNG is IMO best option, because otherwise we could get issues from broken migrations and so on.

PR #628, opened 2021-12-20
[doc] engine-table: sort by Disabled and Name

not-my-profile:
## What does this PR do?

Sorts enabled before disabled and then by engine name.

Before:

image

After:

image

Makes it easier to see which engines are enabled / disabled by default. Makes it easier to find engines by name.

Run `make docs` and open `dist/docs/admin/engines/configured_engines.html` in your browser.


return42 (approved):
Thanks a lot!

Issue #623, opened 2021-12-20
Why are engines disabled by default?


not-my-profile:
I can gather that there can be the following reasons why an engine is disabled by default:

* it only serves a non-English language (e.g. duden.de)
* it would duplicate search results (e.g. it doesn't make sense to enable invidious when youtube is enabled)
* it doesn't work well?

But that does not explain why e.g. hoogle & pypi are enabled by default whereas npm & rubygems are disabled by default.

return42:
TBH we do not have a detailed definition for this decision.

With the defaults we want to serve a moderate setup that works *good* for most common use cases ... what ever the most common use cases are is the first variable in our definition.

But I think there are simple rules that might help to find a decision:

- The engine should not be enabled by default if many issues expected by this engine
- The engine should not be enabled by default if an API key ore more setup is needed
- Engines in the 'general' category enabled by default should support different languages 
- ...

> But that does not explain why e.g. hoogle & pypi are enabled by default whereas npm & rubygems are disabled by default.

These are in category `IT`, there is no need for language support.

For me `npm & rubygems` do work reliable .. if there a no other bad UX with this both engines   we should consider to activate `npm & rubygems` also by default.

What do other think? .. should I prepare a PR for `npm & rubygems` .. any other suggestions to improve the defaults?

mrpaulblack:
Tbh for my public instance the default engines do not really work (either to slow, a lot of IP blocking or the returned results are low quality IMO); These are the changes from SearXNG upstream, that I applied to my fork:
https://github.com/paulgoio/searxng/blob/8d6e7f7e9b6449e73ef5b45e219140a6c59911d4/Dockerfile#L43-L61

(Edit: IMO `use_mobile_ui: true` should be set by default in `settings.yml`)

not-my-profile:
I think it would be neat to document the reasons why the engines are disabled, e.g.:

* `disabled_because: rate-limit`
* `disabled_because: lang:de`
* `disabled_because: duplicates:youtube`

These reasons could then be displayed in the [engine tables](https://searxng.github.io/searxng/admin/engines/configured_engines.html). This would help admins decide whether or not they want to enable a given engine. For example German speakers might want to enable all German engines that are disabled by default ... but currently these aren't easily findable.

return42:
> think it would be neat to document the reasons why the engines are disabled, e.g.:
> .. These reasons could then be displayed in the engine tables.

[...]

PR #619, opened 2021-12-19
Apply black

dalf:
## What does this PR do?

This PR is demo of [black](https://pypi.org/project/black/):
* first commit `black -l 120 -S searx searxng_extra` : reformat except the simple quote.
* second commit : `black -l 120 searx searxng_extra`: reformat including the simple quote.

## Why is this change important?

Normalize code format.

## Author's checklist

* About PEP8 E203: see https://github.com/psf/black/issues/315
* If we use black, we can disable pycodestyle, and just reformat the code (and keep pylint). The tooling is not modified, so `make test` fails for now.
* there is an emacs package: https://github.com/wbolster/emacs-python-black

PR #629, opened 2021-12-21
[doc] introduce about.language and sort engines by it

not-my-profile:
## What does this PR do?

Introduces `about.language` for non-English engines and sorts the tables after (disabled, about.language, name).

Before:

image

After:

image

It helps non-English speakers find engines of their language.

## Related issues

#623

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,7 +50,11 @@ jinja_contexts = {
     },
 }
 jinja_filters = {
-    'sort_engines': lambda engines: sorted(engines, key=lambda engine: (engine[1].disabled, engine[0]))
+    'sort_engines':
+    lambda engines: sorted(
+        engines,
+        key=lambda engine: (engine[1].disabled, engine[1].about.get('language', ''), engine[0])
+    )
 }
 
 # usage::   lorem :patch:`f373169` ipsum
--- a/searx/engines/duden.py
+++ b/searx/engines/duden.py
@@ -16,6 +16,7 @@ about = {
     "use_official_api": False,
     "require_api_key": False,
     "results": 'HTML',
+    "language": 'de',
 }
 
 categories = ['general']
--- a/docs/admin/engines/configured_engines.rst
+++ b/docs/admin/engines/configured_engines.rst
@@ -43,6 +43,9 @@ Explanation of the :ref:`general engine configuration` shown in the table
         - ``!{{mod.shortcut}}``
         - {{mod.__name__}}
         - {{(mod.disabled and "y") or ""}}
+          {%- if mod.about and  mod.about.language %}
+          ({{mod.about.language | upper}})
+          {%- endif %}
         - {{mod.timeout}}
         - {{mod.weight or 1 }}
         {% if mod.engine_type == 'online' %}
--- a/searx/engines/ina.py
+++ b/searx/engines/ina.py
@@ -18,6 +18,7 @@ about = {
     "use_official_api": False,
     "require_api_key": False,
     "results": 'HTML',
+    "language": 'fr',
 }
 
 # engine dependent config
--- a/searx/engines/mediathekviewweb.py
+++ b/searx/engines/mediathekviewweb.py
@@ -14,6 +14,7 @@ about = {
     "use_official_api": True,
     "require_api_key": False,
     "results": 'JSON',
+    "language": "de",
 }
 
 categories = ['videos']
--- a/searx/engines/seznam.py
+++ b/searx/engines/seznam.py
@@ -22,6 +22,7 @@ about = {
     "use_official_api": False,
     "require_api_key": False,
     "results": "HTML",
+    "language": "cz",
 }
 
 base_url = 'https://search.seznam.cz/'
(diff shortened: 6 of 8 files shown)


not-my-profile:
Note that `about.language` is meant as "for which speakers is this engine primarily interesting?". Which is also why I didn't add `language: de` for ccc-tv because it contains many English videos.

docs/conf.py
```diff
@@ -50,7 +50,11 @@
     },
 }
 jinja_filters = {
-    'sort_engines': lambda engines: sorted(engines, key=lambda engine: (engine[1].disabled, engine[0]))
+    'sort_engines':
+    lambda engines: sorted(
+        engines,
+        key=lambda engine: (engine[1].disabled, engine[1].about.get('language', ''), engine[0])
```
return42, line 56: Nit: can you add a space left & right to the equation symbol / even in arguments / I know it is not PEP8, but it is much better to read ;)

    key = lambda engine: (...)

searx/engines/duden.py
```diff
@@ -16,6 +16,7 @@
     "use_official_api": False,
     "require_api_key": False,
     "results": 'HTML',
+    "language": 'de',
```
return42, line 19: Should we better add `languages` (plural) with a python list of languages .. if the engine supports several engines (more than three .. or five?) we can use an entry like `languages = ['i18n', ]`  .. suggestions?

return42 (changes requested):
LGTM thanks a lot! .. I made some remarks, see below .. what do you think?

searx/engines/duden.py
not-my-profile: My idea was that `about.language` specifies the user interface language for engines without multilanguage support (i.e. no `supported_languages`). I'd like to avoid duplicating `supported_languages`.

return42: FYI we have a `supported_languages` field for engines, this could be used 

```
$ ./manage pyenv.cmd python
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import searx.engines
>>> import searx.plugins
>>> searx.engines.load_engines(searx.settings['engines'])
...
>>> searx.engines.engines['google'].supported_languages
...
>>> searx.engines.engines['mediathekviewweb'].supported_languages
```

Sometimes this field is a dict, sometimes it is a list (I don't know why / sorry, I had no time to dig deeper in this topic).

return42: Ops, your comment shows up after I send mine / sorry.

> I'd like to avoid duplicating `supported_languages`.

That's fine I think / OK, lets stay with `language`.

not-my-profile: To clarify: `supported_languages` can be specified with the `language` parameter.

`about.language` on the other hand is explicitly for engines that don't support multiple languages (and the language thus cannot be specified with the `language` parameter).

return42: Thanks!

docs/conf.py
not-my-profile: I'm happy to adhere to any formatting style you choose for your project. I'm however also a fan of consistency and currently the code mostly sticks to the PEP8 style. Given that there's a recent PR to introduce auto-formatting #619, arbitrarily applying manual formatting here seems weird to me.

return42: > I'm happy to adhere to any formatting style you choose for your project. I'm however also a fan of consistency and currently the code mostly sticks to the PEP8 style. 

Our code is (mostly) PEP8 / but not in this one case: What PEP8 says about formatting arguments is good for one-liners but not for a multiline argument list, especial not where the value of an argument takes several lines (IMO PEP8 is pointless in this case).

> Given that there's a recent PR to introduce auto-formatting ..

Yes we have #619 but this PR is in a draft state and I don't know when we have finished a code-style workflow for python like we have done this month for JS.

```
jinja_filters = {
    'sort_engines': lambda engines: sorted(
        engines,
        key = lambda engine: (
            engine[1].disabled, engine[1].about.get('language', ''), engine[0]
        ),
    )
}
```

.. is much better to read / nevertheless what PEP8 says .. AFAIK black won't fix this to PEP8 style.

not-my-profile:
```
$ grep -Pzohr '[(,][\n\t ]*[a-z]+=[^\n]+' --include='*.py' searx  | tr "$'\0'" $'\n' | grep = | wc -l
413
$ grep -Pzohr '[(,][\n\t ]*[a-z]+ =[^\n]+' --include='*.py' searx  | tr "$'\0'" $'\n' | grep = | wc -l
77
```

The PEP8 style seems way more common in the codebase.

>AFAIK black won't fix this to PEP8 style

Actually it does. (Black is in general very strict to prevent such discussions.)

return42: > Actually it does.

Seems you are right .. I will never understand why this arbitrary setting of whitespaces ..

```
dict_a = {
    "foo": "bar",
    "john": "doe",
}

dict_b = dict(
    foo="bar",
    john="doe",
)
```

should be more readable than ..

```
dict_a = {
    "foo" : "bar",
    "john" : "doe",
}

dict_b = dict(
    foo = "bar",
    john = "doe",
)
```
.. anyway if the majority thinks every detail of PEP8 weights more than readability ..

PR #621, opened 2021-12-20
[doc] Improve engines table

not-my-profile:
## What does this PR do?

Reworks the engines table to be more readable and accessible.

| Before | After |
| -- | -- |
| image | image 

## Why is this change important?

Readability and accessibility are important.

## How to test this PR locally?

Run `make docs` and open `dist/docs/admin/engines/configured_engines.html` in your browser.

## Author's checklist

* Please don't squash my commits.

PR #634, opened 2021-12-22
Introduce `categories_as_tabs` & group engines in tabs

not-my-profile:
## What does this PR do?

Previously all categories were displayed as tabs in the UI. This PR changes this so that only categories
listed under `categories_as_tabs` in `settings.yml` are displayed.

This lets us introduce more categories without cluttering up the user-interface.
These new non-tab categories are then:

* used to group the engine tables in the preferences
* used to group the engine tables in the documentation
* searchable with bangs e.g. `!repos youtube-dl`

For details please see the commit messages.

|documentation|preferences|
|--|--|
|image| image

The grouping provides a better overview and the bangs allow for new queries.

## Related issues
#623, #621


not-my-profile:
@return42 this PR initially introduced `about.powered_by` ... I have now generalized this to `about.subcategory`.

Now ready for review :)

return42:
I haven't had time for a review now (coming soon), but adding subcategories makes sense to me, it brings some meaningful grouping in the unordered list of engines we have.

@dalf what do you think; the "about" section has been introduced in a4dcfa0 but never been documented. I think we should add some documentation of the "about" properties to the developer section. If we add subcategories (here in this PR) we should document what are meaningful categories and give some rules at hand to classify subcategories.

return42:
@not-my-profile can you please rebase and place the settings.yml patches from 836ca9b6d2 and 6923f121a in one commit? .. otherwise I can do it, when I am in review with this PR (needs write access on this branch).

not-my-profile:
>can you please rebase

Done :)

>I think we should add some documentation of the "about" properties to the developer section.

I think we should introduce a documentation page `Contributing a new engine` (or sth. like that).

As the last step the page could then say:
>Build the documentation with `make docs` and open `dist/docs/admin/engines/configured_engines.html` in your browser.
Check if your engine shows up in the right section. If it should be in a different subsection, set `about.subcategory` accordingly.

(Because if e.g. `mediawiki` is already setting `subcategory: wikis` you don't need to set it in `settings.yml`.)

dalf:
About the subcategory:
* the subcategory could be a query criteria. For example, the user could query for `!files:app osmand` to search for `osmand` using the engines of the `files` category and `app` subcategory. In this case the subcategory would be a global variable as the category variable (and it requires to change webadapter.py, query.py and search/__init__py).

About powered by:
* Are there engines that use multiple sources?
* A point that we can most probably skip because too much details: the wikipedia page of duckduckgo says that it uses bing and is own index. It's different from startpage that proxy the request to google without its own index. Should we make a difference ?

(Side note: I usually search for `ifconfig.me` in search engines to get the crawler name).

not-my-profile:
>the subcategory could be a query criteria. For example, the user could query for `!files:app osmand`

Yes it could be. I am however unsure if that use case is common enough to warrant the additional complexity. For example if you are looking for an Android app chances are you only want to search you favorite app store so e.g. `!gpa osmand` or `!fd osmand`. The other subcategories also don't really strike me as something you would want to filter by.

>Are there engines that use multiple sources?

Yes but I don't think we want to maintain metadata about meta search engines. That's too meta :D For example we support etools which is a meta search engine for 17 other engines ... I am certain that we do not want 17 powered by ... subsections. If we at one point end up having another general meta search engine we could however introduce a "metasearch engines" subcategory.

> the wikipedia page of duckduckgo says that it uses bing and is own index. It's different from startpage that proxy the request to google without its own index. Should we make a difference ?

IMO the key aspect about "powered by X" is that X has the power to censor results / rank results however it wants. Building your own index lets you rank yourself but if you don't crawl yourself X can still censor results. IMO the risk of censorship is important enough to warrant "powered by X" for Y even if Y has its own index.

not-my-profile:
Ok now that I think about it the IT subcategories would lend themselves to filtering.
For example `!repos youtube-dl` to search for youtube-dl repositories across GitHub, GitLab, Gitea etc.

Note that for UX reasons I think we should avoid complicating our querying syntax ... i.e. no `!it:repos` but just `!repos`.

not-my-profile:
@dalf I changed the PR to no longer introduce a subcategory property but instead use the same `categories` system we currently have (so that secondary categories can still be filtered). I updated the PR description accordingly. Thanks for the idea ... I think this is much better :)

@return42 I am afraid I don't know what's causing the test failure.

return42:
This comes from the `test.robot` / it is recommended to run [make test](https://searxng.github.io/searxng/dev/quickstart.html) before commit.

I haven't had time to look close, but I assume there has been changed something in the `test_preferences`

- https://github.com/searxng/searxng/runs/4609123377?check_suite_focus=true#step:7:78

https://github.com/searxng/searxng/blob/fe751b0ee8af955390ed4caf56e2c9b267f5caa3/tests/robot/test_webapp.py#L34

When I am in review with this PR I will have a look at this.

not-my-profile:
`make test` doesn't work for me locally (see what I wrote on IRC). Yeah I also found the exception but I haven't yet figured out why the checkbox is missing.

not-my-profile:
Ok I figured out why it fails. The dummy engine defined in `test_query.py` isn't configured as `main_categories` so it doesn't show up in the settings. Do you have an idea how to add `settings.yml` configuration for the test?

dalf:
<details>

<summary>A patch to fix the build</summary>

```patch
diff --git a/searx/settings_defaults.py b/searx/settings_defaults.py
index 0b79703c..e6622995 100644
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -20,6 +20,18 @@ OUTPUT_FORMATS = ['html', 'csv', 'json', 'rss']
 LANGUAGE_CODES = ['all'] + list(l[0] for l in languages)
 OSCAR_STYLE = ('logicodev', 'logicodev-dark', 'pointhi')
 SIMPLE_STYLE = ('auto', 'light', 'dark')
+MAIN_CATEGORIES = { k: None for k in [
+    'general',
+    'images',
+    'videos',
+    'news',
+    'map',
+    'music',
+    'it',
+    'science',
+    'files',
+    'social media',
+]}
 STR_TO_BOOL = {
     '0': False,
     'false': False,
@@ -209,7 +221,7 @@ SCHEMA = {
         'off_when_debug': SettingsValue(bool, True),
     },
     'categories_order': SettingsValue(list, []),
-    'main_categories': SettingsValue(dict, {}),
+    'main_categories': SettingsValue(dict, MAIN_CATEGORIES),
     'engines': SettingsValue(list, []),
     'doi_resolvers': {
     },
diff --git a/tests/robot/settings_robot.yml b/tests/robot/settings_robot.yml
index e0d01bb3..79837152 100644
--- a/tests/robot/settings_robot.yml
+++ b/tests/robot/settings_robot.yml
@@ -33,6 +33,10 @@ outgoing:
   request_timeout: 1.0  # seconds
   useragent_suffix: ""
 
+main_categories:
+  general:
+  dummy:
+
 engines:
   - name: general dummy
     engine: dummy
diff --git a/utils/brand.env b/utils/brand.env
index 31afce53..84827acb 100644
--- a/utils/brand.env
+++ b/utils/brand.env
@@ -1,5 +1,5 @@
 export SEARXNG_URL=''
 export SEARXNG_PORT='8888'
+export GIT_URL='https://github.com/dalf/searxng'
 export GIT_BRANCH='master'
```

</details>

The patch adds a default value for `main_categories`.
Also why is it a dictionary ?

not-my-profile:
Ah thanks, I just fixed it :)

> Also why is it a dictionary ?

Because we probably want to add a `label` field in the future so that e.g. `it` is displayed as IT. `description` might also be a good addition in the future.

dalf:
> Ah thanks, I just fixed it :)

The patch includes a default value for the `main_categories` settings, so even an old `settings.yml` would still work. The idea is to make the upgrade painless, especially since we use a rolling release.

> Because we probably want to add a name field in the future so that e.g. it is displayed as IT.

Basically this for now:
https://github.com/searxng/searxng/blob/fe751b0ee8af955390ed4caf56e2c9b267f5caa3/searx/webapp.py#L163-L175

The new implementation has to be compatible with [/babel.cfg](https://github.com/searxng/searxng/blob/fe751b0ee8af955390ed4caf56e2c9b267f5caa3/babel.cfg), I'm not sure how to do that.

But why the "secondary" categories would have a names ? It can be useful:
* in the documentation (similar to `it` -> `IT`)
* in the preferences, if we display the secondary categories.

not-my-profile:
>The idea is to make the upgrade painless, especially since we use a rolling release.

Ah that makes sense, I'll add that!

>The new implementation has to be compatible with /babel.cfg, I'm not sure how to do that.

I'm afraid I don't understand the problem? Why is it incompatible?

>But why the "secondary" categories would have a names ?

Once we introduce category metadata we can just introduce another `secondary_categories` setting. I'd like to keep that for another PR.

return42:
> @dalf I changed the PR to no longer introduce a subcategory property but instead use the same categories system we currently have (so that secondary categories can still be filtered). I updated the PR description accordingly. Thanks for the idea ... I think this is much better :)

This c43b9b633 is a incompatible change, the meaning of categories from /etc/searxng/settings.yml will be break.

I haven't understand all you and @dalf discussed but I think the first implementation with subcategory property was better.

not-my-profile:
>This c43b9b6 is a incompatible change, the meaning of categories from /etc/searxng/settings.yml will be break.

 If we implement the default value as suggested by @dalf, existing configurations would still work.

> I think the first implementation with subcategory property was better.

I strongly disagree. From an UX standpoint it's way better to say: "You can filter by categories with `!`" than it is to say "You can filter by categories with `!` but that doesn't work for subcategories." In particular because users looking at the Configured Engines documentation might not realize that they are looking at a subheading and try to filter for it with `!` anyway.

return42:
> If we implement the default value as suggested by @dalf, existing configurations would still work.

I'm out, I do not understand whats going on here.

not-my-profile:
@dalf Any concerns remaining?

return42:
See https://github.com/searxng/searxng/pull/634#issuecomment-999840450 ..

I have the following setup in where I add an engine to more then one category:

```yaml

  - name: mediathekviewweb
    engine: mediathekviewweb
    shortcut: mvw
    categories: general, TV
    disabled: false
```

With this PR I miss the TV category I get in current master branch:

grafik

with this PR I see only the "main" categories:

grafik

not-my-profile:
Yes, that is intentional. The PR defines many new categories ... showing them all as tabs in the UI would be impractical. If you want to define a new "main category" you need to declare it in your `settings.yml`.

I don't see this as a problem because I think that only very few Searx instances define their own categories (and those just need to update their `settings.yml` once). The secondary categories this PR introduces on the other hand however benefit the vast majority of Searx users because they can now search wikis with `!wikis`, repositories with `!repositories`, etc.

I think the additional categories greatly improve the engines documentation page. Yes we could also do that with another `subcategory` property but then subcategories couldn't be queried via bangs (which seems backwards to me and as I explained would confuse users trying to query subcategories they saw in the documentation).

return42:
> Yes, that is intentional.

OK, this was not clear to me .. now I am 'in' again ;) 

> I think that only very few Searx instances define their own categories (and those just need to update their settings.yml once).

.. thanks for clarifying.  Could you please document the new "category" setup method.  I assume adding a `main_categories` section to chapter [Global Settings](https://searxng.github.io/searxng/admin/engines/settings.html#global-settings) would be a good starting point.

return42:
> I don't see this as a problem because I think that only very few Searx instances define their own categories (and those just need to update their `settings.yml` once).

Define own categories is a common used feature.  In that discussions I remember it was always known, that new categories can be defined in every engine setup, like shown here: 

```yaml

  - name: mediathekviewweb
    engine: mediathekviewweb
    shortcut: mvw
    categories: general, TV
    disabled: false
```

AFIU with this PR I added the TV to ..

```yaml

main_categories:
  - general
  - images
  - videos
  - news
  - map
  - music
  - it
  - science
  - files
  - social media
  - TV
```

and I have back my "TV" .. thats fine I think.

But one thing confuse me ...

> The secondary categories this PR introduces on the other hand however benefit the vast majority of Searx users because they can now search wikis with `!wikis`, repositories with `!repositories`, etc.

In the engine setting above the engine is registered in two main categories:

```yaml

       categories: general, TV
```

Now my question: is "TV" also a secondary category .. can I use this secondary category in other engine setups, not in main category "TV" ..  

Another example/question I have, how can I add the engine from above to a secondary category e.g. to `databases` / will this be the right configuration? ...

```yaml

       categories: general, TV, database
```

not-my-profile:
> is "TV" also a secondary category .. can I use this secondary category in other engine setups, not in main category "TV" ..

No, a category is either a main or a secondary category ... never both. When it's listed under `main_categories` it becomes a main category otherwise it's a secondary category. Alternative way to put this: The categories are the same for all engines.

>  will this be the right configuration?

Yes, adding a category to an engine always puts the engine in that category ... no matter if it's a "main category" or not.

An engine can be in multiple main categories and multiple secondary categories. The order in which you specify categories does not matter (except for the documentation page which groups the engines by their first category and then by their second category).

return42:
Thanks! .. very good description you gave .. could you please add such a description to the [Global Settings](https://searxng.github.io/searxng/admin/engines/settings.html#global-settings).

return42:
> they can now search wikis with `!wikis`, repositories with `!repositories`, etc.

One more, I think we do need a description for the users https://searxng.github.io/searxng/user/search_syntax.html

From where can the user get the informations what _secondary categories_ are defined / I can see the groups in the documentation .. but this means that documentation needs to be recompiled each time the categories are configured ... and most of the instance do not serve an own build of the documentation  --> The list _secondary categories_  must be shown somewhere in the property page.

not-my-profile:
@return42 Thanks for kicking me on this :) I documented `main_settings` but more importantly I now updated the preference templates to also group categories by the "second category". (I updated the screenshots in the PR description).

Notice that the engines in the documentation and the settings are now sorted and grouped exactly the same (powered by a new helper function `searx.engines.group_by_subcategory`).

This also means that the engines in the preferences are now sorted by `about.language` (e.g. the German dictionaries are at the end of the dictionary group). I also updated the templates to display the language code. We might want to change `about.language` to `about.country` in the future and display flags instead. But I think that can wait for another PR.

What do you think of the grouped engine preferences? I think it looks really good.

docs/admin/engines/settings.rst
```diff
@@ -219,6 +219,25 @@ Communication with search engines.
...
+    - images
+    - videos
+    - news
+    - map
+    - music
+    - it
+    - science
+    - files
+    - social media
+
```
return42: > I documented `main_settings`

Can you please add a documentation of main- & secondary categories here, like you gave in the comments above https://github.com/searxng/searxng/pull/634#issuecomment-1001010436

I also miss a description for the users / https://github.com/searxng/searxng/pull/634#issuecomment-1001011772 ...

>> they can now search wikis with `!wikis`, repositories with `!repositories`, etc.

> One more, I think we do need a description for the users https://searxng.github.io/searxng/user/search_syntax.html

searx/templates/oscar/preferences.html
```diff
@@ -348,7 +348,11 @@ <h3>{{ _('Engines') }}</h3>
                                     <th scope="col" class="text-right">{{ _("Allow") }}</th>
                                     {% endif %}
                                 </tr>
-                        {% for search_engine in engines_by_category[categ] %}
+                        {% for group, engines in engines_by_category[categ] | group_engines %}
+                        {% if not (group == 'powered by other sources' and loop.index == 1) %}
+                        <tr><th colspan=9>{{_(group)}}</th></tr>
+                        {% endif %}
+                        {% for search_engine in engines %}
```
return42: > From where can the user get the informations what secondary categories are defined ... 

I think we need a list of secondary categories documented somewhere in the preferences .. not just grouping ...

> I now updated the preference templates to also group categories by the "second category".

I am sorry to say, but this sort order (by secondary category) confused me / tooked a while I understand ... and when I look for the setting of an engine (aka module) I miss the alphabetic order ... is it only me?

grafik

searx/engines/google.py
```diff
@@ -41,7 +41,7 @@
 }
 
 # engine dependent config
-categories = ['general']
+categories = ['general', 'powered by google']
```
return42: I understand categories like 'repos', 'wikis', 'database' ... theses categories classify engines ... but I do not understand how 'powered by ...' fits into this scheme of classifications.

We need a documentation of the intention of subcategories, we should name what _default_ set of subcategories SearXNG has and we should give some rules at hand how to classify new added engines well in the right subcategory.

return42 (commented):
Thanks for your additional work .. I left some comments.  Over all I think grouping by "subcateories" is a useful addition, but I really miss a better documentation of this *feature* and the concept behind.  I left some comments below about documentation ... and UX (from my point of view / lets here what other say).

searx/templates/oscar/preferences.html
not-my-profile: I think that's because the Simple theme aligns the new category headings in the center. It think it should look better when it's aligned to the left (and maybe has a different background for emphasis).

not-my-profile: >I think we need a list of secondary categories documented somewhere in the preferences .. not just grouping ...

I don't think that's necessary.

I updated the simple theme, I think the subcategories now look much better:

image

not-my-profile:
I added three commits, the first of which should hopefully address the lack of documentation. The last two update the simple theme so that the subcategories in the preferences look better.

not-my-profile:
Rebased after the black update.

dalf:
Minor bug fix:
this line https://github.com/searxng/searxng/blob/c6922ae7c5e53f695d5f5f8704b10b4e2815cda4/searx/query.py#L225
should be `self._add_autocomplete(first_char + category.replace(' ', '_'))`
so the spaces in category names are parsed as intended.

---

Note about an existing limitation: an category and an engine can't have the same because the bang doesn't make the difference. For example if we have a "wikipedia" category, "!wikipedia" can mean be either the engine or the category.
(the current implementation choose choose the engine).

---

In the master branch, the main usage of a category is to allow the user to send a same query to the enabled engines of this category.
The engines are displayed by category in the preferences so the users can specify where the query goes for each category.

Category choice [can be locked in settings.yml](https://github.com/searxng/searxng/blob/54bce130f9074c3d63009237b014c727a1443cc5/searx/preferences.py#L330): when locked, the users can't choose a different category but they can choose which engines to enable in the allowed categories. I have to double check, it seems that "lock: categories" means allow only the general category ([hard coded](https://github.com/searxng/searxng/blob/54bce130f9074c3d63009237b014c727a1443cc5/searx/preferences.py#L329)), which could be replaced by "load engine from the general category, not the other".

An engine can belong to multiple [main] categories, but it seems that we are going to a different way: "an engine belongs to one and only one [main] category".

---

* the query `!general time` sends the query to enabled engines of the general category.
* the secondary categories "apps", "repos" are categories like the "general" category: "!apps time" seems to be a valid usage.

---

* the secondary category "powered by google" is different: why would the users send a query to all the engines powered by google? (`!powered_by_google time`). The result are slightly different but not much.

We can imagine that the user select google and then select the google provider (the google engine or the startpage engine).
It doesn't work for bing because some engines provide their own index (ddg for example).

Or we can imagine that the user type "!powered_by_bing", and the autocompletion allows to select one of the engines powered by bing (same for "!powered by youtube", etc...)

In both cases I can think of, the "powered by *" are a way to select one engine among a list, not to send a query to multiple engines.

---

The wiktionary engine currently belong to the general, dictionaries and wikis categories.
It seems the /preferences shows the engine main category, and first secondary category but not the second secondary category.
I guess this intended to avoid duplicate entries in the preferences? It can be deceptive when type the query "!wikis time".

The query "!dictionnaries house" seems like "!general house" but for ... dictionnaries.
But what about "!wikis house"? How does it help to choose an engines? there are wiki about games, about compagnies, etc...

Debatable suggestion: replace the wikis category by "wikipedia fundation" category (a long name, "wikipedia" won't work as described above).

---

Special engines:
* "dictzone" & "mymemory translated" translate word. They are enabled with a special query syntax ["<language_from>-<language_to> <query>"](https://github.com/searxng/searxng/blob/c6922ae7c5e53f695d5f5f8704b10b4e2815cda4/searx/search/processors/online_dictionary.py#L12). Queries that doesn't match this pattern are ignored. These two engines **could** be in the "special query" tab or at least in the "translation" category.
* In the same way, "currency" **could** be in the "special query" tab too since it answers only to query matching this pattern ["<number> <currency> to <number> <currency>"](https://github.com/searxng/searxng/blob/c6922ae7c5e53f695d5f5f8704b10b4e2815cda4/searx/search/processors/online_currency.py#L13).

---

About the "database" secondary category: IMO the term is too generic because all engines have a database (except meta search engine).

* [Wikidata element for Erowid](https://www.wikidata.org/wiki/Q1430691) is not helpful.
* [Wikidata element for IMDB](https://www.wikidata.org/wiki/Q37312) belongs to the "social cataloging application".

Suggestion: what about "catalog" instead of "database"?

---

## Sum up

Semantically, it seems the secondary categories are:
* categories like the existing one ("apps", "repos", "dictionaries")
* source ("powered by *"). The words "powered by" has to be translated too.
* tags ("wikis") : an engine can have multiple tags.

For the sources ans tags, the bang are not useful and the auto-completion can suggest the engines that match (either sources or tags).

If we dump all these information, the /preferences will look like searx.space, and that's a no go.

__if__ we add tags and sources, to get the same UI in the preferences, the engines must be grouped by:
* the first secondary category
* or if there is no secondary category, the first tag
* or if there is no tag, the first source

not-my-profile:
Thanks for the in-depth feedback!

I completely reworked the PR to implement the changes we discussed.

* main & secondary categories was confusing -> now it's named `categories_as_tabs` which is much clearer
* the categories for which, as you rightly pointed out, searching all contained engines at the same time doesn't make sense (wikis & databases) have been removed
* a new `web` category has been added so that web search engines aren't mixed up with wikis
* dictionaries, erowid & imdb were moved out of the general category (as you suggested). they are now shown in an "other" tab in the preferences
* powered by is now again a separate engine property (`about.powered_by`) but still used to group the engines

searx/webapp.py
```diff
@@ -168,8 +172,19 @@
...
+    gettext('dictionaries'),
+    gettext('q&a'),
+    gettext('repos'),
+    gettext('software wikis'),
+    gettext('web'),
+    gettext(DEFAULT_GROUP_NAME),
+    gettext(OTHER_CATEGORY),
 )
 
+_powered_by = gettext('powered by %(company)s')
```
dalf: No need for this one since there is `_('powered by %(company)s')` in the .html templates.

searx/templates/oscar/preferences.html
```diff
@@ -298,7 +298,7 @@ <h3>{{ _('Privacy') }}</h3>
             <div class="tab-pane active_if_nojs" id="tab_engine">
                 <!-- Nav tabs -->
                 <ul class="nav nav-tabs nav-justified hide_if_nojs" role="tablist">
-                    {% for categ in all_categories %}
+                    {% for categ in categories_as_tabs %}
```
dalf: It should be `categories_as_tabs + [OTHER_CATEGORY]` too otherwise the "Other" tab is missing (Oscar theme)

searx/engines/__init__.py
```diff
@@ -258,3 +267,28 @@ def load_engines(engine_list):
...
+def group_engines_in_tab(engines):  # pylint: disable=redefined-outer-name
+    def engine_sort_key(engine):
+        return (engine.about.get('language', ''), engine.name)
+
+    def group_sort_key(group):
+        ((groupname, powered_by), _engines) = group
+        return (powered_by == '', groupname == DEFAULT_GROUP_NAME, groupname.lower())
+
+    def get_group(eng):
+        non_tab_engines = [c for c in eng.categories if c not in settings['categories_as_tabs'] + [OTHER_CATEGORY]]
```
dalf: should `non_tab_engines` be `non_tab_categories` ?

searx/engines/__init__.py
```diff
@@ -258,3 +267,28 @@ def load_engines(engine_list):
         if engine:
             register_engine(engine)
     return engines
+
+
+DEFAULT_GROUP_NAME = 'others'
+
+
+def group_engines_in_tab(engines):  # pylint: disable=redefined-outer-name
```
dalf: Usually function dedicated to UI are in `searx/webutils.py`

searx/templates/simple/preferences.html
```diff
@@ -289,12 +289,27 @@ <h2>{{ _('Preferences') }}</h2>
         <th>{{ _("Max time") }}</th>
         <th>{{ _("Reliability") }}</th>
       </tr>
-      {% for search_engine in engines_by_category[categ] %}
+      {% for (group, powered_by), engines in engines_by_category[categ] | group_engines_in_tab %}
+      {% if loop.length > 1 %}
+      <tr><th colspan=9 class=engine-group>
```
dalf: it should be `<tr><th colspan="9" class="engine-group">` with quote

dalf:
I've left some minor comments about the code.

There are still few things that disturb me.

Here three screenshots:
image

image

image

* The red boxes are not categories.
* The green boxes are categories.

There are no visual clue to show what is a category (so when the users can use a bang).

If an engine has two non tab categories, the second one is never shown.

I feel that we introduce categories (with | without tab) and subcategories.

not-my-profile:
Thanks for the review, I addressed your comments about the code :)

>There are no visual clue to show what is a category (so when the users can use a bang).

That is very much intentional. The engine preferences already show much information ... we have to be careful not to overload our UI with too much information. If you think the bangs should be discoverable via the UI, I think a better way to achieve that would be to improve the autocompletion for `!`.

>If an engine has two non tab categories, the second one is never shown.

Also intentional for the same reason. We don't want to show an engine twice in the same tab that would be confusing. And I don't think there's a good place to show that information in the engine table (without overloading the UI).

>I feel that we introduce categories (with | without tab) and subcategories.

Yes the categories without tabs can be used as subcategories that is exactly the point of this PR. The reason I think we should introduce `categories_as_tabs` as opposed to a `subcategories` engine property is that:

* you shouldn't have to remember what's a category / a subcategory when categorizing engines
* it lets admins promote a non-tab-category to a tab category without needing to change all the engine categories

mrpaulblack:
I really really really like this PR! Great Work!

Some things to note:

* I think in the It category: dockerhub, npm, packagist, pypi and rubygems should be part of some sort of packages group
* I dont understand the others category in the settings; It makes logically sense to have it, but its part of the pref. page, but not part of the result page, so for me as a user it is not obvious in what category they are part of
* when I search for this `!dictionaries test` I get no results... Am I doing something wrong?

not-my-profile:
>I really really really like this PR! Great Work!

Thanks! I'm glad you like it :)

>I think in the It category: dockerhub, npm, packagist, pypi and rubygems should be part of some sort of packages group

Done. I added hoogle as well.

>I dont understand the others category in the settings; It makes logically sense to have it, but its part of the pref. page, but not part of the result page, so for me as a user it is not obvious in what category they are part of

Good point! I added a help text:

image

(Yes it does not explain what bangs are ... currently none of our UI does ... we really have to improve our in-UI documentation of bangs ... but I consider that a separate issue, this PR is already big enough).

Just for clarification: The "Other" tab isn't displayed for search results since it's bound to contain miscellaneous engines which you wouldn't want to search at once.

>when I search for this `!dictionaries test` I get no results... Am I doing something wrong?

I assume you only had dictzone enabled. Dictzone is a special search engine that's only active when the search query contains a pattern like `de-en`. Displaying such special search engines right next to regular search engines is obviously confusing, I discussed this with @dalf and we both agree that we should move these special engines to the "Special Queries" tab in the future (but that's also a separate issue). I moved the dictzone engine back into the `general` category for now so that e.g. the query `en-de search` still works (and you don't have to do `!dictionaries en-de search`).

mrpaulblack:
Ok thank you for your quick changes.

2 more things:

* I would enable all engines in the other category by default. They dont show up on normal result page anyway and its good to have them on, when someone is using the bang IMO (maybe leave the language specific ones off...)
* When I reset to default in /preferences and select the simple theme after that and safe; the cookie saves all default engines in enabled like this:

image

I also tested this with my instance which is up to date with master and it does not happen with it...

not-my-profile:
>I would enable all engines in the other category by default.

Good thinking. I enabled them with my last commit.

>the cookie saves all default engines in enabled

Huh, I have no idea why that's happening. Maybe @dalf has an idea?

dalf:
In one hand, there are three types of categories:
1. "main" category as tab
1. "main" category not as tab
1. "subcategory" (category not as tab, but the engine as a "main" category as tab)

In the other hand, the preferences saving function. It save engine that are not enabled by default but enable by the user. The "key" is (engine name and the category). The category is included because an engine can behave differently according the category.

That's why the category is set in the [params](https://github.com/searxng/searxng/blob/8eb3167fba1f7aacbd4eb3da8615a3e58f249eb3/searx/search/processors/abstract.py#L149)

It works well for category of type 1 and 2, but not for the categories of type 3.

So now the question is which engines use the feature:
* searx/engines/piratebay.py (piratebay in the music category doesn't do the same query than in the files category ... expect now it belongs only to the files category).
* searx/engines/searx.py

There are two solutions:
* either an engine belongs to only one category, wait no, to only "main" category (type 1 or 2)
* either change the preferences code to skip categories of type 3 in the cookies.

---

Note: a name for the categories of type 1, 2 and 3 would be easier.

docs/admin/engines/settings.rst
```diff
@@ -219,6 +219,26 @@ Communication with search engines.
 ``max_redirects`` :
   30 by default. Maximum redirect before it is an error.
 
+``categories_as_tabs:``
```
return42, line 222: nit: is `category_tabs` a better name?

searx/webutils.py
```diff
@@ -134,3 +136,30 @@ def is_flask_run_cmdline():
     if len(frames) < 2:
         return False
     return frames[-2].filename.endswith('flask/cli.py')
+
+
+DEFAULT_GROUP_NAME = 'others'
+
+
+def group_engines_in_tab(engines):
```
return42: Hard to read/understand, can you please left some hints in a doc-string about the sort order algorithm, the input value and the return value.

searx/engines/__init__.py
```diff
@@ -44,6 +44,9 @@
     "display_error_messages": True,
     "tokens": [],
 }
+# set automatically when an engine does not have any tab category
+OTHER_CATEGORY = 'other'
```
return42, line 49: Nit: would `NON_CATEGORY = 'uncategorized'` better?

searx/webutils.py
```diff
@@ -134,3 +136,30 @@ def is_flask_run_cmdline():
     if len(frames) < 2:
         return False
     return frames[-2].filename.endswith('flask/cli.py')
+
+
+DEFAULT_GROUP_NAME = 'others'
```
return42, line 142: Single-point of definition: should we better use the value from `searx.engines.OTHER_CATEGORY`? --> In webapp.py there is a `from searx.engines import  OTHER_CATEGORY, DEFAULT_GROUP_NAME`.

Do we need two values, one for default one for the "OTHER" tab in the preferences?

searx/settings.yml
```diff
@@ -1076,10 +1083,12 @@ engines:
   - name: qwant
     engine: qwant
     shortcut: qw
-    categories: general
+    categories: [general, web]
     disabled: false
     additional_tests:
       rosebud: *test_rosebud
+    about:
+      powered_by: Bing
```
return42: Isn't it the same for the other qwant engines ('qwant images', 'qwant news' ...)?

Are there any reasons why "powered_by" is sometimes set in the settings.yml and sometimes direct in the engine (aka module)?

return42:
> Note: a name for the categories of type 1, 2 and 3 would be easier.

:+1:

> "subcategory" (category not as tab, but the engine as a "main" category as tab)

Hm, what does this mean? .. I do not understand ... do you mean a engine that is set to an empty list of categories (see 2. below).


-----

From user's POV it is unclear (the arrow --> point to what I suggest):

1. for what is the "OTHER" category (see tab in screenshot below) --> a user-documentation is needed.
2. for what is subsection "others" in tab "OTHER" / AFAIK these are engines not in any category --> a user-documentation is needed. 
3. what is the sort order in "subcategories"? When no individual user setup is active, the sort order is by active/inactive in subcategories, when the user activate an engine there is no longer any sort order (see below) --> sort-order should always alphabetical, I mean we have categories & subcategories and in this subcategory there is really no need to sort by active/inactive.

grafik


----

>> I think we need a list of secondary categories documented somewhere in the preferences .. not just grouping ...

> I don't think that's necessary.

I would insist .. From admin's POV it is unclear ... (repost what I asked  9 days ago / see comments above) ..

We need a documentation of the intention of subcategories, **we should name what default set of subcategories SearXNG has builtin and we should give some rules at hand how to classify new added engines well in the right subcategory.**

----

Looks confusing to have "others" and "others (powered by ...)" --> "powered by" is a good information, should be shown in a tooltip or something else ... but not a sort order!

grafik

If we take "powered by" as sort order once we will get a messy, here is an example where "powered by Bing" was added to qwant-video (https://github.com/searxng/searxng/pull/634#discussion_r778008300):

grafik

IMO it makes no sense to have a group "powered by Bing" where only one engine is listed in while bing is not in this group ..

**I really like this PR but the several sort criteria do end in a mess!**

docs/admin/engines/settings.rst
not-my-profile: No.

searx/settings.yml
not-my-profile: >Isn't it the same for the other qwant engines ('qwant images', 'qwant news' ...)?

For web engines it's easy to verify by searching for a site that returns the User-Agent. For images and news it's not so easy to verify.

>Are there any reasons why "powered_by" is sometimes set in the settings.yml and sometimes direct in the engine (aka module)?

Yes it follows the existing convention of defining metadata in the module if possible (e.g. categories are currently defined in the engines when possible).

return42: > For web engines it's easy to verify by searching for a site that returns the User-Agent. For images and news it's not so easy to verify.

I do not understand this answer .. perhaps I do not understand what "powered by" means .. I thought, the results of this engine (e.g. qwant) are powered by e.g. Bing .. I am wrong? .. what do you mean "User-Agent"?

> Yes it follows the existing convention of defining metadata in the module if possible (e.g. categories are currently defined in the engines when possible).

I asked, because you defined "powered by" of qwant in settings.yml but in bing (itself) and ddg you defined it in the engine module:

grafik

docs/admin/engines/settings.rst
return42: I asked because a conjunction (_x_as_y_) in a name of an option is not common.

not-my-profile:
> a user-documentation is needed

I consider categories a configuration detail. Users should not be concerned about categories at all. As far as users are concerned there are tabs they can click on and bangs they can type. The fact that bangs are currently not discoverable via the UI is a problem, but a separate issue that is out of the scope of this PR.

>what is the sort order in "subcategories"? 

They are sorted alphabetically: first by `about.language` (with English engines first) and then by name. I agree that this currently isn't obvious but I think it should become more obvious when the language codes are replaced with country flags so e.g.: :de: and :poland:. I also think that this should be done in a follow up PR.

>We need a documentation of the intention of subcategories, we should name what default set of subcategories SearXNG has builtin and we should give some rules at hand how to classify new added engines well in the right subcategory.

Yes more documentation would be nice but I also think that this should be done in the follow up PR. Mainly for the reason that the lack of documentation about categories is an existing problem. (Yes this PR adds new categories but "subcategories" really aren't as important as the tab categories which currently aren't documented ...)

>"powered by" is a good information, should be shown in a tooltip or something else ... but not a sort order!

Oh yes it most definitely should be used to group engines. That's the point of this PR! As I have explained multiple times it does not make much sense to enable multiple engines powered by the same data source (e.g. Google and Startpage or YouTube and Invidious). The engines should therefore absolutely be grouped by `about.powered_by` to help users make that decision.

>IMO it makes no sense to have a group "powered by Bing" where only one engine is listed in while bing is not in this group ..

Good point, I updated `group_engines_in_tab` to account for that.

>I really like this PR but the several sort criteria do end in a mess!

I strongly disagree.

docs/admin/engines/settings.rst
not-my-profile: `categories_as_tabs` is way clearer than `category_tabs` because it explains that you should list categories there and these categories function as tabs. `category_tabs` does not communicate any of that. "Category tabs" could be something completely different than categories.

searx/settings.yml
not-my-profile: >I thought, the results of this engine (e.g. qwant) are powered by e.g. Bing .. I am wrong? .. what do you mean "User-Agent"?

Search engines like Qwant don't actually disclose in detail which results are powered by what. One way to find out is to find the User-Agent of the result crawler in their search result descriptions.

>I asked, because you defined "powered by" of qwant in settings.yml but in bing (itself) and ddg you defined it in the engine module

Yes because I wanted to avoid categorizing Qwant Video & Images with powered by since I actually don't know that.

searx/webutils.py
not-my-profile: I think for the tab "Other" fits better because general, news, map, music, it, science, social media are all singular.
Whereas the subcategories are almost all plural, e.g. packages, repos & others, apps & others, dictionaries & others.

searx/webutils.py
not-my-profile: Valid criticism. I refactored the method and added a docstring + type hints.

searx/engines/__init__.py
not-my-profile: I don't think so. Firstly "uncategorized" would be completely misleading since engines with categories can be in the "OTHER_CATEGORY". Secondly naming the variable something other than the display text just makes the mental mapping more complicated.

searx/webutils.py
return42: But we do not need a "DEFAULT_GROUP_NAME" twice I think.

PLEASE: do not close my remarks as "resolved conversation" .. I will do it .. and I think @dalf, @mrpaulblack should have a chance to read my remarks ...

not-my-profile: >But we do not need a "DEFAULT_GROUP_NAME" twice I think.

OTHER_CATEGORY and DEFAULT_GROUP_NAME fulfill different functions. Using a single variable for different things seems like bad style to me. (Besides that if we want different UI text like I explained we need different variables).

>PLEASE: do not close my remarks as "resolved conversation"

Right, sorry about that. I think it's a bit annoying how GitHub intermixes code review with PR discussion, it makes the comments on the PR hard to follow.

return42: >  it makes the comments on the PR hard to follow.

Yes, its hard to follow .. I understand.

>> But we do not need a "DEFAULT_GROUP_NAME" twice I think.
> OTHER_CATEGORY and DEFAULT_GROUP_NAME fulfill different functions. 

Ooops sorry, my fail: I thought you have defined DEFAULT_GROUP_NAME twice .. but thats not the case.

searx/webutils.py
return42: Thanks / but does not really describe the sorting algorithm (sort order criteria) ... anyway.

searx/settings.yml
return42: Qwant is powerd by Bing .. but both have different results for the same search term ... 

TBH we do not know which engine is "powered by" which search engine .. AFIAK qwant mix results from bing and from a qwant index.

No matter from which point I look on this "powered by" .. it should never be a sort criteria, since it could never be a serious criteria.

not-my-profile: > it should never be a sort criteria, since it could never be a serious criteria

Of course it can (notice the bingbot):

image

return42:
> The fact that bangs are currently not discoverable via the UI is a problem, but a separate issue that is out of the scope of this PR.

OK

> They are sorted alphabetically: first by about.language (with English engines first) and then by name. 

OK, thanks for clarify.

> Yes more documentation would be nice but I also think that this should be done in the follow up PR.

Might be Ok to describe categories in a second PR, but in this PR we use (unknown) rules to set categories and it should be clear why _this and that_ category is used _here and there_ and what the meaning of the category is.  Otherwise we are not able to review how we have used the categories.

> Oh yes it most definitely should be used to group engines. That's the point of this PR! 

I have a different view on this PR, this might be the cause of the long discussions:

For me the major change is, that we have categories that are not tabs .. and we can select these categories by a !bang ... this is also the reason why I think it is needed to give the user a list of categories (at least those are no tabs) and the admin & developers need a description what categories are exist .. and what are the rules for a category.

> As I have explained multiple times it does not make much sense to enable multiple engines powered by the same data source (e.g. Google and Startpage or YouTube and Invidious).

Next point I have a different opinion:

It makes sense .. see my comment about qwant .. usual we do not know how a engine collects its results.

EDIT: and search engines can change there resources from day to day without publish this information .. "powered by ..." is just a nice info but it is not a warranted characteristic.

>I really like this PR but the several sort criteria do end in a mess!

I strongly disagree.

@dalf & @mrpaulblack what do you think about sorting? .. especially the "powered by" grouping in the UI ... https://github.com/searxng/searxng/pull/634#issuecomment-1004736125

not-my-profile:
>For me the major change is, that we have categories that are not tabs .. and we can select these categories by a !bang

Yes that's one of the major changes introduced by this PR.

> this is also the reason why I think it is needed to give the user a list of categories (at least those are no tabs)

I agree. I think that can be nicely done by improving the autocompletion in another PR.

> EDIT: and search engines can change there resources from day to day without publish this information .. "powered by ..."

Yes of course it can change anytime ... in practice it does not though. DDG has been powered mainly by Bing for several years.

>is just a nice info but it is not a warranted characteristic

Data sources are very important. If DDG mainly gets its results from Bing it means that if Bing censors results they also won't show up in DDG. I think that completely warrants the characteristic.

not-my-profile:
While trying to make sense of `searx.preferences` I ended up refactoring it quite a bit ... I have now removed those commits from this PR since it's already large enough (and will open a separate PR for that in the future).

Anyway I fixed the superflous cookies.

As far as I am concerned this PR is ready to be merged. I am not willing to split it up even further into smaller PRs ... I already spent too much  time on this.

searx/settings.yml
return42: >  (notice the bingbot):

Sorry if I ask a dumb question, but how reliable is the assumption that ddg uses (only) results from a Bing partnership?

not-my-profile: [from their website](https://help.duckduckgo.com/duckduckgo-help-pages/results/sources/):

>We also of course have more traditional links in the search results, which we also source from multiple partners, **though most commonly from Bing**

Please note that I never implied that they "only" use Bing ... but it's their main data source (as can be verified with the User-Agent trick).

return42:
> As far as I am concerned this PR is ready to be merged. I am not willing to split it up even further into smaller PRs ... I already spent too much  time on this.

Me and @dalf also spent a lot of time in review this PR.  I support @dalf in the opinion that this PR must be split.  But this does not mean, that all that work is up to you.  Perhaps @dalf or me start with a smaller PR and cherry pick from here.

We have already done this (splitting a PR) in the past and most often it took a while but for the community it is easier to follow .. and most often the results are better .. but "yes", it is a lot of work.

not-my-profile:
> this PR must be split

This PR adds 272 lines and removes 96 lines ... that's really not much. And the changes are done in atomic commits that are easy to review.

not-my-profile:
Ok I can split `about.powered_by` into a separate PR.

return42:
> Ok I can split `about.powered_by` into a separate PR.

lets wait what @dalf will suggest. you spent a lot of time and you got a lot of comments from mine and others.  I think it is time we reverse the game and me or @dalf give suggestions (PR) and you can comment.

I have respect for your achievement and I can understand that this whole process is very exhausting and also a bit frustrating for you.

not-my-profile:
> lets wait what @dalf will suggest

Too late. I already rebased it. Only took me 10 minutes. :face_exhaling: (Reconstructing `about.powered_by` for another PR will take longer). And I updated the PR description for the ninth time :sweat_smile:

not-my-profile:
I rebased the PR to also add the "web" category for image & video search engines that scrape the entire web. E.g. bing images, duckduckgo images, google images, qwant images are now in that category as opposed to devianart or flickr (which host their own photos).

image

The `!web` bang can now be used to search the entire scraped web for anything and returns a mixture of pages, images and videos ... which I think is very cool!

I opened the issue #690 to track what needs to be addressed after this PR is merged.

EDIT: I also added a `lyrics` category for azlyrics and genius.

return42 (approved):
Thanks a lot for you hard work on this topic and thanks for removing "powered by" and opening https://github.com/searxng/searxng/issues/690

@dalf @mrpaulblack I think this PR is now a good starting point for #690 .. lets merge.

I slightly modified (the order of) the topics in #690

dalf:
thank you for this awesome contribution!

not-my-profile:
Thanks to you guys as well! It wouldn't have gotten this awesome without your continuous feedback.

PR #640, opened 2021-12-22
Drop dead engines

### LINKED ISSUE #639: whaleslide.com is dead
Added half a year ago in #178 (from https://github.com/searx/searx/pull/2861) the website (both https://whaleslide.com/ and  https://search.whaleslide.com) appear to be completely offline. Should we just remove it from `settings.yml`?

not-my-profile:
Closes #639.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -720,25 +720,6 @@ engines:
       requires_api_key: false
       results: JSON
 
-  - name: geektimes
-    engine: xpath
-    paging: true
-    search_url: https://habr.com/en/search/page{pageno}/?q={query}
-    url_xpath: //article[contains(@class, "post")]//a[@class="post__title_link"]/@href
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -720,25 +720,6 @@ engines:
...
-    disabled: true
-    shortcut: gt
-    about:
-      website: https://geektimes.ru/
-      wikidata_id: Q50572423
-      official_api_documentation:
-      use_official_api: false
-      require_api_key: false
-      results: HTML
-
```
return42, line 741: In #639 only whalside is named as offline / whats wrong with geektimes.ru? FYI geektimes.ru is redirected to habr.com (see 45f0e1a859fa).

not-my-profile: We already have a search engine for habr.com ... we don't need two (besides they are both broken anyway since they changed their search page ... but that's a different issue).

Issue #332, opened 2021-09-17
The standalone scripts do no longer work


return42:
The _standalone_ scripts do no longer work:: 

```
./manage pyenv.cmd python searx_extra/google_search.py foo
Traceback (most recent call last):
  File "searx_extra/google_search.py", line 10, in <module>
    from searx.search import default_request_params
ImportError: cannot import name 'default_request_params' from 'searx.search' (searx/searx/search/__init__.py)
```

```
/manage pyenv.cmd python searx_extra/standalone_searx.py foo
Traceback (most recent call last):
  File "searx_extra/standalone_searx.py", line 215, in <module>
    sys.stdout.write(dumps(
  File "/usr/lib/python3.8/json/__init__.py", line 234, in dumps
    return cls(
  File "/usr/lib/python3.8/json/encoder.py", line 201, in encode
    chunks = list(chunks)
...
  File "searx_extra/standalone_searx.py", line 138, in json_serial
    raise TypeError("Type ({}) not serializable".format(type(obj)))
TypeError: Type (<class 'datetime.timedelta'>) not serializable
```

not-my-profile:
This issue was fixed in #658.

PR #651, opened 2021-12-24
remove no longer maintained scripts from searxng_extra

return42:
## What does this PR do?

remove no longer maintained scripts from searxng_extra

tidy up

## Related issues

- https://github.com/searxng/searxng/issues/332


dalf:
IMO we should keep `searxng_extra/standalone_searx.py`: it can be useful to have a command line to run the app.
* For example here and there I've seen request to make a equivalent of [Google Alert](https://en.wikipedia.org/wiki/Google_Alerts). `standalone_searx.py` can be starting point.
* It can also be useful to have an search engine available in command line (using `coloredlogs` or `rich`)
* Or just pipe the output to some other commands.

return42:
> it can be useful to have a command line to run the app.

Yes I also like to have a command line .. but one that works :-) .. try:

    ./manage pyenv.cmd python searxng_extra/standalone_searx.py --help

For me it is not clear what use case of this implementation is / is the command line started like shown above? .. or do I need to install (`make install`).

We can implement a command line but then it should be placed in a folder ./cli (or similar).  At the of the reorganization of SearXNG I see the following root folders in the repository:

- `./client` --> where the HTML clients are located
- `./server` --> where the webapp is located
- `./cli` --> where a command line client is located

If you want, we can implement a command line, but this broken code in standalone_searx.py want help us I think.

dalf:
> but one that works :-) .. try:

A quick fix

```patch
diff --git a/searxng_extra/standalone_searx.py b/searxng_extra/standalone_searx.py
index e243d9cb..de8a0d77 100755
--- a/searxng_extra/standalone_searx.py
+++ b/searxng_extra/standalone_searx.py
@@ -196,9 +196,14 @@ def parse_argument(
 
 
 if __name__ == '__main__':
-    searx.search.initialize()
+    settings_engines = searx.settings['engines']
+    searx.search.load_engines(settings_engines)
     engine_cs = list(searx.engines.categories.keys())
     prog_args = parse_argument(category_choices=engine_cs)
+    searx.search.initialize_network(settings_engines, searx.settings['outgoing'])
+    searx.search.check_network_configuration()
+    searx.search.initialize_metrics([engine['name'] for engine in settings_engines])
+    searx.search.initialize_processors(settings_engines)
     search_q = get_search_query(prog_args, engine_categories=engine_cs)
     res_dict = to_dict(search_q)
     sys.stdout.write(dumps(
```

What happens with `--help`: the engines are initialized with background threads, and the main thread quickly displays the help message and quit. Since the main thread is exiting, [the network are closed](https://github.com/searxng/searxng/blob/6ffa741e7a739a8b7df210054300dda8c33d75b7/searx/network/network.py#L361-L378) by the [atexit](https://docs.python.org/3/library/atexit.html) handler, so the background threads find that the HTTP clients are closed which throw an exception.

The fix is to load the engines but to initialize them only after the arguments parsing.

The engines must be loaded first to be able to display the category list in the help message.

> ./client --> where the HTML clients are located

I don't understand what goes into this folder.

> ./cli --> where a command line client is located

An alternative is to move `standalone_searx.py` to `searx/__main__py`, so `python -m searx[ng]` works and it can be installed using `setup.py` --> `./searx[ng] "time"` works too.

return42:
> A quick fix

@dalf thanks .. but:

> An alternative is to move `standalone_searx.py` to `searx/__main__py`, so `python -m searx[ng]` works ..

I would vote to implement a generic command line which can be started by `python -m searx[ng]`.  The command line can group commands and can be extended to the _needs_.  But I think we can drop `standalone_searx.py` since I do expect some different _needs_.

Its up to your decision, if want, we can keep `standalone_searx.py` ... just give me a short feedback if you want to keep ... / if so please send your PR to fix `standalone_searx.py`.

----

>> ./client --> where the HTML clients are located

> I don't understand what goes into this folder.

Somewhere in the future, the source of the HTML-CSS-JS clients could be moved into this folder, e.g.:

    rename ./searx/static/themes/simple/src    ./client/simple
    rename ./searx/static/themes/__common__    ./client/__common__

dalf:
>  I do expect some different needs.

What are they?

>  I do expect some different needs.

I would like to keep it, so we can take time to implement a new script. If someone wants a command line the existing script still provides some values.

return42:
>>  I do expect some different needs.

> What are they?

- A command line **client** / not a command line **server like** `standalone_searx.py` it is.  The big drawback of this solution is, that each time I use this command line my IP is send to all active engines (google, yahoo ..) and can be tracked. --> this is against of the privacy concept of SearXNG.

- A output that is rendered for the command line .. not a JSON dump.

The `standalone_searx.py` script is good for developers, to see how things in SearXNG do work and to experiment with .. but `standalone_searx.py` is nothing that should be used by end-users.

dalf:
Replaced by https://github.com/searxng/searxng/pull/658

PR #663, opened 2021-12-28
changes about the secret_key

dalf:
## What does this PR do?

Two commits about the secret_key:
* the first commit uses [hmac.compare_digest](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) to prevent timing attack.
* the second commit remove dead code: the secret_key is always a str, never bytes.

* security fix
* clean code

* in the master branch and on this branch, set the secret_key to something like `!!binary NGViODljMDY2YWJhNGVhNDZkNjA2NzAxZjlkMmY0`  : the server doesn't start.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1067,8 +1067,9 @@ def image_proxy():
     if not url:
         return '', 400
 
-    h = new_hmac(settings['server']['secret_key'], url.encode())
-    if h != request.args.get('h'):
+    h_url = new_hmac(settings['server']['secret_key'], url.encode())
+    h_args = request.args.get('h')
+    if len(h_url) != len(h_args) or not hmac.compare_digest(h_url, h_args):
         return '', 400
 
     maximum_size = 5 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


return42 (approved):
LGTM / thank you!

I pushed 2e44867 on top of your branch, if you think it doesn't help, please remove this commit.

The intention of 2e44867 is to have both function to _generate & compare_ a key at one place.  From where other applications (routes) can make use of it.

searx/webapp.py
```diff
@@ -1067,8 +1068,7 @@ def image_proxy():
     if not url:
         return '', 400
 
-    h = new_hmac(settings['server']['secret_key'], url.encode())
-    if h != request.args.get('h'):
+    if not is_hmac_of(settings['server']['secret_key'], url.encode(), request.args.get('h', '')):
```
dalf, line 1071: Bug fix: `request.args.get('h', '')` instead of `request.args.get('h')` so the value is always a string.

Issue #498, opened 2021-11-13
tabs don't act like tabs without JavaScript


return42:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

https://github.com/searxng/searxng/commit/7a9dda

**How did you install SearXNG?**

    make run

**What happened?**

There is a "click on the .."  text direct under the categories

![grafik](https://user-images.githubusercontent.com/554536/141643149-78d5944c-3598-4865-830a-6a6cd9da5872.png)

**How To Reproduce**

~It is very strange:~ The issue appears when I enable plugins / see https://github.com/searxng/searxng/issues/498#issuecomment-968061989 and 6debda2 :

```yaml
enabled_plugins:
  - "Hostname replace"  # see configuration below

# "Hostname replace" plugin configuration example:
hostname_replace:
    'spam\.example\.com':             false  # remove results from spam.example.com
```
---

**Another issue I observe since  7a9dda**

There is a line break in the category column that overlaps in the column size of the table layout ..

![grafik](https://user-images.githubusercontent.com/554536/141643291-8b1323b4-baac-4bf1-b441-2adcf7c0710b.png)

mrpaulblack:
So the first issue is that the instant search when pressing a category is actually a plugin; You can also reproduce this by disabling javascript, which also disables the plugin. So the fix would be: enable plugins and add the search on category select plugin to the list.

> There is a line break in the category column that overlaps in the column size of the table layout ..

Yeah I noticed this one on tablet layout on the result page two.

dalf:
> There is a "click on the .." text direct under the categories

This text is shown on purpose:
* when javascript is disabled
* the "Search on Catagory select" is disabled: when disabled it is possible to select multiple category and then send the search.
![image](https://user-images.githubusercontent.com/1594191/141644329-d56753e6-ea72-432f-9080-a812b032b602.png)

It should be shown as a tooltip.

return42:
> This text is shown on purpose:

In my use case javascript is enabled .. but now I see  "Search on category select" is disabled when I enable a plugin in the settings, like  i showed above .. strange.

return42:
> So the first issue is that the instant search when pressing a category is actually a plugin; You can also reproduce this by disabling javascript, which also disables the plugin. So the fix would be: enable plugins and add the search on category select plugin to the list.

Sorry @MrPaulBlack I should have read your comment first :)

not-my-profile:
[...]

PR #664, opened 2021-12-29
[mod] manage: add themes.live command (rebuild on modification)

return42:
## What does this PR do?

[mod] manage: add themes.live command (rebuild on modification)

To get live builds while editing the HTML client of a theme.  The first argument of the command is the theme name:

    ./manage themes.live simple

To get live builds on `make run` use environment variable:

    LIVE_THEME=simple make run

Start developer session:

    LIVE_THEME=simple make run

and edit CSS (LESS) & JS of the simple theme.

## Related issues

- https://github.com/searxng/searxng/pull/473#discussion_r745371694
- https://github.com/searxng/searxng/issues/498#issuecomment-1002286730

### DIFF
--- a/manage
+++ b/manage
@@ -732,7 +732,6 @@ themes.oscar() {
 }
 
 themes.simple() {
-    local static="searx/static/themes/simple"
     (   set -e
         build_msg GRUNT "theme: simple"
         npm --prefix searx/static/themes/simple run build
--- a/manage
+++ b/manage
@@ -511,12 +511,15 @@ gecko.driver() {
     dump_return $?
 }
 
-node.env() {
+nodejs.ensure() {
     if ! nvm.min_node "${NODE_MINIMUM_VERSION}"; then
         info_msg "install Node.js by NVM"
         nvm.nodejs
     fi
+}
 
+node.env() {
+    nodejs.ensure
     (   set -e
 
         build_msg INSTALL "searx/static/themes/oscar/package.json"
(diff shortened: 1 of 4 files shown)


manage
```diff
@@ -698,14 +712,32 @@ themes.all() {
...
 }
 
+themes.live() {
+    local theme
+    case "$1" in
+        simple|oscar)
+            theme=searx/static/themes/$1
+            ;;
+        *)
+            die 42 "missing theme argument"
```
not-my-profile: That error message also comes when misspelling a theme name in which case "missing argument" is misleading.

Also in the case of an error the error message is printed but since `themes.live` runs in a background job the `die` doesn't terminate the server ... which is a bit suboptimal since developers might miss the error message because with the default DEBUG startup output it will probably scroll of screen.

not-my-profile:
I left a comment regarding the error message. Otherwise LGTM :)

manage
return42: @not-my-profile thanks for your helping remarks ..

> That error message also comes when misspelling a theme name

I changed the argument handling ...

```
$ ./manage themes.live xyz
ERROR: ./manage: line 792: themes.live(): unknown theme 'xyz' // [simple|oscar]

$ ./manage themes.live 
ERROR: ./manage: line 792: themes.live(): missing theme argument
```

> Also in the case of an error the error message is printed but since `themes.live` runs in a background job the `die` doesn't terminate the server 

I rearranged the processes and kill the process group if `themes.live` dies

```
$ LIVE_THEME=xyz make run
INFO    searx                         : load the default settings from /x/share/searxng/searx/settings.yml
INFO    searx                         : max_request_timeout=None
...
ERROR: ./manage: line 124: themes.live(): unknown theme 'xyz' // [simple|oscar]
make: *** [Makefile:28: run] Terminated
$
```

return42: @not-my-profile is the issue you named solved by my solution https://github.com/searxng/searxng/pull/664#discussion_r776499181?

not-my-profile: Yes, thanks!

not-my-profile (approved):

Issue #542, opened 2021-11-23
[simple theme] autocomplete results not clickable


tiekoetter:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

commit dae165eaa02ff51f6c1aec8c1ab33dfe4aed3797

**How did you install SearXNG?**

https://searxng.github.io/searxng/admin/installation.html

**What happened?**

When using simple theme I am not able to click on the autocomplete suggestions the list just instantly closes and the "data-autocomplete-old-value" value just stays the same as before. If I use the arrow keys to select one of the suggestions and hit enter the "data-autocomplete-old-value" gets refreshed as expected.

This problem occurs in Chrome / Chromium (Windows, Linux (Arch), macOS and Android) and Safari (iOS and macOS)

**How To Reproduce**

Installing SearXNG and trying to use the autocomplete feature.

**Expected behavior**

autocomplete behaving like it does when using arrow keys and enter.

---

Sorry if it is some trivial issue. I just discovered it while using SearXNG on a phone where arrow keys are (mostly) not available.

return42:
@tiekoetter sounds strange .. I can't reproduce this issue, do you have any cookies set? /  please go into the preferences and try to reset all settings.

Do you have any other setting changed in the settings.yml?

What happens if you clone SearXNG and run a local instance by `make run`?

tiekoetter:
@return42 I have some strange news: In a fresh Windows VM with Firefox and Google Chrome I encountered this issue again with Google Chrome. Firefox seems to work though.

**settings.yml**
autocomplete: 'google'

**preferences**
I only selected simple as the default theme and didn't change anything else.

By your suggestion I tried to run a local instance with ``make run``. This led to the same result as above.

dalf:
I confirm I have the issue on Chrome on Windows and Chrome on Android.

mrpaulblack:
Yeah also happens on Safari on iOS

return42:
@not-my-profile copied from https://github.com/searxng/searxng/issues/693#issuecomment-1005467686

> the bug apparently occurs on all browsers

I can't reproduce this issue (this is also the reason why I can't fix it) ... tested on FFFox & Chrome Desktop and on my android device where I use Fennec.

tiekoetter:
[...]

Issue #351, opened 2021-09-28
simple theme: /preferences: layout issue on mobile.


dalf:
On smartphone (real one or mobile view on FF with touch simulation), the engine response time is displayed at the wrong place:
![image](https://user-images.githubusercontent.com/1594191/135041955-5dd14dc4-4edd-4714-88b1-d650d055b1a5.png)

return42:
It also minifies the font size .... the screenshot is from FFox emulation but I see the same effect on my smartphone

![grafik](https://user-images.githubusercontent.com/554536/135060943-7904f95c-2634-4ad3-b9c0-15fe5e9e2227.png)

Another issue I see is the value in the column "Response Time" (have a look at the 0.4 value from quant) ..

![grafik](https://user-images.githubusercontent.com/554536/135061308-e1db7f7d-fa96-42a3-902c-fdd73caf1bc4.png)

when I drag the table to the left side, the 0.4 value is fixed at the same position (does not move with its column) ..

![grafik](https://user-images.githubusercontent.com/554536/135061946-4e424734-2a16-42f2-8546-98be876a020b.png)

return42:
Can't remember in which PR but this issue is fixed / I close this issue @dalf if you think different, please reopen.

PR #665, opened 2021-12-29
[mod] switch default theme from oscar to simple

return42:
## What does this PR do?

switch default theme from oscar to simple

Thanks to the fabulous work from @dalf and @mrpaulblack the simple theme is now ready to become the default.  I tested the simple theme over 3 weeks in my daily usage and I can't find a reason why oscar should be longer the default.

## Open issues

- [ ] [Simple theme as the default theme](https://github.com/searxng/searxng/projects/1)
- [x] #463
- [x] #542
- [x] #368
- [x] #351

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -76,7 +76,7 @@ ui:
   # it decreases the privacy, since the browser can records the page titles.
   query_in_title: false
   # ui theme
-  default_theme: oscar
+  default_theme: simple
   # Default interface locale - leave blank to detect from browser information or
   # use codes from the 'locales' config section
   default_locale: ""
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -177,7 +177,7 @@ SCHEMA = {
     'ui': {
         'static_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'static')),
         'templates_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'templates')),
-        'default_theme': SettingsValue(str, 'oscar'),
+        'default_theme': SettingsValue(str, 'simple'),
         'default_locale': SettingsValue(str, ''),
         'theme_args': {
             'oscar_style': SettingsValue(OSCAR_STYLE, 'logicodev'),
--- a/searxng_extra/standalone_searx.py
+++ b/searxng_extra/standalone_searx.py
@@ -105,7 +105,7 @@ def get_search_query(
         "language": args.lang,
         "time_range": args.timerange,
     }
-    preferences = searx.preferences.Preferences(['oscar'], engine_categories, searx.engines.engines, [])
+    preferences = searx.preferences.Preferences(['simple'], engine_categories, searx.engines.engines, [])
     preferences.key_value_settings['safesearch'].parse(args.safesearch)
 
     search_query = searx.webadapter.get_search_query_from_webapp(preferences, form)[0]
--- a/tests/unit/settings/user_settings.yml
+++ b/tests/unit/settings/user_settings.yml
@@ -27,10 +27,10 @@ server:
 ui:
   static_path: ""
   templates_path: ""
-  default_theme: oscar
+  default_theme: simple
   default_locale: ""
   theme_args:
-    oscar_style: logicodev
+    simple_style: auto
 
 engines:
   - name: wikidata
--- a/tests/unit/test_preferences.py
+++ b/tests/unit/test_preferences.py
@@ -119,7 +119,7 @@ class TestPreferences(SearxTestCase):
     def test_encode(self):
         from searx.preferences import Preferences
 
-        pref = Preferences(['oscar'], ['general'], {}, [])
+        pref = Preferences(['oscar', 'simple'], ['general'], {}, [])
         url_params = (
             'eJx1VMmO2zAM_Zr6YrTocujJh6JF0QEKzKAz7VVgJNohLIseUU7ivy-VcWy5yyGOTVGP73GLKJNPYjiYgGeT4NB8BS9YOSY'
             'TUdifMDYM-vmGY1d5CN0EHTYOK88W_PXNkcDBozOjnzoK0vyi4bWnHs2RU4-zvHr_-RF9a-5Cy3GARByy7X7EkKMoBeMp9CuPQ-SzYMx'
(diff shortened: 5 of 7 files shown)


mrpaulblack (approved):
There is still ARIA, RTL layout missing and stuff like settings page is not mobile friendly... But I think we should make it default to get more user feedback for the simple theme and to make it more straight forward for new instance maintainer what the big changes of SearXNG in comparison to SearX are... LGTM :+1: 

[EDIT] Ping @dalf are u ok with changing to the simple by default?

return42:
I forgot to fix the unit tests :facepalm: .. coming soon ..

tiekoetter:
I think search autocomplete suggestions should be fixed in Chrome (and Safari) first, before making simple the default theme of SearXNG.

https://github.com/searxng/searxng/issues/542

not-my-profile:
>There is still ARIA [..] missing

I'd consider that a blocker. Accessibility by default is important.

return42:
>> There is still ARIA [..] missing

> I'd consider that a blocker. Accessibility by default is important. 

What quality would be sufficient, is it enough to sprinkle a few attributes here and there?

If ARIA is a blocker then we might as well close this PR right now. I don't have ARIA knowledge, I don't know anyone who has practical experience with or anyone who can test and grade ARIA additions .. do you?

[No ARIA is better than Bad ARIA](https://www.w3.org/TR/wai-aria-practices/#no_aria_better_bad_aria)

not-my-profile:
Yes I have some experience. The fundamental basics of ARIA are actually very simple ... there is no excuse not to implement them.

E.g. the settings icon image has no alternative text for a screen reader (which can be fixed by simply adding an `aria-label`).

It's worth noting that `aria-label`s should also be translated so e.g. adding  `aria-label="{{_("Preferences")}}"` to the link would do the job.

Nowadays both Firefox and Chromium have accessibility issue reporting built into their developer tools, which can help spot accessibility problems. E.g. they report that the simple theme has contrast problems. We could also set up automatic accessibility testing in our CI with [lighthouse](https://github.com/GoogleChrome/lighthouse) (or [axe-core](https://github.com/dequelabs/axe-core) for more control). But it's worth noting that many accessibility problems cannot be detected automatically.

>What quality would be sufficient?

From the top of my head (this is accessibility in general not just ARIA):

* Every interactive UI element should have a label.
* Every visible UI element should be usable with only the keyboard (e.g. focusable via `Tab`).
* Text should have at least a contrast ratio of 4.5 (can be easily inspected with the developer tools).
* Always try to use the semantically appropriate HTML tags.
* The document should have a proper semantic hierarchy (e.g. with `h1` and `h2`).
* The behavior of interactive UI elements should always match the semantics implied by their markup (e.g. #498 violates that since the checkbox behavior is overridden via JavaScript without adding an appropriate ARIA tag ... which is a good example why it's generally a bad idea to just change behavior of semantic tags via JavaScript)

return42:
> It's worth noting that `aria-label`s should also be translated so e.g. adding  `aria-label="{{_("Preferences")}}"` to the link would do the job.

I don't know .. should it be such easy? ... is `role='button'` sufficient enough? ...  W3C says [All interactive ARIA controls must be usable with the keyboard.](https://www.w3.org/TR/aria-in-html/#3rdrule)

EDIT: I wrote this comment while you edited your comment :smile:

not-my-profile:
> All interactive ARIA controls must be usable with the keyboard.

`<a href="somewhere">some text</a>`

is 100% keyboard accessible. You can focus it and press `Enter` to open it.

>is `role='button'` sufficient enough?

`role=button` should generally **not** be added to proper links (with a working `href`). If the preferences link instead was `<a href=# onclick=openPreferences()>preferences</a>` then `role=button` would be appropriate ... since it's not an actual working link. (for such cases it's generally easier to simply use `<button>` instead though).

return42:
May be I have to describe what my intention is ... I think the requirements to SearXNG are different:

1. There are use cases where a API is needed / this is covered by the JSON interface (my wet dream is a SearXNG network hosted by the [SearXNG community](https://searx.space/)  where every query is dispatched to a different endpoint-server).
2. The majority will use a smooth WEB interface with dynamic elements (JS) / this is the place where I see the simple theme.
3. There are use cases where a pure HTML & CSS interface is needed / for what we have to implement a different theme.
4. And we have uses cases where assistive technologies are used / for what we have to implement a different theme.

> The fundamental basics of ARIA are actually very simple ... there is no excuse not to implement them.

Overlaps exist at 2., 3. and 4. but I think there is more closeness between 3. and 4. than between 2. and 4.

not-my-profile:
Your implication that being accessible requires a separate user interface is wrong (both for `noscript` as well as for people with disabilities), see [progressive enhancement](https://en.wikipedia.org/wiki/Progressive_enhancement) and WCAG. I strongly oppose the idea of implementing separate "accessible" themes for the simple reason that maintaining two themes is more work than maintaining one theme (which unless I am mistaken is the very reason we want to consolidate our two themes into one in the long run).

>I think the requirements to SearXNG are different

If you'd like me to keep contributing, accessibility by default will have to be part of our requirements.

dalf:
IMO, it is too soon:
* the RTL languages are not supported (we just had some translations). There is already a searxng-rtl.css file. A solution is some less macros. For example: `.ltr-left(5rem)` becomes `left: 5rem` in RTL languages and  `right: 5rem` in LTR language).
* at least fix some ARIA issues (for example, the preference icon doesn't have a label).

mrpaulblack:
@return42 when you have time can you please rebase this PR?

unixfox:
There is no infinite scrolling in the simple theme though, something I really like.

mrpaulblack:
Yes that is true; Tho there is a PR for this: https://github.com/searxng/searxng/pull/812

But eventually we just need to make the move, since there is always going to be something that is not finished with the theme. We can fix this after making simple the default IMO. I mean its RN still possible to switch back to oscar for people that really want to use it still...

return42:
> @return42 when you have time can you please rebase this PR?

OK, done ..

mrpaulblack:
ok thanks; I am ok with merge... @dalf @unixfox @tiekoetter are u also ok with this merge? (thumbs up or down on this comment is enough)

unixfox:
I half agree, ok for having by default in searxng but I'm going to keep oscar by default on my instance until infinite loading is available on simple theme.

PR #666, opened 2021-12-29
[docs] move searxng_extra/ stuff to the developer section.

return42:
## What does this PR do?

[docs] move searxng_extra/ stuff to the developer section.

Stuff in folder `./searxng_extra` is not suitable for normal users and should only be used by developers.

The script `searxng_extra/standalone_searx.py` must not give the impression that it improves privacy. [1]

## Related issues

[1] https://github.com/searxng/searxng/pull/651#issuecomment-1001389726


searxng_extra/standalone_searx.py
```diff
@@ -23,6 +23,14 @@
 
     $ python3 searxng_extra/standalone_searx.py rain
 
+.. danger::
+
+   Be warned, using the ``standalone_searx.py`` won't give you privacy!
+
+   On the contrary, this script behaves like a SearXNG server: your IP is
+   exposed and tracked by all active engines (google, bing, qwant, ... ), with
+   every query ... again and again and again!
```
dalf: IMO, there is no need for `... again and again and again`

dalf (commented):
I wrote one comment / note.
Otherwise, LGTM

Thank you for the clarification.

PR #679, opened 2022-01-02
searxng.org: update setup.py & settings.yml

dalf:
## What does this PR do?

* reference `docs.searxng.org` instead of `searxng.github.io/searxng`.
* setup.py: update `author` and `author_email`.

branding

### DIFF
--- a/.config.sh
+++ b/.config.sh
@@ -26,7 +26,7 @@
 #       SEARXNG_SETTINGS_PATH : /etc/searxng/settings.yml
 #       SEARX_SRC             : /usr/local/searx/searx-src
 #
-# [1] https://searxng.github.io/searxng/admin/engines/settings.html
+# [1] https://docs.searxng.org/admin/engines/settings.html
 
 # utils/filtron.sh
 # ----------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -180,7 +180,7 @@ New settings.yml
 - ``ui.advanced_search`` - add preference for displaying advanced settings ( #2327 )
 - ``server.method: "POST"`` - Make default query submission method configurable ( #2130 )
 - ``server.default_http_headers`` - add default http headers ( #2295 )
-- ``engines.*.proxies`` - Using proxy only for specific engines ( #1827 #2319 ), see https://searxng.github.io/searxng/dev/engine_overview.html#settings-yml
+- ``engines.*.proxies`` - Using proxy only for specific engines ( #1827 #2319 ), see https://docs.searxng.org/dev/engine_overview.html#settings-yml
 - ``enabled_plugins`` - Enabled plugins ( a05c660e3036ad8d02072fc6731af54c2ed6151c )
 - ``preferences.lock`` - Let admins lock user preferences ( #2270 )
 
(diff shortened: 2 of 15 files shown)


mrpaulblack (approved):
LGTM

[EDIT] I think it would be good to drop the `.html` end in the paths tbh

not-my-profile:
Related note: As long as there isn't anything on https://searxng.org/ it should probably redirect to https://docs.searxng.org/.

mrpaulblack:
@not-my-profile I can do a redirect from http://searxng.org to https://docs.searxng.org with godaddy, but its going to be from http://searxng.org; If we want from https://searxng.org we would need to do this with github pages to get a cert or if one of us setups certs and the redirect on their VPS...

not-my-profile:
Yeah I think setting that up on a VPS would probably be the best option since then we can do a proper 302 HTTP redirect (which afaik isn't possible with GitHub pages).

return42:
ATM we deploy on gh-pages by this CI:

https://github.com/searxng/searxng/blob/b630c5d7bc0bf5a4281ad402bb32adc7f6ab257f/.github/workflows/integration.yml#L82-L91

how do we deploy on searxng.org?

mrpaulblack:
We need another repo in the org named `searxng.github.io`; We can deploy from there to `searxng.org`

See: https://github.com/encode/encode.github.io/blob/master/_config.yml

[EDIT] And I think we can use CNAME files in the repo/branches to declare docs and landing page on github pages AFAIK

dalf:
> how do we deploy on searxng.org?

We need to create a project named `searxng.github.io`:
* the master branch is the content of the `https://searxng.github.io`.
* `https://searxng.org` is already configured to display the content of `https://searxng.github.io`

See:
* for example: https://github.com/encode/encode.github.io
* documentation: https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site

return42:
And how is it updated/deployed today? .. by example, today I have merge https://github.com/searxng/searxng/pull/666 and now it is deployed at https://docs.searxng.org/dev/searxng_extra/standalone_searx.py.html

dalf:
> And how is it updated/deployed today?

There is a DNS record: 
* `docs.searxng.org.       3588    IN      CNAME   searxng.github.io.`
* `searxng.org` is configured as a custom domain for the GH pages of the searxng/searxng project ( https://github.com/searxng/searxng/settings/pages )

> Yeah I think setting that up on a VPS would probably be the best option since then we can do a proper 302 HTTP redirect (which afaik isn't possible with GitHub pages).

searx.space is hosted in a Hetzner VPS. There is a room to store the content of searxng.org. Bonus: we can provide a .onion URL. But, as a first step I thought GitHub was enough.

not-my-profile:
>But, as a first step I thought GitHub was enough.

My point was that if we use `searxng.org` for branding, then it should actually redirect somewhere instead of returning a 404. Ideally the redirect should happen via a 302 status code.

mrpaulblack:
Ok redirects that point from `www.searxng.org` and `searxng.org` to `docs.searxng.org` are in place now; see: https://github.com/searxng/searxng.github.io (its an HTML redirect)...

return42 (approved):
LGTM .. thanks :+1:

I pushed d7b43064 .. it is just a nit .. if you want, I can merge d7b43064 with 76cbfbbd

PR #686, opened 2022-01-03
Add redis DB and connector

return42:
### add redis connector

The redis connector is defined in the `settings.yml`:

    redis:
      url: ...

The default DB connector is a socket at::

    unix:///usr/local/searxng-redis/run/redis.sock?db=0

### script to build & install redis DB

A script to build & install a simple & isolated redis service, dedicated to SearXNG and connected via Unix socket.

    $ ./manage redis.help
    redis.:
      devpkg    : install essential packages to compile redis
      build     : build redis binaries at /800GBPCIex4/share/SearXNG/dist/redis/6.2.6/amd64
      install   : create user (searxng-redis) and install systemd service (searxng-redis)
      remove    : delete user (searxng-redis) and remove service (searxng-redis)
      shell     : start bash interpreter from user searxng-redis
      src       : clone redis source code to <path> and checkput 6.2.6
      useradd   : create user (searxng-redis) at /usr/local/searxng-redis
      userdel   : delete user (searxng-redis)
      addgrp    : add <user> to group (searxng-redis)
      rmgrp     : remove <user> from group (searxng-redis)

To set up a redis instance simply use::

    $ ./manage redis.build
    $ sudo -H ./manage redis.install

To get access rights to this instance, your developer account needs to be added to the *searxng-redis* group::

    $ sudo -H ./manage redis.addgrp "${USER}"
    # don't forget to logout & login to get member of group

A redis DB connect can be tested by:

    >>> from searx.shared import redisdb
    >>> redisdb.init()
    True
    >>> db = redisdb.client()
    >>> db.set("foo", "bar")
    True
    >>> db.get("foo")
    b'bar'

## Related issues

- https://github.com/searxng/searxng/pull/618

### DIFF
--- a/searx/shared/redisdb.py
+++ b/searx/shared/redisdb.py
@@ -1,0 +1,41 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Implementation of the redis client (redis-py_).
+
+.. _redis-py: https://github.com/redis/redis-py
+
+This implementation uses the :ref:`settings redis` setup from ``settings.yml``.
+A redis DB connect can be tested by::
+
+  >>> from searx.shared import redisdb
+  >>> redisdb.init()
+  True
+  >>> db = redisdb.client()
+  >>> db.set("foo", "bar")
+  True
+  >>> db.get("foo")
+  b'bar'
+  >>>
+
+"""
+
+import logging
+import redis
+from searx import get_setting
+
+logger = logging.getLogger('searx.shared.redis')
+
+
+def client():
+    return redis.Redis.from_url(get_setting('redis.url'))
+
+
+def init():
+    try:
+        c = client()
+        logger.info("connected redis DB --> %s", c.acl_whoami())
+        return True
+    except redis.exceptions.ConnectionError as exc:
+        logger.error("can't connet redis DB ...")
+        logger.error("  %s", exc)
+    return False
--- a/searx/shared/redisdb.py
+++ b/searx/shared/redisdb.py
@@ -24,10 +24,16 @@ import redis
 from searx import get_setting
 
 logger = logging.getLogger('searx.shared.redis')
+_client = None
 
 
 def client():
-    return redis.Redis.from_url(get_setting('redis.url'))
+    global _client  # pylint: disable=global-statement
+    if _client is None:
+        # not thread safe: in the worst case scenario, two or more clients are
+        # initialized only one is kept, the others are garbage collected.
+        _client = redis.Redis.from_url(get_setting('redis.url'))
+    return _client
 
 
 def init():
--- a/utils/lib_redis.sh
+++ b/utils/lib_redis.sh
@@ -1,0 +1,348 @@
+#!/usr/bin/env bash
+# -*- coding: utf-8; mode: sh indent-tabs-mode: nil -*-
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# Tools to build and install redis [1] binaries & packages.
+#
+# [1] https://redis.io/download#installation
+#
+# 1. redis.devpkg (sudo)
+# 2. redis.build
+# 3. redis.install (sudo)
+#
+# systemd commands::
+#
+#    sudo -H systemctl status searxng-redis
+#    sudo -H journalctl -u searxng-redis
+#    sudo -H journalctl --vacuum-size=1M
+#
+# Test socket connection from client (local user)::
+#
+#    $ sudo -H ./manage redis.addgrp "${USER}"
+#    # logout & login to get member of group
+#    $ groups
+#    ... searxng-redis ...
+#    $ source /usr/local/searxng-redis/.redis_env
+#    $ which redis-cli
+#    /usr/local/searxng-redis/.local/bin/redis-cli
+#
+#    $ redis-cli -s /usr/local/searxng-redis/redis.sock
+#    redis /usr/local/searxng-redis/redis.sock> set foo bar
+#    OK
+#    redis /usr/local/searxng-redis/redis.sock> get foo
+#    "bar"
+#    [CTRL-D]
+
+
+# shellcheck disable=SC2091
+# shellcheck source=utils/lib.sh
+. /dev/null
+
+REDIS_GIT_URL="https://github.com/redis/redis.git"
+REDIS_GIT_TAG="${REDIS_GIT_TAG:-6.2.6}"
+
+REDIS_USER="searxng-redis"
+REDIS_HOME="/usr/local/${REDIS_USER}"
+REDIS_HOME_BIN="${REDIS_HOME}/.local/bin"
+REDIS_ENV="${REDIS_HOME}/.redis_env"
+
+REDIS_SERVICE_NAME="searxng-redis"
+REDIS_SYSTEMD_UNIT="${SYSTEMD_UNITS}/${REDIS_SERVICE_NAME}.service"
+
+# binaries to compile & install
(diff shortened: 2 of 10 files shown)


dalf:
Why not use redis from the distos?

return42:
> Why not use redis from the distos?

1. This PR adds an **isolated** redis service, dedicated to SearXNG
2. Building from sources is easy and works on all distros (incl. all architectures)

.. no need to care about distro & distro version specifics on installations & redis versions

dalf:
* How this Redis version is upgraded? Looking at https://github.com/redis/redis/releases the last versions fix multiple CVE.
* What about the log?

The use of Redis from the distro requires more work, but the maintenance is done by the distro maintainers.

searx/shared/redisdb.py
```diff
@@ -0,0 +1,41 @@
...
+
+import logging
+import redis
+from searx import get_setting
+
+logger = logging.getLogger('searx.shared.redis')
+
+
+def client():
+    return redis.Redis.from_url(get_setting('redis.url'))
```
dalf: What about this:
```python
_client = None

def client():
    global _client
    if _client:
        return _client
    # not thread safe: in the worst case scenario,  two or more clients are initialized
    # only one is kept, the others are garbage collected.
    _client = redis.Redis.from_url(get_setting('redis.url'))
    return _client
```
?

So each call to the `client()` skip the overhead of the connection initialization.

AFIU, the client is [thread safe](https://github.com/redis/redis-py#thread-safety) and there is a [connection pool](https://github.com/redis/redis-py#connection-pools).

return42: > AFIU, the client is thread safe and there is a connection pool.

`redis.Redis.from_url` just wraps `ConnectionPool.from_url` ..

https://github.com/redis/redis-py/blob/15f315a496c3267c8cbcc6d6d9c6005ea4d4a4d5/redis/client.py#L852-L853

> So each call to the `client()` skip the overhead of the connection initialization.

AFAIK there is no need for, since `redis.Redis.from_url` just wraps `ConnectionPool.from_url` all the connection management and garbage collection is done by the ConnectionPool.

dalf: `ConnectionPool.from_url` creates a new `ConnectionPool` at each call:
https://github.com/redis/redis-py/blob/15f315a496c3267c8cbcc6d6d9c6005ea4d4a4d5/redis/connection.py#L1200-L1206

I don't see any cache here.

return42: Hm, seems you are right .. I will have a look ..

return42:
> The use of Redis from the distro requires more work, but the maintenance is done by the distro maintainers.

TBH I wont to get rid of all the distro dependencies we have in our scripts / it is very hard to maintain I remember a centos-7 and it quirks, the different apache bundles distros build, how they handle uwsgi and more ..  Each distro and distro-version has its own software-package build from different software versions .. its not such simple to maintain them all well.

> How this Redis version is upgraded? 

We can lift `REDIS_GIT_TAG="${REDIS_GIT_TAG:-6.2.6}"` to the next tag and we can add a function somewhere (e.g. in [searxng_check.py](https://github.com/searxng/searxng/blob/master/utils/searxng_check.py)) that checks the version of the installed redis server.

TBH this is a starting point, that gives an easy method to developers to start with redis development.  I haven't thought all to its end and tbh I don't want make the wrong decisions to early / redis is optional and it is free to the users to connect to a redis TCP connection from whatever url.

searx/shared/redisdb.py
return42: @dalf: thanks for your hint / fixed in dca83944

return42:
@dalf do you have any serious reasons that would prevent us from merging this PR?

I have some features in mind where a storage is needed for / redis seems perfect for this.

utils/lib_redis.sh
```diff
@@ -0,0 +1,348 @@
...
+        "armv6" | "armv7l") ARCH=armv6l ;;
+        "armv8") ARCH=arm64 ;;
+        .*386.*) ARCH=386 ;;
+        ppc64*) ARCH=ppc64le ;;
+    *)  die 42 "ARCH is unknown: $(command uname -m)" ;;
+    esac
+    echo "${ARCH}"
+}
+
+# TODO: move this to the right place ..
```
dalf, line 332: where?

dalf:
The big issue with utils/lib_redis.sh is the update process: even if we manually upgrade this line https://github.com/searxng/searxng/blob/dca83944b588be3ec9e49486daea6cf15ef58f78/utils/lib_redis.sh#L42 , nothing prevent an admin to forget about this custom redis installation.... but as far I understand, this custom redis installation is not mandatory.

If this is the case, then LGTM

utils/lib_redis.sh
return42: > where?

`lib.sh` .. On the long term I will split (reorganize) the `lib.sh` in modules (we talked about a few month ago) .. ATM a func name `bash.cmd` does not fit good in the `lib.sh` ... ATM we have more urgent tasks, but I will an eye on.

return42:
> The big issue with `utils/lib_redis.sh` is the update process ..
> but as far I understand, this custom redis installation is not mandatory.

Right, it is not mandatory!

The intention is to have a client [searx/shared/redisdb.py](https://github.com/searxng/searxng/pull/686/files#diff-30b71865ec101531bb273e7914fb5db534f88d43fc0643b172bcd189ccec22fb) that coneects to a redis DB by an URL [`redis:`](https://github.com/searxng/searxng/pull/686/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802f)

In use cases (and for developer purpose) where no redis DB exist, one can be installed by `utils/lib_redis.sh`.  This redis DB is connected by a unix socket and isolated from a redis installation of the OS package manager that might be exists on the same host.

This redis DB is a [service](https://github.com/searxng/searxng/pull/686/files#diff-043c2feab6acfb6fa0280f7ee4766ac80a761df56e725e9547753bb96fa491ea) that runs under the user `REDIS_USER="searxng-redis"`

If this is OK for you I will going to merge this PR.

searx/shared/redisdb.py
```diff
@@ -0,0 +1,47 @@
...
+def client():
+    global _client  # pylint: disable=global-statement
+    if _client is None:
+        # not thread safe: in the worst case scenario, two or more clients are
+        # initialized only one is kept, the others are garbage collected.
+        _client = redis.Redis.from_url(get_setting('redis.url'))
+    return _client
+
+
+def init():
```
dalf, line 39: Where this function is called?

dalf:
In https://github.com/dalf/searxng/commit/a8639f15545392cf22193e6cd8bc999fdc220418 shows how it can be integrated with the existing code. This is WIP

searx/shared/redisdb.py
return42: We do not have a redis app right now, so it is not called.  Since redis DB is optional the intention is something like this:

```python
if redisdb.init():
   # read / write DB
else:
   # no DB available
```

return42:
> In https://github.com/dalf/searxng/commit/a8639f15545392cf22193e6cd8bc999fdc220418 shows how it can be integrated with the existing code.

Thanks a lot / this is similar to what I had in mind when I implemented this PR.

If you give me your approved to this PR I will have look about a `shared_redis` implementation.

dalf (approved):

return42:
@dalf when you have time could you have a look why the docker build https://github.com/searxng/searxng/runs/4780168620?check_suite_focus=true#step:8:838 exit with error?  I re-run the but it doesn't helped.

PR #689, opened 2022-01-03
[simple theme] use flexbox instead of js for positioning images

### LINKED ISSUE #368: Image category: layout using flexbox instead of javascript

mrpaulblack:
## What does this PR do?

Drop JS to position images in image category. In this PR flexbox is used to psotion images in an responsive layout.

Drop JS and simplify theme.

```make run```

Navigate to images category as well as to the social media category (reddit engine returns an image raster...)

## Author's checklist

Desktop: 
image

(Tablet looks like desktop; but the detail modal overlays the entire view)

Mobile:
image

## Related issues

Closes #368

### CODE REVIEW

File: searx/static/themes/simple/src/less/style.less [hunk 1]
```diff
@@ -330,20 +330,25 @@ article[data-vim-selected].category-social {
...
+  flex-grow: 10;
+}
+
 .result-images {
-  display: inline-block;
-  margin: 0;
+  flex-grow: 1;
+  margin: 0.5rem !important;
   padding: 0;
   position: relative;
```
dalf: I think `position: relative` can be removed, and perhaps it fixes the aspect ratio problem.

dalf:
This line can be removed:
https://github.com/searxng/searxng/blob/8eb3167fba1f7aacbd4eb3da8615a3e58f249eb3/searx/static/themes/simple/gruntfile.js#L102

Note: `!general !images time` most probably break the layout (untested), and require an element to group the images.

return42:
> This line can be removed: https://github.com/searxng/searxng/blob/8eb3167fba1f7aacbd4eb3da8615a3e58f249eb3/searx/static/themes/simple/gruntfile.js#L102

Does it make sense to move JS from ..

- https://github.com/searxng/searxng/tree/master/searx/static/themes/__common__/js

to oscar

- https://github.com/searxng/searxng/tree/master/searx/static/themes/oscar/src/js

If so, I can prepare a patch / or we do it in another PR.

----

EDIT:

its just a nit, but to avoid [bisect](https://en.wikipedia.org/wiki/Bisection_(software_engineering)) issues I recommend to merge this two commits in one:

- 7eb4d108 [simple theme] position images with flexbox
- ce837458 [simple theme] drop image_thumbnail_layout js script

return42:
> the images all take up the same size since they all have the same flexbox growth value; IDK how to fix that (so that they keep their original aspect ratio)

I don't know exactly what I prefer ... in case of doubt a would say, it is better to have a preview of the whole image ... e.g. if there is a text banner on the top, it might be not shown in the new presentation.

**old** 

grafik

**new**

grafik

----

When I looked at both, side by side, I would vote for a solution that supports the original aspect ratio.


**old**

grafik

**new**

grafik

---

what do other think about _original aspect ratio_?

mrpaulblack:
@return42 Yeah I agree that we should keep the original aspect ratio of the image in this PR. Basically I want the changes in this PR to be invisible to the end user. So same layout and everything, but under the hood CSS instead of the JS stuff. That is why this PR is still a draft. Since I am still trying to fix that...

mrpaulblack:
@dalf I think this is ready for review;

The major things that are still broken:

* image category mixed with other categories is broken, since the flexbox definition is on the URL element in the image category; a fix could be to put the image articles in another div with display: flexbox; applied
* [EDIT] maybe also lazy load is broken for image category..? sometimes it works for me and sometime it does not...

(But if you are ok with it I would like to merge this PR and fix this in another one...)

searx/static/themes/simple/src/less/style.less
```diff
@@ -330,38 +334,38 @@ article[data-vim-selected].category-social {
...
-    // color: @color-result-image-span-font;
-    background-color: var(--color-result-image-span-background-hover);
-    font-size: 0.7em;
+    font-size: 0.9rem;
+    color: var(--color-result-image-span-font);
+    padding: 0.5rem 0 0 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    max-width: 12rem;
```
dalf: It may not work with narrowed image.

A possible solution:
```less
.result-images {
  & > a {
    position: relative;

    & > span.title {
	  position: absolute; // combined with relative above, width: 100% doesn't overflow the a element.
   	  width: 100%;
    }
  }
}

```

mrpaulblack: Thanks; I updated the commit with these changes included :+1:

mrpaulblack:
I also changed mobile and made the row height smaller to 6rem and included the title as well on mobile; see my updated image in the Pr description...

dalf:
The last image is too wide on wide screen:
image

<details>

<summary>A patch to fix `!general !images lanscape`, not really clean but it works</summary>

/!\ There is a `<style>...</style>` in searx/templates/simple/results.html to move to the .less sources.

```patch
diff --git a/searx/results.py b/searx/results.py
index 6ab751c5..673b8c1b 100644
--- a/searx/results.py
+++ b/searx/results.py
@@ -264,12 +264,18 @@ class ResultContainer:
             result['parsed_url'] = result['parsed_url']._replace(scheme="http")
             result['url'] = result['parsed_url'].geturl()
 
+        # avoid duplicate content between the content and title fields
+        if result.get('content') == result.get('title'):
+            del result['content']
+
+        # make sure there is a template
+        if 'template' not in result:
+            result['template'] = 'default.html'
+
         # strip multiple spaces and cariage returns from content
         if result.get('content'):
             result['content'] = WHITESPACE_REGEX.sub(' ', result['content'])
 
-        return True
-
     def __merge_url_result(self, result, position):
         result['engines'] = set([result['engine']])
         with self._lock:
diff --git a/searx/templates/simple/results.html b/searx/templates/simple/results.html
index 72e10e07..b84922fd 100644
--- a/searx/templates/simple/results.html
+++ b/searx/templates/simple/results.html
@@ -16,7 +16,7 @@
 {% if results and results|map(attribute='template')|unique|list|count == 1 %}
   {% set only_template = 'only_template_' + results[0]['template']|default('default')|replace('.html', '') %}
 {% else %}
-  {% set unique_template = '' %}
+  {% set only_template = '' %}
 {% endif %}
 
 <div id="results" class="{{ only_template }}">
@@ -130,14 +130,19 @@
     </div>
     {% endif %}
 
+    <style>
+      .template_group_images {
+        display: flex;
+        flex-wrap: wrap;
+      }
+    </style>
+
     <div id="urls">
     {% for result in results %}
+        {% if result.open_group and not only_template %}<div class="template_group_{{ result['template']|replace('.html', '') }}">{% endif %}
         {% set index = loop.index %}
-        {% if result['template'] %}
-            {% include get_result_template('simple', result['template']) %}
-        {% else %}
-            {% include 'simple/result_templates/default.html' %}
-        {% endif %}
+        {% include get_result_template('simple', result['template']) %}
+        {% if result.close_group and not only_template %}</div>{% endif %}
     {% endfor %}
     {% if not results and not answers %}
         {% include 'simple/messages/no_results.html' %}
index 3df9f287..34904d55 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -696,6 +696,8 @@ def search():
     request.timings = result_container.get_timings()  # pylint: disable=assigning-non-slot
 
     # output
+    current_template = None
+    previous_result = None
     for result in results:
         if output_format == 'html':
             if 'content' in result and result['content']:
@@ -731,6 +733,17 @@ def search():
                 else:
                     result['publishedDate'] = format_date(result['publishedDate'])
 
+        #
+        if current_template != result['template']:
+            result['open_group'] = True
+            if previous_result:
+                previous_result['close_group'] = True
+        current_template = result['template']
+        previous_result = result
+
+    if previous_result:
+        previous_result['close_group'] = True
+
     if output_format == 'json':
         x = {
             'query': search_query.query,
```

</details>

[EDIT] the patch breaks the keyboard navigation, but I think this is right path to fix the issue.

return42:
@mrpaulblack is it OK when I rebase this PR?

mrpaulblack:
Yeah go ahead; I am just waiting for review :+1:

return42:
thanks! .. do you know how https://github.com/searxng/searxng/issues/368 is related to this PR? .. is #386 obsolete with this PR or do they handle different tasks?

EDIT: in the project board https://github.com/searxng/searxng/projects/1 #368 is for review but this PR is not listed on the board.

mrpaulblack:
Well this PR implements #368. In this PR  I dropped the JS image lib and replaced it completely with CSS. The issue that gets fixed in this PR is part of the board and therefore the PR that links to the issue.

So if this gets merged we can close #368 basically.

return42 (changes requested):
Wow looks really good!

I like the description below the thumb ..

I also tested images in the general category by adding (e.g.) artic to the general category .. (_mixed results_)

```yaml
  - name: artic
    engine: artic
    shortcut: arc
    timeout: 4.0
    categories: general
```


I checked the _mixed results_ on a desktop & a small touch device .. everything works except two issue I have seen:

grafik

The image in the last row is shown landscape, the origin is portrait ...

grafik

The other issue I have seen is with the forward/backward functionality ..

When I click on the last image in the _mixed results_ from the first screenshot above the image is shown, when I now press the forward button once the _image-view_ is closed, because it is the last image of this series.  What me wonders .. when I press the backward button I can press this button endless without the _image-view_ closes.

The behaviour is hard to describe with my pure English .. Can you please test the forward/backward buttons in a _mixed result_, then you will see what I mean.

mrpaulblack:
@dalf @return42 let me know if you have an idea why unit tests is failing.

I tried to debug:
Currently the build fails at
```
  File "/home/local/code/external/searxng/tests/unit/test_webapp.py", line 118, in test_search_html
    result = self.app.post('/search', data={'q': 'test'})
```

This is the error:
```
  File "/home/local/code/external/searxng/searx/templates/simple/results.html", line 6, in top-level template code
    <input type="hidden" name="engine_data-{{ engine_name }}-{{ k|e }}" value="{{ v|e }}" />
  File "/home/local/code/external/searxng/searx/templates/simple/base.html", line 56, in top-level template code
    {% block content %}
  File "/home/local/code/external/searxng/searx/templates/simple/results.html", line 136, in block 'content'
    {% include get_result_template('simple', result['template']) %}
  File "/home/local/code/external/searxng/searx/webapp.py", line 370, in get_result_template
    themed_path = theme_name + '/result_templates/' + template_name
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'template'
```

I think this is happening because some results are not getting a template name added, even tho it is in results.py Line 286. So when jinja2 calls get_result_template('simple', result['template']) in results.html Line 136 it can not pass on the param to the function and the whole thing crashes?

return42:
> I think this is happening because some results are not getting a template name added, 

The mock-up results from the unit test do not have a template .. see my WIP c4b495ad

mrpaulblack:
@dalf @return42 ready for review :D

dalf:
I like the facts:
* we don't rely on JS to display the images.
* the label is displayed at bottom of each thumbnail.

However I'm not sure how it is possible to avoid cropping:
image

The selected thumbnail is cropped at the bottom and at the top.

With `object-fit: fill;` the aspect ratio is lost:
image

I don't know how to fix the PR to solve these issues (thumbnail with the same aspect radio and not cropped)

---

[EDIT] The query to get the same images : `!images landscape`, but any query will show the issue

mrpaulblack:
Looking at https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit the only way for us to have the proper aspect ratio while not cropping the images at all is to use:

```css
object-fit: contain;
```

This way there is a letter box around the object to make it keep its aspect ratio, be contained in the available space, while keeping the aspect ratio and the height of 12rem. The problem is the width: 100% on the image with this tho, because when the height is defined and the aspect ratio also, then with width: 100% the width should be of the image, but instead its putting either a letter box around the image or cropping it...

I will try around with this a little and lets see if I find out how to fix this...

dalf:
To sum up:
* it works without javascript, a really nice improvement for the no javascript version
* the layout is faster (as soon as a row of images is ready, the images are displayed with the right layout)
* the titles at the bottom of the images are great too
* the images are cropped at the top and bottom (sometimes not at all, sometimes a lot according to the other images)

IMO, the last point really the UX:
* there are some duplicate images in the results: it makes them harder to find.
* the user is not sure about the aspect ratio. For example if you look for a book cover, it is easy to skip images that don't match the aspect ratio of a book cover. With this PR, the user has to focus on each image or even look at the full size image.

If there are no objections, I'm going to merge the pull request so we can get feedback from the users.
In a second time, I may restore a JS version on top of this change.

ping @return42

dalf (approved):

mrpaulblack:
Thanks! The JS implementation can just adjust the height of each article IMO to set a height for each row and then everything else should just work with the current CSS to keep aspect ratio...

PR #695, opened 2022-01-05
[fix] startpage engine / modified API

### LINKED ISSUE #692: Bug: startpage engine does not show any results
Branch: master
Version: 1.0.0-1392-00d3a747

 searxng-docker

**What happened?**
The startpage engine returns no results (but no specific error message)

**How To Reproduce**
!sp test 

I have tried on other instances (older or even searx instances) and I get the same thing. Maybe a change on the startpage side that prevents parsing the results?

return42:
## What does this PR do?

- [fix] startpage engine - avoid captcha
- [pylint] Startpage engine

Startpage has introduced new anti-scraping measures that make SearXNG instances run into captchas:

1. some arguments has been removed and a new `sc` has been added.
2. search path changed from `do/search` to `sp/search`
3. POST request is no longer needed

    make run

try to query `!sp foo`

## Related issues

Closes: https://github.com/searxng/searxng/issues/692

### DIFF
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -1,17 +1,20 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Startpage (Web)
+
 """
- Startpage (Web)
-"""
+
+import re
 
 from urllib.parse import urlencode
-
-from lxml import html
-from dateutil import parser
-from datetime import datetime, timedelta
-import re
 from unicodedata import normalize, combining
+from datetime import datetime, timedelta
+
+from dateutil import parser
+from lxml import html
 from babel import Locale
 from babel.localedata import locale_identifiers
+
 from searx.utils import extract_text, eval_xpath, match_language
 
 # about
@@ -135,10 +138,11 @@ def response(resp):
 
 # get supported languages from their site
 def _fetch_supported_languages(resp):
-    # startpage's language selector is a mess
-    # each option has a displayed name and a value, either of which may represent the language name
-    # in the native script, the language name in English, an English transliteration of the native name,
-    # the English name of the writing script used by the language, or occasionally something else entirely.
+    # startpage's language selector is a mess each option has a displayed name
+    # and a value, either of which may represent the language name in the native
+    # script, the language name in English, an English transliteration of the
+    # native name, the English name of the writing script used by the language,
+    # or occasionally something else entirely.
 
     # this cases are so special they need to be hardcoded, a couple of them are mispellings
     language_names = {
@@ -152,7 +156,15 @@ def _fetch_supported_languages(resp):
     }
 
     # get the English name of every language known by babel
-    language_names.update({name.lower(): lang_code for lang_code, name in Locale('en')._data['languages'].items()})
+    language_names.update(
+        {
+            # fmt: off
+            name.lower(): lang_code
+            # pylint: disable=protected-access
+            for lang_code, name in Locale('en')._data['languages'].items()
+            # fmt: on
+        }
+    )
 
     # get the native name of every language known by babel
     for lang_code in filter(lambda lang_code: lang_code.find('_') == -1, locale_identifiers()):
@@ -177,8 +189,8 @@ def _fetch_supported_languages(resp):
         if isinstance(lang_code, str):
             supported_languages[lang_code] = {'alias': sp_option_value}
         elif isinstance(lang_code, list):
-            for lc in lang_code:
-                supported_languages[lc] = {'alias': sp_option_value}
+            for _lc in lang_code:
+                supported_languages[_lc] = {'alias': sp_option_value}
         else:
             print('Unknown language option in Startpage: {} ({})'.format(sp_option_value, sp_option_text))
 
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -95,7 +95,7 @@ def get_sc_code(headers):
             # suspend startpage API --> https://github.com/searxng/searxng/pull/695
             raise SearxEngineResponseException(
                 suspended_time=7 * 24 * 3600, message="PR-695: query new sc time-stamp failed!"
-            )
+            ) from exc
 
         sc_code = href[5:]
         sc_code_ts = time()
@@ -107,10 +107,19 @@ def get_sc_code(headers):
 # do search-request
 def request(query, params):
 
+    # pylint: disable=line-too-long
+    # The format string from Startpage's FFox add-on [1]::
+    #
+    #     https://www.startpage.com/do/dsearch?query={searchTerms}&cat=web&pl=ext-ff&language=__MSG_extensionUrlLanguage__&extVersion=1.3.0
+    #
+    # [1] https://addons.mozilla.org/en-US/firefox/addon/startpage-private-search/
+
     args = {
         'query': query,
         'page': params['pageno'],
         'cat': 'web',
+        # 'pl': 'ext-ff',
+        # 'extVersion': '1.3.0',
         # 'abp': "-1",
         'sc': get_sc_code(params['headers']),
     }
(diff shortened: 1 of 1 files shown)


searx/engines/startpage.py
```diff
@@ -46,24 +51,23 @@
...
-    params['method'] = 'POST'
-    params['data'] = {
+    args = {
         'query': query,
         'page': params['pageno'],
         'cat': 'web',
-        'cmd': 'process_search',
-        'engine0': 'v1all',
+        # 'abp': "-1",
+        'sc': 'Mj4jZy61QETj20',
```
return42: This `sc` argument is new.  I copied the value from an interactive startpage session.  I don't know what is encoded in the string.  This string might be invalid at one day.

EDIT: 

- In an interactive startpage session, the value of `sc` changes on each request (incl. _next_ page request).
- May be the IP of the SearXNG server, combined with a date signature is encoded in this string / we need some test from different IPs

not-my-profile: Yeah, I think we should test this string again in 24 hours. We could also retrieve the `sc` string on startup by extracting it from the `hidden` input at https://startpage.com/ (requesting that page returns a new `sc` parameter each time).

dalf: in case the `sc` parameter needs to be fetched: gigablast.py fetches the extra parameters every hour
https://github.com/searxng/searxng/blob/00d3a7477c01981a6ef2d396332670ad5b831fa6/searx/engines/gigablast.py#L81-L84

return42: I installed the patch on my server / lets see if this link will work in 24h https://darmarit.org/searx/search?q=%21sp%20Salvadore%20dali

return42: FYI `sc` is no longer valid / I will look for a solution like @dalf suggested in https://github.com/searxng/searxng/pull/695#discussion_r778790532

return42: To get an actual `sc` time stamp from startpage's home page I pushed b3ea061 .. from my local IP it works, but I did not have luck when I tested this patch on my server .. I have to dig a little bit deeper :-(

not-my-profile: Maybe they already put your server IP on some kind of blacklist so that it doesn't get new `sc` values?

return42: This is what I also assume .. I will wait a day and see if it works, if not I have to analyze what the reason is.

Stratos42: > To get an actual `sc` time stamp from startpage's home page I pushed [b3ea061](https://github.com/searxng/searxng/commit/b3ea061a608af854614da067f83789f0001c68ff) .. from my local IP it works, but I did not have luck when I tested this patch on my server .. I have to dig a little bit deeper :-(

@return42 I just tried your patch on my server and it seems to work !

return42: @Stratos42 thanks for your feedback ... I did not have luck on my server, may be IP is blocked since a longer time .. may be something other went wrong .. before I merge this PR I will invest in more research .. 

Please keep me informed if the state of your server change / thanks!

not-my-profile:
>The startpage's API has changed

`!sp foo` still works for me on master ... which I guess means that the API in fact did not change (contrary to what your commit message says)? how am I supposed to check what this PR improves?

return42:
> which I guess means that the API in fact did not change (contrary to what your commit 

Can't say why it works for you, but others have issues with the old API / see https://github.com/searxng/searxng/issues/692

The API has changed, you can see it when you analyse the request and response of an interactive search session at startpage.com ... use developer tools from your browser and do some _reverse engineering_, than you will see, that a `sc` value is backpack on every search request.

not-my-profile:
Right ... can you update your first commit? Because currently it is misleading ... it claims to fix the problem while it actually does not. (Merging your 3rd commit into your 1st should do the job).

Saying the API has changed kinda implies that the old one isn't working anymore which isn't the case.

Maybe change "The API has changed" to "Startpage has introduced new anti-scraping measures that make SearXNG instances run into captchas".

return42:
> can you update your first commit? Because currently it is misleading 

I rebased branch and BTW fixed the commit message.

FYI: a haven't finished my work on this PR .. I have to test 21e884f on my server / which might take a week or longer.

not-my-profile:
To clarify: I meant that your first commit doesn't fix it because it just hardcodes an invalid `sc` parameter ... (so I think your `get an actual sc argument` commit should be merged with your first fix commit)

return42:
You should have write access to the branch of this PR .. feel free to squash, reorder as you like .. just leave comment when you have done / thanks!

return42:
@dalf may you can shoot an answer from the hips (don't invest to much time) ...

In 21e884f I added a suspend time of 7 days

```python
raise SearxEngineCaptchaException(suspended_time=7 * 24 * 3600)
```

Do you know if WSGI is able to store this for 7 days?

I use the uwsgi config from the templates:

https://github.com/searxng/searxng/blob/6420322cd1a22da50d7af9c0bb1b17bd1f370777/utils/templates/etc/uwsgi/apps-available/searxng.ini#L86-L87

I do not have a special apache setup / I just using the one from the templates:

https://github.com/searxng/searxng/blob/6420322cd1a22da50d7af9c0bb1b17bd1f370777/utils/templates/etc/httpd/sites-available/searxng.conf:uwsgi#L25

Here is the issue with suspend time on my server: https://darmarit.org/searx/stats?engine=startpage

My aim is to get my servers IP unlocked from startpage / I made some tests last weekend and it seems my IP is somewhere on the _abuse_ list of startpage.com.

not-my-profile:
>You should have write access to the branch of this PR 

Should this work via `pull/695/head`? cause it doesn't seem to work for me.

return42:
When I review, I typically add the remote repo (e.g. your https://github.com/not-my-profile/searxng/) to my remotes, I checkout the branch of the PR, edit or add new commits and then I push the modifications to the remote branch (e.g. on your https://github.com/not-my-profile/searxng/)

Does it work for you, when you add `git@github.com:return42/searxng.git` to your remotes and (force) push on fix-sp branch?

-----

I know that I can also pull to review a PR by:

    pull upstream refs/pull/695/head

But I have never tested if I can push on that.

----

Write access is given to the maintainers by a checkbox in the sidebar of a PR:

grafik

dalf:
> Do you know if WSGI is able to store this for 7 days?

This is not stored using `searx.shared`. It should.

Current implementation:
* https://github.com/searxng/searxng/blob/7cdb86ba951766d807130f4e8af5190789c2e690/searx/search/processors/abstract.py#L70
* https://github.com/searxng/searxng/blob/7cdb86ba951766d807130f4e8af5190789c2e690/searx/search/processors/abstract.py#L23-L32

So if there are 4 workers (for example), each worker will see a CAPTCHA and then stop sending queries for 7 days as long the server is not restarted.

dalf:
It works except this: it uses the default network instead of the startpage one and I can't figure out why.

See the log:
```
DEBUG   searx.network.default         : HTTP Request: GET https://startpage.com/ "HTTP/1.1 200 OK" (text/html; charset=utf-8)
DEBUG   searx.engines.startpage       : new value is: 1RYqsdYWAfEJ20
DEBUG   searx.network.default         : HTTP Request: GET https://startpage.com/sp/search?query=time&page=1&cat=web&sc=1RYqsdYWAfEJ20&language=francais&lui=francais "HTTP/1.1 200 OK" (text/html; charset=utf-8)
DEBUG   searx.webapp                  : http%3A//localhost%3A8888/search%3Fq%3D%2521sp%2Btime uses locale `fr`
```

It should not be `searx.network.default` but `searx.network.startpage`.

return42:
> it uses the default network instead of the startpage one and I can't figure out why.

Line 87 in function `get_sc_code(headers):`: 

```python
resp = network.get(base_url, headers=headers)` 
```

https://github.com/searxng/searxng/pull/695/files#diff-5a07bba22897433118bad961f089c47ae0446fe9fe75a0e1b95300957c82916cR87

dalf:
I have pushed a commit to fix the issue described in #762

dalf (approved):
Thank you!

PR #699, opened 2022-01-05
[mod] adjust dockerfiles/uwsgi.ini

dalf:
## What does this PR do?

* create %k process (%k = number of core), instead of the hard coded value 4.
* create 4 threads per process.
* increase the request buffer size.
* enable `auto-procname`: automatically set processes name to something meaningful

More threads per process helps the embedded image proxy.
Only `uwsgi.ini` from the docker image is replaced. If this is confirm, other uwsgi.ini version have to be updated.

* `make docker` (check the image name)
* start the image https://docs.searxng.org/admin/installation-docker.html#searxng-searxng
* enable the embedded image proxy
* check everything works fine

## Author's checklist

Open question (perhaps another issue): should the uwsgi log be displayed on `/dev/stdout` like many other Docker images?

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -4,7 +4,8 @@ uid = searxng
 gid = searxng
 
 # Number of workers (usually CPU count)
-workers = 4
+workers = %k
+threads = 4
 
 # The right granted on the created socket
 chmod-socket = 666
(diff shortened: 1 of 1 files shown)


mrpaulblack (approved):
Since these changes (except the increased buffer) are already running in my fork for months now -> LGTM :+1: 

IMO increased buffer can be integration tested

dalf:
> IMO increased buffer can be integration tested

Does you mean to add a test?

mrpaulblack:
Well no; I meant that if it is going to cause problems on the long run a instance maintainer like Unixfox or me is going to notice and we can roll back; Otherwise its OK IMO

[EDIT] I probably should not have said integration testing; Because I actually meant that this can be only tested for sure, by running it for a while on an instance IMO

[EDIT2] If you want to I can run this for a week on my fork... (Otherwise I think merge is ok)

PR #700, opened 2022-01-05
[mod] GitHub workflow: use cache

dalf:
## What does this PR do?

GitHub workflow: 
* use cache to speed up build
* documentation: fetch the whole git history, so the Documentation version is `1.0.0-1041-5cb20c67` instead of `2022.01.05.0ebad82`.

See above

N/A

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -56,6 +56,17 @@ jobs:
       uses: actions/checkout@v2
     - name: Install Ubuntu packages
       run: sudo ./utils/searx.sh install buildhost
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+        architecture: 'x64'
+    - name: Cache Python dependencies
+      id: cache-python
+      uses: actions/cache@v2
+      with:
+        path: ./local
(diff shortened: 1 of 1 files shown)


return42 (approved):
The title does not say it but I assume it is about caching python environment for the build of the "Documentation" .. and the caching is similar to what we have in the _python matrix_ (`python-version:  ["3.7", "3.8", "3.9", "3.10"]`)

I'm not familiar with GH caching, but this PR LGTM .. and it seems to work https://github.com/searxng/searxng/pull/700/checks#step:5:1

Lets merge.

OT: is there a way to cache the whole Linux distro?

return42:
> * documentation: fetch the whole git history, so the Documentation version is `1.0.0-1041-5cb20c67` instead of `2022.01.05.0ebad82`.

Sorry I do not understand this sentence .. can you give me some additional hints?

dalf:
> OT: is there a way to cache the whole Linux distro?

I haven't found how to do that.

> Sorry I do not understand this sentence .. can you give me some additional hints?

image

You can see the HTML title of page: `Welcome to SearXNG &#8212; SearXNG Documentation (2022.01.06.93c6829)` 

With the propose change this title will be something like `Welcome to SearXNG &#8212; SearXNG Documentation (1.0.0-1041-5cb20c67)` which match the version shown at the bottom of the SearXNG pages.

PR #705, opened 2022-01-06
[fix] prevent leaflet.{css,js} from polluting git grep results

### LINKED ISSUE #704: `leaflet.js` files pollute `git grep <term> searx` searches
I like to search our code with `git grep <term> searx` ... while my Git is clever enough to detect some of our vendored files as binary files, e.g:

```
Binary file searx/static/themes/oscar/img/searxng.svg matches
Binary file searx/static/themes/oscar/js/bootstrap.bundle.min.js matches
Binary file searx/static/themes/oscar/js/bootstrap.min.js matches
Binary file searx/static/themes/oscar/js/jquery.min.js matches
```

for some reason though `searx/static/themes/oscar/js/leaflet.js` is not detected as a binary file.

You can reproduce this by running `git grep zoom searx` in the root directory of the repository.

I'm using the latest Git 2.34.1.

not-my-profile:
The files polluted "git grep <term> searx" results.

Fixes #704.

### DIFF
--- a/searx/static/themes/.gitattributes
+++ b/searx/static/themes/.gitattributes
@@ -1,0 +1,2 @@
+leaflet.css -diff
+leaflet.js -diff


return42:
Question: Is it possible to add a `.gitattribute` file in the folder that contain these files? .. and in the root we have only wildcards like `*.xyz`?

not-my-profile:
Yes that would work ... I don't really see the point in having multiple `.gitattributes` files though ... it just makes it more difficult to check if something is already ignored.

return42:
> I don't really see the point in having multiple `.gitattributes` files

It is the same with "ignore": we have multiple `.gitignore` files and we have one at the root folder.

I like to have relevant sources (even if it is a context like git) at one point or at least "near by".  That's why I think it is good to have _wildcards_  somewhere in the root 

https://github.com/searxng/searxng/blob/93c6829b27d5319ca85af798015f1e11da566f1f/.gitignore#L3-L6

and explicitness near by the files which are the targets:

https://github.com/searxng/searxng/blob/93c6829b27d5319ca85af798015f1e11da566f1f/searx/static/themes/simple/.gitignore#L1-L2

not-my-profile:
>I like to have relevant sources (even if it is a context like git) at one point or at least "near by".

Fair enough. I moved the attributes to `searx/static/themes/.gitattributes` (this is the closest I can put them without needing to repeat them (since the files are in both the simple and the oscar themes).

return42 (approved):
Works like a charm / thanks!

```
  $ git grep leaflet searx/static/themes/
searx/static/themes/.gitattributes:leaflet.css -diff
searx/static/themes/.gitattributes:leaflet.js -diff

... no more binary in the list ...

searx/static/themes/simple/src/js/main/mapresult.js:        var map = L.map(leaflet_target);
```

PR #706, opened 2022-01-06
[fix] remove broken ? search operator

### LINKED ISSUE #702: `?` search syntax is broken
The [user documentation says](https://docs.searxng.org/user/search_syntax.html):
  
>**Prefix: `?`**
>to add engines and categories to the currently selected categories

But to me it is unclear how it is different from the `!` bang syntax. The following queries all yield exactly the same results:

* `!bing test !images`
* `!bing test ?images`
* `?bing test !images`

Only `?bing test ?images` for some reason only yields images??? :confused: 

Can somebody explain to me why somebody would want to use `?` over `!`?

If there is no satisfying answer to that question, I vote that we remove (or at least deprecate) the `?` syntax.

not-my-profile:
The ? search operator has been broken for some time and
currently only raises the question why it's still there.

## Context ##

The query "Paris !images" searches for "Paris" in the "images" category.

Once upon a time Searx supported "Paris ?images" to search for "Paris"
in the currently enabled categories and the "images" category.

The feature makes sense ... the ? syntax does not.
We will hopefully introduce a +!images syntax in the future.

Fixes #702.

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -302,7 +302,7 @@ engine is shown.  Most of the options have a default value or even are optional.
   search engine.
 
 ``shortcut`` :
-  Code used to execute bang requests (in this case using ``!bi`` or ``?bi``)
+  Code used to execute bang requests (in this case using ``!bi``)
 
 ``base_url`` : optional
   Part of the URL that should be stable across every request.  Can be useful to
--- a/docs/user/search_syntax.rst
+++ b/docs/user/search_syntax.rst
@@ -14,9 +14,6 @@ Prefix ``!``
 Prefix: ``:``
   to set language
 
-Prefix: ``?``
-  to add engines and categories to the currently selected categories
-
 Abbrevations of the engines and languages are also accepted.  Engine/category
 modifiers are chainable and inclusive (e.g. with :search:`!it !ddg !wp qwer
 <?q=%21it%20%21ddg%20%21wp%20qwer>` search in IT category **and** duckduckgo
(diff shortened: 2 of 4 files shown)


return42:
OT: but while we are modifying query.py and we need to test, does it make sense that we BTW add this file to the pylint process by adding a comment ..

```diff
git diff
diff --git a/searx/query.py b/searx/query.py
index f5f62882..cf5717ab 100644
--- a/searx/query.py
+++ b/searx/query.py
@@ -1,4 +1,10 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=missing-class-docstring
+"""Module that implements parsers to parses the *query* from different contexts.
+
+FIXME: .. add here some words about ':' (lang), '!' (engine, categ) and the parsers? ...
+"""
```

With the `# lint: pylint` addition we have a more elaborate pylint check ...

```
make test.pylint
TEST      [pylint] $PYLINT_FILES
************* Module searx.query
searx/query.py:93:12: C0103: Variable name "lc" doesn't conform to '(([a-z][a-zA-Z0-9_]{2,30})|(_[a-z0-9_]*)|([a-z]))$' pattern (invalid-name)
searx/query.py:134:12: C0103: Variable name "lc" doesn't conform to '(([a-z][a-zA-Z0-9_]{2,30})|(_[a-z0-9_]*)|([a-z]))$' pattern (invalid-name)
searx/query.py:199:8: R1715: Consider using dict.get for getting values from a dict if a key is present or a default if not (consider-using-get)
```

There is no need to do that in this PR ... but I often add files to pylint when we have to touch and test them anyway / it is a good opportunity.

not-my-profile:
Thanks, I didn't know about `# lint: pylint`! I'll still pass on this opportunity for now ... I already have too many other changes on my mind.

return42 (approved):

PR #708, opened 2022-01-06
Refactor `preferences`

not-my-profile:
## What does this PR do?

Fixe some hacks & misplaced inheritance and adds type hints.

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -8,8 +8,14 @@
 from base64 import urlsafe_b64encode, urlsafe_b64decode
 from zlib import compress, decompress
 from urllib.parse import parse_qs, urlencode
+from typing import Iterable, Dict, List, Set
+from dataclasses import dataclass
+
+import flask
 
 from searx import settings, autocomplete
+from searx.engines import Engine
+from searx.plugins import Plugin
 from searx.locales import LOCALE_NAMES
 from searx.webutils import VALID_LANGUAGE_CODE
 from searx.engines import OTHER_CATEGORY
@@ -21,31 +27,20 @@ ENABLED = 1
 DOI_RESOLVERS = list(settings['doi_resolvers'])
 
 
-class MissingArgumentException(Exception):
-    """Exption from ``cls._post_init`` when a argument is missed."""
-
-
 class ValidationException(Exception):
 
-    """Exption from ``cls._post_init`` when configuration value is invalid."""
+    """Exption from ``cls.__init__`` when configuration value is invalid."""
 
 
 class Setting:
     """Base class of user settings"""
 
-    def __init__(self, default_value, locked=False, **kwargs):
+    def __init__(self, default_value, locked: bool = False):
         super().__init__()
         self.value = default_value
         self.locked = locked
-        for key, value in kwargs.items():
-            setattr(self, key, value)
 
-        self._post_init()
-
-    def _post_init(self):
-        pass
-
-    def parse(self, data):
+    def parse(self, data: str):
         """Parse ``data`` and store the result at ``self.value``
 
         If needed, its overwritten in the inheritance.
@@ -59,7 +54,7 @@ class Setting:
         """
         return self.value
 
-    def save(self, name, resp):
+    def save(self, name: str, resp: flask.Response):
         """Save cookie ``name`` in the HTTP reponse obect
 
         If needed, its overwritten in the inheritance."""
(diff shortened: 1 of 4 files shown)


return42 (commented):
Wow another gorgeous work from you .. thanks!

I just want to leave some comments from my review .. and before we merge I have a question, see below ..

----

from the both commits:

> Previously the Setting classes used a horrible _post_init hack that prevented proper type checking.
> [refactor] remove pointless tuple

Oh yes .. very old artefact from fe691a09 / thanks for refactoring this.

> [refactor] rename SwitchableSetting to FeatureSet

I think the name 'SwitchableSetting' is more precise here (even if that name sounds crude to me).

The *Switchable* indicates a on/off (enable/disable) state of such a setting.  Same with the names 'Choice' vs. 'Feature' .. I don't really like 'SwitchableSetting' and 'Choice' but I think 'Feature' and 'FeatureSet' sounds too generall in my ears.

This is just my opinion ... What do you think?

> [refactor] stop FeatureSet from subclassing Setting

I'm wondering about the old implementation (added in 99435381) .. now, when I look at, the old implementions looks crude to me. / thanks!

searx/preferences.py
```diff
@@ -200,19 +200,19 @@ def save(self, name: str, resp: flask.Response):
...
 
-    default_on: bool
     id: str
+    default_on: bool
 
 
-class SwitchableSetting(Setting):
-    """Base class for settings that can be turned on && off"""
+class FeatureSet(Setting):
+    """A set of features that can be either enabled or disabled."""
```
return42: What about `Choice` --> `Toggle` and `SwitchableSetting` --> `ToggleSet`?

not-my-profile:
> I think 'Feature' and 'FeatureSet' sounds too generall in my ears.

Valid criticism! I don't really like toggle or checkbox since these are UI metaphors. I have now settled with `BooleanChoices`.

Check out my last commit ... I've now refactored the `SwitchableSetting` class even more extensively.

searx/preferences.py
```diff
@@ -185,34 +177,34 @@ def parse(self, data):
 class MapSetting(Setting):
     """Setting of a value that has to be translated in order to be storable"""
 
-    def _post_init(self):
-        if not hasattr(self, 'map'):
-            raise MissingArgumentException('missing argument: map')
-        if self.value not in self.map.values():  # pylint: disable=no-member
+    def __init__(self, default_value, map: Dict[str, object], locked=False):  # pylint: disable=redefined-builtin
```
return42, line 180: sorry for bikesheeding again ...

`map` is a ~keyword~ builtin .. I'm often at the same situation where a name that really fits is a keyword or a builtin  e.g. `map` or `dict` .. I usually then choose a name like `kv` (key/value) ... `kvmap` ..

Here I would decide to simply use `values` .. what also fits to the usage:

grafik

compare with ...

grafik


But it is a nit .. if you think `map` is good enough leave it as is.

not-my-profile: nit: `map` isn't a keyword ... it's just a built-in function (otherwise this would be a syntax error).

Yeah I'm currently not really interested in fixing such pylint errors.

return42: > nit: map isn't a keyword ... it's just a built-in function (otherwise this would be a syntax error).

right. Sorry I'm fighting with my github editor .. I edited my initial comment ..

not-my-profile: >map is a keyword

[No it's not.](https://docs.python.org/3/reference/lexical_analysis.html#keywords) as I explained it's a [built-in function](https://docs.python.org/3/library/functions.html).

The difference is that `map = 1` works while e.g. `for = 1` is a syntax error.

(But yes I know it's bad practice to shadow built-ins ... it's just not important enough for me to fix right now)

return42 (approved):
Thanks for your clean up!

`map` is OK for me .. but `values` fits IMO better when we look at the usage of this argument .. and is not a builtin.

But as I said, it is just bikesheeding.

return42:
@not-my-profile if you have time, could you have a look at ... otherwise I can do it, but I don't have time today (sorry).

I assume that the refactoring of the `default_on` caused that all properties are stored and not only the change ones / https://github.com/searxng/searxng/issues/746#issuecomment-1011960917

If I'm not wrong this is responsible for the issues reported here:

- https://github.com/searxng/searxng/issues/746
- https://github.com/searxng/searxng/pull/738

PR #709, opened 2022-01-06
[fix] drop etools engine module

### LINKED ISSUE #696: Bug: etools engine
Branch: master
Version: 1.0.0

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Percentage: 100

return42:
## What does this PR do?

[fix] drop etools engine module

The implementation of the etools engine is poor.  No date-range support, no
language support .. since etools is just a metasearch engine which uses engines
we already have in SeaarXNG I vote to drop this engine entirely from the source
tree. [1]

[1] https://github.com/searxng/searxng/issues/696#issuecomment-1005855499

## Related issues

Closes: https://github.com/searxng/searxng/issues/696


not-my-profile:
>etools is just a metasearch engine which uses engines we already have in SeaarXNG

Actually now that I looked into it, according to its website etools does search 9 search engines that we don't support otherwise:

* https://de.ask.com/
* https://www.base-search.net/
* https://www.exalead.com/search/
* https://www.fastbot.de/
* https://search.lilo.org/
* https://www.moose.at/
* https://web.search.ch/
* https://tiger.ch/ (ironically yet another metasearch engine)
* https://yandex.com/

I agree that ideally we would support these engines directly but right now we don't and dropping etools would mean dropping support for these engines, so I'm not sure about this.

return42:
> I agree that ideally we would support these engines directly but right now we don't and dropping etools would mean dropping support for these engines, so I'm not sure about this.

formally correct, but I see this from a practical standpoint:

1. etools is broken / there is a captcha blocking the requests and I don't know a way around this captcha
2. etools "as is" does not have date-range support, nor language support .. what me brings to the quality ..
3. Quantity is not quality: etools will return results which do not fit e.g. to the selected language ..

In practice: in SearXNG the _worse_ etools results will be mixed with _good_ results from other engines we have (as long as there is no captcha).  

At best case, what we win with etools is in e.g. results from https://de.ask.com/ in a query from a german request .. in all other cases _worse_ results are bubble up in SearXNG's result list.

IMO if there is a request for one of the engines we do not support right now, we should better invest our time in implementing a module for this engine.

not-my-profile:
> etools will return results which do not fit e.g. to the selected language ..

Why don't we just add `language: de-CH` to etools? This way it would only show up specifically for people looking for that language.

return42:
> This way it would only show up specifically for people looking for that language.

etools has a lot of options incl. language support ..

grafik

not-my-profile:
>etools has a lot of options incl. language support ..

Yes but our implementation doesn't support it right now. So we should categorize our implementation according to the settings it has hardcoded.

return42:
> Yes but our implementation doesn't support it right now. So we should categorize our implementation according to the settings it has hardcoded.

Again .. less quality and broken by captcha .. I don't want to maintain such lovelessly implemented search engines .. if you want to maintain this engine please send your alternative PR and close this one.

not-my-profile:
Ok, you're argument that it's broken and not worth fixing is convincing. Please update your commit message to include that reasoning (it currently doesn't mention the Captcha situation or why no language support is a problem).

>etools is just a metasearch engine which uses engines
we already have in SeaarXNG

is wrong please correct that saying sth like "While etools does support several engines we currently don't support directly, support for  them should be added directly to SearXNG if there is demand."

not-my-profile (approved):

Issue #572, opened 2021-12-01
Simple theme: dark / light mode switch


dalf:
**Is your feature request related to a problem? Please describe.**

Be able to switch between light / dark / auto in `/` and `/search`.

**Describe the solution you'd like**
Something similar to https://spot.ecloud.global/ :

![image](https://user-images.githubusercontent.com/1594191/144191635-7ae94030-c846-4968-a708-b3c9669600fd.png)

Note about the spot implementation:
* there is no auto mode
* on mobile phone, the setting is hidden.

References:
* https://gitlab.e.foundation/e/cloud/my-spot/-/commit/f103ddca0254e00843fad60c7127721f19106757
* https://gitlab.e.foundation/e/cloud/my-spot/-/blame/master/searx/webapp.py#L843-851

**Describe alternatives you've considered**

**Additional context**

https://github.com/searxng/searxng/pull/521#issuecomment-974264824

joysriramsarkar:
not yet implemented.

vojkovic:
Now that we have the black theme (https://github.com/searxng/searxng/pull/3841) and potentially more themes in the future, I don't think a toggle like this makes sense anymore.

What do you think @return42, can we close the issue?

return42:
> What do you think [@return42](https://github.com/return42), can we close the issue?

Sure 👍

PR #713, opened 2022-01-07
Fix tab checkbox confusion

### LINKED ISSUE #498: tabs don't act like tabs without JavaScript

not-my-profile:
## What does this PR do?

See the commit messages.

## Related issues

Fixes #498


return42:
@not-my-profile FYI I rebased the PR (your branch)

return42:
@mrpaulblack @dalf .. could you have a look at this PR .. IMO checkboxes for categories give bad UX ...

grafik

Beside it is not rendered well in different screen-size, it is hard for me to switch the category from e.g. "general" to "images" .. if I'm not wrong, I need at least three clicks:

- click to enable images categ
- click to disable general categ
- click on the magnifier

In master branch I just need a single click on the `images` tab (categ)

not-my-profile:
> IMO checkboxes for categories give bad UX

No having stuff that doesn't look like checkboxes act like checkboxes gives bad UX. This PR fixes the UX problem.

>Beside it is not rendered well in different screen-size

I cannot reproduce this.

> I need at least three clicks [..] In master branch I just need a single click on the images tab (categ)

My PR doesn't change how many clicks it takes. I also cannot reproduce this.

return42:
> ... maybe your rebase messed up my PR?

I did not change anything functional, I just rebased and made a new `make static.build.commit` commit on top .. usual procedure when there is a conflict in the client (/static) sources .. but I can't exclude side-effects when we merge this PR with the PRs from you of the last (two) days.  Could you please have a look / thanks.

not-my-profile:
Nvm, I also cannot reproduce the problems you observed with your rebased branch. Maybe your browser still has the old CSS cached? Or can you try another browser? And tell me exactly how to reproduce the problems you experience?

return42:
> And tell me exactly how to reproduce the problems you experience?

In your developer studio, activate "touch device" grafik

return42:
> No having stuff that doesn't look like checkboxes act like checkboxes gives bad UX. This PR fixes the UX problem.

And the price are 3 clicks to switch from one to another categ?

not-my-profile:
>And the price are 3 clicks to switch from one to another categ?

No as I explained my PR is not meant to change how many clicks it takes. If it does so that's a bug that I'll fix.

dalf:
> @mrpaulblack @dalf .. could you have a look at this PR .. IMO checkboxes for categories give bad UX ...
>  I need at least three clicks: ...

this is the default behavior since searx exists, especially when javascript is disabled.
that's why there is a [search_on_category_select.js](https://github.com/searxng/searxng/blob/master/searx/static/plugins/js/search_on_category_select.js) plugin.

The plugin send the query as soon as a category is selected or any other fields is changed (language, date range, safe field).

---

Note about the simple theme:
It doesn't include search_on_category_select.js because:
  * it relies on jQuery which is not include in the simple theme (main reason).
  * the HTML elements are different.

<details>
<summary>related source code</summary>

  * https://github.com/searxng/searxng/blob/ced656606fbc7a5d5dda8b1e4efe16a9eb88e57b/searx/templates/simple/base.html#L22-L25
  * https://github.com/searxng/searxng/blob/ced656606fbc7a5d5dda8b1e4efe16a9eb88e57b/searx/static/themes/simple/src/js/head/00_init.js#L18-L22
  * https://github.com/searxng/searxng/blob/ced656606fbc7a5d5dda8b1e4efe16a9eb88e57b/searx/static/themes/simple/src/js/main/search.js#L90-L110
</details>

---

I've never used this feature: when I need multiple categories in result page I usually use:
* either the bangs. For example `!general !images something`. 
* either I look one category and then another one (because it requires to disable the plugin, save the preferences, send the query, checks the results, go back to preferences, change the settings again, and finally save).

IMO, the default behavior should be the "category on select", so the UX would be same when JS is disabled.
And if this feature is really important, then the HTML should be changed when "category on select" is disabled.

not-my-profile:
>because it requires to disable the plugin, save the preferences, send the query, checks the results, go back to preferences, change the settings again, and finally save

The obvious solution to that is to allow the plugin to be quickly toggled from the results page.

return42:
> The obvious solution to that is to allow the plugin to be quickly toggled from the results page.

If we add a toggle to the result page, we should also consider that we possibly need more toggles / I have https://github.com/searxng/searxng/issues/572 in mind.

mrpaulblack:
I dont really understand this PR; It adds the actual checkbox UI for the categories on the result page with JS enabled as well as JS disabled; Also this breaks search on select, which means that the user needs to select the category and than start a search manually.

I would suggest the following (personal opinion):
* when JS enabled -> UI should be same as in current master
* when JS disabled than there are checkboxes in the same colomn as the label and visible on the left on LTR and on the right on RTL

(Also the search on select does not work because it expects the help text? `if (qinput !== null && d.querySelector('.help') != null && searxng.search_on_category_select)`)

not-my-profile:
Ah yeah my bad ... I missed that the simple theme does not use the plugin JS ... I guess I'll have to fix that first (I opened #769 to track that).

PR #721, opened 2022-01-07
Add user documentation "How to search"

not-my-profile:
## What does this PR do?

Replaces the previous [search syntax](https://docs.searxng.org/user/search_syntax.html) documentation with a [more thorough How to search](https://share.push-f.com/searxng/how-to-search/user/how-to-search.html) documentation.

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -353,6 +353,30 @@ engine is shown.  Most of the options have a default value or even are optional.
    engines, and so won't be described here.
 
 
+Example: Multilingual Search
+----------------------------
+
+SearXNG does not support true multilingual search.
+You have to use the language prefix in your search query when searching in a different language.
+
+But there is a workaround:
+By adding a new search engine with a different language, SearXNG will search in your default and other language.
+
+Example configuration in settings.yml for a German and English speaker:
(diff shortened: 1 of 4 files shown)


return42:
@dalf @mrpaulblack  I don't know what is better ..

My very personal opinion: 

commits e401a384 and 73382661 are worth to merge. About 

- 6559a017 [doc] create "How to search" user documentation

I have a doubt. IMO it is to text-heavy and users might prefer the more simple description we currently have .. but as I said; I don't know.

not-my-profile:
>I don't know what is better ..

We can have both!

>IMO it is to text-heavy and users might prefer the more simple description we currently have

The current documentation fails to explain many important details that aren't obvious.

But you're right that we should also have a short syntax reference somewhere.
I think ideally it should be shown as a popup in the UI when clicking a link like `Syntax?` next to the search box.
When clicked it would show a brief syntax overview in a popup, like what we currently have:

image

and then there would be a `Learn more` link that links to the more extensive user documentation I introduce in this PR.

not-my-profile:
Ok, I'm closing this for now. I think we should address #722 first.

dalf:
A feature without documentation: https://github.com/searx/searx/pull/1640

PR #726, opened 2022-01-07
[fix] Documentation of the builtin plugins

### LINKED ISSUE #637: Documentation of the builtin plugins is broken
The documentation of the builtin plugins is broken, see https://searxng.github.io/searxng/admin/plugins.html

return42:
## What does this PR do?

[fix] Documentation of the builtin plugins

     make docs

Have a look in the admin section, in master the documentation is missed: https://docs.searxng.org/admin/plugins.html

## Related issues

Closes: https://github.com/searxng/searxng/issues/637

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,6 +3,7 @@
 
 import  sys, os
 from pallets_sphinx_themes import ProjectLink
+from flask import Flask
 
 from searx import get_setting
 from searx.version import VERSION_STRING, GIT_URL, GIT_BRANCH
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -41,6 +41,7 @@
 import searx.plugins
 import searx.webutils
 searx.engines.load_engines(searx.settings['engines'])
+searx.plugins.initialize(None)
```
not-my-profile: `app=None` would be more readable, otherwise LGTM

return42: Even if it is not needed by the plugins we have right now, I decided to import the Flask app from webapp.py / @not-my-profile what do you think (613fb155)?

PR #729, opened 2022-01-09
fix(security): filtered_engines were not correctly evaluating validate_token()

k2s:
## What does this PR do?

This PR fixes coding error in lambda that was used to filter engines protected with tokes.

It would be good to write tests at least for /stats and /stats/errors in the future.

Engines protected with `tokens` parameter were always available in Preferences page due to error in lambda. 

It was also available to all users on `/stats` and `/stats/errors` endpoints which is security issue.


return42 (approved):
Thanks for your contribution!

// seems issue has been added in 7acd7ffc0

PR #734, opened 2022-01-10
Small fixes to Search syntax doc

not-my-profile:


return42 (approved):
Makes sense .. pushed 44118e9c on top / drop this commit if you think it is not needed.  Merge PR if you think it is ready to merge.

not-my-profile:
Thanks! I dropped the admonition:

> Configuring additional engines like shown in the example above will release additional HTTP request to this engines.

because I think it's too much of a blanket statement. The config example doesn't show categories ... if they are in different categories there won't be additional HTTP requests ... and it of course also depends on if the engines are enabled or disabled.

PR #737, opened 2022-01-10
[fix] autocomple.js: register li.onmousedown instead li.onclick

### LINKED ISSUE #542: [simple theme] autocomplete results not clickable

return42:
## What does this PR do?

[fix] autocomple.js: register li.onmousedown instead li.onclick

I don't really know why the `li.onclick` event is not fired in Chrome, but `li.onmousedown` seems to work.  May be it is related to the outer `<form>`?

Activate autocompletion and try to click one item from the suggestions.

## Related issues

Closes: https://github.com/searxng/searxng/issues/542


tiekoetter:
Autocomplete now also works correctly in Chrome / Chromium (Linux [Arch], Windows and macOS) and Safari (iOS & macOS).

I tested this on a local instance running on return42/searxng (fix-542).

mrpaulblack (approved):
LGTM works on mobile chrome and safari as well as firefox desktop :+1:

PR #738, opened 2022-01-10
[fix] filtron.service: set -read-buffer-size 65536

return42:
## What does this PR do?

[fix] filtron.service: set -read-buffer-size 65536

## Why is this change important?

Increase filtron buffer size to prevent `Too big request header` errors.

See https://github.com/searxng/filtron/issues/6#issuecomment-1009367442

## How to test this PR locally?

To test a filtron installation [`sudo -H ./utils/filtron.sh install all`](https://docs.searxng.org/utils/filtron.sh.html) is needed.

I tested this installation on my public SearXNG instance.

----

@dalf do we need to fix [docker-compose.yaml `filtron: command:`](https://github.com/searxng/searxng-docker/blob/c7b3f004eb5b304a5695dccea2d7c975a6cc6c43/docker-compose.yaml#L32) also?

PR #749, opened 2022-01-13
[fix] make BooleanChoices only set cookies that vary from the default

### LINKED ISSUE #746: Searxng 1.0.0-1487-687bdef4 can't save preferences after updating
Hi all.

After updating SearxNG to 1.0.0-1487-687bdef4, it seems impossible to save the preferences.
The preference pages opens correctly, but upon pressing "Save" the attached message appears from nginx (page unavailable).

It was working correctly before the update (sorry, can't remember which version I was using back then).

I guess it's not really a bug, but most likely some kind of misconfiguration on my part.
May someone suggest me what to check to make it work again?

SearxNG was installed using the installation script on Arch Linux, without Morty or Filtron, as this is a purely personal instance.

I can give the URL of the instance, if needed.

Thanks a lot in advance.

EDIT: nginx logs say the following

` *36 upstream sent too big header while reading response header from upstream, client: 147.xxx.xxx.xxx, server: searx.xxx.xyz, request: "POST /preferences HTTP/1.1", upstream: "uwsgi://unix:/run/uwsgi/app/searx/socket:", host: "searx.xxx.xyz"
`

image

not-my-profile:
The bug was inadvertently introduced by the refactor in 180d4d0.

## Related issues

- https://github.com/searxng/searxng/issues/746
- https://github.com/searxng/searxng/pull/738

Closes: #746

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -205,6 +205,7 @@ class BooleanChoices:
         self.name = name
         self.choices = choices
         self.locked = locked
+        self.default_choices = dict(choices)
 
     def transform_form_items(self, items):
         # pylint: disable=no-self-use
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks a lot for your fast response!

Issue #369, opened 2021-10-01
Simple theme: keyboard shortcuts: allow translation


dalf:
When "Vim-like hotkeys" is activated:
![image](https://user-images.githubusercontent.com/1594191/135592550-f500258d-f5c0-4f1b-8ebf-fac00de146b3.png)

It is possible to display a help message pressing the `h` key:
![image](https://user-images.githubusercontent.com/1594191/135592678-7641c133-8a57-440d-bbe4-00e69e77fac3.png)

See:
https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/static/themes/simple/src/js/main/searx_keyboard.js#L22-L119

These string should go to https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/webapp.py#L394-L400

See also for usage example:
https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/static/themes/simple/src/js/main/searx_preferences.js#L13

return42:
I'm not experienced ... does it make sense to have a `translation/<component>` URL architecture?  I mean something similar to what we have here:
https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/webapp.py#L1152-L1158

and on the client side ..

https://github.com/searxng/searxng/blob/458575c221abcdbe3c0ffb633dd06f5f6d05bf2e/searx/static/themes/simple/src/js/main/searx_preferences.js#L4-L9

PR #750, opened 2022-01-13
Simple theme: implement RTL

### LINKED ISSUE #747: RTL layout requires a few corrections
**How To Reproduce**

Switch to Arabic, Persian or Hebrew

**Expected behavior**

1) Preferences link should be at the left side.
2) The send search button and search form are properly placed; It's only the frame shapes that need to be Flipped Horizontally.

**Screenshots & Logs**

image

dalf:
## What does this PR do?

This PR implements the RTL text direction in the simple theme.

Modern browsers do most of the work, but margin / padding / border and some other properties have to be manage manually.

The usual method seems to be:
```css
#q {
  border-right: none;
}

#q [dir="rtl] {
  border-right: inherit;
  border-left: none;
}
```

IMO, this is error prone and require some copy / paste 

There is the [`margin-inline-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start ) property, but it seems to be supported only recently (except for FF), and it doesn't solve all cases (for example `float: right`).

So this PR uses some LESS mixins:

In LTR version:
```less
.ltr-left(@offset) {
  left: @offset;
}
```

In this RTL version:
```less
.ltr-left(@offset) {
  right: @offset;
}
```

and so on for the different CSS properties.

Add support for RTL languages in the Simple theme.

image

* check everything in a LTR language
* check everything in a RTL language

## Author's checklist

Issue list:
* [x] URL in the result
* [x] Vim hotkeys help
* [x] LTR text in the result are displayed correctly.

For the last one: I dont' know if there is a way to display LTR text aligned on the left, and RTL text aligned on the right only using CSS.

Currently, the only way I see is to [detect the text direction in Python](https://docs.python.org/3.8/library/unicodedata.html#unicodedata.bidirectional), and then add the `dir` HTML attribute.

From that, it is possible to add some CSS to display the text either aligned on the right, either aligned on the left.

ping @GreenLunar 

## Related issues

Close https://github.com/searxng/searxng/issues/747


searx/static/themes/simple/gruntfile.js
```diff
@@ -117,7 +117,7 @@ module.exports = function (grunt) {
           outputSourceFiles: true,
         },
         files: {
-          "css/searxng.min.css": "src/less/style.less",
+          "css/searxng.min.css": "src/less/style-ltr.less",
```
return42: I miss the `src/less/style-ltr.less` file .. or do I misunderstand this?

```
Running "less:production" (less) task
>> Destination css/searxng.min.css not written because no source files were found.
>> 1 stylesheet created.
>> 1 sourcemap created.
```
I wondering that the build `make themes.simple` ends without an error .. :confused:

dalf: Face palm, I forgot to add this file. Fixed.

mrpaulblack:
@dalf I added some small fixes on top of your branch; feel free to change/throw away/whatever :+1: 

The things that still need attention:
* about page has a `direction=ltr` on it; changing it to rtl works. but punctuation moves to the left on English text, which it should not. The solution would be a translation of that section in all RTL languages so the punctuation matches the text or to only align the text to the right without changing from LTR to RTL
* the vim help is still LTR

dalf:
The URL are displayed left to right but aligned on the right.

Google displays each URL "part" from right to left: 
image

Bings displays the full URL from left to right (or is it broken?):
image

The Google version requires to modify [get_pretty_url](https://github.com/searxng/searxng/blob/599d882d0a43b91d6c700edb2f39a900eabbdd5c/searx/webapp.py#L420-L424) and the CSS (I'm not sure how to do that).

> the vim help is still LTR

https://github.com/searxng/searxng/issues/369 needs to be fixed first

searx/static/themes/simple/.stylelintrc.json
```diff
@@ -1,3 +1,7 @@
 {
-  "extends": "stylelint-config-standard"
+  "extends": "stylelint-config-standard",
+  "rules": {
+    "declaration-empty-line-before": null,
+    "no-invalid-position-at-import-rule": null
+  }
```
return42, line 6: IMO this should be done in a separated commit, with a commit message why we want to activate:

- https://stylelint.io/user-guide/rules/list/declaration-empty-line-before/
- https://stylelint.io/user-guide/rules/list/no-invalid-position-at-import-rule/

searx/static/themes/simple/src/less/style.less
```diff
@@ -33,6 +33,7 @@
   vertical-align: bottom;
   line-height: 1;
   text-decoration: inherit;
+  .ltr-transform();
```
return42, line 36: > In LTR version:

```less
.ltr-left(@offset) {
  left: @offset;
}
```

> In this RTL version:

```less
.ltr-left(@offset) {
  right: @offset;
}
```

> and so on for the different CSS properties.

Why do we choose the prefix `.ltr-` when it stands for RTL and LTR replacements, isn't it better we use a prefix like `.text-direction-` or _shorter_ `.direction-` or anything other that fits better?  .. The prefix `.rtl-` might confuse here a litle bit.

return42 (approved):
Except `/preferences` the RTL text orientation LGTM.

I think this PR is a huge improvement to what we have.

I left some comments with my thoughts ...

searx/static/themes/simple/src/less/style.less
dalf: My reasoning:
* We need to describe one version: either the RTL either the LTR version. From this version, the LESS mixins can create the other version.
* `.ltr-left(@offset)` means `left: @offset` in the RTL version --> when `rtl` then `left` (and adjust when `ltr`) --> `.rtl-left`.
  The idea is make clear that `left` applies to the RTL version and it has to be changed in the LTR version.

`direction-left` is telling `left` for the RTL version implicitly or it can tell `left` for LTR version and `right` in the RTL version.

searx/static/themes/simple/.stylelintrc.json
dalf: With `declaration-empty-line-before` enabled:
```less
#results.image-detail-open article.result-images[data-vim-selected] .detail {
  display: flex;
  flex-direction: column;
  position: fixed;
  .ltr-left(60rem);
  .ltr-right(0);
  top: @search-height;
```

has to be written
```less
#results.image-detail-open article.result-images[data-vim-selected] .detail {
  display: flex;
  flex-direction: column;
  position: fixed;

  .ltr-left(60rem);

  .ltr-right(0);
  top: @search-height;
```

---

With `no-invalid-position-at-import-rule` disabled, it allows to have these lines after the mixin declarations:
https://github.com/searxng/searxng/blob/739f0355f9f0120791f6f59ed0c5499097cf20bf/searx/static/themes/simple/src/less/style-rtl.less#L57

https://github.com/searxng/searxng/blob/739f0355f9f0120791f6f59ed0c5499097cf20bf/searx/static/themes/simple/src/less/style-rtl.less#L57

I can create a commit for this change.

return42: > I can create a commit for this change.

this would be cool / thanks a lot!

dalf: Done.
I've merged the other commits (I have a backup if for reason it was not a good choice).

return42 (approved):
I think this PR is a huge improvement to what we have.

Lets merge this PR, the `/preferences` page can be fixed in another PR.

> I have a backup if for reason it was not a good choice

from my side is no longer need for.

mrpaulblack (approved):
LGTM :+1: 

(We should work on making LTR text in an RTL layout to be displayed as LTR in another PR)

GreenLunar:
Looks awesome!

GreenLunar:
> Bings displays the full URL from left to right (or is it broken?):

In bing, it's reversed (that is, not rtl-wise)

> The URL are displayed left to right but aligned on the right.
> 
> Google displays each URL "part" from right to left: image
> 
> Bings displays the full URL from left to right (or is it broken?): image
> 
> The Google version requires to modify [get_pretty_url](https://github.com/searxng/searxng/blob/599d882d0a43b91d6c700edb2f39a900eabbdd5c/searx/webapp.py#L420-L424) and the CSS (I'm not sure how to do that).
> 
> > the vim help is still LTR
> 
> #369 needs to be fixed first

GreenLunar:
Looks awesome!

> Bings displays the full URL from left to right (or is it broken?):

In bing, it's reversed. URL must always be LTR and NOT affected by RTL

PR #754, opened 2022-01-14
[theme] /preferences: rename "Method" to "HTTP Method"

### LINKED ISSUE #751: /preferences: rename "Method" to "HTTP Method".
image

Should Method be renamed "HTTP Method" ?

Source code:
* https://github.com/searxng/searxng/blob/74847fe9029187e792d44eba49ef5a2cbc6c514a/searx/templates/oscar/preferences.html#L262
* https://github.com/searxng/searxng/blob/master/searx/templates/simple/preferences.html#L237

return42:
Closes: https://github.com/searxng/searxng/issues/751

### DIFF
--- a/searx/templates/oscar/preferences.html
+++ b/searx/templates/oscar/preferences.html
@@ -259,7 +259,7 @@
                 <fieldset>
                     <div class="container-fluid">
                         {% if 'method' not in locked_preferences %}
-                        {% set method_label = _('Method') %}
+                        {% set method_label = _('HTTP Method') %}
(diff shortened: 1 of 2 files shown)


not-my-profile:
How about also fixing the description? "Search language" isn't right.

return42:
> How about also fixing the description? "Search language" isn't right.

Oops .. / thanks for pointing out / fixed in 08b38a8b, copied the description from oscar theme.

not-my-profile:
I think while we have the Oscar theme it should be changed there as well for consistency.

return42:
> I think while we have the Oscar theme it should be changed there as well for consistency.

Sorry I don't know what you mean by .. both themes have the same "HTTP Method" and the same description.

not-my-profile:
Ah, my bad ... I misread the diff.

Issue #722, opened 2022-01-07
Integrate the user documentation into the application


not-my-profile:
Currently the user documentation lives at https://docs.searxng.org/user/. Most SearXNG instances do host their own documentation. There are several problems with that:

1. Currently users are unlikely to find the user documentation.
   (https://docs.searxng.org/ is only linked from the bottom of the About page)

2. When they do find the user documentation it might not fit with the version of SearXNG they are using.
    (because some instances of SearXNG run an outdated version)

3. The official user documentation cannot link example queries to the search page since it doesn't know which instance the user is using. (#659)

4. The user documentation cannot embed information about the instance (e.g. how many engines it has).

5. The user documentation cannot be translated via Weblate to match the UI language.

Moving the user documentation into the application would solve all of these problems.
While we're at it I think we should convert the user documentation to Markdown because I think it's easier to integrate into a web application than reStructuredText. (RST is mainly useful for documenting API which we don't need for user documentation)

return42:
solved by https://github.com/searxng/searxng/pull/930

PR #758, opened 2022-01-15
Introduce `searx/help/`

not-my-profile:
## What does this PR do?

See the commits. This is the first PR to address #722.

## Why Markdown?

* Translators on Weblate are more likely to know Markdown than they're to know reST (since Markdown is used on GitHub, Reddit, Stackoverflow) etc.
* Weblate is likely to gain a preview feature https://github.com/WeblateOrg/weblate/issues/3106.


searx/webapp.py
```diff
@@ -1329,6 +1332,9 @@ def page_not_found(_e):
     return render('404.html'), 404
 
 
+userdoc.render(app)
```
dalf: Nit: the userdoc is initialized twice in debug mode. To avoid that, and to group the initialization code in place, can you move this call just after `search_initialize(...)` (line 1352).

dalf:
from commit 9b9fd802b8c7b78fc2cf863aa8dcc87b1a485495 in this PR:

> Wrapping every String into {{ _('...') }} to enable translation would of course be too cumbersome (and also not flexible enough). 

👍 

> We however also don't want to execute strings from Weblate as Jinja templates since that would be a security nightmare.

👍 

How to translate this about page? With multiple files? (`about.en.md`, `about.de.md`... )

not-my-profile:
>How to translate this about page? With multiple files?

My initial idea was to just do it via Babel's `gettext`. Can Babel map an identifier to a string? So that if there's a minor change in the English version the translated versions aren't lost?

return42:
I don't know where my fail is, but I got broken links ...

    make run

the *Stats page* links to:  `http://127.0.0.1:8888/url_for:stats` 

grafik

I would expected `http://127.0.0.1:8888/about`

not-my-profile:
> I got broken links

Nice catch! That's my fail .. I wrote `[Stats page](url_for:stats)` instead of `[Stats page][url_for:stats]`. I'll force push a fix (along with fixing the double initialization mentioned by dalf).

dalf:
> My initial idea was to just do it via Babel's gettext. 
> Can Babel map an identifier to a string?

Is the idea to assign an identifier to the whole content of `about.md`?

Babel doesn't seem to support multiline content:
* Weblate shows one line of text not multilines
* https://github.com/elixir-gettext/gettext/issues/229

not-my-profile:
>Is the idea to assign an identifier to the whole content of about.md?

Yes that was my idea. Apparently Weblate calls such identifiers "keys".

>Babel doesn't seem to support multiline content

I'm not well versed with neither Babel nor Weblate but I think both support multiline content. The issue you linked is just about the encoding of the newlines ... I don't think it matters for us since I don't think we care much about how the `.po` files look ... encoding newlines as `\n` should still work fine for our purposes.

Anyway before we enable translation I think we should firstly improve our English user documentation ... so I think we should wait with enabling translation for a future PR anyway.

dalf:
> I think both support multiline content.

yes, I misread sorry.

And there is a "new line button" in Weblate: 
image

Related issue in Weblate: https://github.com/WeblateOrg/weblate/issues/2110

> I think we should wait with enabling translation for a future PR anyway.

Do you intend to add other markdown pages?

not-my-profile:
>Do you intend to add other markdown pages?

Yes. I envision the next steps to be:

1. convert the user documentation currently in `docs/user/*.rst` to Markdown files in `searx/userdoc/` and add a new route / UI to access the user documentation right from the user interface
2. improve & extend the existing user documentation
3. enable translation

return42:
I think this PR is going into the right direction, what we need are the abilities

- to translate entire files written in a plain text markup
- to included these files as pages in the WEB app.
- to include runtime variables of the WEB app in theses files

This is needed independent from the content (about-Page-, help-Page ..) we have one day (I have a doubt that the about page will "as is" for a longer time ;)

This PR fulfill all of the points listed above.  There is only one point I see a problem:

> convert the user documentation currently in docs/user/*.rst to Markdown files in searx/userdoc/ and add a new route / UI to access the user documentation right from the user interface

We have already a plain text markup ([reST](https://docs.searxng.org/dev/reST.html)), we should not add another markup (Markdown). Instead of using [mistletoe](https://pypi.org/project/mistletoe/) we can use [docutils](https://pypi.org/project/docutils) to generate HTML from single pages.  If I have the time today, I will prepare a patch today

not-my-profile:
I really dislike the reST link syntax, I find it very awkward, so I'd prefer us to use Markdown for the userdoc.

I also think that Markdown would help with translation since translators are more likely to be familiar with Markdown than with reST.

>we should not add another markup (Markdown)

I really don't see the problem. We are all already familiar with Markdown (it's what we use to write GitHub comments here).

return42:
> I really dislike the reST link syntax, I find it very awkward, so I'd prefer us to use Markdown for the userdoc.

I have a different experience about reST vs Markdown markup, I think it is a question of what one is used to use.  One big pro I see on the reST side; reST has one definition while there are several Markdown dialects out in the wild.

> I really don't see the problem. We are all already familiar with Markdown (it's what we use to write GitHub comments here).

I am not very familiar with Markdown, often I stumble with github UI, Matrix UI .. , because all use different dialects and support a different range of markups.

Adding a second plaint text format also means we need to maintain second tool chain and require additional tools (docutils is already in our tool chain, since it is a requirement of Sphinx).

not-my-profile:
> I think it is a question of what one is used to use

Exactly ... and my point is that more people are used to Markdown than there are people who have ever used reST.

> reST has one definition while there are several Markdown dialects out in the wild

There is a standard for Markdown called [Commonmark](https://commonmark.org/). It is used by GitHub, Reddit and Stack Overflow. I chose mistletoe because it implements Commonmark, meaning it behaves just like you're used to from these big platforms. (Granted GitHub does add a couple of extensions to Commonmark but mistletoe does support the important one which is tables).

>we [also] need to maintain second tool chain

There is no tool chain (tool chain implies several tools involved which isn't the case here). It's just a single dependency in `requirements.txt` and a single line to convert Commonmark to HTML. We don't need to maintain anything about that.

return42:
> Exactly ... and my point is that more people are used to Markdown than there are people who have ever used reST.

Not in this project is my point.  We used to use reST everywhere in the source and in the documentation  / [We at SearXNG are using reStructuredText (aka reST) markup for all kind of documentation](https://docs.searxng.org/dev/reST.html)

>  It's just a single dependency in requirements.txt and a single line to convert Commonmark to HTML. We don't need to maintain anything about that.

We need to maintain all our requirements and their sub-sub-requiments, this becomes very clear when I look at grunt for example, but we also had it in the past with docutils & sphinx and IMO it is only a question of time we will have to deal with this library.

ATM this PR uses only simple markups but in the future I expect we will add more and more functionality, e.g. figures or images ([Commonmark](https://spec.commonmark.org/0.30/#images) / [reST](https://docutils.sourceforge.io/docs/ref/rst/directives.html#image)) which can be classified.  I don't know how to solve such a request with a Commonmark toolchain, in the reST tool chain we can use `:class: user-doc-images` to classify and we can use LESS to format these class.

The above is only one example, we can assume other functional request that will come up to this markup and the produced HTML (& CSS). Even if it looks initial very simple, the toolchain will grow up .. at its worst we end up in a situation where we have to implement the same functionality in the context of two different markup tool chains.

I urge us to avoid such a constellation, it is a very big request from me / thanks

EDIT: it is not only me, also others on github recommend to use reST:

[ .. On the other hand, reStructuredText is more extensible and powerful, with native support (not just embedded HTML) for tables, as well as things like automatic generation of tables of contents. Unless you are embedding text into a programming language (especially Python doc comments, for which purpose reStructuredText was originally developed) you will probably choose to use Markdown for a small document (like this Gist once was). But as you add more sections and features, you may decide that you want the power of reStructuredText.](https://gist.github.com/dupuy/1855764#file-common-md-rst-md)

not-my-profile:
> Not in this project is my point.

I meant that translators from Weblate are way more likely to know Markdown than reST.

> We at SearXNG are using reStructuredText (aka reST) markup for all kind of documentation

Yes that's a documentation of the status quo that you wrote yourself (bee19a76f7dc8b86af3ef342acd3f4db72f51543), I don't see how it's relevant to the discussion of whether or not we should change the status quo.

>ATM this PR uses only simple markups

And we should keep it that way to facilitate translation. I am quite certain that we don't need any of the advanced features reST provides for our user documentation.

> it is not only me, also others on github recommend to use reST

Of course it's not just you ... though the gist you link completely disregards Commonmark and is thus largely irrelevant because Commonmark irons out nearly all of the original Markdown's quirks.

Btw Weblate will [probably soon support Markdown](https://github.com/WeblateOrg/weblate/issues/3106), which is yet another reason why we should use it over reST.

While I have not found your reasoning for reST convincing, I do understand that you're afraid that Commonmark won't suffice for our use cases and we'll end up extending it with custom code that we have to maintain ourselves. So how about the following?

We try using Commonmark for the user documentation and see how far we get ... if it turns out that we absolutely need a feature Commonmark doesn't provide (which I doubt) we convert our documentation to reST with Pandoc and continue using reST. Does that put your mind at ease?

return42:
> Yes that's a documentation of the status quo that you wrote yourself (bee19a7), I don't see how it's relevant to the discussion of whether or not we should change the status quo.

The reST markup is the default plain-text markup in the python ecosystem and the searx project uses reST from the beginning.

I wrote the majority of the documentation .. I wrote down what was a common agreement and bee19a7 has been approved.

>  I am quite certain that we don't need any of the advanced features reST provides for our user documentation.

I have a doubt .. the `:class: ..` directive I described above is obvious

> While I have not found your reasoning for reST convincing,

I gave hints in the comments above .. and it is not clear to me, for what reason we need to add Commonmark instead of using what we have.

Another drawback I see, it is not easy to C&P ore move partial documentation from one part of the source to another part of the sources.

It is OT here, but I think https://github.com/searxng/searxng/commit/07b7c917abd8bd7eb6729325b4c3e1bef0d1eb78 is an example how easy we can add tests to the CI when we stay with reST.  If we add Markdown (Commonmarkup) we also need to add implementations & tools to our tool-chains to validate the Commonmarkup

mrpaulblack:
I am on @not-my-profile site for this PR. I think markdown is a way better option for this than the `.rst` files. I personally for example did not want to touch our `README.rst` since I did not understand the syntax. I think in general a lot more people know markdown compared to reST and even tho I understand that you like reST more than markdown @return42, I still think markdown easier for other first time contributers than reST for example.

Since we want to AFAIK remove userdoc from SearXNG docs at `docs.searxng.org` we can keep the admin and dev docs in reST and change the userdoc to markdown as a compromise IMO.

return42:
> I personally for example did not want to touch our README.rst since I did not understand the syntax.

In the README.rst is a reST Markup that gihub understands / Github does not really compile or render reST.  It is made to layout good in the github HTML .. this file is a very worse example if we are talking about reST.

reST & Markdown are very, very, very simple and intuitive markups, they become more complex if you use more and more special markups.

> I think in general a lot more people know markdown compared to reST and even tho I understand that you like reST more than markdown @return42, I still think markdown easier for other first time contributers than reST for example.

But now you are turning the tables :-) .. It is not about what I like!!

@mrpaulblack you did not know reST because you are not familiar with the python ecosystem (AFAIK you do not implement python code).  I know that Markdown is most often the default in the JS area .. in python, reST is the default markup to document the project.

Saying _"he or she don't like/understand reST, so lets use Markdown"_, is like saying _"I don't like/understand python lets drop that slowly python crap and use go"_.

Adding a second technology if for the same purpose already one technology exists requires solid arguments.  The only argument I have read until here is, that some (not in this project working) people do not understand reST or don't like it. This is not really an argument in my ears.

Related: https://github.com/searxng/searxng-docker/pull/417#discussion_r2262939120

not-my-profile:
>reST [is] very, very, very simple and intuitive

Hahaha ... no. Please compare the following:

```
This is a paragraph that contains `a link <link target>`_.

.. _link target: https://domain.invalid/
```

with:

```
This is a paragraph that contains [a link][link target].

[link target]: https://domain.invalid/
```

You cannot seriously argue that the reST link syntax is intuitive. It is very much the opposite of intuitive.

>The only argument I have read until here is, that some (not in this project working) people do not understand reST or don't like it.

@mrpaulblack and I work for this project but I see that you're happy to disregard us.

For the last time ... the syntax should be accessible for **translators on Weblate**, who are most likely not involved with any Python programming and unfamiliar with reST. But I'm seriously tired of arguing about this with you since your just ignoring my main argument and pulling out one straw man after the other.

mrpaulblack:
Ok yeah I looked at some other projects at https://github.com/psf and https://github.com/python/ and they all use `.rst` files for markup; (At least for their doc), but there is also the theme with the CSS, JS and so on as well as the templates in html in our project which are all not python code.
I mean yeah this repo is mostly python code and the rest is only a small percentage compared, but I know that I and Martin both like markdown more (as you said since I am coming more from a webdev background, rather than a python coding background).
So can we just give this a try; We are a rolling release and if this is really not going to work than we can still switch. Its not like our userdoc is going to be really big or anything, so its easy to rewrite the markdown to `reST`, if this PR does not work out. Our general doc is still `reST`...

return42:
> Hahaha ... no. Please compare the following:

?!?!  don't lets start this benchmark .. I had this in 2016 on the linux kernel mailing list when we introduced a reST process for the linux kernel documentation / we had a lot of programmers arguing that this or that detail is "better" in Markdow .. both have their pros and cons .. overall reST markup has a definition from the beginning while there are many Markdown dialects, CommonMarkup tries to _normalize_ this situation .. but Markdown & the tools are available for, suffer from the circumstance that there was no schema definition at the beginning and users are not willing to agree on one / every parser has its own quirks when it comes to the more complex markups.

> @mrpaulblack and I work for this project but I see that you're happy to disregard us.

I have a different impression, my arguments are disregard by you ...

> the syntax should be accessible for translators on Weblate, who are most likely not involved with any Python programming and unfamiliar with reST. 

Sorry, if I was not clear at this point: I do not see the difference between reST and Markdown like you. If we do not use more markup that we have today, it is simple to write down in both markups.  If we add more functionality both become complex.

Thats why I came to the conclusion, that we can stay with reST and do not need additional tools or markups.

mrpaulblack:
OK tbh I dont feel that strongly about using markdown. `reST` also works for me and from what Markus said, its also well defined compared to markdown. So from my side its ok. I dont really touch the docs anyway and I think the decision should be with the people that are actually maintaining this part of the code. Which is IMO Markus And Martin ATM.

not-my-profile:
Thanks ... I really appreciate that you don't wait for @dalf to weigh in on this and just start changing my PR.

Anyway it was very much intentional of me to remove the jinja templating from the documentation file. We do not want remote code execution via Weblate.

return42:
To demonstrate what I think is a better solution, I rebased this branch an placed two patches on the top.  The two patches are from my https://github.com/return42/searxng/commits/web-user-doc-rst branch

@not-my-profile if you disagree you can drop these two patches and I will open an alternative PR.  

With the first patch d02bff7 we have reST files in the /userdoc folder, these files are jinja templates that are rendered in a flask response context.  I mean we can use jinja templating in the reST files, like we use it in the HTML templates.


With the second patch 4647046 a validation of the reST markup of the files matching `./searx/userdoc/*.rst` has been added to the CI tests.

mrpaulblack:
Well there are just added on top so you can easily remove those commits again. IMO this is common for searxng/searxng that we just push our changes on top in the PRs...

Lets just wait for @dalf to review and leave the PR be for the time being. :+1:

return42:
> Anyway it was very much intentional of me to remove the jinja templating from the documentation file. We do not want remote code execution via Weblate.

May I have overseen something .. you mean code injection from weblate's translations .. right?

EDIT: If so we can change d02bff7 to serve only selected replacements, but IMO an injection from weblate's translation is unlikely since these files are source code files we have an eye on.  For me the reST-template file is not different to a HTML template.

not-my-profile:
Yes the plan was that the files in `searx/userdoc/` can be translated via weblate in the future.

return42:
> Anyway it was very much intentional of me to remove the jinja templating from the documentation file. We do not want remote code execution via Weblate.

I don't know if it is really needed, but in https://github.com/return42/searxng/commit/30445a2bde4205f7c42ded177c5e92a715a3946d I prepared a patch too reduce the security risks.

dalf:
About the translations (not the format sorry), we will need help from translators: the content is intended to be read by the translators.

I've seen two ways to manage the translations of documentation:
* Translation per page. Example: MDN Mozilla:
  * a file https://github.com/mdn/content/blob/main/files/en-us/web/css/font-size/index.md
  * the French version: https://github.com/mdn/translated-content/blob/main/files/fr/web/css/font-size/index.md
* Translation per sentence. Example: Weblate documentation:
  * https://github.com/WeblateOrg/weblate/blob/main/docs/faq.rst
  * the first sentence: https://hosted.weblate.org/translate/weblate/documentation/de/?checksum=fe1b19fc034c7bf0&sort_by=-priority,position#translations
  * Sphinx extracts the sentences: https://www.sphinx-doc.org/en/master/usage/advanced/intl.html
	* It can lead to half translated pages: https://docs.weblate.org/fr/latest/user/translating.html , this is no go for me:
	   * Source code: ```Please see :ref:`workflows` for more info on translation workflow.```
	   * Output: ```Please see Flux de travail de traduction for more info on translation workflow.```
	* also either we call Sphinx at runtime, either after each change ( we add "make userdoc.build.commit" ).

So, I haven't found what we are looking for: use Weblate to translate the pages instead of the sentences. I see two reasons what it is not done:
* translators will have to deal with long text in a textarea, it will require scrolling up and down between the English and the translated version especially when there are some updates.
* if there are some markdown or rst, it will require a review because they won't have a HTML preview in Weblate.

To conclude, right now, I don't think Weblate is the solution to allow the translation of the user documentation, the MDN solution seems more suitable.

unixfox:
Hello,
I just wanted to add my feedback to this issue, I hope it's fine.
I much prefer Markdown too. Also, in the past I already wanted to edit the README myself and I was a bit terrified by the reST syntax.

dalf:
Ok, so if I have to pick a format: it is markdown.

How to translate:
* one directory per language. The file inside this directory.
* translators open PR, we validate.

So we can move the file to `searx/userdoc/en/...`.

return42:
> Ok, so if I have to pick a format: it is markdown.
> So we can move the file to searx/userdoc/en/....

If we write the "userdoc" there is no longer an argument to use reST in the /docs but not in the "userdoc" .. and there is no longer a security reason not using jinja templates.

BTW I do not understand why the majority have problems with reST compared to Markdown 

- [about.md](https://github.com/not-my-profile/searxng/commit/13e16375bdcdd4d57e477e45b7ed2a5c7d78b274#diff-e942f6836ff612dcbea57d072945f6db508b811c191e6a4faf23a55e0528aa09)
- [about.rst](https://github.com/searxng/searxng/commit/d02bff7b29c3a3e0f4135f260d3f0b0235d018b6#diff-3477b017a7843ac03f0419ef326994ab5cbff8d8243ce6b3883710a6a5124a19)

reST is much better to read in the plain-text format, especial when it comes to complex markup .. by example images in Markdown and a figure in reST ...

```markdown
![The San Juan Mountains are beautiful!](images/san-juan-mountains.jpg "San Juan Mountains")

```
AFAIK Narkdown does not know what the difference of an image compared to a figure is .. in reST there are images and figures, here is an example of a figure:

```reST
.. figure:: images/san-juan-mountains.jpg
   :alt: San Juan Mountains

   The San Juan Mountains are beautiful!
```

> I much prefer Markdown too. Also, in the past I already wanted to edit the README myself and I was a bit terrified by the reST syntax.

In deed the README is a very bad example ... it is the reST that guthub-ui is able to render and the README is made to look good in the github-ui.

Please do not think the README file is real reST.

----

We have a [reST primer](https://docs.searxng.org/dev/reST.html), we have a lot of knowledge & tools to work with, it is an extendable standard, the markup schema was defined from the beginning (where we have a lot of dialects of Markup & their tools) .. it is the python standard markup used for doc-strings, it is uses in projects like the kernel documentation, RTD is build up from reST ... how many arguments are more needed? And on the other side I only see the argument that people not working in this project don't like reST or prefer Markdown for any reason. 

I have implemented documentation frameworks for companies and Open-Source projects since 20 years .. I had use several Markups from XML to various plain-text (Markdown), in this years I tested a lot of tools .. all had their limits but reST tools have been with me over 15 years and have always offered me solutions that I have not found in other tools. 

With this experience in mind I am not really able to follow a decision against reST for documentation purpose.

dalf:
we have to make sure `[search engines][url_for:preferences]` is resolved with the right URL when searx is installed with a prefix in the URL ( `https://..../searx` )

searx/user_help.py
```diff
@@ -0,0 +1,39 @@
...
+    We render the user documentation once on startup to improve performance.
+    """
+
+    link_targets = {
+        'brand.git_url': GIT_URL,
+        'brand.public_instances': get_setting('brand.public_instances'),
+        'brand.docs_url': get_setting('brand.docs_url'),
+    }
+
+    with app.test_request_context():
```
dalf: To support installation in a non root path:
`with app.test_request_context(base_url=get_setting('server.base_url')):`

searx/user_help.py
```diff
@@ -0,0 +1,39 @@
...
+        link_targets['url_for:index'] = url_for('index')
+        link_targets['url_for:preferences'] = url_for('preferences')
+        link_targets['url_for:stats'] = url_for('stats')
+
+    define_link_targets = ''.join(f'[{name}]: {url}\n' for name, url in link_targets.items())
+
+    for filename in pkg_resources.resource_listdir(__name__, 'help'):
+        markdown = pkg_resources.resource_string(__name__, 'help/' + filename).decode()
+        markdown = define_link_targets + markdown
+        HELP[filename.split('.')[0]] = mistletoe.markdown(markdown)
```
dalf: `HELP[os.path.splitext(filename)] = ...` to deal with file name with multiple dots.

return42:
@dalf why are you silently ignore my arguments about the markup?

not-my-profile:
Probably because I have already addressed most of your arguments and this discussion is getting tiring.

> If we write the "userdoc" there is no longer an argument to use reST in the [admin/developer] docs

Not true. The admin & developer docs do benefit from the more advanced features of reST. The difference is that we're not planning on translating the admin & developer docs and that the user docs don't require such advanced features. We could look into using [MyST](https://myst-parser.readthedocs.io/) which is Markdown for Sphinx in the future if you want to, but right now this is really not a priority.

> there is no longer a security reason not using jinja templates

Not true. From a security perspective it's always desirable to minimize the area where malicious code could be introduced. And we also don't want to burden translators with Jinja if we don't have to.

Please accept the fact that the majority of people does not have 15 years of experience with reST like you do ... most people don't have any experience with reST. From the 5 people that participated in this discussion everybody except you prefers Markdown. I'd consider continuing this debate to be a waste of time because as you said yourself "it is a question of what one is used to use".

return42:
> reST is the python standard markup used for docs

> Adding a second plaint text format also means we need to maintain second tool chain and require additional tools (docutils is already in our tool chain, since it is a requirement of Sphinx).

> BTW I do not understand why the majority have problems with reST compared to Markdown
...
> reST is much better to read in the plain-text format, especial when it comes to complex markup .. by example images in Markdown and a figure in reST ...

However, if adding a second markup for the same purpose (documentation) is the decision, I cannot follow this conceptual error and will stop maintaining the documentation.

Maybe it is not wrong if this work is taken over by others and I dedicate myself to the topics that I enjoy more. There is enough work to do.

dalf:
@return42 I've read your mail

---

next steps are multi languages support and multi pages support.

Issue #681, opened 2022-01-03
Ideas for a new `settings.yml` format


not-my-profile:
This issue serves to collect ideas for an improved `settings.yml` format (as suggested by @dalf in https://github.com/searxng/searxng/issues/631#issuecomment-999371016).

Ideas so far:

* change `engines` from a list to a mapping using `name` as the key (since it has to be unique)
* rename `engine` property to `module` (because it's confusing that engines have engines)
  Maybe even support `somepackage.somemodule` to load modules outside from `searx.engines`?
* support multiple `shortcuts` per engine (see https://github.com/searx/searx/issues/2229)
* require that `categories` must be a list (currently strings are automatically turned into lists)
* deprecate `proxies` for `network.proxies`
* change `safesearch` (bool) to `content_safety` (safe | unsafe | configurable)  #644

The new format could look like:

```yml
engines:
  yahoo:
    module: yahoo
    enabled_by_default: false  # currently, the default value is double negation "disable: false"
    shortcuts:
      - yh
      - ㄱ
    network:
      proxies:  # no need to specify all:// : ....
        - socks5://127.0.0.1:1337  # socks5 always means DNS resolution made by the proxy
        - socks5://127.0.0.1:1338
```

dalf:
About the plugins (related to #759 )

```yaml
plugins:
  searx.plugins.self_info: 
    load: false
  searx.plugins.hash_plugin:
    enabled_by_default: false
    token: ['secret']  # enable with a token
  searx.plugins.hostname_replace:
    enabled_by_default: true
    config:  # configuration on this plugin (currently at the "root" of the document)
      '(.*\.)?youtube\.com$': 'invidious.example.com'
      '(.*\.)?youtu\.be$': 'invidious.example.com'
      '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
      '(.*\.)?reddit\.com$': 'teddit.example.com'
      '(.*\.)?redd\.it$': 'teddit.example.com'
      '(www\.)?twitter\.com$': 'nitter.example.com'
      # to remove matching host names from result list, set value to false
      # 'spam\.example\.com': false
  mypackage.hostname_replace_extended:
    config:
      file: /a/file/some/where/on/the/file/system
```

There is only one plugin list:
* not to load internal plugin
* to provide plugin configurations
* to load external plugins

not-my-profile:
I think we should use the opportunity to clarify `enabled` as `enabled_by_default`.

return42:
[...]

PR #759, opened 2022-01-15
[enh] add blocked_plugins setting do completly remove plugin from que…

k2s:
## What does this PR do?

There is already `enabled_plugins` and `disabled_plugins` they set the default way how plugins are used. It is not possible to block the plugin from processing queries;

This pull request adds `blocker_plugins` property to completely remove plugin from being used.
It is loading the plugin to obtain the plugin name familiar to users, but then it is not registered.

This pull request improves security and public hosting of searx:

* 'Self Informations' may display internal IP address if server is not correctly configured
* 'Hash plugin' uses server CPU and this functionality doesn't need to be necessary wanted what the public hoster want's to provide to users
* it is possible to disable all engines until valid token is provided, but users are always able to use the instance for some functionality provided by plugins

I am running public instance of seax for a group of people. I provided them with token which makes it for me possible to keep the instance performing well by controlling the number of users with access to my small instance.

Ucommend in settings.yml following:
```yaml
  blocked_plugins:
    - 'Self Informations'
```

Now ip` query should not be processed at all and the plugin will not be listed in /preferences page.

### DIFF
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -226,10 +226,9 @@ def plugin_module_names():
 
 
 def initialize(app):
-    # it is possible to block plugins from being registered into plugin chain
-    blocked_plugins = settings['outgoing'].get('blocked_plugins', [])
-
     for module_name, external in plugin_module_names():
-        plugin = load_and_initialize_plugin(module_name, external, (app, settings))
-        if plugin and plugin.name not in blocked_plugins:
-            plugins.register(plugin)
+        # it is possible to block plugins from being registered into plugin chain in settings.yml
+        if module_name not in settings['blocked_plugins']:
+            plugin = load_and_initialize_plugin(module_name, external, (app, settings))
+            if plugin:
+                plugins.register(plugin)
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -226,7 +226,10 @@ def plugin_module_names():
 
 
 def initialize(app):
+    # it is possible to block plugins from being registered into plugin chain
+    blocked_plugins = settings['outgoing'].get('blocked_plugins', [])
+
     for module_name, external in plugin_module_names():
         plugin = load_and_initialize_plugin(module_name, external, (app, settings))
-        if plugin:
+        if plugin and plugin.name not in blocked_plugins:
             plugins.register(plugin)
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -157,6 +157,12 @@ outgoing:
 #   - plugin2
 #   - ...
 
+# Plugins listed here will be loaded, but will not be registered into plugins chain.
+# The engine will not use them to process queries.
+#
+#  blocked_plugins:
+#    - 'Self Informations'
+
 # Comment or un-comment plugin to activate / deactivate by default.
 #
 # enabled_plugins:
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -160,8 +160,8 @@ outgoing:
 # Plugins listed here will be loaded, but will not be registered into plugins chain.
 # The engine will not use them to process queries.
 #
-#  blocked_plugins:
-#    - 'Self Informations'
+#blocked_plugins:
+#  - searx.plugins.self_info
 
 # Comment or un-comment plugin to activate / deactivate by default.
 #
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -212,6 +212,7 @@ SCHEMA = {
     },
     'plugins': SettingsValue(list, []),
     'enabled_plugins': SettingsValue((None, list), None),
+    'blocked_plugins': SettingsValue(list, []),
     'checker': {
         'off_when_debug': SettingsValue(bool, True),
     },


return42:
Not sure I have understood all, 

ATM we have a list of names stored under `enabled_plugins:`

https://github.com/searxng/searxng/blob/599d882d0a43b91d6c700edb2f39a900eabbdd5c/searx/settings.yml#L162-L173

This PR would add a list of names stored under `blocked_plugins`:

```yaml
blocked_plugins:
   - 'Self Informations'
```

would it help, if we have list of dictionaries

```yaml
enabled_plugins: 
    - 'Hash plugin' : 
         active: true
         tokens: []
```

Plugins are only shown in the preference page, when `active: true` is set, now matter what the defaults are today. 

I don't know if it is ever needed, but it might be useful if we can set a list of `tokens: []` (like known from the engines), which allows activation only for user request with a token in the list.

Drawback of such a solution: 

- downward compatibility of a /etc/searxng/settings.yml will be suffer.
- implementing `token` for plugins needs still more effort compared to the `blocker_plugins:` setting from this PR

dalf:
> 'Hash plugin' uses server CPU and this functionality doesn't need to be necessary wanted what the public hoster want's to provide to users

Note: If you enable the image proxy, for each image, the server computes and later checks a HMAC.

---

Currently, the server loads:
* the plugin in the `searx.plugins` module.
* the plugin referenced by `plugins` in `settings.yml`. Here the comments in `settings.yml` are misleading: this is a list of Python module name, NOT plugin name.

So, you can just delete the two plugins, there is nothing more to configure. Drawback: updates are more difficult to do.

From @return42  ideas, a possible enhanced syntax (related to #681): https://github.com/searxng/searxng/issues/681#issuecomment-1013686674

[settings_defaults.py](https://github.com/searxng/searxng/blob/master/searx/settings_defaults.py) can help to support the current format and a new one: if the current one is detected, it can be converted to the new one. So the code deals only with the new format.

There is no need to implement all of this now.

I see two ways:

### first solution
A possible interim solution is to support two formats:
```yaml
plugins: # list of external plugins
  - mypackage.hostname_replace_extended
```
and
```yaml
plugins: # list of external plugins
  searx.plugins.self_info:
    load: false
  mypackage.hostname_replace_extended:
```

I'm "meh" about this, I prefer a clean new format rather than slowly adding new converters.

### second solution

This PR and then take time to think about a new format.

searx/settings.yml
```diff
@@ -157,6 +157,12 @@ outgoing:
 #   - plugin2
 #   - ...
 
+# Plugins listed here will be loaded, but will not be registered into plugins chain.
+# The engine will not use them to process queries.
+#
+#  blocked_plugins:
+#    - 'Self Informations'
```
dalf: It should be:
- `searx.plugins.self_info`

searx/plugins/__init__.py
```diff
@@ -226,7 +226,10 @@ def plugin_module_names():
 
 
 def initialize(app):
+    # it is possible to block plugins from being registered into plugin chain
+    blocked_plugins = settings['outgoing'].get('blocked_plugins', [])
+
     for module_name, external in plugin_module_names():
         plugin = load_and_initialize_plugin(module_name, external, (app, settings))
-        if plugin:
+        if plugin and plugin.name not in blocked_plugins:
```
dalf: The check should done on the module name:
```python
if module_name not in blocked_plugins:
  plugin = load_and_initialize_plugin(module_name, external, (app, settings))
  if plugin:
```

searx/plugins/__init__.py
```diff
@@ -226,7 +226,10 @@ def plugin_module_names():
 
 
 def initialize(app):
+    # it is possible to block plugins from being registered into plugin chain
+    blocked_plugins = settings['outgoing'].get('blocked_plugins', [])
```
dalf: After this line https://github.com/searxng/searxng/blob/599d882d0a43b91d6c700edb2f39a900eabbdd5c/searx/settings_defaults.py#L214
add
```python
'blocked_plugins': SettingsValue(list), []),
```

It sets a default value to `[]` and checks that the value is a list.

So here, line 230 in `__init__.py`, you can write `blocked_plugins = settings['outgoing']['blocked_plugins']`

k2s:
Thank you that you care. I improved my code.

Best would be to have token support also for plugins.

Don't know too much internals and history of the project. I am pushing changes I need to successfully run searx in my use scenario. I hope seaxNG will not diverge too much  from searx and that the forks will benefit from each other. Uncoordinated change in internal configuration would lead to incompatibilities. Personally I don't see problem to support for example new key `plugins` which will provide all needed settings and load it into existing/old structures.

not-my-profile:
I agree that this is a good addition but I don't like the name `blocked_plugins`. Having three separate lists for plugins is bad: for example: a plugin could be in `enabled_plugins` and `disabled_plugins` at the same time, which just doesn't make any sense. Having both `disabled_plugins` and `blocked_plugins` is very much confusing to everybody seeing both for the first time.

I'd much rather have us introduce a more sensible format like the one suggested by @dalf in https://github.com/searxng/searxng/issues/681#issuecomment-1013686674 than introducing `blocked_plugins` now only to deprecate it again in the near future.

> I hope seaxNG will not diverge too much from searx

The searx `settings.yml` format is problematic for a variety of reasons ... so much so that I don't think we should maintain backwards compatibility with it until the end of time. But that discussion belongs in #681.

k2s:
@not-my-profile Yes, enabled, disabled, blocked is really messy, but that is how it is in code. Changing `settings.yml` format to be more clear for admins would be great and should not be difficult to implement after agreement how the final, human focused format, should look like. 

There is way to map the new format into the old python variables and keep code base compatibility with searx. Rework of all code that uses this settings requires much more effort, testing and it will mage NG incompatible with the original. That was my concern.

In code the blocked and enabled/disabled variables are very distinct. If something is blocked, it doesn't even comes to the part where enabled/disabled is used. Tokens and custom properties for plugins are related to enabled/disabled problematic.

What I want to say is, that introducing `blocked` internally is not a wrong step. To provide admins with more clear `settings.yml` structure will be a great improvement.

return42:
@not-my-profile ping https://github.com/searxng/searxng/pull/759#issuecomment-1013848275

What do you think should we do / first implement a better setup in settings.yml ...

> I'd much rather have us introduce a more sensible format like the one suggested by @dalf in https://github.com/searxng/searxng/issues/681#issuecomment-1013686674 than introducing `blocked_plugins` now only to deprecate it again in the near future.

or is it OK to accept `blocked_plugins` for the interim:

> but I don't like the name `blocked_plugins`.

not-my-profile:
> introducing `blocked` internally is not a wrong step

I'd agree but adding a new setting to `settings.yml` isn't an internal change ... that file is very much part of our public interface.

So yes I'd prefer us to properly fix the already confusing `enabled_plugins` & `disabled_plugins` instead of introducing yet another confusing setting (for which we would need to implement backwards compatibility, if we finally fixed it properly in the future).

But I defer to your judgement @return42.

return42:
> for which we would need to implement backwards compatibility, if we finally fixed it properly in the future

.. that is true ..

> But I defer to your judgement @return42.

... and I play the ball to @dalf :-)

dalf:
In both cases we have to change settings.yml.

* `blocked_plugins` is compatible with the existing configurations, but it adds a third options about the plugins.
* `plugins` breaks the compatibility (`plugins` as suggested in https://github.com/searxng/searxng/issues/681#issuecomment-1013686674)

My initial thought was to make a clean break in an unknown distant date.

But at this point:
* I'm in favor of the `plugins` syntax right now.
* To deal with the old syntax (the current one in the master branch), `searx/settings_defaults.py` can:
  * log a warning
  * translate on the fly the old syntax to the new syntax for a time. The translation seems straightforward, but I may have overlook the complexity. We can set a dead line at different places, and at the dead line remove the translation code (it can be an additional file for example).

If there is no objection and if you think it is doable, the first thing to do is to make sure the syntax is the right one --> go to #681 first

return42:
> But at this point:
> * I'm in favor of the `plugins` syntax right now.

What does it means related to this PR, is adding `blocked_plugins:` OK for now? / can we review & merge this PR?

return42:
@dalf ping .. please see https://github.com/searxng/searxng/pull/759#issuecomment-1016295679

dalf:
I'm in favor of something like this:
```yaml
plugins:
  searx.plugins.self_info:
    load: false  # plugin in searx.plugins are automatically loaded, "load: false" prevent to load on a plugin
  searx.plugins.hash_plugin:
    enabled_by_default: false
  searx.plugins.hostname_replace:
  mypackage.hostname_replace_extended:
```

later we can add other fields.

return42:
Superseded by:

> There is already `enabled_plugins` and `disabled_plugins` they set the default way how plugins are used. It is not possible to block the plugin from processing queries;

... The ``enabled_plugins:`` section in SearXNG's settings do no longer exists. There is no longer a distinction between built-in and external plugin, all plugins are registered via the settings in the ``plugins:`` section.

The implementation is similar to what has been described in this comment: https://github.com/searxng/searxng/pull/4282#issuecomment-2642231325

> adds `blocker_plugins` property to completely remove plugin from being used.

This is no longer necessary with #4538: to load a plugin, it has to be registered via a fully qualified class name.  A configuration (`PluginCfg`) can be transferred to the plugin, e.g. to activate it by default / *opt-in* or *opt-out* from user's point of view.

```YAML

    plugins:
      ...
        active: true
      ...
```

PR #761, opened 2022-01-15
[fix] use defaults of engines protected with tokens on preference save

k2s:
## What does this PR do?

Only list of user disabled search engines is sent to server from preference page. Server doesn't know what engines were part of preferences. Server expects that all known engines which are not disabled are enabled. The problem is that engines protected by tokens are not provided to user preference page and it is not true, that they are enabled.

If user stores its settings and then adds token which enables more engines all of them will be enabled. The configured default should be respected also in this situation not to overload server and provide better user experience.

* protect all or some engines with token, let some of them have `disabled: true`
* open preference page and clear all cookies
* save the preferences
  * you will already now see that all engines are enabled in cookie
* now enter token
  * you will see that all engines protected by the token are enabled regardless of their property `disabled: true`
* this pull request solves this problem


not-my-profile:
Thanks! Please add a comment to the code when the reason you changed something isn't obvious from the diff alone. Otherwise somebody might mistakenly refactor the code to what was written earlier in the future.

k2s:
sorry, this is not good solution it breaks other situations ... looking into it

return42:
> sorry, this is not good solution it breaks other situations ... looking into it

For the interim I convert this PR to a DRAFT ... hope that is OK for you, otherwise give me a note / thanks!

PR #773, opened 2022-01-17
More typing

not-my-profile:


return42:
> https://github.com/microsoft/pyright

Does it make sense when we add static type checking to the CI?

not-my-profile:
Yes I think that's a good idea. I think for CI [mypy](http://mypy-lang.org/) is still the standard.

We currently however have many type checking errors ... so we would either have to fix those first or add `# type: ignore` comments.

return42:
> We currently however have many type checking errors ... so we would either have to fix those first or add # type: ignore comments.

Or initial we do only selected checks .. e.g. in the pylint CI we have a stronger validation (no `${PYLINT_SEARXNG_DISABLE_OPTION}` applied to the pylint command)

https://github.com/searxng/searxng/blob/b972df255a76e043522c9e4f44cbbcaa6b34df41/manage#L661-L682

if the python file contains a `# lint: pylint` comment ..

https://github.com/searxng/searxng/blob/b972df255a76e043522c9e4f44cbbcaa6b34df41/manage#L33-L42

Here is an example:

https://github.com/searxng/searxng/blob/b972df255a76e043522c9e4f44cbbcaa6b34df41/searx/network/client.py#L1-L3

BTW: OT here, but when you create new files (python or any other file), please add the [SPDX](https://spdx.org/licenses/) identifier ... in case of new python file, add the python file to the "stronger" CI test ...

    # lint: pylint 


Example: https://github.com/searxng/searxng/commit/6f1446d55f60a88f0710ce55fe11e9494e258e17

not-my-profile:
Also OT: How do I run a command in the Python virtualenv?

```
make pyenv.cmd pip 
make: *** No rule to make target 'pyenv.cmd'.  Stop.
```

edit: Ah `./manage pyenv.cmd pip` works.

not-my-profile:
Ok, I can look into setting up mypy for our CI.

return42:
> edit: Ah `./manage pyenv.cmd pip` works.

FYI: I often use a bash in the virtualenv

    ./manage pyenv.cmd bash

> How do I run a command in the Python virtualenv?

[All relevant build tasks are implemented in `./manage` and for CI or IDE integration a small Makefile wrapper is available](https://docs.searxng.org/dev/makefile.html)

Use `./manage help` to get a list of the commands available.

In the Makefile we have only those targets who do not need further arguments.  These targets are wrapper around `./manage ...`.  In the command line you can use `make [TABULATOR]` for a command completion (`./manage ...` does not have a completion).

return42:
> Ok, I can look into setting up mypy for our CI.

If you struggle with bash, I can help.

There is also a `make test.shell` target (that is not used in the CI) to validate shell implementations.  Therefore you will need to install `shellcheck` on your OS.

    sudo -H ./utils/searx.sh install buildhost

> :buildhost:  install packages from OS package manager needed by buildhosts

Will install these packages ..

https://github.com/searxng/searxng/blob/b972df255a76e043522c9e4f44cbbcaa6b34df41/utils/searx.sh#L48-L70

searx/webapp.py
```diff
@@ -585,15 +586,14 @@ def post_request(response):
         'render;dur=' + str(round(request.render_time * 1000, 3)),
     ]
     if len(request.timings) > 0:
-        timings = sorted(request.timings, key=lambda v: v['total'])
+        timings = sorted(request.timings, key=lambda t: t.total)
```
return42, line 590: Nit: Is the renaming "v --> t" of this lambda needed? 

IMO it is better we have 'v' in the lambda and 't' in the for loops below.

searx/webapp.py
```diff
@@ -354,14 +355,14 @@ def get_current_theme_name(override=None):
     return theme_name
 
 
-def get_result_template(theme_name, template_name):
+def get_result_template(theme_name: str, template_name: str):
```
return42, line 358: Does it make sense to add the result type `-> str:` like it has been done for `get_current_theme_name() -> str:` above?  It is more a general question that fist also to other functions we have added type hints.

searx/webapp.py
```diff
@@ -224,6 +227,21 @@
...
+    """This class is never initialized and only used for type checking."""
+
+    preferences: Preferences
+    errors: List[str]
+    user_plugins: List[Plugin]
+    form: Dict[str, str]
+    start_time: float
+    render_time: float
+    timings: List[Timing]
+
```
return42, line 240: On the long term we should modularize the webapp.py .. can we place this type definition in a different module or create a new module if nothing fits?

return42 (approved):
LGTM a did not have an type checker right now, so I can't validate all, but what I see in the commits are looking good to me.

I left some comments about a few nits .. nothing what keeps this PR from a merge.

searx/webapp.py
not-my-profile: I like to name lambda parameters after their type so I think `t` for a `Timing` makes sense.

searx/webapp.py
not-my-profile: Pyright does type inference, so it is smart enough to figure out that this function returns a string.

searx/webapp.py
not-my-profile: Right, I'd prefer us to move it when the time comes instead of now.

not-my-profile:
Ok, I looked into it. Pyright is apparently way smarter than mypy (it finds way more errors), so I think we should definitely use Pyright. I'm happy to wait with this PR until I introduce Pyright CI testing in another PR.

return42:
> I'm happy to wait with this PR until I introduce Pyright CI testing in another PR.

Lets get things done .. for me it is OK when you merge this PR first.  When we have pyright in our CI we can improve further I think.

PR #777, opened 2022-01-17
Update XPath for Google engine

### LINKED ISSUE #771: Google engine is broken
**Version**
```
Docker: 1.0.0-1516-96ab5e57
``` 
**What happened?**
No results for query `rclone docker plugin`

**How To Reproduce**
Set google to be the only search engine, then search.

**Screenshots & Logs**
image

This works fine if I search directly in google. I'm confused. There is nothing in the stdout/stderr logs.

My `settings.yaml`:
```
use_default_settings: true

general:
  debug: false

search:
  safe_search: 0
  autocomplete: ''
  default_lang: 'en-GB'
  formats:
    - html
  languages:
   - pt-BR
   - en-GB

```

unixfox:
## What does this PR do?

Fixes #771

It fixes the Google engine.

1. Start SearXNG locally.
2. Enable only Google.
3. Expect it to work.

## Related issues

Closes #771

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -111,8 +111,8 @@ filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 # specific xpath variables
 # ------------------------
 
-# google results are grouped into <div class="g" ../>
-results_xpath = '//div[@class="g"]'
+# google results are grouped into <div class="g ..." ../>
+results_xpath = '//div[contains(@class, "g")]'
 
 # google *sections* are no usual *results*, we ignore them
 g_section_with_header = './g-section-with-header'
@@ -124,8 +124,8 @@ title_xpath = './/h3[1]'
 # href=...>
 href_xpath = './/div[@class="yuRUbf"]//a/@href'
 
-# in the result group there is <div class="IsZvec" ../> containing he *content*
-content_xpath = './/div[@class="IsZvec"]'
+# in the result group there is <div class="VwiC3b ..." ../> containing the *content*
+content_xpath = './/div[contains(@class, "VwiC3b")]'
 
 # Suggestions are links placed in a *card-section*, we extract only the text
 # from the links not the links itself.


dalf:
Thank you!

return42:
@dalf did you tested this PR .. or just merged?  Right now I tested and see a lot of debug messages .. did you tested and did you now if it is OK to have them? 

```
DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
```

unixfox:
> @dalf did you tested this PR .. or just merged?  Right now I tested and see a lot of debug messages .. did you tested and did you now if it is OK to have them? 
> 
> ```
> DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
> DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
> DEBUG   searx.engines.google          : ingoring <div class="g" ../> section: missing title
> ```

Which keyword did you use?

dalf:
I saw that the results were back on the search results, but you right the log are at least misleading or most probably there is an issue.

return42:
> Which keyword did you use?

Just my default `!go foo`

I Assume the debug messages are from a sidebar or similar, but I didn't had time to debug it in deep .. and now I have a meeting :-(

unixfox:
The XPAth doesn't include the sidebar, but some results may have no titles. I'll look into it tonight if I've time, I would say that at least Google works again, so it's fine to merge this PR for now.

return42:
>  I would say that at least Google works again, so it's fine to merge this PR for now.

Yes merge is OK for now, but I will have a look about what we can improve .. hope I have time for this today ..

return42:
Pew, google has only 5 or 6 real links in its response .. the majority of the links are embedded adds I think .. but I don't really know it / by example the page shown below: 

There are three real links on the top, then 4 links to amazon (embedded adds). After the amazon links there is one real link (www.eventim.com) more .. about the last two links I assume these are also adds (?).

In sum google gives 4 or may be 6 real links on one page .. could it be?

Or I am wrong and google just form results from the same domain (e.g. amazon.de) in one block list?

What do you expect as results in SearXNG .. all links shown in the "result-column" (shown in the screenshot below).. or just the 4 or 6 I described?

Here is my google selector to get just the minority of links (IMO the _real links_)

```python
results_xpath = '//div[@id="search"]//div[contains(@class, "g ")]'
```

EDIT for `use_mobile_ui: true` a different selector is needed .. I will prepare a patch ...

```python
results_xpath_mobile_ui = '//div[contains(@class, "g ")]'
```

grafik

unixfox:
There are no ads on the results list, the ads are on the top or displayed with "Ad" next to it:
image

The 3 links inside Amazon are just additional links for the main link, I don't remember the exact name. But we don't have a way to display these inside Searx like it's displayed on Google so IMO we can skip them.

So in the end we have 7 good links to include into the SearX results.

PR #781, opened 2022-01-18
[fix] google engine: remove adds and fix mobile_ui selector

return42:
## What does this PR do?

1. Fix issue reported in comment [1]
2. Fix XPath selector for the response of google's mobile UI, reported in comment [2]

Filter out add-links [2] from google and fix `use_mobile_ui: true` [2] / we need different selectors to filter out adds [1] 

[1] https://github.com/searxng/searxng/pull/777#issuecomment-1015121322
[2] https://github.com/searxng/searxng/pull/777#issuecomment-1015236238

    !go foo

with `use_mobile_ui: true` and `use_mobile_ui: false` 

## Related issues

- #777

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -112,7 +112,8 @@ filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 # ------------------------
 
 # google results are grouped into <div class="g ..." ../>
-results_xpath = '//div[contains(@class, "g")]'
+results_xpath = '//div[@id="search"]//div[contains(@class, "g ")]'
+results_xpath_mobile_ui = '//div[contains(@class, "g ")]'
 
 # google *sections* are no usual *results*, we ignore them
 g_section_with_header = './g-section-with-header'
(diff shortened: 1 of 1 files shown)


unixfox:
I'll test that on my instance.

return42:
> I'll test that on my instance.

BTW: could you please have a look at my analysis in https://github.com/searxng/searxng/pull/777#issuecomment-1015236238 .. do you agree / disagree?

return42:
> There are no ads on the results list, the ads are on the top or displayed with "Ad" next to it:
> image
>
> So in the end we have 7 good links to include into the SearX results.

This PR will result in 6 good links ..

> The 3 links inside Amazon are just additional links for the main link, I don't remember the exact name. 

It is very complicated to scrap the main (amazon.de) link .. this will need a special XPath selector (I haven't found) and AFIK an additional loop.

Hope this is this acceptable .. ?

unixfox:
7 good links not 6 links, I've marked the good links:
image

> It is very complicated to scrap the main (amazon.de) link .. this will need a special XPath selector (I haven't found) and AFIK an additional loop.
> 
The XPath for fetching the link `Medicine At Mightnight - ....` is not the same for `Jia Jin Foo - Senior ...`?

return42:
> The XPath for fetching the link `Medicine At Mightnight - ....` is not the same for `Jia Jin Foo - Senior ...`?

Yes, that is exactly what I wanted to express / if my analysis is not wrong.

unixfox:
> > The XPath for fetching the link `Medicine At Mightnight - ....` is not the same for `Jia Jin Foo - Senior ...`?
> 
> Yes, that is exactly what I wanted to express / if my analysis is not wrong.

Then don't bother trying to include it.

return42:
> Then don't bother trying to include it.

OK, thanks .. then I like to merge this PR / could you give me your github-approved for?

unixfox (approved):

PR #784, opened 2022-01-18
Fix `js_dependencies` & `css_dependencies`

### LINKED ISSUE #769: Convert JS & CSS dependencies to user preferences
These  [plugins](https://docs.searxng.org/dev/plugins.html) are theme dependent so they should be implemented in themes (instead of via externally injected CSS / JS)

* [x] convert `infinite_scroll.py` to user preference  #812
* [x] convert `vim_hotkeys.py` to user preference
* [x] convert `search_on_category_select.py` to user preference

not-my-profile:
## What does this PR do?

This PR fixes our plugin system, which was previously broken (it was impossible to e.g. add features to the simple theme by adding a plugin). See the commit messages for details.

## Related issues

Fixes #769.


mrpaulblack:
So from a QA and operator side of things (Lets leave code review dalf and return42) what I like is that the plugins are now impl. properly and that new plugins can also be loaded into the simple theme.

What I really want to get around tho is that we have an extra request for each JS file for the plugins (which is IMO not optimal). I dont know a solution for that, just wanted to leave my thoughts about that here...

not-my-profile:
Fair point. I added a commit to bundle the plugin `js_dependencies` on the server-side.

return42:
@dalf @not-my-profile 

> [When a feature should be a embedded plugin ? By embedded plugin, I mean a plugin in this repository. What is the purpose of a embedded plugin compare to normal code ? ... So I would convert the infinite scroll and vim shortcuts plugins to normal preferences.](https://github.com/searxng/searxng/issues/769#issuecomment-1015774980)

With the above comment in mind .. we have three plugins that inject CSS and/or JS:

1. infinite_scroll (CSS & JS)
2. vim_hotkeys (JS & CSS)
3. search_on_category_select (JS)

The 1. and 2. should be converted to normal preferences.  About the 3. I think we should also convert it to a normal preference.

If we do so, we do no longer need the js_dependencies & css_dependencies and can drop this CSS and JS injection entirely ... 

> [When we talk about plugins I think we need to differentiate between client (HTML, CSS, JS) and server site. ATM we have a uncontrolled mix of both, the "infinity scroll" is a good example to demonstrate, that this what we have is conceptional insufficient.](https://github.com/searxng/searxng/issues/769#issuecomment-1015894473)

What do you think, should we drop the js_dependencies & css_dependencies from the "plugin API"?

What left is a clear and simple API that haves defined [entry points](https://docs.searxng.org/dev/plugins.html#plugin-entry-points) we are able to maintain.  

TBH the current architecture with js_dependencies & css_dependencies can't be maintained by us.

----

By example: if there is a CSS in a (real) external plugin that modifies the theme style and the theme style LESS or the HTML template of the theme is changed by us, who guarantees that our change does not break the external plugin?

From admins POV: every time the admin updates SearXNG he needs to test his "external" plugins .. if there is an issue the plugin has to be fixed according to the modified (e.g.) HTML or modified CSS.   If he doesn't and a user has activated this plugin in his cookie, there is a chance that SearXNG becomes unusable to the user.

This is what we have today with js_dependencies & css_dependencies and thats why I think we should drop JSS & CSS from the plugins.

not-my-profile:
Fair point. I'll address this in another PR.

PR #789, opened 2022-01-20
Remove `js_dependencies` & `css_dependencies` 

### LINKED ISSUE #769: Convert JS & CSS dependencies to user preferences

not-my-profile:
Fixes #769.


dalf:
I don't follow the idea here.

For example, since [searx/plugins/search_on_category_select.py](https://github.com/searxng/searxng/blob/a4907db90185ce6141d9052f44e1b2fcc0f329ac/searx/plugins/search_on_category_select.py) becomes nothing more than a switch a label and a description, what is the different between this plugin and an user preference?

```python

        self.key_value_settings: Dict[str, Setting] = {
            # fmt: off
            'categories': MultipleChoiceSetting(
                ['general'],
                locked=is_locked('categories'),
                choices=categories + ['none']
            ),
            ....
            'categories_on_select': MapSetting(
                True,
                locked=is_locked('categories_on_select'),
                map={
                    '': settings['ui']['categories_on_select'],
                    '0': False,
                    '1': True,
                    'True': True,
                    'False': False
                }
            ),
            ....
```

not-my-profile:
Turning these plugins into user preferences can be the next step. This is the first step.

dalf:
I think it should be done right away, one plugin at a time makes the review easier.

not-my-profile:
Just to clarify, you think of three additional commits right?

searx/static/themes/oscar/src/js/plugins/infinite_scroll.js
```diff
@@ -0,0 +1,66 @@
...
+        width: 30px;
+        border: 8px solid #666;
+        border-right-color: transparent;
+        border-radius: 50% !important;
+        margin: 0 auto;
+    }
+    #pagination button {
+        visibility: hidden;
+    }
+    `;
```
dalf, line 64: If the intent is to merge the plugins into the themes, why the CSS is not in the theme (in the less directory)?

not-my-profile: Because the CSS should only be active when the plugin is active. I didn't want to spend time integrating the CSS into the less when we want to deprecate the Oscar theme anyway.

not-my-profile:
Are we concerned about loosing saved user preferences when converting these plugins to user preferences?

dalf:
> Just to clarify, you think of three additional commits right?

see https://github.com/dalf/searxng/tree/pref_infinite_scroll

not-my-profile:
Closing in favor of @dalf's approach of converting these to user settings.

PR #793, opened 2022-01-20
Fix Tor

### LINKED ISSUE #790: Tor over http proxy leads to internal server error
I know this is a really niche issue, however, I am using privoxy to proxy requests over an http proxy to tor OR a vpn depending on whether there is a .onion in the url. 

https://github.com/searxng/searxng/blob/96a1f79c6dcaa08a231ad3a28b8c17b677cddc9c/searx/network/network.py#L170

Effectively, I believe the issue lies in this check, if I enable tor networking in settings.yml the server gets an internal server error -- I believe this is because you manually check if the proxy is using remote dns prior to sending the actual request to tor instead of just asking tor if it's a tor IP. 

I am not sure if this needs fixing or not, as I am not sure if my use case is something you guys would want to account for, however, I believe it worth reporting.

tiekoetter:
## What does this PR do?

This PR allows 'using_tor_proxy' to be checked for each engine individually. This allows SearXNG to use Tor engines without routing everything over the Tor network.

Tor engines are "broken" right now and can't be used as intended. This PR would allow using Tor engines without sacrificing speed on all other engines.

- Install Tor proxy
- Add Tor proxy to the engines
Example:
```
  - name: ahmia
    engine: ahmia
    using_tor_proxy: true
    proxies:
      all://:
        - socks5h://127.0.0.1:9050
    categories: onions
    enable_http: true
    shortcut: ah
```
- `make run`
- Search `!ah test`

## Related issues

Closes #790

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -167,9 +167,6 @@ class Network:
         for transport in client._mounts.values():  # pylint: disable=protected-access
             if isinstance(transport, AsyncHTTPTransportNoHttp):
                 continue
-            if not getattr(transport, '_rdns', False):
-                result = False
-                break
         else:
             response = await client.get('https://check.torproject.org/api/ip')
             if not response.json()['IsTor']:
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -255,7 +255,7 @@ def is_engine_active(engine: Engine):
         return False
 
     # exclude onion engines if not using tor
-    if 'onions' in engine.categories and not settings['outgoing'].get('using_tor_proxy'):
+    if 'onions' in engine.categories and not engine.using_tor_proxy:
         return False
 
     return True
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -135,7 +135,6 @@ outgoing:
   #      - http://proxy1:8080
   #      - http://proxy2:8080
   #
-  #  using_tor_proxy: true
   #
   # Extra seconds to add in order to account for the time taken by the proxy
   #
@@ -254,6 +253,7 @@ engines:
   # Requires Tor
   - name: ahmia
     engine: ahmia
+    using_tor_proxy: false
     categories: onions
     enable_http: true
     shortcut: ah
@@ -1272,6 +1272,7 @@ engines:
   # Requires Tor
   - name: torch
     engine: xpath
+    using_tor_proxy: false
     paging: true
     search_url:
       http://xmh57jrknzkhv6y3ls3ubitzfqnkrwxhopf5aygthi7d6rplyvk3noyd.onion/cgi-bin/omega/omega?P={query}&DEFAULTOP=and


searx/network/network.py
```diff
@@ -167,9 +167,6 @@ async def check_tor_proxy(client: httpx.AsyncClient, proxies) -> bool:
         for transport in client._mounts.values():  # pylint: disable=protected-access
             if isinstance(transport, AsyncHTTPTransportNoHttp):
                 continue
-            if not getattr(transport, '_rdns', False):
```
dalf, line 170: Do you have an example to reproduce the bug on this line?
With:
```yaml
outgoing:
  using_tor_proxy: true
```

I have these errors
```
...
ERROR   searx.network.image_proxy     : Error
  File "/home/alexandre/code/searxng/searx/network/network.py", line 289, in check
    await network.get_client()
  File "/home/alexandre/code/searxng/searx/network/network.py", line 205, in get_client
    raise httpx.ProxyError('Network configuration problem: not using Tor')
httpx.ProxyError: Network configuration problem: not using Tor
  File "/home/alexandre/code/anaconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/alexandre/code/anaconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/alexandre/code/searxng/searx/webapp.py", line 1361, in <module>
    search_initialize(enable_checker=True, check_network=True, enable_metrics=settings['general']['enable_metrics'])
  File "/home/alexandre/code/searxng/searx/search/__init__.py", line 32, in initialize
    check_network_configuration()
  File "/home/alexandre/code/searxng/searx/network/network.py", line 298, in check_network_configuration
    raise RuntimeError("Invalid network configuration")
RuntimeError: Invalid network configuration
```

It is expected.

tiekoetter: If you set `using_tor_proxy: true` in `outgoing` you can only use it if everything is routed over Tor.
I have got the same error as you do.
What do you mean by expected?

dalf: `using_tor_proxy: true` means that requests are sent using tor.
To avoid leak, SearXNG (and searx) checks if the HTTP requests are sent using tor. If this is not the case, the SearxNG does not start.

Even if I remove the `if not getattr(transport, '_rdns', False):` checks, I still get the same error.

---

The `_rdns` check is related to this configuration (inherit from [requests](https://docs.python-requests.org/en/latest/)):
* `socks5://127.0.0.1:9050` tells to use the local DNS --> `_rdns = False`. This is not what we want for tor.
* `socks5h://127.0.0.1:9050` tells to use the DNS from the proxy --> `_rdns = True`. This is what we want for tor.

(rdns means "remote DNS")

Out of topic: I wonder if this legacy is required, I mean, is there a valid configuration with a proxy and local DNS resolution.

tiekoetter: If you set a Tor engine to use a socks5 proxy without name resolution at the proxy in SearXNG it won't be able to contact the onion website. Tor doesn't use normal DNS for name resolution, you need a Hidden Service Directory for that.
I think I don't understand what your arguments against checking if a Tor proxy is used for each engine vs only globally are.

Issue #690, opened 2022-01-04
`categories_as_tabs` followups


not-my-profile:
This is a follow up issue for #634, which introduces categories that are not displayed as tabs and only searchable via bangs.

* [ ]  #769
* [x] fix obvious UX problems
    * [x] #702
    * [x] #854
* [x] #722
* [x] Document "How to search"
* [x] Document how to categorize engines.
    * https://github.com/searxng/searxng/pull/1545
* [x] Make "non tab category" bangs discoverable.
  Maybe by updating the autocompletion so that `!` lists the available category bangs and their descriptions / drawback: not all have autocompletion and it requires JS on the client
    * categories are in the autocompletion
    * !bangs are shown in the preferences https://github.com/searxng/searxng/pull/1545
* [x] Introduce `category_labels` with a custom babel extractor to get rid of hardcoded `gettext` calls in `webapp.py`
    * https://github.com/searxng/searxng/pull/859
* [ ] Stop hardcoding icons in the Simple theme (`searx/templates/simple/categories.html`), define them in `categories_as_tabs` in `settings.yml` instead
* [x] Display country flags for language codes in preferences (e.g. DE -> :de:, PL -> :poland:)
* [ ] Introduce `about.powered_by` and group preference tabs by it

return42:
AFAI can say this issue can be closed / if not, ask for reopening.

PR #603, opened 2021-12-15
[mod] simple-theme: on small devices show category icon, not name

return42:
## What does this PR do?

[mod] simple-theme: on small devices show category icon, not name

**before**

grafik

**after**

grafik

## How to test this PR locally?

To test I patched settings.yml

1. add a new category "TV" to see how *new* categories will be rendered (the icon)
2. activate autocomplete, to see autocompletion is positioned well
3. add `Content-Security-Policy` to check CSP

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 305b254c..0ec979cd 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -16,7 +16,7 @@ search:
   # Existing autocomplete backends: "dbpedia", "duckduckgo", "google",
   # "startpage", "swisscows", "qwant", "wikipedia" - leave blank to turn it off
   # by default.
-  autocomplete: ""
+  autocomplete: "google"
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: ""
@@ -48,7 +48,7 @@ server:
 
   secret_key: "ultrasecretkey"  # change this!
   # Proxying image results through searx
-  image_proxy: false
+  image_proxy: true
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
   # POST queries are more secure as they don't show up in history but may cause
@@ -60,6 +60,7 @@ server:
     X-Download-Options: noopen
     X-Robots-Tag: noindex, nofollow
     Referrer-Policy: no-referrer
+    Content-Security-Policy: "default-src 'self'; img-src 'self' data: ;"
 
 ui:
   # Custom static path - leave it blank if you didn't change
@@ -70,7 +71,7 @@ ui:
   # it decreases the privacy, since the browser can records the page titles.
   query_in_title: false
   # ui theme
-  default_theme: oscar
+  default_theme: simple
   # Default interface locale - leave blank to detect from browser information or
   # use codes from the 'locales' config section
   default_locale: ""
@@ -1604,6 +1605,7 @@ engines:
   - name: mediathekviewweb
     engine: mediathekviewweb
     shortcut: mvw
+    categories: TV
 
   # - name: yacy
   #   engine: yacy
```

PR #798, opened 2022-01-21
Preferences Icon Update

MontyQI:
## What does this PR do?

Changes preferences icon to 
settings

This icon will be better associated with settings that previous one.

`make run`
change theme to simple

### DIFF
--- a/docs/admin/engines/sql-engines.rst
+++ b/docs/admin/engines/sql-engines.rst
@@ -136,12 +136,10 @@ find an example configuration below:
 MySQL
 -----
 
-.. _mysql-connector-python: https://pypi.org/project/mysql-connector-python
-
 .. sidebar:: info
 
    - :origin:`mysql_server.py <searx/engines/mysql_server.py>`
-   - ``pip install`` mysql-connector-python_
+   - ``pip install`` :pypi:`mysql-connector-python <mysql-connector-python>`
 
 MySQL is said to be the most popular open source database. Before enabling MySQL
 engine, you must install the package ``mysql-connector-python``.
--- a/docs/admin/installation-switch2ng.rst
+++ b/docs/admin/installation-switch2ng.rst
@@ -7,7 +7,7 @@ Switch from searx to SearXNG
 .. sidebar:: info
 
    - :pull:`456`
-   - `A comment about rolling release <https://github.com/searxng/searxng/pull/446#issuecomment-954730358>`_
+   - :pull:`A comment about rolling release <446#issuecomment-954730358>`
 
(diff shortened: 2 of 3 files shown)


return42 (approved):
Thanks for your PR! Other search engines use the hamburger menu, but I also like the gear icon you suggest and like you, I think the association of a gear with a setting is more obvious.

I approve this PR but I won't judge what is better, lets hear what other say ...

tiekoetter:
I personally think the gear icon looks better.

mrpaulblack (changes requested):
Thanks for the PR. I also like the gear icon more. But please change your commits:

* one commit with the correct icon set from `node_modules/ionicons/dist/svg` in gruntfile (you can replace the current one): https://github.com/searxng/searxng/blob/f3d4b25e739164ea6af087379536b3ef3538d63c/searx/static/themes/simple/gruntfile.js#L180
* one commit with the build theme; for that just run `make static.build.commit` which will auto create the commit :+1: 

[EDIT] you can do this by using `git reset --hard` than making your changes and force pushing those changes on this branch to override this already existing commit with `git push origin -f`

return42:
@mrpaulblack thanks for your hint about `svg2jinja`!

Its OT here but while we are talking about I want to ask you and @dalf if we really need this grunt task:
 
https://github.com/searxng/searxng/blob/f3d4b25e739164ea6af087379536b3ef3538d63c/searx/static/themes/simple/gruntfile.js#L200

Isn't it simpler to copy the svg icons once into the html / AFAIK the ionicons will never modified(?)

dalf:
About the icon, I prefer the hamburger menu.

> Isn't it simpler to copy the svg icons once into the html 

I don't understand when do you want to make this copy.
Once the theme is compiled, ionicons is not available (it is in node_modules)
`svg2jinja` minifies the content of the selected icons, and inline the results in the HTML file

return42:
> I don't understand when do you want to make this copy.

@dalf I haven't thought to its end but I am asking for because of configurable icons in the categories 

- [Stop hardcoding icons in the Simple theme (searx/templates/simple/categories.html), define them in categories_as_tabs in settings.yml instead](https://github.com/searxng/searxng/issues/690) from @not-my-profile and
- https://github.com/searxng/searxng/pull/603#issuecomment-994929745 from mine

> Once the theme is compiled, ionicons is not available (it is in node_modules)

May be we copy the SVG files from ionicons once, like we can do it from any other source.

What I mean: when those icons are configurable any other SVG file should be good enough. (I ignore request/response cycles for now / if, then they need to optimized on the server code).

PR #799, opened 2022-01-21
Add autocompleter from Brave 

return42:
## What does this PR do?

Add autocompleter from Brave / BTW add SPDX tag to `autocomplete.py`

one more alternative

    make run

Activate brave auto completion and test completion in various languages

## Related issues

Initial commit was cherry picked from https://github.com/searx/searx/pull/3109 (and slightly modified to fit SearXNG needs)

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -36,6 +36,20 @@ def get(*args, **kwargs):
     return http_get(*args, **kwargs)
 
 
+def brave(query, lang):
+    # brave search autocompleter
+    url = 'https://search.brave.com/api/suggest?{query}'
+    resp = get(url.format(query=urlencode({'q': query})))
+
+    results = []
+
+    if resp.ok:
+        data = loads(resp.text)
+        for item in data[1]:
+            results.append(item)
+    return results
+
+
 def dbpedia(query, lang):
     # dbpedia autocompleter, no HTTPS
     autocomplete_url = 'https://lookup.dbpedia.org/api/search.asmx/KeywordSearch?'
@@ -128,6 +142,7 @@ backends = {
     'swisscows': swisscows,
     'qwant': qwant,
     'wikipedia': wikipedia,
+    'brave': brave,
 }
 
 
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -1,32 +1,22 @@
-'''
-searx is free software: you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""This module implements functions needed for the autocompleter.
 
-searx is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU Affero General Public License for more details.
+"""
 
-You should have received a copy of the GNU Affero General Public License
-along with searx. If not, see < http://www.gnu.org/licenses/ >.
-
-(C) 2013- by Adam Tauber, <asciimoo@gmail.com>
-'''
-
-
-from lxml import etree
 from json import loads
 from urllib.parse import urlencode
 
+from lxml import etree
 from httpx import HTTPError
-
 
 from searx import settings
 from searx.data import ENGINES_LANGUAGES
 from searx.network import get as http_get
 from searx.exceptions import SearxEngineResponseException
+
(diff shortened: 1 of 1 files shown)


dalf:
Work as expected but 23156f2b692ec110e0140df0767a9f5927018426 is no go for me: 
* the right solution is to have `fetch_supported_languages` function for the brave engine.
* since the brave engine uses the xpath module and inheritance is not implemented, there is no straightforward solution.

So for now, I would
* remove language support
* find a way to specify a `fetch_supported_languages` function the engines using the xpath or json modules.

searx/autocomplete.py
```diff
@@ -36,7 +25,26 @@ def get(*args, **kwargs):
...
+    country = 'all'
+    if lang in _brave:
+        country = lang
+    kwargs = {'cookies': {'country': country}}
+    resp = get(url, **kwargs)
+
+    results = []
+
+    if resp.ok:
+        data = loads(resp.text)
```
dalf: `data = resp.json()` doesn't trigger the charset detection for each request (see the logs).

the `json()` method has shortcut to detect UTF encoding: https://github.com/encode/httpx/blob/321d4aa5097fe7f24cdfed7191c44de589294780/httpx/_models.py#L1514

dalf (changes requested):
Work as expected but 23156f2b692ec110e0140df0767a9f5927018426 is no go for me because it requires manual update of searx/data/brave_languages.json

the right solution is to have `fetch_supported_languages` function for the brave engine.
However, since the brave engine uses the xpath module and inheritance is not implemented, there is no straightforward solution.

So for now, I would
* remove language support
* find a way to specify a `fetch_supported_languages` function the engines using the xpath or json modules.

return42:
> the right solution is to have fetch_supported_languages function for the brave engine.

I think that the whole effort is not worth it.  To me autocompletion response from brave feel slow.

> find a way to specify a fetch_supported_languages function the engines using the xpath or json modules.

I switch this PR to DRAFT, if there is the day we fetch-languages for XPath & JSON engines we may reactivate this PR:

dalf:
> I think that the whole effort is not worth it. To me autocompletion response from brave feel slow.

The support of `fetch_supported_languages` for the xpath engine could be helpful for many other engines.
But there is no need to do that now.

Since brave seems slow, can we skip the language support now and support `fetch_supported_languages` later?

return42:
> Since brave seems slow, can we skip the language support now and support `fetch_supported_languages` later?

Not sure I understand what you mean, isn't it the same I said? .. or do you want to add brave auto-completion with the country=all? (means "all regions")

--- 

EDIT: OK, now I understand :-)  .. I will send a patch and re-activate this PR.

searx/autocomplete.py
return42: > see the logs ..

I debugged ..

```
(Pdb) l
 39  	    results = []
 40  	
 41  	    if resp.ok:
 42  	        import pdb
 43  	        pdb.set_trace()
 44  ->	        data = loads(resp.text)
 45  	        for item in data[1]:
 46  	            results.append(item)
 47  	    return results
 48  	
 49  	
(Pdb) resp.text
DEBUG   charset_normalizer            : override steps (5) and chunk_size (512) as content does not fit (153 byte(s) given) parameters.
INFO    charset_normalizer            : ascii passed initial chaos probing. Mean measured chaos is 0.000000 %
DEBUG   charset_normalizer            : ascii should target any language(s) of ['Latin Based']
INFO    charset_normalizer            : ascii is most likely the one. Stopping the process.
```

what means this logs are not related to json parser.  I assume the messages are related to a typecast of `resp.text` .. but I don't really know what I can do ...

return42: To avoid such issues from httpx.Response, we need to use httpx.Response.json() / fixed in e9588b70

@dalf: Should we replace `json.load(resp.text)` by `resp.json()` entirely? (I can do it in another PR)

searx/autocomplete.py
```diff
@@ -36,7 +26,26 @@ def get(*args, **kwargs):
...
 
-def dbpedia(query, lang):
+def brave(query, _lang):
+    # brave search autocompleter
+    url = 'https://search.brave.com/api/suggest?'
+    url += urlencode({'q': query})
+    country = 'all'
+    # if lang in _brave:
+    #    country = lang
+    kwargs = {'cookies': {'country': country}}
```
dalf, line 36: I see this cookie `ui_lang=en-us` but not `country`.

dalf (approved):
I left a comment about the cookie, not important now since the language selection is not supported.

Thank you.

searx/autocomplete.py
return42: When I go to https://search.brave.com/ and trace the suggestion request, I see:

grafik

dalf: I see
image

may be I use the wrong setting.

PR #805, opened 2022-01-22
Fix Tor

### LINKED ISSUE #790: Tor over http proxy leads to internal server error

tiekoetter:
## What does this PR do?

This PR allows 'using_tor_proxy' to be checked for each engine individually. This allows SearXNG to use Tor engines without routing everything over the Tor network.

Tor engines are "broken" right now and can't be used as intended. This PR would allow using Tor engines without sacrificing speed on all other engines.

- Install Tor proxy
- Add Tor proxy to the engines
Example:
```
  - name: ahmia
    engine: ahmia
    using_tor_proxy: true
    proxies:
      all://:
        - socks5h://127.0.0.1:9050
    categories: onions
    enable_http: true
    shortcut: ah
```
- `make run`
- Search `!ah test`

## Related issues

Closes #790

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -42,6 +42,7 @@ ENGINE_DEFAULT_ARGS = {
     "safesearch": False,
     "time_range_support": False,
     "enable_http": False,
+    "using_tor_proxy": False,
     "display_error_messages": True,
     "tokens": [],
     "about": {},
@@ -230,8 +231,8 @@ def set_language_attributes(engine: Engine):
         )
 
 
-def update_attributes_for_tor(engine):
-    if settings['outgoing'].get('using_tor_proxy') and hasattr(engine, 'onion_url'):
+def update_attributes_for_tor(engine: Engine) -> bool:
+    if using_tor_proxy(engine) and hasattr(engine, 'onion_url'):
         engine.search_url = engine.onion_url + getattr(engine, 'search_path', '')
         engine.timeout += settings['outgoing'].get('extra_proxy_timeout', 0)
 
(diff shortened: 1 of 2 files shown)


dalf:
ping @MarcAbonce

MarcAbonce:
I'm trying this on local and it works for me.

I just think that this `use_tor_proxy` setting is getting even more confusing now. Seeing `use_tor_proxy: false` in an onion engine looks so weird because an onion engine by definition needs Tor. And also, if I enable `use_tor_proxy` globally (under outgoing) it ignores the engine's  `use_tor_proxy: false`.
For example:
```
...
outgoing:
  proxies:
    ...
  using_tor_proxy: true
...
engines:
  - name: ahmia
    use_tor_proxy: false
...
```
Works fine for Ahmia, even though that setting is a contradiction. Which is functionally fine, but it's very confusing, in my opinion.

I think the root problem is that  `use_tor_proxy` means two different things at the same time:
 1. It enables the onion engines.
 2. It checks that the engines are correctly being proxied through Tor.

Now I think that these should be two separate settings. One setting to enable the onions category and `use_tor_proxy` should only be used to do the proxy check. In this case, onion engines would always automatically have  `use_tor_proxy: true` without even needing to specify so in the settings file because like I said, it doesn't even make sense for an onion engine not to use Tor.

But maybe this could be a separate PR for later. Like I said, this works fine.

dalf:
See commit 3a241d9ef444b599bd2d515749edb10b9c4bb5ff : 
* it removes the `using_tor_proxy` from settings.yml (the value is False by default)
* fix a in bug in `searx/network/network.py`

@tiekoetter feel free to cherry-pick 3a241d9ef444b599bd2d515749edb10b9c4bb5ff  and merge with your existing commits. Otherwise I will have a look in the coming days.

PR #812, opened 2022-01-23
Convert the infinite_scroll plugin as a preference

dalf:
## What does this PR do?

* oscar theme: done
* simple theme: done except keyboard navigation

TODO:
* [x] in the category images, the scrollbar stay at the bottom of the page
* [x] check `searxng.http` still works in the generic cases
* [x] vim keyboard navigation doesn't work  --> https://github.com/searxng/searxng/issues/836
* [x] ~(optional / debatable) convert on the fly the preferences from the old format = enable infinite scroll when the cookie `enabled_plugins` contains `searx.plugins.infinite_scroll`.~

* Implement the infinite scroll for the simple theme

* `make themes`
* with different browser, at least Firefox and Chrom* : 
   * in the simple theme, and in the oscar theme:
     * check a request in the general category
     * check a request in the images category
     * check a request `!general !images time`

## Related issues

Part of #769


dalf:
With the PR, the "infinite scroll" preferences will be lost.
I prefer not to implement an on the fly convention to avoid to add new bugs (`preferences.py` is already enough complicated).

return42 (approved):
Wow, very cool work .. 

I tested the simple theme, two things struck me

1. Images that are loaded by `loadNextPage()` are not lazy loaded (the lazy attribute is set, but seems ignored).
2. The `searxng.image_thumbnail_layout.align();` in `loadNextPage()`  aligns not all images ..

Bildschirmfoto vom 2022-02-04 17-55-28

After a *resize* event the images are aligned wit correct aspect ratio ...

Bildschirmfoto vom 2022-02-04 17-59-39

I tried to figure out what is the cause .. but gave up after a few hours ..

dalf:
> Images that are loaded by loadNextPage() are not lazy loaded (the lazy attribute is set, but seems ignored).

A possible explanation: 

The current PR has a hack: https://github.com/dalf/searxng/blob/09e7abd2a2d131ae2d0c38db36fe6c658626b6d4/searx/static/themes/simple/src/js/main/infinite_scroll.js#L43

Without it, the viewport stay at the bottom of the page, and the infinite scroll never stops loading the next pages.
But even with this hack, the whole list of images are "displayed" (since the viewport is at bottom of the page [between lines 39 and 43](https://github.com/dalf/searxng/blob/09e7abd2a2d131ae2d0c38db36fe6c658626b6d4/searx/static/themes/simple/src/js/main/infinite_scroll.js#L39-L43)). Consequence: images are not lazy loaded.

A workaround is display the pagination at the end of page:

```less
#main_results div#results.only_template_images {
  margin: 1rem @results-tablet-offset 0 @results-tablet-offset;
  display: grid;
  grid-template-columns: 100%;
  grid-template-rows: min-content min-content min-content 1fr min-content;
  gap: 0;
  grid-template-areas:
    "corrections"
    "answers"
    "sidebar"
    "urls"
    "pagination";
```

The issue with this workaround is the URL at the top of the page just after the filters (it would be better on the side for example).


> The searxng.image_thumbnail_layout.align(); in loadNextPage() aligns not all images ..

https://github.com/dalf/searxng/tree/pref_infinite_scroll2

This branch contains various changes:
* `searxng.image_thumbnail_layout` uses a [Mutation Observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver) to monitor new added images --> this fix the issue.
* the infinite scroll implementation uses  a [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

return42:
Nit: the plugin comment can be removed (in both branches) .. I'm still testing your new branch ..

https://github.com/searxng/searxng/blob/9cd9dfd26e07552419941a34369b084f2adc30f4/searx/settings.yml#L172

return42:
@dalf I had a look at your [pref_infinite_scroll2](https://github.com/dalf/searxng/tree/pref_infinite_scroll2) branch and the interesting article you linked ... I haven't understand all in deep / can't image what it means in every detail.

[pref_infinite_scroll2](https://github.com/dalf/searxng/tree/pref_infinite_scroll2) fixes the aligning of the images and as far I can say the [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) might also help to fix the _lazy load issue_.

I pushed a WIP commit on my clone of your https://github.com/return42/searxng/commits/pref_infinite_scroll2 branch with some issues I found.

Long words, short question: what do you think how we should proceed .. should we squash my WIP commit in your [pref_infinite_scroll2](https://github.com/dalf/searxng/tree/pref_infinite_scroll2) and then ~rebase~ reset this PR branch pref_infinite_scroll on pref_infinite_scroll2? .. If so, you should create a backup branch of this pref_infinite_scroll branch.

---- 

Side note: I also found a small issue in the images_layout.js when it is used in the oscar theme --> I think we should not look to close to oscar theme and we should not implement quirks just to handle issues in oscar. (we should better drop oscar ASAP / its implementations are in the way whatever we do and it cause a lot of work to maintain something we want to get rid anyway).

dalf:
I've fixed some issues in the [pref_infinite_scroll2](https://github.com/dalf/searxng/tree/pref_infinite_scroll2) branch, see https://github.com/dalf/searxng/commit/e57bb885a126d7b64630be25dedd7641e55c9eea
@return42 , I don't see anymore your WIP commit in this pref_infinite_scroll2 branch?

If the "Search URL" can be on the right side or at least not at the top, I think this version 2 can replaced the version in this PR
Screenshot:
image

return42:
> @return42 , I don't see anymore your WIP commit in this pref_infinite_scroll2 branch?

It is not in your branch, I can't push commits on your branches (except the branch is a PR what is not the case for pref_infinite_scroll2).

You find my WIP commit here in my clone of your branch ..

**in my repository**: https://github.com/return42/searxng/commits/pref_infinite_scroll2

PR #813, opened 2022-01-24
Move WolframAlpha away from Science category

MarcAbonce:
## What does this PR do?

I just noticed #634 and I loved it.
I've always thought that WolframAlpha was a little bit out of place in the Science category, which is otherwise exclusively used for scientific articles, so if you wanted to use the `!science` bang for papers you had to remember to disable Wolfram first.

For now I just left the category blank, but technically it could be part of a Calculator or Mathematics category if there were more engines of that type.

Related-ish issue: https://github.com/searx/searx/issues/2472

1. Go to Preferences and then to Engines.
1. WolframAlpha should be in Other/Other section instead of Science.
1. Run a query with `!science` bang.
1. WolframAlpha shouldn't be called.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1455,7 +1455,7 @@ engines:
     # Or you can use the html non-stable engine, activated by default
     engine: wolframalpha_noapi
     timeout: 6.0
-    categories: science
+    categories: []
 
   - name: dictzone
     engine: dictzone


return42:
> I've always thought that WolframAlpha was a little bit out of place in the Science category, which is otherwise exclusively used for scientific articles, so if you wanted to use the `!science` bang for papers you had to remember to disable Wolfram first.

I don't know, but is category `science` really the wrong category?

These are my thoughts about ...

When I search for an equation (e.g. `E=m*c^2`) I would do it in the general category or in the `science` category.  Equations are most often *related to science* I think, or with other words; who has said that there are only white papers in the `science` category.

An alternative to `science` for WolframAlpha might be `general`, but as you wrote:

> [My only concern is that Wolfram has a way longer timeout compared to other engines, so if WolframAlpha is on by default, the general category would feel slower than before. So I think it should be disabled by default if moved to the general category.](https://github.com/searx/searx/issues/2472#issuecomment-761387959)

I don’t have a final opinion yet, what do you think?

dalf:
I'm in favor of this PR (no category), because WolframAlpha  
* can answer generic question like `length of Broklyn bridge`
* can be slow

I feel it is more a tool that has to be called on purpose rather receiving any queries (from the general or science categories).

return42 (approved):

geckolinux:
Hi there, I really think that the original SearX properly placed Wolfram Alpha under the Science category, because some users might want to enable it by default (without using bangs). Please see my reasoning for this here:
https://github.com/searxng/searxng/issues/966

xieve:
As I have mentioned in the issue above, I would suggest that WolframAlpha be placed in the general category, while being disabled by default. This would mean every user could opt to use it for instant answers on any public instance while the default config would still be fast. Bangs would also still work. I have just set up an instance with that configuration because I really wanted to have WolframAlpha instant answers on every query.

An even cooler solution would be if instant answers could be lazy-loaded, such that the search results appear as soon as available, while WolframAlpha can take its time and drop the instant answer in later.

PR #815, opened 2022-01-24
[enh] add security policy

mrpaulblack:
## What does this PR do?

This PR adds a security policy to the SearXNG org as well as a way for people to contact us if a vuln. is found.

Make the SearXNG project more secure, by enabling users to contact us without disclosing a security issues found to the public.

Check out the commit text. The text is going to be  displayed in the `Security` tab on github after merge.

## Author's checklist

ping @not-my-profile

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,10 @@
+# Security Policy
+
+We love responsible reports of (potential) security issues in SearXNG.
+
+You can contact us at security@searxng.org.
+
+Be sure to provide as much information as possible and if found
+also reproduction steps of the identified vulnerability. Also
+add the specific URL of the project as well as code you found
+the issue in to your report.


return42 (approved):

not-my-profile:
Good idea to take advantage of GitHub's `SECURITY.md` feature!

How about splitting it up into a couple of paragraphs? Right now it's a bit of a wall of text.

>Please do not open a public issue, if you found a vulnerability

I would spin it more positively, how about:

 We love responsible reports of (potential) security issues in SearXNG.

Note the (potential), I think it's an important detail. We still want to know about security problems even when the reporter isn't sure if it's actually exploitable.

>Also add the specific project and code you found the issue in

I would explicitly ask for a link to the repository (because the SECURITY.md will also end up in forks).

>as well as the commit where the issue was found in

Asking for that detail seems a bit superfluous, when given a description of a vulnerability and the repository it should be trivial for us to find it.

dalf:
Should the email address be listed in the Contact section of README.rst?

mrpaulblack:
> Should the email address be listed in the Contact section of README.rst?

No, I dont think so; `security@searxng.org` is only meant for the purpose of reporting security vuln. in SearXNG. Every other request should go through `contact@searxng.org` IMO. So we should also keep the security Email to SECURITY.md file IMO.

mrpaulblack:
@not-my-profile I changed the first sentence, added paragraphs, removed the requirement of adding the commit sha as well as adding the notice to include the project URL.

PR #817, opened 2022-01-24
Pyright 01

not-my-profile:
This PR enables Pyright in three files and introduces a new commit naming convention:

* `[pyright:strict] <module>` -- pyright: strict was enabled for the module
* `[pyright:basic] <module>` -- pyright: basic was enabled for the module

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "dependencies": {
     "eslint": "^8.7.0",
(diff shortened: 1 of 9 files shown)


not-my-profile:
@dalf Can we drop support for Python 3.7? I want to use `typing.TypedDict` and it's only available since Python 3.8.

dalf (approved):

PR #783, opened 2022-01-18
[enh] add pyrightconfig.json and integrate pyright into CI

not-my-profile:
By adding this file Pyright automatically detects the packages from our Python virtual environment.

This can be tested by using the Pyright extension for VS Code or by running `npx pyright`.

PR #820, opened 2022-01-25
[test.pyright] suppress warnings about intentional monkey patching

return42:
The warnings:

- "logger" is not defined'
- "supported_languages" is not defined'
- "language_aliases" is not defined'

are very verbose and superfluous, since these messages are related to
intentional monkey patching.

[1] https://github.com/searxng/searxng/pull/783#issuecomment-1019818178

-----

I think it is enough when we remove the messages about monkey patching and leave other messages in the output:

```
Searching for source files
Found 202 source files
  searx/version.py:94:10 - warning: Import "searx.version_frozen" could not be resolved (reportMissingImports)
  searx/engines/mongodb.py:8:6 - warning: Import "pymongo" could not be resolved (reportMissingImports)
  searx/engines/mysql_server.py:9:8 - warning: Import "mysql.connector" could not be resolved (reportMissingImports)
  searx/engines/postgresql.py:9:8 - warning: Import "psycopg2" could not be resolved from source (reportMissingModuleSource)
  searx/engines/xpath.py:179:28 - warning: "categories" is not defined (reportUndefinedVariable)
  searx/search/__init__.py:184:82 - warning: "flask" is not defined (reportUndefinedVariable)
  searx/shared/__init__.py:8:12 - warning: Import "uwsgi" could not be resolved (reportMissingImports)
  searx/shared/shared_uwsgi.py:4:8 - warning: Import "uwsgi" could not be resolved (reportMissingImports)
0 errors, 86 warnings, 0 informations 

```

### DIFF
--- a/manage
+++ b/manage
@@ -696,10 +696,14 @@ test.pyright() {
     node.env.devtools
     # We run Pyright in the virtual environment because Pyright
     # executes "python" to determine the Python version.
-    pyenv.cmd npx --no-install pyright -p pyrightconfig-ci.json
+    build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"
+    pyenv.cmd npx --no-install pyright -p pyrightconfig-ci.json \
+        | grep -v ".py$" \
+        | grep -v '/engines/.*.py.* - warning: "logger" is not defined'\
+        | grep -v '/engines/.*.py.* - warning: "supported_languages" is not defined' \
+        | grep -v '/engines/.*.py.* - warning: "language_aliases" is not defined'
     dump_return $?
 }
-
 
 test.black() {
     build_msg TEST "[black] \$BLACK_TARGETS"


not-my-profile:
Can you add a comment to the code saying why these warnings are filtered?

return42:
> Can you add a comment to the code saying why these warnings are filtered?

good point .. I added a message:

    build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"

dalf:
I see it is possible to add `# type: ignore` at the end of the line we want pyright to ignore, but it requires to change a lot of engine files.
If we keep the `grep`, can we at least check the warning is about a file in `searx/engines`?

So for example, `make test.pyright` will report an issue if the variable `logger` is not defined in `searxx/network/__init__.py`.

return42:
> can we at least check the warning is about a file in searx/engines?

good point, fixed.

manage
```diff
@@ -696,11 +696,15 @@ test.pyright() {
     node.env.devtools
     # We run Pyright in the virtual environment because Pyright
     # executes "python" to determine the Python version.
-    pyenv.cmd npx --no-install pyright -p pyrightconfig-ci.json
+    build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"
+    pyenv.cmd npx --no-install pyright -p pyrightconfig-ci.json \
```
not-my-profile, line 700: When the pyenv isn't installed yet the "suppress warnings" message is printed long before the warnings are actually suppressed.

return42: Yes, what do you suggest?

not-my-profile: How about:

```
pyenv.activate
build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"
npx --no-install pyright ...
```

Alternatively you can just change the build_msg to a shell comment ... I don't think printing it is necessary ... I just wanted it to be documented in the code.

return42: > How about:
>    pyenv.activate  

and what about the build messages above, its also printed "long before" ..

> it to be documented in the code.

I think this requirement is fulfilled.

not-my-profile: >what about the build messages above

It makes sense to me ... in a log it makes sense to print why sth is installed before you install it.

dalf (approved):
LGTM.

manage
dalf: If I understand correctly, the issue appears when someone starts `make test.pyright` without a virtualenv or if the deppendencies are changed? Otherwise there is no issue, for example `make test` is going to first install the dependencies?

return42: Its not really an issue .. its common that (underlying) tools spit out their messages and this might cause that there is a distance between one "build" message and the next "build" message.

Here in this case @not-my-profile suggested ...

```
pyenv.activate
build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"
npx --no-install pyright ...
```

instead of ..

```
build_msg TEST "[pyright] suppress warnings related to intentional monkey patching"
pyenv.cmd npx --no-install pyright ...
```

Consequently, this would mean we can replace nearly all `pyenv.cmd` what makes no sense to me .. tbh I do not understand this discussion and merged this PR.

@not-my-profile if you think there is a change needed, please open a PR for.

not-my-profile: Right ... that could have been avoided using a subshell. But you're right that it's not really that important.

PR #564, opened 2021-11-29
 [fix] normalize SearXNG color from #487cff to #3050ff 

return42:
## What does this PR do?

 [fix] normalize SearXNG color from `#487cff` to `#3050ff`

## Why is this change important?

The blue color of the simple theme is [#3050ff](https://color-hex.org/color/3050ff) ...

https://github.com/searxng/searxng/blob/0311eba538df589bc11a4d76d3c94b104c5b764d/searx/static/themes/simple/src/less/definitions.less#L29

## How to test this PR locally?

**before:**

grafik

**after:**

grafik

grafik

PR #825, opened 2022-01-26
Simple theme uppercase buttons

MontyQI:
## What does this PR do?
~~Changes simple theme. For now, changed search bar and colors.~~
Changes buttons text in simple theme to uppercase.

It makes simple theme look more modern.

``make node.env``
``make themes.simple``
``make run``
change to simple theme
## Additional resources
Search bar and some colors are from [https://github.com/paulgoio/searxng](url)

**Screenshots**
Screenshot_2022-01-26-12-14-20_2560x1440
Screenshot_2022-01-26-12-30-47_2560x1440

### DIFF
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -401,9 +401,9 @@
     {{ _("These cookies serve your sole convenience, we don't use these cookies to track you.") }}
   </p>
 
-  <input type="submit" value="{{ _('save') }}" />
+  <input type="submit" value="{{ _('Save') }}" />
   <div class="{% if rtl %}left{% else %}right{% endif %} preferences_back"><a href="{{ url_for('clear_cookies') }}">{{ _('Reset defaults') }}</a></div>
-  <div class="{% if rtl %}left{% else %}right{% endif %} preferences_back"><a href="{{ url_for('index') }}">{{ _('back') }}</a></div>
+  <div class="{% if rtl %}left{% else %}right{% endif %} preferences_back"><a href="{{ url_for('index') }}">{{ _('Back') }}</a></div>
 
 </form>
 {% endblock %}
--- a/searx/templates/simple/results.html
+++ b/searx/templates/simple/results.html
@@ -162,7 +162,7 @@
                   <input type="hidden" name="safesearch" value="{{ safesearch }}" >
                   <input type="hidden" name="theme" value="{{ theme }}" >
                   {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
-                  <button type="submit">{{ icon_small('chevron-left') }} {{ _('previous page') }}</button>
+                  <button type="submit">{{ icon_small('chevron-left') }} {{ _('Previous page') }}</button>
                 </div>
             </form>
         {% endif %}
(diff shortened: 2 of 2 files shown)


mrpaulblack:
Well this PR is basically changes from my fork. In my fork I am using the colors from google... I dont think its a good idea to use those colors in this project upstream tbh (google probably has the rights to those colors or something). Also the impl. of the search bar in my fork is also pretty hacky tbh and I think this would need to be reworked to be usable in SearXNG upstream.

What I like is in https://github.com/searxng/searxng/pull/825/commits/4715e2f8046f3fcc288255f7b88b8de91738c943 changing the label to uppercase

But yeah all in all I am meh about this PR. I dont think its a good idea to use googles colors in a big project like SearXNG and I agree that the search bar needs to be reworked, but we should not use the hacky impl. of my fork. I already started refactoring the search bar here a while ago: https://github.com/mrpaulblack/searxng/commits/simple-theme-search-form

Also using material design buttons dont really fit in with the simple theme. There are hover effects, shadows and so on with those elements and with the simple theme we try to avoid those effects (Tho I think a shadow around search bar and autocomplete would be nice tbh)

MontyQI:
I didn't know it was your fork. I think at least the changes of preferences colors are good idea. Now they look a bit yellow.

return42:
> Now they look a bit yellow.

yellow? .. could you give an example?

FYI in https://github.com/searxng/searxng/pull/564 I (we) came to the conclusion that we leaf the colors unchanged.  If there comes the day we change the colors we should harmonize them with the SVGs.

Personally I would not apply this PR (sorry).

mrpaulblack:
> yellow? .. could you give an example?

There is more blue in the google colors. For example the main background is blue-ish tinted with google when using the dark theme, while upstream in SearXNG the dark colors dont have the blue tint and appear more yellow-ish when comparing IMO.

How about this.. Can you change this PR to only include the changes to make the label uppercase, because I really like that change.

This evening I will make a draft PR for the search bar and we can discuss redesign there @MontyQI ?



Also just FYI: We dont need to touch the current blue accent colors. What we could do is just make all the dark colors slightly more blue upstream...
If you want to you can propose a PR with where you change those colors, but without using google colors @MontyQI

MontyQI:
Ok I'll change it.

mrpaulblack (approved):
LGTM

dalf:
Note: it breaks the existing translations in weblate.

I think weblate can make some magic to figure out that may be the translation for "Save" should be the one "save" with a minor change, but nothing of this is configure.

I don't know if we should uppercase the labels using CSS?

return42:
> I don't know if we should uppercase the labels using CSS?

No, please not .. do not apply any magic on strings within the rendering layer.

> I think weblate can make some magic to figure out that may be the translation for "Save" should be the one "save" with a minor change, but nothing of this is configure.

IMO we should get a try and merge.  If all goes wrong, we have a really good and active translator community.

mrpaulblack:
> I don't know if we should uppercase the labels using CSS?

I did not think about this; Hmm we would be more flexible in the future if we want to CAPS LOCK the whole button label for example, but is the CSS Uppercase going to work properly for every language LTR as well as RTL?

return42:
>  but is the CSS Uppercase going to work properly for every language LTR as well as RTL?

No .. any magic on strings in the rendering will raise a lot of new questions and quirks .. please lets avoid such complications.

return42:
> I think weblate can make some magic to figure out that may be the translation for "Save" should be the one "save" with a minor change, but nothing of this is configure.

@dalf: there is a "automatic suggestion" function ..

grafik

dalf:
For information, there are the two versions "Save" and "save" : 
* https://github.com/searxng/searxng/blob/master/searx/templates/oscar/preferences.html#L492
* --> https://weblate.bubu1.eu/translate/searxng/searxng/en/?q=%22Save%22&sort_by=-priority%2Cposition&offset=2

Issue #472, opened 2021-11-03
[simple theme] remove shaddow effects and other blur


return42:
Some of the elements from the simple theme have shadow effects / some of these effects are only visible in light mode

IMO the simple theme should be "flat and simple" .. no shadows, no 3D effects, no blur effects.

* back to top button (shadow effect):

![grafik](https://user-images.githubusercontent.com/554536/140034711-9d0f6fb2-7b18-4668-983c-a89517439997.png)

* toggle button (shadow effect / only viewable in light mode)

![grafik](https://user-images.githubusercontent.com/554536/140034895-9698e9e6-a393-4bfa-b0c5-18ee0352664f.png)

![grafik](https://user-images.githubusercontent.com/554536/140035179-2ed2cd12-3361-4d67-a742-0c0e5782e78d.png)

mrpaulblack:
What about transitions? I was planning on adding transitions to the btn hover effects the search box and stuff like this... (Right now there is only a transition on the backtotop element...)

[EDIT] also: there are shadows on these elements in the dark theme but since they are darker than the background they are very faint (but still there; check out dark theme result page on mobile for example...)

return42:
IMO transitions are OK, my remark is only about the "flat and simple" design. By example; it bits my eyes to have a shadow at the right button but none at the left ... for me, the right one is placed in a third dimension (3D) as the button casts a shadow ...

![grafik](https://user-images.githubusercontent.com/554536/140034711-9d0f6fb2-7b18-4668-983c-a89517439997.png)

... if we want to have a shadow, I think the shadow should be added to all buttons.  But before adding shadows all over there I think we should reduce the complexity and should stay by a very _flat_ design (no 3D effects, all items are 2D).

My English is not very elaborated, hope I could clarify what I mean (2D vs. 3D)

>  there are shadows on these elements in the dark theme but since they are darker

is exactly what I wanted to say :)

return42:
Closed by https://github.com/searxng/searxng/pull/474/commits/ca971bd5e5c1421777696a3c7f315880110ffc65

PR #829, opened 2022-01-26
[WIP] refactor search form

mrpaulblack:
## What does this PR do?

Change the search form. This is planned:
* autocomplete has the same width as search form
* shadow underneath search input as well as autocomplete
* (maybe use gridlayout tot drop positioning through autocomple.js)
* ...

Cleanup search form and make it more recognizable as the main point of the view.

```make themes.all```

```make run```

## Author's checklist

ping @MontyQI feel free to add suggestions in this Pr thread

## Related issues

TODO https://github.com/MontyQI/searxng/commit/1d8a7bb02e71395de5887bb268890bdb50845d97
TODO I think that on mobile there shouldn't be send_search button. It's on keyboard so it isn't necessary.

### DIFF
--- a/searx/static/themes/simple/src/js/main/search.js
+++ b/searx/static/themes/simple/src/js/main/search.js
@@ -70,9 +70,6 @@
           Delay: 300,
           _Position: function () {
             this.DOMResults.setAttribute("class", "autocomplete");
-            this.DOMResults.style.top = (this.Input.offsetTop + this.Input.offsetHeight) + "px";
-            this.DOMResults.style.left = this.Input.offsetLeft + "px";
-            this.DOMResults.style.width = this.Input.clientWidth + "px";
           },
           _Open: function () {
             var params = this;
@@ -85,12 +82,6 @@
             });
           },
         }, "#" + qinput_id);
-
-        // hack, see : https://github.com/autocompletejs/autocomplete.js/issues/37
-        w.addEventListener('resize', function () {
-          var event = new CustomEvent("position");
-          qinput.dispatchEvent(event);
-        });
       }
 
       qinput.addEventListener('focus', placeCursorAtEndOnce, false);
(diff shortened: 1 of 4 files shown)


MontyQI:
What about shadow like that? I don't know which color would be good for dark theme.
Screenshot_2022-01-26-07-34-52_2560x1440

return42:
FYI: 
- https://github.com/searxng/searxng/issues/472

MontyQI:
> * shadow underneath search input as well as autocomplete

So why is it here?

return42:
> So why is it here?

I don't know .. this PR is in Draft, commit is WIP .. lets wait what @mrpaulblack will do.

mrpaulblack:
Yeah I think #472 was a step in the right direction. Tho I am unhappy with the current impl. The search form is IMO not highlighted enough since its the center of the search. This is why I would like to add a shadow for the search form as well as the autocomplete back to really highlight this part of the UI. Also I like the first design from @MontyQI , but IMO the shadow is a little extreme...

return42:
> The search form is IMO not highlighted enough since its the center of the search.

Is it? .. I do not have a problem to find the input field ;-)

I think other search fields (e.g. ddg, github ..) are similar to what we have. 

> This is why I would like to add a shadow for the search form as well as the autocomplete back to really highlight this part of the UI. Also I like the first design from @MontyQI , but IMO the shadow is a little extreme...

"simple" is flat 2D, adding a shadow to an element gives an optical third dimension, IMO this is a style break that does not fit to the rest.

But I think we can better discuss a layout when we see your design .. don't let hesitate you by my remarks.

MontyQI:
I found one bug with change to outline of autocomplete. Under clear search button there is white space without outline.
Screenshot_2022-01-27-11-56-12_2560x1440

mrpaulblack:
Ok thanks :+1: ; I will work on this PR next week; Dont have a lot of time for SearXNG right now...

mrpaulblack:
Closing in favor of #963

PR #830, opened 2022-01-26
[fix] docs - don't add plugins twice

### LINKED ISSUE #828: Plugins are listed twice in the documentation
See https://docs.searxng.org/admin/plugins.html

return42:
The PluginStore is already initalized when the application is initalized

    searx.plugins.initialize(application)

Closes: https://github.com/searxng/searxng/issues/828

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,7 +3,6 @@
 
 import  sys, os
 from pallets_sphinx_themes import ProjectLink
-from flask import Flask
 
 from searx import get_setting
 from searx.version import VERSION_STRING, GIT_URL, GIT_BRANCH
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -47,7 +47,6 @@
 from searx.webapp import application
```
not-my-profile: The `application` import is now unused. Could you please add a comment to explain that this import is necessary to initialize the plugins?

return42: done

not-my-profile: Ah yeah I think changing the import to `import searx.webapp` would be better (since the application isn't needed in the scope).

return42: opps sorry / done

PR #475, opened 2021-11-04
[enh] engine - add Tineye reverse image search

return42:
## What does this PR do?

[enh] engine - add Tineye reverse image search

This is just a prototype to have a basis for discussion, nothing more!

Typical Tineye _search URL_:

- **HTML search form** https://www.tineye.com/search/?&url=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg 

- **result_json** https://tineye.com/result_json/?page=0&url=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg

Alternatives:

- **google**  https://www.google.com/searchbyimage?image_url=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg
- **bing** https://www.bing.com/images/searchbyimage?FORM=IRSBIQ&cbir=sbi&imgurl=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg
- **yandex** https://yandex.com/images/search?source=collections&&url=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg&rpt=imageview
- **sogou** https://pic.sogou.com/ris?flag=1&drag=0&query=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg
- **baidu** https://graph.baidu.com/details?isfromtusoupc=1&tn=pc&carousel=0&image=&image=https%3A%2F%2Fhekint.org%2Fwp-content%2Fuploads%2F2019%2F11%2F1989.1_Fountain-of-Milk_web-768x667.jpg

# Why is this change important?

Add an engine for reverse image searching.

## How to test this PR locally?

    !tin https://hekint.org/wp-content/uploads/2019/11/1989.1_Fountain-of-Milk_web-768x667.jpg

## Related issues

The initial commit (https://github.com/allendema/searx/commit/455b2b4460cd830ac1f7e62e824040e2fe648de9) comes from @allendema's PR 3040 at searx  project

[...]

PR #835, opened 2022-01-28
Introduce Next-Gen Engines

not-my-profile:
This commit introduces a new class-based API to implement online engines.
The new API is designed so that it is backwards compatible with our current module-based engine API.

The Wikipedia engine is converted to the new API as a proof of concept.

For details please refer to the commits.

Planned followups:

* convert more engines to see if the API is flexible enough or needs to be changed
* introduce `Engine` subclasses for online_dictionary & online_currency
* add more docstrings for classes & fields
* document the new API in the developer documentation

### DIFF
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -25,74 +25,74 @@ supported_languages_url = 'https://meta.wikimedia.org/wiki/List_of_Wikipedias'
 language_variants = {"zh": ("zh-cn", "zh-hk", "zh-mo", "zh-my", "zh-sg", "zh-tw")}
 
 
-# set language in base_url
-def url_lang(lang):
-    lang_pre = lang.split('-')[0]
-    if lang_pre == 'all' or lang_pre not in supported_languages and lang_pre not in language_aliases:
-        return 'en'
-    return match_language(lang, supported_languages, language_aliases).split('-')[0]
+    # set language in base_url
+    def url_lang(lang):
+        lang_pre = lang.split('-')[0]
+        if lang_pre == 'all' or lang_pre not in supported_languages and lang_pre not in language_aliases:
+            return 'en'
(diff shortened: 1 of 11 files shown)


dalf:
Unbaked ideas slightly off topic and related to https://github.com/searxng/searxng/pull/475 : 

* The processors are some kind of super classes for the engines. See the implementations of : https://github.com/searxng/searxng/blob/2a6d84dab5fa22a98718afa7450c7a35b987e9c1/searx/search/processors/abstract.py#L162-L164

* The plugin `hash_plugin` handles `sha512 The quick brown fox jumps over the lazy dog`. This plugin is no much different from an offline engine:
  * https://github.com/searxng/searxng/blob/master/searx/plugins/hash_plugin.py
  * https://github.com/searxng/searxng/blob/master/searx/engines/sqlite.py (one offline engine)

There is one major difference: the `hash_plugin` sees only the queries, whereas `sqlite` has to be triggered by a category selection or a bang. In addition there are [`answers`](https://github.com/searxng/searxng/tree/master/searx/answerers) (the different with the plugins: the engines don't see the user query, if there is an answer).

We can imagine some kind of engines that see only the queries, but as triggered by some query syntax. These engines could be in the "Special queries" category.

In this case we can rewrite:
* the `hash_plugin` and `self_info` plugins can be written as engines.
* the various instances of the online* processors can be abstract engines.

In addition we can imagine a workflow configuration:
* send the query to these categories first and engine list
* if there is no answer, send the query to these other groups of engines
* etc...

May be we need a wiki page / etherpad / whatever to gather the ideas?

searx/engines/wikipedia.py
```diff
@@ -1,16 +1,29 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+# pyright: basic
```
return42, line 2: nit: can we add new & re-implemented py-files also to `# lint: pylint`?

PR #839, opened 2022-01-29
[mod] add documentation about searx.utils

dalf:
## What does this PR do?

This module is a toolbox for the engines.
Is should be documented.

In addition, searx/utils.py is checked by pylint.

Documentation about this module can helpful to newcomers.

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -1,7 +1,13 @@
-# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pyright: basic
+"""Utility functions for the engines
+
+"""
 import re
 import importlib
 
+from typing import Optional, Union, Any, Set, List, Dict, MutableMapping, Tuple, Callable
 from numbers import Number
 from os.path import splitext, join
 from random import choice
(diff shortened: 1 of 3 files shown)


searx/utils.py
```diff
@@ -28,37 +33,49 @@
...
-class NotSetClass:
-    pass
+class NotSetClass:  # pylint: disable=too-few-public-methods
+    """Internal class for this module, do not create instance of this class.
+    Replace the None value, allow explicitly pass None as a function argument"""
 
 
 NOTSET = NotSetClass()
 
+STORAGE_UNIT_VALUE = {
```
not-my-profile: Two things I think we should start doing:

* Naming stuff that is not intended to be accessed from outside the module with a leading `_`.
* Putting global variables that are only used by one method directly above the method declaration (I think in general putting variable declarations as close as possible to their first use makes code more readable).

searx/utils.py
```diff
@@ -538,7 +531,7 @@ def get_xpath(xpath_spec):
     raise TypeError('xpath_spec must be either a str or a lxml.etree.XPath')
 
 
-def eval_xpath(element, xpath_spec):
+def eval_xpath(element: ElementBase, xpath_spec: str) -> Any:
```
not-my-profile: I think in general no type annotation is better than `Any` (because if there is no type annotation Pyright does type inference).

dalf: Removed

searx/utils.py
dalf: > Naming stuff that is not intended to be accessed from outside the module with a leading _.

Done.

> Putting global variables that are only used by one method directly above the method declaration (I think in general putting variable declarations as close as possible to their first use makes code more readable).

As soon as two or more function use a global variable, it becomes difficult to know where the variable is declared.

searx/utils.py
```diff
@@ -586,7 +581,7 @@ def eval_xpath_list(element, xpath_spec, min_len=None):
     return result
 
 
-def eval_xpath_getindex(elements, xpath_spec, index, default=NOTSET):
+def eval_xpath_getindex(elements: ElementBase, xpath_spec: XPathSpecType, index: int, default: Any = _NOTSET) -> Any:
```
not-my-profile: I'd remove the `Any` return type here.

dalf: Done

searx/utils.py
```diff
@@ -321,7 +320,7 @@ def int_or_zero(num):
     return convert_str_to_int(num)
 
 
-def is_valid_lang(lang):
+def is_valid_lang(lang) -> Optional[Tuple]:
```
not-my-profile: When specifying `List` or `Tuple` type hints, I think we should whenever possible specify the element types e.g. `Tuple[bool, str, str]`. For this method Pyright infers `tuple[Literal[True], str, str] | None` revealing that the boolean is always True ... depending on what the method is used for we could remove the boolean ... I dunno.

dalf: > t the boolean is always True

Yes.. I would like to avoid too much code changes in this PR.
To remove this always `True` value, it requires to change some engines:
* https://github.com/searxng/searxng/blob/f240a67bd7df88f38ebadf93dcccaa0f557bddaf/searx/search/processors/online_dictionary.py#L31-L38
* https://github.com/searxng/searxng/blob/f240a67bd7df88f38ebadf93dcccaa0f557bddaf/searx/engines/dictzone.py#L30

searx/utils.py
return42: > Naming stuff that is not intended to be accessed from outside the module with a leading _.

I don't think so / I don't see a difference between function names and names of other objects .. names that can accessed from outside are listed in `__all__` // [Setting `__all__` to an empty list indicates that the module has no public API.](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)

searx/utils.py
not-my-profile: That's understandable.

searx/utils.py
not-my-profile: When looking at a function I want to be able to immediately tell whether it is private or public. I don't want to have to scroll up to look for a `__all__`. Besides prefixing `_` makes it obvious that sth shady is going on when there is a `import _foo from bar`.

return42: I can only repeat what is common practice, I don't know what a single person prefer.

return42: BTW: it is also a common practice to say a (function-) name that is not documented (by a docstring) is not public.

not-my-profile: Prefixing underscores is definitely common practice. It's also covered by PEP 8:

> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose names start with an underscore.

But even more importantly there is a [Pyright check](https://github.com/microsoft/pyright/blob/main/docs/configuration.md) for it that we can enable:

>reportPrivateUsage [boolean or string, optional]: Generate or suppress diagnostics for incorrect usage of private or protected variables or functions. Protected class members begin with a single underscore (“_”) and can be accessed only by subclasses. Private class members begin with a double underscore but do not end in a double underscore and can be accessed only within the declaring class. Variables and functions declared outside of a class are considered private if their names start with either a single or double underscore, and they cannot be accessed outside of the declaring module. The default value for this setting is 'none'.

searx/utils.py
```diff
@@ -475,23 +471,23 @@ def ecma_unescape(s):
...
+    string = _ECMA_UNESCAPE4_RE.sub(lambda e: chr(int(e.group(1), 16)), string)
     # "%20" becomes " ", "%F3" becomes "ó"
-    s = ecma_unescape2_re.sub(lambda e: chr(int(e.group(1), 16)), s)
-    return s
+    string = _ECMA_UNESCAPE2_RE.sub(lambda e: chr(int(e.group(1), 16)), string)
+    return string
 
 
-def get_string_replaces_function(replaces):
+def get_string_replaces_function(replaces: Dict[str, str]) -> Callable:
```
not-my-profile: `Callable` is more ambiguous than what Pyright can infer without type annotations (`((text: Unknown) -> str)`).

dalf: Fixed (?)
Feel free to add other commits in this PR to fix the typing issues.

dalf:
@not-my-profile is there anything I can improve or change in this PR?

return42 (approved):
Thanks!

PR #840, opened 2022-01-29
[mod] user_help: support translations

dalf:
## What does this PR do?

replace "user_help.HELP" by a new function "get_help_for_locale".
This function the help a one locale, and fall back to English if the
translation is not available.

The user documentation is written using a markup language to allow translation.

* create a directory in `searx/user_help/` with a local name ( `de` for example).
* start the server, go to `/about?locale=de`, check the English version is displayed
* copy `searx/user_help/en/about.md` to `searx/user_help/de/about.md`
* make some changes
* start the server, go to `/about?locale=de`, check the new version is displayed


searx/user_help/__init__.py
```diff
@@ -37,12 +38,24 @@ def render(app: flask.Flask):
...
     define_link_targets = ''.join(f'[{name}]: {url}\n' for name, url in link_targets.items())
 
-    for filename in pkg_resources.resource_listdir(__name__, 'help'):
-        rootname, ext = os.path.splitext(filename)
-
-        if ext != '.md':
+    for locale_name in pkg_resources.resource_listdir(__name__, '.'):
+        if locale_name not in LOCALE_NAMES:
             continue
+        HELP[locale_name] = {}
```
not-my-profile, line 44: this violates the type annotation

searx/user_help/__init__.py
```diff
@@ -6,8 +6,9 @@
...
 import mistletoe
 
-from . import get_setting
-from .version import GIT_URL
+from .. import get_setting
+from ..version import GIT_URL
+from ..locales import LOCALE_NAMES
 
 HELP: Dict[str, str] = {}
 """ Maps a filename under help/ without the file extension to the rendered HTML. """
```
not-my-profile, line 14: the type annotation and docstring should be updated

not-my-profile:
I think before we support translations we should support multiple files. Via `/help/<filename>` (and turning `/about` into a redirect).

Also note that I think with translations I think it would be better if the language code was always in the URL, that is `/help/en/about`. This way the help pages can always be shared via the URL and whoever you send the URL to will see them in the same language (which I think is generally best practice for content in general).

dalf:
so I close this PR (and keep the branch in case it is still useful after the change about the endpoints).

PR #843, opened 2022-01-30
[simple] page with header

not-my-profile:
See the commit messages.


mrpaulblack (changes requested):
If you put the margin on the body -> the footer is broken;

For example the preferences page footer (same for about page):
image

[EDIT] the margin should be on the main so that the footer can take the whole width. If you add the page_with_header class to the main element this PR could work :+1:

mrpaulblack (approved):
LGTM :+1:

PR #847, opened 2022-01-31
[simple] preferences link

not-my-profile:
See the commit messages.


mrpaulblack (changes requested):
Could you please change the right value of the pref link to 1.8rem in style.less line 577 and remove it from index.less? I think this would look better on about and so on...

Everything else LGTM

PR #848, opened 2022-01-31
Introduce `/help` route

not-my-profile:
## What does this PR do?

Makes the Markdown files in `searx/help/` available under `/help/en/<pagename>` if `<pagename>` is listed in `searx.user_help._TOC`. Translation will be implemented in the future.

`/about` is redirected to `/help/en/about`.

1. Create a new `foobar.md` with content `# foo bar` in `searx/help`.
2. Add `'foobar'` to `searx.user_help._TOC`
3. `make run`
4. Go to http://127.0.0.1:8888/help/en/foobar.

You will see:

image image

The tabs are automatically generated (and their order depends on the order of the `_TOC` list).

## Related issues

#722


searx/webapp.py
```diff
@@ -877,8 +877,21 @@ def __get_translated_errors(unresponsive_engines: Iterable[UnresponsiveEngine]):
 
 @app.route('/about', methods=['GET'])
 def about():
-    """Render about page"""
-    return render('about.html', help=user_help.HELP)
+    """Redirect to about page"""
+    return redirect(url_for('help_page', pagename='about'))
```
dalf, line 881: Note: The link in the base.html templates can be `{{ url_for('help_page', pagename='about') }}` to avoid a redirect. It can be in a second time since the link may change again with the i18n support.

searx/user_help.py
```diff
@@ -37,12 +46,16 @@ def render(app: flask.Flask):
...
-    for filename in pkg_resources.resource_listdir(__name__, 'help'):
-        rootname, ext = os.path.splitext(filename)
-
-        if ext != '.md':
-            continue
-
-        markdown = pkg_resources.resource_string(__name__, 'help/' + filename).decode()
-        markdown = define_link_targets + markdown
-        HELP[rootname] = mistletoe.markdown(markdown)
+    for pagename in _TOC:
```
dalf, line 49: In which case a file exists in `searx/help` but is not declared in `_TOC`?
Instead:
* keep `pkg_resources.resource_listdir(__name__, 'help')`
* pass the `_TOC` values to `help.html` templates, so only `{% for name, page in all_pages %}` can be `{% for name, page in toc_pages %}`

But I may misunderstood the intent

searx/webapp.py
```diff
@@ -877,8 +877,21 @@ def __get_translated_errors(unresponsive_engines: Iterable[UnresponsiveEngine]):
...
+
+
+@app.route('/help/en/<pagename>', methods=['GET'])
```
dalf, line 884: `/robots.txt` should include this URL path.

dalf (commented):
Few comments and questions.

searx/webapp.py
not-my-profile: Our robots.txt is:

```
User-agent: *
Allow: /
Allow: /about
Disallow: /stats
Disallow: /preferences
Disallow: /*?*q=*
```

The `Allow: /about` is unnecessary, the first two lines allow everything by default. So we don't need to add anything.

searx/user_help.py
not-my-profile: If a file would exist in `searx/help` without being listed in `_TOC` it would simply be ignored, which as far as I'm concerned isn't a problem. Do you find it problematic?

dalf: > Do you find it problematic?

Nope.

dalf (approved):
Thank you!

PR #849, opened 2022-01-31
[fix] engines description - currency_convert.py

### LINKED ISSUE #787: Currency engine has DuckDuckGo metadata
In the engine selector of the preferences window, the currency search engine has the same metadata and wikidata url as duckduckgo, I'd assume there should be a difference of some sort there clarifying what source the currency uses or, if it's a duckduckgo service, at least clarifying that it's a currency service by duck duck go.

return42:
## What does this PR do?

- [fix] engines description - currency_convert.py
- [pylint] engines/currency_convert.py

Closes https://github.com/searxng/searxng/issues/787

### DIFF
--- a/searx/engines/currency_convert.py
+++ b/searx/engines/currency_convert.py
@@ -13,6 +13,7 @@ about = {
     "use_official_api": False,
     "require_api_key": False,
     "results": 'JSONP',
+    "description": "Service from DuckDuckGo.",
 }
 
 engine_type = 'online_currency'
--- a/searx/engines/currency_convert.py
+++ b/searx/engines/currency_convert.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- currency convert (DuckDuckGo)
+# lint: pylint
+"""Currency convert (DuckDuckGo)
 """
 
 import json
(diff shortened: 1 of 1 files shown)


searx/engines/currency_convert.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- currency convert (DuckDuckGo)
+# lint: pylint
+"""currency convert (DuckDuckGo)
+
```
not-my-profile: What's the point of this change to the docstring? I think previously it looked better.

return42: No space before title, then one empty line, then body.

not-my-profile: Right ... I think the blank line looks weird. And looking at [PEP 257](https://www.python.org/dev/peps/pep-0257/) I don't think it's required ... since there is no body.

return42: Anything more in this PR?

not-my-profile: No, otherwise LGTM.

return42: OK, fixed blank line.

PR #859, opened 2022-02-03
[mod] add i18n infrastructure for SearXNG message files (searxng.msg)

### LINKED ISSUE #814: Weblate doesn't show the "other"/"others" categories as translatable strings
61853aa0fbcd2573f50e9c092f61cde1ad71492a (current master)

**What happened?**

Weblate doesn't show the "other"/"others" categories as translatable strings. They do not show up in any of the [messages.po](https://github.com/searxng/searxng/blob/master/searx/translations/en/LC_MESSAGES/messages.po) files either, even though all the other new subcategories already appear in there.

**How To Reproduce**
1. Go to Weblate.
1. Try to find the "other" string.

**Expected behavior**
1. Strings should be translatable.

**Additional context**

I think that the script used to generate the messages.po files wasn't able to pick up those strings because they are defined outside the gettext, see: https://github.com/searxng/searxng/blob/master/searx/webapp.py#L189-L190
```python
...
    gettext('web'),
    gettext(DEFAULT_GROUP_NAME),
    gettext(OTHER_CATEGORY)
)
```

return42:
## What does this PR do?

With this patch ``searxng.msg`` files can be added to SearXNG.  In ``searxng.msg`` files messages can be defined which are not captured by babel's gettext, like the generic names of the categories or messages that are stored in constants.

## Related issues

Closes: https://github.com/searxng/searxng/issues/814

### DIFF
--- a/babel.cfg
+++ b/babel.cfg
@@ -1,4 +1,7 @@
+[extractors]
+searxng_msg = searx.babel_extract.extract
 [ignore: **/node_modules/**]
 [python: **.py]
 [jinja2: **/templates/**.html]
 extensions=jinja2.ext.autoescape,jinja2.ext.with_
+[searxng_msg: **/searxng.msg]
--- a/docs/dev/translation.rst
+++ b/docs/dev/translation.rst
@@ -16,6 +16,7 @@ Translation
 
 .. sidebar:: |translated|
 
+   - :ref:`searx.babel_extract`
    - Weblate_
    - SearXNG `translations branch`_
    - SearXNG `Weblate repository`_
--- a/docs/src/searx.babel_extract.rst
+++ b/docs/src/searx.babel_extract.rst
@@ -1,0 +1,8 @@
+.. _searx.babel_extract:
+
+===============================
+Custom message extractor (i18n)
+===============================
+
+.. automodule:: searx.babel_extract
+  :members:
--- a/manage
+++ b/manage
@@ -30,7 +30,7 @@ GECKODRIVER_VERSION="v0.30.0"
 export NODE_MINIMUM_VERSION="16.13.0"
 # SPHINXOPTS=
 BLACK_OPTIONS=("--target-version" "py37" "--line-length" "120" "--skip-string-normalization")
-BLACK_TARGETS=("--exclude" "searx/static,searx/languages.py" "searx" "searxng_extra" "tests")
+BLACK_TARGETS=("--exclude" "searx/static,searx/languages.py" "--include" "searxng.msg" "searx" "searxng_extra" "tests")
 
 pylint.FILES() {
 
@@ -41,6 +41,7 @@ pylint.FILES() {
     # These py files are linted by test.pylint()
 
     grep -l -r --include \*.py '^#[[:blank:]]*lint:[[:blank:]]*pylint' searx searxng_extra tests
+    find . -name searxng.msg
 }
 
 YAMLLINT_FILES=()
--- a/searx/babel_extract.py
+++ b/searx/babel_extract.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""This module implements the :origin:`searxng_msg <babel.cfg>` extractor to
+extract messages from:
+
+- None
+
+The ``searxng.msg`` files are selected by Babel_, see Babel's configuration in
+:origin:`babel.cfg`::
+
+    searxng_msg = searx.babel_extract.extract
+    ...
+    [searxng_msg: **/searxng.msg]
+
+A ``searxng.msg`` file is a python file that is *executed* by the
+:py:obj:`extract` function.  Additional ``searxng.msg`` files can be added by:
+
+1. Adding a ``searxng.msg`` file in one of the SearXNG python packages and
+2. implement a method in :py:obj:`extract` that yields messages from this file.
+
+.. _Babel: https://babel.pocoo.org/en/latest/index.html
+
+"""
+
+from os import path
+
+SEARXNG_MSG_FILE = "searxng.msg"
+_MSG_FILES = []
+
+
+def extract(
+    # pylint: disable=unused-argument
+    fileobj,
+    keywords,
+    comment_tags,
+    options,
+):
+    """Extract messages from ``searxng.msg`` files by a custom extractor_.
+
+    .. _extractor:
+       https://babel.pocoo.org/en/latest/messages.html#writing-extraction-methods
+    """
+    if fileobj.name not in _MSG_FILES:
+        raise RuntimeError("don't know how to extract messages from %s" % fileobj.name)
+
+    namespace = {}
+    exec(fileobj.read(), {}, namespace)  # pylint: disable=exec-used
+
+    for name in namespace['__all__']:
+        for k, v in namespace[name].items():
+            yield 0, '_', v, ["%s['%s']" % (name, k)]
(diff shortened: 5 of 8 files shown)


not-my-profile:
I think the category names should rather be defined in `settings.yml` (with a custom babel extractor), since they're instance-specific configuration which is best placed in `settings.yml`. Which leaves just five strings in `searxng.msg` which I think isn't quite enough to warrant a custom extractor / file format.

return42:
> I think the category names should rather be defined in `settings.yml` (with a custom babel extractor), since they're instance-specific configuration which is best placed in `settings.yml`. 

I don't think so; we can't translate what is customized in an instance, we can only translate what is defined somewhere in the SearXNG project.

return42:
> Which leaves just five strings in `searxng.msg` ..

Forgot to mention: Not in this PR but my intention is to add more messages than we really use. By example I think about common messages (categories) like "radio", "TV" and more we can add to our translation workflow and instance maintainers can organize their (e.g.) categories in this range of translated names like they want.

return42:
@dalf @mrpaulblack what do you think about this PR .. the main intention is to have a infrastructure that consist of ``searxng.msg`` files to define strings that are *generic* and not covered by the translation workflow right now.

In commit 784bf9e is an example of such an ``searxng.msg`` file, that also fixes:

- #814

If it is not clear what I mean by *generic strings* or it is not clear what the intention is (e.g. https://github.com/searxng/searxng/pull/859#issuecomment-1029971059), please ask me.

mrpaulblack:
So if I get this right:
In the `searxng.msg` file in this PR there are (and we will add more in the future) category names and style names for the simple theme defined. Which means these English names are not covered by the babel translations? Meaning: It does not matter what interface language a SearXNG instance is set to, it will always display the generic English name?

return42:
> Meaning: It does not matter what interface language a SearXNG instance is set to, it will always display the generic English name?

Yes, we have variables like `DEFAULT_GROUP_NAME`, `OTHER_CATEGORY` which can't fetched by the translation framework (gettext can only fetch strings):

```python
...
    gettext('web'),
    gettext(DEFAULT_GROUP_NAME),
    gettext(OTHER_CATEGORY)
)
```

Another use case: we want to translate category names that are only listed/defined in the configuration (settings.yml).

ATM: we have this workaround .. the dictionary `_category_names` is never used in the code, it exists only to get our category names into the translation workflow.

https://github.com/searxng/searxng/blob/5fc53b41f8a0c6e1e1dea02f1427100c0a760beb/searx/webapp.py#L167-L191

return42:
PR waits  since 6 weeks for a review, this PR is a bugfix, if there is no veto I am going to merge this PR.

MarcAbonce (approved):
LGTM.

I locally added a dummy string to `searxng.msg`, ran pybabel and it worked.

mrpaulblack (approved):
I dont fully understand everything in this PR, but from a QA view LGTM :+1: 

(I tested with different interface languages and the category names worked correctly)

dalf:
It works.

just wondering the usage `CONSTANT_NAMES`, `DEFAULT_GROUP_NAME`, `OTHER_CATEGORY`, `CATEGORY_NAMES`, `FILES`, etc... ? Why not a list:

```python
CATEGORY_NAMES = [
    'files',
    'general',
    'music',
    'social media',
    'images',
    'videos',
    'it',
    'news',
    'map',
    'onions',
    'science',
]
```

---

If categories were defined in one it could look like this:
```yaml
categories:
  tabs:
    - label: General
       icon: ionicons/dist/svg/search-outline.svg
    - label: Images
       icon: ionicons/dist/svg/image-outline.svg
    ...
  no_in_tabs:
    - label: apps
    - label: dictionnaries
    - label: lyrics
    - label: packages
    - label: q&a
    - label: repos
    - label: software wikis
    - label: web
  no_tab_category:
    label: Other
  no_category:
    label: Others
```

return42:
> Why not a list:

I thought it is better to have names.  This is a general solution, may be there comes the day we need to use them:
```python
foo = [CATEGORY_NAMES]['FILES']
```

> If categories were defined in one it could look like this:

If we know how we will arrange the names, we can reorder the dictionaries in `searxng.msg`  files

dalf:
What is the advantage of python file over a yaml file?

return42:
A python file is needed to import ..

> [... messages that are stored in constants.](https://github.com/searxng/searxng/pull/859#issue-1123304556)

https://github.com/searxng/searxng/blob/3201aa1b3fe855b6332ccb63644a62ce2d6312b6/searx/searxng.msg#L6

https://github.com/searxng/searxng/blob/3201aa1b3fe855b6332ccb63644a62ce2d6312b6/searx/searxng.msg#L16-L20

dalf:
It can be the other way around:
* define all the names inside the yaml file
* load the yaml fiile when the application starts
* DEFAULT_GROUP_NAME and OTHER_CATEGORY can be replaced by references to the content of this yaml file.

return42:
> It can be the other way around:

Thats true, but this means we need to parse the yaml files at runtime .. to define _constants_ in python code coming from these yaml files .. I think this way (import constants from python code into searxng.msg) keeps the runtime performant and the code more clean.

dalf:
>  to define constants in python code coming from these yaml files ..

settings.yml doesn't seem to be an issue on this aspect.
I don't understand the performance issue.

return42:
> settings.yml doesn't seem to be an issue on this aspect.

- It is one file .. `searxng.msg` files can be many
- when you want to define a _constant_ on a module level from a yaml file, you need to parse a yaml file. 

This means whenever python modules containing such _constants_ are imported (first time) a yaml parser is started, this slows down the `import foo`.

It is really bad practice to read & parse config files on an import statement (we have this construct in webapp.py and other modules and it caused several problem, e.g the secret_key hacks we need when we want to run tasks other then webapp.py).

dalf:
> This means whenever python modules containing such constants are imported (first time) a yaml parser is started, this slows down the import foo.

I see the category definition in the same way than the translations, the data ( `searx.data` ).

> It is really bad practice to read & parse config files on an import statement

I don't understand: currently, `searxng.msg` is not referenced by the Python code.

return42:
> I don't understand: currently, `searxng.msg` is not referenced by the Python code.

Yes, that is the big pro of the solution in this PR. In https://github.com/searxng/searxng/pull/859#issuecomment-1073193895 you suggest _"It can be the other way around"_ and this suggestion needs to parse a yaml file while importing a module where a _constant_ is defined.

"Define all the names inside the yaml file" .. and a module "foo.py" that defines a constant "foo":

```python
from os import abspath, join
import yaml

yaml_file = join(abspath(dirname(__file__)), 'translation_strings.yml'))

with open(yaml_file, 'r', encoding='utf-8') as f:
    translation_strings = yaml.safe_load(settings_yaml)

foo = translation_strings['foo']
```

This PR and a module named "foo.py" that defines a constant "foo"

```python
foo = "hello"
```

As you can see, the second one is what a python developer expects, the first one is complicated by translation aspects, not related to any runtime code.

But it is not only complicated it is also ugly ..

> I see the category definition in the same way than the translations, the data ( `searx.data` ).

I think it is different; `searx.data` is loaded by calling a function this function-call can be placed (e.g.) in a try/except block to catch any issue (reading, file permissions .. whatever issue can happen) while loading the yaml file.

In the example above we define a constant on the module level. To catch any issue while loading the yaml file we need to place the `import foo` statements in try/except blocks .. 

That is the reason why I called it a "bad practice" to run code that reads from filesystem while importing the module.  At best the import of a module contains only definitions and non active code.

.. This is not only limited to python, it is always bad practice to execute code at the import time of a library.

PR #862, opened 2022-02-04
[fix] solidtorrents engine - JSON API no longer exists

### LINKED ISSUE #858: Bug: solidtorrents engine
**Version of SearXNG**
Branch: master
Version: 1.0.0 (searx.be)
Version: 1.0.0-1638-9fae80fa (search.mdosch.de)
Version: 1.0.0-1537-f0ee5103 (homeserver)

searxng-docker (for my own instance)

**What happened?**
Solidtorrent is not working. I tested it with my homeserver instance, searx.be, search.mdosch.de.

**How To Reproduce**
Select solidtorrents engine in the "files" category.
Search and select the files tab.

**Technical report**
(from searx.be)

Error
  * Error: httpx.HTTPStatusError
  * Percentage: 100
  * Parameters: `('404', 'Not Found', 'solidtorrents.net')`

Checker
  * simple: `httpx.HTTPStatusError `,
  * paging: `httpx.HTTPStatusError `,

return42:
The API endpoint, we where using does not exist anymore.  This patch is a
rewrite that parses the HTML page.

Related: https://github.com/paulgoio/searxng/issues/17
Closes: https://github.com/searxng/searxng/issues/858

### DIFF
--- a/searx/engines/solidtorrents.py
+++ b/searx/engines/solidtorrents.py
@@ -29,17 +29,16 @@ about = {
 categories = ['files']
 paging = True
 
-base_url = ''
-base_url_rand = ''
+# base_url can be overwritten by a list of URLs in the settings.yml
+base_url = 'https://solidtorrents.net'
 
 
 def request(query, params):
-    global base_url_rand  # pylint: disable=global-statement
     if isinstance(base_url, list):
-        base_url_rand = random.choice(base_url)
+        params['base_url'] = random.choice(base_url)
     else:
-        base_url_rand = base_url
-    search_url = base_url_rand + '/search?{query}'
+        params['base_url'] = base_url
+    search_url = params['base_url'] + '/search?{query}'
     page = (params['pageno'] - 1) * 20
     query = urlencode({'q': query, 'page': page})
     params['url'] = search_url.format(query=query)
@@ -70,7 +69,7 @@ def response(resp):
             'seed': seed,
             'leech': leech,
             'title': title,
-            'url': base_url_rand + url,
+            'url': resp.search_params['base_url'] + url,
             'filesize': filesize,
             'magnetlink': magnet,
             'torrentfile': torrentfile,
--- a/searx/engines/solidtorrents.py
+++ b/searx/engines/solidtorrents.py
@@ -13,6 +13,7 @@ from searx.utils import (
     extract_text,
     eval_xpath,
     eval_xpath_getindex,
+    eval_xpath_list,
     get_torrent_size,
 )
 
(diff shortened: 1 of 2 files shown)


searx/engines/solidtorrents.py
```diff
@@ -1,51 +1,85 @@
...
 
 categories = ['files']
 paging = True
 
-base_url = 'https://www.solidtorrents.net/'
-search_url = base_url + 'api/v1/search?{query}'
+base_url = ''
+base_url_rand = ''
+
+units = {"B": 1, "KB": 2 ** 10, "MB": 2 ** 20, "GB": 2 ** 30, "TB": 2 ** 40}
```
dalf: Note: [`get_torrent_size`](https://github.com/searxng/searxng/blob/de32d543bc5e7fac968a11c7cdd0e3b8f349a94d/searx/utils.py#L263-L301) in `searx.utils` convert the file size to bytes. The current implementation of this function misses the `B` unit.

The PR https://github.com/searxng/searxng/pull/839 improves the implementation: https://github.com/searxng/searxng/blob/2d5929cc59f30ab4c151c69d797da64c87df2a0a/searx/utils.py#L41-L48

searx/engines/solidtorrents.py
```diff
@@ -1,51 +1,85 @@
...
+        filesize = size2int(extract_text(stats[1]))
+        leech = extract_text(stats[2])
+        seed = extract_text(stats[3])
+        magnet = eval_xpath_getindex(result, './div//a[contains(@class, "dl-magnet")]/@href', 0, None)
+
+        params = {
+            'seed': seed,
+            'leech': leech,
+            'title': title,
+            'url': base_url_rand + url,
```
dalf: Note without consequence / can be ignored: the `base_url` here might be different the `base_url` in the `request` function because two or more user query can run in parallel.

To keep exactly the same value `base_url` must be stored in `params` and then retrieve using `resp.search_params["base_url"]`.

return42: > To keep exactly the same value `base_url` must be stored in `params` and then retrieve using `resp.search_params["base_url"]`.

AFAIK request & response are in the same process (task) environment .. isn't it?

I copied this method from

https://github.com/searxng/searxng/blob/de32d543bc5e7fac968a11c7cdd0e3b8f349a94d/searx/engines/invidious.py#L43-L49

https://github.com/searxng/searxng/blob/de32d543bc5e7fac968a11c7cdd0e3b8f349a94d/searx/engines/invidious.py#L67-L73

dalf: Two different threads ( = two different user queries) can call the `request` function in a row and then the `response` function. The namespace will be same since this is the same engine.

searx/engines/solidtorrents.py
return42: Thanks for pointing out [get_torrent_size](https://github.com/searxng/searxng/blob/de32d543bc5e7fac968a11c7cdd0e3b8f349a94d/searx/utils.py#L263-L301) I haven't seen / fixed.

searx/engines/solidtorrents.py
```diff
@@ -58,7 +56,8 @@ def response(resp):
         title = extract_text(a)
         url = eval_xpath_getindex(a, '@href', 0, None)
         stats = eval_xpath(result, './div//div[contains(@class, "stats")]/div')
```
dalf: Nit: it is possible to write `stats = eval_xpath_list(result, './div//div[contains(@class, "stats")]/div', min_len=5)` so when the HTML changes, instead of crashing with an `IndexError` few lines later, it crashes with `searx.exceptions.SearxEngineXPathException: ./div//div[contains(@class, "stats")]/div len(xpath_str) < 5`.

dalf (approved):
Thank you!

searx/engines/solidtorrents.py
return42: :facepalm: thanks for clarify, I fixed it here in solidtorrents engine and in invidious engine https://github.com/searxng/searxng/pull/867

searx/engines/solidtorrents.py
return42: good point / I fixed it to use min_len

@dalf Thanks a lot for your pragmatic suggestions & comments! .. I'm going to merge this PR ..

PR #870, opened 2022-02-04
preferences: Set autocomplete=off for form

not-my-profile:
Otherwise you can change the value of a select,
refresh the page and the preferences stay changed,
leaving the wrong impression that they were saved.


return42 (approved):
Thanks!

PR #872, opened 2022-02-05
Integrate `searx.user_help` into Sphinx documentation

### LINKED ISSUE #659: Broken links (to example.org) in official documentation
The documentation links example searches like `` :search:`query` ``. `conf.py` however fallsback to `example.org`:

https://github.com/searxng/searxng/blob/6ffa741e7a739a8b7df210054300dda8c33d75b7/docs/conf.py#L17

Users clicking on those links will be confused.

not-my-profile:
## What does this PR do?

See the commit messages.

It lets us host a version of the user help pages on docs.searxng.org.

1. `make run` and view http://127.0.0.1:8888/help/en/search-syntax
2. `make docs` and view the user documentation (don't forget to click one of the instance-specific links!)

## Related issues

#722, fixes #659


return42:
IMO rendering & adding HTML to the output is the wrong method. A Markdown to reST conversion needs to be done in a pre-build step we can add here:

https://github.com/searxng/searxng/blob/10e6881ca861cfce70e99ededf2c864a9274b5dd/manage#L413-L424

Such files (in `"${DOCS_BUILD}/includes/`) can be included by a [`.. kernel-include::`](https://return42.github.io/linuxdoc/linuxdoc-howto/kernel-include-directive.html#kernel-include-directive) directive.  If needed a `:start-after:` and `:start-after:` can be used ..

https://github.com/searxng/searxng/blob/10e6881ca861cfce70e99ededf2c864a9274b5dd/docs/admin/installation-searxng.rst#L17-L22

Hope my explanation was clear and understandable, otherwise please ask.

----

EDIT: I do not have much time ATM, but if you want, I can prepare a draft for.

not-my-profile:
No we certainly don't **need** Markdown -> ReST -> HTML. Using ReST as an intermediary format does not make any sense as far as I am concerned and would just introduce additional complexity and an additional dependency for the Markdown to ReST conversion. I think we should keep things simple and just convert Markdown -> HTML for which we already have the mistletoe dependency.

return42:
> does not make any sense as far as I am concerned and would just introduce additional complexity 

Violating the the build architecture adds complexity .. from your current standpoint it seems a easy solution but with additional requests it becomes much more complex.  Adding raw (HTML) content to a markup is hackish and limits the flexibility since it breaks the architecture.  To name a few examples: how can we link from the docs to a chapter in the raw HTML? .. what will happen when we add markups to include images or tables? .. what is about the [Theme & CSS](https://github.com/not-my-profile/searxng/blob/help-in-docs/docs/_themes/searxng/static/searxng.css)?

As I said before: the inclusion of markdown looks simple as long as we do not use any markup beside chapters or lists and as long we do not have more markup there is no difference between the skills needed for Markdown or reST. 

That is why I had a doubt since we start this Markdown discussion, but my doubts are not taken seriously.

not-my-profile:
I am not violating anything (except maybe your ReST-centric world view). I used three public APIs (`sphinx.parsers`, `docutils.nodes` and `mistletoe`) to achieve exactly what we need in a minimal amount of code.

>Adding raw (HTML) content to a markup is hackish

Haha yes ... using something the way it's intended to be used (Markdown is intended to be converted to HTML) is a hack and adding more components to a pipeline makes it more robust!

> how can we link from the docs to a chapter in the raw HTML?

The user help pages are not meant to be long ... they are meant to be short. Linking specific sections therefore shoudn't be necessary, linking to help pages should suffice ... which I implemented.

> what will happen when we add markups to include images?

We have to copy the images. Using ReST as an intermediary format wouldn't change that.

> what will happen when we add markups to include tables?

They will just work. HTML is a standard

> what is about the Theme & CSS?

They look fine.

> there is no difference between the skills needed for Markdown or reST

Could you please stop beating that dead horse? I thought you accepted the outcome of the discussion despite your disagreement.

return42:
> They look fine.

As long as you only use chapter & lists the gap is small, when you include a table you will see that you have add more and more complexity ...

>> Adding raw (HTML) content to a markup is hackish

> Haha yes ... using something the way it's intended to be used (Markdown is intended to be converted to HTML) is a hack and adding more components to a pipeline makes it more robust!

Please try to understand my concerns about architectures, I just fighting with "raw HTML" content in another topic .. https://github.com/searxng/searxng/issues/873

> Could you please stop beating that dead horse? I thought you accepted the outcome of the discussion despite your disagreement.

Pew .. even if I think it was not a good decision I try to find a solution / I mean it is one thing to use Markdown, but it is another thing to break the architecture.

Please try to understand that I am looking for a solution that fits into the tools and architectures we use .. and please be more constructive and don't hit me with your dictator style.

We are not the first who want to use Markdown in a Sphinx build chain. I haven't had time to look at, but may be this is a solution that fits https://www.sphinx-doc.org/en/master/usage/markdown.html .. You and others have decided to use Markdown and I think myst-parser ore a similar solution is the next consistent step.

not-my-profile:
> please be more constructive

Do you think that accusing a contributor of violating/breaking the "build architecture" is constructive?

> don't hit me with your dictator style

It's funny you bring that up considering the vote for Markdown vs ReST was 4:1. Anyway I'm done contributing to this project for the foreseeable future. I am not spending my free time contributing only to be accused and compared to a dictator. I know that you didn't mean it that way and I know that you appreciate my contributions ... but this is still not something I'm willing to put up with.

return42:
> Do you think that accusing a contributor of violating/breaking the "build architecture" is constructive?

I have my concern about this solution not about the author. Sorry for the "dictator" but I wanted to express, that I want to find a solution that fits while you insist on your solution as the only true and do not want to discuss alternatives.

not-my-profile:
> while you insist on your solution as the only true and do not want to discuss alternatives

Yes I have a string opinion ... as you do. I am happy to discuss ... I just expect respect. If you cannot mind your language, I'll leave.

> IMO rendering & adding HTML to the output is the wrong method. A Markdown to reST conversion needs to be done

You started out by telling me that what I'm doing is wrong and that a conversion to reST **needs** to be done. Note how you initially did not state a single reason why you thought it was wrong. Note how you used an authoritative **needs** leaving no room for alternatives.

> I think myst-parser ore a similar solution is the next consistent step

Using MyST would mean adding a third parser, which I think we can both agree on is undesirable. So we would either have to:

* use MyST for the help pages
* use MyST for the docs/

I don't think we want to use MyST for the help pages because I think we should keep their syntax to standardized Markdown only and MyST adds non-standardized extensions for everything that reST can do. I don't think we want to convert docs/ to Markdown either.

> I just fighting with "raw HTML" content in another topic

Yes using a template or API for iframes makes sense, but that's an entirely different topic.

> when you include a table you will see that you have add more and more complexity

Ok point taken. The tables do look a bit broken when they lack the `docutils` CSS class. mistletoe's HTML renderer however is extensible so adding that CSS class is trivial. I think by contributing to mistletoe to improve their API it should be possible to achieve both linkable headings and properly rendered tables in about 10 lines of code.

I think that should address your two main concerns that sections aren't linkable and that tables need a CSS class. I still don't consider both matters to be that important since I am not convinced that we actually need section links (we can just use short pages) and I don't think it matters much if the tables on `docs.searxng.org/user/` aren't rendered perfectly.

This PR solves two real UX problems:
* users don't find the user documentation since it's on a different website
* the user documentation cannot link instance specific pages (e.g. the example queries aren't linked to the instance /search)

IMO solving these UX problems is more important than fussing about table rendering and heading links we might need.

return42:
close in favor of https://github.com/searxng/searxng/pull/930

PR #877, opened 2022-02-06
[fix] docker generated new settings.yml does not work

### LINKED ISSUE #876: [question] docker generated new settings.yml does not contain instance_name/base_url of the commandline
I stated this as question as I do not know whether this is supposed to happen, but when I start 
docker run -d --name=test -p 8898:8080  -v "${PWD}/test:/etc/searxng" -e INSTANCE_NAME=test  searxng/searxng

the settings.yml does not contain the instance_name (or base_url)  settings as supplied on the line above, which I would have expected as with searx/searx this is done. 

after some digging I came to the conclusion that commit    
[yamllint] searx/settings.yml    5a7b12ee6f6285dbf5f0be164f93460839d99c98
changed the yml formating, and the sed expressions in docker_entrypoint.sh do no longer match and therefore do not replace the content anymore

I guess settings.yml or docker_entrypoint.sh should be changed to match again

return42:
In commit 5a7b12e we normalized settings.yml and c6a5cc019 and de5a8ee7d we
changed instance name to ``instance_name: "SearXNG"``.

This patch adjust the sed expressions to modify:

    dockerfiles/docker-entrypoint.sh

Closes: https://github.com/searxng/searxng/issues/876
Suggested-by:  @neiaberau

----

I have no docker here right now, I just tested the sed expressions:

```
export INSTANCE_NAME=test123
export AUTOCOMPLETE=google
export BASE_URL='https://example.org'
sed -i \
        -e "s|base_url: false|base_url: ${BASE_URL}|g" \
        -e "s/instance_name: \"SearXNG\"/instance_name: \"${INSTANCE_NAME}\"/g" \
        -e "s/autocomplete: \"\"/autocomplete: \"${AUTOCOMPLETE}\"/g" \
        -e "s/ultrasecretkey/$(openssl rand -hex 32)/g" \
       searx/settings.yml

sed -i -e "s/image_proxy: false/image_proxy: true/g" searx/settings.yml

```

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -71,23 +71,24 @@ patch_searxng_settings() {
     export BASE_URL="${BASE_URL%/}/"
 
     # update settings.yml
-    sed -i -e "s|base_url : False|base_url : ${BASE_URL}|g" \
-       -e "s/instance_name : \"searxng\"/instance_name : \"${INSTANCE_NAME}\"/g" \
-       -e "s/autocomplete : \"\"/autocomplete : \"${AUTOCOMPLETE}\"/g" \
-       -e "s/ultrasecretkey/$(openssl rand -hex 32)/g" \
-       "${CONF}"
+    sed -i \
+        -e "s|base_url: false|base_url: ${BASE_URL}|g" \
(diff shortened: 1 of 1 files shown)


dalf:
More or less related to https://github.com/searxng/searxng-docker/issues/8

PR #878, opened 2022-02-06
Fix wikidata info box images

tiekoetter:
## What does this PR do?

This PR fixes the wikidata info box images when the image proxy is enabled.

## Why is this change important?

Without the fix info box images from wikidata are not loaded correctly.

## How to test this PR locally?

- `make run`
- Search `!wd test`

## Related issues

Closes #875

Issue #875, opened 2022-02-06
Images in instant answer are not loaded if "image proxy" is enables


catfluoride:
SearXNG Version: 1.0.0-1669-10e6881c

Simple theme is used

**How did you install SearXNG?**

SearxNG was installed by "installation scripts", without Morty or Filtron.

**What happened?**

If  "image proxy" is enabled, images searches work correctly. However, images in "Instant Answers" under the  "General" category are absent and just a broken image placeholder is shown.

For instance, if I search for "Enrico Fermi" in the General category, with the Image Proxy enabled Wikipedia's image is broken.

**How To Reproduce**
This bug is reproducible. Just enable "Image Proxy" an settings GUI and perform the aforementioned search

**Expected behavior**

An image should be displayed in place of the broken image placeholder

**Screenshots & Logs**

Expected result:
![expected](https://user-images.githubusercontent.com/25054477/152677025-85823d66-728c-4f6b-850a-9722c67bc474.png)

Actual result:
![actual](https://user-images.githubusercontent.com/25054477/152677023-0821723f-bae7-44db-a9a8-c46d47e75ebb.png)

Inspection of the page source gives the following:

`<img src="/image_proxy?url=https%3A%2F%2Fcommons.wikimedia.org%2Fwiki%2FSpecial%3AFilePath%2FEnrico%2520Fermi%25201943-49.jpg%3Fwidth%3D500%26height%3D400&amp;h=dff389296b41618681172e6ca9f6a4e50b6ad32554484ff828889e64f4a419ab" title="Enrico Fermi" alt="Enrico Fermi">`

Somewhat "image_proxy" is misbehaving on the home page.

return42:
I can confirm that the image is not loaded:

```
INFO    werkzeug                      : 127.0.0.1 - - [06/Feb/2022 12:24:41] "GET /static/themes/simple/js/searxng.min.js HTTP/1.1" 304 -
DEBUG   searx.network.image_proxy     : HTTP Request: GET https://commons.wikimedia.org/wiki/Special:FilePath/Enrico%20Fermi%201943-49.jpg?width=500&height=400 "HTTP/1.1 302 Found" (text/html; charset=utf-8)
DEBUG   searx.webapp                  : image-proxy: wrong response code: 302
[...]

PR #880, opened 2022-02-07
[fix] openstreatmap: load thumbnail from uploads.wikimedia.org

return42:
Openstreatmap images are now loaded from uploads.wikimedia.org instead of
commons.wikimedia.org to prevent redirects.

With `image_proxy` enabled images from commons.wikimedia.org cant be loaded
since they are redirected.  We already discussed this issue [875] and
@tiekoetter fixed this issue in PR [878].

Related-to:
- [875] https://github.com/searxng/searxng/issues/875
- [878] https://github.com/searxng/searxng/pull/878

-----

With `image_proxy` enabled, search for `!osm paris`

before:

grafik

after:

grafik

### DIFF
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -14,7 +14,7 @@ from flask_babel import gettext
 from searx.data import OSM_KEYS_TAGS, CURRENCIES
 from searx.utils import searx_useragent
 from searx.external_urls import get_external_url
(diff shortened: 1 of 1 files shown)


return42:
@tiekoetter if you have time for, would you test this PR?

tiekoetter:
@return42 will do.

tiekoetter (approved):
The changes look good and work correctly 👍

Issue #881, opened 2022-02-07
Peertube URL is offline


return42:
`base_url` of the peertube engine does no longer response: https://peer.tube/ 

https://github.com/searxng/searxng/blob/5cb6301fbc0c71ee1154f62916bacb496ddc4fdc/searx/settings.yml#L1578-L1582

----

EDIT: is now redirected to https://sepiasearch.org/

PR #882, opened 2022-02-07
[fix] replace embedded HTML by data_src

### LINKED ISSUE #873: Embeded HTML breaks SearXNG architecture (e.g. hostname_replace)
We have embedded content in some of our engines e.g. the iframe we use for embedded videos:

https://github.com/searxng/searxng/blob/10e6881ca861cfce70e99ededf2c864a9274b5dd/searx/engines/youtube_noapi.py#L35-L39 

This raw HTML content breaks with the architecture of SearXNG, e.g. 

- Theme: the size and with should be set by CSS
- The URL of the embedded content is not covered by SearXNG: https://github.com/searxng/searxng/blob/10e6881ca861cfce70e99ededf2c864a9274b5dd/searx/engines/youtube_noapi.py#L153

To give one example; the `hostname_replace` settings are ignored by engine's response handler

```yaml
hostname_replace:
  # youtube --> Invidious
  '(.*\.)?youtube-nocookie\.com': 'invidio.xamh.de'
  '(.*\.)?youtube\.com$': 'invidio.xamh.de'
```

In general: embedding raw HTML content in the result list opens the door for various issues and can't be handled by the SearXNG architecture. It does not only break current functionality (hostname_replace) it also prevents from solving requests like https://github.com/searxng/searxng/issues/140

What we need is a HTML template that renders a `<iframe>` tag for embedded URLs.

return42:
## What does this PR do?

[mod] result_templates/**videos.html** replace embedded HTML by data_src
[mod] result_templates/**default.html** replace embedded HTML by data_src

Embedded HTML breaks SearXNG architecture.  To modularize, HTML is generated in
the templates (oscar & simple) and result parameter 'embedded' is replaced by
'data_src', an URL for embedded content (<iframe>).

[mod] hostname_replace: replace hostnames in result's data_src param

The (iframe & audio) templates of the following engines has been touched (oscar & simple theme)

- searx/engines/dailymotion.py
- searx/engines/invidious.py
- searx/engines/peertube.py --> https://github.com/searxng/searxng/issues/881
- searx/engines/sepiasearch.py
- searx/engines/vimeo.py
- searx/engines/youtube_api.py
- searx/engines/bandcamp.py
- searx/engines/deezer.py
- searx/engines/freesound.py
- searx/engines/mixcloud.py
- searx/engines/soundcloud.py

To test you need to redirect embedded videos (e.g.) from youtube to a invidios
instance.  Search for videos using engine `!youtube lebowski`.  The result URLs
and the embedded videos should link to the invidios instance.

Here is an example of such a `hostname_replace` configuration::

```yaml

    hostname_replace:

      # youtube --> Invidious

      '(.*\.)?youtube-nocookie\.com': 'invidio.xamh.de'
      '(.*\.)?youtube\.com$': 'invidio.xamh.de'
      '(.*\.)?invidious\.snopyta\.org$': 'invidio.xamh.de'
      '(.*\.)?vid\.puffyan\.us': 'invidio.xamh.de'
      '(.*\.)?invidious\.kavin\.rocks$': 'invidio.xamh.de'
      '(.*\.)?inv\.riverside\.rocks$': 'invidio.xamh.de'
```

## Related issues

Closes: https://github.com/searxng/searxng/issues/873


searx/engines/vimeo.py
```diff
@@ -64,7 +57,7 @@ def response(resp):
                 'content': '',
                 'template': 'videos.html',
                 'publishedDate': publishedDate,
-                'embedded': embedded,
+                'data_src': "https://player.vimeo.com/video/{videoid}" + videoid,
```
dalf: `{videoid}` must be removed

searx/plugins/hostname_replace.py
```diff
@@ -28,5 +28,15 @@ def on_result(request, search, result):
                 return False
             result[parsed] = result[parsed]._replace(netloc=pattern.sub(replacement, result[parsed].netloc))
             result['url'] = urlunparse(result[parsed])
+        if result.get('data_src', False):
```
dalf: Nit: `if result.get('data_src'):` should be enough since `if None` is evaluate as `if False`.

searx/plugins/hostname_replace.py
```diff
@@ -28,5 +28,15 @@ def on_result(request, search, result):
...
+            parsed_data_src = urlparse(result['data_src'])
+            if pattern.search(parsed_data_src.netloc):
+                parsed_data_src = parsed_data_src._replace(netloc=pattern.sub(replacement, parsed_data_src.netloc))
+                result['data_src'] = urlunparse(parsed_data_src)
+        if result.get('audio_src', False):
```
dalf: Nit: `if result.get('audio_src'):`

searx/static/themes/simple/src/less/style.less
```diff
@@ -296,10 +296,26 @@ article[data-vim-selected].category-social {
   padding: 0 5px 25px 0 !important;
 }
 
+.audio-control audio {
+  width: 100%;
```
dalf, line 301: See comment below

searx/templates/oscar/result_templates/default.html
```diff
@@ -3,13 +3,13 @@
...
+{%- if result.data_src -%}
     <small> &bull; <a class="text-info btn-collapse collapsed cursor-pointer media-loader disabled_if_nojs" data-toggle="collapse" data-target="#result-media-{{ index }}" data-btn-text-collapsed="{{ _('show media') }}" data-btn-text-not-collapsed="{{ _('hide media') }}" aria-labelledby="result-{{loop.index}}">{{ icon('music') }} {{ _('show media') }}</a></small>
 {%- endif -%}
 
-{%- if result.embedded -%}
-<div id="result-media-{{ index }}" class="collapse">
-   {{- result.embedded|safe -}}
+{% if result.data_src -%}
+<div id="result-media-{{ index }}" class="embedded-content invisible">
+  <iframe data-src="{{result.data_src}}" frameborder="0" allowfullscreen></iframe>
```
dalf: I would add `scrolling="no"` to avoid scroll bars.

The attribute is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-scrolling), but `overflow: hidden` is not enough for the iframe.

searx/templates/oscar/result_templates/videos.html
```diff
@@ -3,15 +3,15 @@
...
+{%- if result.data_src -%}
     <small> &bull; <a class="text-info btn-collapse collapsed cursor-pointer media-loader disabled_if_nojs" data-toggle="collapse" data-target="#result-video-{{ index }}" data-btn-text-collapsed="{{ _('show video') }}" data-btn-text-not-collapsed="{{ _('hide video') }}" aria-labelledby="result-{{loop.index}}">{{ icon('film') }} {{ _('show video') }}</a></small>
 {%- endif -%}
 
-{%- if result.embedded -%}
-<div id="result-video-{{ index }}" class="collapse">
-   {{- result.embedded|safe -}}
+{% if result.data_src -%}
+<div id="result-video-{{ index }}" class="embedded-video collapse">
+  <iframe data-src="{{result.data_src}}" width="540" height="304" frameborder="0" allowfullscreen></iframe>
```
dalf: I would add `scrolling="no"` to avoid scroll bars.

searx/templates/simple/result_templates/default.html
```diff
@@ -15,9 +15,14 @@
   </p>
 {% endif -%}
 {{- result_sub_footer(result, proxify) -}}
-{% if result.embedded -%}
-<div id="result-media-{{ index }}" class="invisible">
-   {{- result.embedded|safe -}}
+{% if result.data_src -%}
+<div id="result-media-{{ index }}" class="embedded-content invisible">
+  <iframe data-src="{{result.data_src}}" frameborder="0" allowfullscreen></iframe>
```
dalf: I would add `scrolling="no"` to avoid scroll bars.

searx/static/themes/simple/src/less/style.less
```diff
@@ -296,10 +296,26 @@ article[data-vim-selected].category-social {
...
+  padding: 10px 0 0 0;
+}
+
 .result-videos .content {
   overflow: hidden;
 }
 
+.result-videos .embedded-video iframe {
+  width: 100%;
+  aspect-ratio: 16 / 9;
```
dalf, line 316: The property is not supported by one year old browser: https://caniuse.com/?search=aspect-ratio

A possible fallback:

```less
@supports not (aspect-ratio: 1 / 1) {
  .result-videos .embedded-video iframe {
    height: calc(@results-width * 9 / 16); // the right value
  }

  @media screen and (max-width: @tablet) {
    .result-videos .embedded-video iframe {
      height: 20rem; // A guess, the video should be displayed full screen anyway...
    }
  }

  @media screen and (max-width: @phone) {
    .result-videos .embedded-video iframe {
      height: auto; // A should, the video should be displayed full screen anyway...
    }
  }
}
```

For reference: https://css-tricks.com/aspect-ratio-boxes/

dalf:
> To modularize, HTML is generated in the templates (oscar & simple)

💯 👍 

> result parameter 'embedded' is replaced by 'data_src', an URL for embedded content (<iframe>).

IMO: `embedded`  is clearer than `data_src`. Why choose `data_src`?

---

The vertical video are fine:
image

---
There are issues with the height of the audio media:

In the case of audio, there is a blank space at the bottom of the iframe:
image

In some other cases the iframe is too small: 
image

Possible solution: the engine returns an additional field named `embedded_height`.

dalf (changes requested):
See my comments

return42:
> > result parameter 'embedded' is replaced by 'data_src', an URL for embedded content (<iframe>).

> IMO: `embedded`  is clearer than `data_src`. Why choose `data_src`?

IMO embedded is not clear, at least when we look at the `default.html` template ...

```html
{% if result.data_src -%}
<div id="result-media-{{ index }}" class="embedded-content invisible">
  <iframe data-src="{{result.data_src}}" frameborder="0" allowfullscreen></iframe>
</div>
{%- endif %}
{% if result.audio_src -%}
<div id="result-media-{{ index }}" class="audio-control">
  <audio controls><source src="{{result.audio_src}}"></audio>
</div>
{%- endif %}
```

... where we have an _embedded_ `<iframe>` and an _embedded_ `audio` control.

That's why I was looking for a name that better _fits_ to the template. I don't know if `data_src` and `audio_src` are good names.  If you have better suggestions, we can change the names, but whatever names we choose they should point out their usage in the templates.  And we should consider that we might have other _embedded_ content in the future, content we do not think about today.  All in all I think _embedded_ as a name is to general.

dalf:
what about `iframe_src`  and `audio_src`?
or `embedded_iframe` and `embedded_audio`?

`data-src` is already used to lazy load images:
https://github.com/searxng/searxng/blob/a4942213a1dc6404616e239e0d2fb55a6e19a499/searx/templates/simple/result_templates/images.html#L12

searx/templates/simple/result_templates/default.html
return42: Set `scrolling="no"`  in oscar is OK for me, but in simple theme I think we should control the layout throughout in CSS.

searx/static/themes/simple/src/less/style.less
return42: > The property is not supported by one year old browser ..

I think ...

```yaml
@supports not (aspect-ratio: 1 / 1) {
  .result-videos .embedded-video iframe {
    height: calc(@results-width * 9 / 16); // the right value
  }
}
```

is an good addition to support older browsers (thanks for this suggestion).

But I don't want to implement different screen solutions, at least not in this PR.  The intention of this PR is not to get perfect layout we hadn't before, the intention is mostly to modularize HTML in templates.  We can improve the layout afterwards and discuss better layouts, BTW the aspect ratio of videos is not always 16:9 and on small devices it might be better when the "show video" button shows the video at a full screen.

searx/plugins/hostname_replace.py
return42: The entire commit needed to be reimplemented, read commit message from 55e2daf8b28

searx/templates/oscar/result_templates/videos.html
return42: fixed in a rebase https://github.com/searxng/searxng/pull/882/files#diff-149bcd9955091769a827c269387a306bba70bcf9e8e619cf4c6840611b846d51R12

searx/static/themes/simple/src/less/style.less
return42: FYI I added edc31d84

return42:
> what about iframe_src and audio_src?

I renamed data_src to iframe_src in 9c468561

> Possible solution: the engine returns an additional field named `embedded_height`.

I don't want the layout hard coded by the engines / therefore I added 1d72ada6

----

@dalf: thanks for your review and helping comments .. I fixed all your requests or leaved a comment at least.

PR is now ready for the next review cycle ..

searx/static/themes/simple/src/less/style.less
```diff
@@ -324,6 +324,26 @@ article[data-vim-selected].category-social {
...
+iframe.www\.mixcloud\.com {
+  // the embedded player from mixcloud has some quirks: initial there is an
+  // issue with an image URL that is blocked since it is an a Cross-Origin
+  // request. The alternative text (<img alt='Mixcloud Logo'> then cause an
+  // scrollbar in the inner of the iframe we can't avoid.  Another quirk comes
+  // when pressing the play button, somtimes the shown player has an height of
+  // 200px, somtimes 250px.
+  height: 250px;
+}
+
```
dalf: Without additional CSS classes on iframe:
```less
a.result.soundcloud iframe {
  height: 120px;
}

a.result.deezer iframe {
  height: 94px;
}

a.result.mixcloud iframe {
  height: 250px;
}
```

assertion: if the `hostname_replace` plugin replace some URL, the widget will have the same height.

If the property is based on the hostname, then each time the `hostname_replace` plugin configuration change, the theme has to be recompiled to add the new host.

---

There is a bug that can be fixed in another PR:
https://github.com/searxng/searxng/blob/a4942213a1dc6404616e239e0d2fb55a6e19a499/searx/templates/simple/macros.html#L22

the spaces of the engine name must be replaced by underscore.

return42: > assertion: if the hostname_replace plugin replace some URL, the widget will have the same height.

I think both concept do not really scale with players from a hostname-replace. These media players from different hostname might have a different layout (compare youtube an invidious video players).

I added a commit 2f806ea9 to have a custom-layer for CSS / I know the price is one request more for the additional `css/custom.css` but I do not know a better solution that fits into any concept.  IMO the layout of players is related to their URL (netloc) and by this we need to be able to layout by URL (netloc).

searx/webapp.py
```diff
@@ -496,6 +496,7 @@ def render(template_name: str, override_theme: str = None, **kwargs):
 
     # helpers to create links to other pages
     kwargs['url_for'] = url_for_theme  # override url_for function in templates
+    kwargs['urlparse'] = urllib.parse.urlparse
```
dalf: If some reason the  CSS attribute selector are not a good choice (?), we can reintroduce the `extract_domain` filter:
https://github.com/searx/searx/blob/0248777f959ba9ee26fc8f1c367ea4ef2cb5d7cc/searx/webapp.py#L285-L288

searx/static/themes/simple/src/less/custom.less
```diff
@@ -1,4 +1,4 @@
-iframe.soundcloud\.com {
+iframe.w\.soundcloud\.com {
```
dalf: The host names and other CSS class names are overlapping. While the chance there is an actual overlap is small, it would be better to avoid to take the risk.
 
It is possible to use CSS attribute selector: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

```less
iframe[src^="https://w.soundcloud.com"] {
  height: 120px;
}

iframe[src^="https://www.deezer.com"] {
  // The real size is 92px, but 94px are needed to avoid an inner scrollbar of
  // the embedded HTML.
  height: 94px;
}

iframe[src^="https://www.mixcloud.com"] {
  // the embedded player from mixcloud has some quirks: initial there is an
  // issue with an image URL that is blocked since it is an a Cross-Origin
  // request. The alternative text (<img alt='Mixcloud Logo'> then cause an
  // scrollbar in the inner of the iframe we can't avoid.  Another quirk comes
  // when pressing the play button, somtimes the shown player has an height of
  // 200px, somtimes 250px.
  height: 250px;
}
```

So there is no need
* to add `kwargs['urlparse'] = urllib.parse.urlparse`
* to call `urlparse` from `video.html` and `default.html` (iframe and audio element).

( changes done in commit 1d72ada65dafafeca99cbca45a549e384897f18f )

searx/plugins/hostname_replace.py
```diff
@@ -17,16 +17,30 @@
...
+        if parsed in result:
+            if pattern.search(result[parsed].netloc):
+                # to keep or remove this result from the result list depends
+                # (only) on the 'parsed_url'
+                if not replacement:
+                    return False
+                result[parsed] = result[parsed]._replace(netloc=pattern.sub(replacement, result[parsed].netloc))
+                result['url'] = urlunparse(result[parsed])
+
+        if not replacement:
```
dalf: Currently it is okay.
However, some engines return YouTube videos (like qwant, google video, bing video ?)
Currently the `iframe_src` attribute is not set for these engines.
If in the future, this is attribute is set , this condition is going to prevent the hostname replacement in the `iframe_src` attribute.

searx/plugins/hostname_replace.py
```diff
@@ -17,16 +17,30 @@
...
+            continue
+
+        for url_field in _url_fields:
+            if result.get(url_field):
+                url_src = urlparse(result[url_field])
+                if pattern.search(url_src.netloc):
+                    url_src = url_src._replace(netloc=pattern.sub(replacement, url_src.netloc))
```
dalf:
```python
if pattern.search(url_src.netloc):
  if remplacement:
    del result[url_field]
  else:
    url_src = url_src._replace(netloc=pattern.sub(replacement, url_src.netloc))
    result[url_field] = urlunparse(url_src)
```

otherwise the code is going to crash on line 43 if `replacement` is `False` (which is allowed in the configuration).

searx/static/themes/simple/gruntfile.js
```diff
@@ -138,6 +138,12 @@ module.exports = function (grunt) {
             nonull: true,
             filter: file_exists,
           },
+          {
+            src: ['src/less/custom.less'],
+            dest: 'css/custom.css',
+            nonull: true,
+            filter: file_exists,
```
dalf: I can't see the benefit of an additional CSS file: `custom.css` is built file like the other built files => `custom.less` won't be customize because it requires to build the themes, so `git pull` is not going to work.

So, despite what I've said before, and after a second thought, I suggest to make it simple:
* rename custom.less to embedded.less
* move the CSS rules about embedded media from style.less to embedded.less ( `.embedded-content iframe {`, `.audio-control audio {`, etc...)
* include embedded.less in the normal build process (if it is not already done).
* leave the customization for later.

So everything about embedded media is in one place.

dalf (changes requested):
After few days rereading the code, I end up to think there is no need to for a custom.less file (for now).

The code inside this file can simplified, see the comments.

After that, I think the PR is ready.

searx/plugins/hostname_replace.py
return42: > If in the future, this is attribute is set , this condition is going to prevent the hostname replacement in the iframe_src attribute.

Not sure I fully understand you .. but:

- if there is no replacement (``if not replacement``)  we can't replace a URL in the iframe
- the hostname_replace plugin can only drop results from the result list, but it can't drop an iframe content.

searx/webapp.py
return42: I didn't aware of CSS attribute selector / thanks for pointing out.

I removed this commit,

searx/static/themes/simple/src/less/custom.less
return42: Thanks, implemented in 8acf540f

searx/plugins/hostname_replace.py
return42: I rewrote this commit, see 795e8af6

searx/plugins/hostname_replace.py
return42: I rewrote this commit, see 795e8af6

searx/static/themes/simple/gruntfile.js
return42: > include embedded.less in the normal build process (if it is not already done).

A ``@import "embedded.less"`` in `searx/static/themes/simple/src/less/style.less` is needed  .. see 8acf540f

return42:
> The code inside this file can simplified, see the comments.
> After that, I think the PR is ready.

@dalf I dropped several commits and rewrote others, all I changed is documented in the answers to your comments above.

searx/static/themes/simple/gruntfile.js
dalf: Nit: I would have move the CSS rules about embedded media from style.less to embedded.less ( .embedded-content iframe {, .audio-control audio {, etc...)

dalf (approved):
Thank you for this contribution.
It fixes an long standing issue.

---

In addition to the test described, I have check this configuration:
```yaml
hostname_replace:
  '(.*\.)?youtube-nocookie\.com': false
```
👍 it works

PR #884, opened 2022-02-07
[mod] disable deezer engine by default

return42:
To play content from deezer a account is needed, the majority of the SearXNG won't have.


return42:
Strange: sometimes I can listen content from deezer without an account, sometimes not.

Does anyone know how deezer works?

PR #885, opened 2022-02-08
[mod] simple theme: define a '.show-content-button'

return42:
## What does this PR do?

With LESS function `.show-content-button` all the *Show-Media* links becomes bottons.

The *Show-Media* button is decent and becomes highlighted when mouse is over, here is an example from the video category ...

grafik

Here is an example of the torrent category with buttons for the magnet-link and the torrent-link ...

grafik

And here is an example for show/hide button of a media that is rendered by the `result_templates/default.html` template..

grafik


dalf:
IMO:
* `Show video` and `Hide video` are buttons. 👍 
* `Torrent` and `Magnet Link` are links: it would be better to display them as links.

return42:
> * `Torrent` and `Magnet Link` are links: it would be better to display them as links.

OK .. IMO the `torrent.html` needs more polish, not only the links but we can discuss this in another PR.

I removed  the `torrent.html` hunk and rebased the commit to avoid /static conflicts with master.

dalf (approved):
It is better especially on mobile where the tiny link was not easy to target.

Ideally, IMO, the video should be opened in the same way than the full size image: on the right side or full screen.

return42:
> Ideally, IMO, the video should be opened in the same way than the full size image: on the right side or full screen.

Sounds good to me .. but first lets make simple theme the default theme .. and drop oscar ;) .. we can talk about this tomorrow ..

PR #892, opened 2022-02-12
plugin limiter: check only /image_proxy and /search

dalf:
## What does this PR do?

* check only `/image_proxy` and `/search` : other path would make sense if the limiter is external tool like filtron to limit a DOS.
* also adjust the number of req/time according to Paul & my experiences.

Don't get blocked even on a personal instance.

## Related issues

https://github.com/searxng/searxng/pull/618


dalf:
The limiter is still experimental.
(seen with @mrpaulblack)

PR #893, opened 2022-02-12
[mod] mediathekviewweb engine: add data_src and use videos template

return42:
## What does this PR do?

[mod] mediathekviewweb engine: add data_src and use videos template

Look videos from the german-speaking public broadcasters directly in the result list:

grafik

Search for `!mvw anstalt` and press _show video_ button


mrpaulblack:
This also needs a change in the CSP for searxng-docker btw. Otherwise loading the content for the CSP is blocked by current CSP...

return42:
I don't know what what the value of CSP is in docker .. tbh we need some CSP handling in SearXNG core.

Issue #818, opened 2022-01-25
Simple theme: Web Accessibility


dalf:
The simple theme should implement the ARIA recommendations.

Here some items & links that can help to the basic steps:
* https://developer.mozilla.org/en-US/docs/Tools/Accessibility_inspector
* https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/
* on a google.com result page, press tab various accessibility links including "skip to content": https://css-tricks.com/how-to-create-a-skip-to-content-link/
* the preferences, engine tab should have a "disable all engines" / "enable all engines" (in the tab, not globally)

----

Some of the issues:
* [ ] The tabs, the categories and the checkboxes (in the preferences) can't get the focus ( [`display: none`](https://github.com/searxng/searxng/blob/a4942213a1dc6404616e239e0d2fb55a6e19a499/searx/static/themes/simple/src/less/search.less#L33) must be replaced by [`appearance: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance)):
   * [x] checkbox: https://codepen.io/dalf-the-encoder/pen/VwrQKKm
   * [ ] tabs 
* [x] #1316

dalf:
Some experiment with the checkbox: https://codepen.io/dalf-the-encoder/pen/VwrQKKm?editors=1100 
* the checkbox can get the focus (it is not possible currently because of [`visibility: hidden;`](https://github.com/searxng/searxng/blob/12b5f9e5611d40d80de7efdfc1fd6afae6c8cdaf/searx/static/themes/simple/src/less/toolkit.less#L405))
* the focus is shown
* the checkbox can be disabled
* it is possible to click anywhere to uncheck/check (the circle or the horizontal bar).
* there is a checkmark when the checkbox is checked (perhaps useless)

![image](https://user-images.githubusercontent.com/1594191/154773774-9f794521-e3ea-4cdc-9d4b-c9a2cba1ff0d.png)

... may be the usual checkboxes are better (but styled with CSS so it is not browser dependent).

What are your opinions?

PR #894, opened 2022-02-12
[enh] simple: basic ARIA fixes

dalf:
## What does this PR do?

Fix some ARIA issues reported by FF devtools / Accessibility tab.
For reference: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/complementary.html

* Check FF devtools / Accessibility tab / Check for issues
* Check the result of
  * https://addons.mozilla.org/en-US/firefox/addon/wave-accessibility-tool/
  * https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/
* Try a screen reader:
  * Orca on Linux
  * NVDA on Windows
  * [ChromeVox](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn/related?hl=en-GB) for Chrome

## Author's checklist

I didn't wanted to change the CSS in this PR, but there is a need to do in another PR to fix these issues:
* some focusable element are not hightlighted for example.
* the ["Skip to content"](https://css-tricks.com/how-to-create-a-skip-to-content-link/) link is missing because it requires to change the CSS.

accessibility / a11y is whole universe, and I'm not an expert: I've tried to keep the changes that seems to improve the UX with a screen reader. The idea was to add as few as possible rather too much and counterproductive. However I'm have failed here and there.

I'm not sure the usage of `fieldset` HTML element in the preferences:
* the `legend` HTML element is not a `label`
* a FF & Chrome browser bug doesn't allow to the change the `display` property of the `legend` inside a `fieldset`: https://stackoverflow.com/questions/62692452/cant-position-html-legend-tag-with-css-grid

## Related issues

Related to  #818

### DIFF
--- a/searx/templates/simple/filters/languages.html
+++ b/searx/templates/simple/filters/languages.html
@@ -1,4 +1,4 @@
-<select class="language" id="language" name="language" tabindex="1">{{- '' -}}
+<select class="language" id="language" name="language" aria-label="{{ _('Search language') }}">{{- '' -}}
 	<option value="all" {% if current_language == 'all' %}selected="selected"{% endif %}>{{ _('Default language') }}</option>
 	{%- for lang_id,lang_name,country_name,english_name in language_codes | sort(attribute=1) -%}
 	<option value="{{ lang_id }}" {% if lang_id == current_language %}selected="selected"{% endif %}>
--- a/searx/templates/simple/filters/safesearch.html
+++ b/searx/templates/simple/filters/safesearch.html
@@ -1,5 +1,5 @@
-<select  name="safesearch" id="safesearch" class="safesearch" tabindex="3">
-    <option value="2" {% if safesearch == '2' %}selected="selected"{% endif %}>{{ _("SafeSearch") + ": " + _('Strict') }}</option>
-    <option value="1" {% if safesearch == '1' %}selected="selected"{% endif %}>{{ _("SafeSearch") + ": " + _('Moderate') }}</option>
-    <option value="0" {% if safesearch == '0' %}selected="selected"{% endif %}>{{ _("SafeSearch") + ": " + _('None') }}</option>
+<select  name="safesearch" id="safesearch" class="safesearch" aria-label="{{ _('SafeSearch') }}">
+    <option value="2" {% if safesearch == '2' %}selected="selected"{% endif %} aria-label="{{ _('Strict') }}">{{ _("SafeSearch") + ": " + _('Strict') }}</option>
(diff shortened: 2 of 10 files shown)


dalf:
ping @mrpaulblack 

another PR is required to update the tabs in the preferences:
* so the user can use the `Tab` key to navigate
* to fix https://github.com/searxng/searxng/issues/945

return42 (approved):
I tested this PR (roughly / I'm not aware of all ARIA aspects) and I think it is huge step forward --> Lets merge this PR

FYI I rebased on current master.

mrpaulblack:
Agreed. Sorry I put the review off again and again so after Markus tested it I am also for merging (Looking at the files the aria labels seems clear IMO :+1: and the reset works)

mrpaulblack:
@dalf After some more testing the new reset impl. does not clear the input filed when the user already searched for something. It will only reset to the search term that the user searched for...

dalf:
That's the expect behavior of a reset button

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/reset : 

> [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) elements of type reset are rendered as buttons, with a default [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event handler that resets all of the inputs in the form to their initial values

I don't know how to clear the input without JS.

Fix for the JS version:

```patch
diff --git a/searx/static/themes/simple/src/js/main/search.js b/searx/static/themes/simple/src/js/main/search.js
index 6ef95f5b..1fb0a7cb 100644
--- a/searx/static/themes/simple/src/js/main/search.js
+++ b/searx/static/themes/simple/src/js/main/search.js
@@ -31,10 +31,11 @@
 
     // update status, event listener
     updateClearButton();
-    cs.addEventListener('click', function () {
+    cs.addEventListener('click', function (ev) {
       qinput.value = '';
       qinput.focus();
       updateClearButton();
+      ev.preventDefault();
     });
     qinput.addEventListener('keyup', updateClearButton, false);
   }
```

mrpaulblack:
only happens with the non JS version tho;
I am reading: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input maybe there is some attribute for input to get around this?

dalf:
[<input type="search"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) may include a clear button (on Chrom(e|ium) browser, but not on FF).

mrpaulblack:
@dalf so first of all I misunderstood an that issue happens in both JS and non JS version in current master. I tried around for a while in HTML, but looks like this is not really possible since it will always reset to value. If you have time could you please make a PR with the JS fix? I mean this way it at least works with JS and kinda works with no JS :+1:

[EDIT] qwant has the same problem without JS: https://www.qwant.com/?q=time

PR #899, opened 2022-02-17
[limiter] update

dalf:
## What does this PR do?

New update of the limiter code / rules based on testing on paulgo public instance and on my personal instance.

Changes:
* The limitation only applies to `/search`
* the two sliding window counter are always updated even if the request is going to denied

While it is far for perfect, it seems to be at least comparable to filtron.

* configure redis
* enable the limiter
* checks the various rules:
  * curl
  * send multiple requests (use the `sha256 dummy` to avoid a quick response without sending outgoing HTTP request)

## Author's checklist

In this case a lot of thing can be simplified:
* remove `searxng/utils/filtron.sh` and `searxng/utils/morty.sh`.
* remove filtron & morty from the searxng-docker project, and add redis (so the project supports multiple architectures)
* perhaps cleanup [uwsgi.ini](https://github.com/searxng/searxng/blob/288e3086d2a67aaee45576f13b032f31ab782b45/dockerfiles/uwsgi.ini#L43-L44)

What can be fixed later:
* I'm not sure the limiter should be a plugin: even if the user could enable / disable the plugin, this wouldn't matter: the code hooks the Flask application at loading. 
* The current code initialize the redis connection: it works because this is the only module to use redis. The initialization of the connection should be done in the `searx.shared` module.

## Related issues

* https://github.com/searxng/searxng/pull/618
* https://github.com/searxng/searxng/pull/892

### DIFF
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -42,19 +42,16 @@ def is_accepted_request(inc_get_counter) -> bool:
             return False
         return True
 
-    if request.path == '/search' and ('q' in request.args or 'q' in request.form):
-        c = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
-        if c > 30:
-            return False
-
-        c = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
-        if c > 300:
+    if request.path == '/search':
+        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
+        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
+        if c_burst > 15 or c_10min > 150:
             return False
 
         if re_bot.match(user_agent):
             return False
 
-        if 'Accept-Language' not in request.headers:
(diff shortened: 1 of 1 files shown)


FarisZR:
Quick question, filtron is no longer needed for seaxng after this PR?

mrpaulblack:
Yes I think it got already removed from the scripts and there is a PR that will also remove filtron as well as morty from searxng-docker: https://github.com/searxng/searxng-docker/pull/12

FarisZR:
> Yes I think it got already removed from the scripts and there is a PR that will also remove filtron as well as morty from searxng-docker: [searxng/searxng-docker#12](https://github.com/searxng/searxng-docker/pull/12)

Oh great,
the PR also adds Redis, It's still not clear to me what it does exactly, does it cache results ? And if it does what is the cache invalidation policy ?

return42:
Yes, ATM you can still use filtron or you can use the [limiter plugin](https://docs.searxng.org/src/searx.plugins.limiter.html) that needs a redis DB, what answers your question https://github.com/searxng/searxng/discussions/993

Sorry for the poor documentation, I am working on a PR to modify shell scripts and improve the documentation https://github.com/searxng/searxng/compare/master...return42:nginx-searxng

But I still haven't had time to test & fix all and to pull a PR.

PR #916, opened 2022-02-20
Convert the infinite_scroll plugin as a preference (second version)

### LINKED ISSUE #759: Infinite scroll broken without scrollbar
Infinite scroll does not work when the results page fits on the screen. This happens from time to time and I think the reason is fewer results being returned or the result text being minimal. This can also be reproduced when zooming out until the scrollbar disappears. I think this could possibly cause problems with different screen sizes/resolutions. Maybe there's a way to keep loading results until a scrollbar appears or detect when the user uses the scroll wheel?

dalf:
## What does this PR do?

Difference with #812 :

Use IntersectionObserver
* most probably fix https://github.com/searx/searx/issues/759
* for reference 
  * https://caniuse.com/?search=IntersectionObserver 
  * https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
  * a YT video about this subject: https://www.youtube.com/watch?v=2IbRtjez6ag

When there are only images, the `#sidebar` is hidden

The infinite scroll is a long missing feature of the simple theme.

Same protocol than PR #812 : 
* with different browser, at least Firefox and Chrom* : 
   * in the simple theme, and in the oscar theme:
     * check a request in the general category
     * check a request in the images category
     * check a request `!general !images time`

## Author's checklist

Extension: if `searxng.infinite_scroll_supported == false`, the preferences can disabled the infinite scroll preferences.

## Related issues

* close #812
* Part of https://github.com/searxng/searxng/issues/769


return42 (approved):
Could you remove no longer needed searx/static/plugins/js/infinite_scroll.js

Lazy load of images does not work / see https://github.com/searxng/searxng/pull/812#issuecomment-1037306233

For me it is OK to merge infinite scroll and fix the lazy load of images afterwards.

dalf:
> Could you remove no longer needed searx/static/plugins/js/infinite_scroll.js

done

mrpaulblack (approved):
LGTM :+1:

PR #917, opened 2022-02-21
[mod] version formatting

### LINKED ISSUE #915: Switch versioning format since SearXNG is rolling release
Right now we use the shortened git commit SHA as version. This can be hard for people to know which version an instance is running on.

Instad of displaying version 1.0.0-commit_sha we should do a version like this for example: `2022-02-20-1` This way its more straight forward what version and instance is running and how old this version is. The `1.0.0` is not really needed in rolling release IMO.

Use the shortened commit SHA as version but still drop the `1.0.0`.

**Additional context**

tiekoetter:
## What does this PR do?

Change SearXNG version formatting to YYYY.MM.DD-commit_hash

- `make run`
- Check version

Closes #915

### DIFF
--- a/searx/version.py
+++ b/searx/version.py
@@ -2,7 +2,6 @@
 # lint: pylint
 # pylint: disable=,missing-module-docstring,missing-class-docstring
 
-import re
 import os
 import shlex
 import subprocess
@@ -59,25 +58,8 @@ def get_git_url_and_branch():
 
 
 def get_git_version():
-    try:
-        tag = subprocess_run("git describe HEAD")
-        # a. HEAD is on tag name, example: tag = "v1.0.1"
-        # b. HEAD is not a tag name, example "<tag>-<distance>-g<commit>"
-        tag_version, tag_distance, tag_commit = (tag.split("-") + ["", ""])[:3]
-        if re.match(r"v[0-9]+\.[0-9]+\.[0-9]+", tag_version):
-            # tag_version "v1.0.0" becomes "1.0.0" (without the v)
-            # other patterns are kept untouched
-            tag_version = tag_version[1:]
-        # remove "g" prefix from tag_commit
-        if tag_commit and tag_commit[0] == "g":
-            tag_commit = tag_commit[1:]
-        # set git_version to "1.0.0-590-0686e274" or '1.0.0'
-        git_version = "-".join(filter(bool, [tag_version, tag_distance, tag_commit]))
-    except subprocess.CalledProcessError:
-        # fall back to "YYYY.MM.DD.Hash" if there is no tag at all
-        git_version = subprocess_run(r"git show -s --format='%as-%h'")
-        # PEP 440: replace - with .
-        tag_version = git_version = git_version.replace("-", ".")
+    git_commit_date_hash = subprocess_run(r"git show -s --format='%cs-%h'").replace("-", ".", 2)
+    tag_version = git_version = git_commit_date_hash
 
     # add "-dirty" suffix if there are uncommited changes except searx/settings.yml
     try:


searx/version.py
```diff
@@ -59,25 +58,9 @@ def get_git_url_and_branch():
...
-            tag_commit = tag_commit[1:]
-        # set git_version to "1.0.0-590-0686e274" or '1.0.0'
-        git_version = "-".join(filter(bool, [tag_version, tag_distance, tag_commit]))
-    except subprocess.CalledProcessError:
-        # fall back to "YYYY.MM.DD.Hash" if there is no tag at all
-        git_version = subprocess_run(r"git show -s --format='%as-%h'")
-        # PEP 440: replace - with .
-        tag_version = git_version = git_version.replace("-", ".")
+    git_commit_date = subprocess_run(r"git show -s --format='%as'").replace("-", ".")
+    git_commit_hash = subprocess_run(r"git show -s --format='-%h'")
```
return42: Instead of the Auth-Date [%as](https://git-scm.com/docs/git-show#Documentation/git-show.txt-emasem) we should use the Commit-Date [%cs](https://git-scm.com/docs/git-show#Documentation/git-show.txt-emcsem).  

When a fork is rebased, the Author-Date is unchanged while the Commit-Date is updated. By example my production branch [darmarit.org](https://github.com/return42/searxng/commits/darmarit.org) has a [ [brand] commit ](https://github.com/return42/searxng/commit/0ce7fa3f161b083c6716aaeda3a65e4fed555993) on top.  When I update this branch, I rebase this branch on [SearXNG's master](https://github.com/searxng/searxng/tree/master). 

```
$ git show -s --format='%as'
2021-10-27

$ git show -s --format='%cs'
2022-02-21
```

@dalf: what do you think about `-` versus `.` in the date (year)? .. I prefer the `-` from the git command:

```
$ git show -s --format='%cs-%h'
2022-02-21-0ce7fa3f
```

.. the dot-nation can be more easily confused with the local date notation of the countries I think.

return42 (changes requested):

searx/version.py
mrpaulblack: After reading https://stackoverflow.com/questions/11856983/why-git-authordate-is-different-from-commitdate I like author date more;

If a user has a fork and they add commits from SearXNG upstream the version string is going to update to a newer version when using commit date, even tho there are no new features in the commit compared to the older version string in SearXNG upstream. With the author date we make sure that the version date stays the same even in forks for a commit...

Also I would propose as a version string format: `YYYY-MM-DD_hash`...

[EDIT] Markus convinced me that commit date is better :+1:

dalf: > @dalf: what do you think about - versus . in the date (year)? .. I prefer the - from the git command:

The dot notation is used by
* Arch: https://archlinux.org/releng/releases/
* Invidious: https://api.invidious.io/
* I can't find a project using the dash notation.

I guess the dot notation is expected for a version.

So, I would choose the dots rather than the dash.

---

lines 61 and 62 can be written in one line:

```python
tag_version = subprocess_run(r"git show -s --format='%as-%h'").replace("-", ".", 2)
```
* or with `replace("-", ".", 2)` is the dash notation is chosen
* `%cs` or ` %as`

return42: > convinced me that commit date is better +1

and @dalf convinced me that dot notation for year notation is better :smiley: 

@tiekoetter could you update the PR?

tiekoetter: Changes:
- Change author date to commit date
- Keep YYYY.MM.DD-commit_hash
- Rewrite line 61 and 62 into one line.

Is this correct?

return42 (approved):
Thanks!

PR #927, opened 2022-02-25
[simple] make checkboxes accessible

dalf:
## What does this PR do?

* The checkboxes can get the focus
* The checkboxes are clickable on 
* In addition, engine tab in the preferences:
  * the checkboxes get an ARIA label
  * the engine name is label : the user can click on a engine name to enable / disable the engine.
* trim down the number of HTML element to show a checkbox.
* remove some spaces here and there in `preferences.html`
* the `/preferences` page size is divided by two (140ko instead of 299ko).

image

image

---

Do note that checkboxes in the engines tab are displayed reversed (checked is displayed unchecked) because of:
* https://github.com/searxng/searxng/blob/3408d061aab9abc6168fec9bbc6deab71b236dac/searx/templates/simple/preferences.html#L313
* code in `preferences.py`

The CSS  class `reversed-checkbox` allows to reverse the checkbox meaning.
* Implementation : https://github.com/searxng/searxng/blob/31b6118e3f8d8553ea4267b6760fc3659c8b7a0b/searx/static/themes/simple/src/less/toolkit.less#L427-L453
* Usage : https://github.com/searxng/searxng/blob/31b6118e3f8d8553ea4267b6760fc3659c8b7a0b/searx/templates/simple/preferences.html#L101

Improve the web accessibility

* check the `/preferences` page (light and dark versions):
   * change the engine selections, save, go back to the preferences, check 
   * click on a engine name
   * use tab, shift+tab

## Related issues

Related to https://github.com/searxng/searxng/issues/818


return42:
> Do note that checkboxes in the engines tab are displayed reversed (checked is displayed unchecked) because of:
> * https://github.com/searxng/searxng/blob/3408d061aab9abc6168fec9bbc6deab71b236dac/searx/templates/simple/preferences.html#L313
> * code in `preferences.py`

Not sure I have understand all, but why do we need this "reverse" meaning? Is this `not in` an alternative:

    checkbox_onoff(engine_id, (search_engine.name, categ) not in disabled_engines) 

.. if not, I do not understand the meaning of `disabled_engines`.

dalf:
> .. if not, I do not understand the meaning of disabled_engines.

For some reason `disabled_engines` here seems to contains the enabled engines: the `not in` alternative shows the enabled engines as disabled.

I don't know where the swap between enabled and disabled engines happens (it's in the Python code), that's why I have added a CSS class to visually swap checked / unchecked.

The issue must be fixed, but that's the topic of another PR, IMO.

searx/static/themes/simple/src/less/toolkit.less
```diff
@@ -378,92 +378,127 @@ select {
...
+    position: absolute;
+    top: -0.5em;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    font-size: 0.75em;
+    width: 1.875em;
+    height: 1.875em;
+    border-radius: 50%;
+    transition: left 0.25s;
```
mrpaulblack: can you please add the transition to animations.less instead?

mrpaulblack (changes requested):
Everything LGTM (the ticket AC work correctly and the CSS looks a ok to me); please just move the transition to the dedicated animations.less (I think its best to keep those all in 1 file to make it easier to maintain transitions and animations in the future...)

searx/static/themes/simple/src/less/toolkit.less
dalf: done

mrpaulblack (approved):
LGTM :+1:

Issue #928, opened 2022-02-27
.


throwaway242685:

dalf:
Thank you for the detail report, it is really helpful. We want to drop the Oscar theme to maintain only one theme ;
but the issues you are reporting are all valid, so the idea is fix the Simple theme.

> the 'Clear' button of the Simple theme doesn't seem to do anything with JS disabled.

it is going to be fixed with https://github.com/searxng/searxng/pull/894 (commit 6352496c30c1653b523942eb9a838df921ea1d9c)

> On mobile (with JS disabled). 

Currently when Javascript is enabled:
| Simple | Oscar |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/1594191/155874482-604e576c-1998-41e7-aada-00a19e67cd54.png) | ![image](https://user-images.githubusercontent.com/1594191/155874496-ccb7a3b8-be25-4865-b9cc-3785257b2a77.png) |

... I see that it can works without javascript (need some tests).
So a pull request is going to fix that (see https://github.com/searxng/searxng/pull/929 )

> On desktop. 

What is your screen resolution?
It is seems related to https://github.com/searxng/searxng/issues/874

Can you try to set the zoom at 90%?

GreenLunar:
Simple is much nicer, especially for people not liking JS, IMHO.  See Image results as an example for it.
I suggest to keep Simple as default and fix it where Oscar does better and wherever else needed.

GreenLunar:
Proposed fixes:

> Desktop
> ![sshot-2022-02-26-19-04-49](https://user-images.githubusercontent.com/85243826/155862806-67d63abf-9cdd-4899-a077-e9f962160bdd.png)

Place suggestions in a single column to the left.

> On mobile (with JS disabled)
> ![screen_20220226-190638](https://user-images.githubusercontent.com/85243826/155862852-53de0a11-986b-4ee8-9940-30d794e072e3.png)

Place categories in a single row.
Keep suggestions untouched.

GreenLunar:
I see what the problems are when JS is disabled and when using mobile interface.
Fix the Simple theme, don't replace it.  Simple theme is one of the few the,es (I guess 2), that work properly with and without JS.  Oscar must have JS enabled to work properly, especialy in image search.

dalf:
@worthless133 on desktop, can you try to set the zoom at 90%? The issue seems related to #874 , and there is a fix for this issue.

worthless133:
[...]

PR #929, opened 2022-02-27
[fix] simple: categories: use media query detect touch screen

dalf:
## What does this PR do?

In the master branch, touch screen are detected using javascript.

Unfortunately, mobile browser with JS disable are not taken into account.

This PR uses media query to detect touch screen:
```
@media screen and (max-width: @tablet) and (hover: none) {
```

In devtools, it still requires to reload the page.

For reference:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1486960

---

A second change when the message "Click on the magnifier to perform search" is displayed:
* in the master branch, this message is always displayed when JS is disabled.
* in this PR, the message is displayed  only when the mouse is over the categories.

Better UX in mobile without JS.

* in devtools, without JS, view a result page with and without support for touch event.

## Author's checklist

A CSS variable could be improvement like in open-props:  https://open-props.style/#media-queries

About `searx/static/themes/simple/src/less/search.less` : the same elements are referenced either by id, either by class name. It makes the code not easy to follow. Example:
* `#categories_container {  position: relative;  }` is equivalent to `.search_categories { position: relative; }` (see at the end of search.less).
* so `.search_categories:hover .help { position: absolute; left: -0.1rem; }` works as expected (move an absolute position inside a relative position).

## Related issues

related to https://github.com/searxng/searxng/issues/928


mrpaulblack (approved):
LGTM :+1: works great on mobile without as well as with JS. Also helps the layout a lot on mobile devices that dont support touch input...
image

With touch input:
image

Issue #808, opened 2022-01-23
What is the content of the user help?


dalf:
Related to  #758

What is the content of the user help?

I see:
* the about page
* https://docs.searxng.org/user/search_syntax.html
* https://docs.searxng.org/user/own-instance.html
* may be https://docs.searxng.org/dev/search_api.html
* may be a page about the difference SearXNG and searx

@return42 @not-my-profile @mrpaulblack can you make a list of what you expect for the user help?
It can be section title in a page, a page title, a topic title (topic that cover multiple pages) = there is no need for write down the actual content, just the titles ; the purpose of this issue is to have an overview.

not-my-profile:
I was planning on:

1. https://docs.searxng.org/user/
2. better documentation on categories / how to search / the search syntax

I don't think `/dev/search_api.html` belongs in the user docs. And explaining differences between SearXNG and searx in the README should suffice.

return42:
I think only content that is related to the end user

- "about page"
- "search syntax"

This both parts can (or will) be reorganized.

I think these are not user docs ..

- "Why use a private instance?" --> its more dedicated to devOps
- "Search API" --> not related to an end user
- May be a page about the difference SearXNG and searx --> NO .. but a hint about searx.space might be good.

About the content: please take in mind, its an end user who is looking for some help, contact information and information related to the instance he is using at this moment.

He has its own reason about the instance he uses ("I trust my friend hosting this instance"), there is no need for a comparison with other (meta-) search engines (searx, ddg, ...).

not-my-profile:
>"Why use a private instance?" --> its more dedicated to devOps

I think the user documentation should educate users about trust. So they can choose between:

* continuing to use the current instance
* using a different instance (=> link to searx.space)
* setting up their own instance (=> link to the admin docs)

return42:
> I think the user documentation should educate users about trust.

Hm .. I think 95% of the users I invited to use my instance will not understand anything about .. they just trust me / another reason why thy use my instance: their experience is, that the results of the instance are better than what they get from google !!! :-)

not-my-profile:
Sure but there almost certainly are also users who are just using some random instance without having thought about their trust relationship. I think for such users having something written about trust in the user docs would be nice.

return42:
Other question: will we remove userdoc entirely from https://docs.searxng.org/ ?

[...]

PR #930, opened 2022-02-27
Integrate the user documentation into the application

### LINKED ISSUE #659: Broken links (to example.org) in official documentation

return42:
## What does this PR do?

Integrate the user documentation into the application

Online documentation: rename /help to /doc

Build HTML documentation and visit http://0.0.0.0:8000/user/search-syntax.html

    make docs.clean docs.live

Start `make run` session and visit pages:

    http://127.0.0.1:8888/doc/about
    http://127.0.0.1:8888/doc/search-syntax

## Related issues

- https://github.com/searxng/searxng/pull/872
- #722 
- #659

Closes: #659, #722


dalf:
image

* It would be better if the "About SearXNG" was the tab.
* `--color-doc-code-background` is too dark compare to the other elements in the document. (`#443333` or `#442222` might better but not the best choice) 

I'm not sure this is my setup, `make docs` / `make docs.live` doesn't work:

```
  File "/home/alexandre/code/searxng/local/py3/lib/pypy3.8/site-packages/sphinx/config.py", line 340, in eval_config_file
    exec(code, namespace)
  File "/home/alexandre/code/searxng/docs/conf.py", line 49, in <module>
  File "/home/alexandre/code/searxng/searx/engines/__init__.py", line 292, in load_engines
    engine = load_engine(engine_data)
  File "/home/alexandre/code/searxng/searx/engines/__init__.py", line 139, in load_engine
    set_loggers(engine, engine_name)
  File "/home/alexandre/code/searxng/searx/engines/__init__.py", line 152, in set_loggers
RuntimeError: dictionary changed size during iteration

  File "/home/alexandre/code/searxng/searx/network/network.py", line 399, in done
    future.result(3)
  File "/home/alexandre/.pyenv/versions/pypy3.8-7.3.7/lib/pypy3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/home/alexandre/.pyenv/versions/pypy3.8-7.3.7/lib/pypy3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/home/alexandre/.pyenv/versions/pypy3.8-7.3.7/lib/pypy3.8/asyncio/tasks.py", line 282, in __step
    result = coro.throw(exc)
  File "/home/alexandre/code/searxng/searx/network/network.py", line 275, in aclose_all
....
```

return42:
> `RuntimeError: dictionary changed size during iteration`

From time to time I have seen this issue in the CI / but could not reproduce it local. Can't remember exactly when I have seen this issue the first time, but I guess it is about two month ago.

manage
```diff
@@ -419,6 +419,7 @@ docs.prebuild() {
         ./utils/searx.sh doc   | cat > "${DOCS_BUILD}/includes/searx.rst"
         ./utils/filtron.sh doc | cat > "${DOCS_BUILD}/includes/filtron.rst"
         ./utils/morty.sh doc   | cat > "${DOCS_BUILD}/includes/morty.rst"
+        pyenv.cmd python -m searx.user_help search-syntax > docs/user/search-syntax.md
```
dalf: If we need to change the `manage` each time there is a new page, that's not scalable.
In `searx.user_help`, why not parse everything in the TOC?

searx/help/en/search-syntax.md
```diff
@@ -0,0 +1,35 @@
...
+Prefix `!` to set category and engine names.
+
+Prefix: `:` to set the language.
+
+Abbrevations of the engines and languages are also accepted.  Engine/category
+modifiers are chainable and inclusive.  E.g. with {{search('!map !ddg !wp paris')}}
+search in map category **and** duckduckgo **and** wikipedia for
+`paris`.
+
+See the {{link('preferences', 'preferences')}} for the list of engines,
```
dalf: It requires translators to deal Markdown and Jinja2: they have to know that the first `preferences` can be translated, not the second one.

`[preferences]({{ link('preferences') }})` is complicated

searx/user_help.py
```diff
@@ -24,7 +25,7 @@ class HelpPage(NamedTuple):
 
 
 # Whenever a new .md file is added to help/ it needs to be added here
-_TOC = ('about',)
+_TOC = ('search-syntax', 'about')
```
dalf: To put the about page as the first tab:
`_TOC = ('about', 'search-syntax')`

searx/webapp.py
```diff
@@ -896,19 +897,28 @@ def __get_translated_errors(unresponsive_engines: Iterable[UnresponsiveEngine]):
...
+
+
+_ONLINE_DOC = _doc.OnlineDoc(_doc.Mistletoe())
 
 
-@app.route('/help/en/<pagename>', methods=['GET'])
-def help_page(pagename):
-    """Render help page"""
-    page = user_help.PAGES.get(pagename)
+@app.route('/doc/<pagename>', methods=['GET'])
```
dalf: IMO it is important to keep the language in URL, otherwise different contents are going to be shown for the same URL: this is going to lead to confusion when the links are shared.

searx/doc.py
```diff
@@ -0,0 +1,290 @@
...
+    #
+    # Closing network won't help ..
+    #   from searx.network import network
+    #   network.done()
+
+    # waiting some seconds before ending the comand line was the only solution I
+    # found ..
+    import time
+
+    time.sleep(3)
```
dalf: If I understand correctly, the typical use case is:
* we load the app to resolve the URL like `/search`.
* we render the page by discarding the URL (we don't need the app).
* once rendered, we wait for the app to finish the background initialization

So do we load the application?

searx/doc.py
```diff
@@ -0,0 +1,290 @@
...
+        self.markup = markup or Markup()
+        self.CACHE: typing.Dict[tuple, typing.Union[HTMLPage, str]] = {}
+
+        # future: could be set from settings.xml
+        self.folder: str = _DOC_FOLDER
+        self.i18n_origin: str = 'en'
+        self.l10n: typing.List = [
+            'en',
+        ]
+        self.toc: typing.List = ['search-syntax', 'about']
```
dalf: Related to the change in `./manage` : the TOC (*) is declared twice in two different locations.

(*) once including `about`, once without.

dalf (commented):
See the comments.
From them, I've made the commit https://github.com/searxng/searxng/commit/e5bc94f07afff465580acc9c15fb505585369881 in this branch: https://github.com/dalf/searxng/tree/user-doc-3

searx/webapp.py
```diff
@@ -896,19 +897,28 @@ def __get_translated_errors(unresponsive_engines: Iterable[UnresponsiveEngine]):
...
+    page = _ONLINE_DOC.render(pagename, locale)
+    if page is None:
+        page = _ONLINE_DOC.render(pagename)
     if page is None:
         flask.abort(404)
 
     return render(
-        'help.html', page=user_help.PAGES[pagename], all_pages=user_help.PAGES.items(), page_filename=pagename
+        'doc.html',
+        all_pages=_ONLINE_DOC.all_pages,
```
dalf: The call to `all_pages` is going to render all the pages for the selected language.

Why not make the code simpler:
* pre-render all the pages when the application starts.
* server the rendered page from this endpoint

docs/user/.gitignore
```diff
@@ -0,0 +1 @@
+search-syntax.md
```
dalf: Would `*.md` be better?
So when there is a new page, there is no need to update this file

searx/doc.py
```diff
@@ -0,0 +1,290 @@
...
+    def get_ctx(cls):
+
+        ctx = Markup.get_ctx()
+        ctx['link'] = lambda *args, **kwargs: '`%s`' % args[0]
+        ctx['search'] = lambda *args, **kwargs: '`%s`' % args[0]
+
+        return ctx
+
+
+_DOC_FOLDER = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'doc')
```
dalf: Is there a problem to use the `pkg_resources` instead of the `os` package and `open`?

searx/doc.py
return42: > So do we load the application?

You asking "why do we load the app"? ..  because the `url_for` command needs the  flask context. See the comment above https://github.com/searxng/searxng/pull/930/files#diff-b439f9f6bc445fcbb1be1c7d0314a845126d56ba235e52361c9918d943d5733aR84-R90


```
    # To render the online documentation a Flask application context is needed.
    # The url_for functions in the jinja templates need all routes to be
    # registered in the Flask app.
```

searx/user_help.py
return42: IMO users will more often look for the "search syntax" then for "about", that was my reason to place it this order, but I don't really care.  We can use the order you prefer.

searx/doc.py
return42: > (*) once including `about`, once without.

My intention is different to yours: IMO both documentations are not related to each other! There will be only some files in the online documentation we want to share in the project documentation.

IMO 'search-syntax' and 'about' is what we have now but I think in the future we will rework the about page and we will have more pages and some of them will not be a part of the project documentation.

This PR is just the starting point to share files from the online doc in the project doc.

searx/doc.py
return42: My intention is different.  IMO (not in this PR but may be in the future) the location is taken from settings.yml / see https://github.com/searxng/searxng/pull/930/files#diff-b439f9f6bc445fcbb1be1c7d0314a845126d56ba235e52361c9918d943d5733aR252-R258

```
        # future: could be set from settings.xml
        self.folder: str = _DOC_FOLDER
        self.i18n_origin: str = 'en'
        self.l10n: typing.List = [
            'en',
        ]
        self.toc: typing.List = ['search-syntax', 'about']
```

By this instance maintainer can write their own online documentation.  By example in my instance I use *hostname replace* to link all youtube results to invidious instances .. and I want to document this setup to my users.

searx/webapp.py
return42: > The call to `all_pages` is going to render all the pages for the selected language.

Yes, we need the titles (in the tabs) of all pages in the selected languages.

> pre-render all the pages when the application starts.

IMO rendering Markdown to HTML is an *expensive* task .. the parser needs to be started and the output (of all pages in all languages) need to be rendered.  I think a lazy rendering on demand combined with caching is much *cheaper* since users will not visit very often the online documentation and IMO no user will read all online doc pages in all languages.

docs/user/.gitignore
return42: See my comment above https://github.com/searxng/searxng/pull/930#discussion_r820058125

manage
return42: > why not parse everything in the TOC

I have a different intention about sharing pages from online doc in project doc / see https://github.com/searxng/searxng/pull/930#discussion_r820058125

searx/help/en/search-syntax.md
return42: That is true / I did not had an idea how to make it smarter for translators.

For search links we have:

    {{search('!wp paris')}}

Instead of:

    {{link('preferences', 'preferences')}}

we can use:

    {{link('preferences --> preferences')}}

but I do not know if it is better / any other suggestion for a markup of links that needed to be rendered in a Flask context?

return42:
> From them, I've made the commit https://github.com/searxng/searxng/commit/e5bc94f07afff465580acc9c15fb505585369881 in this branch: https://github.com/dalf/searxng/tree/user-doc-3

I tried ...

    ./manage pyenv.cmd python -m searx.doc docs/user

but it does not generate any file .. may I do something wrong / did you tested the command line?

searx/webapp.py
return42: Can we improve the routes in another PR? .. I want to touch to much in this PR and this PR already touches a lot.

dalf:
> may I do something wrong / did you tested the command line?

You did nothing wrong, I've tested then I've changed the code and did not tested again.

Add `userdoc.render()` just before `for page_name, _, page_content in userdoc.iter('en'):`

searx/doc.py
dalf: >  I use hostname replace to link all youtube results to invidious instances

This need exists for everyone using this plugin.
IMO, the plugin should provide a documentation or at least some information in the `/preferences` page (there is no way to do the later currently)

searx/doc.py
dalf: Can we avoid multiple declaration of the same information?

Currently, there are lists in:
* `searx/doc.py`
* `manage`
* `docs/user/.gitignore`

The last two are the same, the first one is different according to what you describe.

My idea is to include all the pages and includes two sections in `index.rst`: 
   * one with generic content (syntax.md, ...)
   * one with the documentation dedicated for the instances (about.md, etc...). Alternative: we just ignore these files (there are compiled but not included in the TOC).
--> Most importantly: in one .rst file, we can decide what belongs to the Sphinx documentation.

---

This is related to https://github.com/searxng/searxng/issues/808

searx/doc.py
dalf: Call sequence:
* `pyenv.cmd python -m searx.doc search-syntax Offline`
* Pick the [Offline](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L64)
* [`main()`: `doc = OnlineDoc(Offline())`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L75)
* [`main()`:  Initialize `searx.webapp`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L93-L96)
* [`main()`: `page = doc.render(pagename)`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L102-L104)
* [`OnlineDoc.render(...)`: `page = self.markup.render(content)`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L283)
* [`Markup.render(...)`: `ctx = self.get_ctx()`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L153-L157)
* [`Offline.get_ctx()`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L228-L235) : here we don't use `url_for`
* [`Markup.render(...)`: `return template.render(**ctx)`](https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L157) : so `url_for` won't be called, so there is no need for the app to be initialized.

searx/doc.py
return42: This plugin was just an example .. I think there is a need for having a online documentation to describe the intention of the setup of that instance.

By example: [... but all results (incl. yahoo) are redirected to an instance in DE where the majority of my users is located](https://github.com/searxng/searxng/discussions/911)

searx/doc.py
return42: > so `url_for` won't be called, so there is no need for the app to be initialized.

Not for the `url_for` (`link` & `search`) function but may be in the future we want to use other information or functions from the Flask context? .. if we will never use any information from the Flask context in the project documentation (aka Offline) .. we can simplify the code a lot (like you have done in your branch).

What do you think, should I simplify the code of this PR?

dalf: > should I simplify the code

Yes!

Currently, I feel `searx/doc.py` is a tool that should be in `searxng_extra`.
IMO, for the runtime, there are too many abstractions.

searx/doc.py
return42: > Most importantly: in one .rst file, we can decide what belongs to the Sphinx documentation.

If you place .md files in the /doc but don't include them in the TOC, the Sphinx-doc parser will warn you that there are files not included in any TOC .. may be this warnings can be suppressed by a conf.py setting .. but then we did have won anything about your initial question:

> Can we avoid multiple declaration of the same information?

searx/doc.py
return42: Hm .. just I have another thought (sorry) .. I think the Flask context is also useful when the project documentation is build.  Assume the `base_url` is set:

https://github.com/searxng/searxng/blob/927aa71133839ce5713b742b43285953301adf36/searx/settings.yml#L45

then we can also compile links in the the project documentation / that is linked by

https://github.com/searxng/searxng/blob/927aa71133839ce5713b742b43285953301adf36/searx/settings.yml#L9

Since we do not have a base_url in our project, we do not need this, but forks who build their own online documentation will find it useful I think.

With said in mind, what do you think about?

----

EDIT:

>  Currently, I feel `searx/doc.py` is a tool that should be in `searxng_extra`.
> IMO, for the runtime, there are too many abstractions.

In a rework of this PR I can simplify the code and see what we can move to `searxng_extra` .. but first I want to clarify with you if it make sense to have a flask context in the build process of the project doc.

searx/doc.py
dalf: > By example: https://github.com/searxng/searxng/discussions/911

Where this documentation should be?
Currently it requires to change `searx/doc.py` to include the additional pages in the tab.

Idea:
* keep the git repository clean
* allow to add additional documentation repository: all pages declared here are included in the tab.

So the administrator just have to declare the directory, add some files nothing more.

(perhaps it could be better, simpler)

searx/webapp.py
dalf: > IMO rendering Markdown to HTML is an expensive task ..

To have an idea, https://github.com/jlevy/the-art-of-command-line/blob/master/README.md (40KB) takes few seconds on my laptop (slow mode, power cable unplugged). Multiple by the number of the language this can be a problem if we have the same amount of document (once again, we don't know what will be the user documentation which makes the decision taking not easy).

The rendering time issue exists also in this PR, because any request to the documentation is going to render all pages  for this language.

Call sequence:
* https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/webapp.py#L907
* https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L266
* https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/webapp.py#L917-L922
* https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/templates/simple/doc.html#L5
* https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L260
* for each pages of a language: 
  * https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L263
  * https://github.com/searxng/searxng/blob/5d45cd6eb18095e0bb3e5715c58072e68c4aa727/searx/doc.py#L266 

Idea:
* load the raw content of all the pages
* parse the raw content to get the title and the markdown content
* then do lazy rendering of the markdown content

So we have real lazy rendering.

searx/doc.py
return42: > So the administrator just have to declare the directory, add some files nothing more.

Sounds good to me .. 

In this PR I want to prepare a boilerplate that is able to do so, but I would like to implement the functionality (where the values are taken from settings.yml) in another PR .. is it OK for you?

searx/webapp.py
return42: > So we have real lazy rendering.

Sound good to me, I will give it a try in a rewrite of this PR.

searx/doc.py
dalf: IMO, let's do that in another PR and keep the focus on the rendering of the `searx/doc` content.

searx/doc.py
dalf: I'm not sure to understand.

Do you suggest that some forks may wish to create links from the Sphinx documentation to their own instance?
So in this case `url_for` would be useful for them.

[EDIT] I can't say, but I have NOT seen ANY requests in that way and so let's skip this for now.

return42: > Do you suggest that some forks may wish to create links from the Sphinx documentation to their own instance?

yes.

dalf:
Some naming questions:
* why rename `help` into `doc`? two doc(s) directories can lead to some confusion.
* why `./docs` with a "s" and `./searx/doc` without "s"?

return42:
> * why rename `help` into `doc`? two doc(s) directories can lead to some confusion.

`doc` because it is not a *help* .. its documentation including *help*, *about* .. and other documentation.  I don't know if it lead in confusion, because the one documentation is online (searx/doc) and the other is a project documentation (/docs).

> * why `./docs` with a "s" and `./searx/doc` without "s"?

I don't have a clear opinion on what is better .. /docs has been exists for the project documentation .. when I rewrote the "help" I thought it is not a *help* (as described), so I asked myself what is a better URL route to a online documentation and I came to the route named /doc .. If you prefer a better name for online documentation, I am open for suggestions (except /help what is IMO wrong).

requirements-dev.txt
```diff
@@ -8,12 +8,14 @@ splinter==0.17.0
 selenium==4.1.2
 twine==3.8.0
 Pallets-Sphinx-Themes==2.0.2
+docutils>=0.17
 Sphinx==4.4.0
 sphinx-issues==3.0.1
-sphinx-jinja==1.4.0
-sphinx-tabs==3.2.0
+sphinx-jinja @ git+http://github.com/return42/sphinx-jinja.git
```
return42: Interim solution / see https://github.com/tardyp/sphinx-jinja/pull/30 and https://github.com/tardyp/sphinx-jinja/issues/28#issuecomment-1060701874

requirements-dev.txt
```diff
@@ -8,12 +8,14 @@ splinter==0.17.0
...
+sphinx-tabs @ git+https://github.com/executablebooks/sphinx-tabs
```
return42: Interim solution / see https://github.com/executablebooks/sphinx-tabs/pull/144#issuecomment-1060717377

return42:
@dalf I rebased on current master and added a WIP-commit c136a12c

Could you have a rough look over c136a12c / just to confirm that you are fine with the (class) model .. than I can squash the WIP-commit and polish _here and there_

Lets have an eye on a boilerplate that renders / other topics we discussed can be added or fixed in other PRs.

> Currently, I feel searx/doc.py is a tool that should be in searxng_extra.

I simplified the (class) model and pushed code to generate project docs to `searxng_extra/docs_prebuild`.

> So we have real lazy rendering.

done

> Most importantly: in one .rst file, we can decide what belongs to the Sphinx documentation.

We now have a `*.md` handling (that scales) .. all `searx/docs/en/*.md` files are compiled into `/docs/user/*.md` and in `/docs/user/index.rst` we decide what to import:

```reST
.. include:: search-syntax.md
   :parser: myst_parser.sphinx_
```

[Include Markdown files into an rST file](https://myst-parser.readthedocs.io/en/latest/sphinx/use.html#include-markdown-files-into-an-rst-file) requires docutils `0.17` what lead me in a dependency hell .. therefore I added two interim dependencies to the `requirements-dev.txt` / see my comments:

- https://github.com/searxng/searxng/pull/930/files#r820784816
- https://github.com/searxng/searxng/pull/930/files#r820786256

.. I assume that they can be solved soon.

----

To test online user-doc use `make run` and have a look at the online user documentation.

To test project documentation run `make docs.live` .. if you are fine with, patch your settings.yml by (e.g.)

```diff
@@ -42,7 +42,7 @@ server:
   # instance's enviroment (make buildenv)
   port: 8888
   bind_address: "127.0.0.1"
-  base_url: false  # Possible values: false or "https://example.org/location".
+  base_url: https://darmarit.org/searx  # Possible values: false or "https://example.org/location".
   limiter: false  # rate limit the number of request on the instance, block some bots
```

and run `make docs.live` .. and the links in the project documentation (search syntax) should now link to the instance from `base_url`.

searx/doc.py
```diff
@@ -0,0 +1,199 @@
...
+        self._content = None
+        self._title = None
+
+    @property
+    def raw_content(self):
+        """Raw content of the page (without any jinja rendering)"""
+        if self._raw_content is None:
+            with open(self.fname, 'r', encoding='utf-8') as f:
+                self._raw_content = f.read()
+        return self._raw_content
```
dalf: With [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) :
```python
from functools import cached_property

...
    @cached_property
    def raw_content(self):
        """Raw content of the page (without any jinja rendering)"""
        with open(self.fname, 'r', encoding='utf-8') as f:
            return f.read()
```

searx/doc.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # future: could be set from settings.xml
+
+        self.folder: str = _DOC_FOLDER
+        """location of the Markdwon files"""
+
+        self.i18n_origin: str = 'en'
+        """default language"""
+
+        self.l10n: typing.List = [
```
dalf: AFIU, `l10n` means `localisation`, here the values are languages.

searx/doc.py
```diff
@@ -0,0 +1,199 @@
...
+        self.page_class = page_class
+        self.base_url = base_url
+        self.CACHE: typing.Dict[tuple, DocPage] = {}
+
+        # future: could be set from settings.xml
+
+        self.folder: str = _DOC_FOLDER
+        """location of the Markdwon files"""
+
+        self.i18n_origin: str = 'en'
```
dalf: Why not `default_language` or `fallback_language`?

dalf:
> If you prefer a better name for online documentation, I am open for suggestions (except /help what is IMO wrong).

* `doc` without a `s` is confusing
* `docs` is still confusing
* `userdocs` ?
* `help`: search engines use this word:
  * https://help.duckduckgo.com/ : the front page is an about page.
  * https://support.google.com/websearch/?hl=en&rd=2#topic=3378866
  * https://search.brave.com/help (okay empty)

(related to https://github.com/searxng/searxng/issues/808 )

searxng_extra/docs_prebuild
```diff
@@ -0,0 +1,83 @@
...
+
+_doc_user = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'docs', 'user'))
+
+def main():
+
+    DOC = None
+    base_url = get_setting('server.base_url', None)
+
+    if base_url:
+        DOC = _render_all_with_flask_ctx(base_url)
```
dalf, line 21: Do you plan to use this?
In which context?
If on, let's drop the support for this feature (if base_url is not None, then the URL are rendered in the Sphinx documentation).

return42:
> (related to https://github.com/searxng/searxng/issues/808 )

Please decide what you prefer, my suggestion was `doc` if you prefer `help`, `about` or .. lets take it.

searxng_extra/docs_prebuild
```diff
@@ -0,0 +1,83 @@
...
+            f.write(page.content)
+
+def _render_all():
+    DOC = doc.OnlineDoc(OfflinePage)
+    for pagename, page in DOC.all_pages('en'):
+        page.render()
+    return DOC
+
+
+def _render_all_with_flask_ctx(base_url):
```
dalf: Untested simpler version:

```python
#!/usr/bin/env python
# lint: pylint

"""Script that implements some prebuild tasks needed by target docs.prebuild
"""

import sys
import os.path
import time
import typing

from searx import settings, get_setting, doc

_doc_user = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'docs', 'user'))


class OfflinePage(doc.DocPage):

    def get_ctx(self):  # pylint: disable=no-self-use
        """Jinja context to render :py:obj:`DocPage.content` for offline purpose (no
        links to SearXNG instance)"""

        ctx = super().get_ctx()
        ctx['link'] = lambda name, url: '`%s`' % name
        ctx['search'] = lambda query: '`%s`' % query

        return ctx


class OfflineDoc(doc.OnlineDoc):

    def __init__(self, base_url):
        super().__init__(doc.DocPage)
        self.base_url = base_url  # so the base_url parameter can be removed from OnlineDoc

    def all_pages(self, locale: typing.Optional[str] = None):
        settings['server']['secret_key'] = "x"
        from searx.webapp import app

        with app.test_request_context(base_url=self.base_url):
            yield from super().all_pages(locale)

        time.sleep(3)


def main():
    base_url = get_setting('server.base_url', None)
    DOC = OfflineDoc(base_url) if base_url else doc.OnlineDoc(OfflinePage)
    for pagename, page in DOC.all_pages('en'):
        fname = os.path.join(_doc_user, os.path.basename(page.fname))
        with open(fname, 'w') as f:
            f.write(page.content)


if __name__ == '__main__':
    sys.exit(main())
```

searx/doc.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        if self._html is None:
+            self._html = mistletoe.markdown(self.content)
+        return self._html
+
+
+_DOC_FOLDER = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'doc')
+
+
+class OnlineDoc:  # pylint: disable=too-few-public-methods
```
dalf: With the line in mind ( from `searxng_extra/docs_prebuild` ):
```
DOC = doc.OnlineDoc(OfflinePage)  # It requires a careful reading to understand that there is no cat in a box that may be alive or dead.
```

`UserDoc`, `UserDocumentation` or a name describing the content would be better.

dalf:
>  Please decide what you prefer, my suggestion was doc if you prefer help, about or .. lets take it.

* `searx.doc.OnlineDoc` --> `searx.user_help.UserHelp`

Optional:
* `searx/user_help.py` --> `searx/user_help/__init__.py`
* `searx/user_help/en`, `searx/user_help/de`, etc...

searx/doc.py
return42: `localisation` refers to the adaptation of software and documentation to a specific country .. we use the term "locale" very often when we mean a language.  But if you want, I can rename it to `languages`.

searx/doc.py
return42: Because this is the origin document, the other languages are just translations.

searx/doc.py
dalf: in settings.yml and preferences.py :
* `locale` is the language UI + region most probably [because of babel](https://github.com/searxng/searxng/blob/e1319fc21c713cfb9675389ba38f37e10cf77cd2/searx/webapp.py#L257)
* we will have to `zh_Hans_CN` and `zh_Hant_TW` for example.
* `languages` is the query language (in preferences.py and webapp.py)

So I vote for `self.locale`

searx/doc.py
dalf: weblate uses ["source language"](https://docs.weblate.org/en/weblate-4.11/admin/projects.html#component-source-language).

Since we use locale everywhere (except for the query language), I suggest `self.locale_source`

return42:
@dalf: Sorry for being annoying in this question, what do you think about ..

* `searx.doc.OnlineDoc` --> `searx.info.InfoPage`

Optional:

* `searx/user_help.py` --> `searx/infopage/__init__.py`
* `searx/info/en`, `searx/info/de`, etc...

My first thoughts when I see a URL `/user_help` I would ask if there is another help e.g. a `/admin_help` 

If you insist on `user_help` it is OK for me .. just leave a no/yes about my `/info` suggestion.

I hope I have time this weekend to implement .. thanks for being patient with me.

searx/static/themes/simple/src/less/info.less
```diff
@@ -0,0 +1,13 @@
+.info-page {
+  font-family: sans-serif;
+  font-size: 1.3em;
```
dalf, line 3: `1em` seems better IMO

searx/templates/simple/base.html
```diff
@@ -58,7 +58,7 @@
   </main>
   <footer>
     <p>
-    {{ _('Powered by') }} <a href="{{ url_for('help_page', pagename='about') }}">searxng</a> - {{ searx_version }} — {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
+    {{ _('Powered by') }} <a href="{{ url_for('info', pagename='about') }}">searxng</a> - {{ searx_version }} — {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
```
dalf, line 61: There is no `locale` here : `url_for_theme` is going to add the locale, so even in a 404 page, the link is going to work.

searx/webapp.py
```diff
@@ -382,6 +383,11 @@ def url_for_theme(endpoint: str, override_theme: Optional[str] = None, **values)
         if file_hash:
             values['filename'] = filename_with_theme
             suffix = "?" + file_hash
+    if endpoint == 'info' and 'locale' not in values:
```
dalf, line 386: see `simple/base.html`  and `oscar/nav.html`.

allow to have a working link to the about page even in a 404 page.

searx/webapp.py
```diff
@@ -898,19 +906,36 @@ def __get_translated_errors(unresponsive_engines: Iterable[UnresponsiveEngine]):
...
+
+@app.route('/info/<locale>/<pagename>', methods=['GET'])
+def info(pagename, locale):
+    """Render page of online user documentation"""
+
+    page = _INFO_PAGES.get_page(pagename, locale)
     if page is None:
         flask.abort(404)
 
+    def all_pages():
```
dalf, line 924: Iterator over all the page of the TOC including:
* pagename
* page
* locale

If the locale is `de` and `de/about.md` exists (and no other translated page), this function is going to yield:
* `search_syntax`, `PageInfo(....)`, `en`
* `about`, `PageInfo(....)`, `de`

searxng_extra/docs_prebuild
```diff
@@ -0,0 +1,84 @@
...
+    if base_url:
+        infopageset_ctx = _instance_infosetset_ctx(base_url)
+    else:
+        infopageset_ctx = _offline_infosetset_ctx()
+
+    with infopageset_ctx as infopageset:
+        for _, page in infopageset.all_pages('en'):
+            fname = os.path.join(_doc_user, os.path.basename(page.fname))
+            with open(fname, 'w') as f:
+                f.write(page.content)
```
dalf, line 30: no need to call `render()`, we just need the markdown content.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        page = self.CACHE.get(cache_key)
+
+        if page is not None:
+            return page
+
+        # not yet instantiated
+
+        fname = os.path.join(self.folder, locale, pagename) + '.md'
+        if not os.path.exists(fname):
+            logger.info('file %s does not exists', fname)
```
dalf, line 166: A translator doesn't have to translate all the pages: this is not an error if a page doesn't exist. The `info` endpoint in `webapp.py` deals with missing pages.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+            self.CACHE[cache_key] = None
```
dalf, line 167: Cache the fact the page was not found

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+    def title(self):
+        """Title of the content (without any markup)"""
+        t = ""
+        for l in self.raw_content.split('\n'):
+            if l.startswith('# '):
+                t = l.strip('# ')
+        return t
+
+    @cached_property
+    def html(self):
```
dalf, line 73: Remove the render function and keep only a `html` property.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+
+        ctx = {}
+        ctx['GIT_URL'] = GIT_URL
+        ctx['get_setting'] = get_setting
+        ctx['link'] = _md_link
+        ctx['search'] = _md_search
+
+        return ctx
+
+    def __repr__(self):
```
dalf, line 101: Might be useful for debugging.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        .. _mistletoe: https://github.com/miyuchina/mistletoe
+
+        """
+        return mistletoe.markdown(self.content)
+
+    def get_ctx(self):  # pylint: disable=no-self-use
+        """Jinja context to render :py:obj:`InfoPage.content`"""
+
+        def _md_link(name, url):
+            url = url_for(url, _external=True)
```
dalf, line 86: Use `_external`, so there is no need to add a `base_url` as parameter of the constructor of the class.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # future: could be set from settings.xml
+
+        self.folder: str = info_folder or _INFO_FOLDER
+        """location of the Markdwon files"""
+
+        self.locale_default: str = 'en'
+        """default language"""
+
+        self.locales: typing.List = [locale for locale in os.listdir(_INFO_FOLDER) if locale in LOCALE_NAMES]
```
dalf, line 129: Read the existing directories and make sure there are actual locales: a new translation doesn't require to change the code.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __init__(
+        self, page_class: typing.Optional[typing.Type[InfoPage]] = None, info_folder: typing.Optional[str] = None
+    ):
+        self.page_class = page_class or InfoPage
+        self.CACHE: typing.Dict[tuple, typing.Optional[InfoPage]] = {}
+
+        # future: could be set from settings.xml
+
+        self.folder: str = info_folder or _INFO_FOLDER
```
dalf, line 123: Tests don't need to hook _INFO_FOLDER (tests... that are not written)

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+        if not os.path.exists(fname):
+            logger.info('file %s does not exists', fname)
+            self.CACHE[cache_key] = None
+            return None
+
+        page = self.page_class(fname)
+        self.CACHE[cache_key] = page
+        return page
+
+    def all_pages(self, locale: typing.Optional[str] = None):
```
dalf, line 174: Simple iterator here: the complexity is moved to `searx.webapp` because it requires various parameter from the current HTTP request. May be not the best solution.

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+
+        :param pagename: name of the page, a value from :py:obj:`InfoPageSet.toc`
+        :type pagename: str
+
+        :param locale: language of the page, e.g. ``en``, ``zh_Hans_CN``
+                       (default: :py:obj:`InfoPageSet.i18n_origin`)
+        :type locale: str
+
+        """
+        locale = locale or self.locale_default
```
dalf, line 149: Fallback to the default value (before the fallback was set after `if locale not in self.locales` which is always `False` if `locale is None`

searx/infopage/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+from ..compat import cached_property
+from ..version import GIT_URL
+from ..locales import LOCALE_NAMES
+
+
+logger = logging.getLogger('searx.infopage')
+_INFO_FOLDER = os.path.abspath(os.path.dirname(__file__))
+
+
+class InfoPage:
```
dalf, line 44: Keep only one class for the pages: `render()` was never called in `InfoPage`.

docs/user/index.rst
```diff
@@ -2,9 +2,14 @@
 User documentation
```
dalf, line 2: We should not name the same thing with different names.
`User documentation` must be replaced by `Information` or `User Information` if `Information` is the name in the webapp.

dalf (commented):
I've added a commit and some comments to explain the changes.

return42:
> I've added a commit and some comments to explain the changes.

Thanks!

Issue #140, opened 2021-06-10
Manage HTTP headers


dalf:
Currently, the default HTTP headers are set in [settings.yml](https://github.com/searxng/searxng/blob/73dbee45a649bdb6363717be0670bf4f35bf06a9/searx/settings.yml#L47-L52)

Another version in [searx-docker](https://github.com/searx/searx-docker/blob/abd5079b87e8c50e87c23e7b72cfcfcaa087fa47/Caddyfile#L34-L59) includes more headers. 

* See https://owasp.org/www-project-secure-headers/ 
   * "Best Practices" tab
   * [X-XSS-Protection](https://owasp.org/www-project-secure-headers/#x-xss-protection): _it is recommended to set the header as X-XSS-Protection: 0_

So currently:
* Content-Security-Policy is set in searx-docker but otherwise except if the administrator look at different locations.
* Administrators who do not update their settings.yml, won't have a secure setup.
* Some other administrator will set theirs HTTP headers in the reverse proxy once for all.

Looking at owasp.org, the HTTP headers need regular maintenance.

----

This issue suggests to define the default headers directly a place that is keep up to date: the code, so the configuration is not spread into different locations.

An option can let the administrator override them if requires.

About Content-Security-Policy: 
* a default can be set at first, see [Whoogle](https://github.com/benbusby/whoogle-search/blob/904091f44032961c6d494eb667862c49b5a695f7/app/routes.py#L100-L102)
* in a second time, the engines can returns a list of external website to include in the CSP header.

return42:
Just want to leave some notes and links here ...

- [Make HTML templates more CSP compliant](https://github.com/searxng/searxng/issues/57)
- engines might have their own CSP requirements, e.g. `img-src 'self' data:` from  google.-images (and other): [`Header set Content-Security-Policy "img-src 'self' data: ;"`](https://github.com/searx/searx/blob/5f7056cf37a776e7aae151a4d1f5560f1bd0472f/searx/engines/google_images.py#L7-L12)

dalf:
> engines might have their own CSP requirements

* simple solution: set the right value in settings.yml
* more elaborate: the result of an engine can add value to the CSP header (in addition to normal result, infoboxes, engine-data).

return42:
> more elaborate: the result of an engine can add value to the CSP header (in addition to normal result, infoboxes, engine-data).

[...]

PR #932, opened 2022-02-28
static files: add a hash as a query

dalf:
## What does this PR do?

When the app starts, the hash of each static file (*) is read (sha1).

Then the URL of the static files contains this hash as query.

(*) hidden files, the directories `src` and `node_modules` are ignored.

---

Also, in a second commit, `uwsgi.ini` in the docker image is updated to ask the browser to cache the static files for one year.

* Avoid disappointing UX (most important)
* (???) reduce bandwidth

## Author's checklist

The best solution is to change the file name at build time, but the grunt plugins are rather old (4 years old for the most recent one).

## Related issues

This PR is an updated version of https://github.com/searx/searx/pull/2242

- see [label:"clear browser cache"](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -44,9 +44,9 @@ logto = /var/log/uwsgi/uwsgi.log
 add-header = Connection: close
 
 # uwsgi serves the static files
-# expires set to one day as Flask does
+# expires set to one year since there are hashes
 static-map = /static=/usr/local/searxng/searx/static
-static-expires = /* 864000
+static-expires = /* 31557600
 static-gzip-all = True
 offload-threads = %k
 
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -16,7 +16,7 @@ from timeit import default_timer
 from html import escape
 from io import StringIO
 import typing
-from typing import List, Dict, Iterable
+from typing import List, Dict, Iterable, Optional
 
 import urllib
 import urllib.parse
@@ -348,7 +348,7 @@ def code_highlighter(codelines, language=None):
     return html_code
 
 
-def get_current_theme_name(override: str = None) -> str:
+def get_current_theme_name(override: Optional[str] = None) -> str:
     """Returns theme name.
 
     Checks in this order:
@@ -373,14 +373,16 @@ def get_result_template(theme_name: str, template_name: str):
     return 'result_templates/' + template_name
 
 
-def url_for_theme(endpoint: str, override_theme: str = None, **values):
+def url_for_theme(endpoint: str, override_theme: Optional[str] = None, **values):
+    suffix = ""
     if endpoint == 'static' and values.get('filename'):
         theme_name = get_current_theme_name(override=override_theme)
         filename_with_theme = "themes/{}/{}".format(theme_name, values['filename'])
-        if filename_with_theme in static_files:
+        file_hash = static_files.get(filename_with_theme)
+        if file_hash:
             values['filename'] = filename_with_theme
-    url = url_for(endpoint, **values)
-    return url
+            suffix = "?" + file_hash
+    return url_for(endpoint, **values) + suffix
 
 
 def proxify(url: str):
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -1,12 +1,13 @@
 # -*- coding: utf-8 -*-
 import os
+import pathlib
 import csv
 import hashlib
 import hmac
 import re
 import inspect
 import itertools
-from typing import Iterable, List, Tuple
+from typing import Iterable, List, Tuple, Dict
 
 from io import StringIO
 from codecs import getincrementalencoder
@@ -58,13 +59,29 @@ def get_themes(templates_path):
     return themes
 
 
-def get_static_files(static_path):
-    static_files = set()
-    static_path_length = len(static_path) + 1
-    for directory, _, files in os.walk(static_path):
-        for filename in files:
-            f = os.path.join(directory[static_path_length:], filename)
-            static_files.add(f)
+def get_hash_for_file(file: pathlib.Path) -> str:
+    m = hashlib.sha1()
+    with file.open('rb') as f:
+        m.update(f.read())
+    return m.hexdigest()
+
+
+def get_static_files(static_path: str) -> Dict[str, str]:
+    static_files: Dict[str, str] = {}
+    static_path_path = pathlib.Path(static_path)
+
+    def walk(path: pathlib.Path):
+        for file in path.iterdir():
+            if file.name.startswith('.'):
+                # ignore hidden file
+                continue
+            if file.is_file():
+                static_files[str(file.relative_to(static_path_path))] = get_hash_for_file(file)
+            if file.is_dir() and file.name not in ('node_modules', 'src'):
+                # ignore "src" and "node_modules" directories
+                walk(file)
+
+    walk(static_path_path)
     return static_files
 
 
--- a/tests/unit/test_webapp.py
+++ b/tests/unit/test_webapp.py
@@ -23,6 +23,11 @@ class ViewsTestCase(SearxTestCase):
         webapp.app.config['TESTING'] = True  # to get better error messages
         self.app = webapp.app.test_client()
 
+        # remove sha for the static file
+        # so the tests don't have to care about the changing URLs
+        for k in webapp.static_files:
+            webapp.static_files[k] = None
+
         # set some defaults
         test_results = [
             {


mrpaulblack:
The caching header is not enough. I think it must also contain the `immutable` directive (https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#freshness) so the browser does not do 304 requests for the resource to check if it changed (even tho it expires only after a year)...

Everything else LGTM from testing :+1:

dalf:
You mean
```
Cache-Control: public, max-age=31557600, immutable
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

So in `uwsgi.ini`, it should be something like this:
```
route = ^/static/ addheader:Cache-Control: public, max-age=31557600, immutable
```

according to https://uwsgi-docs.readthedocs.io/en/latest/InternalRouting.html#the-first-example

mrpaulblack:
yes exactly :+1:  (I know from testing in ff that the browser makes 304 requests for a resource if the immutable is missing)

dalf:
> I know from testing in ff that the browser makes 304 requests for a resource if the immutable is missing

I confirm it works with FF (using the docker image).

mrpaulblack:
I also tried the docker image and I get a 304 on firefox 97 on linux (I unchecked the disable cache option in dev tools); I can also see the immutable cache-control header:
image


idk why tho... are you sure its not making any request to the static resources in your setup (after the first one)?

dalf:
Sorry my comment lack of precision:  I got 304 one the second request and then everything from the cache

And I can see the header `Cache-Control: public, max-age=31557600, immutable`

---
I haven't checked if the header is absent from the other paths ( `/`, `/search`, etc...)

mrpaulblack:
yeah the header is only on /static requests...

I would like a setup where we dont even do the 304 requests for static files tho (that is already the case in current setup with expires header...)

I will check whats possible; maybe a limitation with ff..?

[EDIT] wait so after 2nd request your browser does not do the 304 anymore?

dalf:
> [EDIT] wait so after 2nd request your browser does not do the 304 anymore?

If I refresh the `/` page, I continue to get `304` responses.
If I do send a query, `/search` get the static files from the browser cache (without HTTP request).

It seems immutable is only effective after the URL has changed (this is a guess...)

[EDIT] F5 / refresh sends HTTP request, if I navigate using links or submit button, the browser uses the cache.

dalf:
Related bug : https://bugzilla.mozilla.org/show_bug.cgi?id=1267474

dalf:
Chrome doesn't have the issue: F5 or not, the `/static` URLs come from the cache after the first request.

mrpaulblack:
Yeah I just tested with chromium and it worked fine...

Ok so you were right; when changing pages in ff it uses the files from cache (the http200 color is a slightly more light green), when doing a refresh it requests the resource with a 304 again; which is IMO fine...

I just forgot sry: can you also please add caching headers to all other requests? So for the /image_proxy like a public 1 day cache and everything else no cache (if possible)...

Also if someone uses a prefix like [domain]/searx/ are the regex rules in uwsgi.ini for static and so on still going to work?

dalf:
Sum up of the discussion on Matrix:
* `/image_proxy` should not be cached for privacy reason  --> the header for this URL is topic of another PR.
* the current regex in uwsgi.ini doesn't work for URL with a prefix (for example `https://..../searx`) --> it has to be fixed in the PR.

dalf:
Caddyfile content:
```caddy
http://localhost:9000 {
    log {
            output stdout
    }

    handle_path /searx/* {
        reverse_proxy localhost:8080
    }
}
```

```sh
docker run --rm -d -e BASE_URL="http://localhost:9000/searx" -p 8080:8080 searxng/searxng
caddy run
```

I can see the `immutable` value in the `Cache-Control` header in this use case.

---

Another way to look at it: this PR solves two things:
1. cache busting so the browser don't use an old version version of the static file
1. make sure the static files are cached for a very long time.

We can keep the focus on the first item, and deal with the second item in another PR (actually related to https://github.com/searxng/searxng/issues/140 ). In this case, I remove the `Cache-Control: public, max-age=31557600, immutable` from the uwsgi.ini (and let the reverse proxy manages the HTTP header as it is case now).

dalf:
ping @mrpaulblack

mrpaulblack:
Well I am ok with it either way: having the immutable cache-control header in this PR in the `uwsgi.ini` or to manage all headers once and for all once a decision is made in #140. Either way I really like this cache busting and would like to test it on my instance... So I leave the decision about cache-control header in this PR to you :+1:

dalf:
Ok, I keep the focus on the cache busting in this PR, then in another PR we can deal the HTTP headers.

return42:
@mrpaulblack @dalf in our documentation we recommend to wrap the static files directly from HTTP server.  Below you find a screenshot of such a description (this is just one example, we recommend to wrap `/static` at multiple places).

If I'm not wrong, this will no longer be able since this PR is merged .. right? .. Should I fix the documentation? 

EDIT: now I see there is `?` in the URL `searxng.min.css?1e8` .. its not clear to me how URL arguments are handled by static `locations`s 

.. and I ask myself what is better, serve static file by HTTP server or by uWSGI

```
# uwsgi serves the static files
# expires set to one year since there are hashes
static-map = /static=${SEARX_SRC}/searx/static
static-expires = /* 31557600
...
```

I'm a little bit unclear about but I would say HTTP server is the right place to wrap static files .. ?



image

dalf:
> I ask myself what is better, serve static file by HTTP server or by uWSGI

The HTTP server is a better choice: we can keep this configuration.

uWSGI is convenient way to serve the static file from the docker image. 

> its not clear to me how URL arguments are handled by static locationss

True, from my memory there is no issue, but I haven't tried when I submitted the PR.
It reminds me that the best solution is not `searxng.min.css?5fde34a74bc438c7b56ec8c6501e131cc9914bd8` but `searxng-5fde34a74bc438c7b56ec8c6501e131cc9914bd8.min.css`.

return42:
> The HTTP server is a better choice: we can keep this configuration.

OK ...

> It reminds me that the best solution is not `searxng.min.css?5fde34a74bc438c7b56ec8c6501e131cc9914bd8` but `searxng-5fde34a74bc438c7b56ec8c6501e131cc9914bd8.min.css`.

Hm, this is not a name of a file in the static folder that can be passed through from HTTP server.

dalf:
> Hm, this is not a name of a file in the static folder that can be passed through from HTTP server.

The idea is:
* `make static.build.commit` creates:
   * `css/searxng-5fde34a74bc438c7b56ec8c6501e131cc9914bd8.min.css`
   * `mapping.json` with this content
```json
{
  "css/searxng.min.css": "css/searxng-5fde34a74bc438c7b56ec8c6501e131cc9914bd8.min.css"
}
```

so `webapp.py` can load `mapping.json` and then map the URL in the `url_for_theme` function.

There are some grunt plugins to do that, but there are all rather old.

return42:
`make static.build.commit` and https://github.com/searxng/searxng/issues/964 sounds a bit overloaded to me and complicates setup on different http servers .. but I don't have a better idea at hand .. except:

An alternative and much more simple to me is a expire time of 1 day or 10 hours without having hash-extensions in the URL.  This will not reduce the requests of static files by 100% but may be by 90% .. is simple to realize and is transparent to http servers of any kind.

Or I am wrong with my assumptions?

mrpaulblack:
Nah; I am also a big fan of cache busting, since all other methods will either result in broken sites for users for at least some time or make the site a lot slower on end user devices...

So we IMO keep current impl. and hope that HTTP servers ignore the param for static files, which should work or we do the impl. with changing the actual file name for the static files (which I agree is a cleaner solution then the param...)

return42:
> will either result in broken sites ... or make the site a lot slower on end user devices...

OK, I understand your arguments.  But then we should disable hash values in URL as long we do not have them in the file names (since it breaks existing HTTP and uWSGI setups).

OT: IMO `!image` search by image-proxy could be optimized, the potential to save bandwidth is > 10^5 ... since it is a proxy we can optimize bandwidth in client & server direction.

mrpaulblack:
> OK, I understand your arguments. But then we should disable hash values in URL as long we do not have them in the file names (since it breaks existing HTTP and uWSGI setups).

Oh man the URL param works really well in the docker setup and also for my instance tbh... But I understand that it breaks non docker setups so I am ok with reverting the PR for now, until we have the grunt workflow -_- -> feel free to do a PR

> OT: IMO !image search by image-proxy could be optimized, the potential to save bandwidth is > 10^5 ... since it is a proxy we can optimize bandwidth in client & server direction.

I was also thinking about caching the image proxy on client or server side, but either way its bad for privacy...

return42:
> I am ok with reverting the PR for now, until we have the grunt workflow 

May we make it optional for the first.

> I was also thinking about caching the image proxy on client or server side, but either way its bad for privacy...

May I have not a clear view on .. what are your concerns about privacy?

mrpaulblack:
Well the cached image can lead to the query the user was searching for, either on server side or in the browser on client side... (the cached static resources are on the other hand generic enough to not leak any privacy invading information)...

> May we make it optional for the first.

Like a feature flag in settings.yml? I think that is a great idea, we can leave it false by default..

return42:
> Like a feature flag in settings.yml? I

Yes .. If I have time for I will send a PR .. but first I want to test various HTTP & uWSGI setups.

> Well the cached image can lead to the query the user was searching for,

Really .. by example this is one URL from a image_proxy request https://darmarit.org/searx/image_proxy?url=https://tile.loc.gov/storage-services/service/pnp/pga/14300/14320_150px.jpg&h=723ad5ed40820ed12d7d85ee1b2806086c40cbac337e7c2226f6e1c740c9eb29

On server side the image_proxy can cache the contents of `https://tile.loc.gov/storage-services/service/pnp/pga/14300/14320_150px.jpg` .. 

When image_proxy stores BLOBs by URL names in a key/value storage the key is anonymous, I don't know how that could be related to a user .. or do I miss something?

mrpaulblack:
Well its not related to a user in that way directly. But there are still pictures cached in general. On a public instance I dont really see the problem, but on a smaller (for example private) instance, you could make guesses what users where searching for, by looking at the pictures I guess?

I mean storing images client side is IMO out of the questions since this way its easy to find out what someone was searching for, by looking at cached pictures.

The question is: Does it make sense to store the pictures server side, like you suggested? On the one hand you could make educated guesses, what in general was searched for when caching the images on server side, on the other hand it could make the image category a lot faster, since the image proxy could serve out images really quick...
I am not sure what is the best option tbh.

return42:
Thanks for explanation .. now I think we have a similar view on this topic ..

> I am not sure what is the best option tbh.

My 5 cent .. we can make it optional to cache images on server side .. to limit access on, the storage is only readable by the SearXNG service account .. like any other common known proxy caching.

mrpaulblack:
> My 5 cent .. we can make it optional to cache images on server side .. to limit access on, the storage is only readable by the SearXNG service account .. like any other common known proxy caching.

Agreed; We could even have like a new cache category in settings.yml for the static resource caching and the image proxy caching for example. Lets wait what @dalf has to say about it tho...

dalf:
> Really .. by example this is one URL from a image_proxy request

The image is the first result of [`!loc Le Louvre`](https://darmarit.org/searx/search?q=!loc%20le%20louvre): the image doesn't tell your query but your point of interest. If I push the reasoning about the POST / GET discussion, the image proxy should use  encrypted image URL ([Whoogle](https://github.com/benbusby/whoogle-search) does that)

> I don't know how that could be related to a user

It's related to the number of user. If there is one user and the cache contains picture of some holidays destination, and medical issue, and some other thing, it can be enough to find the individual (with time and determination, but the less user there are, the more cached data there is, the easier it is). If the instance deals with 10000 different users, it is far more complicated or impossible.

IMO and at first sight a cache module should be an external (a plugin or whatever).
Note: the current implementation is not fully stateless because of the Python nature. Some strings, URL might stay in RAM even after a call to the garbage collector after each request.

my-spot fork implements a cache of the HTTP response: https://gitlab.e.foundation/e/cloud/my-spot/-/blob/master/searx/network/cache.py . I remember that @MarcAbonce had some issue using it (? to confirm).

MarcAbonce:
I dropped that feature from my instance a while ago, but I think that the issue I had was caused by the image_proxy rather than the Redis cache.

Unfortunately I don't have the stack traces anymore, but the error I would get was about the maximum number of open connections that httpx could have (a 100) and requesting many images could easily surpass that limit.

PR #938, opened 2022-03-02
Update update_languages.py

MontyQI:
## What does this PR do?

- Changes update_languages.py to include flags
- Adds flags to languages.py
- Displays flags in language filter
flags

Makes searxng better looking.

`make run`
search anything
see flags in language filter


MontyQI:
Why do the checks tell that it would reformat update_languages.py? 
`would reformat searxng_extra/update/update_languages.py`

return42:
We have a code formatter: run `make format.python` before commit the python code.

Sorry, I think we forgot to update our quickstarter guide https://docs.searxng.org/dev/quickstart.html

searxng_extra/update/update_languages.py
```diff
@@ -113,13 +114,84 @@ def join_language_lists(engines_languages):
...
-                language_list[short_code]['countries'][lang_code] = {'country_name': country_name, 'counter': set()}
+                language_list[short_code]['countries'][lang_code] = {'country_name': country_name, 'flag': '', 'counter': set()}
 
             # count engine for both language_country combination and language alone
             language_list[short_code]['counter'].add(engine_name)
             if lang_code != short_code:
                 language_list[short_code]['countries'][lang_code]['counter'].add(engine_name)
-
+    flags = {
+        'ar': '🇸🇦',
```
dalf: Is the Saudi Arabia flag representative of all arabic language?

https://emojipedia.org/flag-saudi-arabia/

searxng_extra/update/update_languages.py
```diff
@@ -113,13 +114,88 @@ def join_language_lists(engines_languages):
...
+        'be': '🇧🇾',
+        'bg-BG': '🇧🇬',
+        'cs-CZ': '🇨🇿',
+        'da-DK': '🇩🇰',
+        'de': '🇩🇪',
+        'de-AT': '🇦🇹',
+        'de-CH': '🇨🇭',
+        'de-DE': '🇩🇪',
+        'el-GR': '🇬🇷',
+        'en': '🇬🇧',
```
dalf, line 137: This is especially true for `en`: it represents the English language whatever the region is (UK, US, NK, SG, etc...).

The UK flag can be confusing.

searxng_extra/update/update_languages.py
```diff
@@ -113,13 +114,88 @@ def join_language_lists(engines_languages):
...
+        'en-AU': '🇦🇺',
+        'en-CA': '🇨🇦',
+        'en-GB': '🇬🇧',
+        'en-IE': '🇮🇪',
+        'en-MY': '🇲🇾',
+        'en-NZ': '🇳🇿',
+        'en-US': '🇺🇸',
+        'es': '🇪🇸',
```
dalf, line 145: I guess the reasoning can be applied to `es` : it represents the Spanish language whatever the region is.

searx/templates/simple/filters/languages.html
```diff
@@ -1,8 +1,8 @@
 <select class="language" id="language" name="language" tabindex="1">{{- '' -}}
 	<option value="all" {% if current_language == 'all' %}selected="selected"{% endif %}>{{ _('Default language') }}</option>
-	{%- for lang_id,lang_name,country_name,english_name in language_codes | sort(attribute=1) -%}
+	{%- for lang_id,lang_name,country_name,english_name,flag in language_codes | sort(attribute=1) -%}
 	<option value="{{ lang_id }}" {% if lang_id == current_language %}selected="selected"{% endif %}>
-		{{- lang_name }} {% if country_name %}({{ country_name }}) {% endif %}- {{ lang_id -}}
+		{% if flag %}{{ flag }} {% endif%} {{- lang_name }} {% if country_name %}({{ country_name }}) {% endif %}
```
dalf: Nitpicking:

`-` after `{%` and before `%}` remove the new lines and the HTML smaller:

```python
{%- if flag %}{{ flag }} {% endif%} {{- lang_name }} {% if country_name %}({{ country_name }}) {% endif -%}
```

searxng_extra/update/update_languages.py
MontyQI: Yes because there isn't better flag. We can have this or none.

return42:
FYI: I don't know what is better but I experiment with [Regional indicator symbol](https://en.wikipedia.org/wiki/Regional_indicator_symbol)

See my "WIP" commit on my branch: https://github.com/return42/searxng/commits/language-filter

> This is especially true for `en`: it represents the English language whatever the region is (UK, US, NK, SG, etc...).
>
> The UK flag can be confusing.

It is more a question in general: can (search) language be associated by a flag of a region?

Some other examples where a language is not related to only one country (and its flag) are "German", "Spain", "Hebrew" ... "French" ..

May be we should only use flags where country and language are given (en-US, fr-BE ..)

MontyQI:
I think we should leave it as it is, after deleting flag for arabic language, because it isn't good. 
The flags are to help recognise languages and I think that most people will associate UK flag with English and France flag with French etc.

MontyQI:
> See my "WIP" commit on my branch: https://github.com/return42/searxng/commits/language-filter

 It is cleaner than my code, but some flags are wrong displayed (In Belarus --> Belgium flag, Slovenia flag --> Sierra Leone flag) or displays as unknown flag (🇸🇼)

return42:
@MontyQI sorry for the delay, it toked me a while before I got it done: in https://github.com/searxng/searxng/pull/967 I implemented an alternative solution for the flags. The implementation is a little bit simpler / the flags are added at the end, when the `write_languages_file(..)` is called. And the implementation is more generic .. 

If it is OK for you I would like to close this PR in favour of https://github.com/searxng/searxng/pull/967 and we can discuss any further there.

MontyQI:
It's ok

PR #946, opened 2022-03-06
README.rst: add a link to the screenshots

dalf:
## What does this PR do?

Add a link to the screenshots.
Feel free to suggest better wording.

Some images are better than just text.

N/A

## Author's checklist

May be we should keep https://github.com/searxng/searxng-browser-tests as a dev tool, and reference only few screenshots in README.md

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,6 +92,7 @@ User experience
   * usable on desktop, tablet and mobile
   * light and dark versions (you can choose in the preferences)
   * support right-to-left languages
+  * `see the screenshots <https://dev.searxng.org/screenshots.html>`_
 
 - the translations are up to date, you can contribute on `Weblate`_
 - the preferences page has been updated:


unixfox:
Some values set by default would be great I think.

Like firefox 1280 for browser and index for test case.

mrpaulblack:
I really like when projects have screenshots and with the help of the tool we actually have up to date screenshots; I think what we can do is link to the site itself (also we should maybe put this on a another host like `dev.searxng.org` instead of the path IMO...)

And maybe we can also insert a few screenshots as URL directly into the readme to always have up to date screenshots in the readme. This also helps people coming from SearX to immediately see the SearXNG frontend changes...

dalf:
> we should maybe put this on a another host like dev.searxng.org instead of the path IMO.

Ok. Tell me when dev.searxng.org exists.

---

See https://searxng.org/searxng-browser-tests/screenshots.html

By default it shows some results in general category using FF.
The URL is updated when the user selects another browser and/or test case (= the URL can be copied / pasted).

> And maybe we can also insert a few screenshots as URL directly into the readme

Some symbolic links can be helpful here, so we can have https://dev.searxng.org/images/firefox_1920_search_general.png for example.

mrpaulblack:
`dev.searng.org` has a CNAME now that points to searxng.github.io :+1:

dalf:
PR updated with the new links:
https://dev.searxng.org/
https://dev.searxng.org/screenshots.html

mrpaulblack (approved):
LGTM :+1: 

Thanks :D

return42:
I got a 404 on https://dev.searxng.org/screenshots.html

dalf:
> I got a 404 on https://dev.searxng.org/screenshots.html

For some reason, new run of the workflow revert the location to https://searxng.org/searxng-browser-tests/

There are persistent URL to the screenshots (perhaps the links need to be updated if the dev.searxng.org is not fixed):
```md
![image](https://dev.searxng.org/screenshots/firefox_1920/search_general/search_general.png)
![image](https://dev.searxng.org/screenshots/firefox_1920/search_images/search_images_detail.png)
![image](https://dev.searxng.org/screenshots/firefox_1920_-_dark_mode/search_videos/search_videos_player.png)
```

image
image
image

unixfox:
Still 404 for me too.

dalf:
> Still 404 for me too.

Fixed in commit https://github.com/searxng/searxng-browser-tests/commit/3e556c78c990f798526468990487ed5e1213644c (the GH workflow makes  sure there is a file named `CNAME`).

PR #963, opened 2022-03-13
Mobile bar google-like

MontyQI:
## What does this PR do?
 Changes search bar on mobile to google-like one.
bar

https://user-images.githubusercontent.com/79922836/158064568-707561ef-7628-48e5-b18d-ec8354572e77.mp4

Makes seaxng better on mobile and easier to use.

`make run`

## Related issues
https://github.com/searxng/searxng/issues/176

## Problems
~~1. Search bar gets focus automatically. It shouldn't especially on result page. See https://github.com/searxng/searxng/issues/176~~
2. If you want to clear query, it gets focus. It could be fixed with JS implementation.


mrpaulblack:
@MontyQI I am reviewing this PR and would like to rebase the PR on master branch as well as push the focus fix on top in this branch/PR... Is that ok with you?

MontyQI:
Yeah, ok.

mrpaulblack:
> Search bar gets focus automatically. It shouldn't especially on result page. See https://github.com/searxng/searxng/issues/176

I tried to remove the autofocus on the result page from the search input form (the `q` input field), but for whatever reason it still focuses the input form on page reload...

@MontyQI @dalf If you have an idea why feel free to change the WIP commit or to write a comment, I just cannot figure out why it would not work... (see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input)

(also the other params on the input are from whoogle so we have the updated params for the input in general in this PR too)

[EDIT] it is part of some JS; when disabeling JS it works on the result page and does not set focus to the search form.

[EDIT2] Nvm I figured it out in JS there was a line that would autofocus the search input field when loading the script -> which would happen on every page load...

return42 (approved):
Thanks!

mrpaulblack:
Not done yet tbh...

Still need to remove autofocus on mobile landing page (maybe?) and some cleanup...

return42:
> Still need to remove autofocus on mobile landing page (maybe?) and some cleanup...

Ah, OK you want to see the landing page first and when touching the search field the view should switch to the "google mobile" layout?

mrpaulblack:
yes, but at the same time I would like to keep the autofocus on desktop on the landing page...

return42:
> but at the same time I would like to keep the autofocus on desktop on the landing page...

OK, if you would not said it, it had been my next question :-) 

> [EDIT2] Nvm I figured it out in JS there was a line that would autofocus the search input field when loading the script -> which would happen on every page load...

But you removed it [in JS](https://github.com/searxng/searxng/pull/963/files#diff-533047a3107c9356db35bd9842440c37ba0e8fa48c1efb3239bce84aaf3a1915L97) and the auto-focus is still there .. right?

mrpaulblack:
So yeah we had set the autofocus for each page reload in general in JS. Then we have `search_simple.html` for the landing page and `search.html` for the result page. In `search_simple.html` is the autofocus attribute on the input field in `search.html` it is not.

So right now the landing page gets autofocus because of that HTML attribute and the result page not.

A fix could be to also remove the autofocus on search_simple and detect in JS if the device is a desktop and set the autofocus only on the landing page (otherwise its not possible to set and detect the autofucus only on desktop with pure HTML and CSS AFAIK)

return42:
@mrpaulblack @MontyQI  I pushed commit 8ca4fedca on top of your branch, I think this commit should fix the auto-focus issue on mobile devices / please test .. if not fine with, please drop my commit.

mrpaulblack:
Woow great thanks @return42 will clean up and prep the merge :tada: 

(I did not even think about the screen width :facepalm: )

dalf:
Two notes:
* A text browser is going to show the two input fields (otherwise the UI is nearly okay).
* In there one or two q parameters in the HTTP request?

return42:
> A text browser is going to show the two input fields (otherwise the UI is nearly okay).

Haven't this use case in mind .. what do you think, is it a KO criteria / do we official support text browsers (no CSS) in general?

> in there one or two q parameters in the HTTP request?

I haven't seen an issue .. this is what other says: https://stackoverflow.com/questions/2906793/is-it-valid-to-have-two-input-elements-with-the-same-name

mrpaulblack:
> I haven't seen an issue .. this is what other says: https://stackoverflow.com/questions/2906793/is-it-valid-to-have-two-input-elements-with-the-same-name

I just tried at and it seems to send to q params in ff with dev tools set to mobile screen...:
```http://127.0.0.1:8888/search?q=&q=time```

> Haven't this use case in mind .. what do you think, is it a KO criteria / do we official support text browsers (no CSS) in general?

In the past we kinda tried to not break them at least AFAIK (for example the clear and search button are text in the no JS version) -> for example with lynx or elinks SearXNG is compared to most other sites actually kinda usable (apart from being blocked by the limiter plugin on public instances...)

return42:
> I just tried at and it seems to send to q params in ff with dev tools set to mobile screen...: ```http://127.0.0.1:8888/search?q=&q=time```

Seems HTML standard to handle multiple text fields like multiple choice checkboxes. But does not look very good for `q` search terms I think.

> for example the clear and search button are text in the no JS version)

No-JS is clear to me but my question was more about: "can we use CSS tricks?"

mrpaulblack:
> No-JS is clear to me but my question was more about: "can we use CSS tricks?"

I think it depends. I mean in general those will break the site in text browsers, but on the other hand we can get around JS impl. sometimes..

I think in this case the problem is that the server reads the first q value and ignores the second one. On mobile the second q input has the value and therefor the search does not work (since the first one is empty).

So in our case I would say its better to have a JS solution for getting autofocus on desktop landing (or to just dont have autofocus at all, but idk all other search engines have autofocus on their landing page) and if JS is disabled just dont have autofocus at all..., what do you think?

MontyQI:
Do we really need changes from https://github.com/searxng/searxng/pull/963/commits/8ca4fedca238fb5a25d37a5f323cfcc76a032d5d ? As you wrote, it breaks searching, and I don't think that autofocus on pc is that important. Can't we just disable autofocus completely?

return42:
> I think in this case the problem is that the server reads the first q value and ignores the second one. 

This can be solved (in the jinja template or alternatively in the server code) .. but both solutions will still break the site in text browsers.

So I agree with you: it might be better to set autofocus in JS .. :( ..  I removed my commit from your branch.

> and I don't think that autofocus on pc is that important. 

I think it is .. at least for keybord centric users like me it is much more comfortable with autofocus ..

mrpaulblack:
@return42 wait let me at least recheck the PR for a second than you can merge :D

return42:
I rebased the branch and removed autofocus from [searx/templates/simple/simple_search.html](https://github.com/searxng/searxng/pull/963/commits/42305bc719c0be4f972390327f039544350d5683#diff-ab4ec909183e0ba44b26c47a14fedd4dd4a288878323afdd7c34184efe51b387)

But (sometimes) I see an issue with auto completion:
  
grafik

mrpaulblack:
Yes that is because autocomplete position is still set by the autopcomplete.js lib that is setting it at the original input position, also the focus within is on the search form, but rather should be on the search box wich contains the input... will push a fix commit on top... (Sorry for late review)

mrpaulblack:
I reworked and tested this PR for a few hours with JS, dark, light, mobile, tablet, desktop, RLT as well as LTR and the only things that are still broken AFAIK:

* Not a blocker: the settings button is part of the search box outline on mobile when the q input is empty (but since this is a CSS feature there is nothing really that can be done about this IMO; we could use the before impl. of putting a border on each element in the search box, but not really clean IMO) -> it is also broken on RTL, but this was already before this PR that way
* Definitely a blocker: pressing on the search logo on mobile sets the focus within on the search header div and therefor the link does not work and it instead opens the new google like search input... I dont know how to fix this :/ -> links will always set the focus

Everything else should work now this PR is a big improvement ui/ux wise, thanks @MontyQI

[EDIT] ok looks like I found a fix -> putting the inert html attribute on search-logo makes the element non focus able, while still being able to use keyboard tab to reach it... 

[EDIT2] looks like there is no browser support for inert tho sadly: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

[EDit3] TODO:
* the li items on mobile for autocomplete are too small to press
* the categories on mobile are to small to press
* ~~add a margin on top so the search input is not right against screen top~~
* on safari mobile the outline around the search box does not get the border radius applied
* the touch input detection does not work on android chrome mobile  the categories dont have a overflow and vertical scrollbar
* turns out border is actually part of height and width and outline not -> change everything back to border

MontyQI:
With changes from https://github.com/searxng/searxng/pull/829 the outline is broken
bar
It takes clear_search button's margin as part of it.

MontyQI:
To fix this problem with logo, we could move it to main, the same as link to settings already is.

mrpaulblack:
Oh well I also did fixes for everything, but did not push yet -_- sry...

I want to keep the logo in the grid tbh since I think its cleaner to have it in the grid... My solution was to just wrap the search box in another div and use that for the inner focus... If that is ok for you I would like to force push on top later today...

MontyQI:
I think it's ready to merge. It works well and the biggest problem has been fixed.

mrpaulblack:
Nope there is still one little cleanup thing left to do:
* put some CSS selectors on the search_view instead of .search_box ... Will do this today and then we can merge I think :+1:

mrpaulblack:
@MontyQI @return42 can you please retest? I tested this for a few hours on multiple different devices and all in all it seems to work... The things I noticed are (and IMO these are not blockers, but just to write it down):

* On Samsung devices the touch input detection does not work -> there is no scrollbar on the category overflow
* on iPhone safari the border-bottom from the search box is overlayed by autocomplete only when RTL languages are used
* on iPhone safari the new search form gets scrolled in on slightly when receiving focus (the user can scroll out again tho)
* on firefox the width of the search box and autocomplete box dont match by a pixel
* on chromium there is a small gap between the search box and autocomplete

mrpaulblack (approved):
LGTM

return42 (approved):
@MontyQI and @mrpaulblack thanks for all the efforts on this PR, for me it works like a charm :-) 

@mrpaulblack https://github.com/searxng/searxng/pull/963#issuecomment-1078405025 .. if you think we should track some of the issues, could you please open an issue for / thanks!

PR #967, opened 2022-03-16
[mod] add flags to the languages filter

return42:
## What does this PR do?

Alternative implementation of https://github.com/searxng/searxng/pull/938

The flags are more generic than in https://github.com/searxng/searxng/pull/938

Generic flags are needed since the language list is not a static list ..

https://github.com/searxng/searxng/blob/cd92a7eacdd6b4c333e57111e1503b8455628067/searxng_extra/update/update_languages.py#L126-L127


return42:
Just wondering the CI didn't run ...

dalf:
> Just wondering the CI didn't run ...

There was an incident on GH Action:
https://www.githubstatus.com/incidents/fpk08rxnqjz2

dalf:
First I'm sorry for the delay to provide answer.

IMO, flags are for the regions & the countries not for the languages. Some countries have multiples official languages, some languages are spoken in many countries. From that, I don't think this is a good idea to add flags to language only options ( `en`, `es`, `de`).

However we can the flags where the region is specified.

---

The fact there is only the native names & native regions is better.

----

Search engines show flags when the country choice, not when it's a language choice.

DDG:
image

Brave:
image

Qwant mixes region & language too, but the country is shown first, and after the language:
image

return42:
> I don't think this is a good idea to add flags to language only options ( `en`, `es`, `de`).

I understand, but I have a different opinion about.  I share the opinion of @MontyQI 

[The flags are to help recognise languages and I think that most people will associate UK flag with English and France flag with French etc.](https://github.com/searxng/searxng/pull/938#issuecomment-1058290221)

> The fact there is only the native names & native regions is better.

Sorry I do not understand this sentence .. could you help me and explain?

What do you think about sorting the menu, listing the languages (`en`, `es`, `de` ..) on top and the country codes (`en-EN`, `en-US` ..) at the end?

dalf:
Isn't strange to add a UK flag for an Indian user (or for a user from Hong Kong):
image

From https://en.wikipedia.org/wiki/List_of_countries_by_English-speaking_population :
image

In all these cases the UK flag is going to be irrelevant (except for UK of course).

From https://en.wikipedia.org/wiki/Nigeria :
> The [official language](https://en.wikipedia.org/wiki/Official_language) is [English](https://en.wikipedia.org/wiki/Nigerian_English), chosen to facilitate linguistic unity at the national level

From https://en.wikipedia.org/wiki/India
> [English](https://en.wikipedia.org/wiki/English_language) is used extensively in business and administration and has the status of a "subsidiary official language";[[5]](https://en.wikipedia.org/wiki/India#cite_note-FOOTNOTEMinistry_of_Home_Affairs_1960-6) it is important in [education](https://en.wikipedia.org/wiki/Education_in_India), especially as a medium of higher education.

return42:
> Isn't strange to add a UK flag for an Indian user (or for a user from Hong Kong)

It isn't what we do .. e.g. from DDG (is what this PR do):

grafik

.. the flag is meant to be an eye catcher with a high recognition value to select a language (with or without a region subtag).

dalf:
DDG shows regions not languages: Schweiz / Switzerland is country ; and when it makes sense the different languages spoken within a country.

SearXNG shows two different things:
* region: "de-CH" is the equivalent of "Schweiz (de)" in DDG.
* languages : "en" doesn't say anything about the location. It is up the engines to do their best to select English results without a region.

---

## Engine implementations of `en`

### google engine

<details>
<summary>detail</summary>

The first two return values of `get_lang_info` in of google.py:
https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/google.py#L160-L164

`get_lang_info` uses only `params['language']` from the user query.

If there is no country the code falls back to `US`:
https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/google.py#L197-L203

</details>

So `en` is `en-US` in the case of the google engine.

### duckduckgo engine

<details>

<summary>detail</summary>

https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/duckduckgo.py#L102-L105

The region list:
```javascript
    regions: {
        "wt-wt": "All regions",
        "ar-es": "Argentina",
        "au-en": "Australia",
        "at-de": "Austria",
        "be-fr": "Belgium (fr)",
        "be-nl": "Belgium (nl)",
        "br-pt": "Brazil",
        "bg-bg": "Bulgaria",
        "ca-en": "Canada (en)",
        "ca-fr": "Canada (fr)",
        "ct-ca": "Catalonia",
        "cl-es": "Chile",
        "cn-zh": "China",
        "co-es": "Colombia",
        "hr-hr": "Croatia",
        "cz-cs": "Czech Republic",
        "dk-da": "Denmark",
        "ee-et": "Estonia",
        "fi-fi": "Finland",
        "fr-fr": "France",
        "de-de": "Germany",
        "gr-el": "Greece",
        "hk-tzh": "Hong Kong",
        "hu-hu": "Hungary",
        "in-en": "India (en)",
        "id-en": "Indonesia (en)",
        "ie-en": "Ireland",
        "il-en": "Israel (en)",
        "it-it": "Italy",
        "jp-jp": "Japan",
        "kr-kr": "Korea",
        "lv-lv": "Latvia",
        "lt-lt": "Lithuania",
        "my-en": "Malaysia (en)",
        "mx-es": "Mexico",
        "nl-nl": "Netherlands",
        "nz-en": "New Zealand",
        "no-no": "Norway",
        "pk-en": "Pakistan (en)",
        "pe-es": "Peru",
        "ph-en": "Philippines (en)",
        "pl-pl": "Poland",
        "pt-pt": "Portugal",
        "ro-ro": "Romania",
        "ru-ru": "Russia",
        "xa-ar": "Saudi Arabia",
        "sg-en": "Singapore",
        "sk-sk": "Slovakia",
        "sl-sl": "Slovenia",
        "za-en": "South Africa",
        "es-ca": "Spain (ca)",
        "es-es": "Spain (es)",
        "se-sv": "Sweden",
        "ch-de": "Switzerland (de)",
        "ch-fr": "Switzerland (fr)",
        "tw-tzh": "Taiwan",
        "th-en": "Thailand (en)",
        "tr-tr": "Turkey",
        "us-en": "US (English)",
        "us-es": "US (Spanish)",
        "ua-uk": "Ukraine",
        "uk-en": "United Kingdom",
        "vn-en": "Vietnam (en)"
    },
```

At runtime `get_region_code` maps `en` to `us-en`.

</details>

So `en` is `en-US` in the case of the duckduckgo engine.

### bing engine

<details>
<summary>detail</summary>

The bing engine ignores the region and keep only the language.

With `en` as language, the user query `time` is sent as `language:en time` to bing.

https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/bing.py#L44-L72

</details>

So `en` is `en` in the case of the bing engine (no specified region).


### startpage engine

<details>
<summary>detail</summary>

Startpage allows to select a region see https://www.startpage.com/do/settings 

However, the engine is ignores it and handle only the language selection, see:
https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/startpage.py#L248

https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/searx/engines/startpage.py#L108-L136

With the `en` language, the HTTP request is `sp/search?query=time&page=1&cat=web&sc=XXXXX&language=english&lui=english`


</details>

So `en` is `en` in the case of the startpage engine (no specified region).

### Conclusion

IMO

* The USA flag would be better *for some engines*.
* In all other cases, with the `en` language, the engines select results written in English without a specified region. For someone outside the UK or even the USA, the UK flag is misleading.

## Language selection UI

In Google, the languages selection, there is no flag:
image

In DDG:
image

mrpaulblack:
Markus, Leon and I had a discussion in matrix and came to the same conclusion as you Alexandre...

That we should have flags for region and languages and no flags for only languages. AFAIK Markus is reworking the PR to include a generic icon for only languages and the flags for country + language. I think that is a good compromise...

return42:
> In all other cases, with the `en` language, the engines select results written in English without a specified region. For someone outside the UK or even the USA, the UK flag is misleading.

Thanks for the detail description .. and sorry for the efforts you had to explain me your concerns .. I pushed efdd0c7c on top to show 🌐 instead of a flag when there is only a language without a region (territory) identifier.  By example for the language "English":

grafik

.. does the fix meets your expectations?

dalf:
> to show 🌐 instead of a flag when there is only a language without a region (territory) identifier. 

Way more clear.

The language & region selection is not easy to deal with on many levels. For example if the user selects only google & ddg in the general category, it would make sense to show the US flag instead of 🌐, but that's nitpicking...

...and I realize that most of the choices includes a region but the region is not specify in `languages.py`. This exists before this PR, I'm not sure why.

---

~Do you know why these choices has been removed?~ --> see https://github.com/searxng/searxng/pull/986
* ('af-ZA', 'Afrikaans', '', 'Afrikaans'),
* ('be-BY', 'Беларуская', '', 'Belarusian'),
* ('fa-IR', 'فارسی', '', 'Persian'),
* ('hi-IN', 'हिन्दी', '', 'Hindi'),
* ('is-IS', 'Íslenska', '', 'Icelandic'),
* ('sr-RS', 'Српски', '', 'Serbian'),
* ('sw-TZ', 'Kiswahili', '', 'Swahili'),
* ('vi-VN', 'Tiếng Việt', '', 'Vietnamese'),

return42:
FYI I rebased to get bugfix from https://github.com/searxng/searxng/pull/986

> The language & region selection is not easy to deal with on many levels.

Don't look to close, otherwise it becomes  a more general discussion, we had similar about time range and safe-search: _the user can select an option and the meta-search engine do its best to comply with_.  This PR is about adding a symbol (flag) for visual orientation / nothing more.

I assume you are talking about entries like this ... ?

https://github.com/searxng/searxng/blob/7625da9fa02d0c2bb2c2bc618a41b245b9e2987f/searx/languages.py#L38-L48

return42:
> ...and I realize that most of the choices includes a region but the region is not specify in `languages.py`.

To fix, I pushed 2e4557f3 on top.

jmontane:
IMHO, it's a bad practice using flags to represent locales (language+country codes).
Flags represent territories/countries only, they don't represent languages neither locales. Using flag to represent locales is a [source of problems](https://issuetracker.google.com/issues/37002516?pli=1).

Another issue is inconsistency with description text (language) and setting (flag+locale) in Searchng general preferences:
search-language-en

return42:
See my post at https://github.com/searxng/searxng/pull/1029#issuecomment-1086824911

To shorten up: **flags are only shown if there is a region code**

The problem is that the region name is missed in the screenshot .. but this issue is not limited to catalan / we have to fix it and show the region name when there is a region code.

Trendyne:
Why does it only seem to be search language and not interface language?
image
image

Also for multi-region languages I'd recommend doing something similar to Tatoeba
https://tatoeba.org/en/sentences/show/7443501
https://tatoeba.org/en/sentences/index

PR #968, opened 2022-03-16
[simple theme] Normalize info page

mrpaulblack:
## What does this PR do?

Remove the 1.3em font-size and the doubled sans-serif font definition from CSS.

I think it is better to handle this in one place on the body instead of making the text way bigger in a single part of the site.

```make run```

Visit the info page.


return42 (approved):
Thanks!

PR #979, opened 2022-03-17
Simple upgrade dependencies

dalf:
## What does this PR do?

Upgrade some dependencies except:

* less-plugin-clean-css: no updated version.
  @wikipedia/less-plugin-clean-css might be an alternative.

* stylelint & stylelint-config-standard
  the new versions require configuration and source code changes

About the update packages:
* https://github.com/ionic-team/ionicons/releases/tag/v6.0.1
* https://github.com/eslint/eslint/releases
* https://github.com/less/less.js/releases
* https://sharp.pixelplumbing.com/changelog

Try the simple theme...

## Author's checklist

Since `package-lock.json` is not commited, depdendabot doesn't work for the NPM dependencies.


searx/static/themes/simple/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "devDependencies": {
-    "eslint": "^8.0.1",
+    "eslint": "^8.11.0",
```
return42, line 3: Does it make sense to upgrade eslint in the root also:

https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/package.json#L3

dalf: Actually package-lock.json should be commited, so dependabot can work.

dalf:
search.brave.com:

image

A bright background makes sense.

searx/static/themes/simple/package.json
return42: My idea:

1. drop oscar
2. merge node env from simple theme with the one in root `.`
3. commit `package-lock.json`

dalf: Does this block this PR?

return42: No, but while we are, we should also update 

https://github.com/searxng/searxng/blob/6ccd557650a0892a3c59ac157a781f9a12beb56e/package.json#L3

return42 (approved):
I rebased branch and updated [eslint in ./package.json](https://github.com/searxng/searxng/pull/979/commits/6eae697c73cd79807c5da6d10d843aacfc9f759f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)

PR #985, opened 2022-03-19
docker: log to stdout

dalf:
## What does this PR do?

previously the log (only the exceptions) were log
into /var/log/uwsgi/uwsgi.log

this is disturbing for the admins:
* they see an internal error on HTTP port
* no log where they are expected (docker logs)

this commit fixes this issue

Many times admins can't figure out why the application is not starting.

## Author's checklist

We may want to add a log rotation in searxng-docker:
https://docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,6 @@ ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.
 EXPOSE 8080
 VOLUME /etc/searx
 VOLUME /etc/searxng
-VOLUME /var/log/uwsgi
 
 ARG SEARXNG_GID=977
 ARG SEARXNG_UID=977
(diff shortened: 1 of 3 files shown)


mrpaulblack:
Originally I was against this since I thought this would spam the docker logs, but I agree that this is the better options since exceptions should be easily visible and not hidden in a file somewhere :+1:

mrpaulblack (changes requested):
I think this needs to be removed too:

https://github.com/searxng/searxng/blob/b6920351200dc7d56ba4529758073ea2da219004/Dockerfile#L6

https://github.com/searxng/searxng/blob/b6920351200dc7d56ba4529758073ea2da219004/dockerfiles/docker-entrypoint.sh#L168-L169

mrpaulblack (approved):
I tested with:
`make docker`
`docker run -it --rm -p 8888:8080 dalf/searxng:2022.03.19-12d3fe32`

and UWSGI log now works correctly with `docker logs`

LGTM :+1: (also the docker image worked for me)

mrpaulblack:
Oh I just saw that you also requested a review from unixfox; hope its ok that I already merged -_-

PR #990, opened 2022-03-19
settings.yml: allow to base_url with the SEARXNG_BASE_URL env variable

dalf:
## What does this PR do?

In https://github.com/searxng/searxng-docker/pull/12 everything is configured using settings.yml except:
* the hostname
* the email address for the certificate

From the hostname, docker-compose.yml provides the base_url.
The idea is not to ask the admin to repeat the same information.

However, the calls to sed in [docker-entrypoint.sh](https://github.com/searxng/searxng/blob/master/dockerfiles/docker-entrypoint.sh) to copy and replace base_url becomes obsolete.

This PR allows to set the SEARXNG_BASE_URL environment variable to define the base_url.

One step further to remove morty / filtron.

* set SEARXNG_BASE_URL  to `http://127.0.1.1/`
* start the application
* checks the links use `http://127.0.1.1/`

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -165,7 +165,7 @@ SCHEMA = {
         'bind_address': SettingsValue(str, '127.0.0.1', 'SEARXNG_BIND_ADDRESS'),
         'limiter': SettingsValue(bool, False),
         'secret_key': SettingsValue(str, environ_name='SEARXNG_SECRET'),
-        'base_url': SettingsValue((False, str), False),
+        'base_url': SettingsValue((False, str), False, 'SEARXNG_BASE_URL'),
         'image_proxy': SettingsValue(bool, False),
         'http_protocol_version': SettingsValue(('1.0', '1.1'), '1.0'),
         'method': SettingsValue(('POST', 'GET'), 'POST'),


mrpaulblack:
I dont quite get the point of this PR... Why do we not just set the base url in settings.yml but rather keep it as ENV var? (If I get it right since you want to keep base_url as env var you want to make it easier to set it without the use of `sed` in the entrypoint script?)

dalf:
searxng-docker needs the hostname in two locations:
* https://github.com/searxng/searxng-docker/blob/51fee80abb0d7a50b3a98ab20276cea9fa4efa8c/Caddyfile#L5
* settings.yml in a way or another : https://github.com/searxng/searxng-docker/blob/51fee80abb0d7a50b3a98ab20276cea9fa4efa8c/searxng/settings.yml

The hostname is defined in .env: https://github.com/searxng/searxng-docker/blob/51fee80abb0d7a50b3a98ab20276cea9fa4efa8c/.env

In docker-compose.yml:
* https://github.com/searxng/searxng-docker/blob/51fee80abb0d7a50b3a98ab20276cea9fa4efa8c/docker-compose.yaml#L14
* https://github.com/searxng/searxng-docker/blob/51fee80abb0d7a50b3a98ab20276cea9fa4efa8c/docker-compose.yaml#L48

> Why do we not just set the base url in settings.yml but rather keep it as ENV var?

How Caddy will get the hostname?

Currently there is `start.sh` scripts to check the setup and then start the app, IMO it should be removed and replaced by `docker-compose up` : keep it stupid simple.

mrpaulblack:
Oh ok did not even thought about this mb sry. So yeah makes sense to have the hostname in `.env` for that case :+1: 

> Currently there is start.sh scripts to check the setup and then start the app, IMO it should be removed and replaced by docker-compose up : keep it stupid simple.

Agreed I think the start, util and stop script should be removed (as well as the service) -> docker daemon will take care of running the container and `docker-compose up -d` is better KISS solution :+1:

return42 (approved):
Checked there is non side effect on the shell scripts .. LGTM :+1:

PR #991, opened 2022-03-19
Fix 894

dalf:
## What does this PR do?

The <input type="reset"> introduced in the PR 894, restores the default value.
It works in the index page, but it doesn't work in the /search page:
the reset button restore the initial query.

This PR:
* fix the JS version: the reset button clear the text
* keep the clear button in the / page
* hide the clear button in the /search page

### without JS

The clear button works here:
image

The clear button is hidden on the result page (behavior copied from the Oscar theme):
image

### with JS

image

Fix UI / UX issue.

## Related issues

Related to #894


mrpaulblack (approved):
LGTM :+1: 

(You could also remove the Clear hide-if-js span from `search.html`, but not really a blocker for the PR IMO -> I leave merge up to you)

PR #994, opened 2022-03-20
[doc] add some documentation about the limiter plugin (and redis)

return42:
Requested-by: https://github.com/searxng/searxng/discussions/993#discussioncomment-2396914

To test run `make docs.clean docs.live` and visit the modified pages.

- admin/filtron.html
- src/searx.plugins.limiter.html

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -226,6 +226,17 @@ Global Settings
 
 .. _Redis.from_url(url): https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url
 
+A redis DB can be connected by an URL, in :py:obj:`searx.shared.redisdb` you
+will find a description to test your redis connection in SerXNG.  When using
+sockets, don't forget to check the access rights on the socket::
+
+  ls -la /usr/local/searxng-redis/run/redis.sock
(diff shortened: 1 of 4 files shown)


mrpaulblack (approved):
LGTM :+1: (I think eventually we should remove old stuff like filtron and morty from the documentation, but for now I am ok with the tip)

return42:
>  think eventually we should remove old stuff like filtron and morty from the documentation, but for now I am ok with the tip

Yes, I will remove the documentation as soon I have the limiter Plugin in the shell scripts to install.

My plan is to remove

- https://github.com/searxng/searxng/blob/master/utils/morty.sh and
- https://github.com/searxng/searxng/blob/master/utils/filtron.sh

entirely .. since the major part of the following documentation is generated by theses scripts there will be more files that can be dropped or even cleaned up ..

- https://docs.searxng.org/admin/installation-nginx.html
- https://docs.searxng.org/admin/installation-apache.html
- https://docs.searxng.org/utils/filtron.sh.html
- https://docs.searxng.org/utils/morty.sh.html

And https://docs.searxng.org/admin/filtron.html can also be removed (or reduced to a limiter installation)

Its a lot of work I haven't had time for .. hope I can address this task in the near future.

Issue #964, opened 2022-03-14
Since PR 932 static files can't be delivered by HTTP server any longer


dalf:
( from https://github.com/searxng/searxng/pull/932#issuecomment-1067039518 )

Since https://github.com/searxng/searxng/pull/932 , the URL of the static files contains their hashes:
`/static/themes/simple/css/searxng.min.css?5fde34a74bc438c7b56ec8c6501e131cc9914bd8`

According to https://serverfault.com/questions/934244/have-files-with-querystring-included-in-their-name-need-nginx-to-ignore-queryst to the Content-Type header is missing with nginx with the default configuration.

return42:
IMO it is better to fix this issue as you suggested here https://github.com/searxng/searxng/pull/932#issuecomment-1067216426

return42:
just to remember : since https://github.com/searxng/searxng/pull/995 has been merged, this issue only takes place when `static_use_hash` is active (by default this feature is inactive):

https://github.com/searxng/searxng/blob/b94fdd0df0c01099554e04e7ae53c2b135e6a1fd/searx/settings.yml#L72-L75

PR #995, opened 2022-03-20
[interim fix] static files can't be delivered by HTTP.

return42:
Since PR 932 [1][2] static files can't be delivered by HTTP server any longer.

This patch makes the hash paramter in the URL of static files:

    /static/themes/simple/css/searxng.min.css?5fde34a74bc438c7b56ec8c6501e131cc9914bd8

optional.  By default the hash parameter is disabled.

HINT:

  Instances that do not deliver static files by their HTTP server and have a
  long expire time [3] should enable this option.

----

This is only a interim solution, on the long run:

    make static.build.commit

creates files including the file name:

    css/searxng-5fde34a74bc438c7b56ec8c6501e131cc9914bd8.min.css

and a mapping.json with this content[4]

[1] https://github.com/searxng/searxng/issues/964
[2] https://github.com/searxng/searxng/pull/932#issuecomment-1067039518
[3] https://github.com/searxng/searxng/pull/932/commits/55833364401ed72c30b3e85c16824a75de6b547a
[4] https://github.com/searxng/searxng/pull/932#issuecomment-1067216426

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -70,6 +70,7 @@ redis:
 ui:
   # Custom static path - leave it blank if you didn't change
   static_path: ""
+  static_use_hash: false
   # Custom templates path - leave it blank if you didn't change
   templates_path: ""
   # query_in_title: When true, the result page's titles contains the query
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -176,6 +176,7 @@ SCHEMA = {
     },
     'ui': {
         'static_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'static')),
+        'static_use_hash': SettingsValue(bool, False),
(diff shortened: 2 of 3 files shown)


return42:
@dalf @mrpaulblack I assume this option should be enabled in the docker image .. is there anything we can do for the docker image in this PR?

mrpaulblack:
Yes in the Dockerfile we could set the value to true in settings.yml with sed? I do this also in my fork a lot: https://github.com/paulgoio/searxng/blob/ee18acc423b9676f91a2843321bf14dfc5fbf783/Dockerfile#L31

mrpaulblack (approved):
LGTM :+1: 

Actually after thinking about this again, we can just set this option to true in searxng-docker since we are going to maintain some `settings.yml` changes for docker there anyway

PR #997, opened 2022-03-20
 [fix] Text can't be selected when using simple theme with JS enabled

### LINKED ISSUE #794: Text can't be selected when using simple theme with JS enabled

return42:
## What does this PR do?

[fix] Text can't be selected when using simple theme with JS enabled

    make run

In simple theme try to select text in results of the result page, with and without vim keys enabled.

## Related issues

Closes: https://github.com/searxng/searxng/issues/794


mrpaulblack (approved):
LGTM :+1:  (I tested with desktop ff)

PR #998, opened 2022-03-21
Add rate limiting per engine

MarcAbonce:
## What does this PR do?

Add a plugin that limits how many outgoing requests is SearXNG going to send per engine.

This is very similar to the [limiter](https://github.com/searxng/searxng/blob/master/searx/plugins/limiter.py) plugin and I actually based my code on that, but this one works in the opposite direction.

This plugin is always "on", but it doesn't do anything unless Redis is used and an engine has a `rate_limit` setting. If an engine doesn't have the setting, the plugin should skip (always pass) that engine without writing anything in Redis.

Some engines are more strict than others when dealing with our traffic. For example, you can throw Wikipedia over a hundred requests per second with no problems, but other engines will throw you a CAPTCHA as soon as you send a few requests in rapid succession *\*cough\*cough\*Google\**.

1. Run Redis.
1. Point the `redis: url` setting to your Redis db in `settings.yml`.
1. In `settings.yml` add the following to any engine:
```yml
  # This would limit that engine to a maximum of 2 requests per minute.
  rate_limit:
    - max_requests: 2
      interval: 60
```
4. Run queries with both the rate limited engine and an engine without the setting.
5. Rate limited engine should stop sending requests after reaching limit and other engines should still work as usual.
6. Engine should be usable again after time interval has passed.

## Author's checklist
* Are the plugin and setting names clear enough? Could instance maintainers be confused between this and the `limiter` plugin?
* ~~Am I using the correct formatting in the documentation? The ref link doesn't do anything for me, but nor do any of the other ones in the same page.~~
* I've only tested this locally with very light traffic.

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -422,6 +422,9 @@ engine is shown.  Most of the options have a default value or even are optional.
      max_connections: 100
      max_keepalive_connections: 10
      keepalive_expiry: 5.0
+     rate_limit:
+        max_requests: 200
+        interval: 60
      proxies:
        http:
          - http://proxy1:8080
@@ -486,6 +489,15 @@ engine is shown.  Most of the options have a default value or even are optional.
 
   - ``ipv4`` set ``local_addresses`` to ``0.0.0.0`` (use only IPv4 local addresses)
   - ``ipv6`` set ``local_addresses`` to ``::`` (use only IPv6 local addresses)
+
+``rate_limit``: optional
+  Limit how many outgoing requests is SearXNG going to send to the engines.
+  Requires :ref:`Redis <settings redis>`.
+
+  - ``max_requests`` is the maximum number of requests that will be sent to this
+    engine per interval.
+  - ``interval`` (optional) is the number of seconds before this engine's rate
+    limiter is reset. Defaults to 1 second if unspecified.
 
 .. note::
 
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -423,8 +423,8 @@ engine is shown.  Most of the options have a default value or even are optional.
      max_keepalive_connections: 10
      keepalive_expiry: 5.0
      rate_limit:
-        max_requests: 200
-        interval: 60
+        - max_requests: 200
+          interval: 60
      proxies:
        http:
          - http://proxy1:8080
--- a/searx/plugins/engine_rate_limiter.py
+++ b/searx/plugins/engine_rate_limiter.py
@@ -11,9 +11,6 @@ Enable ``settings.yml``:
 - ``interval: ...`` number of seconds before rate limiting resets (optional, by default 1 second)
 """
 
-import hmac
-
-from searx import settings
 from searx.engines import engines
 from searx.shared import redisdb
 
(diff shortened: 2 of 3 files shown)


docs/admin/engines/settings.rst
```diff
@@ -377,6 +380,14 @@ engine is shown.  Most of the options have a default value or even are optional.
   * ``ipv4`` set ``local_addresses`` to ``0.0.0.0`` (use only IPv4 local addresses)
   * ``ipv6`` set ``local_addresses`` to ``::`` (use only IPv6 local addresses)
 
+``rate_limit``: optional
+  Limit how many outgoing requests is SearXNG going to send to the engines.
+  Requires redis, see :ref:`redis db`.
+  * ``max_requests`` is the maximum number of requests that will be sent to this
+    engine per interval.
+  * ``interval`` (optional) is the number of seconds before this engine's rate
+    limiter is reset. Defaults to 1 second if unspecified.
```
return42: > Am I using the correct formatting in the documentation? 

Sorry, the issue was already in the "network" paragraph before your paragraph , I fixed it in the master branch:

https://github.com/searxng/searxng/blame/master/docs/admin/engines/settings.rst#L450-L455

In reST paragraphs, lists .. are all "block" elements.  

To start a new block (e.g. a list) a blank line is needed.

> The ref link doesn't do anything for me, but nor do any of the other ones in the same page.

for me the link works, did you use::

    make docs

or for live build:

    make docs.live

MarcAbonce: That worked, thanks.

return42:
@MarcAbonce sorry for the delay .. today I want to review this PR .. is it OK when I push a "WIP" commit with my suggestions on top of your branch? update: and I like to rebase this branch ..

MarcAbonce:
I rebased my branch and made two changes:

1. Based on [this comment](https://github.com/MarcAbonce/searxng/commit/69d588a0f3db2ef1fe4bca5d4238389c2b66b38d#r69242842) I set the key in plain text.
2. This setting is now a list that allows for more than one rate limiter per engine. So now you can rate limit by second and by minute, for example.

> is it OK when I push a "WIP" commit with my suggestions on top of your branch? update: and I like to rebase this branch ..  

Yeah, sure.

return42:
> 1. Based on [this comment](https://github.com/MarcAbonce/searxng/commit/69d588a0f3db2ef1fe4bca5d4238389c2b66b38d#r69242842) I set the key in plain text.

I am just wondering that I can't see the comment from @dalf here in the PR.

> 2. This setting is now a list that allows for more than one rate limiter per engine. So now you can rate limit by second and by minute, for example.

OK, ATM I am in the review and it could take its time .. in that time please do not change the implementation.

dalf:
I feel this should not be a plugin.

Example:
* an engine has multiple configured proxies, I guess the expected result is to set the limit per proxy, not for the whole engine.
* the code can even skip the proxies that can't send request anymore and use available proxies.

@MarcAbonce what do you think?

---

Related: https://github.com/searxng/searxng/blob/dec04c0ed640caac3b901787aca2a14452734fe7/searx/search/processors/abstract.py#L24-L57 should be stored in redis, so when an engine is suspended, it is for all the workers. And same: an engine can be suspended with one proxy, but not the others. (so the suspended status should stored in the Network class).

However Redis should remains optional IMO.

So:
* an idea is to have an very light abstraction in https://github.com/searxng/searxng/tree/master/searx/shared to store states in memory instead of Redis, if Redis is not installed.
* another idea is to have two implementation for `searx.network.network.Network`: the current one, and subclass that relies on Redis (this option seems easier).

searx/plugins/engine_rate_limiter.py
```diff
@@ -0,0 +1,74 @@
...
+        return count
+    else
+        local newCount = redis.call('INCR', engine)
+        if newCount == 1 then
+            redis.call('EXPIRE', engine, interval)
+        end
+        return newCount
+    end
+    """
+    script_sha = redis_client.script_load(lua_script)
```
dalf, line 41: If `script_sha` is initialized in the `init` function, then this Python function doesn't have upload the script at each call.

return42: I am on the way to polish the implementation .. give me the time I need for.

MarcAbonce:
Yeah, I wasn't sure if this should be a plugin. I initially based my work on the limiter plugin, but if this feature is always going to be on, then I agree that it would make sense to move it somewhere else.

I also agree that limiting and blocking by proxy would be nice, although that would take me longer to implement.

return42:
> I feel this should not be a plugin.

Your arguments are weighty, yet I would like to ask if we can first review this PR as is .. and in a second PR we can move it to an core component to support multiple proxies (I hope I can continue my review next weekend).

@dalf would this a way we could go?

dalf:
I would like to avoid an intermediate implementation if possible, except if someone requires this feature quickly.

>  if this feature is always going to be on, then I agree that it would make sense to move it somewhere else.

If redis is not available, this feature will be either disable or operate in a degraded manner (the state will be not be shared between the workers).

return42:
> I would like to avoid an intermediate implementation if possible

OK, then I stop my review / but it is not clear to me what the future of this PR is .. 

>  although that would take me longer to implement.

may then you @dalf implement the functionality since me and @MarcAbonce do not have time and the needed knowledge to do it right in all manner.

MarcAbonce:
This feature isn't urgent, though, so there's no hurry. I could start re-writing it and open a new draft PR sometime next month.

MarcAbonce:
I opened #1273 with a rewrite of this feature, so I'm going to close this PR.

PR #996, opened 2022-03-20
[docs] improve settings.yml documentation (admin/engines/settings)

return42:
## What does this PR do?

improve settings.yml documentation

## How to test this PR locally?

    make docs.clean docs.live

visit modified file ``admin/engines/settings.html``

PR #999, opened 2022-03-21
[docs] document settings.yml: search.languages

return42:
Requested-by: @dalf https://github.com/searxng/searxng/pull/996#discussion_r830858139

I did not documented the language option `all` since it is unclear to me /  or it is not implemented in all engines in a common manner.

What is the meaning of `all`in the `search.languages` setting? ..

- Does it mean: _"the default language from the browser information is used"_ or
- does it mean: _"the `search.default_language` is used"_ or
- does it mean: _"if there is no language from browser information the `search.default_language` is used"_ or
- ...

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -123,6 +123,24 @@ Global Settings
   Default search language - leave blank to detect from browser information or
   use codes from :origin:`searx/languages.py`.
 
+``languages``:
+  List of available languages - leave unset to use all codes from
+  :origin:`searx/languages.py`.  Otherwise list codes of available languages.
+  The ``all`` value is shown as the ``Default language`` in the user interface
+  (in most cases, it is meant to send the query without a language parameter ;
+  in some cases, it means the English language) Example:
+
+  .. code:: yaml
+
+     languages:
+       - all
+       - en
+       - en-US
+       - de
+       - it-IT
+       - fr
+       - fr-BE
+
 ``ban_time_on_fail``:
   Ban time in seconds after engine errors.
 
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -21,6 +21,15 @@ search:
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: ""
+  # Available languages
+  # languages:
+  #   - all
+  #   - en
+  #   - en-US
+  #   - de
+  #   - it-IT
+  #   - fr
+  #   - fr-BE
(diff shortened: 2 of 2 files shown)


MarcAbonce:
Oh yeah, the good old `all` language (aka "Default language" and previously "Automatic").

It means different things on each engine, but in most cases it's meant to send the query without a language parameter, so we instead let the engine guess the language or do whatever it wants. So it's more like a "None" option.

But some engines require a language to be passed, so I think in those cases it's hardcoded to English.

Some comments I've had about this over the years:
https://github.com/searx/searx/discussions/3102
https://github.com/searx/searx/pull/973
https://github.com/searx/searx/issues/1085#issuecomment-345467928
https://github.com/searx/searx/issues/1236#issuecomment-381483831

dalf (approved):
I've updated the commit to add `all`.

return42:
> I've updated the commit to add `all`.

Since it is an real option, I think it is correct to add `all` here but without an explanation it might confuse the reader what `all` means in this list.

```reST
``languages``:
  List of available languages - leave unset to use all codes ...
  .. code:: yaml
     languages:
       - all
```

from https://github.com/searxng/searxng/pull/999#issuecomment-1074592605 :

> It means different things on each engine, but in most cases it's meant to send the query without a language parameter, so we instead let the engine guess the language or do whatever it wants. So it's more like a "None" option.

I think we should have a better description for this `all` value, more like @MarcAbonce described in his comment.  May be:

```reST
``languages``:
  List of available languages - leave unset to use all codes from
  :origin:`searx/languages.py`.  Otherwise list codes of available
  languages.  The special value ``all`` sends the search request
  without a language parameter and let the engine guess the language
  or do whatever it wants. Example:

  .. code:: yaml
     languages:
       - all
```
 
There is only one thing left that confuse me: where does the language detection from the browser comes into play .. is it related to this `all` value or where is it related to?

dalf:
>There is only one thing left that confuse me: where does the language detection from the browser comes into play .. 

https://github.com/searxng/searxng/blob/ef8af285a22db78a3d28d0f13228f0487ce9fe8b/searx/webapp.py#L545-L555

This code above run after reading the preferences from the cookies and parameters (GET & POST).
* If there is no `language` (the search language), then the value is read from the HTTP header `Accept-Language`.
* If there is no `locale` (the UI language), then then the value is read from the HTTP header `Accept-Language`.

So `all` does not do anything here.

All is shown as "Default language" in the UI:
image

dalf:
I've updated the commit

PR #1004, opened 2022-03-24
Search_url only if method is POST

MontyQI:
## What does this PR do?

Shows search_url only if method is set to POST.

**POST:**
image

**GET:**
image

Doesn't show search_url if it isn't needed (if you have GET method, you can copy url from address bar)

`make run`
## Related issues

### DIFF
--- a/searx/templates/simple/results.html
+++ b/searx/templates/simple/results.html
@@ -70,8 +70,11 @@
             {% for suggestion in suggestions %}
             <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
               <input type="hidden" name="q" value="{{ suggestion.url }}">
-              <input type="hidden" name="time_range" value="{{ time_range }}">
+              {% for category in selected_categories %}
+              <input type="hidden" name="category_{{ category }}" value="1">
+              {% endfor %}
               <input type="hidden" name="language" value="{{ current_language }}">
+              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="safesearch" value="{{ safesearch }}">
               <input type="hidden" name="theme" value="{{ theme }}">
               {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
@@ -82,10 +85,12 @@
         </div>
         {% endif %}
 
+        {% if method == 'POST' %}
         <div id="search_url" role="complementary" aria-labelledby="search_url-title">
             <h4 class="title" id="search_url-title">{{ _('Search URL') }} :</h4>
             <div class="selectable_url"><pre>{{ url_for('search', _external=True) }}?q={{ q|urlencode }}&amp;language={{ current_language }}&amp;time_range={{ time_range }}&amp;safesearch={{ safesearch }}{% if pageno > 1 %}&amp;pageno={{ pageno }}{% endif %}{% if selected_categories %}&amp;categories={{ selected_categories|join(",") | replace(' ','+') }}{% endif %}{% if timeout_limit %}&amp;timeout_limit={{ timeout_limit|urlencode }}{% endif %}</pre></div>
         </div>
+        {% endif %}
         <div id="apis" role="complementary" aria-labelledby="apis-title">
           {% if search_formats %}
           <h4 class="title" id="apis-title">{{ _('Download results') }}</h4>
@@ -97,8 +102,8 @@
               <input type="hidden" name="category_{{ category }}" value="1">
               {% endfor %}
               <input type="hidden" name="pageno" value="{{ pageno }}">
-              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="language" value="{{ current_language }}">
+              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="safesearch" value="{{ safesearch }}">
               <input type="hidden" name="format" value="{{ output_type }}">
               {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
@@ -115,15 +120,18 @@
       <h4 id="corrections-title">{{ _('Try searching for:') }}</h4>
       {% for correction in corrections %}
       <div class="left">
-	<form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" role="navigation">
+	      <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" role="navigation">
+          {% for category in selected_categories %}
+          <input type="hidden" name="category_{{ category }}" value="1">
+          {% endfor %}
           <input type="hidden" name="q" value="{{ correction.url }}">
-          <input type="hidden" name="time_range" value="{{ time_range }}">
           <input type="hidden" name="language" value="{{ current_language }}">
+          <input type="hidden" name="time_range" value="{{ time_range }}">
           <input type="hidden" name="safesearch" value="{{ safesearch }}">
           <input type="hidden" name="theme" value="{{ theme }}">
           {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit }}" >{% endif %}
           <input type="submit" role="link" value="{{ correction.title }}">
-	</form>
+	      </form>
       </div>
       {% endfor %}
     </div>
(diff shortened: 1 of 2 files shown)


mrpaulblack (approved):
LGTM; Woow great idea, thanks!

~~(Tho one could argue that the search URL contains the entire form params like the theme and so on so the result page always stays the same when using the link. With the URL in the browser address bar there are only the params which differ from default settings... But I still agree its better to hide this when using the get method)~~

Just in case I will also asign @dalf to review...

dalf:
If I focus only what is doing the "Search URL", from the front page, the URL contains only the `q` parameter. It should contains `language`, `time_range`, `safesearch`, `theme`, `timeout_limit` .

Here a possible update of searx/templates/simple/simple_search.html :
```jinja
<form id="search" method="{{ method or 'POST' }}" action="{{ url_for('search') }}" role="search">
  <div id="search_header">
    <div id="search_view">
      <div class="search_box">
        <input id="q" name="q" type="text" placeholder="{{ _('Search for...') }}" autocomplete="off" autocapitalize="none" spellcheck="false" autocorrect="off" dir="auto" value="{{ q or '' }}">
        <button id="clear_search" type="reset" aria-label="{{ _('clear') }}"><span class="hide_if_nojs">{{ icon_big('close') }}</span><span class="show_if_nojs">{{ _('clear') }}</span></button>
        <button id="send_search" type="submit" aria-label="{{ _('search') }}"><span class="hide_if_nojs">{{ icon_big('search-outline') }}</span><span class="show_if_nojs">{{ _('search') }}</span></button>
      </div>
    </div>
  </div>
  {% for category in selected_categories %}
  <input type="hidden" name="category_{{ category }}" value="1" >
  {% endfor %}
  <input type="hidden" name="language" value="{{ current_language }}" >
  <input type="hidden" name="time_range" value="{{ time_range }}" >
  <input type="hidden" name="safesearch" value="{{ safesearch }}" >
  <input type="hidden" name="theme" value="{{ theme }}" >
  {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
</form>
```

<details>
<summary>possible patch to normalize the parameters order & values (at least a little better)</summary>

```diff
diff --git a/searx/templates/simple/results.html b/searx/templates/simple/results.html
index 3063a915..b05841bb 100644
--- a/searx/templates/simple/results.html
+++ b/searx/templates/simple/results.html
@@ -70,8 +70,11 @@
             {% for suggestion in suggestions %}
             <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
               <input type="hidden" name="q" value="{{ suggestion.url }}">
-              <input type="hidden" name="time_range" value="{{ time_range }}">
+              {% for category in selected_categories %}
+              <input type="hidden" name="category_{{ category }}" value="1">
+              {% endfor %}
               <input type="hidden" name="language" value="{{ current_language }}">
+              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="safesearch" value="{{ safesearch }}">
               <input type="hidden" name="theme" value="{{ theme }}">
               {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
@@ -99,8 +102,8 @@
               <input type="hidden" name="category_{{ category }}" value="1">
               {% endfor %}
               <input type="hidden" name="pageno" value="{{ pageno }}">
-              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="language" value="{{ current_language }}">
+              <input type="hidden" name="time_range" value="{{ time_range }}">
               <input type="hidden" name="safesearch" value="{{ safesearch }}">
               <input type="hidden" name="format" value="{{ output_type }}">
               {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
@@ -117,15 +120,18 @@
       <h4 id="corrections-title">{{ _('Try searching for:') }}</h4>
       {% for correction in corrections %}
       <div class="left">
-	<form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" role="navigation">
+	      <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" role="navigation">
+          {% for category in selected_categories %}
+          <input type="hidden" name="category_{{ category }}" value="1">
+          {% endfor %}
           <input type="hidden" name="q" value="{{ correction.url }}">
-          <input type="hidden" name="time_range" value="{{ time_range }}">
           <input type="hidden" name="language" value="{{ current_language }}">
+          <input type="hidden" name="time_range" value="{{ time_range }}">
           <input type="hidden" name="safesearch" value="{{ safesearch }}">
           <input type="hidden" name="theme" value="{{ theme }}">
           {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit }}" >{% endif %}
           <input type="submit" role="link" value="{{ correction.title }}">
-	</form>
+	      </form>
       </div>
       {% endfor %}
     </div>
@@ -151,16 +157,16 @@
             <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" class="previous_page">
                 <div class="{% if rtl %}right{% else %}left{% endif %}">
                   <input type="hidden" name="q" value="{{ q|e }}" >
-                  {{- engine_data_form(engine_data) -}}
                   {% for category in selected_categories %}
                   <input type="hidden" name="category_{{ category }}" value="1" >
                   {% endfor %}
                   <input type="hidden" name="pageno" value="{{ pageno-1 }}" >
-                  <input type="hidden" name="time_range" value="{{ time_range }}" >
                   <input type="hidden" name="language" value="{{ current_language }}" >
+                  <input type="hidden" name="time_range" value="{{ time_range }}" >
                   <input type="hidden" name="safesearch" value="{{ safesearch }}" >
                   <input type="hidden" name="theme" value="{{ theme }}" >
                   {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
+                  {{- engine_data_form(engine_data) -}}
                   <button role="link" type="submit">{{ icon_small('chevron-left') }} {{ _('Previous page') }}</button>
                 </div>
             </form>
@@ -168,16 +174,16 @@
         <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}" class="next_page">
             <div class="{% if rtl %}left{% else %}right{% endif %}">
               <input type="hidden" name="q" value="{{ q|e }}" >
-              {{- engine_data_form(engine_data) -}}
               {% for category in selected_categories %}
               <input type="hidden" name="category_{{ category }}" value="1" >
               {% endfor %}
               <input type="hidden" name="pageno" value="{{ pageno+1 }}" >
-              <input type="hidden" name="time_range" value="{{ time_range }}" >
               <input type="hidden" name="language" value="{{ current_language }}" >
+              <input type="hidden" name="time_range" value="{{ time_range }}" >
               <input type="hidden" name="safesearch" value="{{ safesearch }}" >
               <input type="hidden" name="theme" value="{{ theme }}" >
               {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
+              {{- engine_data_form(engine_data) -}}
               <button role="link"  type="submit">{{ _('Next page') }} {{ icon_small('chevron-right') }}</button>
             </div>
         </form>
diff --git a/searx/templates/simple/simple_search.html b/searx/templates/simple/simple_search.html
index 26d25130..30a6c2ad 100644
--- a/searx/templates/simple/simple_search.html
+++ b/searx/templates/simple/simple_search.html
@@ -8,4 +8,12 @@
       </div>
     </div>
   </div>
+  {% for category in selected_categories %}
+  <input type="hidden" name="category_{{ category }}" value="1" >
+  {% endfor %}
+  <input type="hidden" name="language" value="{{ current_language }}" >
+  <input type="hidden" name="time_range" value="{{ time_range }}" >
+  <input type="hidden" name="safesearch" value="{{ safesearch }}" >
+  <input type="hidden" name="theme" value="{{ theme }}" >
+  {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
 </form>
```
</details>

---

Open questions:
* what is the purpose of this "Search URL" box?
* when do the user copy / paste the URL from the browser?

I see three use cases:
* the user would like to report an issue using the URL.
* the user would like to share the results to another user.
* the user would like to bookmark / save the query.

In the first scenario, it would be useful to get all the parameters, include the engine selection (basically the very long value of `preferences` in `/preferences`).

In the second scenario, the user would like to share the results, and the receiver would like to keep the already configured preferences. So what is important here is the engine selection, and search parameters (language, time_range, safesearch, etc...)

The third scenario is similar to the second one.

Currently, neither the browser URL nor the "search URL" contains the engine selection.

----

The parameters that should be added as hidden input field (or not according the use case):
* UI
  * simple_style
  * infinite_scroll
  * results_on_new_tab
* search
  * doi_resolver 
  * autocomplete
* privacy:
  * image_proxy
  * query_in_title

mrpaulblack:
OK I agree. But this opens opens up a question about how we handle settings... Since we basically have some settings as search params and some settings as Cookie/preferences string encoded in the URL. How would it be handled if the user uses a preferences string in the URL to set s specific theme and then sets the param for that in the search form to something different?

dalf:
Right now I don't know :-)

That's why I suggest to skip these questions for now: I've rewritten my last comment a few times, and I forgot to mention that these questions are not blocker for this PR.

The only thing I [would like to] change in the PR is the update of `searx/templates/simple/simple_search.html`, and then we can merge IMO.

[EDIT] Add [would like to]

return42:
I think we should not enlarge the use cases .. the URL was maid with the aim to share & bookmark search queries ..

> the user would like to share the results to another user.
> the user would like to bookmark / save the query.

IMO the open question is:

> So what is important here is the engine selection, and search parameters (language, time_range, safesearch, etc...)

Most users (like me) won't select or deselect engines or any other preference and will be happy with sharing or bookmarking searches like

- https://darmarit.org/searx/search?q=maceo%20parker&language=de&time_range=None&safesearch=2&categories=music
- https://darmarit.org/searx/search?q=%21google%20foo&language=de&time_range=None&safesearch=2&categories=none

To sum up: I think it is good enough what we have and as long there is no request from user side we should not increase the complexity of the implementation.

mrpaulblack:
Ok I will add the patch for the `simple_search.html` on top later today and merge :+1: 

[EDIT] or @MontyQI is faster, Thanks :D

dalf (approved):
Thank you!

PR #1010, opened 2022-03-25
[fix] pylint issue in py3.10

return42:
searxng_extra/update/update_firefox_version.py:16:0: W0402:
Uses of a deprecated module 'distutils.version' (deprecated-module)

[1] https://github.com/searxng/searxng/pull/1007


return42:
Pew .. CI again refuses to work and did not start the needed tests ..

PR #1013, opened 2022-03-25
[mod] add test.pybabel to cover issues in translation tasks

return42:
We need to cover issue related to pybabel/translation tasks.  By example there
was an issue [1] (fixed [2]) when upgrading jinja2 from 3.0.3 to 3.1.0 [3] that
has not been covered by the CI tests.

- [1] https://github.com/searxng/searxng/runs/5688624325?check_suite_focus=true#step:6:348
- [2] https://github.com/searxng/searxng/pull/1011
- [3] https://github.com/searxng/searxng/pull/1008

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ search.checker.%: install
 	$(Q)./manage pyenv.cmd searx-checker -v "$(subst _, ,$(patsubst search.checker.%,%,$@))"
 
 PHONY += test ci.test test.shell
-ci.test: test.yamllint test.black test.pyright test.pylint test.unit test.robot test.rst
+ci.test: test.yamllint test.black test.pyright test.pylint test.unit test.robot test.rst test.pybabel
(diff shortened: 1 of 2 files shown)


return42:
FYI: just rebased .. I would like to merge this PR, see https://github.com/searxng/searxng/pull/1134#issuecomment-1106213950

dalf:
* `make test` takes already a very long time to run.
* `babel` is updated with dependabot

--> I suggest to add `test.pybabel`  to `test.ci` but not `test`.

return42:
> --> I suggest to add `test.pybabel`  to `test.ci` but not `test`.

good point / I removed `test.pybabel` from Makefile target `test`

https://github.com/searxng/searxng/runs/6139051410?check_suite_focus=true#step:7:147

dalf (approved):
👍

PR #1022, opened 2022-03-27
Add calculator on results page

MontyQI:
## What does this PR do?
Adds calculator if query is math problem.

For now it looks like that:
image

It's basic feature in many search engines and it's useful.

`make run`
search: 2+2

## Notes
It won't work without js.


mrpaulblack:
Thank you for bringing this up in general. I really like this and I think we should in general implement instant answers like this more frequent. The only problem I see is that we are maybe not ready yet to shift our work to features like this yet, since there is still mopre fundamental things to patch in the theme, docker and so on... And I want this to be done properly.

Here are is what I was thinking about. In the future it would make sense to have more and more instant answers. We would need like an API or something so facilitate those like with the engine files and the plugins for example. I also think that instant answers should not b e build into the themes, but be a bit more modular, otherwise the CSS and JS could get really big if we have a lot of them... So here are basically the challange I am seeing with instant answers in general:

* We need some general regex to match the instant answers, if we have a lot of them this could (?) slow down search times maybe?
* I think we also want multiple different instant answers so I think they should ship with their own CSS, JS and HTML that can be just included with like an iframe or server side in the theme (this would make them independent, on the other hand we could maybe rely on for example the colors from `definition.less` to make this work with the current theme) -> there is the question of how maintainable is this? if we update the theme how do we update these instant answers?
* I also think it the majority of tose will only work with JS enabled and that is AFAIK ok since its just an extra feature and not breaking result page or anything...


So yeah all in all cool idea, but I think its to early to start impl. this? ping @dalf what do you think about this? Looking at our roadmap I dont think its the right time to impl. this IMO...

return42:
> So yeah all in all cool idea, but I think its to early to start impl. this? ping @dalf what do you think about this? Looking at our roadmap I dont think its the right time to impl. this IMO...

I fully agree with .. I also want to have features like a calculator or a dictionary but SearXNG's boilerplate isn't at that point we can implement such stuff future proof and isolated on.

dalf:
I remember that a fork has implemented multiple instant answer like this one using the answers:

https://github.com/searxng/searxng/blob/dec04c0ed640caac3b901787aca2a14452734fe7/searx/templates/simple/results.html#L29

Instead:
```<span>{{ answer.answer | safe }}</span> ```

Then each answer calls Jinja2 to render the HTML, CSS and all.

I'm not saying we should do that, rather is there a way to make room for this contribution (asked multiple times)? I need to think about it.

MontyQI:
I also wondered if it could slow the results. I agree with so I'm closing this pr.

return42:
For me it is OK to leave it open as DRAFT .. @MontyQI should I reopen the PR as DRAFT?

dalf:
For reference: https://gitlab.e.foundation/e/infra/spot/-/blob/master/searx/plugins/calculator.py

mrpaulblack:
From matrix about having translation instant answers:

`
has anyone thought about implementing a machine translation support into SearXNG? I'm thinking about e.g. Apertium developed by the University of Alacante. It has tens of language pairs and has a light python server to handle requests. I was wondering if it was possible to add a function somewhat similar to Google's when you search something like "translate word/phrase". List of pairs: https://wiki.apertium.org/wiki/Main_Page The server: https://wiki.apertium.org/wiki/Apertium-apy
`

SecularSteve:
> From matrix about having translation instant answers:
> 
> `has anyone thought about implementing a machine translation support into SearXNG? I'm thinking about e.g. Apertium developed by the University of Alacante. It has tens of language pairs and has a light python server to handle requests. I was wondering if it was possible to add a function somewhat similar to Google's when you search something like "translate word/phrase". List of pairs: https://wiki.apertium.org/wiki/Main_Page The server: https://wiki.apertium.org/wiki/Apertium-apy`

There's also LingvaTranslate if you want a tracker free Google Translate instance:
https://lingva.ml/
https://github.com/thedaviddelta/lingva-translate

Speaking of which, is this calculator idea still active?

WreckingBANG2:
Is this still active?

return42:
@WreckingBANG2: I closed your issue as "duplicate" since it is not good to spread one topic over several issues --> lets collect all upcoming comments here. To answer your question: Its not active right now, but we might come back to this later / see https://github.com/searxng/searxng/pull/1022#issuecomment-1080034213

sqwwqw5:
I relied on the Google search calculator a lot, it's one of the two reasons I still have Google installed as a windowless "PWA", the other reason is obviously to have it as a backup, this would be a very nice feature to have.

lee-b:
Would suggest that this just be a standard search engine plugin that returns the result for math queries, rather than a special tool on the search page.

upintheairsheep:
> > So yeah all in all cool idea, but I think its to early to start impl. this? ping @dalf what do you think about this? Looking at our roadmap I dont think its the right time to impl. this IMO...
> 
> I fully agree with .. I also want to have features like a calculator or a dictionary but SearXNG's boilerplate isn't at that point we can implement such stuff future proof and isolated on.

and it'd be nice to add some small HTML/JS Easter Eggs and possibly games

PR #1024, opened 2022-03-28
Fix answer box bug

### LINKED ISSUE #710: [theme] hash-plugin: sha512 values overwrite the answers box
Try search term `sha512 The quick brown fox`

grafik

MontyQI:
## What does this PR do?
Fixes answer box bux of too long words.
image
image

`make run`
## Related issues
Closes  #710


searx/static/themes/simple/src/less/style.less
```diff
@@ -540,6 +540,10 @@ article[data-vim-selected].category-social {
   h4 {
     display: none;
   }
+
+  span {
```
mrpaulblack, line 544: Can you please do instead the selector from dalf in the issue?:
```css
#answers span {
  overflow-wrap: anywhere;
}
```

I just want to make sure that this will only get applied to spans in the `#answer` box just in case...

mrpaulblack (changes requested):
Please fix that one small thing; Everything else LGTM :+1: (and great thinking with that `anywhere` option)

searx/static/themes/simple/src/less/style.less
mrpaulblack: nvm I did not read properly :facepalm:

mrpaulblack (approved):
LGTM :+1: thanks

PR #1026, opened 2022-03-29
Update simple theme

MontyQI:
## What does this PR do?

Changes colors, removes border in some places, updates autocomplete look.

Makes searxng better looking.

`make run`
search for something

## Todo:

Dark theme:
- [x] preferences page

Light theme:
- [x] cleanup in names

General:
- [x] Autocomplete (search_box border_radius)
- [x] Some mobile changes
- [x] General cleanup

## Screenshots
_Dark theme:_
image
image
image

_Light theme:_
image
image
image


return42:
@MontyQI sorry for being pedantic, but can you use more meaningful commit messages in this PR?

- https://docs.searxng.org/dev/contribution_guide.html#code
- https://wiki.openstack.org/wiki/GitCommitMessages

By example for [fb1bc2e](https://github.com/searxng/searxng/pull/1026/commits/fb1bc2e62aadf0e47cb16e02dbb4e54862fb0b48) I would prefer something like

    [mod] simple theme: Change colors, autocomplete, delete border in some places

MontyQI:
It's almost done, but still need to make autocomplete look good.
For now it has bug:
image
While it should look like that:
image

The problem is that autocomplete is below q, send_search and clear_search, so this doesn't work https://github.com/searxng/searxng/blob/3149038fa6c236086959abd044e22a0152df6eee/searx/static/themes/simple/src/less/autocomplete.less#L64

searx/static/themes/simple/src/less/autocomplete.less
```diff
@@ -51,15 +51,32 @@
...
     z-index: 100;
-    margin-top: 3.2rem;
+    margin-top: 3.1rem;
+    border-radius: 0 0 10px 10px;
 
     &:empty {
       display: none;
     }
+
+    & ~ #q {  // Doesn't work --> autocomplete must be above #q, #send_search and #clear_search
```
mrpaulblack: we are using the `autocomplte.js` lib, which is quite old, this is AFAIK not possible, but maybe @dalf knows more...

[EDIT] an alternative could be the qwant approach of just making the margin between search box and autocomplete bigger and not changing their border radius:
image

MontyQI:
The autocomplete now looks as @mrpaulblack suggested and send_search on hover has gradient:
image
image

Maybe I should change search box background in dark theme, because now it can be hardly visible.

mrpaulblack:
I will do the full review in a week, but this is just a quick comment. I am not a big fan of the gradient; Markus said this somewhere else, but we should keep the simple theme well simple. I am ok with the shadow on search-box and autocomplete since its the centerpiece of the site, but everything else should be flat and simple, so only bg colors and borders basically...

About the background color of the search box in dark mode. I think we should see accessibility features just as features that should be impl. by default. In ff dev tools in the accessibility tab, you can check that the contrast between search-box background and the overall background is high enough :+1:

return42 (approved):
Looks great / thanks a lot!

return42:
@mrpaulblack I approved this PR / from my side the PR is ready to merge .. if you do not veto, I would merge the PR.

---

FYI: I rebased PR to solve conflicts with current master.

mrpaulblack:
Please dont merge this PR yet. I will need more time to do my review; thanks.

mrpaulblack (approved):
LGTM :+1: 

Thank you for being patient with me and thanks for this great color rework of the simple theme :tada: 

(I made some slight changes and fixes and just force pushed them; hope thats ok...)

PR #1029, opened 2022-03-31
Fix Catalan language code for Bing search engines

jmontane:
"ca-CA" means Catalan from Canada. Bing just uses "ca" for Catalan searches.

## What does this PR do?

Replace "ca-CA" langague with "ca" for all Bing search engines

"ca-CA" means Catalan from Canada, it's wrong.  And it's not used by Bing. Bing uses "ca".


dalf:
Thank you for this PR.
However `searx/data/engines_languages.json` is generated automatically. Each month `searxng_extra/update/update_languages.py` updates this file (locally, you can call `make data.languages` to call the script). The script calls the function `_fetch_supported_languages` of each engine to get the values you see in `searx/data/engines_languages.json`

I've made the PR #1031 : 
* it update the function [_fetch_supported_languages](https://github.com/searxng/searxng/blob/d1334beb4fb16beec298f894d75e94b241499e06/searx/engines/bing.py#L129).
* then I've called `make data.languages` to update : 
  * `searx/data/engines_languages.json`
  * `searx/languages.py`

Can you check that #1031 actually fix the issue ?

jmontane:
Hi, @dalf
Thank you for your quick response. Unfortunately, I'm not a developer guy, I just make simple changes. So, I can't check #1031, but if it removes ca-CA locale, then it's an improvement and fine for me. So, I close this PR

Another different but related issue, is using Spanish flag emoji for Catalan language. Yes, I know it represents "ca-ES" locale, but I wonder if "ca" locale (without country part) can be used, and then, use earth emoji instead. Thanks.

dalf:
@jmontane,  I've merged #1031  : it should fix your issue.

About the flag: point taken. For information see https://github.com/searxng/searxng/pull/967 (there is a discussion on matrix about this subject too)

return42:
> Another different but related issue, is using Spanish flag emoji for Catalan language. Yes, I know it represents "ca-ES" locale, but I wonder if "ca" locale (without country part) can be used, 

The problem is, that only `ca-ES` is in the https://github.com/searxng/searxng/blob/master/searx/languages.py

This list is generated by a scoring and since ddg and qwant still use `ca-ES` it scores over bing's `ca`.

To me it seems ddg and qwant do not use `ca` (compared to bing where @dalf fixed an issue on SearXNG site).

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/engines/duckduckgo.py#L181-L182

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/engines/qwant.py#L250-L254

dalf:
The values from 3 engines could be grouped this way (by row):
| bing           | duckduckgo | qwant |
|----------------|------------|-------|
| ca             | ca-CT      |       |
| ca-es-valencia | ca-ES      | ca-ES |

First row:
* catalan result.
* catalan result in Spain.

Only a manual curation can merge `ca` and `ca-CT`  **_if_** it makes sense(I guess `CT` stands for Catalonia). Perhaps there is the same question for other (country, region).

It's the same old issues:
* there is a mapping between the language + region selection to the engine. Sometimes the region is ignored, sometimes the region is enforced. How to display that?
* the autocompletion still show the full list of language is the user types `!qwant :` (this can be fixed).

jmontane:
`CT` is an unassigned country code. I don't know why DDG uses such code.

The issue here about Spanish flag and Catalan language is: flags represent territories, not languages. So, seeing Spanish flag, when language name is written only, is weird. And, from a Catalan point of view, it's wrong (yep, it's a political matter).

return42:
> I don't know why DDG uses such code.

SearXNG is a meta-search engine and has to use whatever the origin search engine is going to use.

This is the engine-side, on the other hand we have the user side, where the user wants to select a language and/or a region, this selection is used for a all the engines that are active.

Whatever the user will select, each engine implementation will do its best to narrow to what the user has selected.  But in the UI it won't help when SearXNG offer a menu with all languages & region codes from all engines, that is why there is a scoring to show only the most common ..

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searxng_extra/update/update_languages.py#L182-L184

To be clear: **flags are only shown if there is a region code**

@dalf I think the issue is the half of the part what @jmontane says:

> So, seeing Spanish flag, when language name is written only, is weird.

I miss the region in ..

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/languages.py#L9

In other items with a region, the region is named, e.g in:

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/languages.py#L13-L15

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/languages.py#L35-L38

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/languages.py#L18-L24

**But this issue is not limited to catalan** .. if we look at the languages.py we will find  lot of items with region code, but without a entry in the region name, by example:

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/languages.py#L59-L62


Can't say if it is due to babel or due to our `/searxng_extra/update/update_languages.py` script.

return42:
We can discuss further in https://github.com/searxng/searxng/pull/1051

PR #1030, opened 2022-03-31
(feat) add jisho.org

### LINKED ISSUE #1016: Jisho.org (Japanese <-> English dictionary)
**Working URL to the engine**
https://jisho.org

It's the go-to online dictionary for Japanese <->English. It's basically what https://www.naver.com is for Korean, and there's already an engine entry for Naver.

**Features of this engine**
I'm not really sure what I should put here... Uhm, I guess it's fast? And it's a nicely and matter-of-factly designed website? And it's indispensable for anyone having to do anything with Japanese?

**How can Searx fetch the information from this engine?**
Again, not really sure what kind of technical data need to go here because I don't really know how Searx talks to its engines internally, but I suppose this could be a starting point: https://jisho.org/docs

General; or maybe Science, seeing as it's of an academic nature, but all the other dictionaries are in General, so I see no reason to break the tradition.

**Additional Context**
I'm cross-posting this from searx/searx#3194. I'm not entirely sure how SearX and SearXNG inter-relate nowadays, but I can see that many issues are being cross-linked, fixed and cherry-picked between the two projects; so I thought I'd ask this in both trackers and see where that gets me :)

austinhuang0131:
## What does this PR do?

Add https://jisho.org as an engine. Note that this only supports searching words (either English or Japanese input).

example

A longer infobox:

infobox

¯\\_(ツ)\_/¯

`make run test`

## Related issues

Closes #1016

### DIFF
--- a/searx/engines/jisho.py
+++ b/searx/engines/jisho.py
@@ -3,7 +3,6 @@
 Jisho (the Japanese-English dictionary)
 """
 
-import json
 from urllib.parse import urlencode, urljoin
 
 # about
--- a/searx/engines/jisho.py
+++ b/searx/engines/jisho.py
@@ -17,7 +17,6 @@ about = {
 }
 
 categories = ['dictionaries']
-engine_type = 'online_dictionary'
 paging = False
 
 URL = 'https://jisho.org'
@@ -34,19 +33,19 @@ def request(query, params):
 
 def response(resp):
     results = []
-    infoboxed = False
+    first_result = True
 
     search_results = resp.json()
-    pages = search_results.get('data', [])
 
-    for page in pages:
+    for page in search_results.get('data', []):
         # Entries that are purely from Wikipedia are excluded.
-        if page['senses'][0]['parts_of_speech'] != [] and page['senses'][0]['parts_of_speech'][0] == 'Wikipedia definition':
+        parts_of_speech = page.get('senses') and page['senses'][0].get('parts_of_speech')
+        if parts_of_speech and parts_of_speech[0] == 'Wikipedia definition':
             pass
+
         # Process alternative forms
-        japanese = page['japanese']
         alt_forms = []
-        for title_raw in japanese:
+        for title_raw in page['japanese']:
             if 'word' not in title_raw:
                 alt_forms.append(title_raw['reading'])
             else:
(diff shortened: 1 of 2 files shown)


searx/engines/jisho.py
```diff
@@ -0,0 +1,125 @@
...
+from urllib.parse import urlencode, urljoin
+
+# about
+about = {
+    "website": 'https://jisho.org',
+    "wikidata_id": 'Q24568389',
+    "official_api_documentation": "https://jisho.org/forum/54fefc1f6e73340b1f160000-is-there-any-kind-of-search-api",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
```
dalf, line 15: Can you add:
`    "language": 'jp',`

searx/engines/jisho.py
```diff
@@ -0,0 +1,125 @@
...
+    params['url'] = SEARCH_URL.format(query=query)
+    logger.debug(f"query_url --> {params['url']}")
+    return params
+
+
+def response(resp):
+    results = []
+    infoboxed = False
+
+    search_results = json.loads(resp.text)
```
dalf: It can simplified:
`search_results = resp.json()`

searx/engines/jisho.py
```diff
@@ -0,0 +1,125 @@
...
+            for pos, engdef, extra in definitions:
+                if pos == 'Wikipedia definition':
+                    infobox_content += '</ul><small>Wikipedia, CC BY-SA 3.0.</small><ul>'
+                if pos == '':
+                    infobox_content += f"<li>{engdef}"
+                else:
+                    infobox_content += f"<li><i>{pos}</i>: {engdef}"
+                if extra != '':
+                    infobox_content += f" ({extra})"
+                infobox_content += '</li>'
```
dalf: Does it always close an `<i>` ?
What about `infobox_content += f" ({extra})"` and `infobox_content += '</ul><small>Wikipedia, CC BY-SA 3.0.</small><ul>'` ?

searx/engines/jisho.py
```diff
@@ -0,0 +1,125 @@
...
+                    ', '.join(defn_raw['parts_of_speech']),
+                    '; '.join(defn_raw['english_definitions']),
+                    extra
+                ))
+            content = ''
+            infobox_content = '''
+                <small><a href="https://www.edrdg.org/wiki/index.php/JMdict-EDICT_Dictionary_Project">JMdict</a> 
+                and <a href="https://www.edrdg.org/enamdict/enamdict_doc.html">JMnedict</a> 
+                by <a href="https://www.edrdg.org/edrdg/licence.html">EDRDG</a>, CC BY-SA 3.0.</small><ul>
+                '''
```
dalf: This is the topic of another issue and PR : the infobox template is missing a field about the license:
https://github.com/searxng/searxng/blob/master/searx/templates/simple/infobox.html

searx/engines/jisho.py
```diff
@@ -0,0 +1,125 @@
...
+            for pos, engdef, extra in definitions:
```
dalf: Would it make sense to use the attributes of the infoboxes: https://github.com/searxng/searxng/blob/d1334beb4fb16beec298f894d75e94b241499e06/searx/templates/simple/infobox.html#L7-L16 ?

dalf (commented):
Some comments here and there

searx/engines/jisho.py
austinhuang0131: You mean `<li>`

> What about `infobox_content += f" ({extra})"`

There always exist `engdef`. Plus `<li>` is always appended since I have both an if and an else

> and `infobox_content += '</ul><small>Wikipedia, CC BY-SA 3.0.</small><ul>'`

At the very top I exclude any entries whose first entry is Wikipedia. My understanding is that the dictionary entries are always before Wikipedia entries.

searx/engines/jisho.py
austinhuang0131: This should be documented imo

searx/engines/jisho.py
austinhuang0131: should it be `ja`?

searx/engines/jisho.py
austinhuang0131: Is it the tables? If so I don't see how this fits into a table...

I generally follow wordnik (`!def`) for the format

return42:
Sorry I haven't had time to test in deep yet, but a first try results in a IndexError:

```
  File "SearXNG/searx/search/processors/online.py", line 143, in search
  File "SearXNG/searx/search/processors/online.py", line 131, in _search_basic
  File "SearXNG/searx/engines/jisho.py", line 43, in response
    if page['senses'][0]['parts_of_speech'][0] != 'Wikipedia definition':
IndexError: list index out of range
```

-----

Should it be

```
engine_type = 'online_dictionary'
```

searx/engines/jisho.py
dalf: It is not a table, rather a list of definition: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

After a quick update, it looks like this:
image.

After a second thought:
* it requires to update worknik too.
* `Noun` should not be inside [<dt> element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt) 

So forget about that.

dalf:
> Should it be `engine_type = 'online_dictionary'` 

yes and no.
* yes, because it would be nice to `en-ja house` to pick the right engine.
* no, because both `!jisho house` or `!jisho 家` and there is no syntax to say "auto detect" to "English" for example (no user query syntax and no implementation).
* no because wordnik and other dictionnaries don't use the `online_dictionnary`.
--> there is a global task about the dictionaries.

@austinhuang0131 > I have refactored the code in https://github.com/dalf/searxng/commit/0e2754e7c6696a6e6fe88d4b800114195c507b36 . Can you have a look?

austinhuang0131:
That's good, you can use it as-is.

dalf:
Merged. Thank you!

Nocifer:
Wow, that's what I'd call a very prompt response. Thank you all for the effort!

PR #1047, opened 2022-04-02
Add a redis library to generalize DB functions we need in SearXNG.

return42:
## What does this PR do?

Add a redis library to generalize DB functions we need in SearXNG.

- single point of definition
- we do not need multiple e.g. counter functions that stores key/values in the DB (this PR is inspired by https://github.com/searxng/searxng/pull/998 and https://github.com/searxng/searxng/pull/618)
- to not pollute the redis DB with keys, we preifx all keys from the SearXNG schema with `SearXNG_` (there is a searx.redislib.purge_by_prefix() function to drop the SearXNG data)

In the second commit, the limiter plugin is modified to use `searx.redislib.incr_counter` .. the limiter is a good test case, further compile the documentation and have a look at `src/searx.redislib.html`.

## Author's checklist

I made several tests .. also with the limiter plugin .. @dalf I am not a lua expert but I think the lua script from the (old) limiter plugin is compareable / replaceable by the `searx.redislib.incr_counter`. May I have overseen something / could you place have a look at the second commit .. thanks!

### DIFF
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -13,11 +13,11 @@ Enable the plugin in ``settings.yml``:
 - ``redis.url: ...`` check the value, see :ref:`settings redis`
 """
 
-import hmac
 import re
 from flask import request
 
 from searx.shared import redisdb
+from searx.redislib import incr_sliding_window
 
 name = "Request limiter"
 description = "Limit the number of request"
@@ -36,8 +36,9 @@ re_bot = re.compile(
 )
 
 
-def is_accepted_request(inc_get_counter) -> bool:
+def is_accepted_request() -> bool:
     # pylint: disable=too-many-return-statements
+    redis_client = redisdb.client()
     user_agent = request.headers.get('User-Agent', '')
     x_forwarded_for = request.headers.get('X-Forwarded-For', '')
 
@@ -47,70 +48,45 @@ def is_accepted_request(inc_get_counter) -> bool:
         return True
 
     if request.path == '/search':
-        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
-        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
+        c_burst = incr_sliding_window(redis_client, 'IP limit, burst' + x_forwarded_for, 20)
+        c_10min = incr_sliding_window(redis_client, 'IP limit, 10 minutes' + x_forwarded_for, 600)
         if c_burst > 15 or c_10min > 150:
+            logger.debug("to many request")  # pylint: disable=undefined-variable
             return False
 
         if re_bot.match(user_agent):
+            logger.debug("detected bot")  # pylint: disable=undefined-variable
             return False
 
         if len(request.headers.get('Accept-Language', '').strip()) == '':
+            logger.debug("missing Accept-Language")  # pylint: disable=undefined-variable
             return False
 
         if request.headers.get('Connection') == 'close':
+            logger.debug("got Connection=close")  # pylint: disable=undefined-variable
             return False
 
         accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
         if 'gzip' not in accept_encoding_list or 'deflate' not in accept_encoding_list:
+            logger.debug("suspicious Accept-Encoding")  # pylint: disable=undefined-variable
             return False
 
         if 'text/html' not in request.accept_mimetypes:
+            logger.debug("Accept-Encoding misses text/html")  # pylint: disable=undefined-variable
             return False
 
         if request.args.get('format', 'html') != 'html':
-            c = inc_get_counter(interval=3600, keys=[b'API limit', x_forwarded_for])
+            c = incr_sliding_window(redis_client, 'API limit' + x_forwarded_for, 3600)
             if c > 4:
+                logger.debug("API limit exceeded")  # pylint: disable=undefined-variable
                 return False
     return True
 
 
-def create_inc_get_counter(redis_client, secret_key_bytes):
-    lua_script = """
-    local slidingWindow = KEYS[1]
-    local key = KEYS[2]
-    local now = tonumber(redis.call('TIME')[1])
-    local id = redis.call('INCR', 'counter')
-    if (id > 2^46)
-    then
-        redis.call('SET', 'count', 0)
-    end
-    redis.call('ZREMRANGEBYSCORE', key, 0, now - slidingWindow)
-    redis.call('ZADD', key, now, id)
-    local result = redis.call('ZCOUNT', key, 0, now+1)
-    redis.call('EXPIRE', key, slidingWindow)
-    return result
-    """
-    script_sha = redis_client.script_load(lua_script)
-
-    def inc_get_counter(interval, keys):
-        m = hmac.new(secret_key_bytes, digestmod='sha256')
-        for k in keys:
-            m.update(bytes(str(k), encoding='utf-8') or b'')
-            m.update(b"\0")
-        key = m.digest()
-        return redis_client.evalsha(script_sha, 2, interval, key)
-
-    return inc_get_counter
-
-
-def create_pre_request(get_aggregation_count):
-    def pre_request():
-        if not is_accepted_request(get_aggregation_count):
-            return '', 429
-        return None
-
-    return pre_request
+def pre_request():
+    if not is_accepted_request():
+        return '', 429
+    return None
 
 
 def init(app, settings):
@@ -122,8 +98,5 @@ def init(app, settings):
         logger.error("init limiter DB failed!!!")  # pylint: disable=undefined-variable
         return False
 
-    redis_client = redisdb.client()
-    secret_key_bytes = bytes(settings['server']['secret_key'], encoding='utf-8')
-    inc_get_counter = create_inc_get_counter(redis_client, secret_key_bytes)
-    app.before_request(create_pre_request(inc_get_counter))
+    app.before_request(pre_request)
     return True
--- a/searx/redislib.py
+++ b/searx/redislib.py
@@ -1,0 +1,213 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""A collection of convenient functions and redis/lua scripts.
+
+This code was partial inspired by the `Bullet-Proofing Lua Scripts in RedisPy`_
+article.
+
+.. _Bullet-Proofing Lua Scripts in RedisPy:
+   https://redis.com/blog/bullet-proofing-lua-scripts-in-redispy/
+
+"""
+
+import hmac
+
+from searx import get_setting
+
+PURGE_BY_PREFIX = """
+local prefix = tostring(ARGV[1])
+for i, name in ipairs(redis.call('KEYS', prefix .. '*')) do
+    redis.call('EXPIRE', name, 0)
+end
+"""
+
+
+def purge_by_prefix(client, prefix: str = "SearXNG_"):
+    """Purge all keys with ``prefix`` from database.
+
+    Queries all keys in the database by the given prefix and set expire time to
+    zero.  The default prefix will drop all keys which has been set by SearXNG
+    (drops SearXNG schema entirely from database).
+
+    The implementation is the lua script from string :py:obj:`PURGE_BY_PREFIX`.
+    The lua script uses EXPIRE_ instead of DEL_: if there are a lot keys to
+    delete and/or their values are big, `DEL` could take more time and blocks
+    the command loop while `EXPIRE` turns back immediate.
+
+    :param prefix: prefix of the key to delete (default: ``SearXNG_``)
+    :type name: str
+
+    .. _EXPIRE: https://redis.io/commands/expire/
+    .. _DEL: https://redis.io/commands/del/
+
+    """
+    script = client.register_script(PURGE_BY_PREFIX)
+    script(args=[prefix])
+
+
+def secret_hash(name: str):
+    """Creates a hash of the ``name``.
+
+    Combines argument ``name`` with the ``secret_key`` from :ref:`settings
+    server`.  This function can be used to get a more anonymised name of a Redis
+    KEY.
+
+    :param name: the name to create a secret hash for
+    :type name: str
+    """
+    m = hmac.new(bytes(name, encoding='utf-8'), digestmod='sha256')
+    m.update(bytes(get_setting('server.secret_key'), encoding='utf-8'))
+    return m.hexdigest()
+
+
+INCR_COUNTER = """
+local limit = tonumber(ARGV[1])
+local expire = tonumber(ARGV[2])
+local c_name = KEYS[1]
+
+local c = redis.call('GET', c_name)
+
+if not c then
+    c = redis.call('INCR', c_name)
+    if expire > 0 then
+        redis.call('EXPIRE', c_name, expire)
+    end
+else
+    c = tonumber(c)
+    if limit == 0 or c < limit then
+       c = redis.call('INCR', c_name)
+    end
+end
+return c
+"""
+
+
+def incr_counter(client, name: str, limit: int = 0, expire: int = 0):
+    """Increment a counter and return the new value.
+
+    If counter with redis key ``SearXNG_counter_<name>`` does not exists it is
+    created with initial value 1 returned.  The replacement ``<name>`` is a
+    *secret hash* of the value from argument ``name`` (see
+    :py:func:`secret_hash`).
+
+    The implementation of the redis counter is the lua script from string
+    :py:obj:`INCR_COUNTER`.
+
+    :param name: name of the counter
+    :type name: str
+
+    :param expire: live-time of the counter in seconds (default ``None`` means
+      infinite).
+    :type expire: int / see EXPIRE_
+
+    :param limit: limit where the counter stops to increment (default ``None``)
+    :type limit: int / limit is 2^64 see INCR_
+
+    :return: value of the incremented counter
+    :type return: int
+
+    .. _EXPIRE: https://redis.io/commands/expire/
+    .. _INCR: https://redis.io/commands/incr/
+
+    A simple demo of a counter with expire time and limit::
+
+      >>> for i in range(6):
+      ...   i, incr_counter(client, "foo", 3, 5) # max 3, duration 5 sec
+      ...   time.sleep(1) # from the third call on max has been reached
+      ...
+      (0, 1)
+      (1, 2)
+      (2, 3)
+      (3, 3)
+      (4, 3)
+      (5, 1)
+
+    """
+    script = client.register_script(INCR_COUNTER)
+    name = "SearXNG_counter_" + secret_hash(name)
+    c = script(args=[limit, expire], keys=[name])
+    return c
+
+
+def drop_counter(client, name):
+    """Drop counter with redis key ``SearXNG_counter_<name>``
+
+    The replacement ``<name>`` is a *secret hash* of the value from argument
+    ``name`` (see :py:func:`incr_counter` and :py:func:`incr_sliding_window`).
+    """
+    name = "SearXNG_counter_" + secret_hash(name)
+    client.delete(name)
+
+
+INCR_SLIDING_WINDOW = """
+local expire = tonumber(ARGV[1])
+local name = KEYS[1]
+local current_time = redis.call('TIME')
+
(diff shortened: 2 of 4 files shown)


return42:
@MarcAbonce in this PR I implemented a counter that might also be useful in your _"rate limiting per engine"_ .. if you have time for, could you please review this PR .. thanks!

dalf:
I note a change of the algorithm: 
* `create_inc_get_counter` in limiter.py implements a sliding window counter.
* `INCR_COUNTER` in redislib.py is a counter that expired: one request every (limit - 1) never reset the counter.

dalf:
> Add a redis library to generalize DB functions we need in SearXNG

From this implementation, 
* does it make redis mandatory?
* how to implement the following function for example?
https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/search/processors/abstract.py#L40-L50
?

---

If redis is mandatory, a lua implemented won't use any lock.
* Pros: it's going to be fast
* Cons: it adds another language

---

A full python version require a distributed lock:
```python
try:
    with redis_client.lock('something', blocking_timeout=5) as lock:
      # calls the redis_client methods
except LockError:
    # the lock wasn't acquired
```

It makes redis also mandatory.

---

https://github.com/yahoo/redislite 
> Redislite is a self contained Python interface to the Redis key-value store.

Basically it compiles and embed an Redis server. Unfortunately, the project lack of maintenance.

---

<details>

<summary>I've tried to write an implementation of "searx.shared.shared_abstract.SharedDict" with two new methods for the counters (whatever the implementation)</summary>

The implementation hides the Redis client (initialization and calls).
SharedRedis provides an API, but except the constructor the implementation could use anything: the intent is to be able to write a pure Python version of this class without call to Redis. The requirement is to provide enough method in this class to implement the SuspendedStatus class, etc...

```python
# lint: pylint

import logging
import redis
import hmac
from searx import get_setting, settings
from . import shared_abstract
from typing import Optional


logger = logging.getLogger('searx.shared.redis')

shared_instance: Optional[shared_abstract.SharedDict] = None


class SharedRedis(shared_abstract.SharedDict):

    __slots__ = ('_client', '_secret_key_bytes', '_key_prefix_counter_sw_hmac', '_key_prefix_counter_sw', '_script_sha')

    LUA_SW_SCRIPT = """
local slidingWindow = KEYS[1]
local key = KEYS[2]
local now = tonumber(redis.call('TIME')[1])
local id = redis.call('INCR', 'counter')
if (id > 2^46)
then
    redis.call('SET', 'count', 0)
end
redis.call('ZREMRANGEBYSCORE', key, 0, now - slidingWindow)
redis.call('ZADD', key, now, id)
local result = redis.call('ZCOUNT', key, 0, now+1)
redis.call('EXPIRE', key, slidingWindow)
return result
"""

    def __init__(self, client, secret_key, key_prefix = b"SearXNG_"):
        self._client = client
        self._secret_key_bytes = bytes(secret_key, encoding='utf-8')
        self._key_prefix_counter_sw_hmac = key_prefix + b'_counter_sw_hmac_'
        self._key_prefix_counter_sw = key_prefix + b'_counter_sw_'
        self._script_sha = self._client.script_load(SharedRedis.LUA_SW_SCRIPT)

    def get_int(self, key: str) -> Optional[int]:
        encoded_value = self._client.get(key)
        if encoded_value is None:
            return encoded_value
        return int.from_bytes(encoded_value, 'big')

    def set_int(self, key: str, value: int):
        encoded_value = value.to_bytes(4, 'big')
        self._client.set(key, encoded_value)

    def get_str(self, key: str) -> Optional[str]:
        return self._client.get(key)

    def set_str(self, key: str, value: str):
        encoded_value = value.encode('utf-8')
        self._client.set(key, encoded_value)

    def inc_get_counter_sw_hmac(self, interval, *keys) -> int:
        m = hmac.new(self._secret_key_bytes, digestmod='sha256')
        for k in keys:
            m.update(bytes(str(k), encoding='utf-8') or b'')
            m.update(b"\0")
        key = self._key_prefix_counter_sw_hmac + m.digest()
        return self._client.evalsha(self._script_sha, 2, interval, key)

    def inc_get_counter_sw(self, interval, *keys) -> int:
        return self._client.evalsha(self._script_sha, 2, interval, self._key_prefix_counter_sw  + b"\0".join(keys))


def init():
    global shared_instance
    try:
        client = redis.Redis.from_url(get_setting('redis.url'))
        logger.info("connected redis DB --> %s", client.acl_whoami())
        secret_key_bytes = bytes(settings['server']['secret_key'], encoding='utf-8')
        shared_instance = SharedRedis(client, secret_key_bytes)
        return True
    except redis.exceptions.ConnectionError as exc:
        logger.error("can't connet redis DB ...")
        logger.error("  %s", exc)
    return False
```

</details>

----

In all these implementation the only way I see to make redis optional is to have two implementations.

return42:
>  sliding window counter.

Ah, OK .. I have to read about this algorithm

> does it make redis mandatory?

No

> how to implement the following function for example?
> ...
> If redis is mandatory, a lua implemented won't use any lock.

Without a deep look at: ... I think lua is not the solution but may be we can use the EXPIRE timer from redis and store. 

```python
def suspend(self, suspended_time, suspend_reason):
    ....
    # a lock for this python object
    redis_client.setex(self.id, suspended_time, value="X")
```

and the property ...

https://github.com/searxng/searxng/blob/765504ceca4ce9e69b02b213b2f9a3036915055d/searx/search/processors/abstract.py#L35-L37 

needs to become something like

```python
def is_suspended(self): 
    return bool(redis_client.get(self.id))
```

> a lua implemented won't use any lock.
>
>    Pros: it's going to be fast

OT: don't lets mix lua & redis .. lua is running on the redis DB (and will lock the redis DB).

But here in the python process: can't say if it is faster .. a redis DB communication will take its time and how long it would be depends on the redis DB and the way to the redis DB.

(we just moved the problem to another layer, but may be it can help here to not to lock the python threading model .. but I do not know).

> It makes redis also mandatory.

I'm not really happy about making redis mandatory .. I think it should be optional.

May we can start in the CI to make it mandatory for some CI tests.

> Redislite is a self contained Python interface to the Redis key-value store.
> Basically it compiles and embed an Redis server. Unfortunately, the project lack of maintenance.

Looks interesting from the point of "deploy redis along with in SearXNG" .. but I am not really a fan of using not so common solutions .. I think we should stay with the origin: most often the origin solution scales better and is more error proof.

> I've tried to write an implementation of "searx.shared.shared_abstract.SharedDict" with two new methods for the counter

I think we should discuss this in another PR .. for me redis & lua is new and I have to do it step by step .. first I want to understand the sliding window algorithm, then I will fix this PR to have one implementation of this algorithm here in the redislib.py .. in parallel I can have a look about your ShareDict solution.

searx/plugins/limiter.py
```diff
@@ -47,70 +48,44 @@ def is_accepted_request(inc_get_counter) -> bool:
...
+            c = incr_sliding_window(redis_client, 'API limit' + x_forwarded_for, 3600)
             if c > 4:
+                logger.debug("API limit exceeded")  # pylint: disable=undefined-variable
                 return False
     return True
 
-
-def create_inc_get_counter(redis_client, secret_key_bytes):
-    lua_script = """
-    local slidingWindow = KEYS[1]
```
return42, line 80: FYI: slidingWindow is an argument not a KEY.

searx/plugins/limiter.py
```diff
@@ -47,70 +48,44 @@ def is_accepted_request(inc_get_counter) -> bool:
...
-    local key = KEYS[2]
-    local now = tonumber(redis.call('TIME')[1])
-    local id = redis.call('INCR', 'counter')
-    if (id > 2^46)
-    then
-        redis.call('SET', 'count', 0)
-    end
```
return42, line 87: Instead of `id` I use the concatenation of `current_time[1] .. current_time[2]`

return42:
> * `create_inc_get_counter` in limiter.py implements a sliding window counter.

@dalf I implemented a sliding window counter in `redislib.py`. The implementation is a little bit simpler, there is no need for a `id` counter when we use the string of the current time as value .. see comment above.

return42:
ping @dalf

return42:
FYI: rebased to solve conflicts with curretn master / no functional change since https://github.com/searxng/searxng/pull/1047#issuecomment-1090871670

return42:
FYI: rebased to solve conflicts with current master / no functional change since https://github.com/searxng/searxng/pull/1047#issuecomment-1090871670

return42:
FYI: rebased to solve conflicts with current master / no functional change since https://github.com/searxng/searxng/pull/1047#issuecomment-1090871670

return42:
@dalf can we please review/merge this PR? .. If you have a doubt in the concept of this PR in general please inform me.

searx/redislib.py
```diff
@@ -0,0 +1,201 @@
...
+      2
+      3
+      3
+      3
+      >>> time.sleep(3)  # wait until expire
+      >>> incr_sliding_window(client, "foo", 3)
+      1
+
+    """
+    script = client.register_script(INCR_SLIDING_WINDOW)
```
dalf: The current pattern is:
* to call `redis_client.script_load` once when the application starts
* to call `return redis_client.evalsha(script_sha, ...)` afterward, so the script is not uploaded to Redis each time.

---

Here the function creates a `Script` instance for each call.
Why not defined the script as a global variable?

For reference the `Script` class (from https://redis.readthedocs.io/en/v4.3.1/_modules/redis/commands/core.html?highlight=register_script )

```python

class Script:
    """
    An executable Lua script object returned by ``register_script``
    """

    def __init__(self, registered_client, script):
        self.registered_client = registered_client
        self.script = script
        # Precalculate and store the SHA1 hex digest of the script.

        if isinstance(script, str):
            # We need the encoding from the client in order to generate an
            # accurate byte representation of the script
            encoder = self.get_encoder()
            script = encoder.encode(script)
        self.sha = hashlib.sha1(script).hexdigest()
```

searx/redislib.py
```diff
@@ -0,0 +1,201 @@
...
+    client.delete(name)
+
+
+INCR_SLIDING_WINDOW = """
+local expire = tonumber(ARGV[1])
+local name = KEYS[1]
+local current_time = redis.call('TIME')
+
+redis.call('ZREMRANGEBYSCORE', name, 0, current_time[1] - expire)
+redis.call('ZADD', name, current_time[1], current_time[1] .. current_time[2])
```
dalf, line 176: The purpose of the ID is to make sure to count every request even if there are at the exact same timestamp.

searx/plugins/limiter.py
```diff
@@ -47,70 +48,45 @@ def is_accepted_request(inc_get_counter) -> bool:
         return True
 
     if request.path == '/search':
-        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
-        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
+        c_burst = incr_sliding_window(redis_client, 'IP limit, burst' + x_forwarded_for, 20)
+        c_10min = incr_sliding_window(redis_client, 'IP limit, 10 minutes' + x_forwarded_for, 600)
```
dalf, line 52: nitpicking: since `name` is not a Redis key anyway, it can be tuple so we avoid a string concat.

searx/redislib.py
```diff
@@ -0,0 +1,201 @@
...
+# https://redis.com/blog/bullet-proofing-lua-scripts-in-redispy/
+
+import hmac
+
+from searx import get_setting
+
+PURGE_BY_PREFIX = """
+local prefix = tostring(ARGV[1])
+for i, name in ipairs(redis.call('KEYS', prefix .. '*')) do
+    redis.call('EXPIRE', name, 0)
```
dalf, line 48: Why not `redis.call('DEL', name)` ?

searx/redislib.py
```diff
@@ -0,0 +1,201 @@
...
+      ...
+      (0, 1)
+      (1, 2)
+      (2, 3)
+      (3, 3)
+      (4, 3)
+      (5, 1)
+
+    """
+    script = client.register_script(INCR_COUNTER)
```
dalf: See the comment about `incr_sliding_window`.

searx/plugins/limiter.py
```diff
@@ -36,8 +36,9 @@
 )
 
 
-def is_accepted_request(inc_get_counter) -> bool:
+def is_accepted_request() -> bool:
     # pylint: disable=too-many-return-statements
+    redis_client = redisdb.client()
```
dalf, line 41: The Redis client implements:
* a retry helper: https://redis.readthedocs.io/en/v4.3.1/retry.html
* a connection pool: https://pypi.org/project/redis/ (see "Connection Pools")

IMO, we can create one client and share it between the different threads.

So, for example, `incr_sliding_window(client, name: str, duration: int)` becomes `incr_sliding_window(name: str, duration: int)`.

dalf (commented):
See the comments

searx/redislib.py
return42: EXPIRE removes the object asynchronous while DEL is executed synchronous.

EDIT: synchronous DB manipulation blocks the python execution loop.

searx/plugins/limiter.py
return42: I know, but I think it is better not to bind a library to *one* DB connection.  IMO: The management of the DB connections is out of the scope of a library that implements of functions. .. may be one day we have multiple connections or a plugin want to use its own connection.

----

EDIT: ...

> we can create one client and share it between the different threads.

More or less we do this already in the `global _client` variable in `shared/redis.py`:

https://github.com/searxng/searxng/blob/e93385e6de8d77b1e8285547d642ca9f0d4021dc/searx/shared/redisdb.py#L30-L36

We can discuss if `shared/redis.py` is the right place to define the **one** DB connection of SearXNG, but I don't wanted to change connection management in this PR.  In this PR I thought it is better to focus on the function library ... and the function library should not limited to a single DB connection.

searx/plugins/limiter.py
return42: Not sure I understand you: the `name` argument is a string that identifies the counter / it is the name of the counter .. why should the name of a counter a tuple? (or did I misunderstood you?)

searx/redislib.py
return42: See my comment above: why should the `name` of a counter a tuple?

searx/redislib.py
```diff
@@ -0,0 +1,201 @@
...
+    The return value is the amount of items in the sorted set (ZCOUNT_), what
+    means the number of calls in the sliding window.
+
+    .. _Sorted sets in Redis:
+       https://redis.com/ebook/part-1-getting-started/chapter-1-getting-to-know-redis/1-2-what-redis-data-structures-look-like/1-2-5-sorted-sets-in-redis/
+    .. _TIME: https://redis.io/commands/time/
+    .. _ZADD: https://redis.io/commands/zadd/
+    .. _EXPIRE: https://redis.io/commands/expire/
+    .. _ZREMRANGEBYSCORE: https://redis.io/commands/zremrangebyscore/
+    .. _ZCOUNT: https://redis.io/commands/zcount/
```
return42, line 220: > The purpose of the ID is to make sure to count every request even if there are at the exact same timestamp.

Not sure I understand your comment, but here in the doc-string the LUA script is described.  The [TIME](https://redis.io/commands/time/) returns tuple and the `..` is a concatenation of the two values in this tuple (to get an unique name).

searx/redislib.py
return42: See https://github.com/searxng/searxng/pull/1047#discussion_r884233029

searx/redislib.py
return42: > Here the function creates a Script instance for each call.

Yes, the price is the calculation of the sha1 value on each request .. not optimal.

> Why not defined the script as a global variable?

AFAIK there is no function or a Script-constructor to create Script objects without a redis-client ... wait, now  I see there is one in the  core command:

https://redis.readthedocs.io/en/v4.3.1/commands.html?highlight=register_client#redis.commands.cluster.RedisClusterCommands.register_script

@dalf: 

I haven't tested this core `register_script` if you want I can give it a try.

But one question remains: when should this Script objects created .. while import `searx.redislib`?

searx/redislib.py
dalf: I couldn't find information about `DEL` vs `EXPIRE 0`.

According to [this micro benchmark](https://gist.github.com/dalf/d43a06ad4e1c5c3913ff57b3278da22a#file-redis_expire_vs_del-py),  `EXPIRE` is more or less equivalent to `DEL` is the best case scenario, otherwise `DEL` is faster especially when there are a lot of keys. May be, I have this point. Could you document the choice of EXPIRE instead of DEL?

searx/redislib.py
dalf: > when should this Script objects created

https://github.com/searxng/searxng/blob/38403ee515ba09fb675bbc8d544ea579436aeee6/searx/webapp.py#L1414-L1416

or

https://github.com/searxng/searxng/blob/38403ee515ba09fb675bbc8d544ea579436aeee6/searx/search/__init__.py#L27-L36

?

searx/redislib.py
dalf: * `redisdb.client()` connects to `localhost:6379`.
* the instantiation of a Redis client for each request has a performance impact according to [this micro benchmark](https://gist.github.com/dalf/d43a06ad4e1c5c3913ff57b3278da22a#file-reuse_redis_client-py)

searx/plugins/limiter.py
dalf: * `redisdb.client()` connects to `localhost:6379`.
* the instantiation of a Redis client for each request has a performance impact according to [this micro benchmark](https://gist.github.com/dalf/d43a06ad4e1c5c3913ff57b3278da22a#file-reuse_redis_client-py)

searx/redislib.py
return42: Its not about what command is faster .. its about transaction frames in the DB itself .. if there is an open transaction frame in the DB the `DEL` has to wait until the frame is closed, then `DEL` can delete the key .. when you use `EXPIRE` the call of the `EXPIRE` turns back immediately and the python command loop can continue .. the deletion (expire) of the key in the DB is managed by the DB internally.

In line 5 of `searx/redislib.py` I gave a link to [Bullet-Proofing Lua Scripts in RedisPy](https://redis.com/blog/bullet-proofing-lua-scripts-in-redispy/) .. its a very interesting article, may you like to read it.

searx/plugins/limiter.py
return42: > * `redisdb.client()` connects to `localhost:6379`.

It is connected to whatever redis.url from settings point to: `redis.Redis.from_url(get_setting('redis.url')) `

> * the instantiation of a Redis client for each request has a performance impact

Not sure I understand, but why do you think that we *"the instantiation of a Redis client for each request"* .. if you look at the code sequence shown above, there is a global variable `_client` that is set only once (if it is None).

Or do I missed something?

searx/redislib.py
return42: OK, tomorrow I will give [register_script](https://redis.readthedocs.io/en/v4.3.1/commands.html?highlight=register_client#redis.commands.cluster.RedisClusterCommands.register_script) a try / Implement a init() function for the `searx/redislib.py`.

return42: @dalf rebased this branch and pushed a commit on top to cache `Script` objects.

searx/redislib.py
```diff
@@ -0,0 +1,241 @@
...
+    """Purge all keys with ``prefix`` from database.
+
+    Queries all keys in the database by the given prefix and set expire time to
+    zero.  The default prefix will drop all keys which has been set by SearXNG
+    (drops SearXNG schema entirely from database).
+
+    The implementation is the lua script from string :py:obj:`PURGE_BY_PREFIX`.
+    The lua script uses EXPIRE_ instead of DEL_: if there are a lot keys to
+    delete and/or their values are big, `DEL` could take more time and blocks
+    the command loop while `EXPIRE` turns back immediate.
```
return42, line 63: Here I added a description why we use `EXPIRE` instead of `DEL`.

return42:
Any questions open with this PR .. or can we merge?

dalf:
[comment delete]

searx/plugins/limiter.py
dalf: I'm sorry I've read `redis.Client()` instead of `redisdb.client()`

dalf (approved):
Discussed on Matrix --> let's merge.

PR #1051, opened 2022-04-03
[fix] add missing territory (country) name

return42:
Add missing country name to languages filter.

Related-to: https://github.com/searxng/searxng/pull/1029#issuecomment-1086824911


dalf:
From https://github.com/searxng/searxng/pull/1029#issuecomment-1086824911 :
> Whatever the user will select, each engine implementation will do its best to narrow to what the user has selected

It most of the case the engine ignore the region with the exception of the google engines.

---

Before the flags were added, the language with only one country where shown without the country.

On a searx instance:
image

---

Looking at https://github.com/searxng/searxng/pull/1029 and the fact the country is not always use (`ca-ES` becomes `ca`), my opinion is:
* remove the flag and the region when there is only one region : `ca-ES` becomes `ca`.
* keep the flags when there are multiple regions for one language.

It's avoid the political aspect.

return42:
> remove the flag and the region when there is only one region : ca-ES becomes ca.

Yes, that s true .. TBH I do not really understand 

https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searxng_extra/update/update_languages.py#L117-L118

and 

https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searxng_extra/update/update_languages.py#L182

I think, we need some rework of these two functions (but this is not related to flags and regions)

I can send a PR to suggest a filter alternative, that reduce to languages without territories.

> It's avoid the political aspect.

It's curious to me, things like ca-ES are accepted but at the moment we add a flag to a region-code there is a political discussion.

I think the whole flags discussion is a discussion about a region issue in our languages filter that didn't aware before we add a flag to a region.

dalf:
> I can send a PR to suggest a filter alternative, that reduce to languages without territories.

I think this is the solution.
More clearly:
* if there is no multiple countries for one language, then the country should be removed.
* `en`, `en-GB`, `en-US`, etc...
* `ca`  (and not `ca-ES`) because there is no other entry for `ca`.

> I think the whole flags discussion is a discussion about a region issue in our languages filter that didn't aware before we add a flag to a region.

It seems a symbol bring more the political aspect than two letters, make sense in a way.

return42:
>> I can send a PR to suggest a filter alternative, that reduce to languages without territories.

> I think this is the solution.
> More clearly:
> * if there is no multiple countries for one language, then the country should be removed.
> * `en`, `en-GB`, `en-US`, etc...
> * `ca`  (and not `ca-ES`) because there is no other entry for `ca`.

:+1: .. but before I want to review all languages provider (engines) .. as closer I look to the languages topic I found more and more issues .. we will see that the list of languages & regions will be cleaned up when we review the providers .. 

The entire work is done in my ([branch mod-languages](https://github.com/return42/searxng/commits/mod-languages)) from where I release PRs .. ATM there are the following PRs that should be reviewed &  merged in that order: 

- https://github.com/searxng/searxng/pull/1071
- https://github.com/searxng/searxng/pull/1072

I am going to merge this PR now.

dalf:
> but before I want to review all languages provider (engines)

how is it related to this PR since the target is in the opposite direction (remove countries instead of adding them) ?

return42:
Sorry now I realize, that we misunderstood each other ..

> remove countries instead of adding them

We cant remove regions codes  entirely .. we have engines that do expect  languages and other expect a region code .. whatever we do, the search language filter in the UI will be a combination of languages (e.g. `zh`) and regions (`zh-HK`, `zh-TW` and `zh-CN`)

> how is it related to this PR

The language lists we have right now is partial crap .. e.g.:

    fetched 7839 languages from engine dailymotion

thats why I want to clean up the language provider first.  Before we improve the language list in the UI we need to know what languages and regions are needed by the engines.  It makes no sense to optimize a list, that is build up on partial crap.

dalf:
> we have engines that do expect languages and other expect a region code

AFIU, for example, the `ca` is going to be mapped to `ca_ES` with the `match_language` function.
(`ca` is the user preference, `ca_ES` is the accepted locale value by the engine).

return42:
If there is at the end no engine that supports `ca` why should the UI offer `ca` to the user when there is only `ca_ES`?

--- 

May be this is where we misunderstood each other .. do you think from the UI? ... I think from what  the engines do support ..

dalf:
> May be this is where we misunderstood each other .. do you think from the UI?

We can only support what the engines support.

Two weeks ago, I found that most of the engine ignore the country except the google engines (see below). Based on that, it would make sense to display only the language except when there are multiple region for one language. From the UX point of view: the user selects a language in the field named "languages" and not a region.

Hypothetical example 1:
* some engines supports `ca`, some other `ca-ES`
* we can display `ca` and map the value either to `ca` or `ca-ES` according to the selected engine.

Hypothetical example 2:
* all engines supports only `ca-ES` (and we are sure this is the actual value in the HTTP request: the update scripts can know that).
* we can display `ca-ES`, so the user knows that actual value (SearXNG is a "spokesperson" of the engines choices).

---

I've made this table two weeks ago (I've look at what the engines really send in the HTTP requests):
image
[searxng.language.support.ods](https://github.com/searxng/searxng/files/8457053/searxng.language.support.ods)

May be there are some errors, and there are some updates since then (the bing engines for example).
However, the conclusion at that moment was:
* the google engines supports the country
* the other engines, no so much.

return42:
> I've made this table two weeks ago (I've look at what the engines really send in the HTTP requests):

Strange .. when I look at google `uk-UA` I see:

```
>>> from searx import settings
>>> from searx.engines import load_engines
>>> engines = load_engines(settings['engines'])
>>> bing = engines['bing']
>>> bing.supported_languages
['af', 'am', 'ar', 'as', 'az-latn', 'be', 'bg', 'bn', 'bs-latn', 'ca', 'ca-es-valencia', 'chr-cher', 'cs', 'cy', 'da', 'de', 'el', 'en', 'es', 'et', 'eu', 'fa', 'fi', 'fil', 'fr', 'ga', 'gd', 'gl', 'gu', 'ha-latn', 'he', 'hi', 'hr', 'hu', 'hy', 'id', 'ig', 'is', 'it', 'ja', 'ka', 'kk', 'km', 'kn', 'ko', 'kok', 'ku-arab', 'ky', 'lb', 'lo', 'lt', 'lv', 'mi', 'mk', 'ml', 'mn-Cyrl-MN', 'mr', 'ms', 'mt', 'nb', 'ne', 'nl', 'nn', 'nso', 'or', 'pa-arab', 'pa-guru', 'pl', 'prs', 'pt-BR', 'pt-PT', 'quc', 'quz', 'ro', 'ru', 'rw', 'sd-arab', 'si', 'sk', 'sl', 'sq', 'sr-cyrl', 'sr-latn', 'sv', 'sw', 'ta', 'te', 'tg-cyrl', 'th', 'ti', 'tk', 'tn', 'tr', 'tt', 'ug', 'uk', 'ur', 'uz-latn', 'vi', 'wo', 'xh', 'yo', 'zh-Hans', 'zh-Hant', 'zu']
>>> 
>>> google = engines['google']
>>> google.match_language('uk-UA', google.supported_languages, google.language_aliases)
'uk'
```

dalf:
The `supported_languages` is not good indicator:
* the value returned by `match_languages` can be ignored partially.
* the `languages` parameter can be parsed outside of the `match_languages` function.

Another way to log the HTTP request: after this line https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/search/processors/online.py#L117 add `self.logger.debug(repr(params))` (or something equivalent to display the HTTP request including the HTTP headers)

---

About the google engine:
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/google.py#L188
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/google.py#L197-L203
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/google.py#L212
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/google.py#L244-L252

So the HTTP header `Accept-Language` contains the country without filtering by `match_languages`.

BTW, line 201 creates a bug: `ca` the languages becomes `CA` the country.

---

About bing:

https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/bing.py#L55-L60

Line 60 is the only line to use the `lang` variable.

return42:
> The `supported_languages` is not good indicator:

Now I am confused completely .. `supported_languages`  is the result of `_fetch_supported_languages()` of the same engine.  This is the list of languages and/or regions the engine supports.

> Another way to log the HTTP request: after this line

Yes but I do not understand how writing a log of something that is wrong can help.

I reviewed two engines and I found issues in both (and you found an issue in the google engine).

IMO the language & region handling has various issues and `match_languages` is the evil that gives the illusion there is a general solution for all this issues.

Thats why I want to review all aspects of languages and regions 

To repeat what I said, my plan is:

1. review of all `_fetch_supported_languages()` and 
2. repair the request of engines that do support languages / regions.
3. improve `searxng/searxng_extra/update/update_languages.py`

.. if you think I am wrong we can revert this PR and I stop my activity on this topic.

dalf:
> something that is wrong can help.

I don't understand what do you mean by "wrong" here: 
* bing.py sends the language in the query because this is the best(*) way that have been found. By best I mean the best solution we have found until now. What is wrong with it?
* google.py sends some HTTP headers and query parameters in the same way: this is the best way that we have found.

> match_languages is the evil that gives the illusion there is an general solution for all this issues.

match_languages doesn't have to be the scapegoat: this is the only tool that try to deal with this subject. we can improve it, for example to deal with the `zh-Hant-HK` value, we can add another functions. Removing it doesn't solve the need to map the language code from the user preference to the engine accepted values. Perhaps there is a better to map the values, but it is difficult to say without a careful analysis of the engine code. 

The plan I see:

* first step:
  * make a review of the mapping between the `languages` preferences and the actual sent value.
  * fix the mapping bugs.
  * adjust the engine code according to the finding: in the case of bing, there is no need to fetch the countries for example in  `_fetch_supported_languages`.
* second step: refactor some code according to the changes made in the first step.

return42:
> match_languages doesn't have to be the scapegoat: 

it is evil believe me :-)

> this is the only tool that try to deal with this subject. 

.. I have a doubt we can generalize what is individual for every engine. 

With more and more fixes we can improve `match_languages` to fit for all engines we have, but each fix needs to be tested in all engines that use `match_languages` .. that is why I think this function is evil.  

We should not only look at what we have, we will add other different engines in the future.
In a future proof concept only engines can map user's selections at its best, there is no way to generalize questions we have in the future .. and this is the other reason why I think this function is evil.

Myself tapped in this pitfall, when I fixed & implemented engines in the past I thought `match_languages` is the magic I need to handle languages in the engine I implemented.

But now, when I look closer to this topic I see that this was just an illusion and now I have to fix the issues I implemented in the past (e.g. in goggle & bing and more engines).

dalf:
> it is evil believe me :-)

The example you gave doesn't convince me, but I've experimented myself the function, so anyway.

What I suggest: in experimental branch, we push various commits to implement the mapping from the `languages` parameter to the expect values from the engines. The implementations doesn't use `match_languages`. Note: the usage of `babel.Locale` is not enough, the code has to deal with `languages` values that does not belong to accepted values.

Once few engines have been rewritten, we can see:
* if the code is enough simple, there is no need for a `match_languages` function
* if the code is complex, we can see if a common function would be useful

Does that sound workable for you?

return42:
I don't know .. I have the [branch mod-languages](https://github.com/return42/searxng/commits/mod-languages)

personally I prefer to go from engine to engine and like to review engine by engine .. each PR has its own aspects where the language topic is an aspect of all engines.

If you look at https://github.com/searxng/searxng/pull/1071 by example,  while I was working on this PR it becomes clear to me that the entire API did not work and that there more than 7000 languages in the list we don't need .. the reduce list will directly save memory in the runtime and code becomes a (little) bit faster .. therefore I have a high interest to merge this improvements to the master branch

ATM I'm working on the bing engines and I see other issues in the request function not related to to the language topic but i need to fix .. and I think it make not really sense to split the fix of this issues from the language topic ... without doubling the work. 

We can discuss it tomorrow.

PR #1056, opened 2022-04-04
Unquote path on result page.

### LINKED ISSUE #1055: Feature Request: Unquote url path on result page.
**Description**
By default, SearXNG will display the search results with Percent Encoded Url, just like:

image

Meanwhile, Google will display the results in decoded path which is easier to be read:
image

**Desired Solution**
Is it possible to make SearXNG to show decoded path by default just like Google?

capric98:
## Related issues
Closed: https://github.com/searxng/searxng/issues/1055

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -20,7 +20,7 @@ from typing import List, Dict, Iterable, Optional
 
 import urllib
 import urllib.parse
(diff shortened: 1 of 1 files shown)


dalf:
thank you!

dalf:
thank you!

PR #1058, opened 2022-04-05
[simple theme] netloc in image result article

mrpaulblack:
## What does this PR do?

Add netloc as a span to image article results.

Makes it easier to make out the source of an image without opening the detail modal and without JS.

```make run```

## Author's checklist

desktop simple dark mode:
image

mobile simple light mode:
image

### CODE REVIEW

File: searx/templates/simple/result_templates/images.html [hunk 1]
```diff
@@ -2,6 +2,7 @@
         <a {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">{{- "" -}}
                 <img class="image_thumbnail" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} src="{% if result.thumbnail_src %}{{ image_proxify(result.thumbnail_src) }}{% else %}{{ image_proxify(result.img_src) }}{% endif %}" alt="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{- "" -}}
                 <span class="title">{{ result.title|striptags }}</span>{{- "" -}}
+                <span class="title source">{{ result.parsed_url.netloc }}</span>{{- "" -}}
```
dalf: I wouldn't reuse the `title` class:
`<span class="source">`  and then 

```less
span.source,
span.title {
  ...
}
```

so it is easy to pick either the title or the source from javascript or CSS.

dalf:
TLDR: Except the nitpicking comment, LGTM.

---

The image template has a `source` field (not documented  : https://docs.searxng.org/dev/engine_overview.html?highlight=images%20html#media-types )

https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searx/engines/flickr_noapi.py#L81

The usage is not clear, but most of the time it contains the hostname, sometimes the user name (flick), or the channel name (youtube):
* https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searx/templates/simple/result_templates/images.html#L23
* https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searx/engines/google_images.py#L220
* https://github.com/searxng/searxng/blob/fcb626dedd3965de4fa6c95c47a34fdd14348afd/searx/engines/bing_images.py#L86

A nice improvement:
* by default the source field contains the hostname (as it is your PR)
* the engine can override (*) the value to include the channel name / user name etc... 

(*) override or add a prefix ( `<source>@<hostname>`)

This is the topic of another PR IMO.

searx/templates/simple/result_templates/images.html
mrpaulblack: done.

mrpaulblack:
Ok yeah I also tried out `result.source` I think you are right and we should make `result.source` return the netloc by default and otherwise engines can override this. But I think its for another PR, going to merge this one for now :+1:

PR #1061, opened 2022-04-06
limiter fixes && implement general.log_level

capric98:
### 1. Fix: 429 Too Many Requests when using Cloudflare
When SearXNG is deployed behind Cloudflare CDN, all search requests will be rejected with code 429 because Cloudflare seems to use `Connection: close` by default.
https://github.com/searxng/searxng/blob/4135a7400143cc934480afced3f6858c0d671fc6/searx/plugins/limiter.py#L61-L62

I am not sure whether `Connection: close` is set by Cloudflare or my browser, but there are some if conditions under `request.path == '/search'` making me feel uncertain... Considering they may not be that related to "rate limit", I also commented them arbitrarily. :P

### 2. Implement general.log_level in settings.yml
During fix 2, I added some debug&warning logs, and then I found that I cannot change log level conviniently using `settings.yml`.

Currently, SearXNG only have 2 log levels -- debug & warning(as LOG_LEVEL_PROD), and it is controlled by `settings.general.debug`. I added a new item `settings.general.log_level`, so `settings.general.debug` is deprecated.

To make other things consistent, `settings.general.debug` is now set as `settings.general.log_level=="debug"` implicitly:
```python
if "log_level" not in user_settings['general']: user_settings['general']['log_level'] = 'info'
user_settings['general']['debug'] = (user_settings['general']['log_level']=="debug")
```

### DIFF
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -36,10 +36,29 @@ re_bot = re.compile(
 )
 
 
+def cdn_reported_ip(headers: dict) -> str:
+    # some commonly used CDNs:
+    cdn_keys = [
+        "CF-Connecting-IP", # Cloudflare
+        "Fastly-Client-IP", # Fastly
+        "True-Client-IP",   # Akamai
+        "ar-real-ip",       # ArvanCloud
+    ]
+
+    for key in cdn_keys:
+        if (v := headers.get(key)):
+            return v
+
+    return ""
+
+
 def is_accepted_request(inc_get_counter) -> bool:
     # pylint: disable=too-many-return-statements
     user_agent = request.headers.get('User-Agent', '')
-    x_forwarded_for = request.headers.get('X-Forwarded-For', '')
+    client_ip  = request.headers.get('X-Forwarded-For', '')
+
+    # uncomment below if you want to use the CDN reported client ip:
+    # client_ip = cdn_reported_ip(request.headers)
 
     if request.path == '/image_proxy':
         if re_bot.match(user_agent):
@@ -47,31 +66,37 @@ def is_accepted_request(inc_get_counter) -> bool:
         return True
 
     if request.path == '/search':
-        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
-        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
+        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', client_ip])
+        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', client_ip])
+
         if c_burst > 15 or c_10min > 150:
             return False
 
         if re_bot.match(user_agent):
             return False
 
-        if len(request.headers.get('Accept-Language', '').strip()) == '':
+        if request.headers.get('Accept-Language', '').strip():
             return False
 
-        if request.headers.get('Connection') == 'close':
-            return False
+        # If SearXNG is behind Cloudflare, all requests will get 429 because
+        # Cloudflare uses "Connection: close" by default.
+        # if request.headers.get('Connection') == 'close':
+        #     return False
 
-        accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
-        if 'gzip' not in accept_encoding_list or 'deflate' not in accept_encoding_list:
-            return False
+        # "Accept-Encoding: gzip" will result in 429 because 'deflate' is not in accept_encoding_list...
+        # accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
+        # if 'gzip' not in accept_encoding_list or 'deflate' not in accept_encoding_list:
+        #     return False
 
         if 'text/html' not in request.accept_mimetypes:
             return False
 
-        if request.args.get('format', 'html') != 'html':
-            c = inc_get_counter(interval=3600, keys=[b'API limit', x_forwarded_for])
-            if c > 4:
-                return False
+        # IDK but maybe api limit should based on path not format?
+        # if request.args.get('format', 'html') != 'html':
+        #     c = inc_get_counter(interval=3600, keys=[b'API limit', client_ip])
+        #     if c > 4:
+        #         return False
+
     return True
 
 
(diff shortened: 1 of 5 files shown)


searx/plugins/limiter.py
```diff
@@ -36,42 +36,67 @@
...
-        if len(request.headers.get('Accept-Language', '').strip()) == '':
+        if request.headers.get('Accept-Language', '').strip():
             return False
 
-        if request.headers.get('Connection') == 'close':
-            return False
+        # If SearXNG is behind Cloudflare, all requests will get 429 because
+        # Cloudflare uses "Connection: close" by default.
+        # if request.headers.get('Connection') == 'close':
+        #     return False
```
mrpaulblack, line 99: Idk about this one... It boils down to what is more important: Having it work behind cloudflare (which we actively discourage public instances from using) or having the additional bot protection.

Tbh I dont think this helps a lot so I am ok with removing it...

searx/plugins/limiter.py
```diff
@@ -36,42 +36,67 @@
...
 
-        accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
-        if 'gzip' not in accept_encoding_list or 'deflate' not in accept_encoding_list:
-            return False
+        # "Accept-Encoding: gzip" will result in 429 because 'deflate' is not in accept_encoding_list...
+        # accept_encoding_list = [l.strip() for l in request.headers.get('Accept-Encoding', '').split(',')]
```
mrpaulblack, line 102: This should be kept. Basically all browsers send this and this is IMO a great way to protect against basic bots... Again if cloudflare does not support this I am actually ok with it... (Since with cloudflare the bulk of bots gets already blocked, even without this limiter plugin)

searx/plugins/limiter.py
```diff
@@ -36,42 +36,67 @@
...
+        #     return False
 
         if 'text/html' not in request.accept_mimetypes:
             return False
 
-        if request.args.get('format', 'html') != 'html':
-            c = inc_get_counter(interval=3600, keys=[b'API limit', x_forwarded_for])
-            if c > 4:
-                return False
+        # IDK but maybe api limit should based on path not format?
```
mrpaulblack, line 110: So basically the limiter plugin is only protecting paths where engines could rate limit or IP block the SearXNG instance... So the API paths like `/stats/errors` or `config` are not covered by the plugin and it is IMO ok since the plugin is built into SearXNG core and is not going to protect against DOS attacks for example. This is only protecting API requests on `/search` basically.

So I think you are right; this should basically block `/search` GET and POST requests for JSON, CSV and RSS requests... (since `/search` does not return an file ending; also by default only HTML results are allowed)

mrpaulblack (commented):
Thanks for your PR, I added a few comments to the limiter plugin changes... I mostly like your changes, just a few nits...

searx/plugins/limiter.py
```diff
@@ -36,42 +40,79 @@
...
         return True
 
     if request.path == '/search':
-        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', x_forwarded_for])
-        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', x_forwarded_for])
-        if c_burst > 15 or c_10min > 150:
+        c_burst = inc_get_counter(interval=20, keys=[b'IP limit, burst', client_ip])
+        c_10min = inc_get_counter(interval=600, keys=[b'IP limit, 10 minutes', client_ip])
+
+        logger.debug(f"counts of [{client_ip}]: c_burst={c_burst}, c_10min={c_10min}")
```
mrpaulblack, line 82: Also I dont think its a good idea to log the client IP. We are a privacy focused search engine and logging IP addresses speaks against that... -> please remove the IP logging from this PR, thanks :+1:

searx/plugins/limiter.py
capric98: Seems like `CF-Connecting-IP` header is protected by cloudflare, so I change this to
```python
if request.headers.get('Connection') == 'close' and "CF-Connecting-IP" not in request.headers:
    return False
```

searx/plugins/limiter.py
capric98: OK, but again, cloudflare rewrites `Accept-Encoding: gzip, deflate, br` to `Accept-Encoding: gzip`... Maybe changing `or` to `and` is a simple solution.

searx/plugins/limiter.py
capric98: Thanks for your explanation, discarded the change for this.

searx/plugins/limiter.py
capric98: Removed the IP logging, and uses `client_ip_masked` to log rejected requests.

capric98:
Added `SimpleResp` part, now the rejected request will get a simple error message - `Too Many Requests` or `Bad Request`, and corresponding status code.

return42:
Can you please send a separate PR for b4e84c309e .. AFIK it is a bugfix not related to the limiter modifications in this PR / thanks!

capric98:
> Can you please send a separate PR for [b4e84c3](https://github.com/searxng/searxng/commit/b4e84c309e2835c04c926f72bba5e25690fc513a) .. AFIK it is a bugfix not related to the limiter modifications in this PR / thanks!

Deleted commit b4e84c3 and created a new PR :)

mrpaulblack:
hi @capric98 ,
I was checking back with the other project maintainers and we basically agreed on the following: SearXNG is a privacy focused search engine. Supporting cloudflare or making changes to the limiter plugin which lessens the bot protection only for the CF support is not an option for our project. Logging IP addresses is also against what this project stands for.

I am sorry that I already did a partial review of this PR, but we all agreed after some discussion that its best not to merge this PR. The bug fix is tracked seperatly and will probably get merged, but the limiter changes are against what we stand for and the logger changes are also not the full story.

Turns out that the debug option does more then setting the logger to debug:
This is not (well) documented, general.debug = True does more than logging.

Here it activate auto reloading of the code when it is changed:
https://github.com/searxng/searxng/blob/4135a7400143cc934480afced3f6858c0d671fc6/searx/webapp.py#L1406

Here it disable the checking of the secret_key:
https://github.com/searxng/searxng/blob/4135a7400143cc934480afced3f6858c0d671fc6/searx/webapp.py#L128-L130

Here the checker is disabled:
https://github.com/searxng/searxng/blob/4f82ab36a90db39d21201c11df3f4a437188e8d0/searx/search/checker/background.py#L154-L156

For those reason I will close this PR, thanks for the interest in the project tho :+1: If you want to use cloudflare, you dont really need the limiter plugin anyway, since CF is already a WAF and if you really want to use your changed limiter plugin you can make the changes for your instance locally or in your own fork; Closing.

capric98:
@mrpaulblack Thanks for your review and discussion with other maintainers!
Though setting `general.log_level` to `debug` has the same effect as `general.debug == true`, it may not be that useful considering SearXNG is a privacy focused search engine. Actually I did not realize it before and was wondering why there is no precisely leveled logs, and now it makes sense.

Anyway, great thanks to all of your brilliant works on this project!

PR #1064, opened 2022-04-07
fix: %2F in path will be converted to " › " in searx/webapp.py

capric98:
Seperated from [#PR 1061](https://github.com/searxng/searxng/pull/1061)


mrpaulblack (approved):
LGTM, thanks :+1:

PR #1071, opened 2022-04-08
[fix] dailymotion engine: filter by language & country

### LINKED ISSUE #1065: engine dailymotion: filter by language vs. filter by localization
The API of the dailymotion is able to filter by 

**language**: [ref](https://developers.dailymotion.com/api/#result-filtering)
    
    For instance, the following request /videos?channel=tech&language=en 
    will list all English-speaking videos in the ‘technology’ channel.

or by **localization**: [Change the default localization of the user.](https://developers.dailymotion.com/api/#global-parameters)

    This will affect results language and content selection.

ATM our engine query filters by **localization**:

https://github.com/searxng/searxng/blob/4135a7400143cc934480afced3f6858c0d671fc6/searx/engines/dailymotion.py#L38-L40

This is IMO OK, but there is an issue when we fetch the "languages" ...

https://github.com/searxng/searxng/blob/4135a7400143cc934480afced3f6858c0d671fc6/searx/engines/dailymotion.py#L28

----

**BUG**: We can't fetch **languages** and pass them to a **localization** argument in the query.

----

IMO this needs to be fixed to fetch the localization (even if the function is named *fetch_supported_languages* what might be a little bit confusing at this point).

from [Detecting and retrieving locales](https://developers.dailymotion.com/api/#locale)

- https://api.dailymotion.com/locales

What do other think about: filter by language vs. filter by localization

----

This reminds me that we do not select only languages in our UI, we often (want) to select  "locales" (aka regions, territory) .. some engines support 'language' filter, some support 'region' filter and others like dailymotion support both and we have to decide for one.

return42:
## What does this PR do?

[fix] dailymotion engine: filter by language & country

- fix the issue of fetching more the 7000 *languages*
- improve the request function and filter by language & country
- implement time_range_support & safesearch
- add more fields to the response from dailymotion (allow_embed, length)
- better clean up of HTML tags in the 'content' field.

This is more or less a complete rework based on the '/videos' API from [1].
This patch cleans up the language list in SearXNG that has been polluted by the
ISO-639-3 2 and 3 letter codes from dailymotion languages which have never been
used.

[1] https://developers.dailymotion.com/tools/

Checkout branch and run a local instance:

    make run

Try some dailymotion searches:

    !dailymotion foo bar

Test various languages & regions, test various time range.

You can also SafeSearch but as far I can say, there is no major change in the results.

## Related issues

Closes: https://github.com/searxng/searxng/issues/1065


return42:
FYI: this is on PR of my complete review ([branch mod-languages](https://github.com/return42/searxng/commits/mod-languages)) of the handling of languages and locales (aka regions) in SearXNG. https://github.com/searxng/searxng/pull/1051 is also a part of this series. More PRs will follow.

searx/engines/dailymotion.py
```diff
@@ -24,21 +25,32 @@
...
-    if params['language'] == 'all':
-        locale = 'en-US'
-    else:
-        locale = match_language(params['language'], supported_languages)
 
+    language = params['language']
+    if language == 'all':
+        language = 'en-US'
+
+    locale = babel.Locale.parse(language, sep='-')
```
dalf, line 87: With this version, `supported_languages` is ignored.
Do you intended to remove `match_language`?

For example `ca` (Catalan)
* becomes `en` with the `match_language`
* remains `ca` with this line --> I don't know how dailymotion deals with this unexpected value.

return42: > Do you intended to remove `match_language`?

May be .. while I review engine by engine I see that this function does not fit to the language that is needed for the query send to the search engine.

> * remains `ca` with this line --> I don't know how dailymotion deals with this unexpected value.

Oops, my fail .. I will fix it. But this will not improve the result list:

> In contrast to the documentation of the api [1], dailymotion's API does not care
about a 'localization' argument or a 'language' filter[2] (may be there is an
API account needed to get in use of this options).

dalf: I ask the here, but I see the same logic in #1073
Do you intend to send the language & region unfiltered to the engines?
At first sight, I think it is regression (`match_language` is going to map `ca-ES` to `ca` or `en`, so there is no unexpected value ; same `de-DE` is going to be mapped to `de` or `en`).

searx/engines/dailymotion.py
```diff
@@ -24,21 +25,39 @@
...
+    if localization in supported_languages:
+        params['url'] = search_url.format(
+            query=urlencode({'search': query, 'localization': localization }),
+            pageno=params['pageno']
+        )
+    else:
+        params['url'] = search_url.format(
+            query=urlencode({'search': query}),
+            pageno=params['pageno']
+        )
```
return42: > * remains `ca` with this line --> I don't know how dailymotion deals with this unexpected value.

fixed

> Do you intend to send the language & region unfiltered to the engines?
At first sight, I think it is regression (`match_language` is going to map `ca-ES` to `ca` or `en`, so there is no unexpected value ; same `de-DE` is going to be mapped to `de` or `en`).

That is the issue of using this function here, the engine expect only one of the values from `supported_languages` .. here, dailymotion expect a locale from 

https://github.com/searxng/searxng/blob/74cf47fa1422384eb4e6fb047a9ff41aeeed5e5f/searx/data/engines_languages.json#L430-L481 

.. there is no language 'ca', 'en' .. available and it is wrong to send a language in the query.

dalf: So in the case of `ca` (*), what should be done?
* should the engine skip the language parameter? (the results are going to be related to the instance location).
* should the engine use `en-US`? or even `es_ES`? (`all` is mapped to `en-US`)
* should the engine returns zero result?

(*) currently [`ca-ES`](https://github.com/searxng/searxng/blob/5e7f46991104a6583676c6c9319a81b6f9a4b151/searx/languages.py#L9) but it is going to be `ca` [as far I understand](https://github.com/searxng/searxng/pull/1051#issuecomment-1093996357).

searx/engines/dailymotion.py
return42: We can't send `ca` to a engine that only except `ca_ES`.

dalf: So we need a mapping between the language preference and the engine expected value.
* In the master branch, the `match_language` function maps the user preference to the engine expected values.
* In the PR, the mapping is done inside the engine code. However it may send unexpected value: `ca`or `ca_ES` are neither accepted by Dailymotion (see comment below).

> I see that this function does not fit to the language that is needed for the query send to the search engine.

Can you say more and/or give examples?

return42: > Can you say more and/or give examples?

E.g. bing, the script Hant & Hans

```py
>>> import babel
>>> from searx import settings
>>> from searx.engines import load_engines
>>> engines = load_engines(settings['engines'])
>>> bing = engines['bing']
>>> bing.supported_languages
['af', 'am', 'ar', 'as', 'az-latn', 'be', 'bg', 'bn', 'bs-latn', 'ca', 'ca-es-valencia', 'chr-cher', 'cs', 'cy', 'da', 'de', 'el', 'en', 'es', 'et', 'eu', 'fa', 'fi', 'fil', 'fr', 'ga', 'gd', 'gl', 'gu', 'ha-latn', 'he', 'hi', 'hr', 'hu', 'hy', 'id', 'ig', 'is', 'it', 'ja', 'ka', 'kk', 'km', 'kn', 'ko', 'kok', 'ku-arab', 'ky', 'lb', 'lo', 'lt', 'lv', 'mi', 'mk', 'ml', 'mn-Cyrl-MN', 'mr', 'ms', 'mt', 'nb', 'ne', 'nl', 'nn', 'nso', 'or', 'pa-arab', 'pa-guru', 'pl', 'prs', 'pt-BR', 'pt-PT', 'quc', 'quz', 'ro', 'ru', 'rw', 'sd-arab', 'si', 'sk', 'sl', 'sq', 'sr-cyrl', 'sr-latn', 'sv', 'sw', 'ta', 'te', 'tg-cyrl', 'th', 'ti', 'tk', 'tn', 'tr', 'tt', 'ug', 'uk', 'ur', 'uz-latn', 'vi', 'wo', 'xh', 'yo', 'zh-Hans', 'zh-Hant', 'zu']
>>> 
>>> bing.match_language('zh', bing.supported_languages, bing.language_aliases)
'zh-Hans'
>>> babel.Locale.parse('zh', sep = '-')
Locale('zh')
>>> 
>>> bing.match_language('zh-CN', bing.supported_languages, bing.language_aliases)
'zh-Hans'
>>> babel.Locale.parse('zh-CN', sep = '-')
Locale('zh', territory='CN', script='Hans')
>>> 
>>> bing.match_language('zh-TW', bing.supported_languages, bing.language_aliases)
'zh-Hans'
>>> babel.Locale.parse('zh-TW', sep = '-')
Locale('zh', territory='TW', script='Hant')
>>> 
>>> bing.match_language('zh-HK', bing.supported_languages, bing.language_aliases)
'zh-Hans'
>>> babel.Locale.parse('zh-HK', sep = '-')
Locale('zh', territory='HK', script='Hant')

```

return42:
@dalf according to what we discussed today I wanted to ask if it is OK when we merge this PR .. or do you see some parts here in this PR we need to improve before merge?

searx/engines/dailymotion.py
```diff
@@ -21,23 +23,73 @@
...
-        query=urlencode({'search': query, 'localization': locale}), pageno=params['pageno']
-    )
+    language = params['language']
+    if language == 'all':
+        language = 'en-US'
+    locale = babel.Locale.parse(language, sep='-')
+
+    query_args = {
+        'search': query,
+        'languages': locale.language,
```
dalf: When the language preference is `ca-ES` (or just `ca`), this value is sent to dailymotion even if dailymotion does not support `ca` = there is no safe guard between the preferences and the value sent to the engine. It works, but it may raise an error in the future. For example a new value in `languages.py` may raise an error in dailymotion.

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
...
         url = res['url']
-        content = html_to_text(res['description'])
-        thumbnail = res['thumbnail_360_url']
+        video_id = res['id']
+
+        content = res['description']
+        if content:
+            content = content.replace('\\r', '')
+            document = lxml.html.document_fromstring(content)
+            content = " ".join(lxml.etree.XPath("//text()")(document))
```
dalf: Why not use `searx.utils.html_to_text` ?

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
...
         publishedDate = datetime.fromtimestamp(res['created_time'], None)
 
+        length = time.gmtime(res.get('duration'))
+        if length.tm_hour:
+            length = time.strftime("%H:%M:%S", length)
+        else:
+            length = time.strftime("%M:%S", length)
+
+        thumbnail = res['thumbnail_360_url']
         # http to https
```
dalf: I would remove this comment: the following line is obvious.

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
...
+            'content': content,
+            'publishedDate': publishedDate,
+            'length': length,
+            'thumbnail': thumbnail,
+        }
+
+        # HINT: no mater what the value is, without API token videos can't shown
+        # embedded
+        allow_embed = res['allow_embed']
+        if allow_embed:
```
dalf:
```python
if res['allow_embed']:
```
is still readable and shorter.

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
...
-        name = language['native_name']
-        if name:
-            supported_languages[language['code']]['name'] = name
-        english_name = language['name']
-        if english_name:
-            supported_languages[language['code']]['english_name'] = english_name
+    for item in response_json['list']:
+        supported_languages.append(item['locale'])
 
     return supported_languages
```
dalf: The whole function can be written this way: 
```python
def _fetch_supported_languages(resp):
    response_json = resp.json()
    return [
        item['locale']
        for item in response_json['list']
    ]
```

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
...
+            'publishedDate': publishedDate,
+            'length': length,
+            'thumbnail': thumbnail,
+        }
+
+        # HINT: no mater what the value is, without API token videos can't shown
+        # embedded
+        allow_embed = res['allow_embed']
+        if allow_embed:
+            item['iframe_src'] = "https://www.dailymotion.com/embed/video/" + video_id
```
dalf: Currently hard coded URLs are set at the top of the files (here `iframe_url`, see qwant.py)

searx/engines/dailymotion.py
```diff
@@ -54,45 +106,61 @@ def response(resp):
```
dalf, line 127: There is no list when there is an error: we can log this error (see qwant.py), so debugging is going to be easier.

dalf (changes requested):
I left some comments.
These two commits fix the various issues except the mapping between the languages preference (`'languages': locale.language,`):
* https://github.com/dalf/searxng/commit/cdecb87478bb114b4f74f418ed4afb244e32b138
* https://github.com/dalf/searxng/commit/32990f6d554ba270133904a68579c4747ec961b4
* https://github.com/dalf/searxng/commit/f4ed1110f2b1635e9847e36d997a924eb8c0bbca

(form https://github.com/searxng/searxng/compare/master...dalf:return42-fix-lang-dailymotion )

searx/engines/dailymotion.py
return42: I have seen issues with `&lt;` and similar HTML notations.

searx/engines/dailymotion.py
dalf: Fix in commit https://github.com/searxng/searxng/commit/f4ed1110f2b1635e9847e36d997a924eb8c0bbca

searx/engines/dailymotion.py
dalf: I can't reproduce the bug: `searx.utils.html_to_text('<span>test &lt; </span>') == 'test <'`.

It would be better to fix the `html_to_text` function rather than introducing a new implementation here.

return42: Strange .. can't reproduce the `&lt;` issue nor the issue with `\\r` .. will switch back to `html_to_text(..)` / thanks!

return42:
@dalf thanks for your review and the commit .. I fixed according to your remarks .. if you think PR is fine, press merge button :-)

searx/engines/dailymotion.py
dalf: From cdecb87478bb114b4f74f418ed4afb244e32b138 :

```python
def request(query, params):
    ....
    params['raise_for_httperror'] = False

    return params

def response(resp):
    results = []

    search_res = resp.json()

    # check for an API error
    if 'error' in search_res:
        raise SearxEngineAPIException(search_res['error'].get('message'))

    # raise for other errors
    raise_for_httperror(resp)
```

How to test: query for `!dailymotion`, since the query sent to dailymotion is empty, the API returns an error (the bug should not be fix in this PR).

Do note the usage of `search_res['error'].get('message')` which provides a meaningful error message.

This implementation is similar to `qwant.py`.

searx/engines/dailymotion.py
```diff
@@ -50,49 +116,56 @@ def response(resp):
...
         return []
 
     # parse results
     for res in search_res['list']:
+
         title = res['title']
         url = res['url']
+        video_id = res['id']
+
         content = html_to_text(res['description'])
```
dalf, line 134: See the commit https://github.com/dalf/searxng/commit/32990f6d554ba270133904a68579c4747ec961b4 : `<br />` is replaced by a space, so `html_to_text("This is a<br />text") == "This is a text"`.

How to test: `!dailymotion :en weather`.
One of the result is video `https://www.dailymotion.com/video/x7t2nzh`, the description is `전국 건조한 날씨 계속<br /><br />Good evening.`. With the suggested change, the text is `전국 건조한 날씨 계속 Good evening` with a space between 속 and "Good".

searx/engines/dailymotion.py
return42: Oops I have overseen your additional commits ..

> Do note the usage of `search_res['error'].get('message')` which provides a meaningful error message.

I squashed your https://github.com/searxng/searxng/commit/cdecb87478bb114b4f74f418ed4afb244e32b138 into my commit 3bb62823 and modified the request function ... 

> How to test: query for !dailymotion, since the query sent to dailymotion is empty,

Instead of sending an unsupported request (empty query string) to daylimotion, additionally I return `False` from the request function if query string is empty ..  https://github.com/searxng/searxng/pull/1071/files#diff-f3739fa1a440898cb603ad679a81c9429d1e74409852c45dd2c8e5f595c98aa0R81-R82

---

If you think PR is now ready .. go ahead and merge ..

searx/engines/dailymotion.py
return42: I cherry picked you commit in 4224607c / thanks!

searx/engines/dailymotion.py
```diff
@@ -21,23 +26,89 @@
...
-    if params['language'] == 'all':
-        locale = 'en-US'
-    else:
-        locale = match_language(params['language'], supported_languages)
 
-    params['url'] = search_url.format(
-        query=urlencode({'search': query, 'localization': locale}), pageno=params['pageno']
-    )
+    if not query:
+        return False
```
dalf, line 82: nit: I would fix this issue in the `searx.search` page: engines should not received an empty query.

dalf (approved):

PR #1072, opened 2022-04-08
[mod] update_languages.py - review of fetch_supported_languages()

return42:
This patch improves the fetch_supported_languages() from update_languages.py .. its one more patch of my complete review languages and locales aka regions ([branch mod-languages](https://github.com/return42/searxng/commits/mod-languages))

This PR is based on // **we should merge in this order!!!**

- https://github.com/searxng/searxng/pull/1051
- https://github.com/searxng/searxng/pull/1071


searx/engines/__init__.py
```diff
@@ -219,18 +218,6 @@ def set_language_attributes(engine: Engine):
...
-    # assign language fetching method if auxiliary method exists
-    if hasattr(engine, '_fetch_supported_languages'):
-        headers = {
-            'User-Agent': gen_useragent(),
-            'Accept-Language': "en-US,en;q=0.5",  # bing needs to set the English language
-        }
-        engine.fetch_supported_languages = (
-            # pylint: disable=protected-access
-            lambda: engine._fetch_supported_languages(get(engine.supported_languages_url, headers=headers))
-        )
```
return42, line 231: @MarcAbonce I moved the closures (lambda) from your commit 9b6ffed06 to the `searxng_extra/update/update_languages.py` .. IMO these manipulation of engine's namespace is not needed here when initialize the engine (what happens e.g. when a SearXNG service starts).

The lambda function is only called in the `update_languages.py`, that's why I think it is better and more clear when we move this part to the `update_languages.py`.

What do you think, did I missed something?

dalf: I agree, this code could be in `update_languages.py` (so the comment `# bing needs to set the English language` has the proper context).

dalf:
About the cache in `update_languages.py`: it avoids few requests every month.
I haven't tried to use [`@cache`](https://docs.python.org/3/library/functools.html#functools.cache)  annotation.
but IMO it is not worth the additional complexity.

---

An unfinished idea: the engines can define a `supported_languages_from` variable.

Then instead of 
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/__init__.py#L182-L191

we can have:
```python
    if engine.supported_languages_from:
        # we have have to make sure the referenced engine is loaded (that's the difficult part, and most probably show stopper for this idea)
        engine.supported_languages = ENGINES_LANGUAGES[engine.supported_languages_from]
    if engine.name in ENGINES_LANGUAGES:
        engine.supported_languages = ENGINES_LANGUAGES[engine.name]
    elif engine.engine in ENGINES_LANGUAGES:
        engine.supported_languages = ENGINES_LANGUAGES[engine.engine]
```

and, for example, in `bing_images.py`, we can remove these lines 
https://github.com/searxng/searxng/blob/7217e430706ce36897b7da0889247f70ab420fce/searx/engines/bing_images.py#L14-L17
and replace them by
`supported_languages_from = 'bing'`.

return42:
> I haven't tried to use [`@cache`](https://docs.python.org/3/library/functools.html#functools.cache)  annotation. but IMO it is not worth the additional complexity.

New in version py3.9.

> An unfinished idea: the engines can define a `supported_languages_from` variable.

We should not complicate the core of SearXNG for something that can be solved by a simple import statement .. BTW engines do not only names of the language topic, they might also support and share other functions and names.

---

EDIT is there anything we need to fix in this PR or is this PR ready to merge?

dalf:
> We should not complicate the core of SearXNG for something that can be solved by a simple import statement 

Yes I agree, that's why I wrote "unfinished" and "deal breaker". I wasn't able to find a proper solution (if engine were Python class, perhaps it could be different).

> is there anything we need to fix in this PR or is this PR ready to merge?

IMO, there is no need to cache the result in `update_languages.py` : it makes the code complicate to avoid less than 10 HTTP requests per month.

return42:
> Yes I agree, that's why I wrote "unfinished" and "deal breaker".

In your comment was only a hint "unfinished" .. and it was not clear to me if it was a request you want to solve in this PR or not.

> it makes the code complicate to avoid less than 10 HTTP requests per month.

The intention of the cache was more not to get blocked when fire multiple identical request in less a few second. But if you want, I can remove the cache.

dalf:
>  it was not clear to me if it was a request you want to solve in this PR or not.

the "deal breaker" expression is lot in the comments, sorry.

> The intention of the cache was more not to get blocked when fire multiple identical request in less a few second. 

As far I remember, we never had an issue. But yes, each month I see these multiple requests, and the copy / paste result in the json file, and ask myself is there a better way?

> But if you want, I can remove the cache.

Yes to keep the code simple. `fetch_supported_languages` can be declared in the `searxng_extra/update/update_languages.py` script.

return42:
FYI: I removed the caching implementation and rebased this PR on current master.

return42:
Switched to DRAFT: needs to be rebased when 

- https://github.com/searxng/searxng/pull/1126

is merged.

PR #1076, opened 2022-04-09
[fix] Search language or region

return42:
The name 'language' is misleading, in the filter the user selects a language or a region.


dalf:
Let's modify this label once the code of the various engines has been updated.

return42:
Switched to DRAFT: needs to be rebased when 

- https://github.com/searxng/searxng/pull/1126

is merged.

PR #1084, opened 2022-04-11
don't reinstall existing packages on arch

Jabster28:
## What does this PR do?

adds the `--needed` flag to the pacman install script, which tells pacman not to fetch and reinstall packages that are already on the system.

it might not be important for most uses, but downloading ~200mb of packages that were already installed can be troublesome for slow networks

execute `sudo -H ./utils/searx.sh install all` on an archlinux installation.

## Author's checklist

N/A

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -1535,7 +1535,7 @@ pkg_install() {
             ;;
         arch)
             # shellcheck disable=SC2068
-            pacman --noprogressbar -Sy --noconfirm $@
+            pacman --noprogressbar -Sy --noconfirm --needed $@
             ;;
         fedora)
             # shellcheck disable=SC2068


mrpaulblack:
Thanks @Jabster28 :D

@return42 I tested this on my manjaro install and it works perfectly (as described skips packages that are already installed). From my side LGTM, ok to merge?

return42:
I haven't tested by myself, but makes sense .. while we are at pacman options I read this article [Avoid certain pacman commands](https://wiki.archlinux.org/title/System_maintenance#Avoid_certain_pacman_commands)

@Jabster28 I am not an archlinux expert .. what do you think, should we modify to `pacman -Syu`: 

    Always upgrade (with pacman -Syu) before installing a package.

Jabster28:
> should we modify to `pacman -Syu`?

since arch is a rolling release distro, there's the potential for hundreds of packages to be updated from a system update (look at this!), most of which would be irrelevant to installing searxng
Screenshot 2022-04-12 at 10 43 03


while everyone says you **should** update before installing packages, i rarely see any installation scripts do it. every now and then i install packages without updating and i don't think i've had any problems with it.

also, it wouldn't make much sense to an end user. `"I just want to install a python project, why is my entire system updating?"`

Jabster28:
however, i think removing the `-y` flag could help. removing it would make pacman yell at the user if the repos are too out of date, which gives the user the choice of either updating fully or doing a partial upgrade.

return42:
> however, i think removing the `-y` 

not from a script point of view .. theses scripts are running in background on all distros:

https://github.com/searxng/searxng/blob/27f8fa6fe066c15dfc60bc2c50533ef87663ce29/utils/lib.sh#L1532-L1547

> while everyone says you **should** update before installing packages, i rarely see any installation scripts do it. every now and then i install packages without updating and i don't think i've had any problems with it.

thanks for the info .. I do have similar experience .. with this in mind I think we do not need to update when installing packages.

I'm going to merge this PR / thanks a lot!

PR #1087, opened 2022-04-12
Remove search.js MinChars & Implement server-side autocomplete limit.

capric98:
## What does this PR do?
1. By default, theme simple has a autocomplete limit that it will not make `/autocompleter?q=xxx` requests unless `len(query)>=4`. I removed this restriction, and it will bring little overhead to the server, but it should be fine.
2. Implement server-side autocomplete limit, in case of potential abuse. The extra configuration is like:
```yaml
server:
  autocomplete_limit:
    min: 0
    max: 20
    # overwrite:
    #   "{{ backend_name }}":
    #     min: 4
    #     max: 16
```


mrpaulblack:
Can you please fix the test? You can run the test locally by running `make test` in your local dev env. It should be something with the `autocomplete.js` after a quick look: https://github.com/searxng/searxng/runs/5993807709?check_suite_focus=true

Also if I get this right: In this PR its possible to define a min and max char limit on server side for the autocomplete feature in general and also for each autocomplete backend independently?

I also think you should set the values from settings.yml also in the settings default please :+1:

return42:
AFAIK the client fires autocomplete request to the instance for every character that is typed in the search field .. but the instance only response suggestions if the number of characters is in the min/max range .. right?

The intention of having 4 characters at minimum was to not fire to much request to the instance .. or with other words, the limit needs to be implemented fully on the client side.

capric98:
> Can you please fix the test? You can run the test locally by running `make test` in your local dev env. It should be something with the `autocomplete.js` after a quick look: https://github.com/searxng/searxng/runs/5993807709?check_suite_focus=true
> 
> Also if I get this right: In this PR its possible to define a min and max char limit on server side for the autocomplete feature in general and also for each autocomplete backend independently?
> 
> I also think you should set the values from settings.yml also in the settings default please 👍

Maybe I could take a look at the test tomorrow.
Btw I added a default value from 0-50 just now.

capric98:
> AFAIK the client fires autocomplete request to the instance for every character that is typed in the search field .. but the instance only response suggestions if the number of characters is in the min/max range .. right?
> 
> The intention of having 4 characters at minimum was to not fire to much request to the instance .. or with other words, the limit needs to be implemented fully on the client side.

The `autocomplete.js` seems to have a default 300ms delay to make a new request, so I think it should be fine?
https://github.com/searxng/searxng/blob/1388b44f8b1397687660331c607cdf85a5cf785a/searx/static/themes/simple/src/js/main/search.js#L72
It means to a normal user, the autocomplete request will be sent out after idling for more than 300ms(I guess).

And currently, if I makeup lots of requests mannually rather than the browser, the server will proxy each queries to the backend and may cause abuse, maybe we could consider to add `/autocompleter` to `limiter.py`?

return42:
> so I think it should be fine?

sorry to say but no it is not OK to fire useless requests to the instance .. SearXNG invest a lot of work to be sparse with instance requests.

capric98:
> > so I think it should be fine?
> 
> sorry to say but no it is not OK to fire useless requests to the instance .. SearXNG invest a lot of work to be sparse with instance requests.

But  oscar theme just has no min-chars restriction now, and it uses `typeahead.js`.

We could have more strict default value of `autocomplete_limit`, or add it to `limiter.py`.
Currently `/autocompleter` is vulnerable to cause abuse, and add js-side restrictions won't help IMO.

mrpaulblack:
I think its ok if we have a limit of 0 to 50 chars on client side as well as server side. It will IMO not add a lot more requests to an instance, because of the 300ms idle offset. And on top this could even lower the requests and even be somewhat of a privacy feature with an upper limit. If for example someone with SearXNG as default search provider copy pasts an URL or a unwanted string into the search bar its not sent to the instance anymore since its probably above the 50 char limit -> thus saving requests and bandwidth...

Also FYI I dont think we need a limiter impl. for autocomplete endpoint. Its not getting rate limited by upstream providers, so why should we? Rate limiting in its current form cannot prevent DOS for example and also I think we should keep rate limiting and blocking with SearXNG to a minimum; its already really annoying when other upstream engines do this :upside_down_face: ...

dalf:
Technical note about a30d757087be595015acf5d1039a30c004d8de11:
* can you remove this commit.
* call `make static.build.commit` : it is going to compile the themes & create a commit. So when we merge branches, it is easy to recognize the build commit and remove it.

https://docs.searxng.org/dev/quickstart.html?highlight=static%20build%20commit

capric98:
> remove commit [a30d757](https://github.com/searxng/searxng/commit/a30d757087be595015acf5d1039a30c004d8de11) and call `make static.build.commit`.

Done, thx for reminding!

mrpaulblack:
@capric98 I would like to clean up the commits and rebase the PR, is that ok for you?

capric98:
> @capric98 I would like to clean up the commits and rebase the PR, is that ok for you?

It's okay, seems like I did rebase falsely before and I tried to rebase it again.
If there is still any mistakes, just feel free to fix it. Thx!

mrpaulblack:
I rebased and squashed the commits for now. I think after https://github.com/searxng/searxng/pull/1091 is merged and some small work on this PR we can get this into master :+1:

dalf:
I understand the min/max global configuration (but as you said, there is filter at all in oscar).

Do you have an use case where the admin set different limits per backend?

If a backend requires a min / max, there is no need for a configuration: we can set the limit in the code (the maintenance is done by the developers).

return42:
IMO there is no need for a limit per backend .. I think there is a need not to fire requests below the global limit is reached / see https://github.com/searxng/searxng/pull/1087#issuecomment-1096964222 .. for this I implemented 7893de5a in #1091

capric98:
> Do you have an use case where the admin set different limits per backend?

For example, if I prefer to set `autocomplete_min: 0` implemented in #1091 and it is okay because I use google backend, but this setting may cost additional requests for others using different backend and lead to a presumable ban from the backend server. Having an option to easily "overwrite" default `autocomplete_min` may be useful in this situation.

However, as you mentioned, the developers can also set the limit to the specific backend in the code. I think @return42 's implementation is pretty good, and if we need some further "overwrite" options, we could have a discussion in a new issue or under PR #1091 and the corresponding changes should be based on that PR. So close this. 🎉

PR #1091, opened 2022-04-14
[mod] client_settings: pass settings from server to JS client 

return42:
## What does this PR do?

[mod] client_settings: pass settings from server to JS client 

- The first commit implements a convenient method to pass settings from the server to the JS client.
- The second commit demonstrates how easy it is to add a setting on server side and use it in the client (JS).

The second commit is related to https://github.com/searxng/searxng/discussions/1082 and an alternative of https://github.com/searxng/searxng/pull/1087

In current master branch it is hard to add a setting and pass it to the client (JS).  ATM we share some settings and properties from server to client, but the implementation for each value is spread over the HTML templates, the render function and needs to be fetched on the JS side.  Without going into details, here is e.g. what is needed on the JS side:

https://github.com/searxng/searxng/blob/09230bfad9dea35a6ed112ea7096935d1c821bd9/searx/static/themes/simple/src/js/head/00_init.js#L17-L29

IMO that is to complicated if we need to share additional settings with the client side .. and https://github.com/searxng/searxng/pull/ shows me, that there is a need for.

## Author's checklist

To spare a additional request / response cycle (_for lees the 1kByte data_) the settings are shipped to the client in a base64 encoded JSON string in an attribute of the `base.html` template.

> the base64 encoding creates a magic value in the HTML page, it's not good for the transparency.

The alternative is to add an additional `<script src="client_settings.js">` to the `base.html` template in which the settings are transferred to client side .. but this will add an addition request / response cycle on each page load (take into mind: `client_settings.js` is not a static file / can't be cached on client side). 

Use `make run`, activate autocomplete and change the value of `autocomplete_min` .. observe how autocomplete works in your client. 

## Related issues

- https://github.com/searxng/searxng/discussions/1082
- alternative of https://github.com/searxng/searxng/pull/1087


dalf:
There are a lot of things in this PR. In this comment, I'm going to focus on the first commit (and not deal with the base64 encoding).

---

`searxng.get_setting('theme.static_path')` is complicated to read compare to `searxng.static_path` or even `searxng.settings.static_path`

Looking at the `get_client_settings` function, the values are always defined.
The fallback value of the `get_setting` function is never assigned. 

In addition, do we need a tree of preferences rather than a simple key / value mapping? The user won't see these values except looking at the HTML source code.

The choices are debatable:
 * instead of `autocomplete.provider` why not `ui.autocomplete.provider`? since the autocompletion is an preferences (value of the Preferences Python class).
 * instead of `ui.*`, why not `preferences.*`? (since there are user preferences) or just `*`?
    * `ui` : since this is the front end, this is already the UI.
 * instead of `http.method`, why not `backend.method` or `server.method`?

---

However I like the fact we can define the various preferences in one Python function.

return42:
> Looking at the `get_client_settings` function, the values are always defined.
The fallback value of the `get_setting` function is never assigned. 

ATM yes, but to have a more future proof implementation a central "get" function is IMO a better design.

> do we need a tree of preferences rather than a simple key 

not really needed if you prefer a flat key/value setting just say it.

> `searxng.get_setting('theme.static_path')` is complicated to read compare to `searxng.static_path` or even `searxng.settings.static_path`

I prefer a get_setting function; is `searxng.get_setting('static_path')` OK for you?

return42:
> Looking at the `get_client_settings` function, the values are always defined.
The fallback value of the `get_setting` function is never assigned. 

ATM yes, but to have a more future proof implementation a central "get" function is IMO a better design.

> do we need a tree of preferences rather than a simple key 

not really needed if you prefer a flat key/value setting just say it.

> `searxng.get_setting('theme.static_path')` is complicated to read compare to `searxng.static_path` or even `searxng.settings.static_path`

I prefer a get_setting function; is `searxng.get_setting('static_path')` OK for you?

dalf:
> a central "get" function is IMO a better design.

How is better compare to the dot notation?

JS pushes the dot notation even for the default values:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
* https://caniuse.com/?search=Optional%20chaining%20operator%20

---

About the base64 encoding, if the CSP header is managed by SearXNG, then we can create an nonce at each request: `<script nonce="iX+pwo1749/JrGDUOlL5GQ==">` for example.

return42:
> a central "get" function is IMO a better design.
> How is better compare to the dot notation?

Sorry I did mixed `?.` with the `?:` operator I know from C .. 

TLDR;

I am not an experienced JS developer and I don't know what is common in the JS-world:

For me there is a (fundamental) difference between the "properties of an object" and a "setup of an application" .. to set and get properties of an object directly by referencing the name of the property (in a dot notation) is common for properties .. when it comes to applications settings I prefer getter and setter function .. at least to have a layer in between the "caller" and the application settings (e.g. a 'deprecation' log can be implemented when a setting is renamed).

I don't know what is common in the JS world .. if you think there is no need to differentiate between properties and settings of an application we can implement it like you suggested: _Settings are properties of an `searxng.settings` object and the caller directly access the names of the `settings` object._

dalf:
> I prefer getter and setter function
> ... at least to have a layer in between the "caller" and the application settings (e.g. a 'deprecation' log can be implemented when a setting is renamed).

* The settings we are talking about are generated by code ( `searx.webapp.get_client_settings` ).
* The `searx.webapp.get_client_settings` relies on `searx.settings`, the values are normalized by `searx.settings_default` (the purpose of this module is already to avoid undefined values). If this is not enough we can add safe guard in `searx.webapp.get_client_settings`.
* The Python and Javascript are always sync.
* Where would the deprecation warning would go?

I'm not saying we should never use getter for the settings, but for now a simple key/value storage seems enough for now.

return42:
> I'm not saying we should never use getter for the settings, but for now a simple key/value storage seems enough for now.

I rebased and then fixed according to your remarks ..

> About the base64 encoding, if the CSP header is managed by SearXNG, then we can create an nonce at each request: `<script nonce="iX+pwo1749/JrGDUOlL5GQ==">` for example.

For know base64 is good enough I think.

return42:
FYI: rebased to solve conflict with master / no functional change since https://github.com/searxng/searxng/pull/1091#issuecomment-1106451608

return42:
FYI: rebased to solve conflict with master / no functional change since https://github.com/searxng/searxng/pull/1091#issuecomment-1106451608

@dalf can we merge this PR?

dalf:
I've was about to push the same thing with few minor changes:
* `get_client_settings` just below `get_translations`
* `<script src="{{ url_for('static', filename='js/searxng.head.min.js') }}" data-client-settings="{{ client_settings }}"></script>` <-- `data-client-settings` instead of `client_settings` ( https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes )

dalf:
Wait that's all good.

return42:
Did I misunderstood you? .. thought you will send your own PR? .. or do you wanted to push on top?

dalf:
> Did I misunderstood you? .. thought you will send your own PR? .. or do you wanted to push on top?

We can keep this PR as it is, no big deal.
Just wondering about the leaftlet.css changes

dalf:
May be it is related to this PR https://github.com/searxng/searxng/pull/1137 : 
* uglify has been updated
* but theme hasn't been rebuilt
-->
* the `[build] /static` commit contains the change of the PR 1137 too.

Just a guess

searx/webapp.py
```diff
@@ -444,17 +445,44 @@ def get_pretty_url(parsed_url: urllib.parse.ParseResult):
...
 
+def get_client_settings():
+    req_pref = request.preferences
+    return {
+        'autocomplete_provider': req_pref.get_value('autocomplete'),
+        'autocomplete_min' : get_setting('search.autocomplete_min'),
+        'http_method':  req_pref.get_value('method'),
+        'infinite_scroll': req_pref.get_value('infinite_scroll'),
+        'translations': get_translations(),
+        'search_on_category_select': req_pref.plugins.choices['searx.plugins.search_on_category_select'],
```
dalf, line 456: I forgot this one (and same for hotkeys): In the master branch the code is similar to `'searx.plugins.search_on_category_select' in request.user_plugins`

I'm not sure `req_pref.plugins.choices['searx.plugins.search_on_category_select']` behaves in the same way looking at 
https://github.com/searxng/searxng/blob/9e97a6d4ea3423ef818bf9720398f6d2e3724284/searx/webapp.py#L566-L572

return42:
> Just wondering about the leaftlet.css changes

I send https://github.com/searxng/searxng/pull/1183

dalf:
> I send https://github.com/searxng/searxng/pull/1183

Merged.

The last thing is the comment about `'search_on_category_select': req_pref.plugins.choices['searx.plugins.search_on_category_select'],`

searx/webapp.py
```diff
@@ -444,17 +445,44 @@ def get_pretty_url(parsed_url: urllib.parse.ParseResult):
...
+                encoding='utf-8',
+            )
+        ), encoding='utf-8',
+    )
+
+    # obsolete, only needed by oscar
+    kwargs['autocomplete'] = request.preferences.get_value('autocomplete')
+    kwargs['method'] = request.preferences.get_value('method')
+    kwargs['infinite_scroll'] = request.preferences.get_value('infinite_scroll')
+    kwargs['translations'] = json.dumps(get_translations(), separators=(',', ':'))
```
return42, line 477: Since we dropped oscar, we can drop these lines

dalf: It will be included in another PR nearly ready that removes the `__common__ ` directories and other oscar related stuff.

dalf:
@return42 thank you!

PR #1096, opened 2022-04-15
fix: FutureWarning from lxml

capric98:
https://github.com/searxng/searxng/blob/7e7d6f77427ad9d469ad2f2e1752bdd2b7e05156/searx/engines/yahoo.py#L145-L146
```
FutureWarning: The behavior of this method will change in future versions.
Use specific 'len(elem)' or 'elem is not None' test instead.
```
Some information about lxml: https://stackoverflow.com/a/24893800


return42 (approved):
Thanks for pointing out!

I rebased your branch and modified the commit, there was a C&P issue with the spaces: 

grafik

PR #1101, opened 2022-04-16
[enh] Allow passing headers/cookies from settings.yml

allendema:
Example:

```
   - name: example
     engine: xpath
     search_url: example.org
     headers: {'example_header': 'example_header'}
     cookies: {'safesearch': 'off'}
```
## What does this PR do?

Allows passing cookies or headers to xpath engine from ```settings.yml```.

Some engine which use xpath can not use safesearch or pass headers to receive different results.
This will add that option.

You can temporary print out params on ```engines/xpath.py```.

## Author's checklist

N/A

## Related issues
N/A

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -85,6 +85,11 @@ suggestion_xpath = ''
 cached_xpath = ''
 cached_url = ''
 
+cookies = {}
+headers = {}
+'''Some engines might offer different result based on cookies or headers.
+Possible use-case: To set safesearch cookie or header to moderate.'''
+
 paging = False
 '''Engine supports paging [True or False].'''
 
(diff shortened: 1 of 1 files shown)


searx/engines/xpath.py
```diff
@@ -166,6 +171,9 @@ def request(query, params):
         'safe_search': safe_search,
     }
 
+    params['cookies'] = cookies
+    params['headers'] = headers
```
dalf: It should be
```python
    params['cookies'].update(cookies)
    params['headers'].update(headers)
```

because the User Agent is already set.

---

https://github.com/searxng/searxng/blob/c474616642890af97ce79700baff7dd47b2406b1/searx/search/processors/online.py#L57-L61

allendema:
Thanks for your review.

PR #1102, opened 2022-04-16
[mod] Pass desired ebay domain in settings

allendema:
Make it possible to add local ebay domain.
Which must be given in ```settings.yml``` as ```base_url```

Example:
```base_url: 'https://www.ebay.de'```

## What does this PR do?

Make it possible to add local ebay domain.

More options.

Enable ebay and search with ```!ebay test```

## Author's checklist
N/A

### DIFF
--- a/searx/engines/ebay.py
+++ b/searx/engines/ebay.py
@@ -20,8 +20,10 @@ about = {
 categories = ['shopping']
 paging = True
 
-url = 'https://www.ebay.com'
-search_url = url + '/sch/i.html?_nkw={query}&_sacat={pageno}'
+# Set base_url in settings.yml in order to
+# have the desired local TLD.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: searx/engines/ebay.py [hunk 1]
```diff
@@ -20,8 +20,10 @@
 categories = ['shopping']
 paging = True
 
-url = 'https://www.ebay.com'
-search_url = url + '/sch/i.html?_nkw={query}&_sacat={pageno}'
+# Set base_url in settings.yml in order to
+# have the desired local TLD.
+base_url = None
```
dalf, line 25: Can you either set a default URL here, or set the base_url in settings.yml ?

So the engine works with the default configuration.

dalf:
Thank you!

PR #1126, opened 2022-04-19
"regions & languages" boilerplate

### LINKED ISSUE #1081: Bug: startpage engine
Branch: master
Version: 2022.04.08

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: 2 redirects, maximum: 0
  * Percentage: 70
  * Parameters: `('200', 'OK', 'www.startpage.com')`
  * File name: `searx/search/processors/online.py:105`
  * Code: `count_error(`

Error
  * Error: 1 redirects, maximum: 0
  * Percentage: 30
  * Parameters: `('200', 'OK', 'www.startpage.com')`
  * File name: `searx/search/processors/online.py:105`
  * Code: `count_error(`

Error
  * Error: searx.exceptions.SearxEngineCaptchaException
  * Percentage: 0
  * Parameters: `('CAPTCHA',)`
  * File name: `searx/engines/startpage.py:66`
  * Function: `raise_captcha`
  * Code: `raise SearxEngineCaptchaException(suspended_time=7 * 24 * 3600)`

return42:
## What does this PR do?

This PR fixes the startpage CAPTCHA and more .. for this SearXNG needs a differentiated view on the languages and regions. In the first and second commit a boilerplate to handle languages & regions has been added 

- [mod] engines_languages.json: add new type engine_properties
- [fix] replace language_support by a language/region view

In the third commit the startpage engine is fixed:

- [fix] startpage engine: language/region & time support & fix CAPTCHA

---

One reason for the often seen CAPTCHA of the startpage requests are the
incomplete requests SearXNG sends to startpage.com.  To avoid CAPTCHA we need to
send a well formed HTTP POST request with a cookie, we need to form a request
that is identical to the request build by startpage.com itself:

- in the cookie the **region** is selected
- in the POST arguments the **language** is selected

Based on the *engine_properties* boilerplate, SearXNG's startpage engine now
implements a `_fetch_engine_properties()` function to fetch regions & languages
from startpage.com.

This patch is a complete new implementation of the request() function, reversed
engineered from the startpage.com page.  The new implementation adds

- time-range support
- save-search support

to the startpage engine which has been missed in the past.

## Why is a "regions & languages" boilerplate important?

Most engines response best results if a region is selected, most often choosing
a language has a minor effect on the result list (e.g. startpage needs a region
to return results that match a language).

To summarize:

Some engines have language codes (e.g. `ca`) in their properties, some have
region codes (e.g. `ca-ES`), some have regions and languages in their properties
and other engine do not have any language or region support.

In the past we generalized *language* over all kind of engines without taking
into mind that most engines gave best result when there is a region selected.

  This *language-centric* view in SearXNG is misleading when we need
  region-codes to parameterize engine request!

The "regions & languages" boilerplate replaces the *language-centric* view by a "language / region" view.

Conclusions:

With regions we can't say any longer that a engine supports *this or that*
language, by example: when the user selects 'zh' and a engine supports only
region codes like 'zh-TW' or 'zh-CN' we do not know what results the user expects /
similar with 'en' or 'fr' when the engine needs a region tag.

- Since it is unclear what the user expects by his language selection, we can't
  assert a property that says: "supports_selected_language"

  The feature is replaced in the UI by the wider sense of "language_support",
  what stands for:

  _The engine has some kind of language support, either by a region tag or by a language tag._

- A list of "supported_languages" does not make sense when there are regions
  responsible for the result of an engine.

  The "supported_languages" has been removed from the /config URL

- The `has_language` test in the `searx/search/checker/impl.py` has been removed
  since it does not cover engines with region support.

  If there is a need for such a test we can implement new tests after all
  engines with language (region) support has been moved to the *supported
  properites* scheme (see searxng_extra/update/update_languages.py)

Start `make run` and test `!startpage foo` queries in different languages & regions.  Since autocomplete has been touched also, don't forget to test the autocompletion from startpage.

## Related issues

Closes: https://github.com/searxng/searxng/issues/1081

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -13,13 +13,14 @@ usage::
 
 import sys
 import copy
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Any
 
 from os.path import realpath, dirname
 from babel.localedata import locale_identifiers
 from searx import logger, settings
 from searx.data import ENGINES_LANGUAGES
 from searx.utils import load_module, match_language
+from searx.locales import SupportedLocales
 
 
 logger = logger.getChild('engines')
@@ -35,7 +36,7 @@ ENGINE_DEFAULT_ARGS = {
     "timeout": settings["outgoing"]["request_timeout"],
     "shortcut": "-",
     "categories": ["general"],
-    "language_support" : False,
+    "language_support": False,
     "paging": False,
     "safesearch": False,
     "time_range_support": False,
@@ -56,11 +57,13 @@ class Engine:  # pylint: disable=too-few-public-methods
     engine: str
     shortcut: str
     categories: List[str]
+    supported_languages: List[str]
+    supported_locales: SupportedLocales
+    # language support, either by selecting a region or by selecting a language
+    language_support: bool
     about: dict
     inactive: bool
     disabled: bool
-    # language support, either by selecting a region or by selecting a language
-    language_support: bool
     paging: bool
     safesearch: bool
     time_range_support: bool
@@ -141,25 +144,6 @@ def load_engine(engine_data: dict) -> Optional[Engine]:
 
     return engine
 
-def engine_properties_template():
-    """A dictionary with languages and regions to map from SearXNG' languages &
-    region tags to engine's language & region tags::
-
-      engine_properties = {
-          'type' : 'engine_properties',
-          'regions': {
-              # 'ca-ES' : <engine's region name>
-          },
-          'languages': {
-              # 'ca' : <engine's language name>
-          },
-      }
-    """
-    return {
-        'type' : 'engine_properties',
-        'regions': {},
-        'languages': {},
-    }
 
 def set_loggers(engine, engine_name):
     # set the logger for engine
@@ -197,10 +181,10 @@ def update_engine_attributes(engine: Engine, engine_data):
 def set_language_attributes(engine: Engine):
     # assign supported languages from json file
 
-    supported_properties = None
+    data: Any = None
 
     if engine.name in ENGINES_LANGUAGES:
-        supported_properties = ENGINES_LANGUAGES[engine.name]
+        data = ENGINES_LANGUAGES[engine.name]
 
     elif engine.engine in ENGINES_LANGUAGES:
         # The key of the dictionary ENGINES_LANGUAGES is the *engine name*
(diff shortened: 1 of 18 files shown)


tiekoetter (commented):
I can confirm that this fixes the startpage engine even on a previously blocked (CAPTCHA) IP.

searx/engines/__init__.py
```diff
@@ -143,6 +141,25 @@ def load_engine(engine_data: dict) -> Optional[Engine]:
 
     return engine
 
+def engine_properties_template():
```
dalf: Why the function is declared here, but used elsewhere (in the update script !)

searx/engines/__init__.py
```diff
@@ -179,56 +196,60 @@ def update_engine_attributes(engine: Engine, engine_data):
...
+        supported_properties = ENGINES_LANGUAGES[engine.engine]
+
+    if not supported_properties:
+        return
+
+    if isinstance(supported_properties, dict) and supported_properties.get('type') == 'engine_properties':
+        engine.supported_properties = supported_properties
+        engine.language_support = len(supported_properties['languages']) or len(supported_properties['regions'])
+
+    else:
```
dalf: * either make two functions (new / old format)
* either use `if: .... return` without `else:` to avoid indentation hell.

searx/engines/__init__.py
```diff
@@ -179,56 +196,60 @@ def update_engine_attributes(engine: Engine, engine_data):
...
+        # depricated: does not work for engines that do support languages
+        # based on a region.
+        engine.supported_languages = supported_properties
+        engine.language_support = len(engine.supported_languages) > 0
+
+        if hasattr(engine, 'language'):
```
dalf: This feature is useful even with the new format.

searx/webapp.py
```diff
@@ -1058,20 +1057,17 @@ def preferences():
...
-            elif safesearch and checker_test_name == 'safesearch':
+            if safesearch and checker_test_name == 'safesearch':
                 safesearch = '?'
             elif time_range_support and checker_test_name == 'time_range':
                 time_range_support = '?'
         supports[e.name] = {
-            'supports_selected_language': supports_selected_language,
             'safesearch': safesearch,
             'time_range_support': time_range_support,
+            'language_support': e.language_support
```
dalf, line 1070: In the template `preferences.html`, the column name is `Supports selected language`. That's why there is a `_is_selected_language_supported` function.

searx/engines/startpage.py
```diff
@@ -104,42 +109,99 @@ def get_sc_code(headers):
...
+    if params['pageno'] == 1:
+        args['abp'] = ['-1', '-1']
+
+    else:
+        args['page'] = params['pageno']
+        args['abp'] = '-1'
+
+    # build cookie
+    lang_homepage = 'english'
+    cookie = OrderedDict()
```
dalf, line 150: `{}` should be enough. See https://docs.python.org/3/library/collections.html#ordereddict-objects

Or even 
```python
cookie = {
  'date_time': 'world',
  ...
}
```

searxng_extra/update/update_languages.py
```diff
@@ -197,6 +263,7 @@ def filter_language_list(all_languages):
         engine_name
         for engine_name in engines.keys()
         if 'general' in engines[engine_name].categories
+        and hasattr(engines[engine_name], 'supported_languages')
         and engines[engine_name].supported_languages
```
dalf, line 263: Line 266 and 267 can be combined:
`and getattr(engines[engine_name], 'supported_languages', None)`

searx/engines/__init__.py
```diff
@@ -143,6 +141,25 @@ def load_engine(engine_data: dict) -> Optional[Engine]:
...
     return engine
 
+def engine_properties_template():
+    """A dictionary with languages and regions to map from SearXNG' languages &
+    region tags to engine's language & region tags::
+
+      engine_properties = {
+          'type' : 'engine_properties',
+          'regions': {
+              # 'ca-ES' : <engine's region name>
```
dalf: Another example would be better like `fr-BE` to avoid a political aspect.

searx/engines/__init__.py
```diff
@@ -179,56 +196,60 @@ def update_engine_attributes(engine: Engine, engine_data):
...
-            # pylint: disable=protected-access
-            lambda: engine._fetch_supported_languages(get(engine.supported_languages_url, headers=headers))
-        )
+        supported_properties = ENGINES_LANGUAGES[engine.engine]
+
+    if not supported_properties:
+        return
+
+    if isinstance(supported_properties, dict) and supported_properties.get('type') == 'engine_properties':
+        engine.supported_properties = supported_properties
```
dalf: `supported_properties` seems generic, `supported_locales` is more descriptive.

searx/templates/oscar/preferences.html
```diff
@@ -344,7 +344,6 @@ <h3>{{ _('Engines') }}</h3>
                                     <th scope="col">{{ _("Allow") }}</th>
                                     <th scope="col">{{ _("Engine name") }}</th>
                                     <th scope="col">{{ _("Shortcut") }}</th>
-                                    <th scope="col" class="col-stat">{{ _("Selected language") }}</th>
```
dalf, line 347: So the column is removed in the oscar theme?

dalf:
I've pushed a commit.

It adds a new class `searx.locale.SupportedLocales`:
* to loads / dumps values from the `searx/data/engines_languages.json`
* get the mapping from a language to a Locale, a language and a region (from the .json)

Since this class exists, the function `engine_properties_template` is not required anymore

The processor [calls the get method of this class](https://github.com/searxng/searxng/blob/7f78b69e86c2b60ba5fba7d5dfb37ad938e85dc1/searx/search/processors/abstract.py#L161-L171) for each engine, so `params["locale"]`, `params['engine_language']`, `params['engine_region']` always exist (but may be `None`).

The `all` value (the default language) is mapped a language (see https://github.com/searxng/searxng/blob/7f78b69e86c2b60ba5fba7d5dfb37ad938e85dc1/searx/engines/startpage.py#L274 ).

---

Notes:
* this commit requires some change on comments and on`./manage`.
* I have not checked neither the preferences nor the other engines (I have just checked startpage).

searx/templates/oscar/preferences.html
return42: > So the column is removed in the oscar theme?

Yes, oscar will be removed the next days ..

PR #1139, opened 2022-04-22
[fix] add back missing languages & regions (followup of PR #1071)

return42:
In PR #1071 the language catalog of dailymotion has been cleaned up, before
there had been over 7000 "languages" in the catalog.

As a side effect of this clean-up the language & region catalog in SearXNG has
been reduced [1].

This patch reduce the ``min_engines_per_lang`` from 13 to 12 to get the missed
languages back in language & region catalog of SearXNG.

[1] https://github.com/searxng/searxng/pull/1071/commits/3bb62823ec3af0e67bd2d959bec20c4791ee3bac#diff-f3f00db0f87f95b882624a192e0aac21525638af0b18c9514e765fcf1991678d

Requested-by: @tiekoetter in a Matrix chat


searx/languages.py
```diff
@@ -56,12 +59,15 @@
...
     ('sv-SE', 'Svenska', 'Sverige', 'Swedish', '\U0001f1f8\U0001f1ea'),
+    ('sw-TZ', 'Kiswahili', 'Tanzania', 'Swahili', '\U0001f1f9\U0001f1ff'),
     ('th-TH', 'ไทย', 'ไทย', 'Thai', '\U0001f1f9\U0001f1ed'),
     ('tr-TR', 'Türkçe', 'Türkiye', 'Turkish', '\U0001f1f9\U0001f1f7'),
     ('uk-UA', 'Українська', 'Україна', 'Ukrainian', '\U0001f1fa\U0001f1e6'),
+    ('vi-VN', 'Tiếng Việt', 'Việt Nam', 'Vietnamese', '\U0001f1fb\U0001f1f3'),
     ('zh', '中文', '', 'Chinese', '\U0001f310'),
     ('zh-CN', '中文', '中国', 'Chinese', '\U0001f1e8\U0001f1f3'),
-    ('zh-HK', '中文', '中國香港特別行政區', 'Chinese', '\U0001f1ed\U0001f1f0'),
+    ('zh-HK', '中文', '中國香港', 'Chinese', '\U0001f1ed\U0001f1f0'),
```
tiekoetter, line 71:
```suggestion
    ('zh-HK', '中文', '香港特別行政區', 'Chinese', '\U0001f1ed\U0001f1f0'),
```
The status of Hong Kong is denoted as "special administrative region". We should use the official name.
Source: https://en.wikipedia.org/wiki/Hongkong

tiekoetter (changes requested):

searx/languages.py
return42: This is caused by the babel update in the PR https://github.com/searxng/searxng/pull/1134 

before:

```python
>>> import babel
>>> l = babel.Locale.parse('zh-HK', sep='-')
>>> l.get_territory_name()
'中國香港特別行政區'
```

Since babel v2.10.1:

```python
>>> import babel
>>> l = babel.Locale.parse('zh-HK', sep='-')
>>> l.get_territory_name()
'中國香港'
>>> 
```

Babel uses definitions from the unicode project CLDR and has update to CLDR v41 in https://github.com/python-babel/babel/pull/853 .. 

TBH, I don't know if this upgrade is responsible for the change from the _official name_ to the _short alternative name_ :

https://github.com/unicode-org/cldr-json/blob/31cacc90735a28d8ec9d156bf7339dca2fbfcba8/cldr-json/cldr-localenames-modern/main/zh-Hant-HK/territories.json#L151-L152

To sum up: we should stay with the values from babel (CLDR), otherwise we have discussions about various regions and their names.

tiekoetter: Agree.

tiekoetter (approved):
As discussed we should use the Babel definitions to avoid potential problems with our language and region names.

PR #1141, opened 2022-04-22
[fix] since git v2.35.2 root can't read from unsafe repository

return42:
Git v2.35.2 closes a security issue, it is no longer possible that root uses a
git repo that is owned by someone else, the error message is::

    fatal: unsafe repository ('/share/darmarit.org/cache/searxng' is owned by someone else)

The fix is to run the `git diff --name-only` not as root in a sudo command.

[1] https://github.blog/2022-04-12-git-security-vulnerability-announced/

### DIFF
--- a/utils/lib_install.sh
+++ b/utils/lib_install.sh
@@ -92,6 +92,10 @@ init_SEARX_SRC_INIT_FILES(){
 
     local fname
     local msg=""
+    local _prefix=""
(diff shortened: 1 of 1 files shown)


return42:
Its hard to test this PR without an actual git .. I tested this PR on my updated archlinux system / I "approved" this PR by myself .. I am going to merge / I hope this is OK for once.

PR #1144, opened 2022-04-23
[enh] Add pagination to Brave

allendema:
Also added ```&spellcheck=1``` because now it is disabled by default, not returning any ```suggestion_xpath```.

## What does this PR do?

Add pagination support, just like the upstream engine itself.

Set ```paging: true``` and made sure spellchecks (suggestions) are returned with ```&spellcheck=1```.

!brave rfind()

## Author's checklist

See also:
https://brave.com/discussions-in-brave-search/

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1667,8 +1667,9 @@ engines:
   - name: brave
     shortcut: brave
     engine: xpath
-    paging: false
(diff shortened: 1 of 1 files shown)


dalf:
Thank you!

--

Note: the `offset=0` is not set by the HTML page on the first page, but it works.

PR #1145, opened 2022-04-27
[fix] library genesis engine: xpath and url

tiekoetter:
## What does this PR do?
Changed the xpath to use the official download mirror.

Changed the URL to libgen.fun as this is the official instance. (https://en.wikipedia.org/wiki/Library_Genesis)

Fixing the library genesis engine.

1. `make run`
2. Search for `!lg test`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -813,8 +813,8 @@ engines:
 
   - name: library genesis
     engine: xpath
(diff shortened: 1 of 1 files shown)


mTvare6:
Should there be two separate entries one for .fun and .rs, .fun is mostly down these days and both are nearly same size

PR #1153, opened 2022-04-30
[theme] peel out oscar from SearXNG development

### LINKED ISSUE #162: Remove themes

return42:
## What does this PR do?

drop oscar theme

not maintainable any more  / is in the way of innovations and costs valuable resources

## Related issues

Closes: https://github.com/searxng/searxng/issues/162


mrpaulblack:
Just an FYI: We still need a patch for `searx.space` to use simple xpath for SearXNG instances...

return42:
@dalf I assume it was github that closed this PR not you :-) .. I reopen this PR.

dalf:
Yes, sorry, the searx-space PR fixes the issue mentioned by @mrpaulblack so this PR can be merged.

mrpaulblack:
Did anybody do a quick test before the merge already? If not than I will do quick QA this evening, otherwise its ok for me to merge, thanks :D

return42:
@dalf & @mrpaulblack please review and test before merge .. there are many places to fail: e.g. the test.robot was based on the oscar theme and I had to modify the tests / just to name one major change beside removing oscar's files.

mrpaulblack (approved):
I smoke tested the simple theme on ff desktop and everything LGTM :+1: 

I also ran most of the make targets. Including all tests, the themes, py build env, the docs and docker. Everything worked as before :+1: 

The only thing I noticed is that when oscar is set as default in settings it will break the site, until simple is set. This will probably lead to some issue tickets, but should be fine, since its an easy fix :+1: (But now I spent around 2 hours testing everything out, I think its ok :D )

mrpaulblack (changes requested):
nvm I ran `make test` and this was returned by shellcheck:

```
make: shellcheck: No such file or directory
make: *** [Makefile:56: test.shell] Error 127
```

mrpaulblack (approved):
nvm its not from this PR and shellcheck also fails on master for me. I am going to to merge; LGTM :+1: 

This is a huge step for SearXNG as project so thanks @return42 for the PR and congrats to everyone for the great work :smile:  :tada:

davidcollini:
I noticed that the theme button is still in settings with only one option for simple, unless you're planning on adding different themes in the future, that should be removed thanks for developing SearXNG :)

TheFrenchGhosty:
>not maintainable any more / is in the way of innovations and costs valuable resources

Really sad to see it removed, I find `simple` worse: why is everything aligned to the left, it make is so the whole right of the screen is empty :(

davidcollini:
There should be an option to center results in the simple theme for users that want that feature

mruczek:
Is there a way to add oscar back to an instance? It was easier to use for me :)

return42:
> Is there a way to add oscar back to an instance?

You can maintain your own branch.

return42:
> and this was returned by shellcheck:
>
> ```
> make: shellcheck: No such file or directory
> make: *** [Makefile:56: test.shell] Error 127
> ```

you need to install the shellcheck command .. see https://docs.searxng.org/admin/buildhosts.html?highlight=shellcheck

masterofobzene:
I'm here just to say that oscar is much better to me. Cleaner interface, no broken features and less google-oriented (google adds a lot of unnecessary "results guessing" and icons on the screen). I really don't like the google-like "simple" theme. Also I'm experiencing text selection bugs on that theme.

I see no reason to drop oscar theme. I just works. I simply don't buy the "not maintainable any more / is in the way of innovations and costs valuable resources". Why isn't it maintainable anymore? how can a simple theme be "resource intensive" being more basic? what "innovations"? if you are planning to add more "we think for you" stuff, please drop that idea. Sometimes devs overdo the "what I think the user wants" idea. Please don't add trivial stuff to it, it just makes more unnecesary connections, gives no real advantages to end users, slows the whole thing down and it will become "resource intensive" after all. If it works, keep it simple... and the "simple" theme is just more complex in a bad way.

return42:
> I'm here just to say that oscar is much better to me.

I fully understand this opinion .. but the rest of your comment is questionable to me:

>  what "innovations"? 

just to give one example: try the image category in the simple theme and compare it to oscar.

> google adds a lot of unnecessary "results guessing" and icons on the screen

There is no difference in the content, oscar & simple show the same results and the same suggestions .. 

>  if you are planning to add more "we think for you" stuff, please drop that idea. 

Don't know what you mean .. are you sure you don't mix theme and content?

> Please don't add trivial stuff to it, it just makes more unnecesary connections, 
..
> If it works, keep it simple... and the "simple" theme is just more complex in a bad way.

Sorry but this is where you wrong / the simple theme is lightweight and fast (less connections / its exactly what you request), oscar is complex and the library it uses are no longer maintainable nor safe to use .. if you think we are wrong, you are welcome to send your contribution where you update oscar .. good luck.

mrpaulblack:
Markus already explained it really well, so thanks :D ...

You can also read more about our decision about dropping oscar here: https://github.com/searxng/searxng/discussions/1167 .

Also just as an FYI there is an ongoing effort of doing a oscar nostalgia theme on top of the simple theme here: https://github.com/searxng/searxng/issues/1169 :+1: 

(Also just as a FYI: While the simple theme may look really similar to google, its not google. If you check out the CSS, JS and HTML, we wrote it completely from scratch: https://github.com/searxng/searxng/tree/master/searx/static/themes/simple/src)

dalf:
@masterofobzene may I ask you to list two or three broken features ?

masterofobzene:
Since you asked, here are the things I don't like:

Cannot select text correctly:
realtime

Look at the cursor behavior: 
_(if you do this with double click to select a word, the whole text is selected, do this fast enough and you will lose the whole search and have to type all again)_
realtime2

Aesthetics:
Untitled

------------------------------------------------------------

For Comparison:

UntitledOK

realtimeOK

------------------------------------------------------------

REGARDING "IMAGES" SECTION:


imagesBAD

imagesOK

return42:
> Since you asked, here are the things I don't like:
> Cannot select text correctly:

@masterofobzene thanks for pointing out :+1: .. tbh this is a long standing issue we have (#794) .. I send a PR to fix .. would you like to test https://github.com/searxng/searxng/pull/1322 .. and "sorry" for the long response time on this serious issue.

masterofobzene:
> > Since you asked, here are the things I don't like:
> > Cannot select text correctly:
> 
> @masterofobzene thanks for pointing out 👍 .. tbh this is a long standing issue we have (#794) .. I send a PR to fix .. would you like to test #1322 .. and "sorry" for the long response time on this serious issue.

Thanks, its ok... not so serious just annoying. Still... I insist that the new theme "simple" is not so good (aesthetically), I could live with the icons and the google results colors and layout but it would be cool if you could move away the intrusive wiki (or others if you plan to add them). To me the best search engine is a really simple search engine with focused results. Hope there will be a "refreshed" oscar-like theme sometime.

Have a nice day @return42 .

dalf:
@masterofobzene thank you for your very detail answers. It is really helpful.

From today ( https://github.com/searxng/searxng/pull/1287 ), there is an new option "Center Alignement" .... "Oscar layout":
image

Here a screenshot (width 1000px). It is not as compact as it as in the Oscar theme, but I think it is an improvement:
image

masterofobzene:
> @masterofobzene thank you for your very detail answers. It is really helpful.
> 
> From today ( #1287 ), there is an new option "Center Alignement" .... "Oscar layout"
> 

Nice! thank you

PR #1183, opened 2022-05-07
[build] /static

return42:
## What does this PR do?

Static build of current master

    make static.build.commit

see https://github.com/searxng/searxng/pull/1091#issuecomment-1120229034


return42:
Argh .. big fail of mine .. my local npm libs had not been updated .. I think we need to revert this PR .. sorry.

PR #1186, opened 2022-05-07
Theme: remove __common__ directories

dalf:
## What does this PR do?

* Remove the `__common__` directories.
* Remove `plugins/js/*` and `plugins/css/*` files.
* Remove `searx.webapp.global_favicons` : they were used in the oscar theme.
* `/autocompleter` : the string results are always escaped.

One step closer to move the theme source to the `src` directory.

check the following item works as expected:
* the button "Submit a new issue on Github including the above information"
* the bar charts on `/stats`
* `/favicon.ico`
* `/opensearch.xml` with and without autocompletion
* autocompletion
* the "Vim-like hotkeys" preference
* the "Search on category select" preference


return42 (approved):
Wow .. thank you, its huge step forward .. I reviewed the code, made the tests you suggested plus some randomly other tests with CSP, image-proxy, some  plugins and more .. everything works like a charm .. lets merge!

dalf:
Thank you for the review!

Issue #1169, opened 2022-05-03
oscar nostalgia style for the simple theme


unixfox:
**Is your feature request related to a problem? Please describe.**
As a fan of oscar theme, I prefer oscar instead of the simple theme because everything is centered. And the colors of the dark theme are more pleasing to the eyes.

**Describe the solution you'd like**
It would be great to have an "oscar" style like for the simple theme, like a nostalgia theme so that people that are coming from the oscar can still feel at home.

Both available in dark and white theme.

In the style:
- Everything would be centered like on the oscar theme:
  <img src="https://user-images.githubusercontent.com/4016501/166516904-ba78d095-a555-4c2b-8a57-02c10aa8f01e.png" width="600">

- All the content would take as much space as possible, but without taking all the screen, like the oscar theme it would have a small border.
A good example is the GitHub wide: https://github.com/mdo/github-wide

- For the dark theme the colors would be the same as the oscar theme:
![image](https://user-images.githubusercontent.com/4016501/166517348-2ee26e04-c5fb-45e2-b1a8-667f4c818fb7.png)
No pure white texts like it's currently the case on the dark theme of the simple theme.

- (Nice to have) No icons for the links in the menu, including the preferences button.
![image](https://user-images.githubusercontent.com/4016501/166518028-86317d1d-820a-4efc-8a79-9f739291f2b1.png)

- Add back the instance name at the top left corner.
  Described in https://github.com/searxng/searxng/issues/3156

mrpaulblack:
There should also be an option to center results for simple theme on desktop layout; maybe this can be part of the oscar nostalgia theme in the future...

dalf:
An attempt to center the result page:

![image](https://user-images.githubusercontent.com/1594191/168423655-6660ee26-1883-44b0-b389-c5163671707b.png)

CSS (comment are existing CSS rules):

```css
main {
  display: flex;
  flex-direction: column;
  align-items: center;
}

#search {
  width: 100%;
  display: flex;
  flex-direction: column;
  align-items: center;
}

#results {
  /* margin-left: 10rem; */
  margin-left: 2rem;
}

.search_filters {
  /* margin-right: 0; */
  /* margin-left: 10.6rem; */
  margin-right: 2rem;
  margin-left: 2rem;
  width: 74rem;
}

#search_header {
  /* grid-template-columns: 3rem 1fr; */
  /* grid-template-areas: "logo search" "spacer categories"; */
[...]

PR #1188, opened 2022-05-08
theme: move searx/templates/simple to searx/templates

dalf:
## What does this PR do?

* move searx/templates/simple to searx/templates
* the `theme` preferences is removed.
* update the documentation (but do not changes the Makefile target names)
* [x] ~the following line has to be comment out~: https://github.com/searxng/searxng/blob/6db568bf69b7145451ff90e14da0e49fb5b6269b/tests/robot/test_webapp.py#L32

## Author's checklist

In another PR, the source files of the static directory will be moved to the `src` directory.

---

The way to change the theme is going be in settings.yml:
https://github.com/searxng/searxng/blob/6db568bf69b7145451ff90e14da0e49fb5b6269b/searx/settings.yml#L76-L79

The admin can have a second repository, then `static_path` and `templates_path` can reference this repository.


dalf:
@return42 FYI, it's a draft but the code won't change much (just fix found issue).

searx/preferences.py
```diff
@@ -292,7 +292,7 @@ def transform_form_items(self, items):
 class Preferences:
     """Validates and saves preferences to cookies"""
 
-    def __init__(self, themes: List[str], categories: List[str], engines: Dict[str, Engine], plugins: Iterable[Plugin]):
+    def __init__(self, categories: List[str], engines: Dict[str, Engine], plugins: Iterable[Plugin]):
```
return42, line 293: Should we really remove the theme option entirely. Maybe we want to support (external) themes one day?

I don't have a finial opinion, perhaps it is better after at all,  to remove entirely first and think about a support of (external) themes when we have a need for.

return42: After thinking twice .. its more clear if we remove the theme from the preferences entirely / https://github.com/searxng/searxng/pull/1153#issuecomment-1117428609

mrpaulblack: I also agree. For now we only have the simple theme and in the future we can think about external themes / a new impl. of themes as only CSS for example...

dalf: In the master branch, to add a new theme, a maintainer creates a directory in the `templates` and `static` directories ... and then the existing theme (the `simple` theme) is forgotten ; which make sense from the maintainer point of view.

See my-spot fork for example: 
* https://gitlab.e.foundation/e/infra/spot/-/commits/master/searx/templates/etheme
* https://spot.ecloud.global <-- the `theme` parameter doesn't work.

With this PR, we make sure only the maintained theme is available.

The remaining question: how my-spot deals with its own theme? (my-spot is an example here)
* either reuse the `templates` and `static` directories.
* either use two new directories, for example `searx/etheme/templates` and `searx/etheme/static`:
   * it is easier to backport commits (no conflicts)
   * the history of the `searx/etheme` directory is clean.

I'm not sure this is a good solution, but this PR:
* make the code more straightforward / easier to understand for someone new to the project.
* enforce the fact we won't support more than one theme.

---
note: my-spot is based of searx anyway.

mrpaulblack:
FYI this needs to be rebased on master before merge... Stuff like the `products.html` template is missing.

dalf:
Before this PR, I would like to implement https://github.com/searxng/searxng/issues/1169#issuecomment-1126696026 so the oscar's users can get back this layout.

PR #1191, opened 2022-05-10
[fix] PR 1186: GET method

tiekoetter:
PR https://github.com/searxng/searxng/pull/1186 prevented the GET method from being selected.

1. `make run`
2. select `GET` in preferences -> privacy -> HTTP Method


mrpaulblack (approved):
LGTM :+1: (will merge, if there is a better impl. we can do another PR IMO)

Issue #857, opened 2022-02-02
Use proxy button on result without morty


mrpaulblack:
**Is your feature request related to a problem? Please describe.**
Since there is a ongoing effort to remove morty/filtron the proxy button for each result also gets lost with morty.

**Describe the solution you'd like**
Bring back the proxy button but instead of using morty just point to a defined proxy service.

**Describe alternatives you've considered**
Use morty only as a proxy for results.

**Additional context**
See: https://github.com/paulgoio/searxng/issues/23

friedemannsommer:
I think this is a good idea, but how would that be accomplished? If the direct relation between Morty and SearXNG is removed what should the parameters for a (unknown) proxy look like? (Since `mortyurl` and `mortyhash` are rather specific)

To solve the HTTP parameter "problem", I can (currently) think of two solutions: "URL placeholders" or "Parameter mapping". There are probably other (and better) approaches and I would welcome a discussion about them.

<details>
<summary>Config examples</summary>

**Placeholder**:
```yaml
result_proxy:
  url: https://www.example.com/?url={URL}&hash={URL_HASH}
  key: !!binary "example"
  proxify_results: true
```

**Mapping**:
```yaml
result_proxy:
  url: https://www.example.com/
  key: !!binary "example"
  proxify_results: true
  proxy_parameters:
    url: "proxy_url_param"
    hash: "proxy_hash_param"
```
</details>

I would also prefer to have the option to use a separate proxy for images, instead of the integrated SearXNG proxy. This would allow to define a separate image proxy, just like a result proxy. (But this would require a separate `image_proxy` key)

<details>
<summary>Config examples</summary>

**Placeholder**:
```yaml
image_proxy:
  url: https://www.example.com/?url={URL}&hash={URL_HASH}
  key: !!binary "example"
```

**Mapping**:
```yaml
image_proxy:
  url: https://www.example.com/
  key: !!binary "example"
  proxy_parameters:
    url: "proxy_url_param"
    hash: "proxy_hash_param"
```
</details>

PR #1201, opened 2022-05-14
Draft: remove utils/morty.sh and utils/filtron.sh

dalf:
## What does this PR do?

Remove `utils/morty.sh` and `utils/filtron.sh`

TODO:
* [ ] update the documentation
* [ ] test the scripts

SearXNG can now works without morty and filtron.

N/A


return42:
FYI I work on a complete new script ``utils/searxng.sh`` / in my [searxng-install](https://github.com/return42/searxng/commits/searxng-install) branch.

My Intention was a little bit different / I will implement the new script and when it is ready I will remove all the old scripts (searx.sh, filtron.sh, morty.sh) and the generic build of the documentation.

dalf:
So I can close this PR.

Is there a quick way to update the documentation first ? So the users don't install Morty & Filtron any, so we don't have support these two repository. 

However, I don't see a lot requests about this subject so may be this can way too.

3nprob:
@dalf: While morty is not a strict dependency, what is the motivation for not removing the existing support for it? It has a lot of use-cases that are not in scope for image_proxy

mrpaulblack:
For example it does not (in its current state) run on other architectures. Also I believe its always better to keep the stack as simple as possible, so using the image_proxy makes the stack simpler... If you mean that there is the proxy button with morty there is a tracking issue here: https://github.com/searxng/searxng/issues/857


(Also nobody really worked on morty for quite some time; we already updated our docker stack to drop morty so it only makes sense to also drop it from the scripts and tbh we are a small team and providing one installation method instead of multiple onse makes support also a lot easier...)

return42:
> However, I don't see a lot requests about this subject so may be this can way too.

For the moment I will wait to remove the docs .. I haven't thought all the scenarios to its end yet, but I assume I will offer a way to upgrade old filtron & morty installations (or at least  some warnings when an old filtron and/or morty installation is detected in a SearXNG instance).

PR #1207, opened 2022-05-15
[enh] Add alexandria.org engine

### LINKED ISSUE #1200: Alexandria.org
**Working URL to the engine**

https://alexandria.org

It's an open source full-text general web search engine which I think works pretty darn well. It tends to return fewer but more relevant results than other general search engines, and sometimes may result in zero matches if the query is too specific, which in my opinion gives more reason to include it in SearXNG.

**Features of this engine**

- Non-profit and ad-free
- No tracking
- Has a public API
- Fast

They provide an API: https://github.com/alexandria-org/alexandria-api#readme

General

**Additional context**

From their about page:

> We are still at an early stage of development and running the search engine on a shoestring budget.

You may want to leave it off by default to avoid overloading it until it's in a more robust state.

allendema:
official_api_documentation: https://github.com/alexandria-org/alexandria-api/raw/master/README.md

## What does this PR do?
Adds another engine.

More engines to choose from.
Engine claims to use [https://en.wikipedia.org/wiki/Common_Crawl](https://en.wikipedia.org/wiki/Common_Crawl).

Search with ```!alx stockfish```

## Author's checklist
N/A

## Related issues

Closes #1200

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1377,6 +1377,25 @@ engines:
     about:
       website: https://wiby.me/
 
+  - name: alexandria
+    engine: json_engine
+    shortcut: alx
+    categories: general
+    paging: true
+    search_url: https://api.alexandria.org/?a=1&q={query}&p={pageno}
+    results_query: results
+    title_query: title
+    url_query: url
+    content_query: snippet
+    timeout: 1.5
+    disabled: true
+    about:
+      website: https://alexandria.org/
+      official_api_documentation: https://github.com/alexandria-org/alexandria-api/raw/master/README.md
+      use_official_api: true
+      require_api_key: false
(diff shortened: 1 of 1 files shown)


dalf:
Thank you !

As far I understand `a=1` stands for anonymous request:

https://www.alexandria.org/settings :

> The default behaviour on alexandria.org is to store the searches together with your IP address. 
> We will never sell or with intent share this data with a third party.
> We value privacy, so you can disable all tracking and use our service completely anonymously by unticking the checkboxes below.

When `Store IP address with search queries` is disabled, the `a` parameter is set to `1`.

allendema:
No problem.

That is correct, that is why it is set to on by default.

PR #1211, opened 2022-05-16
[fix] remove .github/ISSUE_TEMPLATE/discuss.md

return42:
In 2021 we did not used github-discussions, see commit message of 272c9d6b.
Theses days we use github-discussions and have a Q&A category in there [1].

[1] https://github.com/searxng/searxng/discussions/categories/q-a


tiekoetter:
I think this is a good idea because most questions are better suited to a discussion instead of an issue.

Issue #941, opened 2022-03-04
Bing returns low quality results since a few days


mrpaulblack:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

```
67fd4e139dd7a6d577c48df9af6a8aeb909189e9
2022.03.02-67fd4e13
```

**How did you install SearXNG?**

docker image

**What happened?**
Bing returns low quality results for a query.

**How To Reproduce**
search for something with bing for example `!bing time`: https://paulgo.io/search?q=%21bing%20time&language=en-US&time_range=None&safesearch=1&categories=none

**Expected behavior**
The same results as on bing frontend for the same search query: https://www.bing.com/search?q=time&pq=

**Screenshots & Logs**

**Additional context**

dalf:
FWIW, I don't have the issue on my personal instance

return42:
@dalf did you test different languages? When I use "en" I got low quality, when I use en-GB I got better results .. anyway I think bing needs a review .. I will pick up this issue as soon I have time for.

dalf:
Still on my instance:

| :fr | :de |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/1594191/156752384-4635e564-b55a-45a5-a3c3-3f071c38e3de.png) | ![image](https://user-images.githubusercontent.com/1594191/156752432-53210380-245e-42d9-b6e5-6ffb0bbb1616.png) |

But I've experienced some HTTP Connection Error

mrpaulblack:
Actually looking at https://searx.tiekoetter.com/search?q=%21bing+time&language=en-US&time_range=&safesearch=0&theme=simple it seems to be only my instance, because results for Leons instance are also fine...

[EDIT] maybe its time for me to finally setup some HTTP proxys since I also had problems with qwant a few weeks ago -_-

return42:
FYI: I tried to fix the issue but did not had luck. 

https://github.com/searxng/searxng/compare/master...return42:fix-bing

.. in the commit you can see up there on my branch I was able to clean up the implementation a bit, but the selectors for the results stayed the same (bing didn't change anything there).  My impression is rather that the request we send there is not so good ... sometimes it returns a blank page ... 

All I have right now is just a cleanup of the bing engine .. I don't know if it make sense to create a PR for the cleanup.

mrpaulblack:
Yeah the issue seems to be exclusive to my instance/VPS IP... bing also works correctly in my local dev env...

[...]

PR #1219, opened 2022-05-21
bing.py: remove redirection links

dalf:
## What does this PR do?

The purpose of this PR is to validate a possible solution to get the real URL from bing.

This PR make a quick & dirty change in `searx.network` to allow to send multiple URL in parallel (also not all exception are catched). So all redirect URL from bing are resolved at the same time. The total response time of the bing engine is similar to the google engine with this change on my laptop.

[EDIT] based on #1220 , the engine uses the `<cite>` tag when possible for short URL.

If it works on a public instance, an update of this PR should provide a more clean change of `searx.network`.

Fix the bing engine

## How to test this PR ~locally~?

can @return42 @mrpaulblack @tiekoetter @unixfox  or anyone try this PR ?

## Author's checklist

Most probably baidu can be integrated in the same way.

## Related issues

See https://github.com/searxng/searxng/discussions/1209

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -8,7 +8,8 @@
 import re
 from urllib.parse import urlencode, urlparse, parse_qs
 from lxml import html
-from searx.utils import eval_xpath, extract_text, match_language
+from searx.utils import eval_xpath, extract_text, eval_xpath_list, match_language
+from searx.network import multi_requests, Request
 
 about = {
     "website": 'https://www.bing.com',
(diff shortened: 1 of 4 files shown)


Linerly:
Tried patching this PR to my instance, it works
Screenshot
...another screenshot with another result
Another screenshot

searx/engines/bing.py
```diff
@@ -93,16 +94,34 @@ def response(resp):
...
+            else:
+                url = url_cite
+
         # append result
         results.append({'url': url, 'title': title, 'content': content})
 
+    # follow HTTP Bing URL and update the result URLs
+    for i, redirect_response in enumerate(http_get(url_to_resolve, allow_redirects=False)):
+        if not isinstance(redirect_response, Exception):
+            results[url_to_resolve_index[i]]['url'] = redirect_response.headers['location']
```
return42: > can @return42 @mrpaulblack @tiekoetter @unixfox or anyone try this PR ?

The patch itself works but I expect a lot of IO by this loop: when I query about a resource with long URLs e.g. `!bing heise.de bug :de` I see a lot of HTTP requests fired to bing.com .. IDK an other solution but I think fire one HTTP request for one link is to much .. IDK how long it takes bing blocks such requests and bans the server IP (CAPTCHA).

Since I see also other issues [1], personally I tend to drop the bing-web engine.

[1] On my server I need to repeat each query twice to get results, otherwise I get bad results similar to https://github.com/searxng/searxng/issues/941

dsh2dsh: > Since I see also other issues [1], personally I tend to drop the bing-web engine.

May be the best solution would be disable bing by default and enable duckduckgo back, because, AFAIK, DDG uses bing?

dalf:
For what it's worth, I've deployed this PR on my instance, and the bing engine works fine since then (it is enabled in all my browsers).

searx/network/__init__.py
```diff
@@ -83,11 +84,27 @@ def request(method, url, **kwargs):
...
-        raise httpx.TimeoutException('Timeout', request=None) from e
+    if isinstance(url, str):
+        future = asyncio.run_coroutine_threadsafe(network.request(method, url, **kwargs), get_loop())
+        try:
+            response = future.result(timeout)
+        except concurrent.futures.TimeoutError as e:
+            raise httpx.TimeoutException('Timeout', request=None) from e
+    elif isinstance(url, list):
+        future_list = [asyncio.run_coroutine_threadsafe(network.request(method, u, **kwargs), get_loop()) for u in url]
+        responses = []
```
return42: > For what it's worth, I've deployed this PR on my instance, and the bing engine works fine since then (it is enabled in all my browsers).

OK, but I think this part of the `network.request` where the same `**kwargs` are send to a list of URLs is not a normal use case .. as you said: "quick & dirty" :) .. I think we should not merge this patch of `network.request`.

dalf:
For ref: https://github.com/searx/searx/commit/a3ad9f9b3481024c25b570316327475d42193d53

unixfox:
> For ref: [searx/searx@a3ad9f9](https://github.com/searx/searx/commit/a3ad9f9b3481024c25b570316327475d42193d53)

I bet it's just a temporary fix and this will break in a few months. Just like when for months we chased for the user agent that would restore the functionality of Google when it broke.
See https://github.com/searx/searx/pulls?q=is%3Apr+sort%3Aupdated-desc+google+old+ui+is%3Aclosed+author%3Aunixfox

mrpaulblack (approved):
LGTM :+1: 

I tested this for 2 days on my public instance. I am not blocked by bing, URL still resolve to the correct ones and CPU and RAM usage also did not increase :+1:

PR #1245, opened 2022-05-28
[translation] add DE translation of the info pages

return42:
## What does this PR do?

DE translation of the info pages.

Start local instance `make run`, in preferences set UI language to DE and visit the info pages ("Suchbegriffe" and "Über SearXNG").


tiekoetter (approved):
LGTM.

Issue #1270, opened 2022-06-02
Search bar auto focus issue make it unusable for tiling window users


amalgame21:
Actually I use keyboard-driven qutebrowser
But I use chromium here for clear demostration since the dark theme of qutebrowser is hard to disable 
The outcome are the same

No mouse click action below, all keyboard shortcut only:
![out2](https://user-images.githubusercontent.com/93383215/171522949-8d6b741f-4382-487a-8a38-a3c51613fcd1.gif)

As you can see, when there is only one window (browser), and I perform a search, in the search result page, the cursor is automatically focused on the search bar.
When I use keyboard shortcut to open another window, and focus back to the browser, the search result become hidden by the huge auto suggestion block. 

However, If I open another window first, and go back to the browser, then perform a search afterwards, there will be no auto focus issue.

I think the "auto focus search bar" only be triggered while loading the search result with wide screen, and the focus persist when the screen become narrow, which actually is not a problem. The big problem is the search result are blocked. In qutebrowser, I have to press "Tab" key few times to get rid of the blank area and go on browsing.

oscar theme did not have this problem, but I know the theme was discontinued, very sad. 

Tiling window manager I used: i3

SearXNG installed by AUR searxng-git package.

dalf:
Possible solution: the autocompletion takes the whole screen only on mobile, not on desktop. 

Implementation:
[open-props](https://open-props.style/#media-queries) has a media query to detect touch screen : `@custom-media --touch    (hover: none) and (pointer: coarse);`

So the fix would replace the media query for this block https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/static/themes/simple/src/less/search.less#L284-L309

for this one:
```less
@media screen and (max-width: @phone) and (hover: none) and (pointer: coarse) {
    ...
}
```

so the results are only hidden on a device with a small screen and with a coarse pointer without the hover feature = on mobile.

mrpaulblack:
No I think the issue is about the focus on the input field for the search, right? so right now the autofocus is not set on mobile and always set on desktop via JS. I think a fix would be to either not set autfocus at all (which I would not like), or to only set focus to the search input field on the index page on desktop (that is IMO a must)...

https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/static/themes/simple/src/js/main/search.js#L89-#L91

dalf:
[...]

PR #1272, opened 2022-06-02
Theme: enable autofocus of #q only on the index page and when the width > 50em

dalf:
## What does this PR do?

Currently the search input is auto focused except on mobile.

See issue #1270 : on the result page, in a small window on desktop, the result are hidden.

This PR enable the autofocus of the search input only on the index page and when the width is above 50em.

---

About `.index_endpoint`, the value is set in base.html :

https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/templates/simple/base.html#L30

The `endpoint` Jinja2 variable is set to [`request.endpoint`](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Request.endpoint) in the `render` function :

https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/webapp.py#L451

So the value is
* `index_endpoint` for `/`
* `preferences_endpoint` for `/preferences`
* `results_endpoint` for `/search` (the expect value is [`search_endpoint`](https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/webapp.py#L668), the `results_endpoint` value is a legacy from the time where `/?q=%s` was not a redirect ; this does not matter in this PR).

Improve the UX.

## Related issues

Related to #1270


dalf:
Note about the change on leaftet in this PR:
* package.json specifies the version 1.7.1 : https://github.com/searxng/searxng/blob/96dc4369d535dc20178a3e7df793c8f2427d2c79/searx/static/themes/simple/package.json#L25
* it seems the last commit on the master branch was built using leaftlet 1.8.0
* this PR reverts to the version 1.7.1

mrpaulblack (approved):
LGTM (autofocus works as described and leaflet also works, thanks)

PR #1273, opened 2022-06-03
Add rate limiting per engine (using SharedDict)

MarcAbonce:
## What does this PR do?

Add a feature that limits how many outgoing requests is SearXNG going to send per engine.

This is a rewrite of #998 using our SharedDict objects instead of Redis.

Some engines are more strict than others when dealing with our traffic. For example, you can throw Wikipedia over a hundred requests per second with no problems, but other engines will throw you a CAPTCHA as soon as you send a few requests in rapid succession *\*cough\*cough\*Google\**.

1. In `settings.yml` add the following to any engine:
```yml
  # This would limit that engine to a maximum of 2 requests per minute.
  rate_limit:
    - max_requests: 2
      interval: 60
```
2. Run queries with both the rate limited engine and an engine without the setting.
1. Rate limited engine should stop sending requests after reaching limit and other engines should still work as usual.
1. Engine should be usable again after time interval has passed.
1. This is supposed to work both when running SearXNG locally or when it's running as `uwsgi` threads.

## Author's checklist

I've only tested this locally with very light traffic.

This only works by blocking an engine as a whole, it doesn't block by proxy. I tried to do that, but I couldn't find a way to find out which proxy is being used by our network client before a request. The code that picks which proxy to use is buried too deep down into `httpx` itself.

## Related issues

Rewrite of #998

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -45,6 +45,7 @@ ENGINE_DEFAULT_ARGS = {
     "using_tor_proxy": False,
     "display_error_messages": True,
     "send_accept_language_header": False,
+    "rate_limit": [{"max_requests": float('inf'), "interval": 1}],
     "tokens": [],
     "about": {},
 }
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -45,7 +45,7 @@ ENGINE_DEFAULT_ARGS = {
     "using_tor_proxy": False,
     "display_error_messages": True,
     "send_accept_language_header": False,
-    "rate_limit": [{"max_requests": float('inf'), "interval": 1}],
+    "rate_limit": [{"max_requests": None, "interval": 1}],
     "tokens": [],
     "about": {},
 }
--- a/searx/search/__init__.py
+++ b/searx/search/__init__.py
@@ -98,6 +98,10 @@ class Search:
             if request_params is None:
                 continue
 
+            # stop request if it exceeds engine's rate limit
+            if processor.exceeds_rate_limit():
+                continue
+
             counter_inc('engine', engineref.name, 'search', 'count', 'sent')
 
             # append request to list
--- a/searx/search/__init__.py
+++ b/searx/search/__init__.py
@@ -98,10 +98,6 @@ class Search:
             if request_params is None:
                 continue
 
-            # stop request if it exceeds engine's rate limit
-            if processor.exceeds_rate_limit():
-                continue
-
             counter_inc('engine', engineref.name, 'search', 'count', 'sent')
 
             # append request to list
--- a/searx/search/processors/abstract.py
+++ b/searx/search/processors/abstract.py
@@ -15,6 +15,7 @@ from searx.engines import engines
 from searx.network import get_time_for_thread, get_network
 from searx.metrics import histogram_observe, counter_inc, count_exception, count_error
 from searx.exceptions import SearxEngineAccessDeniedException, SearxEngineResponseException
+from searx.shared import storage, run_locked
 from searx.utils import get_engine_from_settings
 
 logger = logger.getChild('searx.search.processor')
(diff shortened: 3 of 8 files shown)


searx/shared/shared_uwsgi.py
```diff
@@ -17,9 +18,11 @@ def get_int(self, key: str) -> Optional[int]:
         else:
             return int.from_bytes(value, 'big')
 
-    def set_int(self, key: str, value: int):
+    def set_int(self, key: str, value: int, expire: Optional[int] = None):
         b = value.to_bytes(4, 'big')
         uwsgi.cache_update(key, b)
```
dalf: uwsgi.cache_update & uwsgi.cache_set have an expire parameter: https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html#uwsgi.cache_update

so the `self._expire` method is not required.

searx/engines/__init__.py
```diff
@@ -44,6 +44,7 @@
     "enable_http": False,
     "using_tor_proxy": False,
     "display_error_messages": True,
+    "rate_limit": [{"max_requests": float('inf'), "interval": 1}],
```
dalf: Usually undefined value are stored as a `None` instead of `float('inf')`.

searx/shared/shared_simple.py
```diff
@@ -16,14 +16,28 @@ def __init__(self):
...
-    def set_str(self, key: str, value: str):
+    def set_str(self, key: str, value: str, expire: Optional[int] = None):
         self.d[key] = value
+        if expire:
+            self._expire(key, expire)
+
+    def _expire(self, key: str, expire: int):
+        t = threading.Timer(expire, lambda k, d: d.pop(k), args=[key, self.d])
+        t.daemon = True
+        t.start()
```
dalf: It would be better to have one thread that loop over the keys and delete the expired one.

One thread per call to `set_*` is going to create a bottleneck when there are a lot of requests.

searx/search/processors/abstract.py
```diff
@@ -137,6 +138,36 @@ def extend_container_if_suspended(self, result_container):
             return True
         return False
 
+    def exceeds_rate_limit(self):
+        def check_rate_limiter(engine_name, max_requests, interval):
```
dalf: It would be better to add the function to SharedDict, something like `incr_counter(name: str, limit: int = 0, expire: int = 0)`.

The function can deal with the lock, the limit etc...

So it is easy to write a Redis implementation without changing the code in abstract.py ( the Redis implementation can be written in Lua, see the function `incr_counter` in #1047 ).

searx/search/__init__.py
```diff
@@ -97,6 +97,10 @@ def _get_requests(self):
             if request_params is None:
                 continue
 
+            # stop request if it exceeds engine's rate limit
+            if processor.exceeds_rate_limit():
```
dalf: I suggest to move this call to [extend_container_if_suspended](https://github.com/searxng/searxng/blob/c17c7dc0504daa6be675de6ac08fed03ef3246f7/searx/search/processors/abstract.py#L132-L138), so the user can get a message.

Something like this:
```python
    def extend_container_if_suspended(self, result_container):
        if self.suspended_status.is_suspended:
            result_container.add_unresponsive_engine(
                self.engine_name, self.suspended_status.suspend_reason, suspended=True
            )
            return True
        if self.exceeds_rate_limit():
            result_container.add_unresponsive_engine(
                self.engine_name, "Exceed rate limit", suspended=True
            )
            return True
        return False
```

dalf (commented):
Thank you!

We are going to switch to Redis in the future, so I have made different suggestions to allow the switch from Uwsgi to Redis easily.

> I tried to do that, but I couldn't find a way to find out which proxy is being used by our network client before a request. The code that picks which proxy to use is buried too deep down into httpx itself.

I agree, I've tried to store the suspended status in a SharedDict and failed. I think the logic / the wires has to be changed, so `searx.network` provide a HTTP client, and then the retry logic is additional layer (not buried in an internal method).

searx/shared/shared_uwsgi.py
MarcAbonce: `uwsgi.cache_update` has an `expires` argument, however the `expires` is always reset every time you call the update function.

The python function `py_uwsgi_cache_update`:
https://github.com/unbit/uwsgi/blob/d58a832c81c2c96ae0f6e72614e1cc47f4b5d332/plugins/python/uwsgi_pymodule.c#L3009-L3034

Which calls `uwsgi_cache_magic_set` with the `UWSGI_CACHE_FLAG_UPDATE` flag:
https://github.com/unbit/uwsgi/blob/master/core/cache.c#L1827

Which then builds a new struct in `uwsgi_cache_prepare_magic_update`:
https://github.com/unbit/uwsgi/blob/master/core/cache.c#L1604

Which resets the `expires` every time a cache item is updated.

MarcAbonce:
Since `SharedDict` was removed in #1856 this PR no longer works.

dalf:
I add the "area: network" label not to loose this PR. 
There is are pending needs regarding the network stack since a long time.

PR #1281, opened 2022-06-03
Theme: add a preference to center the results

dalf:
## What does this PR do?

Add a new preference "Center alignment":
image

This preferences works only for the desktop layout ( `@media screen and (min-width: @tablet) {` ).

When enabled, the results are centered instead of being in the left (or right) side of the screen.

image

image

image

image

A new CSS is added  in base.html:
```jinja2
<html class="no-js theme-{{ preferences.get_value('simple_style') or 'auto' }} center-aligment-{{ preferences.get_value('center_aligment') and 'yes' or 'no' }}" lang="{{ locale_rfc5646 }}" {% if rtl %} dir="rtl"{% endif %}>
```

There is a new LESS file : searx/static/themes/simple/src/less/style-center.less

Highly requested feature.

## Author's checklist

The result width is hard coded:
* `grid-template-columns: calc(50% - 37rem) 3rem 68rem;`
* `left: calc(50% - 48rem + 56.3rem);`

Another can add another preference to set the result width.
In improvement / a fix of this PR is to use less variables instead of hard coded values (especially since there is definitions.less)

For reference in DDG:
image

## Related issues

Related to #1169


dalf:
I'm sorry I've pushed force to update the PR: on the previous version:
* everything was centered including the other pages (preferences, info pages, etc...).
* the back to top button was not in the right place

Now the PR is ready.

return42 (changes requested):
With results on left side, the results are in-line with the seaerch-box:

grafik

in the centred layout the search-box is no longer in-line withe the result list / the entire header is moved to the right:

grafik

When images are shown in a cateory different to "images" ..

grafik


the result list on the left side is shown only partial:

grafik

dalf:
Updated:

> With results on left side, the results are in-line with the seaerch-box

image

> When images are shown in a cateory different to "images" ..

image

( style-center.less is still full of magic values, but at least if we agree on the layout that's a start )

return42:
Sorry for being pedantic .. its more a question:  does it makes sense / is it possible to move result list to the left when there is an image shown?

Or with other words .. whenever I image is shown the layout of the left result list and the image box on the right should be the same nevertheless what the value of `center_aligment` is.

ATM with `center_aligment=true` it looks like this  ..

grafik

there is a lot of *unused* space on the left side .. when disabling `center_aligment=false`, the space is better used:

grafik

dalf:
>  does it makes sense / is it possible to move result list to the left when there is an image shown?

Yes that's doable. I'm not sure what is the best option.

Detail closed:

image

Detail opened (notice the search box & the filters are still centered):

image


<details>

<summary>searx/static/themes/simple/src/less/style-center.less</summary>

Notice the `--center-screen-width` variable to adjust the width (implement another request of the issue #1169 )

```less
@media screen and (min-width: @tablet) {
  .center-aligment-yes #main_results {
    display: flex;
    flex-direction: column;
    align-items: center;

    --center-screen-width: 75rem; // override the less variable @results-width = 45rem

    #search {
      width: 100%;
      display: flex;
      flex-direction: column;
      align-items: center;
    }

    #search_header {
      grid-template-columns: calc(50% - 4.5rem - var(--center-screen-width) / 2) 3rem var(--center-screen-width);
      grid-template-areas: "na logo search" "na spacer categories";
      width: 100%;
      padding-left: 0;
      padding-right: 0;
    }

    .search_filters {
      .ltr-margin-left(0.5rem);
      width: var(--center-screen-width);
    }

    #results {
      grid-template-columns: calc(var(--center-screen-width) - 30rem) 25rem;

      &.image-detail-open {
        grid-template-columns: @results-width 25rem;
      }

      &.only_template_images,
      &.image-detail-open {
        align-self: flex-start;
      }

      &:not(.only_template_images):not(.image-detail-open) {
        .ltr-margin-left(1.5rem);

        #backToTop {
          .ltr-left(calc(50% - 30rem + var(--center-screen-width) / 2));
        }
      }

      .result .content {
        max-width: inherit;
      }
    }
  }
}
```
</detail>

return42:
> notice the search box & the filters are still centered

Its not a drawback .. its a feature :-) .. IMO it is better the search box resist on its position .. when the search box moves its position the user has to look out where it is ..

Do you like to patch this PR?

dalf:
> Do you like to patch this PR?

Done.

There is a new CSS variable `--center-screen-width` to declare the screen width and to remove the magic values (see issue https://github.com/searxng/searxng/issues/1169 ). The layout works as expect if the value is changed using the devtools (it is going to be an user preference in the future). This is half backed IMO because:
* it mixes CSS variables and LESS variables.
* the sidebar width is hardcoded to 25rem without a variable, so it is copy / paste.

return42 (approved):
> This is half backed IMO because:

Its perfect for now .. lets merge :+1:

dalf:
I would like @mrpaulblack  's opinion on the code.

@unixfox feel free to add a comment

mrpaulblack:
Is it ok if I review this tomorrow? The idea I have: I would like to have the center result option which will also make the results larger in width at the same time. This should IMO be one toggle in the preferences setting both at the same time... (having a custom field to specify the width of the results is IMO to much) :+1:

dalf:
> I would like to have the center result option which will also make the results larger in width at the same time.

does the user have to center the results to choose the page width?
or would the different combo exists?
* not center - normal
* not center - large
* center - normal
* center - large
?

mrpaulblack:
I think there should be 2 options:
* not center - normal
and
* center - large

(center result width should be not bigger then the total width of the screen when switching to tablet UI IMO, otherwise we would need another `large desktop` size or something). I think with this we would not overload the user with options (that we also need to maintain btw), while still providing something for oscar and simple fans alike...

dalf:
Ok, make sense.

In Oscar:

| @media (min-width: ... )| container width |
| - | - |
| 1280px | 1170px |
| 992px | 970px |
| 768px | 750px |

`style-center.less` can do that.

dalf:
1280px = 80rem --> only the width of 1170px is going to be used, otherwise the theme switches to the tablet mode.

https://github.com/searxng/searxng/blob/c17c7dc0504daa6be675de6ac08fed03ef3246f7/searx/static/themes/simple/src/less/definitions.less#L257-L262

( I think what Oscar's users would like is to be able to disable the tablet mode )

dalf:
[EDIT] the style-oscar has replaced the branch of this PR.

Original comment below

---

I've created another branch (*) : where the tablet layout is disabled (only when "Center alignment" is enabled)

Screenshoots with a with of 1000px (the Tor Browser default width):

| style-oscar branch | Oscar | master & style-center branches |
| - | - | - |
| image |  image | image |


---

* https://github.com/dalf/searxng/tree/style-oscar
* https://github.com/searxng/searxng/commit/fce1130e0cf113137d57a9902d805646b829d996

unixfox:
Thank you @dalf, it's already a good step towards my wishes from #1169!

Now that I discovered the DDG settings page (https://duckduckgo.com/settings), I really like the way the search bar is centered like you also shown in https://github.com/searxng/searxng/issues/1169#issuecomment-1126696026

image

-------

> I've created another branch (*) : where the tablet layout is disabled (only when "Center alignment" is enabled)
> 
> Screenshoots with a with of 1000px (the Tor Browser default width):
> style-oscar branch 	Oscar 	master & style-center branches
> image 	image 	image
> 
>     * [https://github.com/dalf/searxng/tree/style-oscar](https://github.com/dalf/searxng/tree/style-oscar?rgh-link-date=2022-06-04T08%3A16%3A01Z)
> 
>     * [13a8930](https://github.com/searxng/searxng/commit/13a893022cf3c6ada4180b7fba6da26bdc4b2472)

I do agree with that, a screen of 1000px is already small so if we can have as much space taken by the texts as possible (like it's already possible on oscar) that would be great.

searx/templates/simple/preferences.html
```diff
@@ -213,6 +213,16 @@ <h1>{{ _('Preferences') }}</h1>
...
     </fieldset>
+    <fieldset>
+      <legend id="pref_center_aligment">{{ _('Center Alignment') }}</legend>
+      <p class="value">
+        <select name="center_aligment" aria-labelledby="pref_center_aligment">
+            <option value="1" {% if preferences.get_value('center_aligment') %}selected="selected"{% endif %}>{{ _('On') }}</option>
+            <option value="0" {% if not preferences.get_value('center_aligment') %}selected="selected"{% endif %}>{{ _('Off')}}</option>
+        </select>
+      </p>
+      <div class="description">{{ _('Displays results in the center of the page (Oscar layout).') }}</div>
```
dalf, line 224: The description contains a reference to the Oscar layout.

searx/static/themes/simple/src/less/definitions.less
```diff
@@ -243,6 +243,7 @@
 
 /// General Size
 @results-width: 45rem;
+@results-sidebar-width: 25rem;
```
dalf, line 246: New variable, so style-center.less can reuse this value

dalf:
The PR has been updated:
* starting from the @phone screen width, the layout is similar to the Oscar theme.
* most of the magic values have been replaced by variables.
* the description in the preferences makes a reference to the Oscar theme.

> I really like the way the search bar is centered like you also shown in https://github.com/searxng/searxng/issues/1169#issuecomment-1126696026

I'm not sure how to center the filters (language, date, safe search), so for now the PR keeps the Oscar layout.

---

For reference https://github.com/dalf/searxng/tree/style-center-original contains the original code (which doesn't keep the Oscar layout for screen width below `@tablet` )

unixfox:
IMO the images section should be like when "centered alignment" is not enabled but with the menu centered.

image

This feels weird.

dalf:
> This feels weird.

that was not intended.

I've updated the PR, it should be working now.

unixfox (approved):

amalgame21:
Minor question
center_aligment, seems missing "n"?
Just figure it out from the setting.yml file

return42:
> center_aligment, seems missing "n"?

@amalgame21 its not clear to me what you mean by / this is a YAML file and the value of this option is a boolean (`true` or `false`).

But your comment reminds me that the documentation of this option is missed .. thanks :-) .. I created https://github.com/searxng/searxng/issues/1420

dalf:
@return42 , it's everywhere:

https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/settings.yml#L95

https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/settings_defaults.py#L188

https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/templates/simple/base.html#L2

It should be "center_alignment" with a `n`.

We can fix the variable names, the settings name is more tricky: administrator will have hard time to figure out the typo.

unixfox:
> @return42 , it's everywhere:
> 
> https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/settings.yml#L95
> 
> https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/settings_defaults.py#L188
> 
> https://github.com/searxng/searxng/blob/4ef62637256d4b2966ff7f0be0051e02d545f756/searx/templates/simple/base.html#L2
> 
> It should be "center_alignment" with a `n`.
> 
> We can fix the variable names, the settings name is more tricky: administrator will have hard time to figure out the typo.

I think the feature is pretty new that it's ok to fix this typo. This will affect a very small amount of people.

return42:
> I think the feature is pretty new that it's ok to fix this typo. This will affect a very small amount of people.

Ooops, sorry I haven't seen the typo `center_aligment` needs to be fixed --> `center_alignment`

@dalf do you have time to prepare a patch? .. otherwise I will take the time today ..

dalf:
> do you have time to prepare a patch?

I won't have time today.

PR #1283, opened 2022-06-05
Add helm chart for SearXNG

### LINKED ISSUE #1256: Kubernetes Helm Chart for SearxNG

I'm using the Kubernetes Helm Chart from here: https://github.com/k8s-at-home/charts/tree/master/charts/stable/searx
searxNG was not usable with this chart by only changing the image to searxng/searxng.

A Kubernetes Helm Chart for SearxNG woud be great :-)

Use a Kubernetes Deployment

**Additional context**

@unixfox according to the chat in #searxng:matrix.org I should ping you :-)

unixfox:
## What does this PR do?

This adds a helm chart for easily installing SearXNG on Kubernetes.

Make things easy for Kubernetes users.

1. Install a Kubernetes cluster. You can use k3s: https://k3s.io/
2. Install Helm: https://helm.sh/docs/intro/install/
3. Install the helm chart by following https://github.com/searxng/searxng/blob/c52fb8a65787e50d0ca7bf0d739d6fd93a5eca12/kubernetes/chart/README.md#installing-the-chart
4. The helm chart should be deployed, you can test it by following the NOTES at the end of the output of the command.

## Author's checklist

The chart is located in kubernetes/chart because it's commonly placed like this in most of the projects that have an helm chart. 

I didn't write any documentation for docs.searxng.org because I'm a bit lazy (and I don't really like reST), but feel free to add a page that points towards the README file of the helm chart.

## Related issues

Closes #1256

### DIFF
--- a/kubernetes/chart/.helmignore
+++ b/kubernetes/chart/.helmignore
@@ -1,0 +1,26 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
+# OWNERS file for Kubernetes
+OWNERS
+# helm-docs templates
+*.gotmpl
--- a/kubernetes/chart/Chart.yaml
+++ b/kubernetes/chart/Chart.yaml
@@ -1,0 +1,27 @@
+apiVersion: v2
+appVersion: latest
+description: SearXNG is a free internet metasearch engine which aggregates results from various search services and databases.
+name: searxng
+version: 1.0.0
+kubeVersion: ">=1.16.0-0"
+keywords:
+  - searx
+  - searxng
+  - search engine
+  - privacy
+icon: https://github.com/searxng/searxng/blob/master/searx/static/themes/simple/img/favicon.png?raw=true
+sources:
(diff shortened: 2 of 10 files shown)


mrpaulblack:
Would you think it makes sense to add this to searxng/searxng-docker (we could even rename it to something more generic for example) or to create a seperate repo for this?

This way it would be easier to discover by new users of the project, it would also mean that people dont need to git clone the entire SearXNG core repo and it would also separate SearXNG core development from the helm charts in a way...

unixfox:
Separating repository would be a great idea because I would be able to do some GitHub actions thing in order to push the helm chart to a repository served on HTTP(S).

But the downside is that people will have to dig a bit more in order to find the chart, but this could be solved by advertising it directly in the README of the searxng main repository. Something that I don't understand why we are not already advertising the Docker image too :).

mrpaulblack:
yeah agreed a install section in the readme could be a great thing to advertise docker, helm and the script installation; If you create the helm repo we can also pin it like the docker repo in the org :+1:

dalf:
> Separating repository would be a great idea because I would be able to do some GitHub actions thing in order to push the helm chart to a repository served on HTTP(S).

When to run it? After each commit on the master branch of searxng/searxng ?

unixfox:
> > Separating repository would be a great idea because I would be able to do some GitHub actions thing in order to push the helm chart to a repository served on HTTP(S).
> 
> When to run it? After each commit on the master branch of searxng/searxng ?
> 

The github actions workflow is just to update the chart files to our helm repository when I push some changes to it.

This has nothing to do with the main searxng/searxng repo

dalf:
So I agree, we can have another repository for the helm chart, and then add link to it in README and the documentation.

unixfox:
Migrated to https://github.com/searxng/searxng-helm-chart/tree/main/searxng

PR #1295, opened 2022-06-07
[mod] source code url: GIT_URL + '/tree/' + GIT_BRANCH

return42:
The link 'Source code' in the footer should link to the branch URL:

    GIT_URL + '/tree/' + GIT_BRANCH

This URL fits at least for github and gitlab.

BTW clear the naming and replace:

- `searx_git_url` and `searx_git_branch` by `GIT_URL` and `GIT_BRANCH`
- `searx_version` by `searxng_version`

----

To test create a clone of this PR:

```
$ cd /tmp
$ git clone https://github.com/return42/searxng/
$ cd searxng/
$ git checkout git-url-branch
$ make buildenv
$ git diff
```

```diff
diff --git a/utils/brand.env b/utils/brand.env
index 31afce53..6fd66ebe 100644
--- a/utils/brand.env
+++ b/utils/brand.env
@@ -1,5 +1,5 @@
 export SEARXNG_URL=''
 export SEARXNG_PORT='8888'
 export SEARXNG_BIND_ADDRESS='127.0.0.1'
-export GIT_URL='https://github.com/searxng/searxng'
-export GIT_BRANCH='master'
+export GIT_URL='https://github.com/return42/searxng/'
+export GIT_BRANCH='git-url-branch'
```

```
make run
```

The source code link in the footer should link to this branch https://github.com/return42/searxng/tree/git-url-branch

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -448,8 +448,9 @@ def render(template_name: str, **kwargs):
     # values from settings
     kwargs['search_formats'] = [x for x in settings['search']['formats'] if x != 'html']
     kwargs['instance_name'] = get_setting('general.instance_name')
-    kwargs['searx_version'] = VERSION_STRING
-    kwargs['searx_git_url'] = GIT_URL
+    kwargs['searxng_version'] = VERSION_STRING
+    kwargs['GIT_URL'] = GIT_URL
+    kwargs['GIT_BRANCH'] = GIT_BRANCH  # url: GIT_URL + '/tree/' + GIT_BRANCH
     kwargs['get_setting'] = get_setting
     kwargs['get_pretty_url'] = get_pretty_url
 
@@ -1241,7 +1242,6 @@ def stats():
         engine_stats = engine_stats,
         engine_reliabilities = engine_reliabilities,
         selected_engine_name = selected_engine_name,
-        searx_git_branch = GIT_BRANCH,
         # fmt: on
     )
 
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -450,7 +450,8 @@ def render(template_name: str, **kwargs):
     kwargs['instance_name'] = get_setting('general.instance_name')
     kwargs['searxng_version'] = VERSION_STRING
     kwargs['GIT_URL'] = GIT_URL
-    kwargs['GIT_BRANCH'] = GIT_BRANCH  # url: GIT_URL + '/tree/' + GIT_BRANCH
+    kwargs['GIT_BRANCH'] = GIT_BRANCH
+    kwargs['git_url_format'] = get_setting('general.git_url_format')
     kwargs['get_setting'] = get_setting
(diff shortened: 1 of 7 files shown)


searx/webapp.py
```diff
@@ -1273,7 +1274,8 @@ def get_key(engine_stat):
         engine_stats = engine_stats,
         engine_reliabilities = engine_reliabilities,
         selected_engine_name = selected_engine_name,
-        searx_git_branch = GIT_BRANCH,
+        GIT_URL = GIT_URL,
+        GIT_BRANCH = GIT_BRANCH,
```
dalf: I haven't tested, but it seems these two variables are defined in the `render`  function, so there are not needed here ?

return42: > I haven't tested, but it seems these two variables are defined in the `render`  function, so there are not needed here ?

Oops .. thanks for pointing out .. [fixed](https://github.com/searxng/searxng/compare/7b43d96fc69ab0e273bbfd9c842b4c6239eb22c0..e34782e0b9b1776b95ed418bfc850a552ccd2909)

dalf:
Two things:
* There is a link to the source code in the `searx/infopage/*/about.md` pages.
* More importantly, it works only with GitHub and GitLab services
   * actually it is `/-/tree` for GitLab, but `/tree/` works too, see for example https://gitlab.e.foundation/e/infra/spot/-/tree/unit-conversions ).
   * It does not work with Gitea (error 404): the prefix is `src/branch/` not `/tree`.
   * I haven't try other services.

Basically with this PR, we take the risk to provide a broken link that can't be fixed by the admin (even for the master and main branches).

I don't know what is the best option.

At first sight I would leave the GIT_URL as it is, and display the branch name is additional text area (`tooltip="...."` if the branch is neither master nor main).

return42:
> Two things:
> * There is a link to the source code in the `searx/infopage/*/about.md` pages.

Thanks for pointing out .. I fixed it.

> * More importantly, it works only with GitHub and GitLab services

Yes, this is really a drawback .. what do you think about adding a `git_url_format` option to the settings.yml like I have done in my last commit 830224c26:

https://github.com/searxng/searxng/commit/830224c26138b56419f6df00b618b0bd1b72662d#diff-6a82f4161887c10ef8e86cf63056ae5832896914bd0240628ec076cd075c4fff

dalf:
> what do you think about adding a git_url_format option to the settings.yml like I have done in my last commit 

It can fail silently: the link is going to be broken, so it won't be compliant with the AGPL license silently.
The ways I see:
* add the branch as additional information.
* keep only one URL : the one which includes the branch. The problem moves to searx-stats2 which have to find the git URL to close the repository.

return42:
> The problem moves to searx-stats2 which have to find the git URL to close the repository.

return42:
> keep only one URL : the one which includes the branch. 

OK, then I think we can close this approach.

PR #1306, opened 2022-06-10
[fix] move locale code from webapp.py to locales.py and fix #1303

### LINKED ISSUE #1303: Bug: Chinese language header problem
I'm using a public instance. All instances behaves the same way.

If you visit searxng in Hong Kong, the interface language defaults to Chinese Simplified, whereas HK people usually use Chinese Traditional or just English.

**Expected behavior**
Interface should be Chinese Traditional or just English (Generally it's ok to use the Taiwan version of the translation for Hong Kong.)

**Screenshots & Logs**
My browser's Accept-Language header is zh-HK,zh-Hant;q=0.9

Reference: https://github.com/searxng/searxng/issues/540#issuecomment-1152106517

return42:
To improve modularization this patch:

- moves *locale* related implementation from the webapp.py (application code) to the   locale.py module.

- The initialization of the locales is now done in the application (webapp) and   is no longer done while importing searx.locales.

In the searx.locales module a new dictionary named `LOCALE_BEST_MATCH` has been added.  In this dictionary we can map languages without a translation to languages we have a translation of.

To fix #1303 zh-HK has been mapped to zh-Hant-TW (we do not need additional translations of traditional Chinese).

To test start SearXNG by `make run` and check new locale `zh-HK` (e.g. by setting `Accept-Language: zh-HK,zh-Hant;q=0.9` header) .. the interface language should be traditional Chinese

Closes: https://github.com/searxng/searxng/issues/1303

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -40,11 +40,8 @@ from flask import (
     send_from_directory,
 )
 from flask.wrappers import Response
-from flask.ctx import has_request_context
 from flask.json import jsonify
 
-from babel.support import Translations
-import flask_babel
 from flask_babel import (
     Babel,
     gettext,
@@ -114,11 +111,16 @@ from searx.metrics import (
 )
 from searx.flaskfix import patch_application
 
-# renaming names from searx imports ...
+from searx.locales import (
+    LOCALE_NAMES,
+    RTL_LOCALES,
+    localeselector,
+    locales_initialize,
+)
 
+# renaming names from searx imports ...
 from searx.autocomplete import search_autocomplete, backends as autocomplete_backends
 from searx.languages import language_codes as languages
-from searx.locales import LOCALE_NAMES, RTL_LOCALES
 from searx.search import SearchWithPlugins, initialize as search_initialize
 from searx.network import stream as http_stream, set_context_network_name
 from searx.search.checker import get_result as checker_get_result
@@ -148,7 +150,6 @@ STATS_SORT_PARAMETERS = {
     'time': (False, 'total', 0),
     'reliability': (False, 'reliability', 100),
 }
-_INFO_PAGES = infopage.InfoPageSet()
 
 # Flask app
 app = Flask(__name__, static_folder=settings['ui']['static_path'], template_folder=templates_path)
@@ -192,10 +193,6 @@ exception_classname_to_text = {
 }
 
 
-# monkey patch for flask_babel.get_translations
-_flask_babel_get_translations = flask_babel.get_translations
-
-
 class ExtendedRequest(flask.Request):
     """This class is never initialized and only used for type checking."""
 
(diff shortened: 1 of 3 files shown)


searx/webapp.py
```diff
@@ -114,16 +111,22 @@
...
+# renaming names from searx imports ...
 from searx.autocomplete import search_autocomplete, backends as autocomplete_backends
 from searx.languages import language_codes as languages
-from searx.locales import LOCALE_NAMES, RTL_LOCALES
 from searx.search import SearchWithPlugins, initialize as search_initialize
 from searx.network import stream as http_stream, set_context_network_name
 from searx.search.checker import get_result as checker_get_result
 
 logger = logger.getChild('webapp')
+initialize_locales()
```
dalf: I would move this line just between lines 1417 and 1418 https://github.com/searxng/searxng/blob/59ef9b9287f1beda12f7b9a20b93cbc378a22bac/searx/webapp.py#L1417-L1419


so all the initialization calls are grouped in one place (there is no issue for delaying the initialization according, at least when I've tried).

---

nitpicking: rename the function to `locales_initialize` to fit the two other function names.

dalf (commented):
except one minor comment, LGTM.

At least wikidata *seems* to return traditional Chinese when the search language is "中文 (中國香港)", but I'm not Chinese reader.

dalf:
~I remember to have try "make test" and it worked.~

Ok... now even if `locales_initialize()` is at the top of the file, it doesn't work.

[EDIT] for `make test.robot`, the server log shows:
```
INFO    werkzeug                      : 127.0.0.1 - - [12/Jun/2022 10:35:27] "GET /info/en/about HTTP/1.1" 404 -
```

searx/webapp.py
return42: > I would move this line just between lines 1417 and 1418

Done .. but now we have [this issue](https://github.com/searxng/searxng/runs/6848643803?check_suite_focus=true#step:7:163) in the robot test .. TBH the inititialization of the app is not clear to me / do you have an idea what the problem is in robot-test?

> At least wikidata seems to return traditional Chinese when the search language is "中文 (中國香港)", but I'm not Chinese reader.

FYI: #1303 is about UI language .. and this PR adds a dictionary to map *best match*:

> In the searx.locales module a new dictionary named `LOCALE_BEST_MATCH` has been added. In this dictionary we can map languages without a translation to languages we have a translation of.

return42:
> Ok... now even if `locales_initialize()` is at the top of the file, it doesn't work.

Hm .. my experience is different with this patch below `make test.robot` runs without an issue.

```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index 0469c8149..67fb3995e 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -126,6 +126,7 @@ from searx.network import stream as http_stream, set_context_network_name
 from searx.search.checker import get_result as checker_get_result
 
 logger = logger.getChild('webapp')
+locales_initialize()
 
 # check secret_key
 if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
@@ -1384,7 +1385,6 @@ werkzeug_reloader = flask_run_development or (searx_debug and __name__ == "__mai
 
 # initialize the engines except on the first run of the werkzeug server.
 if not werkzeug_reloader or (werkzeug_reloader and os.environ.get("WERKZEUG_RUN_MAIN") == "true"):
-    locales_initialize()
     plugin_initialize(app)
     search_initialize(enable_checker=True, check_network=True, enable_metrics=settings['general']['enable_metrics'])
```

> I remember to have try "make test" and it worked. 

If you implement something, please push it on top of my PR .. a "WIP" commit message is a good choice / we can squash the WIP commits later before merge.

dalf:
> Hm .. my experience is different with this patch below make test.robot runs without an issue.

Yes, I moved `locales_initialize` on the top but below `_INFO_PAGES = infopage.InfoPageSet()`, that's why.

Possible fix:

```python
# initialize the engines except on the first run of the werkzeug server.
if not werkzeug_reloader or (werkzeug_reloader and os.environ.get("WERKZEUG_RUN_MAIN") == "true"):
    locales_initialize()
    _INFO_PAGES = infopage.InfoPageSet()
    plugin_initialize(app)
    search_initialize(enable_checker=True, check_network=True, enable_metrics=settings['general']['enable_metrics'])
```

because the constructor of `searx.infopage.InfoPageSet`  reads `searx.locale.LOCALE_NAMES`.

return42:
> because the constructor of `searx.infopage.InfoPageSet`  reads `searx.locale.LOCALE_NAMES`.

Thanks :+1: ... this reminds me that we need to redesign the application to have one place where the initialization takes place ;-)

return42:
Sorry .. pressed the wrong button :-)

dalf:
no problem :-)

> we need to redesign the application to have one place where the initialization takes place 

at least initialization code in webapp.py is grouped in one place.

I was sure to have run `make test` in my first review, it seems it was not the case, sorry.

return42:
> at least initialization code in webapp.py is grouped in one place.

"in one module?" :-) .. just joking .. TBH I think the problem already starts with the import of the searx module, this module imports settings module and **loads** the settings:

https://github.com/searxng/searxng/blob/2455f1d06a3b05e4eb86be274c308160262e4a87/searx/__init__.py#L25

The settings are the core of an application .. by this the settings should only be loaded in application's initialization procedure and not when the top most namespace is imported.

Issue #1304, opened 2022-06-10
Bug: text overlap


BernieHuang2008:
![Screenshot_20220610.png](https://user-images.githubusercontent.com/88757735/173038106-4e0c1d81-315b-4a29-81f0-722ed1fd5c2a.png)
I just found some results like this.

PS: I'm using the highest version, and I found it in almost all the version.

mrpaulblack:
Hi @BernieHuang2008 ,
can you please share your device and screen resolution? IF your browser does not use your full screen resolution, how big is your browser window?

BernieHuang2008:
Hi , thanks for your reply to this issue.
My screen resolution is 2560x1600 , and I'm using my Android pad (Android 10).
and I'm using Microsoft Edge that time .
Hope these will help you.
:)

BernieHuang2008:
Hi~
If I found the right file, then I think I solved this bug.
Please look for [PR#1312](https://github.com/searxng/searxng/pull/1312).

return42:
@BernieHuang2008 does the issue persist?

BernieHuang2008:
@return42 

Sorry, but it seems that the bug is still exist ...

![Screenshot_20230906_175940_com.microsoft.emmx.png](https://github.com/searxng/searxng/assets/88757735/49febbe2-1451-40dd-b213-c7c5ef7afc5c)

I'm still interested in this problem, so if you need my help, ask any time please.

dalf:
Which browser do you use?

BernieHuang2008:
Still MS Edge, version 116.0.1938.58, on Android phone.

return42:
Both are flex containers 

![image](https://github.com/searxng/searxng/assets/554536/effeada2-7444-43db-99d2-4c77a1383b5a)

![image](https://github.com/searxng/searxng/assets/554536/1be9a542-1fe4-41d5-b3cb-24e3cdcfe9cf)

CSS Flexible Box Layout Module is Fully Supported on Microsoft Edge, seems Edge for mobile has issues with / I can't find Edge for mobile on any compatibility list .. not sure what we can do.

@dalf do you have a clue? .. otherwise I will close this issue as related to Edge for mobile.

BernieHuang2008:
I think you can close it. If I found any valuable clues, I will open a new issue.

Thank you for all your help!

dalf:
I can't reproduce the bug on Edge Mobile 116.0.1938.64 (Android 13).
[...]

PR #1309, opened 2022-06-10
Sove the issue #1304(text overlap)

BernieHuang2008:
I think this PR will solve the issue#1304.

## What does this PR do?

Solve the [issue#1304](https://github.com/searxng/searxng/issues/1304).

I think this PR solve the [issue#1304](https://github.com/searxng/searxng/issues/1304), if I found the right file.
It worked completely well on my computer, like this:

Before Change
Before change, it looks like this 👆

After Change
After change, it looks likes this 👆

I just test it in my browser.

Find the `<span class="url_i2">` and change the `css` code

```css
.result .url_wrapper .url_o2 .url_i2 {
	float: right;
}
```

to

```css
.result .url_wrapper .url_o2 .url_i2 {
	float: right;
	overflow: hidden;
	text-overflow: ellipsis;
	max-width: 100%;
}
```

Or just replace the code to 
## Related issues
I think this PR will solve the [issue#1304](https://github.com/searxng/searxng/issues/1304).


dalf:
Thank you for your contributions.

The git repository contains the source code & the compiled code. Your change is going to be erased on the next change to the theme.

1. You should make the change here:
https://github.com/searxng/searxng/blob/59ef9b9287f1beda12f7b9a20b93cbc378a22bac/searx/static/themes/simple/src/less/style.less#L257-L259
2. Then commit
3. Then call `make static.build.commit` : it will build the file that you have change and create a commit
4. Then you can push force your change to your branch

See https://docs.searxng.org/dev/quickstart.html?highlight=static%20build%20commit for more details.


---

> I just test it in my browser.

So Microsoft Edge on Android ?

BernieHuang2008:
Thanks, @dalf .
With your guide, I created a new PR, [#1312 ](https://github.com/searxng/searxng/pull/1312).
Hope you will check it, thanks.

PR #1312, opened 2022-06-11
Solve the issue#1304(text overlap)

BernieHuang2008:
Solve issue#1304

## What does this PR do?

Solve the [issue#1304(text overlap)](https://github.com/searxng/searxng/issues/1304) by add code in
https://github.com/searxng/searxng/blob/b22c149478871547d31f75ae0a399382d8e8de64/searx/static/themes/simple/src/less/style.less#L259-L261

This PR solve the [issue#1304](https://github.com/searxng/searxng/issues/1304) .
It worked completely well on my computer, like this:

Before Change
Before change, it looks like this 👆

After Change
After change, it looks likes this 👆

By change the `css` code from

```css
.url_i2 {
	float: right;
}
```

to

```css
.url_i2 {
	float: right;
	overflow: hidden;
	text-overflow: ellipsis;
	max-width: 100%;
}
```

## Related issues
This PR will solve the [issue#1304](https://github.com/searxng/searxng/issues/1304).

### DIFF
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -256,6 +256,9 @@ article[data-vim-selected].category-social {
 
       .url_i2 {
         float: right;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	max-width: 100%;
       }
     }
   }
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -256,9 +256,9 @@ article[data-vim-selected].category-social {
 
       .url_i2 {
         float: right;
-	overflow: hidden;
-	text-overflow: ellipsis;
-	max-width: 100%;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        max-width: 100%;
       }
     }
   }


BernieHuang2008:
Hello, can anyone tell me what's happen?

BernieHuang2008:
Why I can't pass the check?

tiekoetter:
@BernieHuang2008 You have some indentation errors.

```
Error: unner/work/searxng/searxng/searx/static/themes/simple/src/less/style.less:259:2: Expected indentation of 8 spaces (indentation) [error]
Error: unner/work/searxng/searxng/searx/static/themes/simple/src/less/style.less:260:2: Expected indentation of 8 spaces (indentation) [error]
Error: unner/work/searxng/searxng/searx/static/themes/simple/src/less/style.less:261:2: Expected indentation of 8 spaces (indentation) [error]
```

BernieHuang2008:
Oh,thanks @tiekoetter .

mrpaulblack:
Hi @BernieHuang2008 ,
First of all sorry for the long wait. This feature is supposed to cut off the text in the middle and not on the right. So if the screen is to small you should be able to see the domain on the left and the end of the path on right. The middle part should be cut off. Which works correctly. The only problem is that there is no ellipsis in the middle part (on the left of the path). I will close this PR for now, but if you have a solution to get the ellipsis in the middle between the domain and path part of the URL please open another PR.

Again sorry for the long wait with this and I hope this does not discourage you from contributing to SearXNG and other OSS projects in the future :+1: :D

dalf:
Currently the URL is split into `[host, path]`, in the `path` the `/` are replaced by ` › ` see:
https://github.com/searxng/searxng/blob/2bfb269f0a2a2c35d3e4752cb839c38ddcaf805d/searx/webapp.py#L383-L387

What can be done is if `path` is not empty:
* to add a ` › ` at the end of `host`
* to remove the ` › ` at the beginning of `path`.

So when there is a text overlap, the `path` disappear below ` › ` instead of the `host`.
I think it would improvement.

Issue #774, opened 2022-01-17
[question] activate google's use_mobile_ui by default?


return42:
From time to time we hear that SearXNG admins do not get google results, most often we recommend to activate `use_mobile_ui: true` .. may be it is time, we set the default to this value? ...

https://github.com/searxng/searxng/blob/96ab5e57ff92aba4ab0fedeb4e5fbfa09e9fc2e2/searx/settings.yml#L623-L627

return42:
@unixfox @tiekoetter @mrpaulblack do you use google mobile UI .. should we change the default to use it?

You can vote with thumbs up/down.

unixfox:
no it should stay off by default or google will notice it and block this method. google on searx works well locally without any workaround so there is no point in having use_mobile_ui by default

unixfox:
Now that I think again about it, maybe we could print in the logs a help message explaining that the user may need to enable `use_mobile_ui`?

If we find a too many requests error, then we print the help message.

samsapti:
> Now that I think again about it, maybe we could print in the logs a help message explaining that the user may need to enable `use_mobile_ui`?
> 
> If we find a too many requests error, then we print the help message.

@unixfox I think that's a good idea.

PR #1313, opened 2022-06-11
add "warn" message to notify the admin that use_mobile_ui parameter exist

unixfox:
## What does this PR do?

This notifies the user of the existence of the `use_mobile_ui` parameter.

Too many people don't know about this parameter.

1. Run searx on a server where google is blocked (usually servers in datacenters).
2. Expect the message help message in the logs.

## Related issues

Related https://github.com/searxng/searxng/issues/774#issuecomment-1152284014

### DIFF
--- a/searx/search/processors/online.py
+++ b/searx/search/processors/online.py
@@ -162,6 +162,8 @@ class OnlineProcessor(EngineProcessor):
             self.handle_exception(result_container, e, suspend=True)
             self.logger.exception('CAPTCHA')
         except SearxEngineTooManyRequestsException as e:
+            if "google" in self.engine_name:
+                self.logger.warn("We recommend enabling the use_mobile_ui parameter if google is blocked for you.")
             self.handle_exception(result_container, e, suspend=True)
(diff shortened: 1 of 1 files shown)


return42:
> Too many people don't know about this parameter.

@unixfox thanks for your suggestion, but personally I think this could only be a workaround .. the user can't modify the server setup, so I think the user is the wrong addressee.

dalf:
@return42 , the message is displayed on the server log which is the right place IMO.
Perhaps  in the future we can other messages for other engines. In this case we can refactor the code.

return42:
> @return42 , the message is displayed on the server log which is the right place IMO.

Ooops .. I just read the title of the PR .. sorry.

unixfox:
> > Too many people don't know about this parameter.
> 
> @unixfox thanks for your suggestion, but personally I think this could only be a workaround .. the user can't modify the server setup, so I think the user is the wrong addressee.

sorry in my mind, the user = the one that installed searx. I wasn't talking about the user of a searx instance.

dalf (approved):

PR #1315, opened 2022-06-12
Autocomplete: don't suggest the current query.

dalf:
## What does this PR do?

Example of minor issue before this commit:
the autocompletion can suggest "Test" if the query is "Test".

Remove an useless choice from the autocompletion.

* enable google autocompletion
* type `test`: the autocompletion list should not contain a `test` item (as it is now the case).


return42 (approved):
Thanks!

PR #1319, opened 2022-06-13
[fix] text cant be selected in results while JS is enabled

### LINKED ISSUE #794: Text can't be selected when using simple theme with JS enabled

mrpaulblack:
## What does this PR do?

This is a dirty fix to make text selectable in in results while JS is enabled. The image details also dont break through this PR. Keep in mind: Its a dirty fix and there is like a small border around each selected image `<a>` tag I think.

Fix a highly reported bug.

`make run`

* check that you can select text in results

* check that image details open and close and that the source image does not get opened when pressing a image multiple times.

## Author's checklist

* I just realized that this can also be done the other way around, meaning: Just add a click even for images with `e.preventDefault();` (it was called before since on mouseup event the link was not prevented by JS I think)

## Related issues

Closes #794


mrpaulblack:
Refactored in https://github.com/searxng/searxng/pull/1322; Closing.

PR #1320, opened 2022-06-13
Theme fix click on search input

dalf:
## What does this PR do?

See https://github.com/searxng/searxng/pull/1153#issuecomment-1154247988 cursor behavior:
* when the user clicks on the search input field, the cursor moves to the end of selection.

This PR removes that (the intent was to move the cursor at the end after the page load, not after each user click).

The cursor is move at the end of the text only when the user press "i" on the keyboard.

Fix UX.

see description

## Related issues

Related to See https://github.com/searxng/searxng/pull/1153#issuecomment-1154247988


mrpaulblack (approved):
LGTM :+1: 

click on search input does not move the cursor to the end of the input anymore :+1:

PR #1322, opened 2022-06-14
[fix] keyboard.js - highlightResult: don't steal focus on click event

### LINKED ISSUE #794: Text can't be selected when using simple theme with JS enabled

return42:
For keyboard navigation the highlightResult() function in keyboard.js steals the
focus.  On a mouse click event (non keyboard action) the focus should resist
where it is, otherwise a marked region gets lost.  This is the reason why text
can't be selected when using simple theme with JS enabled.

Closes: https://github.com/searxng/searxng/issues/794


dalf:
LGTM

The only drawback I see is this scenario:
* ther user click on a result but not on the link
* then press enter
* in the master branch, the browser follow the link (image 2)
* in this branch, nothing happens (image 1)

but if this user uses the keyboard, then the link is focused (image 2).

Image 1:
image

Image 2 (the title is not underlined):
image

---

The image category works as expected:
* the keyboard navigation works
* a click on a thumbnail details the detail

---

DDG UX:
* a click anywhere on the result follow the result link
* the user can select the text --> perhaps (?), we can set the focus on mouseup. But we can do that when we refactor the JS code.


Thank you !

dalf (approved):

mrpaulblack (approved):
LGTM :+1: 

(tested on ff and chrome desktop)

PR #1343, opened 2022-06-17
Fix google play apps engine

liimee:
## What does this PR do?

Fixes the google play apps engine

the google play apps engine doesn't return any results when it should, so this pr tries to fix it

search `!gpa test`

### DIFF
--- a/searx/engines/google_play_apps.py
+++ b/searx/engines/google_play_apps.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""
+  Google Play Apps
+"""
+
+from urllib.parse import urlencode
+from lxml import html
+from searx.utils import (
+    eval_xpath,
+    extract_url,
+    extract_text,
+    eval_xpath_list,
+    eval_xpath_getindex,
+)
+
+about = {
+    "website": "https://play.google.com/",
+    "wikidata_id": "Q79576",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "HTML",
+}
+
+categories = ["files", "apps"]
+search_url = "https://play.google.com/store/search?{query}&c=apps"
+
+
+def request(query, params):
+    params["url"] = search_url.format(query=urlencode({"q": query}))
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    dom = html.fromstring(resp.text)
+
+    if eval_xpath(dom, '//div[@class="v6DsQb"]'):
+        return []
+
+    spot = eval_xpath_getindex(dom, '//div[@class="ipRz4"]', 0, None)
+    if spot is not None:
+        url = extract_url(eval_xpath(spot, './a[@class="Qfxief"]/@href'), search_url)
+        title = extract_text(eval_xpath(spot, './/div[@class="vWM94c"]'))
+        content = extract_text(eval_xpath(spot, './/div[@class="LbQbAe"]'))
+        img = extract_text(eval_xpath(spot, './/img[@class="T75of bzqKMd"]/@src'))
+
+        results.append({"url": url, "title": title, "content": content, "img_src": img})
+
+    more = eval_xpath_list(dom, '//c-wiz[@jsrenderer="RBsfwb"]//div[@role="listitem"]', min_len=1)
+    for result in more:
+        url = extract_url(eval_xpath(result, ".//a/@href"), search_url)
+        title = extract_text(eval_xpath(result, './/span[@class="DdYX5"]'))
+        content = extract_text(eval_xpath(result, './/span[@class="wMUdtb"]'))
+        img = extract_text(
+            eval_xpath(
+                result,
+                './/img[@class="T75of stzEZd" or @class="T75of etjhNc Q8CSx "]/@src',
+            )
+        )
+
+        results.append({"url": url, "title": title, "content": content, "img_src": img})
+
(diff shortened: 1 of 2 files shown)


searx/engines/google_play_apps.py
```diff
@@ -0,0 +1,60 @@
...
+    results = []
+
+    dom = html.fromstring(resp.text)
+
+    if eval_xpath(dom, '//div[@class="v6DsQb"]'):
+        return []
+
+    spot = eval_xpath(dom, '//div[@class="ipRz4"]')
+    if spot:
+        spot = spot[0]
```
dalf: Nitpicking, these 3 lines can be replaced by:
```python
    spot = eval_xpath_getindex(dom, '//div[@class="ipRz4"]', 0, None)
    if spot is not None:
        url = extract_url(eval_xpath(spot, './a[@class="Qfxief"]/@href'), search_url)
```

This is useful when the value is mandatory (not the case here): the engine error log reports the faulty XPath rather than IndexError.

dalf:
LGTM.

---

Note: when the users search for "!gpa candy crush saga":
* the engine returns "candy crush saga" as the first result
* in the HTML page, the first result is NOT "candy crush saga"

Explanation:
* the engine returns some duplicate results.
* these duplicate results are merged with a higher score than the first result. See https://github.com/searxng/searxng/blob/62e88e63a1d22c6d2298eec1aa1225801c0ca636/searx/results.py#L130-L139
* so `results.py` puts these merged duplicated result before the initial first result: https://github.com/searxng/searxng/blob/62e88e63a1d22c6d2298eec1aa1225801c0ca636/searx/results.py#L353-L359

I don't think this issue should be fixed in this PR.

liimee:
> * the engine returns some duplicate results.

yes, I think it's because search result has two sections (and the engine took both of them):
Screenshot_20220618_174306

should we take only the 'More results' section? The 'Similar apps' section doesn't have the developer name too, so some results the engine returns doesn't have a content.

return42:
> should we take only the 'More results' section?

:+1: .. I think "similar Apps" fits better into `suggestion` / here is an example how google appens its suggestions:

https://github.com/searxng/searxng/blob/de0c4d7cd8bf598d8ec69834f806a3dc80326ad6/searx/engines/google.py#L369-L372

liimee:
but not all search query (like 'game') returns 'similar apps'... is that ok?

return42:
>  is that ok?

:+1: yes, its fine .. if there are no hits, nothing needs to be append.

return42 (approved):
@liimee thanks a lot, excellent work :+1:

I rebased your branch and squashed the commits into one .. I hope this is OK for you .. I am going to merge, just waiting the CI finalze its tests ..

----

FYI: I tested suggestions by `!gpa tetris` and it works like a charm :-)

liimee:
(just found this out)

fyi: some results doesn't have a content because the app name is the same as the developer name (like GitHub). I think this is the cause:
https://github.com/searxng/searxng/blob/62e88e63a1d22c6d2298eec1aa1225801c0ca636/searx/results.py#L282-L283

return42:
@liimee good catch / thanks for pointing out .. can we add some additional info to the `content` or at least empty the `content` if its equal to the `title`:

https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/google_play_apps.py#L54-L55

liimee:
~~I don't think there's anything to add~~ I think we can add ratings, but not all apps has it.

return42:
When both are equal, we can empty the `content` variable ..

liimee:
> I think we can add ratings, but not all apps has it.

@return42 how about this?

return42:
I haven't tested but this should help:

```diff
index 577e4e114..0b0080efb 100644
--- a/searx/engines/google_play_apps.py
+++ b/searx/engines/google_play_apps.py
@@ -53,6 +53,8 @@ def response(resp):
         url = extract_url(eval_xpath(result, ".//a/@href"), search_url)
         title = extract_text(eval_xpath(result, './/span[@class="DdYX5"]'))
         content = extract_text(eval_xpath(result, './/span[@class="wMUdtb"]'))
+        if content == title:
+            content = None
         img = extract_text(
             eval_xpath(
                 result,
```

A `title`is mandatory but the `content` is optional in a result item.

liimee:
~~does that remove the result itself? setting it to `""` keeps it~~

doesn't work when setting `content` to `None`:

```
DEBUG   searx                         : result: invalid content: {'url': 'https://play.google.com/store/apps/details?id=com.github.android', 'title': 'GitHub', 'content': None, 'img_src': 'https://play-lh.googleusercontent.com/PCpXdqvUWfCW1mXhH1Y_98yBpgsWxuTSTofy3NGMo9yBTATDyzVkqU580bfSln50bFU=s64', 'engine': 'google play apps'}
```

___

I think keeping it just like this (not emptying the content if it title = content) would be fine.

return42:
if you set to an empty string?

liimee:
it does work when I set it to an empty string, but I think keeping it just like this (not changing anything) would be fine

return42:
> some results doesn't have a content

Sorry I didn't read your first post seriously .. I thought the result is removed from the result list .. my fail.

>  but I think keeping it just like this (not changing anything) would be fine

Yes, that's fine / and sorry again for the confusion.

PR #1344, opened 2022-06-18
Fix: use sys.modules.copy() to avoid RuntimeError

### LINKED ISSUE #1342: RuntimeError: dictionary changed size during iteration
Really rare but from time to time we have a 

    RuntimeError: dictionary changed size during iteration

in the CI [1].  The full traceback is shown below.  The issue appears while building the documentation `make docs`.

First reported by @dalf [2] in March 2022, but I have been seeing this error for a bit longer [3]

[1] https://github.com/searxng/searxng/runs/6931810805?check_suite_focus=true#step:6:624
[2] https://github.com/searxng/searxng/pull/930#issuecomment-1057304259
[3] https://github.com/searxng/searxng/pull/930#issuecomment-1058413786

```
  File "/home/runner/work/searxng/searxng/local/py3/lib/python3.9/site-packages/sphinx/config.py", line 332, in eval_config_file
    exec(code, namespace)
  File "/home/runner/work/searxng/searxng/docs/conf.py", line 49, in <module>
  File "/home/runner/work/searxng/searxng/searx/engines/__init__.py", line 292, in load_engines
    engine = load_engine(engine_data)
  File "/home/runner/work/searxng/searxng/searx/engines/__init__.py", line 139, in load_engine
    set_loggers(engine, engine_name)
  File "/home/runner/work/searxng/searxng/searx/engines/__init__.py", line 152, in set_loggers
RuntimeError: dictionary changed size during iteration
```

dalf:
## What does this PR do?

use sys.modules.copy() to avoid "RuntimeError: dictionary changed size during iteration"
see https://github.com/python/cpython/issues/89516
and https://docs.python.org/3.10/library/sys.html#sys.modules

The documentation suggests two ways to copy the dictionary: `.copy()` and `tuple(...)`.
I choose `.copy()` because `tuple(...)` could be overlooked.

Avoid crash when the server stats

??

## Related issues

close https://github.com/searxng/searxng/issues/1342

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -149,7 +149,11 @@ def set_loggers(engine, engine_name):
     engine.logger = logger.getChild(engine_name)
     # the engine may have load some other engines
     # may sure the logger is initialized
-    for module_name, module in sys.modules.items():
+    # use sys.modules.copy() to avoid "RuntimeError: dictionary changed size during iteration"
+    # see https://github.com/python/cpython/issues/89516
+    # and https://docs.python.org/3.10/library/sys.html#sys.modules
+    modules = sys.modules.copy()
+    for module_name, module in modules.items():
         if (
             module_name.startswith("searx.engines")
             and module_name != "searx.engines.__init__"


return42 (approved):
Thank you!

Nit; the additional name `modules` is not needed, when we write:

```python
for module_name, module in sys.modules.copy().items():
```

dalf:
The additional line is to highlight the fact a copy is required: same as `tuple(...)` that could be overlooked especially if a PR change the for loop.
May be it's too much...

Issue #1326, opened 2022-06-15
searxng.png file incorrectly preloaded


ononoki1:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
```
2022.06.14-d7a76f37
``` 

**How did you install SearXNG?**
use the official wiki

**What happened?**
searxng.png file incorrectly preloaded with `static_use_hash` on

**How To Reproduce**
1. enable `static_use_hash` option in settings.yml
2. open SearXNG instance's homepage (e.g. https://search.ononoki.org)

**Expected behavior**
Preload link is correct.

**Screenshots & Logs**
![image](https://user-images.githubusercontent.com/82757890/173726736-fb44136e-4189-41b8-a237-da31db9ca69b.png)
![image](https://user-images.githubusercontent.com/82757890/173726767-d7edc5b9-44ac-40dc-9258-4bb45e9e5734.png)

**Additional context**
The searxng.png file is loaded via searxng.min.css file, which is static, so it cannot add hash. Maybe we should load searxng.png file in html.

return42:
Just for my knowledge ..

For performance reason, we have a [<link> preload type](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload) ...

https://github.com/searxng/searxng/blob/36184b2f29187e37061590b40ce404b6a2a20c55/searx/templates/simple/index.html#L4

and the `url_for` adds the hash value ..

    <link rel="preload" href="/static/themes/simple/img/searxng.png?23ea6859f10998ae25df76e92b9d288746776640" ..>

This *preloaded* image is (only) used in the CSS ..

https://github.com/searxng/searxng/blob/36184b2f29187e37061590b40ce404b6a2a20c55/searx/static/themes/simple/src/less/index.less#L9

but in the CSS we do not have the hash value ..

> The searxng.png file is loaded via searxng.min.css file, which is static, so it cannot add hash. Maybe we should load searxng.png file in html.

By this we have two request from the client for the identical image ..

1. the one from the preload link `searxng.png?23ea6859f10998ae25df76e92b9d288746776640`
2. and one from the CSS `searxng.png`

Since the first one is never used, the client reports ..

    The resource at “https:// .... /searxng.png?23ea6859f10998ae25df76e92b9d288746776640”
    preloaded with link preload was not used within a few seconds. Make sure all attributes
    of the preload tag are set correctly.

If I am not wrong ..

[...]

PR #1345, opened 2022-06-18
Workaround: index.html: remove image preload

dalf:
## What does this PR do?

URL for the logo is referenced twice:
* in index.html for preloading: it contains the hash for cache busting (when there is `static_use_hash: true` in `settings.yml`)
* in searxng.min.css: to actually display the image. The URL doesn't contain the hash.

So the image preload actually loads twice the same image.

This commit removed the image preloading.
This is workaround: the real fix is to be able to use the URL with the hash in the CSS.

When there is `static_use_hash: true` in `settings.yml`, the logo is loaded twice.

* set `static_use_hash: true`
* check the logo is loaded only once.

## Related issues

Related to https://github.com/searxng/searxng/issues/1326

### DIFF
--- a/searx/templates/simple/index.html
+++ b/searx/templates/simple/index.html
@@ -1,8 +1,5 @@
 {% extends "simple/base.html" %}
 {% from 'simple/icons.html' import icon_big %}
-{% block meta %}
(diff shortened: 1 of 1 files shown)


return42 (approved):
This workaround is a good first choice I think :+1: .. thank you!

PR #1346, opened 2022-06-18
Theme: dependencies upgrade

dalf:
## What does this PR do?

Upgrade the JS & node dependencies : `make static.build.commit` does not create a commit.

Note: the master branch already contains [leaflet 1.8.0](https://github.com/searxng/searxng/blob/master/searx/static/themes/simple/js/leaflet.js) (we need to commit package-lock.json to avoid flaky dependency versions).

Stylelint requires an update of `.stylelintrc.json` and perhaps some other files.

Up to date build dependencies


searx/static/themes/simple/package.json
```diff
@@ -11,18 +11,18 @@
     "grunt-eslint": "^24.0.0",
     "grunt-stylelint": "^0.16.0",
     "grunt-image": "^6.4.0",
-    "ionicons": "^6.0.1",
-    "less": "^4.1.2",
+    "ionicons": "^6.0.2",
+    "less": "^4.1.3",
```
return42, line 15: I sometimes confused by semantic versioning and the benefit of all / for my information, why do we need to increase patch (and minor) release?

AFIK: in NPM's [semantic versioning](https://docs.npmjs.com/about-semantic-versioning) all patch- and minor-releases are covered by (e.g.) `"^8.0.0"`.

I can understand that we insist on a minor level (e.g. `"leaflet": "^1.8.0"`) when we need some features of this minor release, but I do not understand why we need to insist on patch-level .. may be I missed something in the complexity of package management .. but from my current knowledge the patch level can always set to zero when we use `^` to fetch latest minor release (`"less": "^4.1.3"` --> `"less": "^4.1.0"` should both end in the last minor & patch release).

searx/static/themes/simple/package.json
```diff
@@ -32,8 +32,7 @@
     "eslint": "grunt eslint",
     "eslint-fix": "grunt eslint --fix",
     "watch": "grunt watch",
-    "webfont": "grunt webfont",
-    "clean": "rm -Rf node_modules package-lock.json ion.less",
+    "clean": "rm -Rf node_modules package-lock.json",
```
return42, line 35: Does this `ion.less` not exists (anymore) / is the `rm ion.less` a leftover from past?

dalf: `ion.less` will used to create the webfont : https://github.com/searx/searx/blob/master/searx/static/themes/simple/less/ion.less

searx/static/themes/simple/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "devDependencies": {
-    "eslint": "^8.11.0",
-    "grunt": "~1.5.2",
+    "eslint": "^8.18.0",
```
return42, line 3: We should also update eslint from root's `package.json`

https://github.com/searxng/searxng/blob/62e88e63a1d22c6d2298eec1aa1225801c0ca636/package.json#L3

searx/static/themes/simple/package.json
dalf: Some static build commits were built with leaftlet version 1.7.1 some with 1.8.0:
* https://github.com/searxng/searxng/commits/master/searx/static/themes/simple/js/leaflet.js
* 2c62360e6e13e427342260dd72880179b55fa3c8 : 1.8.0
* 7e8a60fb275bc8252b3f95c6e2c9b810c8802a38 : 1.7.1
* 3f18761adb722dff36cdfeb2d867e40652257f26 : 1.8.0
* 0f18bfc8a55bae0cca70f72cc77fa87bfe073335 : 1.7.1

At least if we enforce the last version it should be less flaky.

searx/static/themes/simple/package.json
dalf: done

searx/static/themes/simple/package.json
return42: isn't it better to pin the version, like we do in python?

return42 (approved):

searx/static/themes/simple/package.json
dalf: We should lock the dependencies with packages-lock.json : this is what dependabot uses to open pull request (currently dependabot ignores packages.json because packages-lock.json is missing).

PR #1353, opened 2022-06-20
[fix] clean up templates / remove import of result_footer_rtl macro

return42:
A macro named 'result_footer_rtl' does not exists.


dalf (approved):
Not tested, but make sense: there is no reference to this macro.

PR #1356, opened 2022-06-21
[mod] mudularize & document searx.results

return42:
## What does this PR do?

The intention of this patch is to improve modularization & documentation of the implementations about the *result* items.

The module `searx.results` is changed into a python package with modules.

> **Note**
> This patch does not contain any functional change!

The result types, the fields of result items and the media types of the result list are  elementary for any development of engines and templates and should therefore be documented in detail.

- Compile docs `make docs.live` and jump to "Source-Code" and read "Engine Results".
- Start session `make run` and test some queries (nothing should be changed / no functional change)

### DIFF
--- a/docs/dev/engine_overview.rst
+++ b/docs/dev/engine_overview.rst
@@ -120,6 +120,11 @@ module:
 Making a Request
 ================
 
+.. sidebar:: info
+
+   - Demo of :py:obj:`request(query, params)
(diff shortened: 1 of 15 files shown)


return42:
I close this PR in favorite of PR https://github.com/searxng/searxng/pull/1412

(PR #1412 contains this PR)

PR #1373, opened 2022-06-27
Dockerfile: use alpine 3.16

dalf:
## What does this PR do?

* Use Alpine 3.16 : https://www.alpinelinux.org/posts/Alpine-3.16.0-released.html
* Use Python 3.10 instead of 3.9 currently in Alpine 3.15

Up to date dependency

## Author's checklist

After some tests locally, it doesn't show any issues.


mrpaulblack (approved):
Works perfectly on my instance LGTM :+1:

PR #1381, opened 2022-06-29
openstreetmap: localized results.

dalf:
## What does this PR do?

Master branch:
image

This PR:
image

* The results are the language the user has selected (English if the query language is English and if there is an English name, otherwise it falls back to the native name)
* HTML template: the wikidata & wikipedia links contains only the titles not the full link.

Improve the usuability

* search for `!osm seoul`
* Note: I'm not sure about the traditional Chinese if well supported.

### DIFF
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -29,6 +29,7 @@ about = {
 # engine dependent config
 categories = ['map']
 paging = False
+language_support = True
 
 # search-url
 base_url = 'https://nominatim.openstreetmap.org/'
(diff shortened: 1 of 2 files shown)


return42:
> Note: I'm not sure about the traditional Chinese if well supported.

AFAIK "no" .. e.g. wikipedia links for `paris`:

- traditional Chinese (zh-TW): https://zh-classical.wikipedia.org/wiki/%E5%B7%B4%E9%BB%8E
- simplified Chinese: (zh-CN): https://zh.wikipedia.org/wiki/%E5%B7%B4%E9%BB%8E

.. I will have a look, what we can improve ..
 
---

didn't find a solution for / its not such trivial to me since I am not familiar with sparql

And what we win is questionable: the articles on zh-classical.wikipedia.org are less maintained compared to zh.wikipedia.org

return42 (approved):
Thanks a lot this is a real improvement! .. FYI I rebased and placed a commit (5831c15) on top to fix a typo.

PR #1382, opened 2022-06-29
fix searxng_extra/update/update*.py scripts

dalf:
## What does this PR do?

call searx.locales.locales_initialize before using LOCALE_NAMES

Bug fix

## Related issues

Related to https://github.com/searxng/searxng/pull/1306


return42 (approved):
Thanks for fix the update scripts / FYI I rebased the branch and add two commits updating currencies and engine descriptions.

PR #1412, opened 2022-07-02
mudularize & document searx.results and fix #1348

### LINKED ISSUE #1348: Links in the infobox are not modified by hostname_replace
The `hostname_replace` list is not applied to the URLs of links and images in the infobox:

- `infobox.img_src`
https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/templates/simple/infobox.html#L3

- `for attribute in infobox.attributes` --> `attribute.image.src`
https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/templates/simple/infobox.html#L7-L11

- `for url in infobox.urls` --> `url.url`
https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/templates/simple/infobox.html#L23-L25

IDK if the above is a complete list of all URLs in the infobox / maybe I have overseen some URLs.

return42:
## What does this PR do?

Fix issues with hostname_replace where URLs in answers and infobox not handled.

In preparation for the bugfix, the first commit mudularize & document searx.results.

```
- [mod] mudularize & document searx.results
- [fix] hostname_replace: modify URLs in the infobox
```

Modularize code, add documentation and fix an issue.

First check the documentation `make docs.clean docs.live` and visit pages:

- dev/engine_overview.html (read "Making a Response")
- src/searx.results.html

Second test hostname_replace plugin by redirecting youtube and twitter:

```yaml
hostname_replace:
  # youtube --> Invidious

  '(.*\.)?youtube-nocookie\.com': 'invidio.xamh.de'
  '(.*\.)?youtube\.com$': 'invidio.xamh.de'

  # twitter --> nitter

  '(www\.)?twitter\.com$': 'nitter.net'

  # test "!cc 1 EUR to dollar" --> link should be removed

  '(.*\.)?duckduckgo\.com$': false
```

make a search, in the infobox the links should be redirected.

## Related issues

Closes: https://github.com/searxng/searxng/issues/1348
Supersede: #1356

### DIFF
--- a/docs/dev/engine_overview.rst
+++ b/docs/dev/engine_overview.rst
@@ -120,6 +120,11 @@ module:
 Making a Request
 ================
 
+.. sidebar:: info
+
+   - Demo of :py:obj:`request(query, params)
+     <searx.engines.demo_online.request>` function.
+
 To perform a search an URL have to be specified.  In addition to specifying an
 URL, arguments can be passed to the query.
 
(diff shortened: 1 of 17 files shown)


return42:
FYI: rebased PR's branch (82 commits)

Can we please force merge (review) of this PR? .. beside the bugfix in the second commit, the modularization and documentation from the first commit is a prerequisite for further development of the response types / thanks!

return42:
FYI rebased (0a11478 --> 7a48dc7) to solve conflict with current master.

dalf:
After a long thinking, I'm "meh" about this PR:
* why not use the [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) instead of documentation?  I prefer code as documentation rather documentation that might get out of sync. We even imagine to have some dedicated classes ( [here a prototype](https://gist.github.com/dalf/972eb05e7a9bee161487132a7de244d2) )
* most of the functions inside the new modules could be methods (some as class methods). For example `is_answer` ... could be a static method, so it is super easy to add a new result type (see [query.py](https://github.com/searxng/searxng/blob/50ec821e8c5b099947d657722b1879fc8c03bcfa/searx/query.py#L43-L45) and see line 241).
* why "standard" result are not in a new module?
* why not split this PR in two: 
  * one to fix the hostname_replace on infoboxes 
  * one where we take time to rewrite results.py 
  ?

return42:
> why not use the [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) instead of documentation? I prefer code as documentation 

This is often what I hear .. but it fails in two ways .. 

- first, what when the code has a bug .. by saying the code is the documentation no one can check whats going wrong in the function.

- how can code document the relation of objects to each other when the code uses ellipses / where is the "Interface" description of the result-types on which a UI developer can build up UI interface (HTML templates).

We have a templates .. these templates are undocumented, split up in files and macros .. and we have multiple result types .. how should a newbie understand all this? .. I toked me a week ore more to find all the relations and to write it down .. IMO the lack of so many needed documentation is the reason why this project is so hard to get into .. to be honest, several of your implementations I don't understand because there is no documentation .. that is why I ask dump questions like the one today https://github.com/searxng/searxng/discussions/1559 .. just one small hint in a docstring of the method of the abstract class about what the meaning is of returning `None` is needed to save a lot of time and make things clear.


> rather documentation that might get out of sync.

IDK a better way to document, the doc-strings are in the source code, a few lines above the code that implements what the doc-string described .. compile the docs `make docs.live` and open http://0.0.0.0:8000/src/searx.results.html to see what I mean.

Issue #1325, opened 2022-06-14
I'd like a translate plugin


BernieHuang2008:
I usually use web browser to translate some words or sentence, like : `translate xxx`

I found that use `plugin` is a good solution.

It's already has the `hash plugin` and the input format is as same as it.

If you want google translate API, I will give you soon.

PR #1434, opened 2022-07-04
Adds Lingva/Google Translate engine

liimee:
## What does this PR do?

adds the lingva engine (which grabs data from google translate)

it'd be convenient to add more translation engines to searxng, especially as the data comes from google translate

search `!lv en-ar test`

## Related issues

#1325 may be related

___

as Lingva grabs data from Google Translate, should we name the engine google translate instead?


return42 (approved):
I rebased your branch and fixed some minor issues in your commit.

I also added two new commits on top: in the infobox, common [`<dl>` Description List elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) had not been rendered well.  With this fix we can use `<dl>` elements instead of the unnumbered list elements (`<ul>`).

I enabled it by default and placed it in the 'general' category / user can now type (e.g.) `en-de assembly` and will get the translation .. this is feature I missed since a long time :+1: 

Thanks a lot for adding this convenient translation to SearXNG .. very cool!

liimee:
I think it'd be great if it supports queries with spaces, like `hello there`. Seems to be a searx thing:

https://github.com/searxng/searxng/blob/5c1c0817c3996c5670a545d05831d234d21e6217/searx/search/processors/online_dictionary.py#L12

PR #1443, opened 2022-07-04
[fix] online_dictionary: regular expression

return42:
The query term of a engine-type `online_dictionary` can consist of more than one word.

### DIFF
--- a/searx/search/processors/online_dictionary.py
+++ b/searx/search/processors/online_dictionary.py
@@ -9,7 +9,7 @@ import re
 from searx.utils import is_valid_lang
 from .online import OnlineProcessor
 
-parser_re = re.compile('.*?([a-z]+)-([a-z]+) ([^ ]+)$', re.I)
+parser_re = re.compile('.*?([a-z]+)-([a-z]+) (.+)$', re.I)
 
 
 class OnlineDictionaryProcessor(OnlineProcessor):


searx/search/processors/online_dictionary.py
```diff
@@ -9,7 +9,7 @@
 from searx.utils import is_valid_lang
 from .online import OnlineProcessor
 
-parser_re = re.compile('.*?([a-z]+)-([a-z]+) ([^ ]+)$', re.I)
+parser_re = re.compile('.*?([a-z]+)-([a-z]+) (.*)$', re.I)
```
liimee: just curious, why not use `.*?([a-z]+)-([a-z]+) (.+)$` instead?

dalf: Not all languages use the latin alphabet.
Even in English that's not so easy, for example `o'clock`

liimee: sorry, but what do you mean? I only edited the `(.*)` to `(.+)` and it seems to work...

dalf: > sorry, but what do you mean?

I've read too quickly, sorry. 

I agree with `(.+)`, it avoids empty queries.

return42: > just curious, why not use `.*?([a-z]+)-([a-z]+) (.+)$` instead?

fully agree .. thanks for pointing out :+1:

PR #1456, opened 2022-07-05
[fix] engine tineye: handle 422 response of not supported img format

### LINKED ISSUE #1449: Bug: tineye engine
Branch: master
Version: 2022.06.14-e67ed894

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: httpx.HTTPStatusError
  * Percentage: 100
  * Parameters: `('422', 'Unprocessable Entity', 'tineye.com')`

return42:
## What does this PR do?

Tineye: handle 422 response of not supported img format

grafik

## Related issues

Closes: https://github.com/searxng/searxng/issues/1449

### DIFF
--- a/searx/engines/tineye.py
+++ b/searx/engines/tineye.py
@@ -17,6 +17,7 @@ billion images `[tineye.com] <https://tineye.com/how>`_.
 
 from urllib.parse import urlencode
 from datetime import datetime
+from flask_babel import gettext
 
 about = {
     "website": 'https://tineye.com',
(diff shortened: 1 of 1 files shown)


dalf:
I've tried various URL, here two examples:

* `https://i.ytimg.com/an_webp/gusXBTyg1O4/mqdefault_6s.webp?du=3000&sqp=CNjCn5YG&rs=AOn4CLBNM0alsGkVAM7rKJ5CydtKcVY2kQ`
  * `suggestions.key`: "Too Simple"
  * Status Code: 400
* `https://tineye.com/result_json`
  * `suggestions.key`: "Download Error"
  * Status Code: 422

Looking at tineye PIP package, it seems the result is always a JSON: https://github.com/TinEye/pytineye/blob/d51c637a4b3e9b50b47739b540ab908243bf4f91/pytineye/api.py#L277 (Note: the entry point is different, we can't be sure)

So, the following code deals with more situation:

```python
    if resp.is_error:
        # handle error codes from Tineye
        if resp.status_code in (400, 422):
            error = json_data.get('error')
            s_key = json_data.get('suggestions', {}).get('key', '')
            if s_key.startswith("Invalid image URL"):
                message = format_not_supported
            elif s_key:
                message = "TinEye:" + s_key
            else:
                message =  "TinEye:" + error
            results.append({'answer': message})
            return results
        resp.raise_for_status()
```

---

I feel we divert the usage of the answers or there are some missing features: looking at the code, we need to specify the engine name because the answers don't provide the engine.
The result is more an error, except we don't provide a way to return that.

For now:
* I would just return an empty result
* open an issue to allow engine to return usage error.
* once implemented, we can update tineye.py

---

Out of topic: when I've search for the query `https://tineye.com/result_json python`, tineye got the query and returned an error. We can avoid this if we
* move tineye to the images category.
* match only full URL without space.
* provide a category for URL search: `!url https://example.com`

return42:
> For now:
> * I would just return an empty result
> * open an issue to allow engine to return usage error.

first lets merge https://github.com/searxng/searxng/pull/1412

> * once implemented, we can update tineye.py

OK

> * provide a category for URL search: `!url https://example.com`

No need for, it is a online_url_search engine:

https://github.com/searxng/searxng/blob/50d714d82952417f3db837a176b751d3845c0cc8/searx/search/processors/online_url_search.py#L10-L14

> when I've search for the query `https://tineye.com/result_json python`, tineye got the query and returned an error. 

The `https://tineye.com/result_json` URL is responsible for the error / its not an image URL.

Issue #1405, opened 2022-07-02
Bug: google engine


linuxtyp:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2022.06.15-36184b2f

**How did you install SearXNG?**
searxng-docker on a raspberry pi 3

**What happened?**
error occurs when i search something

**How To Reproduce**

**Expected behavior**
I don't think the error has consequences, but it bothers me...

**Screenshots & Logs**
Error! Engines cannot retrieve results:
google ( [Suspended: too many requests](https://notgoogle.live/stats?engine=google) )

**Additional context**
I tried to fix this Problem with a solution for this error but it still throws the error.
ui:
  use_mobile_ui: true
  static_use_hash: true

i added the use_mobile_ui: true line in the settings.yml

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineTooManyRequestsException
  * Percentage: 5
  * Parameters: `('Too many request',)`
  * File name: `searx/search/processors/online.py:96`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

tiekoetter:
@linuxtyp Which AS is your servers IP(s) on?

linuxtyp:
Thank you for your quick reply, but i don't understand what you mean by AS.

tiekoetter:
https://en.wikipedia.org/wiki/Autonomous_system_(Internet)

Basically what ISP it uses.

linuxtyp:
No it is just in my Local Network. From my home ip address.

tiekoetter:
Ok that is strange. I suppose you can use Google normally on your network? Is there anything you do on your network which could let google believe that you are a "bot"?

linuxtyp:
Yes i can use google normally. On the same Raspberry pi runs Pihole it could be that that blocks the requests. Thank you for your help, much appreciated. I will look at the query log and whitelist the safebrowsing url. If i have any more problems i will post again.

Razuuu:
This error is on notgoogle.live too and on my own searxng.

tiekoetter:
@Razuuu Same question for you: What AS is the IP(s) of your instance on?

The notgoogle.live instance was probably flooded and therefore blocked by google. Most top public instances at https://searx.space don't have this issue.

Razuuu:
Its Deutsche Telekom AG

[...]

PR #1460, opened 2022-07-05
Better explanation for the use of use_mobile_ui

unixfox:
## What does this PR do?

Better explanation for the use of use_mobile_ui

People don't seem to bother understanding where to put the `use_mobile_ui` parameter: https://github.com/searxng/searxng/issues/1405#issuecomment-1175516356

This PR aims to reduce the amount of useless issues.


return42:
run `make format.python` to solve the issue from black https://github.com/searxng/searxng/runs/7204615657?check_suite_focus=true#step:7:42

return42 (approved):
Thanks!

PR #1478, opened 2022-07-08
Engine error log: Submit a new issue button: ask the user to check if there are some existing issues.

dalf:
## What does this PR do?

Add some friction before the user create an issue on GitHub:

image

When clicked:
image

After a few seconds after the first check has been checked (works without JS, only CSS):
image

When the second checkbox is clicked:
image

User click on the "Submit issue on GH", click "Submit" to easily.

* make a search using a broken engine: btdigg for example.
* go to the `/stats` page for this engine.
* go to the `/stats` page for a working engine.

## Author's checklist

A template to issue an bug can be additional measure or a alternative to this PR or ...


return42 (approved):
Wow .. very cool .. and without JS :+1: .. this will reduce the noise a lot .. don't lets wait to merge / thanks!!!

PR #1481, opened 2022-07-08
feat: Add tor check plugin

Solirs:
Add tor_check plugin which allows convenient tor checking trough searxng

## What does this PR do?

This PR adds a tor_check plugin to conveniently check if you are using tor,
the plugin returns whether you are using tor or not if you search for "tor", "tor check", or "check tor" as well as your own ip
like check.torproject.org does.
Sort of like the self_info plugin but for tor checking.

It's IMO convenient to be able to check if you're using tor from your search engine,
like it is convenient to be able to check your IP or user agent from it.
I think it would be overall a nice thing to add and would improve the UE of some people including me.

Uncomment "Tor check plugin" in settings.yml
(optional) Set the IP variable to a TOR exit node's ip address if you don't want to port forward to access your test instance from tor.
make run.
Search for "tor".

### DIFF
--- a/searx/plugins/tor_check.py
+++ b/searx/plugins/tor_check.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""A plugin to check if the ip address of the request is a TOR exit node if the
+user searches for ``tor-check``.  It fetches the tor exit node list from
+https://check.torproject.org/exit-addresses and parses all the IPs into a list,
+then checks if the user's IP address is in it.
+
+Enable in ``settings.yml``:
+
+.. code:: yaml
+
+  enabled_plugins:
+    ..
+    - 'Tor check plugin'
+
+"""
+
+import re
+from flask_babel import gettext
+from httpx import HTTPError
+from searx.network import get
+
+default_on = False
+
+name = gettext("Tor check plugin")
+'''Translated name of the plugin'''
+
+description = gettext(
+    "This plugin checks if the address of the request is a TOR exit node, and"
+    " informs the user if it is, like check.torproject.org but from searxng."
+)
+'''Translated description of the plugin.'''
+
+preference_section = 'query'
+'''The preference section where the plugin is shown.'''
+
+query_keywords = ['tor-check']
+'''Query keywords shown in the preferences.'''
+
+query_examples = ''
+'''Query examples shown in the preferences.'''
+
+# Regex for exit node addresses in the list.
+reg = re.compile(r"(?<=ExitAddress )\S+")
+
+
+def post_search(request, search):
+
+    if search.search_query.pageno > 1:
+        return True
+
+    if search.search_query.query.lower() == "tor-check":
+
+        # Request the list of tor exit nodes.
+        try:
+            resp = get("https://check.torproject.org/exit-addresses")
+            node_list = re.findall(reg, resp.text)
+
+        except HTTPError:
+            # No answer, return error
+            search.result_container.answers["tor"] = {
+                "answer": gettext(
+                    "The TOR exit node list (https://check.torproject.org/exit-addresses) is unreachable."
+                )
+            }
+            return True
+
+        x_forwarded_for = request.headers.getlist("X-Forwarded-For")
+
+        if x_forwarded_for:
+            ip_address = x_forwarded_for[0]
+        else:
+            ip_address = request.remote_addr
+
+        if ip_address in node_list:
+            search.result_container.answers["tor"] = {
+                "answer": gettext(
+                    "You are using TOR. Your IP address seems to be: {ip_address}.".format(ip_address=ip_address)
+                )
+            }
+        else:
+            search.result_container.answers["tor"] = {
+                "answer": gettext(
+                    "You are not using TOR. Your IP address seems to be: {ip_address}.".format(ip_address=ip_address)
+                )
+            }
+
+    return True
--- a/docs/src/searx.plugins.tor_check.rst
+++ b/docs/src/searx.plugins.tor_check.rst
@@ -1,0 +1,9 @@
+.. _tor check plugin:
+
+================
+Tor check plugin
+================
+
+.. automodule:: searx.plugins.tor_check
+  :members:
+
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -187,6 +187,7 @@ outgoing:
 #   - 'Hostname replace'  # see hostname_replace configuration below
 #   - 'Open Access DOI rewrite'
 #   - 'Vim-like hotkeys'
+#   - 'Tor check plugin'
 
 # Configuration of the "Hostname replace" plugin:
 #


Solirs:
@return42 @dalf Anyone here to review this ?

return42:
> @return42 @dalf Anyone here to review this ? 

Sorry for the delay .. I do not have much time this weekend .. but its on the list ..

Solirs:
> > @return42 @dalf Anyone here to review this ?
> 
> Sorry for the delay .. I do not have much time this weekend .. but its on the list ..

I understand, thank you in advance for your time

return42:
@Solirs can you add SPDX and pylint tag: 

https://github.com/searxng/searxng/blob/4de4a213a6fa4b1755a3c8bc55cdf79a0c98875d/searx/plugins/limiter.py#L1-L2

on top of the new created python file / thanks!

Solirs:
Done!

tiekoetter:
@Solirs Please squash your commits together and add a commit message.

Solirs:
Done!

Solirs:
Would it be better enabled by default or is it fine like that?

tiekoetter:
@Solirs I personally would enable it but I don't know what the others think.

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,57 @@
...
+            return True
+
+        x_forwarded_for = request.headers.getlist("X-Forwarded-For")
+
+        if x_forwarded_for:
+            ip_address = x_forwarded_for[0]
+        else:
+            ip_address = request.remote_addr
+
+        if ip_address in node_list:
```
tiekoetter, line 75: Maybe it should be reworded. Regardless "Tor" should be capitalized.

Solirs: You're right. I'm going to improve that

tiekoetter: Sorry I didn't highlight the correct lines. Either I forgot how this works or my browser is acting up.

tiekoetter (approved):
LGTM.

tiekoetter:
@Solirs Please squash again. Thanks for your work 👍

Waiting for the others to add their review.

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+
+Enable in 'settings.yml' (uncomment "Tor check plugin"
+in the 'enabled_plugins' section)
+"""
+
+import re
+from urllib.request import urlopen
+from urllib.error import HTTPError
+
+name = "Tor check plugin"
```
dalf: To allow translation you have to call gettext:
```python
from flask_babel import gettext

...

name = gettext("Tor check plugin")
```

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+It fetches the tor exit node list from https://check.torproject.org/exit-addresses
+and parses all the IPs into a list, then checks if the user's ip address
+is in it.
+
+Enable in 'settings.yml' (uncomment "Tor check plugin"
+in the 'enabled_plugins' section)
+"""
+
+import re
+from urllib.request import urlopen
```
dalf:
```
from httpx import HTTPError
from searx.network import get
```

* `get` is going to work like [`httpx`](https://www.python-httpx.org/).
* but `get` takes care to provide a HTTP client with [the network configuration described in `settings.yml`](https://docs.searxng.org/admin/engines/settings.html#outgoing)

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+        x_forwarded_for = request.headers.getlist("X-Forwarded-For")
+
+        if x_forwarded_for:
+            ip_address = x_forwarded_for[0]
+        else:
+            ip_address = request.remote_addr
+
+        if ip_address in node_list:
+            search.result_container.answers["tor"] = {
+                "answer": f"You are using TOR. Your IP address seems to be: {ip_address}."
```
dalf: f-string are not supported by babel: https://github.com/python-babel/babel/issues/594

So
```python
gettext("You are using TOR. Your IP address seems to be: {ip_address}.").format(ip_address=ip_address)
```

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+def post_search(request, search):
+
+    if search.search_query.query.lower() in ("tor", "tor check", "check tor"):
+
+        # Request the list of tor exit nodes.
+        try:
+            with urlopen("https://check.torproject.org/exit-addresses") as ex_nodes:
+                node_list = re.findall(reg, str(ex_nodes.read()))
+
+        except HTTPError:
```
dalf, line 59: still `except HTTPError` but from httpx.

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+default_on = False
+
+# Regex for exit node addresses in the list.
+reg = re.compile(r"(?<=ExitAddress )\S+")
+
+
+def post_search(request, search):
+
+    if search.search_query.query.lower() in ("tor", "tor check", "check tor"):
```
dalf: Idea : use the same trigger than the `self_info` plugin.

So the query `ip` (nothing else) triggers `self_info` and this plugin. 
The users can pick the one they want (or both but it is not useful).

Alternative: include this code into `self_info`, but I guess some people don't want to fetch content from `check.torproject.org` when they want their IP address.

---

https://github.com/searxng/searxng/blob/44f2eb50a5566e8e14a1e970f1ed775bf025c1ea/searx/plugins/self_info.py#L37

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+This plugin checks if the address of the request is a TOR exit node,
+and informs the user if it is, like check.torproject.org but from searxng.
+"""
+default_on = False
+
+# Regex for exit node addresses in the list.
+reg = re.compile(r"(?<=ExitAddress )\S+")
+
+
+def post_search(request, search):
```
dalf, line 47: See `self_info.py` :
https://github.com/searxng/searxng/blob/44f2eb50a5566e8e14a1e970f1ed775bf025c1ea/searx/plugins/self_info.py#L35-L36

so the plugin is triggered only on the first page.

searx/plugins/tor_check.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if search.search_query.query.lower() in ("tor", "tor check", "check tor"):
+
+        # Request the list of tor exit nodes.
+        try:
+            with urlopen("https://check.torproject.org/exit-addresses") as ex_nodes:
+                node_list = re.findall(reg, str(ex_nodes.read()))
```
dalf:
```python
resp = get("https://check.torproject.org/exit-addresses")
node_list = re.findall(reg, str(resp.text))
```

Just to make sure: does it makes sense to cache the response?
It seems no looking how often the content change ( https://collector.torproject.org/recent/exit-lists/ )

dalf (changes requested):
Thank you for this PR.
I left some comments.

searx/plugins/tor_check.py
Solirs: I don't really know what else i can do.
Do you mean something like:
`node_list = re.findall(reg, get("https://check.torproject.org/exit-addresses"))` ?
Or 
`if ip_address in re.findall(reg, get("https://check.torproject.org/exit-addresses")):` ?

searx/plugins/tor_check.py
Solirs: I think it would help with the convenience to keep it like that and on by default, its more intuitive to type "tor" to check if you're using tor and "ip" to just check your IP and have this work out of the box than having  them both in the same place.

Solirs:
@dalf @return42 Hey, any followup to this?

return42:
Hi @Solirs, I am very sorry about the delay .. I had a lot work the last days .. I hope I have time next weekend.

@dalf I made rough review and the PR LGTM.  From my side we can merge this PR.  You had some remarks @Solirs solved .. about your https://github.com/searxng/searxng/pull/1481#discussion_r917423742 I think we can optimize (cache) in new PR.

Solirs:
Alright, thank you for your time, i see no more improvements i can make on this so i think i'm done with this PR unless more improvements are suggested.

searx/plugins/tor_check.py
return42: > Just to make sure: does it makes sense to cache the response?

I don't know a rule for, I leave it un-cached  for its first / when we have a rule to update the cache, we can implement a new PR.

searx/plugins/tor_check.py
return42: I reduced it to ``query_keywords = ['tor-check']``

return42:
@Solirs thanks a lot for your PR, in my review I changed the following (hope it is OK for you):

- rebased on master branch
- squashed the 5 commits into one
- added `query_keywords = ['tor-check']` and `preference_section = 'query'` // with the plugin is shown in the preferences ..

image

- add *automodule* to the documentation of the source code // see https://docs.searxng.org/src/searx.plugins.tor_check.html

PR #1486, opened 2022-07-09
Add infinite scroll as a setting in settings.yml

vojkovic:
## What does this PR do?

Added the `infinite_scroll:` bool to searx/settings.yml. This is currently a hidden setting.

Would help the people asking how to enable infinite scroll in the settings such as in this discussion: https://github.com/searxng/searxng/discussions/1333.

It also makes it easier to enable `infinite_scroll` in the settings.yml with tools like sed.

Nothing should be change as infinite scroll is still disabled by default.

Changing the setting to true will enable infinite scroll.

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -230,6 +230,7 @@ Global Settings
    ui:
      default_locale: ""
      query_in_title: false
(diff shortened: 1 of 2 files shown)


tiekoetter (approved):
LGTM

PR #1488, opened 2022-07-10
config page

Xavierhorwood:
## What does this PR do?

Adds `security.config_page`

When the value is `false` `/config` will return 403

When the value is `true` `/config` will return the config as it did before

Some administrators may not want there config values to be public.

`make run` without changing settings.yml: check the `/config` page is still accessible 
Change for security.config_page: false: the `/config` page will return 403

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -225,7 +225,7 @@ SCHEMA = {
     'doi_resolvers': {},
     'security': {
         'config_page': SettingsValue(bool, True),
-    }
+    },
 }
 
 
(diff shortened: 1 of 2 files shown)


mrpaulblack:
Hi @Xavierhorwood ,
After some internal discussion we decided not to merge this PR. One of the problems with public instances is, that the end user needs to trust the instance maintainer to (for example) not log the users search query. Giving instance maintainers the option to disable the `/config` endpoint removes more transparency from an instance on how it is configured. Also on private instances its not really an issue IMO, since its private and should only be accessible by trusted people... Also I dont really see the `/config` endpoint as a security issue; Removing it would be security by obfuscation, which is IMO not a real security option.

Therefor we will not merge this PR. I hope this does not discourage you to open new PRs in the future :D

PR #1489, opened 2022-07-10
[mod] allow brand.issue_url to overwrite the step1 url

tiekoetter:
## What does this PR do?

Allow brand.issue_url to overwrite the step1 url

A changed brand.issue_url and/or brand.new_issue_url would be ignored in the step1 issue search url.

1. `make run`
2. Change `brand.issue_url` in settings.yml
3. Check if the step1 url changed.

### DIFF
--- a/searx/templates/simple/new_issue.html
+++ b/searx/templates/simple/new_issue.html
@@ -63,7 +63,7 @@ or manually by executing the searx/webapp.py file? -->
     <input type="checkbox" id="step1">
     <label for="step1">{{ _('Start submiting a new issue on GitHub') }}</label>
     <div class="step1 step_content">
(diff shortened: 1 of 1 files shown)


dalf:
Same issue than https://github.com/searxng/searxng/pull/1295 we are not sure about the URL. If the repository is neither GitHub nor GitLab nor Gitea. perhaps the URL is `get_setting('brand.issue_url') + "/search?q="` (example).

But here the query (`?q=...`) most probably won't hurt and it falls back to the issue list which is what we want.

dalf (approved):

mrpaulblack (approved):
LGTM :+1:

Issue #662, opened 2021-12-27
PY_SSIZE_T_CLEAN macro must be defined for '#' formats


HLFH:
Repo: https://github.com/dalf/searxng
Branch: upgrade-httpx > https://github.com/dalf/searxng/tree/upgrade-httpx
Last commit of this branch.
Related to: https://github.com/searxng/searxng/pull/596

```
déc. 27 17:24:10 arch-server uwsgi[2672526]: 2021-12-27 17:24:10,806 ERROR:searx: Error while getting the version: fatal: not a git repository (or any of the parent directories): .git
déc. 27 17:24:10 arch-server uwsgi[2672526]: 2021-12-27 17:24:10,813 ERROR:searx: Error while getting the git URL & branch: fatal: not a git repository (or any of the parent directories): .git
```
What shall I do on this one?

```
déc. 27 17:24:11 arch-server uwsgi[2672526]: 2021-12-27 17:24:11,865 ERROR:searx.shared: uwsgi.ini configuration error, add this line to your uwsgi.ini
déc. 27 17:24:11 arch-server uwsgi[2672526]: cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1
```
Scheduler is enabled in `/etc/searxng/settings.yml` and requested config line is in `/etc/uwsgi/vassals/searxng.ini `. 
What shall I do more on this one? 

EDIT: 

```
python -m searx.version freeze
```

To fix the first issue.

HLFH:
cc @dalf 

How to fix the `uwsgi.ini configuration error` issue?

dalf:
I can't figure out why there is a new error.

Can you try https://github.com/dalf/searxng/tree/test-uwsgi ?
This is the master branch + upgrade-httpx + small change to log the uwsgi exception (in addition to the existing generic message).

After running the server, uwsgi.ini should contain a more useful message.

HLFH:
```
déc. 28 11:03:55 arch-server uwsgi[1731476]: 2021-12-28 11:03:55,287 ERROR:searx.shared: uwsgi.ini configuration error, add this line to your uwsgi.ini
déc. 28 11:03:55 arch-server uwsgi[1731476]: cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1
déc. 28 11:03:55 arch-server uwsgi[1731476]: Traceback (most recent call last):
[...]

PR #1513, opened 2022-07-15
Add searx/shared/shared_redis

dalf:
## What does this PR do?

Implements:
* a scheduler on Redis: this is minimal implementation, the purpose is not to schedule hundred of callbacks.
* shared_abstract.SharedDict

`searxng.shared` choose:
* first, Redis
* then uwsgi if it works
* then a simple dictionary.

So uwsgi won't be use anymore if Redis is available.

After a time, we can remove the uwsgi implementation.

currently we use:
* Redis for the limiter plugin
* uwsgi for the result of the checker

Using docker:
* build the image
* enable the checker
* wait for the checker to have finished
* check `/stats/checker`

Otherwise this code should run:
```python
  from .shared import schedule
  from time import time

  p = time()
  q = time()

  def h():
      nonlocal p
      n = time()
      print('hh', n - p)
      p = n

  def g():
      nonlocal q
      n = time()
      print('gg', n - q)
      q = n

  schedule(4, h)
  schedule(10, g)
```

## Author's checklist

`make test` should use something like https://pypi.org/project/fakeredis

## Related issues

`searx/shared/shared_uwsgi.py` seems to creates some issues when deployed: 
* https://github.com/searxng/searxng-docker/issues/50 
* https://github.com/searxng/searxng-docker/issues/29
* https://github.com/searxng/searxng-docker/issues/37
* https://github.com/searxng/searxng-docker/issues/31
* https://github.com/searxng/searxng/discussions/1422

Also related to:
* https://github.com/searxng/searxng/pull/1273

### DIFF
--- a/searx/shared/__init__.py
+++ b/searx/shared/__init__.py
@@ -1,39 +1,50 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 import logging
-import importlib
 
 logger = logging.getLogger('searx.shared')
 
-__all__ = ['SharedDict', 'schedule']
+__all__ = ["storage", "schedule"]
+
 
 try:
-    uwsgi = importlib.import_module('uwsgi')
-except:
-    # no uwsgi
-    from .shared_simple import SimpleSharedDict as SharedDict, schedule
+    # First: try to use Redis
+    from .redisdb import client
 
-    logger.info('Use shared_simple implementation')
-else:
+    client().ping()
+    from .shared_redis import RedisCacheSharedDict as SharedDict, schedule
+
+    logger.info('Use shared_redis implementation')
+except Exception as e:
+    # Second: try to use uwsgi
     try:
-        uwsgi.cache_update('dummy', b'dummy')
-        if uwsgi.cache_get('dummy') != b'dummy':
-            raise Exception()
+        import uwsgi
     except:
-        # uwsgi.ini configuration problem: disable all scheduling
-        logger.error(
-            'uwsgi.ini configuration error, add this line to your uwsgi.ini\n'
-            'cache2 = name=searxngcache,items=2000,blocks=2000,blocksize=4096,bitmap=1'
-        )
-        from .shared_simple import SimpleSharedDict as SharedDict
+        # Third : fall back to shared_simple
+        from .shared_simple import SimpleSharedDict as SharedDict, schedule
 
-        def schedule(delay, func, *args):
-            return False
-
+        logger.info('Use shared_simple implementation')
     else:
-        # uwsgi
-        from .shared_uwsgi import UwsgiCacheSharedDict as SharedDict, schedule
+        # Make sure uwsgi is okay
+        try:
+            uwsgi.cache_update('dummy', b'dummy')
+            if uwsgi.cache_get('dummy') != b'dummy':
+                raise Exception()  # pylint: disable=raise-missing-from
+        except:
+            # there is exception on a get/set test: disable all scheduling
+            logger.exception(
+                'uwsgi.ini configuration error, add this line to your uwsgi.ini\n'
+                'cache2 = name=searxngcache,items=2000,blocks=2000,blocksize=4096,bitmap=1\n'
+            )
+            from .shared_simple import SimpleSharedDict as SharedDict
 
-        logger.info('Use shared_uwsgi implementation')
+            def schedule(delay, func, *args):
+                return False
+
+        else:
+            # use uwsgi
+            from .shared_uwsgi import UwsgiCacheSharedDict as SharedDict, schedule
+
+            logger.info('Use shared_uwsgi implementation')
 
 storage = SharedDict()
--- a/searx/shared/__init__.py
+++ b/searx/shared/__init__.py
@@ -1,50 +1,45 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Initialization of a *shared* storage.
+
+The *shared* :py:obj:`storage` is a dictionary type that is stored:
+
+- in a redis DB.  If a redis DB is not available it is stored
+- in a uWSGI cache. If no uWSGI is not available it is stored
+- in a simple python dictionary.
+"""
 
 import logging
+from . import redisdb
 
 logger = logging.getLogger('searx.shared')
 
 __all__ = ["storage", "schedule"]
 
-
-try:
+if redisdb.init():
     # First: try to use Redis
-    from .redisdb import client
 
-    client().ping()
     from .shared_redis import RedisCacheSharedDict as SharedDict, schedule
 
     logger.info('Use shared_redis implementation')
-except Exception as e:
+
+else:
     # Second: try to use uwsgi
+
+    logger.error("can't connect redis DB, try to init uWSGI cache")
     try:
-        import uwsgi
-    except:
-        # Third : fall back to shared_simple
+        import uwsgi  # type: ignore
+
+        logger.info('Use shared_uwsgi implementation')
+        uwsgi.cache_update('dummy', b'dummy')
+        if uwsgi.cache_get('dummy') != b'dummy':
+            logger.error('found issues when using uWSGI, fall back to python dictionary')
+            from .shared_simple import SimpleSharedDict as SharedDict, schedule
+
+    except ModuleNotFoundError:
+        # Third: fall back to shared_simple
+
+        logger.error("uWSGI is not available, fall back to python dictionary")
         from .shared_simple import SimpleSharedDict as SharedDict, schedule
-
-        logger.info('Use shared_simple implementation')
-    else:
-        # Make sure uwsgi is okay
-        try:
-            uwsgi.cache_update('dummy', b'dummy')
-            if uwsgi.cache_get('dummy') != b'dummy':
-                raise Exception()  # pylint: disable=raise-missing-from
-        except:
-            # there is exception on a get/set test: disable all scheduling
-            logger.exception(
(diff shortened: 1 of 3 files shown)


return42:
> `make test` should use something like https://pypi.org/project/fakeredis

FYI: an alternative might be `./manage`, see [Tip for developers](https://docs.searxng.org/admin/engines/settings.html#redis) .. but the redis build takes a long time .. we can discuss a solution for `make test` in an other PR.

searx/shared/__init__.py
```diff
@@ -4,33 +4,47 @@
...
+        # Make sure uwsgi is okay
+        try:
+            uwsgi.cache_update('dummy', b'dummy')
+            if uwsgi.cache_get('dummy') != b'dummy':
+                raise Exception()  # pylint: disable=raise-missing-from
+        except:
+            # there is exception on a get/set test: disable all scheduling
+            logger.exception(
+                'uwsgi.ini configuration error, add this line to your uwsgi.ini\n'
+                'cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1'
```
return42: name of the cache --> `searxngcache` / https://github.com/searxng/searxng-docker/issues/31#issuecomment-1177418360

dalf: fixed

searx/shared/__init__.py
```diff
@@ -4,33 +4,47 @@
...
+        except:
+            # there is exception on a get/set test: disable all scheduling
+            logger.exception(
+                'uwsgi.ini configuration error, add this line to your uwsgi.ini\n'
+                'cache2 = name=searxngcache,items=2000,blocks=2000,blocksize=4096,bitmap=1\n'
+            )
+            from .shared_simple import SimpleSharedDict as SharedDict
+
+            def schedule(delay, func, *args):
+                return False
```
return42: Why not using (import) `.shared_simple.schedule` here? In line 24 `schedule` is imported but not here .. what is the difference?

searx/shared/__init__.py
```diff
@@ -4,33 +4,47 @@
...
-        # uwsgi
-        from .shared_uwsgi import UwsgiCacheSharedDict as SharedDict, schedule
-
-        logger.info('Use shared_uwsgi implementation')
+        # Make sure uwsgi is okay
+        try:
+            uwsgi.cache_update('dummy', b'dummy')
+            if uwsgi.cache_get('dummy') != b'dummy':
+                raise Exception()  # pylint: disable=raise-missing-from
+        except:
```
return42: Don't catch general Exceptions .. don't raise an (general) Exception when the if condition is clear.

return42 (commented):
I rebased and added a commit on top.

I haven't reviewed the redis part of this PR / first I wan't to review `searx/shared/__init__.py` and I want to understand where in the process architecture the scheduler is placed ..

image

AFAIK there is a scheduler implementation somewhere in the depths of `search.checker.background`

https://github.com/searxng/searxng/blob/04e342702279a4aee3e9ad7a06eafa92ce7dba3f/searx/search/checker/background.py#L171

TBH I haven't looked deep and I have not understand how it works and where in the architecture shown above these threads are placed.

I think this (threading) architecture was owed to uwsgi and this architecture was designed when did not had redis .. to me its very complicated .. isn't it better we drop the uWSGI support and install a new process for the scheduler.  This will simplify the code a lot.

The schedule loop can be controlled in a crontab (of the [searxng account](https://docs.searxng.org/admin/installation-searxng.html#create-user)) or we add an additional systemd unit file that starts our own scheduler implementation.

---

By example:

- https://gitlab.com/doctormo/python-crontab/ is a very simple solution using Linux's crontab
- https://pypi.org/project/scheduler/ is a in-process scheduler 

.. I would prefer a solution from OS (crontab or systemd) over a in-process solution / IMO schedule solutions from OS are made for such tasks, why implementing a own main-loop for schedule?

searx/shared/__init__.py
dalf: In some setup, `uwsgi.cache_update('dummy', b'dummy')` triggers a `SystemError`:
Related to https://github.com/searxng/searxng/issues/662#issuecomment-1002021671

searx/shared/__init__.py
dalf: the scheduler is disabled because:
* the app is managed by uwsgi (we can import `uwsgi`), so there are multiple processes.
* with the simple implementation, the scheduled task is going to run on each process.

So the best thing to do is:
* either stop the app and ask the admin to fix the issue.
* or disable the scheduler because it is only required by the checker.

return42: > So the best thing to do is:

What do think about my comment above [.. to me its very complicated .. isn't it better we drop the uWSGI support and install a new process for the scheduler. This will simplify the code a lot.](https://github.com/searxng/searxng/pull/1513#pullrequestreview-1064421701)

dalf: > This will simplify the code a lot.

* the scheduling is configured in [settings.yml](https://github.com/searxng/searxng/blob/925f3ff48748e287ed596783658a4d15df257e30/searx/settings.yml#L216-L218). How to implement that with a cron without a bash script?
* the setup script will have to deal with the different distro / OS.
* currently, if an admin stops the server, it stops everything. It won't be the case with a cron or systemd unit.
* in the end I'm afraid it will move the complexity from us to the admins.

---

Alternatively, we can remove the embedded checker and make an external project again ([previous PR about this](https://github.com/searx/searx/pull/2419) for reference).

---

the Python packages python-crontab and scheduler don't solve the requirement to run the task once per instance (not per worker). It is implemented in this PR in 110 lines of code using Redis, so the uwsgi dependency can be removed in a near future.

return42: > the Python packages python-crontab and scheduler don't solve the requirement to run the task once per instance (not per worker). It is implemented in this PR in 110 lines of code.

May be it is not clear what I have in mind about .. let me explain:

From the architecture point of view, I think it is wrong to have an *in-process* in the WEB-App where you have to be careful not to mix it with the workers (threads or whatever) of the WEB-App:

We have a WEB-App and we have other services. For the (temporary) persistence we use redis and for the scheduler we also need a service.

> in the end I'm afraid it will move the complexity from us to the admins.

It's just a question of how we orchestrate these services;

- redis is e.g. run by the operating system (systemd),
- our WEB app is also managed by systemd and
- the scheduler should also be managed  by the operating system.

The orchestration using the crontab is really a bit complicated, but it is also quite simple using systemd and define `WantedBy=` units.

By example: we have one central systemd unit and two `WantedBy=` units, one is the WEB-App (wsgi) and the other is a scheduler process.

The `WantedBy=` units can be controlled via the central service (start/stop), additionally the single services can be switched off or disabled via systemd.

I think this is an *extension* of the possibilities without the admin having to pay more attention than before.

Here is a detail example: https://alesnosek.com/blog/2016/12/04/controlling-a-multi-service-application-with-systemd/

image

dalf: > The orchestration using the crontab is really a bit complicated, but it is also quite simple using systemd and define WantedBy= units.

Some OS do not have systemd like [devuan](https://www.devuan.org/) or MacOS.

> I think it is wrong to have an in-process in the WEB-App 

Yes, I agree, that's why there is [Celery](https://docs.celeryq.dev/en/stable/) for example in the Python ecosystem.

The purpose is to remove the uwsgi requirement since multiple users reports issues with it.

This PR prepare a replacement by a Redis implementation of 110 lines [\*] that are not by the book for sure, but do the job.
[\*] lines of Python code, so we stay in the Python ecosystem without external dependencies (to the OS, to pip huge package).

return42: > Some OS do not have systemd like [devuan](https://www.devuan.org/) or MacOS.

.. and MS-Win.  IMO this is not an argument since we do not give any support on deployment on this OS .. uWSGI support by example is also unknown to us.

> Yes, I agree, that's why there is [Celery](https://docs.celeryq.dev/en/stable/) for example in the Python ecosystem.

Do we really need to implement a task queues?  The only argument - I see - could be to be independent from OS .. but is this really a job of the SearXNG app?

I would like to avoid that we always reach for new tools only because we are convinced that SearXNG has to take over tasks that are actually done by an OS.

> This PR prepare a replacement by a Redis implementation of 110 lines

If you think this PR is a good first solution that fixes an existing issue, I don't want to block the PR from a merge and we can implement a different scheduler in another PR.

But to me it becomes harder and harder to understand .. besides the complexity of being in a uWSGI environment or not, this PR adds the complexity of additional redis scripts. 

TBH I do understand only the half of all / the ` RedisCacheSharedDict` is simple, but the scheduler part is hard for me.

Anyway this might only be my problem .. if you think this PR is a valid solution, lets merge.

dalf:
The current PR can't be merged for the issue reported above (it will crash with a SystemError the app if uwsgi does not work)

With this comment in mind https://github.com/searxng/searxng-docker/issues/31#issuecomment-1221499315
I suggest to remove right away the uwsgi support: there is no solution to make it work in the docker image.

In clear text: if we do that, the checker requires redis.
ping @HLFH

return42:
@dalf can we isolate the `RedisCacheSharedDict` from this PR and implement code to store stats in the redis DB.  Its only the scheduler I am not so happy with.

dalf:
> @dalf can we isolate the RedisCacheSharedDict from this PR and implement code to store stats in the redis DB. Its only the scheduler I am not so happy with.

There is no point to do that: the checker won't work.

dalf:
uwsgi can start a background process: https://uwsgi-docs.readthedocs.io/en/latest/Mules.html

Issue #1503, opened 2022-07-13
Can't change a privacy option 


ghost:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Searx.be instance 
**What happened?**
If I set "Query in the page's title" to Disabled and save, it automatically reverts to Enabled
**Expected behavior**
My selected option should be kept in the cookies and not revert

unixfox:
I won't fix that because the oscar patched theme on my instance is in an unstable state but I'm waiting on https://github.com/searxng/searxng/issues/1169 to drop the oscar theme on my instance.

Closing as unfortunately there is no fix planned.

PR #1518, opened 2022-07-17
[fix] preferences: missing option value

tiekoetter:
## What does this PR do?

Fix missing option value "0".

Image proxy and Query in the page's title setting could not be disabled by the user.

1. `make run`
2. change Image proxy and/or Query in the page's title setting

## Related issues

https://github.com/searxng/searxng/issues/1503


dalf (approved):
LGTM.

I understand correctly, currently in the master branch, "Disabled" actually means "default value in settings.yml". So the user can only select between Enabled and default value.

With this PR, "Disabled" works as expected.

I understand correctly the users don't have to update their settings (= cookies).

PR #1529, opened 2022-07-19
searx.network: use SSL_CERT_FILE and SSL_CERT_DIR when they are defined

dalf:
## What does this PR do?

Each network can define a verify option:
* false to disable certificate verification
* a path to existing certificate.

SearXNG uses SSL_CERT_FILE and SSL_CERT_DIR when they are defined.
This is the default behavior of httpx and requests:
see https://www.python-httpx.org/environment_variables/#ssl_cert_file

Allow custom network configuration.

* Install [mitmproxy](https://pypi.org/project/mitmproxy/)

### first test
* Start mitmproxy
* in settings.yml, inside the `outgoing` section:
   * set `verify: ~/.mitmproxy/mitmproxy-ca-cert.cer`
   * set `proxy: http://localhost:8080`
* start SearXNG
* make some requests
* check mitmproxy display the requests

### second test

* start `mitmproxy`
* in settings.yml, inside the `outgoing` section:
   * remove the `verify` option
   * set `proxy: http://localhost:8080`
* `export SSL_CERT_FILE=~/.mitmproxy/mitmproxy-ca-cert.cer`
* start SearXNG
* make some requests
* check mitmproxy display the requests

### thrid test
* with the default configuration (SSL_CERT_FILE not set, no value for `verify`).
* configure an engine to send HTTP request to https://badssl.com/
   for example set this line to `base_url: 'https://badssl.com/` :
https://github.com/searxng/searxng/blob/691c0ed6b97c2480fdc1245fd09231b0b384bbf0/searx/settings.yml#L338
* make a request to the wikipedia engine: check the request fails with a SSL error.

## Author's checklist

The PR is near ready: it only requires an updated of https://docs.searxng.org/admin/engines/settings.html#outgoing

## Related issues

Issue reported on searx by @cyclaero

### DIFF
--- a/searx/network/client.py
+++ b/searx/network/client.py
@@ -26,9 +26,6 @@ else:
 logger = logger.getChild('searx.network.client')
 LOOP = None
 SSLCONTEXTS: Dict[Any, SSLContext] = {}
-TRANSPORT_KWARGS = {
-    'trust_env': False,
-}
 
 
 def get_sslcontexts(proxy_url=None, cert=None, verify=True, trust_env=True, http2=False):
@@ -74,7 +71,7 @@ def get_transport_for_socks_proxy(verify, http2, local_address, proxy_url, limit
         rdns = True
 
     proxy_type, proxy_host, proxy_port, proxy_username, proxy_password = parse_proxy_url(proxy_url)
-    verify = get_sslcontexts(proxy_url, None, True, False, http2) if verify is True else verify
+    verify = get_sslcontexts(proxy_url, None, verify, True, http2) if verify is True else verify
     return AsyncProxyTransportFixed(
         proxy_type=proxy_type,
         proxy_host=proxy_host,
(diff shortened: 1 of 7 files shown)


searx/network/client.py
```diff
@@ -26,9 +26,6 @@
 logger = logger.getChild('searx.network.client')
 LOOP = None
 SSLCONTEXTS: Dict[Any, SSLContext] = {}
-TRANSPORT_KWARGS = {
-    'trust_env': False,
```
dalf, line 30: Always true now = use the environment variables SSL_CERT_FILE and SSL_CERT_DIR when they exists.

searx/search/processors/online.py
```diff
@@ -29,7 +30,6 @@ def default_request_params():
         'data': {},
         'url': '',
         'cookies': {},
-        'verify': True,
```
dalf, line 32: Don't override the value provide by the network = the value defined in `settings["outgoing"]["verify"]`

PR #1530, opened 2022-07-20
Experimental: add baidu engine

dalf:
## What does this PR do?

This is a prototype to add the baidu engine.

Warning: the engine may never work because of CAPTCHA (it works for about ten or twenty requests).
The default behavior is to send the previous queries which baidu.py does not do.

One more engine

* `!baidu 时间`

### DIFF
--- a/searx/engines/baidu.py
+++ b/searx/engines/baidu.py
@@ -1,0 +1,169 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Bing (Web)
+
+- https://github.com/searx/searx/issues/2019#issuecomment-648227442
+"""
+
+import re
+from urllib.parse import urlencode
+from lxml import html
+from searx.utils import eval_xpath, extract_text, eval_xpath_list, eval_xpath_getindex
+from searx.network import raise_for_httperror, multi_requests, get, Request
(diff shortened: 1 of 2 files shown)


dalf:
I close this PR because this implementation, Baidu throws a CAPTCHA after few requests.
Also, there is no request to have this engine in SearXNG.

For reference:
  * https://github.com/searx/searx/issues/2842  (the related issue is in web.archive.org)
  * https://github.com/entropage/mijisou/blob/devel/searx/engines/baidu.py

return42:
@dalf I reopne this PR and mark it as DRAFT .. if someone ever comes back to this topic, we have a good starting point here / a closed PR may be overlooked.

EDIT: it was already a DRAFT :)

dalf:
See my question here: https://github.com/searxng/searxng/discussions/2684#discussioncomment-6982745

codeido:
Do you consider other Chinese search engines?

dalf:
> Do you consider other Chinese search engines?

which one do you have in mind?
If any, please create an issue, or open an discussion if you have a broader request regarding the Chinese search engines.

dalf:
Superseded by https://github.com/searxng/searxng/pull/4121

PR #1332, opened 2022-06-16
Upgrade installation scripts and documentation

return42:
A preview of the documentation about the upgrade is available from here:

https://return42.github.io/searxng/admin/update-searxng.html#migrate-and-stay-tuned

**HINT:  last patch on top needs to be removed before this PR is merged!** 

    [brand] branch: searxng-install https://github.com/return42/searxng

This brand is needed to test the installation procedures as long this PR is not merged into master.

----

## What does this PR do?

```
- [utils/searxng.sh] implement new script to install SearXNG
- [doc] update documentation of the installation procedures
- [clean up] drop obsolete searx, filtron and morty install scripts
- [docs] add section "Migrate and stay tuned!"
```

## Why is this change important?

The infrastructure of SearXNG has changed dramatically over the past year. The installation scripts and documentation have not been adapted so far.

## How to test this PR locally?

----

First read the admin documentation, to see what have been changed ..

    make docs.live

----

Test new `./utils/searxng.sh` script ..

in LXCs, first build containers and install SearXNG by `install all`:

    sudo -H ./utils/lxc.sh build
    sudo -H ./utils/lxc.sh cmd -- FORCE_TIMEOUT=0 ./utils/searxng.sh install all
    
Choose to `install nginx`:

    sudo -H ./utils/lxc.sh cmd -- FORCE_TIMEOUT=0 ./utils/searxng.sh install nginx

or `install apache` HTTP server:

    sudo -H ./utils/lxc.sh cmd -- FORCE_TIMEOUT=0 ./utils/searxng.sh install apache
     
Run `instance check`:

    sudo -H ./utils/lxc.sh cmd -- FORCE_TIMEOUT=0 ./utils/searxng.sh instance check

Now you can test the SearXNG instances from the conatiners, to get the URLs of the containers use:

    sudo -H ./utils/lxc.sh show suite | grep SEARXNG_URL
    ...
    [searxng-ubu2004]    SEARXNG_URL          : http://10.146.148.203/searxng
    [searxng-ubu2110]    SEARXNG_URL          : http://10.146.148.219/searxng
    [searxng-fedora35]    SEARXNG_URL          : http://10.146.148.64/searxng
    [searxng-archlinux]    SEARXNG_URL          : http://10.146.148.135/searxng

----

Settings see https://github.com/return42/searxng/blob/searxng-install/utils/templates/etc/searxng/settings.yml#L3-L22

Features:

[...]

PR #1538, opened 2022-07-22
pass morty key as env variable

tristanwagner:
## What does this PR do?

It modify the way we give the MORTY_KEY to morty service, previously we were passing it as a command line argument, now we are passing it as an environment variable.

I installed a searxng instance and image proxying wasn't working, all the images were returning 403 with this error : 

```
Error: invalid "mortyhash" parameter
```

so it meant that the hash verification wasn't working and I couldn't figure out why.

I checked that the key setup in settings.yml was the same in the searx service and the morty service and it was, but still didn't work.

After some research I found out that you could pass this key to morty as an environment variable, I tested it and it worked for me.

Also passing it as a command line argument will likely be logged by the system somewhere, so it's probably better to do it this way ?

### DIFF
--- a/utils/templates/lib/systemd/system/morty.service
+++ b/utils/templates/lib/systemd/system/morty.service
@@ -10,10 +10,10 @@ Type=simple
 User=${SERVICE_USER}
 Group=${SERVICE_GROUP}
 WorkingDirectory=${SERVICE_HOME}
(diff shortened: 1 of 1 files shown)


tristanwagner:
Closing this because apparently it just disable validation lol and I thought morty was mandatory for image proxying because the install docs mentionned to install it, but in README it says it's not. Seems a bit confusing to me but after commenting the result_proxy part in settings.yml it works out of the box so all good.

return42:
> because the install docs mentionned to install it

@tristanwagner sorry we haven't modified the docs and installation procedures right now, but its on the way see https://github.com/searxng/searxng/pull/1332

Issue #1555, opened 2022-07-25
Investigate the cookie CONSENT for Google


unixfox:
We need to investigate the correct parameters to send to Google so that the consent dialog is not being returned and doesn't cause an issue in parsing the results.

Currently, I don't know why, but  for google images just adding the query parameter `'ucbcb': 1` in the URL only works on Firefox but not on SearXNG. But it works in both cases for the Google engine.

Useful information:
- https://github.com/iv-org/invidious/pull/2207
- https://github.com/TeamNewPipe/NewPipeExtractor/issues/592
- https://old.reddit.com/r/youtube/comments/mvcr5q/what_is_the_ucbcb1_tag_added_to_yt_urls/
- https://github.com/searxng/searxng/pull/1487
- https://github.com/searxng/searxng/pull/1552

The ultimate goal is to send a `PENDING` consent to Google so that we don't give Google the legal rights to collect information about us.

return42:
> The ultimate goal is to send a PENDING consent to Google so that we don't give Google the legal rights to collect information about us.

AFAIK: setting `YES+` ...

https://github.com/searxng/searxng/blob/50ec821e8c5b099947d657722b1879fc8c03bcfa/searx/engines/google_images.py#L144

prevents from redirects to (e.g.) `consent.google.fr` .. compare `Error` from https://github.com/searxng/searxng/issues/1494 ...

```
Error
  * Error: 1 redirects, maximum: 0
  * Percentage: 100
  * Parameters: `('200', 'OK', 'consent.google.fr')`
  * File name: `searx/search/processors/online.py:105`
  * Function: `_send_http_request`
  * Code: `count_error(`
```

and citation from https://github.com/TeamNewPipe/NewPipeExtractor/issues/592

```
Google change the way for consent about YouTube cookies agreement
in EU countries ...
Instead of showing a popup in the website, YouTube redirects the user to a new
webpage at consent.youtube.com domain ...
Fix for this is to put a cookie CONSENT with YES+ value ...
```

This comment is about YouTube .. but (AFAIK) the cookies are the same in all google services.

unixfox:
But `YES+` tell Google that they can legal collect data about the searx instance. That's the issue.

Where as they legally can't if we send `PENDING+`

PR #1556, opened 2022-07-25
[fix] google engine: results XPath

tiekoetter:
## What does this PR do?

Fix the google engine results XPath.

Google would not working anymore.

1. `make run`
2. Search `!go time`

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -112,7 +112,7 @@ filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 # ------------------------
 
 # google results are grouped into <div class="jtfYYd ..." ../>
-results_xpath = '//div[@class="jtfYYd"]'
+results_xpath = '//div[contains(@class, "jtfYYd")]'
 
 # google *sections* are no usual *results*, we ignore them
 g_section_with_header = './g-section-with-header'
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -287,7 +287,6 @@ def request(query, params):
                 'oe': "utf8",
                 'start': offset,
                 'filter': '0',
-                'ucbcb': 1,
                 **additional_parameters,
             }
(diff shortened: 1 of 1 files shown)


return42:
First: Since it does not have a effect, I reverted 747cf1a24 (in c72d70d45) and pushed it on the branch of this PR.  @tiekoetter @unixfox I hope this is OK for you.

**I think we have two different problems with google ..**

- one is CONSENT cookie we track in https://github.com/searxng/searxng/issues/1555
- the other is the XPath problem (what XPath fits to a google response?).

This PR is about the XPath problem and **this PR is a good interim solution but I think we will need to put more work on the XPath**.

**About XPath:**

My experience is, that google has two different HTML results, there is the "first page" and the "following pages"

- The "first page" includes (more) advertising and various layouts for grouping topics (e.g. "Headlines", "maps" and more). These layouts have two effects:
  - topics mess the list of results we are looking for and make it hard to find the "real results" we are looking for
  - since the first page is full of topics (and advertising) there are less results, sometimes I see only three or four "real results" on google's first page and all the other links are related to topics and advertising. 

- The "following pages" are more like a *"real result list"* that contains less advertising and topics

This becomes clear if you search for *comercially interesting terms* on google, by example:

In SearxNG search for `!go smartphone` / in my language (DE) SearXNG build this google request:

- https://www.google.de/search?q=smartphone&hl=de&lr=lang_de&ie=utf8&oe=utf8&start=0&filter=0

.. be warned the result from google also depends on the region code your browser places in the HTTP header (my FFox sends `Accept-Language: de,en;q=0.5`).

If you open the link on google and compare the "first page" with the "following pages" you see what I described above.

**To summarize:**

1. This PR is huge step forward to improve the google results.
2. We need to invest more time in the **CONSENT cookie** and the **XPath problem**

I'm going to merge this PR ..

PR #1560, opened 2022-07-26
[mod] add 'Accept-Language' HTTP header to online processores

### LINKED ISSUE #1321: Chinese Search language problem
**Version of SearXNG**
I am using a public instance, all instances behave the same

**What happened?**
It seems that most search results under search language zh-HK 🇭🇰 are from China (in Simplified Chinese and not specific to the Hong Kong region), whereas the results under zh-TW 🇹🇼 are actually more accurate for HK users.

**How To Reproduce**

For instance, to look up "https://www.price.com.hk/" with the search term "price+hk", the first result under zh-HK is a site from China, whereas the first result under Taiwan is (closer to) the correct result.

**Expected behavior**
Search result should be from Hong Kong region when search language is set to zh-HK

**Screenshots & Logs**
173271627-091c4b2d-e371-4f80-8f9e-87bfc90fa45f
173271834-e5c3087a-f87b-476e-ae17-668059edbe3f

**Additional context**
No sure if this is a SearXNG problem or that of the engines

### LINKED ISSUE #1567: Bing news returns results in English when language set to Italian
2022.07.26-50ec821e

install SearXNG using the official wiki

**What happened?**
Bing news from SearXNG returns results in English and German when it is supposed to return them in Italian according to setting.

**How To Reproduce**
Bing news from SearXNG returns results in English: https://search.ononoki.org/search?q=!bin+meloni&category_news=on&language=it-IT
Official Bing news returns results in Italian: https://www.bing.com/news/search?q=meloni&setlang=it

**Expected behavior**
Bing news from SearXNG returns results in Italian.

return42:
## What does this PR do?

[mod] add 'Accept-Language' HTTP header to online processors

Most engines that support languages (and regions) use the Accept-Language from the WEB browser to build a response that fits to the language (and region).

    make run

compare the results of engines that have language support (e.g. google) before and after the patch.  Select languages (regions) like zh-TW and zh-CN.

The results should be slightly improved (at least in my tests and on my IP).

Closes: https://github.com/searxng/searxng/issues/1567
Closes: https://github.com/searxng/searxng/issues/1321

### DIFF
--- a/searx/engines/duckduckgo_definitions.py
+++ b/searx/engines/duckduckgo_definitions.py
@@ -27,6 +27,8 @@ about = {
     "results": 'JSON',
 }
 
+send_accept_language_header = True
+
 URL = 'https://api.duckduckgo.com/' + '?{query}&format=json&pretty=0&no_redirect=1&d=1'
 
 WIKIDATA_PREFIX = ['http://www.wikidata.org/entity/', 'https://www.wikidata.org/entity/']
@@ -62,7 +64,6 @@ def request(query, params):
     params['url'] = URL.format(query=urlencode({'q': query}))
     language = match_language(params['language'], supported_languages, language_aliases)
     language = language.split('-')[0]
-    params['headers']['Accept-Language'] = language
     return params
 
 
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -30,6 +30,7 @@ about = {
 categories = ['map']
 paging = False
 language_support = True
+send_accept_language_header = True
 
 # search-url
 base_url = 'https://nominatim.openstreetmap.org/'
@@ -142,9 +143,8 @@ def request(query, params):
     params['url'] = base_url + search_string.format(query=urlencode({'q': query}))
     params['route'] = route_re.match(query)
     params['headers']['User-Agent'] = searx_useragent()
-
-    accept_language = 'en' if params['language'] == 'all' else params['language']
-    params['headers']['Accept-Language'] = accept_language
+    if 'Accept-Language' not in params['headers']:
+        params['headers']['Accept-Language'] = 'en'
     return params
 
 
--- a/searx/engines/wikipedia.py
+++ b/searx/engines/wikipedia.py
@@ -19,6 +19,9 @@ about = {
     "results": 'JSON',
 }
 
+
+send_accept_language_header = True
+
 # search-url
 search_url = 'https://{language}.wikipedia.org/api/rest_v1/page/summary/{title}'
 supported_languages_url = 'https://meta.wikimedia.org/wiki/List_of_Wikipedias'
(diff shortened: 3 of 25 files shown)


return42:
@dalf FYI I rebased the PR and added https://github.com/searxng/searxng/pull/1560/commits/5015d43d049fec2eb3e6999abbe7a46f57585f04 on top / according to what we discussed on matrix.

searx/engines/openstreetmap.py
```diff
@@ -143,8 +143,6 @@ def request(query, params):
     params['route'] = route_re.match(query)
     params['headers']['User-Agent'] = searx_useragent()
 
-    accept_language = 'en' if params['language'] == 'all' else params['language']
-    params['headers']['Accept-Language'] = accept_language
```
dalf, line 147: * in the master branch "all" means "en" in HTTP header.
* in this PR there is no header at all.

==> 
The query `!osm :all tokyo` returns:
* `Tokyo, Japan` in the master branch
* `東京都, 日本` in this PR

Can be fixed with:
```python
if 'Accept-Language' not in params['headers']:
    params['headers']['Accept-Language'] = 'en'
```

searx/engines/wikipedia.py
```diff
@@ -41,9 +41,6 @@ def request(query, params):
     language = url_lang(params['language'])
     params['url'] = search_url.format(title=quote(query), language=language)
 
-    if params['language'].lower() in language_variants.get(language, []):
```
dalf, line 44: LGTM 👍 
I see a difference between `!wp 東京都 :zh` and `!wp 東京都 :zh-hk`

searx/engines/duckduckgo_definitions.py
```diff
@@ -62,7 +62,6 @@ def request(query, params):
     params['url'] = URL.format(query=urlencode({'q': query}))
     language = match_language(params['language'], supported_languages, language_aliases)
     language = language.split('-')[0]
-    params['headers']['Accept-Language'] = language
```
dalf, line 65: `!ddd paris :fr` returns:
* a result in French in the master branch
* a result in English in this PR. I don't see the "Accept-Language" header in the log.

With `send_accept_language_header = True` it works as before.

searx/settings.yml
```diff
@@ -322,24 +322,29 @@ engines:
   - name: wikipedia
     engine: wikipedia
     shortcut: wp
+    send_accept_language_header: true
```
dalf: Most of the engines have the configuration in the Python code rather than the .yml file
It would be better to add `send_accept_language_header = True` to wikipedia.py (and to other engines).

searx/network/network.py
```diff
@@ -265,6 +265,7 @@ async def call_client(self, stream, method, url, **kwargs):
         kwargs_clients = Network.extract_kwargs_clients(kwargs)
         while retries >= 0:  # pragma: no cover
             client = await self.get_client(**kwargs_clients)
+            self._logger.debug("HTTP request arguments--> %s", kwargs)
```
dalf: Can we discuss about this change in another PR?

* httpx might add to some additional HTTP header
* It makes the log verbose

dalf (changes requested):
except the few issues, LGTM .

searx/engines/openstreetmap.py
return42: Thanks, forgot to implement the `all` case / I will fix it.

It is OT here but what me often confuse is  the designation "Default language" and the `:all` tag ...

image

return42:
> except the few issues, LGTM .

Thanks for review .. I fixed the issues you reported.

PR #1565, opened 2022-07-27
Increase lingva engine timeout

liimee:
## What does this PR do?

This PR increases the timeout of the lingva engine

3 seconds isn't enough for lingva (returns timeout), so this PR increases it to 5 seconds

search `!lv en-ja test`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -889,6 +889,7 @@ engines:
   - name: lingva
     engine: lingva
     shortcut: lv
+    timeout: 5.0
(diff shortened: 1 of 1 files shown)


return42:
Hi @liimee, 

> 3 seconds isn't enough for lingva (returns timeout), so this PR increases it to 5 seconds

On my instance lingva has a (median) response time of 1.2 sec.  I think you should better set the timeout in your `/etc/searxng/settings.yml`.

PR #1595, opened 2022-07-31
[fix] uWSGI: increase buffer-size

tiekoetter:
## What does this PR do?

4096 as buffer-size is too small and will result in the preference urls not working.

1. Install with uWSGI
2. Try importing the "Search URL of the currently saved preferences" and/or "URL to restore your preferences in another browser" URL.


return42 (approved):
Thanks a lot / I tested this patch on my instance / FYI:

- I fixed the `:socket` templates also
- I added a documentation on top / see: 48968bf46

PR #1623, opened 2022-08-06
[mod] engine yep.com: show all 100 results yep.com has

### LINKED ISSUE #1619: Add pagination for Yep
After Bing, Yandex and Google, Yep is in my opinion the best independent search engine. Mojeek's results may sometimes be better ordered given certain queries but Yep's index is no doubt larger as it is powered by the data of Ahrefs which is one of the biggest SEO companies in the world and thus Yep has a huge backlink index. Engines like Mojeek and Gigablast take a lot of time to get the latest results but Yep fetches/indexes them pretty quickly. For copyrighted works or results that would normally be censored at Google or even Yandex, Yep is really great. Unfortunately, we only get around 10 results or something for most queries on SearXNG with Yep and the tremendous potential of this engine is therefore neglected. Adding pagination to Yep would improve SearXNG by a lot and you would notice it, believe me. Especially if you use Yep with the "query in the page's title" parameter enabled which helps sort out content unrelated to your search query.

The next set of results is attained by clicking the "Show more" button. Yep uses some javascript trickery to fetch the results in a roundabout way using the API. **You might be surprised to learn, however, that we can get 100 results in a *single* shot through a link as follows:**

https://api.yep.com/fs/1/?client=web&limit=100&no_correct=false&q=searxng&type=web

If you open this click in Firefox, the browser will order the JSON source code for you with all the 100 links. If we set the 'limit' parameter to 10, we may get 10 results, if we set it to '20', we may get 20. And we can keep doing this until we reach 100 and then we won't see more results than 100 at that point. But if we use the "show more" button on the website, sometimes it's possible to get a lot more results than 100 but it depends. If there's a delay between each time we click the button, we can sometimes get more results than 100 but this is becoming increasingly rare according to my experience, perhaps someone else can confirm or check.

**Nonetheless, even 100 results from Yep would be a great improvement to SearXNG and this should be added as soon as possible.**

Here is the javascript file which the API uses for fetching the results when we click on the "show more button": https://cdn.yep.com/en/assets/js/62.1c84b76d.js

Nothing.

**Additional context**
Perhaps some Javascript master can come up with a creative trick or a solution to fetch more than 100 results using the API...

return42:
## What does this PR do?

engine yep.com: show all 100 results yep.com has

yep.com is still in beta, the api.yep.com does not have paging support.  There
is only a 'limit' argument with a maximum of 100 results.

yep.com seems fast; there is nor need for a timeout of 12 sec.

The API returns JSON nevertheless what the HTTP header is, the "show more"
button on yep.com's web site does not set a special HTTP Accept header.

FYI: The index does not support languages, the WEB UI does not offer a language
selection of the results and the entire index seems in English.

Closes: https://github.com/searxng/searxng/issues/1619

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -422,18 +422,13 @@ engines:
     categories: general
     disabled: true
     paging: false
-    page_size: 10
     content_html_to_text: true
     title_html_to_text: true
-    search_url: https://api.yep.com/fs/1/?type=web&q={query}&no_correct=false
+    search_url: https://api.yep.com/fs/1/?type=web&q={query}&no_correct=false&limit=100
     results_query: 1/results
     title_query: title
     url_query: url
     content_query: snippet
-    timeout: 12.0
-    headers:
-      'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8'
-      'Sec-Fetch-Dest': 'document'
     about:
       website: https://yep.com
       use_official_api: false


return42:
@allendema could you have a look at this PR / thanks!

allendema:
Yes, tested it and the engine works.

P.s They include a result which redirects to another upstream engine ```!go```.

return42:
> Yes, tested it and the engine works.

Thanks for your test :+1:

> P.s They include a result which redirects to another upstream engine ```!go```.

Sorry, I'm a little slow on the uptake, what do you mean by that?

allendema:
```!yep turing test```

When you search for something -  in results will most likely be a result which says:
 
"Not what you're looking for?"
https://www.g00gle.com/search?q=turing+test

return42:
Ah, OK .. seems yep index hasn't sorted out crap .. TBH I think the index is still in beta (IMO alpha) .. without a language selection, this is not the much-cited "competitor to google" .. lets have a look how it involves over the time :crossed_fingers:

PR #1633, opened 2022-08-08
[fix] google engine: results XPath

tiekoetter:
## What does this PR do?

Fix google engine results XPath.

## How to test this PR locally?

1. `make run`
2. Search `!go test`

Closes: https://github.com/searxng/searxng/issues/1628

PR #1643, opened 2022-08-10
Revert Google XPath.

vojkovic:
## What does this PR do?

Reverts the XPath from yesterday. Google changed it back recently and the Google Engine is currently not working.

Reverts https://github.com/searxng/searxng/pull/1633

1. `make run`
2. `Search !go test`

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -113,7 +113,7 @@ filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 # ------------------------
 
 # google results are grouped into <div class="jtfYYd ..." ../>
-results_xpath = '//div[contains(@class, "kvH3mc")]'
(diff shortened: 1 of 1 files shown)


tiekoetter:
@vojkovic Could you mention the PR or the commit hash of the reverted change in your commit message?

tiekoetter (approved):

blahgeek:
Google changed it again (not sure if it's rolled out globally)

return42:
To be more robust in eb02cc77 I selected ``results_xpath = '//div[contains(@class, "MjjYud")]'`` / two lines above the one you marked in your screen shot.

Do you have any issue with with a SearXNG instance?

blahgeek:
Ah thanks for your reply. I checked and the current version works. I was using an old version and manually applied the patch.

PR #1648, opened 2022-08-10
output format protobuf to HTML for google mobile

unixfox:
## What does this PR do?

This switch the output format from protobuf to HTML for the Google Mobile UI.

Previously with the protobuf output, the HTML parser had to take into account things that were not HTML data. With this new parameter the output is a "clean" HTML output that the parser can "easily" parse.

On top of that the size returned is so much smaller, like 4 times smaller, but the response time is still the same.

Launch SearXNG with the parameter `use_mobile_ui` set to `true` for google engine:
```yml
- name: google 
   engine: google 
   shortcut: go
   # see https://docs.searxng.org/src/searx.engines.google.html#module-searx.engines.google 
   use_mobile_ui: true 
```

## Additional info

The URL used by SearXNG in the `use_mobile_ui` mode (example: `https://www.google.com/search?q=ok&start=0&asearch=arc&async=use_ac:true,_fmt:html`) can now also be used in order to diagnose any breakage in the XPath parsing.

## Author's checklist

N/A

## Related issues

N/A

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -112,21 +112,14 @@ filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 # specific xpath variables
 # ------------------------
 
-# google results are grouped into <div class="jtfYYd ..." ../>
-results_xpath = '//div[contains(@class, "jtfYYd")]'
+results_xpath = '//div[contains(@class, "MjjYud")]'
+title_xpath = './/h3[1]'
+href_xpath = './/a/@href'
+content_xpath = './/div[@data-content-feature=1]'
 
 # google *sections* are no usual *results*, we ignore them
 g_section_with_header = './g-section-with-header'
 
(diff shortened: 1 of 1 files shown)


tiekoetter (approved):

return42:
@unixfox thanks a lot .. this is a great help .. based on your fix I tried to improve the XPath selectors .. I pushed the commit on top of your branch .. would you (and @tiekoetter) please give it a try?

---

with the modified XPath selectors now I have a result list for e.g. "!go Paris :de" .. before I did not have any result ..

tiekoetter:
> now I have a result list for e.g. "!go Paris :de"

Can confirm that this works now.

unixfox:
> with the modified XPath selectors now I have a result list for e.g. "!go Paris :de" .. before I did not have any result ..

Works fine for me and works great with the search `!go Paris :de`

PR #1650, opened 2022-08-10
Add lib.rs engine

v3rmine:
## What does this PR do?

Adds the lib.rs engine

It would be convenient for Rust users to search for package directly.

search `!lrs fuse`

## Related issues


return42 (approved):
works like a charm :+1: .. thanks a lot!

Issue #431, opened 2021-10-22
[question] which (mobile) browsers support OpenSearch?


return42:
[Autodiscovery of search plugins](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#autodiscovery_of_search_plugins) gives the user the ability to easily add alternative (and privacy-respecting) search engines, including autocompletion (and other) features.

Common desktop browsers (e.g. [Firefox](https://support.mozilla.org/en-US/kb/add-or-remove-search-engine-firefox),  [Chromium](https://www.chromium.org/tab-to-search) or [MS-Edge](https://support.microsoft.com/en-us/microsoft-edge/change-your-default-search-engine-f863c519-5994-a8ed-6859-00fbc123b782)) offer [Autodiscovery of search plugins](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#autodiscovery_of_search_plugins) based on the [OpenSearch protocol](https://github.com/dewitt/opensearch).  But it is very hard to find a mobile browser that can do that too ...

I opened this issue to collect more informations about browser that support OpenSearch, especially I am looking for moobile (Android) alternatives, since my favourite browser sadly does not [[1]](https://github.com/mozilla-mobile/fenix/issues/21883) [[2]](https://github.com/mozilla-mobile/fenix/issues/8889) [[3]](https://f-droid.org/en/packages/org.mozilla.fennec_fdroid/) [[4]](https://github.com/mozilla-mobile/fenix/issues/6615#issuecomment-1027947583).

The intention of this issue is to collect some recommendations we can add to our documentation / hints are welcome ...

Related:

- https://github.com/searxng/searxng/pull/1655
- https://github.com/searxng/searxng/discussions/1288
- https://github.com/searxng/searxng/issues/1613#issuecomment-1214293533
- POST vs GET --> some broswers don't like HTTP POST requests in a OpenSearch query / e.g. Chrome on iOS : https://app.element.io/#/room/#searxng:matrix.org/$4_hkVaUiVBW93F2uh8YcDY3J3SKIh5ERs8pkUqyQP0E
- https://github.com/searxng/searxng/issues/2086#issuecomment-1374804339
- https://github.com/searxng/searxng/discussions/1866
- https://github.com/searxng/searxng/discussions/2133#discussioncomment-4956291
- https://github.com/searxng/searxng/discussions/1288
- https://github.com/searxng/searxng/issues/2190
- https://github.com/searxng/searxng/issues/546
[...]

PR #1655, opened 2022-08-11
[fix] improve OpenSearch description

return42:
## What does this PR do?

Inspired by the OpenSearch description from qwant and with informations from the
specification[4] the ``opensearch.xml`` has been *improved*.

- convert `<Url>` methods from lower case to upper case (`POST`|`GET`)
- add `<moz:SearchForm>` and `xmlns:moz="http://www.mozilla.org/2006/browser/search/"`
- add `<Query role="example" searchTerms="SearXNG" />`  [4]
  OpenSearch description documents should include at least one Query element of `role="example"` that is expected to return search results. Search clients may use this example query to validate that the search engine is working properly.

- modified `<LongName>` to SearXNG
- modified `<Description>` the word 'hackable' scares uninitiated users and was removed
- add the `type="image/png"` to `<Image>`

[1] https://app.element.io/#/room/#searxng:matrix.org/$xN_abdKhNqUlgXRBrb_9F3pqOxnSzGQ1TG0s0G9hQVw
[2] https://github.com/searxng/searxng/issues/431
[3] https://developer.mozilla.org/en-US/docs/Web/OpenSearch
[4] https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-query-element

Some HTTP-Clients do have issues with the ``opensearch.xml`` from SearXNG
(related [1][2]) while other OpenSearch descriptions[3] (e.g. from qwant) work
flawles.

   make run

Visit http://127.0.0.1:8888/ and add the search engine to your WEB-Browser / test with different WEB-Browser from desktop and Smartphones (are there any iOS user here, please test on Safari and Chrome).

### DIFF
--- a/searx/templates/simple/opensearch.xml
+++ b/searx/templates/simple/opensearch.xml
@@ -1,22 +1,21 @@
 <?xml version="1.0" encoding="utf-8"?>
-<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
   <ShortName>{{ instance_name }}</ShortName>
-  <Description>a privacy-respecting, hackable metasearch engine</Description>
+  <LongName>SearXNG metasearch</LongName>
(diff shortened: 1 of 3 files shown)


tiekoetter (approved):
LGTM

PR #1661, opened 2022-08-14
Add twitter engine

liimee:
## What does this PR do?

Adds the twitter (user + post search) engine

this would be a nice addition imo, especially since twitter is a huge platform

search `!tw elon musk`

## Author's checklist

I don't think this will return good results, but it works :man_shrugging:

### DIFF
--- a/searx/engines/twitter.py
+++ b/searx/engines/twitter.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Twitter (microblogging platform)"""
+
+from json import loads
+from urllib.parse import urlencode
+from datetime import datetime
+
+about = {
+    "website": 'https://twitter.com',
+    "wikidata_id": None,
+    "official_api_documentation": 'https://developer.twitter.com/en/docs/twitter-api',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['social media']
+
+url = "https://api.twitter.com"
+search_url = (
+    "{url}/2/search/adaptive.json?{query}&tweet_mode=extended&query_source=typed_query&pc=1&spelling_corrections=1"
+)
+
+
+def request(query, params):
+    params['url'] = search_url.format(url=url, query=urlencode({'q': query}))
+
+    params['headers'] = {
+        # https://github.com/zedeus/nitter/blob/5f31e86e0e8578377fa7d5aeb9631bbb2d35ef1e/src/consts.nim#L5
+        'Authorization': (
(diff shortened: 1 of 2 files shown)


dalf:
LGTM, thank you!

It is possible to add thumbnails but the layout is not great (the thumbnails should be on the right side, side for the news, etc..., it can be improved in a second time):
```python
        img_src = tweet.get('extended_entities', {}).get('media', [{}])[0].get('media_url_https')
        if img_src:
            img_src += "?name=thumb"
        results.append(
            {
                'url': 'https://twitter.com/i/web/status/' + tweet['id_str'],
                'title': (text[:40] + '...') if len(text) > 40 else text,
                'content': text[display[0] : display[1]],
                'img_src': img_src,
                'publishedDate': datetime.strptime(tweet['created_at'], '%a %b %d %H:%M:%S %z %Y'),
            }
        )
```

dalf (approved):

liimee:
why not use `'thumbnail': img_src` instead?

searx/engines/twitter.py
```diff
@@ -0,0 +1,63 @@
...
+def request(query, params):
+    params['url'] = search_url.format(url=url, query=urlencode({'q': query}))
+
+    params['headers'] = {
+        # https://github.com/zedeus/nitter/blob/5f31e86e0e8578377fa7d5aeb9631bbb2d35ef1e/src/consts.nim#L5
+        'Authorization': (
+            "Bearer AAAAAAAAAAAAAAAAAAAAAPYXBAAAAAAACLXUNDekMxqa8h%2F40K4moUkGsoc%3DTYfbDKb"
+            "T3jJPCEVnMYqilB28NHfOPqkca3qaAxGfsyKCs0wRbw"
+        )
+    }
```
return42: @liimee beside the link to nitter sources, can you explain Twitter’s Bearer authentication token a little?

liimee: I believe that's what twitter.com uses, but I don't know anything beyond that. will add it

dalf: In a private browser window (so without being logged), with the network console opened, go to https://twitter.com/searxng or any twitter URL).

Display only the XHR requests, for example the one to `https://twitter.com/i/api/1.1/attribution/event.json`.
In the sent HTTP headers, you will see the `authorization` header.

The value of token has never changed as far I know (or may be once a long time ago).

Without this header, the API doesn't work.

I realize this is not the value from Browser which starts with `AAAAAAAAAAAAAAAAAAAAANRIL` and this value was used by nitter 7 months ago ( https://github.com/zedeus/nitter/blame/79b98a80817f91e1c4cd76bcb97a15b52b885f0e/src/consts.nim ).
I'm not sure why, all I know is that this value rarely changes

dalf (approved):
LGTM

PR #1667, opened 2022-08-14
"regions & languages" boilerplate

### LINKED ISSUE #1081: Bug: startpage engine

return42:
## What does this PR do?

This PR is a revise of PR https://github.com/searxng/searxng/pull/1126 / for any details, please read the initial comment in this PR https://github.com/searxng/searxng/pull/1126 

The main difference to https://github.com/searxng/searxng/pull/1126 is:

- To fetch languages & regions from startpage, this patch makes use of the `EngineProperties` implemented in 7bf0d46c

- To get Startpage's locale & language for requests, the function get_engine_locale from 9ae409a is used.

To test select various languages and regions in startpage queries / by example::

    !sp paris :zh-TW

You should not get a CAPTCHA and you should see differences in the result lists from (e.g )`zh-TW` and `zh-CN`

---

@dalf in PR 1126 you added commit https://github.com/searxng/searxng/pull/1126/commits/7f78b69e86c2b60ba5fba7d5dfb37ad938e85dc1

     Add searx.locales.SupportedLocales 

on top, you described in https://github.com/searxng/searxng/pull/1126#issuecomment-1107457521 I did not applied this commit here in this branch since it breaks the property aspect what is IMO not directly bound to a language and region handling (done in your `SupportedLocales` class). 

> Since this class exists, the function engine_properties_template is not required anymore

Instead of `SupportedLocales` I added the commit:

    [mod] engines_languages.json: add new type EngineProperties

which implements a class for the properties of an engine.  If you think there is also a need for a SupportedLocales lets discuss it in another task .. this PR is already very overloaded :-)

BTW: in this PR, beside startpage I also normalized "qwant" to the make use of the EngineProperties 

---

@tiekoetter the custom commit in your master branch at https://github.com/tiekoetter/searxng/commits/master 

    [WIP] [fix] startpage engine
    
    Squashed commits from PR 1126 by Markus Heiser <markus.heiser@darmarit.de>

Can be dropped when this PR is merged (since this PR is just a revise of PR-1126). 

---

In the last commit:

Closes: https://github.com/searxng/searxng/issues/1081


searx/engines/__init__.py
```diff
@@ -52,22 +51,52 @@
...
+        'ca' : <engine's language name>
+    },
+    """
+
+    def asdict(self):
+        return {
+            'type': 'engine_properties',
+            'regions': self.regions,
+            'languages': self.languages,
+        }
```
dalf: It would be better to group dumps to & loads from JSON in one place.

AFIK, the usual pattern is to have a dedicate class or module to do that (`json.loads` & `json.dumps` for example, same for pickle).
It is possible to take a shortcut and define the dumps & loads as two methods of the EngineProperties class (it is actually nearly what you wrote with `asdict`):

```python
@dataclasses.dataclass
class EngineProperties:
  regions: Dict[str, str] = dataclasses.field(default_factory=dict)
  """...."""
  languages: Dict[str, str] = dataclasses.field(default_factory=dict)
  """...."""

  def dumps(self): 
      # this is "asdict()" renames into "dumps".
      # after a second thought , the name asdict() might be good in this context (I will not update all the comments, feel free to pick the name you want). In this case the method "loads" must be renamed into "fromdict".
      # this method serializes all the properties of the instance: there is no need to update this method when a new field is added
      return {
          'type': 'engine_properties',
          **dataclasses.asdict(self)
      }

  @classmethod
  def loads(cls, obj):
    if not isinstance(obj, dict) or obj.get("type") != "engine_properties":
        # return None can simplify the usage (see below)
        raise ValueError('EngineProperty: invalid format')
    # deserialize all the properties
    # avoid side effect: we don't change "obj"
    return cls(**{k:v for k,v in obj.items() if k != 'type'})
```

<details>
    <summary>( alternative with a dedicated class )</summary>

Here the methods loads and dumps are move to a dedicated class to follow the usual pattern: `json.dumps`, `json.loads`, `pickle.dumps`, `pickle.loads`.

```python
class EnginePropertyJson:
  @staticmethod
  def dumps(self, obj): 
    # serialize all the properties of the instance: there is no need to update this method when a field is added
    return {
        'type': 'engine_properties',
        **obj.as_dict()
    }

  @staticmethod
  def loads(obj):
    if not isinstance(obj, dict) or obj.get("type") != "engine_properties":
        raise ValueError('EngineProperty: invalid format')
    # deserialize all the properties
    return EngineProperties(**{k:v for k,v in obj.items() if k != 'type'})
```

</details>

Then `update_languages.py` can rely on that, so instead of:
```python
def fetch_supported_languages():
    ...
    engine_properties = engine_properties.asdict()
    ...

def join_language_lists(engines_languages):
    ...
    if isinstance(engine_properties, dict) and engine_properties.get('type') == 'engine_properties':
        # engine_properties is an EngineProperties instance stored as dictionnary.
    else:
        # engine_properties is NOT an EngineProperties
```
the code is
```python
def fetch_supported_languages():
    ...
    engine_properties = engine_properties.dumps()
    ...

def join_language_lists(engines_languages):
    ...
    try:
        engine_properties = EngineProperties.loads(engine_properties)
    except ValueError:
        # engine_properties is NOT an EngineProperties
    else:
        # engine_properties is an EngineProperties instance stored as dictionnary.
```

The drawback is the more verbose try..except..else blocks than a simple if.

Also, see below for a change on `set_language_attributes` in this file.

searx/engines/__init__.py
```diff
@@ -52,22 +51,52 @@
...
     about: dict
     inactive: bool
     disabled: bool
+    # language support, either by selecting a region or by selecting a language
     language_support: bool
     paging: bool
     safesearch: bool
     time_range_support: bool
     timeout: float
+    properties: EngineProperties
```
dalf: according to line 234, it must be:
`supported_properties: EngineProperties`

---

Actually the name `supported_properties` is confusing: `paging` could be seen a supported property because it tells whether or not the engine supports paging. However paging is set manually by us, the content of `EngineProperties` is updated automatically.

Without strong conviction, I think `EngineData` would better since it contains data from the `searx.data` module.

```python
class Engine:
    ...
    data: EngineData
```

searx/engines/__init__.py
```diff
@@ -184,56 +213,60 @@ def update_engine_attributes(engine: Engine, engine_data):
...
-            # pylint: disable=protected-access
-            lambda: engine._fetch_supported_languages(get(engine.supported_languages_url, headers=headers))
-        )
+        supported_properties = ENGINES_LANGUAGES[engine.engine]
+
+    if not supported_properties:
+        return
+
+    if isinstance(supported_properties, dict) and supported_properties.get('type') == 'engine_properties':
+        engine.supported_properties = supported_properties
```
dalf: This comment is only on the line 234: this makes the dataclass definition useless because with this line `engine.supported_properties` is a dictionary instead of an instance of  the `EngineProperty` class.

To create an instance of EngineProperty, the line 234 must be:
```python
engine.supported_properties = EngineProperty.loads(supported_properties)
```
(the code is based on the suggestion above, but the point here is to create an instance of EngineProperty by whatever mean).

So:
* `EngineProperty` doesn't have to inherit from `dict`.
* `supported_properties['regions']` can be replaced by `supported_properties.regions` which pyright is going to check (and IDE are going to suggest auto-completion).

searxng_extra/update/update_languages.py
```diff
@@ -21,32 +35,67 @@
...
-            print("fetched %s languages from engine %s" % (len(engines_languages[engine_name]), engine_name))
-            if type(engines_languages[engine_name]) == list:  # pylint: disable=unidiomatic-typecheck
-                engines_languages[engine_name] = sorted(engines_languages[engine_name])
+        engine = engines[engine_name]
+        fetch_languages = getattr(engine, '_fetch_supported_languages', None)
+        fetch_properties = getattr(engine, '_fetch_engine_properties', None)
+
+        if fetch_properties is not None:
+            resp = network.get(engine.supported_properties_url, headers=headers)
+            engine_properties = EngineProperties()
```
dalf: \<Nit\>: I guess the purpose is avoid the engine to have `from searx.engines import EngineProperties` to then create an instance of this class ; however it would be nice to have typing on the `_fetch_engine_properties`  function.

Currently it can be in the engine:
```python
from searx.engines import EngineProperties

def _fetch_engine_properties(resp, engine_properties: EngineProperties):
```

This is close to

```python
from searx.engines import EngineProperties

def _fetch_engine_properties(resp) -> EngineProperties:
  data = EngineProperties()
  ...
  return data
```

\</nit\>

You can close this comment without any further comment if you want to because the `request` and `response` functions don't have typing anyway.

searx/search/processors/online.py
```diff
@@ -212,16 +212,6 @@ def get_default_tests(self):
                 'test': ['unique_results'],
             }
 
-        if getattr(self.engine, 'supported_languages', []):
```
dalf, line 215: I don't understand why these tests are removed.

For example, the language selection of the Qwant engine can fail.

searx/engines/__init__.py
```diff
@@ -184,56 +213,60 @@ def update_engine_attributes(engine: Engine, engine_data):
...
-        engine.fetch_supported_languages = (
-            # pylint: disable=protected-access
-            lambda: engine._fetch_supported_languages(get(engine.supported_languages_url, headers=headers))
-        )
+        supported_properties = ENGINES_LANGUAGES[engine.engine]
+
+    if not supported_properties:
+        return
+
+    if isinstance(supported_properties, dict) and supported_properties.get('type') == 'engine_properties':
```
dalf: To take advantage of `EngineProperty.loads` :

```python
def set_language_attributes(engine: Engine):
	...
	try:
		engine.supported_properties = EngineProperty.loads(supported_properties)
		engine.language_support = len(supported_properties['languages']) or len(supported_properties['regions'])
	except ValueError:
		set_language_attributes_deprecated_way(engine, supported_properties)
		
def set_language_attributes_deprecated_way(engine: Engine, supported_properties: dict):
	# depricated: does not work for engines that do support languages
	# based on a region.
```

In the future, we can remove `set_language_attributes_deprecated_way`  easily.

dalf (changes requested):
I've tried to make each request for changes independent of each other, but I failed way too often.
Perhaps to get the whole picture, it would be helpful if I add a commit, even if we remove or rewrite it later on.

All in all, this opens the door to better support of the regions & locales.

searx/engines/__init__.py
return42: > It would be better to group dumps to & loads from JSON in one place.

After a second thought I have a different opinion:

This is not the loader of the JSON file, this is a set of engine-properties

hint: "engine data" fits better and is less misleading as you described .. I will use the term engine-data in the following (later I will send a commit to rename EngineProperty to Engine Data).

The sum of engine-data items from all engines is stored in a JSON file but on engine level this is a `@dataclasses.dataclass`.  

> AFIK, the usual pattern is to have a dedicate class or module to do that (`json.loads` & `json.dumps` for example, same for pickle). It is possible to take a shortcut and define the dumps & loads as two methods of the EngineProperties class (it is actually nearly what you wrote with `asdict`):

Store/load JSON files is not in the scope of this dataclass (the dataclass is per engine, the JSON file stores dataclasses of all engines). 

The intention of `EngineProperties.asdict` here is to add an additional data (the type):

```python
    def asdict(self):
        return {
            'type': 'engine_data',
            ...
        }
```

Later in the JSON file, we need this type to distinguish between the vintage JSON entries and the new entries of EngineData.

> 
> ```python
> @dataclasses.dataclass
> class EngineProperties:
>   regions: Dict[str, str] = dataclasses.field(default_factory=dict)
>   """...."""
>   languages: Dict[str, str] = dataclasses.field(default_factory=dict)
>   """...."""
> 
>   def dumps(self): 
>       # this is "asdict()" renames into "dumps".
>       # after a second thought , the name asdict() might be good in this context 
>       # (I will not update all the comments, feel free to pick the name you want).
>       # In this case the method "loads" must be renamed into "fromdict".
> ```

There is no need of a load function (this is dataclass not JSON) and the engine-data object is build by the constructor `EngineData(**engine_data)`  .. while `engine_data` is the dict from one item (engine) in the JSON file.

> ```python
>       # this method serializes all the properties of the instance:
>       # there is no need to update this method when a new field is added
>       return {
>           'type': 'engine_properties',
>           **dataclasses.asdict(self)
>       }
> ```

Good point .. I will fix the method :+1:


> ```python
>   @classmethod
>   def loads(cls, obj):
>     if not isinstance(obj, dict) or obj.get("type") != "engine_properties":
>         # return None can simplify the usage (see below)
>         raise ValueError('EngineProperty: invalid format')
>     # deserialize all the properties
>     # avoid side effect: we don't change "obj"
>     return cls(**{k:v for k,v in obj.items() if k != 'type'})
> ```
> 
> ( alternative with a dedicated class )

I think we should stay with dataclass, dataclass fits very well to this use case.

> Then `update_languages.py` can rely on that, so instead of:
> ...
> the code is
> 
> ```python
> def fetch_supported_languages():
>     ...
>     engine_properties = engine_properties.dumps()
>     ...
> ```

As you said in your comment above: _after a second thought , the name asdict() might be good in this context_

> ```python
> 
> def join_language_lists(engines_languages):
>     ...
>     try:
>         engine_properties = EngineProperties.loads(engine_properties)
>     except ValueError:
>         # engine_properties is NOT an EngineProperties
>     else:
>         # engine_properties is an EngineProperties instance stored as dictionnary.
> ```
> 
> The drawback is the more verbose try..except..else blocks than a simple if.

In the ``fetch_supported_languages(..)`` function there is already build a instance of `EngineData`:

```python
def fetch_supported_languages():
    ....
    for engine_name in names:
        ...
        fetch_data = getattr(engine, '_fetch_engine_data', None)

        if fetch_data is not None:
            resp = network.get(engine.supported_properties_url, headers=headers)
            engine_data = EngineData()
```

And in the ``join_language_lists(..)`` we have this instance of `EngineData` here:

```python
def join_language_lists(engines_languages):
    ...
    for engine_name in engines_languages:
        engine = engines[engine_name]
        engine_data = engines_languages[engine_name]

        if isinstance(engine_data, dict) and engine_data.get('type') == 'engine_data':
            # items of type 'engine_data' do have regions & languages, the
            # list of engine_codes should contain both.
            engine_codes = engine_data.regions or {}  # <---- !!! EngineData instance !!!
```

IMO: We can rewrite (simplify) the ``searxng_extra/update/update_languages.py`` when all engine languages are ported to `EngineData`.

> Also, see below for a change on `set_language_attributes` in this file.

I will do it .. but I think your comment there based on the same confusion of EngineData (per engine) and JSON (persistence of all engine-data objects).

Issue #1642, opened 2022-08-09
Google search internal API with JSON results


unixfox:
After reverse engineering the `Google images for Google go` internal API in #1641, I had a better understanding of the undocumented parameters used by Google.

So I looked again at the issue #159, and I found that if you add the parameter `_fmt:json` in the query string `async=` you basically get the Google search results in JSON.

Here is how to do the request, it's very similar to #159, only the parameter `_fmt:json` need to be added:

You need to send a GET request to: `https://www.google.com/search?q=yeah&start=0&asearch=arc&async=use_ac:true,amw:true,_fmt:json` and with the headers:
- user-agent: `Mozilla/5.0 (X11; Linux x86_64; rv:91.0) Gecko/20100101 Firefox/91.0`
- accept: `*/*`

Need to understand the meaning of `amw:true`

Then you get a bunch of JSON objects all over the place, mainly what we are interested is in the object `arcResponse => search_results => result_list`. I've taken an example object of the interesting data for one result:
```json
"payload":{
   "encrypted_docid":"PuQ9p5ome90J",
   "has_snippet":true,
   "has_video":false,
   "is_first_result_in_group":false,
   "is_first_result_in_whole_page_cluster":false,
   "is_leaf_result":false,
   "is_video_result":false,
   "primary_link":{
      "is_clickable":true,
      "news_result_signals":{
         "host_newsiness":0
      },
      "should_truncate_title":true,
      "title":"\u003cb\u003eJSON\u003c\/b\u003e.parse() - \u003cb\u003eJavaScript\u003c\/b\u003e - MDN Web Docs",
      "url":"https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse",
      "volt_result_data":{
         "is_volt":true
      }
   }
}
```

Other parameters in the query strings can be used too, all the ones from the current google engine implementation.

More services from Google like Google videos, Google News and Google images could be used on this API:

- Google videos, add `tbm=vid`
- Google News: `tbm=nws`
- Google images: `tbm=isch`

Additional note: Getting the results in JSON result in increase latency, maybe we could look into integrating protobuf so that we can process the results in protobuf format in order to decrease the latency by around 300-400ms. We don't have the [proto files](https://developers.google.com/protocol-buffers/docs/proto3), but we can reconstruct them from the JSON results.

-----

[...]

PR #1679, opened 2022-08-19
[mod] do not consent to tracking when using google

return42:
## What does this PR do?

do not consent to tracking when using google

Cherry picked from https://github.com/searx/searx/commit/5b50d7

Test various google engines an the youtube (noapi) engine / the patch is applied on my instance:

- [!go london](https://darmarit.org/searx/search?q=%21go+london)
- [!goi london](https://darmarit.org/searx/search?q=%21goi+london)
- [!gov london](https://darmarit.org/searx/search?q=%21gov+london)
- [!gos  heisenberg](https://darmarit.org/searx/search?q=%21gos+heisenberg)
- [!yt london](https://darmarit.org/searx/search?q=%21yt+london)

You might get a "too many request" on my instance / this issue is not related to this PR and discussed here https://github.com/searxng/searxng/issues/1642

## Related issues

- https://github.com/searxng/searxng/issues/1555

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -26,6 +26,7 @@ The google WEB engine itself has a special setup option:
 """
 
 from urllib.parse import urlencode
+from random import random
 from lxml import html
 from searx.utils import match_language, extract_text, eval_xpath, eval_xpath_list, eval_xpath_getindex
 from searx.exceptions import SearxEngineCaptchaException
@@ -282,7 +283,7 @@ def request(query, params):
         query_url += '&' + urlencode({'safe': filter_mapping[params['safesearch']]})
     params['url'] = query_url
 
-    params['cookies']['CONSENT'] = "YES+"
+    params['cookies']['CONSENT'] = "PENDING+" + str(random() * 100)
     params['headers'].update(lang_info['headers'])
     if use_mobile_ui:
         params['headers']['Accept'] = '*/*'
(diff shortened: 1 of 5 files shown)


unixfox:
I tried to experiment with the PENDING cookie, but it doesn't always work unfortunately. In some country you still get the consent page. That's why I preferred to set it to `YES+` until we find a solution over setting `PENDING+` and having a broken experience in some countries.

Anyway with https://github.com/searxng/searxng/issues/1642 you don't need to pass any CONSENT cookie anymore.

dalf:
While the issue remains, I suggest to close this PR, the `CONSENT` cookie has no effect (or nearly) when the value is `PENDING`.

return42:
In my analyses at the time, I had the impression that the behavior had improved.

> In some country you still get the consent page

I think @unixfox has much more experience than I do and I would rely on his assessment.

But one question remains on my side: if it doesn't solve the problem in all countries, isn't it at least an improvement in some countries?

@unixfox, @dalf: I'm sorry I do not have time ATM to look closer and I'm not sure anymore, please close the PR if you think it's not an improvement.

unixfox:
> In my analyses at the time, I had the impression that the behavior had improved.
> 
> > In some country you still get the consent page
> 
> I think @unixfox has much more experience than I do and I would rely on his assessment.
> 
> But one question remains on my side: if it doesn't solve the problem in all countries, isn't it at least an improvement in some countries?
> 

I mean the reliable method that works for all the countries is to send CONSENT YES. Sending CONSENT PENDING may not work for all the countries.

PR #1688, opened 2022-08-22
Add 9gag engine

liimee:
## What does this PR do?

Adds the 9gag engine

it'd be nice if searxng could have more social media engines

search `!9g test`

### DIFF
--- a/searx/engines/9gag.py
+++ b/searx/engines/9gag.py
@@ -34,9 +34,9 @@ def request(query, params):
 def response(resp):
     results = []
 
-    json_results = loads(resp.text)['data']['posts']
+    json_results = loads(resp.text)['data']
 
-    for result in json_results:
+    for result in json_results['posts']:
         result_type = result['type']
 
         if result_type == 'Photo':
@@ -62,5 +62,9 @@ def response(resp):
                     'thumbnail': result['images']['imageFbThumbnail']['url'],
                 }
             )
+
+    if 'tags' in json_results:
+        for suggestion in json_results['tags']:
+            results.append({'suggestion': suggestion['key']})
 
     return results
--- a/searx/engines/9gag.py
+++ b/searx/engines/9gag.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=C0103
+"""9GAG (social media)"""
+
+from json import loads
+from datetime import datetime
+from urllib.parse import urlencode
+
+about = {
+    "website": 'https://9gag.com/',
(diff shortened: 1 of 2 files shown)


searx/engines/9gag.py
```diff
@@ -0,0 +1,66 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+# pylint: disable=C0103
```
return42: FYI: We prefer to use the named option: `invalid-name` instead of `C0103` .. so it is easier to see what has been switched off.

liimee: done

searx/engines/9gag.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+        elif result_type == 'Animated':
+            results.append(
+                {
+                    'template': 'videos.html',
+                    'url': result['url'],
+                    'title': result['title'],
+                    'content': result['description'],
+                    'publishedDate': datetime.utcfromtimestamp(result['creationTs']),
+                    'thumbnail': result['images']['imageFbThumbnail']['url'],
```
dalf: With
`'iframe_src' : result['images'].get('image460sv', {}).get('url'),`

it is possible to view the videos:
image

However, since it is managed by the browser without additional javascript, the user have to stop the first video and then play the second one (otherwise both are played at the same time).

searx/engines/9gag.py
```diff
@@ -0,0 +1,70 @@
...
+                }
```
dalf, line 70: To get the not cropped version of the image when the image height is not too important:

```python
            if result['images']['image700']['height'] > 400:
                thumbnail = result['images']['imageFbThumbnail']['url']
            else:
                thumbnail = result['images']['image700']['url']

...
            'thumbnail': thumbnail
```

This is useful when the thumbnail contains some text.
Search for `!9gag nasa` for example

dalf (commented):
LGTM, some possible improvements in the comments.

[EDIT] these suggestions are optional, there is no problem to commit the current version.

searx/engines/9gag.py
liimee: @dalf just for videos or both?

searx/engines/9gag.py
liimee: done

searx/engines/9gag.py
dalf: I've tried only for videos, but it seems it could work for images too.

liimee: done

dalf (approved):
Thank you!

PR #1699, opened 2022-08-24
add apple app store engine

liimee:
## What does this PR do?

Adds the apple app store engine

I just feel that this engine is missing

search `!aps twitter`

## Related issues

___

sorry for submitting two PRs at the same time :sweat_smile:

### DIFF
--- a/searx/engines/apple_app_store.py
+++ b/searx/engines/apple_app_store.py
@@ -41,10 +41,11 @@ def response(resp):
             {
                 'url': result['trackViewUrl'],
                 'title': result['trackName'],
-                'content': result['sellerName'] + ': ' + result['description'],
+                'content': result['description'],
                 'img_src': result['artworkUrl100'],
                 'thumbnail': result['screenshotUrls'][0],
                 'publishedDate': parse(result['currentVersionReleaseDate']),
+                'author': result['sellerName'],
             }
         )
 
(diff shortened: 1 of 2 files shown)


liimee:
They seem to provide price data for paid apps -- but I don't think it would work without setting the template to `products.html`

searx/engines/apple_app_store.py
```diff
@@ -0,0 +1,58 @@
...
+    json_result = loads(resp.text)
+
+    for result in json_result['results']:
+        results.append(
+            {
+                'url': result['trackViewUrl'],
+                'title': result['trackName'],
+                'content': result['description'],
+                'img_src': result['artworkUrl100'],
+                'thumbnail': result['screenshotUrls'][0],
```
dalf: The other engines in the `app` category doesn't provide a screenshot : 
image

This make the "apple_app_store" engine huge compare to others.

So, can you remove the thumbnail?

dalf (changes requested):
Except the thumbnail, LGTM to me, thank you!

searx/engines/apple_app_store.py
liimee: I thought it'd be nice if we could get a quick peek of the app before actually opening the URL, but ok

liimee: done

dalf: > I thought it'd be nice if we could get a quick peek of the app before actually opening the URL

I agree; the thing is, I don't know how to display these screenshots.
I thought about something similar to the image: when the users click on a result, the screenshot(s) are displayed on the right side, so the users can still browse the result quickly without too much information in the result list.

Of course it requires more development (the embedded videos could be displayed on the right side too).

PR #1705, opened 2022-08-26
Theme: add a paper.html template and update of the science engines

dalf:
## What does this PR do?

See the commits:
* add `paper.html` in addition to `images.html`, `map.html` ...
   * fields (most comes from bibtex format : https://www.bibtex.com/g/bibtex-format/ ):
* add a `scientific publications` non tab category.
* update science engines: arxiv, crossref, google_scholar, pubmed, semantic scholar, springer
* oa_doi_rewrite : the found DOI is added to the result, so `paper.html` can show it (useful for google scholar for example because this engine does not the DOI).

Field list (see https://www.bibtex.com/g/bibtex-format/ and https://en.wikipedia.org/wiki/BibTeX#Field_types ) :
| field | Python type | description |
| -- | -- | -- |
| title | str | |
| content | str | abstract |
| comments | str | free text display in italic below the content |
| tags | List[str] | free tag list |
| publishedDate | datetime.datetime | last publication date |
| authors | List[str] | list of authors of the work (authors with a `s`) |
| editor | str | list of editors of a book |
| publisher | str | name of the publisher |
| journal | str | name of the journal or magazine the article was published in |
| volume | str | volume number |
| pages | str | |
| number | str | number of the report or the issue number for a journal article |
| doi | str | DOI number (like 10.1038/d41586-018-07848-2) |
| issn | str | ISSN number (like 1476-4687) |
| isbn | str | ISBN number (like 9780201896831) |
| pdf_url | str | URL to the full article, the PDF version |
| html_url | str | URL to full article, HTML version |

Status of the engines in the master branch:
| Engine | status |
| ------ | --------------------------- |
arxiv		| ok 	  |
crossref 	| timeout |
google_scholar | ok   |
pubmed      | only 1 result (indent problem in the code) | 
semantic scholar | HTTP status 400 |

Provide a more usable science category

image

* check each engines
* the `paper.html` layout in various configuration ( desktop, browser, centered or not, etc...) 

## Author's checklist

* google scholar provides an useful link "cited by #### papers" : it is difficult to proxy this link currently. An example: `https://scholar.google.com/scholar?cites=11448658805290839292&as_sdt=2005&sciodt=0,5&hl=en`. The results have the  same layout.
* In addition to Altemtric, crossmark can be useful:
  * https://www.crossref.org/services/crossmark/
  * example: https://crossmark.crossref.org/dialog?doi=10.1073%2Fpnas.2213287119&domain=www.pnas.org&uri_scheme=https%3A&cm_version=v2.0
* And also http://retractiondatabase.org/RetractionSearch.aspx to check if a paper is retracted.


searx/webutils.py
```diff
@@ -138,6 +141,22 @@ def highlight_content(content, query):
     return content
 
 
+def searxng_format_date(dt: datetime):  # pylint: disable=invalid-name
```
return42: In the commit message there is a typo (*day, day*):

    When the month, day, hour, day and second are zero, 
    the function returns only the year.

should be something like:

    When the month, day, hour, minute and second are zero,
    the returned string only indicates how many years ago the date was.

BTW this information should be in a doc-string of the `searxng_format_date()` definition.  Additional in this doc-string it should be described that this functions returns string with a human readable time distance from *now* to a date/time object in the past.

```python

def searxng_l10n_timespan(dt1: datetime):
    """"Returns a human-readable and translated string indicating how long ago
    a date was in the past / the time span of the date to the present.

    When the month, day, hour, minute and second of the :py:obj:datetime: object
    are zero, the returned string only indicates how many years ago the date was.
    """
    ...    
```

searx/engines/springer.py
```diff
@@ -41,32 +41,30 @@ def response(resp):
...
                 'url': record['url'][0]['value'].replace('http://', 'https://', 1),
+                'type': record.get('contentType'),
                 'content': content,
                 'publishedDate': published,
-                'metadata': metadata,
+                'authors': authors,
+                'doi': record.get('doi'),
+                'journal': record.get('publicationName'),
+                'start_page': record.get('start_page'),
+                'end_page': record.get('end_page'),
```
return42: Should we consider to use the notation (syntax) of the [BibTeX field: pages](https://www.bibtex.com/f/pages-field/) .. The notation is much more flexible than having a start and an end page number:

The pages field is used to store page numbers or a page range. For a journal article it's the page range where the article can be found in the specific issue of that journal, or for a section in a book it could be a series of page numbers.

```
pages = "723--729"
pages = "723--29"
pages = "III-IX"
pages = 23
pages = "34,37--39"
```

In Germany (IDK if other languages know this) we have [*f.* and *ff.*](https://de.wikipedia.org/wiki/F.) .. the `f` stands for *following page* (DE singular: "folgende") and `ff` *following pages* (DE plural: "folgenden").

searx/engines/semantic_scholar.py
```diff
@@ -60,22 +60,47 @@ def response(resp):
...
-        if venue:
-            metadata.append(venue)
-        if metadata:
-            item['metadata'] = ', '.join(metadata)
+        # pick for the first alternate link, but not from the crawler
+        pdf_url = None
+        for doc in result.get('alternatePaperLinks', []):
+            if doc['linkType'] != 'crawler':
+                pdf_url = doc['url']
+                break
```
return42: When I search for `!se :en foo` some PDF links point to a PDF and some of the PDF link points to a page that contains a PDF but not to the PDF itself.  By example my first hit in the result list was 

image


- Button [PDF] links to https://ieeexplore.ieee.org/document/9374435 
- PDF itself is: https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9374435

searx/settings.yml
```diff
@@ -398,23 +397,9 @@ engines:
...
+    engine: crossref
     shortcut: cr
-    about:
-      website: https://www.crossref.org/
-      wikidata_id: Q5188229
-      official_api_documentation: https://github.com/CrossRef/rest-api-doc
-      use_official_api: false
-      require_api_key: false
-      results: JSON
+    timeout: 10
```
return42: I think we should disable this engine by default / in my tests the response time was 30sec and more.  It does not make a difference if I use the SearXNG engine `!crossref :en physics` or directly https://search.crossref.org/?q=physics&from_ui=yes

searx/engines/arxiv.py
```diff
@@ -27,6 +28,23 @@
...
+xpath_title = XPath('.//atom:title', namespaces=arxiv_namespaces)
+xpath_id = XPath('.//atom:id', namespaces=arxiv_namespaces)
+xpath_summary = XPath('.//atom:summary', namespaces=arxiv_namespaces)
+xpath_author_name = XPath('.//atom:author/atom:name', namespaces=arxiv_namespaces)
+xpath_doi = XPath('.//arxiv:doi', namespaces=arxiv_namespaces)
+xpath_pdf = XPath('.//atom:link[@title="pdf"]', namespaces=arxiv_namespaces)
+xpath_published = XPath('.//atom:published', namespaces=arxiv_namespaces)
+xpath_journal = XPath('.//arxiv:journal_ref', namespaces=arxiv_namespaces)
+xpath_category = XPath('.//atom:category/@term', namespaces=arxiv_namespaces)
+xpath_comment = XPath('./arxiv:comment', namespaces=arxiv_namespaces)
```
return42, line 46: I like to have the XPath selectors in the code, it makes code reading and fixing expression a little easier / but this is my very own opinion.

return42 (approved):
Cool work, thanks a lot.  I left some comments .. and I added a commit on top with the documentation (in `/docs`) given in the initial comment of this PR .. I think code readers will not read PR comments, it is better to have all the needed descriptions in the code, the commit message or in the online documentation.

searx/engines/semantic_scholar.py
dalf: I have changed to check:
`if doc['linkType'] not in ('crawler', 'doi'):`
but the result format does provide this information.

searx/webutils.py
dalf: done

searx/engines/springer.py
dalf: done

searx/settings.yml
dalf: done

searx/engines/arxiv.py
dalf: Yes me too, except the namespaces must be specified.

Currently (I pick two lines among others):
```python
for entry in eval_xpath_list(dom, xpath_entry):
    title = eval_xpath_getindex(entry, xpath_title, 0).
```

Usually we have:
```python
for entry in eval_xpath_list(dom, '//atom:entry'):
    title = eval_xpath_getindex(entry, './/atom:title', 0).
```

Except it won't work because we have to specify the namespaces.
However the `eval_xpath_*` functions do not accept the `namespaces` parameter. But we can pass a XPath instance instead of a string, so the code would be:
```python
for entry in eval_xpath_list(dom, XPath('//atom:entry', namespaces=arxiv_namespaces)):
    title = eval_xpath_getindex(entry, XPath('.//atom:title', namespaces=arxiv_namespaces), 0).
```

IMO it is less readable compare to the current version.

searx/webutils.py
dalf: Edit: I have update the comment to
```
    """Returns a human-readable and translated string indicating how long ago
    a date was in the past / the time span of the date to the present.

    On January 1st, midnight, the returned string only indicates how many years
    ago the date was.
    """
```

docs/dev/engine_overview.rst
```diff
@@ -311,3 +311,88 @@ the parameter ``template`` must be set to the desired type.
...
+     - list of authors of the work (authors with a "s")
+
+   * - editor
+     - :py:class:`str`
+     - list of editors of a book
+
+   * - publisher
+     - :py:class:`str`
+     - name of the publisher
+
```
return42, line 362: I am just in translation of the strings .. and I miss a description of the [`result.type` field](https://github.com/searxng/searxng/blob/master/searx/templates/simple/result_templates/paper.html#L24) .. is the field intended to characterize the medium (e.g. pdf, book, paper)?

dalf: It sounds like what you have described.

It is mix between the file format and "entry type" (in the BibTex vocabulary):
* springer: I can't remember what the [`contentType` field](https://github.com/searxng/searxng/blob/bafebe91ac9812c4ac243b0717f411b1ed0a66a9/searx/engines/springer.py#L55) is. It sounds like the file format.
* crossref: `book-chapter`, `book`, `other`, `component`
* google schoolar: `book` but also `pdf`, `html`. This is the word before the title. In the following example, it is "book":
image

return42:
Copied from the initial comment ..

| field | Python type | description |
| -- | -- | -- |
| issn | str | ISSN number (like 1476-4687) |
| isbn | str | ISBN number (like 9780201896831) |

Sure that these are `str` types and not `List[str]`?

Issue #1695, opened 2022-08-23
Some symbols are not kept on externals bangs & autocomplete


LoicBersier:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
4dc70b08 

**How did you install SearXNG?**
searxng-docker

**What happened?**
When searching something that contain a special symbol (Only tried with & # +) and use an external bang or on autocomplete the symbol will be lost

**How To Reproduce**
Search "1+1 !!g", it will show "1 1" in google
Search for "1+1 !" and the autocomplete will show up missing the plus symbol

**Expected behavior**
Show "1+1" in google & in the autocomplete

**Screenshots & Logs**
![image](https://user-images.githubusercontent.com/36051669/186251182-e54e4ea3-2e2c-4737-91d1-649ecc1741f3.png)

**Additional context**
I think the issue comes from the query not being url encoded.

dalf:
I confirm the bugs:
* autocompletion of `1+1 !` is `1 1 !images`. The bug exists only with the POST method, not the GET method.
* `1+1 !!g` actually search `1 1` on Google. 

---
Technical notes

First point:
   https://github.com/searxng/searxng/blob/5284de9137c43c2a343958c5c3fccf97b72d65a3/searx/webapp.py#L899
   `request.form.get('q', '')` does not contains the `+`

Second point:
   https://github.com/searxng/searxng/blob/5284de9137c43c2a343958c5c3fccf97b72d65a3/searx/external_bang.py#L42

```patch
diff --git a/searx/external_bang.py b/searx/external_bang.py
index a56737c7..d24ba34f 100644
--- a/searx/external_bang.py
+++ b/searx/external_bang.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+from urllib.parse import quote_plus
 from searx.data import EXTERNAL_BANGS
 
 LEAF_KEY = chr(16)
@@ -39,7 +40,11 @@ def get_bang_definition_and_ac(external_bangs_db, bang):
 
 def resolve_bang_definition(bang_definition, query):
     url, rank = bang_definition.split(chr(1))
-    url = url.replace(chr(2), query)
+    chr2_pos = url.find(chr(2))
+    if url.find('?') < chr2_pos or url.find('#') < chr2_pos:
+        url = url.replace(chr(2), quote_plus(query))
+    else:
+        url = url.replace(chr(2), query)
     if url.startswith('//'):
[...]

PR #1706, opened 2022-08-27
Fix: autocomplete with the POST method: url encode the user query

dalf:
## What does this PR do?

See the commit and the comment in the code: currently, the endpoint `/autocompleter` receives `1 1` when the users type `1+1`. This is a bug in autocomplete.js.

Since the project seems [abandoned](https://github.com/autocompletejs/autocomplete.js/issues/78), the PR monkey patch a function of the project. 

With the POST, and then with the GET method: check the autocompletion of `1 + 1 !`.
You should see in both cases `1 + 1 !images` and not `1 1 !images` as the first suggestion.

## Author's checklist

https://www.npmjs.com/package/@tarekraafat/autocomplete.js might be a good replacement for autocomplete.js

## Related issues

Partially related to https://github.com/searxng/searxng/issues/1695


return42 (approved):
> Monkey patch autocomplete.js to fix a bug

Is there anything left over from autocomplete.js we did not have fixed :-)

.. thanks!!

dalf:
True...
I've tried to include to embed it so it is easier to fix it ; but it is written in typescript and I don't want to add an additional language.

PR #1707, opened 2022-08-27
External bang: bug fix: URL encode the query so "!!g 1+1" works as intended

dalf:
## What does this PR do?

Currently `!!g 1+1` is read as `1 1` by Google because the query is not URL encoded.

Bug fix

Some URL includes the user query:
* as a parameter: encoding is mandatory ( `!!g 1+1` )
* a part of the URL path.: some website can deal with it ( `!!1122 1+1`)
* a fragment ( `#...`) : I can't find a relevant example search for `#` in `searx/data/external_bangs.json`.

## Related issues

Partially related to #1695

### DIFF
--- a/searx/external_bang.py
+++ b/searx/external_bang.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+from urllib.parse import quote_plus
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks!

PR #1522, opened 2022-07-17
Convert result_proxy.key into bytearray if necessary

SuperSandro2000:
## What does this PR do?
Convert result_proxy.key into bytearray if necessary

## Why is this change important?
When generating the settings.yml from NixOS code ``!!binary`` cannot easily be set.

## How to test this PR locally?

Remove ``!!binary`` from settings.yml.

PR #1708, opened 2022-08-27
settings.yml: set default values for result_proxy

dalf:
## What does this PR do?

* initialize `settings["result_proxy"]` with searx/settings_defaults.py
* allow `result_proxy.key` to be a string (convert into bytes by searx/settings_defaults.py)

Related to #1522 : the string is base64 decoded, since Morty read the key as a base64 encoded string. 

In #1522 , the string is utf-8 encoded. 

## Related issues

#1522

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -121,6 +121,7 @@ ui:
 #
 # result_proxy:
 #   url: http://127.0.0.1:3000/
+#   # the key is a base64 encoded string, the YAML !!binary prefix is optional
 #   key: !!binary "your_morty_proxy_key"
 #   # [true|false] enable the "proxy" button next to each result
 #   proxify_results: true
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -9,6 +9,7 @@ import numbers
 import errno
 import os
 import logging
+from base64 import b64decode
 from os.path import dirname, abspath
 
 from searx.languages import language_codes as languages
@@ -117,6 +118,15 @@ class SettingsDirectoryValue(SettingsValue):
         return super().__call__(value)
 
 
+class SettingsBytesValue(SettingsValue):
+    """str are base64 decoded"""
+
+    def __call__(self, value: typing.Any) -> typing.Any:
+        if isinstance(value, str):
+            value = b64decode(value)
+        return super().__call__(value)
+
+
 def apply_schema(settings, schema, path_list):
     error = False
     for key, value in schema.items():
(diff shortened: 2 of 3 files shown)


searx/settings.yml
```diff
@@ -121,6 +121,7 @@ ui:
 #
 # result_proxy:
 #   url: http://127.0.0.1:3000/
+#   # use !!binary to write the key in base64 otherwise use an usual string
```
return42: Hint: did you see comment started in line 117 .. in cfcbc3a5c3 you added a note about `!!binary` notation where you point to commit af77ec3 .. but this commit does not exists.

I would prefer ..

    # the key is a base64 encoded string, the YAML !!binary prefix is optional

and line 119 needs to be fixed: *(keep the !!binary notation)* .. is no longer valid.

Do we need a documentation of this setting? / morty is no longer needed as image proxy and we keep the option to configure `proxy` link silently.

return42 (changes requested):
LGTM / Except the nits in my comment.

searx/settings.yml
dalf: It has two different meaning.
Both are the string `foobar`:
* `!!binary "Zm9vYmFy"` <-- string base64 encoded, https://yaml.org/type/binary.html
* `"foobar"` <-- not base64 encoded

So `!!binary` is not optional, up to the user to choose between raw string or base64 encoded string.

return42: > So `!!binary` is not optional, up to the user to choose between raw string or base64 encoded string.

Oops sorry, but if so, how can it help to solve the issue #1522 ...

> When generating the settings.yml from NixOS code ``!!binary`` cannot easily be set.

If I understand @SuperSandro2000 correctly there has been a base64 string generated `"Zm9vYmFy"` this base64 string is used when the morty service is started (`morty -key Zm9vYmFy ...`)

The morty key is always a base64 string / see https://github.com/asciimoo/morty#usage

And he wants to use this string notation in settings.yml, whithout the need to add a `!!binary` prefix

```yaml
result_proxy:
  url: http://127.0.0.1:3000/
  key: Zm9vYmFy
  proxify_results: true
```

.. or I am wrong?

dalf: > The morty key is always a base64 string

yes, now I remember that is has created issues few year ago. For example:
* user were writing the key as plain text in morty, which was decoded as base64.
* then copy / paste the key in settings.yml  and it didn't worked.

As far I understand Nix can't generated the `!!binary` prefix, so this PR should solve that : just skip the prefix, it works the same.

I have updated the PR in this way.

return42: > I have updated the PR in this way.

OK, LGTM .. there is a question leftover from my initial comment 

[Do we need a documentation of this setting? / morty is no longer needed as image proxy and we keep the option to configure `proxy` link silently.](https://github.com/searxng/searxng/pull/1708#discussion_r956693217)

What do you think?

dalf: > Do we need a documentation of this setting?

No need for documentation, settings.yml is good enough here: 
* it doesn't break existing installation
* this PR makes the life easier with Nix, nothing more

return42: For me a feature is official when it is documented .. if there is no documentation, only a comment in a config file it is an hidden feature (without version stability).

Since `result_proxy` is based on a no longer maintained software (morty) its fine for me to keep it unofficial.

return42 (approved):

PR #1709, opened 2022-08-27
Fix: The environment variable SEARX_SETTINGS_PATH was ignored

dalf:
## What does this PR do?

Refactor searx.settings_loader.py.get_user_settings_path
(SearxSettingsException is never raised)

SearXNG searches the user settings is these directories (by descending priority):
* `environ['SEARXNG_SETTINGS_PATH']`
* `'/etc/searxng/settings.yml'`
* `environ['SEARX_SETTINGS_PATH']`
* `'/etc/searx/settings.yml'`

## Why is this change important?

* Don't ignore SEARX_SETTINGS_PATH
* Simplify the code of searx.settings_loader (`SearxSettingsException` is never raised).

## How to test this PR locally?

Define two custom settings.yml in the various locations:
* SEARXNG_SETTINGS_PATH and SEARX_SETTINGS_PATH : The SearXNG must be chosen.
* SEARXNG_SETTINGS_PATH and /etc/searx/settings.yml : The SearXNG must be chosen.
* SEARX_SETTINGS_PATH and /etc/searx/settings.yml : SEARX_SETTINGS_PATH must be chosen.

## Related issues

Close https://github.com/searxng/searxng/issues/1519

PR #373, opened 2021-10-02
SearXNG: environment variables

dalf:
## What does this PR do?

SearXNG brand

https://github.com/searxng/searxng/blob/b268d672a6b4ccf5f2e4f4752c68b066b1563f47/searx/settings_defaults.py#L45-L52

The code uses the `SEARX_*` variables, if the `SEARXNG_*` variables are not defined.

This PR is only about these variables (not `SEARX_SRC` etc...)

PR #1712, opened 2022-08-28
Remove usage of SEARX environment variables

### LINKED ISSUE #1519: SEARX_SETTINGS_PATH fallback is ignored for uwsgi

7bf4e8d12d1d0ee53bf71f7c3a4010ef936f25d9

NixOS module
**What happened?**

``SEARX_SETTINGS_PATH`` is being ignored when running via uwsgi vessel and the default config is being used. Using SEARXNG_SETTINGS_PATH instead works.
**How To Reproduce**

Run searx via uwsgi and supply the config only via SEARXNG_SETTINGS_PATH.
**Expected behavior**

**Screenshots & Logs**

```
Jul 17 21:18:12 hydrogen uwsgi[4009531]: DAMN ! worker 1 (pid: 42032) died :( trying respawn ...
Jul 17 21:18:12 hydrogen uwsgi[4009531]: worker respawning too fast !!! i have to sleep a bit (2 seconds)...
Jul 17 21:18:14 hydrogen uwsgi[4009531]: Respawned uWSGI worker 1 (new pid: 43476)
Jul 17 21:18:14 hydrogen uwsgi[43476]: SEARXNG_SETTINGS_PATH uses value from SEARX_SETTINGS_PATH
Jul 17 21:18:14 hydrogen uwsgi[43476]: 2022-07-17 21:18:14,905 ERROR:searx: git is not found, fallback to the default version
Jul 17 21:18:14 hydrogen uwsgi[43476]: 2022-07-17 21:18:14,991 ERROR:searx.shared: uwsgi.ini configuration error, add this line to your uwsgi.ini
Jul 17 21:18:14 hydrogen uwsgi[43476]: cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1
Jul 17 21:18:14 hydrogen uwsgi[43476]: 2022-07-17 21:18:14,994 ERROR:searx.webapp: server.secret_key is not changed. Please use something else instead of ultrasecretkey.
Jul 17 21:18:15 hydrogen uwsgi[4009531]: DAMN ! worker 1 (pid: 43476) died :( trying respawn ...
Jul 17 21:18:15 hydrogen uwsgi[4009531]: Respawned uWSGI worker 1 (new pid: 43615)
Jul 17 21:18:15 hydrogen uwsgi[43615]: SEARXNG_SETTINGS_PATH uses value from SEARX_SETTINGS_PATH
Jul 17 21:18:15 hydrogen uwsgi[43615]: 2022-07-17 21:18:15,928 ERROR:searx: git is not found, fallback to the default version
Jul 17 21:18:15 hydrogen systemd[1]: Stopping uwsgi.service...
``` 

**Additional context**

dalf:
## What does this PR do?

Remove usage of SEARX environment variable.

See comment on https://github.com/searxng/searxng/pull/1709

## Author's checklist

A public announcement would be good: some users may rely on existing packages that relies on this SEARX variables.

## Related issues

* Close #1519 as won't fix
* Related to #373

### DIFF
--- a/searx/settings_loader.py
+++ b/searx/settings_loader.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+from typing import Optional
 from os import environ
 from os.path import dirname, join, abspath, isfile
 from collections.abc import Mapping
@@ -13,7 +14,7 @@ from searx.exceptions import SearxSettingsException
 searx_dir = abspath(dirname(__file__))
 
 
-def check_settings_yml(file_name):
+def existing_filename_or_none(file_name: str) -> Optional[str]:
     if isfile(file_name):
         return file_name
     return None
@@ -30,29 +31,29 @@ def load_yaml(file_name):
 
 
 def get_default_settings_path():
-    return check_settings_yml(join(searx_dir, 'settings.yml'))
+    return existing_filename_or_none(join(searx_dir, 'settings.yml'))
 
 
-def get_user_settings_path():
-    # find location of settings.yml
+def get_user_settings_path() -> Optional[str]:
+    """Get an user settings file.
+    By descending priority:
+    1. ``environ['SEARXNG_SETTINGS_PATH']``
+    2. ``/etc/searxng/settings.yml`` except if ``SEARXNG_DISABLE_ETC_SETTINGS`` is ``true`` or ``1``
+    3. ``None``
+    """
+
+    # check the environment variable SEARXNG_SETTINGS_PATH
+    # if the environment variable is defined, this is the last check
     if 'SEARXNG_SETTINGS_PATH' in environ:
-        # if possible set path to settings using the
-        # enviroment variable SEARXNG_SETTINGS_PATH
-        return check_settings_yml(environ['SEARXNG_SETTINGS_PATH'])
+        return existing_filename_or_none(environ['SEARXNG_SETTINGS_PATH'])
 
+    # if SEARXNG_DISABLE_ETC_SETTINGS don't look any futher
     if environ.get('SEARXNG_DISABLE_ETC_SETTINGS', '').lower() in ('1', 'true'):
         return None
 
-    # if not, get it from searx code base or last solution from /etc/searxng
-    try:
-        return check_settings_yml('/etc/searxng/settings.yml')
-    except SearxSettingsException as e:
-        # fall back to searx settings
-        try:
-            return check_settings_yml('/etc/searx/settings.yml')
-        except SearxSettingsException:
-            # if none are found, raise the exception about SearXNG
-            raise e  # pylint: disable=raise-missing-from
+    # check /etc/searxng/settings.yml
+    # (continue with other locations if the file is not found)
+    return existing_filename_or_none('/etc/searxng/settings.yml')
 
 
 def update_dict(default_dict, user_dict):
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -41,16 +41,6 @@ STR_TO_BOOL = {
 }
 _UNDEFINED = object()
 
-# compatibility
-SEARX_ENVIRON_VARIABLES = {
-    'SEARX_DISABLE_ETC_SETTINGS': 'SEARXNG_DISABLE_ETC_SETTINGS',
-    'SEARX_SETTINGS_PATH': 'SEARXNG_SETTINGS_PATH',
-    'SEARX_DEBUG': 'SEARXNG_DEBUG',
-    'SEARX_PORT': 'SEARXNG_PORT',
-    'SEARX_BIND_ADDRESS': 'SEARXNG_BIND_ADDRESS',
-    'SEARX_SECRET': 'SEARXNG_SECRET',
-}
-
 
 class SettingsValue:
     """Check and update a setting value"""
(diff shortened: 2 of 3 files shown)


searx/settings_loader.py
```diff
@@ -30,29 +31,32 @@ def load_yaml(file_name):
 
 
 def get_default_settings_path():
-    return check_settings_yml(join(searx_dir, 'settings.yml'))
+    return existing_filename_or_none(join(searx_dir, 'settings.yml'))
```
return42, line 34: Do we really need a function ``existing_filename_or_none`` for such a trivial task?

BTW: I would change `searx_dir` in `/share/SearXNG/searx/settings_loader.py` to an `pathlib.Path` type.

searx/settings_loader.py
```diff
@@ -30,29 +31,32 @@ def load_yaml(file_name):
...
+def get_user_settings_path() -> Optional[str]:
+    """Get an user settings file.
+
+    By descending priority:
+    * environ['SEARXNG_SETTINGS_PATH']
+    * '/etc/searxng/settings.yml'
+    * None
+
+    Note: check only SEARXNG_SETTINGS_PATH if environ['SEARXNG_SETTINGS_PATH'] is 1 or true
+    """
```
return42: I would prefer a numbered list (and reST) 

```
Get user ``settings.yml`` file.

By descending priority:

1. ``environ['SEARXNG_SETTINGS_PATH']``
2. ``/etc/searxng/settings.yml``
3. ``None``

.. note::

   Checks only ``SEARXNG_SETTINGS_PATH`` if ``environ['SEARXNG_SETTINGS_PATH']``
   is 1 or ``True``
```

TBH: I don't understand the meaning of the *note* .. for me it is clear, that there needs to be a valid filename of an existing file in ``SEARXNG_SETTINGS_PATH`` .. or do I misunderstood  the meaning the *note*?

searx/settings_loader.py
```diff
@@ -30,29 +31,32 @@ def load_yaml(file_name):
...
+    # check the environment variable SEARXNG_SETTINGS_PATH
+    # if the environment variable is defined, this is the last check
     if 'SEARXNG_SETTINGS_PATH' in environ:
-        # if possible set path to settings using the
-        # enviroment variable SEARXNG_SETTINGS_PATH
-        return check_settings_yml(environ['SEARXNG_SETTINGS_PATH'])
+        return existing_filename_or_none(environ['SEARXNG_SETTINGS_PATH'])
 
+    # if SEARXNG_DISABLE_ETC_SETTINGS don't look any futher
     if environ.get('SEARXNG_DISABLE_ETC_SETTINGS', '').lower() in ('1', 'true'):
```
return42, line 51: Should we document `SEARXNG_DISABLE_ETC_SETTINGS` [here](https://docs.searxng.org/admin/engines/settings.html?#settings-yml-location) .. and that `SEARXNG_SETTINGS_PATH` takes precedence?

IDK what is better: should `SEARXNG_DISABLE_ETC_SETTINGS` take precedence over `SEARXNG_SETTINGS_PATH` / or with other words: should `SEARXNG_DISABLE_ETC_SETTINGS` disable all kind of custom settings?

return42 (changes requested):
LGTM  / beside the comments I added.

searx/settings_loader.py
dalf: Comment updated

searx/settings_loader.py
dalf: It makes the code of `get_user_settings_path` super easy to read.

with:
```python
    if 'SEARXNG_SETTINGS_PATH' in environ:
        return existing_filename_or_none(environ['SEARXNG_SETTINGS_PATH'])
```

without:
```python
    if 'SEARXNG_SETTINGS_PATH' in environ:
        return environ['SEARXNG_SETTINGS_PATH'] if isfile(environ['SEARXNG_SETTINGS_PATH']) else None
```

---

Regarding `pathlib.Path` : 
* there are a lot places where it can replace `os.path`. 
* it blurs what the PR does.

so, I prefer to have a PR focused on that like "Drop Python 3.6 support".

searx/settings_loader.py
dalf: This is an internal variable for the tests, so `/etc/searxng/settings.yml` does not break the tests:
* https://github.com/searxng/searxng/blob/8bdc6986a1488c6ef72b6ab3512d570f5868f4a7/tests/__init__.py#L7
* https://github.com/searxng/searxng/blob/8bdc6986a1488c6ef72b6ab3512d570f5868f4a7/tests/unit/test_settings_loader.py#L56-L122

==> 
* there is no need to document this.
* `SEARXNG_DISABLE_ETC_SETTINGS` must not skip `SEARXNG_SETTINGS_PATH` : the tests check `SEARXNG_SETTINGS_PATH` is working as expected.

searx/settings_loader.py
```diff
@@ -30,29 +31,29 @@ def load_yaml(file_name):
...
 
-def get_user_settings_path():
-    # find location of settings.yml
+def get_user_settings_path() -> Optional[str]:
+    """Get an user settings file.
+    By descending priority:
+    1. ``environ['SEARXNG_SETTINGS_PATH']``
+    2. ``/etc/searxng/settings.yml`` except if ``SEARXNG_DISABLE_ETC_SETTINGS`` is ``true`` or ``1``
+    3. ``None``
+    """
```
return42, line 43: Compare with https://github.com/searxng/searxng/pull/1712#discussion_r956701030

---

For the future:  reST is a plain text markup. Plain text markups are made to be readable in plain text and with a pinch of salt and pepper it can be parsed and rendered into another format ... 

Empty lines are used to introduce the next block (paragraph, list, ..). A single newline does not introduce a new block, it just exists to wrap the line in the plain text.

```
Get an user settings file.
By descending priority: ..
```

is rendered in one paragraph / it is the same like 

```
Get an user settings file. By descending priority: ..
```

The (numbered) list is a new block, to help the parser its recommended to add a empty line that introduce the new block (the list):

```
Get an user settings file. By descending priority:

1. foo
2. bar
```

return42 (approved):

PR #1714, opened 2022-08-28
[mod] option 'server: archive_url:' to configure internet archive

return42:
## What does this PR do?

The URL of the internet archive is hard coded in the template: make it configurable.

Allow to configure alternative web archives.

Check documentation and test internet archive by setting `https://archive.today/`

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -240,6 +240,7 @@ Global Settings
      query_in_title: false
      infinite_scroll: false
      center_alignment: false
+     cache_url: https://web.archive.org/web/
      default_theme: simple
      theme_args:
        simple_style: auto
@@ -266,6 +267,15 @@ Global Settings
   When enabled, the results are centered instead of being in the left (or RTL)
   side of the screen.  This setting only affects the *desktop layout*
   (:origin:`min-width: @tablet <searx/static/themes/simple/src/less/definitions.less>`)
+
+.. cache_url:
+
+``cache_url`` : ``https://web.archive.org/web/``
+  URL prefix of the internet archive or cache, don't forgett trailing slash (if
+  needed).  The default is https://web.archive.org/web/ alternatives are:
+
+  - https://webcache.googleusercontent.com/search?q=cache:
+  - https://archive.today/
 
 ``default_theme`` :
   Name of the theme you want to use by default on your SearXNG instance.
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -95,6 +95,8 @@ ui:
   default_theme: simple
   # center the results ?
   center_alignment: false
+  # URL prefix of the internet archive, don't forgett trailing slash (if needed).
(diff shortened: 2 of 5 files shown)


dalf (commented):
LGTM except one thing: the name of the setting `server.archive_url`.

First: on the screen `archive_url` is display as `Cached`, so `cached_url` would fit what is displayed on the screen (?)

Second: the `server` section puzzle me.

The existing sections of the yaml file:
* `search` describes the default values for the search form
   * exception: `ban_time_on_fail` and `max_ban_time_on_fail` which more or less belong between the `outgoing` and the `engine` sections (said differently: these two settings are not default values for the search form).
* `outgoing` describes the outgoing network configuration.
* `ui` describes how to display the results
* `server` describes the technical configuration between the user and the instance (which port, which HTTP headers, etc...)
  * exception: `image_proxy` and `method` are displayed in the "Privacy" tab of the preferences.

So `ui.cached_url` seems better: SearXNG just makes some links in the HTML results.

---

For reference: https://github.com/searxng/searxng/issues/681 (however, the issue does not deal with these sections)

return42:
I rebased to solve a conflict with master ..

> LGTM except one thing: the name of the setting `server.archive_url`.
> 
> First: on the screen `archive_url` is display as `Cached`, so `cached_url` would fit what is displayed on the screen (?)
> So `ui.cached_url` seems better: SearXNG just makes some links in the HTML results.

I placed it to `ui.archive_url:` ... "cache" is the wrong word and I added a new commit:

    [mod] rename 'cache' to 'web archive' 

    Users do not associate a web archive with the word "cache".  A "cache" is
    diametrically opposed to an "archive"; the former is short-term storage, while
    the latter is long-term persistence for historical purposes.

What do you think about "cache" vs "archive" in general, is it OK to rename it?

If you want, we can move the commit `[mod] rename 'cache' to 'web archive'` (and the `/static`) in a follow up PR.

> Second: the `server` section puzzle me.

Oh yes that is true ..

dalf:
> What do you think about "cache" vs "archive" in general, is it OK to rename it?

* Google, Bing and Gigablast use the word "Cached".
* Yandex uses "Saved copy".

In all cases, it refers to the copy of the pages by the search engine: https://en.wikipedia.org/wiki/Search_engine_cache

Web archive refers to what archive.is archive.today do : https://en.wikipedia.org/wiki/Web_archiving

So on the paper "web archive" seems better, but
* nearly everywhere the user see "cached" including SearXNG and searx.
* If the value for `archive_url` is `https://webcache.googleusercontent.com/search?q=cache:`, then it is not more a "web archive", it is the cached version from Google...

So I would stick to " Cached".

return42:
>  it is the cached version from Google...

Oops you remind me .. thanks to SearXNG its a long time ago I used google and I have forgotten that there is an cache from the indexed page.

> So I would stick to " Cached".

With said in mind .. done / thanks!

PR #1729, opened 2022-08-31
Disable brave by default

unixfox:
Brave is too unstable and will often not work by default. As seen in many issues: https://github.com/searxng/searxng/issues?q=is%3Aissue++sort%3Aupdated-desc+brave+label%3Abug+


dalf:
On searx.space, engines tab :
image

PR #1743, opened 2022-09-02
Update about the metrics

dalf:
## What does this PR do?

Two commits:
* the first one changes `/stats` to display on the score per result instead of the result.
* the second one hides to empty information when `general.enable_metrics: false`:
   * hide the link to `/stats` at the bottom of the pages
   * in /preferences, hide the columns "Response time" and "Reliability" 

* first commit: more privacy
* second commit: hide useless information when metrics are disabled.

With `general.enable_metrics: true`
* make some queries
* go the `/stats`, check the column score, compare to the master branch. The scores should be around 1.
* click on the "Score" column, check the scores are sorted in the right order

With `general.enable_metrics: false`:
* make some queries
* check there is no link to `/stats` at the bottom of the pages
* go the preferences: check the columns "Response time" and "Reliability" are missing

### DIFF
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -62,9 +62,9 @@
   <footer>
     <p>
     {{ _('Powered by') }} <a href="{{ url_for('info', pagename='about') }}">searxng</a> - {{ searx_version }} — {{ _('a privacy-respecting, hackable metasearch engine') }}<br/>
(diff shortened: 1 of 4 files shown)


return42 (approved):
LGTM / reminds me that I have never fully understood the interrelationships --> https://github.com/searxng/searxng/issues/1747

PR #1749, opened 2022-09-05
Switch to internal API for major google engines

### LINKED ISSUE #175: Pagination is broken for Google Images engine
**Version of Searx, commit number if you are using on master branch and stipulate if you forked Searx**
Commit number: `056550d523ddf4e06821957e13623fc322cf4675`

**How did you install Searx?**
Manually with `python -m searx.webapp`

**What happened?**
The Google Images engine doesn't have multiple pages, it only has one page. In contrary to Bing images which have multiple pages and the button "next page" is available.

**How To Reproduce**
1. `!goi test`
2. No "next page" button.

**Expected behavior**
Have multiple pages with "next page" button available.

**Screenshots & Logs**

<table>
<tr>
	<td>Google images
	<td>Bing images
<tr>
	<td>
	<td>
</table>

**Additional context**

### LINKED ISSUE #1642: Google search internal API with JSON results

unixfox:
## What does this PR do?

This PR uses the internal API of Google found in #1642

- More results, this PR properly handle the normal links, imbricated links (links related to a main link) and video links.
- Stable API (somewhat, I mean more stable than HTML...)

1. Run SearXNG
2. Type `!go yourquery`

## Author's checklist

- [x] Need to implement the suggestions fetching
- [x] Need to implement the "answer" fetching
- [x] Need to implement the "spelling check" fetching
- [ ] Need to implement the number of results fetching
- [ ] Need to introduce some functions in order to reduce code duplication, like when finding `result_index` element.
- [x] Need to implement this for Google news
- [ ] Need to implement this for Google videos
- [ ] Add comments?
- [ ] Skip parsing object if exception found but throw warning in the logs.
- [ ] Check with the different user-agent firefox and chrome
- [ ] Test for some keywords:
	- [ ] volodymyr zelenskyy

### Refactor
- [ ] ? Remove the fact that the parse_item methods are in a class. Also for `item_types`

## Related issues

Closes #1642
Closes #1749
Closes #175

## Additional info
How to quickly get a real JSON output from the Google internal API for the keyword "ok": https://www.google.com/search?q=ok&start=0&asearch=arc&async=use_ac:true,amw:false,_fmt:json

Initially I wanted to communicate with the google internal API through protobuf, but it was just a nightmare to write from scratch a [proto file](https://developers.google.com/protocol-buffers/docs/proto3) due to the big amount of different kind of fields so I gave up for the moment.

### DIFF
--- a/searx/engines/google_internal_search.py
+++ b/searx/engines/google_internal_search.py
@@ -1,0 +1,286 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""This is the implementation of the google WEB engine using the google internal API used on the mobile UI.
+This internal API offer results in
+- JSON (_fmt:json)
+- Protobuf (_fmt:pb)
+- Protobuf compressed? (_fmt:pc)
+- HTML (_fmt:html)
+- Protobuf encoded in JSON (_fmt:jspb).
+
+Some of this implementations are shared by other engines:
+The implementation is shared by other engines:
+
+- :ref:`google images internal engine`
+- :ref:`google news internal engine`
+- :ref:`google videos internal engine`
+
+"""
+
+from urllib.parse import urlencode
+from json import loads, dumps
+from searx.utils import html_to_text
+
+# pylint: disable=unused-import
+from searx.engines.google import (
+    get_lang_info,
+    detect_google_sorry,
+    supported_languages_url,
+    time_range_dict,
+    filter_mapping,
+    _fetch_supported_languages,
+)
+
+# pylint: enable=unused-import
+
+# about
+about = {
+    "website": 'https://www.google.com',
+    "wikidata_id": 'Q9366',
+    "official_api_documentation": 'https://developers.google.com/custom-search/',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+# engine dependent config
+categories = ['general', 'web']
+paging = True
+time_range_support = True
+safesearch = True
+send_accept_language_header = True
+
+
+def request(query, params):
+    """Google search request"""
+
+    offset = (params['pageno'] - 1) * 10
+
+    lang_info = get_lang_info(params, supported_languages, language_aliases, True)
+
+    # https://www.google.de/search?q=corona&hl=de&lr=lang_de&start=0&tbs=qdr%3Ad&safe=medium
+    query_url = (
+        'https://'
+        + lang_info['subdomain']
+        + '/search'
+        + "?"
+        + urlencode(
+            {
+                'q': query,
+                **lang_info['params'],
+                'ie': "utf8",
+                'oe': "utf8",
+                'start': offset,
+                'filter': '0',
+                'asearch': 'arc',
+                'async': 'use_ac:true,_fmt:json',
+            }
+        )
+    )
+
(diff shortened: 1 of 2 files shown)


dalf:
With the commit b184a77d087f3238a35733a11ccfc883651117b4 , the Google image engine uses the internal API ; however pagination is still broken.

unixfox:
On hold because google is not returning any meaningful results but instead a bunch of `result_args_cache`.

unixfox:
Closing as google doesn't allow the usage of this API anymore.

PR #1762, opened 2022-09-11
Re-use morty's default config while Docker startup

der-domi:
## What does this PR do?

This PR enables morty's default configuration `result_proxy:` in [settings.yml](https://github.com/searxng/searxng/blob/eb3d185e66dea1add88ef8d3a3c6bb74b8a3b166/searx/settings.yml#L124) instead of adding a duplication at the end of the file while docker starts-up and updating a configuration.
The default config is commented which will just be uncommented and the given parameter `${MORTY_URL}` and `${MORTY_KEY}` will be applied.

As an additional change the option `proxify_results` will be automatically enabled.

This avoids a confusing configuration.

1. Run SearXNG via Docker and force to update the settings
   * e.g. via renaming `settings.yml`
2. compare `settings.yml` with origin one.

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -81,15 +81,16 @@ patch_searxng_settings() {
     # Morty configuration
 
     if [ -n "${MORTY_KEY}" ] && [ -n "${MORTY_URL}" ]; then
-        sed -i -e "s/image_proxy: false/image_proxy: true/g" \
+        sed -i \
+            -e "s/image_proxy: false/image_proxy: true/g" \
+            -e "s|# result_proxy:|result_proxy:|g" \
+            -e "s|#   url: http://127.0.0.1:3000/|  url: ${MORTY_URL}|g" \
+            -e "s|#   # the key is a base64 encoded string|  # the key is a base64 encoded string|g" \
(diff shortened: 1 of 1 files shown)


return42:
Do we need a morty config in the docker distro? .. there is already the image proxy active.

der-domi:
No. The morty infrastructure is already given in [docker-entrypoint.sh](https://github.com/searxng/searxng/blob/master/dockerfiles/docker-entrypoint.sh). The config will be applied, once the environment variables `MORTY_URL` and `MORTY_KEY` are give via `docker-compose.yaml`.

der-domi:
Ah. Maybe I missunderstood your question. For `proxify_results` the same morty proxy will be used.

mrpaulblack:
Hi @der-domi ,
I checked your PR and the regex in your sed command in its current form does not work. Merging this PR would also mean that with every `settings.yml` change we would also need to crosscheck that the docker entrypoint script still works... The recommended way is to use a custom settings.yml file mounted as a volume for example inside the docker container where you only need to add the changes that you want to make from the default config, example here: https://github.com/searxng/searxng-docker/blob/master/searxng/settings.yml

For those reasons we decided to not merge your PR, but dont feel discouraged to open up another PR if you want to improve the project in any other way :+1: ; Closing.

PR #1801, opened 2022-09-24
Hong Kong is Hong Kong, not Hong Kong China

sdip15fa:
## What does this PR do?

Changed "China Hong Kong" to "Hong Kong" in languages.py.

Hong Kong should not be "Hong Kong China". This wording simply supports the totalitarian government in china. To be honest, I feel being offended by this word.

I am not saying Hong Kong is not China (this is illegal anyway, ~~in Hong Kong saying Hong Kong is not China is a crime and has a maximum penalty of a life sentence~~), but you don't call things like "New York, United States", or "Glasgow, United Kingdom".

### DIFF
--- a/searx/languages.py
+++ b/searx/languages.py
@@ -68,6 +68,6 @@ language_codes = (
     ('vi-VN', 'Tiếng Việt', 'Việt Nam', 'Vietnamese', '\U0001f1fb\U0001f1f3'),
     ('zh', '中文', '', 'Chinese', '\U0001f310'),
     ('zh-CN', '中文', '中国', 'Chinese', '\U0001f1e8\U0001f1f3'),
-    ('zh-HK', '中文', '中國香港', 'Chinese', '\U0001f1ed\U0001f1f0'),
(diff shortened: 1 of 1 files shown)


return42:
hint: the file `searx/languages.py` is generic and automatic generated by using tools like babel.

> Hong Kong should not be "Hong Kong China". This wording simply supports the totalitarian government in china. To be honest, I feel being offended by this word.

We can fully understand your upset on this, but this naming come from babel [1] and babel itself gets its territory names from the Common Locale Data Repository (CLDR), a project of the Unicode Consortium to provide locale data (including translations for territory and country names).

We are sorry, but we will have to refer you to the appropriate consortium on this issue. I hope you understand that we close this issue.



[1] https://babel.pocoo.org/en/latest/dev.html#tracking-the-cldr
[2] https://en.wikipedia.org/wiki/Common_Locale_Data_Repository

JasonHK:
@return42 CLDR does have an alternative, shorter display name for territory code `HK`, i.e. "Hong Kong" instead of "Hong Kong SAR China". So it should be possible to configure the generator script to spit out "Hong Kong" instead. If I found the correct configuration, would you accept a PR for this?

PR #1805, opened 2022-09-24
[mod] paper.html: add links to doi resolver

return42:
## What does this PR do?

paper.html: add links to doi resolver

Check link of items that have a DOI ..

grafik


dalf:
I was thinking about that, thank you.
I was wondering if the URL and the reference URL by the DOI were different: it seems this seems to be the case quite often.

dalf (approved):

PR #1806, opened 2022-09-24
fix some leftovers in the science category

return42:
## What does this PR do?

- paper.html: simplify template by using result_link macro
- core.ac.uk: add cetgory 'scientific publications'

Do some test in the science category


dalf (approved):
👍

PR #1830, opened 2022-10-01
Use official sci-hub mirrors

MohamedElashri:
## What does this PR do?

This PR makes SearxNG use the official mirrors.

Modify `doi_resolvers` to include only the `sci-hub` offical mirrors
`sci-hub.ru`
`sci-hub.se`
`sci-hub.st`

Many of the mirrors provide different wallets for donation (Alexandra consider [this ](https://twitter.com/ringo_ring/status/1431376651691114501)theft/fraud).  Also some of the current links does not work

Nothing special, I built custom docker image with the change and currently using it for sometime on my instance


dalf:
Thank you !

* `.ru` and `.se` give me the PDF for the DOI `10.1086/383302`
* however I can't find a result on the `.st` domain. Can you confirm that?

---

[EDIT] Everywhere I see reference to the ru,se and st, so I'm going to merge.

MohamedElashri:
> Thank you !
> 
> 
> 
> * `.ru` and `.se` give me the PDF for the DOI `10.1086/383302`
> 
> * however I can't find a result on the `.st` domain. Can you confirm that?
> 
> 
> 
> ---
> 
> 
> 
> [EDIT] Everywhere I see reference to the ru,se and st, so I'm going to merge.

Hi @dalf 

sci-hub.st works normally for me. If you  random old physical review article (not open access) like [this](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.103.115106) it will work. 

I am not sure this is the mirror behavior or something with US ISPs but I cannot open '.st' from US even with some US vpns (tried a couple of mullvad and adguard vpn US servers). But when I switch to non-US servers it works. 

Try to reproduce that. Thank you for quick merger. 


Edit, I have been able to access the DOI paper you mentioned also on `.st`. Try accessing this directly and see 
https://sci-hub.st/10.1086/383302. 

PS: This mirror is protected via `DDos-Guard` and you might have extension or DNS server that is having problem with that.

PR #1851, opened 2022-10-09
feat(engines): Add google_cs

whs:
## What does this PR do?

Add [Google Custom Search JSON API](https://developers.google.com/custom-search/v1/overview) search engine.

I've been wondering if Google Custom Search JSON API would be a less-hacky but costly approach to a privacy search engine, instead of scraping Google. This feature was previously discussed in original searx's https://github.com/searx/searx/issues/2622 .

Turns out the search API did use different search implementation than Google.com's, so it should be considered as a different engine altogether. Also, users may configure the Custom Search to search a specific site and use Searx as a site-specific search engine instead.

1. Enable the google_cs engine following the documentation
2. Run `make run` and search

## Other notes

- ~~I did not test the quota exceeded exception - turns out once you added a billing project you can't remove it. I believe the exception should be similar to other googleapis.com APIs, so that's what I'm trying to check.~~
- I believe since it is a googleapis endpoint, you could use a Google service account to access it (so the API key is not required when running on GCE). As this do not use official Google SDK, this feature is not supported.

## Related issues

https://github.com/searx/searx/issues/2622 from original searx

### DIFF
--- a/docs/src/searx.engines.google.rst
+++ b/docs/src/searx.engines.google.rst
@@ -30,6 +30,14 @@ Google WEB
 .. automodule:: searx.engines.google
   :members:
 
+.. _google custom search engine:
+
+Google Custom Search
+====================
+
+.. automodule:: searx.engines.google_cs
+  :members:
+
 .. _google images engine:
 
 Google Images
--- a/searx/engines/google_cs.py
+++ b/searx/engines/google_cs.py
@@ -37,8 +37,10 @@
 """
 from urllib.parse import urlencode
 
+from dateutil.parser import isoparse
+
 from searx.engines.google import get_lang_info
-from searx.exceptions import SearxEngineAPIException, SearxEngineTooManyRequestsException
+from searx.exceptions import SearxEngineTooManyRequestsException
 from searx.network import raise_for_httperror
 
 about = {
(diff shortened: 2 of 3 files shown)


pedrojreis:
This would be amazing to have it merged. @return42  think this would be a possibility?

wei840222:
In fact, I did the same customization in my local deployment to obtain Google search results. If this could be merged, I wouldn't have to do it myself.
https://github.com/wei840222/searxng/commit/cf6e6ed9ef72cd657003fea7dfcb8be0ef5767fc

allixx:
What's maintainers take on this? Should CSE be added as a separate engine, or integrated into existing one?
Assuming they are not against CSE in principle.

return42:
@whs Unfortunately, I can't explain how your PR could have been overlooked at the time.

Due to recent events, your PR has received renewed attention (see messages above). Are you still available for a review process? If so, please perform a rebase. 

*Thank you very much, and please accept my apologies for overlooking your PR such a long time / next time feel free to ping me.*

whs:
I think this PR was done as my Hacktoberfest project at the time, presumably, the project had additional workload at the time?

I'll try to make time to rebase if it's trivial to do, but I make no promise. I don't intend to use this PR myself so it is very low priority - if you have used CSE yourself you'll realize it's no Google - the search quality is worse than Bing or DDG imo (as I've written in the PR description), but the only way I could've know that is to write this PR. Eventually this led me to realize that paying for Kagi (and its API) is the right choice - they're probably using some private Google API and not CSE.

If there's no update from me in a month, anyone else can feel free to take over (or actually you can do that now if you want to expedite the process)

whs:
Looking at my projects on hand I don't think I'll have time to finish this. If someone else would like to adopt this PR, please feel free.

whs:
I guess this PR could be closed now.

I just learned that on https://developers.google.com/custom-search/v1/overview Google is shutting down Custom Search JSON API at the end of this year. They also have closed down signups. There's no other alternative offering from Google except for a "call for business partnership" product.

return42:
> I guess this PR could be closed now.

Nevertheless, thanks a lot for having an eye on this topic and sharing your knowledge with the community over such a long time :+1:

PR #1856, opened 2022-10-11
The checker requires Redis

dalf:
## What does this PR do?

Remove the abstraction in searx.shared.SharedDict.
Implement a basic and dedicated scheduler for the checker using a Redis script.


searx/shared/redisdb.py
```diff
@@ -26,26 +26,31 @@
...
-        # initialized only one is kept, the others are garbage collected.
-        _client = redis.Redis.from_url(get_setting('redis.url'))
-    return _client
 
+def client() -> redis.Redis:
+    return _CLIENT
 
-def init():
+
+def initialize():
```
return42, line 40: Why is this function renamed? / at least the doc-string in line 11 needs to be modified

https://github.com/searxng/searxng/blob/ee4475ff4b7f1d7d687606664c3b1dd4cafe5a99/searx/shared/redisdb.py#L10-L13

searx/webapp.py
```diff
@@ -120,6 +120,7 @@
 # renaming names from searx imports ...
 from searx.autocomplete import search_autocomplete, backends as autocomplete_backends
 from searx.languages import language_codes as languages
+from searx.shared.redisdb import initialize as redis_initialize
```
return42, line 123: Instead of having different names for the same object (``import .. as ..``) I would prefer:

```python
from searx.shared import redisdb
...
redisdb.initialize()
```

this will also reduce the imported names (if more names from that imported lib are needed).

PR #1896, opened 2022-10-26
Reduce Docker image size

ngosang:
## What does this PR do?

* Copy only source code and Docker entrypoint
* Avoid updating packages. The base image is updated frecuently

* Docker image size reduced by 3 MB (195 MB -> 192 MB uncompressed)

Deploy and check everything is working as before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,7 @@ WORKDIR /usr/local/searxng
 
 COPY requirements.txt ./requirements.txt
 
-RUN apk upgrade --no-cache \
- && apk add --no-cache -t build-dependencies \
+RUN apk add --no-cache -t build-dependencies \
     build-base \
     py3-setuptools \
     python3-dev \
@@ -47,8 +46,7 @@ RUN apk upgrade --no-cache \
     brotli \
  && pip3 install --no-cache -r requirements.txt \
  && apk del build-dependencies \
- && rm -rf /root/.cache \
- && find /usr/lib/python*/ -name '*.pyc' -delete
+ && rm -rf /root/.cache
 
 COPY --chown=searxng:searxng dockerfiles ./dockerfiles
 COPY --chown=searxng:searxng searx ./searx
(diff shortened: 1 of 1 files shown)


ngosang:
@return42 how long it takes the review process? I'm a bit concerned about the project health.

return42:
> I'm a bit concerned about the project health.

Sorry for the delay .. I'm not the docker expert, so I requested ([here](https://github.com/searxng/searxng/pull/1896#event-7689327681)) a review of this PR from our docker experts @dalf and @mrpaulblack.

ngosang:
@return42 I want to open more PRs:
* Update Alpine 3.17 => Fix this issue https://github.com/searxng/searxng-docker/issues/31
* Update Python 3.11 and include it in CI/CD => Performance gains https://www.phoronix.com/review/python-311-benchmarks/4

I'm an expert in Docker and I will be reachable for any issue caused by my PRs, but I won't open more PRs until I see this merged, or at least some activity.

return42:
@dalf & @mrpaulblack you are the maintainer of the docker part, can you please have a look at this PR .. @ngosang is still waiting since over a month without getting any feedback.

dalf:
Quick answer:
Related to https://gitlab.alpinelinux.org/alpine/aports/-/issues/11906#note_232487
I was hoping the issue to be fixed in Alpine 3.17.
Some benchmarks are needed on SBC like Raspberry Pi.

unixfox (approved):

unixfox (changes requested):
No obvious issues. Good idea to chown the searx directory.
Can you prove that removing the compiled sources doesn't affect the startup time by too much?

ngosang:
@unixfox You are right. I just did some tests and these are the results. This is the time since I execute "docker run" in my machine until uwsgi initialization ends. My computer is really fast.
* Docker image with compiled sources => 2.61 seconds
* Docker image without compiled sources => 6.31 seconds

I reverted the changes related to compiled files in this PR. The changes in size are just 3 MB but the image will build a bit faster too.

unixfox (approved):
Thank you for the test @ngosang 
@return42 good to merge

Issue #1680, opened 2022-08-19
Make SearXNG default search in Firefox


slashgob:
Version: 
SearXNG 2022.08.19-6f28a69f
Firefox 103.0

**How did you install SearXNG?**
Installed using Docker image provided by xthursdayx on UNRAID. Self hosted using NGINX as proxy. https://search.slashgob.com/

**What happened?**
I cannot make SearXNG the default search engine in Firefox. Reading Firefox documentation, I should be able to click on the URL and add SearXNG to the list of search engines if it supports Opensearch

**How To Reproduce**
navigate to SearXNG search page and click on URL in Firefox

**Expected behavior**
Clicking on URL should allow you to add SearXNG to the list of search engines in Firefox

return42:
Seems Cloudflare adds some crap to your side

![grafik](https://user-images.githubusercontent.com/554536/185621116-dea7adfe-6c37-4d22-aa66-4fc5f82f86d5.png)

Beside this, Cloudflare can read everything (because they MITM your https connection) what is a disaster for the privacy of the user .. don't use Cloudflare!

dalf:
@return42 are you sure this is not related to https://github.com/searxng/searxng/pull/1655 ?

return42:
No it isn't / #1655 works by example on my instance and on @tiekoetter and @unixfox instances

But the first I get when I open https://search.slashgob.com/ is a CSP violation 

---

EDIT IDK why all guys think they need Cloudflare .. they do MITM and are the culprit of dozens of issues . IMO we should no longer give any support to Cludlfare users.

LalleSX:
If you already have a SearxNG in your search engine list on Firefox you can either remove that or change the name of your instance in the `/etc/searxng/settings.yml`.

LalleSX:
Just looked at your site its definitely the title, but its the lack of one. 
If you look at the tab it just says the URL _search.slashgob.com_ where for example [tiekoetter](https://searx.tiekoetter.com/) has a title _SearXNG @ searx.tiekoetter.com_

So edit `/etc/searxng/settings.yml`
And make it look something like this:
![pic-selected-221028-1658-52](https://user-images.githubusercontent.com/29478339/198664195-f1d8f267-1014-4920-afe4-56327019f22e.png)

mrx23dot:
Please provide a way to accept **queries in URL string**, so it could by added to any browser by substituting that string **/%s**.

[...]

PR #1903, opened 2022-10-28
Added info about adding the enginge to the brownser

LalleSX:
## What does this PR do?

It adds information about a problem that can occur when adding the site to your browser.
Specify about trying to add one when there is already one added with the same title/name   

Well i personally wasted a lot of hours trying to know what was wrong with my instance, when it was just the name.

There is no special code its markdown so i am 99.9% sure it works the same.

## Author's checklist
Works on my machine

## Related issues
[Slashbobs issue](https://github.com/searxng/searxng/issues/1680#issue-1344200084)
The issue in that case was the lack of a title but it is still sort of the same issue


return42 (approved):
@LalleSX sorry for the late response on your PR .. I rebased, slightly modified the text and added DE and ID translations.

Thanks :+1:

PR #1918, opened 2022-10-31
Remove unused volume

LinuxMeow:
Docker file used old project volume instruction which is not existent in current one. It used to create clutter in created volumes without names and purposes polluting the docker environment.

Volume doesn't point to existing directory inside container and has no actual use other then creating docker volume mess each time container is run recreated or potentially restarted.


dalf:
Thank you!

PR #1929, opened 2022-11-04
GH workflow: various improvements

dalf:
## What does this PR do?

GitHub displays this warning with the actions/setup-python@2 :
> Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.
For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This PR : 
* avoid GH warnings:
  * use actions/setup-python@v4 (previously v3)
  * use actions/cache@v3 (previously v2)
* in addition to `./local`, now `./.nvm` and `./node_modules` are cached : faster build time.
* add Python 3.11
* set content permission to read by default.

* see above

N/A

### remaining warning

I'm not sure why, this warning remains
```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout
```

Setting the node version doesn't fix the issue.
```yaml
- uses: actions/setup-node@v3
  with:
    node-version: 16
```

Note: Ubuntu 20.04 has [node version 10](https://packages.ubuntu.com/focal/nodejs).

### Ubuntu-22.04

I've tried to use `ubuntu-22.04` instead of `ubuntu-20.04` : it does not work by default because Firefox is a snap in Ubuntu 22.04 --> geckodriver can't connect to it (also, the Ubuntu package `firefox-geckodriver` doesn't exist for the version 22.04 and later).

The [geckodriver version 0.32.0](https://github.com/mozilla/geckodriver/releases/tag/v0.32.0) should fix that but it requires additional changes. The alternative is to download the last Firefox version from mozilla.org.

### sudo ./utils/searxng.sh install packages

`./utils/searxng.sh install packages` provides the Ubuntu packages to build the Python & Node dependencies.
However, if the GH cache provides them ( actions/cache ), the Ubuntu packages are not required.
We can write
```yaml
    - name: Install Firefox
      run: sudo apt install firefox
    ...
    - name: Install Ubuntu packages
      if: steps.cache-python.outputs.cache-hit != 'true'
      run: |
        sudo ./utils/searxng.sh install packages
```

This should speed up the build time by ~ 30 seconds (tested in my fork).

I haven't do the change in this PR, I'm afraid it breaks something.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -28,9 +28,12 @@ jobs:
         architecture: 'x64'
     - name: Cache Python dependencies
       id: cache-python
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
-        path: ./local
+        path: |
+          ./local
+          ./.nvm
+          ./node_modules
         key: python-${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements*.txt', 'setup.py') }}
     - name: Install Python dependencies
       if: steps.cache-python.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


return42 (approved):
I am not familiar with https://github.com/actions/setup-python/releases .. but it seems to work --> LGTM

dalf:
There are still warnings: https://github.com/searxng/searxng/actions/runs/3392352948

I'm doing some tests here to avoid notification spam : https://github.com/dalf/searxng/pull/172

dalf:
I have updated the description of the PR and the commits

PR #1934, opened 2022-11-05
User agent: don't include the patch number in the Firefox version

### LINKED ISSUE #1914: SearXNG Firefox user-agent doesn't match real Firefox
2022.10.29-fc9986de
Script
**What happened?**
SearXNG will send a user-agent such as:  
`Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0.2) Gecko/20100101 Firefox/106.0.2`
However, real Firefox looks like this:  
`Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0) Gecko/20100101 Firefox/106.0`

Firefox does not send servers the patch version the user is running.

**How To Reproduce**
See https://github.com/searxng/searxng/blob/fc9986de0a21b01c1c8e342a9c4c8c6b96cc7ea2/searx/utils.py#L69-L74 and https://github.com/searxng/searxng/blob/fc9986de0a21b01c1c8e342a9c4c8c6b96cc7ea2/searx/data/useragents.json

To confirm Firefox's behavior, use a Firefox version with non-zero patch version and check its user-agent.

**Expected behavior**
SearXNG creates a User-Agent header that matches the format of the browser it is pretending to be.

**Additional context**
Firefox seems to have stopped sending the patch version with the release of Firefox 16: https://web.archive.org/web/20181123134532/https://www.fxsitecompat.com/en-CA/docs/2012/ua-string-no-longer-contains-patch-level-version-number/

dalf:
## What does this PR do?

The Firefox version in the user agent doesn't include the patch version: 106.0 not 106.0.2

Make SearXNG closer Firefox regarding the user agent.

## Related issues

Close #1914


return42 (commented):
Shouldn't we offer more than two versions?

dalf:
> Shouldn't we offer more than two versions?

As far I remember, some engines don't provide any result (or behave differently) for old Firefox versions.
But sure, the reduced number of user agent bother me, that's why I didn't merge the pull request.

However currently, the user agent header yells "this is not a really Firefox" (in addition to the different headers, different TLS & curve signatures).

Perhaps we can include MacOS but it requires some testing because some engines may return different results.

PR #1938, opened 2022-11-06
Update network.py

ghost:
## What does this PR do?

This PR extends the proxy timeout for checking Tor to 60 seconds. This could be a good compromise between the default value of 10 and totally disabling the timeout. 

On some systems 10 seconds could be a short time at boot and SearxNG would crash because of a proxy timeout. 

## Related issues

Closes #1537

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -173,7 +173,7 @@ class Network:
             ):
                 continue
             return False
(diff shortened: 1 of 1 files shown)


dalf:
Thank you!

ghost:
> Thank you!

You are welcome 😊it seems to work for me when using SearxNG over Tor.

PR #1945, opened 2022-11-10
[fix] follow up of PR-1856

return42:
- `` url: false`` breaks with existing installations   https://github.com/searxng/searxng/commit/e92755d358df5b34b0181f48f8ba02c7f2939e8f#r89577339

- PR-1856: https://github.com/searxng/searxng/pull/1856

- Comments from review: https://github.com/searxng/searxng/pull/1856#pullrequestreview-1176402810

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -305,7 +305,7 @@ In this example read/write access is given to the *searxng-redis* group.  To get
 access rights to redis instance (the socket), your SearXNG (or even your
 developer) account needs to be added to the *searxng-redis* group.
 
-``url``
+``url``: ``unix:///usr/local/searxng-redis/run/redis.sock?db=0``
   URL to connect redis database, see `Redis.from_url(url)`_ & :ref:`redis db`::
 
     redis://[[username]:[password]]@localhost:6379/0
@@ -314,20 +314,14 @@ developer) account needs to be added to the *searxng-redis* group.
 
 .. admonition:: Tip for developers
 
-   To set up a local redis instance, first set the socket path of the Redis DB
-   in your YAML setting:
-
-   .. code:: yaml
-
-      redis:
-        url: unix:///usr/local/searxng-redis/run/redis.sock?db=0
-
-   Then use the following commands to install the redis instance ::
+   To install a local redis instance that is connected by the default URL use
+   the following commands::
 
      $ ./manage redis.build
      $ sudo -H ./manage redis.install
      $ sudo -H ./manage redis.addgrp "${USER}"
-     # don't forget to logout & login to get member of group
+
+   Don't forget to logout & login to get member of group.
 
 
 .. _settings outgoing:
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -76,9 +76,9 @@ server:
     X-Robots-Tag: noindex, nofollow
     Referrer-Policy: no-referrer
 
-redis:
-  # https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url
-  url: false
+# redis:
+#   # https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url
+#   url: unix:///usr/local/searxng-redis/run/redis.sock?db=0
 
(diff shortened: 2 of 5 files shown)


dalf:
> url: false breaks with existing installations

Do the installation scripts use this following file  ?
https://github.com/searxng/searxng/blob/3a765113eefc7ac0b1ee94e9a0ac0db8069ec336/utils/templates/etc/searxng/settings.yml#L18-L19

If yes, the existing installation should be fine.

> it breaks all developer scenarios.

It doesn't work for developer having a Redis instance on the default TCP port.

There is `SEARXNG_DEBUG` for `general.debug`.
It is easy to add `SEARXNG_REDIS_URL` for `redis.url`:
* it avoids to push development settings to all installations
* and most importantly, if `redis.url` is defined, the code can be the URL should work, there is no "here an exception, but actually ignore that or may be not".

return42:
> If yes, the existing installation should be fine.

Yes, sorry .. I didn't have used this template in my instance so I had the issue.

> It is easy to add `SEARXNG_REDIS_URL` for `redis.url`:

OK .. when I have time I will have a look and update this PR / thanks!

return42:
> * and most importantly, if `redis.url` is defined, the code can be the URL should work, there is no "here an exception, but actually ignore that or may be not".

What if redis server is not up and running?

dalf:
> What if redis server is not up and running?

If the user has configured an URL and it doesn't work, then this is an error.
In this scenario, SearXNG can exit.

However if the value is `unix:///usr/local/searxng-redis/run/redis.sock?db=0`, SearXNG can not decide.

> OK .. when I have time I will have a look and update this PR / thanks!

<details>

<summary>Here a patch based on the master branch</summary>

```patch
diff --git a/manage b/manage
index c887826e2..3879087b2 100755
--- a/manage
+++ b/manage
@@ -31,6 +31,8 @@ export NODE_MINIMUM_VERSION="16.13.0"
 # SPHINXOPTS=
 BLACK_OPTIONS=("--target-version" "py37" "--line-length" "120" "--skip-string-normalization")
 BLACK_TARGETS=("--exclude" "searx/static,searx/languages.py" "--include" 'searxng.msg|\.pyi?$' "searx" "searxng_extra" "tests")
+DEFAULT_DEV_REDIS_SOCKET="/usr/local/searxng-redis/run/redis.sock"
+
 
 pylint.FILES() {
 
@@ -144,6 +146,11 @@ webapp.run() {
         sleep 3
         xdg-open http://127.0.0.1:8888/
     )&
+    # set SEARXNG_REDIS_URL if it is not defined and "{DEFAULT_DEV_REDIS_URL}" exists.
+    if [ -S "${DEFAULT_DEV_REDIS_SOCKET}" ] && [ -z "${SEARXNG_REDIS_URL}" ]; then
+        export SEARXNG_REDIS_URL="unix://${DEFAULT_DEV_REDIS_SOCKET}?db=0"
+        build_msg RUN "Use Redis URL ${SEARXNG_REDIS_URL}"
+    fi
     SEARXNG_DEBUG=1 pyenv.cmd python -m searx.webapp
 }
 
diff --git a/searx/settings_defaults.py b/searx/settings_defaults.py
index 470290f92..cfa1bb47c 100644
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -174,7 +174,7 @@ SCHEMA = {
         'default_http_headers': SettingsValue(dict, {}),
     },
     'redis': {
-        'url': SettingsValue((None, False, str), False),
+        'url': SettingsValue((None, False, str), False, 'SEARXNG_REDIS_URL'),
     },
     'ui': {
         'static_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'static')),
```

</details>

* `SEARXNG_REDIS_URL=redis://localhost?db=0 make run` works.
* if `/usr/local/searxng-redis/run/redis.sock` is a socket, `make run` connect automatically to it EXCEPT if SEARXNG_REDIS_URL is already defined.

(`make test.shell` doesn't show any error, but the script might need some adjustment)

return42:
closed in favorite of https://github.com/searxng/searxng/pull/1954

PR #1954, opened 2022-11-12
[fix] follow up of PR-1856

dalf:
## What does this PR do?

* This first commit is the patch in this comment: https://github.com/searxng/searxng/pull/1945#issuecomment-1312001620
* The second commit moves `searx.shared.redisdb` to `searx.redisdb`.

Clean up

## Author's checklist

In https://github.com/searxng/searxng/pull/1945#issuecomment-1312001620 one idea is stop SearXNG is the Redis connection can't be established. Reading the code, I'm meh :

```python
    except redis.exceptions.RedisError as e:
        _CLIENT = None
        _pw = pwd.getpwuid(os.getuid())
        msg = "[%s (%s)] can't connect redis DB ..."
        if redis_url == OLD_REDIS_URL_DEFAULT_URL and isinstance(e, redis.exceptions.ConnectionError):
            logger.exception(msg, _pw.pw_name, _pw.pw_uid)
            logger.info(
                "You can safely ignore the above Redis error if you don't use Redis. "
                "You can remove this error by setting redis.url to false in your settings.yml."
            )
        else:
            logger.fatal(msg, _pw.pw_name, _pw.pw_uid)
            sys.exit(1)
```

In webapp, the change from `redis_initialize()` to `redisdb.init()` can be done in another PR.

## Related issues

Related to:
* #1856
* #1945

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -294,7 +294,7 @@ Global Settings
 
 .. _Redis.from_url(url): https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url
 
-A redis DB can be connected by an URL, in :py:obj:`searx.shared.redisdb` you
+A redis DB can be connected by an URL, in :py:obj:`searx.redisdb` you
 will find a description to test your redis connection in SerXNG.  When using
 sockets, don't forget to check the access rights on the socket::
 
--- a/docs/admin/installation-uwsgi.rst
+++ b/docs/admin/installation-uwsgi.rst
@@ -239,8 +239,8 @@ For example on Fedora (RHEL): If you try to install a redis DB with socket
 communication and you want to connect to it from the SearXNG uWSGI, you will see a
 *Permission denied* in the log of your instance::
 
-  ERROR:searx.shared.redis: [searxng (993)] can't connect redis DB ...
-  ERROR:searx.shared.redis:   Error 13 connecting to unix socket: /usr/local/searxng-redis/run/redis.sock. Permission denied.
+  ERROR:searx.redisdb: [searxng (993)] can't connect redis DB ...
+  ERROR:searx.redisdb:   Error 13 connecting to unix socket: /usr/local/searxng-redis/run/redis.sock. Permission denied.
   ERROR:searx.plugins.limiter: init limiter DB failed!!!
 
 Even if your *searxng* user of the uWSGI process is added to additional groups
--- a/docs/admin/update-searxng.rst
+++ b/docs/admin/update-searxng.rst
@@ -132,5 +132,4 @@ to see if there are some left overs.  In this example there exists a *old*
    INFO:  [OK] (old) account 'filtron' does not exists
    INFO:  [OK] (old) account 'morty' does not exists
    ...
-   INFO    searx.shared                  : Use shared_simple implementation
-   INFO    searx.shared.redis            : connected redis DB --> default
+   INFO    searx.redisdb            : connected redis DB --> default
--- a/docs/admin/update-searxng.rst
+++ b/docs/admin/update-searxng.rst
@@ -132,4 +132,5 @@ to see if there are some left overs.  In this example there exists a *old*
    INFO:  [OK] (old) account 'filtron' does not exists
    INFO:  [OK] (old) account 'morty' does not exists
    ...
-   INFO    searx.redisdb            : connected redis DB --> default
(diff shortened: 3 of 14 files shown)


return42 (changes requested):
I added two WIP commits on top we can squash .. I left some remarks in the commit messages.

dalf:
About 1b67922655edd28b36c6145b1a33ae059bed24d5 : 

> The redis URL should be logged before connect, otherwise when connection fails the URL is not shown in the log.

This surprised me, but `redis.Redis.from_url` does not create any connection :

```python
c = redis.Redis.from_url('redis://nowhere')
print('c=', c, '(no exception in the line above)')
c.ping()
print('unreachable code, exception in the line above') 
```

> OLD_REDIS_URL_DEFAULT_URL is not needed; when redis.url is set and the redis DB is not available an error should be logged.

* Previously, Redis connection errors were not shown in the logs except if the limiter was enabled.
* With #1942 , an error is shown if `redis.url` is set and is not a valid URL.
* There are some existing installations where the value of `redis.url` is `unix:///usr/local/searxng-redis/run/redis.sock?db=0`. These installations are going to display an error if Redis is not installed and without context admins are going to open issues about that.

About `redis.exceptions.RedisError` instead of `redis.exceptions.ConnectionError` : since there is a ping command, the Redis client might raise an exception more generic than ConnectionError (?)

return42:
> This surprised me, but `redis.Redis.from_url` does not create any connection :

Yes, that's why you have added the ping .. was it a question or is it not clear why I want to log the URL before ping?

> Redis client might raise an exception more generic than ConnectionError (?)

I don't know, but I don't want to fetch to general exceptions .. to fetch a ConnectionError is OK since we can handle it (a redis connect is not needed to drive SearXNG / as long we do not make redis mandatory .. what we should do in the next future).

> Previously, Redis connection errors were not shown in the logs except if the limiter was enabled.

On the way to make redis mandatory one day, we should log an error right now (as long redis.url is not set to `false`).

> With https://github.com/searxng/searxng/pull/1942 , an error is shown if redis.url is set and is not a valid URL.

Its the same with the changes from mine --> whenever redis.url is set but not available, an error should be logged (very simple rule).

> There are some existing installations where the value of redis.url is unix:///usr/local/searxng-redis/run/redis.sock?db=0. These installations are going to display an error if Redis is not installed and without context admins are going to open issues about that.

By default a local redis is installed by the scripts and this redis.url (the socket) is in the searxng.yml template.

Even if some admins will add an issue to github's issue tracker .. this is good opportunity to encourage our admins to install redis .. and we can give the hint that redis becomes mandatory in the near future.

BTW: the error is logged but SearXNG continues to run / as redis is not mandatory.

dalf:
> was it a question or is it not clear why I want to log the URL before ping?

Sorry my words are not clear: I don't understand why the URL should be displayed before `redis.Redis.from_url` since `redis.Redis.from_url` does not create any connection ?

`redis.Redis.from_url` is convenient because it allows to display the parameters as seen by redis-py and without the Redis password (otherwise, if someone copy paste the logs without looking carefully, the password is going to be published on GitHub).

Also:
```
>>> redis.Redis.from_url('redis:///somewhere.com')  # triple slash instead of double
Redis<ConnectionPool<Connection<host=localhost,port=6379,db=0>>>
```

> The redis URL should be logged before connect, otherwise when connection fails the URL is not shown in the log.

A first connection is established in `_CLIENT.ping()`, not before.

Side note: I can't find documentation about the fact that `redis.Redis.from_url` does not create a connection, but 
* `c = redis.Redis.from_url('redis://nowhere')` actually does not trigger an exception
* the constructors don't make any call to `Connection.get_connection` (or any subclasses of Connection):
  * https://github.com/redis/redis-py/blob/fa45fb13730c6ece6b08184b4f475d841895b4d4/redis/connection.py#L1236-L1305
  * https://github.com/redis/redis-py/blob/fa45fb13730c6ece6b08184b4f475d841895b4d4/redis/client.py#L860-L1038

return42:
> A first connection is established in _CLIENT.ping(), not before.

Sorry my fail .. IDK why but I had the impression the connection was established by ``_CLIENT.get_connection_kwargs()``.

Anyway I think masking `username:password@` by a ``re.sub(r':\/\/(.*?)\@', r'://xxx:xxx@', redis_url)`` is much simpler (no additional knowledge/technology is needed to understand what happens here, it needs less lines of code and we can use this re here and in any other place where we want to mask `username:password@` from an URL).

BTW: Placing the debug message outside of the try/except block reduce the code in the try/except (those blocks should contain only the critical code).

> I can't find documentation about the fact that redis.Redis.from_url does not create a connection, but

.. late evaluating is a common concept I think.

PR #1974, opened 2022-11-24
[fix] Deserialize preferences which are explicitly blank strings

### LINKED ISSUE #1568: Setting autocomplete to “-“ (none) does not save the preference in preference URL

2022.07.26

Using a public instance 
**What happened?**

When setting the autocomplete to “-“ (none/disable), the setting is not saved in the preferences URL but resets to the default autocomplete engine when pasting the preferences URL in a clean browser window. I use the custom preferences URL function and would like to disable autocomplete.

**How To Reproduce**

1. Go to Searx Preferences
2. Set autocomplete to “-“ (none)
3. Go to cookies tab
4. Copy preferences URL (the bottom one)
5. Paste into new browser/private window
6. Go to Searx preferences in new browser/private window
7. Autocomplete shows default setting (in my case “google”) instead of “-“

**Expected behavior**
It should be possible to deactivate autocomplete completely when using the preferences URL.

fehho:
Default behavior of urllib.parse_qs() is to treat a blank string as an undefined value. parse_qs() is used when preferences are read in from a URL, causing a failure to set user preference when the user's preference is a blank string.

## What does this PR do?

Change parse_qs in preference URL de-serialization to respect blank values.
Does not change behavior of undefined values (when preference has not changed by user.)

A blank string is a valid choice for some preferences (autocomplete and private engine token)

1. Set a preference which gets stored as a blank string (disable autocomplete or have no private engine token) and save.
2. Go to cookies and copy down a preferences URL.
3. Go back to preferences and change the earlier setting to something different and save.
4. Use the URL and see the old preference restored.
5. Repeat above without PR and see the old preference not restored.

## Related issues

Closes #1568

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -441,7 +441,7 @@ class Preferences:
         """parse (base64) preferences from request (``flask.request.form['preferences']``)"""
         bin_data = decompress(urlsafe_b64decode(input_data))
         dict_data = {}
-        for x, y in parse_qs(bin_data.decode('ascii')).items():
+        for x, y in parse_qs(bin_data.decode('ascii'), keep_blank_values=True).items():
             dict_data[x] = y[0]
         self.parse_dict(dict_data)
 


return42:
@dalf can you merge this PR?

It seems I do not have any longer write acces to the SearXNG repo (why?) ..

grafik

dalf (approved):
I'm sorry for the delay.
Thank you !

PR #1988, opened 2022-11-29
Copy client connection kwargs before mutating

ryankask:
## What does this PR do?

It copies the redis connection kwargs before mutating them.

If Redis authentication is required, a password won't be used because it's been removed from the connection dict so it isn't logged.

Set up redis with a password and check that the client still connects.


return42:
@dalf this issue is also solved by the commit 1b67922655ed in https://github.com/searxng/searxng/pull/1954 .. should we merge this PR first to get a real bug fixed .. or do you want to continue with #1954 ?

dalf (approved):

PR #1736, opened 2022-09-01
[dev] Add EditorConfig support and fix duplicate key in simple theme ESLint configuration

FunctionalHacker:
## What does this PR do?
I was just starting to get familiar with the searxng source code and thought I would get my feet wet contributing some fixes for the development experience.

Main addition is the [EditorConfig](https://editorconfig.org/) configuration file. This allows people using different editors have some project wide settings applied, such as indent style, indent size, eol, charset etc.

For now, only configuration is eol and charset for all files and indent settings for JavaScript files. More can be added later.

At the same time, I noticed a duplicate key in the simple theme ESLint configuration so I included that fix here too.

## Why is this change important?

Improves developer experience and makes contributing for new developers easier

## How to test this PR locally?
Use an editor [with EditorConfig support built in](https://editorconfig.org/#pre-installed) or [install a plugin for your current editor](https://editorconfig.org/#download) and try if the settings work.

PR #1989, opened 2022-11-29
Add PyCharm configuration files to .gitignore

ngosang:
## What does this PR do?

Add PyCharm configuration files to .gitignore

PyCharm creates a folder called `.idea` in the project. This PR excludes those files from Git so they are not commited by mistake.

Tested by me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ gh-pages/
 
 /package-lock.json
 /node_modules/
+
+.idea/


return42:
Does it make sense to add a PyCharm configuration which might help others to get up a running PyCharm environment? 

By example for emacs users I added https://github.com/searxng/searxng/blob/master/.dir-locals.el and  there is also a PR for a `.editorconfig` https://github.com/searxng/searxng/pull/1736

ngosang:
I never commit the IDE configuration into the repositories for these reasons:
* In the `.idea` folder the user-configuration and project-configuration are mixed. My configuration might not work for other users (for example, the Python version, Python path...). If I change something in my IDE, Git detects the change and I could commit by mistake.
* Pycharm and other IDEs are able to create all the configuration from scratch when you open a new project.
* There are too many IDEs / text editors to commit the configuration for all of them.

Said that, I'm open to any solution.

return42:
> In the `.idea` folder the user-configuration and project-configuration are mixed. 

Thanks for clarify .. than we should not add the PyCharm config .. the `.dir-locals.el` and the `.editorconfig` I mentioned above include only settings related to the local SearXNG project and do not contain any user specific settings or absolute pathnames.

dalf:
Related to https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000154070-How-to-gitignore-idea-files

PR #2019, opened 2022-12-11
Replace langdetect with fasttext (followup of #1969)

ArtikusHG:
## What does this PR do?

This PR replaces langdetect with fasttext. Since fasttext is both faster and more accurate than langdetect, and we're using fasttext in #1969 anyway, we don't need two language detection libraries.

## Why is this change important?

This change is important to make sure SearXNG doesn't use unnecessary resources and libraries, and provides accurate language detection.

## How to test this PR locally?

Just run the code with my changes.

## Author's checklist

~This PR depends on #1969.~ already merged.

~Note: I didn't do extensive testing *yet*, this is just a PoC. Nothing broke catastrophically on my side though so that's good. Suggestions and improvements are absolutely welcome.~ Now that I have access to checks, I fixed some things and they all seem to pass now.

PR #2027, opened 2022-12-17
Add "auto" as a language.

### LINKED ISSUE #2018: Users language selection should take precedence over autodetection
### requested feature

The core of SearXNG should provide the ability to distinguish between explicit language selection by the user and language suggestions from a plugin.

### background of the request

In PR-1969 we ..

- https://github.com/searxng/searxng/pull/1969

Autodetection of the language from the search term is a great feature often requested by our users.  It should become active by default but it is not enabled by default since it has one drawback --> **The autodetection takes precedence over user's selection and this can make SearXNG unusable for the user.**

----

copied from https://github.com/searxng/searxng/pull/1969#issuecomment-1345294431

SearXNG's locale of a query comes from (*highest wins*):

1. The ``Accept-Language`` header from user's HTTP client.
2. The user select a locale in the preferences.
3. The user select a locale from the menu in the query form (e.g. ``:zh-TW``)
4. Autodetection plugin is activated in the preferences and the locale (only the language code / none region code) comes from the fastText's language detection.

Conclusion: There is a conflict between the language selected by the user and the language from language detection of this plugin.  For example, the user explicitly selects the German locale via the search syntax to search for a term that is identified as an English term. 

Try ``:de-DE thermomix``, for example ..

In this screenshot the user selects the `de-DE` locale but the autodetect plugin has switched the language to `en` because the name of the product the user looks for is a English word

grafik

dalf:
## What does this PR do?

Add a language choice "auto":

image

When selected, the choice remains "auto", however the result page displays which language was detected:

image

The user can use `:auto` as a language. In this screenshot, when the user sends the query, the language is going to detected (the screenshot is before sending the query, after the language is going to be `auto (en)`):
image

See #2018  : in the master branch, in the preferences, the user can select search language and at the same enable autodetection. If autodetection is enabled, the language selection should be disabled, however plugin does not allow to do that. This PR makes language auto detection a first class citizen.

## Related issues

Close #2018


searx/utils.py
```diff
@@ -638,7 +638,28 @@ def _get_fasttext_model() -> "fasttext.FastText._FastText":
...
+    no oc or os pa pam pfl pl pms pnb ps pt qu rm ro ru rue sa sah sc scn sco sd
+    sh si sk sl so sq sr su sv sw ta te tg th tk tl tr tt tyv ug uk ur uz vec vep
+    vi vls vo wa war wuu xal xmf yi yo yue zh
+
+    .. _fastText: https://fasttext.cc/
+    .. _python fasttext: https://pypi.org/project/fasttext/
+    .. _language identification model: https://fasttext.cc/docs/en/language-identification.html
+    .. _Bag of Tricks for Efficient Text Classification: https://arxiv.org/abs/1607.01759
+    .. _`FastText.zip: Compressing text classification models`: https://arxiv.org/abs/1612.03651
+    """
```
return42: This is not the function description .. I miss a description of function's arguments .. for what e.g. is `min_probability`?

AFAIK: `min_probability` is the same as treshold / in ed901ab18 I replaced `lang[1][0] >= 0.3` by `threshold=0.3`

In your review of https://github.com/searxng/searxng/pull/2019 you added `min_probability` back, but it was not clear to me why .. did I have overseen or missed something?

searx/search/__init__.py
```diff
@@ -37,6 +39,38 @@ def initialize(settings_engines=None, enable_checker=False, check_network=False,
...
+    code that is used by an engine.
+    b. SearXNG and most of the engines do not support all the languages from
+    language model and there might be also a discrepancy in the ISO-639-3 and
+    ISO-639-2 handling (:py:obj:`searx.locales.get_engine_locale`).  Further
+    more, in SearXNG the locales like ``zh-TH`` (``zh-CN``) are mapped to
+    ``zh_Hant`` (``zh_Hans``).
+    Conclusion: This function  does only auto-detect the languages a user can select
+    in the language menu (:py:obj:`supported_langs`).
+    .. hint::
+    """
```
return42: This function description confuses me .. its partial C&P of the plugin description but the markup is broken and the content is different to what the function does.

By example: *is harmonized with the SearXNG's language (locale) model.* --> IMO it isn't .. in the plugin we had 

```python
    if lang in supported_langs:
        search.search_query.lang = lang
        try:
            search.search_query.locale = babel.Locale.parse(lang)
        except babel.core.UnknownLocaleError:
            pass
```

and now we have no longer `if lang in supported_langs`.

HINT: if a language detected by fasttext is not in the `supported_langs` then the probability that this language is not supported by an engine is very high, all engines that do not support this language fall back to their default and this is mostly `en`.

searx/plugins/autodetect_search_language.py
```diff
@@ -1,97 +0,0 @@
-# SPDX-License-Identifier: AGPL-3.0-or-later
-# lint: pylint
-"""Plugin to detect the search language from the search query.
```
return42, line 3: The autodoc file of this source can also be deleted https://github.com/searxng/searxng/blob/master/docs/src/searx.plugins.autodetect_search_language.rst

searx/search/__init__.py
dalf: Updated: 
* the signature of `searx.utils.detect_language` is now:
```python
def detect_language(text: str, threshold: float = 0.3, only_search_languages: bool = False) -> Optional[str]:
```
* `searx/search/__init__.py` sets `only_search_languages` to `True`.

searx/utils.py
dalf: > In your review of https://github.com/searxng/searxng/pull/2019 you added min_probability back, but it was not clear to me why .. did I have overseen or missed something?

You are right, `min_probability` there is no need. I removed it.

The docstring is updated too.

Note: simplified and traditional Chinese are both detected as `zh`.

return42 (approved):

searx/search/__init__.py
```diff
@@ -37,18 +40,57 @@ def initialize(settings_engines=None, enable_checker=False, check_network=False,
...
+      The default value is "all" when no language is detected.
+    * `search_query.locale` is updated accordingly
+
+    Use :py:obj:`searx.utils.detect_language` with `only_search_languages=True` to keep
+    only languages supported by the engines.
+    """
+    if search_query.lang != 'auto':
+        return
+
+    detected_lang = detect_language(search_query.query, threshold=0.0, only_search_languages=True)
```
return42, line 57: @dalf why do we set the threshold to 0.0? .. the default is (was) 0.3 (with 0.0 we get 'en' for a single letter).

PR #2074, opened 2023-01-04
Correct a small mistake

julianfairfax:
## What does this PR do?

This corrects a small mistake in the docs

It's not, but there's no reason not to make the change.

I don't think you have to?

## Author's checklist

No notes

## Related issues

No issues for this are open as far as I know.

### DIFF
--- a/docs/src/index.rst
+++ b/docs/src/index.rst
@@ -2,7 +2,7 @@
 Source-Code
 ===========
 
-This is a partial documentation of our source code. We are not aiming to document
+This is a partial documentation of our source code.  We are not aiming to document
 every item from the source code, but we will add documentation when requested.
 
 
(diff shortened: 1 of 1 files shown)


docs/src/index.rst
```diff
@@ -2,7 +2,7 @@
 Source-Code
 ===========
 
-This is a partial documentation of our source code.  We are not aim to document
+This is a partial documentation of our source code. We are not aiming to document
```
return42: The double space after the dot of sentence end was intended. Typographic tools use the two spaces after the end of a sentence to distinguish from the period of an abbreviation.

return42 (changes requested):
Thanks for your patch / I have a nit to remark

docs/src/index.rst
julianfairfax: Oh oops, that's my mistake. Fixed.

return42 (approved):
Thanks!

PR #2084, opened 2023-01-08
Bugfix typos in searxng.sh

Milad-Laly:
## What does this PR do?

This PR corrects text typos spotted in utils/searxng.sh

Correcting text typo in Searxng.sh

### DIFF
--- a/utils/searxng.sh
+++ b/utils/searxng.sh
@@ -395,7 +395,7 @@ searxng.remove.all() {
     wait_key
 
     if service_is_available "${SEARXNG_URL}"; then
-        MSG="** Don't forgett to remove your public site! (${SEARXNG_URL}) **" wait_key 10
+        MSG="** Don't forget to remove your public site! (${SEARXNG_URL}) **" wait_key 10
     fi
 }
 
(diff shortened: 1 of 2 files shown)


utils/searxng.sh
```diff
@@ -688,7 +688,7 @@ To install uWSGI use::
         rst_para "\
 The configured redis DB is not available: If your server is public to the
 internet, you should setup a bot protection to block excessively bot queries.
-Bot protection requires a redis DB.  About bot protection visit the official
+Bot protection requires a redis DB. About bot protection visit the official
```
return42: The double space after the dot of sentence end was intended. Typographic tools use the two spaces after the end of a sentence to distinguish from the period of an abbreviation.

return42 (changes requested):

utils/searxng.sh
Milad-Laly: Ah Okay, I'll change it now thank you

return42 (approved):
Thanks / FYI: I squashed the commits into one.

PR #2097, opened 2023-01-11
Adjust uWSGI workers and threads from env

### LINKED ISSUE #447: docker & installation scripts: uwsgi.ini : avoid out of memory.
Currently the number of uwsgi worker doesn't take into account the memory limit : each worker takes about 256MB of memory. If the memory is low ( < 512MB ), the docker image & the installation scripts start more than one worker. Consequence: there is either OOM or swap usage. From the user point of view, the application is really slow.

The default configuration must avoid OOM.

The uwsgi worker count can be defined using:
```python
import psutil
import multiprocessing

# 256MB without the checker per worker, 320MB with the checker 
memory_total = psutil.virtual_memory().total
worker_count_from_memory = int(memory_total / 320)
worker_count_from_core = multiprocessing.cpu_count()
worker_count = math.min(worker_count_from_memory, worker_count_from_core)
```
(or the bash version of this code).

?

**Additional context**

### LINKED ISSUE #2096: Override uWSGI worker count from env
The [current worker count setting is `%k`](https://github.com/searxng/searxng/blob/master/dockerfiles/uwsgi.ini), which is the amount of cores detected, but since uWSGI workers consume about 150-200M of RAM, a system with lots of cores can explode in RAM consumption. 

Please make worker count overridable through an environment variable: 
- https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#environment-variables
- https://stackoverflow.com/questions/53283263/is-it-possible-to-override-uwsgi-ini-file-with-environment-variables
- https://stackoverflow.com/questions/47091122/uwsgi-set-configuration-depending-on-environment-variable

I understand that I can supply a whole `uwsg.ini` file but I don't want to do that, I have no problem with the other default settings. Maybe `workers` and `threads` could be configurable both but `workers` is a must. I don't want to use ConfigMap in k8s, a simple env setting should suffice.

**Additional context**
There is already a logic to override searxng `settings.yml` values with env vars, maybe this can be applied too, should the `patch_uwsgi_settings` function contain more logic as `patch_searxng_settings` does: https://github.com/searxng/searxng/blob/master/dockerfiles/docker-entrypoint.sh

Memory consumption on a 6 core k8s node run from the container shell (`rss` is the non-paged pyhsical memory, resident set size):

```
/usr/local/searxng # ps -o pid,user,vsz,rss,comm,args
PID   USER     VSZ  RSS  COMMAND          COMMAND
    1 root      816    4 tini             /sbin/tini -- /usr/local/searxng/dockerfiles/docker-entrypoint.sh
    7 searxng   28m  14m uwsgi            {uwsgi} uWSGI master
   14 searxng  200m 139m Thread-4 (initi  {Thread-4 (initi} uWSGI worker 1
   16 searxng  209m 149m 4103b35d-3dae-4  {4103b35d-3dae-4} uWSGI worker 2
   22 searxng  205m 146m fb0c5ef1-207c-4  {fb0c5ef1-207c-4} uWSGI worker 3
   32 searxng  215m 155m Thread-4 (initi  {Thread-4 (initi} uWSGI worker 4
   36 searxng  200m 140m Thread-4 (initi  {Thread-4 (initi} uWSGI worker 5
   42 searxng  200m 139m Thread-4 (initi  {Thread-4 (initi} uWSGI worker 6
  139 root     1692 1124 sh               sh
  156 root     1616 1012 ps               ps -o pid,user,vsz,rss,comm,args
```

image

immanuelfodor:
## What does this PR do?

Makes the uWSGI worker count (and thread count) configurable from an env variable. According to https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html#environment-variables even more variables could be added to this later on if needed.

The [current worker count setting is %k](https://github.com/searxng/searxng/blob/master/dockerfiles/uwsgi.ini), which is the amount of cores detected, but since uWSGI workers consume about 150-200M of RAM, a system with lots of cores can explode in RAM consumption.

Rebuild the Docker image and set any of the `UWSGI_WORKERS` or `UWSGI_THREADS` env variables. See the replacement result at `/etc/searxng/uwsgi.ini`. 

Or you can run the sed manually from the container's shell. No matter if you define 0, 1 or both variables, the sed always work, it keeps the defaults when no values are assigned to the env vars.

## Author's checklist

I tested the sed in a running container.

## Related issues

Closes https://github.com/searxng/searxng/issues/2096
Closes https://github.com/searxng/searxng/issues/447

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -61,6 +61,12 @@ echo "SearXNG version ${SEARXNG_VERSION}"
 # helpers to update the configuration files
 patch_uwsgi_settings() {
     CONF="$1"
+
+    # update uwsg.ini
+    sed -i \
+        -e "s|workers = .*|workers = ${UWSGI_WORKERS:-%k}|g" \
+        -e "s|threads = .*|threads = ${UWSGI_THREADS:-4}|g" \
+        "${CONF}"
 }
 
 patch_searxng_settings() {


dalf (approved):
Thank you!

Note: once uwsgi.ini exists, the file is not updated.

immanuelfodor:
Thanks! Since it's not a mount point in my setup, it's no problem, recreated after a new pod is scheduled

PR #2109, opened 2023-01-15
Standalone words highlighting for query result in non-CJK characters

### LINKED ISSUE #2095: Search query got emphasized even in the middle of another, unrelated word
**Version of SearXNG**

2023.01.09-afd71a6c

Installed using docker with clone, docker build and run.

**What happened?**

Query letters got emphasized even in the middle of another unrelated word.

**How To Reproduce**

Do some search using alphabetic, English words with all language flag, such as `the :all`, `java :all`, `master :all`. 

**Expected behavior**

Emphasization should occur on queries found in standalone words only, such as `java` query only emphasize '**java**' instead of '**java**script', or `master` only in '**Master**' or '**master**' instead of 'grand**master**'.

**Screenshots & Logs**

| screenshot-localhost_8080-2023 01 10-21_55_30 |
| --- |
| In the word 'o**the**rwise' |
 
| screenshot-localhost_8080-2023 01 10-21_55_08 |
| --- |
| In the word '**The**saurus' and '**the**ir' |

| screenshot-localhost_8080-2023 01 10-22_02_31 |
| --- |
| In the word '**master**ful', '**master**s', 'grand**master**' |

**Additional context**

Likely happened because the regex being used does not isolate the query for occurrences in standalone words and instead it looks for all occurrences in the whole text without requiring the presence of spaces before or after it. This regex actually works well for the emphasization of queries in Chinese characters, for example:

| screenshot-localhost_8080-2023 01 10-22_09_43 |
| --- |
| Query used: ’村上春樹’ |

ahmad-alkadri:
## What does this PR do?

Modifying the content highlight to become only full or standalone words for query results in non-CJK (Chinese, Japanese, and Korean) characters. 

During searches using non-CJK languages/characters, such as English, query letters got emphasized even in the middle of another unrelated word on the search page. Examples can be found when doing some searches using alphabetic, English words with all language flag, such as `the :all`, `java :all`, `master :all`. The expected result should be that only standalone words are highlighted.

For example, `master` in `master of science` should be highlighted, while  `master` in `grandmaster` shouldn't.

By modifying the `highlight_content()` function in `webutils.py`, and adding a function that will recognize if the query contains CJK characters or not, the query highlight will still work for words enclosed by other characters for the CJK letters, whereas for the other languages, it will only be highlighted if the words aren't enclosed by other characters.

Some screenshots of the results:

| screenshot-127 0 0 1_8888-2023 01 15-21_19_06 |
| --- |
| query: `the :all`. Notice that the `the` in `otherwise` is no longer highlighted |

| screenshot-127 0 0 1_8888-2023 01 15-21_21_40 |
| --- |
| query: `master :all`. Notice the `master` instances in the `MasterPlus+` content aren't highlighted because they're not standalone words |

| screenshot-127 0 0 1_8888-2023 01 15-21_24_50 |
| --- |
| query: `村上春樹 books`. Notice the query word in kanji characters are highlighted even when they are enclosed by other letters, while the `books` is not. |

Pull the branch with the latest commit of this PR locally, then run `make run`.

Alternatively, build a docker image locally and run it:

```shell
docker build -t $IMGNAME -f Dockerfile .
docker run -p $PORT:$PORT $IMGNAME:latest
```

Then, do some searches using queries such as `the :all`, `java :all`, and `master :all`.

## Author's checklist

+ Because of the change of the `highlight_content()` function, the test codes for the `test_highlight_content` (done during `make test`) were also modified

## Related issues

Closes #2095

### DIFF
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -124,13 +124,14 @@ def contains_cjko(s: str) -> bool:
     Returns:
         bool: True if the input s contains the characters and False otherwise.
     """
-    unicode_ranges = ('\u4e00-\u9fff' # Chinese characters
-                      '\u3040-\u309f' # Japanese hiragana
-                      '\u30a0-\u30ff' # Japanese katakana
-                      '\u4e00-\u9faf' # Japanese kanji
-                      '\uac00-\ud7af' # Korean hangul syllables
-                      '\u1100-\u11ff' # Korean hangul jamo
-                      )
+    unicode_ranges = (
+        '\u4e00-\u9fff'  # Chinese characters
+        '\u3040-\u309f'  # Japanese hiragana
+        '\u30a0-\u30ff'  # Japanese katakana
+        '\u4e00-\u9faf'  # Japanese kanji
+        '\uac00-\ud7af'  # Korean hangul syllables
+        '\u1100-\u11ff'  # Korean hangul jamo
+    )
     return bool(re.search(fr'[{unicode_ranges}]', s))
 
 
@@ -168,7 +169,9 @@ def highlight_content(content, query):
     querysplit = query.split()
     queries = []
     for qs in querysplit:
-        queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
+        qs = qs.replace("'", "").replace('"', '').replace(" ", "")
+        if len(qs) > 0:
+            queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
     if len(queries) > 0:
         for q in set(queries):
             content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{q}</span>', content)
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -113,31 +113,65 @@ def prettify_url(url, max_length=74):
         return url
 
 
+def contains_cjko(s: str) -> bool:
+    """This function check whether or not a string contains Chinese, Japanese,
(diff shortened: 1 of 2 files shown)


ahmad-alkadri:
Hi @dalf, this is the PR that should resolve #2095, which is the content highlight issue. 

Please let me know what you think; I'm open to follow-up discussions on this subject. Thanks in advance.

dalf (approved):
It works well.  I'm going to merge the PR.

I realize that some code can be refactored:
* `highlight_content` splits the query, but [RawTextQuery](https://github.com/searxng/searxng/blob/13b0c251c45c3d14700723b25b601be56178e8df/searx/query.py#L238) already does that.
* so `highlight_content` can use [raw_text_query.user_query_parts](https://github.com/searxng/searxng/blob/13b0c251c45c3d14700723b25b601be56178e8df/searx/webapp.py#L680) directly.
* in addition the query is parsed multiple times to create the regex. Another way to do that : 
  * create an `Highlighter` class
  * the constructor creates the regex in the constructor from a RawTextQuery instance
  * a method `highlight(content)`
  * `regex_highlight_cjk` is a method of this class

I'm may change my mind about this refactor idea.

Anyway, thank you for your PR!

ahmad-alkadri:
Thanks for reviewing this @dalf ! I just had a quick look at the class that you mentioned (`RawTextQuery`), and I can agree with you on this refactoring idea. I think we can put the `Highlighter` class on the `webutils.py`, passing the split query from `RawTextQuery`... Definitely a discussion for another enhancement Issue.

Happy to work on this!

PR #2122, opened 2023-01-22
SearXNG version: YYYY.MM.DD without leading zero

dalf:
## What does this PR do?

See https://github.com/searxng/searxng/discussions/2120 : leading zero in the version number doesn't comply with PEP440.

Avoid bugs in the update procedure.

* `make install`

## Related issues

Related to https://github.com/searxng/searxng/discussions/2120

### DIFF
--- a/searx/version.py
+++ b/searx/version.py
@@ -59,7 +59,7 @@ def get_git_url_and_branch():
 
 
 def get_git_version():
-    git_commit_date_hash = subprocess_run(r"git show -s --date='format:%Y.%m.%d' --format='%cd+%h'")
+    git_commit_date_hash = subprocess_run(r"git show -s --date='format:%-Y.%-m.%-d' --format='%cd+%h'")
     tag_version = git_version = git_commit_date_hash
 
     # add "+dirty" suffix if there are uncommited changes except searx/settings.yml


searx/version.py
```diff
@@ -59,7 +59,7 @@ def get_git_url_and_branch():
 
 
 def get_git_version():
-    git_commit_date_hash = subprocess_run(r"git show -s --date='format:%Y.%m.%d' --format='%cd+%h'")
+    git_commit_date_hash = subprocess_run(r"git show -s --date='format:%-Y.%-m.%-d' --format='%cd+%h'")
```
return42, line 62: `YYYY-MM-DD` is not a valid **number** for a (major) release in sense of semantic versioning --> `YYYY-MM-DD` is  a string --> only **numbers** (integers) can be selected by version specifiers

- https://peps.python.org/pep-0440/#version-specifiers
- https://peps.python.org/pep-0440/#public-version-identifiers

dalf: The `-` removes the leading zero

return42: Oops sorry. :facepalm:

return42 (approved):

PR #1967, opened 2022-11-22
settings.yml: add search.suspended_times

tiekoetter:
## What does this PR do?

Make suspended_time changeable in settings.yml
Allow different values to be set for different exceptions.

## Why is this change important?

See above

## How to test this PR locally?

Check if changes in `settings.yml` are reflected in the defined keys.

```
SearxEngineAccessDenied_suspended_time
SearxEngineCaptcha_suspended_time
SearxEngineTooManyRequests_suspended_time
cf_SearxEngineCaptcha_suspended_time
cf_SearxEngineAccessDenied_suspended_time
recaptcha_SearxEngineCaptcha_suspended_time
```

The effects of these changes can only be tested if an engine already triggers one (or more) of these exceptions regularly.
Setting a key to 0 should "disable" the specific exception from triggering a suspension of the engine.

## Related issues

Closes #1966

PR #2132, opened 2023-01-28
search.suspended_time settings: bug fixes

dalf:
## What does this PR do?

* fix type in settings.yml: replace suspend_times by suspended_times
* always use delay defined in settings.yml:
  * HTTP status 402 and 403: read the value from settings.yml instead of using the hard coded value of 1 day.
  * startpage engine: CAPTCHA suspend the engine for one day instead of one week

see https://github.com/searxng/searxng/pull/1967#issuecomment-1407154163

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -110,6 +110,13 @@ Global Settings
      default_lang: ""
      ban_time_on_fail: 5
      max_ban_time_on_fail: 120
+     suspended_times:
+       SearxEngineAccessDenied: 86400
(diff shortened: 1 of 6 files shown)


return42 (approved):
I placed three commits on top, mainly to document exception stuff / please check and drop what is not needed.

dalf:
Thank you.
I prefer to keep the values in settings.yml without comments: usually the values are declared in settings.yml even if there are default values in `settings_defaults.py`.
(even if there are some inconsistencies between settings.yml and settings_defaults.py).

PR #2081, opened 2023-01-06
Add HTTP Referer header to DuckDuckGo requests

rudism:
## What does this PR do?

Adds an HTTP Referer header to DuckDuckGo requests

## Why is this change important?

DuckDuckGo started rejecting requests without this header.

## How to test this PR locally?

Enable DuckDuckGo and do a search, you should see no "access denied" error message, DDG result should be showing up, and you should be able to page through results.

## Author's checklist

N/A

## Related issues

Closes #2080

PR #2161, opened 2023-02-03
[fix-2146] set different HTTP Referer header to DuckDuckGo requests

### LINKED ISSUE #2146: Bug: duckduckgo engine
Branch: master
Version: 2023.1.29+9d102fb0

Running docker image searxng/searxng, image ID 5cac8b4bbf70
**What happened?**
DDG still returns `Error! Engines cannot retrieve results: duckduckgo ( Suspended: access denied )` even though the fix from #2081 is definitely applied (I checked by running `docker exec -it searxng_searxng_1 sh` and then in the container:

```
/usr/local/searxng # cat searx/engines/duckduckgo.py | grep Referer
    params['headers']['Referer'] = 'https://lite.duckduckgo.com/'
```

**How To Reproduce**
Run a search with DDG enabled

**Expected behavior**
No error returned by DDG engine

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Percentage: 5
  * File name: `searx/search/processors/online.py:113`

return42:
For what ever reasons, ddg-lite don't like the Referer

    https://lite.duckduckgo.com/

In an interactive session in the WEB browser  the Referer has exactly this value, but ddg-lite don't like this value when the request is build up by SearXNG.  The new value is:

    https://google.com/

What fakes a user comes from a google link.

Related: https://github.com/searxng/searxng/pull/2081

Closes: https://github.com/searxng/searxng/issues/2146

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -73,7 +73,7 @@ def request(query, params):
     # link again and again ..
 
     params['headers']['Content-Type'] = 'application/x-www-form-urlencoded'
-    params['headers']['Referer'] = 'https://lite.duckduckgo.com/'
+    params['headers']['Referer'] = 'https://google.com/'
 
     # initial page does not have an offset
     if params['pageno'] == 2:


ahmad-alkadri (commented):
Tested in my local instance with a lot of queries; DDG works well and so does the other engines.

Some examples:

screenshot-127 0 0 1_8888-2023 02 03-09_34_34

screenshot-127 0 0 1_8888-2023 02 03-09_34_47

Really surprising that they'd be blocking based on the Referer of their own website to be honest.

return42:
> Really surprising that they'd be blocking based on the Referer of their own website to be honest.

Like it is often in hacking .. it was just a typo that brought it to my attention.  But I could not explain myself why this is so :smiley: 

> DDG works well and so does the other engines.

Other engines are not related to this patch .. if they work for you in your test this doesn't say anything about a daily usage on a public instance .. e.g. the startpage CAPTCHA will come back at some point.

ahmad-alkadri:
> Like it is often in hacking ..

Somebody's probably good at it at DDG then 👀
 
> Other engines are not related to this patch .. if they work for you in your test this doesn't say anything about a daily usage on a public instance .. e.g. the startpage CAPTCHA will come back at some point.

You're right and I admit I should've known this but it was my OCD+trauma talking tbh 😂

Thanks a lot for this patch!

PR #2178, opened 2023-02-10
[fix] engine brave: fix XPath of title element

### LINKED ISSUE #2020: Bug: Brave has mismatch between links and descriptions due to advertisement
Branch: master
Version: 2022.12.12-966e9c3c

script
**What happened?**
Results from Brave have mismatches between their URLs and their titles/descriptions

**How To Reproduce**
1. `!brave hatch act enforcement`
2. Look at results

**Expected behavior**
Results match titles/descriptions

**Screenshots & Logs**
chrome_ClS315bTV8

Note that the first result is an ad for a crypto wallet but the URL is to `osc.gov`. All the real results after are off by one.

return42:
Brave has mismatch between links and descriptions due to advertisement:

Suggested by: @allendema & @AlyoshaVasilieva
Closes: https://github.com/searxng/searxng/issues/2020

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1849,7 +1849,7 @@ engines:
     time_range_url: "&tf={time_range_val}"
     search_url: https://search.brave.com/search?q={query}&offset={pageno}&spellcheck=1{time_range}
     url_xpath: //a[@class="result-header"]/@href
-    title_xpath: //span[@class="snippet-title"]
+    title_xpath: //*[contains(@class,"snippet fdb")]/*[@class="result-header"]/*[@class="snippet-title"]
     content_xpath: //p[1][@class="snippet-description"]
     suggestion_xpath: //div[@class="text-gray h6"]/a
     time_range_map:


AlyoshaVasilieva:
Bad news, PR doesn't fix it. I think they've changed format? (think Brave blocked me so I switched engines)

Tested by removing custom title_xpath from settings, `!brave kraken`

Official:
krakenoff

This PR (master is also broken):
krakenon

It's grabbing the descriptions from every result (including the ad and sub-items), titles from main results (not including the ad or sub-items), and every URL (including the sub-items, not including the ad). I think I got that explanation right.

Pasted the HTML for the official site search for "kraken" (after redacting geolocation): https://bin.snopyta.org/?f2eb3063c9dded1a=#2dCnsj91HMo2nuKBAWMJcz4rVAggoQTEhU8gWQvrtpDn

I'm trying to come up with xpaths for title, content, url, but I haven't worked with them before. In very brief testing, these almost work:

```yaml
title_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]/*[@class="result-header"]/*[@class="snippet-title"]
content_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]/*[@class="snippet-content"]/p[1][@class="snippet-description"]
url_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]/a[@class="result-header"]/@href
```

These avoid reading the extra items (e.g. "NFT", "Fee Structure", "Prices", "Staking" items) because they don't seem very useful in this one instance that I've seen them.

Descriptions go wrong, though, with those xpaths.

I attempted to test:

```yaml
title_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]//*[@class="result-header"]/*[@class="snippet-title"]
content_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]//*[@class="snippet-description" or contains(@class,"description")]
url_xpath: //*[contains(@class,"snippet fdb") and not(@data-type="ad")]//a[@class="result-header"]/@href
```

... and hit "too many requests", so that's the end of my trial-and-error coding for today. Will return tomorrow, don't want to set another instance up when I should be going to sleep.

Searches I was testing were "kraken", "bitcoin", "NHL".

Cryptocurrency-related searches are the most reliable way to see ads from Brave to test ad avoidance. Some special results, such as video results, have slightly different formatting and the NHL search has at least one.

return42:
> Bad news, PR doesn't fix it. 

Thanks for testing :+1:

> Will return tomorrow, don't want to set another instance up when I should be going to sleep.

Can you send a PR when finished / thanks!

I close this PR.

AlyoshaVasilieva:
Well, with further research, I don't know if XPath can do this. Results that contain products on Amazon ('Airborne Toxic Event' search) have different description format with no mention of `description` in their classes. I *can* manage to grab them, but it contains an inline `<style>` that turns into a large block of text in SearXNG.

I might be wrong, of course, as I have essentially no experience with XPaths, but I don't think I can come up with a PR for this, because I *think* Brave needs to be a custom engine written in Python. (I might attempt this, but I've never written any Python.)

PR #2189, opened 2023-02-13
Spelling: Tor, SearXNG

comradekingu:
"Tor", and "SearXNG" as per https://www.torproject.org/ and https://docs.searxng.org/

## What does this PR do?

Sentence structure and proper nouns.

More professional and better explained.

Saw the translation, thought to do something about it.
Didn't.

### DIFF
--- a/searx/plugins/tor_check.py
+++ b/searx/plugins/tor_check.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
-"""A plugin to check if the ip address of the request is a TOR exit node if the
+"""A plugin to check if the ip address of the request is a Tor exit-node if the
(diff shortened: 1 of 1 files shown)


searx/plugins/tor_check.py
```diff
@@ -75,13 +75,13 @@ def post_search(request, search):
         if ip_address in node_list:
             search.result_container.answers["tor"] = {
                 "answer": gettext(
-                    "You are using TOR. Your IP address seems to be: {ip_address}.".format(ip_address=ip_address)
+                    "You are using Tor. It looks like you have this external IP address: {ip_address}.".format(ip_address=ip_address)
```
comradekingu: An assumption on my part. It isn't clear whether the original meaning is that of the actual or the exit-node IP as observed by the network.

return42 (approved):
LGTM / thanks!

PR #2205, opened 2023-02-19
[mod] video result: replace HTML <iframe> by <video> tag

return42:
video result: replace HTML `<iframe>` by `<video>` tag

The HTML `<video>` tag reduces the bandwidth massively compared to the `<iframe>`.

When a video is embedded in a `<iframe>`, control over that video transfers to the
`<iframe>`.  The `<iframe>` downloads the entire video.  With the `<video>` tag,
control stays in the main frame and only the passages that the user chooses are
downloaded.


mrpaulblack:
Just leaving this as a note here for future reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster

return42:
Hm, I set this PR to DRAFT .. without a source type (or a file ending) the video tag won't work

```html
<video controls>
  <source src="myVideo.webm" type="video/webm" />
```

Not all engines can scrap the source type  .. this needs much more work..

dalf:
It does not work with the Youtube engine because the URL of the `iframe_src` field contains a HTML document.

image

For MediathekView, `<video>` works, so a fix to the mention issue is:
* to keep `iframe_src` as it is now in master
* to add another field `video_src` which is render with a `<video>` HTML element.

```jinja2
{% if result.iframe_src or result.video_src %}
<div id="result-video-{{ index }}" class="embedded-video invisible">
{% if result.iframe_src %}
  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen></iframe>
{%- endif %}
{% if result.video_src %}
  <video controls>
    <source src="{{result.video_src}}">
  </video>
{%- endif %}
</div>
{%- endif %}
```

Perhaps a better alternative:
* rename `iframe_src` to `embedded_src`
* add a field `embedded_type` which can take two values: `iframe` (default) or `video`.

```jinja2
{% if result.embedded_src %}
<div id="result-video-{{ index }}" class="embedded-video invisible">
  {% if result.embedded_type is None or  result.embedded_type  == 'iframe' %}
  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen></iframe>
  {% elif if result.embedded_type == 'video' %}
  <video controls>
    <source src="{{result.video_src}}">
  </video>
  {%- endif %}
</div>
{%- endif %}
```

PR #2217, opened 2023-02-24
fix: storage unit value

NV6bxpYSKb9h9O8uDIm6hw86EDg7HnbcQXOovhw:
## What does this PR do?

- add missing storage unit value
- Gibibyte and Gigabyte were confused

- fix nyaa engine filesize

### searxng
searxng

### nyaa
nyaa

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -39,12 +39,14 @@ _ECMA_UNESCAPE4_RE = re.compile(r'%u([0-9a-fA-F]{4})', re.UNICODE)
 _ECMA_UNESCAPE2_RE = re.compile(r'%([0-9a-fA-F]{2})', re.UNICODE)
 
 _STORAGE_UNIT_VALUE: Dict[str, int] = {
-    'TB': 1024 * 1024 * 1024 * 1024,
-    'GB': 1024 * 1024 * 1024,
-    'MB': 1024 * 1024,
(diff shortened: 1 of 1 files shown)


searx/utils.py
```diff
@@ -39,12 +39,14 @@
...
-    'TB': 1024 * 1024 * 1024 * 1024,
-    'GB': 1024 * 1024 * 1024,
-    'MB': 1024 * 1024,
-    'TiB': 1000 * 1000 * 1000 * 1000,
-    'MiB': 1000 * 1000,
-    'KiB': 1000,
+    'TB': 1000 * 1000 * 1000 * 1000,
+    'GB': 1000 * 1000 * 1000,
+    'MB': 1000 * 1000,
+    'kB': 1000,
```
dalf, line 45: `k` in lower case: is it intended ?

return42: On base 10^n (metric/decimal) it is ``kB`` on base 2^n (binary) it is ``KB``--> https://en.wikipedia.org/wiki/Kilobyte

Since we are talking about bytes I prefer a 2^10 (1024/binary) factor and the ``KB`` notation.

NV6bxpYSKb9h9O8uDIm6hw86EDg7HnbcQXOovhw:
I found that different torrent sites have different definitions for storage unit values, have no idea how to solve this.

Issue #2216, opened 2023-02-24
Bug: gentoo engine


ppfeufer:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2023.2.20+d669da81

**How did you install SearXNG?**

searxng-docker

**What happened?**

Searching and switching to the "IT" tab.

**How To Reproduce**

Searching and switching to the "IT" tab.

**Expected behavior**

No timeout warning.

**Screenshots & Logs**

![image](https://user-images.githubusercontent.com/2989985/221288567-349a9f39-4f94-4191-9035-df720ff25306.png)

**Additional context**

**Technical report**
![image](https://user-images.githubusercontent.com/2989985/221288686-9d91469c-d4bb-4c6d-bdbd-696cbc6b3ebc.png)

  * Error: httpx.TimeoutException
  * Percentage: 100
  * Parameters: `(None, None, None)`
  * File name: `searx/search/processors/online.py:114`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Solirs:
Can reproduce and confirm, ill try to look into it.

Solirs:
It seems like simply setting a nicer timeout fixes it all for me.
In settings.yml find:
  - name: gentoo
    engine: gentoo
    shortcut: ge

And append:
    timeout: 10.0
to it

I feel like this should be the default because of this bug so i may make a PR later

return42:
closed by #2220

PR #2220, opened 2023-02-28
Increase timeout for gentoo wiki engine

Solirs:
## What does this PR do?

This PR increases the timeout for the gentoo engine

See issue #2216 
This PR fixes the bug.

Run SearXNG.
Search for one of the terms that cause the issue (i.e "funtoo" did it for me)
Go to the "IT" tab
Check if the gentoo engine returns a timeout error.

## Related issues
#2216


dalf:
Thank you

PR #2237, opened 2023-03-05
[fix] Remove unexpected language restriction in Bing

9Ninety:
## What does this PR do?
Fixed unusable Bing result for user who using language "Default language (all)" with non English search keywords.
> The origin code add `language: EN` to keywords if language was set to `all`

The origin code:
- Make search without language restriction not possible
- No result, or add no sense result when keywords was not English and `default_lang` set to `all`

## Screenshot
**Origin:**
bing_origin

**With this PR:**
bing_fix

1. Use origin code from master branch
2. Enable Bing only
3. Set `default_lang` to `all` or manually switch to "Default language" in result page
4. Search by using any language other than English, CJK was preferred
3. Take a screenshot
6. Checkout code to this PR
7. Search again, compare result quality with screenshot

>Example keyword: 人と、地球の、明日のために, copied from Toshiba website

## Author's checklist
- [x] Tested with(out) `default_lang`
- [x] Tested with different language in dropdown

## Related issues

#941 (Only related to title)

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -55,12 +55,10 @@ def request(query, params):
     if offset == 1:
         search_string = inital_query
 
-    if params['language'] == 'all':
-        lang = 'EN'
-    else:
+    # Add language filter in query if specific language was given
+    if params['language'] != 'all':
         lang = match_language(params['language'], supported_languages, language_aliases)
-
-    query = 'language:{} {}'.format(lang.split('-')[0].upper(), query)
+        query = 'language:{} {}'.format(lang.split('-')[0].upper(), query)
 
     search_path = search_string.format(query=urlencode({'q': query}), offset=offset)
 


return42:
@9Ninety TLDR: language value 'all' is most often a bad choice:

The bing.com search needs a locale (or in SearXNG at least a language).  If you go bing.com  and do not choose a locale for the search results bing.com uses the `Accept-Language` header from your browser to select the locale for you.

When SearXNG sends a request to bing.com, SearXNG needs to set a locale and a language, otherwise SearXNG will not get good results or even any results:

https://github.com/searxng/searxng/blob/767f22eca143022f37d31818e4c04c2a76569fa0/searx/engines/bing.py#L122-L131

I recommend not to set any language in SearXNG, then SearXNG will use the default setting from your WEB browser as locale (SearXNG can also evaluate the Accept header).

For those user using in different languages (like I do, I search sometimes in DE, sometimes in EN) I recommend to use the language detection (`:auto`) / by example: `!bing :auto 人と、地球の、明日のために`

grafik


----

Additional informations: 

bing.com does not have an *"all locale"* 

https://github.com/searxng/searxng/blob/767f22eca143022f37d31818e4c04c2a76569fa0/searx/data/engine_traits.json#L54-L55

while www.bing.com/news has an *"all locale"*:

https://github.com/searxng/searxng/blob/767f22eca143022f37d31818e4c04c2a76569fa0/searx/data/engine_traits.json#L246-L247

Other search provider like google.com or DDG do have also a "all" support:

https://github.com/searxng/searxng/blob/767f22eca143022f37d31818e4c04c2a76569fa0/searx/data/engine_traits.json#L491-L492

https://github.com/searxng/searxng/blob/767f22eca143022f37d31818e4c04c2a76569fa0/searx/data/engine_traits.json#L983-L984

If we now take a look at all search engines, we have to realize that there are some that can handle a value like "all", but others cannot.  Those who cant handle 'all' will fallback to 'en' and mix English results into our result list ..

That is why I think using 'all' is not a good option in the common use cases.

It is a bit more complex, I hope you could understand my explanations.  With this background I come to the conclusion that we cannot implement a reasonable handling for 'all' for bing and it will remain with the fallback 'en'.

I will therefore close this PR ... But maybe I missed a possibility, then we can open the PR again if necessary.

9Ninety:
You can **keep it as closed** since the latest code improved how language was handled, and "all" now works on most case.

Auto detect doesn't work well when the search keyword is English or mixed languages or I need results in two or three different languages. That's a complex problem and I have no idea about how to improve it at this time...

**Example**, English has been detect as French, description was missing, much worse when the keyword is CJK character
Snipaste2

**Bing with Japan IP address, browser language set to English**
Snipaste3

Issue #2256, opened 2023-03-15
Search leading to Internal Error


keithdanielll:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
<!-- If you are running on master branch using git execute this command
in order to fetch the latest commit ID:
```
2023.3.2+abd2d860

**How To Reproduce**
<!-- How can we reproduce this issue? (as minimally and as precisely as possible) -->
Take any of the top rated public instances and insert this search:
`WARNING: /plugins/Live/VisitorProfile.php(24): Deprecated - Creation of dynamic property Piwik\Plugins\Live\VisitorProfile::$idSite is deprecated - Matomo 4.13.3 `

... will result in "internal server error"

**Expected behavior**
<!-- A clear and concise description of what you expected to happen. -->
Some search result, or masking of "forbidden" characters

keithdanielll:
[Seems to be related](https://github.com/searxng/searxng/issues/2250)

Solirs:
I found a way to patch this by running re.escape() on the query inside of the problematic function call.
`content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{re.escape(q)}</span>', content)`
It fixes it for both your issue and the other one so i'll PR it.

Issue #2250, opened 2023-03-12
Search for Laminas\XmlRpc\Client


r0073rr0r:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

**How did you install SearXNG?**

**What happened?**

**How To Reproduce**
Try searching for:  Laminas\XmlRpc\Client

Search Laminas\XmlRpc\Client
**Expected behavior**

**Screenshots & Logs**

Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
**Additional context**

Solirs:
Tested and can confirm from latest commit
Full error:
```
Traceback (most recent call last):
  File "/usr/lib64/python3.10/sre_parse.py", line 1051, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\X'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 2551, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/Solirs/Documents/code/searxng/searx/flaskfix.py", line 71, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/werkzeug/middleware/proxy_fix.py", line 187, in __call__
    return self.app(environ, start_response)
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 2531, in wsgi_app
    response = self.handle_exception(e)
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 2528, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/Solirs/Documents/code/searxng/local/py3/lib64/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
[...]

PR #2257, opened 2023-03-15
re.escape() the query in highlight_content to prevent a server side error.

Solirs:
## What does this PR do?

This PR replaces
`content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{q}</span>', content)`  
by
`content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{re.escape(q)}</span>', content)`  
in the highlight_content method to prevent a bad escape error.

When searching for something containing what can look like an escape code (i.e Laminas\XmlRpc\Client) sre_parse can misinterpret it as a bad escape and throw a fatal error preventing the search.

make run
Search for "Laminas\XmlRpc\Client"
No error should be thrown and the search should go on normally.

## Author's checklist

N/A

## Related issues
Closes both
#2250 
and
#2256

### DIFF
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -174,7 +174,9 @@ def highlight_content(content, query):
             queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
     if len(queries) > 0:
         for q in set(queries):
-            content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{q}</span>', content)
+            content = re.sub(
+                regex_highlight_cjk(q), f'<span class="highlight">{q}</span>'.replace('\\', r'\\'), content
+            )
     return content
 
 


return42 (changes requested):
Hi @Solirs thanks a lot for your patch .. the doc of [re.escape](https://docs.python.org/3/library/re.html#re.escape) says:

> This function must not be used for the replacement string in [sub()](https://docs.python.org/3/library/re.html#re.sub) and [subn()](https://docs.python.org/3/library/re.html#re.subn), only backslashes should be escaped.

Here is what I would suggest to fix the issue ..

```diff
diff --git a/searx/webutils.py b/searx/webutils.py
index 7b9a8045c..00929bebe 100644
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -174,7 +174,10 @@ def highlight_content(content, query):
             queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
     if len(queries) > 0:
         for q in set(queries):
-            content = re.sub(regex_highlight_cjk(q), f'<span class="highlight">{q}</span>', content)
+            content = re.sub(
+                regex_highlight_cjk(q), f'<span class="highlight">{q}</span>'.replace('\\', r'\\'), content
+            )
     return content
```


To test you can also use ``!qwant ascii \x20``

Solirs:
Applied the patch and squashed everything, i think all is good.

Issue #2249, opened 2023-03-11
Public Usecase


bijsven:
**How did you install SearXNG?**
searxng-docker

**about**
Hi, is there a way to change the logo and disable editing the settings?

Solirs:
Both of those features are unsupported for now.
You can change the logo by replacing facivon.svg and favicon.png in searx/static/themes/simple/img (/usr/local/searxng/searx/static/themes/simple/img in docker). Make sure to comment the static-map line in your uwsgi.ini as searxng handles that already and it causes issues with this. You might also need to hard refresh your searxng window to see the change. I'll probably PR a way to change icons via settings.yml in the future.

For disabling editing the settings, i'm pretty sure there is no way to do that without changing the source code.

unixfox:
changing the logo is not supported

PR #2276, opened 2023-03-22
Add setting to change  favicon and logo

Solirs:
## What does this PR do?

This PR adds two settings in settings.yml, one to set SearXNG's logo (like the one that appears in page_with_header.html) and one to set its favicon. Those settings can either be a full path, a relative path (to the app root) or a URL to pull the image from.
It also adds a new endpoint (/logo) and rewrites /favicon.ico to respect those new settings.
Finally it rewires base.html and page_with_header.html to use those new endpoints.
It also adds a function in webapp.py used by both the new endpoints (get_favicon_or_logo).

It has been requested several times in the past without anything coming out of it (see related issues).
I personally think it's a good feature to make SearXNG more hackable/customisable, especially for public instances. Since the current solution of changing the image files is a pain and is not practical at all.
It's one more nice feature IMO.

1. Set logo and favicon in settings.yml to files of your choice
2. make run
3. Check /logo and /favicon.ico
4. Repeat with favicon and logo set to a URL to an image, a path relative to SearXNG's root, and an invalid path/url.
And check the behavior.

## Author's checklist

I am aware of the repeated code problem in  this PR and i know that the pulling image from a URL should be made a function shared by both get_favicon_or_logo and the image proxy. However i'm not sure whether this is in the scope of this PR because image_proxy would be modified.
Should i commit it here or make another pull request if this is merged to fix that?

## Related issues

- #2249 
- https://github.com/searxng/searxng/discussions/2032#discussioncomment-4467735

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -14,6 +14,7 @@ general:
   enable_metrics: true
 
 brand:
+  # Both favicon and logo are relative to SearXNG's root path unless they are absolute paths
   favicon: static/themes/simple/img/favicon.png
   logo: static/themes/simple/img/searxng.png
   new_issue_url: https://github.com/searxng/searxng/issues/new
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -14,6 +14,8 @@ general:
   enable_metrics: true
 
 brand:
+  favicon: static/favicon.ico
+  #logo: searx/static/themes/simple/img/searxng.png
   new_issue_url: https://github.com/searxng/searxng/issues/new
   docs_url: https://docs.searxng.org/
   public_instances: https://searx.space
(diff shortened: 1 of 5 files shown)


tessus:
This seems nice. Is there a reason why this hasn't been merged?

return42:
> Is there a reason why this hasn't been merged?

Lack of time to review / the PR brings some code that is questionable and I don't have time to elaborate alternatives --> custom logo & favicon do not have the first prio.

----

**update:** please read [this comment](https://github.com/searxng/searxng/pull/4228#issuecomment-2612141470)

tessus:
Thanks for the info.

return42:
Closing due to inactivity / see https://github.com/searxng/searxng/pull/2276#issuecomment-1556640900

PR #2302, opened 2023-03-31
[upd] httpx v0.21.2 --> v0.23.3 & httpx-socks v0.7.2 --> v0.7.5

return42:
[Dependabot alerts](https://github.com/searxng/searxng/security/dependabot) but did not update the httpx dependencies.

@dalf what do you think, can upgrade httpx packages?

I test this PR on my instance https://darmarit.org/searx/ .. so far I could not detect any issues.

Related: https://github.com/searxng/searxng/security/dependabot/4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,10 +7,10 @@ lxml==4.9.2
 pygments==2.14.0
 python-dateutil==2.8.2
 pyyaml==6.0
-httpx[http2]==0.21.2
+httpx[http2]==0.23.3
 Brotli==1.0.9
 uvloop==0.17.0
-httpx-socks[asyncio]==0.7.2
+httpx-socks[asyncio]==0.7.5
 setproctitle==1.3.2
 redis==4.5.4
 markdown-it-py==2.2.0


dalf:
Most probably, this code needs to be updated:
https://github.com/searxng/searxng/blob/27611470bfff6806f56b574f9b045d52ff6de95c/searx/network/client.py#L63-L126

Without diving into the httpx code, a simple check is to upgrade, start a with mitmproxy for example and configure SearXNG to use it. If the requests are sent through the proxy that's good. Note that socks4, socks5, http and https proxies need to be tested. And then other parameters need to be checked.

I was hoping the [version 0.24](https://github.com/encode/httpx/discussions/2534) to be released sooner. 

I will have a look at it tomorrow.

return42:
> I was hoping the [version 0.24](https://github.com/encode/httpx/discussions/2534) to be released sooner. 

From my side there is no need to hurry ... should we wait? --> https://github.com/encode/httpx/discussions/2534#discussioncomment-5445474

> Note that socks4, socks5, http and https proxies need to be tested. And then other parameters need to be checked. .. I will have a look at it tomorrow.

Thanks a lot, I would be overwhelmed with this for now / wouldn't know how to assemble a setup for a test :-o

requirements.txt
```diff
@@ -7,10 +7,10 @@ lxml==4.9.2
...
 pyyaml==6.0
-httpx[http2]==0.21.2
+httpx[http2]==0.23.3
 Brotli==1.0.9
 uvloop==0.17.0
-httpx-socks[asyncio]==0.7.2
+httpx-socks[asyncio]==0.7.5
 setproctitle==1.3.2
 redis==4.5.4
 markdown-it-py==2.2.0
```
return42, line 16: copy @dalf's comment from https://github.com/return42/searxng/commit/d86ec70877d24272c41f8c8e2ccde012791b9540#commitcomment-107513541 to here ..

> The important thing to test is the proxy configuration (http and socks). When I have tried with mitmproxy and `verify: ~/.mitmproxy/mitmproxy-ca-cert.cer` I had an error. It worked before.

dalf:
Another note: in new version httpx supports socks and http proxy but not https proxy.
For https proxy, httpx-socks is still required.

See:
* https://github.com/encode/httpx/issues/203
* https://github.com/encode/httpx/issues/1434#issuecomment-979193104
* https://github.com/encode/httpx/issues/2632

mrpaulblack:
Closing in favor of https://github.com/searxng/searxng/pull/2644; Closing.

PR #2304, opened 2023-04-01
[mod] limiter: minor improvements

return42:
## What does this PR do?

- requests without HTTP header ~`Connection` or missing~ `User-Agent` will be blocked by the limiter

- re_bot is related to `User-Agent` and has been renamed to block_user_agent

---- 

To trace limiter in production I use:

     sudo -H FORCE_SELECTION=Y ./utils/searxng.sh instance inspect | grep limiter

### DIFF
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -26,13 +26,17 @@ default_on = False
 preference_section = 'service'
 logger = logger.getChild('limiter')
 
-re_bot = re.compile(
+block_user_agent = re.compile(
     r'('
-    + r'[Cc][Uu][Rr][Ll]|[wW]get|Scrapy|splash|JavaFX|FeedFetcher|python-requests|Go-http-client|Java|Jakarta|okhttp'
+    + r'unknown'
+    + r'|[Cc][Uu][Rr][Ll]|[wW]get|Scrapy|splash|JavaFX|FeedFetcher|python-requests|Go-http-client|Java|Jakarta|okhttp'
     + r'|HttpClient|Jersey|Python|libwww-perl|Ruby|SynHttpClient|UniversalFeedParser|Googlebot|GoogleImageProxy'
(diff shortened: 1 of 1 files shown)


return42:
With this PR the limiter will block requests where HTTP header `Connection` or `User-Agent` is unset.  I tested this on my instance over a long period without an issue in my (very small) user base.

In order to get a more representative evaluation we are looking out for testers to test this patch in live operation on their SearXNG instances.

Related:  https://docs.searxng.org/admin/installation-nginx.html#nginx-s-searxng-site

unixfox:
Please add me as a reviewer.

searx/plugins/limiter.py
```diff
@@ -62,6 +67,10 @@ def is_accepted_request() -> bool:
             logger.debug("BLOCK %s: got Connection=close", x_forwarded_for)
             return False
 
+        if not request.headers.get('Connection'):
+            logger.debug("BLOCK %s: got Connection=", x_forwarded_for)
```
mrpaulblack: AFAIK the connection header is not used in http2 -> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection . IMO we should not filter for this header.

mrpaulblack (changes requested):

searx/plugins/limiter.py
unixfox: I'm also against this change, I don't think it's wise to block based on this missing HTTP header.

return42: Thanks for having an eye on .. I removed [`if not request.headers.get('Connection'):`](https://github.com/searxng/searxng/compare/553a01171ad48653cf20455cd7f41aa5bba5b662..66810ce71122183f52446cb110dc76e4cae9b7ba).

unixfox (approved):

mrpaulblack (approved):
Thanks; LGTM :+1:

Issue #2314, opened 2023-04-03
./manage : call to the delete file searxng_extra/update/update_languages.py


dalf:
https://github.com/searxng/searxng/blob/3ab85db5ddd12e1d17dab7d72ea8a7cb7ca961b5/manage#L403-L412

* The function name should be `data.traits()`
* The script to call is `searxng_extra/update/update_engine_traits.py`

Related to https://github.com/searxng/searxng/pull/2269

return42:
@dalf thanks for pointing out, this brings me to the next issue

- https://github.com/searxng/searxng/pull/2316

----

We can close this issue (#2314) when #2316 is merged.

return42:
Finally closed by https://github.com/searxng/searxng/pull/2316

PR #2269, opened 2023-03-20
Revision of the locale- and language- handling in SearXNG

return42:
## What does this PR do?

Drop-in replacement of the locale- and language- handling in SearXNG.

## Why is this change important?

From its roots, SearXNG supports only languages and the localizations (language & region) were partially built in.

The implementations so far are partly deficient and contain errors, such as mixing GUI language and localization in some of the search engines.

The trouble become obvious when trying to improve e.g. the support of languages and regions like zh, zh-Hant and zh-Hans based on a user selection of zh, zh-TW or zh-CN.  Or setting the locale parameter for a search engine to get optimal results, depending on the user selection of a language vs. a region.

This PR implements a unified boilerplate for locale support in all levels including a rework of all search engines that do support languages.

## How to test this PR locally?

Test all languages and locales in all engines.

## Related issues

I developed and tested this patch series in my instance the last six month.  We have already made several attempts to address this issue in the past. The experiences from these attempts have been incorporated into the development of this PR.

- https://github.com/searxng/searxng/pull/1667
- https://github.com/searxng/searxng/pull/1832
- https://github.com/searxng/searxng/issues/2011
- https://github.com/searxng/searxng/issues/1822
- https://github.com/searxng/searxng/issues/1769
- https://github.com/searxng/searxng/issues/1864
- https://github.com/searxng/searxng/issues/1965
- https://github.com/searxng/searxng/issues/1213 ?
- https://github.com/searxng/searxng/issues/1359 .. should also be fixed
- https://github.com/searxng/searxng/issues/2287 .. is fixed.
- https://github.com/searxng/searxng/issues/941 ? .. bing is hard to domesticate, has it improved?
- https://github.com/searxng/searxng/issues/1081 ? .. startpage is hard to domesticate, has it improved?
- https://github.com/searxng/searxng/issues/1959
- https://github.com/searxng/searxng/issues/1966
- https://github.com/searxng/searxng/issues/2141

PR #2316, opened 2023-04-05
[fix] searxng_extra/update/update_engine_descriptions.py 

return42:
Follow up of #2269

The script to update the descriptions of the engines does no longer work since  #2269 has been merged.

Related: https://github.com/searxng/searxng/issues/2314#issuecomment-1497129035

---

searx/engines/wikipedia.py

  The update_engine_descriptions.py needs a list of all wikipedias. The
  implementation from #2269 had only a reduced list:

  - https://meta.wikimedia.org/wiki/Wikipedia_article_depth
  - https://meta.wikimedia.org/wiki/List_of_Wikipedias

searxng_extra/update/update_engine_descriptions.py

  Before PR #2269 there was a match_language() function that did an
  approximation using various methods.  With PR #2269 there are only the types
  in the data model of the languages, which can be recognized by babel.  The
  approximation methods, which are needed (only here) in the determination of
  the descriptions, must be replaced by other methods.

-----------

DRAFT: the modified script in this PR does it work but the yield of descriptions is still significantly lower than it used to be at the level of commit 64fea2f9c (before #2269 was merged).  It still needs some rework


dalf:
The wikipedia engine seems to have different issues.

I've pushed a commit where update_engine_descriptions.py doesn't use the wikipedia engine. In a way, that makes sense since the Wikipedia article names are returned by a SPARQL query on wikidata. With a minor update the query can return the Wikipedia article URL. From that, the wikipedia engine can be bypass.

With this change, `searx/data/engine_descriptions.json` is updated without issues except the language `zh_Hant`. See https://github.com/searxng/searxng/issues/2330

return42:
FYI: my work is not yet finished .. I will come back to this PR the next days ...

PR #2323, opened 2023-04-06
Provide template (files.html) for results in file category

OliveiraHermogenes:
## What does this PR do?

It adds a template for file results (i.e. results from searching in the file category) in the simple theme. 

It is [a requirement](https://github.com/searxng/searxng/blob/64e221426a4c27caa030cfa89b099636851c400b/searx/engines/recoll.py#L71) for the [recoll engine](https://docs.searxng.org/admin/engines/recoll.html).

Just drop `files.html` into `searx/templates/simple/result_templates/` and restart SearXNG.

## Author's checklist

This was basically copied and adapted from the oscar theme without much thinking about how templates work in SearX or a complete understanding of the simple theme. Please, review carefully.

### DIFF
--- a/searx/templates/simple/result_templates/files.html
+++ b/searx/templates/simple/result_templates/files.html
@@ -1,0 +1,56 @@
+{% from 'simple/macros.html' import result_header, result_sub_header, result_sub_footer, result_footer, result_link with context %}
+{% from 'simple/icons.html' import icon_small %}
+
+{{- result_header(result, favicons, image_proxify) -}}
+{{- result_sub_header(result) -}}
+
+{%- if result.embedded -%}
+    <small> &bull; <a class="text-info btn-collapse collapsed cursor-pointer media-loader disabled_if_nojs" data-toggle="collapse" data-target="#result-media-{{ index }}" data-btn-text-collapsed="{{ _('show media') }}" data-btn-text-not-collapsed="{{ _('hide media') }}">
+    {%- if result.mtype == 'audio' %}{{ icon_small('musical-notes-outline') -}}
+    {%- elif result.mtype == 'video' %} {{ icon_small('play-outline') -}}
+    {%- endif %} {{ _('show media') }}</a></small>
+{%- endif -%}
+
+{%- if result.embedded -%}
+<div id="result-media-{{ index }}" class="collapse">
+   {{- result.embedded|safe -}}
+</div>
+{%- endif -%}
+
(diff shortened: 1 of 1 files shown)


return42:
Notes from my research:

- `file.html` is only used by the [`recoll.py` engine](https://github.com/searxng/searxng/blob/master/searx/engines/recoll.py)
- [`recoll.py` engine](https://github.com/searxng/searxng/blob/master/searx/engines/recoll.py) has been added in commit 4a36a3044
- commit 4a36a3044 added only a oscar template & theme / never added to the simple theme.

As far as I can tell, the UI for the recoll engine (a `file.html` result) still needs to be developed, unfortunately we can't use the existing template because it is tailored to the oscar theme and miss a simple theme CSS.

Not an easy task, nevertheless we have to do it to get recoll back into SearXNG.

OliveiraHermogenes:
OK. So I'll have a closer look into the CSS and try to come up with something more adequate to the simple theme. I'll try to provide some screenshots.

OliveiraHermogenes:
Hi,

With respect to the [comments](https://github.com/searxng/searxng/issues/2957#issuecomment-1779726033) by @virtadpt in #2957, the URL truncation behavior is inherited from [url_wrapper](https://github.com/searxng/searxng/blob/b0b2efdc8dbfe323ed3d2e10c309aacac315febf/searx/static/themes/simple/src/less/style.less#L233C3-L265C6). The same truncation strategy seems to be applied overall, where the hostname and tail path is prioritized over internal path directories. So, perhaps this is to be expected? I personally see no problems here.

Similarly, the padding in the metadata table seems to come from [searxng.min.css](https://github.com/searxng/searxng/blob/b0b2efdc8dbfe323ed3d2e10c309aacac315febf/searx/static/themes/simple/css/searxng.min.css#L1C54406).

I added some commits with minor corrections. I think it is working well. Here are some screenshots:

searx-sample-screen

Sample results page on Firefox 121.0

searx-sample-screen-video

Sample results page on Firefox 121.0 with video after clicking "show media" button.

return42:
@virtadpt If you have the opportunity, can you please check this PR? / thanks :+1:

virtadpt:
@return42 It looks pretty solid.  I tested it with all of my Recoll instances on Leandra and it works nicely.  It even pages in a useful manner.

return42:
@virtadpt thanks for taking the time and test .. I don't have a Recoll instance .. if it is OK for you I think its ready to merge :)

OliveiraHermogenes:
closes #2957

Issue #1604, opened 2022-08-02
Improve usability and document the bang (!) features


return42:
Improve usability and document the bang (!) features

In #1545 I tried to polish the rough edges of `categories_as_tabs` but the discussion we had in the review makes it clear to me that we need to improve the documentation and the usability of the bang features we have.

First let us sort what we have, we use bang (``!`` and ``!!``) for:

1. categories shown in the tab (*first class categories* / e.g. `!images dali`)
2. categories to group engines (*secondary categories* / e.g. `!dictionaries smart`)
3. engine names (e.g. `!google x^2`)
4. engine shortcuts (e.g. `!go x^2`)
5. external bangs (e.g. `!!wolframalpha x^2`)

## Categories shown in the tab

Are configured in the `settings.yml` under section `categories_as_tabs`:

https://github.com/searxng/searxng/blob/7c9c112484ff091a7c11df1dbb645616b57d662f/searx/settings.yml#L252-L262

Each item is a dictionary what scales good for adding new properties.

More or less (un-) documented in [Configured Engines](https://docs.searxng.org/admin/engines/configured_engines.html#configured-engines) / the documentation is grouped in subsections that are build from `categories_as_tabs`.

The default *first class categories* SearXNG uses are added to the translation:

https://github.com/searxng/searxng/blob/7c9c112484ff091a7c11df1dbb645616b57d662f/searx/searxng.msg#L22-L34

## Categories to group engines

Adding a engine to a category that is not in `categories_as_tabs` add this engine to that *group*.  By example:

```yaml
engines:
  ...
  - name: mediathekviewweb
    categories: general, tv
```

will add the `mediathekviewweb` to the *first class category* `general` that is shown as a tab in the UI and it adds the engine to *secondary category* `tv`.

The *secondary categories* are more or less (un-) documented in [Configured Engines](https://docs.searxng.org/admin/engines/configured_engines.html#configured-engines) / the sub-sub sections are the *groups* that are spread in the subsections (from `categories_as_tabs`). By example we have a group named `web` that groups the common WEB-search engines, we have this group in several *first class categories* like `general`, `images`, `videos` .. and more.

The default *secondary categories* SearXNG uses are added to the translation:

https://github.com/searxng/searxng/blob/7c9c112484ff091a7c11df1dbb645616b57d662f/searx/searxng.msg#L36-L46

Limitation of the *secondary categories* are:

- there is only the "name" not a dictionary with properties (like we know from `categories_as_tabs`)

[...]

PR #1545, opened 2022-07-24
Improve categories_as_tabs & group engines in tabs

return42:
## What does this PR do?

Fix several issues and nits of the PR:

- https://github.com/searxng/searxng/pull/634

for details read the commit messages.

## Why is this change important?

Fix issues and improve UX.

## `[doc] add description of categories_as_tabs (settings.yml)`

    make docs.live

.. and read documentation of `categories_as_tabs`

## `[fix] don't show category if there is not active engine in`

go to preference and disable all engines of one category (e.g. social media):

before ..

image

after ..

image

## `[fix] categories can't be removed from UI (categories_as_tabs)`

In your `/etc/searxng/settings.yml` activate `use_default_settings: true` and reduce the list of in `categories_as_tabs`:

```yaml
use_default_settings: true

categories_as_tabs:
  general:
  images:
  videos:
```

image

## `[mod] rename DEFAULT_GROUP_NAME = 'ungrouped'`

image

## `[mod] show !bang of (secondary) categories in the preference`

image

## Related issues

Related to: 
- https://github.com/searxng/searxng/issues/690
- https://github.com/searxng/searxng/pull/1562

Closes: https://github.com/searxng/searxng/issues/1020

PR #2326, opened 2023-04-07
[mod] clarify the difference of the default category and subgrouping

return42:
This PR does no functional change it is just an attempt to make more clear in
the code, what a default category is and what a subcategory is.  The previous
name 'others' leads to confusion with the **category 'other'**.

If a engine is not assigned to a category, the default is assigned::

    DEFAULT_CATEGORY = 'other'

If an engine has only one category and this category is shown as tab in the user
interface, this engine has no further subgrouping::

    NO_SUBGROUPING = 'without further subgrouping'

Related:

- https://github.com/searxng/searxng/issues/1604
- https://github.com/searxng/searxng/pull/1545

### DIFF
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -298,18 +298,18 @@
     <p>{{ _('Currently used search engines') }}</p>
     {{ tabs_open() }}
      {% set ns = namespace(checked=true) %}
-     {% for categ in categories_as_tabs + [OTHER_CATEGORY] %}
+     {% for categ in categories_as_tabs + [DEFAULT_CATEGORY] %}
      {{ tab_header('enginetab', 'category' + categ, _(categ), ns.checked )}}
      {% set ns.checked = false %}
-   {% if categ == OTHER_CATEGORY %}
-      <p>{{_('This tab does not show up for search results, but you can search the engines listed here via bangs.')}}</p>
-    {% endif %}
+     {% if categ == DEFAULT_CATEGORY %}
+      <p>{{_('This tab dues not exists in the user interface, but you can search in these engines by its !bangs.')}} <a href="{{ url_for('info', pagename='search-syntax') }}">&#9432;</a></p>
+     {% endif %}
     <div class="scrollx">
     <table class="striped table_engines">
       <tr>{{- "" -}}
         <th class="engine_checkbox">{{ _("Allow") }}</th>{{- "" -}}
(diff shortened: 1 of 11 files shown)


searx/templates/simple/preferences.html
```diff
@@ -298,18 +298,18 @@ <h1>{{ _('Preferences') }}</h1>
...
+     {% if categ == DEFAULT_CATEGORY %}
+      <p>{{_('This tab dues not exists in the user interface, but you can search in these engines by its !bangs.')}} <a href="{{ url_for('info', pagename='search-syntax') }}">&#9432;</a></p>
+     {% endif %}
     <div class="scrollx">
     <table class="striped table_engines">
       <tr>{{- "" -}}
         <th class="engine_checkbox">{{ _("Allow") }}</th>{{- "" -}}
         <th class="name">{{ _("Engine name") }}</th>{{- "" -}}
-        <th class="shortcut">{{ _("Shortcut") }}</th>{{- "" -}}
+        <th class="shortcut">{{ _("!bang") }}</th>{{- "" -}}
```
dalf, line 312: Does that mean the engine name is not a bang? What would be the name?

Should this section be adjusted?

https://github.com/searxng/searxng/blob/393e14965acbf0c9a1942a249a55705be9881918/searx/infopage/en/search-syntax.md?plain=1#L7-L26

Or should the !bang column display, for example, `!archive_is`, `!ai` for the archive.is engine?

return42: > Does that mean the engine name is not a bang? What would be the name?

No, I haven't thought that way when I changed from "shortcut" to "!bang" in the column header.

> Should this section be adjusted?

I don't think so, the engine name is still a valid !bang .. 

My thought process was: I think it's hard to show all possibilities in the user interface (preferences) without overloading the UI ... the users will usually use the *shortcut bang*.  On the info page  can read that the *engine bang* is also a valid search syntax.

> Or should the !bang column display, for example, `!archive_is`, `!ai` for the archive.is engine?

Yes this is what I intend, but in a follow up PR.  Now, when I read your comment, think it was not a good idea of mine to add the change from "shortcut" to "!bang" in this PR -- > its not related to this PR / sorry.

return42: > Or should the !bang column display, for example, `!archive_is`, `!ai` for the archive.is engine?

In https://github.com/searxng/searxng/pull/2327 the !bang from shortcut, categories and the engine name are displayed in the tool-tip.

Issue #1899, opened 2022-10-28
Bug: duckduckgo engine


Universalizer:
![chrome_screenshot_Oct 28, 2022 10_57_00 AM GMT+05_30](https://user-images.githubusercontent.com/116723165/198510982-114e2033-e850-4f8b-9112-def48ca9cf5a.png)
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2022.10.25-affd8f75

**How did you install SearXNG?**

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: httpx.ReadTimeout
  * Percentage: 0
  * Parameters: `(None, None, 'lite.duckduckgo.com')`
  * File name: `searx/search/processors/online.py:113`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Error
  * Error: searx.exceptions.SearxEngineAccessDeniedException
  * Percentage: 0
  * Parameters: `('HTTP error 403',)`
  * File name: `searx/search/processors/online.py:113`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

return42:
duplicate https://github.com/searxng/searxng/issues/1854 and several others https://github.com/searxng/searxng/issues?q=is%3Aissue+%22Bug%3A+duckduckgo+engine%22

PR #2333, opened 2023-04-10
[mod] enable autocomplete from duckduckgo by default

### LINKED ISSUE #1889: [Feature request] Enable autocomplete by default
By default autocomplete is disabled.
It was easy to find the flag and make the change, but, it worked only in the SearXNG web UI, not in Firefox search autocomplete. Even if I enable it in the Web UI preferences, the autocomplete is not working in Firefox.

I spent too much time fixing the issue:
1. Enable the autocomplete in the setting.yml
2. Remove SearXNG from Firefox
3. Restart Firefox
4. Add SearXNG again

Enabling autocomplete by default will fix this issue for new users.

return42:
## What does this PR do?

enable autocomplete from duckduckgo by default

- https://github.com/searxng/searxng/issues/1889

## Related issues

Closes: https://github.com/searxng/searxng/issues/1889

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -26,7 +26,7 @@ search:
   # Existing autocomplete backends: "dbpedia", "duckduckgo", "google", "yandex",
   # "seznam", "startpage", "swisscows", "qwant", "wikipedia" - leave blank to turn it off
   # by default.
-  autocomplete: ""
+  autocomplete: "duckduckgo"
   # minimun characters to type before autocompleter starts
   autocomplete_min: 4
   # Default search language - leave blank to detect from browser information or


dalf:
A more private alternative is to add a no-backend autocompletion (named "autocomplete bangs only"):
* what the user types is not sent to send to third party
* bangs and external bangs autocompletion work
* installation in Firefox works

return42:
> A more private alternative is to add a no-backend autocompletion (named "autocomplete bangs only"):

OK, I understand its a more privacy default but I'm not sure it meets user's expectations described in https://github.com/searxng/searxng/issues/1889

@ngosang you are the opener of #1889 .. what do you think about a default auto-completion that only completes !bangs (and !!bangs) .. for a real auto completion we need to enable a back-end (like it is done here in the PR / ddg auto-completion is activated by default).

ngosang:
I think it's a good idea to enable one backend for autocomplete as default (for example ddg). The user will use an useful feature from day one. The user could disable (and enable again) the autocomplete and Firefox will still works. Disabling autocomplete improves privacy but SearX already has more than 3 search engines enabled by default, I see some incoherence so I prefer autocomplete enabled by default. This PR will close my issue.

unixfox:
Mixed feelings in enabling a default engine for the autocompletion, especially because there are a lot of also great engines that would fit for the autocompletion.

Unlike a simple search, you can't have results from multiple engines for the autocompletion, so "forcing by default" one engine and not the other feels wrong to me.

I would much rather print a warning in the logs that the user has not configured an autocompletion setting.

ngosang:
The Firefox issue could be resolved without enabling autocomplete by default => https://github.com/searxng/searxng/issues/1889#issuecomment-1565310385

I still think it's a good idea to enable autocomplete by default. If the issue is we don't want to favor one search engine above others we can implement support for multiple autocomplete providers enabled. It will be the same as search, we make requests to all providers, de-suplicate the results and return to the client. It could be a bit slower but it will be more resilient if some provider stops working. You can still configure just one. We can enable 2 or 3 as default (same as search engines).

return42:
Not sure what we can do, see my https://github.com/searxng/searxng/issues/1889#issuecomment-1565366772

Since we open the user the ability to choose a backend, the user has first to select the backend and then add the SearXNG instance to the WEB-browser.

But as described in #1899 this is also the issue, when the user wants to change the preferred autocompletion backend, the user has to remove the instance from the WEB-browser and re-add the SearXNG  instance (with the new selected backend).

If the user did not remove/re-add the SearXNG instance from the WEB-browser the backend used in autocompletion in the URL bar is the previous selected backend, and the backend in the search field of SearXNG is the new selected backend / the later one is showed in the preferences, the first one is hidden to the user.

Does anyone have a clue how we can improve the situation?

-------

To avoid my users from this conflict, on my instance I pre-selected and locked the autocompletion on server side .. `/etc/searxng/settings.yml`:

```yaml
search:
  ..
  autocomplete: 'google'
  ..

preferences:
  lock:
    - autocomplete
```

I did this at that time because I thought that the simple user does not know any criteria according to which he could make a good selection.  I had in mind that there are backends that support languages and those that do not // Its unknown to the user that 

- brave, dbpedia, seznam, swisscows, yandex do not support languages
- duckduckgo, google, startpage, qwant and wikipedia support languages

See https://github.com/searxng/searxng/blob/master/searx/autocomplete.py

return42:
Related: 

- https://github.com/searxng/searxng/discussions/4262#discussioncomment-12005746

PR #2348, opened 2023-04-16
Fix Israeli flag emoji in locale dropdown

MarcAbonce:
## What does this PR do?

The locale dropdown shows the flag of Iran (🇮🇷) instead of the flag of Israel (🇮🇱) for the Hebrew language.

See: https://emojipedia.org/flag-israel/

UI bug.

1. Go to `/preferences`
1. Look at the `Search language` dropdown.
1. Hebrew option (עברית) should have the correct flag.


return42 (approved):
Ops, big fail / thanks a lot for your fix :+1:

PR #2129, opened 2023-01-25
implement rate limiter whitelisting

blob42:
## What does this PR do?

Add the possibility to whitelist IPs or subnets from the rate limiter. 

Adds the options under the server section in `settings.yml`:
`server.limiter_whitelist_ip`
`server.limiter_whitelist_subnet`

## Why is this change important?

This is a proposal for #2127. I am trying to add Searx as a search engine source for [langchain](https://github.com/hwchase17/langchain/pull/854). Using the json endpoint quickly hits the rate limiter. People usually enable the rate limiter, adding this feature allows the instance to be used publicly as well as a for development purposes. 

## How to test this PR locally?

I added proper unit tests and tried running the instance with valid and invalid parameters.

## Author's checklist

TODO:
- [x] add tests
    - validate the ip address/subnet format 
- [x] update documentation

## Related issues

Fixes #2127

Issue #1237, opened 2022-05-26
[limiter] Rate limiting does not work


mrpaulblack:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

```
Leons instance: 2022.05.24-fddbc5ed (39077fc0)
ALso my instance: 2022.05.08-85c1c14f 
``` 

**How did you install SearXNG?**

Using the scripts as well as with docker.

**What happened?**

The limiter plugin does not rate limit requests and uses the `X-Forwarded-For` header which could be easily set by a client request.

**How To Reproduce**
Make 15 `/search` requests in 20sec to trigger rate limiting.
Make another `search` request, which is going to get blocked with a HTTP 429 response.
Make another `/search` request which is going to get not blocked again.

**Expected behavior**
The IP should get blocked for a specific amount of time with a timeout value for example. So a offending host could get blocked for like 2 hours for example.

**Screenshots & Logs**

**Additional context**
Another problem is that the IP is read from a header. In a improper setup the IP could maybe not be in that header or a client could abuse this and set the header themselves. The IP should be gotten from a trusted source... (Maube the request lib has something to get the IP?)

mrpaulblack:
for vis: @dalf @return42 @tiekoetter

tiekoetter:
Tested on my instance :

IPv4:
- Refresh about 15 times -> 429
- Wait about 10 seconds -> 200
- Refresh about once a second after getting blocked initially -> 429

IPv6:
- Refresh about 15 times -> 429
- Wait about 10 seconds -> 200
- Refresh about once a second after getting blocked initially -> 429

dalf:
> Another problem is that the IP is read from a header.

* Caddy: https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults
* Traefik: https://doc.traefik.io/traefik/v2.1/routing/entrypoints/#forwarded-header

return42:
> **How To Reproduce**
> Make 15 `/search` requests in 20sec to trigger rate limiting.
> Make another `search` request, which is going to get blocked with a HTTP 429 response.
> Make another `/search` request which is going to get not blocked again.

"is going to get not blocked again." .. I can't reproduce this behaviour .. in my tests, the *sliding window* works and blocks the following request  .. @paul, can you please verify your test.

mrpaulblack:
[...]

PR #2357, opened 2023-04-19
[mod] limiter: reduce request rates for requests without a ping

return42:
## What does this PR do?

Adds a very simple method to improve the limiter / *send a ping from the UI*.

Too many bots pass the limiter and with their requests they worsen the acceptance of SearXNG engines.  In the issue Tracker we find many reports and comments about CAPTCHA and blocking of engines.  The requests of the engines can't be further improved to avoid CAPTCHA.

As long as bot requests are passed through from SearXNG, SearXNG will be classified as a bot by the origin search-engines.

The only way to improve this state is to develop methods that allow the limiter to better detect bots.

## Author's checklist

- We can add a token to the limiter URL
- This binds the limit of request rates to the `simple` theme.


searx/templates/simple/base.html
```diff
@@ -17,6 +17,9 @@
   {% else %}
   <link rel="stylesheet" href="{{ url_for('static', filename='css/searxng.min.css') }}" type="text/css" media="screen" />
   {% endif %}
+  {% if get_setting('brand.public_instances') %}
+  <link rel="stylesheet" href="/limiter{{ limiter_token }}.css" type="text/css" media="screen" />
+  {% endif %}
```
return42, line 22: @unixfox: here the limiter is bounded to the UI (I added a random token to the URL).

unixfox:
Need a lot of checking that everything works fine for all the possibles browsers. For example through browserstack or similar.

A list that came to my mind when I developed https://github.com/unixfox/antibot-proxy:
- modern browsers will cache CSS and JS resources and may not request them again to the server. this will need to be checked
- this will block browsers that don't request CSS like lynx or elinks (terminal browsers)

About your implementation:
- This will obviously increase the response time as searxng will check in the redis DB for every request. Maybe this need an internal memory cache?
- When redis is not configured, I didn't understand, you still generate the "ping" CSS resource but it is predictable as the hash key is always `12345678`? Seems odd to do that because then most bots can easily bypass the requests when the limiter is not configured.

searx/plugins/limiter.py
```diff
@@ -40,6 +42,39 @@
...
 )
 
+PING_KEY = 'SearXNG_limiter.ping'
+TOKEN_KEY = 'SearXNG_limiter.token'
+
+
+def ping():
+    redis_client = redisdb.client()
+    user_agent = request.headers.get('User-Agent', 'unknown')
+    x_forwarded_for = request.headers.get('X-Forwarded-For', '')
```
unixfox: what happen when `X-Forwarded-For` is empty like for local usage with a request directly to searxng without passing through a reverse proxy?

searx/plugins/limiter.py
```diff
@@ -40,6 +42,39 @@
...
+    if not redis_client:
+        # This function is also called when limiter is inactive / no redis DB
+        # (see render function in webapp.py)
+        return '12345678'
+    token = redis_client.get(TOKEN_KEY)
+    if token:
+        token = token.decode('UTF-8')
+    else:
+        token = ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(8))
+        redis_client.set(TOKEN_KEY, token, ex=600)
```
return42: Here the token expires after 600 sec

searx/webapp.py
```diff
@@ -648,6 +649,13 @@ def health():
     return Response('OK', mimetype='text/plain')
 
 
+@app.route('/limiter<token>.css', methods=['GET', 'POST'])
```
return42: Here the token is expected in the URL / instead of an (empty) css we can return a (empty) html.

searx/plugins/limiter.py
return42: > what happen when `X-Forwarded-For` is empty like for local usage with a request directly to searxng without passing through a reverse proxy?

`X-Forwarded-For` is a prerequisite of the limiter.  If this value is empty, the requests are counted in the context of an empty IP (like any other IP). / This behavior is the same, with or without this PR.

searx/templates/simple/base.html
```diff
@@ -17,6 +17,9 @@
   {% else %}
   <link rel="stylesheet" href="{{ url_for('static', filename='css/searxng.min.css') }}" type="text/css" media="screen" />
   {% endif %}
+  {% if get_setting('server.limiter') %}
+  <link rel="stylesheet" href="/limiter{{ limiter_token }}.css" type="text/css" media="screen" />
```
return42: If css is not loaded by browsers like lynx .. will a `<iframe>` be loaded?

searx/templates/simple/base.html
```diff
@@ -17,6 +17,9 @@
   {% else %}
   <link rel="stylesheet" href="{{ url_for('static', filename='css/searxng.min.css') }}" type="text/css" media="screen" />
   {% endif %}
+  {% if get_setting('server.limiter') %}
```
return42, line 20: > When redis is not configured,

The limiter request is only inserted if the limiter is activated.

return42:
> modern browsers will cache CSS and JS resources and may not request them again to the server. this will need to be checked

additionally in the static folder wsgi sets expire period to 365 days (what is questionable but OT here) .. 

https://github.com/searxng/searxng/blob/98387e291059d662dfb1da0ba131ca25da8b8d3c/utils/templates/etc/uwsgi/apps-available/searxng.ini#L85

But the URL of the limiter is changed every 10 min by a random token / compare my comments above.

> This will obviously increase the response time as searxng will check in the redis DB for every request.

Believe me: no!  .. Not a redis request is the bottleneck (we already have multiple) ... the additional HTTP request from client counts more. But thinking about optimizations is a very different task .. we have the requests & timeouts to the engines, to name just one point that counts most.

> Maybe this need an internal memory cache?

The communication is stateless (no sessions); we de not have "one" memory where we can store values (we have this "hackish" wsgi cache that is shared by workers, but since this "shared memory" depends on the architecture how SearXNG is setup, we decided to use the redis DB for the limiter).

> When redis is not configured, 

Since limiter can't be used without redis (all the limiter code is based on redis) this static URL will never be used (compare my comments in the code above).

> Need a lot of checking that everything works fine 

:+1: .. I will do more tests, your comments helped my a lot .. if more comes into mind, please leave a note here / thanks.

searx/templates/simple/base.html
unixfox: lynx doesn't automatically load the iframe nor it loads images nor javascript. the only way to detect that it's a terminal-based browser is to fingerprint the browser with methods like https://github.com/searxng/searxng/discussions/2264. as the user agent header is not trustworthy

unixfox:
> But the URL of the limiter is changed every 10 min by a random token / compare my comments above.

That doesn't really matter, if the user does multiple searches in a row then for the second time he visits the search page the "ping" CSS resource will be cached, and the issue may appear.

searx/plugins/limiter.py
```diff
@@ -40,6 +42,39 @@
...
+TOKEN_KEY = 'SearXNG_limiter.token'
+
+
+def ping():
+    redis_client = redisdb.client()
+    user_agent = request.headers.get('User-Agent', 'unknown')
+    x_forwarded_for = request.headers.get('X-Forwarded-For', '')
+
+    ping_key = PING_KEY + user_agent + x_forwarded_for
+    redis_client.set(secret_hash(ping_key), 1, ex=600)
```
return42: >> But the URL of the limiter is changed every 10 min by a random token / compare my comments above.

> It doesn't really matter, if the user does multiple searches in a row then for the second time he visits the search page the "ping" CSS resource will be cached, and the issue may appear.

The token is also stored for 600 sec.  Only when the token changes after 10min, it must be "unlocked" again.

searx/plugins/limiter.py
```diff
@@ -56,9 +91,20 @@ def is_accepted_request() -> bool:
...
     if request.path == '/search':
 
+        c_burst_max = 2
+        c_10min_max = 10
+
+        ping_key = PING_KEY + user_agent + x_forwarded_for
+        if redis_client.get(secret_hash(ping_key)):
+            logger.debug('got a ping')
+            c_burst_max = 15
+            c_10min_max = 150
```
return42: here the token, that have been stored (for 10min) is reused.

searx/plugins/limiter.py
unixfox: Mmh OK I understand the implementation. Then it's pretty fine in this case.

searx/plugins/limiter.py
```diff
@@ -56,9 +91,20 @@ def is_accepted_request() -> bool:
...
+        else:
+            logger.debug('missing a ping')
+
         c_burst = incr_sliding_window(redis_client, 'IP limit, burst' + x_forwarded_for, 20)
         c_10min = incr_sliding_window(redis_client, 'IP limit, 10 minutes' + x_forwarded_for, 600)
-        if c_burst > 15 or c_10min > 150:
+        if c_burst > c_burst_max or c_10min > c_10min_max:
             logger.debug("BLOCK %s: to many request", x_forwarded_for)
             return False
```
unixfox: `BLOCK %s: to many request` => `BLOCK %s: too many requests`

unixfox (changes requested):

unixfox:
Couple of questions:

- What happen when multiple people behind the same public IP address do in a row some search requests on the instance? Wouldn't that block the users for too many requests? The browser of the users may not fetch the CSS resource in time for not getting blocked.

----------

Additional idea about the limiter, I feel like we shouldn't block entirely the user but present a CAPTCHA in case the limiter wrongly block the user. This improves the user experience because the user wouldn't be without any option for solving this issue.
The CAPTCHA should obviously be without javascript.

return42:
> Couple of questions:
> 
>  * What happen when multiple people behind the same public IP address do in a row some search requests on the instance? 

Yes / more words to explain what the limiter does today and what this PR adds on top:

Assume you have a subnet with 100 or more PCs and all the users use the same SearXNG instance. Then all these user have (from SearXNG's point of view) the same IP.  The limiter is counting `if c_burst > 15 or c_10min > 150:` on this IP --> the probability that the limits are reached is greater the more clients from the same subnet make requests to the same instance.

This is a known drawback of the limiter implementation and this PR does not improve this situation.  To improve this situation we need to store sessions (in the request / response cycle and no longer based on an IP address) .. but this is OT of this PR .. and it is nothing we want to do / we do not want to store a user context / we do not want to track the user.

This PR tries to identify requests from bots; if there is no ping, the limits (for this IP) are reduced to `if c_burst > 2 or c_10min > 10:`.  However, if the bot request came from a subnet from which a real user requests also came, then the limits would be increased again.

This evaluation takes place in a time window that lasts a maximum of 10min, after which the situation is again as if no request had ever come from this subnet (the game starts again from the beginning).

The probability that a bot is in the same subnet as a user is very low, so that we can assume that the bots from whose IP no ping comes must work with the reduced rates.

The intention of this is not to block bots completely!  The bots should only not make as many requests as other subnets. Or in other words: The intention of this PR is to prevent bots that are badly programmed (bots that run crazy) from passing the limiter and then degrading the engines.  From the description of this PR -->

> Too many bots pass the limiter and with their requests they worsen the acceptance of SearXNG engines.

-----

About CAPTCHA ..

> Additional idea about the limiter, I feel like we shouldn't block entirely the user but present a CAPTCHA in case the limiter wrongly block the user. This improves the user experience because the user wouldn't be without any option for solving this issue. The CAPTCHA should obviously be without javascript.

Here I see two problems:

1. I do not know a CAPTCHA implementation we can use and that is not tracking the user.

2. We need to store session data .. what we do not want to do.

The limiter should not store any user information if possible, so currently it only works on the basis of the IP, whose hash value is stored for a maximum of 10min.  With this PR the user agent header is also stored in a hash value (max 10min).  But we never want to store data (even if only hash values are stored) that allow to build a session context.  So we have to be careful with what we store ... and how we store it (hash values, never reproducible data).

Without session context we will never be able to implement a perfect bot-blocker, but we can significantly improve the situation for the engines with our limiter ... and that is the goal of this PR.

unixfox:
Understood that it's to improve the user experience and that it's not fundamentally changing the way the limiter treats the amount of requests that can be done per seconds/minutes.

One issue remain, the user experience for terminal based browsers will be worsened. How do we deal with this issue?
Multiple options:

- We do not care about them. This is sad for users that use terminal based browsers, while I understand it's a very niche user base. They are free to run their own instance with the limiter disabled though.
- We print a message when we detect the user agents of the terminal based browsers saying to contact the maintainer for whitelisting the user agents or turning on an option in the settings for allowing the terminal based browsers.   
   The added advantage is that the bad bots can't know in advance if the terminal based browsers are whitelisted on the instance, it's up to the maintainer to decide or not to whitelist them.  
   We need the ability to whitelist user agents for that to work.
- We whitelist the user agents of the terminal based browsers. But this would effectively allow bots to easily bypass the limiter.
- We only support terminal based browsers that support images like w3m. Instead of a CSS resource, we would ask the browser to load an image for validating the ping.

------

Note about searx.space:
We will need to adapt it because the CSS or image PING resource will be classified as customization and all the instances will receive the `C` letter.

return42:
> One issue remain, the user experience for terminal based browsers will be worsened. How do we deal with this issue? Multiple options:

We can whitelist IPs like 

- https://github.com/searxng/searxng/pull/2129

suggest .. I haven't reviewed this PR since I'm unsure it is the best solution.

I haven't thought to its end but what comes in mind; we have ["Private Engines (tokens)"](https://docs.searxng.org/admin/engines/private-engines.html) .. why not implement the same to pass the limiter by such a token.

unixfox:
But ideally, we would like to find a solution that doesn't require the maintainer action or at least not a solution that require the maintainer to have personal data identifier like the IP address for whitelisting the user.

I'm talking about the most common case, when a user uses the terminal based browser on a public instance.

unixfox:
How do we move forward here? I would be tempted at first to have these protections only available under a specific setting called: 

```
limiter:
  public_instance: true
```

So that it doesn't affect the private/single used instances though.

return42:
> How do we move forward here? 

Sorry for the delay .. I need to do some more thorough testing first and see if there are any simpler solutions.  I will take your comments into account.  But I will also need one or two weeks until I have the time to build up a test.

unixfox:
Can we allow the instances to beta test? We merge the code and allow the public instances to turn on a setting for testing.

return42:
FYI : I test this branch at https://darmarit.org/searx/

To become a bot you can repeat this command three times:

    curl -H 'Accept-Language: de-DE,en-US;q=0.7,en;q=0.3' \
         -H 'Accept: text/html' \
         -H 'User-Agent: xyz'  \
         -H 'Accept-Encoding: gzip' \
         'https://darmarit.org/searx/search?q=foo'

Then you should get `Too Many Requests` and your IP is blocked for the next 24 hours .. to release your IP, open https://darmarit.org/searx in your WEB-browser .. as long your WEB-browser is not a [terminal based browsers](https://github.com/searxng/searxng/pull/2357#issuecomment-1521856728) that do not load CSS files, you can continue with your requests.

unixfox:
Awesome!

-------

Just wanted to say that with IPv6, you get a very huge range of IP addresses and this could easily defeat the anti bot solutions.

Usually that's why services will straight ban the /48 subnet of the IPv6 address. See more: https://github.com/scipag/RIPv6 and https://github.com/ycd/ipv6-rotator

return42:
Forgot to say, the documentation is available from here https://return42.github.io/searxng/src/searx.botdetection.html

As long you did not activate the [`link_token` method](https://return42.github.io/searxng/src/searx.botdetection.html#module-searx.botdetection.link_token) .. the limiter works as usual .. to activate the method see [`ip_limit` method](https://return42.github.io/searxng/src/searx.botdetection.html#method-ip-limit) --> ...

image

unixfox:
Why is this a separate config file though?

return42:
The settings.yml file is already overloaded today.  If we want to implement more features like https://github.com/searxng/searxng/pull/2129#issuecomment-1551258452 we need a separate file for that.

Further I want to isolate the limiter from the SearXNG core. With the limiter.toml a customizing of the limiter is possible ... without this file the limiter runs in the default as it always did before.

----

Setting up new options in the [settings.yml](https://github.com/searxng/searxng/blob/79c02456c1a852ecab5548a840d09e9642565615/searx/settings_defaults.py#L139-L242) is a bit complex.  For the limiter we do not need this complexity .. implement a new option is nothing more than a new entry in the [schema file](https://github.com/return42/searxng/blob/limiter-ping/searx/botdetection/limiter.toml)


---

OT: in the long run, the engines should also be moved out of settings.yml

return42:
> Usually that's why services will straight ban the /48 subnet of the IPv6 address.

thanks for the hint ... this still needs to be implemented ... I will first continue to run my tests on IPv4 and think about something for IPv6

image

return42:
In the following example the bot rotates (v4) IPs 

```
May 27 23:07:00  WARNING searx.botdetection.link_token : missing ping (IP: 140.238.172.132) / request: 68612c619f764707ec156dec366e007b020ebbbb1d738e28d60fd05bc3773384
...
May 27 23:07:00  WARNING searx.botdetection.link_token : missing ping (IP: 140.238.172.132) / request: 68612c619f764707ec156dec366e007b020ebbbb1d738e28d60fd05bc3773384
...
May 27 23:11:52  WARNING searx.botdetection.link_token : missing ping (IP: 140.238.172.132) / request: 68612c619f764707ec156dec366e007b020ebbbb1d738e28d60fd05bc3773384
...
May 27 23:14:23  WARNING searx.botdetection.link_token : missing ping (IP: 140.238.172.132) / request: 68612c619f764707ec156dec366e007b020ebbbb1d738e28d60fd05bc3773384
May 27 23:14:23  ERROR   searx.plugins.limiter         : BLOCK (IP 140.238.172.132): bot detected, too many request from 140.238.172.132 in SUSPICIOUS_IP_WINDOW
```

the bot in the example reached the end of its IP list about every three minutes and started over again.

After 7 minutes, it was identified as a bot.

There are bots with many more IPs that may make requests over several hours with ever new IPs ... then it takes not 7min but several hours until the bot is recognized as such.

I now let the test run for 24h and then look at what my instance has identified as a bot.

return42:
With this PR applied and and *link_token* method enabled my instance begin to block requests from searx.space

image

This issue can be solved by something like:

- https://github.com/searxng/searxng/pull/2129

unixfox:
One thing that could trigger the limiter is "Infinite scroll" feature. I did test, and you get 15 requests in a short amount of time (see column start time) until you get blocked.

I know that's a lot, but it's something to keep in mind, as the infinite scroll script won't make any request to the PING resource.

image

return42:
> I did test, and you get 15 requests in a short amount of time (see column start time) until you get blocked.

I debug while you did your test ...

    DEBUG   searx.botdetection.link_token : found ping for this request: 549553c3025039c76752a279925ab71522dbe3216171f6a8644b0e71c4a4f127
    ERROR   searx.plugins.limiter         : BLOCK (IP 120.166.34.102, 176.150.35.61): bot detected, too many request from 120.166.34.102, 176.150.35.61 in BURST_MAX

The BURST_MAX window allows 15 request .. This should be the same when current master is used and limiter is activated ~except in current master the value is 20~ (`c_burst > 15`):

https://github.com/searxng/searxng/blob/3ca97cf5e3f432b761553c912582449d12e43017/searx/plugins/limiter.py#L62 

... I can reproduce this issue with instances based on current master and limiter active.

----

OT here but the log from above shows two IPs ``BLOCK (IP 120.166.34.102, 176.150.35.61)`` in the `X-Forwarded-For` ?!?

The apache config is from here:

https://github.com/searxng/searxng/blob/3ca97cf5e3f432b761553c912582449d12e43017/utils/templates/etc/httpd/sites-available/searxng.conf#L32

And this config seems to work for all requests, excepet yours .. your request is the first where I see two IPs in the http header .. where is my fail?

searx/botdetection/link_token.py
```diff
@@ -0,0 +1,126 @@
...
+import string
+import random
+import flask
+
+from searx import logger
+from searx import redisdb
+from searx.redislib import secret_hash
+
+TOKEN_LIVE_TIME = 600
+"""Livetime (sec) of limiter's CSS token."""
```
return42, line 58: > as the infinite scroll script won't make any request to the PING resource.

@unixfox: The ping is requested with the first query and is hold in the DB for 10min.

If the user stops scrolling for 10min and then continue he comes into trouble with the limiter.

To fix this, the ping can be renewed when a token is validated ...

    DEBUG   searx.botdetection.link_token : found ping for this request: 549553c3025039c76752a279925ab71522dbe3216171f6a8644b0e71c4a4f127

unixfox:
> OT here but the log from above shows two IPs `BLOCK (IP 120.166.34.102, 176.150.35.61)` in the `X-Forwarded-For` ?!?
> 
> The apache config is from here:
> 
> https://github.com/searxng/searxng/blob/3ca97cf5e3f432b761553c912582449d12e43017/utils/templates/etc/httpd/sites-available/searxng.conf#L32
> 
> And this config seems to work for all requests, excepet yours .. your request is the first where I see two IPs in the http header .. where is my fail?

Haha, it's due to this extension: https://github.com/jake-cryptic/AbsoluteDoubleTrace, it sends a fake public ip for trying to avoid being tracked easily. You are probably not passing correctly the IP address to SearXNG, you should never use append but "overwrite" the header `X-Forwarded-For`.

At least with NGINX it's set correctly: https://github.com/searxng/searxng/blob/3ca97cf5e3f432b761553c912582449d12e43017/utils/templates/etc/nginx/default.apps-available/searxng.conf#L14

searx/botdetection/link_token.py
unixfox: Yes it's a good idea to have the "Infinite scroll" feature renewing the ping resource.

return42:
> At least with NGINX it's set correctly:
> 
> https://github.com/searxng/searxng/blob/3ca97cf5e3f432b761553c912582449d12e43017/utils/templates/etc/nginx/default.apps-available/searxng.conf#L14

Thats incorrect, both (nginx and apache setup) add the the client / see [proxy_add_x_forwarded_for](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#var_proxy_add_x_forwarded_for) .. and now I remember: in `X-Forwarded-For` the (proxy) chain is hold, separated by a comma .. and in `X-Real-IP` the IP of the *real* client (it's subnet router) is hold.

By example::

    curl -H 'X-Forwarded-For: 1.2.3.4' \
         -H 'X-Real-IP: 4.3.2.1' \
         'http://10.146.148.103/searxng/search?q=foo'

Looks on server side:

    X-Forwarded-For: 1.2.3.4, 10.146.148.1 || X-Real-IP: 10.146.148.1

`10.146.148.1` is the IP of the client (its subnet router).

In the limter we have used `X-Forwarded-For` / see https://github.com/searxng/searxng/issues/1237#issuecomment-1147564516 .. but it is a bug in the limiter that we do not split the list 

     X-Forwarded-For: <client>, <proxy1>, <proxy2>

I think the problem is, that we can not determine the number of proxies without knowledge of the infrastructure ...

return42:
@unixfox I add two new patches and rebased this branch on SearXNG's master.

- 571f2eb53
- a2c1a6dab

> Then you should get `Too Many Requests` and your IP is blocked for the next 24 hours .. to release your IP, open https://darmarit.org/searx in your WEB-browser ..

When your IP is rated as *suspicious* instead of a 429 `Too Many Requests` a 302 redirect to the intro page is returned .. by this, a common user (WEB-browser) should never see `Too Many Requests`.

The BURST_WINDOW and LONG_WINDOW that count requests of an IP are often false-positive when a number of users from the same subnet do SearXNG queries in a row .. and when a user uses *infinite scroll* .. and scrolls to fast :-)

To avoid these drawbacks, BURST_WINDOW and LONG_WINDOW are not used when the request is verified by the `link_token` method.

----

Forgot to say: its deployed on https://darmarit.org/searx  and can be tested there

return42:
> its deployed on https://darmarit.org/searx  and can be tested there

Beside the bot detection from the vanilla limiter, the `link_token` method detected round about 100 additional bots (100 IPs) in 8h.

return42:
> ust wanted to say that with IPv6, you get a very huge range of IP addresses and this could easily defeat the anti bot solutions. Usually that's why services will straight ban the /48 subnet of the IPv6 address. 

@unixfox see ..

- https://github.com/searxng/searxng/issues/2477

[281e36f](https://github.com/searxng/searxng/pull/2357/commits/281e36f4b7848374535d5e953050ae73423191ca) in this PR fixes the IPv6 issue / see https://github.com/searxng/searxng/pull/2357/commits/281e36f4b7848374535d5e953050ae73423191ca#diff-437a93b109d23719837e77f803998a188751dcc7f60cf711996939653defa967R7-R11

I think this PR now ready for review (no longer a draft).

searx/botdetection/ip_limit.py
```diff
@@ -0,0 +1,146 @@
...
+def filter_request(
+    network: IPv4Network | IPv6Network,
+    request: flask.Request,
+    cfg: config.Config,
+) -> werkzeug.Response | None:
+
+    # pylint: disable=too-many-return-statements
+    redis_client = redisdb.client()
+
+    if network.is_link_local and not cfg['botdetection.ip_limit.filter_link_local']:
```
unixfox, line 101: Where is the list of link local addresses?

searx/botdetection/_helpers.py
```diff
@@ -63,7 +89,9 @@ def get_real_ip(request: flask.Request) -> str:
     forwarded_for = request.headers.get("X-Forwarded-For")
     real_ip = request.headers.get('X-Real-IP')
     remote_addr = request.remote_addr
-    logger.debug("X-Forwarded-For: %s || X-Real-IP: %s || request.remote_addr: %s", forwarded_for, real_ip, remote_addr)
+    # logger.debug(
```
unixfox, line 92: Are those comments for debug or is it going to be implemented in the final code?

searx/botdetection/ip_limit.py
return42: The object `network` is a instance of class [IPv4Network](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network) or [IPv6Network](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv6Network). These classes are python builtin types to handle IP addresses & networks.  The documentation:
-  [is_link_local](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_link_local) --> constant definition in the code for [IPv4](https://github.com/python/cpython/blob/77d25795862f19c6e3d647b76cfb10d5ce1f149c/Lib/ipaddress.py#L1542-L1570) and [IPv6](https://github.com/python/cpython/blob/77d25795862f19c6e3d647b76cfb10d5ce1f149c/Lib/ipaddress.py#L2262-L2292)

In the configuration [searx/botdetection/limiter.toml](https://github.com/searxng/searxng/pull/2357/commits/281e36f4b7848374535d5e953050ae73423191ca#diff-437a93b109d23719837e77f803998a188751dcc7f60cf711996939653defa967) link local addresses are unfiltered by default .. in a follow-up PR we can add an additional option for [is_private](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_private)

searx/botdetection/_helpers.py
return42: I want to have this comments in the code .. its just for convenience to activate additional debug messages on my instance / I will do a lot of tests in the next weeks (also after a merge), sometimes it can be helpful to activate these additional (and very verbose) debug messages.  I will remove these comments from the code at some point.

searx/botdetection/ip_limit.py
unixfox: Shouldn't we offer the ability to add more link local networks? Some internal networks use public ranges, that's actually very common in IPv6 as there is usually no [NAT](https://en.wikipedia.org/wiki/Network_address_translation) involved and the "local" network is the public range of the network of computers behind the "modem" (internet connection).

return42: > Some internal networks use public ranges, 

Thanks for sharing your experience (I'm new to IPv6) .. very helpful :+1: 

When I added the option, I assumed that there would be many exceptions (such as private addresses or the cases you mentioned).  But link-local will always be link-local, that's irrevocable (right?).

With this knowledge in mind I decided to add the option link_local, other options I wanted to discuss in other PRs.

I see two possibilities for this PR:

a. we leave the link-local option as it is now.
b. we take it out of this PR again and move it to another PR where we discuss the question you asked "Shouldn't we offer the ability to add more link local networks?" in detail.

What do you think is the best way to proceed?

unixfox: Yes I've used the word link local, but I actually meant private networks.

Maybe you should rename the variable and function names to "private_networks" if all the addresses located here are "whitelisted": https://github.com/python/cpython/blob/77d25795862f19c6e3d647b76cfb10d5ce1f149c/Lib/ipaddress.py#L2262-L2292

We can open another pull request for adding a support for more private networks.

return42: > Maybe you should rename the variable and function names to "private_networks" if all the addresses located here are "whitelisted":

Sorry, my link was not clear, "link local" is only:

- `IPv6Network('fe80::/10')`
- `IPv4Network('169.254.0.0/16')`

by this, the name `link_locale` is OK / I think.

> We can open another pull request for adding a support for more private networks.

:+1:

Do you see any other questionable in this PR?

unixfox: Oh then the other IP addresses are not whitelisted?
```python
   _multicast_network = IPv6Network('ff00::/8')

    _private_networks = [
        IPv6Network('::1/128'),
        IPv6Network('::/128'),
        IPv6Network('::ffff:0:0/96'),
        IPv6Network('100::/64'),
        IPv6Network('2001::/23'),
        IPv6Network('2001:2::/48'),
        IPv6Network('2001:db8::/32'),
        IPv6Network('2001:10::/28'),
        IPv6Network('fc00::/7'),
        IPv6Network('fe80::/10'),
        ]

    _reserved_networks = [
        IPv6Network('::/8'), IPv6Network('100::/8'),
        IPv6Network('200::/7'), IPv6Network('400::/6'),
        IPv6Network('800::/5'), IPv6Network('1000::/4'),
        IPv6Network('4000::/3'), IPv6Network('6000::/3'),
        IPv6Network('8000::/3'), IPv6Network('A000::/3'),
        IPv6Network('C000::/3'), IPv6Network('E000::/4'),
        IPv6Network('F000::/5'), IPv6Network('F800::/6'),
        IPv6Network('FE00::/9'),
    ]

    _sitelocal_network = IPv6Network('fec0::/10')
```

return42: No, options for `_private_networks` and  `_reserved_networks` I would add in a follow up PR, I just linked them all to show, that there other properties to check [network.is_private](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_private) and [netwirk.is_reserved](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_reserved) ...

unixfox: Ok then we can tackle this issue in another PR.

unixfox:
There is still a remaining issue, searx.space can't properly scan your instance. See the column `Search response time` is empty.

return42:
> There is still a remaining issue, searx.space can't properly scan your instance.

This is not an issue of this PR .. to whitelist IPs (and searx.space) we need another PR / remember my comment from above https://github.com/searxng/searxng/pull/2357#issuecomment-1566091597 .. on which you gave your thumbs-up :-)

First we have to merge this PR .. then I can implement a whitelist method .. and when both are merged maintainers can (official) activate the `link_token` method .. in the meantime my instance is the test instance ..

unixfox:
> > There is still a remaining issue, searx.space can't properly scan your instance.
> 
> This is not an issue of this PR .. to whitelist IPs (and searx.space) we need another PR / remember my comment from above [#2357 (comment)](https://github.com/searxng/searxng/pull/2357#issuecomment-1566091597) .. on which you gave your thumbs-up :-)
> 
> First we have to merge this PR .. then I can implement a whitelist method .. and when both are merged maintainers can (official) activate the `link_token` method .. in the meantime my instance is the test instance ..

To my eyes, implementing a whitelist method should be part of this PR because then you are delivering a code that doesn't work for the specific target which are public instances.

But if you really want to do it in another PR then fine up to you...

return42:
> To my eyes, implementing a whitelist method should be part of this PR because then you are delivering a code that doesn't work for the specific target which are public instances.

From a functional POV you are right .. but its not activated by default and "whitelist" is a topic with its own questions and complexity, we better discuss in a new thread.

return42:
Small statistic of the ``link_token`` method after 30h test run:

The ``link_token`` method on my instance (IPv4 only) blocked **3587 requests** (~120/h)...

    $ sudo -H journalctl -u "uwsgi@searxng" | grep -o 'BLOCK: .* SUSPICIOUS_IP_WINDOW' | wc -l 
    3587

from **492 unique IPs** ...

    $ $ sudo -H journalctl -u "uwsgi@searxng" | grep -o 'BLOCK: .* SUSPICIOUS_IP_WINDOW' | sort -u | wc -l
    492

----

In this IP list there are IPs like https://cleantalk.org/blacklists/188.130.210.231 also listed in the http://iplists.firehol.org/?ipset=firehol_abusers_30d#

unixfox:
You did merge the PR but I haven't had the time to approve it.

For instance I've talked that it would be better to change the PING resource from CSS to an image so that some terminal based browser may still able to pass the test.

Further more the scroll feature doesn't work great with the limiter. I'll open an issue for that.

return42:
> You did merge the PR but I haven't had the time to approve it.

Aargh .. I'm very sorry .. I thought its OK for you to merge ..

> For instance I've talked that it would be better to change the PING resource from CSS to an image so that some terminal based browser may still able to pass the test.

:+1: 

> Further more the scroll feature doesn't work great with the limiter. I'll open an issue for that.

Did you see the issue I posted yesterday https://github.com/searxng/searxng/issues/2480 .. on my instance I have such a configuration .. on my instance this issue exists.

unixfox:
> > Further more the scroll feature doesn't work great with the limiter. I'll open an issue for that.
> 
> Did you see the issue I posted yesterday #2480 .. on my instance I have such a configuration .. on my instance this issue exists.

Oops did not, I'm submerged by issues lately (not just from SearXNG repo, [by Invidious too](https://github.com/iv-org/invidious/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed+The+video+returned+by+YouTube+isn%27t+the+requested+one.+)).

But this is not about the fact that the limiter blocks the autoscroll feature after a while.

return42:
> Oops did not, I'm submerged by issues lately (not just from SearXNG repo, by Invidious too

Oh yes I see .. prioritize Invidious.

> But this is not about the fact that the limiter blocks the autoscroll feature after a while.

This issue has been solved 

[The BURST_WINDOW and LONG_WINDOW that count requests of an IP are often false-positive when a number of users from the same subnet do SearXNG queries in a row .. and when a user uses infinite scroll .. and scrolls to fast :-) To avoid these drawbacks, BURST_WINDOW and LONG_WINDOW are not used when the request is verified by the link_token method.](https://github.com/searxng/searxng/pull/2357#issuecomment-1567536555)

And from the commit message of b8c7c2c9aa604fd

```
- renew a ping-key on validation [2], this is needed for infinite scrolling,
  where no new token (CSS) is loaded. / this does not fix the BURST_MAX issue in
  the vanilla limiter
```
.. this issue can't be fixed in the vanilla limiter since there is only a BURST_WINDOW and a LONG_WINDOW and no verification like done in the `link_token` method.

return42:
> For instance I've talked that it would be better to change the PING resource from CSS to an image so that some terminal based browser may still able to pass the test.

@unixfox: I mplemented a POC in:

- https://github.com/searxng/searxng/pull/2485

searx/botdetection/limiter.toml
```diff
@@ -0,0 +1,22 @@
...
+
+# Number of values to trust for X-Forwarded-For.
+
+x_for = 1
+
+# The prefix defines the number of leading bits in an address that are compared
+# to determine whether or not an address is part of a (client) network.
+
+ipv4_prefix = 32
+ipv6_prefix = 48
```
SuperSandro2000, line 11: I have some doubt that this ipv6 prefix is generally correct and that we can find a general correct prefix. My provider for example hands me out /56

unixfox: You can get /48 on some providers so that's why we consider this range to be a single network.

return42: > and that we can find a general correct prefix. 

@SuperSandro2000: that is true, but I would prefer /48 prefix over (e.g.) /56 .. there are two use cases we need to consider:

1. A bot in a /48 subnet where a user like you with a /56 prefix is also in (which will probably very rarely be the case).  --> the /48 subnet (and the bot) gets unblocked as soon a /56 user opens SearXNG in the WEB-Browser

2. A bot that owns a complete /48 subnet --> bot is permanently blocked

If we would use a /56 prefix, the bot could run through all its addresses in use case 2. and this would be in my opinion the much worse choice compared to a /48 prefix as described in use case 1.

return42:
> For instance I've talked that it would be better to change the PING resource from CSS to an image so that some terminal based browser may still able to pass the test.

@unixfox do we want to handle this request in a new issue? .. it's been a while, but I had already implemented a POC back then

-  https://github.com/return42/searxng/tree/link_token-image -->  e0a3dee3bf

.. if there is interest we could reactivate the topic.

unixfox:
> > For instance I've talked that it would be better to change the PING resource from CSS to an image so that some terminal based browser may still able to pass the test.
> 
> @unixfox do we want to handle this request in a new issue? .. it's been a while, but I had already implemented a POC back then

I think I did a year ago here: https://github.com/searxng/searxng/issues/2486

return42:
> I think I did a year ago here: #2486

I am a fool, I had lost sight of the issue .. and the PR https://github.com/searxng/searxng/pull/2485 .. :facepalm:

PR #2369, opened 2023-04-24
Go to main instead of search page when external bang query is empty

### LINKED ISSUE #2368: Double bangs without query redirect to search URL instead of main page
`2023.4.21+98387e29`

`searxng-docker`

**What happened?**
Double bangs without query redirect to search URL instead of main page

**How To Reproduce**
Use external bang without any search items e.g. `!!gh` and we land on `https://github.com/search?utf8=%E2%9C%93&q=` 

**Expected behavior**
`!!gh` should redirect to `https://github.com` if no search items

Jorenar:
## What does this PR do?

Added the check if the query for the external bang is empty.
If yes, then the URL is stripped to base domain.

DDG bangs are also convenient shortcuts to websites.  
With this change, the behaviour is consistent with the original.

1. Search `!!gh`
2. Verify it redirects to https://github.com instead of https://github.com/search?utf8=%E2%9C%93&q=

## Related issues

Closes #2368

### DIFF
--- a/searx/external_bang.py
+++ b/searx/external_bang.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
-from urllib.parse import quote_plus
+from urllib.parse import quote_plus, urlparse
 from searx.data import EXTERNAL_BANGS
 
 LEAF_KEY = chr(16)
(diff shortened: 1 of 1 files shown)


return42:
@Jorengarenar thanks a lot for this PR :+1:

> then the URL is stripped to base domain

To strip to base domain, we better have to use `urlparse` and use `url = o.scheme + '://' + o.netloc` to build a URL to the base domain / I think.

I pushed a "WIP" commit on your branch see 25b85381d1 (we can squash after this PR is accepted).

I tested with various external bangs:

When we strip to base domain, some external bangs will not fully functional .. by example, there are various external bangs linking to deeepl.com and setting a language preference:

- https://duckduckgo.com/bangs?c=Translation&sc=General&q=deepl

While we reduce the URL to the domain, all these bangs will lost the preferred language settings:

- `!!deeplfr` 
- `!!deeplpten` 
- `!!deeplenes`

But I think this is an acceptable limitation .. or?

Can you please make more tests with the modified implementation (with the WIP commit) .. Can you think of anything else that we haven't thought of yet?

Jorenar:
Thanks!

I don't see any issue with the https://github.com/searxng/searxng/commit/25b85381d1716218a1769ac2dcae867781cb6858 change.

>But I think this is an acceptable limitation .. or?

That's the behaviour of DDG as well. 
For the sake of consistency, convenience and simplicity I believe it should keep working like that.

PR #2387, opened 2023-04-28
Swedish translation error

truecrunchyfrog:
Original translation *"det"* means it, as in *"**it** is bad"* and not as in *"I work in **IT**"*.

## What does this PR do?

Fix translation error

because it may cause confusion

---

### DIFF
--- a/searx/translations/sv/LC_MESSAGES/messages.po
+++ b/searx/translations/sv/LC_MESSAGES/messages.po
@@ -73,7 +73,7 @@ msgstr "videor"
 #. CATEGORY_NAMES['IT']
 #: searx/searxng.msg
 msgid "it"
-msgstr "det"
+msgstr "it"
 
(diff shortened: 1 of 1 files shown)


return42:
Hi @javaveryhot / thanks a lot for your translation fix. TBH we do not fix translations directly in the code here.  Please visit our translation tool at https://translate.codeberg.org/projects/searxng/searxng/sv/ and leave a suggestion (no need to login) .. there are also some other unfinished strings, would be cool if you can give us your translations suggestions for .. thanks again for making SearXNG better / I'm going to close this PR:

return42:
@javaveryhot : today I accepted the new Swedish translation / I assume they are coming from you / thanks a lot .. 

FYI: unfortunately, these translations have not made it into today's merge.  But they will be in the merge automatically next weekend.

truecrunchyfrog:
Yes I think I added all of them. Happy to help :)

PR #2448, opened 2023-05-19
use logger.warning

pankwings:
## What does this PR do?

`logger.warn()` is depricated, use `logger.warning()` which is already being used in some files.

`logger.warn()` is depricated.

Before using this patch depreciation warnings were being logged. 
```
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```
with this patch, this warning is removed.


return42 (approved):
Thanks for your patch :+1:

Issue #2011, opened 2022-12-08
Bug: qwant engine


samsapti:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2022.12.05-67eea86b
Instance: https://search.sapti.me

**How did you install SearXNG?**

Custom `docker-compose.yml` file (based on the one from searxng-docker).

**What happened?**

Qwant cannot retrieve results, and gives the following error:

```
qwant ( server API error )
```

**How To Reproduce**

Search for something like `!qwant test`.

**Expected behavior**

Qwant should return results.

**Screenshots & Logs**

![image](https://user-images.githubusercontent.com/26819407/206529858-6d51153f-6dfd-440b-aac8-dc47ec4b24f8.png)

```
searxng  | 2022-12-08 18:03:18,047 WARNING:searx.engines.qwant: ErrorContext('searx/engines/qwant.py', 128, "raise SearxEngineAPIException('API error::' + msg)", 'searx.exceptions.SearxEngineAPIException', None, ('API error::unknown',)) False
searxng  | 2022-12-08 18:03:18,047 ERROR:searx.engines.qwant: exception : API error::unknown
searxng  | Traceback (most recent call last):
searxng  |   File "/usr/local/searxng/searx/search/processors/online.py", line 160, in search
searxng  |     search_results = self._search_basic(query, params)
searxng  |   File "/usr/local/searxng/searx/search/processors/online.py", line 148, in _search_basic
searxng  |     return self.engine.response(response)
searxng  |   File "/usr/local/searxng/searx/engines/qwant.py", line 128, in response
searxng  |     raise SearxEngineAPIException('API error::' + msg)
searxng  | searx.exceptions.SearxEngineAPIException: API error::unknown
```

**Additional context**

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineAPIException
  * Percentage: 100
  * Parameters: `('API error::unknown',)`
  * File name: `searx/engines/qwant.py:128`
  * Function: `response`
  * Code: `raise SearxEngineAPIException('API error::' + msg)`

return42:
FYI: I can't reproduce the issue on my instance [`!qwant test`](https://darmarit.org/searx/search?q=%21qwant+test), but I can see the same error in my [stats](https://darmarit.org/searx/stats?engine=qwant) .. compared to your 100% my percentage is somewhere by 0%:

Error
[...]

PR #2451, opened 2023-05-22
[doc] answer CAPTCHA from server's IP

return42:
Related: https://github.com/searxng/searxng/issues/2011#issuecomment-1553317619


unixfox:
Can you print this documentation link in the error logs when searxng encounters an issue related to qwant?

return42:
The described method to solve a CAPTCHA from server's IP is not only related to qwant .. you can use this method to solve IP blocking for whatever engine is blocked (rated as a bot by the origin search engine).

Our CAPTCHA detection is not precise enough to print this message when a CAPTCHA is detected; by example when qwants response a CAPTCHA we get a `qwant ( server API error )` / compare #2011.

unixfox:
Hum no it won't work on most engines, most engines requires passing the cookie containing the captcha validation. That's actually impressive that qwant doesn't require that.

So I suggest to still print a message when an error is occurring in the qwant engine.

return42:
> most engines requires passing the cookie containing the captcha validation.

I know google requires a cookie but AFIK ddg does not have a cookie .. and others?

> So I suggest to still print a message when an error is occurring in the qwant engine.

Lets wait what issue reports come in .. when I have a clear view on more than two or three engines we can print a message whenever it makes sense.

PR #2485, opened 2023-06-03
[POC] limiter: change PING of link_token method from CSS to <img>

return42:
while PR #2357 [1] was being implemented the question came up:

    would be better to change the PING resource from CSS to an image so that
    some terminal based browser may still able to pass the test [1]

This patch implements a POC in where a ```` tag is loaded instaed a CSS.

To test this patch activate limiter and link_token method [3] and start a developer instance::

    make run

In your terminal browser open http://127.0.0.1:8888/search?q=foo

If the browser is suitable for the link_token method, it loads the image and the following messages appear::

    DEBUG   searx.botdetection.limiter    : OK 127.0.0.1/32: /clientft61aak7fzyu6o6v.svg ...
    DEBUG   searx.botdetection.link_token : token is valid --> True
    DEBUG   searx.botdetection.link_token : store ping_key for (client) network 127.0.0.1/32 (IP 127.0.0.1) -> SearXNG_limiter.ping[...]

Browsers that do not load images will be blocked: If you try by example::

    lynx http://127.0.0.1:8888/search?q=foo

you will see a WARNING message like::

    WARNING searx.botdetection.link_token : missing ping (IP: 127.0.0.1/32) / request: SearXNG_limiter.ping[...]

----

[1] https://github.com/searxng/searxng/commit/80aaef6c95b572df1fa3a8c30b7fdc1538d7a306
[2] https://github.com/searxng/searxng/pull/2357#issuecomment-1574898834
[3] activate limiter and link_token method

```diff
diff --git a/searx/botdetection/limiter.toml b/searx/botdetection/limiter.toml
index 71a231e8f..7e1dba755 100644
--- a/searx/botdetection/limiter.toml
+++ b/searx/botdetection/limiter.toml
@@ -17,6 +17,6 @@ ipv6_prefix = 48
 filter_link_local = false

 # acrivate link_token method in the ip_limit method
-link_token = false
+link_token = true

diff --git a/searx/settings.yml b/searx/settings.yml
index a82a3432d..e7b983afc 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -73,7 +73,7 @@ server:
   # public URL of the instance, to ensure correct inbound links. Is overwritten
   # by ${SEARXNG_URL}.
   base_url: false  # "http://example.com/location"
-  limiter: false  # rate limit the number of request on the instance, block some bots
+  limiter: true  # rate limit the number of request on the instance, block some bots

   # If your instance owns a /etc/searxng/settings.yml file, then set the following
   # values there.
```

### DIFF
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -17,9 +17,6 @@
   {% else %}
   <link rel="stylesheet" href="{{ url_for('static', filename='css/searxng.min.css') }}" type="text/css" media="screen">
   {% endif %}
-  {% if get_setting('server.limiter') or get_setting('server.public_instance') %}
-  <link rel="stylesheet" href="{{ url_for('client_token', token=link_token) }}" type="text/css">
-  {% endif %}
   {% block styles %}{% endblock %}
   <!--[if gte IE 9]>-->
   <script src="{{ url_for('static', filename='js/searxng.head.min.js') }}" client_settings="{{ client_settings }}"></script>
@@ -86,5 +83,8 @@
   <!--[if gte IE 9]>-->
   <script src="{{ url_for('static', filename='js/searxng.min.js') }}"></script>
   <!--<![endif]-->
+  {%- if get_setting('server.limiter') -%}
+    <img class='invisible' src="{{ url_for('client_token', token=link_token) }}">
+  {%- endif -%}
 </body>
 </html>
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -617,10 +617,12 @@ def health():
     return Response('OK', mimetype='text/plain')
 
 
-@app.route('/client<token>.css', methods=['GET', 'POST'])
+@app.route('/client<token>.svg', methods=['GET', 'POST'])
 def client_token(token=None):
     link_token.ping(request, token)
-    return Response('', mimetype='text/css')
+    resp = Response('<svg></svg>', mimetype='image/svg+xml')
+    resp.headers['Cache-Control'] = 'no-store'
+    return resp
 
 
 @app.route('/search', methods=['GET', 'POST'])


unixfox:
I'm not sure if it's a good idea to load an image in SVG format, I would prefer it to be in a more general format like PNG or JPG.

I can see a lot of browsers that do not support SVG: https://caniuse.com/svg (this is just a part of all the existing browsers around the world)

-----

Now that I think of it, the PING resource shouldn't be cached by the browser or the browser may not request it again. You should probably add this header:
```
Cache-Control: no-store
```

return42:
> I can see a lot of browsers that do not support SVG: https://caniuse.com/svg (this is just a part of all the existing browsers around the world)

Not sure what browsers do not support SVG and what "SVG (basic support)" means: 

    Method of displaying basic Vector Graphics features using the embed or object elements. 

I assume most (all?) of the browser load and render SVGs like they load and render other images (e.g. PNG images) from a `` tag  / these browser have a svg images rendering library, but you can't select `<svg>` with CSS selectors since these XML tags are not in the elements tree. 

They do not have a XML namespace for `xmlns:svg="http://www.w3.org/2000/svg"` but this is not related to this PR which implements a `` tag which is a part of `xmlns="http://www.w3.org/1999/xhtml"`

return42:
> Now that I think of it, the PING resource shouldn't be cached by the browser or the browser may not request it again. You should probably add this header:
> 
> ```
> Cache-Control: no-store
> ```

Fixed in last force push --> [diff](https://github.com/searxng/searxng/compare/ec3eead721d001fec0ca61bce401c030028c0d4c..08ef986e5864616bc9ce6eebc0069b204f3e300f)

However, the more I look into it, the more doubts I have this change from CSS to `` will bring any improvement at all for users of (e.g.) terminal browser.  Browsers like lynx will neither request the CSS nor the `` tag while more modern terminal browser like browsh do request CSS & `` ...

- https://github.com/searxng/searxng/issues/2486

unixfox:
Well maybe not lynx but others do like w3m: https://askubuntu.com/a/1248415, same for kitty: https://github.com/kovidgoyal/kitty and netsurf according to this comment: https://github.com/browsh-org/browsh/issues/86#issue-339769124

I'm really not a terminal based browser user, so I have very little experience, but that's what I found from a few searches.

return42:
see https://github.com/searxng/searxng/issues/2486#issuecomment-2452887830

Issue #2515, opened 2023-06-24
Google is actively blocking SearXNG instances - SearxEngineTooManyRequestsException: Too many requests


unixfox:
# Description about the issue

It appears that Google is now trying to block SearXNG instances. If you install a fresh SearXNG instance and then perform 5 searches in a row, your instance gets instantly blocked.

The used technique is probably TLS and/or HTTP2 fingerprinting, as restarting SearXNG brings back the ability to perform queries because on startup, like implemented in #2266 SearXNG rotate the ciphers list. There is also the possibility that Google think that SearXNG is some kind of botnet and if one SearXNG instance is doing too many requests or sending "bots" queries, then it affects all the other legitimate SearXNG instances.

Enabling the new advanced limiter (#2357) is probably not going to solve anything, as there will still be public instances without the limiter enabled or people maliciously using SearXNG locally/privately.

I do not have any metrics for the global workability of the Google engine over the time (maybe it would be a good thing to monitor?), but currently it seems like Google now works on a lot less SearXNG public instances than before. Only 25 instances out of 91 have Google engine working ([saved archive](https://archive.is/sMUvA)).
According to web archive, a few days ago (19th June) it was 66 out of 91: http://web.archive.org/web/20230619001339/https://searx.space/

# Possible solutions for fixing this issue

- Periodically rotate the cipher list, like every 5 requests. May not work very well because there is a finite amount of ciphers that you can rotate to.
- Bring back the old ability to not use the mobile UI (this worked fine on private/local SearXNG instance) and set it by default like before. And hope it will reduce the amount of requests to that endpoint from SearXNG instances and thus Google will rate limit less the SearXNG instances.  
   Now that Google is aware, I don't think this will help except maybe for private/local SearXNG instance where the default endpoint (not mobile UI) worked very good.
- Try to look like as close as possible to a Firefox browser in terms of TLS fingerprint (and maybe HTTP2 fingerprint). I shared in https://github.com/searxng/searxng/pull/2266#issuecomment-1581162826 a couple of python libraries for doing that.
- Use HTTP3

I'll in the coming days try to find other solutions, understand a bit better the root cause, as this deeply affect me because I'm used to only use Google engine on SearXNG for the great results...

# Opinion about switching off the ability to switch between normal Google endpoint and the mobile UI endpoint

[...]

Issue #2513, opened 2023-06-23
Unable to set additional parameters for google engine


unixfox:
If you try to set a proxy for btdigg for example it will send the request to a proxy:
```
- name: btdigg
    engine: btdigg
    shortcut: bt
    proxies:
       all://:
        - socks5h://127.0.0.1:9050
```

But for google it doesn't work:

```
- name: google
    engine: google
    shortcut: go
    proxies:
      all://:
        - socks5h://127.0.0.1:9050
```

You can also try any other parameter like `enable_http2: false`, HTTP2 will still be used.

I'm not sure if it affects other engines too.

@return42 do you know what is going on?

return42:
About `enable_http2`: this option can't be set in the engine.  This option can (only) be set in the [`outgoing:` section](https://docs.searxng.org/admin/engines/settings.html#outgoing).

AFAIK its the same with the `proxies:` setting .. or did I miss something?

return42:
Not sure if this issue is related to https://github.com/searxng/searxng/issues/762 ..

unixfox:
Well if it can help you, it did work at the time of this random commit: https://github.com/searxng/searxng/commit/53a4a90d0c9bb69b995e3957bbe77aaba22370d2

I'll try to pinpoint the commit where it broke.

return42:
I assume its broken (the google engine) since 2499899554 .. in this commit the import of the name `network`

https://github.com/searxng/searxng/blob/2e4a435134e0f677fbe24853dd81453a54770674/searx/engines/google.py#L26

has been moved from a function level to the module level .. 

This is what I pointed out in my comment of #762 .. but we have several other modules that do import this name .. I think its more a problem in general that we have conflicts on names in the module level of an engine module.

return42:
@unixfox can you please test (review) https://github.com/searxng/searxng/pull/2517 to see if it fix the issue you reported here in this thread / thanks :+1:

unixfox:
Solved in #2517

Issue #762, opened 2022-01-15
engine doesn't use the defined network with `import searx.network` 


dalf:
Each engine has its own [network](https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/network/network.py#L37). See the [searx.network.network.initialize](https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/network/network.py#L298) function.

Some engine sends additional requests.
For example [seznam](https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/engines/seznam.py#L33).

The code:
```python
from searx.network import get

...
  response_index = get(base_url, headers=params['headers'], raise_for_httperror=True)
```

SearXNG uses the `seznam` network as intended.

Another way to written the code:
```python
from searx import network

...
  response_index = network.get(base_url, headers=params['headers'], raise_for_httperror=True)
```

In this second case, SearXNG uses the `default` network. This should not happen.

---

I guess it is related to the usage of `threading.local()`:
https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/network/__init__.py#L43

https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/network/__init__.py#L65-L70

https://github.com/searxng/searxng/blob/32874dbf7f1a538112345e604bc3f898e4c62734/searx/network/__init__.py#L105

return42:
> In this second case, SearXNG uses the default network. This should not happen.

For me, the `from searx.network import get` workaround is good enough.

return42:
@dalf as far I can say there is another conflict with the name `network` in an engine module / engine setting.

For example the google engine (and several other engines) import the name `network`:

https://github.com/searxng/searxng/blob/2e4a435134e0f677fbe24853dd81453a54770674/searx/engines/google.py#L26

In the settings of an engine there is also a setting with name `network` --> https://docs.searxng.org/admin/engines/settings.html#engine-settings .. In the serax.network.network module there is a `iter_networks()` function that seems to use this name ..

https://github.com/searxng/searxng/blob/2e4a435134e0f677fbe24853dd81453a54770674/searx/network/network.py#L366

I am just at the beginning of my analyses of the bug reported by @unixfox in Related: https://github.com/searxng/searxng/issues/2513

[...]

PR #2517, opened 2023-06-25
[fix] engine & network issues / documentation and type annotations

return42:
This patch fixes some quirks and issues related to the engines and the network. Each engine has its own network and this network was broken for the following engines[1]:

- archlinux
- bing
- dailymotion
- duckduckgo
- google
- peertube
- startpage
- wikipedia

Since the files have been touched anyway, the type annotaions of the engine modules has also been completed so that error messages from the type checker are no longer reported.

The documentation of the engine and outgoing options has also been fixed and completed.

Related and (partial) fixed issue:

- [1] https://github.com/searxng/searxng/issues/762#issuecomment-1605323861
- [2] https://github.com/searxng/searxng/issues/2513
- [3] https://github.com/searxng/searxng/issues/2515

### DIFF
--- a/docs/admin/engines/settings.rst
+++ b/docs/admin/engines/settings.rst
@@ -397,14 +397,26 @@ Communication with search engines.
   Global timeout of the requests made to others engines in seconds.  A bigger
   timeout will allow to wait for answers from slow engines, but in consequence
   will slow SearXNG reactivity (the result page may take the time specified in the
-  timeout to load). Can be override by :ref:`settings engine`
+  timeout to load).  Can be override by ``timeout`` in the :ref:`settings engine`.
 
 ``useragent_suffix`` :
   Suffix to the user-agent SearXNG uses to send requests to others engines.  If an
   engine wish to block you, a contact info here may be useful to avoid that.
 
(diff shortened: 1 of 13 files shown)


docs/admin/engines/settings.rst
```diff
@@ -429,15 +441,6 @@ Communication with search engines.
   Number of retry in case of an HTTP error.  On each retry, SearXNG uses an
   different proxy and source ip.
 
-``retry_on_http_error`` :
```
unixfox, line 432: Why removing this documentation? It still works per engine right?

unixfox:
Apart from my comment about `retry_on_http_error`, the ability to set proxies, disable HTTP2 now works again. Thank you

docs/admin/engines/settings.rst
return42: This documentation has been moved from the `outgoing:` section to the [`engines:` section](https://github.com/searxng/searxng/pull/2517/files/cc953f8efb6b219f4bdda3c7a908dbb9177dfbb3#diff-ac4f9c85d7e2e0f93a7c75f1b89a274590e65c98fa0c4f0b3da13daaa248efd9R636) (moved to line 636).

Issue #2345, opened 2023-04-15
Bug: z-library engine


fabsau:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2023.4.15+0adfed19

**How did you install SearXNG?**
Docker

**What happened?**
Z-Library was seized and now they are back to the clearnet. The old engine is not adapted to work with the updated z-library page. Every user gets a custom domain and has to signin using their password.
I tested it with the following config:
    base_url: https://newprivatedomain.example.com/s/{query}
    password: mysecretpassword

It would be nice if searxng could update the engine. Thank you very much in advance

Edit: More info about z-library:
https://www.reddit.com/r/zlibrary/comments/zx6z62/megathread_how_to_access_zlibrary_on_tor_app

return42:
I would give it a try, but I don't want to register .. do you have a domain & password for me /  send me an eMail markus.heiser@darmarit.de

fabsau:
> I would give it a try, but I don't want to register .. do you have a domain & password for me / send me an eMail [markus.heiser@darmarit.de](mailto:markus.heiser@darmarit.de)

Mail sent. Thank you!

return42:
> Mail sent. Thank you!

@fabsau thanks :-)  

.. I did a short analysis and I see few drawbacks:

1. With the login a session is opened and the session data (cookie) hast to send with each request .. this is a technical requirement that could still be managed with a few lines of code.  However, I see a disadvantage for the user that we cannot solve:

2. The Z-library links in the Searx result list would be of no use to the user: He would have to log in again at the destination address to be able to load the content. (the download request would need the same cookie) .. This problem could be solved theoretically by implementing a proxy for z-library (like for images).  But even then there would be another problem: 

3. The downloads would all go through this one account and the download rate for one account is limited.

I haven't had time to look at z-library via tor yet, I don't know if you need a login there too ... if not, then maybe we should look for a solution that goes via tor.

fabsau:
> > Mail sent. Thank you!
> 
> @fabsau thanks :-)
> 
> .. I did a short analysis and I see few drawbacks:
> 
>     1. With the login a session is opened and the session data (cookie) hast to send with each request .. this is a technical requirement that could still be managed with a few lines of code.  However, I see a disadvantage for the user that we cannot solve:
> 
[...]

PR #2519, opened 2023-06-25
[feat] zlibrary engine use zlibrary-global.se

paolobasso99:
## What does this PR do?

This PR:
1. re-enables z-library as the new domain zlibrary-global.se is now available from the open web. The announcement of the domain: https://www.reddit.com/r/zlibrary/comments/13whe08/mod_note_zlibraryglobalse_domain_is_officially/ (It is an official domain, it requires to log in to the "personal" subdomain only to download files, but the search works.)
2. changes the result template of zlibrary to paper.html, filling the appropriate fields
3. implements language filtering for zlibrary
4. implement zlibrary custom filters (engine traits)
5. refactor and document the zlibrary engine

z-library is very useful and it is back.

Search with !zlib

## Related issues

https://github.com/searxng/searxng/issues/2345#issuecomment-1517319202


paolobasso99:
Converting to draft, I will try implementing engine traits.

paolobasso99:
Done, I have updated the initial PR comment accordingly.

return42:
@paolobasso99 I rebased your branch on SearXNG's master branch and squashed the commits .. ut I haven't finished my review .. I will continue the next days ...

return42:
@paolobasso99 great work :+1: .. sorry for the git-push mess from mine .. I was in a hurry the last days, but now I finished my review.

To summarize what I have done:

- I rebased your branch on SearXNG's master branch and squashed your commits into one (see comment above)
- I placed a WIP commit on top of your branch

In the WIP commit:

1. Move the .rst to the new location ... in #2537 I reordered the docs / engines documented now [here](https://docs.searxng.org/dev/engines/index.html)

2. response: added i18n to terms "Language", "Book rating" and "File quality" (in the UI of z-library its named "File quality" not "Book quality")
3. response: tried do condense the code for better readability .. "readability" is in the eye of the beholder, the new code is strongly influenced by my subjective views.

Can you please have a look at the WIP commit / is 3. OK for you? If you think different, we can revert it.

Otherwise I would squash my WIP commit into you commit and merge this PR into master branch.

----

Further: what do you think about the following points:

- I have noticed timeouts, e.g. when I search for terms like YBCO `!zlib ybco :en`.  But I don't know if it would be good to increase the timeout from 7 to e.g. 10.
- The z-library is currently in the category 'files', I don't know if it would make sense to add it to another category like 'general' (the drawback is the long timeout).

I have left it as it is for now, but I would be interested in your opinion.

paolobasso99:
Changes are ok for me, thanks.
It makes sense to increase the timeout but, honestly, I don't know exactly what are the implications of doing so.
About adding the engine to other categories, in my personal use, I prefer to have it only on files but that is just me.

return42:
> It makes sense to increase the timeout but, honestly, I don't know exactly what are the implications of doing so.

Let's look at the process flow

1. User sends search term to SearXNG
2. SearXNG sends this request **in parallel** to a "group of engines" and **waits** for the responses
3. SearXNG aggregates the results from engines and sends a result list back to the user

In 2. we have the bottleneck, SearXNG has to wait until the last engine has answered ... but what if there is one among the engines that never response or it takes extremely long until it finally response?

To limit the impact of this bottleneck on the process flow, each engie has a timeout.  If it does not respond within this period, its response is ignored and the process flow can continue once the maximum timeout has expired.

The "group of engines" (by example) is built up by the active engines in category "general". If the user wants to be more selective, he can select engines or groups of engines directly via the search syntax [(single bang)](https://docs.searxng.org/user/configured_engines.html) in addition to the categories that are selectable as tabs in the UI.

In the SearXNG default setup we have an priority on a fast response in the category "general".  Engines in this category that are active by default have a maximum timeout of 3sec. This ensures that the user receives a result list after approx. 3sec at the latest (as long he is in the "general" category).

There are several other engines in this category that the user can activate, but these have longer timeouts and the user must be prepared that his search queries will usually take longer than in the default configuration.

-----

In the light of the explanation above, it may also become clearer why I consider the activation of all engines (including the saving of this setting) to be wrong and why I rejected https://github.com/searxng/searxng/issues/2536#issuecomment-1616143729

paolobasso99:
Understood, in this case, I would leave z-lib in the files category only.

PR #2535, opened 2023-06-30
theme: fix "#backToTop" button always being clickable

KiruPoruno:
## What does this PR do?

The button to scroll back to the top is always clickable, even if you've not scrolled down, it'll just have it's opacity set to `0`. This means hovering over where the button is, will cause the cursor/pointer to become that of the `pointer` style, yet nothing will happen when clicking, as you're already at the top.

I can imagine that a button having pointer events when it does nothing could trigger some problems in terms of accessibility clients. Further there's no good reason why it should be able to clicked or hovered (from what I know).

 1. Start the dev server (`make run`)
 2. Make a search query
 3. Scroll slightly
 4. Hover over the "back to top" button
 5. Scroll back to top, whilst keeping the mouse still
 6. The button is no longer clickable, and the cursor changes accordingly

Do the same before this PR, and it'll then still be clickable and hoverable, and have it's cursor indicate it's clickable.


mrpaulblack (approved):
Thanks for the PR, and sorry for the long wait while reviewing. The change works great!
I split up your commit in the actual change and another commit for the static files of the theme that got generated from that change. Next time please run `make static.build.commit` to commit the static files of the theme (see -> https://docs.searxng.org/dev/quickstart.html) :D

LGTM :+1:

PR #2555, opened 2023-07-17
[feat] engine: implementation of Piped

### LINKED ISSUE #952: piped video search engine
Hello, like invidious is it possible to add piped on engines videos selection please ?
> https://github.com/TeamPiped/Piped

Thank you.

Bnyro:
## What does this PR do?
* Add support for the Piped API for searching videos

* Works similar to Invidious and the tracking issue for adding Piped didn't seem to be rejected, so I thought it'd be worth adding

1. Move to preferences -> engines -> video
2. Uncheck all engines and enable Piped
3. Search something in the video category

## Related issues

closes #952

cc @FireMasterK

### DIFF
--- a/docs/dev/engines/online/piped.rst
+++ b/docs/dev/engines/online/piped.rst
@@ -1,0 +1,13 @@
+.. _piped engine:
+
+=====
+Piped
+=====
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.piped
+  :members:
--- a/searx/engines/piped.py
+++ b/searx/engines/piped.py
@@ -32,6 +32,18 @@ named `piped` and are used by all piped engines
     shortcut: ppdm
     piped_filter: music_songs
     ...
+
+Known Quirks
+============
+
+The implementation to support :py:obj:`paging <searx.enginelib.Engine.paging>`
+is based on the *nextpage* method of Piped's REST API / the :py:obj:`frontend
+API <frontend_url>`.  This feature is *next page driven* and plays well with the
+:ref:`infinite_scroll <settings ui>` setting in SearXNG but it does not really
+fit into SearXNG's UI to select a page by number.
+
+Implementations
+===============
 """
 
 from __future__ import annotations
@@ -54,7 +66,7 @@ about = {
 
 # engine dependent config
 categories = []
-paging = False
+paging = True
 
 # search-url
 backend_url: list | str = "https://pipedapi.kavin.rocks"
@@ -95,18 +107,29 @@ def _frontend_url() -> str:
 
 def request(query, params):
 
-    query = urlencode({'q': query})
-    params["url"] = _backend_url() + f"/search?{query}&filter={piped_filter}"
+    args = {
+        'q': query,
+        'filter': piped_filter,
+    }
 
+    path = "/search"
+    if params['pageno'] > 1:
+        # don't use nextpage when user selected to jump back to page 1
+        nextpage = params['engine_data'].get('nextpage')
+        if nextpage:
+            path = "/nextpage/search"
+            args['nextpage'] = nextpage
+
+    params["url"] = _backend_url() + f"{path}?" + urlencode(args)
     return params
 
 
 def response(resp):
     results = []
 
-    search_results = resp.json()["items"]
+    json = resp.json()
 
-    for result in search_results:
+    for result in json["items"]:
         publishedDate = parser.parse(time.ctime(result.get("uploaded", 0) / 1000))
 
         item = {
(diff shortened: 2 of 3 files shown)


unixfox:
Note:

- https://pipedapi.kavin.rocks: using cloudflare
- https://pipedapi-libre.kavin.rocks: not using cloudflare
- https://pipedapi.adminforge.de: not using cloudflare

Bnyro:
> Note:
> 
>     * https://pipedapi.kavin.rocks: using cloudflare
> 
>     * https://pipedapi-libre.kavin.rocks: not using cloudflare
> 
>     * https://pipedapi.adminforge.de: not using cloudflare
>
Are there any preferences or limitations from searx's side concerning using cloudflare or not using it?

return42:
> Are there any preferences or limitations from searx's side concerning using cloudflare or not using it?

AFAIK Cloudflare has a CAPTCHA in front ..

unixfox:
We have some engines that are behind cloudflare but it's not by choice.

If we can not use cloudflare that's much better. We thrive to avoid the users using by default services hosted on cloudflare.

For example on searx-instances we banned all usage of instances hosted at cloudflare.

`frontend_url` should be then switched to the non cloudflare URL.

> > Are there any preferences or limitations from searx's side concerning using cloudflare or not using it?
> 
> AFAIK Cloudflare has a CAPTCHA in front .. 

That's not entirely true. It's only when cloudflare detect the user as a bot based on multiple factors. And this can be disabled per website, something that kavin (owner of piped.video) probably did.

Bnyro:
> We have some engines that are behind cloudflare but it's not by choice.
> 
> If we can not use cloudflare that's much better. We thrive to avoid the users using by default services hosted on cloudflare.
> 
> For example on searx-instances we banned all usage of instances hosted at cloudflare.
> 
> `frontend_url` should be then switched to the non cloudflare URL.
> 
That makes sense, unfortunately I don't think Kavin is hosting a non-nightly frontend somewhere that's not behind Cloudflare, but we can still use any other public frontend (I'll discuss that with Kavin privately).

> That's not entirely true. It's only when cloudflare detect the user as a bot based on multiple factors. And this can be disabled per website, something that kavin (owner of piped.video) probably did.
>
Indeed, Kavin's instances don't use any Captchas.

Bnyro:
We now use <https://srv.piped.video> for the frontend, which doesn't use Cloudflare.

unixfox (approved):

return42:
@Bnyro I'm sorry for the delay / your PR is perfect but I have a suggestion in mind I need to implement first .. I hope I have time next weekend .. thanks for being patient with me :+1:

return42:
Hi @Bnyro, I pushed two commits on top of your branch.  In the first I added some documentation and in the second I split the engine into two, one for videos, one for music.

The second commit is WIP .. don't worry about the errors reported by CI .. I haven't had much time, but I will continue my work next week .. in the meantime you could have a look at the commit with my suggestions ..

I think there is some space for improvements .. I also want to implement the next-page feature / by using the `nextpage` value from the JSON response:

image

But for now I have to go .. / sorry.

return42:
Hi @Bnyro, I implemented ``nextpage`` feature but still have problems with / I will continue the next days ..

Bnyro:
Any further page must be fetched from `/nextpage/search` instead of `/search`. `nextpage` must be passed as a query parameter, as well as the filter and query (`q`), similar to the normal search request.
That's not part of the API docs yet so I wasn't sure if you're aware of it.
I might be able to finish this PR too, I'll see whether I can make it work when I find the time before you.

Bnyro:
Pagination seems to work now from my tests, it's been as simple as using `nextpage/search` instead of `search` for additional pages :)

unixfox:
(Note: Would be interesting to add these features to the invidious engine for feature parity 😃)

return42:
> (Note: Would be interesting to add these features to the invidious engine for feature parity smiley)

@unixfox I think there is more we can improve .. whenever I look into I think we should review the invidious engine .. but then I forget it :-) --> Do you like to open an issue for a review / thanks!

----

@Bnyro I do not have much time this week / but I will review this PR in the next days.

----

NOTE: **next page** versus **page by number**

This feature is *next page* driven .. it does not really fit into SearXNG's schema to select a page by number.

return42:
@Bnyro I added a doc about the *next page* versus *page by number* quirk .. to soften the quirk a bit I implemented a condition not use the nextpage token when the users selects the first page (page no. 1).

Further in my tests I have seen some results had been ignored by SearXNG since the `content` field was set to `None` .. its fixed in 898ede0

If you think my modifications are OK and you do not have further suggestion I would like merge the PR .. what do you think, is PR ready to merge?

Bnyro:
The changes look good to me, thanks for continuing work on it!

return42:
> thanks for continuing work on it!

I thank you, your PR was really well done :+1:

PR #2557, opened 2023-07-19
[fix] remove disabled: false from engine definitions in settings.yml

mrpaulblack:
## What does this PR do?

Remove `disabled: false` from engine definitions in `settings.yml`.

This setting is not needed, by default engines are disabled and when I want to enable engiens in my downstream patch repo I first need to remove this statement. This basically makes the sed comnmands downstream for me easier -> https://github.com/paulgoio/searxng/blob/dc58b8c4dd0daf0f2309fc858a480c69226c80f9/Dockerfile#L56

`make run`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -626,7 +626,6 @@ engines:
     first_page_num: 1
     shortcut: et
     categories: [dictionaries]
-    disabled: false
     about:
       website: https://www.etymonline.com/
       wikidata_id: Q1188617
@@ -1204,7 +1203,6 @@ engines:
     engine: qwant
     shortcut: qw
     categories: [general, web]
-    disabled: false
     additional_tests:
       rosebud: *test_rosebud
 
@@ -1213,7 +1211,6 @@ engines:
     engine: qwant
     shortcut: qwn
     categories: news
-    disabled: false
     network: qwant
 
   - name: qwant images
@@ -1221,7 +1218,6 @@ engines:
     engine: qwant
     shortcut: qwi
     categories: [images, web]
-    disabled: false
     network: qwant
 
   - name: qwant videos
@@ -1229,7 +1225,6 @@ engines:
     engine: qwant
     shortcut: qwv
     categories: [videos, web]
-    disabled: false
     network: qwant
 
   # - name: library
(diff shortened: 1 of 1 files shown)


return42 (approved):
> This setting is not needed, by default engines are disabled 

May be you are confused by double negation? .. By default engines are *not disabled* :-)

https://github.com/searxng/searxng/blob/51c531d4503f6ebb96a90cfba2b7c263a220164b/searx/engines/__init__.py#L42

.. anyway, the settings are superfluous and can be deleted :+1:

bucketcat:
@return42 I think this commit has caused the GUI toggles for enabled/disabled to become inverted for me. So setting an engine to "Allow: " as checked will cause the opposite behaviour. Could be easily solved by adding a "Not" before the Allow label. But currently, it does the opposite. Can also be verified by looking at the parametrized settings under the Cookie section.

return42:
> I think this commit has caused the GUI toggles for enabled/disabled to become inverted for me.

I recommend to drop the cookies / further see https://github.com/searxng/searxng/discussions/2591#discussioncomment-6631859

> Can also be verified by looking at the parametrized settings under the Cookie section.

On my instance I can't reproduce your issue / by example I disabled a engine [patched by this PR](https://github.com/searxng/searxng/pull/2557/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fL629)

image

and here is the cookie content:

image

bucketcat:
@return42 Cheers, I think that there might be some issues with firefox Browser cache being either corrupted or tampered with. Sadly, browser cache is the only thing I can't clear, for firefox, as there's a recent bug that disables the ability to selectively clear cache (per-site basis). I had learned this after some nasty data loss where it all got wiped. Also, cache is cleared every firefox update by itself, for some reason. Not sure if that information is of any use to you.

But since I am aware of the settings being inverted, and can't clear cache in this specific browser, and really don't want to go through every possible bool that could have been flipped, I can simply adjust my workflow as the issue itself is relatively minor ;). I'll get around to doing a cache backup later, and can try then.

It's not related to the cookies, I cleared them together with DOM, for what it's worth.

Issue #966, opened 2022-03-16
Please allow engines ( Wolfram Alpha) from "Other" category by default for search results


geckolinux:
Hi there, I enjoy the improvements made to the SearXNG theme compared to the original SearX "oscar" theme. But I think that there is a usability/feature regression in SearXNG by moving some engines to the "Other" category and only making them accessible via bangs, especially Wolfram Alpha. I personally don't like bangs, and for many searches I need the excellent results that Wolfram Alpha gives. With the original SearX I always enabled the "Science" category by default and in that category disabled all the engines except for Wolfram Alpha. This always gave me a very informative and useful sidebar. With SearXNG it could be even better than that by allowing the "Other" category by default and keeping the "Science" category available when needed for actual scientific queries. Thanks for taking this into account!

![Screenshot from 2022-03-16 09-17-00](https://user-images.githubusercontent.com/2692138/158612638-fdb1695f-2f0e-4964-8ae7-ef9863d24878.png)

![Screenshot from 2022-03-16 09-18-07](https://user-images.githubusercontent.com/2692138/158612666-38ad6922-49a7-4200-b5b5-b9be3c9ad2c6.png)

![Screenshot from 2022-03-16 09-19-49](https://user-images.githubusercontent.com/2692138/158612680-26347833-718f-4f60-9977-7710ae0a9727.png)

![Screenshot from 2022-03-16 09-30-46](https://user-images.githubusercontent.com/2692138/158613523-3d452306-f95c-437a-b25c-7db6a71c27ed.png)

geckolinux:
I'm not very good coding, but I could probably find a way to add Wolfram Alpha back to the Science category. If I could do that would a pull request even be considered? I'm just asking before I dedicate a lot of time to it. Thanks a lot.

mrpaulblack:
Hi @geckolinux ,
First sorry for the long wait for an answer. You already read our reasoning for moving wolframalpha to the other "category" here: https://github.com/searxng/searxng/pull/813. But at the same time I also understand your argument of not wanting to use bangs for wolframalpha all the time.

Maybe we can find a compromise. Instead of moving it back to the science category we can move it to a different category or create a new one for the current other category. But whatever we do I would like to keep it disabled by default upstream since its really slow...

What you can do for now is move it to the science category in your instances settings.yml file :+1:

geckolinux:
Hi there @mrpaulblack , appreciate the response, and thanks for taking into account this usage scenario!

[...]

PR #2570, opened 2023-07-27
Move Wolfram Alpha into General category and disable by default

geckolinux:
As suggested in https://github.com/searxng/searxng/pull/813#issuecomment-1574851572 as a good compromise between https://github.com/searxng/searxng/pull/813#issuecomment-1020581765 and https://github.com/searxng/searxng/issues/966#issue-1171107451.

This will help users that want Wolfram Alpha instant answers by default with every query and don't mind the slowness. It will be disabled by default so it's fast for all users, and those that need it can enable it as a user-configurable option without requiring setting up their own instance to change the Wolfram Alpha categorization.

## Would close:
https://github.com/searxng/searxng/issues/966

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1602,7 +1602,8 @@ engines:
     # Or you can use the html non-stable engine, activated by default
     engine: wolframalpha_noapi
     timeout: 6.0
-    categories: []
+    disabled: true
+    categories: general
 
   - name: dictzone
     engine: dictzone


unixfox:
Can you edit the PR title with a better title?

Same for the commit name.

Thank you

geckolinux:
> Can you edit the PR title with a better title?
> 
> Same for the commit name.

Ah, sorry about that. I changed the PR title but it doesn't let me change the commit name (I'm just using the Github web interface).

unixfox:
> > Can you edit the PR title with a better title?
> > 
> > Same for the commit name.
> 
> Ah, sorry about that. I changed the PR title but it doesn't let me change the commit name (I'm just using the Github web interface).

Use the git command line then...

PR #2576, opened 2023-07-28
add bt4g engine and disable by default btdigg

### LINKED ISSUE #2553: [Engine request] bt4g
**Working URL to the engine**

https://bt4gprx.com/

bt4g is a BitTorrent DHT search engine it has a better UI than [btdig](https://btdig.com), and it shows the number of Seeders and Leechers, No annoying google captchas, its faster in crawling new torrents
**Features of this engine**

fast, good UI, simple to use, showes the numbre of Seeders and Leechers

**Additional context**

unixfox:
## What does this PR do?

This adds the bt4g engine.

Implement engine requested in https://github.com/searxng/searxng/issues/2553

I did not parse the HTML page because there is an API in XML, but it provides fewer data like amount of seeders/leechers and the files in the torrent file. It's a tradeoff for a "stable" engine as the XML content will change way less than the HTML page.

## Screenshot 

image

## Related issues

Closes https://github.com/searxng/searxng/issues/2553

### DIFF
--- a/searx/engines/bt4g.py
+++ b/searx/engines/bt4g.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""
+ BT4G (Videos, Music, Files)
+"""
+
+import re
+from datetime import datetime
+from urllib.parse import quote
+
+from lxml import etree
+
+from searx.utils import get_torrent_size
+
+# about
+about = {
+    "website": 'https://bt4gprx.com',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'XML',
+}
+
+# engine dependent config
+categories = ['files']
+paging = True
+time_range_support = True
+
+# search-url
+url = 'https://bt4gprx.com'
+search_url = url + '/search?q={search_term}&orderby={order_by}&category={category}&p={pageno}&page=rss'
+bt4g_order_by = 'relevance'  # relevance, size, seeders, time
+bt4g_category = 'all'  # all, audio, movie, doc, app, other
+
+
+def request(query, params):
+
+    order_by = bt4g_order_by
+    if params['time_range']:
+        order_by = 'time'
+
+    params['url'] = search_url.format(
+        search_term=quote(query),
+        order_by=order_by,
+        category=bt4g_category,
+        pageno=params['pageno'],
+    )
+    return params
+
+
+def response(resp):
+    results = []
+
+    search_results = etree.XML(resp.content)
+
+    # return empty array if nothing is found
+    if len(search_results) == 0:
+        return []
+
+    for entry in search_results.xpath('./channel/item'):
+        title = entry.find("title").text
+        link = entry.find("guid").text
+        fullDescription = entry.find("description").text.split('<br>')
+        filesize = fullDescription[1]
+        filesizeParsed = re.split(r"([A-Z]+)", filesize)
+        magnetlink = entry.find("link").text
+        pubDate = entry.find("pubDate").text
+        results.append(
+            {
+                'url': link,
+                'title': title,
+                'magnetlink': magnetlink,
+                'seed': 'N/A',
+                'leech': 'N/A',
+                'filesize': get_torrent_size(filesizeParsed[0], filesizeParsed[1]),
+                'publishedDate': datetime.strptime(pubDate, '%a,%d %b %Y %H:%M:%S %z'),
+                'template': 'torrent.html',
+            }
+        )
+
+    return results
--- a/searx/engines/bt4g.py
+++ b/searx/engines/bt4g.py
@@ -1,7 +1,40 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
-"""
- BT4G (Videos, Music, Files)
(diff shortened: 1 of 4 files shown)


return42:
@unixfox do you missed to add the patch of the settings.yml to your commit?

unixfox:
@return42 yes sorry I forgot it. please review again :)

searx/engines/bt4g.py
```diff
@@ -0,0 +1,67 @@
...
+    "results": 'XML',
+}
+
+# engine dependent config
+categories = ['files']
+paging = True
+
+# search-url
+url = 'https://bt4gprx.com'
+search_url = url + '/search?q={search_term}&p={pageno}&page=rss'
```
return42: The rss feed does not support paging ...

> I did not parse the HTML page because there is an API in XML,

Not sure if this is a good decision, the RSS feed seems to be sorted "BY TIME" .. I think we need "BY RELVANCE" for a search in SearXNG ..

image

I tried some arguments on the RSS feed but I did not had luck: I didn't find a way to control the RSS result list. --> May be its better to parse the HTML output?

return42: Wait, seems `orderby=`  does work:

- https://bt4gprx.com/search?q=foo&orderby=time&page=rss 
- https://bt4gprx.com/search?q=foo&orderby=relevance&page=rss

return42: @unixfox I pushed a WIP commit on top of your branch / this commit is just to show how it works / what I have tested ..

The option `orderby=` and `category` seem to work in the RSS feed, the only drawback of the RSS feed is the fixed limit of 15 result per query term / not sure if this is good enough for a torrent search.

@Bnyro, @unixfox .. what do you think about this limited result list?

Bnyro: I think that's decent if we enable paging as well. That's possible by passing the page number as `p` query parameter.

For example: https://bt4gprx.com/search?q=foo&orderby=relevance&page=rss&p=2

And yes, ordering by relevance is the very best choice for that engine in my opinion.

Bnyro: (Oh, I just saw that's what you just disabled, then I'll need to rethink my answer ...)

Bnyro: Pagination still seems to work when sorting by relevance, so I don't think there's any drawback if we don't sort by the time.

unixfox: > The rss feed does not support paging ...

That's not true, with the `p` query parameter it does support paging.

I'm okay with setting it to "relevant" type.

About whether using RSS or HTML, if we want to have less breakage and less burner of maintaining this engine I think we should prioritize the RSS type. It provides most of the features we need even though we miss some functionalities.

return42: > That's not true, with the `p` query parameter it does support paging.

Not sure where my issue is but I get always page one in my reader, even I set `p=2`.

> Pagination still seems to work when sorting by relevance

Yes, seems my feed reader is kidding me ...

> I'm okay with setting it to "relevant" type.

I will give it a new try today .. if paging works, we should use XML feed :+1:

searx/engines/bt4g.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+# do search-request
+def request(query, params):
+    params['url'] = search_url.format(search_term=quote(query), pageno=params['pageno'] - 1)
```
return42: > Not sure where my issue is but I get always page one in my reader, even I set p=2.

I found my issue: it is `params['pageno'] - 1` --> there is no page number `0` in bt4g's rss  .. so I get for page one and page two in SearXNG the same result / we have to start with `p=1` --> `pageno=params['pageno']` 

@unixfox  I will fix this issue with some other nits I have found .. coming soon.

return42:
> @unixfox I will fix this issue with some other nits I have found .. coming soon.

I squashed my nits into your commit / see [diff](https://github.com/searxng/searxng/compare/1ed1ef4c5256af91a7736cec2e62f949ec5817fb..4bc1ede0c)  --> NOTE: When *time_range* is activate, the results always orderd by ``time``.

And I added commit 1f5b1fc12 on top to document the engine.

@unixfox @Bnyro from my side PR is ready to merge .. what do you think?

unixfox (commented):
Good for me

Bnyro (approved):

return42:
FYI: not sure you already know ... 

The documentation of the engines is available from here: https://docs.searxng.org/dev/engines/index.html

In your development branch you can build (& view) the documentation by Makefile targets [`make docs.clean docs.live`](https://docs.searxng.org/dev/contribution_guide.html#documentation).  BTW for most CI and development tasks we have [Makefile targets](https://docs.searxng.org/dev/makefile.html)

Bnyro:
> FYI: not sure you already know ...
> 
> The documentation of the engines is available from here: https://docs.searxng.org/dev/engines/index.html
> 
> In your development branch you can build (& view) the documentation by Makefile targets [`make docs.clean docs.live`](https://docs.searxng.org/dev/contribution_guide.html#documentation). BTW for most CI and development tasks we have [Makefile targets](https://docs.searxng.org/dev/makefile.html)
>
Oh right, that looks great!

Issue #855, opened 2022-02-01
Yacy local mode


Dalethium:
**Describe the solution you'd like**
Adding a boolean config toggle for "global" or "local" here: https://github.com/searxng/searxng/blob/07f7289e23c3ed28bb5e4f71d47e0da554ca9aea/searx/engines/yacy.py#L38

**Describe alternatives you've considered**
I tried setting a tupple via the config to override the URL but it would not work. Currently, I am just manually editing the engine file.

**Additional context**
The yacy P2P search engine has a "local" mode and "global" mode when you are part of the DHT -- this is an important difference as things like loadtime, and content, can be very different. Your local yacy peer may have content that you have **specifically** curated and crawled, whereas the global p2p engine functions more analogous to google. This is all fine and well, however, seeing how it's just a boolean toggle we'd need, it seems like a fair thing to add considering the "nicheness" of the situation is the engine itself, and not the config I am proposing.

allendema:
Can you try this patch?
https://github.com/searxng/searxng/compare/master...allendema:searxng:yacy-modes

return42:
closed by https://github.com/searxng/searxng/pull/2587

PR #2587, opened 2023-08-02
add option to change yacy search mode

zutto:
## What does this PR do?

Removes hardcoded "global" mode from Yacy engine and gives user an option to change the scope of the search instead.

Yacy has two search modes, peer-to-peer search and local peer search. This PR enables the mode to be changed in the configuration.

`search_mode='global'` Does the peer-to-peer network search. This is the default search mode in Yacy.
`search_mode='local'` Does local peer only search.

In Yacy's default UI, this option is basically the top left "peer-to-peer" and "privacy" toggle.
image
And sometimes it's called Stealth Mode, because why not :grin:.
image

Documentation for reference https://wiki.yacy.net/index.php/Dev:APIyacysearch

### DIFF
--- a/docs/dev/engines/online/yacy.rst
+++ b/docs/dev/engines/online/yacy.rst
@@ -1,0 +1,13 @@
+.. _yacy engine:
+
+====
+Yacy
+====
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.yacy
+  :members:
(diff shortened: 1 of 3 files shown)


zutto:
Also found this, looks like someone else has also made this same change earlier. https://github.com/searxng/searxng/issues/855

return42 (changes requested):
Please fix the bugs reported by CI / I recommend to read [Development Quickstart](https://docs.searxng.org/dev/quickstart.html#development-quickstart).

`make format.python` should do the trick

zutto:
done

return42:
Hi @zutto, thanks a lot for your PR, I squashed your two commits and added a 4fb14cd64 for a documentation.

Yacy sounds interesting, sadly the development stopped in 2016 / last release is [1.92](https://github.com/yacy/yacy_search_server/releases/tag/Release_1.92)

Issue #1221, opened 2022-05-21
Wikimedia Commons


julianfairfax:
**Working URL to the engine**
https://commons.wikimedia.org

**Why do you want to add this engine?**
Out of all of the Wikimedia projects, Wikimedia Commons is one of only two to not appear in any engine category in SearXNG, with the other being Wikispecies.

**Features of this engine**
It has a collection of [82,886,704](https://commons.wikimedia.org/wiki/Special:Statistics) [freely usable](https://commons.wikimedia.org/wiki/Commons:Reusing_content_outside_Wikimedia) media files.

**How can SearXNG fetch the information from this engine?**
`https://commons.wikimedia.org/w/index.php?search=%s` with `%s` being what you want to search.

**Applicable category of this engine**
General, files, images, music, videos.

dalf:
There are various categories: Images, Audio, Video, Other media, Categories and pages.

So I guess we need an engine per category (and most probably one Python module).

---

* commons.wikimedia.org has a SPARQL endpoint but it requires a login
* the result page `https://commons.wikimedia.org/w/index.php?search=time&title=Special:MediaSearch&go=Go&type=image` make a XHR request to display the second page `https://commons.wikimedia.org/w/api.php?action=query&format=json&uselang=en&generator=search&gsrsearch=filetype:bitmap|drawing -fileres:0 time&gsrlimit=40&gsroffset=40&gsrinfo=totalhits|suggestion&gsrprop=size|wordcount|timestamp|snippet&prop=info|imageinfo|entityterms&inprop=url&gsrnamespace=6&iiprop=url|size|mime&iiurlheight=180&wbetterms=label`.

PR #2588, opened 2023-08-03
[feat] implementation of Wikimedia commons for images

Bnyro:
## What does this PR do?
* Add support for the Wikimedia Commons API for searching images

* Wikimedia Commons provides a large amount of images, that are copyleft or free to use
* The API supports much more, such as videos, audio files, and general/text content, but it's most important category are images

1. Move to preferences -> engines -> images
2. Uncheck all engines and enable `wikicommons.images`
3. Search something in the images category

## Related issues

addresses #1221

(example url as note to self: https://commons.wikimedia.org/w/api.php?action=query&generator=search&gsrsearch=filetype:bitmap|drawing%20cat&gsrlimit=20&gsrnamespace=6&format=json&gsrprop=snippet&prop=info|imageinfo&gsrnamespace=6&iiprop=url|size|mime&iiurlheight=180&inprop=url)

### DIFF
--- a/searx/engines/wikicommons.py
+++ b/searx/engines/wikicommons.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Wikimedia Commons (images)
+
+"""
+
+from urllib.parse import urlencode
+
+# about
+about = {
+    "website": 'https://commons.wikimedia.org/',
+    "wikidata_id": 'Q565',
+    "official_api_documentation": 'https://commons.wikimedia.org/w/api.php',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+base_url = "https://commons.wikimedia.org"
+search_prefix = (
+    '?action=query'
+    '&format=json'
+    '&generator=search'
+    '&gsrnamespace=6'
+    '&gsrprop=snippet'
+    '&prop=info|imageinfo'
+    '&iiprop=url|size|mime'
+    '&iiurlheight=180'  # needed for the thumb url
+)
+paging = True
+number_of_results = 10
+
+
+def request(query, params):
+    language = 'en'
+    if params['language'] != 'all':
+        language = params['language'].split('-')[0]
+
+    args = {
+        'uselang': language,
+        'gsrlimit': number_of_results,
+        'gsroffset': number_of_results * (params["pageno"] - 1),
+        'gsrsearch': "filetype:bitmap|drawing " + query,
+    }
+
+    params["url"] = f"{base_url}/w/api.php{search_prefix}&{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    results = []
+    json = resp.json()
+
+    if not json.get("query", {}).get("pages"):
+        return results
+
+    for item in json["query"]["pages"].values():
+        imageinfo = item["imageinfo"][0]
+        title = item["title"].replace("File:", "").rsplit('.', 1)[0]
+        result = {
+            'url': imageinfo["descriptionurl"],
+            'title': title,
+            'content': item["snippet"],
+            'img_src': imageinfo["url"],
+            'img_format': f'{imageinfo["width"]} x {imageinfo["height"]}',
+            'thumbnail_src': imageinfo["thumburl"],
+            'template': 'images.html',
+        }
+        results.append(result)
+
+    return results
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1599,6 +1599,12 @@ engines:
       website: https://www.wikivoyage.org/
       wikidata_id: Q373
 
+  - name: wikicommons.images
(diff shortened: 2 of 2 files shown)


return42:
> Uncheck all engines and enable wikicommons.images

FYI: its not needed to modify the settings, you can access a engine directly by SearXNG's [search syntax](https://docs.searxng.org/user/search-syntax.html) / by example: 

`!wc birds` or `!wikicommons.images birds`

Bnyro:
That's cool, thanks for the tip!

return42:
First we should review https://github.com/searxng/searxng/pull/2589 ..

My intention is to make the [mediawiki engine](https://github.com/searxng/searxng/blob/a96b6f3a31f315a77e65016b8be57fd4f486fc9c/searx/engines/mediawiki.py) as generic as possible to be able to implement engines like this `wikicommons.images` by configuration alone.

I haven't quite thought this through yet, but the [Xpath engine](https://docs.searxng.org/dev/engines/xpath.html) is an example of how this could looks like / I will send a PR with a proposal for a more generic implementation / I think we will find a way, this PR is a good excercise.

searx/engines/wikicommons.py
```diff
@@ -0,0 +1,72 @@
...
+
+    params["url"] = f"{base_url}/w/api.php{search_prefix}&{urlencode(args)}"
+    print("url", params["url"])
+
+    return params
+
+
+def response(resp):
+    json = resp.json()
+    print("json", json)
```
dalf: a leftover of the debug process, I guess

dalf (approved):
LGTM: the engine works well as it is, and the code simple.
It can be improved later on IMO.

searx/engines/wikicommons.py
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    params["url"] = f"{base_url}/w/api.php{search_prefix}&{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    json = resp.json()
+    results = []
+    for pageid in json["query"]["pages"]:
```
Bnyro:
```suggestion
    for item in json["query"]["pages"].values():
```
should work too I think. (but I can't test it)

searx/settings.yml
```diff
@@ -1599,6 +1599,12 @@ engines:
       website: https://www.wikivoyage.org/
       wikidata_id: Q373
 
+  - name: wikicommons.images
+    engine: wikicommons
+    shortcut: wc
+    categories: images
+    number_of_results: 10
```
Bnyro, line 1606: I think we should comment that out by default, or completely remove it?

searx/engines/wikicommons.py
Bnyro: yeah, seems so :)

return42:
> LGTM: the engine works well as it is, and the code simple.
> It can be improved later on IMO.

Its OK for me .. if you think PR is ready for merge .. don't hesitate :)

searx/engines/wikicommons.py
```diff
@@ -50,7 +50,7 @@ def request(query, params):
 def response(resp):
     json = resp.json()
     results = []
-    for pageid in json["query"]["pages"]:
+    for item in json["query"]["pages"].values():
         item = json["query"]["pages"][pageid]
```
Bnyro: This can be dropped now :)

dalf: Done

searx/settings.yml
dalf: Some other engines specify the number_of_results.
I'm inclined to keep this setting without much conviction.

return42:
can we squash the commits?

return42:
search for `!wc fodasdd` ends in:

```
  File "/800GBPCIex4/share/SearXNG/searx/search/processors/online.py", line 165, in search
  File "/800GBPCIex4/share/SearXNG/searx/search/processors/online.py", line 153, in _search_basic
  File "/800GBPCIex4/share/SearXNG/searx/engines/wikicommons.py", line 53, in response
    for item in json["query"]["pages"].values():
```

searx/engines/wikicommons.py
```diff
@@ -0,0 +1,67 @@
...
+        'gsrsearch': "filetype:bitmap|drawing " + query,
+    }
+
+    params["url"] = f"{base_url}/w/api.php{search_prefix}&{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    json = resp.json()
+    results = []
```
Bnyro, line 51:
```suggestion
    results = []
    
    if "query" not in json.keys():
        return results
    
```

return42: give me 5min .. its on the way ;-)

Bnyro: Oh, nice :)

Bnyro: I'm getting more and more experienced in editing code on a mobile phone :p

PR #2589, opened 2023-08-03
[feat] new engine: wikispecies

### LINKED ISSUE #1222: Wikispecies
**Working URL to the engine**
https://species.wikimedia.org

Out of all of the Wikimedia projects, Wikimedia Species is one of only two to not appear in any engine category in SearXNG, with the other being Wikimedia Commons.

**Features of this engine**
It has [803,813](https://species.wikimedia.org/wiki/Special:Statistics) articles, covering [Animalia](https://species.wikimedia.org/wiki/Animalia), [Plantae](https://species.wikimedia.org/wiki/Plantae), [Fungi](https://species.wikimedia.org/wiki/Fungi), [Bacteria](https://species.wikimedia.org/wiki/Bacteria), [Archaea](https://species.wikimedia.org/wiki/Archaea), [Protista](https://species.wikimedia.org/wiki/Protista) and all other forms of life.

`https://species.wikimedia.org/w/index.php?search=%s` with `%s` being what you want to search.

General, science.

Bnyro:
## What does this PR do?
* adds wikispecies as an engine
* wikispecies is just a standard wikimedia api

* it contains articles about different animals/species and other forms of life.

* disable all engines except and enable wikispecies

## Related issues

closes  #1222

### DIFF
--- a/searx/engines/mediawiki.py
+++ b/searx/engines/mediawiki.py
@@ -3,9 +3,10 @@
  General mediawiki-engine (Web)
 """
 
-from json import loads
 from string import Formatter
 from urllib.parse import urlencode, quote
+
+from searx.utils import html_to_text
 
 # about
 about = {
@@ -69,7 +70,7 @@ def request(query, params):
 def response(resp):
     results = []
 
-    search_results = loads(resp.text)
+    search_results = resp.json()
 
     # return empty array if there are no results
     if not search_results.get('query', {}).get('search'):
@@ -86,7 +87,7 @@ def response(resp):
         )
 
         # append result
-        results.append({'url': url, 'title': result['title'], 'content': ''})
+        results.append({'url': url, 'title': result['title'], 'content': html_to_text(result.get('snippet', ''))})
 
     # return results
     return results
--- a/searx/engines/mediawiki.py
+++ b/searx/engines/mediawiki.py
@@ -1,18 +1,59 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- General mediawiki-engine (Web)
-"""
+# lint: pylint
+"""The MediaWiki engine is a *generic* engine to **query** Wikimedia wikis by
+the `MediaWiki Action API`_.  For a `query action`_ all Wikimedia wikis have
+endpoints that follow this pattern::
 
-from string import Formatter
+    https://{base_url}/w/api.php?action=query&list=search&format=json
+
+.. note::
+
+   In its actual state, this engine is implemented to parse JSON result
+   (`format=json`_) from a search query (`list=search`_).  If you need other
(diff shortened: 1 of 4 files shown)


searx/engines/mediawiki.py
```diff
@@ -86,7 +86,7 @@ def response(resp):
         )
 
         # append result
-        results.append({'url': url, 'title': result['title'], 'content': ''})
+        results.append({'url': url, 'title': result['title'], 'content': result.get('snippet', '')})
```
return42: Wow this is *rocket cool*, it (BTW) fixes the content field of wikibooks, wikinews, wikiquote .. and all the other mediawiki engines :+1:

In my review I will strip the HTML tags from the *snippet* .. coming soon.

return42:
@Bnyro I added two commits on top of your branch .. can you please have a look / thanks :+1: 

1. in 5e58ccf8aeb I added some docs and some additional config options for generic wikimedia engines
2. in a96b6f3 I enabled the engine from the wikimedia foundation 

@Bnyro, @unixfox what do you think about the 2nd patch ... is it OK or does it pollute the result list a bit too much?

Bnyro:
That looks awesome in my opinion.
The wikimedia engines provide great results and therefore having them enabled by default is a good decision in my opinion.

return42:
> @Bnyro, @unixfox what do you think about the 2nd patch ... is it OK or does it pollute the result list a bit too much?

After using this setting for a while I'm no longer sure if it was a good decision to activate theses engines by default (a96b6f3a31) / in the *general* category I often get to many results with a ranking on top of the result list. 

By example: when I query for [markdown](https://searx.be/search?q=markdown) I got to many results I didn't was looking for.

Should we better deactivate these engines by default?

---

Or should we rank ([`weight`](https://docs.searxng.org/admin/settings/settings_engine.html)) them lower?

Bnyro:
I think it's better to weight them lower, since they do in fact provide good results.

unixfox:
> > @Bnyro, @unixfox what do you think about the 2nd patch ... is it OK or does it pollute the result list a bit too much?
> 
> After using this setting for a while I'm no longer sure if it was a good decision to activate theses engines by default ([a96b6f3](https://github.com/searxng/searxng/commit/a96b6f3a31f315a77e65016b8be57fd4f486fc9c)) / in the _general_ category I often get to many results with a ranking on top of the result list.
> 
> By example: when I query for [markdown](https://searx.be/search?q=markdown) I got to many results I didn't was looking for.
> 
> Should we better deactivate these engines by default?

I would prefer to deactivate them, I never really got any meaningful results from them. In fact, I have disabled them on searx.be

I think we should be very careful with what engines we add as enabled by default for the General category, some people will not like if we add too many new kind of engines (4 new engines enabled by default in a row is a lot).

Issue #2294, opened 2023-03-30
Implement a python module for https://search.brave.com/


return42:
We have a XPath implementation of the brave engine (see below).  In the past, however, we found that there were always problems with the XPath implementation, and brave also offers more possibilities than we can realize with an XPath implementation.

In the long run we should implement a Python module for https://search.brave.com/ that is more robust and also supports safe-search and localization.

- https://github.com/searxng/searxng/issues/2020
- https://github.com/searxng/searxng/pull/2178#issuecomment-1426650913
- https://github.com/searxng/searxng/issues/392#issuecomment-938085057

----

https://github.com/searxng/searxng/blob/007bc7322723fd8bf1db50820638528a51f6120d/searx/settings.yml#L1827-L1853

----

Related:

- [`!brave !g something` returns no result](https://github.com/searxng/searxng/issues/392#issuecomment-938085057)
- https://github.com/searxng/searxng/issues/1651#issuecomment-1503277191
- https://github.com/searxng/searxng/issues/2590

return42:
implemented by 

- https://github.com/searxng/searxng/pull/2599

RaptaG:
Brave appears to remain broken #2616

return42:
@RaptaG I know, please lets continue in your issue https://github.com/searxng/searxng/issues/2616

PR #2599, opened 2023-08-05
Brave: support for images, videos and news (and WEB)

### LINKED ISSUE #2590: brave image and video search
**Working URL to the engine**

- https://search.brave.com/images?q
- https://search.brave.com/videos?q

Brave search is an independent search engine 

**Features of this engine**
fast, Doesn't track its users, It has its independent own index

images for image search and videos for video search

**Additional context**

- https://github.com/searxng/searxng/issues/2294
- https://twitter.com/brave/status/1687132143250481152
- https://twitter.com/brave/status/1687132143250481152

Bnyro:
## What does this PR do?
* Add support for the brave search engine for news, videos and images

## Architecture
* the PR parses the js data internally used by Brave using Svelte for rendering the result pages
* therefore I added the `chompjs` dependency, since I couldn't find an other way to parse JavaScript objects to Python dicts
* we can't parse the html for images, because rendering is client side via JS. For videos and news, we could use xpath instead in theory

* Brave has its own index and provides decent results

`!brave{img,vid,news} hello world`

## Related issues

addresses #2294
closes #2590


Bnyro:
Oh, we can't use `match` expressions? Damn.

return42:
To fix the issue reported by CI: `make format.python`

You have accidentally included `utils/brand.env` in the commit.

return42:
> Add support for the brave search engine for news, videos and images

FYI: I work on this PR / in my [brave branch](https://github.com/return42/searxng/commits/brave) I added WIP in wich I also want to implement the WEB `/search` (replace XPath engine for brave) .. and paging support .. and time-range support .. and more? ..but it will take its time / I come back when I finished my POC.

> Terefore I added the `chompjs` dependency, since I couldn't find an other way to parse JavaScript objects to Python dicts

In the past I used `json.loads`:

https://github.com/searxng/searxng/blob/64100db9045ec712afbd8fa1becee14f7109921c/searx/engines/duckduckgo.py#L368-L371

--> lets add the `chompjs` dependency :+1: 

but: [chompjs is a library designed to do that as a more powerful replacement of standard `json.loads`](https://github.com/Nykakin/chompjs#chompjs)

return42:
@Bnyro I placed a commit on top of your branch / in this patch I replaced the XPath configuration for the brave (web) search and implemented this engine in your `brave.py` module.

I also added rudimentary search-language and region (country) support .. in the next days I will implement a `fetch_traits(..)` function for a better locale support.

I added a lot of features (time-range, safe-search) and changed some of your code .. can you please have a look / if you have some for me, comments (in the code) are welcome.

The brave engines needs a lot more testing .. if you have time it would be good to have some tests :-)

In the next days I will implement the `fetch_traits(..)`  .. but I have a lot of other work to do, so I can't promise when I will have time for / be patient :-)

Bnyro:
> @Bnyro I placed a commit on top of your branch / in this patch I replaced the XPath configuration for the brave (web) search and implemented this engine in your `brave.py` module.
> I also added rudimentary search-language and region (country) support .. in the next days I will implement a `fetch_traits(..)` function for a better locale support.
> 
That sounds awesome, can't wait to test it ;)
> I added a lot of features (time-range, safe-search) and changed some of your code .. can you please have a look / if you have some for me, comments (in the code) are welcome.
> 
I'll review and test your changes in the next couple of days, sure! 
> The brave engines needs a lot more testing .. if you have time it would be good to have some tests :-)
> 
I'm not yet familiar on how testing at Brave works, I'll add some tests if I can figure out how to :P
> In the next days I will implement the `fetch_traits(..)` .. but I have a lot of other work to do, so I can't promise when I will have time for / be patient :-)
>
No worries, I actually didn't expect such a fast feedback at any of my PRs, so just take your time and continue whenever you feel like you want to do so :)

searx/engines/brave.py
```diff
@@ -0,0 +1,250 @@
...
+}
+
+base_url = "https://search.brave.com/"
+categories = []
+brave_category = 'search'  # search, images, videos, news
+brave_spellcheck = False
+
+send_accept_language_header = True
+paging = False
+"""By Brave, paging is only supported in :py:obj:`brave_category`
```
Bnyro: I think this should be `At Brave`, or `Brave only supports paging in ...`, `By Brave` sounds very German :P

searx/engines/brave.py
```diff
@@ -0,0 +1,250 @@
...
+``search`` (UI category All)."""
+
+safesearch = True
+safesearch_map = {2: 'strict', 1: 'moderate', 0: 'off'}  # cookie: safesearch=off
+
+time_range_support = False
+"""By Brave, time range search is only supported in :py:obj:`brave_category`
```
Bnyro: Same as above

return42:
> I'm not yet familiar on how testing at Brave works, I'll add some tests if I can figure out how to :P

Sorry I was not clear: I mean do some queries in SearXNG and compare the results with the ones from Brave.

searx/engines/brave.py
```diff
@@ -0,0 +1,250 @@
...
+
+    params["url"] = f"{base_url}{brave_category}?{urlencode(args)}"
+
+    # set preferences in cookie
+    params['cookies']['safesearch'] = safesearch_map.get(params['safesearch'], 'off')
+
+    # ToDo: we need a fetch_traits(..) implementation / the ui_lang of Brave are
+    #       limited and the country handling has it quirks
+
+    eng_locale = params.get('searxng_locale')
```
Bnyro: Isn't that the same as
```python
eng_locale = params['searxng_locale']
```
?
I think we should either use that or provide a default value.
(that's probably something you wanted to change again when implementing `fetch_traits(...)`

searx/engines/brave.py
```diff
@@ -0,0 +1,250 @@
...
+    params['cookies']['useLocation'] = '0'  # the useLocation is IP based, we use 'country'
+
+    if not eng_locale or eng_locale == 'all':
+        params['cookies']['country'] = 'all'  # country=all
+    else:
+        params['cookies']['country'] = eng_locale.split('-')[-1].lower()
```
Bnyro: Not sure how relevant important is to us, we could also do
```python
locale_split = eng_locale.lower().split('-')
params['cookies']['country'] = locale_split[0]
params['cookies']['ui_lang'] = locale_split[1]
```

searx/engines/brave.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def response(resp):
+
+    if brave_category == 'search':
+        return _parse_search(resp)
+
+    datastr = ""
+    for line in resp.text.split("\n"):
+        if "const data = " in line:
```
Bnyro, line 214: maybe we should save "const data = " to a variable called `data_identifier` or something similar since it's used twice

Bnyro:
> Sorry I was not clear: I mean do some queries in SearXNG and compare the results with the ones from Brave.
>
Oh right, that makes sense :)

searx/engines/brave.py
return42: > that's probably something you wanted to change again when implementing `fetch_traits(...)`

Yes, this part is just a POC and will be changed when we have `fetch_traits(...)` implemented.

searx/engines/brave.py
return42: This part is also a part of the POC .. and will be changed.  But to answer your question:

    eng_locale = params['searxng_locale']

The `searxng_locale` parameter (SearXNG’s internal representation of a locale) can be something like `en-US` or `en` .. and in the latter, `locale_split[1]` will raise a ``IndexError: list index out of range``.

-----

A few words about languages & regions since it becomes much more complex as it looks like on a first glance ..

First: SearXNG locale support is based on [babel.Locale](https://babel.pocoo.org/en/latest/locale.html#the-locale-class) class. Babel itself is based on [CLDR](https://cldr.unicode.org/) / we can't support locales unknown to babel (CLDR).

The [fetch_traits](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.fetch_traits) function is intended to map "regions" and "languages" to SearXNG’s internal representation of a locale.

By example here is what startpage offers about regions and languages: 

https://github.com/searxng/searxng/blob/bcaaae699f01d44fa83dd8c251d9b30e5325d5e9/searx/data/engine_traits.json#L3470-L3618

.. the key in this dict is SearXNG's internal representation and the value is the representation the service is used in its API. When a SearXNG user selects locale `de-CH` its mapped to startpage's API by:

- Language `de --> deutsch`
- region `de-CH --> de_CH`

But not all engines do support *languages* or *regions*, some only support one or the other (and the meaning is often different, often the language is only the language of the UI).  Further the range of what a engine supports is different to other engines. Brave by example supports only this (UI) languages:

image

And a lot more regions :

image

The point becomes complicated when a user in SearXNG selects a locale for which is no 1:1 mapping in the engine traits.  Take in mind, SearXNG is a aggregator and the user selects a locale for all engines in this aggregation --> we have to make some assumptions to select a locale that *best fits* to what the user selected and is supported by the engine

This assumption and the approximation rules are implemented in [searx.locales.get_engine_locale](https://docs.searxng.org/src/searx.locales.html#searx.locales.get_engine_locale).

To close the circle again: this function is used by

- [EngineTraits.get_language](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.traits.EngineTraits.get_language)
- [EngineTraits.get_region](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.traits.EngineTraits.get_region)

By example here in the startpage implementation:

https://github.com/searxng/searxng/blob/bcaaae699f01d44fa83dd8c251d9b30e5325d5e9/searx/engines/startpage.py#L260-L263

---

To make it a bit more complex: when user selects a locale like `zh-TW` that is associated with a [Language Script](https://cldr.unicode.org/development/updating-codes/update-language-script-info/language-script-description) / [aka "Likely Subtags"](https://babel.pocoo.org/en/latest/locale.html#likely-subtags) then we must also master this. This can be done in the engine request or (even better) in the `fetch_traits()` of the engine.

To give an Example: here is the description of the startpage regions that map to (babel) locales ..

https://github.com/searxng/searxng/blob/bcaaae699f01d44fa83dd8c251d9b30e5325d5e9/searx/engines/startpage.py#L7-L28

.. in the UI of SearXNG the user can select:

- `zh-CN` which is mapped to babel `zh_Hans_CN` (simplified Chinese) and in startpage API it is `zh-CN_CN`

- `zh-TW` which is mapped to babel `zh_Hant_TW` (traditional Chinese) and in statpage API it is `zh-TW_TW`

searx/engines/brave.py
return42: I don't think this is needed; in this line its used to find a position in a string while the other usage replaces this string with an empty string ... Since this implementation depends on line breaks and strings being found in certain places, it cannot be ruled out that we will have to do some rework here soon (e.g. if the line breaks are set differently one day than they are today).

return42:
> > Sorry I was not clear: I mean do some queries in SearXNG and compare the results with the ones from Brave.
> 
> Oh right, that makes sense :)

Ok, I had some time today to finish my work on this PR / region support now is implement by EngineTraits ..

Can you please have a final test and give me some feedback / from my side this PR ready to merge.

Bnyro:
> Ok, I had some time today to finish my work on this PR / region support now is implement by EngineTraits ..
> Can you please have a final test and give me some feedback / from my side this PR ready to merge.
>
That's awesome!

I've been doing some search requests with it, and the everything seems to be working as expected:
* results are returned in the proper language, as selected
* the results are similar to the ones on Brave (as far I can tell, as they have a different sort order)
* even the top entry on the page (mostly Wikipedia) is displayed correctly

Therefore, looks good to me, ready to merge! Thanks for all your work on that!

PR #2608, opened 2023-08-07
[feat] engine: support for lemmy communities, posts and users

### LINKED ISSUE #698: Add lemmy.ml
**Working URL to the engine**

https://lemmy.ml

From https://join-lemmy.org/docs/en/index.html

> Lemmy is similar to sites like Reddit, Lobste.rs, Raddle, or Hacker News: you subscribe to forums you're interested in, post links and discussions, then vote, and comment on them. Behind the scenes, it is very different; anyone can easily run a server, and all these servers are federated (think email), and connected to the same universe, called the Fediverse.

TLDR : fediverse engine.

**Features of this engine**

https://lemmy.ml/post/89578

* social

**Additional context**

https://github.com/searx/searx/issues/2055

Bnyro:
## What does this PR do?
* add support for lemmy communities, posts and users

* Lemmy is very popular in open source communities

* !leco linux
* !lepo linux
* !leus linux
* !lecom linux

## Related issues
closes #698

@return42: The `content` field is provided as a markdown text by the api. Do we have any way already present in searx to convert that markdown to plain text or html?

### DIFF
--- a/searx/engines/lemmy.py
+++ b/searx/engines/lemmy.py
@@ -56,7 +56,7 @@ about = {
     "results": 'JSON',
 }
 paging = True
-categories = ['general', 'social media']
+categories = ['social media']
 
 base_url = "https://lemmy.ml/"
 """By default, https://lemmy.ml is used for providing the results.  If you want
--- a/searx/engines/lemmy.py
+++ b/searx/engines/lemmy.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""This engine uses the Lemmy API (https://lemmy.ml/api/v3/search), which is
+documented at `lemmy-js-client`_ / `Interface Search`_.  Since Lemmy is
+federated, results are from many different, independent lemmy instances, and not
+only the official one.
+
+.. _lemmy-js-client: https://join-lemmy.org/api/modules.html
+.. _Interface Search: https://join-lemmy.org/api/interfaces/Search.html
+
+Configuration
+=============
+
+The engine has the following additional settings:
+
+- :py:obj:`base_url`
+- :py:obj:`lemmy_type`
+
+This implementation is used by different lemmy engines in the :ref:`settings.yml
+<settings engine>`:
+
+.. code:: yaml
+
+  - name: lemmy communities
+    lemmy_type: Communities
+    ...
+  - name: lemmy users
+    lemmy_type: Users
+    ...
+  - name: lemmy posts
+    lemmy_type: Posts
+    ...
+  - name: lemmy comments
+    lemmy_type: Comments
+    ...
+
+Implementations
+===============
+
+"""
+
+from urllib.parse import urlencode
+from markdown_it import MarkdownIt
+from searx.utils import html_to_text
+
+about = {
+    "website": 'https://lemmy.ml/',
+    "wikidata_id": 'Q84777032',
+    "official_api_documentation": "https://join-lemmy.org/api/",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+paging = True
+categories = ['general', 'social media']
+
+base_url = "https://lemmy.ml/"
+"""By default, https://lemmy.ml is used for providing the results.  If you want
+to use a different lemmy instance, you can specify ``base_url``.
+"""
+
+lemmy_type = "Communities"
+"""Any of ``Communities``, ``Users``, ``Posts``, ``Comments``"""
+
+
+def request(query, params):
+    args = {
+        'q': query,
+        'page': params['pageno'],
+        'type_': lemmy_type,
+    }
+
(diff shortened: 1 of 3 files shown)


return42:
> Do we have any way already present in searx to convert that markdown to plain text or html?

(Not in a XPath engine)

The content field needs plain text / in the info pages I use markdown-it to render HTML from CommonMark(-down)

https://github.com/searxng/searxng/blob/a2108432f79581c8b87ccf456d7f866afa27a50b/searx/infopage/__init__.py#L73C1-L82

>  field is provided as a markdown text by the api. 

Not sure how it looks like, but if we strip all XML tags (`<..>`) it should be (more or less) plain text .. if this does not work may we convert to html and then remove HTML tags?

Can you show an example of how a markdown `content` would looks like?

----

BTW I would prefer a little python module over a XPath config .. what do you think?

Bnyro:
Okay, thanks for the info!

An example description would be
```
Comunidade Lusófona de entusiastas Linux.\n\nBem vindo ao nosso agregado de links sobre Linux.\n\n---\n\n##### Links (Em construção)\n\n\n Geral \n- [Arch Wiki](https://wiki.archlinux.org/)\n\n Básico ao avançado\n- [Guia Foca](https://www.guiafoca.org/guiaonline/)\n\n Materiais\n Linha de Comando\n\n- [LinuxCommand.org](https://linuxcommand.org/lc3_learning_the_shell.php)\n\n Sistema Operacionais\n\n- [Sistemas Operacionais:\nConceitos e Mecanismos](https://wiki.inf.ufpr.br/maziero/lib/exe/fetch.php?media=socm:socm-livro.pdf)\n\nGaming\n\n- [guia do r/linux_gaming](https://linux-gaming.kwindu.eu/index.php?title=Getting_started_with_Linux#Recommendations)
```

> BTW I would prefer a little python module over a XPath config .. what do you think?
>
I agree that it would be a good idea, I'll rewrite the engine later, we need changes for the markdown parsing anyways :)

Bnyro:
The markdown rendering you suggested works great!

return42:
@Bnyro I squashed some nits in your commit (mainly to improve the docs) --> see [diff](https://github.com/searxng/searxng/compare/913040d9df44ff27b9a79090bfed9d055ea9e1e8..3646b5e65e6911366e06fe0843dc3b28401a2d6e)

I'm still in review .. *to be continued*

return42:
FYI: last force-push was a rebase on branch master / just to get https://github.com/searxng/searxng/pull/2615 in this branch.

return42:
@Bnyro I finished my review on this PR / on top of your branch you will find two patches.  The first one adds more fields to the result items, the second one removes lemmy from category 'general' / its better placed in 'social media' and enabled by default I think.

Can you please test my patches .. any suggestions or comments?

Bnyro:
The changes look good to me in general.
I have been taking some time to make sure we don't use any nullable fields and that seems fine!
However, by the change in `_get_posts()` to limit the quality of the thumbnail to 128px, it seems like they're looking kind of broken in the search results (not sure if that's a general issue or only specific to lemmy now)

With `'?format=webp&thumbnail=128'`:
20230810_11h23m51s_grim

Without:
20230810_11h23m20s_grim

searx/engines/lemmy.py
```diff
@@ -0,0 +1,203 @@
...
+    if lemmy_type == "Users":
+        return _get_users(json)
+
+    if lemmy_type == "Posts":
+        return _get_posts(json)
+
+    if lemmy_type == "Comments":
+        return _get_comments(json)
+
+    raise ValueError("Unsupported lemmy type!")
```
Bnyro: I've seen, we usually do `return []` if the selected filter in `settings.yml` is not supported, do you think it's generally better to raise an error or just ignore it and return nothing?

e.g. https://github.com/searxng/searxng/blob/473129031710973e3c02e7178085b751b5e821ed/searx/engines/google_play.py#L51

return42: Since it is a misconfiguration I think it is better to throw an exception .. may we should fix the google engine also.  But we should better include the invalid value in the exception message:

```python
raise ValueError("Unsupported lemmy type: %s", lemmy_type)
```

return42:
> to limit the quality of the thumbnail to `128px`, it seems like they're looking kind of broken in the search results (not sure if that's a general issue or only specific to lemmy now)

Would you like to play with size and increase `128px` to find a value where it looks better?

Bnyro:
> Would you like to play with size and increase `128px` to find a value where it looks better?
>
Sure, I'll do that!

Bnyro:
208px seems to align perfectly in the search results, so I think that's fine like that.

return42:
> 208px seems to align perfectly in the search results, so I think that's fine like that.

Yes, looks much better :+1: 

What do you think about my comment ..

> the second one removes lemmy from category 'general' / its better placed in 'social media' and enabled by default I think.

If this is OK for you, this PR is ready to merge from my side.

Bnyro:
Yes, that's fine by me!

PR #2611, opened 2023-08-08
[feat] engine: implementation of German news, Tagesschau

### LINKED ISSUE #2609: New engine: Tagesschau
**Working URL to the engine**
* https://www.tagesschau.de/api2u/

**Features of this engine**
* provides recent news in German

* https://www.tagesschau.de/api2u/search?searchText=test
* https://github.com/AndreasFischer1985/tagesschau-api
* the implementation is not the issue, it's rather whether it'd be accepted

* news, general

**Additional context**
There are two concerns/questions from my side regarding adding it:
* According to https://github.com/AndreasFischer1985/tagesschau-api, we're not allowed to run more than 60 api requests per hour (though there's no rate-limiting, so it wouldn't be a problem to fetch more), so it's only some legal stuff about which I'm not sure
* It only provides results in German: I don't know whether engines, that only support a single language other than English are fine to become added

Bnyro:
## What does this PR do?
* add Tagesschau engine for German news, using `json_engine`

* the engine provides a variety of news on various topics, unfortunately in German only

* !ts Habeck

## Related issues
closes #2609

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1407,6 +1407,11 @@ engines:
   #      WHERE title LIKE :wildcard OR description LIKE :wildcard
   #      ORDER BY duration DESC
 
+  - name: tagesschau
+    engine: tagesschau
+    shortcut: ts
+    disabled: true
+
   # Requires Tor
   - name: torch
     engine: xpath
--- a/searx/engines/tagesschau.py
+++ b/searx/engines/tagesschau.py
@@ -1,0 +1,101 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""ARD: `Tagesschau API`_
+
+The Tagesschau is a news program of the ARD.  Via the `Tagesschau API`_, current
+news and media reports are available in JSON format.  The `Bundesstelle für Open
+Data`_ offers a `OpenAPI`_ portal at bundDEV_ where APIs are documented an can
+be tested.
+
+This SearXNG engine uses the `/api2u/search`_ API.
+
+.. _/api2u/search: http://tagesschau.api.bund.dev/
+.. _bundDEV: https://bund.dev/apis
+.. _Bundesstelle für Open Data: https://github.com/bundesAPI
+.. _Tagesschau API: https://github.com/AndreasFischer1985/tagesschau-api/blob/main/README_en.md
+.. _OpenAPI: https://swagger.io/specification/
+
+"""
+from typing import TYPE_CHECKING
+
+from datetime import datetime
+from urllib.parse import urlencode
+import re
+
+if TYPE_CHECKING:
+    import logging
+
+    logger: logging.Logger
+
+about = {
+    'website': "https://tagesschau.de",
+    'wikidata_id': "Q703907",
+    'official_api_documentation': None,
+    'use_official_api': True,
+    'require_api_key': False,
+    'results': 'JSON',
+    'language': 'de',
+}
+categories = ['general', 'news']
+paging = True
+
+results_per_page = 10
+base_url = "https://www.tagesschau.de"
+
+
+def request(query, params):
+    args = {
+        'searchText': query,
+        'pageSize': results_per_page,
+        'resultPage': params['pageno'] - 1,
+    }
+
+    params['url'] = f"{base_url}/api2u/search?{urlencode(args)}"
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    json = resp.json()
+
+    for item in json['searchResults']:
+        item_type = item.get('type')
+        if item_type in ('story', 'webview'):
+            results.append(_story(item))
+        elif item_type == 'video':
+            results.append(_video(item))
+        else:
+            logger.error("unknow result type: %s", item_type)
+
+    return results
+
+
+def _story(item):
+    return {
+        'title': item['title'],
+        'thumbnail': item.get('teaserImage', {}).get('imageVariants', {}).get('16x9-256'),
+        'publishedDate': datetime.strptime(item['date'][:19], '%Y-%m-%dT%H:%M:%S'),
+        'content': item['firstSentence'],
+        'url': item['shareURL'],
+    }
+
+
+def _video(item):
+    video_url = item['streams']['h264s']
+    title = item['title']
+
+    if "_vapp.mxf" in title:
+        title = title.replace("_vapp.mxf", "")
+        title = re.sub(r"APP\d+ (FC-)?", "", title, count=1)
+
+    return {
+        'template': 'videos.html',
+        'title': title,
+        'thumbnail': item.get('teaserImage', {}).get('imageVariants', {}).get('16x9-256'),
+        'publishedDate': datetime.strptime(item['date'][:19], '%Y-%m-%dT%H:%M:%S'),
+        'content': item.get('firstSentence', ''),
+        'iframe_src': video_url,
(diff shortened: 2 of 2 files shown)


searx/settings.yml
```diff
@@ -1407,6 +1407,25 @@ engines:
   #      WHERE title LIKE :wildcard OR description LIKE :wildcard
   #      ORDER BY duration DESC
 
+  - name: tagesschau
+    engine: json_engine
+    paging: true
+    search_url: https://www.tagesschau.de/api2u/search?searchText={query}&resultPage={pageno}
```
return42: Seems the API description at https://github.com/AndreasFischer1985/tagesschau-api/blob/69362cd840085ff02fa03ef4db899b304a783145/openapi.yaml#L107-L143 .. is a bit misleading: `description: Suchergebnisse pro Seite (1-30)` --> the `resultPage` starts with `0` while SearXNG's `pageno` starts with `1` / compare:

- https://www.tagesschau.de/api2u/search?searchText={query}&pageSize=10&resultPage=0
- https://www.tagesschau.de/api2u/search?searchText={query}&pageSize=10&resultPage=1

The API is very powerful, we can use much more data from it if we implement a Python module for it.

Can I persuade you to implement a Python module for this? ... it would be enough if you make a basic implementation ... the details I can implement then.

Bnyro: Sure, I can implement a python module for that!
Btw, the docs claim that the default `pageSize` would be 30, although it's 25 actually :p

return42:
Nice to know: https://api.bund.dev/ offers 47 APIs .. one is http://tagesschau.api.bund.dev/ :-)

Open Data: https://github.com/bundesAPI

Bnyro:
Oh, that's awesome!
I'm still trying to figure out how I can get the Tagesschau "Eilmeldungen" as an RSS feed, but unfortunately there's nothing like that in the API or RSS docs, I guess I'll have to use a scraper for that, and self-host it :/

return42:
> I'm still trying to figure out how I can get the Tagesschau "Eilmeldungen" as an RSS feed

RSS feeds are listed here: https://www.tagesschau.de/infoservices/rssfeeds

>  guess I'll have to use a scraper for that, and self-host it

If we can believe http://tagesschau.api.bund.dev/ --> "Ausgewählte aktuelle Nachrichten und Eilmeldungen" you can use:

    curl -v  'https://www.tagesschau.de/api2u/homepage'  -H 'accept: application/json'

note: the swagger APIs from the http://tagesschau.api.bund.dev/  page refer to `www.tagesschau.de/api2` but will be redirected to `www.tagesschau.de/api2u`.

--- 

If we have a python module for this PR, beside `api2u/search` we can implement more APIs in this module (`api2u/homepage` for example in the info box(?)).  But first we need a python module and more experience with these interface and what we can assemble from.

Bnyro:
(the force push changed the commit message title to fit the formatting of the others)

return42:
> the force push changed the commit message title to fit the formatting of the others

Thanks a lot .. it is not mandatory but I prefer the notation with the square bracket .. this notation is nicely highlighted by some git tools ;-)

image

return42:
@Bnyro to keep the commit history short I squashed my [modifications](https://github.com/searxng/searxng/compare/29b1986383207f8cb168ae959792feebe68ad0e3..cdbd5d613c9f27529b848baa08969a2fba6e1d41) directly in you commit / hope that is OK for you.

Can you please test my modifications .. any suggestions or comments?

Bnyro:
> @Bnyro to keep the commit history short I squashed my [modifications](https://github.com/searxng/searxng/compare/29b1986383207f8cb168ae959792feebe68ad0e3..cdbd5d613c9f27529b848baa08969a2fba6e1d41) directly in you commit / hope that is OK for you.
> 
Sure!
> Can you please test my modifications .. any suggestions or comments?
>
Works great with your modifications, showing the thumbnail of the news as well seems to be a really good choice to me.

One thing I noticed: When you search for `!ts Frankreich Marokko` for example, one of the first results is called `APP1630 NIF- FFWM Frankreich - Marokko NEU_vapp.mxf` (video result), and other videos seem to follow a similar naming scheme (e.g. `APP0600 FC- Liveblog Ukraine_vapp.mxf`.
So I wonder if we should remove the the trailing `_vapp.xmf` and the `APP{number_id}` at the beginning of the title to make it easier to read in the results.

And I noticed there's also an `image` type, which we could/should be able to handle as well since it's returned quite commonly.

I can work on both things and squash them if you want to, (or you can do so if you have time :) ).

return42:
> I can work on both things and squash them if you want to, (or you can do so if you have time :) ).

if you have time for .. don't hesitate, start hacking :-)

BTW: the `paging=true` in the settings file can be removed .. and may we have to play with the timeout which is currently set to 6 sec .. I think we should test how 3 sec will work .. or do we often get timeouts? .. needs some practical experience to find a good value ( I want to enable this engine on my instance .. and it is not good to have timeouts > 3sec in the general category).

Bnyro:
> if you have time for .. don't hesitate, start hacking :-)
>
:+1:  
> BTW: the `paging=true` in the settings file can be removed .. and may we have to play with the timeout which is currently set to 6 sec .. I think we should test how 3 sec will work .. or do we often get timeouts? .. needs some practical experience to find a good value ( I want to enable this engine on my instance .. and it is not good to have timeouts > 3sec in the general category).
>
I've made the experience that results timed out when we didn't yet limit the `pageSize` to 10 items. However, it seems to be much faster when fetching less results, so I think we should be able to drop that `timeout` in the config now. I'll test that a bit :)

return42:
One thing I forgot .. can you add a 'language' field to the about properties / like here in MediathekViewWEB:

https://github.com/searxng/searxng/blob/2256ba2ffbf625b03432c54352e5e05f6eeb9150/searx/engines/mediathekviewweb.py#L17

In UI preferences the language is added to the engine name so that the user can decide if he wants to activate an engine in this language:

image

Bnyro:
I'll add the language field too, thanks for the info!

Fyi: there's no `image` category, I mixed it up since the `image` type is being provided for normal search results in the `images` array (that contains the thumbnails of normal results).

However, I found an other category: `webview`, which seems useful too, so I'll add that now :wink:

Bnyro:
Good to merge from my side now :)

Bnyro:
GitHub seems to have some issues currently, CI doesn't work properly and I can't seem to force push anymore.

return42:
> Good to merge from my side now :)

just squashed a little bit doc in the commit [diff](https://github.com/searxng/searxng/compare/a13f168ba13375557c12757b593379898c885a62..c0c63ceeba81df364ebc656201a2e007a9317cdd) .. lets wait until test from CI has passed and then we can merge I think.

Thanks again for the work you put in ... it's a real pleasure for me to work with you on these PRs .. I am already curious about the other PRs .. but unfortunately I have to go for now ... "'I'll be back" :)

Bnyro:
> Thanks again for the work you put in ... it's a real pleasure for me to work with you on these PRs .. I am already curious about the other PRs .. but unfortunately I have to go for now ... "'I'll be back" :)
>
The pleasure is on my side ;)
I initially only wanted to work on the PR for supporting Piped, but it turned out that working on searx engines is much more interesting than I thought it would be :)

Bnyro:
And since I'm on holiday currently, I've enough time to look into some engines or issues that seem interesting :wink:

PR #2613, opened 2023-08-08
[feat] results: show source of suggested answer

### LINKED ISSUE #2236: Show source for the short definition/explainer below the search bar
Not a bug but an enhancement for an existing feature.

The explainer below the searchbar is realy good for a short summary, but often i want to visit this page and cannot because there is no link attached to it. Most of the time it is not the first search result.
grafik

An clickable URL (shown shortend) below the explainertext below the searchbar
grafik

Making the Textfield itself clickable and hide url as to look the same as it currently is.
-> may not be as good if it pulls some description from a sketchy site

**Additional context**

As I said the Explainer often is realy good and it would be nice to cklick on this result to get more information.

Bnyro:
## What does this PR do?
* show the source of the answer provided as a clickable link
* currently, google answers are not supported by this - all other engines should return an url for their answer already

* it allows users to view the source of the provided answer

* search "!brave https vorteile" with language set to german

## Related issues
closes #2236

20230808_18h20m38s_grim


return42:
@Bnyro in https://github.com/searxng/searxng/pull/2613/commits/7ac0a3847936d253c9a2788ccf4cca09c00fc745 you changed a lot of indentations and line breaks not needed for your patch .. can you please reduce your modification to what is needed to patch / so we have a more readable diff .. thanks for for paying attention.

Bnyro:
Oh, sorry, I didn't notice that. I just have been running the formatter as usual, I'll fix that.

return42:
@Bnyro I fixed the answers of two engines more .. on top you see WIP commit with patches we can squash in your first commit (if you agree with these patches) / the patches are:

- instead of `get_url_hostname` place `urlparse` in template's context (if we need to parse URLs in other places in the future)
- instead of a margin on top of the URL, set a margin on the bottom of the URL / here is how it will looks like, when you have more then one answer ..

image

Can you please test my patches / to test I used:

- `10 dollar to euro`
- `!go :en https vs http`

Bnyro:
> instead of get_url_hostname place urlparse in template's context (if we need to parse URLs in other places in the future)
>
That's probably not a bad idea, I'm fine with that :)

> instead of a margin on top of the URL, set a margin on the bottom of the URL / here is how it will looks like, when you have more then one answer ..
>
The bottom margin is reasonable in my opinion, however I think we should also add a small top margin to the url, maybe 5px, so that it's still clear which answer the url belongs to, but that the url and the answer have a small separation to make it obvious that the url is not part of the answer text. (which is not that obvious if the last line of the answer covers almost the full width)

> I fixed the answers of two engines more
>
That's great, that makes the feature much more useful!

> Can you please test my patches / to test I used
>
Tested, all fine here :)

return42:
>  however I think we should also add a small top margin to the url, maybe 5px, so that it's still clear which answer the url belongs to

:+1: .. done / I squashed it into your commit see diff [cd4387448..3e86785c6](https://github.com/searxng/searxng/compare/cd4387448..3e86785c6)

Can you please have a final look at the current implementation / if everything is OK from your side, I would merge.

Have I already said it? that was once again an excellent work :100: .. a true pleasure .. I wish I had more time :)

Bnyro:
After a short testing, I can confirm it's working (and looking) great now!

> Have I already said it? that was once again an excellent work 💯 .. a true pleasure .. I wish I had more time :)
>
Thank you! <3
It was much fun working on that one with you from my side as well :)

PS: Sorry for the delay - I have been really busy the last two days, so I didn't really have the chance to look into it.

PR #2614, opened 2023-08-09
[mod] remove discontinued petalsearch engines

allendema:
## What does this PR do?
Remove two petalsearch engines: web and images.

https://en.wikipedia.org/wiki/Petal_Search#cite_note-1

Less cluttered settings.

Searches with the engines ```!pts !ptsi query```  should not return results.

## Author's checklist
N/A

## Related issues

N/A


return42 (approved):
Hi @allendema, thank you for helping to keep SearXNG clean :+1: .. seems petalsearch.com is dead / but I can't find an announcement  .. anyway, its dead and we can drop the code.

allendema:
Hi there, the engines were not working for some weeks now.
Based on the cited link above, the only info on the website was removal of some services but nothing on discontinuation.

PR #2617, opened 2023-08-09
[mod] settings: show weight of each instance

### LINKED ISSUE #1701: Show weight of engines under preferences
Similiar to safesearch, shortcut, response time etc., the weight of engines should also be shown under "preferences" > "engines".

When using the same  engines on different instances a user would expect almost  identical ranking of results.

But the instance admin could have changed the importance of an engine, say for ```!ask !scc !szn !su``` instead of the default ```1.0```  
a value of ```0.5``` is set and as a user there is no way to tell.

Show the weight of engines under  "preferences" > "engines".

The ```weight``` is a very powerful option in searXNG but almost never used.
Some simplier engines could be downranked by default for example.

This is somehow "hidden" in the docs.

**Additional context**

N/A

Bnyro:
## What does this PR do?
* this PR adds a new "Weight" column to the engine details table in the settings

* it increases the transparency about how results are being generated

* go to settings -> engines -> look at the "weight" tab

## Related issues
closes #1701

### DIFF
--- a/searx/templates/simple/preferences/engines.html
+++ b/searx/templates/simple/preferences/engines.html
@@ -20,6 +20,7 @@
         <th>{{- _("Supports selected language") -}}</th>{{- '' -}}
         <th>{{- _("SafeSearch") -}}</th>{{- '' -}}
         <th>{{- _("Time range") -}}</th>{{- '' -}}
+        <th>{{- _("Weight") }}</th>
         {%- if enable_metrics -%}
           <th>{{- _("Response time") -}}</th>
         {%- endif -%}
@@ -72,6 +73,7 @@
               <td>
                 {{- checkbox(None, supports[search_engine.name]['time_range_support'], true) -}}
               </td>{{- '' -}}
+              <td>{{- search_engine.weight or '1.0' -}}</td>{{- '' -}}
               {%- if enable_metrics -%}
                 {{- engine_time(search_engine.name) -}}
               {%- endif -%}


searx/translations/messages.pot
```diff
@@ -651,6 +651,10 @@ msgstr ""
 msgid "Reliability"
 msgstr ""
 
+#: searx/templates/simple/preferences/engines.html:23
+msgid "Weight"
+msgstr ""
+
```
return42: Can you remove the patch of the .pot file from this commit / thanks!

----

No need to extract translations; we have a [workflow](https://github.com/searxng/searxng/blob/733b795d5327a1be133365d9d46c41d709719109/.github/workflows/integration.yml#L163-L167) for the extraction.  This workflow is triggered when a patch (PR) is merged on branch master. 

Our translation process is more or less completely automated.  At first glance it looks a bit complicated, but it has the advantage that we have completely separated the translation from the master branch and changes to translations are incorporated into the normal review process via automated PRs.  The PRs from the translation tool are created once a week .. the entire process is described and illustrated here:
https://docs.searxng.org/dev/translation.html#searxng-weblate-workflow

return42 (changes requested):

searx/translations/messages.pot
Bnyro: Oh right, that's cool!
I thought they would need to become added once to the root `messages.pot` file and will then be added to the other languages via a workflow, but it's even better that way :)

return42 (approved):
Thanks!

return42:
> this workflow is triggered when a patch (PR) is merged on branch master.

You can see it in the history from weblate .. "1 new string to translate appeared in the translation." --> https://translate.codeberg.org/projects/searxng/searxng/de/#history

I just translated it to DE .. and on Friday we will see a PR coming in with title "Update translation" .

Bnyro:
> You can see it in the history from weblate .. "1 new string to translate appeared in the translation." --> https://translate.codeberg.org/projects/searxng/searxng/de/#history
> 
That's cool, thanks for telling me :)

PR #2618, opened 2023-08-09
[mod] decrease the weighting of wikivoyage, wikibooks, wikisource and wikiversity

return42:
The search engines deliver hits for many search terms [1], but these are usually not the focus of the user. In order to arrange these hits further down in the list, their weighting is reduced.

[1] https://github.com/searxng/searxng/pull/2589#issuecomment-1670915089

@Bnyro, @unixfox: could you please do a review / thanks :+1:

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1520,10 +1520,12 @@ engines:
 
   - name: wikibooks
     engine: mediawiki
+    weight: 0.5
     shortcut: wb
     categories: [general, wikimedia]
     base_url: "https://{language}.wikibooks.org/"
     search_type: text
+    disabled: true
     about:
       website: https://www.wikibooks.org/
       wikidata_id: Q367
@@ -1541,10 +1543,12 @@ engines:
 
   - name: wikiquote
     engine: mediawiki
+    weight: 0.5
     shortcut: wq
     categories: [general, wikimedia]
     base_url: "https://{language}.wikiquote.org/"
     search_type: text
+    disabled: true
     additional_tests:
       rosebud: *test_rosebud
     about:
@@ -1553,6 +1557,7 @@ engines:
 
   - name: wikisource
     engine: mediawiki
+    weight: 0.5
     shortcut: ws
     categories: [general, wikimedia]
     base_url: "https://{language}.wikisource.org/"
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

return42:
@Bnyro @unixfox: additional I set the weight of wikivoyage to 0.5 / in practice I to often have (unrelated to my query) results from wikivoyage in the result list, placed before more related google results.

return42:
> [I think we should be very careful with what engines we add as enabled by default for the General category ...](https://github.com/searxng/searxng/pull/2589#issuecomment-1683677727)

I think there is some truth in your words.  Should I deactivate the engines in this PR?

unixfox:
> > [I think we should be very careful with what engines we add as enabled by default for the General category ...](https://github.com/searxng/searxng/pull/2589#issuecomment-1683677727)
> 
> I think there is some truth in your words. Should I deactivate the engines in this PR?

Maybe I biased because I'm only interested in "Google" results but in my opinion these engines shouldn't be enabled by default, we already have too many ones by default.

Bnyro:
In my opinion, the results of wikiversity and wikibooks are not that useful usually, so I don't really mind if they are disabled by default.
Nevertheless I think that Wikisource has some very helpful resources very often, that you wouldn't find otherwise.
Concerning wikivoyage I'm not that sure - it provides some okay results but the ones from wikipedia are usually quite similar - and wikipedia results appear at the top of most other general search engines too.

return42:
> Maybe I biased because I'm only interested in "Google" 

I think google results (minus adds) is a good reference for a default setting of SearXNG engines.  SearXNG results should be by default something I would name *google plus plus* :)

I tried to find a compromise combining `weight` & `disabled`:

1. all mediawiki engines in category `general` are `weight` by 0.5 (to avoid listed above more relevant hits from google or ddg)

2. all mediawiki engines in category `general` are `disabled` by default, except `wikisource`

If you think this is a good compromise, give me your *thumbs up* .. otherwise give me a patch of the settings.yml file you think is better for a default setting .. for me every setting is OK .. I will customize it in my local instance to my needs :)

mrpaulblack (approved):
LGTM

unixfox (approved):

PR #2620, opened 2023-08-10
Report type misconfig

Bnyro:
## What does this PR do?
* throw an exception if the type or category specified in `settings.xml` is not supported

## Related issues
see https://github.com/searxng/searxng/pull/2608#discussion_r1289857117


return42 (approved):
Wow, that was fast .. thanks a lot for having an eye on this topic.

Bnyro:
I hope I caught all of them, these seemed to be the only ones with issues returned by `grep -Rlw 'return []'`.

PR #2622, opened 2023-08-10
[feat] engine: implementation of seekr for news, images and videos

### LINKED ISSUE #1644: seekr
**Working URL to the engine**
https://www.seekr.com

Seekr is a transparent and independent search engine

**Features of this engine**
includes a Seekr Score of High, Medium, or Low with each result

General, news, images, videos, maps

Bnyro:
## What does this PR do?
* implement seekr as engine for news, videos and images
* uses the official seekr api
* the videos support is very basic due to api limitations - not sure if we can do anything about it
* seekr has two different api versions: v1 (used for images and videos) and v2 (used for news)

* I don't have much experience with seekr, but the results seem to be decent

* !senews World
* !seimg World
* !sevid World

## Related issues
closes #1644

### DIFF
--- a/searx/engines/seekr.py
+++ b/searx/engines/seekr.py
@@ -1,50 +1,120 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
-"""Seekr (images, videos, news)
+"""seekr.com Seeker Score
+
+Seekr is a privately held search and content evaluation engine that prioritizes
+credibility over popularity.
+
+Configuration
+=============
+
+The engine has the following additional settings:
+
+- :py:obj:`seekr_category`
+- :py:obj:`api_key`
+
+This implementation is used by seekr engines in the :ref:`settings.yml
+<settings engine>`:
+
+.. code:: yaml
+
+  - name: seekr news
+    seekr_category: news
+    ...
+  - name: seekr images
+    seekr_category: images
+    ...
+  - name: seekr videos
+    seekr_category: videos
+    ...
+
+Known Quirks
+============
+
+The implementation to support :py:obj:`paging <searx.enginelib.Engine.paging>`
+is based on the *nextpage* method of Seekr's REST API.  This feature is *next
+page driven* and plays well with the :ref:`infinite_scroll <settings ui>`
+setting in SearXNG but it does not really fit into SearXNG's UI to select a page
+by number.
+
+Implementations
+===============
+
 """
 
 from datetime import datetime
 from json import loads
 from urllib.parse import urlencode
+from flask_babel import gettext
 
 about = {
     "website": 'https://seekr.com/',
     "official_api_documentation": None,
-    "use_official_api": True,
+    "use_official_api": False,
     "require_api_key": True,
     "results": 'JSON',
+    "language": 'en',
 }
-paging = True  # news search doesn't support paging
 
 base_url = "https://api.seekr.com"
-# v2/newssearch, v1/imagetab, v1/videotab
-seekr_path = "newssearch"
-seekr_api_version = "v2"
+paging = True
+
 api_key = "srh1-22fb-sekr"
-results_per_page = 10
+"""API key / reversed engineered / is still the same one since 2022."""
+
+seekr_category: str = 'unset'
+"""Search category, any of ``news``, ``videos`` or ``images``."""
+
+
+def init(engine_settings):
+
+    # global paging
+    if engine_settings['seekr_category'] not in ['news', 'videos', 'images']:
+        raise ValueError(f"Unsupported seekr category: {engine_settings['seekr_category']}")
 
 
 def request(query, params):
+
+    if not query:
+        return None
+
     args = {
         'query': query,
         'apiKey': api_key,
-        'limit': results_per_page,
-        'offset': (params['pageno'] - 1) * results_per_page,
     }
 
-    path = f"{seekr_api_version}/{seekr_path}"
-    if seekr_api_version == "v1":
-        path = seekr_path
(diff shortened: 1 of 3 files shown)


searx/engines/seekr.py
```diff
@@ -0,0 +1,104 @@
...
+"""
+
+from datetime import datetime
+from json import loads
+from urllib.parse import urlencode
+
+about = {
+    "website": 'https://seekr.com/',
+    "official_api_documentation": None,
+    "use_official_api": True,
```
return42: When the API is official but not documented, from where did you get your know how of the details from the API .. reversed engendered?

Bnyro: Not sure if the API is official, I didn't search deeper about it - I reverse engineered the API requests in the browser dev tools.
From my research now the API is probably not official, I just thought so because we need to provide an `apiKey`, which is always the same when using it in the browser however.

return42: > I reverse engineered the API requests in the browser dev tools.

This is what I also often do :) .. TBH I haven't thought in deep, but I think the intention the flag `use_official_api` is to show if the engine uses an official API that is documented (and versioned) .. on the opposite we have engines where we know there is an  API but we didn't use the API:

https://github.com/searxng/searxng/blob/c76c999a1eeecae265573b12e81ae74f56988820/searx/engines/google.py#L42-L43

Even the google engine is reversed engineered, it can help to understand the arguments of the API when we follow (more or less) the official API.

I will change the value of the flag in my review ..

searx/engines/seekr.py
```diff
@@ -0,0 +1,104 @@
...
+    "require_api_key": True,
+    "results": 'JSON',
+}
+paging = True  # news search doesn't support paging
+
+base_url = "https://api.seekr.com"
+# v2/newssearch, v1/imagetab, v1/videotab
+seekr_path = "newssearch"
+seekr_api_version = "v2"
+api_key = "srh1-22fb-sekr"
```
return42: I think we still have too little experience with the API to offer official settings for it today.  If we release such settings today, then we also have to maintain them in the future and in the meantime some maintainers may have made their own (local) settings, which would then overwrite our bugfixes again.

If it is OK for you I would like to hard-code these values (as part of my review).

----

I am not sure yet, but maybe we can get the api-key via a fetch method too

Bnyro: Sure, that makes sense!
However, I don't think we can fetch an API key somehow - it's still the same one as when the seekr issue got opened here 1.5 years ago, so I think it's fine to use that one.

searx/engines/seekr.py
```diff
@@ -0,0 +1,104 @@
...
+            }
+        )
+
+    return results
+
+
+def _news_response(json):
+    results = []
+
+    for result in json['expertResponses'][0]['advice']['categorySearchResult']['searchResult']['results']:
```
allendema: This will give out an error if nothing is found.

Maybe use something like this:

```py
results = []

if json['expertResponses'][0]['responseStatus'] == "NoResultsMatched":
       return results
```
...

return42: Hi @allendema thanks for pointing out / I just in review with this PR and fixed this issue already .. give me a few days for the review and I will come back here with a revision of the implementation .. and then it would be great if you test the revision (which will include major changes).

searx/engines/seekr.py
```diff
@@ -53,52 +123,97 @@ def _images_response(json):
...
+        search_results = json.get('advice')
+
     results = []
+    if not search_results:
+        return results
 
-    for result in json['expertResponses'][0]['advice']['categorySearchResult']['searchResult']['results']:
+    for result in search_results['results']:
+
+        gettext("Language")
```
Bnyro: That `gettext` seems to be unused?

return42:
@Bnyro, @allendema : I rebased the branch and pushed commit be2df9a on top .. be2df9a includes some major changes compared to the initial commit ba69c8e .. these changes would need to be thoroughly tested .. if you have time for, it would be great if you cant test the new engines ..

searx/engines/seekr.py
return42: Oops .. I dropped the line :facepalm:

Bnyro:
From a briefly look at your changes, this looks great!
I'll test that as soon as possible, probably tomorrow :)

Bnyro:
Great job, I couldn't find any issues :)
The engine seems much more useful now due to the new fields you added!

One question: How did you figure out how we can display the thumbnail for video and image results? Did you compare the json response with the network tab requests in the devtools? I remember trying something similar when I worked at the engine, but I couldn't really figure out how to do it.

return42:
> One question: How did you figure out how we can display the thumbnail for video 

I figured that the WEB client of seekr also has no more data than the one in the JSON, so the thumbnail URL had to be made from that .. I then looked at the URLs loaded from the WEB client (in FFox WEB Developer):

```
https://media.seekr.com/engine/rp/z5e1ef0c971d6818c8959ea4a89f8a326/?src= bZr9GTwXFFwQn_lwuUYHbfwi4Ni1Ic-fSPGkN-mYxg7OB1dgBsj3VEghjsjZc7jRCpPT2l8U3Kei1Nz_-cYZ1Q8JleI```
```

and I *hacked*  a little around ...

the `src= ` value was given in the `thumbnail` field .. and the `z5e1ef0c971d6818c8959ea4a89f8a326` path was in the `tp` field of `summary`


image

Bnyro:
Cool, thanks!

allendema:
Also tested them and can confirm that the engines worked fine.

PR #2637, opened 2023-08-11
[feat] engine: implementation of Yummly 

### LINKED ISSUE #438: Yummly engine
**Working URL to the engine**

https://www.yummly.com/

I think that it would be cool to have a category for recipes like duckduckgo has.  That way people can find recipes easily.  Not open source or libre that I know of.

**Features of this engine**

Finds recipes for people.

Create a new category called recipes.

**Additional context**

Example of what I think it should look similar to https://duckduckgo.com/?q=shrimp+scampi&ia=recipes&iax=recipes

Bnyro:
## What does this PR do?
* add yummly.com as a new engines for recipes
* currently yummly is uncategorized, since it's the only supported recipes search engine at the moment

* it makes it easy to find recipes in the web

* !yum pancakes

## Author's checklist
* yummly has an official api requiring an api key, however there's also an unofficial one providing the same results without an api key

## Related issues
closes #438

### DIFF
--- a/searx/engines/yummly.py
+++ b/searx/engines/yummly.py
@@ -1,0 +1,79 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Yummly
+"""
+
+from urllib.parse import urlencode
+
+from flask_babel import gettext
+from searx.utils import markdown_to_text
+
+about = {
+    "website": 'https://yummly.com',
+    "wikidata_id": 'Q8061140',
+    # not used since it requires an api key
+    "official_api_documentation": 'https://developer.yummly.com/documentation.html',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+paging = True
+categories = []
+
+api_url = "https://mapi.yummly.com"
+number_of_results = 10
+show_pro_recipes = False
+base_url = "https://www.yummly.com"
+
+
+def request(query, params):
+    args = {
+        'q': query,
+        'start': (params['pageno'] - 1) * number_of_results,
+        'maxResult': number_of_results,
+    }
+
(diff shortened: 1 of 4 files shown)


searx/engines/yummly.py
```diff
@@ -0,0 +1,66 @@
...
+            continue
+
+        content = result['seo']['web']['meta-tags']['description']
+        description = result['content']['description']
+        if description is not None:
+            content = markdown_to_text(description['text'])
+
+        results.append(
+            {
+                'url': result['display']['source']['sourceRecipeUrl'],
```
return42: Sometimes the sourceReceipeUrl seems not to work, in my tests by example I got

- `https://www.yummly.com/private/recipe/Chickpea-Kartoffelpuffer-9656329?layout=prep-steps`

I think it is better to use ..

image

... I will fix it.

Bnyro: `sourceRecipeUrl` shows the original source of the recipe, e.g. the New York times, while the tracking ID only provides the link to the Yummly recipe. So, in order to get the real recipe source, we should use `sourceRecipeUrl or ... + tracking-id`.

return42: > `sourceRecipeUrl` shows the original source of the recipe, e.g. the New York times,

ah! .. thanks for pointing out, I placed a WIP commit on top of your branch, you can look at this after your vacation, now enjoy your vacation and switch off :relaxed:

Bnyro: Seems to work as expected now, thanks for having a look at that!

I've been trying to rebase the changes which somewhat didn't work (because I have been acting a bit stupid), so I will by trying to fix that now.

return42: Sorry I missed a `/` in my WIP commit .. its now [fixed](https://github.com/searxng/searxng/compare/e16abfb815cd75997d08a242ab91734338b3a97d..25d00c52466b2aa3df0971c5fd5e479467437cd7)

PR #2639, opened 2023-08-11
Improve settings.yml

ghost:


searx/settings.yml
```diff
@@ -31,16 +31,15 @@ search:
   autocomplete_min: 4
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
-  default_lang: "auto"
-  # Available languages
-  # languages:
+  # Available languages:
```
return42, line 34: this will break the setting example which is

```yaml
languages:
  - all
  - ..
```

return42:
Hi @ToxyFlog1627 thanks a lot for your PR .. but TBH I don't think there is a big benefit when we use two lines for settings & comments where we have both in one line.

I hope you do not feel discouraged if we do not merge this PR / anyway thanks a lot for your contribution

ghost:
I thought making comments always precede option will make it more consistent. Same idea with languages - why list of available options is AFTER the property? I didn't think about it breaking something.

return42:
> I didn't think about it breaking something.

```yaml
Available languages:
  - all
  - ..
```

versus 

```yaml
languages:
  - all
  - ..
```

ghost:
Oh, and about
 > I hope you do not feel discouraged if we do not merge this PR / anyway thanks a lot for your contribution
I don't, because in my opinion
1. You are the maintainer, so it's up to you
2. You gave a valid argument why not
3. You took your time to look at it

ghost:
Yeah, I kinda forgot it was a property and not just a comment :/

ghost:
Also, why are some properties commented out? I thought it contained only default values

return42:
> Also, why are some properties commented out? I thought it contained only default values

These are example configurations, the settings are documented [here](https://docs.searxng.org/admin/settings/index.html)

Issue #2038, opened 2022-12-23
Route over tor


zlElo:
**Describe the solution you'd like**
A option to route the requests over tor (only a easy check button or switch button in the settings). This function is already implemented in whoogle search.

danktankk:
tor is only as secret as its endpoint.  lol  funny that people dont get that...  haha  if you really think you have done something by using tor to hide yourself, you are truly a fool.

zlElo:
Hi, too bad you can't speak to one normally here.  But whatever.  If you only had the endpoint/exitnode then that would of course be insecure or pointless.  Built for this case there is also the entry node and the middle relay.  Means the exit node only knows that the request comes from the middle relay and that it only comes from the entry node.  everything is encrypted up to the exit node.  Of course, the exit node is the point where the hoster knows what was called, but he doesn't know who called it.  To what extent it now makes sense to only use Tor when searching remains to be seen, but I think you should also try to offer features that other alternatives offer in order to stand up to the competition... (Whoogle has the feature)

Greets

mo-rijndael:
But how this will affect user? User already talks only with Searx server. If you need to hide your IP, you can connect to SearX via Tor. But way how SearX instance connects to engines is server administrator business

zlElo:
Good point.  The question is rather how Whoogle then uses that.  I didn't think about it at all, I have to say, thank you for your comment :) But how about making the feature available for the admins of the servers?  So that the server asks the individual search engines via Tor?  Does this make sense?!  xD

PR #2644, opened 2023-08-12
Bump httpx 0.21.2 from to 0.24.1

dalf:
## What does this PR do?

Update httpx to 0.24.1

Up to date httpx version, so dependabot can create PR.

Check the proxy configuration: http, socks4, socks5, socks5h

All seems to work as expected.
No error after re-connection after a 10 minutes delay.

However, it would be good to run this PR on a public instance with proxies for a few days to check, and then merge.

## Author's checklist

See https://github.com/encode/httpx/blob/master/CHANGELOG.md

Changes that might require some care:

> * Fix optional percent-encoding behaviour. (`#2671`)
> * Use utf-8 as the default character set, instead of falling back to charset-normalizer for auto-detection. To enable automatic character set detection, see [the documentation](https://www.python-httpx.org/advanced/#character-set-encodings-and-auto-detection). (`#2165`)
> * Fix which gen-delims need to be escaped for path/query/fragment components in URL. (`#2701`)
> * Fix Headers.update(...) to correctly handle repeated headers (`#2038`)

Note: 
* httpx 0.24.1 supports socks5h proxy but not socks5 proxy, so with this PR, SearXNG continue to rely on httpx-socks for the SOCKS proxy.
* Also, httpx-socks continues to be useful for proxy with https URL (which requires TLS in TLS connection).

About logging:
* This PR basically disable httpx logging to rely on SearXNG logging.
* See https://www.python-httpx.org/logging/ (this does not exist in the version 0.21.2):

> * The logging behaviour has been changed to be more in-line with other standard Python logging usages. We no longer have a custom TRACE log level, and we no longer use the HTTPX_LOG_LEVEL environment variable to auto-configure logging. We now have a significant amount of DEBUG logging available at the network level. Full documentation is available at https://www.python-httpx.org/logging/ (`#2547`, https://github.com/encode/httpcore/pull/648)

## Related issues

Supersede #2302

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,10 +7,10 @@ lxml==4.9.3
 pygments==2.16.1
 python-dateutil==2.8.2
 pyyaml==6.0.1
-httpx[http2]==0.21.2
+httpx[http2]==0.24.1
 Brotli==1.0.9
 uvloop==0.17.0
-httpx-socks[asyncio]==0.7.2
+httpx-socks[asyncio]==0.7.7
 setproctitle==1.3.2
 redis==4.6.0
 markdown-it-py==3.0.0
--- a/searx/network/client.py
+++ b/searx/network/client.py
@@ -180,7 +180,15 @@ def get_loop():
 
 def init():
     # log
-    for logger_name in ('hpack.hpack', 'hpack.table', 'httpx._client'):
+    for logger_name in (
+        'httpx',
+        'httpcore.proxy',
+        'httpcore.connection',
+        'httpcore.http11',
+        'httpcore.http2',
+        'hpack.hpack',
+        'hpack.table',
+    ):
         logging.getLogger(logger_name).setLevel(logging.WARNING)
 
     # loop
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -409,7 +409,7 @@ def done():
     """Close all HTTP client
 
     Avoid a warning at exit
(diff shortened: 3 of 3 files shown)


return42:
>  However, it would be good to run this PR on a public instance with proxies for a few days to check, and then merge.

I don't know if we can find someone to test this for us in the near future.  Before we postpone the change any further we could just roll-out and see if we get any feedback at all / SearXNG is a rolling release ... we can roll-back the change at any time.

@unixfox @Bnyro what do you think?

Bnyro:
> I don't know if we can find someone to test this for us in the near future. Before we postpone the change any further we could just roll-out and see if we get any feedback at all / SearXNG is a rolling release ... we can roll-back the change at any time.
> 
> @unixfox @Bnyro what do you think?
>
From my testing on a local machine, I couldn't find any issues (though I obviously didn't test any behavior changes regarding TLS locally).

I think it's okay to merge the PR, as you said rolling back is always possible later.

return42:
@mrpaulblack I assume your last [force-push](https://github.com/searxng/searxng/compare/6bf5baeb3fdf9f3e2be610c6d5ff150fe24d7d72..b4e4cfc026fdc287622e0aa5f08db5d1e30ef869) was just a rebase on master .. right?

mrpaulblack:
Yes

mrpaulblack:
After testing this for the last 2 days on my public instance, memory usage seems to be down a bit:
image
CPU usage seems to be about the same:
image

My instance setup with proxies is also working correcly; Merging.

mrpaulblack (approved):
LGTM :+1:

Issue #2643, opened 2023-08-12
Bug: archive.is engine Timeout / blocked by CAPTCHA 


bonswouar:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2023.8.8+bcaaae699

**How did you install SearXNG?**
searxng-docker

**What happened?**
Tried to use archive.is engine, but it always timeout

**How To Reproduce**
Search for "!ai samsung.com"

**Expected behavior**
Shouldn't timeout

**Additional context**
My Searxnng instance is on a dedicated server. 
But I notice I also struggle navigating to archive.is directly : with my home connection _(using a shared Starlink IP)_ it seems to infinite loop on the captcha page. But no problem using cell network.. So I guess they might have weird IP restrictions or something?

**Technical report**

Error
  * Error: httpx.ReadTimeout
  * Percentage: 65
  * Parameters: `(None, None, 'archive.is')`
  * File name: `searx/search/processors/online.py:118`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Error
  * Error: httpx.TimeoutException
  * Percentage: 35
  * Parameters: `(None, None, None)`
  * File name: `searx/search/processors/online.py:118`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

unixfox:
Related information: https://old.reddit.com/r/DataHoarder/comments/13g4htv/cloudflare_dns_blocking_archiveis/

bonswouar:
@unixfox thanks for the link it's interesting! Although it seems my issue isn't exactly the same, I can nslookup/ping archive.is successfully (using Hetzner's dns server apparently hm)
But they probably just have different types of IP restrictions..

return42:
Sadly archive.is is blocked by a CAPTCHA and I don't have a clue how we can avoid this CAPTCHA

![image](https://github.com/searxng/searxng/assets/554536/f956ad33-f7fd-41a6-9a98-f135177e2691)

return42:
@bonswouar I'm sorry, but we do not have a solution for this issue at hand .. in #2645 I will drop the XPath configuration for this engine (makes no sense to hold a configuration, that do no longer work).

[...]

PR #2645, opened 2023-08-13
[drop] engine archive.is / blocked by CAPTCHA [1]

return42:
## What does this PR do?

engine archive.is

Sadly archive.is is blocked by a CAPTCHA that can't be avoid (at least in a XPath engine).

[1] https://github.com/searxng/searxng/issues/2643

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -346,25 +346,6 @@ engines:
     engine: archlinux
     shortcut: al
 
-  - name: archive is
-    engine: xpath
-    search_url: https://archive.is/search/?q={query}
-    url_xpath: (//div[@class="TEXT-BLOCK"]/a)/@href
-    title_xpath: (//div[@class="TEXT-BLOCK"]/a)
-    content_xpath: //div[@class="TEXT-BLOCK"]/ul/li
-    categories: general
-    timeout: 7.0
-    disabled: true
-    shortcut: ai
-    soft_max_redirects: 1
-    about:
-      website: https://archive.is/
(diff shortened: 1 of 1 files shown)


unixfox:
Haven't had the time to review the PR though...

I want to be sure that it's a captcha blockage worldwide or region wise. Because if it's not worldwide there is no point in removing the engine if it works for most of the people.

return42:
> Haven't had the time to review the PR though...

I'm sorry .. if you want, we can revert the merge.  But at https://searx.space/ (on the engine tab) there is no instance without the timeout issue described in https://github.com/searxng/searxng/issues/2643

For those who want to test; you can add the XPath engine to your local settings.yml:

```yaml

  - name: archive is
    engine: xpath
    search_url: https://archive.is/search/?q={query}
    url_xpath: (//div[@class="TEXT-BLOCK"]/a)/@href
    title_xpath: (//div[@class="TEXT-BLOCK"]/a)
    content_xpath: //div[@class="TEXT-BLOCK"]/ul/li
    categories: general
    timeout: 7.0
    disabled: true
    shortcut: ai
    soft_max_redirects: 1
    about:
      website: https://archive.is/
      wikidata_id: Q13515725
      official_api_documentation: https://mementoweb.org/depot/native/archiveis/
      use_official_api: false
      require_api_key: false
      results: HTML
```

unixfox:
Just tested again now and it doesn't seems to work great... I guess we can keep it removed

```
archive is ( timeout )
```

PR #2646, opened 2023-08-13
[fix] preferences: change engine table category colspan to 8

mrpaulblack:
## What does this PR do?

* this is a small fix to increase the colspan of the category in engine preferences from 7 to 8, since there was a column added
=> fixing a small fallout from 473129031710973e3c02e7178085b751b5e821ed

Fixing small visual bug.

`make run` and check out engines tab in preferences page

## Author's checklist

Before this PR:
image

After this PR:
image


return42 (approved):
Oops .. I didn't notice that at all, thanks for the correction :+1:

Bnyro:
Oh, thank you for taking care of that!

PR #2655, opened 2023-08-17
Add pkg.go.dev engine

playday3008:
## What does this PR do?

Add's pkg.go.dev engine

New source to search in

https://github.com/searxng/searxng#codespaces


return42:
Hi @playday3008, thanks a lot for adding go package search to SearXNG :+1: .. squashed your two commits into one and I changed the shortcut from `pgd` to `pgo` which I think is more memorable /**P**ackage**GO**.

Issue #2353, opened 2023-04-17
Preferences aren't saved on page refresh


neurodiverseEsoteric:
searx goes back to defult preferences whenever I try to navigate to another mode (images/videos/it/etc), or the next page, or alter the search query...and it's always in gibberish, there's no way to make human-readable-text-parameters like autocomplete=off, openinnewtab=on, or whatever...

return42:
> searx goes back to defult preferences whenever 

What instance do you use? / `server: base_url` needs to be set --> https://docs.searxng.org/admin/engines/settings.html#server

neurodiverseEsoteric:
[...]

Issue #1365, opened 2022-06-24
Settings-URL dont work / gets ignored, Settings always default from the instance.


solisinvictum:
Hello Guys,

yesterday i found out that searxng have implemented a center view (https://github.com/searxng/searxng/issues/1175#issuecomment-1164997639).

And started using it. But know i found out, that this dont work right.

After some experiments, i found out this is because  the Setting-URL dont work / gets ignored.

Than i thought it is a problem with the instance i choosed (https://searx.tiekoetter.com). 

But sadly this occurs on every instance listed here: https://searx.space/ (i tested all DE ones). 

Saving the Setting with a Cookie works, but on my system the cookie gets isolated in each tab, and removed after i left the site or close it. 

Here is a example: 

[...]

Issue #1316, opened 2022-06-12
Simple theme has inverted checkboxes


ghost:
I have noticed that when I check a box in the settings, the option is actually disabled, while blank boxes are taken as enabled options. This is also happening in the engines selection page. Is this the expected behavior? Thanks.

dalf:
This is a legally from a long time.
It requires to break the preferences to fix the issue as far I remember.
It has to be fix.

(I have updated the issue #818 which was describing this issue among others)

ghost:
@dalf thanks for the quick reply, so I will be waiting for the fix 👍

return42:
I close this issue, the topic is covered by:

- https://github.com/searxng/searxng/issues/818
- https://github.com/searxng/searxng/issues/1440

PR #2660, opened 2023-08-20
[feat] settings: replace boolean select preferences with checkboxes

### LINKED ISSUE #1440: Preferences options aren't consistent.

2022.07.04-7a9beb4f 

Used https://searx.tiekoetter.com/ instance
**What happened?**
Some options have checkboxes, some- "On" "Off" selection menus, while others have "Disabled" "Enabled".

**How To Reproduce**

Go to options page, settings have these three types of the same type of setting (On or Off).
**Expected behavior**

Settings should have the same representation of the same type of option.
**Screenshots & Logs**
image
image

**Additional context**

Bnyro:
## What does this PR do?
* replace boolean select preferences with checkboxes

* currently, boolean preferences that are not about enabling or disabling plugins are visible to the user as a dropdown - however, that's quite inconvenient since checkbox/switches are more common for boolean preferences, and also used for enabling and disabling plugins

* open the settings and navigate to user interface or privacy

## Related issues
closes #1440

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -383,7 +383,8 @@ class Preferences:
                     '0': False,
                     '1': True,
                     'True': True,
-                    'False': False
+                    'False': False,
+                    'on': True
                 }
             ),
             'method': EnumStringSetting(
@@ -412,7 +413,8 @@ class Preferences:
                     '0': False,
                     '1': True,
                     'False': False,
-                    'True': True
+                    'True': True,
+                    'on': True
                 }
             ),
             'doi_resolver': MultipleChoiceSetting(
@@ -432,7 +434,8 @@ class Preferences:
                     '0': False,
                     '1': True,
                     'False': False,
-                    'True': True
+                    'True': True,
+                    'on': True
                 }
             ),
             'advanced_search': MapSetting(
@@ -454,7 +457,8 @@ class Preferences:
                     '0': False,
                     '1': True,
                     'True': True,
-                    'False': False
+                    'False': False,
+                    'on': True
                 }
             ),
             'infinite_scroll': MapSetting(
@@ -465,7 +469,8 @@ class Preferences:
                     '0': False,
                     '1': True,
                     'True': True,
-                    'False': False
+                    'False': False,
+                    'on': True
                 }
             ),
             # fmt: on
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -9,6 +9,7 @@ from base64 import urlsafe_b64encode, urlsafe_b64decode
 from zlib import compress, decompress
 from urllib.parse import parse_qs, urlencode
 from typing import Iterable, Dict, List, Optional
+from collections import OrderedDict
 
 import flask
 import babel
@@ -23,6 +24,18 @@ from searx.engines import DEFAULT_CATEGORY
 
 COOKIE_MAX_AGE = 60 * 60 * 24 * 365 * 5  # 5 years
 DOI_RESOLVERS = list(settings['doi_resolvers'])
+
+MAP_STR2BOOL: Dict[str, bool] = OrderedDict(
+    [
+        ('0', False),
+        ('1', True),
+        ('on', True),
+        ('off', False),
+        ('True', True),
+        ('False', False),
+        ('none', False),
+    ]
+)
 
 
 class ValidationException(Exception):
@@ -192,6 +205,26 @@ class MapSetting(Setting):
             raise ValidationException('Invalid choice: {0}'.format(data))
         self.value = self.map[data]
         self.key = data  # pylint: disable=attribute-defined-outside-init
+
+    def save(self, name: str, resp: flask.Response):
+        """Save cookie ``name`` in the HTTP response object"""
+        if hasattr(self, 'key'):
+            resp.set_cookie(name, self.key, max_age=COOKIE_MAX_AGE)
+
+
+class BooleanSetting(Setting):
+    """Setting of a boolean value that has to be translated in order to be storable"""
+
+    def normalized_str(self, val):
+        for v_str, v_obj in MAP_STR2BOOL.items():
+            if val == v_obj:
+                return v_str
+        raise ValueError("Invalid value: %s (%s) is not a boolean!" % (repr(val), type(val)))
+
+    def parse(self, data: str):
+        """Parse and validate ``data`` and store the result at ``self.value``"""
+        self.value = MAP_STR2BOOL[data]
+        self.key = self.normalized_str(self.value)  # pylint: disable=attribute-defined-outside-init
 
     def save(self, name: str, resp: flask.Response):
         """Save cookie ``name`` in the HTTP response object"""
@@ -375,17 +408,9 @@ class Preferences:
                 locked=is_locked('autocomplete'),
                 choices=list(autocomplete.backends.keys()) + ['']
             ),
-            'image_proxy': MapSetting(
+            'image_proxy': BooleanSetting(
                 settings['server']['image_proxy'],
-                locked=is_locked('image_proxy'),
-                map={
-                    '': settings['server']['image_proxy'],
-                    '0': False,
-                    '1': True,
-                    'True': True,
-                    'False': False,
-                    'on': True
-                }
+                locked=is_locked('image_proxy')
             ),
             'method': EnumStringSetting(
                 settings['server']['method'],
@@ -406,16 +431,9 @@ class Preferences:
                 locked=is_locked('theme'),
                 choices=themes
             ),
-            'results_on_new_tab': MapSetting(
+            'results_on_new_tab': BooleanSetting(
                 settings['ui']['results_on_new_tab'],
-                locked=is_locked('results_on_new_tab'),
-                map={
-                    '0': False,
-                    '1': True,
-                    'False': False,
-                    'True': True,
-                    'on': True
-                }
+                locked=is_locked('results_on_new_tab')
             ),
             'doi_resolver': MultipleChoiceSetting(
                 [settings['default_doi_resolver'], ],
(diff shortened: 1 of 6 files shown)


return42:
Hi @Bnyro, I dropped commit 93e53ce `[build] /static` --> static builds are only needed when static files from [searx/static](https://github.com/searxng/searxng/tree/master/searx/static/themes/simple/src) had been patched.

I placed commit ["WIP: optimze HTML templates for readability and less white space"](https://github.com/searxng/searxng/pull/2660/commits/0b945627ad76f6869dd9faff3f0704856f23e0f6)
on top / it contains some nits for a "better" templates .. if it is OK for you I would squash this commit into your commit.  

On my tests I found an issue / you can reproduce by:

1. ``make run`` opens http://127.0.0.1:8888/ in your WEB browser
2. UI: go to *preferences*, reset cookies
3. UI: activate *infinite scroll* and save cookie
4. UI: go to *preferences*
4.a UI: checkbox of *infinite scroll* should be activated
4.b UI: in "COOKIES" tab cookie (name) ``infinite_scroll`` value is ``on`` (in branch master ``1`` is stored)

5. in settings.yml in ``ui:`` section activate ``infinite_scroll: true``
6. UI: go to *preferences*, reset cookies
7. UI: de-activate *infinite scroll* and save cookie
8. UI: go to *preferences*
8.a UI: checkbox of *infinite scroll* should be de-activated / **BUG:** *infinite scroll* is activated
8.b UI: in "COOKIES" tab cookie (name) ``infinite_scroll`` should contain a false value / **BUG:** is set to ``on`` in branch master ``0`` is stored.

----

IDK if we better subject the module to a review or we fix issue related to this PR .. the `preferences.py` has it quirks (see below) .. I while ago I tried to make a [review](https://github.com/return42/searxng/tree/review-preferences) but didn't finished my work since for lack of time.

Whether we focus on this PR or we do a review, in both cases we have to test a lot, which is why I would prefer the review ... opinions?


----

Some words about the `preference.py` module, we have received several issues related to this module and its usage in the `webapp.py`:

- https://github.com/searxng/searxng/issues/2353#issuecomment-1544534632
- https://github.com/searxng/searxng/issues/1440
- https://github.com/searxng/searxng/issues/1316
- https://github.com/searxng/searxng/issues/1365
- https://github.com/searxng/searxng/issues/818
- https://github.com/searxng/searxng/issues/217
- https://github.com/searxng/searxng/issues/142
- https://github.com/searxng/searxng/discussions/2591
- https://github.com/searxng/searxng/discussions/2559

I have also noticed a few quirks myself, for example sometimes a string value of "1", sometimes "on" or "true" for a bool value is stored in the cookie.  This happens among other things when you load the Properites from a URL.

The implementation in the module has grown over the last 7 years and the class model would need to be revisited.

- On the base class `Settings` for example classes like StringSetting are implemented, there should still be a BoolSetting class. 

- The class BooleanChoices represents a group of bool settings and should consist of an aggregation of BoolSetting.  A more appropriate name for the class BooleanChoices might be BooleanSet.

- class `MapSetting` should be based on a ordered dictionary, in a ordered dict the first match is the normalized value that is stored in the COOKIE, in the HTML form and in the URL.

Here an example how str-values of a boolean can be normalized:

```python
MAP_STR2BOOL = OrderedDict(
    [
        ('0', False),
        ('1', True),
        ('on', True),
        ('off', False),
        ('True', True),
        ('False', False),
        ('', False),  # unset is mapped to False
        ('none', False),  # None is mapped to False
    ]
)

def get_normalized_str(value):
    for v_str, v_obj in MAP_STR2BOOL.items():
        if value == v_obj:
           return v_str



>>> get_normalized_str(True)
'1'
>>> get_normalized_str(False)
'0'
>>> get_normalized_str(MAP_STR2BOOL.get('off'))
'0'
>>> get_normalized_str(MAP_STR2BOOL.get('on'))
'1'
>>> get_normalized_str(MAP_STR2BOOL.get(''))
'0'
>>> get_normalized_str(MAP_STR2BOOL.get('True'))
'1'
>>> get_normalized_str(MAP_STR2BOOL.get('none'))
'0'
```

Bnyro:
> Hi @Bnyro, I dropped commit [93e53ce](https://github.com/searxng/searxng/commit/93e53ced0eb96f820d554473c471c477d4419861) `[build] /static` --> static builds are only needed when static files from [searx/static](https://github.com/searxng/searxng/tree/master/searx/static/themes/simple/src) had been patched.
> 
> I placed commit ["WIP: optimze HTML templates for readability and less white space"](https://github.com/searxng/searxng/pull/2660/commits/0b945627ad76f6869dd9faff3f0704856f23e0f6) on top / it contains some nits for a "better" templates .. if it is OK for you I would squash this commit into your commit.
> 
Alright, fine by me :+1: 
> On my tests I found an issue / you can reproduce by:
> 
Okay, that's unfortunate. I didn't have much time to test the changes yet - I hoped it would just all go well.
I can looks into fixing that issue, that wouldn't be a problem by me.
> IDK if we better subject the module to a review or we fix issue related to this PR .. the `preferences.py` has it quirks (see below) .. I while ago I tried to make a [review](https://github.com/return42/searxng/tree/review-preferences) but didn't finished my work since for lack of time.
> 
> Whether we focus on this PR or we do a review, in both cases we have to test a lot, which is why I would prefer the review ... opinions?
> 
I think there are two different things that should generally be looked into:
1. The preferences logic in `webapp.py`. As you already mentioned, there's a lot of boilerplate code and things could be made simpler and clearer in many places.
2. The templates of single preferences. I wondered why we just don't use macros for simple preference, but instead use the same template for each `select` dropdown with hardcoded values. It would make the template code much easier to overview and fix issues if things don't need to get changed per preference in the dedicated template.

However, I think that this would go beyond the goals / the intention of this pull request. I'd love to look into cleaning up the preference templates with macros, but I think it's better to do that once the changes to rule out the boolean select dropdowns are working well and merged.

Nevertheless I agree that we should already simplify the handling of boolean preferences in `webapp.py` in this PR, since we now have 5 lines per preference to map the possible values to `True` or `False`.

TLDR(imo):
1. Simplify the changes in this PR as much as possible by removing boilerplate code for a boolean preference and fix the issue you found.
2. Address other quirks in separate pull requests.

return42:
> Alright, fine by me 👍

OK, I squashed my modifications into your commit [diff --> bee980f4c..8526af2](https://github.com/searxng/searxng/compare/bee980f4c..8526af2)

>> Whether we focus on this PR or we do a review, in both cases we have to test a lot, which is why I would prefer the review ... opinions?

> Simplify the changes in this PR as much as possible by removing boilerplate code for a boolean preference and fix the issue you found.

I stopped my review for now .. can you please have a look about the issues reported / thanks :+1:

----

> The templates of single preferences. I wondered why we just don't use macros for simple preference, but instead use the same template for each select dropdown with hardcoded values. It would make the template code much easier to overview and fix issues if things don't need to get changed per preference in the dedicated template.

Welcome to the rabbit hole ;) .. we have this ugly *reverse-checkbox* in the preferences

https://github.com/searxng/searxng/blob/4b42644579581699ad1d4b7974610933277dd021/searx/templates/simple/preferences.html#L30-L37

And its CSS ..

https://github.com/searxng/searxng/blob/4b42644579581699ad1d4b7974610933277dd021/searx/static/themes/simple/src/less/toolkit.less#L426-L452

And we have a checkbox macro

https://github.com/searxng/searxng/blob/4b42644579581699ad1d4b7974610933277dd021/searx/templates/simple/macros.html#L55-L63

I can't remember why we have the *reverse-checkbox* in the preferences but AFAIK its needed for checkboxes in the engines tab (and for the plugins).

this is all very confusing ... I've been wanting to clean this up for a long time .. but my revisions often stuck in lack of time.

Bnyro:
> I stopped my review for now .. can you please have a look about the issues reported / thanks 👍
>
The culprit is that the checkbox value is only sent as form data if the checkbox is checked, and otherwise not. We're now checking each boolean preference, and set the value to false when it's not been part of the form data: see [here](https://github.com/searxng/searxng/pull/2660/files#diff-c68b1e3201d656b33513a919fceae6b6fb776aa31f2b8f5aa80ac8eb4da64c0dR527).

It's certainly not really a perfect solution, however that's the only idea I could come up with :/

dalf:
When I look at these two lines : 

In this PR:
https://github.com/searxng/searxng/blob/44e2dd4a389aab9dc2beae666b0b62119f2b08d2/searx/preferences.py#L36

In the master branch:
https://github.com/searxng/searxng/blob/4b42644579581699ad1d4b7974610933277dd021/searx/preferences.py#L464

It seems the meaning of the value `''` should be provided to the BooleanSetting class.

I don't know if this fix the issue.

dalf:
> UI: go to preferences
> 8.a UI: checkbox of infinite scroll should be de-activated / BUG: infinite scroll is activated
> 8.b UI: in "COOKIES" tab cookie (name) infinite_scroll should contain a false value / BUG: is set to on in branch master 0 is stored.

I confirm these bugs are fixed with the commit 30fe04c

searx/preferences.py
```diff
@@ -199,6 +212,26 @@ def save(self, name: str, resp: flask.Response):
...
 
 
+class BooleanSetting(Setting):
+    """Setting of a boolean value that has to be translated in order to be storable"""
+
+    def get_bool_value(self, data):
+        for v_str, v_obj in MAP_STR2BOOL.items():
+            if data == v_str:
+                return v_obj
+        raise ValueError("Invalid value for a boolean setting!")
```
return42: the intention of the OrderedDict that maps str to boolean is to normalize the str representation of the boolean value .. this function does the opposite and can be replaced `MAP_STR2BOOL[data]` .. I think this issue was unintentional :) .. I will fix it in a WIP commit.

When we store `self.key = data` (see line 227) we store the value from the HTML form in the cookie ..

image

what we want to store is the first representation of boolean `True` in the OrderedDict, what is 1:

image

return42:
@Bnyro  see my remark about the OrderedDict in the code .. I squashed my fix into your commit --> [diff](https://github.com/searxng/searxng/compare/30fe04ce6000990f058a851edc563c75ab4458db..5d3a620384c04a5d72284a917a5f1773f417925e)

Before we merge, can please have a final look / thanks!

---

> The culprit is that the checkbox value is only sent as form data if the checkbox is checked, and otherwise not. 

Wow ... that was I struggled all the time but I haven't seen it / thanks for pointing out :+1: 

> It's certainly not really a perfect solution, however that's the only idea I could come up with :/

Don't worry, there more quirks in the prefs we should eliminate / I had given a few [hints above](https://github.com/searxng/searxng/pull/2660#issuecomment-16875979989)  .. but this is not in the scope of this PR :)

searx/preferences.py
Bnyro: Makes sense, thanks for pointing that out!

Bnyro:
@return42 Thanks for the review and your tips, looks good to me from a code perspective now!

I'm in vacation now as announced via Matrix, and don't have my notebook with me (if I had it with me, I'd spend half of the day coding instead of enjoying the vacation, so I don't take it with me anymore on purpose), so I'm unable to test this currently unfortunately.

I'm certain you got it working right :)

PR #2667, opened 2023-08-22
[fix] openstreetmap engine

dalf:
## What does this PR do?

I'm not sure if this is a change of the API: the extratags can be either None or a directory. When the engine was written, extratags was either a dictionnary, or not set ( `"extratags" not in result` ).

This PR avoid a crash when extratags is None in the json response.

Fix openstreetmap engine

Search for `!osm gare du nord`

### DIFF
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -162,8 +162,15 @@ def response(resp):
             }
         )
 
+    # simplify the code below: make sure extratags is a dictionnary
+    for result in nominatim_json:
+        if not isinstance(result.get('extratags'), dict):
+            result["extratags"] = {}
+
+    # fetch data from wikidata
     fetch_wikidata(nominatim_json, user_language)
 
+    # create results
     for result in nominatim_json:
         title, address = get_title_address(result)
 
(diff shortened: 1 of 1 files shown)


searx/engines/openstreetmap.py
```diff
@@ -349,19 +350,23 @@ def get_links(result, user_language):
...
-                {
-                    'label': get_key_label(k, user_language),
-                    'url': url,
-                    'url_label': url_label,
-                }
-            )
-            link_keys.add(k)
+        extratags = result.get('extratags')
+        if not isinstance(extratags, dict):
+            continue
```
dalf: Actually this condition should be before the loop : no need to iterate over VALUE_TO_LINK

searx/engines/openstreetmap.py
```diff
@@ -348,9 +355,13 @@ def get_links(result, user_language):
     """Return links from result['extratags']"""
     links = []
     link_keys = set()
-    for k, mapping_function in VALUE_TO_LINK.items():
-        raw_value = result['extratags'].get(k)
-        if raw_value:
+    extratags = result['extratags']
+    if extratags:
```
Bnyro: Just a small comment that's not too important:

Probably nobody agrees with me on that, but I'd prefer doing 
```python
if not extratags:
    return [], set()
```
instead of doing that if clause here, so that we have one indentation less.

Bnyro:
Tested, works as expected now, well done :)

searx/engines/openstreetmap.py
dalf: Done

PR #2648, opened 2023-08-15
[feat] engine: implementation of mwmbl

Bnyro:
## What does this PR do?
* add a new engine, mwmbl, for general results

## Why is this change important?
* the results are pretty good for a foss non-meta search engine

## How to test this PR locally?
!mwm hello world

## Related issues
closes #1654

PR #2670, opened 2023-08-23
[mod] engine mwmbl: add auto completion and link to official api docs

Bnyro:
## What does this PR do?
* add link to official mwmbl api docs
* add auto completions using mwmbl

* `make run`
* go to settings and set autocomplete to `mwmbl`
* start typing to search for something and observe the suggestions

see https://github.com/searxng/searxng/pull/2648#issuecomment-1689920490

### DIFF
--- a/docs/admin/settings/settings_search.rst
+++ b/docs/admin/settings/settings_search.rst
@@ -35,6 +35,7 @@
   - ``dbpedia``
   - ``duckduckgo``
   - ``google``
+  - ``mwmbl``
   - ``startpage``
   - ``swisscows``
   - ``qwant``
--- a/docs/dev/engines/online/mwmbl.rst
+++ b/docs/dev/engines/online/mwmbl.rst
@@ -1,0 +1,27 @@
+.. _Mwmbl engine:
+
+============
+Mwmbl Engine
+============
+
+.. contents::
+   :depth: 2
+   :local:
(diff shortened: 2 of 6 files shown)


return42:
Hi @Bnyro, I added commit with docs on top of your branch .. I tested the autocomplete but sometimes I get strange results from Mwmbl. By example when there are no results for a sarch term like [fooo](https://mwmbl.org/?q=fooo) there is this one autocomplete value:

image

Bnyro:
> Hi @Bnyro, I added commit with docs on top of your branch .. I tested the autocomplete but sometimes I get strange results from Mwmbl. By example when there are no results for a sarch term like [fooo](https://mwmbl.org/?q=fooo) there is this one autocomplete value:
> 
Oh okay, thanks for spotting that! I only noticed some results that were called something like `go: https://example.com`, but apparently there are more of that. I'll see whether I can find all these possible `extra suggestions` in their code, otherwise we should filter them with a RegEx in my opinion, e.g. `\w+(: )\w+`.

Bnyro:
Okay, they only use `go: ` and `search: ` as special results, for reference: https://github.com/mwmbl/mwmbl/blob/18dc760a3402c74803823a94310ef058a60d57b4/mwmbl/tinysearchengine/rank.py#L135

PR #2671, opened 2023-08-23
[mod] random answerer: add random hex color generator

Bnyro:
## What does this PR do?
* adds a  random hex color generator as answerer

* I'd consider this useful for myself when working on CSS projects to quickly get a random color that could be used

* `make run`
* search `random color`


return42 (approved):
nice :smile:

PR #2681, opened 2023-08-25
[fix] reduce error logs related to X-Forwarded-For & X-Real-IP headers

return42:
To test start a `make run` instance and query multiple times for `ip` (the answerer), you should see only once the error message(s):

    ERROR   searx.botdetection            : X-Forwarded-For header is not set!
    ERROR   searx.botdetection            : X-Real-IP header is not set!

- https://github.com/searxng/searxng-docker/discussions/174
- https://github.com/searxng/searxng-docker/pull/162#issuecomment-1692834457

### DIFF
--- a/searx/botdetection/_helpers.py
+++ b/searx/botdetection/_helpers.py
@@ -58,6 +58,15 @@ def get_network(real_ip: IPv4Address | IPv6Address, cfg: config.Config) -> IPv4N
     return network
 
 
+_logged_errors = []
+
+
+def _log_error_only_once(err_msg):
+    if err_msg not in _logged_errors:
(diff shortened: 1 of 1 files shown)


dalf:
This should work.

I still don't understand why this is an error since the limiter works with only one of the header : this is misleading.

Would it better if an error is logged only:
* if the two headers are missing?
* or if two headers are set to different values? (already done)

dalf (approved):

return42:
> if the two headers are missing?

Yes, you are right .. this seems the best solution .. in your review, would you modify the commit accordingly? .. otherwise I can do it in the next days / thanks for insisting :+1:

PR #2682, opened 2023-08-25
[mod] searx.network: memory optimization

dalf:
## What does this PR do?

Avoid to create a SSLContext in AsyncHTTPTransportNoHttp

See:
* https://github.com/encode/httpx/blob/0f61aa58d66680c239ce43c8cdd453e7dc532bfc/httpx/_transports/default.py#L271
* https://github.com/encode/httpx/issues/2298

Each SSLContext consume more than 500kb fold by the number of network, this PR should reduce the memory by few megabytes.

* check that engine still can't access HTTP
* check the memory usage


Bnyro (approved):
I have to admit that I'm not really familiar with that part of the code yet, but the changes seem reasonable to me and I couldn't find any issues when testing it.

PR #2683, opened 2023-08-25
[dev-env] upgrade Sphinx-doc 7.2.3 and unpin at v7.1.2 on py3.8

return42:
upgrade Sphinx-doc 7.2.3 and unpin at v7.1.2 on py3.8

- Sphinx-doc 7.2.0 drops py3.8 support [1][2]
- last version with py3.8 support is 7.1.2

Many LTS distributions still have py3.8 which EOL is in 2024-10 [3].

To continue to support a development environment on py3.8 the rigid dependency
in the development environment is unpinned in py3.8 / environment markers [4].

To get 7.2.3. work, a fix in sphinx-notfound-page is needed.

[1] https://github.com/searxng/searxng/pull/2658#issuecomment-1684867270
[2] https://github.com/sphinx-doc/sphinx/issues/11621
[3] https://devguide.python.org/versions/#supported-versions
[4] https://peps.python.org/pep-0508/#environment-markers
[5] https://github.com/readthedocs/sphinx-notfound-page/issues/219

----

### notes for the reviewers ...

What do you think, is it OK to unpin Sphinx-doc (in the develop environment)? .. otherwise we have to drop py3.8 support .. or do you know a better solution for this conflict of interest?

----

Closes: https://github.com/searxng/searxng/pull/2680

### DIFF
--- a/docs/_themes/searxng/theme.conf
+++ b/docs/_themes/searxng/theme.conf
@@ -4,3 +4,4 @@ stylesheet = searxng.css
 
 [options]
 touch_icon =
+globaltoc_maxdepth = 5
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,15 +7,16 @@ splinter==0.19.0
 selenium==4.11.2
 twine==4.0.2
 Pallets-Sphinx-Themes==2.1.1
-Sphinx==7.1.2
+Sphinx==7.1.2; python_version == '3.8'
+Sphinx==7.2.3; python_version > '3.8'
 sphinx-issues==3.0.1
 sphinx-jinja==2.0.2
 sphinx-tabs==3.4.1
 sphinxcontrib-programoutput==0.17
 sphinx-autobuild==2021.3.14
-sphinx-notfound-page==0.8.3
+sphinx-notfound-page==1.0.0rc1
 myst-parser==2.0.0
-linuxdoc==20230629
+linuxdoc==20230827
 aiounittest==1.4.2
 yamllint==1.32.0
 wlc==1.13


Bnyro:
I didn't know Python 3.8 is still a thing in some distributions - I already wondered why we're having a CI for that here.

I think it's okay to unpin the dependency then. We just have to watch out that new releases of sphinx don't introduce any breaking changes that would require changes on how we use Sphinx, so that we couldn't support both versions at a time anymore.

We should make sure to drop Python3.8 as soon as it's no longer officially supported (or already a bit earlier) in my opinion, since it's really lacking a bunch of features and fixes.

And I'm still waiting for the day where we can use `match case` here, which Python already supports for multiple years! (since 3.10) :)

Bnyro (approved):

return42:
> We should make sure to drop Python3.8 as soon as it's no longer officially supported (or already a bit earlier) in my opinion, since it's really lacking a bunch of features and fixes.

.. I fully understand the developer site .. its always hard to support "old" ecosystems

> We just have to watch out that new releases of sphinx don't introduce any breaking changes that would require changes on how we use Sphinx

.. welcome in the dependency hell while supporting old stuff :sos:  .. I like to break old habits and keep things simple .. but on the other hand we have SearXNG installations on LTS distributions (e.g. ubu20.04 with py3.8).  I think we would scare away some maintainers if we discontinue py3.8 before the EOL ... if we discontinue, we would have to announce it well in advance .. but it's better to stick to the official EOL / I think.

dalf:
I don't remember exactly, but do Sphinx updates often break the compilation of documentation?

If yes, we can pin the two versions according to the Python version:

```
Sphinx==7.1.2; python_version == '3.8'
Sphinx==7.2.3; python_version > '3.8'
```

I don't know if this is practical / better or not

return42:
> I don't remember exactly, but do Sphinx updates often break the compilation of documentation?

Yes, the doc build is often broken on updates of sphinx .. but its an issue of the dependencies (extensions) we use: 

https://github.com/searxng/searxng/blob/0647f83a3ea3ce2bd8c5f5fc59bb676be2a790e2/requirements-dev.txt#L9-L18

these extensions are often pinned on a patch or minor level of the Sphinx version.

Or if not pinned, they cause an issue like [here in the build](https://github.com/searxng/searxng/actions/runs/5978209391/job/16233376631?pr=2683#step:6:76) :) .. fixed in 

- https://github.com/readthedocs/sphinx-notfound-page/issues/219

----

**update:** you are right, we dropped 3.7 before its EOL

- https://github.com/searxng/searxng/discussions/2356



> If yes, we can pin the two versions according to the Python version:

This is a good suggestion to handle the py-version issue :+1: .. I will give [Environment Markers](https://peps.python.org/pep-0508/#environment-markers) a try.

return42:
> If yes, we can pin the two versions according to the Python version:

done .. further: To get 7.2.3. work, a fix in sphinx-notfound-page is needed.

But still have an issue (see [build](https://github.com/searxng/searxng/actions/runs/5978209391/job/16233376631?pr=2683#step:6:1078)) --> https://github.com/readthedocs/sphinx-notfound-page/issues/219#issuecomment-1694670929

PR #2685, opened 2023-08-26
refactor `searx.network`

dalf:
## What does this PR do?

See https://github.com/dalf/searxng/wiki/Branch-searx_network_rework to get an overview of the architecture change.

In few words:
* remove async code (so we can drop uvloop)
* there is better separation of concern
* it keeps one HTTP client during the execution of one user query in one engine (this is optional). This might help in some scenarios ; if for example an engine returns a cookie linked to one IP address.
* there are different retry mechanisms.

For each network, there is a new parameter `retry_strategy` which describes what to do when an outgoing HTTP request fail and `retries > 1`:
* `DIFFERENT_HTTP_CLIENT` (default value): this is the current implementation in the master branch. If a request fails, then another (proxy, IP address) is tried. If an engine sends two HTTP requests, the first one goes through a proxy, the second request goes through another proxy (if two proxies or more are configured).
* `ENGINE`: the (proxy, IP address) is kept during the process of one user query. If a HTTP request fails, the engine restart the process of the query with another (proxy, IP address).
* `SAME_HTTP_CLIENT`: If a HTTP request fails, the request is sent again with the same (proxy, IP address).

All these strategies respect the user query timeout.

This is unfinished, but I'm not sure if this should be merged once done.

To be written, non-exhaustive list :

* [x] check outgoing connection with one proxy
* [x] check outgoing connection with multiple proxies
* [ ] check outgoing connection with multiple IP addresses
* [ ] check outgoing connection with retries > 1 and failing request, to check retries works as expected
* [ ] check the image proxy does not have a memory leak

## Author's checklist
 
About async code: it code was introduced when the target was to fully switch to async (async code in the engine, use [quart](https://github.com/searx/searx/pull/1724), etc...). However, async won't fit the use case here: async is cooperative multitasking, and between the very small and fast function calls to deal with HTTP/2 and the very long pause to parse the engine result, I'm afraid it will provide slow down the response time (or it will break as soon we are not very careful).

The first commit brings NetworkContext but keep async code. The second commit removes async code.

The only point of async is to send multiple requests at the same time from one engine:
* the bing engine **was** using it to [resolve the redirection URL in the results](https://github.com/searxng/searxng/blob/0647f83a3ea3ce2bd8c5f5fc59bb676be2a790e2/searx/engines/bing.py#L235-L242). https://github.com/searxng/searxng/pull/2686 fixes that.
* the baidu engine: it is not possible to make it work after a few request.

---

This PR brings 1401 new lines of code, but there is far more documentation and tests than before.

## Related

- https://github.com/searxng/searxng/discussions/2684
- https://github.com/searxng/searxng/discussions/3001


searx/network/network.py
```diff
@@ -142,188 +162,83 @@ def iter_proxies(self):
...
-        return await self.call_client(False, method, url, **kwargs)
-
-    async def stream(self, method, url, **kwargs):
-        return await self.call_client(True, method, url, **kwargs)
-
-    @classmethod
-    async def aclose_all(cls):
-        await asyncio.gather(*[network.aclose() for network in NETWORKS.values()], return_exceptions=False)
+    def get_context(self, start_time: int, timeout: float) -> NetworkContext:
+        context_cls = RETRY_STRATEGY.get('engine') or NetworkContextRetryFunction
```
dalf: currently hard coded, it must be parameterisable.

return42:
I would suggest to implement a class for the networks management / what we have is a global dict:

```python
NETWORKS: Dict[str, 'Network'] = {}
```

this dict is initialized with default when the module is imported:

```python
NETWORKS[DEFAULT_NAME] = Network()
```

some parts of a "network management constructor" is done in the initialization ..

```python
def initialize(settings_engines=None, settings_outgoing=None):
    ...
    NETWORKS.clear()
    NETWORKS[DEFAULT_NAME] = new_network({}, logger_name='default')
    NETWORKS['ipv4'] = new_network({'local_addresses': '0.0.0.0'}, logger_name='ipv4')
    NETWORKS['ipv6'] = new_network({'local_addresses': '::'}, logger_name='ipv6')

```

and for the *management* of the networks we have some lose functions like:

```python
def get_network(name=None):
    return NETWORKS[name or DEFAULT_NAME]
...
def check_network_configuration():
    exception_count = 0
    for network in NETWORKS.values():
        ...
```

IMO we should have a class like *NetworkManager* from which a instance `NETWORKS` can be build:

- the initialization of the manager is done in the class constructor
- global functions like `get_network(name=None)` are implemented in methods like `NetworkManager.get(name=None)`

dalf:
The line below makes code outside the webapp works out of the box : no need to do any kind of initialization, just use `searx.network.get(...)` same as `requests.get(...)` :

```python
NETWORKS[DEFAULT_NAME] = Network()
```

For example, the traits are fetched outside the webapp:
https://github.com/searxng/searxng/blob/9fce459c7f6856dcde089eaee0f0c6fbd7cd9b53/searx/engines/google.py#L428

The code can use httpx directly in some places, but not everywhere:
https://github.com/searxng/searxng/blob/9fce459c7f6856dcde089eaee0f0c6fbd7cd9b53/searxng_extra/update/update_currencies.py#L98
https://github.com/searxng/searxng/blob/9fce459c7f6856dcde089eaee0f0c6fbd7cd9b53/searx/engines/wikidata.py#L153-L159

Here  `get` and `post` from `searx.network` need a Network (either the dedicate one in the webapp, either a default one in the standalone scripts).

Instead of using httpx directly, the standalone scripts can initialize `searx.network` with a dedicate initialization (a parameter in the constructor of NetworkManager, or whatever), so this line can be removed (if you see it as an issue):

https://github.com/searxng/searxng/blob/9fce459c7f6856dcde089eaee0f0c6fbd7cd9b53/searx/network/network.py#L427

What do you think?

---

`check_network_configuration(...)` : yes the call should be done from `initialize`.

`NetworkManager.get(name=None)` : who creates and own the `NetworkManager` singleton?

return42:
>  so this line can be removed (if you see it as an issue):

the line is not a big deal .. I just want suggest to have a manager for the task "networks management" in SXNG's network stack.

dalf:
Why not a network management, but I can't answer who creates the NetworkManager, and then how to access the singleton.

return42:
> Why not a network management

was what I suggested / in code-speak ..

```python

class NetworkManager:

  def __init__(self):
      self.networks = {}

  def new_network(self, name):
      ...
  def check_network_configuration(self):
      ...
  def get_network(self, name = None):
      ret_val = self.networks.get(name or DEFAULT_NAME)
      if ret_val:
          return ret_val
      self.networks[name] = self.new_network(name)
      return self.networks[name]

NETWORKS = NetworkManager()
```

dalf:
I've added a NetworkManager as you described, a lot of thing can be improved, but that's a start.

dalf:
the last commit removes the async code.
Perhaps it's worth trying to enable HTTP/2 with the image proxy.

searx/network/context.py
```diff
@@ -0,0 +1,311 @@
...
+    ABCHTTPClient,
+    SoftRetryHTTPException,
+)
+
+P = ParamSpec('P')
+R = TypeVar('R')
+HTTPCLIENTFACTORY = Callable[[], ABCHTTPClient]
+
+
+THREADLOCAL = threading.local()
```
dalf: Since each NetworkContext is per user query and per engine, THREADLOCAL can be removed (?)

dalf: Done

searx/network/__init__.py
```diff
@@ -1,266 +1,137 @@
...
-    @staticmethod
-    def delete(url, **kwargs):
-        return Request('DELETE', url, kwargs)
-
-
-def get(url, **kwargs):
+def request(method, url, **kwargs) -> httpx.Response:
+    """the same as httpx.request"""
+    network_context: NetworkContext = getattr(_THREADLOCAL, _NETWORK_CONTEXT_KEY, None)
+    http_client = network_context.get_http_client() if network_context else NETWORKS.get().get_http_client()
```
dalf: Raise an NetworkContextNotFound when there is no context

searx/network/network.py
```diff
@@ -2,426 +2,446 @@
...
+        for client in self._clients.values():
+            client.close()
+
+    def get_context(self, timeout: float, start_time: Optional[int] = None) -> NetworkContext:
+        """Return a new NetworkContext"""
+        # TO DO : start_time should be optional, timeout however is mandatory
+        context_cls = self._settings.retry_strategy.value
+        return context_cls(self._settings.retries, self.get_http_client, start_time, timeout)
+
+    def get_http_client(self) -> ABCHTTPClient:
```
dalf: Make this method private: the only way to get a HTTP client must be through a NetworkContext (so the retry strategy is applied)

searx/webapp.py
```diff
@@ -1042,8 +1042,9 @@ def image_proxy():
             'Sec-GPC': '1',
             'DNT': '1',
         }
-        set_context_network_name('image_proxy')
-        resp, stream = http_stream(method='GET', url=url, headers=request_headers, allow_redirects=True)
+        network = NETWORKS.get('image_proxy')
+        http_client = network.get_http_client()
```
dalf: Get a context and then a HTTP client

searx/network/context.py
```diff
@@ -0,0 +1,321 @@
...
+class NetworkContext(ABC):
+    """Abstract implementation: the call must defined in concret classes.
+
+    Life time: one engine request or initialization of an engine
+    """
+
+    __slots__ = ('retries', 'http_client_factory', 'http_client', 'start_time', 'total_time', 'timeout')
+
+    def __init__(
+        self, retries: int, http_client_factory: HTTPCLIENTFACTORY, start_time: Optional[float], timeout: float
```
dalf: There are two timeout types:
* per request
* per context

The timeout per context should be optional.
It should already work because self.get_remaining_time() already deal with the None value

`timeout: float` --> `timeout: Optional[float]`

searxng_extra/update/update_engine_descriptions.py
```diff
@@ -206,7 +206,7 @@ def initialize():
 
 def fetch_wikidata_descriptions():
     print('Fetching wikidata descriptions')
-    searx.network.set_timeout_for_thread(60)
+    searx.network.set_networkcontext_for_thread(None, 60.0)
```
dalf: There is no need to specify a network context time out:

`searx.network.set_networkcontext_for_thread(None, None)`

or

`searx.network.set_networkcontext_for_thread()`

?

searx/network/__init__.py
```diff
@@ -1,266 +1,137 @@
...
+    "options",
+    "head",
+    "post",
+    "put",
+    "patch",
+    "delete",
+]
+
+
+_THREADLOCAL = threading.local()
```
dalf, line 84: Make sure the network contexts are deleted when the threads end

searx/network/client.py
```diff
@@ -1,35 +1,58 @@
...
+# pylint: disable=global-statement
+# pyright: basic
+"""Implement various ABCHTTPClient
+
+* OneHTTPClient        wrapper around httpx.Client
+* HTTPClient           httpx.Client accept the verify and max_redirects parameter only in the constructor.
+                       HTTPClient allows to pass these parameter in each query by creating multiple OneHTTPClient.
+* HTTPClientSoftError  Inherit from HTTPClient, raise an error according to retry_on_http_error parameter.
+* TorHTTPClient        Inherit from HTTPClientSoftError, check Tor connectivity
+"""
```
dalf, line 10: Find better names for the classes: it is not clear that `HTTPClientSoftError` is class that should not be use outside this module.

searx/network/client.py
```diff
@@ -78,33 +111,29 @@ def __init__(self, *args, **kwargs):
...
-    ) -> None:
+    def __exit__(self, exc_type, exc_value, traceback) -> None:  # pylint: disable=signature-differs
+        # avoid to import the various type for the signature, but pylint is not happy
         pass
 
 
-class AsyncProxyTransportFixed(AsyncProxyTransport):
-    """Fix httpx_socks.AsyncProxyTransport
+class _ProxyTransportFixed(SyncProxyTransport):
+    """Fix httpx_socks.SyncProxyTransport
```
dalf: It does not fix SyncProxyTransport, it wraps the python_socks exceptions into httpx exceptions, so other modules deals only with httpx exceptions.

* https://github.com/romis2012/python-socks/blob/master/python_socks/_errors.py
* https://www.python-httpx.org/exceptions/

searx/network/__init__.py
```diff
@@ -1,266 +1,137 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
-# pylint: disable=missing-module-docstring, global-statement
-
-import asyncio
+# pylint: disable=global-statement
+"""Provide HTTP clients following the user configuration
+"""
 import threading
```
dalf, line 55: ~~Describe the API: the implementation uses httpx, but the API is mostly compatible with requests.~~

~~**mostly** because, may be there are some missing parts.~~
~~TODO : check https://www.python-httpx.org/compatibility/~~

See the cookie section --> should we drop the requests API emulation and keep only httpx?

searx/search/checker/impl.py
```diff
@@ -71,8 +72,8 @@ def _download_and_check_if_image(image_url: str) -> bool:
         a = time()
         try:
             # use "image_proxy" (avoid HTTP/2)
-            network.set_context_network_name('image_proxy')
-            r, stream = network.stream(
+            http_client = NETWORKS.get('image_proxy').get_http_client()
```
dalf: Create a NetworkContext and then get a HTTP client

searx/network/__init__.py
```diff
@@ -1,266 +1,137 @@
...
+    """A NetworkContext is expected to be set in this thread.
+
+    Use searx.network.set_context_for_thread or searx.network.context_for_thread
+    to set a NetworkContext
+    """
 
 
-THREADLOCAL = threading.local()
-"""Thread-local data is data for thread specific values."""
+def get_http_runtime() -> Optional[int]:
```
dalf: Rename to `get_http_runtime_for_thread` to keep coherency with other method names.

 Move this method after `networkcontext_call`

tests/unit/network/test_network.py
```diff
@@ -78,120 +99,113 @@ def test_get_kwargs_clients(self):
...
+        with patch.object(httpx.Client, 'request', return_value=response):
+            network = Network.from_dict(enable_http=True)
+            http_client = network.get_http_client()
+            response = http_client.request('GET', 'https://example.com/')
             self.assertEqual(response.text, a_text)
-            await network.aclose()
+            network.close()
 
 
 class TestNetworkRequestRetries(SearxTestCase):
```
dalf, line 148: Run this class with the different retry strategies

dalf (commented):
Notes for later

searx/network/network.py
```diff
@@ -2,426 +2,446 @@
...
+                local_addresses=local_addresses,
+                retry_on_http_error=self._settings.retry_on_http_error,
+                hook_log_response=hook_log_response,
+                log_trace=log_trace,
+                logger=self._logger,
+            )
+        return self._clients[key]
 
-    def get_ipaddress_cycle(self):
+    def _get_ipaddress_cycle(self):
```
dalf: rename `_get_local_addresses_cycle`

searx/network/network.py
```diff
@@ -2,426 +2,446 @@
...
-                pattern = PROXY_PATTERN_MAPPING.get(pattern, pattern)
-                if isinstance(proxy_url, str):
-                    proxy_url = [proxy_url]
-                yield pattern, proxy_url
-
-    def get_proxy_cycles(self):
-        proxy_settings = {}
-        for pattern, proxy_urls in self.iter_proxies():
-            proxy_settings[pattern] = cycle(proxy_urls)
+    def _get_proxy_cycles(self):
```
dalf, line 176: Describe the return value because it is not straightforward

dalf: Done

searx/network/__init__.py
dalf: This is the case with some delay which `del` does not improve.

Bnyro:
(typo in third commit message: refactaring -> refactoring)

dalf:
> (typo in third commit message: refactaring -> refactoring)

Thank you.
I have difficulties to make meaningful commits, currently there are more backups.
In the end I see one big commit, please tell me if you see a way to split this massive change.

searx/network/context.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def _get_new_client_from_factory(self):
+        return self._http_client_factory()
+
+    @contextmanager
+    def _record_http_time(self):
+        """This decorator records the code's runtime and adds it to self.total_time"""
+        time_before_request = default_timer()
+        try:
+            yield
```
Bnyro, line 169: Could you please explain why that is needed/what it does? I don't really get that.

dalf: This follow the pattern from documentation: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

The following code 
```python
with _record_http_time():
  do_something()
```

is equivalent to:

```python
time_before_request = default_timer()
try:
  do_something()
finally:
  self._http_time += default_timer() - time_before_request
```

so the elapsed time is measured even with an exception.

searx/network/__init__.py
```diff
@@ -1,266 +1,483 @@
...
+    )
+
+
+def options(
+    url: str,
+    params: Optional[httpx._types.QueryParamTypes] = None,
+    headers: Optional[httpx._types.HeaderTypes] = None,
+    cookies: Optional[httpx._types.CookieTypes] = None,
+    auth: Optional[httpx._types.AuthTypes] = None,
+    allow_redirects: bool = False,
```
Bnyro, line 294: This has been set to `True` by default previously (not sure why though).

searx/network/client.py
```diff
@@ -155,56 +192,378 @@ def get_transport(verify, http2, local_address, proxy_url, limit, retries):
...
+
+    def _is_error_but_retry(self, response):
+        # pylint: disable=too-many-boolean-expressions
+        if (
+            (self.retry_on_http_error is True and 400 <= response.status_code <= 599)
+            or (isinstance(self.retry_on_http_error, list) and response.status_code in self.retry_on_http_error)
+            or (isinstance(self.retry_on_http_error, int) and response.status_code == self.retry_on_http_error)
+        ):
+            return True
+        return False
```
Bnyro:
```suggestion
        return (
            (self.retry_on_http_error is True and 400 <= response.status_code <= 599)
            or (isinstance(self.retry_on_http_error, list) and response.status_code in self.retry_on_http_error)
            or (isinstance(self.retry_on_http_error, int) and response.status_code == self.retry_on_http_error)
        )
```

searx/network/__init__.py
dalf: The default value is different between httpx and requests.
* `True` is the [default value for requests](https://github.com/psf/requests/blob/ee93fac6b2f715151f1aa9a1a06ddba9f7dcc59a/src/requests/api.py#L36).
* `False` is the [default value for httpx](https://github.com/encode/httpx/blob/3ba5fe0d7ac70222590e759c31442b1cab263791/httpx/_api.py#L213
)

Since we use httpx, it is easier to follow httpx default values.
SearXNG uses only / mainly (?) two HTTP methods: GET and POST, so there is no impact in changing the default value for OPTIONS.

Note: the parameter name is `allow_redirects` and it's a legacy from requests. httpx uses `follow_redirects`. I didn't had the courage to rename the parameter in this PR.

searx/network/context.py
```diff
@@ -0,0 +1,387 @@
...
+class NetworkContextRetryFunction(NetworkContext):
+    """When an HTTP request fails, this NetworkContext tries again
+    the whole function with another HTTP client.
+
+    This guarantees that func has the same HTTP client all along.
+    """
+
+    def call(self, func: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
+        try:
+            while self._retries >= 0 and self.get_remaining_time() > 0:
```
Bnyro, line 255: Even if retries is set to `0`, we still start one additional attempt. Is that intended?

searx/network/context.py
```diff
@@ -0,0 +1,387 @@
...
+                self._set_http_client()
+                try:
+                    return func(*args, **kwargs)  # type: ignore
+                except (ssl.SSLError, httpx.RequestError, httpx.HTTPStatusError, SoftRetryHTTPException) as e:
+                    if self._retries <= 0:
```
Bnyro, line 260: I don't think self._retries can be smaller than 0, as we already checked it's `>=0` above? Or is it possible that `_retries` is being changed from an other function that's running at the same time somewhere else in the code?

searx/network/context.py
dalf: I don't think so.

Note: some tests are missing of the retry count.

searx/network/context.py
```diff
@@ -0,0 +1,387 @@
...
+            try:
+                return self.http_client.send(stream, method, url, **kwargs)
+            except SoftRetryHTTPException as e:
+                if retries <= 0:
+                    return e.response
+            except (ssl.SSLError, httpx.RequestError, httpx.HTTPStatusError) as e:
+                if retries <= 0:
+                    raise e
+            retries -= 1
+        return self.http_client.send(stream, method, url, **kwargs)
```
Bnyro: These lines are used about three times with very small modifications only for the `Retry...HTTPClient`'s, I wonder if we can somehow clean that up using a function or make the `Retry...HTTPClient`'s inherit from each other?

searx/network/__init__.py
Bnyro: I see, that makes sense. Thanks for the explanation!

Not really sure why we had even set a different default for the `OPTIONS` request although we don't even use it ...

searx/network/__init__.py
```diff
@@ -1,266 +1,483 @@
...
+* The context manager `networkcontext_for_thread`, for the generic use case.
+
+Inside the thread, the caller can use `searx.network.get`, `searx.network.post` and similar functions without
+caring about the HTTP client. However, if the caller creates a new thread, it must initialize a new NetworkContext.
+A NetworkContext is most probably thread-safe, but this has not been tested.
+
+The overall architecture:
+* searx.network.network.NETWORKS contains all the networks.
+    The method `NetworkManager.get(network_name)` returns an initialized Network.
+* searx.network.network.Network define one network (a set of proxies, local IP address, etc...).
```
Bnyro:
```suggestion
* searx.network.network.Network defines a network (a set of proxies, local IP address, etc...).
```

searx/network/__init__.py
```diff
@@ -1,266 +1,483 @@
...
+    They are defined in settings.yml.
+    The method `Network.get_context()` creates a new NetworkContext.
+* searx.network.context contains three different implementations of NetworkContext. One for each retry policy.
+* searx.network.client.HTTPClient and searx.network.client.TorHTTPClient implements wrappers around httpx.Client
```
Bnyro:
```suggestion
* searx.network.client.HTTPClient and searx.network.client.TorHTTPClient implement wrappers around httpx.Client.
```

searx/network/context.py
dalf: I know, the issue is "small modifications" and readability.

My take, debatable:
* nearly copy / paste are easy to read and understand. It requires attention when the code is changed, but at least all cases are clear and straightforward.
* inheritance, if not carefully crafted, is going to make the code harder to read and modify.

The NetworkContext + the ABCHTTPClient wrappers are already difficult to follow.

Bnyro:
This certainly already has been a ton of work and did cost a lot of time, so I don't want to cause too much additional efforts! Left some comments about things that came into my mind above, but there's certainly not much to complain about here :+1:

Works great from testing it a bit, so there are no bugs/issues that would be blocking a merge, just some small code improvement suggestions/questions.

searx/network/context.py
dalf: It is defensive programming: later, if for some reason `_retries` falls below zero, it creates an (nearly) infinite loop.
As you said, currently, the value can't go below zero.

> ...somewhere else in the code?

In the previous version of the code, it was the case.

searx/network/context.py
dalf: ~~Fixed~~

dalf: Actually: `retries` defines the number of ... retries, but at first the code must try.
Said differently, if for example `retries == 1`, the number of iteration must be `2`.
==> `self._retries >= 0` does not add an additional attempt.

searx/network/__init__.py
```diff
@@ -1,266 +1,483 @@
...
+
+        return func_inner
+
+    return func_outer
+
+
+def request(
+    method: str,
+    url: str,
+    params: Optional[httpx._types.QueryParamTypes] = None,
```
return42, line 187: FYI (no need to change) with `from __future__ import annotations` you can also write:

```python
    params: httpx._types.QueryParamTypes | None = None
```

searx/network/client.py
```diff
@@ -220,18 +577,5 @@ def init():
...
-
-    thread = threading.Thread(
-        target=loop_thread,
-        name='asyncio_loop',
-        daemon=True,
-    )
-    thread.start()
-
 
 init()
```
return42: Isn't it better the init (set log-levels) is done by the application? .. set log-level on import of a module can confuse, when other modules do the same (I know that is not the case in SearXNG, but AFIAK know its a anti-pattern to set log-levels or do other side-effects that are not related directly to the module while importing the module).

dalf: Yes, I removed the uvloop part and left over this part. 
The place I see is `searx/__init__.py` because this is where the logging initialization is done, what do you think ?

return42: > ..  searx/__init__.py because this is where the logging initialization is done

Sadly that is true .. IMO its the same issue: *"set log-levels or do other side-effects that are not related directly to the module while importing the module"*

https://github.com/searxng/searxng/blob/24561370b7702846124ab083701792deee310c5b/searx/__init__.py#L91-L96

> what do you think?

I think all initialization should be done when the application context is build up.  We do not have an *application factory*, we build the application while import `searx.webapp`.

To limit the damage, we should perform initializations only when importing the `searx.webapp` (or where needed, during the tests).

https://github.com/searxng/searxng/blob/24561370b7702846124ab083701792deee310c5b/searx/webapp.py#L1318-L1325

searx/network/client.py
```diff
@@ -155,56 +192,376 @@ def get_transport(verify, http2, local_address, proxy_url, limit, retries):
...
+        for pattern, proxy_url in self.proxies.items():
+            if not self.enable_http and pattern.startswith('http://'):
+                continue
+            if (
+                proxy_url.startswith('socks4://')
+                or proxy_url.startswith('socks5://')
+                or proxy_url.startswith('socks5h://')
+            ):
+                mounts[pattern] = _get_transport_for_socks_proxy(
+                    self.verify, self.enable_http2, self.local_address, proxy_url, limit, 0
```
dalf, line 353: Comment: `0` is the number of retries on purpose set to this value. The `NetworkContext` classes manage the retries (and may use different HTTPClient)

searx/search/processors/online.py
```diff
@@ -9,6 +9,8 @@
 from timeit import default_timer
 import asyncio
 import ssl
+from typing import Dict, List
+
 import httpx
 
 import searx.network
```
return42, line 14: @dalf In context of https://github.com/searxng/searxng/issues/762 I have the question: do we need this `import searx.network` any longer or can we now import by:

```python
from searx import network
```

and by example in line 110 ff

```python
        if params['method'] == 'GET':
            req = network.get
        else:
            req = network.post
```

No need to change, just a question for my knowledge.

---

BTW if so (what I assume), we can close  https://github.com/searxng/searxng/issues/762 when we merge this PR

dalf: This PR does not fix #762 because there is still a conflict between the global variable name `network` defined by `settings.yml` and the `network` module.

The issue exists only the engine modules and for `online.py`. These lines are not an issue because `searx.network` is going to be override:
https://github.com/searxng/searxng/blob/1878f67e642dc6a640c0b3a72fe5c6e060bcbc81/searx/search/processors/online.py#L16
https://github.com/searxng/searxng/blob/1878f67e642dc6a640c0b3a72fe5c6e060bcbc81/searx/search/processors/online.py#L107-L110

---

IMO, the issue #762 should be solved in another way, but that's topic of another discussion.

searx/network/client.py
dalf: > Sadly that is true ..

Do you have time to add the change in your branch?

return42: > Do you have time to add the change in your branch?

I will give it a try .. but will take a while / first I want to read, format and .. **understand** the doc-strings :)

return42:
Just a reminder for myself:

I could solve the issue when I removed all the http client & SSL caching (5eb27b9e4) what exactly the cause is requires more precise clarification.

- patch: [WIP experimantal fix of a bing issue](https://github.com/return42/searxng/commit/5eb27b9e4d3c8c209a71fd41f99d545288846402)

To test in production I merged this PR #2685 and the "WIP"-patch on my instance (https://darmarit.org/searx/) --> [darmarit.org branch](https://github.com/return42/searxng/commits/darmarit.org)

----

EDIT / "WIP-patch" worked a few days ago, but since last merge (today) it does not work any longer? .. needs more investigation!

dalf:
The image proxy is fixed: the streamed response are now closed.

Explanation: the code uses `Flask.Response.call_on_close` to close the response from httpx.
`Flask.Response` was initialized `direct_passthrough=True` to provide some optimization however in this scenario, `call_on_close` has not effect: the httpx response were not closed. That's explain why the number of opened socket were increasing.

ping @mrpaulblack

mrpaulblack:
Thanks for the info. I deployed ffcff586939a548c6cd9b279e2d0badea55a6656 on my instance and will keep you posted if the issue resolved in staging :+1:

return42:
FYI: I rebased my [searx_network_rework](https://github.com/return42/searxng/commits/searx_network_rework) branch and continue my review (sorry for the delay) .. I am currently revising the documentation .. it will still take a while until I have reached a state that I can present here.

return42:
@dalf FYI: I applied 0c1d188f on my server .. but without luck / see my https://github.com/searxng/searxng/issues/2977#issuecomment-1794206939

I think it's better to put the Bing problem on the back burner for now, we should concentrate on this PR (as it is now / without the 0c1d188f commit).

Anyway, I think it was worth to give commit 0c1d188f a try ... but for now let's get back to the network refactoring.

return42:
@dalf is it OK when we remove [0c1d188](https://github.com/searxng/searxng/pull/2685/commits/0c1d188fb05b6fd936d73cd385d15debd0a50ed8) from this PR? (I think we need this commit no longer here and with this commit I get into trouble when I rebase my branch where I do my review of this PR).

BTW there is an issue in 0c1d188f:

```
File "/usr/local/searxng/searxng-src/searx/engines/bing.py", line 83, in get_network_extra_context
     engine_region = traits.get_region(params['searxng_locale'], traits.all_locale)  # type: ignore
NameError: name 'traits' is not defined
```

dalf:
Yes this commit can be dropped

return42:
> Yes this commit can be dropped

FYI: I removed commit 0c1d188 from your branch.

searx/network/client.py
```diff
@@ -155,56 +192,377 @@ def get_transport(verify, http2, local_address, proxy_url, limit, retries):
...
+        # pylint: disable=too-many-arguments
+        self,
+        verify=True,
+        enable_http=True,
+        enable_http2=False,
+        max_connections=None,
+        max_keepalive_connections=None,
+        keepalive_expiry=None,
+        proxies=None,
+        local_addresses=None,
```
return42, line 253: @dalf : Its not entirely clear to me, does a `HTTPClient` really have more than on `local_addresses`? .. I know the `Network` can have more than on local address .. but the `HTTPCLient`?

dalf: Ah good point the parameter name should be `local_address` not `local_addresses` :
* `HTTPClient` have only one `local_addresses`
* `Network` have a mapping between [`(local_address, proxies)`](https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/network.py#L139-L140) and an `HTTPClient`.

return42: > * `Network` have a mapping between [`(local_address, proxies)`](https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/network.py#L139-L140) and an `HTTPClient`.

yes, this was the point where I struggled :) .. I will fix the naming in my reviewhttps://github.com/return42/searxng/commits/searx_network_rework

searx/network/client.py
```diff
@@ -155,56 +192,377 @@ def get_transport(verify, http2, local_address, proxy_url, limit, retries):
...
+
+    The class checks if the client is really connected through Tor.
+    """
+
+    _TOR_CHECK_RESULT = {}
+
+    def __init__(self, proxies=None, local_addresses=None, **kwargs):
+        self.proxies = proxies
+        self.local_addresses = local_addresses
+        super().__init__(proxies=proxies, local_addresses=local_addresses, **kwargs)
```
return42, line 515: @dalf I'm wrong, or can we drop this TOR constructor and simply use the constructor of the parent class? If so, I remove it in my review ..

dalf: Actually yes:
* drop `__init__`
* replace https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/client.py#L516 by
`if not self._is_connected_through_tor(self.default_kwargs.get("proxies"), self.default_kwargs.get("local_addresses")):`

it should work, but I have not tested.

With this change, a proxy is not mandatory: as long the connection is through Tor (with an external router for example), it is accepted.

searx/network/network.py
```diff
@@ -1,427 +1,505 @@
...
-        if not local_addresses:
-            return
-        if isinstance(local_addresses, str):
-            local_addresses = [local_addresses]
-        for address in local_addresses:
-            yield address
+    def get_context(self, timeout: Optional[float] = None, start_time: Optional[float] = None) -> NetworkContext:
+        """Return a new NetworkContext"""
+        context_cls = self._settings.retry_strategy.value
+        return context_cls(self._settings.retries, self._get_http_client, start_time, timeout)
```
return42, line 124: @dalf: Here we pass the arguments `self._settings.retries, self._get_http_client,` to the `RetryStrategy` (aka `NetworkContext`).  Is there any special reason why we not pass the `Network` object (here it is `self`) to the `NetworkContext` constructor.

When we pass the `Network` object as one argument, the function prototypes becomes less verbose and it scales better in future, when we might want to add more features to the context.

IMO it might be good, we pass the network object to Context & Clients .. so we can implement features in Context and Clients without the need to touch code like this one here to pass additional data from network to (e.g.) context.

Or do I oversee something?

dalf: The purpose is to avoid circular import:
* searx.network.network imports searx.network.context
* if searx.network.context imports searx.network.network is going to be messy.

Also, it is easier to write unit tests.

return42: Since the objects are already here, I guess you mean we have a circular import when we add type annotations?

dalf: > Since the objects are already here, I guess you mean we have a circular import when we add type annotations?

Yes (that's the usage I can think of).

Note: it is also easier to write unit test.

return42: > Yes (that's the usage I can think of).

Haven't tested, but I think we can use [forward references](https://peps.python.org/pep-0484/#forward-references) .. where types are noted as strings:

```python

def foo(network: 'searx.network.network.Network'):
    pass
```
However, type-hints should not prevent us from writing the code we need.

> Note: it is also easier to write unit test.

I will have a look .. again: However, unit tests should not prevent us from writing the code we need :)

Thanks for your feedback, at least now I know that there are no hard reasons not to pass the network ... I think this simplifies the code a lot and scales better for future requirements.

dalf: > I think this simplifies the code a lot

how?

The two values are copied:
https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L81-L83

* `self._retries` is not a constant, so this attribute can't be skipped. See for example:
https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L260
* `http_client_factory` is wrapped. May be you have other implementation in mind to manage timeout?

The only thing is when we need additional parameter from the Network: I see that as a way to review the coupling between the two classes ; instead of calling any fields and methods from a Network instance and end up to entangle of two classes.

return42: I have a use case in mind ([the bing issue](https://github.com/return42/searxng/commit/ceba946e2547dabe56335a64cbd6229d731aeab9) / don't cache client, don't reuse SSL context in a bing-network), where I see its hard to pass an setting from the Network to the NetworkContext and from there down to the HTTPClient .. all these objects are working in one context, that is dominated by the Network object ..

Give me a try .. may I come to the conclusion its a bad idea, but for now I think all objects should have access the the central Network (.. to read settings from the config and decide what to do).

Austin-Olacsi:
Maybe we could add something for:
https://github.com/searxng/searxng/pull/2772


To set the correct header? I'm just not sure what the best approach is.

return42:
> Maybe we could add something for: https://github.com/searxng/searxng/pull/2772

@Hackurei I think setting HTTP headers for image load of special engines is not in the scope of   the network stack / BTW it would not help when there is no image proxy and the image needs to be load by the web client itself .. 

TL;DR: #2772 is very special to https://www.pixiv.net/ and its WEB-Clients/WEB-Side communication and cannot be solved sustainably at the level of the network stack of this PR.

searx/network/context.py
```diff
@@ -0,0 +1,404 @@
...
+        self._set_http_client()
+        return self.call(self._get_http_client().request, *args, **kwargs)
+
+    @final
+    def stream(self, *args, **kwargs):
+        """Convenient method to wrap a call to stream inside the call method.
+
+        Use a new HTTP client to wrap a call to the stream method using self.call
+        """
+        self._set_http_client()
```
return42: @dalf  Why is a new client set here? / Actually, the retry strategy (the `self.call`) is already responsible for setting a new client. By example in line 250 a `self._set_http_client()` is set by the `call` method.

```python

class NetworkContextRetryFunction(NetworkContext):
    ...
    def call(self, func: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        try:
            while self._retries >= 0 and self.get_remaining_time() > 0:
                self._set_http_client()

```

searx/network/context.py
```diff
@@ -0,0 +1,404 @@
...
+        return self._http_time
+
+    @final
+    def get_remaining_time(self, _override_timeout: Optional[float] = None) -> float:
+        """Return the remaining time for the context.
+
+        _override_timeout is not intended to be used outside this module.
+        """
+        timeout = _override_timeout or self._timeout or DEFAULT_TIMEOUT
+        timeout += 0.2  # overhead
```
return42, line 137: Do we have a use case / a need for `_override_timeout`? (I haven't found a usage of this argument).  May its better to have an argument `offset=0.2`?

searx/network/context.py
```diff
@@ -0,0 +1,404 @@
...
+        so the NetworkContext implementations can wrap the ABCHTTPClient.
+        """
+        self._http_client = self._get_new_client_from_factory()
+
+    @final
+    def _reset_http_client(self):
+        self._http_client = None
+
+    def _get_new_client_from_factory(self):
+        return self._http_client_factory()
```
return42, line 162: Nit picking: do we really need such a lot of indirections .. where private methods get/set private members?  By example to set a new http client in `self._http_client` 

- we call `self._set_http_client()` which does nothing more than 
- calling a method `self._get_new_client_from_factory()` which does nothing more than
- calling `self._http_client_factory()`

BTW: the reader needs to know that a *private* method / a method with an underline prefix (`_get_new_client_from_factory`) is overwritten in the inheritance .. this needs really a lot of code reading to understand whats going on

I mean (haven't tested), this could be simplified to a property (which can be overwritten in the inheritance:

```python
class NetworkContext:

    _http_client = None

    @property
    def http_client(self):
        if self._http_client is None:
            self._http_client = self._http_client_factory()
        return self._http_client

    @final
    def reset_http_client(self):
        self._http_client = None
```

searx/network/context.py
dalf: https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L228-L231

called by:

https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L211-L219

In English, this allows the user to override the timeout per HTTP request:
```python
with searx.network.networkcontext_manager(timeout=2):
  searx.network.get("https://example.com", timeout=1) # The user asks for a timeout of 1, not 2.
```

searx/network/context.py
dalf: I know: an explanation or a different implementation is needed here.
You will see it crashes if you remove this line, I don't remember why right now. I will look at it later this week / this day.

searx/network/context.py
dalf: > do we really need such a lot of indirections .. where private methods get/set private members?

I don't remember exactly the last audio call, may be I said it was private methods ?
I should have said protected methods rather than private methods.

> the reader needs to know that a private method / a method with an underline prefix 

it seems to be a common practice:
* a single `_` as prefix for protected methods
* a double `__` as prefix for private methods

See:
* https://softwareengineering.stackexchange.com/questions/228216/what-is-the-historical-reason-why-python-uses-the-double-underscore-for-class-pr
* https://discuss.python.org/t/private-protected-modifier-and-notation/22356
* https://stackoverflow.com/questions/6930144/underscore-vs-double-underscore-with-variables-and-methods
* httpx uses that convention
* pylint message : http://pylint-messages.wikidot.com/messages:w0212
* also: https://github.com/pylint-dev/pylint/issues/1802#issuecomment-390159460

---

About indirection: I can't say I'm happy with them, however `_get_new_client_from_factory` is override by the different implementations:
https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L267-L268

https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L315-L316

https://github.com/searxng/searxng/blob/ffcff586939a548c6cd9b279e2d0badea55a6656/searx/network/context.py#L369-L370

Reading the code now, I think:
* the different implementations of `NetworkContext` can set the `self._http_client` directly in `_set_http_client` (so this method is no more final).
* `_get_new_client_from_factory` can be removed.

For example:

```python
class NetworkContextRetryFunction(NetworkContext):
    def call(self, func: Callable[P, R], *args: P.args, **kwargs: P.kwargs) -> R:
        ...

    def _set_http_clientself):
        self._http_client = _RetryFunctionHTTPClient(super()._get_new_client_from_factory(), self)
```

dalf:
I will review your additional commits most probably on Friday ( 5871c78 and d82d06e ).

return42:
> I will review your additional commits most probably on Friday ( [5871c78](https://github.com/searxng/searxng/commit/5871c78ff3c01e989a22caa8e8246a4a6161a33d) and [d82d06e](https://github.com/searxng/searxng/commit/d82d06eeb2d0fcc98a8fe29756b185ce914c1ed5) ).

~No, please not! .. its just half of the code .. and may I will drop it later .. give me some more time to finish my review ..~

searx/network/context.py
return42: > I will look at it later this week / this day.

No need to hurry, please let me first finish my review .. in this phase I will post some of my question here on gh .. if needed, we can comeback later to this topics/quesions.

searx/network/context.py
return42: > I don't remember exactly the last audio call, may be I said it was private methods ?
> I should have said protected methods rather than private methods.

No you didn't said its private, its just me who mixes both terms .. in python the concept of `__private` is just crap and should not been used (as David Goodger recommends / `_Foo__private`).  As a replacement, it has commonly been agreed that it is better to use `_protected`.

.. what I wanted to point out is: I think its not good to mark names as `_protected` and then overwriting these names in inheritance .. we can do it, but personally I think its not a good practice.

> the different implementations of NetworkContext can set the `self._http_client directly` in  ..

Furthermore, I think that `.getter()` and `.setter()` can often be realized by a `@property` (as shown above) / which is late-evaluated as a positive side effect (no need to explicitly call a `.setter()`).

But you should leave everything as it is for now, I will make suggestions in my branch, which we can look at after I have finished my review

searx/network/context.py
dalf: Fixed in ed35731 
The call to `self._get_http_client()` has to be done inside `self.call`, not before.

dalf:
The tests are flaky, this is related to https://github.com/searxng/searxng/issues/2988 not to this PR

The socket leak might be fixed (at least I've tried).

return42:
@dalf @mrpaulblack sorry, I lost the overview, what the state of this PR, the network stack?

dalf:
The broken part is the streaming for the image proxy: the sockets are not always closed.
Everything else works.
I have an idea how to fix that, just need motivation to implement it...

czaky:
Sorry, for intruding on this almost 1y discussion.

Context: I have recently experimented with SearXNG through unreliable proxy networks (particularly TOR)
(reference setup: https://github.com/czaky/turbotor). This resulted in PR: #3491 .

How could we support following features (using a non-asyncio approach)?

- Parallel fan out of requests through multiple proxies (at least in initial phase)?
- Keep alive / connectivity- / latency-determination for existing HTTP connections (in the background)?

### Parallel requests through multiple proxies.

As implemented in #3491, doing the "retries" in parallel (instead of in series) makes this approach usable over networks with increased latency (tail >10s) and high rate of client rejection. This was rather straightforward to implement using asyncio and it minimized issues related to using SearXNG over TOR.

(Motivation: theoretically and practically, waiting for multiple engines increases the probability of hitting the tail latency exponentially. So if the common recommendation for TOR is a timeout of 30s, it is, unfortunately, how long SearXNG takes to answer queries.)

### Keep alive / connectivity- / latency-determination and background connection pools.

In case of TOR each proxy changes the circuits every 10 minutes. Thus, the set of "good" proxies changes rather fast. A background task could proactively determine a smaller subset of proxies that provide the optimal connectivity for each engine.
This would reduce client rejection rates by the target-point and reduce the overall noise. 

(Note, I am more interested in making SearXNG usable over TOR then advocating for any approach.)

dalf:
Regardless of the decision about async vs. sync, this PR is ready with one exception: some sockets are leaking, likely due to response streaming.

Response streaming is required for only one feature: the image proxy.

Therefore, an alternative approach is to:

* Remove response streaming from the code
* Provide a much simpler alternative

But first, we should verify whether the socket leak is indeed caused by response streaming.

PR #2686, opened 2023-08-27
[mod] bing: resolve redirect without additional requests

dalf:
## What does this PR do?

Remove the usage of searx.network.multi_requests
The results from Bing contains the target URL encoded in base64 See the u parameter, remove the first two character "a1", and done.

Also, ~~previously, if the number_of_results was lower than the actual result, all the results were ignored. This commit ignore number_of_results but still provides the results.~~

**[EDIT]** related to https://github.com/searx/searx/pull/1387 (on searx) --> I've added a comment about that

Remove the usage of `searx.network.multi_requests`. Bing is the only working engine to use this function.
From that, we might switch to sync code in `searx.network`.

Also, this PR make bing a little bit faster ; anyway, bing is already on of the fastest engine.

make requests on the bing engine.

## Author's checklist

It seems https://bing.com does not work at all without javascript (?) : If I type a query in the front page, then press enter, nothing happen.

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -29,10 +29,11 @@ inaccuracies there too):
 # pylint: disable=too-many-branches, invalid-name
 
 from typing import TYPE_CHECKING
+import base64
 import datetime
 import re
 import uuid
-from urllib.parse import urlencode
+from urllib.parse import parse_qs, urlencode, urlparse
 from lxml import html
 import babel
 import babel.languages
@@ -179,9 +180,7 @@ def request(query, params):
 
 
 def response(resp):
-    # pylint: disable=too-many-locals,import-outside-toplevel
-
-    from searx.network import Request, multi_requests  # see https://github.com/searxng/searxng/issues/762
+    # pylint: disable=too-many-locals
 
     results = []
     result_len = 0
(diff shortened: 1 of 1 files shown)


searx/engines/bing.py
```diff
@@ -208,38 +204,21 @@ def response(resp):
...
+            # get the first value of u parameter
+            url_query = urlparse(url).query
+            parsed_url_query = parse_qs(url_query)
+            param_u = parsed_url_query["u"][0]
+            # remove "a1" in front
+            encoded_url = param_u[2:]
+            # add padding
+            encoded_url = encoded_url + '=' * (-len(encoded_url) % 4)
+            # decode base64 encoded URL
+            url = base64.b64decode(encoded_url, validate=True).decode()
```
return42: With search term `:en !bing foo` I got an exception ..

```
  File "/share/searxng/searx/engines/bing.py", line 218, in response
    url = base64.b64decode(encoded_url, validate=True).decode()
  File "/usr/lib/python3.10/base64.py", line 86, in b64decode
    raise binascii.Error('Non-base64 digit found')
binascii.Error: Non-base64 digit found
```

dalf: Fixed at least for the query you mentioned

`base64.b64decode` --> `base64.urlsafe_b64decode`

return42 (approved):
> resolve redirect without additional requests 

Seems to work / thanks a lot :+1:  (it was  long time on my hidden ToDo list)

In my tests I realized, that bing engine needs some more review (not related to this PR):

- time range option
- safe search option
- page no

.. do no longer work / I suppose bing has changed its API in a line with which they promote this AI thing :o

PR #2721, opened 2023-09-07
[feat] simple theme: image preview - arrow keys to navigate

### LINKED ISSUE #2720: [FEATURE REQUEST] use the left and right arrow keys to switch between images when the image preview is open
i want to use the left and right arrow keys to switch between images when the image preview is open.

this is a habit of mine, something i have always done, and it works in almost all other search engines.

**Additional context**
i've tried to hunt down the source, and i found this file:
/searxng/searx/static/themes/simple/js/searxng.min.js

and in it, i found result-detail-previous and result-detail-next

i stopped here because i do not know javascript... and i only think i am looking at the right thing...

bonswouar:
## What does this PR do?
Use the left and right arrow keys to switch between images when the image preview is open

Better UX

Search for images, preview one, use arrow keys to navigate

## Author's checklist
I'm not sure this is the most efficient way, feedback is welcome!
I'd have preferred to bind it directly when opening the image preview in `results.js`, but then I'm not sure how to properly unbind them when closing it without unbinding other possible keydown events (vimKeys)

## Related issues
Closes #2720

### DIFF
--- a/searx/static/themes/simple/src/js/main/keyboard.js
+++ b/searx/static/themes/simple/src/js/main/keyboard.js
@@ -32,6 +32,10 @@ searxng.ready(function () {
     return resultElement && resultElement.classList.contains('result-images');
   }
 
+  function hasImageInDetail () {
+    return document.querySelector('#results.image-detail-open') !== null;
+  }
+
   searxng.on('.result', 'click', function (e) {
     if (!isElementInDetail(e.target)) {
       highlightResult(this)(true, true);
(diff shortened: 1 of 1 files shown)


BernieHuang2008 (commented):
I think you can merge your code into L162, this will make the code more readable and beautiful.

return42:
@bonswouar thanks a lot for your PR :+1: .. I'm not the JS expert but I think it is better, we register a even handler on the `.result` class :

```javascript
  searxng.on('.result', 'keydown', function (e) {
    let resultElement = getResultElement(e.target);
    if (isImageResult(resultElement)) {
      if (e.key === 'ArrowLeft') {
        e.preventDefault();
        searxng.selectPrevious();
      } else if (e.key === 'ArrowRight') {
        e.preventDefault();
        searxng.selectNext();
      } else if (e.key === 'Escape') {
        e.preventDefault();
        removeFocus(e);
      }
    }
  });
```

@dalf you wrote: [Currently, you can use `j` and `k` to navigate between the images. `Esc` close the detail,](https://github.com/searxng/searxng/issues/2720#issuecomment-1710315930)

.. AFAIK this is only when the vim-mode plugin is activated .. I think it is good to have a navigation by the arrow keys beside the vim-mode / what do you think about a arrow navigation and my code example from above?

BernieHuang2008:
@return42 improved the code on your advice, see #2724

PR #2723, opened 2023-09-07
[feature] dark theme for code highlighter in the result list 

### LINKED ISSUE #1354: Code highlighter is not suitable for the dark theme
The code highlighter should support dark & light theme of the UI:

grafik

---- 

remark from @dalf on Matrix: the result width / font size ratio is also an issue (IMO).

- First screenshot: Oscar.
- Second screenshot: Simple

grafik

grafik

return42:
## What does this PR do?

dark theme for code highlighter in the result list / BTW: the layout has been improved slightly

- https://pygments.org/styles/

For the light mode we use the "default" style and this PR add the "lightbulb" style for the dark mode.

`make run` and `!scc  hello world python`

image

## Related issues

Closes: https://github.com/searxng/searxng/issues/1354


searx/static/themes/simple/src/less/code.less
```diff
@@ -1,9 +1,51 @@
...
+
+      &::-moz-selection {
+        background: transparent; /* Gecko Browsers */
+      }
+      margin-right: 8px;
+      text-align: right;
+    }
+
+    span.linenos {
+      color: #64708d;
```
return42, line 32: Not sure if it is a good idea to custom the line.nummbers color here / but in the "default" style and in "lightbulb" style the color was to dark .. a value of `#64708d;` fits to both styles .. but when we decide to use other pygments style we have to modify or remove the value here.

searx/templates/simple/result_templates/code.html
```diff
@@ -2,12 +2,31 @@
 
 {{ result_header(result, favicons, image_proxify) -}}
 {{- result_sub_header(result) -}}
-{%- if result.content %}{{ result.content|safe }}{% endif %}</p>
+
+{%- if result.content -%}
+  <p class="content">
+    {{- result.content|safe -}}
+</p>
```
Bnyro:
```suggestion
  </p>
```

Bnyro:
That's cool, I didn't even know that we have support for searching code snippets :wink:

searx/static/themes/simple/src/less/code.less
Bnyro: I agree with you on that.
The line numbers are a bit too highlighted for my taste, it'd be cooler to use a gray color for it.

Nevertheless I don't think that's worth doing any extra patches only to change the line number color, for what its worth most people probably won't notice it anyways :)

searx/templates/simple/result_templates/code.html
return42: fixed / thanks :+1:

PR #2724, opened 2023-09-07
[feature] ArrowKeys to navigate between image results

### LINKED ISSUE #2720: [FEATURE REQUEST] use the left and right arrow keys to switch between images when the image preview is open

BernieHuang2008:
## What does this PR do?

Improved PR #2721.
## Author's checklist
I use Switch-case to make the code more readable and flexible, it's completely same as the example @return42 gived in #2721.

## Related issues
Closes #2720
Closes #2736


BernieHuang2008:
Local check passed, require check approvals.

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -43,6 +43,29 @@ searxng.ready(function () {
...
+  searxng.on('.result', 'keydown', function (e) {
+    var resultElement = getResultElement(e.target);
+    if (isImageResult(resultElement)) {
+      var matchKey = true;
+      switch (e.key){
+      case 'ArrowLeft':
+        searxng.selectPrevious();
+        break;
+      case 'ArrowRight':
+        searxng.selectPrevious();
```
bonswouar: Typo, should be `searxng.selectNext();` :)

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -43,6 +43,29 @@ searxng.ready(function () {
     }
   });
 
+  searxng.on('.result', 'keydown', function (e) {
```
bonswouar: Why not use `.image-detail-open` class directly then?
Plus with that the next 2 lines can be removed (no check for image result needed)

bonswouar (changes requested):

searx/static/themes/simple/src/js/main/keyboard.js
BernieHuang2008: thanks, solved.

searx/static/themes/simple/src/js/main/keyboard.js
BernieHuang2008: that all works same, because `.image-detail-open` contains `.result`.

BernieHuang2008: However, it seems has a new problem:

It's about the focus point: 
If user focus on the image (e.g., click the image), we won't catch the event when binding on `'.result'`
Otherwise, if we bind it on `.detail`(which is the pannel that display the image) but the user didn't do anything, then the focus is still on `.result`, we can't catch it on `.detail`.

BernieHuang2008: maybe we can bind it on `body` and use your `hasImageInDetail` to check whether a image result is opened.

```js
  function hasImageInDetail () {
    return document.querySelector('#results.image-detail-open') !== null;
  }
```
what do you think?

bonswouar: > that all works same, because .image-detail-open contains .result.

Yes but that should be cleaner, as the binding is done only when required (when an image is selected), not when any result is shown.

> It's about the focus point: If user focus on the image (e.g., click the image), we won't catch the event when binding on `'.result'` Otherwise, if we bind it on `.detail`(which is the pannel that display the image) but the user didn't do anything, then the focus is still on `.result`, we can't catch it on `.detail`.

Not sure what you mean exactly, but in my tests locally just binding on `.image-detail-open` seems to work properly.
But yes the user needs to click on one image first (resulting in showing `.detail` and adding `.image-detail-open`), that's the expected behaviour, isn't it?

BernieHuang2008: for the first comment, i agree with you. (committed)
for the second, i think i should make a video to explain this:

https://github.com/searxng/searxng/assets/88757735/efebe378-fde2-4547-8d43-e3fbc48a05ec

You see, that's the problem and we should solve it.

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -43,7 +43,7 @@ searxng.ready(function () {
     }
   });
 
-  searxng.on('.result', 'keydown', function (e) {
+  searxng.on('.image-detail-open', 'keydown', function (e) {
     var resultElement = getResultElement(e.target);
     if (isImageResult(resultElement)) {
```
bonswouar: If you go with the binding on `.image-detail-open` then this condition `if (isImageResult(resultElement))` isn't required anymore (`.image-detail-open` only exists when an image is opened) as said previously

bonswouar (changes requested):

searx/static/themes/simple/src/js/main/keyboard.js
bonswouar: What are the steps? On my end that seems to work as intended, I click on an image and then use the arrow keys. Maybe I missed something?
I don't have any condition on `e.target` locally though, as said in the reviews

Austin-Olacsi:
good job. i tried the code and it seems to work well. IMO what you mentioned in the video is not an important detail, you shouldn't stress it. i think the code is ready.

BernieHuang2008:
Thank you, i think it's ready too.

searx/static/themes/simple/src/js/main/keyboard.js
BernieHuang2008: well, it's really hard to explain, so i think we should put it away right now. 
We can solve it later ?

BernieHuang2008: > What are the steps? On my end that seems to work as intended, I click on an image and then use the arrow keys. Maybe I missed something? I don't have any condition on `e.target` locally though, as said in the reviews

maybe you missed the step of clicking the blank when the image preview is opened.

BernieHuang2008: i think we can start a new issue or discussion to talk about this. how about #2736 ?

BernieHuang2008:
@return42 @dalf @bonswouar @Hackurei 
Code is ready, please review.

bonswouar:
Looks good to me! 
Although we should figure out what to do about https://github.com/searxng/searxng/discussions/2736

As you said the easy fix would be to bind on document, but I feel like this is not very clean (we would have the event catch everywhere on the site every time a key is pressed, with a check for existing div). And I don't really know why binding on `.detail` doesn't work in that edge case..

If someone with more knowledge about js good practises can share it's thought he's more than welcome :)

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -43,6 +43,26 @@ searxng.ready(function () {
     }
   });
 
+  searxng.on('.image-detail-open', 'keydown', function (e) {
```
dalf: Currently `Ctrl-Left`, `Alt-Left` etc... works too.

Can you add something like this:

```js
if (e.ctrlKey || e.altKey || e.shiftKey || e.metaKey) {
  return
}
```

to ignore the event when Ctrl/Alt/Shift/Meta keys are pressed ?

BernieHuang2008: Committed, are there anything else I can do?

dalf:
With all your tests and comments, I've been able to fix the issue when the user clicks on empty space like described in #2736 :
https://github.com/dalf/searxng/tree/left-right-images

BernieHuang2008:
> With all your tests and comments, I've been able to fix the issue when the user clicks on empty space like described in #2736 :
> https://github.com/dalf/searxng/tree/left-right-images
> 

That's cool and beautiful, I like it.

Austin-Olacsi:
> > With all your tests and comments, I've been able to fix the issue when the user clicks on empty space like described in #2736 :
> > https://github.com/dalf/searxng/tree/left-right-images
> 
> That's cool and beautiful, I like it. Shall I close this PR now, or waiting for your new PR?

why not make the changes here?

BernieHuang2008:
i've updated @dalf 's changes, to this PR.

dalf:
@BernieHuang2008 can you squash your commits and call `make static.build.commit` to add the build files (searxng.min.js) ?

BernieHuang2008:
@dalf i've squashed and built.
what should i do next?

Austin-Olacsi:
@BernieHuang2008 can you possibly squash to 1 commit?

BernieHuang2008:
ok, squashed into 1 commit, you can now merge the PR if nothing wrong.

return42:
> ok, squashed into 1 commit, you can now merge the PR if nothing wrong.

Hy @BernieHuang2008, thanks a lot for your work on this topic :+1: .. and sorry for my absent the last days .. 

> and call `make static.build.commit` to add the build files (searxng.min.js) ?

Don't worry about .. I will do it in my review I started right now.

return42:
FYI I [removed the build files](https://github.com/searxng/searxng/compare/9d8c1d17009806e18f56c5ebc2323ed24d138a25..53a8f4249ba032cffe0bcafe9aa35f828fcfaa90) from your commit and [rebased](https://github.com/searxng/searxng/compare/53a8f4249ba032cffe0bcafe9aa35f828fcfaa90..1b461553a5820b0490161d974e6233f4fe63aabd) your commit on SearXNG's master branch / I'm still in review .. *to be continued*

return42:
> [feature] ArrowKeys to navigate between image results 

the hotkeys should not only applicable to image-results / I pushed [3eff63f](https://github.com/searxng/searxng/pull/2724/commits/3eff63fbfa20bec236d2ed253195e3c09230033f) on top of your branch .. read commit message for more details.

The hotkeys now are ...

image

----

Related:

- https://github.com/searxng/searxng/discussions/2736

----

But there is more work to do, there is an old issue:

- https://github.com/searxng/searxng/issues/2751

I will fix it in a follow up PR.

return42:
@BernieHuang2008 @bonswouar can you please test my modifications .. if you think everything is OK, we should merge this PR.

BernieHuang2008:
> @BernieHuang2008 @bonswouar can you please test my modifications .. if you think everything is OK, we should merge this PR.

Sure,please wait a moment ...

BernieHuang2008:
Something wrong with my codespaces, please wait

return42:
> Well, that's something wrong with my codespaces, so I'm sorry that I can't test it until I found out what's happened.

May be you have to checkout your master branch again since I forced pushed to your master branch when I [removed the build files](https://github.com/searxng/searxng/compare/9d8c1d17009806e18f56c5ebc2323ed24d138a25..53a8f4249ba032cffe0bcafe9aa35f828fcfaa90) from your commit.

BernieHuang2008:
Thank you I've found it.

I think it works well.

But i still have the question, mentioned in the review --- using `e.key` is imprecise, an upper case or lower case letter could make it failed. Are you really sure to use key name instead of key code? May I have the reason?

return42:
>  May I have the reason?

I wrote it down in the commit message: https://github.com/searxng/searxng/pull/2724/commits/3eff63fbfa20bec236d2ed253195e3c09230033f

BernieHuang2008:
> I wrote it down in the commit message: https://github.com/searxng/searxng/pull/2724/commits/3eff63fbfa20bec236d2ed253195e3c09230033f

sorry i missed it.

I agree with you now.

bonswouar:
Looks good!
Although a side effect is that even if no result is selected when you press the arrow keys it will highlight it, but that can be a feature :)

bonswouar (approved):

BernieHuang2008:
> Looks good!
> Although a side effect is that even if no result is selected when you press the arrow keys it will highlight it, but that can be a feature :)

I think we sometimes need this, especially in the "general" mode.
Although we can improve it by smoothing the animation or else.

bonswouar:
> Although we can improve it by smoothing the animation or else.

Personally I don't mind no animation!

return42:
Thanks to all who had worked on this topic :+1: .. and made SearXNG better and better :)

I'm going to merge the PR now ..

PR #2728, opened 2023-09-08
[feat] new engine: svgrepo

Bnyro:
## What does this PR do?
* adds a new engine, svgrepo.com, which provides copyleft svg files
* they do have an official api in fact, which requires an API key
* they also have an unofficial api they use internally, but I didn't really figure out how it works and hence scrapping the HTML results seems simpler

* the engine is very useful for designers to get inspiration or normal people like me to quickly find vectors to use for their own website

* `!svg linux`

## Notes
* As described in #2753, this doesn't really fit the images category, since it provides SVG vectors/icons and no images. Not sure if it's better located under the `other` category.

### DIFF
--- a/searx/engines/svgrepo.py
+++ b/searx/engines/svgrepo.py
@@ -1,0 +1,46 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Svgrepo (images)
+"""
+
+from lxml import html
+from searx.utils import extract_text, eval_xpath, eval_xpath_list
+
+about = {
+    "website": 'https://www.svgrepo.com',
+    "official_api_documentation": 'https://svgapi.com',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+paging = True
+categories = ['images']
+base_url = "https://www.svgrepo.com"
+
+results_xpath = "//div[@class='style_nodeListing__7Nmro']/div"
+url_xpath = ".//a/@href"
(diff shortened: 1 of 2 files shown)


return42:
Strange, see my WIP commit, I had to remove the base_url .. and for a search term like `linux tux` the timeout needs to be at 10sec.

Bnyro:
> Strange, see my WIP commit, I had to remove the base_url ..
>
Indeed weird, I'm pretty sure it worked for me when I created the PR. 
I saw you also removed `base_url` from `url`, which leads to `http://svg/521265/world-1` when searching `world`, so I think it's needed for the `url`.

> and for a search term like `linux tux` the timeout needs to be at 10sec.
>
Yes, the engine is very slow, we shouldn't enable that one by default for sure :)

return42:
> I saw you also removed `base_url` from `url`, which leads to `http://svg/521265/world-1`

Oops, sorry .. I will fix it ...

return42:
I squashed my modifications into your commit / see [diff](https://github.com/searxng/searxng/compare/948dac2..788cd639ffbd50c83e13327e130e893a55b82bae) .. from my side this PR is ready to merge ..

Bnyro:
Awesome!
Do you think that the engine fits well into the `images` category or should we maybe move it to the `other` category, since it in fact lets you search for icons and not for images.

return42:
I think icons are images ;) .. `other`more far away from `icons` as `images`.  / I'm going to merge the PR / thanks for contribute this engine :+1:

PR #2732, opened 2023-09-08
[fix] preferences - layout of the "special queries" tab

return42:
## What does this PR do?

fix the layout of the "special queries" tab in the preferences / before: 

image

and after the patch:

image


searx/static/themes/simple/src/less/preferences.less
```diff
@@ -196,6 +190,18 @@
       width: 100%;
     }
   }
+
+  #tab-content-query {
+    table td,
+    table th {
+      text-align: left;
```
dalf: It should be `.ltr-text-align-left();` to deal with RTL languages.

Bnyro:
Looks much better than the current one!

I'd suggest to add a top margin of `10px` and a bottom margin of `5px` to the `pref-group` class, so that we have a visual separation between the members of the different categories.

return42:
> I'd suggest to add a top margin of `10px` and a bottom margin of `5px` to the `pref-group` 

from my side OK .. but we have already a grouping class `.engine-group` in the engines / its rendered without margins:

image

should we add margin to `.engine-group` and `pref-group` ?

Bnyro:
In my opinion, we should add it to both of them, but we might want to wait for some other opinions first.

It'd make it look a bit more organized and less squashed in my eyes, but that might be subjective.

searx/static/themes/simple/src/less/preferences.less
return42: thanks / done!

return42:
> In my opinion, we should add it to both of them, but we might want to wait for some other opinions first.

To reduce the complexity I decided to to have one layout for both tables / I replaced the `.engine-group` by a `tr.pref-group` selector that fits in all tables.

@Bnyro  may you can have a closer look at ..

Bnyro (approved):

PR #2733, opened 2023-09-08
[feat] engine: implementation of internet archive scholar

### LINKED ISSUE #2713: [Engine Request] scholar.archive.org
Hi, would it be possible to add this engine to the science category?

**Working URL to the engine**
https://scholar.archive.org/

It provides lots of resources in full text, including very old papers.

**Features of this engine**
Doesn’t track its users, fast.

I don’t know. The service is still in beta.

Science.

Thanks,

Bnyro:
## What does this PR do?
* add a new engine: scholar.archive.org
* the engine is very slow, hence timeout is set to 10 seconds (which still timeouts in rare cases)

* Internet Archive Scholar allows searching documents on sites from the past

* !ias world

## Related issues
closes #2713

### DIFF
--- a/searx/engines/internet_archive_scholar.py
+++ b/searx/engines/internet_archive_scholar.py
@@ -1,0 +1,72 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Internet Archive scholar(science)
+"""
+
+from datetime import datetime
+from urllib.parse import urlencode
+from searx.utils import html_to_text
+
+about = {
+    "website": "https://scholar.archive.org/",
+    "wikidata_id": "Q115667709",
+    "official_api_documentation": "https://scholar.archive.org/api/redoc",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+categories = ['science', 'scientific publications']
+paging = True
+
+base_url = "https://scholar.archive.org"
+results_per_page = 15
+
+
+def request(query, params):
+    args = {
+        "q": query,
+        "limit": results_per_page,
+        "offset": (params["pageno"] - 1) * results_per_page,
+    }
+    params["url"] = f"{base_url}/search?{urlencode(args)}"
+    params["headers"]["Accept"] = "application/json"
+    return params
+
+
+def response(resp):
+    results = []
+
+    json = resp.json()
+
+    for result in json["results"]:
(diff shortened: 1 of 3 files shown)


return42:
I changed:

- categories from `it` to `science`
- in my queries a timeout of 6sec was never reached
- I removed the unintended patch of utils/brand.env

TBH, the json engine was made for the `default.html` template, the objects from this engine are scientific articles with a DOI, a full biblio entry and an abstract.

image

I think we should better implement a python module that uses the [paper.html template](https://github.com/searxng/searxng/blob/master/searx/templates/simple/result_templates/paper.html)

These engines are good examples:

- https://github.com/searxng/searxng/blob/master/searx/engines/core.py
- https://github.com/searxng/searxng/blob/master/searx/engines/crossref.py

Bnyro:
> in my queries a timeout of 6sec was never reached
>
Interesting, I had response times of 5 seconds sometimes and hence decided to change it to 6s (since it's disabled by default anyways). Though I'm okay with changing it to 4 or 5s.

I didn't know there's a `paper.html` template, apparently there's really a lot of hidden stuff in SearXNG that I didn't find yet :)
I think it's a very good idea to use that one instead, I will work on the necessary changes in the next days.

return42:
> apparently there's really a lot of hidden stuff in SearXNG that I didn't find yet :)

believe me, I also often find new stuff :) .. the honor for this goes to @dalf, who has already searched for comprehensive and flexible solutions in the searx project and has invested a lot of work in it.

Bnyro:
Fyi, I implemented the suggested change to use "paper.html", so the PR is ready for review.

But this one is of rather low priority in general and for me personally, so it's better to look into other PRs first and look into this one when there's not much other stuff going on at this repo.

return42:
@Bnyro I squashed my modifications into your commit / [diff](https://github.com/searxng/searxng/compare/657ddc945971484c3d56bfd22875fb942d77f11b..1bc730fc8e4f4564b1bfe9c416e5f16cca2d5ba7)

I made the field selection a little more robust by using the get-method of the dictionaries (I had some issues).. I added the engine to one more (sub-) category .. and activated the engine by default:

The engine has better results compared to google-scolar .. we always have a PDF and/or DOI .. the biblio is of high quality .. and the timeout of 5 sec fits to other engines we have activated by default in the science category.

This is really an excellent engine for scientific search / thanks for providing it :rocket:

PR #2738, opened 2023-09-09
Replace chompjs with pure Python code

dalf:
## What does this PR do?

The new implementation is good enough for the current usage (brave)

Remove a dependency with C code.

Check the brave image engine.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,4 +17,3 @@ markdown-it-py==3.0.0
 typing_extensions==4.7.1
 fasttext-predict==0.9.2.1
 pytomlpp==1.0.13
-chompjs==1.2.2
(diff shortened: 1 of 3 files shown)


Bnyro (approved):
Works great, nice job!

This is much better than using the C dependency within ChompJS.

allendema:
I found the js_to_json function from yt-dlp which worked fine.

https://github.com/yt-dlp/yt-dlp/blob/5d0395498d7065aa5e55bac85fa9354b4b0d48eb/yt_dlp/utils/_utils.py#L2663

And tests for it: [test_utils.py#L1099](https://github.com/yt-dlp/yt-dlp/blob/5d0395498d7065aa5e55bac85fa9354b4b0d48eb/test/test_utils.py#L1099)

dalf:
@allendema nice, do you think we should replace the current implementation?

or at least it provides an alternative in case the current one is not enough.

allendema:
No, the current implementation can stay.
If new engine has any edgecases, the above function could be added. But right now there is no need for it.

PR #2743, opened 2023-09-10
[feat] implement feeling lucky feature

### LINKED ISSUE #1128: implement "feeling lucky" feature
Hello everyone and thank you for the great work you're doing!

I've started using SearxNG recently and I love it!
In fact I'd like it to be my main search engine, although I'm missing something coming from DuckDuckGo. As you can guess, I wasn't using DDG for the quality of its results, but mainly for its bang capabilities.
While SearxNG allows for even faster bang redirects than DDG (using for example `!!w` for those who don't know), I also make a heavy use of the "feeling lucky" DDG bang syntax, where `[keywords] !` gets you directly to the first result.

[A comment](https://github.com/searx/searx/issues/870#issuecomment-286585153) in the issues of the original *searx/searx* mentionned that it wasn't allowed for "security reasons", namely that you could get redirected to a malicious website if the results happen to be harmfully skewed for the keywords you entered.

Not only have I never had this problem in many years of using it, but maybe it should be the user's choice whether that can happen or not. I was wondering what was your take on this issue. 

Implementing the "feeling lucky" bang would be a great addition that would make me switch to SearxNG as my main search engine, I'm sure many others are feeling this way too.

Thank you for your attention and for the great work!
S.

Bnyro:
## What does this PR do
* implement a feeling lucky feature, which redirects you to the first result automatically when your search contains `!!` as a part of it

* we use a `QueryPartParser` as for bangs, other query parsers had to be modified too so that they don't consume the `!!` shortcut

* It allows you to quickly navigate to the url of the first suggested result

* `!! foo`
* `bar !!`

## Author's checklist
* It might be worth adding a preference to disable that behavior, since you can't be sure that the url of the first result can be trusted. However, since we also have the direct redirect search syntax `!!bing foo` which is not a preference either (afaik), I didn't yet add the preference for it.

## Related issues
closes #1128

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -150,7 +150,7 @@ class LanguageParser(QueryPartParser):
 class ExternalBangParser(QueryPartParser):
     @staticmethod
     def check(raw_value):
-        return raw_value.startswith('!!')
+        return raw_value.startswith('!!') and len(raw_value) > 2
 
     def __call__(self, raw_value):
         value = raw_value[2:]
@@ -177,7 +177,8 @@ class ExternalBangParser(QueryPartParser):
 class BangParser(QueryPartParser):
     @staticmethod
     def check(raw_value):
-        return raw_value[0] == '!'
+        # make sure it's not any bang with double '!!'
+        return raw_value[0] == '!' and (len(raw_value) < 2 or raw_value[1] != '!')
 
     def __call__(self, raw_value):
         value = raw_value[1:].replace('-', ' ').replace('_', ' ')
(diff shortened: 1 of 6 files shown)


return42 (approved):
Thank you for implementing this feature that our users have been waiting for a long time ... and it works like a charm :)

dalf:
This page needs an update : https://github.com/searxng/searxng/blob/master/searx/infopage/en/search-syntax.md

return42:
> This page needs an update : https://github.com/searxng/searxng/blob/master/searx/infopage/en/search-syntax.md

When I was in review I thought about adding documentation .. but then I decided to wait a while and see how this feature works in practice .. may we came to a point where we have a collision in the interests of "feel lucky" and "external bang" .. I tested various use cases and did not find a collision .. but these had only the uses cases I had in mind.

Bnyro:
> I tested various use cases and did not find a collision .. but these had only the uses cases I had in mind.
>
I've tested this one a lot as well, thus I've also added some patches to not run into any issues where the bang parser and feeling lucky are triggered (e.g. https://github.com/searxng/searxng/pull/2743/files#diff-9e829d8223ecd38adcadb6edc1e7b937bf0eb377f30011929ddacbe1b0a7038cR153).

I think we should be fine concerning that, but it'll become interesting if more additional search syntaxes will be added ...

PR #2746, opened 2023-09-11
[mod] library of congress: fix engine

Bnyro:
## What does this PR do
* fix the loc engine
* it seems like the `api/pictures` endpoint no longer exists, hence we must use `api/photos`

* !loc bridges

### DIFF
--- a/docs/dev/engines/online/loc.rst
+++ b/docs/dev/engines/online/loc.rst
@@ -1,0 +1,13 @@
+.. _loc engine:
+
+===================
+Library of Congress
+===================
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.loc
+  :members:
--- a/searx/engines/loc.py
+++ b/searx/engines/loc.py
@@ -1,67 +1,96 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
+"""Library of Congress: query Photo, Print and Drawing from API endpoint_
+``photos``.
+
+.. _endpoint: https://www.loc.gov/apis/json-and-yaml/requests/endpoints/
+
+.. note::
+
+   Beside the ``photos`` endpoint_ there are more endpoints available / we are
+   looking forward for contributions implementing more endpoints.
+
 """
 
- Library of Congress : images from Prints and Photographs Online Catalog
-
-"""
-
-from json import loads
 from urllib.parse import urlencode
-
+from searx.network import raise_for_httperror
 
 about = {
     "website": 'https://www.loc.gov/pictures/',
     "wikidata_id": 'Q131454',
-    "official_api_documentation": 'https://www.loc.gov/pictures/api',
+    "official_api_documentation": 'https://www.loc.gov/api',
     "use_official_api": True,
     "require_api_key": False,
     "results": 'JSON',
 }
 
 categories = ['images']
-
 paging = True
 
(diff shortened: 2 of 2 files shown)


return42 (approved):
I had to fix some some issues in edge cases / [diff](https://github.com/searxng/searxng/compare/c8cdbdbe36f815c9902793f86c20e189995d2f35..8fa702d51aab3d00775af992ac73a85e76967777#diff-b1abe15a83fca33497536aa2e9e55d66daecff2a913dedb7c3619e3ed3de849cL53-R90):

-  ` result['item'].get('summary'),` is often unset and without content field in the result SearXNG ignores the result item --> `result: invalid content: { .. 'url': '//www.loc.gov/pictures/item/2014684027/', 'title': '15th Cavalry, U.S.A.', 'content': None ...`
- when a search term has none results, loc sends a JSON in a HTTP 404 response --> we have to handle HTTP error codes in the response method `params['raise_for_httperror'] = False`
- When a image URL is `https://memory.loc.gov/pp/grp.gif` we can ignore this result item.

I also added the framework for the documentation .. in the hope there comes the day we have time to implement further *endpoints* .. in preparation for this I added the global variable `endpoint` .. with a default to `photos`.

Before we merge, can you please have a final look at my modifications .. thanks :+1:

Bnyro:
> I had to fix some some issues in edge cases / [diff](https://github.com/searxng/searxng/compare/c8cdbdbe36f815c9902793f86c20e189995d2f35..8fa702d51aab3d00775af992ac73a85e76967777#diff-b1abe15a83fca33497536aa2e9e55d66daecff2a913dedb7c3619e3ed3de849cL53-R90):
> 
>    ` result['item'].get('summary'),` is often unset and without content field in the result SearXNG ignores the result item --> `result: invalid content: { .. 'url': '//www.loc.gov/pictures/item/2014684027/', 'title': '15th Cavalry, U.S.A.', 'content': None ...`
> 
Interesting, I didn't know that. Shouldn't that only be the case for general/web searches though? For images, it doesn't really matter if there's a description or not, does it?
> when a search term has none results, loc sends a JSON in a HTTP 404 response --> we have to handle HTTP error codes in the response method `params['raise_for_httperror'] = False`
> 
Good catch, I should have tested more queries that could fail too! I'll remember that for next time.
> When a image URL is `https://memory.loc.gov/pp/grp.gif` we can ignore this result item.
> 
:+1:
> 
> I also added the framework for the documentation .. in the hope there comes the day we have time to implement further _endpoints_ .. in preparation for this I added the global variable `endpoint` .. with a default to `photos`.
> 
That sounds like a great idea, I saw there are about 10 search categories when using the search UI at loc.gov, so this would probably neat to have. The api should hopefully behave very different to the `photos` API.
> Before we merge, can you please have a final look at my modifications .. thanks 👍
>
I've tested it and it seems to work great :+1: . I noticed a few timeouts however, so we should consider bumping the default timeout in the future maybe? But most of the times, there are no timeout issues however, so maybe it's worth seeing how it behaves in the future and decide then.

return42:
> Shouldn't that only be the case for general/web searches though? 

Can't remember why, but it is also the case for images .. 

>  I noticed a few timeouts however, so we should consider bumping the default timeout in the future maybe?

I also noticed the timeouts .. may we can increase the timeout, but I hesitate since deviantart is enabled by default and all other active engines in the `images` category have a max timeout of 3sec (except artic).

image

PR #2747, opened 2023-09-11
Odysee: time range and languages support

### LINKED ISSUE #2706: [FEATURE REQUEST] language filtering and safe search with odysee
No

We can use, for example, language=de in the search URL. en, de-DE, and en-US also seem to work. There is no list of supported languages afaik, we just need to try things out one by one.

for safe search Moderate/Strict we should use nsfw=false in the URL

**Additional context**
The information that you need for this is here: https://github.com/searx/searx/issues/2504

----
Related

- https://github.com/searxng/searxng/pull/2656
- https://github.com/searxng/searxng/issues/590
- [lbr command line](https://gitlab.com/gardenappl/lbt/-/blob/main/lbt?ref_type=heads)
- [LBRY SDK ](https://github.com/lbryio/lbry-sdk/)

Bnyro:
## What does this PR do?
* Odysee: support for time ranges
* Odysee: support for searching by language

* This makes the Odysee engine feature-complete

* !odysee :es test
* !odysee :de test
-> Compare the results

!odysee and set a time range filter -> Observe filtered results

## Related issues
closes #2706


return42:
@Bnyro I reduced your last commit to the patch of the odysse part / when you run `make data.traits` all engine traits are updated .. and when (by example) brave engine is broken for some reason the brave-traits will be removed unintentional .. [diff](https://github.com/searxng/searxng/compare/c94042d390c333bba82e032b79b96543deb10ac0..593671d1fa22b112fb213ec021bba20b5411baff) .. the issues with anas archive and brave engines should be handled in a PR dedicated to the engines.

I will continue my review the next days ...

return42:
> closes 
> - https://github.com/searxng/searxng/issues/2706

@Bnyro I pushed a commit on top to implement safe safe search / would you please have a finla look / thanks!

Bnyro:
> @Bnyro I pushed a commit on top to implement safe safe search / would you please have a finla look / thanks!
>
If we set `nsfw` to true, it seems like we only get nsfw/unsafe results. Since safesearch is disabled by default, I think it's kind of odd to only display nsfw results from Odysee by default.

Bnyro:
Also see my comment here: https://github.com/searxng/searxng/issues/2706#issuecomment-1713303694

return42 (approved):
> Also see my comment here: https://github.com/searxng/searxng/issues/2706#issuecomment-1713303694

Oops I haven't seen .. you are right!

I removed my commit from your branch .. I'm going to merge :+1:

Issue #2719, opened 2023-09-06
Bug: qwant engine


RishabhSahil:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2023.9.5+4fb26cd96

**How did you install SearXNG?**

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineAPIException
  * Percentage: 100
  * Parameters: `('API error::unknown',)`
  * File name: `searx/engines/qwant.py:129`
  * Function: `response`
  * Code: `raise SearxEngineAPIException('API error::' + msg)`

Error
  * Error: httpx.ReadTimeout
  * Percentage: 0
  * Parameters: `(None, None, 'api.qwant.com')`
  * File name: `searx/search/processors/online.py:118`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

PR #2748, opened 2023-09-11
[feat] implementation of qwant lite for web search

Bnyro:
## What does this PR do?
* add qwant lite for general web search as an experiment
* this is a very basic implementation, if it's able to bypass rate limits it will be implemented as its own module if needed
* my goal is to let some public instances with much traffic use it, and thus enable it by default, to see if there's any benefit over using the Qwant API

* see https://github.com/searxng/searxng/issues/2719

* !qwl hello world

## Related issues
see https://github.com/searxng/searxng/issues/2719

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1280,6 +1280,18 @@ engines:
     categories: [videos, web]
     network: qwant
 
+  - name: qwantlite
+    engine: xpath
+    paging: true
+    search_url: https://lite.qwant.com/?q={query}&p={pageno}
+    results_xpath: //section/article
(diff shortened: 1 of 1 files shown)


return42:
@Bnyro a web engine needs support of languages (regions) .. this is best done in the existing qwant module where we have the engine traits .. I implemented a PR:

- https://github.com/searxng/searxng/pull/2781

which replace this PR / I close this PR, hope that is OK for you.

Bnyro:
Sure, my goal was to keep it as simple as possible to not have wasted too much time when it doesn't work in the end - but that's fine by me.

Issue #2641, opened 2023-08-11
Bing results usually don't appear


geckolinux:
Hi, this report is based on at least three different public instances I've tried, currently running an apparently vanilla version of SearXNG `2023.8.11+905ce2a6f`, although another one has this same behavior with version `2023.7.22+65d8b1a31`.

It appears that Bing is either silently failing or is being weighted into oblivion for _most_ but _not all_ searches. For example with Google and Bing enabled, a search for `anamorphic solids` only reveals Google results. Searching for just `anamorphic` or just `solids` does reveal a normal mix of Google and Bing results.

Same thing for `elephant herding` vs just `elephant` or just `herding`.

But then `ram test` shows Bing and Google results, but `elephant test` returns nothing from Bing, so I'm not seeing a consistent pattern here.

In all of these searches there are no failure messages from Bing, and it's showing 100% reliability in the stats.

Thanks, and best regards!

Bnyro:
From my testing the results are being fetched properly, but the results from Google are getting ranked higher than bing results for certain queries.
So I don't think that's really a bug, you can try increasing the `weight` of bing in `settings.yml` if you're self-hosting an instance.

EDIT: It turns out to not be a weighting problem.

geckolinux:
Hi @Bnyro , thanks for the reply.

To make it easier to determine if it's a weighting problem I simply disabled all the other engines except Bing:

![image](https://github.com/searxng/searxng/assets/2692138/7426850c-035c-40b4-887e-99d2634442fe)

---------------

![image](https://github.com/searxng/searxng/assets/2692138/a387a9d8-1999-4f54-b234-58ac230fe695)

-----------------

![image](https://github.com/searxng/searxng/assets/2692138/574644dc-1b52-4a15-8357-c06dec0229cc)

----------------

Upon further investigation my theory is that Bing is silently failing without throwing any error message, and it appears to be limited to specific regions or common providers' IP address ranges in certain regions:
- `2023.2.20+d669da81+dirty` (THE-1984-AS, IS) --> works
- `2023.7.1+5720844f` (TMA Magenta Telekom, AT) --> works
- `2023.7.14+4380fc2b` (SERBIA-BROADBAND-AS Serbia) --> works
- `2023.7.22+65d8b1a31` (OVH, FR) --> works
- `2023.7.22+65d8b1a31` (AS-CHOOPA, US) --> **broken**
- `2023.7.22+65d8b1a31` (DIGITALOCEAN-ASN, US) --> works
[...]

PR #2749, opened 2023-09-11
[mod] engine bing: improve reliability

Bnyro:
## What does this PR do?
* attempt to improve the situation in #2641
* remove some code that expos
* this doesn't fix the issue completely, in many cases there still are no results

* bing often return zero results for no reason

* `!bing libretube reddit` using this PR, and without

## Author's checklist
* should be tested well before merging (if there's even any notable positive impact, maybe I've just been randomly lucky while testing with some queries)

## Related issues
* see  #2641

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -138,7 +138,6 @@ def request(query, params):
     query_params = {
         # fmt: off
         'q': query,
-        'pq': query,
         'cvid': CVID,
         'qs': 'n',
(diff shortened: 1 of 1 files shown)


Bnyro:
Oh, this also breaks searching by region, I'll have to look at it again when I have time for it.

Bnyro:
Turns out that it doesn't have the wanted effect, hence closing.

PR #2753, opened 2023-09-12
[feat] engine: implementation of google icons/material design icons

Bnyro:
## What does this PR do?
* Unfortunately, it doesn't have a search api. However, we can fetch a list of all available icons (response time is really fast) and filter by name, tag and category ourselves, exactly as the search on <https://fonts.google.com/icons> does it
* I know this is sorta not how engines usually work, but I use it very frequently and hence it would be very handy for me to have it integrated into SearXNG

* Quote from its website: `Material Symbols are our newest icons consolidating over 3,061 glyphs in a single font file with a wide range of design variants.`
* That's very useful for designers or people searching for icons to use for their website
* It doesn't really fit the images category, similar to `Svgrepo`, but rather the category `icons`. Would it be better to put it into the `others` category, create a new `icons` category, or keep it in the `images` category?

* !mi account
* !mi account filled
* !mi bird tree

### DIFF
--- a/searx/engines/material_icons.py
+++ b/searx/engines/material_icons.py
@@ -1,0 +1,61 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Material Icons (images)
+"""
+
+import re
+from json import loads
+
+about = {
+    "website": 'https://fonts.google.com/icons',
+    "wikidata_id": 'Q107315222',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+search_url = "https://fonts.google.com/metadata/icons?key=material_symbols&incomplete=true"
+result_url = "https://fonts.google.com/icons?icon.query={query}&selected=Material+Symbols+Outlined:{icon_name}:FILL@0{fill};wght@400;GRAD@0;opsz@24"  # pylint: disable=line-too-long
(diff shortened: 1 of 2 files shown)


return42:
Hi @Bnyro very nice addition, one point I have: when I search for `!mi bird tree` the resulting URL is:

https://fonts.google.com/icons?icon.query=bird%20tree&selected=Material+Symbols+Outlined:flutter_dash:FILL@00;wght@400;GRAD@0;opsz@24

When you click on the link you see that id did not work due to the search term `icon.query=bird%20tree`  .. I haven't found a solution for .. can you please have a look how we can generate a valid URL / thanks!

Bnyro:
> Hi @Bnyro very nice addition, one point I have: when I search for `!mi bird tree` the resulting URL is:
> 
Thanks for having a look at this!
> https://fonts.google.com/icons?icon.query=bird%20tree&selected=Material+Symbols+Outlined:flutter_dash:FILL@00;wght@400;GRAD@0;opsz@24
> 
> When you click on the link you see that id did not work due to the search term `icon.query=bird%20tree` .. I haven't found a solution for .. can you please have a look how we can generate a valid URL / thanks!
>
Good catch, I've force pushed a fix for that :+1:

return42 (approved):
> Good catch, I've force pushed a fix for that 👍

Yeah .. cool, links are fixed now :+1:

PR #2754, opened 2023-09-12
[fix] engine - openstreetmap currency matching

jazzzooo:
## What does this PR do?

The currency tag was being incorrectly parsed, leading to a full on error whenever there was any currency returned in `extratags`.

It should be obvious that the old code never worked :) But it is very hard to test for. I found it by dumb luck.

I haven't looked hard but `!osm hello world` gives "currency:XBT". 
Previously you would receive `TypeError: unhashable type: 'list'`, but now it  works as intended.

### DIFF
--- a/searx/engines/openstreetmap.py
+++ b/searx/engines/openstreetmap.py
@@ -450,7 +450,7 @@ def get_key_label(key_name, lang):
         # https://taginfo.openstreetmap.org/keys/currency#values
         currency = key_name.split(':')
         if len(currency) > 1:
-            o = CURRENCIES['iso4217'].get(currency)
+            o = CURRENCIES['iso4217'].get(currency[1])
             if o:
                 return get_label(o, lang).lower()
             return currency


Bnyro:
Please consider using more expressive branch names, even though GitHub always suggests `patch-1` that doesn't mean it's an accurate branch name for your changes ;)

For this one I would have used `osm-currency` as branch name for example.

jazzzooo:
@Bnyro thank you for the feedback, will do for the next PR  :+1:

return42 (approved):
perfect / thanks a lot :+1:

dalf:
There is an rendering issue:

image

jazzzooo:
nice, will fix this. I found a USD test location too by using Overpass QL: `!osm Mission San Luis de Apalachee`

return42:
I think it is an old issue / not caused by this PR.  The issue is, that an result can have multiple currencies .. see "Examples" in

- https://wiki.openstreetmap.org/wiki/Key%3Acurrency

the current implementation for handling currencies is not suitable for this purpose.

jazzzooo:
In my opinion the openstreetmap code could use some refactoring, to implement all of this properly, instead of doing it tag by tag. But yes the issue wasn't caused by the PR, just revealed by it. For now I'll make a quick PR so it shows `XBT    yes` or more commonly `USD   yes`. I think that was the original intention of the code anyways.

PR #2755, opened 2023-09-12
[fix] engine - pypi redirect

jazzzooo:
## What does this PR do?

Fixes pypi redirecting `/search?q={query}` to `/search/?q={query}`.

The redirect wastes time and produces a warning.

`!pypi requests`
previously this caused a warning `ErrorContext('searx/search/processors/online.py', 127, 'count_error(', None, '1 redirects, maximum: 0', ('200', 'OK', 'pypi.org')) True` due to the redirect.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1235,7 +1235,7 @@ engines:
     shortcut: pypi
     engine: xpath
     paging: true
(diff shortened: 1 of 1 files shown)


return42 (approved):
> The redirect wastes time and produces a warning.

Thats true .. thanks a lot for having an eye on the logs :)

PR #867, opened 2022-02-04
[fix] invidious engine: store random base_url in param

return42:
## What does this PR do?

- [fix] invidious engine: store random base_url in param
  Suggested-by: @dalf https://github.com/searxng/searxng/pull/862#discussion_r799324861

- [pylint] engines/invidious.py

PR #2756, opened 2023-09-13
[fix] engine - invidious thumbnails

jazzzooo:
## What does this PR do?

Some invidious instances give partial thumbnail url's in the form `/vi/abcdefghi/sddefault.jpg`, where most give full url. This leads to thumbnails not displaying.

Lots of instances have this problem, regardless of the version. I'm not sure why...

Put `https://invidious.io.lol` as your only invidious instance in settings.yml, then do `!iv rick astley`.
Previously all thumbnails failed and gave 404's, now they work.

## Related issues

I don't see an issue for it but dalf mentioned it [here](https://github.com/searxng/searxng/pull/867#issuecomment-1030572960)

### DIFF
--- a/searx/engines/invidious.py
+++ b/searx/engines/invidious.py
@@ -5,7 +5,7 @@
 
 import time
 import random
-from urllib.parse import quote_plus
+from urllib.parse import quote_plus, urlparse
 from dateutil import parser
 
 # about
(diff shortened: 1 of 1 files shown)


return42 (approved):
> Put `https://invidious.io.lol` .. then do `!iv rick astley`.

Oh man, you took me back to the pop era of the 80s when I was a young guy :)

// I can only repeat what I said so often: Thanks again for the fix :+1:

PR #2757, opened 2023-09-13
[fix] settings.yml - change default invidious instances

jazzzooo:
## What does this PR do?

Removes two dead instances that lead to half the requests on a default installation failing.
I have added some new instances from [https://api.invidious.io](https://api.invidious.io/?sort_by=health). My criteria were: 1K+ users, 99%+ health and to be running a recent version. I have tested all the instances. They are also sorted by health.

**Why have 8 instances instead of 4, won't it be more maintenance burden if one goes down?**
- If an instance goes down, only 1/8 requests fail, not 1/4 or as would've been now, 3/4. Thus there is less need to update them and less maintenance burden.
- More geographical diversity.
- Some instance will surely be far from any admin, this encourages them to customize the instances.

`!iv what is love`
with default settings.yml this will fail 50% of the time. Now is works every time.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -872,6 +872,7 @@ engines:
       - https://invidious.flokinet.to
       - https://vid.puffyan.us
       - https://invidious.privacydev.net
+      - https://inv.tux.pizza
     shortcut: iv
     timeout: 3.0
     disabled: true
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -865,11 +865,14 @@ engines:
     # Instanes will be selected randomly, see https://api.invidious.io/ for
     # instances that are stable (good uptime) and close to you.
     base_url:
-      - https://invidious.snopyta.org
+      - https://invidious.io.lol
+      - https://invidious.fdn.fr
+      - https://yt.artemislena.eu
+      - https://invidious.tiekoetter.com
+      - https://invidious.flokinet.to
       - https://vid.puffyan.us
-      # - https://invidious.kavin.rocks  # Error 1020 // Access denied by Cloudflare
-      - https://invidio.xamh.de
-      - https://inv.riverside.rocks
+      - https://invidious.privacydev.net
+      - https://invidious.snopyta.org
     shortcut: iv
     timeout: 3.0
     disabled: true
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -872,7 +872,6 @@ engines:
       - https://invidious.flokinet.to
       - https://vid.puffyan.us
       - https://invidious.privacydev.net
-      - https://invidious.snopyta.org
     shortcut: iv
     timeout: 3.0
     disabled: true


unixfox:
Shouldn't it not fail at all, I mean it should fallback to other ones?

searx/settings.yml
```diff
@@ -865,11 +865,14 @@ engines:
...
+      - https://invidious.fdn.fr
+      - https://yt.artemislena.eu
+      - https://invidious.tiekoetter.com
+      - https://invidious.flokinet.to
       - https://vid.puffyan.us
-      # - https://invidious.kavin.rocks  # Error 1020 // Access denied by Cloudflare
-      - https://invidio.xamh.de
-      - https://inv.riverside.rocks
+      - https://invidious.privacydev.net
+      - https://invidious.snopyta.org
```
unixfox: invidious.snopyta.org one is unmaintained and has been in the past out of date for months.

unixfox (changes requested):

jazzzooo:
@unixfox I'm not sure what *should* happen, but on default settings, whenever https://invidio.xamh.de is picked I get `httpx.ConnectError: [Errno -2] Name or service not known` with a traceback, and no further retries. `inv.riverside.rocks` gives `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` after being redirected to a static page.

Maybe you can test and see if you get the same? My setup might be wrong.

unixfox:
> @unixfox I'm not sure what _should_ happen, but on default settings, whenever [invidio.xamh.de](https://invidio.xamh.de) is picked I get `httpx.ConnectError: [Errno -2] Name or service not known` with a traceback, and no further retries. `inv.riverside.rocks` gives `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` after being redirected to a static page.
> 
> Maybe you can test and see if you get the same? My setup might be wrong.

I might be wrong, maybe it's the current implementation that works like this. I have not tested.

Could you try adding this parameter if it does retry with another instance:
```
retries: 1
```

Example:
```
- name: invidious
  engine: invidious
  retries: 1
```

If it does we might want to add this parameter so that it retries with another instance.

jazzzooo:
Hmmm it doesn't seem to retry, maybe it's due to the errors not being handled? Anyhow that's out of scope for this PR, we could make an issue for it tho.

unixfox (approved):

return42:
> Could you try adding this parameter if it does retry with another instance
>> .. it doesn't seem to retry 

FYI: There are no retries, there is only a random choice implemented, when `base_url` is a list:

https://github.com/searxng/searxng/blob/b2fd6304bfe4c7e1855f6b934cfe83d7b22e89c9/searx/engines/invidious.py#L38-L41

Retries are problematic, as they do not conceptually fit into the request & response method/concept (we will try to change this concept of engine implementations in the future / as one preparation for this, @dalf is currently working on a revision of the network stack).

They are also problematic because they all have to run in one timeout without exceeding it ... another point: if there are dead endpoints they have to be removed from the list to keep the response times permanently small.

// thank you for updating the list, I had planned to do this for a long time but never got around to it :-)

return42:
@jazzzooo its OT here but let me ask you: do you have a CSP and `hostname_replace` setup for the randomly instances .. its an old entry but here is what I posted I while ago:

- https://github.com/searxng/searxng/discussions/911

it's about time we did an implementation for this ..  

- https://github.com/searxng/searxng/issues/140

.. if only i had more time :o

jazzzooo:
@return42 oh I don't actually host my own instance (yet). I'm just using the default settings now to find bugs. So no I haven't set anything up :)

PR #2758, opened 2023-09-13
 [fix] engine - openstreetmap currency rendering

jazzzooo:
## What does this PR do?

Render accepted currencies correctly in openstreetmap. Sorry I missed this with the previous PR. This isn't perfect but it is what the original buggy code was meant to do. In my opinion all this code could use a refactor. 

See the screenshot [here](https://github.com/searxng/searxng/pull/2754#issuecomment-1717060044)

`!osm Hello world`


return42 (approved):
Thank you for the quick fix :+1:

PR #2759, opened 2023-09-13
[fix] engine - piped.music incorrect timestamps

jazzzooo:
## What does this PR do?

Piped will return -1 for all upload date timestamps, but only when filtering for music. Thus all dates become 1969. This change will make all -1 timestamps be None instead, thus not rendering the wrong timestamp.

This issue affects upstream, you can see it at [https://piped.video/results?search_query=test&filter=music_videos](https://piped.video/results?search_query=test&filter=music_videos), but they simply don't show a date. (Maybe it's some YouTube issue, doesn't matter for us)

`!ppdm take on me` and every date is Dec 31, 1969. Now the dates are not rendered.

### DIFF
--- a/searx/engines/piped.py
+++ b/searx/engines/piped.py
@@ -130,13 +130,14 @@ def response(resp):
     json = resp.json()
 
     for result in json["items"]:
-        publishedDate = parser.parse(time.ctime(result.get("uploaded", 0) / 1000))
+        # note: piped returns -1 for all upload times when filtering for music
+        uploaded = result.get("uploaded", -1)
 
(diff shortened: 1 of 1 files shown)


Bnyro:
> (Maybe it's some YouTube issue, doesn't matter for us)
>
(It's not a YouTube issue, they rather just don't return the upload date with their API intentionally for YouTube music content because the upload date is not actually relevant for songs.)

Thanks for having a look at that, I missed that when implementing the engine :)

Bnyro (approved):

PR #2760, opened 2023-09-13
[fix] engine - pub.dev

jazzzooo:
## What does this PR do?

pud.dev wasn't working, now it is.

Yes I did use `//` in the XPath, but the performance impact is negligible (<0.2ms), and this should make it more robust to future changes.

`!pd http`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1214,10 +1214,10 @@ engines:
     shortcut: pd
     search_url: https://pub.dev/packages?q={query}&page={pageno}
     paging: true
-    results_xpath: /html/body/main/div/div[@class="search-results"]/div[@class="packages"]/div
+    results_xpath: //div[contains(@class,"packages-item")]
     url_xpath: ./div/h3/a/@href
     title_xpath: ./div/h3/a
-    content_xpath: ./p[@class="packages-description"]
+    content_xpath: ./div/div/div[contains(@class,"packages-description")]/span
     categories: [packages, it]
     timeout: 3.0
     disabled: true


searx/settings.yml
```diff
@@ -1214,10 +1214,10 @@ engines:
     shortcut: pd
     search_url: https://pub.dev/packages?q={query}&page={pageno}
     paging: true
-    results_xpath: /html/body/main/div/div[@class="search-results"]/div[@class="packages"]/div
+    results_xpath: //div[@class="packages-item"]
```
```suggestion
    results_xpath: //div[contains(@class,"packages-item")]
```
This should make it more robust in case they add an additional class to it.

searx/settings.yml
```diff
@@ -1214,10 +1214,10 @@ engines:
...
     url_xpath: ./div/h3/a/@href
     title_xpath: ./div/h3/a
-    content_xpath: ./p[@class="packages-description"]
+    content_xpath: ./div/div/div/span
```
```suggestion
    content_xpath: ./div/div/div[contains(@class,"packages-description")]/span
```
Makes it a bit easier to read and makes sure to pick the actual package description and not any random div they might add in the future, who knows

searx/settings.yml
jazzzooo: nice, thank you!

return42:
@jazzzooo can you squash the commits into one? / we want to keep the [history](https://github.com/searxng/searxng/commits/master) clean.

If you have a issue with squashing commits, leave a note and I can do it for you.

jazzzooo:
@return42 done :+1:

return42 (approved):
perfect / thanks again :+1:

PR #2763, opened 2023-09-14
[fix] engine - solidtorrents

jazzzooo:
## What does this PR do?

- Fix solidtorrents xpaths and clean up the function
- Fix pages to be 1,2,3 and not 0,20,40
- Remove urls which redirect to solidtorrents.to

Solidtorrents wasn't working and the pages were wrong. Also I made it clearer why we skip some results, and made the xpaths more robust.

`!solid test`, [this query](https://solidtorrents.to/search?q=test) also gives some anime results, those aren't torrents and should be skipped when parsing.


return42 (approved):
Perfect .. thanks a lot :+1:

PR #2764, opened 2023-09-14
[fix] engine - rumble redirect

jazzzooo:
## What does this PR do?

Rumble redirects `/search/video?q=test&page=1` to `/search/video?q=test` which wastes a round trip. Now we don't send the page parameter on page 1.

This might not seem like a big deal, but it tested it and it saves 50ms for me, or one ping RTT, so it's pretty significant. 

`!ru test`

### DIFF
--- a/searx/engines/rumble.py
+++ b/searx/engines/rumble.py
@@ -23,9 +23,7 @@ categories = ['videos']
 paging = True
 
 # search-url
-base_url = 'https://rumble.com'
-# https://rumble.com/search/video?q=searx&page=3
-search_url = base_url + '/search/video?{query}&page={pageno}'
+base_url = 'https://rumble.com/'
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

return42 (approved):
@jazzzooo thanks a lot for the fix .. 

I made a little change to your code, just to show how I would assemble URL arguments, see [diff](https://github.com/searxng/searxng/compare/1ec7d870424dd5b4ca74484430810097202720ef..8435022eb8db74dbfd9206a48ccae078f36bb6ac)

Can you have a final look if it is OK for you / thanks!

jazzzooo:
@return42 Yep that's a lot nicer, thanks. I was silly not to put the `base_url` inside the f-string :)

PR #2769, opened 2023-09-14
[feat] implementation of presearch engine (general, images, videos, news)

### LINKED ISSUE #1939: Add presearch engine
**Working URL to the engine**
https://presearch.com/search?q=searxng

presearch is a distributed search engine.

**Features of this engine**
It is a general search engine, but it displays superior results that are different from those of Google and other major engines.

`general`

**Additional context**

Bnyro:
## What does this PR do?
* implement the presearch engine (htttps://presearch.com)

* !ps foo
* !psimg foo
* !psvid foo
* !psnews foo

## Known limitations
* the engine automatically picks the node to use based on the server location
* the specified time range is only supported for general results

## Related issues
closes #1939

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1215,6 +1215,7 @@ engines:
     search_type: search
     categories: [general, web]
     shortcut: ps
+    disabled: true
 
   - name: presearch images
     engine: presearch
@@ -1227,7 +1228,7 @@ engines:
   - name: presearch videos
     engine: presearch
     search_type: videos
-    categories: [videos, web]
+    categories: [general, web]
     timeout: 4.0
     shortcut: psvid
     disabled: true
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1210,6 +1210,36 @@ engines:
   #    query_str: 'SELECT * from my_table WHERE my_column = %(query)s'
   #    shortcut : psql
 
+  - name: presearch
+    engine: presearch
+    search_type: search
+    categories: [general, web]
+    shortcut: ps
+
+  - name: presearch images
+    engine: presearch
+    search_type: images
+    categories: [images, web]
+    timeout: 4.0
+    shortcut: psimg
+    disabled: true
+
+  - name: presearch videos
+    engine: presearch
+    search_type: videos
+    categories: [videos, web]
+    timeout: 4.0
+    shortcut: psvid
+    disabled: true
+
+  - name: presearch news
+    engine: presearch
+    search_type: news
+    categories: [news, web]
+    timeout: 4.0
+    shortcut: psnews
+    disabled: true
+
   - name: pub.dev
     engine: xpath
     shortcut: pd
--- a/searx/engines/presearch.py
+++ b/searx/engines/presearch.py
@@ -1,6 +1,20 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
 """Presearch (general, images, videos, news)
+
+.. hint::
+
+   The results in the video category are most often links to pages that contain
+   a video, for instance many links from preasearch's video category link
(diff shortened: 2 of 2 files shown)


return42:
FYI: just rebased and removed a leftover [diff](https://github.com/searxng/searxng/compare/611142446b73ddff3ea6a8c60e8ba986f007b72c..cdc059c4ede4098e1d526708955ae0997414b9d5) .. still in review ..

return42:
@Bnyro do we have any chance to add a language or at least a region support?

Bnyro:
> @Bnyro do we have any chance to add a language or at least a region support?
>

I've really been trying to figure that out.
Presearch works differently than other engines:
The language of the results depends on the node that we use, they have nodes everywhere across the world.

I just neither have been able to figure out a way to:
- get a list of all nodes and their regions/language code of the country
- set the location properly

My findings:
- You can search for locations via something like https://api.presearch.com/search-location?location=germ
- When you set the location, the `search_results_location` Cookie will be set, for Germ, France, it would be `JTdCJTIyZm9ybWF0dGVkJTIyJTNBJTIyNjUyNDAlMjBHZXJtJTJDJTIwRnJhbmNlJTIyJTJDJTIybGF0JTIyJTNBJTIyNDIuNzklMjIlMkMlMjJsb24lMjIlM0ElMjIwLjQzJTIyJTdE` -> No clue how that's related to the response of the location search link pasted above.

Bnyro:
From https://presearch.com/assets/app.js:
```
saveLocation:function(t,e){if(t){var n=s({},t),o=JSON.stringify(n),r=btoa(encodeURIComponent(o));i.default.set("search_results_location",r,a),this.state.savedLocation=n,this.hideInput(e),window.location.reload()}}
```

return42:
@Bnyro I placed a commit on top of your branch to improve the implementation .. videos can't be shown by SearXNG's `video.html` template ..  see hint in module's doc-string.

After playing a while with this engine I think the results are not worth to investigate a lot more work / may I did already to much  .. anyway, I think the engine is worth to merge :+1: 

Can you have a final look, before we merge / thanks!

Bnyro:
> @Bnyro I placed a commit on top of your branch to improve the implementation .. videos can't be shown by SearXNG's `video.html` template .. see hint in module's doc-string.
> 
Neat, I've tested this for some time now and couldn't find any issues :+1:
The infobox works well too, I didn't even know presearch has support for it ...

> After playing a while with this engine I think the results are not worth to investigate a lot more work / may I did already to much .. anyway, I think the engine is worth to merge 👍
> 
I agree. Couldn't find any issues, ready to merge from my side! Thanks for your investigations :)

unixfox:
Good except no description for presearch videos:
image

Bnyro:
Yes, presearch video results don't have any description unfortunately. E.g. https://presearch.com/videos?q=hey

return42 (approved):
> I agree. Couldn't find any issues, ready to merge from my side! Thanks for your investigations :)

Thanks / approved :+1:

searx/settings.yml
```diff
@@ -1210,6 +1210,36 @@ engines:
   #    query_str: 'SELECT * from my_table WHERE my_column = %(query)s'
   #    shortcut : psql
 
+  - name: presearch
```
unixfox, line 1213: Could you make it so presearch is disabled by default?

We haven't discussed to add more engines by default.

unixfox (changes requested):

searx/settings.yml
return42: good point .. did it https://github.com/searxng/searxng/compare/7ee74452fb8957a9859d9cad961615d8b4bb767c..845778b0c82525a0eca8dbaf6ed298083337fe1c

unixfox (approved):

PR #2772, opened 2023-09-15
Draft: pixiv engine

Austin-Olacsi:
## What does this PR do?

This is my work on adding a pixiv engine to searxng. pixiv is quite similar to deviantart.

another sweet engine

test !pv

## Author's checklist

The problem with this engine is that the images do not load, however the image URLs are correct. As you should see in the terminal, we get 403 forbidden- but use https://www.pixiv.net as the referer and the image URLs work perfectly (tested with wget). i just don't know how to get searxng to do this with each individual picture that it gets.

current state of this engine:

test

### DIFF
--- a/searx/engines/pixiv.py
+++ b/searx/engines/pixiv.py
@@ -1,0 +1,65 @@
+import httpx
+from urllib.parse import urlencode
+from datetime import datetime
+
+# Engine metadata
+about = {
+    "website": "https://www.pixiv.net/",
+    "wikidata_id": None,
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+# Engine configuration
+paging = True
+categories = ['images']
+
+# Search URL
+base_url = "https://www.pixiv.net/ajax/search/illustrations"
+
+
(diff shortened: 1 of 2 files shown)


Bnyro:
The `Referer` header in the image request must be set to `https://www.pixiv.net/`.

You can test it yourself with `curl "https://i.pximg.net/c/250x250_80_a2/img-master/img/2020/04/29/05/21/46/81142134_p0_square1200.jpg" -H "Referer:https://www.pixiv.net/"`

I don't think we have any logic for that in SearXNG yet, so it would require some refactoring to the image proxy, but in theory we could use the image proxy to set that header for this engine. However, the engine won't be usable without the image proxy of the instance being enabled.

Austin-Olacsi:
reminder for myself:

with this modification to webapp.py, this engine works (when the image proxy is enabled):

        request_headers = {
            'User-Agent': gen_useragent(),
            'Accept': 'image/webp,*/*',
            'Referer': 'https://www.pixiv.net/', #added this line!
            'Accept-Encoding': 'gzip, deflate',
            'Sec-GPC': '1',
            'DNT': '1',
        }


i'll try doing some more work on this later.

PR #2776, opened 2023-09-15
[feat] implementation of the movie database

Bnyro:
## What does this PR do?
* implement https://www.themoviedb.org as an xpath engine

## Why is this change important?
* tmdb is a more open alternative to imdb by amazon
* it allows you to search for movies and tv shows

## How to test this PR locally?
* !tmdb christmas

PR #2780, opened 2023-09-15
[feat] engine: implemenation of moviepilot (de)

Bnyro:
## What does this PR do?
* add a new engine: https://www.moviepilot.de

* Moviepilot allows you to search a large catalogue of movies from everywhere in the world, the results are in German however

* !moviepilot tom cruise
* !moviepilot fsk-12 jahrzehnt-2020er genre-abenteuerfilm

## Author notes
* Please read the documentation on top of `moviepilot.py` before reviewing.

### DIFF
--- a/searx/engines/moviepilot.py
+++ b/searx/engines/moviepilot.py
@@ -1,0 +1,116 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Moviepilot is a German movie database, similar to IMDB or TMDB.  It doesn't
+have any official API, but it uses JSON requests internally to fetch search
+results and suggestions, that's being used in this implementation.
+
+Moviepilot additionally allows to discover movies by certain categories
+or filters, hence we provide the following syntax:
+
+- Any normal search query -> Fetch search results by the query
+
+- A query containing one of the category identifiers ``fsk``, ``genre``,
+  ``jahr``, ``jahrzent``, ``land``, ``online``, ``stimmung`` will be used to
+  search trending items by the provided filters, which are appended to the
+  filter category after a ``-``.
+
+Search examples:
+
+- Normal: ``!mp Tom Cruise``
+- By filter: ``!mp person-Ryan-Gosling``
+- By filter: ``!mp fsk-0 land-deutschland genre-actionfilm``
+- By filter: ``!mp jahrzehnt-2020er online-netflix``
+
+For a list of all public filters, observe the url path when browsing
+
+- https://www.moviepilot.de/filme/beste.
+
+"""
+
+from urllib.parse import urlencode
+from searx.utils import html_to_text
+
+about = {
+    'website': "https://www.moviepilot.de",
+    'official_api_documentation': None,
+    'use_official_api': False,
+    'require_api_key': False,
+    'results': 'JSON',
+    'language': 'de',
+}
+paging = True
+categories = []
+
+base_url = "https://www.moviepilot.de"
+image_url = "https://assets.cdn.moviepilot.de/files/{image_id}/fill/155/223/{filename}"
+
+filter_types = ["fsk", "genre", "jahr", "jahrzehnt", "land", "online", "stimmung", "person"]
+
+
+def request(query, params):
+    query_parts = query.split(" ")
+
+    discovery_filters = []
+    for query_part in query_parts:
+        filter_category_and_value = query_part.split("-", 1)
+
+        if len(filter_category_and_value) < 2:
+            continue
+
+        filter_category = filter_category_and_value[0]
+
+        if filter_category in filter_types:
+            discovery_filters.append(query_part)
+
+    params['discovery'] = len(discovery_filters) != 0
+
+    if params['discovery']:
+        args = {
+            'page': params['pageno'],
+            'order': 'beste',
+        }
+        params["url"] = f"{base_url}/api/discovery?{urlencode(args)}"
+        for discovery_filter in discovery_filters:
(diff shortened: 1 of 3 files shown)


Bnyro:
Test fails don't seem to be related to this PR, no clue...

return42:
> Test fails don't seem to be related to this PR, no clue...

not sure what the cause is .. FYI: I review this PR (don't change anything in the meantime) ..I will push my fixes and then we see if CI gets back to work :)

searx/engines/moviepilot.py
```diff
@@ -0,0 +1,110 @@
...
+    'require_api_key': False,
+    'results': 'JSON',
+    'language': 'de',
+}
+paging = True
+
+base_url = "https://www.moviepilot.de"
+image_url = "https://assets.cdn.moviepilot.de/files/{image_id}/fill/155/223/{filename}"
+
+discovery_type = "filme"  # any of filme, serien, person
```
return42: >  any of filme, serien, person

is `person` really a discovery-type?  As far I can see only ...

image

are discovery types and `person` is a filter .. or I'm wrong?

Bnyro: Oh yes, that's right!

I just saw the `type` set to "person" in some search results, so I thought it would be supported too, but it doesn't really make sense to filter by genre or FSK of a person :joy:

Bnyro: It looks like we can completely drop the `type` query in the discovery and search for series and movies at the same time, which makes much more sense!

Bnyro: I thought it would be a required parameter, but I just tested it: If we don't specify the `type`, we get series and movies.

searx/settings.yml
```diff
@@ -1740,6 +1740,12 @@ engines:
       require_api_key: false
       results: HTML
 
+  - name: moviepilot
+    shotcut: mp
+    engine: moviepilot
+    discovery_category: filme
```
Bnyro: Oh, I called that `discover_type` in `moviepilot.py`, that's probably why CI might fail 🤦‍

Bnyro:
Together with #2776, we have 3 different engines that are for movies and TV shows (IMDb, tmdb and Mobiepilot), so we could consider creating a `movies` category for them (as subcategory of the "other" category).

Not sure if 3 engines are enough for a new category, but the two supported weather engines seem to have their own category too.

searx/settings.yml
return42: Yes, I have seen ;) / I mean the typo in  `shotcut` is the issue --> `shortcut`

Bnyro: Oh, that as well 🤦

return42:
I squashed my suggestions / and the fix we discussed into your commit / [diff](https://github.com/searxng/searxng/compare/3dbc1737eb1a07e58be35f9313640f3e4876a6e1..1e971067dab46dcce4be351d8e61ac6bb543870c)

- fixed the doc strings
- moved person to the filters: ``!mp person-Ryan-Gosling``
- Some results do not have a image in the JSON
- added `result['summary']` to the content
- added metadata to the result

Not sure I a have made everything right with the API, please take a serious look at it :)

----

EDIT: and CI works :)

Bnyro:
> fixed the doc strings
> 
Cool, I didn't really notice the `=` chars should have the same length as the name of the engine :)
Good to know for the next time.
>  moved person to the filters: `!mp person-Ryan-Gosling`
> 
That's pretty awesome, I didn't know that's supported. How did you come across that? Via the search bar https://www.moviepilot.de/filme/beste/ by searching for an actor?
> Some results do not have a image in the JSON
> 
I've experienced that too with my recent testing, though we now handle everything properly when using the `discovery` mode after your change :+1:

>  added `result['summary']` to the content
> added metadata to the result
> 
> EDIT: and CI works :)
>
All in all: Thanks for your changes, works great now, and I didn't find any issues! :tada:

return42:
> Good to know for the next time.

In the Python eco-system the RestructuredText markup (aka reST) is used for all kind of documentation / see [PEP-256](https://peps.python.org/pep-0256/#road-map-to-the-docstring-peps) and [PEP-287](https://peps.python.org/pep-0287/) (BTW its also used to write [PEPs](https://raw.githubusercontent.com/python/peps/main/peps/pep-0287.rst) :).

The build system for reST is docutils, but now mostly Sphinx is used. Sphinx is an extension of docutils, with which one can realize complex requirements for extended documentation, so there is e.g. in Sphinx the intersphinx extension with which a mashup of documentation over the Internet is possible.

There are various other plian-text markups, often propagated is e.g. Markdown, which found however only very late with CommonMark a first standardization and which in the Python eco system no meaning attained, because already long before reST with the reference implementation docutils could specify a standard.

If you work longer with python you should get to know reST, it is basically a simple plain-text markup but like any other markup it has its peculiarities and niches.

To give you an easy start, we have written an article describing and demonstrating the most important markups. 

--> [reST primer](https://docs.searxng.org/dev/reST.html#rest-primer)

Bnyro:
That looks like a really extensive documentation on that topic, that must have been a lot of work!

I'm not really familiar with documenting code in general yet, so your explanations are really welcome to me :)
I have some experience with JavaDoc, so reST seems to be pretty much the Python alternative, but it's much more feature-rich apparently.

Thank you for redirecting me to that article, I'll definitely go through it (but probably not read all of it at one time :) ).

return42:
> but probably not read all of it at one time 

its a primer :) .. just bookmark and have a look from time to time .. at the end its just important to document and the markup can be added or fixed in the review cycle :+1:

---

EDIT: reviews are the way we all learn by doing

PR #2781, opened 2023-09-15
[feat] implementation of qwant lite for web search

return42:
## What does this PR do?

- Implements a scrapper for Qwant-Lite `web-lite` in the python module.
- Set `web-lite` as default in the `settings.yml`

Already described in https://github.com/searxng/searxng/pull/2748

Start instance `make run` and query (e.g.) for a trademark `!qw bmw` .. in different regions (languages) / about regions supported by Qwant see below.

## Related issues

Related: https://github.com/searxng/searxng/issues/2719
Replace: https://github.com/searxng/searxng/pull/2748

-------

https://github.com/searxng/searxng/blob/b56db4e04e0f229ba8d3dc7881382d0f1ebdeb7d/searx/data/engine_traits.json#L3595-L3639

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -1,15 +1,29 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
-"""Qwant (Web, News, Images, Videos)
+"""This engine uses the Qwant API (https://api.qwant.com/v3) to implement Qwant
+-Web, -News, -Images and -Videos.  The API is undocumented but can be reverse
+engineered by reading the network log of https://www.qwant.com/ queries.
 
-This engine uses the Qwant API (https://api.qwant.com/v3). The API is
-undocumented but can be reverse engineered by reading the network log of
-https://www.qwant.com/ queries.
+For Qwant's *web-search* two alternatives are implemented:
 
-This implementation is used by different qwant engines in the settings.yml::
+- ``web``: uses the :py:obj:`api_url` which returns a JSON structure
+- ``web-lite``: uses the :py:obj:`web_lite_url` which returns a HTML page
+
+
+Configuration
+=============
+
+The engine has the following additional settings:
+
+- :py:obj:`qwant_categ`
+
+This implementation is used by different qwant engines in the :ref:`settings.yml
+<settings engine>`:
+
+.. code:: yaml
 
   - name: qwant
-    qwant_categ: web
+    qwant_categ: web-lite  # alternatively use 'web'
     ...
   - name: qwant news
     qwant_categ: news
@@ -21,6 +35,9 @@ This implementation is used by different qwant engines in the settings.yml::
     qwant_categ: videos
     ...
 
+Implementations
+===============
+
 """
 
 from datetime import (
@@ -31,10 +48,17 @@ from json import loads
 from urllib.parse import urlencode
 from flask_babel import gettext
 import babel
+import lxml
 
 from searx.exceptions import SearxEngineAPIException
 from searx.network import raise_for_httperror
 from searx.enginelib.traits import EngineTraits
+
+from searx.utils import (
+    eval_xpath,
+    eval_xpath_list,
+    extract_text,
+)
 
 traits: EngineTraits
 
@@ -51,10 +75,11 @@ about = {
 # engine dependent config
 categories = []
 paging = True
-qwant_categ = None  # web|news|inages|videos
+qwant_categ = None
+"""One of ``web``, ``news``, ``images`` or ``videos``"""
 
 safesearch = True
-safe_search_map = {0: '&safesearch=0', 1: '&safesearch=1', 2: '&safesearch=2'}
+# safe_search_map = {0: '&safesearch=0', 1: '&safesearch=1', 2: '&safesearch=2'}
 
 # fmt: off
 qwant_news_locales = [
@@ -67,7 +92,12 @@ qwant_news_locales = [
 # fmt: on
 
 # search-url
-url = 'https://api.qwant.com/v3/search/{keyword}?{query}&count={count}&offset={offset}'
+
+api_url = 'https://api.qwant.com/v3/search/'
+"""URL of Qwant's API (JSON)"""
+
+web_lite_url = 'https://lite.qwant.com/'
+"""URL of Qwant-Lite (HTML)"""
 
 
 def request(query, params):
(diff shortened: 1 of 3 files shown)


unixfox (approved):
just tested, works great thank you

searx/engines/qwant.py
```diff
@@ -76,38 +105,74 @@ def request(query, params):
...
+def parse_web_lite(resp):
+    """Parse results from Qwant-Lite"""
+
+    results = []
+    dom = lxml.html.fromstring(resp.text)
+
+    for item in eval_xpath_list(dom, '//section/article'):
+        results.append(
+            {
+                'url': extract_text(item.xpath('./span')),
```
Bnyro: Sometimes we use `eval_xpath(dom, selector)`, and sometimes we use `item.xpath(selector)`.

Referring to [utils.py](https://github.com/searxng/searxng/blob/7dfcc3386e6787fe2267ae667fefff51cc4286a6/searx/utils.py#L481), it seems like we should use `eval_xpath` instead because it's cached and hence provides a better performance?

Bnyro:
Still seems to work as I intended it to :+1: 

I'm already curious on its impact when it's going live!

searx/engines/qwant.py
return42: Oops you have found a point where I am often sloppy (will fix it) / to answer your question: yes, its better we use the xpath functions from the utils.py module. Theses functions will raise dedicated SearX-Xpath exceptions ..

https://github.com/searxng/searxng/blob/b4e0d2eedc94df2369a7661f9134fcfd8fd68af4/searx/exceptions.py#L111-L112

https://github.com/searxng/searxng/blob/b4e0d2eedc94df2369a7661f9134fcfd8fd68af4/searx/exceptions.py#L41-L42

EDIT: forgot to mentioned / xpath expression is cached:

https://github.com/searxng/searxng/blob/b4e0d2eedc94df2369a7661f9134fcfd8fd68af4/searx/utils.py#L496-L497

Bnyro: Alright, thanks for the explanation :+1:

return42: Fixed the issues / BTW I found another issue ..

```python
        # qwant-lite delivers only 5 pages maximum
        if params['pageno'] > 5:
            return
```

return42:
> I'm already curious on its impact when it's going live!

HINT: `www.qwant.com` and `lite.qwant.com` have the identical bot blocker in front / to unblock my instance's IP I first had to go to https://www.qwant.com/ (attention not the lite URL) and answer the CAPTCHA  / as described here https://github.com/searxng/searxng/issues/2011#issuecomment-1553317619

EDIT: may I was unclear my IP was blocked the last days .. today I deployed this PR on my instance and I realized I have to unblock first.

searx/engines/qwant.py
```diff
@@ -76,38 +106,78 @@ def request(query, params):
...
-    params['raise_for_httperror'] = False
     return params
 
 
 def response(resp):
-    """Get response from Qwant's search request"""
+
+    if qwant_categ == 'web-lite':
+        return parse_web_lite(resp)
+    return parse_web_api
```
return42, line 158: FYI: I have not seen this typo before I merged this PR / typo is fixed in master branch in 7398d525c

unixfox:
> > I'm already curious on its impact when it's going live!
> 
> HINT: `www.qwant.com` and `lite.qwant.com` have the identical bot blocker in front / to unblock my instance's IP I first had to go to [qwant.com](https://www.qwant.com/) (attention not the lite URL) and answer the CAPTCHA / as described here [#2011 (comment)](https://github.com/searxng/searxng/issues/2011#issuecomment-1553317619)
> 
> EDIT: may I was unclear my IP was blocked the last days .. today I deployed this PR on my instance and I realized I have to unblock first.

What's the outcome of this PR then? Does this unblock qwant on most public instances? Or this is to optimize the code so that we use a proper/smaller API?

Bnyro:
See https://github.com/searxng/searxng/pull/2748.

We had the hope to unblock some instances by that, but it seems that it doesn't work out unfortunately.

unixfox:
Should we revert this PR then? I mean if it doesn't unblock the engines then what are the reasons for keeping it?

return42:
> if it doesn't unblock the engines then what are the reasons for keeping it?

Sorry if I was misleading; this implementation (`web-lite`) is an alternative to the first implementation (`web`).

When an instance has used the `web` option, the instance gets blocked after a while and a CAPTCHA is present.  What I wanted to say above: when you then switch from `web` to `web-lite` the CAPTCHA resists, you have to solve it ones.

The question is: does the CAPTCHA comes back when using `web-lite` .. after testing this implementation 14h I got a CAPTCHA :(

> Should we revert this PR then? I

Not revert / we will keep both implementations .. in the settings we can switch back from `web-lite` to `web` .. but it doesn't matter, because both implementations end with the same CAPTCHA.

unixfox:
> Not revert / we will keep both implementations .. in the settings we can switch back from `web-lite` to `web` .. but it doesn't matter, because both implementations end with the same CAPTCHA.

Well this "lite" thing looks a lot like the implementation of google when we had the "normal" way of fetching google results and the "use mobile UI".

Both existed and probably made the code a bit more complex, but at the end we (a bit more you) decided to keep only one implementation out of the two.

So here I don't see the upside of keeping two kinds of implementation for the same engine, even more the new one doesn't solve the CAPTCHA errors. If there is no benefit, people will never use it, so I think we should remove it, it makes the engine code more complex and that's not something we want.

return42:
> way of fetching google results and the "use mobile UI".
> So here I don't see the upside of keeping two kinds of implementation for the same engine,

Its IMO not comparable .. google was much more complex and when I decided to remove the non API solution I did it to reduce the complexity and get rid of a solution that fails more often then the other.  It was at a time I had much experience with google than I have at this time with qwant .. let both solution in the code for a while, just to experiment with one and the other.

>  it makes the engine code more complex and that's not something we want.

Not as much here, the implementation of the two solutions can be well separated and is therefore also well readable. OT: but while we are talking about "complexity of code" .. the PR #2807 shows very clearly how complex the implementations (and causes of issues) can become if we do not know whether there is a persistence (redis) or not ... the code for this can not be separated and is difficult to read because it is littered with conditions if redis foo else bar. But also from the process point of view it makes a difference because we basically have to implement and maintain two procedures / one effective that uses a persistence and one less effective that does not have a persistence.

unixfox:
I'm ok with letting both in the code for the moment for experimentation, but in the future if it doesn't give us any benefit I would like to see only one of the two implementation be kept.

Having additional code and somewhat duplicated goal (web and web-lite give the same results) will always add more complexity for anyone getting into the code for the first time. Maybe for you the code is easily understandable because you have written it, but for someone else that's a different point of view.

PR #2789, opened 2023-09-16
A solution to #2788

### LINKED ISSUE #2788: Problems about Keyboard Navigation
since #2724 .

**What happened?**
When using arrowkey-navigation on first/last result, it will jump to bottom/top  (that is, loop select).
I think it's not good for the UX, but seems you write a special code that cause this "bug".
https://github.com/searxng/searxng/blob/ec540a967a66156baa06797183cc64c4a3e345be/searx/static/themes/simple/src/js/main/keyboard.js#L237-L239

i think it shouldn't act like that, because users will be confused by this sudden jump.

**How To Reproduce**

- Search anything
- focus on the first result
- press `Arrow-Left`
and you will see the page has scrolled to bottom (the last result).

**Expected behavior**

do nothing

BernieHuang2008:
## What does this PR do?

Solve the problem mentioned in #2788 (Loop-scroll while using ArrowKeys to navigate)

Improved UX (if you agree with me on #2788 )

## Related issues
Closes #2788

### DIFF
--- a/searx/static/themes/simple/src/js/main/keyboard.js
+++ b/searx/static/themes/simple/src/js/main/keyboard.js
@@ -235,13 +235,13 @@ searxng.ready(function () {
         case 'down':
           next = current.nextElementSibling;
           if (next === null) {
-            next = results[0];
+            next = current;
           }
           break;
         case 'up':
           next = current.previousElementSibling;
           if (next === null) {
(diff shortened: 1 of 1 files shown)


return42:
Replaced by PR:

- https://github.com/searxng/searxng/issues/2788

PR #2791, opened 2023-09-16
[fix] tagesschau videos

jazzzooo:
## What does this PR do?

Fixes tagesschau crashing when a video result didn't have `h264s` as codec. Now we fall back on higher resolution, and fail the result instead of crashing otherwise.

Should h264s be first priority tho? If not feel free to change.

`!ts video` doesn't errror now.

### DIFF
--- a/searx/engines/tagesschau.py
+++ b/searx/engines/tagesschau.py
@@ -83,7 +83,8 @@ def _story(item):
 
 
 def _video(item):
-    video_url = item['streams']['h264s']
+    streams = item['streams']
+    video_url = streams.get('h264s') or streams.get('h264m') or streams.get('h264l') or streams.get('h264xl')
     title = item['title']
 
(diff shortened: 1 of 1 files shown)


searx/engines/tagesschau.py
```diff
@@ -83,7 +83,8 @@ def _story(item):
 
 
 def _video(item):
-    video_url = item['streams']['h264s']
+    streams = item['streams']
+    video_url = streams.get('h264s') or streams.get('h264m') or streams.get('h264l') or streams.get('h264xl')
```
return42, line 87: No need for, just to your knowledge: this can also be written by:

```python
streams.get('h264s', streams.get('h264m', streams.get('h264l', streams.get('h264xl'))))
```

return42 (approved):
.. yes fixes the issue :+1:

PR #2792, opened 2023-09-16
[mod] js_variable_to_python: add tests, handle more JS syntax 

dalf:
## What does this PR do?

handle more cases
bring tests from chompjs: comment out tests do not pass

The tests from chompjs give some overview of what is not implemented.

`make test.unit`

## Related issues

Related to https://github.com/searxng/searxng/discussions/2785

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -21,3 +21,4 @@ aiounittest==1.4.2
 yamllint==1.32.0
 wlc==1.13
 coloredlogs==15.0.1
+parameterized==0.9.0
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -38,9 +38,14 @@ _BLOCKED_TAGS = ('script', 'style')
 _ECMA_UNESCAPE4_RE = re.compile(r'%u([0-9a-fA-F]{4})', re.UNICODE)
 _ECMA_UNESCAPE2_RE = re.compile(r'%([0-9a-fA-F]{2})', re.UNICODE)
 
-_JS_QUOTE_KEYS_RE = re.compile(r'([\{\s,])(\w+)(:)')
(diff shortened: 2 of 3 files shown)


return42:
@dalf can you change the commit title to are more meaningful :)

return42:
Just a question, no need for .. does it make sense to isolate such JS stuff into a dedicated python module? .. just to prevent utils.py from getting bigger and bigger over the years.

dalf:
> can you change the commit title to are more meaningful 

done (I hope so).

> does it make sense to isolate such JS stuff into a dedicated python module?

I wonder the same question, the tests are already in a different file.

Also, some performance test are missing.
I know pytest-benchmark with pytest, but I don't know what to use with unittest.

>  just to prevent utils.py from getting bigger and bigger over the years.

The current implementation is not sustainable, this PR is also way to show which tests pass and which not. So it gives an idea of reliability of the current implementation.

I have not found a JS grammar for [lark](https://github.com/lark-parser/lark). 
Even if we found it, the response time has to be checked, however lark might provide a far more robust implementation ... at the code of one more dependency.

Note: perhaps it is possible to write this grammar, since this is a limited subset of JS.

return42:
> I have not found a JS grammar for [lark](https://github.com/lark-parser/lark).

I know you are looking for a perfect solution of this question .. and I don't want to stop you, but aren't we possibly overshooting the mark?  :)

dalf:
> I know you are looking for a perfect solution of this question .. and I don't want to stop you, but aren't we possibly overshooting the mark? :)

Actually I'm not even sure this PR should be merged: the current code works. Perhaps the code in the PR has a better readability, I let you judge.

My idea was to spend one day on possible solutions to provide alternative if the current implementation break. This was yesterday.

return42:
Superseded by:

- https://github.com/searxng/searxng/pull/5477

PR #2794, opened 2023-09-16
Auto deploy preview on okteto and add docker-compose support

unixfox:
## What does this PR do?

This PR automatically deploy a searxng instance available for reviewing a pull request.

And it automatically deletes the instance when the PR is closed.

When new changes are pushed, it even updates the preview environment.

This PR also add a docker-compose.yml which in my opinion I don't why we haven't added one yet because it's so much easier for beginners to do git clone then `docker-compose.yml`. This is also needed by okteto to deploy the preview environment.

In order to be more efficient into our workflow, sometimes it takes less time to just visit the URL for validating the change. Instead of having to git clone locally and run a whole searxng environment.

Currently opened test PR: https://github.com/unixfox/searxngpreviewenv/pull/12

## Author's checklist

- [x] Requires to add OKTETO_TOKEN as a secret env in the project settings.

### DIFF
--- a/.github/workflows/preview-env-delete.yml
+++ b/.github/workflows/preview-env-delete.yml
@@ -1,0 +1,20 @@
+name: Delete preview environment
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  closed:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Context
+      uses: okteto/context@latest
+      with:
+        token: ${{ secrets.OKTETO_TOKEN }}
+
+    - name: Destroy preview environment
+      uses: okteto/destroy-preview@latest
+      with:
+        name: pr-${{ github.event.number }}-unixfox
--- a/.github/workflows/preview-env-deploy.yml
+++ b/.github/workflows/preview-env-deploy.yml
@@ -1,0 +1,23 @@
+name: Deploy preview environment
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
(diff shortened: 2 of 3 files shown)


return42:
Wow .. looks great .. Will these jobs are also fired on clones? .. I mean can we avoid that these jobs a fired on clones?

unixfox:
> Wow .. looks great .. Will these jobs are also fired on clones? .. I mean can we avoid that these jobs a fired on clones?

When you clone a repository github actions is disabled by default.

image

I have made this PR in draft because okteto only allow 4 preview environments. This is too few in my opinion. I'll look into deploying okteto somewhere else.

return42:
> When you clone a repository github actions is disabled by default.

Good to know :+1: .. AFAIK this is new / I seem to remember that the workflows were always activated in the past.

unixfox:
@return42 are we able to configure redis from an environment variable in searxng? I wanted to add redis to the docker-compose.yml file so that we get the full searxng experience :).

return42:
> get the full searxng experience

Yeah .. there is one:

https://github.com/searxng/searxng/blob/89fbac576348ad05076d523b2dfb06b8724649ad/searx/settings_defaults.py#L185

may be its better to have a dedicated `/etc/searxng/` folder with setups for a *reference instance* .. in this folder we can copy the [`settings.yml`](https://github.com/searxng/searxng/blob/master/utils/templates/etc/searxng/settings.yml) and a [limiter.toml](https://github.com/searxng/searxng/blob/master/searx/botdetection/limiter.toml)

unixfox:
Feature added in github actions here: https://github.com/searxng/preview-environments/blob/main/.github/workflows/look-new-pr-test.yaml

PR #2795, opened 2023-09-16
[mod] preferences: implement drop-down menu for hotkeys (default, vim)

return42:
## What does this PR do?

Replace the on/off checkbox of the vim-hotkeys in the preferences by a drop-down menu.

image

Start instance `make run`, go to the preferences and select a *"hotkeys"* layout, save the settings, do a query and press 'h' in the result list.

@BernieHuang2008 I [cherry picked your commit](https://github.com/searxng/searxng/pull/2795/commits) in this PR and slightly modified the commit to fix an issue .. and I rewrote the commit message / thanks a lot for your patch :+1:


searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -117,63 +117,66 @@ searxng.ready(function () {
...
+      }, defaultLayout)
+  }
 
+  var keyBindings = keyBindingLayouts.default;
   if (searxng.settings.hotkeys) {
-    // To add Vim-like key bindings, merge the 'vimKeys' into 'keyBindings'.
-    Object.assign(keyBindings, vimKeys);
+    // Change another key binding layout.
+    keyBindings = keyBindingLayouts[searxng.settings.hotkeys];
   }
```
BernieHuang2008: Since the `hotkeys` has a default value ('default' layout), there's no need to have this `if`.
```suggestion
  var keyBindings = keyBindingLayouts[searxng.settings.hotkeys] || keyBindingLayouts.default;
```
(the last line will set a base case althought no one with normal way could trigger it ...)

BernieHuang2008 (changes requested):

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -66,13 +66,13 @@ searxng.ready(function () {
     'ArrowLeft': {
       key: '&#11013;',
       fun: highlightResult('up'),
-      des: 'Use left arrow to select previous search result',
+      des: 'left arrow to select previous search result',
```
Bnyro:
```suggestion
      des: 'select previous search result',
```
This one and the one after are the only key bindings where we mention the name of the key ("left arrow"), I think we should just drop that, even though we have two different bindings for that in vim mode.

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -66,13 +66,13 @@ searxng.ready(function () {
...
       cat: 'Results'
     },
     'ArrowRight': {
       key: '&#10145;',
       fun: highlightResult('down'),
-      des: 'Use right arrow to select next search result',
+      des: 'right arrow to select next search result',
```
Bnyro: see 'ArrowLeft'

Bnyro (approved):
Still works pretty well from my testings, and it's way more convenient than having a plugin for it instead :+1:

searx/static/themes/simple/src/js/main/keyboard.js
return42: > This one and the one after are the only key bindings where we mention the name of the key ("left arrow"),

yes may its better to remove the name of the key / I added it because the symbol arrows were hard to see ... but maybe I just need a new pair of glasses :)

> even though we have two different bindings for that in vim mode.

I will remove the ArrowKey bindings from the vim table / thanks for pointing out :+1:

BernieHuang2008: Yes, only the arrow icon can cause misunderstand .You can use `ArrowLeft` instead of →, it's longer but clearer.

If you insists on using arrow icon,   `← →` seems more clear than  `⬅` which is currently used...

return42: @Bnyro, @BernieHuang2008: thanks for you hints, I fixed the issues you mentioned, see [diff](https://github.com/searxng/searxng/compare/ec19d4f502f47dc4292ec9ca5d7e0a7a96b5e15c..5c68aeee7311496b1d971f356a99e0c7fbbc0538) .. (resolve conversation).

BernieHuang2008 (approved):

BernieHuang2008:
You'll need further reviews and testings.

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -117,65 +105,80 @@ searxng.ready(function () {
...
-    // To add Vim-like key bindings, merge the 'vimKeys' into 'keyBindings'.
-    Object.assign(keyBindings, vimKeys);
+  var keyBindingLayouts = {
+
+    "default": Object.assign(
+      { /* SearXNG layout */
+        'ArrowLeft': {
+          key: '←',
+          fun: highlightResult('up'),
+          des: 'to select previous search result',
```
Bnyro:
```suggestion
          des: 'select previous search result',
```

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -117,65 +105,80 @@ searxng.ready(function () {
...
+          cat: 'Results'
+        },
+        'ArrowRight': {
+          key: '→',
+          fun: highlightResult('down'),
+          des: 'to select next search result',
```
Bnyro:
```suggestion
          des: 'select next search result',
```

BernieHuang2008:
I think mistakes like description/spelling is not so important now, because we will clean the code after this PR merged.

searx/static/themes/simple/src/js/main/keyboard.js
return42: Sorry, I'm an idiot .. thanks for double check / [fixed](https://github.com/searxng/searxng/compare/5c68aeee7311496b1d971f356a99e0c7fbbc0538..f1d9b9680b9391c03923919278d4cf87582e9a65) :+1:

Bnyro:
> I think mistakes like description/spelling is not so important now, because we will clean the code after this PR merged.
>
Sure, but since we're already working on the files that contain the descriptions we can also just make this PR "perfect" and save time that would be needed for creating and reviewing new PRs.

searx/static/themes/simple/src/js/main/keyboard.js
Bnyro: > Sorry, I'm an idiot ..
>
Certainly not :)
> thanks for double check / [fixed](https://github.com/searxng/searxng/compare/5c68aeee7311496b1d971f356a99e0c7fbbc0538..f1d9b9680b9391c03923919278d4cf87582e9a65) 👍
>
You're always welcome!

return42:
> You'll need further reviews and testings.

Yes, I also requested a review from @dalf and @unixfox .. this PR can wait until they have time for.

return42:
Rebased to solve conflict with preferences and build files .. going to merge to avoid further conflicts.

PR #2799, opened 2023-09-17
Update .gitignore to ignor `utils/brand.env`

BernieHuang2008:
## What does this PR do?
`.gitignor` the `utils/brand.env` which is generated automatically.
`utils/brand.env` contains informations for the local test, and shouldn't be uploaded into `master`.

it really annoyed me because i have to remove it manually every time!

Improve the `Developer-Experience`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ dist/
 local/
 gh-pages/
 *.egg-info/
+utils/brand.env
 
 /package-lock.json
 /node_modules/


return42:
> it really annoyed me because i have to remove it manually every time!

Yes, I understand, this is a drawback .. IDK why you have to remove it manually in your developer environment .. in my environment I git-add only those files I want to have in my commit .. and if needed I revert this file / but most often there is no need to revert this file.

> contains informations for the local test, and shouldn't be uploaded into `master`.

Its not needed in SearXNG's master, but when you (like me and may be others) maintain your SearXNG brand from a branch, you need this file for install and maintenance.

By instance in my [darmarit.org](https://github.com/searxng/searxng/compare/master...return42:searxng:darmarit.org) branch I maintain my instance and [here in this commit](https://github.com/searxng/searxng/commit/cf8645e8fa20070f4beeabba1504436b7a16952f) I patched the environment to the needs of my darmarIT instance:

image

@dalf already ask me these days not to modify this file **when we build the developer environment** .. I'm looking forward for a solution but unfortunately I have not had the time yet.

Since git ignore is not the solution for this drawback I will close this issue.

Nevertheless, thanks a lot for having an eye on this :+1:

BernieHuang2008:
Well you know, vscode has a "git" panel which I often used to avoid `git add` a lot of files. And I will remove it from there every time.

Besides, no one will upload or change `brand.env` often whether it's a instance server or master branch, right? 
`.gitignor` will only prevent developers to **upload** it by accident, but we can also modify it in our own branches, git repos, or locally on the server. In special situation, we can upload it by `git add --force` too.

return42:
> will only prevent developers to upload it by accident,

Not sure how it feels in vscode, but in a common git workflow you can git-add files or even small hunks of code you want to have in a commit. 

My experience with vscode is outdated, but I seem to remember that vscode was not able to differentially compile commits as known from git-add.  As far as I remember, you could only add whole files to a commit in vscode back then.

But these are problems in vscode in which the possibilities of git are only poorly implemented.

The common base we all work with (no matter what IDE individuals use) is git.

If we add a file to .gitignore it is ignored by git .. it will not be listed by git-diff or git-status (since it is completely ignored) .. 

if we were to work with gitignore and --force as you suggest, we would only be replacing one flawed solution with another flawed solution.

The problem is different: we don't need to create a brand.env in a developer environment, this must be created if you want to create your own brand.

So it's more a matter of process than technique ... I'll think of a solution to this when I get a chance.

PR #2800, opened 2023-09-17
highlight color changed to red.

### LINKED ISSUE #2787: [feature] highlight keywords
Highlight the user's search keyword improved the UX a lot. User can judge the result quality better.

i think it's easy, because a XPath or HTML Selector can simply find them.
for example, `bing` will highlight keywords by using `<strong>` tags, and `google` will use `<em class="t55VCb">`.

**Additional context**

image

BernieHuang2008:
## What does this PR do?
turn the highlighted-keywords' color to red.

users can find the keywords more easilly

`make run` is enough.

## Author's checklist
I think it's better both in light theme or dark...
image
image

## Related issues
Closes #2787


Austin-Olacsi:
i'm just not sure about this... maybe we can have a toggle, so this can be turned on or off.

BernieHuang2008:
> i'm just not sure about this... maybe we can have a toggle, so this can be turned on or off.

Can I know why?
Is `red` too sharp?

Bet it's ok in light mode. It's too sharp in dark mode indeed, we can change a color.

Austin-Olacsi:
it's just my brain tells me "look here, it's red" and so i move my eyes all over the screen. so would be nice if i could disable it.

BernieHuang2008:
Emmm, this is a problem.
We can ask opinions from others, and I will try to make it a plugin.

return42:
What a good layout looks like is in the eye of the beholder, so it's always a matter of personal taste ... and as is well known, one can argue about tastes.

An objective evaluation is difficult for me, in case of doubt I look at how it is for the masses. In this case I would orient myself on large platforms such as ... 

By instance when the Query term is **foo** ..

google:

image

startpage:

image

DDG:

image

.. and these layouts are in line with SXNG --> personally I don't want to change the the current highlighting.

BernieHuang2008:
Why my google highlight it for me ...?


Oh I know it. in some languages (such as chinese), highlight the keyword with red is better while in other languages it's worse. So google/bing always have 2 themes.

BernieHuang2008:
I may close this PR currently, until I found a better solution.

return42:
> Oh I know it. in some languages (such as chinese), highlight the keyword with red is better while in other languages it's worse. So google/bing always have 2 themes.

Interesting .. good to know :+1:  ..

I guess there are more things that are handled differently in the Chinese language area, if you ever notice something in SearXNG that can be improved in this regard, then write us about it. Here in the ticket system or in the discussions.

BernieHuang2008:
I'm a Chinese and I'm glad to do this.
I want more Chinese to use Searxng because baidu(China's most popular search engine) is full of ADs......

PR #2047, opened 2022-12-26
Use fasttext-predict instead of fasttext(-wheel)

dalf:
## What does this PR do?

* use [fasttext-predict](https://github.com/searxng/fasttext-predict) :
  * fork of fasttext with only the predict method, drop everything else
  * less than 1MB on the disk (there is no external dependency)
  * the package has wheels and source (see #2028 )
* lazy load fasttext-predict: 
  * zero memory impact
  * load fasttext-predict for the checker or the language autodetection plugin
  * once loaded, it takes 5.5MB of RAM per worker (including the model).

## Why is this change important?

* Drop numpy dependency
* Fix installation issue

## How to test this PR locally?

* check the language autodetection works
* check the docker image size

## Author's checklist

* fasttext(-predict) is more than 200 time faster than langdetect. On a fast CPU:
    * fasttext(-predict): 0.02ms per call or even lower
    * langdetect: 3ms. It would surprise me if it raises to 30ms on a SBC like RaspberryPI. This won't be usable to check the languages of the results for example.
* langdetect uses about 2MB of RAM : with 4 workers, the fasttext-predict overhead is 14MB. I think it is reasonable since it is lazy loaded.
* I'm not sure how these issues / PR affect the SearXNG usage: 
  * https://github.com/Mimino666/langdetect/pull/33
  * https://github.com/Mimino666/langdetect/pull/40

* on the down side
    * fasttext-predict has to be compiled, fasttext-predict for various architecture (exceptions: arm/v7 and Apple Arm).
    * fasttext(-predict) has issue with CJK languages. Hopefully, there is a second model to deal with that: https://github.com/currentslab/fastlangid 

<details>
<summary>usage of fastlangid model</summary>

The fastlangid model is 1MB in RAM. So with 4 workers, the overhead is 18MB compare to langdetect (2 models + the fasttext package).

```python
from typing import Optional
import fasttext

UNCERTAIN_SETS = frozenset(('zh', 'ko', 'ja'))
CHINESE_FAMILY_CODES = frozenset(('zh-hant', 'zh-hans', 'zh-yue'))
MODEL_S_THRESHOLD = 0.93

def load_models(fasttext_model_path: str, fastlangid_model_path: Optional[str]) -> None:
    global FASTTEXT_MODEL, FASTLANGID_MODEL
    FASTTEXT_MODEL = fasttext.load_model(fasttext_model_path)
    if fastlangid_model_path:
        FASTLANGID_MODEL = fasttext.load_model(fastlangid_model_path)
[...]

PR #2801, opened 2023-09-17
searx.utils.detect_language : better support for Japanese, Korean, Chinese languages

dalf:
## What does this PR do?

The project fastlangid has a FastText model to achieve higher accuracy for Japanese, Korean, Chinese languages.

This commits includes the model from this project, and change searx.utils.detect_language to rely on that model as a second stage.

The model is lazy loaded when the base model detect Japanese, Korean or Chinese. 

Drawback : the memory footprint increase with these languages.

https://github.com/currentslab/fastlangid

Draft because I've check that the function works with some quick tests, but I have not update the documentation of the functions ; tests are missing too. 

Improve detection of language autodetection.

image

## Related issues

* Related to https://github.com/searxng/searxng/pull/2047 , click on "usage of fastlangid model"
* Related to https://github.com/searxng/searxng/pull/2800#issuecomment-1722409762 , however here [contains_cjko](https://github.com/searxng/searxng/blob/7dfcc3386e6787fe2267ae667fefff51cc4286a6/searx/webutils.py#L233-L252) might be enough.

ping @BernieHuang2008 : there is need now answer now, just keep that in the corner in case it becomes relevant.


BernieHuang2008:
I'm working on a special plugin for CJK languages, when I've done, I will inform yous.

Do you think it's necessary to put all the CJK special code into a plugin/file ?

BernieHuang2008:
I don't have much time in the weekdays, so it will take weeks to complete.

PR #2804, opened 2023-09-17
[fix] engine - qwant wrong error type

jazzzooo:
## What does this PR do?

Returns TooManyRequestsException from qwant when the [undocumented](https://github.com/NullDev/qwant-api/blob/master/DOCUMENTATION.md#error-codes) error code 24 is received.

Now we return the proper error and can suspend the engine accordingly. I suspect we will never raise SearxEngineAPIException and that line could be removed in the future.

Activate the qwant 'web' engine:

```diff
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1256,7 +1256,7 @@ engines:
       results: HTML
 
   - name: qwant
-    qwant_categ: web-lite
+    qwant_categ: web
     engine: qwant
     shortcut: qw
     categories: [general, web]
```

Spam qwant 10-20 times, then do `!qw test`

## Related issues

Related to #2719

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -112,11 +112,11 @@ def request(query, params):
     args = {'q': query}
     params['raise_for_httperror'] = False
 
-    if qwant_categ == 'web-lite':
+    # all qwant engines (incl qwant-lite) delivers only 5 pages maximum
+    if params['pageno'] > 5:
+        return None
 
-        # qwant-lite delivers only 5 pages maximum
-        if params['pageno'] > 5:
-            return None
+    if qwant_categ == 'web-lite':
 
         url = web_lite_url + '?'
         args['locale'] = q_locale.lower()
@@ -130,21 +130,15 @@ def request(query, params):
 
         args['locale'] = q_locale
         args['safesearch'] = params['safesearch']
-
         args['count'] = 50
-        offset = (params['pageno'] - 1) * args['count']
-        # count + offset must be lower than 250
-        args['offset'] = min(offset, 199)
+        args['offset'] = (params['pageno'] - 1) * args['count']
 
     else:  # web, news, videos
 
         args['locale'] = q_locale
         args['safesearch'] = params['safesearch']
-
         args['count'] = 10
-        offset = (params['pageno'] - 1) * args['count']
-        # count + offset must be lower than 50
-        args['offset'] = min(offset, 39)
+        args['offset'] = (params['pageno'] - 1) * args['count']
 
     params['url'] = url + urlencode(args)
 
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -50,7 +50,7 @@ from flask_babel import gettext
 import babel
 import lxml
 
-from searx.exceptions import SearxEngineAPIException
+from searx.exceptions import SearxEngineAPIException, SearxEngineTooManyRequestsException
 from searx.network import raise_for_httperror
 from searx.enginelib.traits import EngineTraits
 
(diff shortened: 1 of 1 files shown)


unixfox:
Will probably need to be rebased once https://github.com/searxng/searxng/pull/2781 is merged.

return42:
> Will probably need to be rebased once https://github.com/searxng/searxng/pull/2781 is merged.

Yes, but no problem ... I will do it in my review :)

searx/engines/qwant.py
```diff
@@ -118,15 +118,10 @@ def response(resp):
...
-                [
-                    'unknown',
-                ],
-            )
-        )
-        raise SearxEngineAPIException('API error::' + msg)
+        error_code = data.get('error_code')
+        if error_code == 24:
+            raise SearxEngineTooManyRequestsException
+        raise SearxEngineAPIException(error_code)
```
return42: If error_code != 24 .. isn't it better we return the (old) `msg` string (`SearxEngineAPIException('API error::' + msg)`?

jazzzooo: I assumed the `msg` was outdated, I never got a `msg` when testing :unamused:

return42: @jazzzooo I tested with: https://api.qwant.com/v3/search/web?q=foo&locale=de_DE&safesearch=0&count=10&offset=39

```
{'message': ['offset must be a factor of 10'], 'error_code': 1}
```

While I was testing the above I see we have also other issues with this engine (in the request function) .. if it is OK for you I will squash my fixes in your commit.

jazzzooo: @return42 yeah no problem :+1:

return42:
@jazzzooo I squashed the fix we discussed into your branch and to fix an issue in the request function I placed a commit on top of you branch .. the latter one takes into account that qwant delivers only 5 pages maximum.

Before we merge, can you please have I final look .. take into account since we also rebased, to test this PR you need modify the settings.yml file --> I edited your inital comment / see "How to test this PR locally?" above.

jazzzooo:
Nice! Tested and looks good to me :+1: 

My linter complains about the parentheses on line 187 being redundant, but that's not really important, I'll wanna clean up stuff like that project-wide in the future.

return42:
> Nice! Tested and looks good to me 👍

Thanks !

> My linter complains about the parentheses on line 187 being redundant

Hm, thats strange .. I added these parantheses since `SearxEngineTooManyRequestsException` is a class ..

https://github.com/searxng/searxng/blob/7398d525c8b08931b375623b94b5df080bcced8d/searx/exceptions.py#L97

and we want to raise a instance of an exception class .. but maybe I oversee something.

jazzzooo:
@return42 I believe raise doesn't care if it's an instance or a type if no arguments are passed. [See here](https://docs.astral.sh/ruff/rules/unnecessary-paren-on-raise-exception/). Also ruff is nice, and I plan to bring up adding it to the pipeline in the future (it's 2000x faster than pylint on our codebase, but it doesn't need to replace pylint).

return42:
> believe raise doesn't care if it's an instance or a type if no arguments are passed. [See here](https://docs.astral.sh/ruff/rules/unnecessary-paren-on-raise-exception/).

ah,sorry my mistake / good to know :+1:

>  Also ruff is nice .. 

Thanks for pointing out, ruff sounds interesting 

> but it doesn't need to replace pylint).

It will never be able to replace pylint entirely ([How does ruff compare to pylint?](https://github.com/astral-sh/ruff/issues/689)), **but** it can be a valuable addition to our ecosystem ... I am looking forward to your contribution :)

PR #2806, opened 2023-09-17
wikipedia wikidata infobox + disable wikisource

unixfox:
## What does this PR do?

By default, wikipedia and wikidata will only display their results into the infobox.

And disable wikisource by default.

This change comes from the discussion in https://github.com/searxng/searxng/discussions/2802.

Here are the reasons for this change:
- The API behind wikipedia, wikidata and wikisource do not have a smart algorithm like a real search engine (google, ddg and more). They are polluting the results by default, they are not aware of the full of context of the query sent by the user.
  We are not trying to build a whole new search engine, we are just a proxy. We shouldn't try to steal the job from search engines by trying to guess what to display from wiki pages.
  Moreover, the existing search engines already provide the results from wiki pages, for instance if you search for the world "ok" you get wikipedia results as the first result.
- By default and if all the default search engines are broken, the user doesn't get any big error message. He gets meaningless results and that's it.
   
- All the wiki engines enabled by default make it so all the public instances are working fine as searx.space see they are returning results but **the actual search engines do not work**:
   
- Top10 of the public instances from https://searx.space have already disabled all of them, or at least 2 wiki engines out of 3. This is a strong factor that the majority of people will never really use these engines, as top public instances have already disabled them because they do not want to have them by default.

In conclusion, the disadvantages in keeping these wiki engines into the results list are numerous against the advantages they offer.

## Possible improvements

I have been thinking to create a new category called "wiki" with the engines like wikipedia, wikidata, wikisource and so on.

I'm not entirely sure if it's a good idea to add again a new category on top of all the existing ones, but this would be a good tradeoff with the removal of these engines from the general tab.

## Description about the implementation of this PR

I'm not entirely sure if it is done correctly, I created two similar engines in the settings for the same engine in order to get either the infobox or just the results in the list.

Also, I'm not sure if it's implemented properly in the engine code.

@return42 any ideas how to improve that? What do you think?

## Related issues

### DIFF
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -31,8 +31,6 @@ if TYPE_CHECKING:
 
 traits: EngineTraits
 
-display_type = ["infobox"]
-
 # about
 about = {
     "website": 'https://wikidata.org/',
@@ -42,6 +40,12 @@ about = {
     "require_api_key": False,
     "results": 'JSON',
 }
+
+display_type = ["infobox"]
+"""A list of display types composed from ``infobox`` and ``list``.  The latter
+one will add a hit to the result list.  The first one will show a hit in the
+info box.  Both values can be set, or one of the two can be set."""
+
 
 # SPARQL
 SPARQL_ENDPOINT_URL = 'https://query.wikidata.org/sparql'
@@ -270,8 +274,9 @@ def get_results(attribute_result, attributes, language):
                 for url in value.split(', '):
                     infobox_urls.append({'title': attribute.get_label(language), 'url': url, **attribute.kwargs})
                     # "normal" results (not infobox) include official website and Wikipedia links.
-                    if "infobox" in display_type and (attribute.kwargs.get('official') or attribute_type == WDArticle):
+                    if "list" in display_type and (attribute.kwargs.get('official') or attribute_type == WDArticle):
                         results.append({'title': infobox_title, 'url': url, "content": infobox_content})
+
                     # update the infobox_id with the wikipedia URL
                     # first the local wikipedia URL, and as fallback the english wikipedia URL
                     if attribute_type == WDArticle and (
@@ -307,7 +312,13 @@ def get_results(attribute_result, attributes, language):
     # add the wikidata URL at the end
     infobox_urls.append({'title': 'Wikidata', 'url': attribute_result['item']})
 
-    if img_src is None and len(infobox_attributes) == 0 and len(infobox_urls) == 1 and len(infobox_content) == 0:
+    if (
+        "list" in display_type
+        and img_src is None
+        and len(infobox_attributes) == 0
+        and len(infobox_urls) == 1
+        and len(infobox_content) == 0
+    ):
         results.append({'url': infobox_urls[0]['url'], 'title': infobox_title, 'content': infobox_content})
     elif "infobox" in display_type:
         results.append(
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
 
 traits: EngineTraits
 
-display_type = "list"
+display_type = ["infobox"]
 
 # about
 about = {
@@ -270,7 +270,7 @@ def get_results(attribute_result, attributes, language):
                 for url in value.split(', '):
                     infobox_urls.append({'title': attribute.get_label(language), 'url': url, **attribute.kwargs})
                     # "normal" results (not infobox) include official website and Wikipedia links.
-                    if display_type == "list" and (attribute.kwargs.get('official') or attribute_type == WDArticle):
+                    if "infobox" in display_type and (attribute.kwargs.get('official') or attribute_type == WDArticle):
                         results.append({'title': infobox_title, 'url': url, "content": infobox_content})
                     # update the infobox_id with the wikipedia URL
                     # first the local wikipedia URL, and as fallback the english wikipedia URL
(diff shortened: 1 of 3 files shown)


Bnyro:
In my opinion we should introduce a third possible `display_type`, e.g. called `all`, which would return both, the info box and the search results, instead of creating a new engine in settings.yml for that.

If an instance admin enables both wikidata engines for example, it does two requests with the exactly same parameters twice asynchronously, which is just additional resource usage that could be easily avoided.

Apart from that, the changes seem like the right move in my opinion, I also experienced getting only results from wikisource which is kinda useless for a normal general search.

unixfox:
> In my opinion we should introduce a third possible `display_type`, e.g. called `all`, which would return both, the info box and the search results, instead of creating a new engine in settings.yml for that.

Maybe in this case, the normal wiki engine should be display_type `all`. This way we do not break existing setups that have wikipedia forced as enabled, people that do not use `use_default_settings: true`

What do you think?

Bnyro:
> Maybe in this case, the normal wiki engine should be display_type `all`. This way we do not break existing setups that have wikipedia forced as enabled, people that do not use `use_default_settings: true`
> 
> What do you think?
>
Hmm, not sure about that. I would probably set the display_type to `infobox` on my own instance, so I might be based, but in my opinion it's the better default to only use it for the infobox. But I'd like to also hear the opinion of @dalf and @return42 concerning that.

unixfox:
I have made the changes, now it's just one engine for both wikidata and wikipedia. (ping @dalf)

Bnyro:
Oh wow, that looks even better than my proposal :+1:

BernieHuang2008:
Wiki contents are usually displayed at the sidebar, at "info".

So you can just remove all wiki result to the sidebar.

return42:
The `display_type` with default to `infobox` is a very good idea :+1: 

What me wonders, I had to change the "if" conditions to get the expected results (is it only me, did I misunderstood the concept?)

I placed a WIP commit on top to illustrate what I had to change to get excepted output .. not sure if my patch violates the initial intention:

> All the wiki engines enabled by default make it so all the public instances are working fine as searx.space see they are returning results but the actual search engines do not work:

@unixfox can you please double check my modifications .. if you think its OK we should squash the commits in this PR in the first commit before merge.

unixfox:
I think I did test wikipedia and not entirely wikidata, so you made the correct modification, the "if case" was incorrect when I refactored to a single engine. I guess that's a proof that we need multiple eyes on a pull request before merging it :smiley: 

> @unixfox can you please double check my modifications .. if you think its OK we should squash the commits in this PR in the first commit before merge.

Ok for me but instead can't you squash the commits when merging?
image

return42:
> Ok for me but instead can't you squash the commits when merging?

Didn't ever know there is a github button for :) .. I do it in my editor / much simpler for me.

I will do it .. and merge. / thanks for double-check :+1:

unixfox:
> > Ok for me but instead can't you squash the commits when merging?
> 
> Didn't ever know there is a github button for :) .. I do it in my editor / much simpler for me.
> 
> I will do it .. and merge. / thanks for double-check 👍

No please keep the commit history in this PR, don't squash this PR, it's always great to be able to view the history! That's why I always use the github button, this way it's only squash for the main branch not the PR.

For example we might need in the future to find some code from the first implementation (where there was two engines): https://github.com/searxng/searxng/pull/2806/commits/299a3fc7b06c68842242c5d86b4a61278521ea90

unixfox:
Found out a bit too late that I needed to add a tag ([feat]) for the commit sorry for my mistake.

return42:
> Found out a bit too late that I needed to add a tag ([feat]) for the commit sorry for my mistake.

No problem next time you can add one ..

> No please keep the commit history in this PR .. That's why I always use the github button, this way it's only squash for the main branch not the PR.

Hm I'm not familiar with these github buttons .. but it looks like its nothing different to squash by git.

unixfox:
> > No please keep the commit history in this PR .. That's why I always use the github button, this way it's only squash for the main branch not the PR.
> 
> Hm I'm not familiar with these github buttons .. but it looks like its nothing different to squash by git.

Well not exactly as you can see in the commit history we now have the PR ID where the commit existed:

image

Second one is from me, and it clearly shows from which PR the commit originated from. This is very useful to easily become back to a discussion about the commit.

return42:
Seems github button modifies the git commit message and adds the PR on the end, thats all (?).. not sure if we really want to make this to a rule of how a commit messages should look like (makes the entire commit/ create PR / review PR / merge PR process more complicated).

BTW the PR is already shown in github below the commit message:

image

Bnyro:
> Seems github button modifies the git commit message and adds the PR on the end, thats all (?).. not sure if we really want to make this to a rule of how a commit messages should look like (makes the entire commit/ create PR / review PR / merge PR process more complicated).
> 
We can customize the commit title via the UI when using GitHubs `Squash merge` feature, so we don't need the PR number to be included.

unixfox:
> Seems github button modifies the git commit message and adds the PR on the end, thats all (?).. not sure if we really want to make this to a rule of how a commit messages should look like (makes the entire commit/ create PR / review PR / merge PR process more complicated).
> 
> BTW the PR is already shown in github below the commit message:
> 
> image

This will need to be discussed, but we can have a middle ground here.

You continue with your workflow but add the PR ID at the end like GitHub. And everyone else can use the squash commit button from GitHub but like I said in another PR I would like to keep the git history, it's very important.

It's much more convenient than having to click on the commit link then click on the PR ID. On top of that, what if theoretically we need to migrate to a new platform that allow to keep the PR IDs? We would lose all connection between the commits and the pull requests!  
Invidious, a somewhat similar project to us, got a cease and desist letter months ago, treating the fact that the repository would still exist on github so this CAN happen to us: https://github.com/iv-org/invidious/issues/3872!

Anyway we are getting a bit off-topic here, we definitively need to discuss that during a meet.

return42:
> Anyway we are getting a bit off-topic here, we definitively need to discuss that during a meet.

Thats true  .. but just one word to think in the meantime: what we discussed here is when we have one (squashed) comit in a PR .. but what when we have two ore more commits in a PR .. should we really modify the commit message of all the commits in a PR? .. IMO things becomes complicated more than needed (we commit & merge this way since 2014 without missing anything).

PR #2775, opened 2023-09-15
[fix] spelling

jazzzooo:
## What does this PR do?

Fixes 98 spelling mistakes :)

## Why is this change important?

It can be a waste of time when someone contributes a fix of a single typo. This should keep those to a minimum.

I don't think rebuilding the themes is necessary for this. There are no functional changes.

This was done with flake8-spellcheck and sphinxcontrib.spelling.

PR #2807, opened 2023-09-17
[fix] engine - duckduckgo vqd edge-case

jazzzooo:
## What does this PR do?

Fixes a crash on non-Redis instances whenever an invalid vqd token is retried. On instances with Redis, this results in no exception, but  also no result returned on seemingly valid queries.

Let me preface that this fix is dumb, and DuckDuckGo is annoying. If someone has a better solution, please suggest.

I wouldn't have bothered with this if it wasn't for the fact that most public instances report this exception, it's one of the most common exceptions right now, and it brings down the reliability and result count of DuckDuckGo. See [searx.space](https://searx.space) section under the engines tab, the error is IndexError. (This also means that a large part of public instances don't use Redis, as this error only occurs when you don't... Maybe we need to advertise Redis better to admins?)

#### Now to explain the issue... And I might've misunderstood it entirely

We fetch a `vqd` token from duckduckgo whenever we do a query. This token is needed to access results on later pages of DuckDuckGo. This token is then cached in Redis for the associated query, if Redis is enabled. 

We perform a search, and when the results run out, searxng will give the appropriate ***Sorry! we didn't find any results.***. But now, the vqd token is no longer valid because we made an invalid request with the token (on the DuckDuckGo UI there would not have been a next page to click on). If Redis is enabled, the token is still in cache, and when we try the same query with a different page and request it, we get no results because the token was invalid. If Redis is not enabled, we re-request the token on every search (but the token is always the same for the same query), now if we request it again, but we invalidated it, we get a captcha when requesting it. This is not handled correctly by get_vqd and we get an IndexError.

To avoid the captcha is very simple, we need to perform a different query, this will give us a new token. Now if we perform the original query, we get a new token and the captcha is gone.

Make sure to test with and without Redis. Search for something uncommon like some random text `!ddg vrylongqrywthrndmtxth2b9IKK5mV9wWufzOdLSz9YPhg`. Get to the end of the results. Now simply press the search button again for the same query and you will get a crash without Redis. With Redis we cache results (right?), so just go to a page that isn't cached and it will fail too.

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -57,13 +57,13 @@ url = 'https://lite.duckduckgo.com/lite/'
 # url_ping = 'https://duckduckgo.com/t/sl_l'
 
 time_range_dict = {'day': 'd', 'week': 'w', 'month': 'm', 'year': 'y'}
-form_data = {'v': 'l', 'api': 'd.js', 'o': 'json'}
 
 
 def cache_vqd(query, value):
-    """Caches a ``vqd`` value from a query.
+    """Caches a ``vqd`` token from a query, if token is None the cached value
+    is deleted.
 
-    The vqd value depends on the query string and is needed for the follow up
+    The vqd token depends on the query string and is needed for the follow up
     pages or the images loaded by a XMLHttpRequest:
 
     - DuckDuckGo Web: `https://links.duckduckgo.com/d.js?q=...&vqd=...`
@@ -72,18 +72,27 @@ def cache_vqd(query, value):
     """
     c = redisdb.client()
     if c:
-        logger.debug("cache vqd value: %s", value)
         key = 'SearXNG_ddg_vqd' + redislib.secret_hash(query)
-        c.set(key, value, ex=600)
+        if value is not None:
+            logger.debug("cache vqd value: %s", value)
+            c.set(key, value, ex=600)
+        else:
+            # remove from cache
+            c.delete(key)
+
+
+def _get_vqd_value(query):
+    res = get('https://lite.duckduckgo.com/lite/?' + urlencode({'q': query}))
+    doc = lxml.html.fromstring(res.text)
+    return eval_xpath_getindex(doc, "//input[@name='vqd']/@value", 0, None)
 
 
 def get_vqd(query):
     """Returns the ``vqd`` that fits to the *query*.  If there is no ``vqd`` cached
-    (:py:obj:`cache_vqd`) the query is sent to DDG to get a vqd value from the
+    (:py:obj:`cache_vqd`) the query is sent to DDG to get a vqd token from the
     response.
 
     """
-    value = None
     c = redisdb.client()
     if c:
         key = 'SearXNG_ddg_vqd' + redislib.secret_hash(query)
@@ -93,12 +102,20 @@ def get_vqd(query):
             logger.debug("re-use cached vqd value: %s", value)
             return value
 
-    query_url = 'https://lite.duckduckgo.com/lite/?{args}'.format(args=urlencode({'q': query}))
-    res = get(query_url)
-    doc = lxml.html.fromstring(res.text)
-    value = doc.xpath("//input[@name='vqd']/@value")[0]
-    logger.debug("new vqd value: %s", value)
-    cache_vqd(query, value)
+    value = _get_vqd_value(query)
+    if not value:
+        # seems we got a CAPTCHA for this query string, send a dummy request to
+        # release the captcha and then fetch the vqd value for the query string
+        # again.
+        logger.warning("vqd token will no longer work, trying to get a new one by sending another query")
+        _get_vqd_value(f'{query[:3]} duckduckgo')
+        value = _get_vqd_value(query)
+
+    if not value:
+        logger.error("was not able to fetch a valid vqd token from DDG")
+    else:
+        logger.debug("new vqd value: %s", value)
+        cache_vqd(query, value)
     return value
 
 
@@ -241,10 +258,10 @@ def request(query, params):
     # initial page does not have additional data in the input form
     if params['pageno'] > 1:
 
-        params['data']['o'] = form_data.get('o', 'json')
-        params['data']['api'] = form_data.get('api', 'd.js')
-        params['data']['nextParams'] = form_data.get('nextParams', '')
-        params['data']['v'] = form_data.get('v', 'l')
+        params['data']['o'] = 'json'
+        params['data']['api'] = 'd.js'
+        params['data']['nextParams'] = ''
+        params['data']['v'] = 'l'
 
     params['data']['kl'] = eng_region
     params['cookies']['kl'] = eng_region
@@ -274,23 +291,19 @@ def response(resp):
         # the layout of the HTML tables is different.
         result_table = result_table[1]
     elif not len(result_table) >= 3:
-        # no more results
+        # no more results / if we have the vqd token in cache, it's no longer
+        # valid and has to be deleted
+        cache_vqd(resp.search_params['data']['q'], None)
         return []
     else:
         result_table = result_table[2]
         # update form data from response
         form = eval_xpath(doc, '//html/body/form/div[@class="filters"]/table//input/..')
         if len(form):
-
-            form = form[0]
-            form_data['v'] = eval_xpath(form, '//input[@name="v"]/@value')[0]
-            form_data['api'] = eval_xpath(form, '//input[@name="api"]/@value')[0]
-            form_data['o'] = eval_xpath(form, '//input[@name="o"]/@value')[0]
-            logger.debug('form_data: %s', form_data)
-
-            value = eval_xpath(form, '//input[@name="vqd"]/@value')[0]
+            value = eval_xpath_getindex(form[0], "//input[@name='vqd']/@value", 0, None)
             query = resp.search_params['data']['q']
-            cache_vqd(query, value)
+            if value:
+                cache_vqd(query, value)
 
     tr_rows = eval_xpath(result_table, './/tr')
     # In the last <tr> is the form of the 'previous/next page' links


jazzzooo:
Also we could save a whole request on page 1 of ddg by using the same request for getting a `vqd` token and the results. I can try doing this later, it's not a big priority tho.

unixfox:
Note about redis: We have internally discussed whenever we want to make redis mandatory or not.

The outcome is that I personally do not want to have redis as a requirement, this will make running searxng much more difficult for a lot of people. So the solution was to embed redis inside searxng, with a fake redis if the user does not have a redis server installed.

We have not yet implemented something about that.

I would be interested in your opinion.

dalf:
I have not checked into the details, but I really feel `engine_data` result can remove the need for redis, see the usage in the `youtube_noapi` engine:

https://github.com/searxng/searxng/blob/7398d525c8b08931b375623b94b5df080bcced8d/searx/engines/youtube_noapi.py#L98-L103

https://github.com/searxng/searxng/blob/7398d525c8b08931b375623b94b5df080bcced8d/searx/engines/youtube_noapi.py#L40-L53

The content of engine_data are serialized into the web page, and sent again on the next page:

https://github.com/searxng/searxng/blob/7398d525c8b08931b375623b94b5df080bcced8d/searx/templates/simple/results.html#L122

https://github.com/searxng/searxng/blob/7398d525c8b08931b375623b94b5df080bcced8d/searx/templates/simple/results.html#L3-L9

jazzzooo:
@unixfox Thanks for the info. I don't have an opinion on Redis yet, I'm not familiar enough with the top-level code. Probably it's a question to be posed to instance admins, not us developers. A lot of them seem to not be running it, that's surprising to me because the default install script and docker both include it.

@dalf Ah that's good to know. And it seems to work even if I go directly to searx page 2, without page 1. I'll look more into this later, I think it can be a topic for a later PR, this PR only intends to fix the error. But yes saving a whole request on every page load after page 1 for non-redis users would be really good... Also since I already pinged you, if you wanna have a look at #2775 when you have a free minute, that'd be :ok_hand:. Thanks!

return42:
@jazzzooo:

> Fixes a crash on non-Redis instances whenever an invalid vqd token is retried.

if a vqd value is requested to often a CAPTCHA is shown (you mentioned it):

```html
  <div class="anomaly-modal__title">Unfortunately, bots use DuckDuckGo too.</div>
  <div class="anomaly-modal__description">Please complete the following challenge to confirm this search was made by a human.</div>
  <div class="anomaly-modal__instructions">Select all squares containing a duck:</div>
```

In case we do not have redis to cache the vqd value we can use the `engine_data` @dalf mentioned .. but in case we have a redis DB we should better store the vqd value in the redis DB. 

@dalf:

> I have not checked into the details, but I really feel `engine_data` result can remove the need for redis

The `engine_data` is dedicated to a request / response / request cycle of a search from one user.

My intention when I used redis to cache the value was a little bit different: in my reverse engineering I have found that a vqd value is only dependent on the search term and has a certain lifetime, after which it must be renewed.  I used a vqd value (with the same search term) from different IPs when I reverse engineered ddg.

TLDR; a vqd value is valid for a search term and has a live time / to save additional requests this value is cached for a duration of 10min in the redis DB.

return42:
> Probably it's a question to be posed to instance admins, not us developers.

OT here but to give a hint: public instances needs effective bot blocking and for this redis is needed .. on the long term (and this is my opinion) instances operated without redis will become extinct.

jazzzooo:
@return42 

> In case we do not have redis to cache the vqd value we can use the `engine_data` dalf mentioned .. but in case we have a redis DB we should better store the vqd value in the redis DB.

Should I try implementing it in this PR? This would be some more complexity, I suggest we keep this PR focused on fixing the error like mentioned, and then in a future PR me or someone else can implement `engine_data` to avoid the extra request.

Also we can take the Redis debate to Discussions. This issue was never about Redis in particular, funny how every response was about Redis :laughing:, must be a hot issue.

return42:
> Should I try implementing it in this PR? This would be some more complexity, I suggest we keep this PR focused on fixing the error like mentioned, and then in a future PR me or someone else can implement `engine_data` to avoid the extra request.

I tried `engine_data`  and yes, it could be a solution .. I also tried the patch from this PR .. and this PR will improve the situation for non Redis instance .. since both solutions add complexity (see below) I spend a lot of time in "what is better, what alternatives do we have" .. but at some point I stopped:

@dalf @unixfox @Bnyro @mrpaulblack ..

In an internal discussion we agreed that it will always happen that certain things work better with a redis DB than without.  These are the disadvantages that a maintainer have to accept if he don't have a redis DB.

Here we have a typical example for the case that an engine runs worse when no redis DB is available.

We can now think about implementing a solution for redis and one without redis. We would then have to do this for DDG-Web, DDG-Images.  If we ever implement DDG-News or similar we have to implement it again for Redis and without Redis. 

We have to test the code for both variants in all those engines ... in case of error messages we have to determine if the error comes from an instance with Redis DB or if the instance runs without Redis .. and whether that makes any difference at all in this error case .. in case of duplicate errors we have to make this distinction as well .. etc .. etc

*Quite a few balls that we have to keep in the air while juggling.*

My personal decision: we leave it as it is and the instances without Redis DB have the problems they have; the solution is to provide a Redis DB.

Opinions?

jazzzooo:
I'd just like to clarify that this issue affects Redis instances too. It wont give an error, but you wont get any results until you "reset" the token with another query. Ofc please verify, I could be wrong.

return42:
> I'd just like to clarify that this issue affects Redis instances too. It wont give an error, but you wont get any results until you "reset" the token with another query. Ofc please verify, I could be wrong.

Sure? .. I can't reproduce the issue with redis .. where is my issue?

There is one issue in all solutions: when you repeat a query like `!ddg asöflksdölfkadfölkdölfk` n-times .. ddg will send a captcha at one point.

jazzzooo:
Oh you're right! That means my assumptions were wrong that redis and non-redis were facing the same issue, so it makes the if redis, delete from cache statements unnecessary. I was simply getting an error on redis because the captcha was triggering from too many queries. I'll update the PR in a sec

jazzzooo:
My bad again, I can't remove the Redis code because then the SearxEngineCaptchaException won't trigger with Redis. I don't have the energy for this right now, I'll convert to draft and have a stab tomorrow :laughing: 

@return42 Should we raise CaptchaException tho? The captcha goes away when you search something else, so we would be timing out the engine too much?

return42:
> I don't have the energy for this right now, I'll convert to draft and have a stab tomorrow 😆

Same on my side .. lets have some more time in this PR .. give the other team members a chance to review .. may they see what we not see since we are to deep in DDG's rabbit hole right now :)

unixfox:
If there is currently a way to make ddg working again for everyone (whenever they are using redis or not) then we should merge this fix.

We haven't advertised everywhere that we are making redis a requirement, so we shouldn't put the blame on the user. It's not their fault if they are not using redis. Once we make redis a requirement, you have "carte blanche" for removing all the code that don't work without using redis!

jazzzooo:
@unixfox I'm not sure it will make it "working for everyone", but it's true that many instances report this Exception, so handling it will make ddg more reliable. From updated instances on searx.space, I see now that 28 have gotten this error.

@return42 or someone else, what would be the appropriate error here? Sure it's a captcha, but it goes away when you search something different, so I don't think timing out the engine is appropriate. Maybe we raise no error and simply return no results?

I'll do that now, and change if someone has a better idea.

jazzzooo:
Haha I somehow messed up so bad in git that the PR got closed. Anyways, I think this is ready to merge. This accomplishes the goal of the PR, to handle the IndexError. We can revisit this if Redis becomes required. It doesn't make sense to work more on this now imo.

searx/engines/duckduckgo.py
```diff
@@ -77,13 +77,12 @@ def cache_vqd(query, value):
         c.set(key, value, ex=600)
 
 
-def get_vqd(query):
+def get_vqd(query, reset_token):
```
return42: @jazzzooo thanks a lot for your perseverance and patience (I know myself how tiring this work can be) .. one nit I have: I think we do not need the `reset_token` argument, since all caller set it to `False` .. may I'm wrong, can you please have another look?

jazzzooo: @return42 yes no worries. The get_vqd function calls itself on line 109 with reset_token=True. This was my fix for non-redis instances. We get a captcha on a search unless we "reset" the token by sending a different query and then resending the original one. 

So to test: turn off redis, search something uncommon, get to the end of the results. Now search the same thing again. On master you get IndexError (because it's a guaranteed captcha). Now we send another query to get a new token, then send the same one again, it takes a second but no more crash and you get results.

return42:
> The get_vqd function calls itself on line 109

:facepalm: sorry sometime I'm slow-witted :)

Nevertheless I think we do not need a recursion with a break condition .. I tried to write it a little bit more readable (I hope so) .. I placed a WIP commit on top of your branch to demonstrate .. could have have look and give me your feedback (if you are fine with, we can squash the WIP commit int your commit.

Today I made a series of tests .. and "yes" .. your patch improves the situation a lot .. but if we send an unusual search term too often in a very short period of time, sooner or later we will be blocked ... even the dummy request is not able to solve this blockage ... but in most cases the method you suggested (dummy request) should reduce the blockages --> DDG should run much better now :+1:

I deleted the `form` data in my WIP commit / they always had no effect.

Further, you were right (with Redis) when you cleared the cache when the hit list was empty --> I am now clearing the cache in my WIP commit .. as you suggested in your very first commit :rocket: 

Can you please have a final look / thanks a lot for tacking the time :+1:

jazzzooo:
@return42 thanks! Looks good to me, feel free to squash and merge.

It's your call to make whether the third retry is necessary, I never saw a second dummy request improve the situation.

return42:
> I never saw a second dummy request improve the situation.

earlier I thought it would help sporadically .. but now, when I tested again it seems you are right .. I will remove the second dummy request and squash-merge / thanks for having a look.

return42:
In practice it shows that ddg does not like the additional dummy request, the IP is put on a ban list and there is no response to a request at ddg ... you can no longer call the ddg website with the web browser ... the own IP must first cool down for a few minutes before ddg responds to requests again.

On my instance I have 100% timeout --> reliability is by zero .. before this patch ddg had a reliability of 80% on my instance.

Issue #2498, opened 2023-06-13
Bug: google engine


NoExplorer:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2023.6.12

**How did you install SearXNG?**

I am using SearXNG through a public instance https://searx.be

**What happened?**

Search engine Google returns no results with the error `Suspended: access denied` when trying to search something. Accessing Google with its own link (`https://www.google.com`) or its alternatives (`https://www.google.fi` and/or `https://www.google.se`) works just fine, on any browser and device I've tried so far.

**How To Reproduce**

1) Visit the instance's page `https://searx.be`
2) Type something in the search box (anything) and click the Search button (Magnifying glass icon)

**Expected behavior**

Search engine returns results for the requested search query, without an error.

**Screenshots & Logs**

![image](https://github.com/searxng/searxng/assets/37076999/8040653e-84c8-421f-a121-2860e0481281)
![image](https://github.com/searxng/searxng/assets/37076999/2657c858-8841-41ee-ba14-b418386325d3)
![image](https://github.com/searxng/searxng/assets/37076999/881f1fbb-b804-46e3-bec4-634dc4168606)

**Additional context**

If the browser could be an issue, I will list its configuration here:
Browser: LibreWolf 113.0.2-1
Add-Ons active: uBlock Origin , ClearURLs, NoScript Security Suite, Decentraleyes

I am getting errors for many other engines as well, for example:
DuckDuckGo, which returns no search results with error `HTTP connection error`
qwant, which returns no search results with error `server API error`
StartPage, which returns no search results with error `Suspended: CAPTCHA` (**Solvable** by _keeping a tab of StartPage open alongside SearXNG._)

I've tried checking if it is a network error, but with every restart I did to my router I got no progress done.

Some other devices (i.e my Android  phone) face the same issue as described above on the same network.

I hope I provided enough details and info, if not, do not hesitate to tell me.

**Technical report**

Error
  * Error: httpx.TimeoutException
  * Percentage: 0
  * Parameters: `(None, None, None)`
  * File name: `searx/search/processors/online.py:118`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Error
  * Error: timeout
  * Percentage: 0
  * Parameters: `()`
  * File name: `searx/search/processors/abstract.py:102`
  * Function: `handle_exception`
[...]

PR #2808, opened 2023-09-17
[mod] better error message when no results found

unixfox:
## What does this PR do?

Better wording when no results are found and give solutions to the user.

When there are no results, the section `Messages from the search engines` is displayed by default so that the user immediately see all the search engines that do not work.

This change comes from the discussion in https://github.com/searxng/searxng/discussions/2802.

This gives solutions to the user in case no results were found. Previously, the user was left with no solution if he is a beginner.

Preview of the change:

image

## Author's checklist

@return42 I hope I rebuilt the themes correctly. Let me know if it's not how it works.

## Related issues

### DIFF
--- a/searx/templates/simple/messages/no_results.html
+++ b/searx/templates/simple/messages/no_results.html
@@ -1,5 +1,11 @@
 {% from 'simple/icons.html' import icon_big %}
-<div class="dialog-error" role="alert">
+<div class="dialog-error-block" role="alert">
   <p><strong>{{ _('Sorry!') }}</strong></p>
-  <p>{{ _("we didn't find any results. Please use another query or search in more categories.") }}</p>
+  <p>{{ _("No results were found. You can try to:") }}</p>
+  <ul>
+    <li>{{ _("Refresh the page.") }}</li>
+    <li>{{ _("Search for another query or select another category (above).") }}</li>
+    <li>{{ _("Change the search engine used in the preferences:") }} <a href="/preferences">/preferences</a></li>
+    <li>{{ _("Switch to another instance:") }} <a href="https://searx.space">https://searx.space</a></li>
+  </ul>
 </div>
--- a/searx/static/themes/simple/src/less/toolkit.less
+++ b/searx/static/themes/simple/src/less/toolkit.less
@@ -176,6 +176,15 @@ div.selectable_url {
   border-color: var(--color-error);
 }
 
+.dialog-error-block {
+  .dialog();
+
+  display: block;
+  color: var(--color-error);
+  background: var(--color-error-background);
+  border-color: var(--color-error);
+}
+
 .dialog-warning {
   .dialog();
 
--- a/searx/templates/simple/elements/engines_msg.html
+++ b/searx/templates/simple/elements/engines_msg.html
@@ -1,5 +1,9 @@
 <div id="engines_msg">
+  {% if not results and not answers %}
+  <details class="sidebar-collapsable" open>
+  {% else %}
   <details class="sidebar-collapsable">
+  {% endif %}
     <summary class="title" id="engines_msg-title">{{ _('Messages from the search engines') }}</summary>
     <div class="dialog-error" role="alert">
       {{ icon_big('warning') }}
--- a/tests/robot/test_webapp.py
+++ b/tests/robot/test_webapp.py
@@ -75,4 +75,4 @@ def test_search(browser):
     browser.visit(url)
     browser.fill('q', 'test search query')
     browser.find_by_xpath('//button[@type="submit"]').first.click()
-    assert browser.is_text_present('didn\'t find any results')
+    assert browser.is_text_present('No results were found')


searx/templates/simple/messages/no_results.html
```diff
@@ -1,5 +1,11 @@
...
-<div class="dialog-error" role="alert">
+<div class="dialog-error-block" role="alert">
   <p><strong>{{ _('Sorry!') }}</strong></p>
-  <p>{{ _("we didn't find any results. Please use another query or search in more categories.") }}</p>
+  <p>{{ _("No results were found. You can try to:") }}</p>
+  <ul>
+    <li>{{ _("Refresh the page.") }}</li>
+    <li>{{ _("Search for another query or select another category (above).") }}</li>
+    <li>{{ _("Change the search engine used in the preferences:") }} <a href="/preferences">/preferences</a></li>
+    <li>{{ _("Switch to another instance:") }} <a href="https://searx.space">https://searx.space</a></li>
```
Bnyro, line 9:
```suggestion
    <li>{{ _("Change the search engine used in the preferences:") }} <a href="/preferences">/preferences</a>.</li>
    <li>{{ _("Switch to another instance:") }} <a href="https://searx.space">https://searx.space</a>.</li>
```
I'd prefer adding trailing `.` here for consistency, but I'm not really sure if that'd make sense with the grammar then.

Bnyro (approved):

searx/templates/simple/messages/no_results.html
unixfox: Yeah that's the issue. I know it's possible to embed a link inside a "text to translate" but I was having some issues.

unixfox:
@return42 I know you don't completely agree on this change but I would like to get your review.

I think this is a very needed change during this time here where 60% of the public instance returns "no results". We should at least guide our user until we have at least fixed one default engine from the general category or added a new one by default.

return42:
LGTM :+1: .. may you can split the code patches and the build of the static files in two commit / see `make static.build.commit` in our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide.

> @return42 I know you don't completely agree on this change but I would like to get your review.

Hm, may you misunderstood .. I think what you suggested here is a good compromise:

> When there are no results, the section `Messages from the search engines` is displayed by default so that the user immediately see all the search engines that do not work. 

I just wanted to avoid that the error messages are always displayed even if there are hits (that's how it used to be and that was not good).

> this time here where 60% of the public instance returns "no results". 

These engines should activate the bot blocker and the new bot blocker method we [discussed](https://github.com/searxng/searxng/issues/2498#issuecomment-1590625541).  May be this is the next step, to activate the new bot blocker method by default.

BernieHuang2008:
How about just a few result found? For example the wiki engine you mentioned.

The trigger shall be something like "50% of engines didn't respond" or "just 10 result displayed".

jazzzooo:
I haven't followed the discussion of this but my 2 cents are that when so many engines are failing, we don't want to spam the user with errors. To developers and admins it might be helpful, but it looks pretty janky to normal users. There's a reason all popular services have errors to the tune of "Ooopsie woopsie we did bad". So yeah like return42 said, showing messages from engines is ok when there are no hits, but otherwise not so good.

Here is my suggestion for making the message even more user friendly:

Sorry!
No results were found. You can try:
Refreshing.
Searching something else or using a different category.
Select different search engines: /preferences.
Try a different instance: searx.space.

dalf:
Disclaimer: I feel very slow compare to all of you to comment and interact in all these topics, I'm going to try to answer quickly here.

I like this PR because it provides some guidance when everything fails.

A possible extension of this is to provide alternate working engines as described here:
https://github.com/searxng/searxng/discussions/2802#discussioncomment-7027935

> There's a reason all popular services have errors to the tune of "Ooopsie woopsie we did bad".

There is an tension between two axes: user friendliness and monitoring (and privacy).

Yes, popular services shows a generic message but they also rely on various probes and monitoring to make sure everything works as intended. And usually they have a status page like https://www.githubstatus.com/ for people who want to dive into the details.

With SearXNG that's way more complicated since we won't be able to know when the backends ( = the engines)  break.
We need the users to report the issues = they need to be involved into the monitoring.
That's less user friendly but the only current way to get feedbacks as described in https://github.com/searxng/searxng/discussions/2802

There was an attempt to automate the monitoring with the embedded checker, see https://github.com/searx/searx/pull/2419
Perhaps this feature lacks some visibility or it miss the target, I don't know ; that's the topic for another discussion.

return42:
> I feel very slow compare to all of you to comment and interact in all this topics

Its not only you .. I am already quite out of breath :)

> I like this PR because it provides some guidance when everything fails.

Thats what I think .. this PR is a first improvement on a longer road that we still have ahead of us.

> A possible extension of this is to provide alternate working engines as described here:
https://github.com/searxng/searxng/discussions/2802#discussioncomment-7027935

Sounds like a valuable solution .. but TBH: I need more time to think about.

I vote to finalize this PR to have a first improvement .. in a second step we should consider the solution @dalf suggested.  Maybe one of us will find the time to implement a POC for the solution @dalf suggested .. maybe one of us will think of an alternative solution as well.

searx/templates/simple/messages/no_results.html
return42: > but I'm not really sure if that'd make sense with the grammar then.

What I have often seen in UI: when a sentence start with a upper letter there is most
often a dot at the end, when the sentence start with a lower letter there is no dot at the end.

In the context of translations its most often easier to handle strings without a dot at the end which why I would prefer less dots :)

BernieHuang2008:
> they also rely on various probes and monitoring to make sure everything works as intended. 
You're right! The only way for SearXNG there's a problem is user report! So can we have a sort of "quick report button" that allows user report this to the admin?

I'm not saying report to github issue, but to the instance admin.
We can have a count in each instance: if many users reports a same error, than we shall inform the instance admin to solve this problem(such as network problem) ASAP.

BernieHuang2008:
> they also rely on various probes and monitoring to make sure everything works as intended. 

You're right! The only way for SearXNG there's a problem is user report! So can we have a sort of "quick report button" that allows user report this to the admin?

I'm not saying report to github issue, but to the instance admin.
We can have a count in each instance: if many users reports a same error, than we shall inform the instance admin to solve this problem(such as network problem) ASAP. 

If the admin solved it quickly, we can mark it as `well-maintained`. We can encourage admins to maintain their engine in this way.
If he don't, we can do some sort of punishment(I don't know if it's good, do or not depends on you).

This will improve the quality of each instance I think.

BernieHuang2008:
Without good maintaining, everything is treating symptoms and not the root cause.
( although I think this is another topic, I strongly advise you to focus on how to encourage admins to maintain their instances.)

unixfox:
Please use the edit button @BernieHuang2008.

I have moved your discussion to https://github.com/searxng/searxng/discussions/2802#discussioncomment-7033140, we are getting offtopic here.

return42 (approved):
@unixfox from my side this PR is ready to merge .. except what I posted in a previous comment:

> LGTM :+1: .. may you can split the code patches and the build of the static files in two commit / see `make static.build.commit` in our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide.

if you have issues with, I can do it .. and merge the PR.

unixfox:
So you want the files with the extension .css, .css.map and .less in a separate commit if I understand well?

unixfox:
Done! I have created a separate PR: https://github.com/searxng/searxng/pull/2811

Bnyro:
I think it would be better to:
* Do the theme changes here in the [better error message when no results found](https://github.com/searxng/searxng/pull/2808/commits/6c36f612537cf3bb86f80c0c1d19bea1c5d748e1) commit
* Run `static.build.commit` as a second commit

I don't think it makes much sense to split this into two PRs, since the changes belong together.

unixfox:
I think I confused PR and commit, to be fair the workflow is really strange, and we probably need to write some kind of guide if we want to do it like this, might be confusing for new users.

I'll close the other PR and add back the theme change in this one.

dalf (approved):

return42:
> to be fair the workflow is really strange, and we probably need to write some kind of guide if we want to do it like this, might be confusing for new users.

I already linked the docs two times in this PR ..

> LGTM :+1: .. may you can split the code patches and the build of the static files in two commit / see `make static.build.commit` in our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide.

unixfox:
> > to be fair the workflow is really strange, and we probably need to write some kind of guide if we want to do it like this, might be confusing for new users.
> 
> I already linked the docs two times in this PR ..
> 
> > LGTM 👍 .. may you can split the code patches and the build of the static files in two commit / see `make static.build.commit` in our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide.

I think this page needs a restructure because it's not defined as exact steps of our workflows.

It's feels like these are just comments explaining what does our Makefile do.

----

I was building the static files right now, I was about to include them into this PR.

return42:
Sorry I accidentally  merge the PR .. I will add the missing static.build in master

----

EDIT: done in 3ac7c40b6

PR #2813, opened 2023-09-19
[mod] Switch back to qwant web normal API

### LINKED ISSUE #2812: Qwant Lite: Incomplete URLs in results where links are too long
2023.9.18+90b0bfd1b
searxng-docker
**What happened?**
If a Qwant result link title has three dots (...), it uses those three dots in the result URL as well.
**How To Reproduce**
Search with Qwant (qwant **lite** must be used in settings.yml, seems like the default as of recent commits)
**Expected behavior**
The full URL should be in the result, not a trimmed one where the link becomes invalid.
**Screenshots & Logs**
image

unixfox:
## What does this PR do?

This switch back to the old qwant web normal API.

The new qwant web lite break the links in the results: https://github.com/searxng/searxng/issues/2812

And we JSON API is much more reliable than parsing HTML.

## Related issues

Avoiding the bug by default in #2812

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1264,7 +1264,7 @@ engines:
       results: HTML
 
   - name: qwant
-    qwant_categ: web-lite
+    qwant_categ: web
     engine: qwant
     shortcut: qw
     categories: [general, web]


Bnyro (approved):
Since Qwant uses redirects as href for Qwant Lite, we can't fix the issue with Qwant Lite unfortunately :/

So I think it's best to disable Qwant Lite or completely remove it.

unixfox:
Note about the PR https://github.com/searxng/searxng/pull/2816, I would still like to have the qwant JSON API set by default because:

- It has worked fine for years (even though there is a captcha on public instances)
- JSON API is much more reliable than the HTML one
- the HTML one doesn't give benefit over the JSON one

I'm not against setting the HTML API (web-lite) by default in the future, but we should let a period of testing for this API before making it mainstream.

return42:
@unixfox the branch of this PR was in the https://github.com/searxng/searxng/ repository .. I think it is better when we use our forks for branching & PR .. PR has been merged .. I now delete this branch.

unixfox:
Yes sorry my mistake, I edited through github. I should be doing that in my repository.

PR #2817, opened 2023-09-19
[fix] Search on category select without JavaScript PR #2740

return42:
Small addendum to #2740; search_on_category_select is now no longer a plugin.

Related: https://github.com/searxng/searxng/pull/2740#issuecomment-1725437584


Bnyro (approved):

Bnyro:
Thank you for fixing this! <3

PR #2818, opened 2023-09-19
[feat] engine: implementation of wallhaven

Bnyro:
## What does this PR do?
* adds a new engine for images: wallhaven
* it's enabled by default because the results are really good and the response times short

* because the engine is very useful if searching for anything nature-related

* !wallhaven sunset

### DIFF
--- a/searx/engines/wallhaven.py
+++ b/searx/engines/wallhaven.py
@@ -1,0 +1,84 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Wallhaven_ is a site created by and for people who like wallpapers.
+
+.. _Wallhaven: https://wallhaven.cc/about#Copyright
+"""
+
+from datetime import datetime
+from urllib.parse import urlencode
+
+about = {
+    'website': 'https://wallhaven.cc/',
+    'official_api_documentation': 'https://wallhaven.cc/help/api',
+    'use_official_api': True,
+    'require_api_key': False,
+    'results': 'JSON',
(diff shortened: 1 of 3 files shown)


searx/engines/wallhaven.py
```diff
@@ -0,0 +1,62 @@
...
+    'require_api_key': False,
+    'results': 'JSON',
+}
+categories = ['images']
+paging = True
+
+base_url = "https://wallhaven.cc"
+
+# using an api key is optional and will increase the rate limit
+api_key = ''
```
return42: @Bnyro do you know where the rate limits are?

Bnyro: 45 requests per minute, see https://wallhaven.cc/help/api#limits

Bnyro: It's possible that the API key doesn't actually impact the rate limit, not sure about that. In their docs, it only says that the api key would be used for users to access "some of their settings" ...

return42 (approved):
> it's enabled by default because the results are really good and the response times short

Yes, engine is very fast .. good default addition to the image category.

FYI: I squashed some doc into your commit [diff](https://github.com/searxng/searxng/compare/384b2d002d5016f6200e8ae87301ee8f1df8e607..7b9aefe0063e196c34d8c8b62a9b48ffb7ff3c46)

Bnyro:
Thanks for the addition!

Indeed, the description of sfw, sketchy & nsfw is worth having its place in the docs :joy:

PR #2819, opened 2023-09-19
[feat] engine: implementation of imgur

Bnyro:
## What does this PR do?
* add a new engine: imgur, via xpath
* there's also an official API, but it requires a key and hence it makes more sense to scrape the HTML results

* imgur is very commonly used, mostly for memes but also for serious image uploads (which we're aiming for)

* !imgur sunset

### DIFF
--- a/searx/engines/imgur.py
+++ b/searx/engines/imgur.py
@@ -1,0 +1,66 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Imgur (images)
+"""
+
+from urllib.parse import urlencode
+from lxml import html
+from searx.utils import extract_text, eval_xpath, eval_xpath_list
+
+about = {
+    "website": 'https://imgur.com/',
+    "wikidata_id": 'Q107565255',
+    "official_api_documentation": 'https://api.imgur.com/',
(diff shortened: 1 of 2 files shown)


return42:
When I search for `!imgur foo :en` I got one result item that dos not have an image:

image

When I click on the link: https://imgur.com/gallery/UNfOD .. I see there are two images (videos) .. that might be the reason (?).

Bnyro:
> When I search for `!imgur foo :en` I got one result item that dos not have an image:
> 
> image
> 
> When I click on the link: https://imgur.com/gallery/UNfOD .. I see there are two images (videos) .. that might be the reason (?).
>
Yes, that actually seems like an issue at imgur's side - the same happens on their site. We now just skip these results, as they're not useful for us.

see [diff](https://github.com/searxng/searxng/compare/9446fe91615cc0778ef5ccacca239379cf7b823d..d5f5a4a42724945d1854ab8e98528185f0224a1f)

return42:
I added the engine to the image category /[diff](https://github.com/searxng/searxng/compare/d5f5a4a42724945d1854ab8e98528185f0224a1f..9c49321ecd9596ee69844d36d5401ba931d00f27)

PR #2820, opened 2023-09-19
[feat] engine: implementation of chefkoch.de

Bnyro:
## What does this PR do?
* new engine: chefkoch.de (German results only)
* there's an undocumented API which is quite straight forward to use

* How should I survive without having some recipes? By eating frozen pizza all day?

* !chefkoch Pfannkuchen


return42 (approved):
> !chefkoch Pfannkuchen

How much I loved them, my mother's Pfannkuchen :)

Engine works like a charm :pancakes:

PR #2824, opened 2023-09-20
[fix] engine - google images error when no results

jazzzooo:
## What does this PR do?

Fixes `KeyError` when there are no results from Google Images.

One of the top currently reported errors on searx.space

`!goi ~^`

### DIFF
--- a/searx/engines/google_images.py
+++ b/searx/engines/google_images.py
@@ -95,7 +95,7 @@ def response(resp):
     json_start = resp.text.find('{"ischj":')
     json_data = loads(resp.text[json_start:])
 
-    for item in json_data["ischj"]["metadata"]:
+    for item in json_data["ischj"].get("metadata", []):
 
         result_item = {
             'url': item["result"]["referrer_url"],


Bnyro (approved):
Fixes the issue, thank you!

PS: It's been a bit difficult to figure out what changed honestly, the formatting changes make it a bit difficult to overview the changes (as in the bing PR).

searx/engines/google_images.py
```diff
@@ -95,8 +91,7 @@ def response(resp):
     json_start = resp.text.find('{"ischj":')
     json_data = loads(resp.text[json_start:])
 
-    for item in json_data["ischj"]["metadata"]:
-
+    for item in json_data["ischj"].get("metadata", []):
```
return42, line 98: @jazzzooo this is the only line that is needed to fix the bug / please revert unneeded reformatted code in this PR / thanks!

return42 (changes requested):

return42 (approved):
Thanks a lot for fixing this issue :+1:

PR #2827, opened 2023-09-22
[fix] harmonize default settings for caching of /static files

return42:
Caching files on the client side for more than a day can confuse the end user when updating static files[1].

Depending on the way of providing a SearXNG instance via HTTP, there are several ways to optimize the access to the /static files.  However, since we don't know what optimization an admin has provided for his static files, we should have moderate settings in the defaults that run robustly in a wide variety of installations.

In this sense, all caches on the client side should be cleared after one day at the latest.  So far the files were cached for one year on client side; as soon as changes are made to the static files (with the option `static_use_hash: true`) the old static files are kept for one year on the CLient side / which can also be evaluated as unnecessary caching.

[1] https://github.com/searxng/searxng/discussions/2821

----

Related:

- https://github.com/searxng/searxng/pull/932#issuecomment-1067039518

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -39,8 +39,8 @@ buffer-size = 8192
 add-header = Connection: close
 
 # uwsgi serves the static files
-# expires set to one year since there are hashes
 static-map = /static=/usr/local/searxng/searx/static
-static-expires = /* 31557600
+# expires set to one day
+static-expires = /* 86400
 static-gzip-all = True
 offload-threads = %k
--- a/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
+++ b/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
@@ -78,7 +78,7 @@ buffer-size = 8192
 #     static_use_hash: true
 #
 static-map = /static=${SEARXNG_STATIC}
-# expires set to one year since there are hashes
-static-expires = /* 31557600
+# expires set to one day
+static-expires = /* 86400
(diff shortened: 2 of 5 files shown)


Bnyro (approved):

return42:
@dalf @unixfox we have more and more users facing the CSS caching issue .. can we merge this PR?

unixfox:
Oops sorry for the lack of replies here.

~~How is that possible that the cache is an issue when changing the static files? Don't we invalidate the cache by changing the version number passed as a query string for each CSS and JS resources?~~

~~Looking at paulgo.io HTTP requests I see there is a hash for each static file:~~
image

~~When we update the static files, the hashes should change in order to invalidate the cache right? Or I am missing something?~~

unixfox:
Update: @dalf explained to me the situation, I wasn't aware that some apache servers are caching the files statically without touching the searxng application at all. This then doesn't work for hashes as a query string.

I'm okay with reducing the cache longevity for the moment, but this doesn't actually solve the whole issue, if a user change the static resources and no hash in the query string is used then the cache will not be instantly invalidated. The user will have to wait 1 day.

I propose to look for a solution for this issue, could be having the hash in the filename like GitHub seems to be doing and this would be compatible with any web server. Once we have implemented a solution for this issue we can then revert the cache longevity to the same values as before this PR.

Bonus part against this PR, reducing the cache longevity will actually hurt public instances where saving some bandwidth will make a difference as they receive a lot of traffic every day.

unixfox (approved):

dalf (approved):

PR #2829, opened 2023-09-22
Revert "[fix] engine - duckduckgo vqd edge-case"

return42:
This reverts commit 102502a4f09e78682cd4f030605be394bc33282c.

## What does this PR do?

Reverts:

- https://github.com/searxng/searxng/pull/2807

In practice it shows that ddg does not like the additional dummy request, the IP is put on a ban list and there is no response to a request at ddg ... https://github.com/searxng/searxng/pull/2807#issuecomment-1730896445

The problem occurs in all public instances that are receiving requests from bots / we should invest in a better defense against bots in the long run

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -57,13 +57,13 @@ url = 'https://lite.duckduckgo.com/lite/'
 # url_ping = 'https://duckduckgo.com/t/sl_l'
 
 time_range_dict = {'day': 'd', 'week': 'w', 'month': 'm', 'year': 'y'}
+form_data = {'v': 'l', 'api': 'd.js', 'o': 'json'}
 
 
 def cache_vqd(query, value):
-    """Caches a ``vqd`` token from a query, if token is None the cached value
-    is deleted.
+    """Caches a ``vqd`` value from a query.
 
-    The vqd token depends on the query string and is needed for the follow up
+    The vqd value depends on the query string and is needed for the follow up
     pages or the images loaded by a XMLHttpRequest:
 
     - DuckDuckGo Web: `https://links.duckduckgo.com/d.js?q=...&vqd=...`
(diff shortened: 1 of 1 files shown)


return42:
In my samples I see how the instances (searx.space) that have ddg enabled by default are degrading ... I will just merge this PR to avoid further degradation / hope thats OK for you all.

jazzzooo:
@return42 mm that's unfortunate. We still need to handle the error caused by getting a Captcha when fetching a vqd tho... Should I make a PR for that? Also don't forget you made changes to the form_data in this commit, that got reverted.

Do you think we should raise captcha exception when we get this captcha or just not return results? I think the latter is better as this captcha goes away on it's own when you search something else later.

return42:
> Also don't forget you made changes to the form_data in this commit, that got reverted.

Thanks for pointing out / I was in a hurry to get ddg engines working back :)

> Do you think we should raise captcha exception when we get this captcha or just not return results? I think the latter is better as this captcha goes away on it's own when you search something else later.

The issue is, that ddg does not send a captcha when we used this patch in practice .. ddg just never response .. the IP needs to cool down for some minutes before ddg continues to response.

Not shure how DDG bot protection works, but I assume when a IP does not answer a captcha n-times, ddg stop to response.

jazzzooo:
I will experiment some and see if i can figure out how it works. Do you think we should raise the default timeout for ddg?

return42:
> Do you think we should raise the default timeout for ddg?

Nut sure what you mean by "default timeout" .. if ddg do not response the engine runs into the timeout of the engine.

> I will experiment some and see if i can figure out how it works. 

Thanks! .. if you think you have a solution we can test the PR on my instance before we merge next time.

But I think we can't improve DDG a lot .. the main problem are the bots on the engines that pass the bot protection of SearXNG .. 

Anyway if you have any new insights, please let us know :+1:

jazzzooo:
Will do :)

I meant that we have 3 seconds timeout for all engines by default right? And then some engines like tineye have a higher timeout. So maybe setting duckduckgo to 4 seconds would help with not timing out when doing dummy requests? It would help in the case that it just takes long to do request+dummy+request again, not in the case that ddg doesn't respond on purpose.

return42:
> 4 seconds would help with not timing out when doing dummy requests? 

no .. may I was not clear; if DDG bans an IP, DDG does not response for 10 minutes or more 

When we test DDG local we never come to this point .. but on instances where you have hundreds or thousands request over a day and DDG sends you a CAPTCHA several times and this CAPTCHA is not answered but new requests are commming from the IP that does not answer the CAPTCHA .. then DDG thinks this IP must be a bot .. and does no longer response to this IP.

PR #2830, opened 2023-09-22
[fix] internet archive scholar: crash when there's no title

### LINKED ISSUE #2823: Bug: internetarchivescholar engine
Branch: master
Version: 2023.9.19+3ac7c40b6

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: KeyError
  * Percentage: 25
  * Parameters: `()`
  * File name: `searx/engines/internet_archive_scholar.py:59`
  * Function: `response`
  * Code: `'title': result['biblio']['title'],`

Bnyro:
## What does this PR do?
* we now shot the journal title if the paper itself doesn't have a title, so that there's no crash and the paper is still visible to the user

* !ias foo bar

## Additional notes
* I noticed Internet archive scholar also provides a thumbnail for some papers, but `paper.html` does not yet support that. Might be something to consider adding in the future.

## Related issues
closes #2823

### DIFF
--- a/searx/engines/internet_archive_scholar.py
+++ b/searx/engines/internet_archive_scholar.py
@@ -56,7 +56,7 @@ def response(resp):
             {
                 'template': 'paper.html',
                 'url': result['fulltext']['access_url'],
-                'title': result['biblio']['title'],
+                'title': result['biblio'].get('title') or result['biblio'].get('container_name'),
                 'content': html_to_text(content),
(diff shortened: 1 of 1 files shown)


return42 (approved):
can confirm / patch fixes the issue / thanks!

PR #2832, opened 2023-09-22
[mod] activate limiter & link_token method by default

return42:
## What does this PR do?

activate limiter & link_token method (aka CSS ping) by default

The SearXNG instances suffer from the massive bot requests

Install SearXNG into a LXC:

    $ sudo ./utils/lxc.sh build searxng-archlinux
    ...
    $ sudo ./utils/lxc.sh install suite archlinux
    ...
    environment /usr/local/searxng/searxng-src/utils/brand.env:
      ...
      SEARXNG_URL          : http://10.186.30.3/searxng
    ...
    $ sudo ./utils/lxc.sh cmd searxng-archlinux ./utils/searxng.sh install nginx
    
Limiter configuration should be installed:

    $ sudo ./utils/lxc.sh cmd searxng-archlinux ls -la /etc/searxng
    ...
    -rw-r--r--  1 searxng searxng  388 Sep 23 11:21 limiter.toml

Open http://10.186.30.3/searxng in your WEB browser and look a the CSS files:

grafik

----

Without Redis you will see an error message::

    ERROR   searx.plugins.limiter         : The limiter requires Redis

.. the instance will run, but the limiter can then not be used.

----

Related:

- https://github.com/searxng/searxng-docker/pull/182

### DIFF
--- a/docs/admin/index.rst
+++ b/docs/admin/index.rst
@@ -15,6 +15,7 @@ Administrator documentation
    installation-apache
    update-searxng
    answer-captcha
+   searx.botdetection
    api
    architecture
    plugins
--- a/docs/admin/searx.botdetection.rst
+++ b/docs/admin/searx.botdetection.rst
@@ -19,6 +19,8 @@ Bot Detection
   :members:
 
 
+.. _botdetection rate limit:
+
 Rate limit
 ==========
 
@@ -28,6 +30,8 @@ Rate limit
 .. automodule:: searx.botdetection.link_token
   :members:
 
+
+.. _botdetection probe headers:
 
 Probe HTTP headers
 ==================
--- a/searx/botdetection/__init__.py
+++ b/searx/botdetection/__init__.py
@@ -2,6 +2,23 @@
 # lint: pylint
 """.. _botdetection src:
 
+The :ref:`limiter <limiter src>` implements several methods to block bots:
+
+a. Analysis of the HTTP header in the request / can be easily bypassed.
+
+b. Block and pass lists in which IPs are listed / difficult to maintain, since
+   the IPs of bots are not all known and change over the time.
+
+c. Detection of bots based on the behavior of the requests and blocking and, if
+   necessary, unblocking of the IPs via a dynamically changeable IP block list.
+
+For dynamically changeable IP lists a Redis database is needed and for any kind
+of IP list the determination of the IP of the client is essential.  The IP of
+the client is determined via the X-Forwarded-For_ HTTP header
+
+.. _X-Forwarded-For:
+   https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+
 X-Forwarded-For
 ===============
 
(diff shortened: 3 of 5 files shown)


dalf:
with this PR, does an instance without Redis fail to start, or is there an only error message in the log?

unixfox:
I'm strongly against activating the limiter by default, mainly because there are people that use searxng locally and these people do not need the limiter, it's just personal usage they are not going to have bots attack.

Moreover, the advanced limiter (link_token) still actually limits terminal based browsers, like we discussed in https://github.com/searxng/searxng/pull/2357#issuecomment-1515216773. And **very probably** more browsers that we do not know about!

I would prefer to have a setting called `public_instance` or a setting under the limiter setting called `exposed_public` like:
```
limiter:
  exposed_public: true
```

This way we can require this setting for public instances and have default settings like (link_token enabled) that only impact them, not everyone else. It's the best tradeoff for everyone, while this may take some time to have this setting propagated against all the instances, once it is applied everywhere we can start experimenting more bot protection **WITHOUT** impacting the private usage.

I did talk about this setting here: https://github.com/searxng/searxng/discussions/2802

Do note that I'm not against enabling the advanced limiter (link_token) in https://github.com/searxng/searxng-docker which is targeted towards searxng instances exposed to the public.

return42:
> does this PR require Redis? or is there an error message but instance continue to work without the limiter?

Instance continues to work without redis / but on long term we should make redis a requirement.

return42:
> mainly because there are people that use searxng locally and these people do not need the limiter, 
> still actually limits terminal based browsers, 

They can opt-out / our public instances have issues ..

> while this may take some time to have this setting propagated against all the instances,

to help our public instance maintainers we should prioritize public instances in our default settings / maintainers of local instances or instances with special needs (terminal browser) can opt-out.

the maintainers of public instances trust that the default settings are optimal for public instances. ... if we deviate from this, then the maintainers always have to become active first if we want to establish new methods.

unixfox:
Users with private instances are usually the one that are the less technical, so they are the group that will surely have difficulties in "opting out". Please remember, we still want people to want to install searxng locally! Public instances are surely not the only way to use searxng!

While for public instances, maintainers are enough knowledge to enable some settings.

> the maintainers of public instances trust that the default settings are optimal for public instances. ... if we deviate from this, then the maintainers always have to become active first if we want to establish new methods.

That's the point of having ONE single setting that they have to activate, and we can target some specific changes just for them :).

Anyway, like I said in https://github.com/searxng/searxng/pull/2807#issuecomment-1724132175 and probably on other places, we are not going to enforce any setting that have a hard requirement on redis if we have not yet announced publicly the requirement of redis for searxng.

return42:
> I'm strongly against activating the limiter by default,

FYI: patch now enable only when using the installation scripts.

unixfox:
Who is the target for the install scripts? Does this allow to have a working install just locally?

If so then we shouldn't enable the limiter in the install scripts, we want it to only be activated on instances that will be exposed to the public.

searxng-docker is a project where you **HAVE TO** use a public domain for using it, hence the limiter being activated make sense.

Anyway, I have plans to add a new parameter in the settings.yml for specifying if it's a public instance or not. That's the ideal solution, once we get everyone on board we can start adding more protection by default just for the public instances like https://github.com/searxng/searxng/pull/2501

return42:
The script installation is an alternative to docker / some on searx.space use this installation method.

PR #2742, opened 2023-09-10
Improve the layout of the search page on mobile

ormai:
## What does this PR do?

Now when a result is selected the content of the card doesn't shift. The preferences button is now visible on the left of the search bar. I've also tweaked margins and paddings so that the content mantains some space against the edges of the viewport.

changes

## Why is this change important?

Improves user experience and the page looks nicer.

## How to test this PR locally?

I've tested with the browser responsive mode.

PR #2833, opened 2023-09-22
Fix mobile search view

Bnyro:
## What does this PR do?
* fix the mobile layout of the search view, which is located at the start page
* the issue has been introduced with https://github.com/searxng/searxng/pull/2742, not sure how we missed that

current master:
20230922_15h49m40s_grim

with this PR/fix:
20230922_16h01m40s_grim


return42 (approved):
> * the issue has been introduced with https://github.com/searxng/searxng/pull/2742, not sure how we missed that

was my fail / sorry and thanks a lot for your quick fix :rocket:

PR #2835, opened 2023-09-22
[feature] added a "copy" button in `search url` sidebar.

BernieHuang2008:
## What does this PR do?
add a `copy` button to the search-url sidebar so that users can copy the search-url with one click.

image

image

added a feature, user now can copy the search-url more easilly

1. `make run`
2. set `preferences -- Privacy -- HTTP Method` to `POST`
3. search anything, and open the `search url` sidebar.


searx/templates/simple/elements/search_url.html
```diff
@@ -1,6 +1,7 @@
 <div id="search_url" role="complementary" aria-labelledby="search_url-title">
   <details class="sidebar-collapsable">
     <summary class="title" id="search_url-title">{{ _('Search URL') }}</summary>
+    <button id="copy_url" type="submit">{{ _('Copy') }}</button>
```
dalf: If you change for
`<button id="copy_url" type="submit" data-copied-label="{{ _('Copied') }}">{{ _('Copy') }}</button>`

then `searxng.on('#copy_url', 'click', function () {` can use `data-copied-label`  and it can be translated.

(`data-copied-label` or whatever attribute name)

BernieHuang2008: thanks, i've improved it. Have a look now @dalf

BernieHuang2008:
Is this ok now? @dalf

BernieHuang2008:
ping @return42 , what do you think of it?

return42:
> ping @return42 , what do you think of it?

not sure .. final UI decisions are up to @dalf and @mrpaulblack ..

Bnyro:
In my opinion we should not change the color to green after the user copied the url, it just kinda looks out of place and doesn't fit the general theming of SearXNG.

It's sufficient to show the "Copied!" text message.

BernieHuang2008:
> In my opinion we should not change the color to green

How about gray, or other color? 
I just thought it can be a more obvious sign after copied ...

Bnyro:
I'd prefer to not introduce any new colors, so I'd prefer keeping it blue.

BernieHuang2008:
> I'd prefer to not introduce any new colors, so I'd prefer keeping it blue.

I've removed the green background. @Bnyro

BernieHuang2008:
BTW, have we removed the sidebar?
image

mrpaulblack:
No, this is the tablet layout, when the screen gets bigger, the sidebar moves to the right in desktop layout (see -> https://github.com/searxng/searxng/blob/526d5c7b308a314f7a118a5ecc3b47013e4a9ce9/searx/static/themes/simple/src/less/definitions.less#L260)

BernieHuang2008:
Not like that before, never mind.

Bnyro (approved):

mrpaulblack (changes requested):
Hi,
Thanks the PR looks really good.

A few nits:
* could you please change the copied button to be only displayed when JS is used in the browser (because it only works with JS)? You could for example put `display:none` on `#search_url button#copy_url` and remove it with JS on load

Everything apart from that nit LGTM :+1:

BernieHuang2008:
@mrpaulblack Sorry for my late reply. 
I've made some changes, please check again thanks.

mrpaulblack (approved):
LGTM :+1:

Thank you, and thank you for being patient with me taking a lot of time for the review :+1:

searx/static/themes/simple/src/js/main/results.js
```diff
@@ -16,6 +16,8 @@
           }
         ));
 
+    d.querySelector('#search_url button#copy_url').style.display = "block";
```
return42, line 19: @BernieHuang2008 @mrpaulblack this line is responsible for https://github.com/searxng/searxng/issues/3003 .. can you please have a look / thanks!

PR #2847, opened 2023-09-24
/config: expose limiter configuration

dalf:
## What does this PR do?

In the API endpoint `/config` expose three new values:
* `enabled` : is the limiter enabled ?
* `botdetection.ip_limit.link_token`: value from `limiter.toml`
* `botdetection.ip_lists.pass_searxng_org`: value from `limiter.toml`

Make sure public instances have enable the limiter and the client token.

* `make run`, check `http://127.0.0.1:8888/config`
* change the limiter configuration
* check again `.../config`

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -57,6 +57,7 @@ from searx import (
 )
 
 from searx import infopage
+from searx.botdetection import limiter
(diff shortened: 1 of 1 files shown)


unixfox (approved):
good small addition thank you

Bnyro (approved):

PR #2848, opened 2023-09-24
[mod] add new parameter called server.public_instance

unixfox:
## What does this PR do?

new parameter:
```
server:
  public_instance: false
```

For enabling by default advanced limiter functions.
And in the future allow us to add features just for the public instances.

This also makes redis mandatory for public instances.

As discussed in https://github.com/searxng/searxng/discussions/2802

This is an important setting that allows to not enable features not needed for local usage, it allows us to develop features specifically for public instances.

It's the best middle ground to have both parties "happy".

And it goes towards a "world" where we are going to make redis mandatory.

- Check if the limiter and the link_token limiter is working.
- Check if you can't start searxng without redis.
- Check if you can see the parameter public_instance in /config
- Check if the limiter and the link_token limiter are still working if the parameter `public_instance` is used.

## Author's checklist

- Check if the limiter and the link_token limiter is working.

image

- Check if you can't start searxng without redis.

image

- Check if you can see the parameter public_instance in /config

image

## Related issues

### DIFF
--- a/searx/__init__.py
+++ b/searx/__init__.py
@@ -104,3 +104,9 @@ if max_request_timeout is None:
     logger.info('max_request_timeout=%s', repr(max_request_timeout))
 else:
     logger.info('max_request_timeout=%i second(s)', max_request_timeout)
+
+if settings['server']['public_instance']:
+    logger.warning(
+        "Be aware you have activated features intended only for public instances. "
+        + "This force the usage of the bot limiter and link_token plugins."
+    )
--- a/searx/botdetection/ip_limit.py
+++ b/searx/botdetection/ip_limit.py
@@ -47,6 +47,7 @@ from ipaddress import (
 import flask
 import werkzeug
 from searx.tools import config
+from searx import settings
 
 from searx import redisdb
 from searx.redislib import incr_sliding_window, drop_counter
@@ -109,7 +110,7 @@ def filter_request(
         if c > API_MAX:
             return too_many_requests(network, "too many request in API_WINDOW")
 
-    if cfg['botdetection.ip_limit.link_token']:
+    if settings['server']['public_instance'] or cfg['botdetection.ip_limit.link_token']:
 
         suspicious = link_token.is_suspicious(network, request, True)
 
--- a/searx/plugins/limiter.py
+++ b/searx/plugins/limiter.py
@@ -3,6 +3,7 @@
 # pyright: basic
 """see :ref:`limiter src`"""
 
+import sys
 import flask
 
 from searx import redisdb
@@ -23,10 +24,15 @@ def pre_request():
 
 
 def init(app: flask.Flask, settings) -> bool:
-    if not settings['server']['limiter']:
+    if not settings['server']['limiter'] and not settings['server']['public_instance']:
         return False
     if not redisdb.client():
-        logger.error("The limiter requires Redis")
+        logger.error(
+            "The limiter requires Redis, please consult the documentation: "
+            + "https://docs.searxng.org/admin/searx.botdetection.html#limiter"
+        )
+        if settings['server']['public_instance']:
+            sys.exit(1)
         return False
     app.before_request(pre_request)
     return True
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -74,6 +74,7 @@ server:
   # by ${SEARXNG_URL}.
   base_url: false  # "http://example.com/location"
   limiter: false  # rate limit the number of request on the instance, block some bots
+  public_instance: false  # enable features designed only for public instances
 
   # If your instance owns a /etc/searxng/settings.yml file, then set the following
   # values there.
@@ -95,7 +96,7 @@ server:
 
 redis:
   # URL to connect redis database. Is overwritten by ${SEARXNG_REDIS_URL}.
-  # https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url
+  # https://docs.searxng.org/admin/settings/settings_redis.html#settings-redis
   url: false
 
 ui:
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -174,6 +174,7 @@ SCHEMA = {
         'port': SettingsValue((int, str), 8888, 'SEARXNG_PORT'),
         'bind_address': SettingsValue(str, '127.0.0.1', 'SEARXNG_BIND_ADDRESS'),
         'limiter': SettingsValue(bool, False),
+        'public_instance': SettingsValue(bool, False),
         'secret_key': SettingsValue(str, environ_name='SEARXNG_SECRET'),
         'base_url': SettingsValue((False, str), False, 'SEARXNG_BASE_URL'),
         'image_proxy': SettingsValue(bool, False),
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1286,6 +1286,7 @@ def config():
             },
             'doi_resolvers': list(settings['doi_resolvers'].keys()),
             'default_doi_resolver': settings['default_doi_resolver'],
+            'public_instance': settings['server']['public_instance'],
         }
     )
 


unixfox:
Official documentation for this new change: https://gist.github.com/unixfox/4e22730769540fe5b9f1943a86439761

With details how to do it for each type of installation.

dalf:
IMO, we should have dedicated code to deal with this new option.

So instead of 
```
if settings['server']['public_instance'] or cfg['botdetection.ip_limit.link_token']:
```
we keep
```python
if cfg['botdetection.ip_limit.link_token
```

And we add a dedicated function in `searx/settings_defaults.py` :
```python
...

from searx.botdetection import limiter
from searx import settings

...

def apply_public_instance_setting(settings, limiter_cfg):
    if not settings['server']['public_instance']:
        return
    settings['server']['limiter'] = True
    limiter_cfg['botdetection.ip_limit.link_token'] = True
    
    
def settings_set_defaults(settings, limiter_cfg):
    apply_schema(settings, SCHEMA, [])
    apply_public_instance_setting(settings, limiter_cfg)
    return settings
```

In `searx/__init__.py`, replace

```python
if settings is not None:
    settings = settings_set_defaults(settings)
```

by

```python

from searx.botdetection import limiter
....

if settings is not None:
    settings = settings_set_defaults(settings, limiter.get_cfg())
```

So anyone can see clearly what this new setting does.
If we need to update the behavior of this settings, it is one place (and not spread in different locations)

[EDIT] Fix various mistakes in the suggested code

dalf (approved):
As discuss with @unixfox , this PR is good enough to provide the settings to everyone.

My suggestion does not work : `limiter_cfg['botdetection.ip_limit.link_token'] = True` is not possible because the result from [pytomlpp.loads](https://pypi.org/project/pytomlpp/) is read only.

Bnyro:
I don't think it's generally a bad idea to extend the amount of instances that use Redis by that, however:
- This setting seems a bit unnecessary to me. Can't we just instead consider enabling the limiter and link_token by default in https://github.com/searxng/searxng/pull/2832 as well as in the Docker image?
- If we have multiple different options that impact the same feature, it's not obvious to admins why changing the value of `limiter` or `link_token` sometimes does not work.

In my opinion, we should make changes to the documentation to make sure every public instance enables these settings - instead of making the code more complex by this change. That's why I've been pretty much in favor of Markus' recent PR concerning making the limiter enabled by default.

EDIT: Linked the wrong url above in my first bullet point: I meant to link https://github.com/searxng/searxng/blob/master/utils/templates/etc/searxng/settings.yml.

return42:
> For enabling by default advanced limiter functions. And in the future allow us to add features just for the public instances.

I fully agree with the https://github.com/searxng/searxng/pull/2848#issuecomment-1732642252, its vague to the admin what we change when and what does it means to its dedicated setup.

unixfox:
> This setting seems a bit unnecessary to me. Can't we just instead consider enabling the limiter and link_token by default in https://github.com/searxng/searxng/pull/2832 as well as in the Docker image?

Check my comment: https://github.com/searxng/searxng/pull/2832#issuecomment-1731127056

It is not a good idea to have the limiter for private instances, people do not need the limiter for a local usage. It would be too much harm for them, see the comment in the link above for why.

That's why there should be a setting to distinguish a private instance from a public instance. Let's not forget that the limiter started as the idea of implementing filtron into the core of SearXNG: https://github.com/asciimoo/filtron. **This tool was only recommended to public instances.**

Moreover, the limiter actually needs some specific criteria for it to work **properly** or else you are going to block yourself:
- it needs various HTTP headers sent from the reverse proxy: https://github.com/searxng/searxng/blob/master/searx/botdetection/_helpers.py#L70

I have been managing the community side of SearX and SearXNG for some years now and I can tell you that having the limiter for everyone will introduce a lot of angry people because they do not understand why their own software is turning against themselves.

One major example is when people run searxng-docker (limiter enabled by default) behind a cloudflare tunnel, this has caused many messages in the matrix room or on other places, proof:
- https://github.com/search?q=repo%3Asearxng%2Fsearxng+cloudflare&type=discussions
- https://old.reddit.com/r/Searx/search?q=cloudflare&restrict_sr=on&include_over_18=on
- public matrix room of SearXNG

----------

> If we have multiple different options that impact the same feature, it's not obvious to admins why changing the value of limiter or link_token sometimes does not work.
> In my opinion, we should make changes to the documentation to make sure every public instance enables these settings - instead of making the code more complex by this change. That's why I've been pretty much in favor of Markus' recent PR concerning making the limiter enabled by default.

The point of this setting is that it is simple for public instances maintainers, they just have to enable one setting and have redis configured, then we do the rest! Have you read the new Matrix room for public instances maintainer recently?

We have maintainers that do not understand much about searxng, do you think they will be able to keep up with every new changes we deploy for protecting the public instances? No they will not and that's where we are going to have a list of instances well protected and instances not protected enough.

I had a very very hard time to contact ALL the public instances maintainers, some DO NOT even check github very often. It's easy to say just look at the docs, but people have things to do outside of the computer, they don't always have unlimited free time. They are most probably running 15 other public services and searxng is just one of them, see example: https://projectsegfau.lt/instances

If we start to have every 6 months new parameters to enable for protecting our public instances, then I can assure you that people are not going to follow our development pace, and we will see people just quitting altogether the hosting of a public instance.

Let's make the life easier for public instances maintainers and let's improve the quality of our list of public instances!

And also make MY life easier because I do spend a lot of time guiding the public instances maintainers (and also the private instances users). You should try to manage them @Bnyro and @return42, just to understand the effort that you have to put into.

return42:
> Let's make the life easier for public instances maintainers and let's improve the quality of our list of public instances!

Therefore I suggest to activate the limiter in the defaults.

We have to decide if we prefer private or public instances in the defaults.

I understand all your arguments but come to a different conclusion: IMO nearly all of your arguments speak in favor of preferring public instances.

>  I can tell you that having the limiter for everyone will introduce a lot of angry people because they do not understand why their own software is turning against themselves.

IMO admins will be angry when a dedicated setup is overwritten by a update of the instance:

>> For enabling by default advanced limiter functions. And in the future allow us to add features just for the public instances.

>  its vague to the admin what we change when and what does it means to its dedicated setup.

TL;DR;

From the same arguments we come to different conclusions, we can discuss this for a long time without agreeing, I am against implementing something in the code that overrides the settings of the admin.  If you think that's a good solution, then implement it... I can live with it, but I will never like it.

jazzzooo:
I actually think a unified parameter `public_instance` would be good, it would make the configuration easier for new users. We could clearly state what settings it enables in the comment "enable features designed only for public instances" -> "enables features: x, y, z needed for public instances".

But also like Markus points out, we should not roll out a setting that overwrites existing configs. To avoid breaking anyone's setup, we can make the parameter optional now, and only have an effect when explicitly configured, so no default true or false, as either would harm one set of users. Then over time we can introduce a warning that is should be set, and eventually make it mandatory. Once we are confident this setting is set for most users, then we can start relying on it more.

To that extent, I think something along the lines of dalf's original suggestion would be appropriate. Having the `public_instance` setting only affect other settings and not anything in the codebase for now.

Buuut I'm not as familiar as you guys with the userbase and project so lmk if I'm wrong.

unixfox:
> we can make the parameter optional now, and only have an effect when explicitly configured

That's actually what does this PR, there is nothing forced here ^^': https://github.com/searxng/searxng/pull/2848/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR77

> But also like Markus points out, we should not roll out a setting that overwrites existing configs.

Markus exactly wants the opposite, he wants to overwrite the limiter config and so make it by default. This PR here is for exactly making some parameters optional and only for public instances maintainers to enable.

The only setting that will be overridden when turning `public_instance` parameter to true are the limiter settings, but that's exactly what is needed by public instances maintainers and what we are going to make mandatory so that public instances are more protected.

Merging this PR will not break ANY existing setups.

> I actually think a unified parameter `public_instance` would be good, it would make the configuration easier for new users. We could clearly state what settings it enables in the comment "enable features designed only for public instances" -> "enables features: x, y, z needed for public instances".

I'm ok with adding a warning during the startup in case someone enable that optional setting for unknown reasons, but this could happen with any other settings in the settings.yml.

unixfox:
@return42 

> https://github.com/searxng/searxng/pull/2848#issuecomment-1732683214

Maybe it could be that you have not spend enough time guiding the various different users of our community (and the previous community at SearX), and thus you can't project the real consequences of doing this change. You only look at your current way of doing, and you think that it's pretty easy to handle all the new changes done in SearXNG.

I propose two ways of handling this situation:

The first one is that we activate the limiter and the link token by default. I let you handle the announcement of asking **every** public instances maintainers to configure redis and making sure that they have correctly configured the limiter before the end of 2023.

Also, you will be tasked to handle every question posts created on Reddit, reply to GitHub discussions and also reply on Matrix room (public instances and public one). I do not want empty replies on GitHub discussions for installation issues like it's already the case: https://github.com/searxng/searxng/discussions

Especially a huge effort will have to be made when we will make redis mandatory and thus the limiter will run by default for everyone.

Or the other way, you trust me on this and we merge this change. I'm tasked to announced to every public instances maintainers to configure redis and making sure that they have correctly configured the new parameter.

I've already prepared a document for that: https://gist.github.com/unixfox/4e22730769540fe5b9f1943a86439761

Which way do you prefer? I think that's the best balance, the one that implement the change have to deal with outcomes of this change.

return42:
I'm not a fan of this PR, but as I said above, if you think it is a good idea, go for it.

dalf (approved):

PR #2849, opened 2023-09-24
[fix] engine - kickass update url, fix parsing, use multiple mirrors

### LINKED ISSUE #2842: bug kickass engine timeout
the domain used for the kickass engine doesn't work anymore: https://github.com/searxng/searxng/blob/master/searx/engines/kickass.py#L26 so we get a timeout.

We should try to use another domain at https://thekickasstorrents.to and fix the xpath parsing.

jazzzooo:
## What does this PR do?

- Adds 5 working kickass mirrors
- Fixes the xpaths and makes them more robust
- Unfortunately, magnet, download links and file counts aren't shown in the result anymore. Code for that is removed.

So kickass is dead:
https://reddit.utsav2.dev/r/kickasstorrents/comments/mue44z/kat_is_dead/
https://thehackernews.com/2016/12/kickass-torrents-site.html
It has gone down, new clones by same staff have appeared and gone down again. The current clone network mirrors are listed on https://thekickasstorrents.to/ .

### Suggestions:
- Kickass can probably be enabled by default now.
- Test and see if the 4 second timeout can be decreased. I'm getting 0.2s P95.
- Maybe the shortcut should be !kat, as is the official abbreviation.

`!kc 12 angry men`

## Related issues

Closes #2842

### DIFF
--- a/searx/engines/kickass.py
+++ b/searx/engines/kickass.py
@@ -1,16 +1,23 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Kickass Torrent (Videos, Music, Files)
-"""
+# lint: pylint
+"""Kickass Torrent (Videos, Music, Files)"""
+
+import random
+from operator import itemgetter
+from urllib.parse import quote
 
 from lxml import html
-from operator import itemgetter
-from urllib.parse import quote, urljoin
-from searx.utils import extract_text, get_torrent_size, convert_str_to_int
+from searx.utils import (
+    eval_xpath,
+    eval_xpath_getindex,
+    eval_xpath_list,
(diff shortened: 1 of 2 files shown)


return42:
@jazzzooo nice work, thanks a lot :+1: .. I just add SearXNG's boilerplate stuff and enabled by default --> [diff](https://github.com/searxng/searxng/compare/881ac5b45854e0bf26f1aea3ebeb142ebd7944cb..9a6cf3f01b42e90354b5427b252f629892d815cb)

> Kickass can probably be enabled by default now.

In category `files` a timeout of 4sec is a very good value .. we have engines enabled in this category with higher timeouts.

PR #2850, opened 2023-09-24
[mod] warn to info for failed ping limiter

unixfox:
## What does this PR do?

Change the log level from warning to info for `searx.botdetection.link_token: missing ping`

This spam a lot in the public instances, and it shouldn't be a warning as it means the limiter is doing its job.


return42:
> This spam a lot in the public instances, and it shouldn't be a warning as it means the limiter is doing its job.

If you want, we can change to info .. my intention was to log warnings & errors to count IPs that are bots.

Thats also why I log to error when a bot *is placed on the block list*:

https://github.com/searxng/searxng/blob/19383921d7e7153a34a29a37c252197772cfd1c8/searx/botdetection/ip_limit.py#L125-L127

dalf (approved):

PR #2852, opened 2023-09-25
[feat] engine: implementation of pinterest

### LINKED ISSUE #1234: Pinterest
**Working URL to the engine**

https://www.pinterest.com/search

Pinterest has great image search

**Features of this engine**

Better image results

Images

**Additional context**

Bnyro:
## What does this PR do?
* add a new engine, pinterest via xpath, as its own module

* !pin bird

## Related issues
closes #1234

### DIFF
--- a/searx/engines/pinterest.py
+++ b/searx/engines/pinterest.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Pinterest (images)
+"""
+
+from json import dumps
+
+about = {
+    "website": 'https://www.pinterest.com/',
+    "wikidata_id": 'Q255381',
+    "official_api_documentation": 'https://developers.pinterest.com/docs/api/v5/',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['images']
+paging = True
+
+base_url = 'https://www.pinterest.com'
+
+
+def request(query, params):
+    args = {
+        'options': {
+            'query': query,
+            'bookmarks': [params['engine_data'].get('bookmark', '')],
+        },
+        'context': {},
+    }
+    params['url'] = f"{base_url}/resource/BaseSearchResource/get/?data={dumps(args)}"
+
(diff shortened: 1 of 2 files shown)


return42 (approved):
Thanks!  This image engine is very fast :rocket:  / should we enable by default?

I squashed some minors in / see [diff](https://github.com/searxng/searxng/compare/9a98b625e538501b690ea811331606aa8a481ff1..0e52e192823174e1330ea66c2287ec87ffb86980)

- `result.get('description')` is most often an empty string
- added `source` to result item
- I renamed `json` to `json_resp` to avoid name collisions / for debugging purpose sometimes I import json:

```diff
diff --git a/searx/engines/pinterest.py b/searx/engines/pinterest.py
index 4ca316698..f7be71a39 100644
--- a/searx/engines/pinterest.py
+++ b/searx/engines/pinterest.py
@@ -37,6 +37,9 @@ def response(resp):
     results = []
 
     json_resp = resp.json()
+    with open("response.json", 'w') as f:
+        import json
+        f.write(json.dumps(json_resp))
 
     results.append(
         {
```

Bnyro:
> Thanks! This image engine is very fast 🚀 / should we enable by default?
> 
I think the engine fulfills all requirements to become enabled by default, I just disabled it because 80% of the engines in the images category are already enabled by default, and because it doesn't have real pagination.
But enabling it by default would be fine by me!
> I squashed some minors in / see [diff](https://github.com/searxng/searxng/compare/9a98b625e538501b690ea811331606aa8a481ff1..0e52e192823174e1330ea66c2287ec87ffb86980)
> `result.get('description')` is most often an empty string
> 
Yeah, I don't think a description is really necessary for an image though.
> added `source` to result item
> 
:+1:
>  I renamed `json` to `json_resp` to avoid name collisions / for debugging purpose sometimes I import json:
> 
:+1:

return42:
> Yeah, I don't think a description is really necessary for an image though.

Why not, when we have one in 

```python
'content': (result.get('rich_summary') or {}).get('display_description') or "",
```

we use it in the gallery:

image

PR #2826, opened 2023-09-21
[fix] update user agent

jazzzooo:
## What does this PR do?

Fixes the User-Agent to be a valid one. Currently the geckoversion rv value is frozen as 109.

## Why is this change important?

As dalf found, [firefox froze their rv value at 109](https://bugzilla.mozilla.org/show_bug.cgi?id=1805967). This made our User-Agent invalid and led to us being detected as a bot by Bing. This could also have negative effects for other engines.

## How to test this PR locally?

Try some searches, make sure no engine is broken by this. Shouldn't be the case tho.

PR #2853, opened 2023-09-26
Fix update engine traits (brave & annas archive)

return42:
## What does this PR do?

fix the fetch traits function of:

- brave
- annas archive

As @dalf pointed out, the gh workflow to fetch the traits from the engines did fails since month --> https://github.com/searxng/searxng/pull/2826#discussion_r1336805362

    make data.traits

## Related issues

- https://github.com/searxng/searxng/pull/2826#discussion_r1336805362


searx/engines/brave.py
```diff
@@ -397,19 +396,26 @@ def fetch_traits(engine_traits: EngineTraits):
 
     # search regions of brave
 
-    engine_traits.all_locale = 'all'
+    resp = get('https://cdn.search.brave.com/serp/v2/_app/immutable/chunks/parameters.734c106a.js', headers=headers)
```
Bnyro, line 399: This could probably break really quick again, but I don't have a better idea unfortunately

Bnyro (approved):

searx/engines/brave.py
return42: Yes, I have checked: the URL is not a href in a `<link>` of the root html from https://search.brave.com/settings .. this doc loads some JS and one of the loaded JS scripts extends the `<link>` list in the root document to load the `parameters.734c106a.js'` .. very complicated.  Instead of having two hard to track indirectness I decided to read directly from `parameters.734c106a.js` .. and see when it breaks :o

return42:
@dalf on my local instance `make data.traits` did work / do you have a tip for me, how I can test in CI? .. I triggered the job in the CI https://github.com/return42/searxng/actions/runs/6322049251 but the job was skipped.

If it is to complicated, can we merge and test in SearXNG's master?

dalf:
You need to add a temporary commit to remove this line:
https://github.com/return42/searxng/blob/3c3ce81f314224bd31647ac01a417f07562c924f/.github/workflows/data-update.yml#L11

Since there is no way to debug the GitHub actions, prepare yourself to `git add .; git commit --amend; git push --force`, and once okay clean the git log of your branch.

> If it is to complicated, can we merge and test in SearXNG's master?

It will create PR in searxng/searxng : if there are broken, it will be a mess. It is better to test in a fork.

dalf (approved):
~~It runs fine in your fork.~~
Once it runs fine in your fork, I let you remove the WIP commit and self merge.

return42:
> You need to add a temporary commit to remove this line ..

Yeah, that did the trick .. PR was created in my clone: https://github.com/return42/searxng/pull/254

> ~~It runs fine in your fork.~~

But I also observed, what I have also seen local on my desktop, [502 Bad Gateway](https://github.com/return42/searxng/actions/runs/6322307330/job/17167683513#step:6:67) from zlibrary-global.se: 

```
httpx.HTTPStatusError: Server error '520 ' for url 'https://zlibrary-global.se/'
```

Seems to me zlibrary-global.se is sometimes unstable .. since this issue is not related to this PR, I'm going to merge this PR.

PR #2854, opened 2023-09-27
[feat] engine: implementation of radio-browser.info

Bnyro:
## What does this PR do?
* add a new foss engine for radio search, https://www.radio-browser.info
* we're currently using the 'videos.html' template because it allows you to directly start playing the radio in the search results page

* we don't have any other search engine for radios yet, and radio-browser.info is the most beloved one
* because @return42 thinks it is a "Nette Sache"

* !rb nrj :de
* !rb i love radio :all
* !rb swr3 :de-DE


searx/data/engine_traits.json
```diff
@@ -4146,6 +4619,7 @@
         "km",
         "kn",
         "ko",
+        "koi",
         "krc",
```
return42: Its not a part of this PR

searx/sxng_locales.py
```diff
@@ -29,13 +29,16 @@
     ('en-AU', 'English', 'Australia', 'English', '\U0001f1e6\U0001f1fa'),
     ('en-CA', 'English', 'Canada', 'English', '\U0001f1e8\U0001f1e6'),
     ('en-GB', 'English', 'United Kingdom', 'English', '\U0001f1ec\U0001f1e7'),
+    ('en-HK', 'English', 'Hong Kong SAR China', 'English', '\U0001f1ed\U0001f1f0'),
     ('en-IE', 'English', 'Ireland', 'English', '\U0001f1ee\U0001f1ea'),
     ('en-IN', 'English', 'India', 'English', '\U0001f1ee\U0001f1f3'),
     ('en-MY', 'English', 'Malaysia', 'English', '\U0001f1f2\U0001f1fe'),
```
return42: Do not commit modifications in this file .. it will change the language menu and I think we can update the lang menu when its needed (not a part of this PR).

searx/engines/radio_browser.py
```diff
@@ -0,0 +1,111 @@
...
+            continue
+
+        conflict = engine_traits.languages.get(sxng_tag)
+        if conflict:
+            continue
+
+        engine_traits.languages[sxng_tag] = lang['name']
+
+        for region in country_list:
+            try:
```
return42: Indentation issue / this for loop is in the `for lang in language_list:` loop.

return42 (changes requested):
seems not to work, on every query I got:

```
  File "/share/searxng/local/py3/lib/python3.10/site-packages/httpx/_models.py", line 749, in raise_for_status
httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://de1.api.radio-browser.info//json/stations/search?name=nrj&order=votes&offset=0&limit=10&hidebroken=true&reverse=true&language=german&countrycode=DE'
```

searx/engines/radio_browser.py
Bnyro: Yes, that's intended. All these countries are supported by radio-browser.info independent of the language used, hence we check if babel knows language + region and if it does, add it to the engine traits. There's no list of countries mapped to their language.

searx/data/engine_traits.json
Bnyro: How can I only generate traits for a specific engine? I've used `make data.traits`, but it does a bunch of changes I don't want it to.

return42: Its not possible :) .. I add only those hunks to the git commit I want in the commit and revert the rest .. Its not a problem, I can do it in my review if you want.

Bnyro: Should be fixed now.

searx/sxng_locales.py
Bnyro: :+1:

Bnyro:
> seems not to work, on every query I got:
> 
> ```
>   File "/share/searxng/local/py3/lib/python3.10/site-packages/httpx/_models.py", line 749, in raise_for_status
>     raise HTTPStatusError(message, request=request, response=self)
> httpx.HTTPStatusError: Client error '404 Not Found' for url 'https://de1.api.radio-browser.info//json/stations/search?name=nrj&order=votes&offset=0&limit=10&hidebroken=true&reverse=true&language=german&countrycode=DE'
> ```
>
That's fixed now :+1:

Though I don't understand why it worked when I tested it locally, I think I switched it to use the German instance of the API in the end, and probably forgot to remove the trailing `/`.

return42:
> because @return42 thinks it is a "Nette Sache"

@Bnyro its really a "Nette Sache" :) .. Thanks a lot .. :headphones: .. I implemented some improvements and committed on top of your branch / could  you have a final look.

Bnyro:
How we handle `countrycodes` looks way better now, I didn't know we can use custom properties for saving country codes :)

I couldn't find any issues with the implementation/new changes, neither from testing nor from looking at the code, so from that perspective: looks good to me!

return42 (approved):
> I didn't know we can use custom properties for saving country codes :)

Thats why I like to collaborate in branches, where we show our solution others and we can learn from each other.

> looks good to me!

OK, then I will merge this PR / thanks for the review :smile:

PR #2855, opened 2023-09-27
[mod] settings.yml: remove plugin settings for plugins that don't exist anymore

Bnyro:
We have been replacing these plugins by normal UI settings recently, but forgot to remove the existing plugin settings in settings.yml where they have been enabled.

### DIFF
--- a/docs/admin/api.rst
+++ b/docs/admin/api.rst
@@ -69,10 +69,6 @@ Sample response
        {
          "enabled": true,
          "name": "HTTPS rewrite"
-       },
-       {
-         "enabled": false,
-         "name": "Vim-like hotkeys"
        }
      ],
      "safe_search": 0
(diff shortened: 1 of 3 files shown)


dalf:
I have pushed an additional commit to remove the reference to the VIM plugin `docs/admin/api.rst`. This page need some update BTW.

---

Side out of topic note: everything is setup to be able to reference plugin by their module name
https://github.com/searxng/searxng/blob/81b72916132376d964fc91a425faff8a3b2cdfb5/searx/plugins/__init__.py#L167

```yaml
enabled_plugins:
  - searx.plugins.hash_plugin
  - searx.plugins.self_info
  - searx.plugins.tracker_url_remover
  - searx.plugins.ahmia_filter
  # - searx.plugins.hostname_replace  # see hostname_replace configuration below
  # - searx.plugins.oa_doi_rewrite
  # - mycustomcode.myplugin
```

dalf (approved):

dalf:
In `utils/templates/etc/searxng/settings.yml`, I've updated the commit to remove a reference to 'Infinite scroll' since this is preference and not any more a plugin.

The PR is good to merge IMO.

PR #2856, opened 2023-09-27
[fix] categories.html: can't select social media category using search on category select

Bnyro:
## What does this PR do?
* fixes a regression from #2740 that we've overseen

* make run
* search anything and click on the "social media" category

### DIFF
--- a/searx/templates/simple/categories.html
+++ b/searx/templates/simple/categories.html
@@ -26,7 +26,7 @@
             {%- if display_tooltip %}<div class="help">{{ _('Click on the magnifier to perform search') }}</div>{% endif -%}
         {%- else -%}
             {%- for category in categories_as_tabs -%}{{- '\n' -}}
-                <button type="submit" name="category_{{ category|replace(' ', '_') }}" class="category category_button {% if category in selected_categories %}selected{% endif %}">
+                <button type="submit" name="category_{{ category }}" class="category category_button {% if category in selected_categories %}selected{% endif %}">
                     {{- icon_big(category_icons[category]) if category in category_icons else icon_big('globe-outline') -}}
                     <div class="category_name">{{- _(category) -}}</div>{{- '' -}}
                 </button>{{- '' -}}


searx/templates/simple/categories.html
```diff
@@ -26,7 +26,7 @@
             {%- if display_tooltip %}<div class="help">{{ _('Click on the magnifier to perform search') }}</div>{% endif -%}
         {%- else -%}
             {%- for category in categories_as_tabs -%}{{- '\n' -}}
-                <button type="submit" name="category_{{ category|replace(' ', '_') }}" class="category category_button {% if category in selected_categories %}selected{% endif %}">
+                <button type="submit" name="category_{{ category }}" class="category category_button {% if category in selected_categories %}selected{% endif %}">
```
dalf, line 29: The line 19 has exactly the same `|replace(' ', '_')` but it is left untouched, is it intended?

Bnyro: Yes, I don't really understand why but if you disable search on category select, it still works fine when selecting social media.

return42: > The line 19 has exactly the same `|replace(' ', '_')` but it is left untouched, is it intended?

In line 19 the *replace* is done on the `id`, but not on the `name` attribute:

`id="checkbox_{{ category|replace(' ', '_') }}" name="category_{{ category }}`

The `name` attribute is used for the preferences .. and if I'm not wrong, in the preferences code we use the category *name* without replacing the white-spaces:

https://github.com/searxng/searxng/blob/e37d775fa2da6de818e0dc73842bbb0bddc615ef/searx/preferences.py#L295-L299

For instance `social media` from here:

https://github.com/searxng/searxng/blob/e37d775fa2da6de818e0dc73842bbb0bddc615ef/searx/settings.yml#L285-L295

return42 (approved):
can confirm, this PR fixes the issue :+1:

PR #2857, opened 2023-09-27
[feat] engine: implementation of mastodon

### LINKED ISSUE #792: Mastodon & Pixelfed
**Working URL to the engine**

I don't know yet. I'll find out later and update.

To encourage libre and decentralized www.

**Features of this engine**

Photo sharing platform.

images, social media

**Additional context**

https://pixelfed.social/
https://pixelfed.org/

Bnyro:
## What does this PR do?
* see the documentation at the top of `mastodon.py`

* we already support lemmy, hence it makes sense to support mastodon too (also as we finally removed Twitter support recently)

* !mau torvalds
* !mah linux

## Related issues
closes #792

### DIFF
--- a/searx/engines/mastodon.py
+++ b/searx/engines/mastodon.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Mastodon_ is an open source alternative to large social media platforms like
+Twitter/X, Facebook, ...
+
+Since it's federated and self-hostable, there's a large amount of available
+instances, which can be chosen instead by modifying ``base_url``.
+
+We use their official API_ for searching, but unfortunately, their Search API_
+forbids pagination without OAuth.
+
+That's why we use tootfinder.ch for finding posts, which doesn't support searching
+for users, accounts or other types of content on Mastodon however.
+
+.. _Mastodon: https://mastodon.social
+.. _API: https://docs.joinmastodon.org/api/
+
+"""
+
+from urllib.parse import urlencode
+from datetime import datetime
+
+about = {
+    "website": 'https://joinmastodon.org/',
+    "wikidata_id": 'Q27986619',
+    "official_api_documentation": 'https://docs.joinmastodon.org/api/',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+categories = ['social media']
+
+base_url = "https://mastodon.social"
+mastodon_type = "accounts"
+
+# https://github.com/searxng/searxng/pull/2857#issuecomment-1741713999
+page_size = 40
+
+
+def request(query, params):
+    args = {
+        'q': query,
+        'resolve': 'false',
+        'type': mastodon_type,
+        'limit': page_size,
+    }
+    params['url'] = f"{base_url}/api/v2/search?{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    results = []
+
+    json = resp.json()
+
+    for result in json[mastodon_type]:
+        if mastodon_type == "accounts":
+            results.append(
+                {
+                    'url': result['uri'],
+                    'title': result['username'] + f" ({result['followers_count']} followers)",
+                    'content': result['note'],
+                    'thumbnail': result.get('avatar'),
+                    'publishedDate': datetime.strptime(result['created_at'][:10], "%Y-%m-%d"),
+                }
+            )
+        elif mastodon_type == "hashtags":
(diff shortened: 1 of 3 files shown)


searx/engines/mastodon.py
```diff
@@ -0,0 +1,75 @@
...
+    return params
+
+
+def response(resp):
+    results = []
+
+    json = resp.json()
+
+    for result in json[mastodon_type]:
+        if mastodon_type == "accounts":
```
allendema, line 57: NP: Consider using .lower() for mastodon_types so the settings.yml can be flexible.

searx/engines/mastodon.py
```diff
@@ -0,0 +1,75 @@
...
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+categories = ['social media']
+
+base_url = "https://mastodon.social"
+mastodon_type = "accounts"
+
+number_of_results = 10
```
allendema: We should consider renaming this var because it is also used for getting the total number of results.

https://github.com/search?q=repo%3Asearxng%2Fsearxng%20number_of_results&type=code

Bnyro: That's true, we should probably rename that to "page_size" for all engines (in a different PR).

return42: > That's true, we should probably rename that to "page_size" for all engines (in a different PR).

Thats what I also have thought often :) .. but we shouldn't:  ``number_of_results`` could be used in the `/etc/searxng/settings.yml` .. we don't know it .. currently we have monkey patching for all the settings of an engine .. may be one day we can establish a config scheme .. that scales .. and is downward compatible :)

searx/engines/mastodon.py
Bnyro: Most settings are only working when being configured as lowercase, we would need to lower every occurrence of mastodon_type which seems a bit unnecessary to me, especially since the two working types are already preconfigured in "settings.yml"

searx/engines/mastodon.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def request(query, params):
+    args = {
+        'q': query,
+        'resolve': 'false',
+        'type': mastodon_type,
+        'limit': page_size,
+    }
+    params['url'] = f"{base_url}/api/v2/search?{urlencode(args)}"
```
return42: Is there no page argument / does the API support paging?

Bnyro: Only when using OAuth, see
https://github.com/searxng/searxng/pull/2857/files#diff-448f76c6b4dc52de868f8b33e441e68d9e34d13caa24c6640c7edeb0eff5c1c7R10

You can see the error messages popping up when trying to scroll down in the search results on mastodon.social without an account.

return42: Ooops,sorry, my eyes were probably already too tired .. :+1:

return42 (approved):
@Bnyro thanks a lot for another engine :+1: .. I squashed some nits into you commit / if you could have a final look .. hint: since we do not have pagination without oAuth, I increased the page size from 10 to 100 .. what do you think?

Bnyro:
@return42 I'll test this later, but especially queries like `!mau Linux` will probably timeout, because the response body (100 results...) is so large (not tested). But if we don't get timeouts by that, it's fine by me.

Bnyro:
(rebased)

Bnyro:
From my tests, it seems like the response times are fine :+1:

I just noticed I can't find a query to get more than about 40 results, it might be the case that they limited it on server side to 40 results per requests maximum?

return42:
> I just noticed I can't find a query to get more than about 40 results, it might be the case that they limited it on server side to 40 results per requests maximum?

I reduced page size to 40

IDK how it could happened, but in you rebase last week from a9dbf8e to c3ca350 you renamed the commit to "[mod] tagesschau: add option to only use tagesschau urls " .. 

In my force push from c3ca350 to b6cbb5c I renamed it back to the title of this PR.

Bnyro:
> I reduced page size to 40
> 
Okay :+1:
> IDK how it could happened, but in you rebase last week from [a9dbf8e](https://github.com/searxng/searxng/commit/a9dbf8ee08caae2ab713bd9426d48aef29445761) to [c3ca350](https://github.com/searxng/searxng/commit/c3ca3509f60d6460457561553ed303b832418cec) you renamed the commit to "[mod] tagesschau: add option to only use tagesschau urls " ..
> 
Oh, no idea what happened there. Git sometimes doesn't behave as I want it too...
Thanks for fixing it!

PR #2860, opened 2023-09-27
[fix] engine - currency fix error and simplify

jazzzooo:
## What does this PR do?

Fixes the currencies engine, as it's broken right now, and simplify the code.

Change in the ddg api.

`!cc 123 btc to eur`


return42 (approved):
can confirm, issue exists and is fixed by this PR .. thanks :+1:

PR #2864, opened 2023-09-28
[feat] engine: implementation of matrixrooms.info

Bnyro:
## What does this PR do?
* implement a new engine: matrixrooms.info

* the engine allows you to quickly search for public Matrix or IRC rooms 

* !mtrx searxng


return42 (approved):
Works like a charm :)

image

dalf:
The paging does not seem to work as expected: the next page removes the first result and add a next result at the end.

Bnyro:
Oh, so we actually need to pass an offset and not a page number. Didn't know that.

PR #2865, opened 2023-09-28
[mod] tagesschau: add option to only use tagesschau urls

Bnyro:
## What does this PR do?

* add a configurable option to the "Tagesschau" engine to return the URL pointing to the URL at tagesschau.de, instead of the real source (e.g. NDR)
* I've built a small alternative frontend and want to redirect all Tagesschau links to it, however that only works when there's one common domain (since I don't want to configure a hostname_replace for each single official public German news media)

* !ts test
* set `use_source_url` to false and search again
* compare the URLs of the results

### DIFF
--- a/searx/engines/tagesschau.py
+++ b/searx/engines/tagesschau.py
@@ -42,6 +42,17 @@ paging = True
 results_per_page = 10
 base_url = "https://www.tagesschau.de"
 
+use_source_url = True
+"""When set to false, display URLs from Tagesschau, and not the actual source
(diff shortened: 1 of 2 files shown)


return42 (approved):
Thanks :+1: .. works like intended / FYI I squashed a note into your commit --> [diff](https://github.com/searxng/searxng/compare/3deb4a74d5a73add6a1e74f2827cd51f8526be9a..6d80bc92eb3ac74846e2e79a8088e5b28e48f868)

Bnyro:
Thank you! LGTM!

PR #2873, opened 2023-09-29
[feature] show the network speed level at each result.

BernieHuang2008:
## What does this PR do?
This PR allows users to see the network between them and the results at a glance, which could help them choose search results that loads faster.

Below are my screenshot, focus on the right-bottom label of each result where i've changed. 

image

Note: all the checks are performed at _user-side_, not _instance-side_, so it's the actuall network speed when user access the result site.
Note2: the check is performed by accessing `favicon.ico` of the result site and they won't trace `favicon.ico`.

give user more information to choose a better result which has a faster speed.

`make run`

### Note3:
i'm still not completely sure about this feature and your thoughts, so it will need further discussion before implement.

What do you think of it?


BernieHuang2008:
manually assign: @return42 @dalf @unixfox  @Bnyro

BernieHuang2008:
people will be informed if they can't access the site by showing "slow" to them.

unixfox:
Unfortunately this won't work like this, not all websites have favicon.ico

Top of that I think this should be a plugin not by default.

searx/static/themes/simple/src/js/main/results.js
```diff
@@ -16,6 +16,36 @@
           }
         ));
 
+    function ping (url, callback) {
+      var start = performance.now();
+      var faviconUrl = new URL('/favicon.ico', url).href;
+      var xhr = new XMLHttpRequest();
```
return42, line 22: I don't think we want to ping every result item .. I see a violation of the privacy aspect in SearXNG .. with this ping the IP from the user is send to every domain in the result list, long before the user has decided to follow a link from the results.

return42 (commented):
TBH I don't see a chance for this feature in SearXNG .. its a massive privacy violation and an (unnecessary) waste of network resources.

I apologize, but I have to decline this PR.

However, thank you for thinking about how we can improve SearXNG for our users :+1:

dalf:
@BernieHuang2008 are you interested in the response time from the instance? For example, how fast is Bing from the instance?

If this is the case, there is already something hidden in the devtools:
https://github.com/searx/searx/pull/1637

From these number it is possible to display the same information from the instance point of view (and not from the browser point of view).

However, this is the kind of option I see as a plugin .... and the plugin framework which does not allow to do that for now.

return42:
> However, this is the kind of option I see as a plugin .... and the plugin framework which does not allow to do that for now.

note: all solutions that ping domains in context of a search request will slow down the SearXNG instance a lot.

return42:
> From these number it is possible to display the same information from the instance point of view (and not from the browser point of view).

I don't think it is the same what @BernieHuang2008 intended with this PR .. https://github.com/searx/searx/pull/1637 measures the response of a search engine .. this PR measures the response time of each domain in the search results.

dalf:
face palm...

|      |       | 
| --- | --- |
| image_proxy | image_proxy |

Still it could be plugin, but there is no way to hook the templates.

Side note: I really like the plugin of [mkdocs](https://github.com/mkdocs/mkdocs).

BernieHuang2008:
You're right, I shall think of another features ...

Thank for your quick response, anyway.

unixfox:
I'm not against a plugin for that. Maybe we could instead use searxng proxy for hiding the ip address when fetching the favicon.

BernieHuang2008:
but if using a proxy, the network speed will be the speed between `instance -- favicon`, not `user -- favicon` and will be meaningless.
btw, why anyone want to trace the favicon ???

what im thinking is what return42 said:
> an (unnecessary) waste of network resources.

i can't find a reason a website wants to monitor on their favicon ...

PR #2874, opened 2023-09-29
[fix] #2866 Bug: Yahoo results for simplified Chinese search sometimes have the first character cut off

### LINKED ISSUE #2866: Bug: Yahoo results for simplified Chinese search sometimes have the first character cut off

The latest commit as of this report, 75587d03a6766c87f2f1964c82080393559ed623

```
make run
```

**What happened?**

As the title says. Current code redirects simplified search to the Hong Kong site, which actually uses traditional Chinese, but marks simplified pages with a `<span class="fz-s fc-obsidian ml-4">簡</span>` after the title text node, at the end:

```html
<h3 style="display:block;margin-top:24px;margin-bottom:2px;" class="title tc"><a style="line-height:26px" class=" d-ib fz-20 lh-26 td-hu tc va-bot mxw-100p" href="https://zh.wikipedia.org/wiki/%E7%B9%81%E4%BD%93%E5%AD%97" referrerpolicy="origin" target="_blank" data-439="6515fd318fbac" aria-label="繁体字 - 维基百科，自由的百科全书"><span class=" s-url d-ib p-abs t-0 l-0 fz-14 lh-20 fc-obsidian wr-bw ls-n pb-4">zh.wikipedia.org<span class=" fc-pewter"> › wiki › <b>繁体字</b></span></span>繁体字 - 维基百科，自由的百科全书<span class="fz-s fc-obsidian ml-4">簡</span></a></h3>
```

This character is extracted by `.xpath('span')` in the first line, which, without knowledge of this situation, actually extracted text from two `span`s, making the second line substring off by one:

https://github.com/searxng/searxng/blob/75587d03a6766c87f2f1964c82080393559ed623/searx/engines/yahoo.py#L139-L140

**How To Reproduce**

Search for some simplified Chinese keyword and look for a Yahoo result with 簡 being the last character of its title.

**Expected behavior**

That mark should be ignored, eliminating the off by one.

**Screenshots & Logs**

problematic search result

You can see that the last breadcrumb is 繁体字, but the title says 体字, cutting off 繁; while there is an extraneous 簡 at the end.

**Additional context**

```diff
diff --git a/searx/engines/yahoo.py b/searx/engines/yahoo.py
index 0fdeacec2..4858e7e3f 100644
--- a/searx/engines/yahoo.py
+++ b/searx/engines/yahoo.py
@@ -136,8 +136,9 @@ def response(resp):
         title = eval_xpath_getindex(result, './/h3/a', 0, default=None)
         if title is None:
             continue
-        offset = len(extract_text(title.xpath('span')))
-        title = extract_text(title)[offset:]
+        offset = len(extract_text(title.xpath('span[1]')))
+        offset_right = -len(extract_text(title.xpath('span[2]'))) or None
+        title = extract_text(title)[offset:offset_right]

         content = eval_xpath_getindex(result, './/div[contains(@class, "compText")]', 0, default='')
         content = extract_text(content, allow_none=True)
```

This is an apparent fix, but I'm not sure it's worth it.

BernieHuang2008:
Co-authored-by: @bnoctis

## What does this PR do?
fix the bug:
yahoo for simplified chinese sometimes cut off the last character.

fixed by: setting a offset_right.

```
make run
```
and search for keyword `!yahoo 繁体字`

### for detail information in test, add these test codes to 
```diff
AT /searx/engines/yahoo.py near L140

         if title is None:
             continue
         offset = len(extract_text(title.xpath('span[1]')))
         offset_right = -len(extract_text(title.xpath('span[2]'))) or None
+        print("LOG1", '\"'+extract_text(title.xpath('span'))+'\"')
+        print("LOG2", '\"'+extract_text(title.xpath('span[1]'))+'\"')
+        print("LOG3", '\"'+extract_text(title.xpath('span[2]'))+'\"')
+        print("LOG4", '\"'+extract_text(title)+'\"')
+        print()
         title = extract_text(title)[offset:offset_right]

```

## Related issues
Closes #2866

### DIFF
--- a/searx/engines/yahoo.py
+++ b/searx/engines/yahoo.py
@@ -159,7 +159,7 @@ def fetch_traits(engine_traits: EngineTraits):
 
     resp = network.get('https://search.yahoo.com/preferences/languages')
     if not resp.ok:
-        print("ERROR: response from peertube is not OK.")
+        print("ERROR: response from yahoo is not OK.")
 
     dom = html.fromstring(resp.text)
     offset = len('lang_')
--- a/searx/engines/yahoo.py
+++ b/searx/engines/yahoo.py
@@ -133,12 +133,7 @@ def response(resp):
             continue
         url = parse_url(url)
 
-        title = eval_xpath_getindex(result, './/h3/a', 0, default=None)
-        if title is None:
-            continue
-        offset = len(extract_text(title.xpath('span')))
-        title = extract_text(title)[offset:]
-
+        title = extract_text(result.xpath('.//h3//a/@aria-label'))
         content = eval_xpath_getindex(result, './/div[contains(@class, "compText")]', 0, default='')
         content = extract_text(content, allow_none=True)
 


return42:
@BernieHuang2008 @bnoctis I think there is a simpler solution that does not need to juggle with offset of a string

- https://github.com/searxng/searxng/pull/2874/files

.. I squashed it into your commit / can you test if it works for you .. thanks :+1:

----

BTW I fixed a `peertube` --> `yahoo` typo.

nc7s:
Taking `aria-label` (or `title`) is indeed better than wrestling offsets. Good find! I should've examined the tags more closely.

(However I don't think the typo fix belongs here, but a separate simple commit.)

BernieHuang2008:
> However I don't think the typo fix belongs here, but a separate simple commit.

Yeah, if you don't mind, I will separate it to another commit. 

I thought a typo commit won't mess the git-tree, will it?

return42:
> (However I don't think the typo fix belongs here, but a separate simple commit.)

Thats true .. because I currently have a lot to do I'm a little sloppy and have packed the exceptionally in the commit ... basically I agree with you but right :100:

PR #2882, opened 2023-09-29
Convert plugin searx.plugin.limiter to normal code

dalf:
## What does this PR do?

The limiter plugin is not a plugin:
* the user can't enable or disable the plugin and there is no point to allow that
* the limiter does not use any of the features provides by the plugin framework

This commit convert the limiter plugin into normal code

Additional change: in `/config`, `limiter.enabled` is `true` only if the limiter is really enabled (Redis is available).

The limiter is shown as a plugin here and there and this is confusing.

* check the app works without the limiter, check `/config`
* check the app works with the limiter and with the token, check `/config`

## Author's checklist

The initialization code is in `searx/botdetection/install.py`, it can be in `searx/botdetection/__init__.py` in the same way that `searx.search` and `searx.redisdb`.
Alternative : move the code from `install.py` to `limiter.py`, but this lead to the next topic: naming ambiguity .

There are two names to reference the same thing:
* limiter
* `searx.botdetection`

Is there a difference between the botdetection and the limiter ?

Since limiter is the known public name, I suggest:
* to rename `searx.botdetection` into `searx.limiter`
* to move the code from `searx/botdetection/limiter.py` and `searx/botdetection/install.py` to `searx/limiter/__init__.py`.

## Related issues

[EDIT] sorry for the numerous edits.

### DIFF
--- a/searx/botdetection/install.py
+++ b/searx/botdetection/install.py
@@ -6,16 +6,15 @@
 import sys
 import flask
 
-from searx import redisdb
-from searx.plugins import logger
+from searx import redisdb, logger
 from searx.botdetection import limiter
 
-name = "Request limiter"
-description = "Limit the number of request"
-default_on = False
-preference_section = 'service'
-
+# the configuration are limiter.toml and "limiter" in settings.yml
+# so, for coherency, the logger is "limiter" even if the module name "searx.botdetection"
 logger = logger.getChild('limiter')
+
+
+_INSTALLED = False
 
 
 def pre_request():
@@ -23,9 +22,15 @@ def pre_request():
     return limiter.filter_request(flask.request)
 
 
-def init(app: flask.Flask, settings) -> bool:
+def is_installed():
+    return _INSTALLED
+
+
+def initialize(app: flask.Flask, settings):
+    """Instal the botlimiter aka limiter"""
+    global _INSTALLED  # pylint: disable=global-statement
     if not settings['server']['limiter'] and not settings['server']['public_instance']:
-        return False
+        return
     if not redisdb.client():
         logger.error(
             "The limiter requires Redis, please consult the documentation: "
@@ -33,6 +38,6 @@ def init(app: flask.Flask, settings) -> bool:
         )
         if settings['server']['public_instance']:
             sys.exit(1)
-        return False
+        return
     app.before_request(pre_request)
-    return True
+    _INSTALLED = True
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -94,7 +94,11 @@ from searx.utils import (
 from searx.version import VERSION_STRING, GIT_URL, GIT_BRANCH
 from searx.query import RawTextQuery
 from searx.plugins import Plugin, plugins, initialize as plugin_initialize
-from searx.botdetection import link_token
+from searx.botdetection import (
+    link_token,
+    initialize as botdetection_initialize,
+    is_installed as botdetection_is_installed,
+)
 from searx.plugins.oa_doi_rewrite import get_doi_resolver
 from searx.preferences import (
     Preferences,
@@ -1288,7 +1292,7 @@ def config():
                 'DOCS_URL': get_setting('brand.docs_url'),
             },
             'limiter': {
-                'enabled': settings['server']['limiter'],
+                'enabled': botdetection_is_installed(),
                 'botdetection.ip_limit.link_token': _limiter_cfg.get('botdetection.ip_limit.link_token'),
                 'botdetection.ip_lists.pass_searxng_org': _limiter_cfg.get('botdetection.ip_lists.pass_searxng_org'),
             },
@@ -1322,6 +1326,7 @@ if not werkzeug_reloader or (werkzeug_reloader and os.environ.get("WERKZEUG_RUN_
     redis_initialize()
     plugin_initialize(app)
     search_initialize(enable_checker=True, check_network=True, enable_metrics=settings['general']['enable_metrics'])
+    botdetection_initialize(app, settings)
 
 
 def run():
--- a/searx/botdetection/__init__.py
+++ b/searx/botdetection/__init__.py
@@ -42,3 +42,6 @@ X-Forwarded-For
 from ._helpers import dump_request
 from ._helpers import get_real_ip
 from ._helpers import too_many_requests
+from .install import initialize, is_installed
+
+__all__ = ['dump_request', 'get_real_ip', 'too_many_requests', 'initialize', 'is_installed']


searx/botdetection/install.py
```diff
@@ -6,33 +6,38 @@
...
+
+
 def pre_request():
     """See :ref:`flask.Flask.before_request`"""
     return limiter.filter_request(flask.request)
 
 
-def init(app: flask.Flask, settings) -> bool:
+def is_installed():
+    return _INSTALLED
```
return42, line 26: Do we need this indirection? .. isn't it simpler to use `botdetection.install.INSTALLED` in line 1295 of webapp.py?

searx/webapp.py
```diff
@@ -94,7 +94,11 @@
 from searx.version import VERSION_STRING, GIT_URL, GIT_BRANCH
 from searx.query import RawTextQuery
 from searx.plugins import Plugin, plugins, initialize as plugin_initialize
-from searx.botdetection import link_token
+from searx.botdetection import (
+    link_token,
+    initialize as botdetection_initialize,
+    is_installed as botdetection_is_installed,
+)
```
return42, line 101: Alias names make it hard to read the code .. instead of lasias names I prefer to import the module name:

    from searx import botdetection

.. and replace the usage of `link_token` by `botdetection.link_token` ...

searx/webapp.py
```diff
@@ -1288,7 +1292,7 @@ def config():
                 'DOCS_URL': get_setting('brand.docs_url'),
             },
             'limiter': {
-                'enabled': settings['server']['limiter'],
+                'enabled': botdetection_is_installed(),
```
return42, line 1295: here we can use `botdetection.initialize.INSTALLED`

searx/botdetection/install.py
```diff
@@ -6,33 +6,38 @@
...
-description = "Limit the number of request"
-default_on = False
-preference_section = 'service'
-
+# the configuration are limiter.toml and "limiter" in settings.yml
+# so, for coherency, the logger is "limiter" even if the module name "searx.botdetection"
 logger = logger.getChild('limiter')
 
 
+_INSTALLED = False
```
return42, line 17: I would prefer `INSTALLED` .. see my comments below.

return42 (changes requested):
Haven't tested .. but left some comments in the code to make it a little bit more readable .. what do you think?

searx/webapp.py
dalf: IMO, either we change all these function names or we keep these alias for now, and make the change in another PR:
https://github.com/searxng/searxng/blob/26fed56d5149e8b56afaa4ecd4dd10332c8900df/searx/webapp.py#L1320-L1324

searx/botdetection/install.py
dalf: It is written following redisdb code:
https://github.com/searxng/searxng/blob/26fed56d5149e8b56afaa4ecd4dd10332c8900df/searx/redisdb.py#L36-L37

See https://github.com/searxng/searxng/pull/686/files#diff-30b71865ec101531bb273e7914fb5db534f88d43fc0643b172bcd189ccec22fbR27

return42:
> Additional change: in `/config`, `limiter.enabled` is `true` only if the limiter is really enabled (Redis is available).

You now, that the limiter can be enabled without redis? .. the methods that [probe HTTP headers](https://docs.searxng.org/admin/searx.botdetection.html#module-searx.botdetection.http_accept) do not need a redis.

searx/webapp.py
return42: > and make the change in another PR

Is Ok for me .. we can do it in a code review.

dalf:
> (Redis is available)

This is misleading, the variable is True only when the limiter is actually installed (Redis installed or not).

https://github.com/searxng/searxng/blob/bee313c61c63aaa42ae859d3fa53041a2d1e67b4/searx/botdetection/install.py#L29-L43

You can see that `_INSTALLED = True` is done at the end after the hook is installed.

dalf:
For information the commit https://github.com/searxng/searxng/commit/c5baf9bfa1d1baf7d370929a4afca903e2e61eae applied on top of this PR move `searx.botdetection` to `searx.limiter`. 
@return42 if you think that make sense, I can add this commit to this PR.

return42:
> if you think that make sense, I can add this commit to this PR.

not sure .. there are also some title or chapter naming that might be better changed from "botdetection" to "limiter" .. lets focus on the code in this PR .. and in a follow up we can discuss how to reorder the documentation.

return42:
> There are two names to reference the same thing:
>
>    limiter
>    searx.botdetection
>
> Is there a difference between the botdetection and the limiter ?

In the past we had said the limiter is a simple builtin that limits request from an IP.  My intention was real *bot-detection* as an (optional) addition to the builtin limiter.  

The `searx.botdetection` namespace has not much (none except redis?) dependency to the rest of the SearXNG core .. As far as possible, I also don't want to bury bot detection too deep in the SearXNG code.  

Bot detection is a task that has nothing to do with the core of SearXNG, this should also be clearly visible on namespace level.  The "module" botdetection should in principle be interchangeable and the limiter is the integration in SearXNG. 

The modularization of limiter and botdetection should keep the possibility open for us to integrate external modules or components into the limiter in the future.

The only thing that would be missing at the moment is the definition of an interface between the limiter and the (external) bot detection.  The currently existing modularization in the code would make this very easy.  If we now merge the limiter with the botdetection this will not be so easy in the future.

TL;DR; just imagine that today's namespace `searx.botdetection` is an external module that you can install.

--> this is also the reason why this module has its own configuration .. 

:thinking: OK we can discuss if `limiter.toml` was a well chosen name :)

dalf:
So the name is `botdetection` not `limiter`.
However, everything outside this module is named `limiter`: the configuration options, configuration file, the API, etc...

There is no need to solve this in this PR: here, the purpose is just to remove the plugin [*] does not make sense.

[*] plugin as they are defined now.

searx/botdetection/install.py
```diff
@@ -6,33 +6,38 @@
...
+def is_installed():
+    return _INSTALLED
+
+
+def initialize(app: flask.Flask, settings):
+    """Instal the botlimiter aka limiter"""
+    global _INSTALLED  # pylint: disable=global-statement
     if not settings['server']['limiter'] and not settings['server']['public_instance']:
-        return False
+        return
```
return42, line 33: This and what the undocumented `is_installed()` returns is unclear to me.

Does it mean `is_installed()` returns true only when redis is available and `settings['server']['public_instance']` is set to `true`?

Sorry, but this option `public_instance` only causes confusion, how is a developer supposed to understand this code without getting all the undocumented background to it?

dalf, line 33: Please Markus, is it possible for you to deal with choice that are not yours, like others do?
Yesterday, you said you want to do your PR, then switch back to my PR when I suggest to create myself a package for botdetection. The underlying question was not what to do, but the fact you want to ownership of the code. This is exhausting, and what remains in mind my mind is the fact that SearXNG is a playground and want to do whatever you want.

dalf:
Supersede by https://github.com/searxng/searxng/pull/2894

PR #2885, opened 2023-09-30
[fix] matrixrooms.info: pagination not working properly

Bnyro:
## What does this PR do?
* matrixrooms.info requires an offset parameter, not a page number

## Related issues
see https://github.com/searxng/searxng/pull/2864#issuecomment-1741716364


return42 (approved):
Fixes the issue / thanks :+1:

PR #2888, opened 2023-09-30
[feat] engine: implementation of podcastindex.org

Bnyro:
## What does this PR do?
* add a new engine for searching podcasts

* we currently only have gpodder to search for podcasts (afaik)

* !podcast linux

## Notes
* not sure if we should add it to the music category or not, currently it's not part of any category

### DIFF
--- a/searx/engines/podcastindex.py
+++ b/searx/engines/podcastindex.py
@@ -1,0 +1,43 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Podcast Index
+"""
+
+from urllib.parse import quote_plus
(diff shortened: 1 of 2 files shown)


unixfox:
image

Also quite big images

Bnyro:
That's fine for podcasts imo.

return42 (commented):
Perfect / thanks a lot :+1:

@unixfox you mentioned ..

> Also quite big images

I think its the way we render podcast items / modifing the layout of podcasts is not related to this PR I think.

Is it OK to merge the PR or are changes still necessary?

return42 (approved):

PR #2889, opened 2023-09-30
[fix] emojipedia: fix engine

Bnyro:
## What does this PR do?
* fix the emojipedia engine

* the html layout of the engine changed a while ago, hence the engine doesn't work anymore

* !em bird
* !em poland

## Author's checklist
* in order to get the description too (about 50% of the emojis have one), we would need to parse the JavaScript code at the bottom of the page, which could break way easier however. Thus I chose to use the solution that is more stable.

### DIFF
--- a/searx/engines/emojipedia.py
+++ b/searx/engines/emojipedia.py
@@ -14,7 +14,6 @@ from lxml import html
 
 from searx.utils import (
     eval_xpath_list,
-    eval_xpath_getindex,
     extract_text,
 )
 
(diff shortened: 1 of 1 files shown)


return42 (approved):
Fixes the issue :+1:

obfuscated-loop:
So sorry! I actually noticed this change the day it happened and forgot to make a PR!

Bnyro:
No worries, since nobody complained before that PR I think that nobody else noticed it either :)

PR #2890, opened 2023-09-30
[feat] engine: implementation of fyyd

Bnyro:
## What does this PR do?
* add a new engine: fyyd, which allows searching for podcasts

* we currently lack podcast search engines, and fyyd provides very good results

* !fyyd linux

### DIFF
--- a/searx/engines/fyyd.py
+++ b/searx/engines/fyyd.py
@@ -1,0 +1,50 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Fyyd (podcasts)
+"""
+
+from datetime import datetime
+from urllib.parse import urlencode
+
+about = {
+    'website': 'https://fyyd.de',
+    'official_api_documentation': 'https://github.com/eazyliving/fyyd-api',
+    'use_official_api': True,
+    'require_api_key': False,
+    'results': 'JSON',
+}
+categories = []
+paging = True
+
+base_url = "https://api.fyyd.de"
+page_size = 10
+
+
+def request(query, params):
+    args = {
+        'term': query,
+        'count': page_size,
+        'page': params['pageno'] - 1,
+    }
+    params['url'] = f"{base_url}/0.2/search/podcast?{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    results = []
+
+    json_results = resp.json()['data']
+
+    for result in json_results:
+        results.append(
+            {
+                'url': result['htmlURL'],
+                'title': result['title'],
+                'content': result['description'],
+                'thumbnail': result['smallImageURL'],
+                'publishedDate': datetime.strptime(result['status_since'], '%Y-%m-%d %H:%M:%S'),
+                'metadata': f"Rank: {result['rank']} || {result['episode_count']} episodes",
+            }
+        )
+
+    return results
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -702,6 +702,12 @@ engines:
     shortcut: frk
     disabled: true
 
+  - name: fyyd
+    engine: fyyd
+    shortcut: fy
+    timeout: 8.0
+    disabled: true
+
   - name: genius
     engine: genius
     shortcut: gen


return42:
FYI: first_page_num is set to zero (SearXNG's default is 1) --> [diff](https://github.com/searxng/searxng/compare/f2e4d6a48e1794d8c91c5b2f5282c5ebd2509751..f663fafa0436d43be6605b7ecff6bde47ab602b9)

Not sure if its worth and needed to implement a py-module for this engine .. I would say: "its worth, the engine has *good* result!" .. however we should consider to support:

### A. the *grouping* of podcast series needs to be supported

For instance with this xpath-engine we fetch the results from the *grouping* (**Coopers Kaffee**)

image

The result from https://fyyd.de/search?search=b%C3%B6hmermann&page=0 are enumerations of podcast episodes, grouped by series.

image

What we need in the result lists is the enumeration of podcast episodes.


### B. url to the stream of the podcast recording 

If we fetch the link to the episode (described in A.) it would be best we have a url to the stream of the podcast recording  / but might not be possible within one request.

I think B. is a "nice to have", but A. is a prerequisite: without links to episodes the result list does IMO not have the expected quality.

I don't know if we can implement A. with an X-Path engine, or if we need a python module after all.

Bnyro:
Thanks for your review on that :+1:

## A
Generally, we have the choice to either
* Only display the most recent episode of each podcast
* Only display the oldest/first episode of each podcast
* Display all listed episodes as a single result, but that could cause a bloated list with results from one and the same podcast (probably a bad idea)
* as suggested by you: only display a list of titles of the episodes (wouldn't that be a bit useless, the descriptions of a single episode usually tell you more about the podcast than all the titles :thinking:)

In my opinion, displaying the most recent episode is probably the best idea.



## B
That would probably be a cool thing, but I don't think that's really possible with fyyd unfortunately (at least I don't have any idea how it should work).

## TLDR

In the end we probably should implement a module to not overcomplicate the xpath configuration, and to allow parsing some additional fields, also when considering B in the future. However I'm not sure when I'll find motivation for that, we'll see when someone is motivated to look deeper into the engine :)

return42:
> In my opinion, displaying the most recent episode is probably the best idea.

Decision  up to you .. I only think that the current grouping is not good.

Bnyro:
I figured out there's a JSON API, not sure how I missed that before ...

The results should be much better now.

unixfox:
image

The images for each results are quite big.

Bnyro:
Well, these are podcasts, so using a larger thumbnail is the right choice in that case imo.

unixfox:
I personally much prefer smaller image, it's easier to compare each results.

And you have more results displayed at the same time without having to scroll.

return42:
> The results should be much better now.

Yeah, much better / thanks :+1: 

But found some other nits ..

image

The link points to https://www.zdf.de/rss/podcast/video/zdf/comedy/neo-magazin-mit-jan-boehmermann which returns a json rsult from zdf.de

image

Other results are linking to RSS feeds like https://www.zdf.de/rss/podcast/audio/zdf/politik/maybrit-illner

Not sure .. can we scrap "better" URLs for the result list?

Bnyro:
For reference: https://api.fyyd.de/0.2/search/podcast?term=b%C3%B6hmermann&count=10&page=0

Apparently the location of the podcast changed and fyyd didn't yet receive an update with the new URL - hence they likely have to fix it.

We could use the `url_fyyd` parameter in theory (e.g. https://fyyd.de/podcast/zdfneo-neo-magazin-royale/0), but this would make the engine kind of unnecessary since you will need to use fyyd itself first before being able to start the podcast anyways.

TL;DR: it's not our issue

return42 (approved):
> We could use the `url_fyyd` parameter in theory (e.g. https://fyyd.de/podcast/zdfneo-neo-magazin-royale/0), but this would make the engine kind of unnecessary since you will need to use fyyd itself first before being able to start the podcast anyways.

Thats true.

> TL;DR: it's not our issue

Anyway .. thanks for your explanation .. 

LGTM .. IMO ready to merge :+1:

PR #2891, opened 2023-10-01
[fix] enable bing by default / was disabled in cea7b71d14

return42:
The engine was deactivated at that time (cea7b71d14 in #1235) because there were problems, but they are fixed now.

- https://github.com/searxng/searxng/pull/1235

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -384,7 +384,6 @@ engines:
   - name: bing
     engine: bing
     shortcut: bi
-    disabled: true
 
   - name: bing images
     engine: bing_images


Bnyro (approved):

unixfox:
There is still this bug which should be fixed before we enable it back: https://github.com/searxng/searxng/issues/2641

Happens on a local instance and public one.

Bnyro:
I disagree on that. Even though that's an issue, the Bing response times are quick enough to not be the slowest engine. So there are no drawbacks from enabling it even if there are no results for some queries.

unixfox:
It's not "some", for me it was 80% of my queries didn't give any results.

unixfox:
Was bing engine fixed? I can't reproduce the issue above.

If so then I'm for enabling bing by default BUT we have to disable another engine by default, otherwise by default the search will be even slower as fetching 5 engines takes way more time than 4.

I'm for disabling either brave or qwant as they are always blocked on public instances.

return42:
> It's not "some", for me it was 80% of my queries didn't give any results.

I agree with @unixfox, considerable problems with bing have become apparent over time:

- [label: bing-engine](https://github.com/searxng/searxng/labels/bing-engine)

under these conditions it no longer makes sense to activate this engine by default / I close the PR.

Issue #2892, opened 2023-10-01
uwsgi adds Connection: close  --> client is rated as a bot


return42:
Related to an old filtron issue, `Connection: close` is added to the HTTP headers .. [The http_connection method evaluates a request as the request of a bot if the header is set to close](https://docs.searxng.org/admin/searx.botdetection.html#method-http-connection). I can't understand how this setting could survive here for so long ... since we replaced filtron with the limiter, clients are classified as bot if they give a `Connection: close`.

https://github.com/searxng/searxng/blob/e509cb7c459a59c6793922bce6ce0cc57006aac6/dockerfiles/uwsgi.ini#L37-L39

Pointed out by @samsapti on [Matrix](https://matrix.to/#/!uotInzBKwlMfhtqZvW:matrix.org/$rvmUvX6jbfzz05XMcruNQjq1jepz1Pzm2vbv7MeMvec?via=matrix.org&via=envs.net&via=frei.chat)

Related:

- https://github.com/searx/searx-docker/issues/24

catfluoride:
I guess this is only true for docker installs?

dalf:
The [Connection header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection) works with the HTTP/1 protocol, it does not work with HTTP/2 and HTTP/3. According to MDN, the 

> * `close` indicates that either the client or the server would like to close the connection. This is the default on HTTP/1.0 requests.
> * any comma-separated list of HTTP headers [Usually `keep-alive` only]. Indicates that the client would like to keep the connection open. Keeping a connection open is the default on HTTP/1.1 requests.

So the check `Connection=closed` means if the HTTP use the HTTP/1.0 protocol and does not use HTTP/1.1 (keep alive), then it is a bot.

Now a workflow:

```mermaid
sequenceDiagram
    Browser->>Nginx: (a) most probably HTTP/2 connection 
    Nginx->>Uwsgi: (b) HTTP/1.0 connection, Nginx sets "Connection:close"
    Uwsgi->>SearXNG: (c) WSGI request
    SearXNG-->>Uwsgi: (d) WSGI response
    Uwsgi-->>Nginx: (e) uwsgi.ini : "add-header = Connection: close"
    Nginx-->>Browser: (f) HTTP/2 response
```

b:
[...]

PR #2893, opened 2023-10-01
[fix] limiter / botdetection: remove http_connection method

return42:
Related:

- https://github.com/searxng/searxng/issues/2892#issuecomment-1742153932

### DIFF
--- a/searx/botdetection/limiter.py
+++ b/searx/botdetection/limiter.py
@@ -51,7 +51,6 @@ from . import (
     http_accept,
     http_accept_encoding,
     http_accept_language,
-    http_connection,
     http_user_agent,
     ip_limit,
     ip_lists,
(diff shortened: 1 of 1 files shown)


unixfox:
While this may fix some issues with uwsgi, removing this rule reduce the effectiveness of the bot protection.

I have seen many many bots sending the connection close header on my server. And this rule have blocked a lot of bots.

searx/botdetection/limiter.py
```diff
@@ -51,7 +51,6 @@
     http_accept,
     http_accept_encoding,
     http_accept_language,
-    http_connection,
     http_user_agent,
     ip_limit,
     ip_lists,
```
return42, line 56: > I have seen many many bots sending the connection close header on my server. And this rule have blocked a lot of bots.

After https://github.com/searxng/searxng/pull/2894 has been reviewed & merged we can consider to make this rules configurable in the `limiter.toml`.

unixfox:
Forgot to say that I'm using NGINX for this rule not the official bot limiter in searxng.

PR #2894, opened 2023-10-02
[mod] isolation of botdetection from the limiter

return42:
This PR is an alternative to https://github.com/searxng/searxng/pull/2882

It mainly consist of two commits ...

#### `[mod] isolation of botdetection from the limiter`

This patch was inspired by the discussion around PR-2882 [2].  The goals of this
patch are:

1. Convert plugin searx.plugin.limiter to normal code [1]
2. isolation of botdetection from the limiter [2]
3. searx/{tools => botdetection}/config.py and drop searx.tools
4. in URL /config, 'limiter.enabled' is true only if the limiter is really
   enabled (Redis is available).

This patch moves all the code that belongs to botdetection into namespace
searx.botdetection and code that belongs to limiter is placed in namespace
searx.limiter.

Tthe limiter used to be a plugin at some point botdetection was added, it was
not a plugin.  The modularization of these two components was long overdue.
With the clear modularization, the documentation could then also be organized
according to the architecture.

[1] https://github.com/searxng/searxng/pull/2882
[2] https://github.com/searxng/searxng/pull/2882#issuecomment-1741716891

#### `[mod] document server:public_instance & remove it out of the botdetection`

- the option server:public_instance lacks some documentation
- the processing of this option belongs in the limiter and not
  in botdetection module

#### To test:

- check the app works without the limiter, check `/config`
- check the app works with the limiter and with the token, check `/config`
- make docs.live .. and read
  - http://0.0.0.0:8000/admin/searx.limiter.html
  - http://0.0.0.0:8000/src/searx.botdetection.html#botdetection
- see https://github.com/searxng/searxng/pull/2882


dalf:
~Supersede by #2894~

searx/limiter.py
```diff
@@ -35,36 +59,72 @@
...
 
+
+Configure Limiter
+=================
+
+The methods of :ref:`botdetection` the limiter uses are configured in a local
+file ``/etc/searxng/limiter.toml``.  The defaults are shown in limiter.toml_ /
+Don't copy all values to your local configuration, just enable what you need by
+overwriting the defaults.  For instance to activate the ``link_token`` method in
+the :ref:`botdetection.ip_limit` you need only to set this option to ``true``:
```
Bnyro:
```suggestion
the :ref:`botdetection.ip_limit` you only need to set this option to ``true``:
```

Bnyro (approved):

searx/limiter.py
return42: done --> [diff](https://github.com/searxng/searxng/compare/5eb3f575ba42ae2a951777beb379fef679661465..76b41d35632e355424a1150b8c6db8b00fc3707c)

searx/limiter.py
```diff
@@ -211,23 +211,33 @@ def pre_request():
...
         return
-    botdetection.init(get_cfg(), redis_client)
-    app.before_request(pre_request)
+
     _INSTALLED = True
+
+    cfg = get_cfg()
+    if settings['server']['public_instance']:
+        # overwrite limiter.toml setting
+        cfg.set('botdetection.ip_limit.link_token', True)
```
dalf, line 240: I was thinking about this either at the top of this function or in settings_default.py:

```python
    if settings['server']['public_instance']:
        # overwrite limiter.toml setting
        cfg = get_cfg()
        cfg.set('botdetection.ip_limit.link_token', True)
        # overwrite settings.yml
        settings['server']['limiter'] = True
```

so:
* b05a155 can be revert
* line 222 can check `settings['server']['limiter']` and there is no need to think about `server.public_instance` anymore in the future.
* the code above describes all the changes made by `server.public_instance` in one place.

return42: > https://github.com/searxng/searxng/commit/b05a15540e1dc2dfb8e4e25aa537b2a68e713844 can be revert

not yet, see the "or" ..

    get_setting('server.limiter') or get_setting('server.public_instance') 

the point is, that `server.public_instance` does not activate the `server.limiter`

I wanted to prepare a PR for this, but first I needed to merge this PR / to prevent a conflict of the new PR and this PR (the new PR needs to be on top of this PR here).

dalf: >the point is, that server.public_instance does not activate the server.limiter

??

ping @unixfox

unixfox: > This new parameter for the moment activate by default both the bot limiter AND the new advanced bot limiter called link_token. 

https://github.com/searxng/searx-instances/discussions/417

So yes it should activate server.limiter

PR #2895, opened 2023-10-02
[feat] Improve logging & allow more control over log levels

y0nei:
## What does this PR do?

This PR aims to change the existing logging functionality by allowing users to have more control over the log level.

Most of the time, users generally wont focus on `DEBUG` or `INFO` logs, instead they would shift their focus on whats actually failing or causing issues. Keeping all log levels as one or two default levels could introduce a lot of clutter in the logs and potentially makes new contributors confused on how to properly debug the app. This change adds clarity to the current system, making it more readable and allowing more fine grained control over the application behavior.

Just add the new `log_level` setting in the `general:` section of the `settings.yml` file, the values should match entries in the `ValidLogLevels` class. Case sensitivity does not matter since the strings will get `.upper()`'ed anyway.


return42:
@dalf what do you think about this PR?  I think that if we are going to tackle the issue of logging, we should move logging to its own module and leave it to the operator to use a [standard configuration file](https://docs.python.org/3/library/logging.config.html#logging-config-fileformat) for logging 

----

UPDATE: see comment ..

- https://github.com/searxng/searxng-docker/discussions/220#discussioncomment-8607684

return42:
I close this PR .. see https://github.com/searxng/searxng/pull/2895#issuecomment-1868248433

PR #2899, opened 2023-10-03
[feat] engine: implementation of hackernews

Austin-Olacsi:
## What does this PR do?

adds the hackernews engine

hackernews was requested

make run, and search with bang: apple !hn

## Author's checklist

it supports time ranges now
additionally if the query is front_page or fp it will get results from the hackernews front page

### DIFF
--- a/searx/engines/hackernews.py
+++ b/searx/engines/hackernews.py
@@ -1,0 +1,91 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Hackernews
+"""
+
+from datetime import datetime
+from urllib.parse import urlencode
+from dateutil.relativedelta import relativedelta
+
+from flask_babel import gettext
+
+# Engine metadata
+about = {
+    "website": "https://news.ycombinator.com/",
+    "wikidata_id": "Q686797",
+    "official_api_documentation": "https://hn.algolia.com/api",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+# Engine configuration
+paging = True
+time_range_support = True
+categories = ["it"]
+results_per_page = 30
+
+# Search URL
+base_url = "https://hn.algolia.com/api/v1"
+
+
+def request(query, params):
+    search_type = 'search'
+    if not query:
+        # if search query is empty show results from HN's front page
+        search_type = 'search_by_date'
+        query_params = {
+            "tags": "front_page",
+            "page": (params["pageno"] - 1),
+        }
+    else:
+        query_params = {
+            "query": query,
+            "page": (params["pageno"] - 1),
+            "hitsPerPage": results_per_page,
+            "minWordSizefor1Typo": 4,
+            "minWordSizefor2Typos": 8,
+            "advancedSyntax": "true",
+            "ignorePlurals": "false",
+            "minProximity": 7,
+            "numericFilters": '[]',
+            "tagFilters": '["story",[]]',
+            "typoTolerance": "true",
+            "queryType": "prefixLast",
+            "restrictSearchableAttributes": '["title","comment_text","url","story_text","author"]',
+            "getRankingInfo": "true",
+        }
+
+        if params['time_range']:
+            search_type = 'search_by_date'
+            timestamp = (datetime.now() - relativedelta(**{f"{params['time_range']}s": 1})).timestamp()
+            query_params["numericFilters"] = f"created_at_i>{timestamp}"
+
+    params["url"] = f"{base_url}/{search_type}?{urlencode(query_params)}"
+    return params
+
+
+def response(resp):
+    results = []
+    data = resp.json()
+
+    for hit in data["hits"]:
+        object_id = hit["objectID"]
+        points = hit["points"] or 0
+        num_comments = hit["num_comments"] or 0
+
+        metadata = ""
+        if points != 0 or num_comments != 0:
+            metadata = f"{gettext('points')}: {points}" f" | {gettext('comments')}: {num_comments}"
+        results.append(
+            {
+                "title": hit["title"] or f"{gettext('author')}: {hit['author']}",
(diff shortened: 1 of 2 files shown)


searx/engines/hackernews.py
```diff
@@ -0,0 +1,98 @@
...
+            if timestamp is not None:
+                query_params["numericFilters"] = f"created_at_i>{timestamp}"
+
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+    return params
+
+
+def calculate_timestamp(time_range):
+    time_ranges = {'year': 365, 'month': 30, 'week': 7, 'day': 1}
+    if time_range in time_ranges:
```
Bnyro: That can't be false, we're already checking if a `time_range` is selected, and `year`, `month`, `week` and `day` are the only time ranges we support.

searx/engines/hackernews.py
```diff
@@ -0,0 +1,98 @@
...
+    units = [("year", 31536000), ("month", 2592000), ("day", 86400)]
+    time_difference = int(datetime.utcnow().timestamp()) - created_at_i
+
+    for unit, seconds_in_unit in units:
+        if time_difference >= seconds_in_unit:
+            count = int(time_difference // seconds_in_unit)
+            return f"{count} {unit}{'s' if count > 1 else ''} ago"
+
+    return "past day"
+
```
Bnyro: We should remove this code and use `publishedDate` instead, see below

searx/engines/hackernews.py
```diff
@@ -0,0 +1,98 @@
...
+        title = f"post by {author}" if title is None or not title else title
+        num_comments = 0 if num_comments is None or not num_comments else num_comments
+        points = 0 if points is None or not points else points
+
+        content = (
+            f"{points} point{'s' if points != 1 else ''} | "
+            f"{author} | "
+            f"{time_ago(created_at_i)} | "
+            f"{num_comments} comment{'s' if num_comments != 1 else ''} | "
+            f"https://news.ycombinator.com/item?id={object_id}"
```
Bnyro: This url shouldn't be part of the content

searx/engines/hackernews.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        results.append({"title": title, "url": url, "content": content})
```
Bnyro:
```suggestion
        results.append({"title": title, "url": url, "content": content, "publishedDate": datetime.utcfromtimestamp(created_at_i)})
```

searx/engines/hackernews.py
```diff
@@ -0,0 +1,98 @@
...
+            hit["points"],
+            hit["author"],
+            hit["created_at_i"],
+            hit["num_comments"],
+        )
+
+        url = f"https://news.ycombinator.com/item?id={object_id}" if url is None or not url else url
+        title = f"post by {author}" if title is None or not title else title
+        num_comments = 0 if num_comments is None or not num_comments else num_comments
+        points = 0 if points is None or not points else points
```
Bnyro: It might make sense to add a small helper function like
```python
def _get_or_default(value, default):
    if value and value is not None:
        return value
    return default
```

searx/engines/hackernews.py
Austin-Olacsi: i actually thought to use this for the url instead, because i like to read the discussion on hackernews, there is no way to find it without. @return42 what do you think?

Austin-Olacsi:
made some changes based on the review... code is shorter and cleaner now... let me know your thoughts

searx/engines/hackernews.py
return42: A simple alternative to expressions like:

```python
title = f"post by {author}" if title is None or not title else title`
```

is:

```python
title = title or f"post by {author}"`
```

searx/engines/hackernews.py
return42: Since this engine is for hackernews, users would expect links to hackernews .. or I'm wrong?

return42:
> made some changes based on the review... code is shorter and cleaner now... let me know your thoughts

Thanks a lot, I placed a WIP commit on top with suggestions from mine.  I added a `search_type = 'search'` variable that is `'search_by_date'` for time related search.

But I think the search results are not what known from the https://hn.algolia.com/?dateRange=all&page=0&prefix=false&query=searxng&sort=byDate&type=story

grafik

compared with the result from the engine implemented in this PR:



grafik

Austin-Olacsi:
added another commit to improve these results, here was before:
before
and here is after:
after
let me know if you need anything else.

return42:
> added another commit to improve these results

1. You added the story_tilte back I removed intentional: the story title does not fit to what the result links. The linked article of the result is one post related to the story title, but it is not the story title. Its related to comment above https://github.com/searxng/searxng/pull/2899#discussion_r1349492639

> Since this engine is for hackernews, users would expect links to hackernews .. or I'm wrong?

2. In my comment above https://github.com/searxng/searxng/pull/2899#issuecomment-1751661876 I asked why the results from this engine are so much different. From SearXNG's engine the user mght expect identical results he gets with the search from hackernews.

Austin-Olacsi:
1. oops... my apologies, i'll change it back


2. thanks for clarifying, i wasn't sure what you where asking in https://github.com/searxng/searxng/pull/2899#issuecomment-1751661876

i figured out how to get identical results... will make another commit soon

Austin-Olacsi:
story_title is removed again, but we should keep the story_text because it is related, and it appears in the hackernews search results when it is available:
story text

Austin-Olacsi:
search results are now  identical:
sxng results
hn results

return42 (approved):
Results are now much better than before / thanks a lot :+1: 

I squashed all your commits into one commit and fixed the commit message slightly, further I squashed gettext into your commit --> [diff](https://github.com/searxng/searxng/compare/2837f7bf72b8b7c730f0d25babf19b62ed9a080b..a9d12e43b5ec197e0b86f2798c4e155813fdfc74) / we need gettext for the translations.

PR #2901, opened 2023-10-05
improve issue search

BernieHuang2008:
## What does this PR do?
As we said in #2896 ,
> > Besides, since the search query is more clear, the `submit issue` button can be shown after **clicking** the `check existing issues` link, forcing the user to check existing issues.
(You know better than me, at present, the button is triggered after waiting a couple of seconds )
> 
> Yes, this is a weak point in the current process ..
> 
> _Originally posted by @return42 in https://github.com/searxng/searxng/discussions/2896#discussioncomment-7196808_

the `search existing issues` link can be improved.

I manually generated a `technical report` (100% the same as in the issue), and add it to the link.
So now, the link will search for issues that contain the same "technical report" and filter out irrelevant issues.

The problem is, i'm sure that this is NOT the most effective way, but it is a good idea.

## screenshots
after this PR, mountains of issues are filted 👇 (only 7 left! )
image

If we could do it cleaner, maybe there would be only 1 issue left, and that one happens to be the same!

### DIFF
--- a/searx/templates/simple/new_issue.html
+++ b/searx/templates/simple/new_issue.html
@@ -63,7 +63,7 @@ or manually by executing the searx/webapp.py file? -->
     <input type="checkbox" id="step1">
     <label for="step1">{{ _('Start submiting a new issue on GitHub') }}</label>
     <div class="step1 step_content">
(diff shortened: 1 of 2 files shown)


return42:
FYI: the issue reported by the [CI](https://github.com/searxng/searxng/actions/runs/6418682784/job/17426967402?pr=2901#step:7:55) can be fixed by code formating:

     make format.python

unixfox (approved):
Good I really like this new way to report bugs.

return42 (approved):
@BernieHuang2008 please excuse my very late review ... I lost sight of your PR. Thanks a lot for your work on this topic :+1:

PR #2902, opened 2023-10-05
[feat] engine: implementation of bpb (de)

Bnyro:
## What does this PR do?
* implement a new German engine, https://www.bpb.de

* there's a lot of misinformation concerning history and politics in and of Germany, hence it's important to have trustworthy sources available

* !bpb deutsche einheit
* !bpb grundrechte

### DIFF
--- a/docs/dev/engines/online/bpb.rst
+++ b/docs/dev/engines/online/bpb.rst
@@ -1,0 +1,13 @@
+.. _bpb engine:
+
+===
+Bpb
+===
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.bpb
+  :members:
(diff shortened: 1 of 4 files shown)


unixfox (approved):

return42 (changes requested):
When tested I got an exception .. can you please have a look / thanks!

```
  File "searx/search/processors/online.py", line 165, in search
  File "searx/search/processors/online.py", line 153, in _search_basic
  File "searx/engines/bpb.py", line 60, in response
    'publishedDate': datetime.utcfromtimestamp(result['extension']['publishingDate']),
KeyError: 'publishingDate'
```

Bnyro:
> When tested I got an exception .. can you please have a look / thanks!
> 
Good catch (and fixed now), thanks for having a look!

PR #2912, opened 2023-10-07
[feat] duckduckgo: support for videos and news

### LINKED ISSUE #1602: DuckDuckGo Videos/News/Maps
**Working URL to the engine**

https://duckduckgo.com/?q=news&t=h_&iar=news&ia=news -- a search for "news"
https://duckduckgo.com/?q=videos&t=h_&iax=videos&ia=videos -- a search for "videos"
https://duckduckgo.com/?q=maps&t=h_&ia=web&iaxm=maps -- a search for "maps"

It's DuckDuckGo and this specific news category of it is not yet implemented in SearX.

**Features of this engine**

DuckDuckGo supposedly doesn't track its users and the news category is fairly reliable in my experience with another library (linked below).

news, map, videos

**Additional context**

It would be great to be able to expand the number of news search engines SearX supports. At the very least, it would be nice to have news. I would love to submit a PR myself, but I wouldn't know how to implement or test it. Could someone point me in the right direction for something like a template for creating new engines?

Bnyro:
## What does this PR do?
* add support for videos and news from DuckDuckGo by reusing the code for the ddg images engine

* duckduckgo has quite good results, so these are worth adding

* !ddv hello world
* !ddn hello world

## Related issues
closes #1602 (I don't think it makes sense/is possible to add maps)


searx/engines/duckduckgo_extra.py
```diff
@@ -69,28 +73,68 @@ def request(query, params):
...
+        'metadata': result.get('uploader'),
+    }
+
+
+def _news_result(result):
+    return {
+        'url': result['url'],
+        'title': result['title'],
+        'content': result['excerpt'],
+        'img_src': result.get('image'),
```
return42: The images in the news category are often very big images and not small thumbs.  Since SearXNG uses this image as a thumb, I think it might be better we remove `img_src` the result items of ddg-news.

return42 (changes requested):
> by reusing the code for the ddg images engine

I didn't expect it to be so easy to do :rocket: 

I fixed minor nits [diff](https://github.com/searxng/searxng/compare/14e0659..84e827c3669183e482dc4a3c0801921831665af5) and left a comment in the code .. what do you think about the `img_src` in the ddg-news?

searx/engines/duckduckgo_extra.py
```diff
@@ -69,28 +73,68 @@ def request(query, params):
...
 
+def _image_result(result):
+    return {
+        'template': 'images.html',
+        'url': result['url'],
+        'title': result['title'],
+        'content': '',
+        'thumbnail_src': result['thumbnail'],
+        'img_src': result['image'],
+        'img_format': '%s x %s' % (result['width'], result['height']),
```
allendema, line 92: One would expect the ```img_format``` to be something like ```png```, ```raw```, ```ẁebp``` etc.

https://docs.searxng.org/dev/engines/engine_overview.html#images
The img_format is not specified in engine docs, see next comment below.

searx/engines/duckduckgo_extra.py
```diff
@@ -69,28 +73,68 @@ def request(query, params):
...
+    return {
+        'template': 'videos.html',
+        'url': result['content'],
+        'title': result['title'],
+        'content': result['description'],
+        'thumbnail': result['images'].get('small') or result['images'].get('medium'),
+        'iframe_src': result['embed_url'],
+        'source': result['provider'],
+        'length': result['duration'],
+        'metadata': result.get('uploader'),
```
allendema, line 107: https://docs.searxng.org/dev/engines/engine_overview.html#videos
We should consider adding ```metdata``` or ```extra``` field to all templates.
Which is a dict and it includes all kinds of metadata, even if they will not be used in the UI.

Also the ```ìframe_src``` should be added to docs.

searx/engines/duckduckgo_extra.py
Bnyro: I agree, but I didn't touch this because it's unrelated to this PR's changes (which only focuses at videos and news).

searx/engines/duckduckgo_extra.py
return42: The macros `result_header`, `result_sub_header` and `result_sub_footer`:

https://github.com/searxng/searxng/blob/c3ab49cd903d27905d2da6f70699a55c9a74593e/searx/templates/simple/macros.html#L20-L48

are included in various templates / for instance in the `video.html` template:

https://github.com/searxng/searxng/blob/c3ab49cd903d27905d2da6f70699a55c9a74593e/searx/templates/simple/result_templates/videos.html#L3-L4

https://github.com/searxng/searxng/blob/c3ab49cd903d27905d2da6f70699a55c9a74593e/searx/templates/simple/result_templates/videos.html#L18-L18

> We should consider adding ```metdata``` or ```extra``` field to all templates.

we have a `metadata` field in the `result_sub_header` macro:

https://github.com/searxng/searxng/blob/c3ab49cd903d27905d2da6f70699a55c9a74593e/searx/templates/simple/macros.html#L34-L39

searx/engines/duckduckgo_extra.py
return42: > I agree, but I didn't touch this because it's unrelated to this PR's changes (which only focuses at videos and news).

@Bnyro when this PR has been merged, can you please post a follow up PR to fix the image issue / thanks!

searx/engines/duckduckgo_extra.py
Bnyro: Okay, fine with me :+1:

searx/engines/duckduckgo_extra.py
Bnyro: Sure!

searx/engines/duckduckgo_extra.py
```diff
@@ -69,28 +73,68 @@ def request(query, params):
         args['p'] = safe_search  # "-1", "1"
 
     logger.debug("cookies: %s", params['cookies'])
+
+    path = 'i'
+    if ddg_category == 'videos':
```
Bnyro: This should be changed to a map from category to url path to make the code easier to read.

searx/engines/duckduckgo_extra.py
Bnyro: (removed)

return42 (approved):

PR #2913, opened 2023-10-07
[fix] kickass: crash when no results

Bnyro:
## What does this PR do?
* fix crashes when there are no results during kickass searches by making sure the table element actually has an anchor element/link to a torrent file

* !kickass genre-ab


return42 (approved):
I can confirm the issue and the fix in this patch / thanks!

PR #2914, opened 2023-10-07
[mod] piped: always show video length if available

### LINKED ISSUE #2911: Time duration for piped videos
Like piped music feature, it il possible to add time video duration please ?

**Additional context**
With music whe can see Length : 
final_music

With video we can not see Length : 
final_video

Could you add this feature please ?
Thanks :)

Bnyro:
## What does this PR do?
* we previously only displayed the length/duration for music results, which doesn't make much sense because it's important for videos too.

!piped hello world

## Related issues
closes #2911


return42 (approved):
I can confirm the issue and the fix in this patch / thanks!

PR #2915, opened 2023-10-07
Add hotkeys option to settings.yml

### LINKED ISSUE #2898: Unable to select Vim-like hotkeys via settings.yml

I'm using docker: searxng/searxng:latest

Docker compose, using searxng/searxng:latest

**What happened?**

Since the removal of the VIM-like hotkeys plugin, there doesn't seem to be a way to select the VIM hotkeys via settings.yml. I'd like to be able to set it in the default config for my instance.

**How To Reproduce**

I have tried the following in settings.yml, both under the `ui` section and at the top-level:

```
ui:
  static_use_hash: true
  # query_in_title: When true, the result page's titles contains the query
  # it decreases the privacy, since the browser can records the page titles.
  query_in_title: true
  # infinite_scroll: When true, automatically loads the next page when scrolling to bottom of the current page.
  infinite_scroll: true
  # Open result links in a new tab by default
  results_on_new_tab: true
  hotkeys: vim

hotkeys: vim
```

**Expected behavior**

The default hotkey setting should be VIM, but it is set to the default SearxNG setting. I can't seem to set the default to 'vim'.

**Screenshots & Logs**

**Additional context**

thezeroalpha:
## What does this PR do?

The change in the hotkey mechanism introduced in 317db5b04f6de919e996df943f7f0a845b8238a9 does not allow configuration via `settings.yml`. This commit adds that functionality. Addresses #2898.

It makes the hotkeys setting admin-configurable.

Add the following in `settings.yml`:

```yaml
ui:
  hotkeys: vim
```

Then, restart SearXNG, and observe that the hotkeys change according to that setting.

## Related issues

Closes #2898.

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -201,6 +201,7 @@ SCHEMA = {
         'infinite_scroll': SettingsValue(bool, False),
         'cache_url': SettingsValue(str, 'https://web.archive.org/web/'),
         'search_on_category_select': SettingsValue(bool, True),
+        'hotkeys': SettingsValue(('default', 'vim'), 'default'),
     },
     'preferences': {
         'lock': SettingsValue(list, []),
--- a/docs/admin/settings/settings_ui.rst
+++ b/docs/admin/settings/settings_ui.rst
@@ -20,6 +20,7 @@
      theme_args:
        simple_style: auto
      search_on_category_select: true
+     hotkeys: default
 
 .. _static_use_hash:
 
@@ -64,3 +65,6 @@
 
 ``search_on_category_select``:
   Perform search immediately if a category selected. Disable to select multiple categories.
+
+``hotkeys``:
+  Hotkeys to use in the search interface: ``default``, ``vim`` (Vim-like).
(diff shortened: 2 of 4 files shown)


Bnyro (approved):

searx/settings_defaults.py
```diff
@@ -201,6 +201,7 @@ def apply_schema(settings, schema, path_list):
         'infinite_scroll': SettingsValue(bool, False),
         'cache_url': SettingsValue(str, 'https://web.archive.org/web/'),
         'search_on_category_select': SettingsValue(bool, True),
+        'hotkeys': SettingsValue(str, 'default'),
```
dalf: You can specify the list of values instead of `str`:
```python
'hotkeys': SettingsValue(('default', 'vim'), 'default'),
```

thezeroalpha: Indeed, that's better. Done!

return42 (approved):
@thezeroalpha thanks a lot for your PR .. works like a charm :)

---

Next time, please create a branch and do not use your master branch for a PR.

PR #2917, opened 2023-10-08
[mod] yacy: use official instance by default and fix crashes

### LINKED ISSUE #2910: YaCy Public Instance
**Working URL to the engine**

https://yacy.searchlab.eu/

Currently it seems support for YaCy is limited to local nodes, using a public instance of YaCy maintained by the team that develops it is a solution for end users who would like to query yacy through their searxng instance of choice without it being set up locally.
**Features of this engine**

YaCy uses a novel peer-to-peer technology which collectively powers a decentralized webcrawler. It does not rely on SEO or a single entity. It is a federated service that prioritzes index by user query and manual feedback.

Hover mouse over 'search interfaces' in top right, at the bottom will be links to numerous types of api calls

general, images, potentially more. Click 'more' next to text and image options at homepage.
**Additional context**

Bnyro:
## What does this PR do?
* add https://yacy.searchlab.eu as default yacy instance
* see #2910

* !yai tree
* !ya hello world

## Related issues
closes #2910

### DIFF
--- a/searx/engines/yacy.py
+++ b/searx/engines/yacy.py
@@ -18,17 +18,27 @@ Configuration
 
 The engine has the following (additional) settings:
 
+- :py:obj:`http_digest_auth_user`
+- :py:obj:`http_digest_auth_pass`
+- :py:obj:`search_mode`
+- :py:obj:`search_type`
+
 .. code:: yaml
 
-   - name: yacy
-     engine: yacy
-     shortcut: ya
-     base_url: http://localhost:8090
-     # Yacy search mode. 'global' or 'local'.
-     search_mode: 'global'
-     number_of_results: 5
-     http_digest_auth_user: ""
-     http_digest_auth_pass: ""
+  - name: yacy
+    engine: yacy
+    categories: general
+    search_type: text
+    base_url: https://yacy.searchlab.eu
+    shortcut: ya
+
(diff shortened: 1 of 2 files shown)


return42:
@Bnyro FYI squashed some nits into your commit --> [diff](https://github.com/searxng/searxng/compare/d3436a522ba544b67e6632d399e092fe3dfb81d5..11c4f9e27216d717b2b4be81f5a9bc199a73135d)

> add https://yacy.searchlab.eu/ as default yacy instance

The service at yacy.searchlab.eu seems broken .. If you enter a term in the search form and send it, after a few minutes you get a response with an error message ... sometimes you get an empty result list ... but the timeouts are more minutes than seconds.

Bnyro:
> The service at yacy.searchlab.eu seems broken .. If you enter a term in the search form and send it, after a few minutes you get a response with an error message ... sometimes you get an empty result list ... but the timeouts are more minutes than seconds.
>
That's strange, I get results within 1 to 2 seconds usually and didn't encounter a timeout yet.

return42:
> That's strange, I get results within 1 to 2 seconds usually and didn't encounter a timeout yet.

May it was a temporary issue .. https://yacy.searchlab.eu/  seems to work now again.

PR #2928, opened 2023-10-11
[fix] imgur - incorrect wikidata id

Austin-Olacsi:
## What does this PR do?
corrects the wikidata id

imgur is using the wikidata id for piped.


Austin-Olacsi:
also side note: we can close the hackernews engine request now.

return42 (approved):
Thanks for the fix :+1: 

> also side note: we can close the hackernews engine request now.

Which request do you mean?

Austin-Olacsi:
i mean issue 1632 can be closed now

PR #2934, opened 2023-10-13
fixing results parsing brave

unixfox:
## What does this PR do?

Fix the results parsing from brave.

Before:

image

After:
image


return42 (approved):
I can confirm: the issue exists and is fixed by this patch / thanks :+1:

PR #2944, opened 2023-10-18
[fix] HTMLParser: undocumented not implemented method

### LINKED ISSUE #2943: Google: Searching "regex exclude" redirects to homepage with an error
2023.10.13+b013cbb4c

searxng-docker

**What happened?**
I searched `regex exclude` and I got redirected to the homepage with a "search error" popping up.

**How To Reproduce**
- Enable the Google engine or **!go regex exclude**. 
- Mobile UI is enabled (not sure if that changes anything but I'll mention it anyways)

**Expected behavior**
- No redirection to homepage with an error. 
- Results on how to exclude keywords from regex.

**Screenshots & Logs**
image

return42:
## What does this PR do?

[fix] HTMLParser: undocumented not implemented method

To be compatible to higher versions (>=py3.10) an error method is implemented which throws an AssertionError exception like the higher Python versions do [3].

In python versions <py3.10 there is an issue with an undocumented method HTMLParser.error() [1][2] that was deprecated in Python 3.4 and removed in Python 3.5.

You need Python <py3.10 AND a result item that is not a HTML but includes a `<` / in #2943 we tested it by this search term:

    :en !go regex exclude

Alternatively in a python interpreter: `./manage pyenv.cmd python`

with the fix ...

```python
>>> from searx.utils import html_to_text
>>> html_to_text(r'regexp: (?<![a-zA-Z]')
'regexp: (?<![a-zA-Z]'
```

without the fix ...

```python
>>> from searx.utils import html_to_text
>>> html_to_text(r'regexp: (?<![a-zA-Z]')
  File "<stdin>", line 1, in <module>
  File "SearXNG/searx/utils.py", line 161, in html_to_text
    s.feed(html_str)
  File "/usr/lib/python3.8/html/parser.py", line 111, in feed
    self.goahead(0)
  File "/usr/lib/python3.8/html/parser.py", line 179, in goahead
    k = self.parse_html_declaration(i)
  File "/usr/lib/python3.8/html/parser.py", line 264, in parse_html_declaration
    return self.parse_marked_section(i)
  File "/usr/lib/python3.8/_markupbase.py", line 159, in parse_marked_section
    self.error('unknown status keyword %r in marked section' % rawdata[i+3:j])
  File "/usr/lib/python3.8/_markupbase.py", line 33, in error
    raise NotImplementedError(
NotImplementedError: subclasses of ParserBase must override error()
```

## Related issues

- [1] https://github.com/python/cpython/issues/76025
- [2] https://bugs.python.org/issue31844
- [3] https://github.com/python/cpython/pull/8562

Closes: https://github.com/searxng/searxng/issues/2943

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -15,6 +15,7 @@ from numbers import Number
 from os.path import splitext, join
 from random import choice
 from html.parser import HTMLParser
+from html import escape
 from urllib.parse import urljoin, urlparse
 from markdown_it import MarkdownIt
 
@@ -88,7 +89,7 @@ class _HTMLTextExtractorException(Exception):
     """Internal exception raised when the HTML is invalid"""
 
 
-class _HTMLTextExtractor(HTMLParser):  # pylint: disable=W0223  # (see https://bugs.python.org/issue31844)
+class _HTMLTextExtractor(HTMLParser):
     """Internal class to extract text from HTML"""
 
     def __init__(self):
@@ -137,6 +138,11 @@ class _HTMLTextExtractor(HTMLParser):  # pylint: disable=W0223  # (see https://b
     def get_text(self):
         return ''.join(self.result).strip()
 
+    def error(self, message):
+        # error handle is needed in <py3.10
+        # https://github.com/python/cpython/pull/8562/files
+        raise AssertionError(message)
+
 
 def html_to_text(html_str: str) -> str:
     """Extract text from a HTML string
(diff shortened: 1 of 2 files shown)


searx/utils.py
```diff
@@ -137,6 +138,11 @@ def handle_entityref(self, name):
     def get_text(self):
         return ''.join(self.result).strip()
 
+    def error(self, message):
+        # error handle is needed in <py3.10
+        # https://github.com/python/cpython/pull/8562/files
+        raise AssertionError(message)
```
dalf, line 144: The tests might not work because of the usage of AssertionError here. A dedicated exception would be better.

return42: Understand .. but AssertionError is what >=py3.10 raises / see code from cpython [implementation](https://github.com/python/cpython/pull/8562/files#diff-7304624ffcc8be1268e5f86ebf1723034477694ce33456dc330f9c03f2d1e84fL134-R132).

BTW: I think the problem in tests is not the raise of AssertionError .. the problem might be that we need an Exception handling of AssertionError (see `except AssertionError` block in line 171)

return42: @dalf After thinking about it again I realized that a ``AssertionError``exception is handled in the test framework like any other exception. But I took the discussion as a reason to add a test for the problematic cases --> [diff](https://github.com/searxng/searxng/compare/01cfc55b5f595c9d5a265646d09035cd934e24f6..89039104a481bde12530b6494986ff71727b5d60)

Bnyro (approved):

dalf (approved):

PR #2946, opened 2023-10-18
[POC] Hostname replace client side

### LINKED ISSUE #1575: Redirect to alternative front-ends option!
image

* I would love to have a similar feature like this picture implemented into searxng. This feature lets you choose FOSS alternative front-ends to replace links with. 
* One thing I would love to be changed from the way librex does it, is to allow us to use custom redirects. (ex. if it doesn't list nitter as an option, i could click "+" to add new row, and type twitter.net in one box, and nitter.net in the other.

### LINKED ISSUE #2351: [Feature] exclude / filter domains from results (client side)
Update from the maintainers: there is a POC based on [Window: localStorage property](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)

-  https://github.com/searxng/searxng/pull/2946

-----

Feature request for excluding/filtering domains from the results on the client side.

Searx related issue: https://github.com/searx/searx/issues/2001

-----

Often when you search, engines like Google and DDG return completely unhelpful results that just shill their own product or don't answer the question or query you were searching for.

The possibility for users to blacklist a website from ever showing up in results again and this gets saved in cookies. Would be great if you can also change the list of blocked websites in preferences.

Maybe instance-wide blocking of websites decided by the admins of the instance. Or a suggested blacklist the user can opt-in, or opt-out of, or modify at will.
Another expansion upon this idea would be a community-maintained blacklist of all SEO/AI-generated/unhelpful websites so we can all put an end to this once and for all.

Bnyro:
## What does this PR do?
- this PR implements a client side version of the `hostname_replace` plugin via JavaScript and [`local storage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
- this does not yet have a UI, I'll continue work on this when I find the time for it and nobody has fundamental arguments against this

- Devtools -> Storage -> Local Storage -> Create a new entry called "rewrites" and set a pattern like in hostname replace as the value, e.g. `(\w+\.)?youtube\.com:piped\.video` -> all YouTube links are converted to Piped links
- do the same but remove the replacement pattern (e.g. use `(\w+\.)?youtube\.com`) -> all YouTube results are filtered out

## Author's checklist
- the python3.12 commit will be dropped before making it ready for review (it's currently needed for me to build SearXNG)

## Related issues
closes #1575
closes #2351


return42:
Yeah cool :rocket: .. may we implement the feature in a dedicated file `rewrite_url.js`, `rewrite_link` or similar?

> and nobody has fundamental arguments against this

from my side you get big fat "GO" :)

Bnyro:
> Yeah cool 🚀 .. may we implement the feature in a dedicated file `rewrite_url.js`, `rewrite_link` or similar?
>
Seems like a good idea, I'll do that when continuing work :+1:

Bnyro:
(rebased only, will continue work on it another day)

Bnyro:
No interest to continue this at the moment.

PR #2950, opened 2023-10-20
[feat] engine: implementation of mdn

Bnyro:
## What does this PR do?
- this PR adds a new engine, the "Mozilla Developer Network" (MDN in short form)

- MSN is well known for informative articles on any topics related to web development such as `HTML`, `CSS`, `JS` and `SVG` (and more), I don't think there's much need to explain it because everyone should be familiar with these docs

* !mdn :has()

### DIFF
--- a/searx/engines/json_engine.py
+++ b/searx/engines/json_engine.py
@@ -8,6 +8,7 @@ from searx.utils import to_string, html_to_text
 
 search_url = None
 url_query = None
+url_prefix = ""
 content_query = None
 title_query = None
 content_html_to_text = False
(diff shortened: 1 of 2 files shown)


searx/engines/json_engine.py
```diff
@@ -129,7 +130,7 @@ def response(resp):
                 content = ""
             results.append(
                 {
-                    'url': to_string(url),
+                    'url': url_prefix + to_string(url),
```
unixfox, line 133: Why is that necessary?

Bnyro: because the url returned by the mozilla mdn api is relative, e.g. `/en-US/docs/Web/CSS/:has`. So we need to prepend a base url.

return42 (approved):
works! .. developer will love it :) .. thanks!

Issue #2844, opened 2023-09-23
Built-in CAPTCHA solver


unixfox:
following the discussion from https://github.com/searxng/searxng/discussions/2458

**Is your feature request related to a problem? Please describe.**
Many engines will give a CAPTCHA in case of too many requests.

I think this is becoming such a big issue on public instances that the only reliable thing would be to add a CAPTCHA automatic solver.

Some engines like qwant and startpage just requires solving a CAPTCHA, then the IP is whitelisted for quite a long time. A CAPTCHA solver would be a huge benefit here.

This will obviously be more difficult for Google as the CAPTCHA is inside a cookie and the validity of a captcha is not very long.

I think it is ok to implement such solution, like a very while ago that was the only viable solution for invidious to work: https://github.com/iv-org/invidious/issues/886

**Describe the solution you'd like**

We should implement respectable services that do not leverage exploited humans in third countries:
- https://capmonster.cloud
- https://captchaai.com
- https://www.capsolver.com/
- https://nopecha.com/
- more... search "captcha solver ai"

For this implementation we have to use redis, so I think it's better to start once we have enforced redis.

LyndonArmitage:
> I think this is becoming such a big issue on public instances that the only reliable thing would be to add a CAPTCHA automatic solver.

Why not pass on the CAPTCHAs to the users directly? At least the simple image based ones. The point of a CAPTCHA is to confirm that a human has made the request, so passing on an image of the capture for the initiator of the search request to solve fulfills this requirement, it bypasses any worries around using exploitative sites and stops public instance maintainers from having to foot the bill for a CAPTCHA solver service.

reCAPTCHA is another kettle of fish entirely, as it does profiling based on how the browser was being used up to the clicking of the checkbox or evaluating of the cookie. The only quick way of solving these without compromising user privacy is the mentioned automated services.

Perhaps there could be a CAPTCHA solver service added to the project? With an implementation that can call out to these services and some implementation that can also hand off to the Searxng user; either throwing up an interactive dialog box when a user's search hits a CAPTCHA or better yet a link to the right of the search results to allow a user an attempt to solve the CAPTCHA.

unixfox:
> Why not pass on the CAPTCHAs to the users directly? At least the simple image based ones.

This has been brought on matrix, and it is possible only for some engines:
- startpage
- qwant: requires to "click" on the captcha, but it's possible to do something with it in order to determine the coordinates to click and send them to qwant backend

ones that will not work:

[...]

PR #2973, opened 2023-10-29
[fix] limit maximum page number of a search query to page 50

### LINKED ISSUE #2972: Bug: google engine --> SearxEngineAccessDeniedException
To reproduce the issue query a high page number / e.g. page no. 500:

      http://127.0.0.1:8888/search?q=!go%20foo&pageno=500

**Technical report**

```
Error
  * Percentage: 100
  * File name: `searx/search/processors/online.py:118`
```

On my instance I have bots going into wild and request a page number 458, here is a request from the bot:

```python
form: {
    'q': 'brown wicker side table',
    'category_general': '1',
    'pageno': '458',   # <--- here !!!
    'language': 'auto',
    'time_range': '',
    'safesearch': '2',
    'theme': 'simple'
}
```

As a result, some engines block, e.g. google responds with an HTTP 403 (Forbidden). As a result a `SearxEngineAccessDeniedException` is thrown and the engine is blocked for a long time (by default 86400 sec / 24h). 

https://github.com/searxng/searxng/blob/b63841f0a1a059a1567d91f8c5c9cb645eb1849d/searx/settings.yml#L48-L60

return42:
To test this PR run a local instance and try to query page 51:

    http://127.0.0.1:8888/search?q=foo&pageno=51

A parameter exception will be raised:

    searx.exceptions.SearxParameterException: Invalid value "51" for parameter pageno

And the client will receive a HTTP 400 (Bad request).

- Closes https://github.com/searxng/searxng/issues/2972

----

FYI: since I merged this PR on my instance I got google back to work (~~and ddg?, qwant? seems to work better / I am not 100% sure but at first glance I have the impression that the limitation of the maximum number of pages also has a positive effect on other engines~~),

### DIFF
--- a/searx/webadapter.py
+++ b/searx/webadapter.py
@@ -45,7 +45,7 @@ def validate_engineref_list(
 
 def parse_pageno(form: Dict[str, str]) -> int:
     pageno_param = form.get('pageno', '1')
-    if not pageno_param.isdigit() or int(pageno_param) < 1:
+    if not pageno_param.isdigit() or int(pageno_param) < 1 or int(pageno_param) > 50:
         raise SearxParameterException('pageno', pageno_param)
     return int(pageno_param)
 


Bnyro (approved):

unixfox:
Review of this PR:

I do not see any improvements in the block of the popular engines. On your instance @return42, qwant, startpage, brave engines are still blocked:

image

Same for all the up-to-date public instances on searx.space.

In fact, I fail to see how this would help in having the engines think that searxng is less of a bot than before this code change. Most engine see that we are doing too many requests and that's it, page number does not matter.

Google and ddg are unblocked because for ddg you fixed it in https://github.com/searxng/searxng/pull/2925 and for Google I have not really seen Google blocked on your instance.

Maybe that was just a coincidence that you restricted the number of page and this unblocked the popular engines? How did you thoroughly test this PR?

About the code change in details, restricting all the users of searxng from doing searches pass a certain page number is for me breaking some part of main feature of searxng: doing the search that you want. We do not know the use case of our users, perhaps some of them want to go pass this number of page for any reason. Remember this apply to all the searxng installation (private AND public)!

Also, in my opinion, any new change that only apply to public instances should **only** be activated when the parameter `server.public_instance` is enabled. This code change has nothing to do in a private instance, there are no bots for private searxng instances.
This is a bot protection feature because it was designed to resist against some bots.

I would be very interested though to know how many requests with a high page number are done per hour. Just to see if it is rare or very common.

return42:
> and for Google I have not really seen Google blocked on your instance.

The bot that does this requests of page numbers > 400 blocked my instance only a day, then I debugged my server found such requests and developed this PR.

> I would be very interested though to know how many requests with a high page number are done per hour. 

I debugged ongoing requests .. from various IPs (that why I assume it was a bot)

>  restricting all the users of searxng from doing searches pass a certain page number is for me breaking some part of main feature 

Yes, but what alternatives do you suggest?

> This is a bot protection feature because it was designed to resist against some bots.

In case of google it is a protection against incorrect usage .. 

AFIK all search services have a maximum page number .. a perfect solution in terms of "protection against incorrect usage" requires a determination of the maximum page number for each engine .. I have decided to limit the number of pages for all engines, maybe it would be better to limit the number of pages only for the online engines ... to determine the individual number of pages for each engine would be a big task for us.

I have no idea which would be the best solution that does not create too much work for us.

I'm asking all reading this comment: suggestions?

unixfox:
> > and for Google I have not really seen Google blocked on your instance.
> 
> The bot that does this requests of page numbers > 400 blocked my instance only a day, then I debugged my server found such requests and developed this PR.

But have you found a correlation between limiting the number of pages and the engines being less blocked? Over multiple days too?

**That's the real question here**, if there is no improvement limiting the number of pages is not useful. Just because some bots are requesting an odd number of pages doesn't mean we need to create a rule against it.

Because if we follow the same logic as this, then hypothetically should we also block the search queries containing the word "banana" because the major visitors that do this query are bots? That doesn't make any sense...

> > restricting all the users of searxng from doing searches pass a certain page number is for me breaking some part of main feature
> 
> Yes, but what alternatives do you suggest?

We just do not do anything about that, that's it. Like explained above.

> > This is a bot protection feature because it was designed to resist against some bots.
> 
> In case of google it is a protection against incorrect usage ..
> 
> I have no idea which would be the best solution that does not create too much work for us.

I think if we want to improve the bot protection or improve the situation about the block of the engines. There are much better solution to implement that I have been proven to work:
- captcha solver: https://github.com/searxng/searxng/issues/2844
- blocking IP addresses: https://github.com/searxng/searxng/pull/2501
- bot protection based on CSS: https://github.com/searxng/internal/issues/31

Your suggestion to block bots that are requesting odd number of pages may block some bots, but these are like droplets into an ocean of bots. **Blocking 0.1% of the bots will not improve the situation about the block of the popular engines.**

return42:
@unixfox test #2979 .. and if you think its OK to revert don't hesitate to merge the roll-back of this PR.

unixfox:
I'm not against any revision of this "idea" in the future, one could also say that it could reduce the load on the public instances if only the bots are doing these odd page number requests.

Like you explained, maybe this bot protection should be per engine. Maybe Google has a hard limit of number of pages that you can load, and thus it doesn't make sense to allow pass a certain number of pages.

For reference:

> Note: Google has always limited the total number of results displayed. I think at one stage it used to be 1,000, then a few years ago it was reduced to 700. It's now about 400. If you want to see more than 400 results, you'll need to use a different search engine — Bing currently stops displaying results at about 700.

- https://support.google.com/websearch/thread/206920626/google-says-search-has-millions-of-results-i-only-get-a-30-pages-of-mayvbe-6-10-results-per-page?hl=en
- https://support.google.com/websearch/thread/199880990/how-do-i-see-all-of-the-searches-beyond-the-usual-limit-of-40-pages-that-that-i-am-shown?hl=en
- https://support.google.com/websearch/thread/8663620/i-can-search-more-than-300-records?hl=en

If we do a quick math, 400 results divided per 20 results per page, then it's 20 pages max for Google.

But for the moment this implementation was rushed and reverting it seems like the better solution until we can improve this idea.

return42:
I don't now the exact numbers .. its a little bit more complex google allows e.g.

  http://127.0.0.1:8888/search?q=!go%20foo&pageno=50

but the result list is empty, but when query

  http://127.0.0.1:8888/search?q=!go%20foo&pageno=500

Google will response with a HTTP 403 (Forbidden) --> in SearXNG 403 will raise a SearxEngineAccessDeniedException and SearXNG will suspend the google engine for 24h (the default suspend time in settings.yml) .. I described it in #2972

unixfox:
For only the Google engine, I'm ok with setting a hard limit of 50 pages.

From my experimentation, with the most popular words in the English dictionary (reference https://en.wikipedia.org/wiki/Most_common_words_in_English). I only get results until the 43th page.

This is a direct relation with what "bluequoll" said on Google forums: https://support.google.com/websearch/thread/206920626/google-says-search-has-millions-of-results-i-only-get-a-30-pages-of-mayvbe-6-10-results-per-page?hl=en

> Note: Google has always limited the total number of results displayed. I think at one stage it used to be 1,000, then a few years ago it was reduced to 700. It's now about 400. If you want to see more than 400 results, you'll need to use a different search engine — Bing currently stops displaying results at about 700.

Last sentence is interesting too because Bing has a hard limit too.

I'm for creating a global issue in searxng where we document the hard limit of each (popular) engine, and we implement the hard limit too in our engine code.

PR #2976, opened 2023-10-30
Check `public_instance` in simple theme

### LINKED ISSUE #2975: Enabling `server.public_instance` without `server.limiter` does not send `link_token` pings
Enabling `server.public_instance` forcefully enables the [`server.limiter`](https://github.com/searxng/searxng/blob/7e2e335dd1e06ea20496028abe58b3ad8b255662/searx/plugins/limiter.py#L27) and limiter [`botdetection.ip_limit.link_token`](https://github.com/searxng/searxng/blob/7e2e335dd1e06ea20496028abe58b3ad8b255662/searx/botdetection/ip_limit.py#L113) features, even if these settings are disabled in their respective settings files. However, because the in-memory values for the associated settings are not updated, other areas that reference these values without also checking `public_instance` will act as if the feature is disabled. This is a particularly important distinction for `link_token` injection in the `simple` theme which checks only `server.limiter`:

https://github.com/searxng/searxng/blob/7e2e335dd1e06ea20496028abe58b3ad8b255662/searx/templates/simple/base.html#L20

This regression results in servers running with `use_default_settings` and `public_instance` detecting every incoming IP as suspicious, which blocks every user for a long time after only a few requests.

I have published a PR that will fix this, but I would highly recommend that instead of checking for `public_instance`, that the values for the desired features be [overridden when loading the config](https://github.com/searxng/searxng/blob/7e2e335dd1e06ea20496028abe58b3ad8b255662/searx/settings_loader.py#L68) so that these kinds of regressions are avoided in the future. This also means that any custom theme templates do not have to be updated to check `public_instance` like the simple theme.

sevmonster:
## What does this PR do?

Adds `link_token` ping injection to simple theme if `server.public_instance` is enabled

Users are blocked as suspicious if `server.limiter` is not explicitly enabled when `server.public_instance` has force-enabled the limiter and the `simple` theme is used

1. Start a server on `master`—or test existing public servers—with `server.limiter` disabled and `server.public_instance` enabled
2. Ping stylesheet is not included in in the response body
3. Do a few searches, get blocked as suspicious
    Logs will show `searx.botdetection.link_token : missing ping` and eventually `BLOCK too many request`
5. Start a server on this PR
6. Ping stylesheet is included in the response body
7. You are longer blocked as suspicious 🎉
    Debug logs will show `token is valid --> True`

## Related issues

Closes #2975

### DIFF
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -17,7 +17,7 @@
   {% else %}
   <link rel="stylesheet" href="{{ url_for('static', filename='css/searxng.min.css') }}" type="text/css" media="screen" />
   {% endif %}
-  {% if get_setting('server.limiter') %}
+  {% if get_setting('server.limiter') or get_setting('server.public_instance') %}
   <link rel="stylesheet" href="{{ url_for('client_token', token=link_token) }}" type="text/css" />
   {% endif %}
   {% block styles %}{% endblock %}


unixfox (approved):
Thank you @sevmonster, we definitively missed that!

unixfox:
@return42 if you have time, could you check if this does indeed cover all the cases where the PING resource need to be loaded when the limiter is activated? If so, could you approve the PR?

Thank you

return42 (approved):

return42:
@return42 if you have time, could you check if this does indeed cover all the cases where the PING resource need to be loaded when the limiter is activated? If so, could you approve the PR?

I think this PR is a solution for an issue we have / my thanks goes to @sevmonster 

I'm going to merge this PR ..

But as @sevmonster already suggested we need a "better" solution ..

> I have published a PR that will fix this, but I would highly recommend that instead of checking for `public_instance`, that the values for the desired features be [overridden when loading the config](https://github.com/searxng/searxng/blob/7e2e335dd1e06ea20496028abe58b3ad8b255662/searx/settings_loader.py#L68) so that these kinds of regressions are avoided in the future. This also means that any custom theme templates do not have to be updated to check `public_instance` like the simple theme.

And we need a better documentation of this feature, currently the doc is spread around e.g. the main documentation is here in an annoncement in the searxng-instances repo https://github.com/searxng/searx-instances/discussions/417

PR #2983, opened 2023-11-02
[fix] brave engine: This site did not provide any description

Austin-Olacsi:
## What does this PR do?

fixes brave "This site did not provide any description" currently this happens with every search result.

now we can have descriptions again.

make run


Austin-Olacsi:
ugh, this doesn't work anymore, because they changed it to desktop-default-regular
closing this PR, because i'll fix this when i add the brave goggles engine.

PR #2990, opened 2023-11-04
[feat] engine: implementation of rotten tomatoes

Bnyro:
## What does this PR do?
- add a new engine, rotten tomatoes

- because everyone loves rotten tomatoes (for its scoring system)

- !rt tom cruise

## Author's note
- pagination would be very complicated for this engine since we would need to use JSON results for that and parse some secret request tokens, which wouldn't be worth it imo


return42 (approved):
Works perfect / thanks again :+1:

PR #2991, opened 2023-11-05
[dev] manage runtime versions with asdf

return42:
## What does this PR do?

Introduce a method to manage versions of runtimes (python, ~~nodejs~~ & shellcheck)

It gives developers the possibility to test SearXNG with different runtime versions (python, ~~nodejs~~, shellcheck) locally.

Build docs and read http://0.0.0.0:8000/dev/rtm_asdf.html

    make docs.clean docs.live

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,2 @@
+python 3.12.0
+shellcheck 0.9.0
--- a/docs/dev/index.rst
+++ b/docs/dev/index.rst
@@ -6,6 +6,7 @@ Developer documentation
    :maxdepth: 2
 
    quickstart
+   rtm_asdf
    contribution_guide
    engines/index
    search_api
--- a/docs/dev/rtm_asdf.rst
+++ b/docs/dev/rtm_asdf.rst
@@ -1,0 +1,121 @@
+==================
+Runtime Management
+==================
+
+The runtimes are managed with asdf and are activated in this project via the
+`.tool-versions <.tool-versions>`_. If you have not yet installed asdf_, then
+chapter :ref:`introduce asdf` may be of help to you.
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+
+Get started
+===========
+
+If you have asdf installed you can install the runtimes of this project by:
+
+.. code:: bash
+
+   $ cd /path/to/searxng
+   $ asdf install          # will install runtimes listed in .tool-versions
+   ...
+
+Manage Versions
+===============
+
+If you want to perform a ``test`` with special runtime versions of nodejs,
+python or shellcheck, you can patch the ``.tool-versions``:
+
+.. code:: diff
+
+   --- a/.tool-versions
+   +++ b/.tool-versions
+   @@ -1,2 +1,2 @@
+   -python 3.12.0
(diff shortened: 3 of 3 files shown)


return42:
@dalf @Bnyro @mrpaulblack do you think its OK adding a ``.tool-versions`` file and the description how runtime versions of python, nodejs, shellcheck can be managed (on developer's host) by asdf?

dalf:
About the versions:  Why the last version of Python, Node?
The Python code might not be compatible with the last version of Python.

Same comment for Node. Also, does that mean that we can remove https://github.com/searxng/searxng/blob/master/.nvmrc ?

Comment on the side: I've seen more than once the usage of `.python-version`. For example in pyenv or [rye](https://github.com/mitsuhiko/rye) (which is really nice but I'm afraid not ready, and [pdm](https://pdm.fming.dev/) does not deal with the python installation).

mrpaulblack (approved):
So if I understand correctly asdf will provide python, Node.js and so on binaries for running the SearXNG source code and its therefor independent of package managers...

Tbh I just prefer to use the python, Node.js and so on version provided by my package manager and do not want to manage those binaries by a third party script in my user space.

Hmm but on the other hand you are right: It makes it possible to test in different python versions and so on locally, but we can also just use docker for that?

So in the end it's another dotfile in the project root, but if you think this can be useful for us in the future I am ok with the change...

return42:
> Comment on the side: I've seen more than once the usage of `.python-version`. For example in pyenv or [rye](https://github.com/mitsuhiko/rye) (which is really nice but I'm afraid not ready, and [pdm](https://pdm.fming.dev/) does not deal with the python installation).

To manage runtime versions, I have tried several solutions inside the python ecosystem e.g. pyenv .. but they have their drawbacks since these solutions are implemented inside the python eco system .. e.g. you need a python interpreter install a python version of the interpreter .. asdf on the other side, is a tool that sits outside of a ecosystem (its implemented in bash and only requires git).

Another point is, while solutions like pyenv and nvm just manage the version of one tool, asdf can mange runtimes of nearly all tools, e.g. shellcheck (for a full list use `$ asdf plugin list all`.

>  does that mean that we can remove https://github.com/searxng/searxng/blob/master/.nvmrc ?

No .. it was not intended (yet) .. may it is better I remove nodjs from the `.tool-versions` file to not confuse ..

> Hmm but on the other hand you are right: It makes it possible to test in different python versions and so on locally, but we can also just use docker for that?

Not sure what you mean with "docker" .. asdf is not a replacement for solutions we have in production .. its just for those of us developers who want or need to test locally with various runtime versions .. e.g. the a new python or shellcheck version that will not be not installed by your systems's package manager (I will remove nodejs from the list, since we managed it already by nvm) ..

return42:
FYI I reduced the dot-file to versioning only python and shellcheck ..

```
python 3.12.0
shellcheck 0.9.0
```

With this setting and asdf installed you can test with up-to-date versions of the tools .. this setup is usefull in LTS distros which package managers often install older versions of theses tools.

Bnyro (approved):
Looks like a pretty handy tool to quickly test compatibility locally, I think that's a pretty good idea!

PR #2992, opened 2023-11-05
Docker: add UWSGI_WORKERS and UWSGI_THREAD environment variables

dalf:
## What does this PR do?

* UWSGI_WORKERS specifies the number of process.
* UWSGI_THREADS specifies the number of threads.

The Docker convention is to specify the whole configuration through environment variables. While not done in SearXNG, these two additional variables allows admins to skip uwsgi.ini

In additional, https://github.com/searxng/preview-environments starts Docker without additional files through searxng-helm-chat. Each instance consumes 1Go of RAM which is a lot especially when there are a lot of instances / pull requests. With `UWSGI_WORKERS=1` each instance consume about 250Mo.

* Easier docker configuration
* Make preview-environments more reliable

* `make docker.build`
* `docker run --rm -ti -p 8080:8080 ${GITHUB_LOGIN}/searxng` should start as many workers as number of CPU cores (same as the current master branch).
* `docker run --rm -ti -p 8080:8080 -e UWSGI_WORKERS=1 ${GITHUB_LOGIN}/searxng` should start one worker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,9 @@ ENV INSTANCE_NAME=searxng \
     MORTY_KEY= \
     MORTY_URL= \
     SEARXNG_SETTINGS_PATH=/etc/searxng/settings.yml \
-    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini
+    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini \
+    UWSGI_WORKERS=%k \
+    UWSGI_THREADS=4
 
 WORKDIR /usr/local/searxng
 
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -4,8 +4,12 @@ uid = searxng
 gid = searxng
 
 # Number of workers (usually CPU count)
-workers = %k
-threads = 4
+# default value: %k (= number of CPU core, see Dockerfile)
+workers = $(UWSGI_WORKERS)
+
+# Number of threads per worker
+# default value: 4 (see Dockerfile)
+threads = $(UWSGI_THREADS)
 
 # The right granted on the created socket
 chmod-socket = 666
--- a/docs/admin/installation-docker.rst
+++ b/docs/admin/installation-docker.rst
@@ -92,6 +92,9 @@ instance using `docker run <https://docs.docker.com/engine/reference/run/>`_:
                 searxng/searxng
    2f998.... # container's ID
 
+The environment variables UWSGI_WORKERS and UWSGI_THREADS overwrite the default
+number of UWSGI processes and UWSGI threads specified in `/etc/searxng/uwsgi.ini`.
+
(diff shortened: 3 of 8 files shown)


unixfox:
Can you add these new envs into the list: https://github.com/searxng/searxng/blob/master/Dockerfile#L18?

Thank you

dalf:
Done.

Important note: this work with new installations, older installation keep the previous version of `uwsgi.ini`. But at least it fixes the issue for the preview-environments.

unixfox (approved):

return42:
Added a commit on top that adds the same environment to the script installation procedure.

One question I have: should we also set ``offload-threads = %k`` from UWSGI_THREAD environment?

AFAIK the ``offload-threads`` setting is only for static files:

- https://uwsgi-docs.readthedocs.io/en/latest/StaticFiles.html#offloading
- https://uwsgi-docs.readthedocs.io/en/latest/OffloadSubsystem.html

dalf:
The number of offload-threads never appeared to be an issue as far I know.
If we drop UWSGI in a near future, we won't have to explain this new env variable is not supported anymore.

return42 (approved):

edevosc2c:
I think these lines can be also removed: https://github.com/searxng/searxng/blob/master/dockerfiles/docker-entrypoint.sh#L65-L70

return42:
> I think these lines can be also removed
> https://github.com/searxng/searxng/blob/76b91a3ef685e288cc54e6a1131760d9c68faff0/dockerfiles/docker-entrypoint.sh#L62-L70

I have seen these lines in my review, but I think we can't  ...

The template is copied with out a replacement from the environment:

https://github.com/searxng/searxng/blob/76b91a3ef685e288cc54e6a1131760d9c68faff0/dockerfiles/docker-entrypoint.sh#L132-L133

Now we have a uwsg.ini with the raw strings `$(UWSGI_WORKERS)` & `$(UWSGI_THREADS)` in ..

https://github.com/searxng/searxng/blob/76b91a3ef685e288cc54e6a1131760d9c68faff0/dockerfiles/uwsgi.ini#L8-L12

After the copy this raw strings will be patched by `patch_uwsgi_settings()` function ..

https://github.com/searxng/searxng/blob/76b91a3ef685e288cc54e6a1131760d9c68faff0/dockerfiles/docker-entrypoint.sh#L132-L134

@dalf: after rethinking about theses lines; its no longer clear to me if it was intended.

When should `$(UWSGI_WORKERS)` & `$(UWSGI_THREADS)` be set .. from the environment when the uwsgi service is started? .. or from the `patch_uwsgi_settings()` when the service is installed?

.. can you have a look / thanks!

unixfox:
I think these lines can be also removed
https://github.com/searxng/searxng/blob/76b91a3ef685e288cc54e6a1131760d9c68faff0/dockerfiles/docker-entrypoint.sh#L62-L70

(reposting with the correct account)

Issue #2982, opened 2023-11-01
[Documentation - Enhancement] All the popular engines number of pages hard limit


unixfox:
This issue was created from the discussion in https://github.com/searxng/searxng/pull/2973#issuecomment-1786868157.

We should document the number of pages hard limit and implement this hard limit inside the engine, not globally. Ideally use the most common words in English to find that: https://en.wikipedia.org/wiki/Most_common_words_in_English

This allows to not send dubious requests to the engine and avoid the engine blocking searxng due to too many unusual search queries.

- Google: 50 pages
   https://support.google.com/websearch/thread/206920626/google-says-search-has-millions-of-results-i-only-get-a-30-pages-of-mayvbe-6-10-results-per-page?hl=en

---------

Changes done https://github.com/searxng/searxng/pull/3002 but will need a refactor, see https://github.com/searxng/searxng/pull/3002#issuecomment-1832192107

Damaj301damaj-lol:
*Query: hi*
Qwant: 5 pages
Trying to do more just results in an improper redirect ( https://lite.qwant.com/?q=hi&p=6 )

Mojeek: 100 pages
Trying to do more will just return to the nearest possible number https://www.mojeek.com/search?q=hi&s=991

Brave: 6 pages
Trying to do more won't return any result and you will most likely be flagged as a bot
https://search.brave.com/search?q=hi&offset=5


Startpage: 9 pages
https://www.startpage.com/sp/search?query=hi&cat=web

Bing: 200 pages (._.)
https://www.bing.com/search?q=hi&first=1991

return42:
@Damaj301damaj-lol thanks a lot for your research .. I added a commit to https://github.com/searxng/searxng/pull/3002 .. can you test the PR?

Damaj301damaj-lol:
> @Damaj301damaj-lol thanks a lot for your research .. I added a commit to https://github.com/searxng/searxng/pull/3002 .. can you test the PR?

Yes, i will pull this PR on my local searxng instance and try

unixfox:
My findings:
- brave 10 pages: https://search.brave.com/search?q=the&offset=8&spellcheck=0
- startpage 18 pages: https://www.startpage.com/sp/search?query=the&cat=web

can replicate the same findings as @Damaj301damaj-lol for the same other engines tested

return42:
I close this issue; the hard page limits we know are documented by the doc-strings of the engines.

[...]

PR #3002, opened 2023-11-13
[mod] add option max_page

### LINKED ISSUE #2972: Bug: google engine --> SearxEngineAccessDeniedException

return42:
Adds new option max_page and set the value of the google engines to 50

Related: https://github.com/searxng/searxng/issues/2982
Closes: https://github.com/searxng/searxng/issues/2972

### DIFF
--- a/searx/engines/brave.py
+++ b/searx/engines/brave.py
@@ -152,6 +152,10 @@ send_accept_language_header = True
 paging = False
 """Brave only supports paging in :py:obj:`brave_category` ``search`` (UI
 category All)."""
+max_page = 10
+"""Tested 9 pages maximum (``&offset=8``), to be save max is set to 10.  Trying
+to do more won't return any result and you will most likely be flagged as a bot.
+"""
 
 safesearch = True
 safesearch_map = {2: 'strict', 1: 'moderate', 0: 'off'}  # cookie: safesearch=off
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -127,6 +127,9 @@ different to the UI language) and a region filter.
 # engine dependent config
 categories = ['general', 'web']
 paging = True
+max_page = 18
+"""Tested 18 pages maximum (argument ``page``), to be save max is set to 20."""
+
 time_range_support = True
 safesearch = True
 
--- a/searx/search/processors/abstract.py
+++ b/searx/search/processors/abstract.py
@@ -150,6 +150,11 @@ class EngineProcessor(ABC):
         if search_query.pageno > 1 and not self.engine.paging:
             return None
 
+        # if max page is reached, skip
+        max_page = self.engine.max_page or settings['search']['max_page']
+        if max_page and max_page < search_query.pageno:
+            return None
+
         # if time_range is not supported, skip
         if search_query.time_range and not self.engine.time_range_support:
             return None
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -169,6 +169,7 @@ SCHEMA = {
             'recaptcha_SearxEngineCaptcha': SettingsValue(numbers.Real, 604800),
         },
         'formats': SettingsValue(list, OUTPUT_FORMATS),
+        'max_page': SettingsValue(int, 0),
     },
     'server': {
         'port': SettingsValue((int, str), 8888, 'SEARXNG_PORT'),
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -45,6 +45,7 @@ ENGINE_DEFAULT_ARGS = {
     "using_tor_proxy": False,
     "send_accept_language_header": False,
     "tokens": [],
+    "max_page": 0,
 }
 # set automatically when an engine does not have any tab category
 DEFAULT_CATEGORY = 'other'
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -59,6 +59,9 @@ about = {
 # engine dependent config
 categories = ['general', 'web']
 paging = True
+max_page = 200
+"""200 pages maximum (``&first=1991``)"""
+
 time_range_support = True
 safesearch = True
 """Bing results are always SFW.  To get NSFW links from bing some age
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -48,6 +48,7 @@ about = {
 # engine dependent config
 categories = ['general', 'web']
 paging = True
+max_page = 50
 time_range_support = True
 safesearch = True
 
--- a/searx/engines/google_images.py
+++ b/searx/engines/google_images.py
@@ -47,6 +47,7 @@ about = {
 # engine dependent config
 categories = ['images', 'web']
 paging = True
+max_page = 50
 time_range_support = True
 safesearch = True
 send_accept_language_header = True
--- a/searx/engines/google_scholar.py
+++ b/searx/engines/google_scholar.py
@@ -51,6 +51,7 @@ about = {
 # engine dependent config
 categories = ['science', 'scientific publications']
 paging = True
+max_page = 50
 language_support = True
 time_range_support = True
 safesearch = False
--- a/searx/engines/google_videos.py
+++ b/searx/engines/google_videos.py
@@ -57,6 +57,7 @@ about = {
 
 categories = ['videos', 'web']
 paging = True
+max_page = 50
 language_support = True
 time_range_support = True
 safesearch = True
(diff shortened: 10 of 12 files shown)


searx/engines/brave.py
```diff
@@ -152,6 +152,10 @@
 paging = False
 """Brave only supports paging in :py:obj:`brave_category` ``search`` (UI
 category All)."""
+max_page = 6
```
unixfox: I can request up to 9 pages on brave: https://search.brave.com/search?q=the&offset=8&spellcheck=0

Could you set max page to 10 instead?

related: https://github.com/searxng/searxng/issues/2982#issuecomment-1810679732

searx/engines/startpage.py
```diff
@@ -127,6 +127,9 @@
 # engine dependent config
 categories = ['general', 'web']
 paging = True
+max_page = 9
```
unixfox: I can request up to 18 pages on startpage: https://www.startpage.com/sp/search?query=the&cat=web
related: https://github.com/searxng/searxng/issues/2982#issuecomment-1810679732

could you set to max 20 pages just to be safe?

unixfox (changes requested):

unixfox:
Thank you for the PR markus, really appreciate.

Hopefully we can have a max number of pages implemented for most of the major engines. I'll continue to do my findings in https://github.com/searxng/searxng/issues/2982

return42:
> Thank you for the PR markus, really appreciate.

:+1: .. BTW fixed the `max_page` as you suggested.

> I'll continue to do my findings

:+1: .. but I think we can merge this PR after your review / to get the google bug fixed (https://github.com/searxng/searxng/issues/2972)

unixfox:
> 👍 .. but I think we can merge this PR after your review / to get the google bug fixed (#2972)

Yes I did not explicitly say in this PR.

unixfox (approved):

Bnyro (approved):
Pretty good solution for the issue if you ask me.

Thank you!

searx/engines/brave.py
```diff
@@ -152,6 +152,10 @@
 paging = False
 """Brave only supports paging in :py:obj:`brave_category` ``search`` (UI
 category All)."""
+max_page = 10
+"""Tested 9 pages maximum (``&offset=8``), to be save max is set to 10.  Trying
+to do more won't return any result and you will most likely be flagged as a bot.
+"""
```
dalf, line 158: There is a contradiction between these two values: 
* `paging=False`
* `max_page=10`.

Or did I miss something?

searx/settings_defaults.py
```diff
@@ -169,6 +169,7 @@ def apply_schema(settings, schema, path_list):
             'recaptcha_SearxEngineCaptcha': SettingsValue(numbers.Real, 604800),
         },
         'formats': SettingsValue(list, OUTPUT_FORMATS),
+        'max_page': SettingsValue(int, 0),
```
dalf, line 172: This default value should be in `settings.yml` too.

searx/search/processors/abstract.py
```diff
@@ -154,6 +154,11 @@ def get_params(self, search_query, engine_category):
         if search_query.time_range and not self.engine.time_range_support:
             return None
 
+        # if max page is reached, skip
+        max_page = self.engine.max_page or settings['search']['max_page']
+        if max_page and max_page < search_query.pageno:
+            return None
```
dalf, line 156: Nitpicking: is it possible to group the two tests about `search_query.pageno` together so it is easy to see there are two tests about search_query.pageno ? (basically just move lines 157 - 160 above line 153).
Nitpicking...

dalf:
I feel the engine settings `paging` is superseded by `max_page` especially if the default value for `max_page` is 0 (as described in `settings_default.py`). For example:
* `max_page=0` means there is no paging (I'm not sure if the page starts counting at 0 or 1?).
* `paging=True` and `max_page=10` means ten pages are supported, but `paging=True` is useless.

==>

A suggestion:
* remove all `paging=True` in the engine, keep only `max_page`.
* as a safety net, `searx/engines/__init__.py` can set `paging=True` for engines which have `max_page > 0`.
* optional: replace easy to do checks about `engine.paging` like here: https://github.com/searxng/searxng/blob/b3d29cb86db4cc1a4e6320016529d1361451e1f1/searx/webapp.py#L1256

If we don't do that, does it make sense to add a check about invalid settings like `paging=False` and `max_page=10`?

unixfox:
@return42 did you had time to look at the comments of @dalf :)?

return42:
> @return42 did you had time to look at the comments of @dalf :)?

Yes but to make it perfect it needs to touch a lot of files .. what will needs lot of tests .. all in all its becomes a "bigger" change and it will take its time .. nothing I can do at the moment (I'm in review of https://github.com/searxng/searxng/pull/2685 what is also a "big" change).

What I would suggest: To have a bugfix for https://github.com/searxng/searxng/issues/2972 we can merge this PR and do the big-change in a follow up PR?

unixfox:
I don't fully understand everything but wouldn't it be possible to at least implement the most important things that @dalf talked about in https://github.com/searxng/searxng/pull/3002#issuecomment-1811183122? Maybe dalf can give some insights about it.
For the less important ones we can look at it later.

I know this PR is important, but I don't want to merge something that is not adequate and never look at it again, staying like this forever. And for the next one that read this new change, he gets confused in how it is implemented.

return42:
> I don't fully understand everything but wouldn't it be possible to at least implement the most important things that @dalf talked about in https://github.com/searxng/searxng/pull/3002#issuecomment-1811183122?

In current master branch a engine has one option related to pages:

```python
paging = True
```

with this PR there comes another option related to pages

```python
max_page = 200
``` 

The pro side of this additional option is, that we have only to patch engines where we want to add a page limit.

@dalf suggested to remove ``paging`` option from all engines --> if an engines supports paging, its enough to define a `max_page`, where `max_page=0` is what `paging=False` today (unlimited pages) .. and if max_page is -1 (?) we have page support without a page limit.

The pro side is, that we do not have two options with an overlapping functionality / the con side is, we need some downward compatible code in the core and we have to touch all engines we have.

--> all in all its becomes a "bigger" task / I don't have time for.

unixfox (approved):
I'm ok with the changes if the refactor is too complex and large to do.

searx/engines/brave.py
return42: This py-module is used for brave (web), brave.video and more brave engines .. only brave (web) supports paging.  By default paging is set to False and for brave (web) its activated in the [settings](https://docs.searxng.org/dev/engines/online/brave.html) .. 

We could discuss moving max_page for brave (web) to the settings as well.  But we could not add a doc-string there, which would also be automatically included in the documentation / this is at least the reason why I decided to implemented the option here on the module level.

return42:
@dalf I fixed the code according to your remarks / see additional commits on the branch

> I'm ok with the changes if the refactor is too complex and large to do.

I would like to leave the decision to @dalf, he has to decide how important this proposal is to him.  Or whether it will be taken up afterwards.  If we do it in this branch, then I would appreciate support in the form of patches applied on this feature branch :+1:

searx/engines/brave.py
dalf: > But we could not add a doc-string there

I don't understand why because the documentation will show the same apparent contradiction.

Actually, `paging=False` in the code and `paging: true` in the configuration are also misleading.

With OOP, we would have one abstract class, and 3 classes, and the configuration would be much cleaner.

Ok. I close this comment.

dalf (approved):
I still think the add a burden to future development of engines.
But okay

return42:
> I still think the add a burden to future development of engines.

I won't disagree with that, but if we only have a compromise in the time available to us, that is still more than having no solution at all. I will merge the PR now to solve the bug in #2972  .. thanks for taking the time and pointing out what can be improved :+1:

PR #3022, opened 2023-11-21
[mod] Capitalize theme names/styles in theme.html view file

### LINKED ISSUE #3014: Use capital letters
In Theme, Style theme, Autocomplete

I think it would look nicer.

obfuscated-loop:
## What does this PR do?

This changes the text within the `preferences/theme.html` view file to capitalize both theme names/styles. The original issue stated engines should also be capitalized however I disagree as I think it's much more logical to leave them as-is due to many engines having capitalized letters across the entire name. Maybe it would be a good idea to link the engines to a 'text-kind' name using a dictionary?

It's not an incredibly important PR by any means, but it also something I have noticed personally and is a little bit of an eyesore.

Goto `Preferences > User Interface` and you will see the capitalized names within the two related selectors/dropdowns.

## Related issues

Closes #3014

### DIFF
--- a/searx/templates/simple/preferences/theme.html
+++ b/searx/templates/simple/preferences/theme.html
@@ -5,7 +5,7 @@
       {%- for name in themes -%}
         <option value="{{ name }}"
                 {%- if name == theme %} selected="selected"{%- endif -%}>
-                {{- name -}}
+                {{- name | capitalize -}}
         </option>
       {%- endfor -%}
     </select>{{- '' -}}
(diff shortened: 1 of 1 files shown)


obfuscated-loop:
You may have to leave this one til last to merge as it's wriggled its way into some of my other PRs

return42 (changes requested):
Thanks for the PR / LGTM :+1: .. one nit I have:

In #3014 the opener also asked for "autocomplete" drop down box .. can you add this also to your PR / thanks!

obfuscated-loop:
> Thanks for the PR / LGTM 👍 .. one nit I have:
> 
> In #3014 the opener also asked for "autocomplete" drop down box .. can you add this also to your PR / thanks!

Hi, I left this out intentionally as I thought it makes a lot more sense to leave the engines all lowercase. Think about such engines as DuckDuckGo which engines' name is `duckduckgo` - it would display as `Duckduckgo`, I.E not grammatically correct. I can change it to capitalized if you really think it would look better but I personally think it would look a bit rough like this.

return42 (approved):

unixfox (approved):
don't have opinion on this, why not

PR #3023, opened 2023-11-21
[mod] Change no results error message based on page number

### LINKED ISSUE #2815: Show error message about no-results only on the first page
After using a while 

- https://github.com/searxng/searxng/pull/2808

I think its better when the error message is only shown on the first page.

We have many engines that do support only one or a few of result pages, by instance qwant does have 5 result pages, when you click on the sixth page the error message is shown.

It might be better when we show a "no more results" message on no-result when page number is greater than one.

obfuscated-loop:
## Dependency
This PR requires https://github.com/searxng/searxng/pull/3024 to be merged before.

## What does this PR do?

Currently, when the page number is above 1 and the results dictionary is empty, a message returns saying 'No results found'. It was proposed to change this message to something that more aligns with the actual error.

It makes more sense to the user, and less inconspicuous errors.

Search for something completely obscure or use a broken engine etc - this is the original message. The best way of getting the 'No more results' message, in my experience, is again using an obscure search term and then click through the page numbers at the bottom.

## Related issues

Closes #2815

### DIFF
--- a/searx/templates/simple/messages/no_results.html
+++ b/searx/templates/simple/messages/no_results.html
@@ -1,7 +1,21 @@
 {% from 'simple/icons.html' import icon_big %}
 <div class="dialog-error-block" role="alert">
-  <p><strong>{{ _('Sorry!') }}</strong></p>
-  <p>{{ _("No results were found. You can try to:") }}</p>
+  <p>
+    <strong>
+      {%- if pageno == 1 -%}
+        {{ _('Sorry!') }}
+      {%- else -%}
+        {{ _("Oops!") }}
+      {%- endif -%}
+    </strong>
+  </p>
+  <p>
+    {%- if pageno == 1 -%}
+      {{ _("No results were found. You can try to:") }}
+    {%- else -%}
+      {{ _("Looks like there isn't any more results. You can try to:") }}
+    {%- endif -%}
+  </p>
   <ul>
     <li>{{ _("Refresh the page.") }}</li>
     <li>{{ _("Search for another query or select another category (above).") }}</li>
--- a/searx/templates/simple/messages/no_results.html
+++ b/searx/templates/simple/messages/no_results.html
@@ -1,25 +1,28 @@
 {% from 'simple/icons.html' import icon_big %}
-<div class="dialog-error-block" role="alert">
-  <p>
+<div class="dialog-error-block" role="alert">{{- '' -}}
+  <p>{{- '' -}}
     <strong>
       {%- if pageno == 1 -%}
         {{ _('Sorry!') }}
-      {%- else -%}
-        {{ _("Oops!") }}
       {%- endif -%}
-    </strong>
-  </p>
+    </strong>{{- '' -}}
+  </p>{{- '' -}}
   <p>
     {%- if pageno == 1 -%}
       {{ _("No results were found. You can try to:") }}
     {%- else -%}
-      {{ _("Looks like there isn't any more results. You can try to:") }}
+      {{ _("There are no more results. You can try to:") }}
     {%- endif -%}
-  </p>
+  </p>{{- '' -}}
   <ul>
-    <li>{{ _("Refresh the page.") }}</li>
-    <li>{{ _("Search for another query or select another category (above).") }}</li>
-    <li>{{ _("Change the search engine used in the preferences:") }} <a href="/preferences">/preferences</a></li>
-    <li>{{ _("Switch to another instance:") }} <a href="https://searx.space">https://searx.space</a></li>
+    {%- if pageno == 1 -%}
+      <li>{{ _("Refresh the page.") }}</li>{{- '' -}}
+      <li>{{ _("Search for another query or select another category (above).") }}</li>{{- '' -}}
+      <li>{{ _("Change the search engine used in the preferences:") }} <a href="/preferences">/preferences</a></li>{{- '' -}}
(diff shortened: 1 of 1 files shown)


return42:
@obfuscated-loop FYI: I squashed the five commits into one.

return42 (approved):
Thanks a lot for your PR .. LGTM :+1:

I pushed 668369811 on top of your branch.  Not sure, but I think its enough to show a "There are no more results.":

image

Just because there are no other results than those on the previous pages, we should not make users feel insecure or refer them to other search engines.  We don't do that in the infinite results list either.

Its what I suggest: but the decision is not up to me.

unixfox:
I think keeping some help for the user would be great.
I think something like this would be ideal:

```
There are no more results. You can try to:

<ul>
<li>Search for another query or select another category (above).</li>
<li>Go back to the previous page using the previous page button.</li>
</ul>
```

Maybe the box shouldn't be red? But another color? Like green or blue?

return42:
@unixfox I changed to what you suggested ..

image

Please take into account: "Go back to the previous page using the previous page button." requires https://github.com/searxng/searxng/pull/3024 first to be merged.

unixfox:
> @unixfox I changed to what you suggested ..
> image

Thank you :)

> Please take into account: "Go back to the previous page using the previous page button." requires #3024 first to be merged.

Yes I forgot to say that this will depend on the other PR :)

-------

By the way, please don't force push, it's not necessary. For instance here it would have been great to see the diff between before my review and after.

searx/templates/simple/messages/no_results.html
```diff
@@ -1,11 +1,28 @@
...
+      {%- if pageno == 1 -%}
+        {{ _('Sorry!') }}
+      {%- endif -%}
+    </strong>{{- '' -}}
+  </p>{{- '' -}}
+  <p>
+    {%- if pageno == 1 -%}
+      {{ _("No results were found. You can try to:") }}
+    {%- else -%}
+      {{ _("There are no more results.") }}
```
unixfox: You forgot the "You can try to: " part :).
```suggestion
      {{ _("There are no more results. You can try to:") }}
```

unixfox (changes requested):

searx/templates/simple/messages/no_results.html
return42: > You forgot the "You can try to: " part :).

:facepalm: .. thanks for pointing out .. fixed in https://github.com/searxng/searxng/compare/ee0caf313079cb94c40808c50284a2765b77b624..875d83674d1d17b2996b7e7c46c9c3eb9afa0602

unixfox:
Waiting for the input from @obfuscated-loop for the new text section.

By the way what would it take to make the box in another color? New CSS rules? Technically this message is not an error per se.

return42:
> Maybe the box shouldn't be red? But another color? Like green or blue?
> By the way what would it take to make the box in another color? New CSS rules? Technically this message is not an error per se.

Thats true! .. we need some CSS to define INFO messages (AFAIK at this time we have only a style for error messages).  Not sure its a task of this PR to define & layout such a INFO-box.

This PR is good enough to fix the issue https://github.com/searxng/searxng/issues/2815 .. layout should be discussed in a different PR / I think.

unixfox:
We are not in a rush to fix https://github.com/searxng/searxng/issues/2815. If we can add this new "info" box in another PR then revisit this one to implement the new box color then I'm in favor of doing it.

return42:
> If we can add this new "info" box in another PR then revisit this one to implement the new box color then I'm in favor of doing it.

Yes, this was my intention .. implement a UI component for messages (that are not error messages) and then go through the application to see where we better use the new UI component instead of the current red box.  The message box in this PR will be one candidate, may we find more ..

obfuscated-loop:
I still think the message should be recognizable as an error of sorts, a blue or green box usually is used to show information with a successful request. Perhaps amber/orange would be a good color to represent a non-hard-error related message? With red signifying an error moreso... Don't know really, I have no problem with the text change but I think there should also be another list item that states `Verify you haven't been blocked from the engine` or something of the likes to signify that the engine itself is just returning no more results, and not a failure within SearXNG..

obfuscated-loop:
> > If we can add this new "info" box in another PR then revisit this one to implement the new box color then I'm in favor of doing it.
> 
> Yes, this was my intention .. implement a UI component for messages (that are not error messages) and then go through the application to see where we better use the new UI component instead of the current red box. The message box in this PR will be one candidate, may we find more ..

Bootstrap has always done a similar thing with alerts, see here: https://getbootstrap.com/docs/5.0/components/alerts/. I think it would be worth recreating this CSS component.

return42:
> This PR requires https://github.com/searxng/searxng/pull/3024 to be merged before.

FYI: rebased on current master to get #3024 in this branch.

image

return42:
> Don't know really, I have no problem with the text change but I think there should also be another list item that states `Verify you haven't been blocked from the engine` or something of the likes to signify that the engine itself is just returning no more results, and not a failure within SearXNG..

Hm .. not sure we confuse the user .. if the user decided to use one engine and the engine is blocked he already gets a message on the first page ..

image

When multiple engines are used (what is common) the engines that are not blocked will give results until there are no more results and then the massage of this PR is shown on a follow up page (that is not the first page).

image

I don't have a clue how we can further improve it with a message like `Verify you haven't been blocked from the engine` ..

I think this PR is a huge improvement to what we have now .. lets merge this PR and if there are more options to improve the messages we can do it in a follow up PR.

@obfuscated-loop thanks a lot for your work to improve SearXNG :+1:

PR #3024, opened 2023-11-22
[mod] Retain page numbers even when there are no results

### LINKED ISSUE #3007: Keep the number of pages display when no results are found when page number is superior to 1
If you go to the next page but no engines return results, you lose the ability to easily come back to the previous page.

Last working page:
image

When you click on the next page (that do not return any results):
image

For a better user experience, it would be ideal to still display the number of pages but only keep the button "previous page".

Something like this:
image

This display would only be available when the user is on another page than the first page because it wouldn't make sense on the first page.

**Additional context**

obfuscated-loop:
## What does this PR do?

It makes sure the page numbers (pagination) is retained, even when there are 0 results. It also hides the *Next page* button but retains the *Previous page* button.

The current behaviour for page numbers is to hide if there is no results - this can be painful for the user, especially when using only POST requests and they have to hit back on their browser and re-send the previous POST request.

Go through the page numbers until there are no results, you will see the pagination stay there, along with the *Next page* button not being outputted.

## Related issues

Closes #3007

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -232,7 +232,7 @@ class ResultContainer:
         if engine_name in engines:
             histogram_observe(standard_result_count, 'engine', engine_name, 'result', 'count')
 
-        if not self.paging and standard_result_count > 0 and engine_name in engines and engines[engine_name].paging:
+        if not self.paging and engine_name in engines and engines[engine_name].paging:
             self.paging = True
 
     def _merge_infobox(self, infobox):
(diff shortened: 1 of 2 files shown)


obfuscated-loop:
I should add: as per the spec from the created issue, the pagination is *not* shown on the first page if there are 0 results.

return42 (approved):
LGTM :+1:

unixfox (approved):
awesome thank you

PR #3025, opened 2023-11-22
[fix] Remove anchor tag outline in image results

obfuscated-loop:
## What does this PR do?

This removes a CSS bug regarding the image results - as you can see from the example below, if using the arrow keys to navigate, then the anchor tag is assigned `:focus-visible` which in-turn renders an outline of the link. The PR applies a `outline: none;` rule to the anchor tag to combat it. 

*This behaviour is apparent in Gecko and Chromium - albeit they have a different rendering process for outlines.* 

image
image

It fixes an eyesore regarding viewing the images in the gallery view - something of which the user most likely notices.

Search for images, and use the arrow keys to navigate and compare with existing instances.


dalf:
Can you update your commit to keep only the change to searx/static/themes/simple/src/less/style.less
and then call `make static.build.commit` to build and commit the built files
?

See https://docs.searxng.org/dev/quickstart.html

obfuscated-loop:
> Can you update your commit to keep only the change to searx/static/themes/simple/src/less/style.less and then call `make static.build.commit` to build and commit the built files ?
> 
> See https://docs.searxng.org/dev/quickstart.html

Epic fail by me, I promise I read the quickstart when Markus sent me it :sob: ... Just not well enough clearly. I will do this now.

obfuscated-loop:
I've royally F'd up my local Git branch. I might have to join the Matrix for help hahaha

mrpaulblack (approved):
LGTM :+1:

return42 (approved):
@obfuscated-loop works like a charm / thanks a lot for this patch :+1: 

FYI: I squashed the 3 commits into one / hope that is OK for you .. going to merge after the rebase has passed the CI tests.

unixfox (commented):
great thank you

PR #3026, opened 2023-11-22
[mod] Move static preference URLs back to General in Preferences and apply CSS

### LINKED ISSUE #2940: Move back Search URL of the currently saved preferences to the General tab
`Search URL of the currently saved preferences` is hidden under the "cookies" tab.

image

It has always been under the general tab for years on oscar: 

image

I'm not sure when that got changed, but this feature, which is awesome for people that do not want to use cookies. If it only displayed under the Cookies tab, hardly anybody will notice about this feature.

I propose to move it back to the General tab for more eyes on it.

The point that people could be afraid of this URL is not valid because people that dig into the preferences are already the same people that are a bit more experienced with IT/SearXNG.

**Additional context**

obfuscated-loop:
## What does this PR do?

This moves the static/saved preference URLs to the *General* tab. See below:
image
image

A lot of people, including myself, do not enable or remember cookies across browsing sessions. The static URL allows for people to set-up their browser in a way where they can search with their desired preferences without ever using any type of local storage (apart from storing the URL and parameters as say a bookmark or search engine within their browser). Therefore, it should not be placed in the *Cookies* tab as it may go unseen.

Goto `Preferences > General` and look at the bottom of the page - I have also implemented a dark-mode background for the card.

## Related issues

Closes #2940


return42 (changes requested):
To clean up not related patches and fix conflicts I rebased this branch.
 
I think it makes sense to isolate `<div class="saved_prefs_url_card">` in a dedicated `url_card.html` file.

I personally think that the URL belongs thematically in the "cookies" tab, the URL is not related to settings in the "general" tab .. but the decision is not up to me :)

unixfox (approved):

return42:
@obfuscated-loop to fix conflicts with current masters I rebased this branch.

return42:
From the technical point of view this PR LGTM .. but I have a doubt we really want to merge this change in the UI ..

@dalf @unixfox @mrpaulblack  from my comment above ..

> I personally think that the URL belongs thematically in the "cookies" tab, the URL is not related to settings in the "general" tab .. but the decision is not up to me :)

I see some drawbacks in this change 

1. the users I know will be confused by the cryptic URLs in the general tab / they don't expect such "technical stuff" on a tab where you set more general things like the languages or categories
2. the color highlighting gives the impression that the user has to deal with it
3. to complete the irritation there are sentences at the bottom like "These settings are stored in your cookies .. "
4. The second URL (*"URL to restore your preferences in another browser:"*) **creates a cookie** and is related to the topic "COOKIES" .. but is not shown in the tab with that name.

I'm sorry but from my side I want to vote against this change .. I would therefore like to ask you to decide on the further progress and merge of this PR.






grafik

unixfox:
Has always worked great like this for 7 years for SearX (first appeared in https://github.com/searx/searx/commit/38d6ba4066a474c1b13e7ccb6f9ea92b43702a4a) and nobody ever complained.

All the claims are valid for newbies users BUT newbies users do not use preferences. They just use the default settings and that's it.

Like I said, if we hide this very useful feature under "cookies" I'm pretty sure that users that do not store cookies permanently won't discover this feature. And this represents a huge chunk of our user base, [Cookie AutoDelete](https://addons.mozilla.org/en-US/firefox/addon/cookie-autodelete/) is one of the most used extensions on the Mozilla addons store.

If one day we get complaints, I'm ok to revisit this feature. Until then, I would like to "revert" to the place where this setting was originally placed before we forked SearX.

unixfox:
Could you please rebase the PR @obfuscated-loop :)?

I'll merge it as soon as it's rebased. Thank you.

Bnyro:
I agree 100% with @return42 on this PR. Additionally we should keep in mind that
1. The general tab already has enough other options that fill the space
2. Preference restore URLs are rather for advanced than casual users -> they know where to find what they need

In my opinion, the Cookies tab would be well suited for being renamed to "Backup and Restore" to be even easier to understand for new users.

Issue #1673, opened 2022-08-16
searxng marginalia integration request shotgun


vlofgren:
Hello. I run search.marginalia.nu. I noticed it was recently added to searxng ( https://github.com/searxng/searxng/pull/1627 ), and since then I've noticed some strange behavior in the api logs.

When the rate limit is exceeded (and the public key only allows 15 requests per minute globally, so not a lot; let me know if you want me to hook you up with a key by the way, the public key is intended for sampling and the odd python script), the api server sends a 503. Maybe there is some bad interaction with searxng that causes it to interprets this as a signal to retry? 

I've attached a redacted sample of the sort of curious traffic behavior I'm seeing. I'd love to be able to integrate with this project, but my search engine can't respond to queries at this rate, you'll get  rate limited immediately (which is unnecessary because it's the exact same query 20 times in a row).  It's not really a problem per se on my end, I just thought I would let you know this is happening.

```
INFO  2022-08-16 23:49:48,327 qtp1892704146-12749  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:45,766 qtp1892704146-12749  ApiService          : RSP 200

INFO  2022-08-16 23:50:20,351 qtp1892704146-12721  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:20,591 qtp1892704146-12749  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:20,635 qtp1892704146-12721  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:20,675 qtp1892704146-12749  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:20,718 qtp1892704146-12721  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
INFO  2022-08-16 23:50:20,762 qtp1892704146-12749  ApiService          : PUBLIC #b6f70eb3: GET /public/api/public/search/[FIRST QUERY]?index=4&count=20
[...]

PR #3036, opened 2023-11-28
[feat] engine: implementation of marginalia

### LINKED ISSUE #3034: Engine request: Marginalia
**Working URL to the engine**

https://search.marginalia.nu

This search engine has a lot of relatively unknown websites, and might help some people to use SearXNG as how Google was at the beginning (being able to actually search for things properly)

**Features of this engine**

Keyword-based results
Special Keywords
The engine searches exactly for the keywords you want.
Not tailored for big websites, can find every niche website

General, News, Science

**Additional context**

The engine might be hard to use, because of its own advantage: Not having large websites on top.
But if you use it to its maximum potential, you can even find very mysterious / obscure things (by that, I mean that they are not usually found anywhere).

Bnyro:
## What does this PR do?
* implement a new open source search engine for general results, marginalia

* explained in #3034

* !ma foo

## Author's checklist
* there's an official API, but the rate limit is very strict - I got ratelimited while implementing the engine (round about 10 requests in 5mins), so I don't think the API is worth it

## Related issues
closes #3034

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1857,6 +1857,23 @@ engines:
   #   timeout: 5.0
   #   disabled: true
 
+  - name: marginalia
+    shortcut: ma
+    engine: xpath
+    categories: general
+    search_url: https://search.marginalia.nu/search?query={query}
+    results_xpath: //section[contains(@class, "search-result")]
+    url_xpath: ./div[contains(@class, "url")]/a/@href
+    title_xpath: ./h2
+    content_xpath: ./p[contains(@class, "description")]/text()
+    disabled: true
+    about:
+      website: https://search.marginalia.nu
+      official_api_documentation: https://www.marginalia.nu/marginalia-search/api/
+      use_official_api: true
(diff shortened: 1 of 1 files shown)


Bnyro:
Oh, I didn't read we removed the engine due to too much traffic. I guess we should comment it out?

return42:
@Bnyro There is an upstream story, please read https://github.com/searxng/searxng/issues/1673 

The maintainer @vlofgren said:

> I've attached a redacted sample of the sort of curious traffic behavior I'm seeing. I'd love to be able to integrate with this project, 

which is why I assume that there is an interest, but we should closely coordinate any integration (of whatever kind) with the operator ... unfortunately I haven't had time to do this myself so far, if you are interested in continuing this PR, I would ask you to coordinate with @vlofgren.

What must not be repeated under any circumstances is that we build up a constellation in which we run the risk of penetrating marginalia again with our various active instances.

Bnyro:
We should close the issue too then.

PR #3045, opened 2023-12-02
Setup.py: fix infopage location

jwijenbergh:
## What does this PR do?

Previously the infopage was under searx/info directory, it seems like this was never changed

About page gives a 404. This doesn't affect the docker image because that just copies over the whole searx directory.

Build a wheel and see that it doesn't contain the info page

## Related issues

Don't think this has been reported yet

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
 """Installer for SearXNG package."""
 
-from setuptools import setup
-from setuptools import find_packages
+from setuptools import setup, find_packages
 
 from searx.version import VERSION_TAG, GIT_URL
 from searx import get_setting
@@ -18,7 +17,7 @@ with open('requirements-dev.txt') as f:
 
 setup(
     name='searxng',
-    python_requires=">=3.7",
+    python_requires=">=3.8",
     version=VERSION_TAG,
     description="A privacy-respecting, hackable metasearch engine",
     long_description=long_description,
@@ -28,7 +27,6 @@ setup(
         "Issue tracker": get_setting('brand.issue_url')
     },
     classifiers=[
-        "Development Status :: 4 - Beta",
         "Programming Language :: Python",
         "Topic :: Internet",
         "Topic :: Internet :: WWW/HTTP :: HTTP Servers",
@@ -39,40 +37,40 @@ setup(
     author='SearXNG dev team',
     author_email='contact@searxng.org',
     license='GNU Affero General Public License',
-    packages=find_packages(exclude=["tests*", "searxng_extra"]),
-    zip_safe=False,
+    packages=find_packages(
+        include=[
+            'searx', 'searx.*', 'searx.*.*', 'searx.*.*.*',
+        ]
+    ),
     install_requires=requirements,
     extras_require={
         'test': dev_requirements
     },
     entry_points={
         'console_scripts': [
-            'searx-run = searx.webapp:run',
-            'searx-checker = searx.search.checker.__main__:main'
+            'searxng-run = searx.webapp:run',
+            'searxng-checker = searx.search.checker.__main__:main'
         ]
     },
     package_data={
         'searx': [
             'settings.yml',
-            '../README.rst',
-            '../requirements.txt',
-            '../requirements-dev.txt',
-            'data/*',
-            'info/*',
-            'info/*/*',
-            'plugins/*/*',
-            'static/*.*',
-            'static/*/*.*',
-            'static/*/*/*.*',
-            'static/*/*/*/*.*',
-            'static/*/*/*/*/*.*',
-            'templates/*/*.*',
-            'templates/*/*/*.*',
-            'tests/*',
-            'tests/*/*',
-            'tests/*/*/*',
-            'translations/*/*/*'
+            '*.toml',
+            '*.msg',
+            'search/checker/scheduler.lua',
+            'data/*.json',
+            'data/*.txt',
+            'data/*.ftz',
+            'infopage/*/*',
+            'static/themes/simple/css/*',
+            'static/themes/simple/css/*/*',
+            'static/themes/simple/img/*',
+            'static/themes/simple/js/*',
+            'templates/*/*',
+            'templates/*/*/*',
+            'translations/*',
+            'translations/*/*',
+            'translations/*/*/*',
         ],
     },
-
 )


return42 (commented):
> Don't think this has been reported yet

Since we do not build the searx package and some of the items (e.g. `test`) do not exists, I assume we can drop `package_data` (its a ten year old leftover :) ..

https://github.com/searxng/searxng/blob/8a4104b9922cbe4c33864b5c0de02e88d3c9665d/setup.py#L54-L75

jwijenbergh:
> > Don't think this has been reported yet
> 
> Since we do not build the searx package and some of the items (e.g. `test`) do not exists, I assume we can drop `package_data` (its a ten year old leftover :) ..
> 
> https://github.com/searxng/searxng/blob/8a4104b9922cbe4c33864b5c0de02e88d3c9665d/setup.py#L54-L75

Hmmm the reason I found out is because I use the package on nixos. That would thus break packages like that as those files would then not be packaged at all. But I understand if no one has the time to maintain it

return42:
One big problem I see is the name collision with the existing package https://pypi.org/project/searx/ .. which, as far I can say has never worked.

dalf:
> One big problem I see is the name collision with the existing package https://pypi.org/project/searx/ .. which, as far I can say has never worked.

It has worked, a long time ago.

I don't think name collision is a problem: if we published a package on pypi it should be named searxng, but there is no problem to include a module named searx inside this searxng package. This happen for some packages with C codes but without wheels, someone else publish another package with a different name but same module.

Exemple:
* https://pypi.org/project/fasttext/
* https://pypi.org/project/fasttext-wheel/
* or even https://pypi.org/project/fasttext-predict/

All these packages create a module named `fasttext`.

return42:
> Hmmm the reason I found out is because I use the package on nixos. 

AFAIK searx package has been dropped on NixOS

- https://github.com/NixOS/nixpkgs/pull/258813

> That would thus break packages like that as those files would then not be packaged at all. 

I'm not an expert, but I don't think NixOS installs the python "searx" package from this repo here into the normal python packages .. if you have issues with the origin NixOS package maintained by the NixOS team @SuperSandro2000 please open a issue in the NixOS repository.

>  if we published a package on pypi it should be named searxng,

@dalf OK thanks for clarify / I think "pypi searxng packaging" will be a new feature and is not in the scope of this PR or what @jwijenbergh wanted to address .. I'm going to close this PR since it does not really solve an issue we have / as long we do not have a searxng package.

jwijenbergh:
> > Hmmm the reason I found out is because I use the package on nixos.
> 
> AFAIK searx package has been dropped on NixOS

Yes the searx package, not the searxng package :/ https://search.nixos.org/packages?channel=23.11&from=0&size=50&sort=relevance&type=packages&query=searx

>     * [ searx: drop  NixOS/nixpkgs#258813](https://github.com/NixOS/nixpkgs/pull/258813)
> 
> 
> > That would thus break packages like that as those files would then not be packaged at all.
> 
> I'm not an expert, but I don't think NixOS installs the python "searx" package from this repo here into the normal python packages .. if you have issues with the origin NixOS package maintained by the NixOS team @SuperSandro2000 please open a issue in the NixOS repository.
> 

It uses the setup.py to determine the package contents... Now we have to patch the package. So I thought the better fix was trying here in upstream first.

FWIW this patch makes me have a fully functioning package. But I understand if you want to close it, if so I would think the best way forward would be to drop the setup.py altogether

return42:
> FWIW this patch makes me have a fully functioning package. But I understand if you want to close it, if so I would think the best way forward would be to drop the setup.py altogether

Not sure what to do next / first of all, I am reopening this PR to keep the discussion alive ..

I have not yet dealt with "how and whether" we should build a Python package ... for the following reason:

SearXNG is a rolling release / see [Migrate and stay tuned!](https://docs.searxng.org/admin/update-searxng.html#id5).

Typically it is installed by setting up a pyenv ...

https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/utils/searxng.sh#L490-L494

... and installing it (in the pyenv) directly from the git clone.  The installation from the shell script is a little bit more complex since we create a dedicated user for the SearXNG process and more .. but from the python's point of view, the installation is just a simple "developer" installation (therefore we need a `setup.py`):

https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/utils/searxng.sh#L508

Since the "developer" installation is nothing more than a link, this installation does not require any packaging instructions and we can deploy updates by a git workflow.

https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/utils/searxng.sh#L570-L577

Since we have not published any other installation methods so far, no one has ever taken care of the packaging ... Packaging always requires release management and we don't have that in SearXNG because, as mentioned above, it is a rolling release.

I don't want to stand in anyone's way, but the only thing I would like to avoid is that the "back door" of packaging gives rise to the claim that SearXNG has releases.

I don't know how best to marry the rolling releases with the release management of a packager-manager like Linux distributors use (apt, pacman, nix ..).

Suggestions?

----

But think what we can (or should) do in any case is to delete the `package_data` completely first, because in this form it is wrong in any case.

https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/setup.py#L54-L76

SuperSandro2000:
> I have not yet dealt with "how and whether" we should build a Python package ... for the following reason:

It is already a python package. You can do ``python setup.py install`` and it mostly works.

> SearXNG is a rolling release / see [Migrate and stay tuned!](https://docs.searxng.org/admin/update-searxng.html#id5).

From a NixOS perspective we don't really care. We just update to latest master branch once in a while and pick up the changes.

> Typically it is installed by setting up a pyenv ...

We are more or less doing the same thing but the python installation is managed by nix, so people don't have to worry about state and things are reproducible and not depended on what was previously installed.

> The installation from the shell script is a little bit more complex since we create a dedicated user for the SearXNG process and more

For NixOS that is all done in the searx(ng) module. We don't want packages to change system wide related things, so that the OS knows what happens and not a giant mess is created.

> but from the python's point of view, the installation is just a simple "developer" installation (therefore we need a setup.py):

Just take away the ``-e`` and the package is done. :)

I don't think symlinks wouldn't work for nixos out of the box. We would need to rewrite them and I would rather avoid that.

> Since we have not published any other installation methods so far, no one has ever taken care of the packaging ... Packaging always requires release management and we don't have that in SearXNG because, as mentioned above, it is a rolling release.

Packaging does not require a release workflow. We just take from master without any tags or uploads to pypi. Actually we usually prefer git downloads because the tarballs on pypi usually miss tests, sometimes source files for generated files or are not exactly created from the tag or are else somehow modified.

> I don't want to stand in anyone's way, but the only thing I would like to avoid is that the "back door" of packaging gives rise to the claim that SearXNG has releases.

The current version in nixpkgs is unstable-2023-10-31, so taken from the date of the commit.

> But think what we can (or should) do in any case is to delete the `package_data` completely first, because in this form it is wrong in any case.

Please don't. That's is truly all we need and We would revert it immediately anyway.

return42:
@SuperSandro2000 thanks a lot for your detailed feedback, given me some insights of NixOS packager :+1: 

> Just take away the ``-e`` and the package is done. :)
> I don't think symlinks wouldn't work for nixos out of the box. We would need to rewrite them and I would rather avoid that.

Its not a symbolic link what ``-e`` crates .. it creates an entry in a ``.pth`` file in the site-package of the pyenv. 

> Please don't. That's is truly all we need and We would revert it immediately anyway.

Do you have a chance to install it with ``-e``?  If not, we have to maintain ``package_data`` / I don't see a other way ..or?

SuperSandro2000:
> Do you have a chance to install it with `-e`?

It is only supported in temporary nix-shells and we would probably need to expand upon that and add some extra logic to rewrite the pth file. I am not really wanting to plan to work on that.

https://github.com/NixOS/nixpkgs/blob/695f62353a45b8dd9a81b39f045a2a32bf38a8ab/doc/languages-frameworks/python.section.md#development-mode-development-mode

> If not, we have to maintain `package_data` / I don't see a other way ..or?

Yeah, but that shouldn't be to hard. Not like There are new directories that often. I also think the current code could be replaced to just tell setuptools to just install all files in a directory, reducing the amount of wildcards it has.

return42:
@jwijenbergh @SuperSandro2000 sorry for the very long delay .. I have lost sight of the PR / sorry.

I force pushed [1fc1d5f](https://github.com/searxng/searxng/pull/3045/commits/1fc1d5fd76647dc0cc0ea10a71d2a991e6c04e96) .. can you please test if it works for you / thanks :+1: 

Here is how/what I tested:

```
$ make clean py.build
...
$ python -m venv test123
$ . ./test123/bin/activate
(test123) $ pip install dist/searxng-2023.7.19*-py3-none-any.whl
(test123) $ SEARXNG_DEBUG=1 searxng-run
```

Open http://127.0.0.1:8888/info/en/about in your browser / do some tests ..

setup.py
```diff
@@ -39,39 +37,41 @@
...
+            'static/themes/simple/css/*',
+            'static/themes/simple/css/*/*',
+            'static/themes/simple/img/*',
+            'static/themes/simple/js/*',
+            'templates/*/*',
+            'templates/*/*/*',
+            'translations/*',
+            'translations/*/*',
+            'translations/*/*/*',
+
```
SuperSandro2000:
```suggestion
```

SuperSandro2000 (approved):
I didn't test much but the `/info/en/about` endpoint is now working on nixos

SuperSandro2000:
> Here is how/what I tested:

Did this on NixOS:

```diff
diff --git a/pkgs/by-name/se/searxng/package.nix b/pkgs/by-name/se/searxng/package.nix
index a6a07a40adfd..7273552ca42f 100644
--- a/pkgs/by-name/se/searxng/package.nix
+++ b/pkgs/by-name/se/searxng/package.nix
@@ -1,6 +1,7 @@
 { lib
-, python3
 , fetchFromGitHub
+, fetchpatch
+, python3
 }:

 python3.pkgs.toPythonModule (python3.pkgs.buildPythonApplication rec {
@@ -14,6 +15,13 @@
     hash = "sha256-bThQgRcENoTIl4EwsjZiLT+f0X7pSV2Ly7X/HX1ECw0=";
   };

+  patches = [
+    (fetchpatch {
+      url = "https://github.com/searxng/searxng/pull/3045.patch";
+      hash = "sha256-sfzTtkCda6VdVPBtKXN8Ak/aeLuo2dI26S+v0vDu/HA=";
+    })
+  ];
+
   postPatch = ''
     sed -i 's/==.*$//' requirements.txt
   '';
```

setup.py
return42: Oops .. I tidied up the patch .. BTW I fixed the python version to py3.8 (py3.7 is not supported for a longer time) --> [diff](https://github.com/searxng/searxng/compare/1fc1d5fd76647dc0cc0ea10a71d2a991e6c04e96..8e67431cd86fa3922bdeca261d6721ecd592a501)

jwijenbergh:
sorry for my late reply. It works perfectly, thank you!

PR #3047, opened 2023-12-04
Revert "[fix] update user agent"

return42:
## What does this PR do?

This reverts commit 3c6549a17f886b90a8e700810f9e1aea86581966.

Related:

- https://github.com/searxng/searxng/pull/2826

rv s no longer fixed at 109

@dalf @jazzzooo can you please verify / thanks!


unixfox (approved):
Good but will need to check that it doesn't break bing engine.
Maybe they have not yet updated their anti bot.

PR #2947, opened 2023-10-19
add subdomain replacement to hostname_replace.py

Austin-Olacsi:
## What does this PR do?

you can now rewrite turtle.example.com to newsite.com/turtle
we can now move the subdomain.

## Why is this change important?

i use this and maybe other people would like to use it also.

## How to test this PR locally?

configure hostname replace like this:
`'(.*\.)?(?P<subdomain>[^.]+)\.fandom\.com$': 'breezewiki.com/\g<subdomain>'`

enable the hostname replace plugin, and search "patrick star"

and you can see, https://spongebob.fandom.com/wiki/Patrick_Star becomes https://breezewiki.com/spongebob/wiki/Patrick_Star

## Related issues

related to #1646

Issue #1646, opened 2022-08-10
Allow hostname_replace to also replace subdirectories


mattague:
**Is your feature request related to a problem? Please describe.**
I would love if the hostname_replace feature allowed/supported replacing subdirectories as well

**Describe the solution you'd like**
The way hostname_replace currently works, it will not recognize subdirectories, so when I want to replace en.wikipedia.org with www.wikiwand.com, I am unable to because I have to replace the /wiki as well, and this is unsupported. I need to turn "https://en.wikipedia.org/wiki/<insert-item-here>" into "https://www.wikiwand.com/en/<insert-item-here>"

**Describe alternatives you've considered**
I tried doing this by including the subdirectory in the regex, but it doesnt get matched still, not sure why to be honest. I had also been using an extension for wikiwand, but something in SearXNG is blocking it, so when I click on the link, i get regular wikipedia, and then if I refresj the page I get a wikiwand result.

**Additional context**
I use wikiwand to get a more modern look on wikipedia. I read wikipedia ***A LOT***, like, almost too much. I spend a couple hours everyday reading wikipedia articles. Because of this I appreciate the better readablility offered by wikiwand as well as the improved navigation.

ACodingGenie:
I would like to see this feature implemented as well. Just haven't made a request for it since you already have lots to worry about. My use case is with [Bibliogram](https://bibliogram.art/) which throws a 404 if `/u` is missing from the path, like in this case: `https://bibliogram.art/leomessi/`. With `/u` included, it works properly: `https://bibliogram.art/u/leomessi` (when not blocked by Instagram, of course).

Thank you!

return42:
I close this feature request since the plugin its based on is outed .. lets discuss further in https://github.com/searxng/searxng/issues/3682

Issue #2619, opened 2023-08-10
[FEATURE REQUEST] external list file for hostname_replace


DeltaTangoLima:
## Is your feature request related to a problem? Please describe.
No

## Describe the solution you'd like
The `hostname_replace` feature is absolutely killer, and the main reason I started using searxng. It means I can totally eliminate all the crap websites out there.

What would be really useful would be if lists of hostnames could be stored in external lists, for replace and block, respectively. Ideally these would be in a simplified format so an admin can quickly (or programmatically) update text files, rather than having to modify the settings yaml file directly.

I guess an `include` capability is really what I'm asking for, in `settings.yaml`, and then breaking the `hostname_replace` section into sub-sections for replace vs block:

```
# Configuration of the "Hostname replace" plugin:
#
hostname_replace:
   replacelist:
      !include replacements.txt
   blocklist:
      !include blocklist.txt
```

### replacelist
This file could store the regex/replace pairs, one per line, just as in the existing yaml today.

### blocklist
This file could just store the regex for the host/site/domain, and searxng can just assume a `: false` at the end of each line, when it reads the file in.

## Describe alternatives you've considered
Writing my own script to modify the yaml. I still might. ;)

## Additional context
None

return42:
This is a feature I was thinking of since a long time :+1: .. I just link issue entries and discussions related to this topic:

- https://github.com/searxng/searxng/issues/3086
- https://github.com/searxng/searxng/pull/3051#issuecomment-1868309357
- https://github.com/searxng/searxng/discussions/918
- https://github.com/searxng/searxng/issues/1348
- https://github.com/searxng/searxng/issues/1646
- https://github.com/searxng/searxng/issues/2351
- https://github.com/searxng/searxng/issues/1575#issuecomment-1198978876
- https://github.com/searxng/searxng/issues/1086
- https://github.com/searx/searx/issues/2001
- https://github.com/searxng/searxng/discussions/970
- https://github.com/searxng/searxng/discussions/911

I think one important point is to have a proper CSP handling for the redirects.

One can't really be handled well without the other.  Unfortunately, I haven't had the time yet to think the whole topic through to the end.

rpdelaney:
There's also [ublacklist](https://iorate.github.io/ublacklist/docs) in the mean time.

return42:
@rpdelaney you linked a browser extension (client side) made for google search.  We discuss a solution for SearXNG that filters out results already on server side.

A discussion about client side filtering:

[...]

PR #3051, opened 2023-12-06
add subdomain replacement to hostname_replace.py

Austin-Olacsi:
## What does this PR do?
you can now rewrite turtle.example.com to newsite.com/turtle
we can now move the subdomain.

i use this and maybe other people would like to use it also.

configure hostname replace like this:
`'(.*\.)?(?P<subdomain>[^.]+)\.fandom\.com$': 'breezewiki.com/\g<subdomain>'`

enable the hostname replace plugin, and search "patrick star"

and you can see, https://spongebob.fandom.com/wiki/Patrick_Star becomes https://breezewiki.com/spongebob/wiki/Patrick_Star

## Related issues

related to #1646
I didn't intend to close this in #2947

### DIFF
--- a/searx/plugins/hostname_replace.py
+++ b/searx/plugins/hostname_replace.py
@@ -30,7 +30,9 @@ def on_result(request, search, result):
                 # (only) on the 'parsed_url'
                 if not replacement:
                     return False
-                result[parsed] = result[parsed]._replace(netloc=pattern.sub(replacement, result[parsed].netloc))
+                subdomain = pattern.search(result[parsed].netloc).group(1) or ''
+                new_netloc = replacement.replace('(*)', subdomain)
+                result[parsed] = result[parsed]._replace(netloc=pattern.sub(new_netloc, result[parsed].netloc))
                 result['url'] = urlunparse(result[parsed])
 
         for url_field in _url_fields:
@@ -40,7 +42,9 @@ def on_result(request, search, result):
                     if not replacement:
                         del result[url_field]
                     else:
-                        url_src = url_src._replace(netloc=pattern.sub(replacement, url_src.netloc))
+                        subdomain = pattern.search(url_src.netloc).group(1) or ''
+                        new_netloc = replacement.replace('(*)', subdomain)
+                        url_src = url_src._replace(netloc=pattern.sub(new_netloc, url_src.netloc))
                         result[url_field] = urlunparse(url_src)
 
     return True


return42:
I don't think we should drill the hostname_replace plugin any further, as the name suggests it was once intended to replace the hostname.   Later the blocking function was added ... we should not put more into it.

I would prefer it if we had a URL rewriter that can apply regular expressions to URLs and that offers other functions such as 

- Import of filter lists like
  - https://github.com/stroobants-dev/ublock-origin-shitty-copies-filter
  - https://github.com/quenhus/uBlock-Origin-dev-filter/tree/main
  - ...  
- Reading out the original URL for tracker links that are intended for forwarding / https://github.com/gorhill/uBlock/wiki/Strict-blocking#ability-to-parse-the-strict-blocked-url

Austin-Olacsi:
So you are asking for a URL rewriter like this:

1. not part of the hostname_relplace plugin
2. can apply regular expressions to URLs
3. can import filter lists from an external source
4. can read tracker links and rewrite them to the intended destination URLs

return42:
> So you are asking for a URL rewriter like this:  ...

yes, that would be cool stuff, we can implement it as a core component or as plugin.

Just brainstorming: 

May we can combine it with stuff from brave's goggle / goggle's syntax https://github.com/brave/goggles-quickstart/blob/main/goggles/quickstart.goggle 

- CopyCats removal: https://github.com/brave/goggles-quickstart/blob/main/goggles/copycats_removal.goggle
- https://github.com/search?q=brave+goggles&type=repositories

Limitation: SearXNG does not have an index --> we can only "profile & rate" URLs on a result page.

NanashiTheNameless:
this does not appear to work anymore, it breaks hostname replace entirely when `  '(.*\.)?(?p&lt;subdomain&gt;[^.]+)\.fandom\.com$': 'antifandom.com/\g&lt;subdomain&gt;'` is included

NanashiTheNameless:
> this does not appear to work anymore, it breaks hostname replace entirely when ` '(.*\.)?(?p&lt;subdomain&gt;[^.]+)\.fandom\.com$': 'antifandom.com/\g&lt;subdomain&gt;'` is included

nevermind, its just github copy being shit

CIPH3RD:
for those stumbling across this thread looking to get AnonymousOverflow to work with SearXNG (properly), I tested the following regex in combination with adding this pull request to my instance (not sure why it hasn't been merged into main yet, it's great) and it works with both "/exchange/" subdomains as well as the regular ones (needed to prevent AnonOverflow from random misdirects and 404s).

 `` '(.*\.)?((?P<subdomain>[^.]+))?\.stackexchange\.com$' : 'anonymousoverflow.example.com/exchange/\g<subdomain>'``
   
 This Regular Expression needs to be used in combination with the regular one to work as a catch-all with AnonymousOverflow.
 
 keep up the great work!

return42:
> with adding this pull request to my instance

Which PR do you refer?

Austin-Olacsi:
@return42 Are you really really sure that you don't want to merge this PR? People are manually adding these changes to their own instances...

I don't foresee anyone else working on another solution...

return42:
> I don't foresee anyone else working on another solution...

I don't know which is the better decision here.   Drilling the plugin will reduce the need for a better, [alternative solution](https://github.com/searxng/searxng/pull/3051#issuecomment-1868309357).  If we don't notice PRs like this, users may never get new features. It's a dilemma ... no matter what you decide, you can only make the wrong decision.

For those coming across here: can you please review the PR and leave your comment .. if we have more votes we should merge this PR :+1:

Sommerwiesel:
I would really like this PR to get reworked for the new version and then merged...
Would also help improve what has been discussed here: https://github.com/searxng/searxng/issues/2619

return42:
I close this PR since the plugin its based on is outed .. lets discuss further in https://github.com/searxng/searxng/issues/3682

PR #3056, opened 2023-12-13
Update "extra_proxy_timeout" in Settings.yml

templehasfallen:
Changed value of "extra_proxy_timeout" from 10.0 to 10 as the variable expects an int. 
Uncommenting this value with a non-int value will throw many errors and crash all engines.

## What does this PR do?

Changes the default values of an optional setting

Uncommenting this setting when attempting to use a proxy will crash all engines and not allow any connectivity.

Uncomment option and set up proxy.

## Related issues

None. Directly troubleshooted and change submitted instead of opening issue.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -189,7 +189,7 @@ outgoing:
   #
   # Extra seconds to add in order to account for the time taken by the proxy
   #
-  #  extra_proxy_timeout: 10.0
+  #  extra_proxy_timeout: 10
   #
   # uncomment below section only if you have more than one network interface
   # which can be the source of outgoing search requests


return42 (approved):
> Changed value of "extra_proxy_timeout" from 10.0 to 10 as the variable expects an int.

@dalf may we should define the type as float?

https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/searx/settings_defaults.py#L227

return42:
> > Changed value of "extra_proxy_timeout" from 10.0 to 10 as the variable expects an int.
> 
> @dalf may we should define the type as float?
> 
> https://github.com/searxng/searxng/blob/6129b161650bf3dd8720bedbb14681e506894896/searx/settings_defaults.py#L227

ping @dalf

PR #3058, opened 2023-12-14
[feat] engine: implementation of bahnhof.de

Bnyro:
## What does this PR do?
* implement support for searching train stations for German regions

* it's useful to find train stations and information about them such as their train connections, accessibility, ...

* !bahnhof Berlin


return42 (approved):
Nice addition for our german users :+1: / works

PR #3066, opened 2023-12-19
[fix] do highlight replacement at once

### LINKED ISSUE #3057: Escaped HTML creeping into search responses

2023.12.11+8a4104b99

The issue shows itself on any public searxng instances with current searxng, for example https://search.hbubli.cc/

**What happened?**

Parts of escaped HTML creep in search responses.

**How To Reproduce**

Perform search for `class attribute` query.

**Expected behavior**

I'd prefer not to see HTML in responses in any form :)

**Screenshots & Logs**

Untitled

**Additional context**

allixx:
Highlights all search queries in search result in one go.

Fixes the case where search query contains word from highlight HTML code, which causes broken HTML to appear in search results.

Closes #3057

### DIFF
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -290,10 +290,8 @@ def highlight_content(content, query):
         if len(qs) > 0:
             queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
     if len(queries) > 0:
-        for q in set(queries):
-            content = re.sub(
-                regex_highlight_cjk(q), f'<span class="highlight">{q}</span>'.replace('\\', r'\\'), content
-            )
+        regex = re.compile("|".join(map(regex_highlight_cjk, queries)))
+        return regex.sub(lambda match: f'<span class="highlight">{match.group(0)}</span>'.replace('\\', r'\\'), content)
     return content
 
 
--- a/tests/unit/test_webutils.py
+++ b/tests/unit/test_webutils.py
@@ -57,6 +57,11 @@ class TestWebUtils(SearxTestCase):
                     ]
                 ),
             ),
+            (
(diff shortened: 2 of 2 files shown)


Bnyro (approved):
Lgtm and I can confirm this fixes the issue, thanks for your investigations!

searx/webutils.py
```diff
@@ -290,10 +290,10 @@ def highlight_content(content, query):
         if len(qs) > 0:
             queries.extend(re.findall(regex_highlight_cjk(qs), content, flags=re.I | re.U))
     if len(queries) > 0:
-        for q in set(queries):
-            content = re.sub(
-                regex_highlight_cjk(q), f'<span class="highlight">{q}</span>'.replace('\\', r'\\'), content
-            )
+        regex = re.compile("|".join(map(regex_highlight_cjk, queries)))
+        return regex.sub(
+            lambda match: f'<span class="highlight">{re.escape(match.group(0))}</span>'.replace('\\', r'\\'), content
```
allixx: I noticed that re.escape'ing match.group(0) introduces undesired effects, for example escaped dashes, which is further worsened by the following backslash replace. I'm not sure escaping is needed there. Opinions?

allixx:
So I've ran this on my instance for more than a month by now and haven't got any breakage or peculiar side effects.

return42 (approved):
>  So I've ran this on my instance for more than a month by now and haven't got any breakage or peculiar side effects.

@allixx sorry for the delay .. I haven't seen, that you force pushed already a month ago.

I can confirm; joining regular expressions into one regex works much better here compared to the iteration over them and replace (again and again) the word `class` in the sub-strings  :+1:

PR #3075, opened 2023-12-27
[fix] Fix safesearch parameter in Google Videos engine (#2762)

### LINKED ISSUE #2762: [Engine:Google Video] Safesearch is not considered when set to None
Brief description by @unixfox

For the Google video engine, if you set Safesearch to None in the preferences it seems like it's not considered by the search engine behind.

With a search like `!gov block porn`, on the second page or first one you get thumbnails blurred.

image

Example URL: https://paulgo.io/search?q=%21gov+block+porn&category_none=1&pageno=2&language=all&time_range=&safesearch=0&theme=simple

---------------

Original message:

Basically as the title says whenever you search for something NSFW or not even NSFW but in google video engine the video thumbnail gets blurred out.
SearXNG google video blurred video thumbnail

**Step to reproduce this issue**
!gov porn
!gov block porn
basically any nsfw related search on google video (!gov)

without blur, video thumbnail of Google video search result

N/A   

**Additional context**
I don't know if this going to help or not, the only working parameters I have found is `&safe`.

`safe` parameter has 3 value: `on`, `off`, `images`(for blurring the images)
**Google Video url without blurred image:** https://www.google.com/search?q=%s&tbm=vid&safe=off

<Table>
<tr>
<th>
without `safe=off`: youtube age restricted video with blurred video thumbnail
</th>
<th>
with `safe=off`: youtube age restricted video with video thumbnail
</th>
</tr>
</Table>

BernieHuang2008:
## What does this PR do?
Fix safesearch parameter in Google Videos engine

As i said in #2762: 
> got it! 
look at here: 
https://github.com/searxng/searxng/blob/8dfc1dbc5c0ed4968a7111909dc597a755cc94c6/searx/engines/google_videos.py#L90-L91
if safesearch is set to 'off', the value of `params['safesearch']` will be `0`, which equals `False` in python, so the `if` codeblock won't be executed!

*PS: i won't put an image here ... you know why ...*

the usual `make run`

## Related issues
I won't close the #2762 , because the problem of ddg hasn't been solved.

### DIFF
--- a/searx/engines/google_videos.py
+++ b/searx/engines/google_videos.py
@@ -87,7 +87,7 @@ def request(query, params):
 
     if params['time_range'] in time_range_dict:
         query_url += '&' + urlencode({'tbs': 'qdr:' + time_range_dict[params['time_range']]})
-    if params['safesearch']:
+    if 'safesearch' in params:
         query_url += '&' + urlencode({'safe': filter_mapping[params['safesearch']]})
     params['url'] = query_url
 


Bnyro (approved):

BernieHuang2008:
have a look please, @return42

return42 (approved):
> have a look please, @return42

Sorry for the delay .. I have somehow overlooked PR :facepalm: 

> won't close the https://github.com/searxng/searxng/issues/2762 , because the problem of ddg hasn't been solved.

- https://github.com/searxng/searxng/issues/2762

.. as the title state .. the issue is related to Goolge-Video .. I think we should close the issue.

It was not clear to me that we have a similar bug in DDG .. but if so we should open an issue for DDG.

FYI: I slightly modified the commit message.

Thanks a lot for having an eye on this issue :+1:

PR #3076, opened 2023-12-28
[fix] display problem in the image detail window

BernieHuang2008:
## What does this PR do?
fix a display problem in the image detail window.

|before|after|
|-|-|
|image|image|


Bnyro (approved):
I can't reproduce the issue on the current master branch, but hiding the offset shouldn't have any negative impact.

return42 (approved):
> can't reproduce the issue on the current master branch, but hiding the offset shouldn't have any negative impact.

Yes, same here / I suppose issue depends on the browser's font (?) .. however .. LGTM :+1: 

@BernieHuang2008 FYI: to solve conflict with current master, I had to rebuild the static files / pushed on your branch.

PR #3087, opened 2023-12-31
Searx to SearXNG change error message

unixfox:
Just changing the error message because the software is called SearXNG and not SearX.

I know it's not the only change needed from changing SearX to SearXNG but this is one step towards fixing the name in the source code, at least for user errors.


return42 (approved):
> I know it's not the only change needed from changing SearX to SearXNG but this is one step towards fixing the name in the source code.

We did migrated most of the names but I don't know if we would ever change the python package name from `searx` to `searxng` ..

Bnyro (approved):

PR #3095, opened 2024-01-03
Fix development quickstart URL

Myzel394:
## What does this PR do?

The url seems to be incorrect; fixed it

The URL should be correct :D


return42 (approved):
Thanks for the fix :+1:

Issue #534, opened 2021-11-21
Satellite projects


dalf:
Looking at

* https://github.com/searxng/searxng/issues/320 :
  >  I could see this feature becoming a way for communities to build their own search engines

* https://github.com/searxng/searxng/issues/530 :
  > it's more a matter of getting content in the Esperanto language.
  
I feel that some communities would like to make some customizations on their own.
For now the current solution from here is to fork, make the customization and merge from time to time the code from here.

Perhaps, we can find a better way.

For example, a satellite project "searxng-esperanto" would contain:
* the engine definitions ( settings.yml )
* the code for engines
* some tests.
* nothing more.

From here (to be define):
* either add the project in requirements.txt (easy because pip takes care of downloading / upgrading the files).
* either make a reference in settings.yml

This would allow to:
* to keep this project with a default generic configuration.
* various communities to have their own customization, with a easy way to integrate in this project.

dalf:
ping @return42

return42:
Nice idea but I do not have capacity for .. it's getting too much for me, I'd like to finish the move from searx to SearXNG first .. there is a loot of work to get this done.  Afterwards I am open for new features and models.

dalf:
About https://github.com/searxng/searxng/issues/725 : I wonder if this list could be manage like the adblock lists: the plugin could use an URL of where to download the list. So various project could manage different lists with different flavor.

In settings.yml:
```yaml
hostname_replace:
  # pages which reflect nothing more than bad translations of other pages
  'qastack\.com\.de': false
  'im-coder\.com': false
  'code\.i-harness\.com': false
  *import: https://example.com/path/hostname_replace.yml
```

In this example `*import` is special syntax that import the content of URL into the current configuration.

This could also works with the engines too.

return42:
Load block list is one thing, load executable code from the net is another thing ... load code from a remote opens the door for exploits.

dalf:
For reference: https://github.com/searxng/searxng/pull/3099#issuecomment-1879782136

return42:
Ultimately, this would mean that we would no longer be allowed to change any function in SearXNG without checking its impact on satellites (every function would potentially be an official interface). This is conceivable if we define internal interfaces (similar to plugins). However, the adjustments would then be limited to specific topics, such as plugins. In any case, the further development of the core product would become more complicated and slower.

SearXNG is the project we manage, and I want us to focus on it.

unixfox:
[...]

PR #3099, opened 2024-01-05
[feat]: implementation of stract (autocompleter)

Bnyro:
## What does this PR do?
* add stract for search and auto completion

* stract is open source and someone recommended it in the matrix room

* !stract foo
* set the autocompleter to stract and start typing in the search bar

## Additional notes
* the engine is in beta and might change soon/in the future
* it only supports 5 or 6 other locales at the moment, so I don't think it's worth adding support for it and adding it to the web category

### DIFF
--- a/docs/dev/search_api.rst
+++ b/docs/dev/search_api.rst
@@ -69,7 +69,7 @@ Parameters
 
 ``autocomplete`` : default from :ref:`settings search`
   [ ``google``, ``dbpedia``, ``duckduckgo``, ``mwmbl``, ``startpage``,
-  ``wikipedia``, ``swisscows``, ``qwant`` ]
+  ``wikipedia``, ``stract``, ``swisscows``, ``qwant`` ]
 
   Service which completes words as you type.
 
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -6,7 +6,7 @@
 # pylint: disable=use-dict-literal
 
 import json
-from urllib.parse import urlencode
+from urllib.parse import urlencode, quote_plus
 
 import lxml
 from httpx import HTTPError
@@ -16,15 +16,24 @@ from searx.engines import (
     engines,
     google,
 )
-from searx.network import get as http_get
+from searx.network import get as http_get, post as http_post
 from searx.exceptions import SearxEngineResponseException
 
 
-def get(*args, **kwargs):
+def update_kwargs(**kwargs):
     if 'timeout' not in kwargs:
         kwargs['timeout'] = settings['outgoing']['request_timeout']
     kwargs['raise_for_httperror'] = True
+
+
+def get(*args, **kwargs):
+    update_kwargs(**kwargs)
     return http_get(*args, **kwargs)
+
+
+def post(*args, **kwargs):
+    update_kwargs(**kwargs)
+    return http_post(*args, **kwargs)
 
 
 def brave(query, _lang):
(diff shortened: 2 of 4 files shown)


utils/brand.env
```diff
@@ -1,5 +1,5 @@
 export SEARXNG_URL=''
 export SEARXNG_PORT='8888'
 export SEARXNG_BIND_ADDRESS='127.0.0.1'
-export GIT_URL='https://github.com/searxng/searxng'
```
dalf: ping @return42 @Bnyro

Bnyro: Can't we just add that file to the .gitignore? :(

return42: This file is needed if you ([like me](https://github.com/searxng/searxng/compare/master...return42:searxng:darmarit.org)) maintain a [brand](https://docs.searxng.org/admin/settings/settings_brand.html) .. its sourced by the [`/utils/searxng.sh`](https://docs.searxng.org/utils/searxng.sh.html).

- [installation](https://docs.searxng.org/admin/installation-scripts.html) and [update](https://docs.searxng.org/admin/update-searxng.html) SearXNG is based on
- [`make buildenv`](https://docs.searxng.org/dev/makefile.html#make-buildenv)
- https://docs.searxng.org/dev/contribution_guide.html#deploy-on-github-io

For those who run a docker instance this file may be annoying in the development phase, for those of us who run a normal installation (without docker) this file is elementary.

Please adjust the git workflow so that only those files are added to the commit where there are changes that belong to the patch / don't just add everything with `git add` without taking a closer look.

[Commits are a firsthand historical record of exactly how and why each line of code came to be. They even come with human-readable messages! As a result, **a repository’s commit history is the best tool a developer can use to explain and understand code** .... Commits are more than just the code they contain. Despite there being no shortage of jokes about them, **commit messages are an extremely valuable – but often overlooked – component of a commit.** Most importantly, they’re an opportunity to speak directly to your audience and explain a change in your own terms.](https://github.blog/2022-06-30-write-better-commits-build-better-projects/)

return42 (approved):
LGTM :+1: / Thanks !

----

> It only supports 5 or 6 other locales at the moment, so I don't think it's worth adding support for it and adding it to the web category

So far I have tested, they support search languages ..

grafik

but at this time they do not have a UI localization nor language support for the autocompleter.

dalf:
The engine is disabled by default so the bots won't send traffic to stract by default. However,  it seems the engine is handled by a small team: it would be nice to ping them before merging.

ping @mikkeldenker , @oeb25

Bnyro:
Maybe we can just remove it from searx/settings.yml? I'd really like to use the engine on my private instance without having to fork the project.

unixfox:
> Maybe we can just remove it from searx/settings.yml? I'd really like to use the engine on my private instance without having to fork the project.

Use git patches? That's what I do for my own instance.

dalf:
Related to https://github.com/searxng/searxng/issues/534 ...

Bnyro:
> Use git patches? That's what I do for my own instance.
>
Is there a way to do that with docker compose?

unixfox:
> > Use git patches? That's what I do for my own instance.
> >
> Is there a way to do that with docker compose?
> 
> 

https://github.com/unixfox/searx-custom if you want have a look at my way of doing.

mikkeldenker:
We would love to have stract implemented into searxng! Thanks a lot for adding it @Bnyro.

Everything noted here about the languages is correct. Currently only very few languages are supported for search and autocomplete only works in english. In general, the engine probably works best for english queries but this is of course something we would very much like to improve in the future.

Also, thanks for being considerate about the amount of traffic you send our way initially. Stract is still very small (currently served from just 2 commodity servers) but I'm pretty confident we can handle the traffic if it just isn't enabled by default.
Please let me know if you need anything from my end to get this merged.

return42:
> We would love to have stract implemented into searxng!

@mikkeldenker Thanks a lot, we are very happy to hear :rocket: .. if you (stract) have any special needs (UA header or whatever) please let us know.

@dalf @unixfox Do you still see reasons not to merge the PR?

Bnyro:
Thanks for your answer @mikkeldenker, and thank you for your work on that amazing project!

If there are any traffic issues in the near future (then they're probably caused by us), feel free to leave a comment here again and we'll find a solution for it, because we definitely don't want small projects to be suffering from us.

I'll be following the progress of stract in the future and see when we can look into adding support for different languages too here :)

PR #3101, opened 2024-01-06
utils/build_env.py run without external dependencies

dalf:
## What does this PR do?

This should allow to create `utils/brand.env` without virtualenv. 

The ultimate goal of the commit is to remove `utils/brand.env` from the git repository. Then the script can create this file when required with Python interpreter without virtualenv.

The code includes a tiny yaml parser that **should** be good enough. 
The code reads `searx/settings.yml` and `searx/version.py` directly (the environment variables are just ignored)

Multiple times `utils/brand.env` ends up in the pull requests because `git add .` usually works.

In addition, `utils/brand.env` contains information that are already known:
* git URL and branch
* some values of `searx/settings.yml` (and not from `/etc/searxng/settings.yml`)

This file should be generated when needed and added to .gitignore

`make buildenv`

## Author's checklist

I wrote the small yaml parser nearly out of thin air, so I have a big concern about it, however yamllint make sure the format is normalized ; so perhaps this is good enough. List are not parsed, hower the dot notation to read the settings does not allow list ( `server.base_url` for example).

[yq](https://github.com/mikefarah/yq/#install) is a more reliable alternative but this require to download a binary from github which is not great.

### DIFF
--- a/utils/build_env.py
+++ b/utils/build_env.py
@@ -4,58 +4,146 @@
 
 # set path
 import sys
-import os
-from os.path import realpath, dirname, join, sep, abspath
+import importlib.util
+import re
 
-repo_root = realpath(dirname(realpath(__file__)) + sep + '..')
-sys.path.insert(0, repo_root)
+from pathlib import Path
 
-# Assure that the settings file from repository's working tree is used to
-# generate the build_env, not from /etc/searxng/settings.yml.
-os.environ['SEARXNG_SETTINGS_PATH'] = join(repo_root, 'etc', 'settings.yml')
+repo_root = Path(__file__).resolve().parent.parent
 
-def _env(*arg, **kwargs):
-    val = get_setting(*arg, **kwargs)
-    if val is True:
-        val = '1'
-    elif val is False:
-        val = ''
-    return val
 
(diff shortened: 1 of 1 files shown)


return42:
@dalf thanks for your PR .. FYI I'm working in parallel in my [brand.env branch](https://github.com/searxng/searxng/compare/master...return42:searxng:brand.env) / my intention is to entirely get rid of brand.env and the workflow behind .. as we discussed; all the needed environment can be generated on the fly with some git commands and your YAML parser from here.

.. if it is OK for you I consider to copy your YAML parser (in a slightly modified version) into my [brand.env branch](https://github.com/searxng/searxng/compare/master...return42:searxng:brand.env)

dalf:
> . if it is OK for you I consider to copy your YAML parser (in a slightly modified version) into my [brand.env branch](https://github.com/searxng/searxng/compare/master...return42:searxng:brand.env)

Sure, no problem.

return42:
I close this PR in favor of .. 

- https://github.com/searxng/searxng/pull/3104

The code to parse YAML file from this PR has been moved to PR-3104 into file `utils/get_setting.py` .. thanks for your implementation of this tiny YAML parser :+1:

PR #3103, opened 2024-01-06
[feat] engine: Mullvad Leta

### LINKED ISSUE #2509: [Feature] Mullvad Leta
**Working URL to the engine**
https://leta.mullvad.net/

Mullvad is a reputable VPN provider. If I could have my local instance also be a proxy to this search engine, that would reduce the number of devices I need connected to Mullvad to run to access this search feature.

**Features of this engine**
The blog post [here](https://leta.mullvad.net/faq) contains the features and FAQ. Their meta search engine uses the official paid Google API, so results are likely quite robust and not error prone to changes.

Technically Mullvad Leta is already a meta search engine on Google - I wonder if they use searx lol (probably not, they use NodeJS).

TBD - will likely need to scrape the front end

general

**Additional context**
You MUST have a Mullvad account to use this search engine. No account, no searches.

There is a limit on 100 non-cached searches (or credits) per valid Mullvad VPN account per 24 hours. You can search the cache without using credits.

**Notes**
I plan on working on this integration when I can, as it directly benefits myself! But anyone else can work on it too.

glanham-jr:
## What does this PR do?

Add a new engine to support Mullvad Leta

### WARNING - Engine Trait Updates require Mullvad VPN Connection

The engine_traits can only be fetched when running the `make traits` script in a network within Mullvad VPN. Otherwise, the traits will never be updated.

Failure to connect to the VPN will only print an error statement.

Mullvad Leta is already a meta-search engine that I would like to leverage on my own searxng instance. Would think this is also useful for others as well.

Pre-Req: Have access to the Mullvad VPN and connect to it.

1. Connect to Mullvad VPN
1. Uncomment the mullvad settings
1. Disable "use_cache" to get a result
1. Run searxng
1. Search for a random phrase X using '!ml'
1. Validate results
1. Enable "use_cache" to test cache gathering
1. Run searxng 
1. Search for the same phrase X with same bang
1. Run the engine trait generator to get the latest traits and ensure no errors thrown
1. Try disabling the VPN
1. Search for a phrase on searxng
1. Check that the error message says "not connected to Mullvad VPN" to ensure easy diagnosing

## Author's checklist
- [X] Test end to end
- [x] Enable option to allow searching out of cache
- [X] Fetch countries dynamically 

## Screenshots
image

Closes #2509

### DIFF
--- a/searx/engines/mullvad_leta.py
+++ b/searx/engines/mullvad_leta.py
@@ -1,9 +1,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # lint: pylint
 
-"""This is the implementation of the Mullvad-Leta meta-search engine
+"""This is the implementation of the Mullvad-Leta meta-search engine.
 
-This engine _REQUIRES_ that searxng operate within a Mullvad VPN
+This engine **REQUIRES** that searxng operate within a Mullvad VPN
 
 If using docker, consider using gluetun for easily connecting to the Mullvad
 
@@ -13,6 +13,12 @@ Otherwise, follow instructions provided by Mullvad for enabling the VPN on Linux
 
 - https://mullvad.net/en/help/install-mullvad-app-linux
 
+.. hint::
+
+   The :py:obj:`EngineTraits` is empty by default.  Maintainers have to run
+   ``make data.traits`` (in the Mullvad VPN / :py:obj:`fetch_traits`) and rebase
+   the modified JSON file ``searx/data/engine_traits.json`` on every single
+   update of SearXNG!
 """
 
 from typing import TYPE_CHECKING
@@ -125,7 +131,8 @@ def response(resp: Response):
 def fetch_traits(engine_traits: EngineTraits):
     """Fetch languages and regions from Mullvad-Leta
 
-    WARNING
+    .. warning::
+
         Fetching the engine traits also requires a Mullvad VPN connection. If
         not connected, then an error message will print and no traits will be
         updated.
--- a/searx/engines/mullvad_leta.py
+++ b/searx/engines/mullvad_leta.py
@@ -1,0 +1,166 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+
+"""This is the implementation of the Mullvad-Leta meta-search engine
+
+This engine _REQUIRES_ that searxng operate within a Mullvad VPN
+
+If using docker, consider using gluetun for easily connecting to the Mullvad
+
+- https://github.com/qdm12/gluetun
+
+Otherwise, follow instructions provided by Mullvad for enabling the VPN on Linux
+
+- https://mullvad.net/en/help/install-mullvad-app-linux
+
+"""
+
+from typing import TYPE_CHECKING
+from httpx import Response
+from lxml import html
+from searx.enginelib.traits import EngineTraits
+from searx.locales import region_tag, get_official_locales
+from searx.utils import eval_xpath, extract_text, eval_xpath_list
+from searx.exceptions import SearxEngineResponseException
+
+if TYPE_CHECKING:
+    import logging
+
+    logger = logging.getLogger()
+
+traits: EngineTraits
+
+use_cache: bool = True  # non-cache use only has 100 searches per day!
+
+search_url = "https://leta.mullvad.net"
+
+# about
+about = {
+    "website": search_url,
+    "wikidata_id": 'Q47008412',  # the Mullvad id - not leta, but related
+    "official_api_documentation": 'https://leta.mullvad.net/faq',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+# engine dependent config
+categories = ['general', 'web']
+paging = True
+max_page = 50
+time_range_support = True
+time_range_dict = {
+    "day": "d1",
+    "week": "w1",
+    "month": "m1",
+    "year": "y1",
+}
+
+
+def is_vpn_connected(dom: html.HtmlElement) -> bool:
+    """Returns true if the VPN is connected, False otherwise"""
+    connected_text = extract_text(eval_xpath(dom, '//main/div/p[1]'))
+    return connected_text != 'You are not connected to Mullvad VPN.'
+
+
+def assign_headers(headers: dict) -> dict:
+    """Assigns the headers to make a request to Mullvad Leta"""
+    headers['Accept'] = "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8"
+    headers['Content-Type'] = "application/x-www-form-urlencoded"
+    headers['Host'] = "leta.mullvad.net"
+    headers['Origin'] = "https://leta.mullvad.net"
+    return headers
+
+
+def request(query: str, params: dict):
+    country = traits.get_region(params.get('searxng_locale', 'all'), traits.all_locale)  # type: ignore
+
+    params['url'] = search_url
+    params['method'] = 'POST'
+    params['data'] = {
+        "q": query,
+        "gl": country if country is str else '',
+    }
+    # pylint: disable=undefined-variable
+    if use_cache:
+        params['data']['oc'] = "on"
+    # pylint: enable=undefined-variable
+
+    if params['time_range'] in time_range_dict:
+        params['dateRestrict'] = time_range_dict[params['time_range']]
+    else:
+        params['dateRestrict'] = ''
+
+    if params['pageno'] > 1:
+        #  Page 1 is n/a, Page 2 is 11, page 3 is 21, ...
+        params['data']['start'] = ''.join([str(params['pageno'] - 1), "1"])
+
+    if params['headers'] is None:
+        params['headers'] = {}
+
+    assign_headers(params['headers'])
+    return params
+
+
+def extract_result(dom_result: html.HtmlElement):
+    [a_elem, h3_elem, p_elem] = eval_xpath_list(dom_result, 'div/div/*')
+    return {
+        'url': extract_text(a_elem.text),
+        'title': extract_text(h3_elem),
+        'content': extract_text(p_elem),
+    }
+
+
+def response(resp: Response):
+    """Checks if connected to Mullvad VPN, then extracts the search results from
+    the DOM resp: requests response object"""
+
+    dom = html.fromstring(resp.text)
+    if not is_vpn_connected(dom):
+        raise SearxEngineResponseException('Not connected to Mullvad VPN')
+    search_results = eval_xpath(dom.body, '//main/div[2]/div')
+    return [extract_result(sr) for sr in search_results]
+
+
+def fetch_traits(engine_traits: EngineTraits):
+    """Fetch languages and regions from Mullvad-Leta
+
+    WARNING
+        Fetching the engine traits also requires a Mullvad VPN connection. If
+        not connected, then an error message will print and no traits will be
+        updated.
+    """
+    # pylint: disable=import-outside-toplevel
+    # see https://github.com/searxng/searxng/issues/762
+    from searx.network import post as http_post
+
+    # pylint: enable=import-outside-toplevel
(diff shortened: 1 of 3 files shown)


Austin-Olacsi:
bear in mind, specifying the xpath engine in the settings.yml means searxng will use the xpath.py (instead of your mullvad-leta.py) 
 which i also think we should rename to mullvad_leta.py

glanham-jr:
Ahhh, thanks for the info, was basing this on looking at other settings (first engine commit). Will test it soon.

searx/settings.yml
```diff
@@ -1192,6 +1192,23 @@ engines:
       require_api_key: false
       results: JSON
 
+  - name: mullvad_leta
```
unixfox: The engine should be commented if it doesn't work out of the box.

unixfox (changes requested):

searx/engines/mullvad-leta.py
```diff
@@ -0,0 +1,320 @@
...
+            'title': h3_elem.text_content(),
+            'content': p_elem.text,
+        }
+    dom = html.fromstring(resp.text)
+    search_results = dom.body.xpath('//main/div[2]')[0].getchildren()[1:]
+    return [extract_result(sr) for sr in search_results]
+
+
+# List of all available countries for the "gl" option
+_countries = [
```
Bnyro: At SearXNG we handle the countries and languages using engine traits, they shouldn't be hardcoded if possible. For example, see https://github.com/searxng/searxng/blob/master/searx/engines/brave.py#L356. Basically, `fetch_traits` always fetches the list of countries from somewhere in the internet and parses them. In case that mullvad leta supports all regions/countries that exist, we should probably use the babel library's inbuilt countries instead of hard coding them here.

mullvad-leta.py
```diff
@@ -0,0 +1,47 @@
...
+        # Mullvad-Leta searches in increments of 10
+        url = "https://leta.mullvad.net"
+        data = {
+            "q": "test",
+            "gl": "",
+            "dateRestrict": "",
+            "oc": "on",
+            "start": "11",
+        }
+        login_response = client.post(url, data=data, headers={
```
Bnyro:
```suggestion
    from searx.network import post
    
        # Mullvad-Leta searches in increments of 10
        url = "https://leta.mullvad.net"
        data = {
            "q": "test",
            "gl": "",
            "dateRestrict": "",
            "oc": "on",
            "start": "11",
        }
        login_response = post(url, data=data, headers={
```

searx/engines/mullvad-leta.py
Bnyro: Oh, I just saw they're part of https://github.com/searxng/searxng/pull/3103/files#diff-fe29dd443723cfce49c4b43476d6311cfb8970e2515dc80955d38e15363a57bd. So we should probably fetch them dynamically from the URL used there instead of hard coding.

glanham-jr: Yes, this is part of the TODO to fetch them with engine traits

searx/settings.yml
glanham-jr: Thanks, after testing I'll do this and resolve the comment

glanham-jr:
@unixfox not sure if I did this correctly, but I want to use an 'use_cache' variable in the settings. I setup a 'use_cache' variable in the engine as well, that should be good enough, right?

Other than that, the initial implementation is now working, just need to do some cleanup.

glanham-jr:
Next question is that I have the number of pages returned from the request, I'm 
 not sure how to return that from the response method yet.

Hackurei:
not sure if you're still having trouble with use_cache, but if you are i would try- if use_cache == True

glanham-jr:
Added detailed testing steps for anyone with a Mullvad VPN if they want to give it a go

return42:
@glanham-jr what is the sate of this PR? .. Since I do not have a Mullvad account I can't test this PR .. but I can make a code review if you want to see this PR merged ..

glanham-jr:
@return42 The PR is ready for review and merging. Given its been >1 month since implementation, I can run through happy path test cases again to ensure its working.

EDIT: Just verified this is working locally again. If you'd prefer to test it yourself, happy to provide a Mullvad account for your use.

The only possible improvement is to return the total number of pages available, which I have code that works to extract the info (see commented code), just not sure how to integrate that info into SearxNG if its used at all.

searx/engines/mullvad_leta.py
```diff
@@ -0,0 +1,200 @@
...
+    headers['Origin'] = "https://leta.mullvad.net"
+    headers['Referer'] = "https://leta.mullvad.net/"
+    headers['Sec-Fetch-Dest'] = "document"
+    headers['Sec-Fetch-Mode'] = "navigate"
+    headers['Sec-Fetch-Site'] = "same-origin"
+    headers['Sec-Fetch-User'] = "?1"
+    headers['Sec-GPC'] = "1"
+    headers['TE'] = "trailers"
+    headers['Upgrade-Insecure-Requests'] = "1"
+    return headers
```
return42: Sure all these headers are needed?

searx/engines/mullvad_leta.py
```diff
@@ -0,0 +1,200 @@
...
+
+def request(query: str, params: dict):
+    """pre-request callback
+    params<dict>:
+      method  : POST/GET
+      headers : {}
+      data    : {} # if method == POST
+      url     : ''
+      category: 'search category'
+      pageno  : 1 # number of the requested page
```
return42: This doc here does not describe what the request function of mulvad_leta does

searx/engines/mullvad_leta.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    def get_country(eng_traits: EngineTraits) -> str:
+        sxng_locale = params.get('searxng_locale', 'all')
+        country = eng_traits.get_region(sxng_locale, eng_traits.all_locale)
```
return42: This func is only used once .. no need to write a func for it.

searx/engines/mullvad_leta.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # def are_more_pages(dom: html.HtmlElement) -> bool:
+    #    previous and next buttons. If only 1, then just previous
+    #    return len(eval_xpath_list(dom, '//main/div[3]/form')) == 2
+    # def get_page_x_of_y(dom: html.HtmlElement) -> Tuple[int, int]:
+    #    page_x_of_y = extract_text(eval_xpath(dom.body, '//main/div[3]/span'))
+    #    if page_x_of_y is None or page_x_of_y == '':
+    #        return (0, 0)
+    #    [_, x, _, y] = page_x_of_y.split(' ')
+    #    return (int(x), int(y))
```
return42: This code seems no longer needed, remove it.

searx/settings.yml
```diff
@@ -1190,6 +1190,20 @@ engines:
...
+  #   use_cache: true  # Only 100 non-cache searches per day, suggested only for private instances
+  #   search_url: https://leta.mullvad.net
+  #   categories: [general, web]
+  #   shortcut: ml
+  #   about:
+  #     website: https://leta.mullvad.net/
+  #     wikidata_id: Q47008412
+  #     official_api_documentation: https://leta.mullvad.net/faq
+  #     use_official_api: false
+  #     require_api_key: false
```
return42: Its already in mullvad_leta.py

searx/engines/mullvad_leta.py
```diff
@@ -0,0 +1,200 @@
...
+    if not isinstance(resp, Response):
+        print("ERROR: failed to get response from mullvad-leta. Are you connected to the VPN?")
+        return
+    if not resp.ok:
+        print("ERROR: response from mullvad-leta is not OK. Are you connected to the VPN?")
+        return
+    dom = html.fromstring(resp.text)
+    if not is_vpn_connected(dom):
+        print('ERROR: Not connected to Mullvad VPN')
+        return
```
return42, line 155: This function will be called by gh worklfows when [updating data](https://github.com/searxng/searxng/blob/master/.github/workflows/data-update.yml) .. when we are not in the Mullvad VPN.

return42 (changes requested):

searx/engines/mullvad_leta.py
glanham-jr: I'll triage removing them and ensuring I get results.

searx/engines/mullvad_leta.py
glanham-jr: Will this even run if Mullvad is commented out in settings? It technically works, so I don't see a reason not to include it.

searx/engines/mullvad_leta.py
glanham-jr: So this is what I was trying to find out - I can get the remaining number of pages available for Leta, so SearxNG doesn't show more pages available than expected. Looking at the codebase I couldn't find if this was a feature or not, so I commented it out.

In the meantime, I'll remove it, as it doesn't seem strictly necessary.

searx/engines/mullvad_leta.py
return42: solved

searx/engines/mullvad_leta.py
return42: If you not know a good documentation, you can remove the doc-string.

searx/engines/mullvad_leta.py
return42: Sorry / forgot that engine is commented out by default .. that means:

**Maintainers have to run `make data.traits` (in the Mullvad VPN) and merge the modified JSON file `searx/data/engine_traits.json` ... on every single update of SearXNG?** /  May this is not clear to maintainers and needs to be documented?

searx/settings.yml
return42: Can you please have a look and remove duplicates already given `searx/engines/mullvad_leta.py`

return42 (changes requested):

searx/engines/mullvad_leta.py
glanham-jr: > Maintainers have to run make data.traits (in the Mullvad VPN) and merge the modified JSON file searx/data/engine_traits.json ... on every single update of SearXNG?

Wasn't aware the maintainers would need to run the update on every update. Let's keep it simple, I'll just remove this method and have no engine traits.

searx/engines/mullvad_leta.py
glanham-jr: I'm just following the format of [other engines I've seen](https://github.com/searxng/searxng/blob/3edaa26dae80b1256c9460d59754475b6d7293ea/searx/engines/duden.py#L34).

With this, I'll assume the engines with these docstrings are actually wrong and just remove my comment.

searx/engines/mullvad_leta.py
glanham-jr: I've included a dedicated WARNING header for this specific issue + a WARNING in code

I feel like its a waste to delete, but it likely won't ever be ran outside of local usage (unless Mullvad wants to sponsor us with a free VPN connection :smile: )

return42:
@glanham-jr I squashed your commits into one and added 707d6270c to inckude the doc-strings in the engine documentation --> https://docs.searxng.org/dev/engines/online/mullvad_leta.html

searx/engines/mullvad_leta.py
allendema: Although now merged: Can you use the logger instead of print statements?

return42: > Although now merged: Can you use the logger instead of print statements?

It's OK the way it is now ..

The print statements are in the `def fetch_traits(..):` function / this function is not called in the server context / we need this print statements when running a `make data.traits`.

allendema: Nice. I thought it was getting called everytime.

glanham-jr:
So it has been a while since I've been using this on my private instance, and wanted to report back that the integration has still been running well 😄I disabled searching in cache, as I personally don't have > 100 searches a day personally.

PR #3104, opened 2024-01-08
[mod] get rid of ./utils/brand.env and its workflow

return42:
## What does this PR do?

All the environments defined in `./utils/brand.env` are generated on the fly, so there is no longer a need to define the brand environment in this file and all the workflows to handle this file.

Simplify workflows to build a brand of SearXNG

Patch server setting ...

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 5e70b7803..205142414 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -67,11 +67,11 @@ search:
 
 server:
   # Is overwritten by ${SEARXNG_PORT} and ${SEARXNG_BIND_ADDRESS}
-  port: 8888
-  bind_address: "127.0.0.1"
+  port: 7777
+  bind_address: "0.0.0.0"
   # public URL of the instance, to ensure correct inbound links. Is overwritten
   # by ${SEARXNG_URL}.
-  base_url: false  # "http://example.com/location"
+  base_url: "http://example.com/location"
   limiter: false  # rate limit the number of request on the instance, block some bots
   public_instance: false  # enable features designed only for public instances
```
and compare the values with the *environment* section of command:

```
$ ./utils/searxng.sh 
usage:
...
environment:
  GIT_URL              : https://github.com/return42/searxng
  GIT_BRANCH           : brand.env
  SEARXNG_URL          : http://example.com/location
  SEARXNG_PORT         : 7777
  SEARXNG_BIND_ADDRESS : 0.0.0.0
```

As you can see `GIT_URL` and `GIT_BRANCH` are coming from the branch that is checked out.

Test installation procedure (LXC):

```
$ sudo -H ./utils/lxc.sh remove searxng-archlinux
$ sudo -H ./utils/lxc.sh build searxng-archlinux
$ sudo -H ./utils/lxc.sh cmd searxng-archlinux FORCE_TIMEOUT=0 \
          SEARXNG_UWSGI_USE_SOCKET=false ./utils/searxng.sh install all
$ sudo -H ./utils/lxc.sh cmd searxng-archlinux FORCE_TIMEOUT=0 \
          SEARXNG_UWSGI_USE_SOCKET=false ./utils/searxng.sh install nginx
```

Now check `proxy_pass` of the nginx `location` ...

https://github.com/searxng/searxng/blob/3dea7e609bac72d4a129979ba15315b966ff020f/utils/templates/etc/nginx/default.apps-available/searxng.conf#L1-L3

has been set to bind address `0.0.0.0` and port `7777`:

```
$ sudo -H ./utils/lxc.sh cmd searxng-archlinux \
          grep proxy_pass /etc/nginx/default.apps-available/searxng.conf
```
```
    proxy_pass http://0.0.0.0:7777;
```

Now probe searxng instance check

```
$ sudo -H ./utils/lxc.sh cmd searxng-archlinux \
          touch /usr/local/searxng/searxng-src/utils/brand.env

$ sudo -H ./utils/lxc.sh cmd searxng-archlinux \ 
          ./utils/searxng.sh instance check
...
/usr/local/searxng/searxng-src/utils/searxng_check.py:36: DeprecationWarning: /usr/local/searxng/searxng-src/utils/brand.env is no longer needed, remove the file
  warnings.warn(msg, DeprecationWarning)
```

## Related issues

Superset of: https://github.com/searxng/searxng/pull/3101
Triggered by: https://github.com/searxng/searxng/pull/3099#discussion_r1443309609


Bnyro:
This looks really great, thank you @dalf and @return42 for looking into this thing (which has really been annoying me for a while already). :tada:

I can confirm that this works well, but I can't say anything about the impact on the automated build scripts (but since CI passes, that's probably a good sign :+1:)

I just want to note that we don't actually need our own yaml parser - we could also have used the command line tool "yq" for example, which allows doing stuff like

```
❯ yq '.server.base_url' searx/settings.yml
false
```
But I assume we want to keep the amount of dependencies as few as possible, so it's just a general tip if you ever want to quickly parse yaml in the command line :+1:

utils/brand.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+GIT_BRANCH="$(git branch | grep '\*' | cut -d' ' -f2-)"
+GIT_REMOTE="$(git config "branch.${GIT_BRANCH}.remote")"
+GIT_URL="$(git config --get "remote.${GIT_REMOTE}.url")"
+if [[ "${GIT_URL}" == git@* ]]; then
+    GIT_URL="${GIT_URL/://}"
+    GIT_URL="${GIT_URL/git@/https://}"
```
dalf: This following part is not back ported, is on purpose ?

https://github.com/searxng/searxng/blob/3dea7e609bac72d4a129979ba15315b966ff020f/searx/version.py#L55-L56

The purpose is to remove the end part of `git@github.com:searxng/searxng.git`.

return42: Thanks for having an eye on .. [fixed it](https://github.com/searxng/searxng/compare/fc5c0060942cba4408efe71a2ff8d51ff75eed72..c881f9f0e1fe9b10733e3097ee4f87e3d9dc07e0)

return42:
Instance check reports "fatal: not a git repository"

```
[root@searxng-archlinux SearXNG]# ./utils/searxng.sh instance check
INFO:  wrapper:  utils/searxng.sh instance _call searxng.check
fatal: not a git repository (or any of the parent directories): .git
```

----

fixed by ...
```bash
pushd "${REPO_ROOT}" &> /dev/null
build.env.export
popd &> /dev/null
```

--> https://github.com/searxng/searxng/compare/c881f9f0e1fe9b10733e3097ee4f87e3d9dc07e0..04fa23029c5659d1bd842240642c353c8163cc55#diff-f4a3b2b2b9b088647631567e52c48be2aa850d710dae3062b131e0bee359fed8R30-R32

return42:
> I just want to note that we don't actually need our own yaml parser - we could also have used the command line tool "yq" for example, which allows doing stuff like
> But I assume we want to keep the amount of dependencies as few as possible, so it's just a general tip if you ever want to quickly parse yaml in the command line 👍

FYI: added a note about `yq` to the commit message / see 1c2e5d6

utils/brand.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# shellcheck source=utils/lib.sh
+. /dev/null
```
dalf, line 5: nitpicking: why is it needed?
(don't block the merge of this PR for this question)

return42: Its needed to prevent an warning [SC2154](https://www.shellcheck.net/wiki/SC2154) "var is referenced but not assigned" from shellcheck ..

Usually you wrote

```bash
# shellcheck source=utils/lib.sh
. utils/lib.sh
```

This will source `utils/lib.sh` and gives shellcheck the hint, that variable names in *this* (`utils/brand.sh`) script are defined in the sourced file.

Since `utils/brand.sh` is sourced by the main script `utils/searxng.sh` which already sources `utils/lib.sh` we don't want to source it again .. but we need to give shellcheck a hint from where environment variables used here are coming from.

--> Thats why we source from `/dev/null` instead:

```
# shellcheck source=utils/lib.sh
. /dev/null
```

There is an ticket in the shellcheck project about:

- https://github.com/koalaman/shellcheck/issues/356#issuecomment-853515285

dalf: Ok I understand this is related to `REPO_URL`.

dalf (commented):
I won't have time to check this PR before the end of the week, but at first sight everything seems fine and simpler.

@Bnyro if you think that's good --> go merge.

utils/brand.sh
return42: > Ok I understand this is related to `REPO_URL`.

You mean `REPO_ROOT` I guess .. then the answer is: *yes*

Bnyro (approved):
Yes, looks all good to me, thank you @dalf and @return42!

(PS: someone else has to merge it for me as I don't have the permissions)

return42:
FYI today I updated my instance (aka brand):

```
Update SearXNG instance
=======================

fetch from https://github.com/return42/searxng and reset to
origin/darmarit.org
cd /usr/local/searxng/searxng-src
git fetch origin "darmarit.org"
git reset --hard "origin/darmarit.org"
  |searxng| From https://github.com/return42/searxng
  |searxng|  * branch              darmarit.org -> FETCH_HEAD
  |searxng|  + fd32d617...d8bc894c darmarit.org -> origin/darmarit.org  (forced update)
  |searxng| HEAD is now at d8bc894c [brand] darmarit.org/searx - brand.git_branch: darmarit.org
```

As we can see (`https://github.com/return42/searxng` and `origin/darmarit.org`), the update is fetched from my brand.

Thanks a lot for your support .. and being patience when I sometimes slow in uptake :)

PR #3123, opened 2024-01-12
[fix] presearch: safesearch, time ranges, crash when no result found

### LINKED ISSUE #3115: Time-frame searches with Presearch are not correctly filtered
**SearXNG 2024.1.9+54a662658**, not forked.

`searxng-docker`

Hi there, I really like the results that I get from Presearch in SearXNG because Presearch itself is a meta search engine that pulls from the three best search indexes (Google, Bing/DuckDuckGo, and Yandex).

Apart from that I have also enabled: Brave, Mojeek, Qwant, Stract, and Yep.

I assume that the other ones except Brave don't support searches filtered by time ranges, because when I select the "Last week" time filter it usually shows only results from Presearch (with one strange exception that I'll show later). However, the results that appear are not correctly filtered.

To test this I searched for some random Linux distribution names from Distrowatch.com that aren't very popular and I filtered by "Last week".

SearXNG shows the original canonical page for the distribution name, and older reviews that mention the product name, but none of those pages have been updated within the last week.

On the other hand, the same query on presearch.com correctly shows only a few results, mainly in things like frequently updated tag / index list pages where the distribution name appeared in an old article.

----

----

----

Bnyro:
Too lazy to write a text, see #3115

closes #3115

### DIFF
--- a/docs/dev/engines/online/presearch.rst
+++ b/docs/dev/engines/online/presearch.rst
@@ -1,0 +1,13 @@
+.. _engine presearch:
+
+================
+Presearch Engine
+================
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.presearch
+   :members:
--- a/searx/engines/presearch.py
+++ b/searx/engines/presearch.py
@@ -30,6 +30,7 @@ about = {
     "results": "JSON",
 }
 paging = True
+safesearch = True
 time_range_support = True
 categories = ["general", "web"]  # general, images, videos, news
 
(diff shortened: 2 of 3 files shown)


return42 (approved):
I pushed three commits on top .. for details read the commit messages and the source code documentation.

@Bnyro what do you think about the [additions](https://github.com/searxng/searxng/compare/bc307afe..999223e) from mine .. can you please test them / thanks!

Bnyro:
> @Bnyro what do you think about the [additions](https://github.com/searxng/searxng/compare/bc307afe..999223e) from mine .. can you please test them / thanks!
>
I've tested and looked through the changes, the region support works very well and the other changes are reasonable too :+1:
Thank you!

unixfox (approved):

PR #2195, opened 2023-02-16
[mod] info page: Privacy Policy

return42:
## What does this PR do?

Initial creation of an info page on the topic "Privacy Policy".

[1] https://github.com/searxng/searxng/issues/1285#issuecomment-1431497644

Suggested-by: @nicfab [1]

## Why is this change important?

- https://github.com/searxng/searxng/issues/1285

## How to test this PR locally?

The "Privacy Policy" is assembled two places:

1. the info pages of the SearXNG instance --> `make run` --> open *About* page, klick on tab "Privacy Policy"

2. in the online documentation --> `make docs.clean docs.live`  --> jump to section "User information // Privacy Policy" 

## Related issues

Closes #1285

PR #3135, opened 2024-01-19
[feat] footer: support for custom entries

### LINKED ISSUE #3131: Custom Links in the Footer

No.

Support for custom footer links. Currently, all that can be set are links to git repos, project-related pages, and instance owner mailto:. I would like to link back to a status page for the things that I host, and any other arbitrary thing. Preferably this would be a key(link name) -> value(url) map somewhere in the config.

Editing the templates manually.

**Additional context**

Bnyro:
## What does this PR do?

* add a new setting which allows appending custom links to the footer

* it's quite common that instance maintainers want to have the possibility to advertise a custom status/donation/instance info page, so I think it's a good step to allow instance maintainers to do so

* go to `searx/settings.yml` and uncomment the `brand.custom` part
* make run -> view footer in browser

## Author's note
* in fact, we could replace all the other entries in that footer with that logic too, however with the downside that the entries can not be translated. Hence I think we should keep the common/default entries as they are.

## Related issues
closes #3131

### DIFF
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -77,8 +77,8 @@
         {% if get_setting('general.contact_url') %}
         | <a href="{{ get_setting('general.contact_url') }}">{{ _('Contact instance maintainer') }}</a>
         {% endif %}
-        {% for title, link in get_setting('brand.custom').items() %}
-        | <a href="{{ link }}">{{ title }}</a>
+        {% for title, link in get_setting('brand.custom.links').items() %}
+        | <a href="{{ link }}">{{ _(title) }}</a>
         {% endfor %}
     </p>
   </footer>
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -77,6 +77,9 @@
         {% if get_setting('general.contact_url') %}
         | <a href="{{ get_setting('general.contact_url') }}">{{ _('Contact instance maintainer') }}</a>
         {% endif %}
+        {% for title, link in get_setting('brand.custom').items() %}
+        | <a href="{{ link }}">{{ title }}</a>
+        {% endfor %}
     </p>
   </footer>
   <!--[if gte IE 9]>-->
--- a/searx/searxng.msg
+++ b/searx/searxng.msg
@@ -11,6 +11,7 @@ __all__ = [
     'CATEGORY_NAMES',
     'CATEGORY_GROUPS',
     'STYLE_NAMES',
+    'BRAND_CUSTOM_LINKS',
 ]
 
 CONSTANT_NAMES = {
@@ -50,4 +51,9 @@ STYLE_NAMES = {
     'AUTO': 'auto',
     'LIGHT': 'light',
     'DARK': 'dark',
+}
+
+BRAND_CUSTOM_LINKS = {
+    'UPTIME': 'Uptime',
+    'ABOUT': 'About',
 }
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -19,10 +19,12 @@ brand:
   public_instances: https://searx.space
   wiki_url: https://github.com/searxng/searxng/wiki
   issue_url: https://github.com/searxng/searxng/issues
-  # Custom entries in the footer: [title]: [link]
   # custom:
-  #  About instance: "https://searxng.org"
-  #  Instance admin: "https://searxng.org"
+  #   maintainer: "Jon Doe"
+  #   # Custom entries in the footer: [title]: [link]
+  #   links:
+  #     Uptime: https://uptime.searxng.org/history/darmarit-org
+  #     About: "https://searxng.org"
 
 search:
   # Filter results. 0: None, 1: Moderate, 2: Strict
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -19,6 +19,10 @@ brand:
   public_instances: https://searx.space
   wiki_url: https://github.com/searxng/searxng/wiki
   issue_url: https://github.com/searxng/searxng/issues
+  # Custom entries in the footer: [title]: [link]
+  # custom:
+  #  About instance: "https://searxng.org"
+  #  Instance admin: "https://searxng.org"
 
 search:
   # Filter results. 0: None, 1: Moderate, 2: Strict
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -151,7 +151,7 @@ SCHEMA = {
         'docs_url': SettingsValue(str, 'https://docs.searxng.org'),
         'public_instances': SettingsValue((False, str), 'https://searx.space'),
         'wiki_url': SettingsValue(str, 'https://github.com/searxng/searxng/wiki'),
-        'custom': SettingsValue(dict, {}),
+        'custom': SettingsValue(dict, {'links': {}}),
     },
     'search': {
         'safe_search': SettingsValue((0, 1, 2), 0),
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -151,6 +151,7 @@ SCHEMA = {
         'docs_url': SettingsValue(str, 'https://docs.searxng.org'),
         'public_instances': SettingsValue((False, str), 'https://searx.space'),
         'wiki_url': SettingsValue(str, 'https://github.com/searxng/searxng/wiki'),
+        'custom': SettingsValue(dict, {}),
     },
     'search': {
         'safe_search': SettingsValue((0, 1, 2), 0),


searx/templates/simple/base.html
```diff
@@ -77,6 +77,11 @@
         {% if get_setting('general.contact_url') %}
         | <a href="{{ get_setting('general.contact_url') }}">{{ _('Contact instance maintainer') }}</a>
         {% endif %}
+        {% for item in get_setting('brand.custom') %}
+          {% for key, value in item.items() %}
+          | <a href="{{ value }}">{{ key }}</a>
+          {% endfor %}
+        {% endfor %}
```
Bnyro: I don't really like that nested loop here, but we first need to iterate over a `list`, and then over the `dict` items inside.

unixfox:
I wonder if we could then deprecate the functionalities privacy policy link, donation url and contact url: https://github.com/searxng/searxng/blob/master/searx/settings.yml#L7

searx/templates/simple/base.html
return42: Haven't tested but I would expect that YAML construct is what you are looking for:

```yaml
custom:
 About instance: "https://searxng.org"
 Instance admin: "https://searxng.org"
```

Bnyro:
> I wonder if we could then deprecate the functionalities privacy policy link, donation url and contact url: https://github.com/searxng/searxng/blob/master/searx/settings.yml#L7
>
Yes, at least in my opinion, that'd be a good idea and reasonable move :+1:

searx/templates/simple/base.html
Bnyro: Works, thank you!

return42:
> > I wonder if we could then deprecate the functionalities privacy policy link, donation url and contact url: https://github.com/searxng/searxng/blob/master/searx/settings.yml#L7
> 
> Yes, at least in my opinion, that'd be a good idea and reasonable move 👍

- https://github.com/searxng/searxng/pull/2195 

.. is related .. the PR stucked at the point that we did not have [`{name of the instance maintaner}` and `{maintainer's email}`](https://github.com/searxng/searxng/pull/2195/files#r1108777544).

return42:
Pushed commit [9ea5e62](https://github.com/searxng/searxng/pull/3135/commits/9ea5e62139ed726bbcd256c7c6f13647dd50a297) on top of the branch.  Adds i18n and moves the key/val dict with links from `brand.custom` to `brand.custom.links`:

```yaml
brand:
  # ...
  custom:
    maintainer: "Jon Doe"
    # Custom entries in the footer: [title]: [link]
    links:
      Uptime: https://uptime.searxng.org/history/darmarit-org
      About: "https://searxng.org"
```

@Bnyro my intention was to leave the namespace `brand.config` open for names/settings we need in a different context / e.g. to implement https://github.com/searxng/searxng/pull/2195

What do you think about the shift of the namespace to `brand.config.links`?

Bnyro:
> What do you think about the shift of the namespace to brand.config.links?
>
I think that's a good idea, that's certainly better and more self-explaining than `brand.custom` :+1:

return42 (approved):

unixfox:
Does this support HTML? Why limiting to only links in the footer? People may want to write other things than links.

Bnyro:
> Does this support HTML? Why limiting to only links in the footer? People may want to write other things than links.
>
What would you want to place there for example?

unixfox:
> > Does this support HTML? Why limiting to only links in the footer? People may want to write other things than links.
> 
> What would you want to place there for example?

I don't have a clear example, but I know that we allowed that on invidious and people started to show useful information on their instances.

I know it's at the top of the page not in the footer, but people for example write informational messages. Example: https://invidious.projectsegfau.lt/feed/popular

return42:
> I don't have a clear example

I think your request is more related to https://github.com/searxng/searxng/issues/3156 and is not in the scope of this PR which is more related to https://github.com/searxng/searxng/issues/3131 (and opened a way for https://github.com/searxng/searxng/pull/2195)

unixfox:
I don't get it why should we implement something that needs to be exactly like the user asked in https://github.com/searxng/searxng/issues/3131?

It's an open discussion. Everyone can offer alternative implementation. We are not obligated to do it exactly like asked by the user.

On top of that my suggestion fits for the feature that the github user wants to have but also allow for more flexibility because anyone can add more than links in the footer.

And people that use this feature are not dumb, they know how to write a link in HTML.

return42:
> It's an open discussion. Everyone can offer alternative implementation. 
>> I don't have a clear example

And what shall we do now with this PR? / Sure, we can discuss it, but we can both only speculate about the additional requirements. But we can also leave the PR until someone comes up with something better ... I personally wouldn't like that, because the PR answers a specific requirement / BTW this PR does not rule out further additions at a later date - as soon as there are clear requirements.

perennialtech:
My two cents as an instance maintainer:

Allowing custom HTML in the footer would be more versatile (and what I would personally prefer). As @unixfox pointed out, we've seen that Invidious instances often have useful information in the banner. While simple links to donation and status pages are common, there are also often short notes that wouldn't warrant a link to an entire external page, where they're also less likely to be read by users.

However, allowing custom HTML also comes with potential security risks, where an instance can technically "run modified source code while still adhering to AGPL". This was covered somewhat in [issue #2166](https://github.com/iv-org/invidious/issues/2166) on the Invidious repo, where sanitising HTML was brought up as a solution, but the issue as a whole seems mostly forgotten about.
Additionally, running a strong CSP (e.g. setting `script-src 'self';`) as a hard requirement for instances to be listed on [searx.space](https://searx.space) should also be considered if this PR is merged with the change to allow custom HTML.

On the other hand, only allowing links may be more intuitive for newer maintainers who might not be familiar with YAML and correctly escaping the characters used for HTML tags (though this could be remedied for the most part via documentation).

return42:
@unixfox @perennialtech thank you very much for your valuable advice.

As is clear from @perennialtech's explanations, this requires some thought ... I have a separate feature request for this purpose ..

- https://github.com/searxng/searxng/issues/3176

In the PR it becomes clear that we can define the topic of "banner" (or whatever we want to call it) a little more broadly.  But that has nothing to do with simple texts or links in a footer ... I would like to separate this PR here -- which was created in the context of #3131 and also opens the way for #2195 -- from the "banner" topic.

unixfox:
Why was this merged???? We didn't agreed on the final implementation...

return42:
Sorry, I thought it would be okay if we evaluated your additional requirements in issue #3176, as I suggested above.  Since I got thumbs-up for the suggestion and the discussion in the new issue has already progressed, I was convinced that this PR is ready for merging.  

I'm very sorry, I wanted to progress / it was not my intention to disappoint you or the community.

unixfox:
But the point is to not have hundred ways to add "things" in the footer.

So I thought it was an ongoing discussion whenever we would implement it or not like in this PR?

If you are unsure to merge or not, just ask?

perennialtech:
Is this change going to be reverted then?

The current implementation is **fine for most cases**, but we were discussing alternatives in [issue #3176](https://github.com/searxng/searxng/issues/3176) that might serve instance maintainers better.

On the other hand, it might be better for this to stay merged and for discussion regarding something like an instance banner that contains more lengthy messages to continue as a separate matter.

unixfox:
> On the other hand, it might be better for this to stay merged and for discussion regarding something like an instance banner that contains more lengthy messages to continue as a separate matter.

The issue with that in my opinion is that this will increase the already many parameters in the settings.yml

perennialtech:
> The issue with that in my opinion is that this will increase the already many parameters in the settings.yml

A bit off-topic, but is it possible to split `settings.yml` into multiple files for easier management (e.g. configuring engines in a separate file)?

I haven't looked too much into the SearXNG codebase, so I'm not sure whether this is feasible, but if so, it could help with managing the many parameters.

return42:
> The issue with that in my opinion is that this will increase the already many parameters in the settings.yml

Yeah, thats a valid point I had also in mind .. I don't know exactly where the journey is going myself at the moment, so it's still difficult for me to reduce something in the settings ... I want to look at the further course of #3176 and I still have to think about a solution for #2195 ... the topics all deal with switches that do similar or the same thing in their function . . it would certainly be better to implement a perfect solution for all these questions once, but the considerations are all very immature at the moment ... sometimes the right ideas come to you when you first take a few steps forward ... even if you may have to take a step back along the way. 

With that in mind, I'd be reluctant to revert this PR ... let's find suitable solutions for #3176 and #2195 ... if it turns out to revert this PR or remove the settings it adds, then so be it.

Issue #984, opened 2022-03-18
Feature Request: preference export/import


Th3FatKat:
Please add some way to export and import changes in the preferences. This would cut down on time spent re-configuring the engines and other settings when moving to a new device/browser.
I am aware of the URL at the bottom of the settings with all of the changes implemented into it, but exporting and importing seems like a more secure option, as to avoid visited pages from resolving the origin URL and potentially tracking the user.

tiekoetter:
I will forward this request to the SearXNG team.

tiekoetter:
I would also agree that this approach should be looked into.

dalf:
What you suggest is to have a string that you can copy / paste from one instance to another?
Is it correct, or do you have another way in mind?

immanuelfodor:
A QR code for mobile would also be a nice touch

dalf:
About QRCode:
* it can't be an URL (?), since the purpose is to import the preferences to another instance = another URL
* so it should be text.

From the UX point of view, it would be better to scan the QRCode from the new SearXNG instance (there are some JS library to do that, not tested), but I really don't want to give access to the camera from an SearXNG instance ( ping @mrpaulblack @return42  @tiekoetter @unixfox ). 

So the solution I see right now:
  * scan the QRCode with an external application
  * copy the text
  * paste the text into a new field of the /preferences

return42:
I don't see any benefit in having a QR code .. a text field with a "copy" button (and "paste" one) gives IMO best UX.

dalf:
> I don't see any benefit in having a QR code

How do you copy / paste text from two different computers?

Currently the preferences string is more than one 1 kilobytes. An example:
```
[...]

PR #3138, opened 2024-01-21
[feat] preferences: button to copy and input to restore preferences hash

Bnyro:
## What does this PR do?
* This PR adds a button to copy the current preferences hash and an input field to restore it

2024-01-21T13:13:08,417108520+01:00

* https://github.com/searxng/searxng/issues/984#issuecomment-1074033782
* Additionally, this is compatible across different instances and hence is more user friendly than the current solution

* Go to the settings -> cookies
* copy the preferences hash
* change some settings
* restore the preferences hash
* the previously changed settings are back to where they were when copying the preferences hash

## Author's checklist
* I could additionally add a button to show a QR code for scanning the preferences hash, so that you can easily scan it with an other device to copy and restore it. However, I wasn't sure if this QR code should be generated via JavaScript (on client side) or via Python (on server side), so I'll wait for feedback on that first.

## Related issues
* addresses #984


return42:
FYI: I haven't looked at the details, but the robot test  `make test.robot` ..

- https://github.com/searxng/searxng/actions/runs/7601148476/job/20700227850#step:7:147

fails at this point 

https://github.com/searxng/searxng/blob/047c7a7cf6d5ffc6df062ff4b9e6b71126dee063/tests/robot/test_webapp.py#L46

Bnyro:
``` Element <input id="copy-hash" type="submit"> could not be scrolled into view ```

From what I figured out, the error message implies that the "copy" button is not reachable/visible on the screen, so that it can't be clicked. However, I tested this on Firefox and Chromium, but it's always visible and clickable for me, so no idea :thinking:

**Comment by return42** [comment]
> However, I tested this on Firefox and Chromium, but it's always visible and clickable for me, so no idea :thinking:

*Nope* .. its not *visible* .. at least *"not this one"* :)

The test clicks on the engine tab (`"tab-engines"`) .. then the test activates a dummy engine ..

https://github.com/searxng/searxng/blob/047c7a7cf6d5ffc6df062ff4b9e6b71126dee063/tests/robot/test_webapp.py#L41-L46

at the end the test wants to press the submit button .. for this, the test selects the !!first!! `input` tag ..

But this PR has added one more `input` tag 

https://github.com/searxng/searxng/blob/4537000fcba4c12a0d4233c66eb9e9b414c2f4fa/searx/templates/simple/preferences/cookies.html#L48-L55


which is now the !!first!! .. further (as you can see in the screenshot below) when the engine-tab is active, the (new) first `input` tag is inactive when we are in the engines tab (not reachable for the user since the !!first!! now sits in the cookie tab).
 
image

**Comment by Bnyro** [comment]
Thank you for your analysis @return42 :+1:

I solved it now by removing the `input[type=submit]` to copying the hash string and replaced it with a `button`. Initially I used `input[type=submit]` in order to achieve the same styling as for the other buttons, that's now solved by adding a new `.button` class which is styled the same way as `input[type=submit]`

File: searx/static/themes/simple/src/less/preferences.less [hunk 1]
```diff
@@ -33,6 +33,21 @@ table {
...
+    background: none repeat scroll 0 0 var(--color-toolkit-select-background);
+    padding: 0.2rem 0.4rem;
+    height: 2rem;
+    .rounded-corners-tiny;
+
+    &:hover,
+    &:focus {
+      background-color: var(--color-toolkit-select-background-hover);
+    }
+  }
```
return42, line 49: > that's now solved by adding a new `.button` class 

then I assume, this can be (also) be moved back to line 55 / back into class `.value` .. right?

Bnyro: No, that's needed for https://github.com/searxng/searxng/pull/3138/files#diff-8d9c16c3672d71c5f2d899909d0fe2552099be09accc86cf187b168da449a1b1R59

return42 (approved):

PR #3149, opened 2024-01-24
[feat] engine: implementation of goodreads

Bnyro:
## What does this PR do?
* implement a new engine, goodreads, for searching books by their names, authors or citations

* we currently only have one single search engine for books (afaik), which is Anna's archive. Anna's archive however has a way smaller library of books than goodreads, which is backed by Amazon

## Notes
* There also is a JSON Api for suggestions, but it only returns 5 items and doesn't support pagination, but it's helpful to get an overview of the engines capabilities: https://www.goodreads.com/book/auto_complete?format=json&q=foo

### DIFF
--- a/searx/engines/goodreads.py
+++ b/searx/engines/goodreads.py
@@ -1,0 +1,58 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Goodreads (books)
+"""
+
+from urllib.parse import urlencode
+
+from lxml import html
+from searx.utils import extract_text, eval_xpath, eval_xpath_list
+
+about = {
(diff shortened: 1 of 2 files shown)


searx/engines/goodreads.py
```diff
@@ -0,0 +1,59 @@
...
+
+    params['url'] = f"{base_url}/search?{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    results = []
+
+    dom = html.fromstring(resp.text)
+    print(resp.text)
```
allendema: Convert this to a debug msg or remove it.

return42 (approved):
Thanks :+1:

PR #3150, opened 2024-01-24
[fix] semantic scholar: bad request

Bnyro:
## What does this PR do?
* fixes that semantic scholar doesn't work currently
* it seems like they removed the params `useFallbackRankerService` and `useFallbackSearchCluster`, which leaded to bad requests


return42 (approved):
works! .. thanks a lot :+1:

PR #3157, opened 2024-01-27
Stats request count

dalf:
## What does this PR do?

Display the request count is percentage:
image

SearXNG has more and more engines, it is difficult to maintain all of them.
In this growing list, we don't know which one are important (except the "obvious" ones).
On a public instance, a ratio of the number of request can be helpful.

Of course, there is the privacy aspect. Some ideas:
* easy: the percentage appears after 1000 requests (or whatever number)
* easy: enable the percentage only on public instances (can be combined the previous idea).
* a lot of code: the public instances and only the public instances send the stats to searx.space which display the totals. The number of request per instance is never public.


unixfox:
just pushed this PR as a test on my instance. I forced push in order to rebase against master

https://searx.be/stats?theme=simple&sort=request

unixfox:
https://searx.be/stats?theme=simple&sort=request

perennialtech:
Some thoughts on this PR:
- There should be a small note or tooltip describing what each column means. Users may be confused by the disparity between the result count and request count, or why the request count is shown as a percentage instead of an actual integer (the lack of units on the stats page makes this even more confusing).
- Allow showing the number of requests as an integer if configured by the instance maintainer, but default to percentages. If user privacy is a concern, the request count may be rounded to avoid exposing specific metrics. For example, if the actual number of requests for an engine is:
  - 2381, round to 2400 instead.
  - 321, round to 300 instead.
- If only percentages allowed to be shown publicly, maintainers should still be allowed to see the actual number of requests, rounded or not. Maybe this can be done via the [administration API](https://docs.searxng.org/admin/api.html) with authorisation via a secret token set in `settings.yml` (similar to how `server.secret_key` is set). This would also allow searx.space to query an instance for these metrics without exposing them publicly.

vojkovic:
I like this idea. What still needs to be done before it can be merged?

thunder-coding:
> Allow showing the number of requests as an integer if configured by the instance maintainer, but default to percentages. If user privacy is a concern, the request count may be rounded to avoid exposing specific metrics.

This opens up a way for side-channel attacks where an attacker can make bogus searches and monitor the statistics to make an estimate of how many searches were actually made.

vojkovic:
> This opens up a way for side-channel attacks where an attacker can make bogus searches and monitor the statistics to make an estimate of how many searches were actually made.

I understand the attack you are talking about, and please correct me if I'm wrong, but I don't see how estimating the number of searches on an public instance will impact the privacy of any users.

IMO this feature should be enabled by default on public instances.

vojkovic:
The Request Count to engines was added in https://github.com/searxng/searxng/pull/3835 hidden behind a http auth.

return42:
FYI: last force push was a rebase on current master to solve conflicts (migration to vitejs)

return42 (approved):
> The Request Count to engines was added in https://github.com/searxng/searxng/pull/3835 hidden behind a http auth.

@vojkovic thanks for reminding me .. 

@dalf I lost this draft out of my scope (due to lack of time I rarely look into drafts / sorry) .. here is what I have done:

- moved commit ae1e73e871f8 from this branch to https://github.com/searxng/searxng/pull/4510
- rebased on current master (migrated to vitejs)
- squashed the two commit 037ab2738 and 3c37cf4 into one commit (and slightly modified the commit message) 

From my point of view, this PR is ready for the merger. Are there still any doubts?

PR #3173, opened 2024-01-31
[mod] remove X-XSS-Protection headers

### LINKED ISSUE #3171: Remove `X-XSS-Protection` headers. (Deprecated header not used by browsers nowadays)

No

Consider the removal of X-XSS-Protection headers in:
https://github.com/searxng/searxng/blob/90072eb6ff2a7c402ab47f2d8c973e6b24116795/searx/settings.yml#L91
https://github.com/searxng/searxng/blob/90072eb6ff2a7c402ab47f2d8c973e6b24116795/tests/unit/settings/user_settings.yml#L22
https://github.com/searxng/searxng/blob/90072eb6ff2a7c402ab47f2d8c973e6b24116795/docs/admin/settings/settings_server.rst#L19

Make use of the `Content-Security-Policy` header instead.

**Additional context**

I will not reinvent the wheel so I will just cite the sources.

*"In modern browsers, X-XSS-Protection has been deprecated in favor of the Content-Security-Policy to disable the use of inline JavaScript. Its use can introduce XSS vulnerabilities in otherwise safe websites. This should not be used unless you need to support older web browsers that don’t yet support CSP. It is thus recommended to set the header as X-XSS-Protection: 0."* 
https://infosec.mozilla.org/guidelines/web_security#x-xss-protection

There is no need to use `X-XSS-Protection` unless we want to support the <2010 version of Chrome or the <2015 version of Microsoft Edge. Other browsers like firefox never implemeneted or make use of this header so it makes no sense to use it nowadays because no browser uses it. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

return42:
Deprecated header not used by browsers nowadays[1]:

"""In modern browsers, X-XSS-Protection has been deprecated in favor of the Content-Security-Policy to disable the use of inline JavaScript. Its use can introduce XSS vulnerabilities in otherwise safe websites. This should not be used unless you need to support older web browsers that don’t yet support CSP. It is thus recommended to set the header as X-XSS-Protection: 0."""[2]

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
[2] https://infosec.mozilla.org/guidelines/web_security#x-xss-protection

Closes: https://github.com/searxng/searxng/issues/3171

### DIFF
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -16,7 +16,6 @@
        image_proxy: false
        default_http_headers:
          X-Content-Type-Options : nosniff
-         X-XSS-Protection : 1; mode=block
          X-Download-Options : noopen
          X-Robots-Tag : noindex, nofollow
          Referrer-Policy : no-referrer
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -88,7 +88,6 @@ server:
   method: "POST"
   default_http_headers:
     X-Content-Type-Options: nosniff
-    X-XSS-Protection: 1; mode=block
     X-Download-Options: noopen
     X-Robots-Tag: noindex, nofollow
     Referrer-Policy: no-referrer
--- a/tests/unit/settings/user_settings.yml
+++ b/tests/unit/settings/user_settings.yml
@@ -19,7 +19,6 @@ server:
   method: "POST"
   default_http_headers:
     X-Content-Type-Options: nosniff
-    X-XSS-Protection: 1; mode=block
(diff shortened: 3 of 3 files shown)


Bnyro (approved):

unixfox (approved):
caniuse show that 98% of the browsers versions used globally support CSP: https://caniuse.com/contentsecuritypolicy

The only "major" browsers that will miss from the XSS protection will be internet explorer.

I'm ok with the change.

PR #3169, opened 2024-01-29
[fix] Improve engines line alignment and formatting

allixx:
## What does this PR do?

Fixes vertical alignment of engines and cached/proxied links. Adds missing space between cached and proxied links.

Replaces seemingly useless left padding of engines div with top padding. Having top padding here really helps to add a bit of visual separation of engines line (it often gets quite crowded with multiple engines) and search result.

## Why is this change important?

Those pixel discrepancies really hurt the eye. No visual separation between cached and proxied link is a cherry on top.

Before:
before

After:
after

PR #3174, opened 2024-01-31
[fix] remove missing link icon macro

allixx:
## What does this PR do?

Removes non-existing 'link' icon macro call from results sub footer expression.

There's no point having a macro call which is actually a dud.


return42:
Since we have merged your PR #3169, to solve the conflict you have to rebase this branch on current master  

> Removes non-existing 'link' icon macro call 

Yeah there is no need for a link-icon and `link` is not in the list of icons .. thanks for having an eye on :+1: 

https://github.com/searxng/searxng/blob/ab8e5383fb6021afd690060c7b718dc505e7d30c/searx/static/themes/simple/gruntfile.js#L173-L193

return42 (approved):

PR #3186, opened 2024-02-03
[fix] remove certifi from requirements.txt

return42:
- Requirement certify was added in 35a2bc5650b

- Since commit 93f7f7eee certifi is no longer needed.  Not sure why 93f7f7eee upgraded certifi while removing the usage of this package from the source code in the same commit.

Closes: https://github.com/searxng/searxng/pull/3182

@dalf do you see any reasons not to remove certifi?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-certifi==2023.11.17
 babel==2.14.0
 flask-babel==4.0.0
 flask==3.0.1


dalf:
certifi is a dependency of httpx, as httpcore, however the package list the sources of trust for HTTPS.
That's why I think we should keep certifi as an explicit dependency.

PR #3189, opened 2024-02-07
[feat] engine: implementation of ask.com

Austin-Olacsi:
## What does this PR do?

adds ask.com search engine

the results are similar to other search engines, but i was not able to find where they get results from.

searxng !ask

### DIFF
--- a/searx/engines/ask.py
+++ b/searx/engines/ask.py
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Ask.com"""
+
+from urllib.parse import urlencode
+import re
+from lxml import html
+
+# Metadata
+about = {
+    "website": "https://www.ask.com/",
+    "wikidata_id": 'Q847564',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "HTML",
+}
+
+# Engine Configuration
+categories = ['general']
+paging = True
+
+# Base URL
+base_url = "https://www.ask.com/web"
+
+
+def request(query, params):
+
+    query_params = {
+        "q": query,
+        "page": params["pageno"],
+    }
+
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+    return params
+
+
+def response(resp):
+
+    text = html.fromstring(resp.text).text_content()
+    urls_match = re.findall(r'"url":"(.*?)"', text)
+    titles_match = re.findall(r'"title":"(.*?)"', text)[3:]
+    content_match = re.findall(r'"abstract":"(.*?)"', text)
+
+    results = [
+        {
+            "url": url,
+            "title": title,
+            "content": content,
+        }
+        for url, title, content in zip(urls_match, titles_match, content_match)
+        if "&qo=relatedSearchNarrow" not in url
+        # Related searches shouldn't be in the search results: www.ask.com/web&q=related
+    ]
+
+    return results
--- a/searx/engines/ask.py
+++ b/searx/engines/ask.py
@@ -3,8 +3,9 @@
 """Ask.com"""
 
 from urllib.parse import urlencode
-import re
+import dateutil
 from lxml import html
+from searx import utils
 
 # Metadata
 about = {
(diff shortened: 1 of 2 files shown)


unixfox:
please disable the engine by default

searx/engines/ask.py
```diff
@@ -0,0 +1,56 @@
...
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+    return params
+
+
+def response(resp):
+
+    text = html.fromstring(resp.text).text_content()
+    urls_match = re.findall(r'"url":"(.*?)"', text)
+    titles_match = re.findall(r'"title":"(.*?)"', text)[3:]
+    content_match = re.findall(r'"abstract":"(.*?)"', text)
```
return42: grep a regexp in a XML (HTML) content is fragile (consider the regexp pattern is in the result itself / or in the search term) .. like in all the other engines we should use xpath selectors.

A good starting point are the xpath selectors I have stolen from @allendema's PR https://github.com/searx/searx/pull/2898 / A XPATH engine that we have never merged into SearXNG (was merged into the old project searx but not here in this project SearXNG)

untested:

```python

    title_xpath = '//a[@class="PartialSearchResults-item-title-link result-link"]'
    url_xpath = '//a[@class="PartialSearchResults-item-title-link result-link"]/@href'
    content_xpath = '//p[@class="PartialSearchResults-item-abstract"]'
    suggestion_xpath = '//span[@class="PartialRelatedSearch-item-link-text"]'
```

return42 (changes requested):

unixfox (approved):

Austin-Olacsi:
is xpath really suited for this? Maybe it is not what we need, because what i am dealing with is similar to this: https://stackoverflow.com/questions/40787982/xpath-javascript-in-python

return42:
> is xpath really suited for this? 

Ah, now I see .. you are parsing JS .. for this we have the very useful `utils.js_variable_to_python()` function.  I pushed commit 11c2d30b4 on top to parse the JS.

Can you please test / is it OK?

return42 (approved):

searx/engines/ask.py
```diff
@@ -0,0 +1,78 @@
...
+    start_tag = 'window.MESON.initialState = {'
+    end_tag = '}};'
+
+    dom = html.fromstring(resp.text)
+    script = utils.eval_xpath_getindex(dom, '//script', 0, default=None).text
+
+    pos = script.index(start_tag) + len(start_tag) - 1
+    script = script[pos:]
+    pos = script.index(end_tag) + len(end_tag) - 1
+    script = script[:pos]
```
Bnyro, line 50:
```suggestion
    start_pos = script.index(start_tag) + len(start_tag) - 1
    end_pos = script.index(end_tag) + len(end_tag) - 1
    script = script[start_pos:end_pos]
```

searx/engines/ask.py
```diff
@@ -0,0 +1,78 @@
...
+    for item in json_resp['search']['webResults']['results']:
+        pubdate_original = item.get('pubdate_original')
+        if pubdate_original:
+            pubdate_original = dateutil.parser.parse(pubdate_original)
+
+        metadata = []
+        for field in ['category_l1', 'catsy']:
+            value = item.get(field)
+            if value:
+                metadata.append(value)
```
Bnyro:
```suggestion
        fields = ['category_l1', 'catsy']
        metadata = [item.get(field) for field in fields if item.get(field)] 
```

searx/engines/ask.py
Bnyro: Or wait, the previous way probably has a slightly better performance because the second `index` call has to read less characters?

return42: > call has to read less characters?

My version asserts that the `end_tag` is not taken from a string before `start_tag` / assume start_tag is `= {`  end_tag is `};` .. then in your version the first `end_tag` is in the first line but the `start_tag` is in the second line 

```
 window.MESON = window.MESON || {};
 window.MESON.initialState = {"siteConfig": ...
  ...}};
```

searx/engines/ask.py
return42: :+1: --> [done](https://github.com/searxng/searxng/compare/11c2d30b456064c22b29d0a47152a87014649bef..b26e76a1e10056fcfc56f8ffbf1902d97f2ff59f)

Bnyro (approved):

searx/engines/ask.py
Bnyro: Oh, that makes sense :+1:

Austin-Olacsi:
LGTM

PR #3190, opened 2024-02-08
Add 2 new language options for Morocco (country specific)

Midou36O:
## What does this PR do?

It adds a 2 new languages, fr-MA and ar-MA, both in the respective lines. Nothing more.

There are country specific queries that cannot be found without using this specific language for some reason. I need it to do some research on local queries.

just run `make run` and pick the option.

## Author's checklist

N/A.

## Related issues

N/A.


return42:
@Midou36O thanks a lot for your PR .. sadly we can't accept the PR / however there is a solution for your task, but first let me explain:

The list of languages and regions you want to modify is auto-generated, based on the languages and regions we fetch from the origin search engines / we call it the [engine_traits](https://github.com/searxng/searxng/blob/master/searx/data/engine_traits.json) ([doc](https://docs.searxng.org/dev/engines/enginelib.html#module-searx.enginelib.traits)).

If you scroll down along the engine_traits lines, you see a lot of languages and regions.  Since it does not make sense to put all of them in the drop down box in the UI we filter the list of languages and regions shown in the UI:

https://github.com/searxng/searxng/blob/df1a774003c285866a96b149bf92412037b4932d/searxng_extra/update/update_engine_traits.py#L100-L104

With the threshold we try to keep the UI usable .. TBH it has already been criticized that the list of languages and regions is still too long.

----

However, it is also possible to use languages or regions other than those shown in the drop-down box / of course, this only makes sense if there is support for the region in one of the search engines.

For those origin engines that do not support a language or region (aka locale) selected by the user, SearXNG tries to *find* a alternative locale of the engine that best matches the one selected by the user [searx.locales.get_engine_locale](https://docs.searxng.org/src/searx.locales.html#searx.locales.get_engine_locale)

> There are country specific queries that cannot be found without using this specific language for some reason. I need it to do some research on local queries.

To select a language or region (aka locale) you can always use the [search syntax](https://docs.searxng.org/user/search-syntax.html) .. let assume you are looking for sides related to the car brand BMW in Maroc: 

- [`:ar-MA BMW المغرب`](https://darmarit.org/searx/search?q=:ar-MA+BMW+%D8%A7%D9%84%D9%85%D8%BA%D8%B1%D8%A8)
- [`:fr-MA BMW Maroc`](https://darmarit.org/searx/search?q=:fr-MA+BMW+Maroc)

I hope the explanation above helps you to get your tasks done / I'm going close this PR without merge and ask for your understanding.

Midou36O:
Understandable, I thought that the list was incomplete with the number of specific countries and languages listed in there. I should've searched longer and made an issue before a PR. Thanks a lot for taking your time and giving me a complete explanation of what I am supposed to do.

PR #3192, opened 2024-02-09
[feat] engine: implementation of pixiv

Austin-Olacsi:
## What does this PR do?

adds the pixiv engine

pixiv is a popular alternative to deviantart

jigglypuff !pv

## Author's checklist

we cannot load the images directly from pixiv, see here: https://github.com/searxng/searxng/pull/2772

However, there is an alternative frontend for pixiv: https://codeberg.org/VnPower/PixivFE
using proxy from a random instance to load the images.

### DIFF
--- a/searx/engines/pixiv.py
+++ b/searx/engines/pixiv.py
@@ -1,0 +1,72 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Pixiv (images)"""
+
+from urllib.parse import urlencode
+import random
+
+# Engine metadata
+about = {
+    "website": 'https://www.pixiv.net/',
+    "wikidata_id": 'Q306956',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+# Engine configuration
+paging = True
+categories = ['images']
+
+# Search URL
+base_url = "https://www.pixiv.net/ajax/search/illustrations"
+pixiv_image_proxies: list = []
+
+
+def request(query, params):
+    query_params = {
+        "word": query,
+        "order": "date_d",
+        "mode": "all",
+        "p": params["pageno"],
+        "s_mode": "s_tag_full",
+        "type": "illust_and_ugoira",
+        "lang": "en",
+    }
+
+    params["url"] = f"{base_url}/{query}?{urlencode(query_params)}"
+
+    return params
+
+
+def response(resp):
+    results = []
+    data = resp.json()
+
+    for item in data["body"]["illust"]["data"]:
+
+        image_url = item["url"]
+        pixiv_proxy = random.choice(pixiv_image_proxies)
+        proxy_image_url = image_url.replace("https://i.pximg.net", pixiv_proxy)
+        proxy_full_image_url = (
+            proxy_image_url.replace("/c/250x250_80_a2/", "/")
+            .replace("_square1200.jpg", "_master1200.jpg")
+            .replace("custom-thumb", "img-master")
+            .replace("_custom1200.jpg", "_master1200.jpg")
+        )
+
+        results.append(
+            {
+                "title": item.get("title"),
+                "url": proxy_full_image_url,
+                'content': item.get('alt'),
+                "author": f"{item.get('userName')} (ID: {item.get('userId')})",
+                "img_src": proxy_full_image_url,
+                "thumbnail_src": proxy_image_url,
+                "source": 'pixiv.net',
+                "template": "images.html",
+            }
+        )
+
+    return results
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1334,6 +1334,31 @@ engines:
     url: https://thepiratebay.org/
     timeout: 3.0
 
+  - name: pixiv
+    shortcut: pv
+    engine: pixiv
+    disabled: true
+    inactive: true
+    pixiv_image_proxies:
+      - pximg.example.org
+      # A proxy is required to load the images. Hosting an image proxy server
+      # for Pixiv:
+      #    --> https://codeberg.org/VnPower/PixivFE/wiki/Hosting-an-image-proxy-server-for-Pixiv
+      # Proxies from public instances.  Ask the public instances owners if they
+      # agree to receive traffic from SearXNG!
+      #    --> https://codeberg.org/VnPower/PixivFE#instances
+      #    --> https://github.com/searxng/searxng/pull/3192#issuecomment-1941095047
+      # image proxy of https://pixiv.cat
+      # - https://i.pixiv.cat
+      # image proxy of https://www.pixiv.pics
+      # - https://pximg.cocomi.eu.org
+      # image proxy of https://pixivfe.exozy.me
+      # - https://pximg.exozy.me
+      # image proxy of https://pixivfe.ducks.party
+      # - https://pixiv.ducks.party
+      # image proxy of https://pixiv.perennialte.ch
+      # - https://pximg.perennialte.ch
+
   - name: podcastindex
     engine: podcastindex
     shortcut: podcast


searx/engines/pixiv.py
```diff
@@ -0,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Pixiv.net"""
```
Bnyro:
```suggestion
"""Pixiv (images)"""
```

searx/engines/pixiv.py
```diff
@@ -24,6 +25,22 @@
...
 
+# Check for working proxies
+def check_proxies(proxies):
+    working_proxies = []
+
+    for proxy in proxies:
+        proxy_test_url = f"{proxy}/c/250x250_80_a2/img-master/img/2024/02/12/17/41/49/115989389_p0_square1200.jpg"
+
+        try:
+            test_response = httpx.get(proxy_test_url)
```
Bnyro: You should use `get` from `searx.network`.

searx/engines/pixiv.py
```diff
@@ -44,13 +61,14 @@ def request(query, params):
 def response(resp):
     results = []
     data = resp.json()
+    working_proxies = check_proxies(Proxy)
```
Bnyro: That will take a lot of time for each request and we don't test whether the proxy is working for engines like Piped or Invidious either, it's up to the instance maintainer to pick reliable proxies.

I think the changes with the latest commit should be undone.

return42: To check while engine is initialized, you can implement a [`init()`](https://docs.searxng.org/dev/engines/demo/demo_online.html) function in your engine module.

Here is by example what elasticsearch engine checks on initialization:

https://github.com/searxng/searxng/blob/11c0651ef47a72395d2d386d7600d951b0987ea0/searx/engines/elasticsearch.py#L59-L64

searx/engines/pixiv.py
return42: When you use the get method from SearXNG's network stack, don't import the `get` / import the complete module and use it like shown in the example below:

```python
from searx import network
network.get(proxy_test_url)
```

searx/settings.yml
```diff
@@ -1334,6 +1334,17 @@ engines:
...
+  - name: pixiv
+    engine: pixiv
+    shortcut: pv
+    disabled: true
+    # A proxy is required to load the images.
+    # Find more or host your own: https://codeberg.org/VnPower/PixivFE
+    Proxy:
+      - https://pixiv.tatakai.top
+      - https://img.chaotic.ninja
+      - https://i.pixiv.cat
```
return42: To avoid naming conflicts/confusion with the [`proxies` settings from SearXNG's](https://docs.searxng.org/admin/settings/settings_engine.html) network stack you should better name the proxies like:

```YAML

    pixiv_image_proxy:
      - https://pixiv.tatakai.top
      - https://img.chaotic.ninja
      - https://i.pixiv.cat
```

searx/engines/pixiv.py
Austin-Olacsi: tried your example but got an error:
`TypeError: type of pixiv.network is a module (searx.network), expected a string
make: *** [Makefile:28: run] Error 1
`

searx/settings.yml
unixfox: Or just "instances:"

unixfox:
My only issue is that this will probably add a lot of load on the instances hardcoded in the code because loading many images at the same time is very bandwidth consuming.

searx/settings.yml
return42: > Or just "instances:"

I vote for a name that makes it clear that this setting is related to the engine / has the engine name a prefix: `pixiv_image_proxy` .. or .. `pixiv_instances` .. or ..

@Austin-Olacsi I can fix it in my review / if you want.

searx/engines/pixiv.py
return42: > tried your example but got an error:

yes, its about the horrible monkey patching we have .. leave it as is now, I will fix it in my review .. BTW: thanks a lot for implementing the `init()` function :)

return42:
> My only issue is that this will probably add a lot of load on the instances hardcoded in the code because loading many images at the same time is very bandwidth consuming.

The engine has paging support / not sure what we can do more to reduce the amount of images in one page .. any suggestions? .. AFAIK we have this issue in several other image engines.

unixfox:
> > My only issue is that this will probably add a lot of load on the instances hardcoded in the code because loading many images at the same time is very bandwidth consuming.
> 
> The engine has paging support / not sure what we can do more to reduce the amount of images in one page .. any suggestions? .. AFAIK we have this issue in several other image engines.

But we do not have images engines that are run by individuals. The only images engines are run by organizations that have many servers behind it and are designed to serve the traffic to many users.

I'm ok with invidious and piped instances because it's low traffic (thumbnails and API results) but here it's around 70 images that are being loaded by each page result.

On top of that, **we have not asked the public instances owners if they agree to receive traffic from SearXNG!**.

Austin-Olacsi:
My recent changes (summary):

1. Made fullsize_img work for all images
2. renamed Proxy to pixiv_image_proxy and told pylint to shut up about the new name
3. commented-out this engine in settings.yml and removed the hard-coded instances. i used an example url instead.
4. removed the init function, since the instance maintainer now needs to pick the proxy himself anyway, it's not necessary to check the proxy for reliability. we can add this function back but i didn't feel like working on it.
5. included a link with some fairly hidden and useful info- you can easily make a pixiv image proxy with only nginx and you do not need to install this alternative frontend for pixiv.

return42:
> My recent changes (summary):

Thanks :+1: ... I'm going to review your PR .. I would like to squash the commits into the first commit .. do you know why 11b649c0f has @Hackurei as Author and Commit

image

Austin-Olacsi:
that was my old github name... in my most recent commit https://github.com/searxng/searxng/commit/09d66dd41a1e2a726c6c506ccb62da1a5a01a007 is it fixed?

searx/settings.yml
```diff
@@ -1334,6 +1334,31 @@ engines:
     url: https://thepiratebay.org/
     timeout: 3.0
 
+  - name: pixiv
+    shortcut: pv
+    engine: pixiv
+    disabled: true
+    inactive: true
+    pixiv_image_proxies:
+      - pximg.example.org
```
Bnyro, line 1343: I think this line should better be removed or commented since it's not an actual working proxy url

Bnyro (approved):

return42:
> that was my old github name... in my most recent commit [09d66dd](https://github.com/searxng/searxng/commit/09d66dd41a1e2a726c6c506ccb62da1a5a01a007) is it fixed?

Yeah is fixed / I squashed some modifications into your commit / see [diff](https://github.com/searxng/searxng/compare/09d66dd41a1e2a726c6c506ccb62da1a5a01a007..2de8bdd06fa69a2b71b40efcd3d75a0802127cc1).

> we have not asked the public instances owners if they agree to receive traffic from SearXNG!

@unixfox I set the engine to `inactive: true` and added a comment to the `settings.yml`.

With setting from below, `make run` and use search syntax `!pv bmw`

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index a8ae4d57c..2ded53c50 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1338,9 +1338,9 @@ engines:
     shortcut: pv
     engine: pixiv
     disabled: true
-    inactive: true
+    # inactive: true
     pixiv_image_proxies:
-      - pximg.example.org
+      # - pximg.example.org
       # A proxy is required to load the images. Hosting an image proxy server
       # for Pixiv:
       #    --> https://codeberg.org/VnPower/PixivFE/wiki/Hosting-an-image-proxy-server-for-Pixiv
@@ -1349,15 +1349,15 @@ engines:
       #    --> https://codeberg.org/VnPower/PixivFE#instances
       #    --> https://github.com/searxng/searxng/pull/3192#issuecomment-1941095047
       # image proxy of https://www.pixiv.net
-      # - https://i.pixiv.cat
+      - https://i.pixiv.cat
       # image proxy of https://www.pixiv.pics
-      # - https://pximg.cocomi.eu.org
+      - https://pximg.cocomi.eu.org
       # image proxy of https://pixivfe.exozy.me
-      # - https://pximg.exozy.me
+      - https://pximg.exozy.me
       # image proxy of https://pixivfe.ducks.party
-      # - https://pixiv.ducks.party
+      - https://pixiv.ducks.party
       # image proxy of https://pixiv.perennialte.ch
-      # - https://pximg.perennialte.ch
+      - https://pximg.perennialte.ch
```

searx/settings.yml
Bnyro: Oh, I guess it's related to the `settings.yml.new` auto-new-changes-adder, right?

return42: If I comment the `pximg.example.org` list item out the indentation of the comments are reported as error:

```
make test
TEST      [yamllint] $YAMLLINT_FILES
searx/settings.yml:1343:7: [warning] comment not indented like content (comments-indentation)
ERROR: test.yamllint exit with error (2)
make: *** [Makefile:95: test.yamllint] Fehler 2
```

Here is the config that produce error from above:

```yaml
  - name: pixiv
    shortcut: pv
    engine: pixiv
    disabled: true
    inactive: true
    pixiv_image_proxies:
      # - pximg.example.org
      # A proxy is required to load the images. Hosting an image proxy server
      # for Pixiv:
      #    --> https://codeberg.org/VnPower/PixivFE/wiki/Hosting-an-image-proxy-server-for-Pixiv
      # Proxies from public instances.  Ask the public instances owners if they
      # agree to receive traffic from SearXNG!
      #    --> https://codeberg.org/VnPower/PixivFE#instances
      #    --> https://github.com/searxng/searxng/pull/3192#issuecomment-1941095047
      # image proxy of https://www.pixiv.net
      # - https://i.pixiv.cat
      # image proxy of https://www.pixiv.pics
      # - https://pximg.cocomi.eu.org
      # image proxy of https://pixivfe.exozy.me
      # - https://pximg.exozy.me
      # image proxy of https://pixivfe.ducks.party
      # - https://pixiv.ducks.party
      # image proxy of https://pixiv.perennialte.ch
      # - https://pximg.perennialte.ch
```

Bnyro: Thanks for the explanation :+1:

return42 (approved):

Austin-Olacsi:
ready 👍

Issue #2568, opened 2023-07-23
Add Shortcut for focusing on search bar, focusing subject tabs, navigating search items with arrow key.


Oymate:
**Is your feature request related to a problem? Please describe.**
Keyboard accessibility is poor in searx currently. Shortcuts will alleviate this along with increasing accessibility.

**Describe the solution you'd like**
Add Shortcut for focusing on search bar, focusing subject tabs, navigating search items with arrow key.

**Describe alternatives you've considered**
Using mouse button which is slow and sometimes inaccessible.

**Additional context**

devloberto:
I would love a shortcut like `Ctrl`+`/` and/or `Ctrl`+`k` (this are common ones) to focus the search bar 🙂

Oymate:
Actually, it turns out the feature has already been implemented, but the shortcuts are a bit wack. Press 'h' to bring up the shortcut guide. Right-Left arrow to navigate, T/O to open results, 'I' (eye) to focus search box, N/P to navigate pages, C to copy result link and R to reload the page.

PR #3199, opened 2024-02-09
Add UI feature: shortcut to focus searchbar

devloberto:
## add "/" shortcut to focus searchbar

focus the searchbar with the cursor placed after the last input character by just typing `/` ("shift" + "7" or numpad "/").
implemented by a few lines of js in `searx/static/themes/simple/src/js/main/search.js`.

It solves partially the feature request [Add Shortcut for focusing on search bar, focusing subject tabs, navigating search items with arrow key.](https://github.com/searxng/searxng/issues/2568)

### Just manual testing
there are various test cases
1. setup:
    - test case 1: open the root page `<base-url>/`, leave the searchbar empty and remove the focus
    - test case 2: open the root page `<base-url>/`, type some text into the searchbar and remove the focus
    - test case 3: open the result page `<base-url>/search?q=searxng`
    - test case 4: open the result page `<base-url>/search?q=searxng`, clear the searchbar and remove the focus
2. execution: type "/"
3. assertion: the searchbar should be focused and the cursor should be placed after the last input character

### DIFF
--- a/searx/static/themes/simple/src/js/main/search.js
+++ b/searx/static/themes/simple/src/js/main/search.js
@@ -37,8 +37,9 @@
 
   searxng.ready(function () {
     qinput = d.getElementById(qinput_id);
+    const wasSearchBarFound = qinput !== null;
 
-    if (qinput !== null) {
+    if (wasSearchBarFound) {
       // clear button
       createClearButton(qinput);
 
@@ -155,7 +156,7 @@
     // filter (safesearch, time range or language) (this requires JavaScript
     // though)
     if (
-      qinput !== null
+      wasSearchBarFound
         && searxng.settings.search_on_category_select
       // If .search_filters is undefined (invisible) we are on the homepage and
       // hence don't have to set any listeners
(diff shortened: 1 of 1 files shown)


return42:
Thanks a lot for your PR :+1: .. haven't looked at the details, but do you have reasons not to add the key binding to the tables in:

https://github.com/searxng/searxng/blob/b0fe0924a3fbfbc6a8e9459fa14a12d1ef4e31d8/searx/static/themes/simple/src/js/main/keyboard.js#L72-L77

Further, to build the commit use `make static.build.commit` / please have a look into our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html)

devloberto:
thx for the hint. tbh i just didn't notice the `keyboard.js` even though it's obvious 😅 

unfortunately adding the `'/'` key map to the `baseKeyBinding` does only work with the numpad "/" because of the check `!e.shiftKey` in [keyboard.js#L187](https://github.com/searxng/searxng/blob/b0fe0924a3fbfbc6a8e9459fa14a12d1ef4e31d8/searx/static/themes/simple/src/js/main/keyboard.js#L187).
```
    '/': {
      key: '/',
      fun: searchInputFocus,
      des: 'focus on the search input',
      cat: 'Control'
    },
``` 
that's not what i wanted to achieve.
furthermore i don't think removing the `shiftKey` check from the _check for modifiers to don't break browser's hotkeys_ was a good idea.
so i will settle for the "i" shortcut and close the pr.

nevertheless, thx for your support 🙂

PR #3200, opened 2024-02-09
[feat] engine: NixOS search / nixpkgs

v411e:
## What does this PR do?

This PR will add a new engine which can be used to search for Nix packages using https://search.nixos.org

I first used the existing Elasticsearch engine, but although it somehow worked, the results were not great.

I would like to quickly search for Nix packages.

Just set up the engine using the provided example in the docs or `settings.yml`. Then search for a package like "firefox".

## Author's checklist

None

## Related issues

None

### DIFF
--- a/searx/engines/nixpkgs.py
+++ b/searx/engines/nixpkgs.py
@@ -1,0 +1,164 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""`Nixpkgs Search <https://search.nixos.org>`_ makes it easy to search the huge
+nixpkgs package index.  The search backend on search.nixos.org is an
+Elasticsearch instance.
+
+Example
+=======
+
+The following is an example configuration for an nixpkgs engine with
+authentication configured.
+
+Credentials are available here: https://github.com/NixOS/nixos-search/blob/main/frontend/src/index.js
+
+.. code:: yaml
+
+   - name: nixpkgs
+     shortcut: nix
+     engine: nixpkgs
+     base_url: https://search.nixos.org/backend
+     username:
+     password:
+     index: latest-42-nixos-unstable
+     channel: unstable
+
+"""
+
+from searx.exceptions import SearxEngineAPIException
+
+about = {
+    "website": "https://search.nixos.org",
+    "use_official_api": False,
+    "require_api_key": True,
+    "results": 'JSON',
+}
+categories = ["it", "packages"]
+
+base_url = "https://search.nixos.org/backend"
+username = ""
+password = ""
+index = "latest-42-nixos-unstable"
+channel = "unstable"
+search_url = f"{base_url}/{index}/_search"
+show_metadata = False
+categories = ["it", "packages"]
+
+
+def init():
+    if not (username and password):
+        raise ValueError("username and password need to be configured.")
+
+
+def request(query, params):
+    params["auth"] = (username, password)
+    params["url"] = search_url
+    params["data"] = _build_query(query)
+    params["headers"]["Content-Type"] = "application/json"
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    resp_json = resp.json()
+    if "error" in resp_json:
+        raise SearxEngineAPIException(resp_json["error"])
+
+    for result in resp_json["hits"]["hits"]:
+        raw_result = {key: value if not key.startswith("_") else value for key, value in result["_source"].items()}
+        r = {
+            "template": "packages.html",
+            "title": raw_result["package_attr_name"],
+            "package_name": raw_result["package_pname"],
+            "version": raw_result["package_pversion"],
+            "content": raw_result["package_description"] or "",
+            "source_code_url": _position_to_github_url(raw_result["package_position"]),
+            "url": f"https://search.nixos.org/packages?channel={channel}&show={raw_result['package_attr_name']}",
+            "maintainer": ", ".join(raw_result["package_maintainers_set"]),
+            "tags": raw_result["package_programs"],
+            "license_name": ", ".join(raw_result["package_license_set"]),
+            "homepage": next(iter(raw_result["package_homepage"]), None),
+        }
+
+        if show_metadata:
+            r["metadata"] = {
+                "index": result["_index"],
+                "id": result["_id"],
+                "score": result["_score"],
+            }
+
+        results.append(r)
+
+    return results
+
+
+def _build_query(query: str):
+    return f"""
+      {{
+        "from": 0,
+        "size": 50,
+        "sort": [
+          {{
+            "_score": "desc",
+            "package_attr_name": "desc",
+            "package_pversion": "desc"
+          }}
+        ],
+        "query": {{
+          "bool": {{
+            "must": [
+              {{
+                "dis_max": {{
+                  "tie_breaker": 0.7,
+                  "queries": [
+                    {{
+                      "multi_match": {{
+                        "type": "cross_fields",
+                        "query": "{query}",
+                        "analyzer": "whitespace",
+                        "auto_generate_synonyms_phrase_query": false,
+                        "operator": "and",
+                        "_name": "multi_match_firefox",
+                        "fields": [
+                          "package_attr_name^9",
+                          "package_attr_name.*^5.3999999999999995",
+                          "package_programs^9",
+                          "package_programs.*^5.3999999999999995",
+                          "package_pname^6",
+                          "package_pname.*^3.5999999999999996",
+                          "package_description^1.3",
+                          "package_description.*^0.78",
+                          "package_longDescription^1",
+                          "package_longDescription.*^0.6",
+                          "flake_name^0.5",
+                          "flake_name.*^0.3"
+                        ]
+                      }}
+                    }},
+                    {{
+                      "wildcard": {{
+                        "package_attr_name": {{
+                          "value": "*{query}*",
+                          "case_insensitive": true
+                        }}
+                      }}
+                    }}
+                  ]
+                }}
+              }}
+            ]
+          }}
+        }}
+      }}
+    """
+
+
+def _position_to_github_url(package_position: str):
+    path, line = package_position.split(":")
+    return f"https://github.com/NixOS/nixpkgs/blob/master/{path}#L{line}"
+
+
+def _get_codelines(package_name: str):
+    code = f"nix-shell -p {package_name}"
+    return [(0, code)]
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -2367,6 +2367,16 @@ engines:
     shortcut: pgo
     disabled: true
 
+# NixOS Package Search
+# - name: nixpkgs
+#   shortcut: nix
+#   engine: nixpkgs
+#   base_url: https://search.nixos.org/backend
+#   username:
+#   password:
+#   index: latest-42-nixos-unstable
+#   channel: unstable
+
 # Doku engine lets you access to any Doku wiki instance:
 # A public one or a privete/corporate one.
 #  - name: ubuntuwiki
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -174,3 +174,4 @@ features or generally made searx better:
 - @micsthepick
 - Daniel Kukula `<https://github.com/dkuku>`
 - Patrick Evans `https://github.com/holysoles`
+- Valentin Rieß `<https://github.com/v411e>`
--- a/docs/dev/engines/online/nixpkgs.rst
+++ b/docs/dev/engines/online/nixpkgs.rst
@@ -1,0 +1,8 @@
+.. _nixpkgs engine:
+
+==============
+Nixpkgs Search
+==============
+
+.. automodule:: searx.engines.nixpkgs
+   :members:


Bnyro:
I personally don't really like the idea of adding a dedicated template only for NixOS - can't you just use the normal existing results template?
The default result template actually supports a lot of stuff, see
https://github.com/searxng/searxng/blob/master/searx/templates/simple/result_templates/default.html
and
https://github.com/searxng/searxng/blob/11c0651ef47a72395d2d386d7600d951b0987ea0/searx/templates/simple/macros.html#L21-L53.

We already have support for other package managers (e.g. conda, archlinux, go) and they all make use of the default template too.

v411e:
Good point, will fix that

dalf:
> We already have support for other package managers (e.g. conda, archlinux, go) and they all make use of the default template too.

It is can be done in a second PR, actually a dedicate template for package would be useful.
For example, if I search for a package using `!pypi` the version is missing, so now I use `!!pypi`. YMMV.

return42:
> It is can be done in a second PR, actually a dedicate template for package would be useful.

Yeah, good point .. created feature request:

- https://github.com/searxng/searxng/issues/3210

Bnyro:
@v411e Whenever you have time, you can rebase your PR to the latest master branch and use the package manager template introduced in https://github.com/searxng/searxng/pull/3229 for this engine :)

v411e:
@Bnyro Uhh this is great. Thank you :)

v411e:
I finally got back to this PR and updated it to use the new template (which looks great btw).
The PR was tested locally and I would say it's ready.

image

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,174 @@
...
+password = ""
+index = "latest-42-nixos-unstable"
+channel = "unstable"
+search_url = base_url + "/" + index + "/_search"
+show_metadata = False
+categories = ["it", "packages"]
+
+
+def init(engine_settings):
+    if not ("username" in engine_settings and "password" in engine_settings):
```
Bnyro:
```suggestion
    if not (username and password):
```

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,174 @@
...
+        raise ValueError("username and password need to be configured.")
+
+
+def request(query, params):
+    search_url = base_url + "/" + index + "/_search"
```
Bnyro:
```suggestion
    search_url = f"{base_url}/{index}/_search"
```

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if username and password:
+        params["auth"] = (username, password)
```
Bnyro: That check is not necessary, you already made sure that they exist in `init()`

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,174 @@
...
+    params["data"] = _build_query(query)
+    params["headers"]["Content-Type"] = "application/json"
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    resp_json = loads(resp.text)
```
Bnyro:
```suggestion
    resp_json = resp.json()
```

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,174 @@
...
+        r["title"] = raw_result["package_attr_name"]
+        r["package_name"] = raw_result["package_pname"]
+        r["version"] = raw_result["package_pversion"]
+        r["content"] = raw_result["package_description"]
+        r["source_code_url"] = _position_to_github_url(raw_result["package_position"])
+        r["url"] = f"https://search.nixos.org/packages?channel={channel}&show={raw_result['package_attr_name']}"
+        r["maintainer"] = ", ".join(raw_result["package_maintainers_set"])
+        r["tags"] = raw_result["package_programs"]
+        r["license_name"] = ", ".join(raw_result["package_license_set"])
+        r["homepage"] = next(iter(raw_result["package_homepage"]), None)
```
Bnyro:
```suggestion
        r = {
            "template": "packages.html",
            "title": raw_result["package_attr_name"],
            ...
        }
```

searx/engines/nixpkgs.py
v411e: Whoops yes, thanks :)

v411e:
Thanks for the review @Bnyro. Should be fixed now :)

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,171 @@
...
+
+def init(engine_settings):
+    if not (username and password):
+        raise ValueError("username and password need to be configured.")
+
+
+def request(query, params):
+    params["auth"] = (username, password)
+    params["url"] = search_url
+    params["method"] = "GET"
```
Bnyro: This can be dropped as GET is the default anyways.

Bnyro (approved):

searx/engines/nixpkgs.py
v411e: Alright, thanks! Fixed it :)

v411e:
Alright checks passed

ddogfoodd:
LGTM

return42:
Before I can review this PR I have to clean up the commit history .. I will document my force push tasks ..

return42:
FYI last force push was a rebase on current master and squashed the 16 commits into one commit .. no code changes!

return42:
FYI: last force push adds docstrings from the engine module to the online documentation / [diff](https://github.com/searxng/searxng/compare/5ab5ea992bd6f5700cc4785f6bd7e35f3e33fa84..f584ae31fabdfa451bbe7f1ebec23f94c3471e7f)

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,164 @@
...
+    "require_api_key": True,
+    "results": 'JSON',
+}
+categories = ["it", "packages"]
+
+base_url = "https://search.nixos.org/backend"
+username = ""
+password = ""
+index = "latest-42-nixos-unstable"
+channel = "unstable"
```
return42, line 41: We should document what the options are .. the values and there meanings.

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,164 @@
...
+search_url = f"{base_url}/{index}/_search"
+show_metadata = False
```
return42, line 43: Not sure showing metadata in the result items should be optional .. and if it is optional, this switch should also be documented.

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return results
+
+
+def _build_query(query: str):
+    return f"""
+      {{
+        "from": 0,
+        "size": 50,
+        "sort": [
```
return42, line 101: Warning: this will produce a JSON string with a lot of whitespaces. Luckily the receiver is able to parse such a string .. but this might change at some day or point.

searx/engines/nixpkgs.py
```diff
@@ -0,0 +1,164 @@
...
+    "website": "https://search.nixos.org",
+    "use_official_api": False,
+    "require_api_key": True,
+    "results": 'JSON',
+}
+categories = ["it", "packages"]
+
+base_url = "https://search.nixos.org/backend"
+username = ""
+password = ""
```
return42, line 39: Do we really need username & password?  The search over https://search.nixos.org also works without a login and I would prefer SearXNG maintainers would not be forced to create accounts at NixOS just to get the search engine active.

By example, we are able to search for packages without login:

- https://search.nixos.org/packages?size=50&sort=relevance&type=packages&query=searxng

The WEB client will create a JSON request ...

```json
{"from":0,"size":50,"sort":[{"_score":"desc","package_attr_name":"desc","package_pversion":"desc"}],"aggs":{"package_attr_set":{"terms":{"field":"package_attr_set","size":20}},"package_license_set":{"terms":{"field":"package_license_set","size":20}},"package_maintainers_set":{"terms":{"field":"package_maintainers_set","size":20}},"package_platforms":{"terms":{"field":"package_platforms","size":20}},"all":{"global":{},"aggregations":{"package_attr_set":{"terms":{"field":"package_attr_set","size":20}},"package_license_set":{"terms":{"field":"package_license_set","size":20}},"package_maintainers_set":{"terms":{"field":"package_maintainers_set","size":20}},"package_platforms":{"terms":{"field":"package_platforms","size":20}}}}},"query":{"bool":{"filter":[{"term":{"type":{"value":"package","_name":"filter_packages"}}},{"bool":{"must":[{"bool":{"should":[]}},{"bool":{"should":[]}},{"bool":{"should":[]}},{"bool":{"should":[]}}]}}],"must":[{"dis_max":{"tie_breaker":0.7,"queries":[{"multi_match":{"type":"cross_fields","query":"searxng","analyzer":"whitespace","auto_generate_synonyms_phrase_query":false,"operator":"and","_name":"multi_match_searxng","fields":["package_attr_name^9","package_attr_name.*^5.3999999999999995","package_programs^9","package_programs.*^5.3999999999999995","package_pname^6","package_pname.*^3.5999999999999996","package_description^1.3","package_description.*^0.78","package_longDescription^1","package_longDescription.*^0.6","flake_name^0.5","flake_name.*^0.3"]}},{"wildcard":{"package_attr_name":{"value":"*searxng*","case_insensitive":true}}}]}}]}}}
```

and send it to https://search.nixos.org/backend/latest-42-nixos-24.05/_search from where the client gets a JSON response 

.. this is similar to what the engine from this PR does .. but without the need of a login.

return42 (changes requested):
I left some comments .. The most important thing for me is that we have an implementation that does not require a login.

Denperidge:
@return42 if I may be as forward; would it be okay for me to create a PR from this branch with the review comments fixed?

return42:
> @return42 if I may be as forward; would it be okay for me to create a PR from this branch with the review comments fixed?

To not break the conversation of this PR I suggest you create a branch from this PR https://github.com/v411e/searxng/tree/master .. add a commit with your fix on top and post the link to your new branch here .. then I can force push your commit on this PR here.

If this is too complicated for you, feel free to create a new PR (which in this case is almost the same process ;)

Denperidge:
That’s a better way to do it! Will do my best 😌

searx/engines/nixpkgs.py
Denperidge: I've been trying to get it to run without username/pass, but it seems to still return 401's. Wondering if it's a no-auth-for-specified-origin thing

Denperidge: There seems to be an auth token given with and visible from search.nixos.org. If keeping that token around doesn't automatically fail within the next day, we can discuss how ethically using that is. If it fails, then our minds rest easier

return42: > There seems to be an auth token given with and visible from search.nixos.org.

Seems not to be an cookie .. where in the communication did you find the token / can you tell us a bit more about?

Denperidge: I've been trying to dissect the nixos-search code despite my unfamiliarity with flakes & elm, but what I could find:
- So, the elasticsearch requires authorisation of base encoded username + password. This doesn't seem to be negotiable, as search.nixos.org itself is *also* logged in
- The Search page on search.nixos.org documents shows options.username & options.password being passed to the backend https://github.com/NixOS/nixos-search/blob/510d42d847600c0f197a233a564b428d7551f510/frontend/src/Search.elm#L1392 
- These options seem to be passed from environment variables (https://github.com/NixOS/nixos-search/blob/510d42d847600c0f197a233a564b428d7551f510/frontend/src/index.js#L12)
- When capturing the _search request through the network tab, check the request headers. This reveals the auth token 
A screenshot of the Firefox dev networking console. A POST request to _search, code 200. The authorization header is highlighted, with the value of it censored

So I just grabbed one of those values and stuck them into the request. Still works after a day, but idk whether it'll keep working and/or whether it's okay to do that

return42: Thanks a lot for your deep analysis .. I think I have a solution in mind .. I will give it a try next week ..

Denperidge: Thanks for hearing me out! Linking here to my current version of the branch in case it’s any help https://github.com/Denperidge-Contrib/searxng/tree/v411e-nixos

SuperSandro2000: I don't think NixOS is giving out tokens for that. IIRC the token is publically available in the source of the website repo and IMO I would just copy paste it here.

searx/settings.yml
```diff
@@ -2367,6 +2367,16 @@ engines:
...
     disabled: true
 
+# NixOS Package Search
+# - name: nixpkgs
+#   shortcut: nix
+#   engine: nixpkgs
+#   base_url: https://search.nixos.org/backend
+#   username:
+#   password:
+#   index: latest-42-nixos-unstable
```
SuperSandro2000, line 2377: I am not aware of any other deployment out in the wild, so we could just omit those 4 options.

searx/engines/nixpkgs.py
Denperidge: Am I overlooking the tokens location in the repo? I saw the elastic login has a default value, but due to its env value base I didn’t assume it’s that

redbeardymcgee:
@Denperidge Sorry for the ping, but I wasn't able to find your updated PR. Is this still a work in progress?

return42:
Closed due to inactivity.

PR #3207, opened 2024-02-11
[feat] engine: implementation of mozhi

Bnyro:
## What does this PR do?
- mozhi is a self-hostable wrapper around different translation APIs such as DeepL, Yandex, Google, DuckDuckGo, ...

- it's what I personally use instead of lingva now, due to the fact lingva is no longer maintained

- !mozhi en-de Hello World
- Change `mozhi_engine` to `google` in settings.yml for example -> redo step 1

## Author notes
- could be interesting for @return42 maybe? :)

### DIFF
--- a/searx/engines/mozhi.py
+++ b/searx/engines/mozhi.py
@@ -1,0 +1,59 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Mozhi (alternative frontend for popular translation engines)"""
+
+import random
+import re
+from urllib.parse import urlencode
+
+about = {
+    "website": 'https://codeberg.org/aryak/mozhi',
+    "wikidata_id": None,
+    "official_api_documentation": 'https://mozhi.aryak.me/api/swagger/index.html',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+engine_type = 'online_dictionary'
+categories = ['general']
+
+base_url = "https://mozhi.aryak.me"
+mozhi_engine = "google"
+
+re_transliteration_unsupported = "Direction '.*' is not supported"
+
+
+def request(_query, params):
+    request_url = random.choice(base_url) if isinstance(base_url, list) else base_url
+
+    args = {'from': params['from_lang'][1], 'to': params['to_lang'][1], 'text': params['query'], 'engine': mozhi_engine}
+    params['url'] = f"{request_url}/api/translate?{urlencode(args)}"
+    return params
+
+
+def response(resp):
+    translation = resp.json()
+
+    infobox = ""
+
+    if translation['target_transliteration'] and not re.match(
+        re_transliteration_unsupported, translation['target_transliteration']
+    ):
+        infobox = f"<b>{translation['target_transliteration']}</b>"
+
+    if translation['word_choices']:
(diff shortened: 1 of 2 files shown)


unixfox:
I think this should support multiple public instances like piped or invidious instance right?

unixfox:
Doesn't work out of the box. It just timeout: https://3207.testpr.searxng.net/search?q=%21mozhi%20en-de%20Hello%20World&language=auto&time_range=&safesearch=0&categories=none

Bnyro:
Thanks for the review @unixfox, I addressed both your suggestions.

unixfox:
Still in timeout. The server is located in USA.

Maybe the engine should act like invidious and piped engines, have the ability to use multiple instances? The one configured is hosted in India, hence the high response time.

Bnyro:
> Maybe the engine should act like invidious and piped engines, have the ability to use multiple instances? The one configured is hosted in India, hence the high response time.
>
Yes, I already added all public instances to the list so that a random one is picked each time.

For me in Germany there are no timeout issues with any of the 3 instances, strange.

searx/engines/mozhi.py
```diff
@@ -0,0 +1,59 @@
...
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+engine_type = 'online_dictionary'
+categories = ['general']
+
+base_url = "https://mozhi.aryak.me"
+mozhi_engine = "yandex"
```
return42: Any reason why you use `yandex` by default? .. for me `yandex` had very slow response and  `deepl`, `goggle` and `mymemory` have good translations and do fast response.

`reverso` did not worked for me / I tested with:

```
!mozhi en-de SearXNG is a free internet metasearch engine which aggregates results from more than 70 search services. Users are neither tracked nor profiled. Additionally, SearXNG can be used over Tor for online anonymity.
```

return42 (approved):

searx/engines/mozhi.py
Bnyro: The only reason is that Yandex suppports additional example sentences for a translation (which I needed for testing). I haven't tested the other supported engines really, there probably is a better alternative, we can change it to whatever translation engine you prefer :)

return42: I vote for google / afaik it has the most translations (languages).

return42:
> Still in timeout. The server is located in USA.

@unixfox can you give it another try / is there anything more that holds PR from merge?

---

UPDATE: I do not understand why https://3207.testpr.searxng.net has timeouts / has this instance been updated to b2f4c4e70714?

UPDATE: @Bnyro I had to increase the timeout to 7sec to not fall (randomly) into the timeouts.

Bnyro:
> UPDATE: @Bnyro I had to increase the timeout to 7sec to not fall (randomly) into the timeouts.
>
Strange, that might be an upstream mozhi issue :thinking:

return42:
> > @Bnyro I had to increase the timeout to 7sec to not fall (randomly) into the timeouts.

May it was an temporary issue / here are the results from today:

```yaml
    base_url:
      - https://mozhi.aryak.me  # 3-4 sec
      # - https://translate.bus-hit.me # 2-3 sec
      # - https://nyc1.mz.ggtyler.dev  # 2-3 sec
```

PR #3211, opened 2024-02-14
[fix] presearch videos: move to videos category

Bnyro:
## What does this PR do?
- the changes are self-explaining

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1375,7 +1375,7 @@ engines:
     engine: presearch
     network: presearch
     search_type: videos
-    categories: [general, web]
+    categories: [videos, web]
     timeout: 4.0
     shortcut: psvid
     disabled: true


return42:
> funny enough that nobody else noticed/complained before :)

it was intended :) --> https://docs.searxng.org/dev/engines/online/presearch.html#search-type-video

Not sure it is a good solution for the video category of presearch / I'm open for suggestions .. but I think in the SaerXNG category "video" user will expect result items from video-template and not of type default-template :shrug:

Bnyro:
Hmm, but when searching in the general category of a web search engine, I expect the results to be of all kinds and not only videos.

In my opinion it'd be better to move the presearch videos engine to videos due to the fact that it only outputs links to videos, but that's just my POV

return42:
> move the presearch videos engine to videos due to the fact that it only outputs links to videos

Does it? .. by example try [!videos !psvid bmw](https://darmarit.org/searx/search?q=%21videos+%21psvid+bmw) on top you will find (real) videos from piped youtube and more .. but when you scroll down you will see items from presearch-video that are not (real) videos:

grafik

If you follow the link https://www.bmw.co.uk/en/index.html you see not a typical page of a video ... its the homepage BMW in UK with an animation in the background .. what presearch classifies as a *video* ..

Not sure if a user wants to see a hompage with a video in the background when he search in the video category .. 

On the other hand presearch also lists a video from youtube .. but the next two results after the youtoube there is a homepage again.

In sense of "videos" I was looking for, presearch gives me to much results that not really videos .. so I thought its results are better placed in the the "general" category .. 

Not sure what is better / its just my feeling :shrug: 

grafik

return42:
Further I think in the video category the results should come from a video template / with the link to the video and a "Show Video" Button.

grafik

Since presearch videos does not give the link to the video we can't use the video template for the results.

Bnyro:
Okay, I think there's no perfect solution here, so we'll keep things as they are right now :+1:

PR #3213, opened 2024-02-14
Set unused engines as inactive

unixfox:
## What does this PR do?

This PR sets a bunch of engines as inactive.

Setting engines as inactive reduces the memory used by SearXNG and reduce the enormous amount of engines listed by default in the preferences page. Thus improving the user experience overall.

One can always reactivate the engine by setting `inactive: false` for the engine.

For the moment I have only set the engines as inactive which do not appear in the stats page of my instance: https://searx.be/stats.

There might be a follow up PR with the engines that are not being used at all by looking at the "requests count" thanks to @dalf PR: https://github.com/searxng/searxng/pull/3157

Run SearXNG as usual.

## Related issues

https://github.com/searxng/internal/issues/36

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -323,6 +323,7 @@ engines:
     timeout: 4.0
     shortcut: apkm
     disabled: true
+    inactive: true
 
   - name: apple app store
     engine: apple_app_store
@@ -349,6 +350,7 @@ engines:
     timeout: 6.0
     shortcut: conda
     disabled: true
+    inactive: true
 
   - name: arch linux wiki
     engine: archlinux
@@ -363,6 +365,8 @@ engines:
     engine: arxiv
     shortcut: arx
     timeout: 4.0
+    disabled: true
+    inactive: true
 
   # tmp suspended:  dh key too small
   # - name: base
@@ -386,6 +390,7 @@ engines:
     engine: bilibili
     shortcut: bil
     disabled: true
+    inactive: true
 
   - name: bing
     engine: bing
@@ -462,6 +467,7 @@ engines:
   - name: chefkoch
     engine: chefkoch
     shortcut: chef
+    inactive: true
     # to show premium or plus results too:
     # skip_premium: false
 
@@ -477,6 +483,7 @@ engines:
     shortcut: cr
     timeout: 30
     disabled: true
+    inactive: true
 
   - name: crowdview
     engine: json_engine
@@ -559,11 +566,13 @@ engines:
     engine: deezer
     shortcut: dz
     disabled: true
+    inactive: true
 
   - name: destatis
     engine: destatis
     shortcut: destat
     disabled: true
+    inactive: true
 
   - name: deviantart
     engine: deviantart
@@ -588,6 +597,8 @@ engines:
     engine: docker_hub
     shortcut: dh
     categories: [it, packages]
+    disabled: true
+    inactive: true
 
   - name: erowid
     engine: xpath
@@ -601,6 +612,7 @@ engines:
     categories: []
     shortcut: ew
     disabled: true
+    inactive: true
     about:
       website: https://www.erowid.org/
       wikidata_id: Q1430691
@@ -674,6 +686,7 @@ engines:
     timeout: 4.0
     shortcut: em
     disabled: true
+    inactive: true
 
   - name: tineye
     engine: tineye
@@ -757,6 +770,7 @@ engines:
     shortcut: fy
     timeout: 8.0
     disabled: true
+    inactive: true
 
   - name: genius
     engine: genius
@@ -884,6 +898,7 @@ engines:
     page_size: 19
     categories: music
     disabled: true
+    inactive: true
     about:
       website: https://gpodder.net
       wikidata_id: Q3093354
@@ -903,6 +918,7 @@ engines:
     categories: it
     timeout: 4.0
     disabled: true
+    inactive: true
     shortcut: habr
     about:
       website: https://habr.com/
@@ -928,6 +944,8 @@ engines:
     page_size: 20
     categories: [it, packages]
     shortcut: ho
+    disabled: true
+    inactive: true
     about:
       website: https://hoogle.haskell.org/
       wikidata_id: Q34010
@@ -952,6 +970,7 @@ engines:
     shortcut: in
     timeout: 6.0
     disabled: true
+    inactive: true
 
   - name: invidious
     engine: invidious
@@ -975,6 +994,7 @@ engines:
     shortcut: js
     timeout: 3.0
     disabled: true
+    inactive: true
 
   - name: kickass
     engine: kickass
@@ -1043,6 +1063,8 @@ engines:
   - name: lingva
     engine: lingva
     shortcut: lv
+    disabled: true
+    inactive: true
     # set lingva instance in url, by default it will use the official instance
     # url: https://lingva.thedaviddelta.com
 
@@ -1108,6 +1130,8 @@ engines:
     engine: json_engine
     categories: [it]
     paging: true
+    disabled: true
+    inactive: true
     search_url: https://developer.mozilla.org/api/v1/search?q={query}&page={pageno}
     results_query: documents
     url_query: mdn_url
@@ -1126,6 +1150,7 @@ engines:
     engine: metacpan
     shortcut: cpan
     disabled: true
+    inactive: true
     number_of_results: 20
 
   # - name: meilisearch
@@ -1266,6 +1291,7 @@ engines:
     categories: files
     timeout: 4.0
     disabled: true
+    inactive: true
     shortcut: or
     about:
       website: https://openrepos.net/
@@ -1285,6 +1311,7 @@ engines:
     content_query: description
     categories: [it, packages]
     disabled: true
+    inactive: true
     timeout: 5.0
     shortcut: pack
     about:
@@ -1343,6 +1370,7 @@ engines:
   - name: podcastindex
     engine: podcastindex
     shortcut: podcast
+    inactive: true
 
   # Required dependency: psychopg2
   #  - name: postgresql
@@ -1401,6 +1429,7 @@ engines:
     categories: [packages, it]
     timeout: 3.0
     disabled: true
+    inactive: true
     first_page_num: 1
     about:
       website: https://pub.dev/
@@ -1413,6 +1442,8 @@ engines:
     engine: pubmed
     shortcut: pub
     timeout: 3.0
+    disabled: true
+    inactive: true
 
   - name: pypi
     shortcut: pypi
@@ -1426,6 +1457,8 @@ engines:
     suggestion_xpath: /html/body/main/div/div/div/form/div/div[@class="callout-block"]/p/span/a[@class="link"]
     first_page_num: 1
     categories: [it, packages]
+    disabled: true
+    inactive: true
     about:
       website: https://pypi.org
       wikidata_id: Q2984686
@@ -1498,6 +1531,7 @@ engines:
     engine: rottentomatoes
     shortcut: rt
     disabled: true
+    inactive: true
 
   # Required dependency: redis
   # - name: myredis
@@ -1551,11 +1585,13 @@ engines:
     engine: searchcode_code
     shortcut: scc
     disabled: true
+    inactive: true
 
   - name: framalibre
     engine: framalibre
     shortcut: frl
     disabled: true
+    inactive: true
 
   # - name: searx
   #   engine: searx_engine
@@ -1569,6 +1605,7 @@ engines:
     engine: semantic_scholar
     disabled: true
     shortcut: se
+    inactive: true
 
   # Spotify needs API credentials
   # - name: spotify
@@ -1657,6 +1694,7 @@ engines:
     content_xpath: .//div[contains(@class,"overview")]
     shortcut: tm
     disabled: true
+    inactive: true
 
   # Requires Tor
   - name: torch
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -323,6 +323,7 @@ engines:
     timeout: 4.0
     shortcut: apkm
     disabled: true
+    inactive: true
```
return42: Should we remove the `disabled: true` when we set `inactive: true`?

unixfox: It's a good idea yes.

unixfox:
Updated the number of inactive engines with new stats where I found some engines that were set to inactive but were used this morning.

stats screenshot: https://github.com/searxng/searxng/assets/4016501/12db3172-2fa9-4d72-b6ca-83ab179fa415

searx/settings.yml
unixfox: I did comment the engine where "disabled: true" is found. This way we keep the history for engines that were disabled.

searx/settings.yml
```diff
@@ -1907,7 +1933,8 @@ engines:
   - name: duden
     engine: duden
     shortcut: du
-    disabled: true
+    # disabled: true
+    inactive: true
```
return42, line 1937: I would vote to leave `duden.de` disabled / but thats because I'm DE :)

searx/settings.yml
```diff
@@ -1880,11 +1903,14 @@ engines:
   - name: dictzone
     engine: dictzone
     shortcut: dc
+    inactive: true
```
return42, line 1906: I vote to leave `dictzone` enabled / its of type [online dictionary](https://docs.searxng.org/dev/engines/index.html#online-dictionary) .. and I often use `en-de hello` for translations.

searx/settings.yml
```diff
@@ -1656,7 +1678,8 @@ engines:
     title_xpath: .//div[contains(@class,"title")]//h2
     content_xpath: .//div[contains(@class,"overview")]
     shortcut: tm
-    disabled: true
+    # disabled: true
+    inactive: true
```
return42, line 1682: Personally I use tmdb rarely .. but I assume there are more user out there using the movie database `!tmdb hitchhiker's guide to the galaxy`.

searx/settings.yml
```diff
@@ -1567,8 +1588,9 @@ engines:
 
   - name: semantic scholar
     engine: semantic_scholar
-    disabled: true
+    # disabled: true
     shortcut: se
+    inactive: true
```
return42, line 1593: Very useful for scientific searches `!semantic_scholar ybco`.

searx/settings.yml
```diff
@@ -1497,7 +1515,8 @@ engines:
   - name: rottentomatoes
     engine: rottentomatoes
     shortcut: rt
-    disabled: true
+    # disabled: true
+    inactive: true
```
return42, line 1519: I know there are some guys prefer rottentomatos over tmbd .. further they are all in the movies group `!movies hitchhiker's guide to the galaxy`

image

searx/settings.yml
```diff
@@ -558,12 +563,14 @@ engines:
   - name: deezer
     engine: deezer
     shortcut: dz
-    disabled: true
+    # disabled: true
+    inactive: true
```
return42, line 567: Deezer has nice music `!deezer maceo parker in time` .. but needs registration to listen.

searx/settings.yml
```diff
@@ -476,7 +480,8 @@ engines:
     engine: crossref
     shortcut: cr
     timeout: 30
-    disabled: true
+    # disabled: true
+    inactive: true
```
return42, line 484: Personally I do not often use it, but has good DOI links `!crossref submarine telegraph cables`.

searx/settings.yml
```diff
@@ -462,6 +465,7 @@ engines:
   - name: chefkoch
     engine: chefkoch
     shortcut: chef
+    inactive: true
```
return42, line 468: :sos: my wive use it .. if we don't want to be responsible for the starvation of nerds like me, we shouldn't inactivate :sos:

return42 (commented):
I left some remarks about the engines set to `inactive: true` on this PR ..

Since the selection is always characterized by a very personal view, it is difficult to specify the best selection for the widest possible range of users.

Perhaps we should also take other factors into account when making our selection ... e.g. if an engine is only suitable for a very narrow language area, then it should perhaps be deactivated in the default setting ... English, Spanish and possibly Chinese are certainly the most frequently used languages ... but German, for example, is only spoken by a few people on this planet.


Maintainers like me who set up an instance for a user base in DE could then activate these languages.

## But maybe it is also better not to inactivate any engine?

At least not for arguments like "reduces the memory" ... if maintainers think their instance would use too much memory, then they should optimize their instance according to their needs.

searx/settings.yml
Bnyro: +1

searx/settings.yml
Bnyro: +1

searx/settings.yml
Bnyro: +1, I personally use this one too

searx/settings.yml
Bnyro: +1

unixfox:
@return42 @Bnyro for every comment in the review that you sent, you are capable in reactivating the engine yourself.

Unless this affect more people I don't see the argument against deactivating them.

Here I'm just taking a non-biased opinion and deactivating all the engine that are not being used on my instance nor on a lot of other public instances.

This means that this won't affect public instance users. For the people that are running private engines they can reactivate the engine because they are in control of the settings.yml.

------

> Perhaps we should also take other factors into account when making our selection ... e.g. if an engine is only suitable for a very narrow language area, then it should perhaps be deactivated in the default setting ... English, Spanish and possibly Chinese are certainly the most frequently used languages ... but German, for example, is only spoken by a few people on this planet.

> Maintainers like me who set up an instance for a user base in DE could then activate these languages.

I'm sorry but this will complicate the workflow of configuring engines... I know dalf talked about the ability to install engines from external sources.

Like on dnscrypt-proxy you can enable servers from an external list.

Maybe this could be a good tradeoff for your issue? I think this was discussed in a previous meeting about externalizing the engines to a different repository.

We can't possibly add every single website/service in the world. We have to put a limit somewhere.  
Or we have multiple repositories, with for example engines for specific regions/languages in the world.

return42:
> Unless this affect more people I don't see the argument against deactivating them.

What I have in mind is more related to the groups we have .. by example: with the bang [`!packages foo`](https://darmarit.org/searx/search?q=%21packages+foo) the user can activate all engines in this group .. compare with screenshot below (most of them are disabled by default) .. this is not a task a user will do often but he will love it when the day comes he wants to look in as much as possible package mangers .. we have similar groups for `!weather`, `!movies` and more .. when we now start to reduce the amount of active (but may be disabled) engines the concept of the groups suffers.

TL:DR;

It feels to me like I'm depriving the user of the engines .. this decision can be made by a maintainer, but I don't think we should decide that in our default settings.

> We have to put a limit somewhere.
> externalizing the engines to a different repository.

Plural reposetories .. and one repo will be the default we maintain (*base*).  We can sort out engines into repositories/groups like *base* and *community* .. but as long we do not have the infrastructure for external engines we have them here and I think we should not decide by default to what engines are available for the end user / for the reasons I mentioned above.

image

unixfox:
> What I have in mind is more related to the groups we have .. by example: with the bang [`!packages foo`](https://darmarit.org/searx/search?q=%21packages+foo) the user can activate all engines in this group .. compare with screenshot below (most of them are disabled by default) .. this is not a task a user will do often but he will love it when the day comes he wants to look in as much as possible package mangers .. we have similar groups for `!weather`, `!movies` and more .. when we now start to reduce the amount of active (but may be disabled) engines the concept of the groups suffers.

But you can see based on the stats that this hasn't really catch on because people already find these kind of information through conventional search engines. It's such a niche for these specific categories like packages or movie that I do not see the point in having them activated by default.

The disavantages outweight the avantages. It's worsening the user experience by having too many engines and the resources needed to have all of these engines just for a very small portion of our user base.

I have always said that we should focus on the average user base and not trying to necesarily fullfill the needs of all the niche use cases.

Anyway, these specific categories will still work. One just have to reactivate the engines, if it's something that is a must for you @return42 or @Bnyro, reactivate the engines on your instances.

But try to think outside of your sole needs and instead think about the average user that will use our software.

Also remember we **ARE NOT** trying to make SearXNG yet another search engine, it's just a meta search engine. The goal is not to replace the conventional search engines. SearXNG is too dumb to order or categorize the results in a way that is useful for the "normies".

glanham-jr:
If memory is an issue, could there be an LRU cache for engines? Load again when needed, dump when not needed. For instances running on minimal RAM, this could would help, just wondering if there is a more dynamic solution.

Also, is there a memory footprint difference before/after the disabling? Might give an idea of how much RAM we are saving.

return42:
> Load again when needed

Lazy loading of the engines is not possible since the engine needs to be initialized before the users can make use of the engine.

> If memory is an issue

If memory is the issue, we should work on https://github.com/searxng/searxng/discussions/2633

> Also, is there a memory footprint difference before/after the disabling? Might give an idea of how much RAM we are saving.

Yes would be cool if someone has the time to measure :+1:

glanham-jr:
> Lazy loading of the engines is not possible since the engine needs to be initialized before the users can make use of the engine.

Is that an issue? Seems exactly like a case for lazy loading. When a user makes a request to that engine, it can use a mutex lock to block all requests that want to use that engine until it's initialized. If it's not part of a default search and say only accessible by bang, then the initial performance hit may not be an issue. Could even customize the cache time per lazy engine.

The idea would be to add the option of lazy loading. Eagerly load the most used engines, and set lazy load + cache as the default for the rest.

---

But without more data, it's hard to know even how much memory is saved. Might be good to come back with metrics before even attempting to above. I might try to do this first, then if we see big savings I'd consider experimenting with lazy caches.

Bnyro:
> The disavantages outweight the avantages. It's worsening the user experience by having too many engines and the resources needed to have all of these engines just for a very small portion of our user base.
>
I don't agree with this. Users don't have any disadvantages from having more engines to choose from. Why should it? Users can either use the default settings and be happy or customize the engines to their needs: more available engines -> higher chances that the user finds all engines they want to use.

The performance thing might be a point, but
1. Compared to other projects I'm self-hosting, SearXNG is definitely one of the more lightweight projects. What's the scale for a good/bad performance?
2. Is this really relevant for the average person that's self-hosting an instance? Most engines are probably private and used by less than 100 people, so they will never come into a situation where the additional 100MB of RAM matter.

The engines that you want to disable in this PR are engines that are not being used **on your instance**. However, each instance has a different community, and I don't see why your community should set the default settings for the whole project.

I can definitely agree with disabling all non-english engines and all these that require an API key by default - so that each instance maintainer can decide whether they have some users from that country/language, but anything else is just pointless as per my point of view. We could endlessly discuss which engines to set inactive and which not to if we don't have any fixed criteria based on which these engines are set to inactive.

Also, this will break many existing setups / make some currently working queries not working -> we need good documentation and announcements about it, so that instance maintainers know they have to set `inactive: false` before being able to allow the engine again. I'm certain that more of 90% of the instance admins are not aware of that.

unixfox:
> The engines that you want to disable in this PR are engines that are not being used **on your instance**. However, each instance has a different community, and I don't see why your community should set the default settings for the whole project.

Like said above, it's not only on my instance but on other popular public instances that I looked at (priv.au for example). But if you want, I can try to prove that these engines that not being used on 80% of the public instances on https://searx.space.

Also, like said above, users of private instances have all the capabilities to reactivate the engines, so this is not an issue here. If you or @return42 are missing some engines due to the deactivation of this PR, then **you have all the capabilities to reactivate them because you have access to the settings.yml**.

-------

For me, this PR aims to solve 3 things:
- reducing the memory usage (even if minor)
- not overwhelm the users by having too many options to choose from
- put a limit somewhere of how many engines we should provide by default. We can't possibly add every single service on the internet, we have to make a choice.
  a good factor is to look at the popularity of the service on https://www.similarweb.com and check if based on the ranking enough people will use it so that it is worth having an engine in the core.

> Also, this will break many existing setups / make some currently working queries not working -> we need good documentation and announcements about it, so that instance maintainers know they have to set inactive: false before being able to allow the engine again. I'm certain that more of 90% of the instance admins are not aware of that.

Very good thing to bring up, I can take the lead on that and announce this change. Like I have done for [`public_instance: true`](https://github.com/searxng/searx-instances/discussions/417), [removal of searx instances](https://github.com/searxng/searx-instances/discussions/336).

Bnyro:
If I counted it correctly, there are about 10 engines that only support a language other than English and about the same amount of engines that require an api key. If we set these to `inactive`, we
- reduce memory usage a bit
- declutter the interface
- put a limitation to only enable english and multilingual engines by default

What speaks against this suggestion, a solution tgat @return42 and me can agree with as the inactive engines are chosen based on strict criteria and it would not break our existing search categories?

Bnyro:
You're still free to disable any additional engines on your instance if you think they're useless, which should certainly be up to "advanced"/experienced instance admins to decide.

return42:
> Setting engines as inactive reduces the memory used by SearXNG

Do we know the approximate quantity / if someone has interest and time for, a rough comparison with https://github.com/searxng/searxng/pull/3443 would be interesting

unixfox:
Closing as the ones mainly maintaining searxng would prefer to have many engines enabled by default.

PR #3221, opened 2024-02-16
Save memory by not calling babel.Locale.parse at runtime

dalf:
## What does this PR do?

babel.Locale.parse loads more than 60MB in RAM.  The only purpose is to get:

    LOCALE_NAMES   - searx.data.LOCALES["LOCALE_NAMES"]
    RTL_LOCALES    - searx.data.LOCALES["RTL_LOCALES"]

This commit calls babel.Locale.parse when the translations are update from weblate and stored in:

    searx/data/locales.json

This file can be build by:

    ./manage data.locales

By store these variables in searx.data when the translations are updated we save round about 65MB (usually 4 worker = 260MB of RAM saved.

## Author's checklist

~This is a draft,~@return42  feel free to update / close / whatever (I'm not sure to have the "resource / time" to update this PR)

## Related issues

- https://github.com/searxng/searxng/discussions/2633#discussioncomment-8490494


searxng_extra/update/update_translation_labels.py
```diff
@@ -0,0 +1,133 @@
...
+            locale = babel.Locale.parse(dirname)
+            LOCALE_NAMES[tag] = get_locale_descr(locale, dirname)
+            if locale.text_direction == 'rtl':
+                RTL_LOCALES.add(tag)
+
+    with open(translations_filename, 'w', encoding='utf-8') as f:
+        content = {
+            "LOCALE_NAMES": LOCALE_NAMES,
+            "RTL_LOCALES": list(RTL_LOCALES),
+            "LOCALE_BEST_MATCH": LOCALE_BEST_MATCH,
```
dalf: wikipedia uses this variable, so it is saved --> ADDITIONAL_TRANSLATIONS is also added, so the change on `searx.locale` is fully compatible with the current master branch.

return42:
> This is a draft, @return42  feel free to update / close / whatever (I'm not sure to have the "resource / time" to update this PR)

Thank you very much! I will continue the PR :+1:

searxng_extra/update/update_translation_labels.py
return42: In the new commit `ADDITIONAL_TRANSLATIONS` and `LOCALE_BEST_MATCH` are back in the `searx.locales`.

return42:
Turned POC into a production-ready implementation and force-pushed it onto the branch here --> PR is ready to review :+1: 

The PR will save round about 65MB (per worker) compared to current master :rocket:

searx/locales.py
```diff
@@ -90,74 +118,37 @@ def get_translations():
...
-    english_language, english_territory = _get_locale_descr(locale, 'en')
-
-    if native_territory == english_territory:
-        english_territory = None
 
-    if not native_territory and not english_territory:
-        if native_language == english_language:
-            return native_language
-        return native_language + ' (' + english_language + ')'
+def get_translation_locales() -> list[str]:
```
dalf, line 124: Why this function is not in `update_locales.py`, this is the only place where it is used.
Looking at `_TR_LOCALES`, I guess the intend is to use this function elsewhere at runtime? 
But `LOCALE_NAMES` already provides the same information, with the labels.
Ot did I miss something?

searxng_extra/update/update_locales.py
```diff
@@ -0,0 +1,103 @@
...
+import babel.core
+
+from searx import searx_dir
+from searx.locales import (
+    ADDITIONAL_TRANSLATIONS,
+    LOCALE_BEST_MATCH,
+    get_translation_locales,
+)
+
+LOCALE_DATA_FILE = Path(searx_dir) / 'data' / 'locales.json'
```
dalf, line 27: The `locales.json` name is confusing: 
* the same data are displayed to the user as `Interface language`.
* there is already a generated file named `sxng_locales.py`.

searx/locales.py
return42: > this is the only place where it is used.

My intention was to have most of the *business logic* in SearXNG / just hide the problematic `babel.Locale.parse` in update files.

> `LOCALE_NAMES` already provides the same information

The number of items in the list of `LOCALE_NAMES` is different from what `get_translation_locales()` returns / there are addition items from `ADDITIONAL_TRANSLATIONS` and `LOCALE_BEST_MATCH`.

If you insist, I can move this function to the update script / but for me it feels better to have it here in the `locale` module.

searxng_extra/update/update_locales.py
return42: > The `locales.json` name is confusing:
> * the same data are displayed to the user as `Interface language`.

What do you had in mind / I have only this part of code in mind

https://github.com/searxng/searxng/blob/c197c0e35e6eaa823d6ba8606df4e5a5c598a07b/searx/webapp.py#L1004

and I would prefer to change the name `locales` in the templates to `locale_names` or better `LOCALE_NAMES`. / but templates are not in the scope of this PR.

searx/locales.py
dalf: I don't understand the usage for `get_translation_locales()`, may I ask you to provide one?

[EDIT] to focus only on `get_translation_locales()`

searxng_extra/update/update_locales.py
dalf: > I would prefer to change the name locales in the templates to locale_names

Including in the UI? That's not locales, that's language for the UI, that's why SearXNG needs the labels : to display a list of languages for the UI.

dalf:
@return42, I wrote some comments, but feel free to merge.

utils/lib_sxng_data.sh
```diff
@@ -49,6 +51,15 @@ data.useragents() {
     dump_return $?
 }
 
+data.locales() {
+    (   set -e
+        pyenv.activate
+        build_msg DATA "update searx/data/locales.json"
+        python searxng_extra/update/update_locales.py
+    )
```
Bnyro, line 59: Just for my understanding, what do these brackets do? Do they just create a new "scope", so that anything running inside these brackets (inner function) will not affect the outer function (data.locales in this case)?

I've already seen this multiple times, so I'm wondering.

Bnyro (approved):
Can confirm this works as intended, thank you for spotting our worst memory-nightmares @dalf :tada:

utils/lib_sxng_data.sh
return42: > what do these brackets do? 

it wraps commands in a subshell / here in this example the option `set -e` is only active in the subshell commands:

[`set -e``](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) .. exit (sub-)shell immediately if a command returns a non-zero status.

With the construct:

```bash
(  set -x
   command1
   command2 # ends a non-zero-status
   command3
)
```

You assert that command3 is not called when command2 ends with a non-zero-status

Bnyro: Good to know, thank you!

searxng_extra/update/update_locales.py
return42: >  That's not locales, that's language for the UI

Sorry, my fail .. you are right; "languages" or similar would be a better name / back to your initial comment:

> The `locales.json` name is confusing: 

IMO its a name that fits too its usage .. it contains values for names that are defined in the `searx.locales` module.

https://github.com/searxng/searxng/blob/ed66ed758dbd1c926296aa227da8a82fff9166e1/searx/locales.py#L54-L63

I had thought about naming the file `locales_data.json` ... but found that a bit superfluous, because the file is already in a folder `data/`

searx/locales.py
return42: > I don't understand the usage for `get_translation_locales()`, may I ask you to provide one?

Its documented in https://docs.searxng.org/src/searx.locales.html

image

From my POV the "how this list is generated?" is a part of our application model and should not been hidden somewhere in a update file / update tools should IMO just call some library methods but should not implement the application model partial.

I understand your point of view, but for me it just feels better if this list is generated by a function that is in the namespace `searx.locales` for the reasons mentioned.    

A newcomer will usually not look in an update script to see how a list of translations of the application is built.

> this is the only place where it is used.

It is also irrelevant to me whether this function is only used once / in the update script.

There are many functions that are only used once, but are therefore not in the namespace in which they are used. Now we have all the `locale` logic bundled in the `searx.locale` module .. where it is also [documented at a single point](https://docs.searxng.org/src/searx.locales.html).

return42:
@dalf again great thanks for finding this waste of memory and the great POC what it made very easy for me  ..  :+1: :rocket:

return42:
> By store these variables in searx.data when the translations are updated 

Tested translation update and it works --> https://github.com/searxng/searxng/commit/bf0d09c9f184cb3728d7dabae6be9c6546faf03b#diff-05e2b08d3a450427044854039874907981dfb640d3e6afc89f2b2311a57dcdfa

PR #3231, opened 2024-02-20
[feat] images: show resolution in overview

### LINKED ISSUE #3177: Image search: show resolution beneath title or url
When searching for images it is quite annoying to click at all the images just to notice their resolution doesn't fit my need. Probably not all engines provide a resolution but for those who do, it would be great to show it beneath the image title or url.

I would like to see the resolution of images on the image search overview where possible. An option to enable/disable would be even better.

Use another search engine for image search or click on every image and hope for the best. A filter function to only show certain resolutions would be another possible solution.

Bnyro:
## What does this PR do?
- the previous `img_format` will now be used for the real format of the image (e.g. jpg, png, webp, ...)
- there now is a new field called `resolution` that should be used with the format `width x height`
- I've also changed the documentation accordingly

see #3177

- search for images, and notice the resolution labels next to the source (only works for the here modified engines)

## Related issues
closes #3177


searx/engines/wallhaven.py
```diff
@@ -76,7 +76,7 @@ def response(resp):
                 'url': result['url'],
                 'img_src': result['path'],
                 'thumbnail_src': result['thumbs']['small'],
-                'img_format': result['resolution'],
+                'resolution': result['resolution'].replace('x', ' x '),
```
Bnyro, line 82: Wallhaven returns something like `1920x1080`as image resolution, while for other engines, we have a gap in between the numbers and the `x`, like `1920 x 1080`. So this change is just for consistency, although I don't think anyone would actually notice the difference really :)

searx/engines/artic.py
```diff
@@ -61,7 +61,7 @@ def response(resp):
                 'content': result['medium_display'],
                 'author': ', '.join(result['artist_titles']),
                 'img_src': image_api + '/%(image_id)s/full/843,/0/default.jpg' % result,
-                'img_format': result['dimensions'],
+                'resolution': result['dimensions'],
```
return42: > there now is a new field called resolution that should be used with the format `width` x `height`

I assume we expect resolution in pixel sizes .. Artic shows the physical sizes of a picture / I moved this size to the `content` field

image

return42 (commented):
I pushed a WIP commit with some suggestions / I also added a field `filesize` to the image template / not yet in the documentation .. I first wanted to hear if this field makes sense / You can test it with `!wh foo` .. due to the additional fields, the lower image area appears squeezed

image

searx/engines/artic.py
Bnyro: Yes, I also wondered what to do about the Artic engine, the content field is probably the best place :+1:

Bnyro:
> I pushed a WIP commit with some suggestions / I also added a field `filesize` to the image template / not yet in the documentation .. I first wanted to hear if this field makes sense / You can test it with `!wh foo` .. due to the additional fields, the lower image area appears squeezed
> 
> image
>
Yes, that seems like a good idea :+1:

However, as you said, we will probably need to slighly reduce the line height and the lower/upper padding of the image details section to make it fit in well again.
Also, do you know why we keep the space free if the field is missing? It would probably look better if we didn't, I'll look into the UI changes later if you don't mind :)

return42:
> Also, do you know why we keep the space free if the field is missing? 

Perhaps this was done so that the eye can always remain in the same position when scrolling through the images in the table shown below? If the table layout changes depending on the existing metadata, could the view of the table become a bit choppy when you scroll between the images?

But then the identifier should have been displayed consistently / even if the value remains empty?

Personally I vote to stay with stable table layout in the bottom .. may we can show the identifier in the left column, even a value is not given.

Bnyro:
> Perhaps this was done so that the eye can always remain in the same position when scrolling through the images in the table shown below? If the table layout changes depending on the existing metadata, could the view of the table become a bit choppy when you scroll between the images?
> 
Ah yes, that makes sense.
> But then the identifier should have been displayed consistently / even if the value remains empty?
> 
> Personally I vote to stay with stable table layout in the bottom .. may we can show the identifier in the left column, even a value is not given.
>
I personally don't think we should show empty values because that would be some unnecessary text for the user.

I've now been
1. decreasing the gap between title and content
2. slightly reducing the line height of the result labels (resolution, format, ...)
3. increased the total height of the image results container a bit

For me this seems to work well on all different kind of resolutions now (on Firefox), but it'd be good if you could test it too (maybe also on a different browser).

return42:
I squashed minor edits / mainly documentation [diff](https://github.com/searxng/searxng/compare/6dd7bab17c0275cb877060caeab7dbf0a8a306da..716f6d04f093f1a10f01bffc84264c7fb62b9636).

Some thoughts I had in my review ..

Should we place "Visit page" as first row? the huge empty space looks sometimes a bit strange .. at least to me:

image

Not related to this PR / when a field value is to long for the rigth col .. a linebreak starts in the left column:

image

If the thumbnail is small or the URL to long, the resolution is no longer shown ..

On DDG the resolution its shown in the image preview .. 

image

Bnyro:
> I squashed minor edits / mainly documentation [diff](https://github.com/searxng/searxng/compare/6dd7bab17c0275cb877060caeab7dbf0a8a306da..716f6d04f093f1a10f01bffc84264c7fb62b9636).
> 
> Some thoughts I had in my review ..
> 
> Should we place "Visit page" as first row? the huge empty space looks sometimes a bit strange .. at least to me:
> 
> image
> 
No, I don't think that's a good idea. It would look kind of odd to start with a very long link in the first line and then the other lines are being much shorter.
Also, the full url is the least important information for me personally - I just want to have the image, anything else doesn't matter :)

> Not related to this PR / when a field value is to long for the rigth col .. a linebreak starts in the left column:
> 
> image
> 
Yes, we could also fix this in this PR by simply limiting the amount of lines to 1, if we're already working on it, why not :)

> If the thumbnail is small or the URL to long, the resolution is no longer shown ..
> 
> On DDG the resolution its shown in the image preview ..
> 
> image
>
I think that's a good idea, I'll see if I can make it work and send a WIP commit :+1:

searx/static/themes/simple/src/less/style.less
```diff
@@ -460,9 +477,7 @@ article[data-vim-selected].category-social {
   span.title,
   span.source {
     display: block;
-    position: absolute;
 
-    width: 100%;
     font-size: 0.9rem;
```
return42: Wow, looks quite pretty :rocket:   .. but there is an issue in the layout:

When we remove these properties from the `span.title` selector the result thumbnail is stretched by the title text ..

image

before this patch it looks like this:

image

Bnyro: Without these changes, the texts didn't show at all, I'm not really sure what's causing this here :shrug:

searx/templates/simple/result_templates/images.html
```diff
@@ -1,10 +1,8 @@
 <article class="result result-images {% if result['category'] %}category-{{ result['category'] }}{% endif %}">{{- "" -}}
         <a {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} href="{{ result.img_src }}">{{- "" -}}
-                <div class="image_thumbnail_container">
-                        <img class="image_thumbnail" {% if results_on_new_tab %}target="_blank" rel="noopener noreferrer"{% else %}rel="noreferrer"{% endif %} src="{% if result.thumbnail_src %}{{ image_proxify(result.thumbnail_src) }}{% else %}{{ image_proxify(result.img_src) }}{% endif %}" alt="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{- "" -}}
-                        {%- if result.resolution %} <span class="image_resolution">{{ result.resolution }}</span> {%- endif -%}
-                </div>
```
return42: > I'm not really sure what's causing this here 

I think this div (container) is the problem .. this div is not needed: when we have a span tag with the resolution directly after the img tag:

```html
<img class="image_thumbnail..> ..</img>
<span class="image_resolution">100x200</span>
```

the absolute position of the span tag is relative to this img tag:

```css
.image_resolution {
    position: absolute;
    right: 0;
}
```

I'm not a CSS expert, but it seems to work as expected .. can you please double check my WIP commit / thanks :+1:

searx/static/themes/simple/src/less/detail.less
```diff
@@ -68,9 +68,6 @@ article.result-images .detail {
       color: var(--color-result-detail-label-font);
       font-size: 0.9rem;
       line-height: 0.9rem;
-      white-space: nowrap;
-      text-overflow: ellipsis;
-      overflow: hidden;
```
Bnyro: This code was meant to fix the overflow for image labels (like resolution, author, ...). Did you intentionally remove this?

searx/templates/simple/result_templates/images.html
Bnyro: Yes, thanks, this does work as intended :+1:

searx/static/themes/simple/src/less/detail.less
return42: Opps sorry, I reverted to much / this what not intended .. its fixed now.

BTW: I had to rebase on master to solve a conflict in the files crated by `make static.build.commit`.

Bnyro: :+1:

searx/static/themes/simple/src/less/detail.less
```diff
@@ -67,7 +67,9 @@ article.result-images .detail {
     p {
       color: var(--color-result-detail-label-font);
       font-size: 0.9rem;
-      line-height: 0.9rem;
+      white-space: nowrap;
+      text-overflow: ellipsis;
+      overflow: hidden;
```
return42: While I was looking into this I fond an old issue .. here I removed the `line-height: 0.9rem;` since setting this value causes the lower part of letters such as g to be cut off:

Before ..

image


After removing line-height ..

image

Bnyro: Hmm, this change was meant to reduce the spacing between the individual entries/rows.
So now, the last entry is sometimes cut off a bit again.

I think we can achieve the same by using a smaller margin, I'll have a look.

Bnyro: Okay, `margin: 0.8rem 0;` fixes this (no cut off entries anymore and no text cut off).

Bnyro:
Ready to merge from my side, thanks for your investigations @return42 :smiley:

return42 (approved):
> Ready to merge from my side, thanks for your investigations @return42 😃

Thanks goes to you .. this PR is a huge improvement for our users searching for images :rocket:

PR #3237, opened 2024-02-22
[fix] opensearch.xml: firefox v123 does not except self-URI (URL with param)

### LINKED ISSUE #3227: SearXNG cannot add  as search engine in Firefox
Firefox gets an error when adding SearXNG as a search engine through the address bar.

图片

SearXNG instance: baresearch.org
SearXNG version：2024.2.18+76845ea42
Firefox version: 123.0b9

This problem also occurs on multiple instances such as priv.au, searx.aleteoryx.me, etc.

return42:
Before this patch we had a address in the rel-self-template with params [1]:

GET:

```xml
    <Url method="GET"
         rel="self"
         template="/opensearch.xml?method=GET&amp;autocomplete=google"
         type="application/opensearchdescription+xml"/>
```

POST:

```xml
    <Url method="POST"
         rel="self"
         template="/opensearch.xml?method=POST&amp;autocomplete=google"
         type="application/opensearchdescription+xml"/>
```

The first one (GET) is no longer accepted by firefox since v123, the latter one was always been wrong (since it mixes GET params with a POST request).

Since Firefox v123 a template syntax [2] in the rel-self-template is no longer supported / Firefox will report an error::

    Firefox could not download the search plugin from:
    http://127.0.0.1:8888/opensearch.xml?method=GET&autocomplete=google

This patch reduce the URL to::

    template="http://127.0.0.1:8888/opensearch.xml"

[1] https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#url-rel-values
[2] https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#opensearch-url-template-syntax

Closes: https://github.com/searxng/searxng/issues/3227

### DIFF
--- a/searx/templates/simple/opensearch.xml
+++ b/searx/templates/simple/opensearch.xml
@@ -15,7 +15,7 @@
   {% if autocomplete %}
     <Url rel="suggestions" type="application/x-suggestions+json" method="{{ opensearch_method }}" template="{{ url_for('autocompleter', _external=True) }}?q={searchTerms}"/>
   {% endif %}
-  <Url rel="self" type="application/opensearchdescription+xml"  method="{{ opensearch_method }}" template="{{ opensearch_url }}" />
+  <Url rel="self" type="application/opensearchdescription+xml"  method="{{ opensearch_method }}" template="{{ url_for('opensearch', _external=True) }}" />
   <Query role="example" searchTerms="SearXNG" />
   <moz:SearchForm>{{ url_for('search', _external=True) }}</moz:SearchForm>
 </OpenSearchDescription>
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -438,16 +438,6 @@ def render(template_name: str, **kwargs):
     kwargs['cache_url'] = settings['ui']['cache_url']
     kwargs['get_result_template'] = get_result_template
     kwargs['doi_resolver'] = get_doi_resolver(request.preferences)
-    kwargs['opensearch_url'] = (
-        url_for('opensearch')
-        + '?'
-        + urlencode(
-            {
-                'method': request.preferences.get_value('method'),
-                'autocomplete': request.preferences.get_value('autocomplete'),
-            }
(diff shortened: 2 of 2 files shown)


Bnyro (approved):

dalf:
For ref: https://github.com/searx/searx/pull/2247 --> I suggest to provide the parameters in the path like this `/os/post/google/opensearch.xml` (or similar).

return42:
> For ref: [searx/searx#2247](https://github.com/searx/searx/pull/2247) --> I suggest to provide the parameters in the path like this `/os/post/google/opensearch.xml` (or similar).

The problem already exists today with the old implementation ... when a user adds the search engine, it is added with the settings he has at that time.  If he changes his settings, this is only saved in the cookies.  However, the cookies are not sent via the search function of the web browser ... e.g. if you enter the search terms in the URL bar.

Assuming the user has set *google for auto-completion & POST* at the time he registered SearXNG in the browser ... then all search queries via the URL bar are searched with *google for auto-completion & POST* .. if he changes the settings in the cookies, e.g. to *ddg for auto-completion & GET*  ... then the search queries from the HTML form are carried out with *ddg for auto-completion & GET* and the search queries from the URL bar are still carried out with *google for auto-completion & POST*.

The only solution I know is to remove the SearXNG instance from the browser, change the settings and then re-add the instance with the new settings .. 

I suspect that hardly any of the users are aware of this problem ..

Furthermore, I have read that the web browsers usually do not perform automatic updates via the self-template ... Chrome [does not seem to have it](https://www.chromium.org/tab-to-search/) and Mozilla does not support it at least with the [AOM](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#supporting_automatic_updates_for_opensearch_plugins)

I'm open for suggestions .. but I think the switch from URL arguments to a URL path does not solve the fundamental problems of "cookies" versus "added `opensearch.xml`" in the browser.

PR #3244, opened 2024-02-25
[feat] engine: implementation of void linux packages

Bnyro:
## What does this PR do?
- add a new engine for Void Linux package search

- because Void Linux is

* `!void zsh`

## Author's checklist
- At Void Linux, sometimes multiple packages share the same template / source code / result url, for example `zsh-dbg` is contained in the `zsh` package template: Hence I've added a new field to skip deduplicating results that have the same url - because otherwise it would just remove packages like `zsh-dbg` from the search results (maybe there's a better way to achieve this, but this seemed like the easiest one)

### DIFF
--- a/searx/engines/voidlinux.py
+++ b/searx/engines/voidlinux.py
@@ -1,0 +1,91 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""SearXNG engine for `Void Linux binary packages`_.  Void is a general purpose
+operating system, based on the monolithic Linux kernel. Its package system
+allows you to quickly install, update and remove software; software is provided
+in binary packages or can be built directly from sources with the help of the
+XBPS source packages collection.
+
+.. _Void Linux binary packages: https://voidlinux.org/packages/
+
+"""
+
+import re
+
+from urllib.parse import quote_plus
+from searx.utils import humanize_bytes
+
+about = {
+    'website': 'https://voidlinux.org/packages/',
+    'wikidata_id': 'Q19310966',
+    'use_official_api': True,
+    'official_api_documentation': None,
+    'require_api_key': False,
+    'results': 'JSON',
+}
+
+categories = ['packages', 'it']
+
+base_url = "https://xq-api.voidlinux.org"
+pkg_repo_url = "https://github.com/void-linux/void-packages"
+
+void_arch = 'x86_64'
+"""Default architecture to search for.  For valid values see :py:obj:`ARCH_RE`"""
+
+ARCH_RE = re.compile('aarch64-musl|armv6l-musl|armv7l-musl|x86_64-musl|aarch64|armv6l|armv7l|i686|x86_64')
+"""Regular expresion that match a architecture in the query string."""
+
+
+def request(query, params):
+    arch_path = ARCH_RE.search(query)
+    if arch_path:
+        arch_path = arch_path.group(0)
+        query = query.replace(arch_path, '').strip()
+    else:
+        arch_path = void_arch
+
+    params['url'] = f"{base_url}/v1/query/{arch_path}?q={quote_plus(query)}"
+    return params
+
+
+def response(resp):
+    """
+    At Void Linux, several packages sometimes share the same source code
+    (template) and therefore also have the same URL.  Results with identical
+    URLs are merged as one result for SearXNG.
+    """
+
+    packages = {}
+    for result in resp.json()['data']:
+
+        # 32bit and dbg packages don't have their own package templates
+        github_slug = re.sub(r"-(32bit|dbg)$", "", result['name'])
+        pkg_url = f"{pkg_repo_url}/tree/master/srcpkgs/{github_slug}"
+
+        pkg_list = packages.get(pkg_url, [])
+        pkg_list.append(
+            {
+                'title': result['name'],
+                'content': f"{result['short_desc']} - {humanize_bytes(result['filename_size'])}",
+                'package_name': result['name'],
+                'version': f"v{result['version']}_{result['revision']}",
+                'tags': result['repository'],
+            }
+        )
+        packages[pkg_url] = pkg_list
+
+    results = []
+    for pkg_url, pkg_list in packages.items():
+
+        results.append(
+            {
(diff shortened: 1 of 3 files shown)


return42:
> * At Void Linux, sometimes multiple packages share the same template / source code / result url, for example `zsh-dbg` is contained in the `zsh` package template: Hence I've added a new field to skip deduplicating results that have the same url - because otherwise it would just remove packages like `zsh-dbg` from the search results (maybe there's a better way to achieve this, but this seemed like the easiest one)

Due to a lack of knowledge about packaging in voidlinux, I haven't quite understood this yet / how could it be that identical URLs can point to different content. To understood by example, I tried `!void zsh-dbg` with the current implementation but I got only one result item.

Can you please explain or give me an example with more than one result / thanks :+1:

Bnyro:
> Due to a lack of knowledge about packaging in voidlinux, I haven't quite understood this yet / how could it be that identical URLs can point to different content. To understood by example, I tried `!void zsh-dbg` with the current implementation but I got only one result item.
> 
> Can you please explain or give me an example with more than one result / thanks 👍
>
Yes, of course!

For example, if you search for `LibreCAD` (reference: https://voidlinux.org/packages/?arch=x86_64&q=LibreCAD), you will get 4 results:
1. The first two results are just duplicated `LibreCAD` entries - no clue why there are actually two of them - in theory this should only be one entry (maybe we have to deduplicate by the title for this engine :thinking:)
2. LibreCAD-32bit: This is exactly the same package as LibreCAD, but compiled for 32bit machines
3. LibreCAD-dbg: This is a package containing debug files - useful for developers or users when debugging crashes or other unwanted behavior

However, the `32bit` and the `dbg` package don't have their own templates (templates are like build instructions for packages - in Void Linux they are just plain bash scripts :) ) - and therefore are located at the same URL as the normal LibreCAD package, which is https://github.com/void-linux/void-packages/tree/master/srcpkgs/LibreCAD.

All of them can be install with the Vod Linux package manager XBPS via `xbps-install LibreCAD LibreCAD-32bit LibreCAD-dbg` though.

return42:
> All of them can be install with the Void Linux package manager XBPS via `xbps-install LibreCAD LibreCAD-32bit LibreCAD-dbg` though.

Thanks for explanation .. than I think it is better we merge the packages under a single result item (based on the URL).

I pushed a WIP with a suggestion how it could by / not sure you are happy with.

Further the WIP implements a regular expression for the query string to choose a architecture / in SearXNG it will looks like this:

image

But what irritates me a little: When I search for https://voidlinux.org/packages/?arch=aarch64&q=librecad on voidlinux.org the UI shows three results ..

image

But both, SearXNG and the UI from voidlinux.org do the same query --> https://xq-api.voidlinux.org/v1/query/aarch64?q=librecad with the identical JSON result that contains only two items ..

image

Why does the UI at voidlinux.org shows three results? .. the first and the second in the screenshot above are identical .. is it a bug in the UI of voidlinux.org/packages ?

searx/engines/voidlinux.py
```diff
@@ -22,30 +22,63 @@
...
+    for pkg_url, pkg_list in packages.items():
+
+        results.append(
+            {
+                'url': pkg_url,
+                'template': 'packages.html',
+                'title': ' | '.join(x['title'] for x in pkg_list),
+                'content': ' | '.join(x['content'] for x in pkg_list),
+                'package_name': ' | '.join(x['package_name'] for x in pkg_list),
+                'version': ' | '.join(x['version'] for x in pkg_list),
```
Bnyro: Packages that are `dbg` or `32bit` always have the same version as the main package, so I think we can just use `pkg_list[0]['version']` here.

Bnyro:
> > All of them can be install with the Void Linux package manager XBPS via `xbps-install LibreCAD LibreCAD-32bit LibreCAD-dbg` though.
> 
> Thanks for explanation .. than I think it is better we merge the packages under a single result item (based on the URL).
> 
> I pushed a WIP with a suggestion how it could by / not sure you are happy with.
> 
Yes, that seems like a good idea to me :+1: I'm just thinking about whether we can reduce the amount of text a bit - the packages always share the same version and an almost identical name (only with "(debug files)" or "(32bit)" appended)
> Further the WIP implements a regular expression for the query string to choose a architecture / in SearXNG it will looks like this:
> 
> image
> 
That's a good idea, agreed, thanks :+1:
> But what irritates me a little: When I search for https://voidlinux.org/packages/?arch=aarch64&q=librecad on voidlinux.org the UI shows three results ..
> 
> image
> 
> But both, SearXNG and the UI from voidlinux.org do the same query --> https://xq-api.voidlinux.org/v1/query/aarch64?q=librecad with the identical JSON result that contains only two items ..
> 
> image
> 
> Why does the UI at voidlinux.org shows three results? .. the first and the second in the screenshot above are identical .. is it a bug in the UI of voidlinux.org/packages ?
>
That's a very good question and I don't have an answer for it either. I suspect that since the query `librecad` matches exactly the package name (also `LibreCAD`), this one is shown a second time intentionally at the top at the void linux website, because it's the most likely result the user has searched for. All the other results are sorted alphabetically, so example if you search for `zsh` the correct result would be right at the bottom - so it makes sense to place a copy of it at the top if there's a package called exactly the query `zsh`. I didn't really test that, but that would be my thought.

searx/engines/voidlinux.py
```diff
@@ -0,0 +1,84 @@
...
+
+    results = []
+    for pkg_url, pkg_list in packages.items():
+
+        results.append(
+            {
+                'url': pkg_url,
+                'template': 'packages.html',
+                'title': ' | '.join(x['title'] for x in pkg_list),
+                'content': ' | '.join(x['content'] for x in pkg_list),
```
Bnyro: The content just differs by an appended `(debug files)` or `(32bit)` - I think it's the best to only use one of the package descriptions (probably using the first one is the easiest way here too).

return42 (approved):
Modified according to the remarks from @Bnyro / squashed all modifications into the first commit.

return42:
Ready to merge?

Bnyro:
From my side, yes :+1:

PR #3247, opened 2024-02-27
[feat] engine: implementation of cppreference

thunder-coding:
## What does this PR do?
Implement cppreference engine

Search for C++ reference documentation

Start dev server and search for `std::vector`, `std::string` or any other C++ reference

## Author's checklist
- [X] search functionality works

Also cppreference does support a bunch of other languages, but they aren't complete. Pages are missing, as well as content may be incomplete. I don't know any of those languages in there. Also I don't have any experience working with localization and stuff, would be nice if anyone else does take care of that.

## Related issues
None

### DIFF
--- a/searx/engines/cppreference.py
+++ b/searx/engines/cppreference.py
@@ -1,0 +1,39 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""Cppreference
+"""
+from lxml import html
+from searx.utils import eval_xpath
+
+
+about = {
+    "website": "https://en.cppreference.com/",
+    "wikidata_id": None,
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+
+categories = ['it']
+url = 'https://en.cppreference.com/'
+search_url = url + 'mwiki/index.php?title=Special%3ASearch&search={query}'
+
+
+def request(query, params):
+    params['url'] = search_url.format(query=query)
+    return query
+
+
+def response(resp):
+    results = []
+    dom = html.fromstring(resp.text)
+    for result in eval_xpath(dom, '//div[contains(@class, "mw-search-result-heading")]'):
+        results.append(
+            {
+                'url': url + eval_xpath(result, './/a/@href')[0],
+                'title': eval_xpath(result, './/a/text()')[0],
+            }
+        )
+    return results
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -477,6 +477,12 @@ engines:
   #   # get your API key from: https://core.ac.uk/api-keys/register/
   #   api_key: 'unset'
 
+  - name: cppreference
+    engine: cppreference
+    shortcut: cpp
+    paging: false
+    disabled: true
+
   - name: crossref
     engine: crossref
     shortcut: cr


return42:
> would be nice if anyone else does take care of that.

`cppreference.com` is a [MediaWiki](https://www.mediawiki.org/wiki/MediaWiki) / we have a (abstract) [MediaWiki engine](https://docs.searxng.org/dev/engines/mediawiki.html) that supports languages:

https://github.com/searxng/searxng/blob/c0b97c65438be44dd5f5e64786576763993ce424/searx/engines/mediawiki.py#L97-L98

May you want to give it a try / its very easy to setup, here is an example from the `settings.yml` file:

https://github.com/searxng/searxng/blob/c0b97c65438be44dd5f5e64786576763993ce424/searx/settings.yml#L735-L745

If you look out for `engine: mediawiki` you will find a lot of other examples.

thunder-coding:
@return42 the problem is that searxng's mediawiki engine assumes a lot of stuff. Like it appends `/w/` to the base url, cppreference uses `/mwiki/` instead of `/w/`, also addition of extra parameters seem to confuse cppreference even if I remove the `/w/` part

unixfox:
Please disable the engine by default.

return42:
> @return42 the problem is that searxng's mediawiki engine assumes a lot of stuff. Like it appends `/w/` to the base url, cppreference uses `/mwiki/` instead of `/w/`, also addition of extra parameters seem to confuse cppreference even if I remove the `/w/` part

TL;DR; API seems not suitable / Here is what I tried (without luck):

```diff
diff --git a/searx/engines/mediawiki.py b/searx/engines/mediawiki.py
index 6a9ac974a..cd7835a35 100644
--- a/searx/engines/mediawiki.py
+++ b/searx/engines/mediawiki.py
@@ -101,6 +101,8 @@ base_url: str = 'https://{language}.wikipedia.org/'
   ISO 639-1 language code (en, de, fr ..) of the search language.
 """
 
+api_path = 'w/api.php'
+
 timestamp_format = '%Y-%m-%dT%H:%M:%SZ'
 """The longhand version of MediaWiki time strings."""
 
@@ -115,9 +117,9 @@ def request(query, params):
         params['language'] = params['language'].split('-')[0]
 
     if base_url.endswith('/'):
-        api_url = base_url + 'w/api.php?'
+        api_url = base_url + f'{api_path}?'
     else:
-        api_url = base_url + '/w/api.php?'
+        api_url = base_url + f'/{api_path}?'
     api_url = api_url.format(language=params['language'])
 
     offset = (params['pageno'] - 1) * number_of_results
diff --git a/searx/settings.yml b/searx/settings.yml
index 72cf4e7ac..4136dc0f6 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1895,6 +1895,16 @@ engines:
     categories: images
     number_of_results: 10
 
+  - name: cppreference
+    engine: mediawiki
+    shortcut: cpp
+    search_type: title
+    categories: [it]
+    base_url: "https://{language}.cppreference.com/"
+    api_path: "mwiki/api.php"
+    srprop: "snippet|timestamp"
+
+
   - name: wolframalpha
```

with I get URLs like https://de.cppreference.com/mwiki/api.php?action=query&list=search&format=json&srsearch=c_str&sroffset=0&srlimit=5&srwhat=title&srprop=snippet%7Ctimestamp&srsort=relevance&srenablerewrites=1

grafik

Not sure, but to me it looks like the mediawiki api of cppreference.com is old

- https://en.cppreference.com/mwiki/api.php?action=help ...

grafik

compared to the actual one:

- https://www.mediawiki.org/w/api.php?action=help

grafik

Issue #3210, opened 2024-02-13
Implement a dedicated template for package managers


return42:
**Is your feature request related to a problem? Please describe.**

> actually a dedicate template for package would be useful.
For example, if I search for a package using `!pypi` the version is missing, so now I use `!!pypi`. YMMV.

_Originally posted by @dalf in https://github.com/searxng/searxng/issues/3200#issuecomment-1940714906_

**Describe the solution you'd like**

We should have an eye on the existing engines we already have for package managers

> We already have support for other package managers (e.g. conda, archlinux, go) 

**Describe alternatives you've considered**

> .. and they all make use of the default template.

**Additional context**

The request comes up in the review of #3200 .. which included initial the [nix-package.html](https://github.com/searxng/searxng/blob/f9671ea712a4f63ec8b9c9acb25686c1ba12e5ec/searx/templates/simple/result_templates/nix-package.html) template / may this template could be a starting point for further considerations (?).

PR #3249, opened 2024-02-28
pypi engine: use packages.html

dalf:
## What does this PR do?

The pypi engine uses the template packages.html

image

Without this change, the version is not displayed

## Author's checklist

The field `name` and the title are redundant.
Why the field `name` is mandatory?

## Related issues

https://github.com/searxng/searxng/issues/3210

### DIFF
--- a/searx/engines/pypi.py
+++ b/searx/engines/pypi.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""pypi.org
+
+"""
+
+from urllib.parse import urlencode
+from dateutil import parser
+
+from lxml import html
+from searx.utils import (
+    eval_xpath_getindex,
+    eval_xpath_list,
+    extract_text,
+)
+
+# about
+about = {
+    "website": "https://pypi.org",
+    "wikidata_id": "Q2984686",
+    "official_api_documentation": "https://warehouse.readthedocs.io/api-reference/index.html",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "HTML",
+}
+
+categories = ['it', 'packages']
+
+
+# engine dependent config
+first_page_num = 1
+base_url = "https://pypi.org"
+search_url = base_url + '/search/?{query}'
+
+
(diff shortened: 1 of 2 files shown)


return42 (approved):
Thanks a lot :+1: / I squashed my minor modifications ([diff](https://github.com/searxng/searxng/compare/39f16600307bbff945f11ac67e9cf7bffba6b38e..66a1cc6a8b2e507f7d5e1fbc769bfc6a9d2bd3db)) / mainly about pylint and types.

> Why the field `name` is mandatory?

you mean `package_name` (?) .. on pypi the name of the *product* is equal to the name of the package .. but on other package providers the package name does not necessarily have to be the same as the name of the product.  In the `packages.html"` the `title` field holds the name of the product while the `package_name` holds the name of the package.

dalf:
I understand, but is there a reason to make it mandatory for pypi?

[EDIT] screenshot if package_name is not specify:

image

[EDIT2] I try to understand why package-name is mandatory. Perhaps the idea is to be able to compare results from different engines using the packages.html template? This lead me to the question, why for some engine the title is not the package name? ping @Bnyro @return42

Bnyro:
Package name refers to the name you use to install the package (e.g. pip install).

The title however should be the project's real name. For example at Go packages, at https://pkg.go.dev/search?q=fiber.

For the first result, the title would be `fiber` while the package name used to install it would be github.com/gofiber/fiber/v2. (`go get github.com/gofiber/fiber/v2`)

Bnyro:
What I described above doesn't fit all engines though, so we might just want to make `title` optional and use `package_name` as title if no `title` is provided?

Bnyro (approved):

dalf:
Thank you, I understand the golang context.

The package name is displayed once on pypi:
image

> What I described above doesn't fit all engines though, so we might just want to make title optional and use package_name as title if no title is provided?

And remove the package_name field in this case?

Bnyro:
It's okay for me to keep the package name field and have it being shown twice, but I dont have a strong opinion it. Your suggestion is fine with me too.

PR #3262, opened 2024-03-01
Run the checker on Friday

dalf:
## What does this PR do?

* Run the checker on Friday.
* The checker runs all the tests and then display the results to make them more readable.

Give an overview of which engines works or not.
Related to the growing number of engines.

* `make search.checker`
* or verbose mode: `./local/py3/bin/searxng-checker -v google`

## Author's checklist

An extension might be this:
* the GH workflows create a json from the results
* this json is included in source code
* the instances can notify the users about the broken the engines (broken according to the checker)

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -1,0 +1,31 @@
+name: "Checker"
+on:
+  schedule:
+    - cron: "0 4 * * 5"
+  workflow_dispatch:
+
+jobs:
+  checker:
+    name: Checker
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Ubuntu packages
+        run: |
+          sudo ./utils/searxng.sh install packages
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          architecture: 'x64'
+
+      - name: Install Python dependencies
+        run: |
+          make V=1 install
+
+      - name: Checker
+        run: |
+          make search.checker
--- a/searx/search/checker/__main__.py
+++ b/searx/search/checker/__main__.py
@@ -55,7 +55,7 @@ def iter_processor(engine_name_list):
             if processor is not None:
                 yield name, processor
             else:
-                stdout.write(f'{BOLD_SEQ}Engine {name:30}{RESET_SEQ}{RED}Engine does not exist{RESET_SEQ}')
+                stdout.write(f'{BOLD_SEQ}Engine {name:30}{RESET_SEQ}{RED}Engine does not exist{RESET_SEQ}\n')
     else:
         for name, processor in searx.search.PROCESSORS.items():
             yield name, processor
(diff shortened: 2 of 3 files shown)


dalf:
Note: I don't know how many engines are going to be blocked on the GitHub runners, but at least it will give an overview.

searx/search/checker/__main__.py
```diff
@@ -64,12 +64,17 @@ def iter_processor(engine_name_list):
 # actual check & display
 def run(engine_name_list, verbose):
     searx.search.initialize()
+    name_checker_list = []
     for name, processor in iter_processor(engine_name_list):
         stdout.write(f'{BOLD_SEQ}Engine {name:30}{RESET_SEQ}Checking\n')
         if not sys.stdout.isatty():
             stderr.write(f'{BOLD_SEQ}Engine {name:30}{RESET_SEQ}Checking\n')
```
return42, line 71: What was the intention to log on stderr additional when it is not a tty?

.github/workflows/checker.yml
```diff
@@ -0,0 +1,31 @@
...
+          python-version: '3.11'
+          architecture: 'x64'
+
+      - name: Install Python dependencies
+        run: |
+          make V=1 install
+
+      - name: Checker
+        run: |
+          make search.checker
```
return42, line 31: Not related to this PR: does it make sense to have an option to store the result in a JSON (for post processing purpose)?

return42 (approved):
For me it works / I left some comments .. what comes into my mind when I reviewed .. not necessarily related to this PR.

----

I see trillions of tracebacks in my tty / may we do not show the full traceback, just one line should be enough .. or we completely suppress these messages that are displayed when loading the images:

```
 File "/share/searxng/local/py3/lib/python3.12/site-packages/httpcore/_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.ReadError

The above exception was the direct cause of the following exception:

  File "/share/searxng/searx/search/checker/impl.py", line 92, in _download_and_check_if_image
    r.close()
```

return42:
> Note: I don't know how many engines are going to be blocked on the GitHub runners, but at least it will give an overview.

Just for reference, here is the output when I run the checker from my private IP: https://gist.github.com/return42/a4538daa4c8676d32edb2b7a5282e51a

Most of the `img_src URL is invalid` errors seem to me to be false positives / I was able to load the images in all my samples / even with a simple curl:

    curl 'https://www.artic.edu/iiif/2//f8fd76e9-c396-5678-36ed-6a348c904d27/full/843,/0/default.jpg' > test.jpg

Here is an excerpt from the report linked above / all images can be loaded:

```
Engine artic                         Error
    found languages: en eo fr
    simple         : img_src URL is invalid https://www.artic.edu/iiif/2//f8fd76e9-c396-5678-36ed-6a348c904d27/full/843,/0/default.jpg (query='life' lang='all' pageno=1 safesearch=0 time_range=None)
    simple         : img_src URL is invalid https://www.artic.edu/iiif/2//a38e2828-ec6f-ece1-a30f-70243449197b/full/843,/0/default.jpg (query='life' lang='all' pageno=1 safesearch=0 time_range=None)
    simple         : img_src URL is invalid https://www.artic.edu/iiif/2//4a04138f-43d8-cd9f-5ac4-478cd8828210/full/843,/0/default.jpg (query='life' lang='all' pageno=1 safesearch=0 time_range=None)
    simple         : img_src URL is invalid https://www.artic.edu/iiif/2//6f513908-03cc-b974-633b-adfce56b7936/full/843,/0/default.jpg (query='life' lang='all' pageno=1 safesearch=0 time_range=None)
    simple         : img_src URL is invalid https://www.artic.edu/iiif/2//d9bde524-38b2-4262-3338-e4d06a50746d/full/843,/0/default.jpg (query='life' lang='all' pageno=1 safesearch=0 time_range=None)
```

dalf:
Some additional general context: the checker was forgotten, the tests are not up to date, the code was broken (see the first commit).

The purpose here is not to fix everything, otherwise the PR won't be merge before next decade. The purpose is rather to have reminder of what thing that might be broken, then we can fix them. If it becomes also forgotten, then we can be remove the embedded checker... If it gets traction, then it might be a good thing for the QA.

.github/workflows/checker.yml
dalf: Yes, clearly.
I remember there are some GH action that annotated the source code (?), we can use that too (in addition to the JSON output).

return42:
> The purpose here is not to fix everything, otherwise the PR won't be merge before next decade. 

Fully agree :+1:  .. from my side this PR is ready to merge / lets merge and see how it progress

searx/search/checker/__main__.py
dalf: `make search.checker > output.log` : 
* /dev/stderr shows the progress
* /dev/stdout records the details into output

PR #3263, opened 2024-03-01
npm engine: use packages.html

dalf:
## What does this PR do?

image

## Author's checklist

@return42 / @Bnyro feel free to take over this PR.

Also the API provides authors, publisher and maintainers: the authors are mapping into the maintainer field inside the packages.html template because this is the expected value (I guess?).

In the API, the field named `flags` might contains `unstable` or `deprecated`. These values are displayed as `tags`.

In the API there are 4 fields about the quality:
* quality
* popularity
* maintenance
* a global score

In packages.html there is only quality, to avoid misunderstanding, all the values are ignored. If quality is included, from the user point of view the question will be "is it the score displayed as quality? or is it only the quality but what about the other values?"

## Related issues

#3210

### DIFF
--- a/searx/engines/npm.py
+++ b/searx/engines/npm.py
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# lint: pylint
+"""npms.io
+
+"""
+
+from urllib.parse import urlencode
+from dateutil import parser
+
+
+about = {
+    "website": "https://npms.io/",
+    "wikidata_id": "Q7067518",
+    "official_api_documentation": "https://api-docs.npms.io/",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ['it', 'packages']
+
+
+# engine dependent config
+paging = True
+page_size = 25
+search_api = "https://api.npms.io/v2/search?"
+
+
+def request(query: str, params):
+
+    args = urlencode(
+        {
+            'from': (params["pageno"] - 1) * page_size,
+            'q': query,
+            'size': page_size,
+        }
+    )
+    params['url'] = search_api + args
+    return params
+
+
+def response(resp):
+    results = []
+    content = resp.json()
+    for entry in content["results"]:
+        package = entry["package"]
+        publishedDate = package.get("date")
+        if publishedDate:
+            publishedDate = parser.parse(publishedDate)
+        tags = list(entry.get("flags", {}).keys()) + package.get("keywords", [])
+        results.append(
+            {
+                "template": "packages.html",
+                "url": package["links"]["npm"],
+                "title": package["name"],
+                'package_name': package["name"],
+                "content": package.get("description", ""),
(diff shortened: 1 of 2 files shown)


dalf:
I've fixed the API usage and the parameters are URL encoded as expected.

return42:
> @return42 / @Bnyro feel free to take over this PR.

Thanks a lot :rocket:

dalf:
I've updated the "Author's checklist" section.

searx/engines/npm.py
```diff
@@ -0,0 +1,75 @@
...
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ['it', 'packages']
+
+
+# engine dependent config
+paging = True
+first_page_num = 0
```
Bnyro: This is no longer needed I think?

Bnyro (approved):

searx/engines/npm.py
```diff
@@ -46,28 +44,23 @@ def response(resp):
...
-                "title": title,
-                'package_name': title,
-                "content": content,
-                "version": version,
-                "maintainer": maintainer,
-                'publishedDate': parser.parse(publishedDate),  # type: ignore
+                "url": package["links"]["npm"],
+                "title": package["name"],
+                'package_name': package["name"],
+                "content": package.get("description", ""),
```
return42, line 57: Field `content` must not be `None` / at least an empty string is needed to show the result in the result list.

return42:
> I've updated the "Author's checklist" section.

Thanks ...

> Also the API provides authors, publisher and maintainers: the authors are mapping into the maintainer field inside the packages.html template because this is the expected value (I guess?).

I think so .. @Bnyro?

> In packages.html there is only quality, to avoid misunderstanding, all the values are ignored. If quality is included, from the user point of view the question will be "is it the score displayed as quality? or is it only the quality but what about the other values?"

I don't have a clear opinion about .. @Bnyro?

-----

I placed a WIP commit that fixes a issue with the `content` field .. but mainly contains nits to get the code more condensed.

Bnyro:
> > Also the API provides authors, publisher and maintainers: the authors are mapping into the maintainer field inside the packages.html template because this is the expected value (I guess?).
> 
> I think so .. @Bnyro?
> 
Yes.
> > In packages.html there is only quality, to avoid misunderstanding, all the values are ignored. If quality is included, from the user point of view the question will be "is it the score displayed as quality? or is it only the quality but what about the other values?"
> 
> I don't have a clear opinion about .. @Bnyro?
> 
You're referring to the `popularity` field? At least at DockerHub, we use the score as count of downloads, at GitHub it's been the amount of stars :shrug:

return42 (approved):
Squashed my WIP commit ([ref](https://github.com/searxng/searxng/pull/3263#pullrequestreview-1912942825)) into your first commit.

About https://github.com/searxng/searxng/pull/3263#issuecomment-1975272636 : I think we leaf it as it now :+1:

PR #3265, opened 2024-03-01
Drop typing-extensions dependency

dalf:
## What does this PR do?

* typing.TypedDict, typing.Literal exist in Python 3.8

Reduce the number of dependencies

## Author's checklist

Draft : Self removed without tests (and this is Friday afternoon, GitHub Action free tier is saturated)

## Related issues

Close #3260

### DIFF
--- a/searx/enginelib/traits.py
+++ b/searx/enginelib/traits.py
@@ -14,16 +14,13 @@ from __future__ import annotations
 import json
 import dataclasses
 import types
-from typing import Dict, Literal, Iterable, Union, Callable, TypeVar, Optional, TYPE_CHECKING
+from typing import Dict, Literal, Iterable, Union, Callable, Optional, TYPE_CHECKING
 
 from searx import locales
 from searx.data import data_dir, ENGINE_TRAITS
 
 if TYPE_CHECKING:
     from . import Engine
-
-
-Self = TypeVar("Self", bound="EngineTraits")  # Python 3.11, to replace by typing.Self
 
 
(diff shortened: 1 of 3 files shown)


dalf:
Seems ok even with Python 3.8

searx/enginelib/traits.py
```diff
@@ -24,6 +23,9 @@
     from . import Engine
 
 
+Self = TypeVar("Self", bound="EngineTraits")  # Python 3.11, to replace by typing.Self
+
```
return42: > Seems ok even with Python 3.8

Nope .. not in py3.8 nor in any other py-version .. the pyright we use in the CI is not based on the python version installed in the CI / pyright is implemented in TypeScript and runs in node.js / BTW we ignore issues from pyright in the CI.

IDK how to test in python 3.8 .. but in my IDE I see that "Self" replacement does not work .. it will be used for `EngineTraits` class and `EngineTraitsMap` in the code below.

Give me a few minutes, I prepare a WIP commit with a [forward references](https://peps.python.org/pep-0484/#forward-references) as mentioned in my comment https://github.com/searxng/searxng/pull/3260#issuecomment-1973280996

dalf: Simpler with the class name instead of Self

return42: > Simpler with the class name instead of Self

OK, feel free to squash the WIP commit into you initial commit .. after we can merge the PR.

PR #2121, opened 2023-01-22
Version: rely on pyproject.toml and setuptools_scm

dalf:
## What does this PR do?

* Move brand information to Python package rather than settings.yml `pip install .` works with the right version.
* dependencies are moved to setup.py
* requirements.txt contains the dev dependencies including `-e .` so SearXNG is installed
* requirements-dev.txt is removed

TODO:
* [ ] fix `make install` : the script should:
  * check if the virtualenv is valid using `setup.py` and `requirements.txt` (currently `requirements.txt` and `requirements-dev.txt`) 
  * create the virtualenv
  * `pip install -r requirements.txt` <-- install SearXNG and dev dependencies
  * nothing more regarding Python (as far I can tell right now)
* [ ] fix `make docker`
* [ ] see if `setup.py` is required: may be `pyproject.toml` is enough with a small custom Python file for the custom version.
* [ ] Dockerfile: use `pip install -e .` to create the version inside
* [ ] use `searx/_version.py` in the code

## Why is this change important?

Currently, there is a lot of specific code to install SearXNG.

This PR:
* uses a more standard way to describe the package --> Contributors will be able to more easily understand how SearXNG is installed, and I hope contribute.
* is going to simplify the code in `manage` and `utils/lib.sh`.

## How to test this PR locally?

Currently in the right virtualenv: 
* `pip install -r requirements.txt` to install everything.
* `pip install .` to install only SearXNG.

PR #3266, opened 2024-03-01
prototype / demo: use rye

dalf:
## What does this PR do?

rye manages:
* the python version (it downloads the required python is needed).
* the virtualenv, make sure to keep sync
* the dependencies inside the virtualenv

See https://rye-up.com/
With [uv](https://astral.sh/blog/uv), which is not activated by default, rye creates the virtualenv very fast.

uv and rye are written by astral who makes [ruff](https://docs.astral.sh/ruff/) (actually rye is written by the [creator of Flask](https://github.com/mitsuhiko), but the [project has joined astral](https://lucumr.pocoo.org/2024/2/15/rye-grows-with-uv/))

To enable uv, see https://rye-up.com/guide/config/#configtoml
* `~/.rye/config.toml`.
* set `use-uv = true` in the `behavior` section.

This PR is demo / prototype to show how we can remove a lot of boilerplate and create a standard pyproject.toml.

* install rye
* use `make test`, `make run` as usual

## Author's checklist

This is a quick & dirty implementation, expect things to break.

npx is not installed, so `make test.pyright` does nothing.

Also rye does not like symbolic links, so I removed them since there are required for the installation phase but not for development (perhaps that's was only for `rye init`, I have not tried after).

The dynamic creation of version file is not implemented, I guess build-system has to be setuptools instead of hatchling (or may be hatchling can do that).

## Related issues

Related to a PR forgotten at the bottom of the ocean : #2121


dalf:
From my point, the tool is rather new, but it seems in good hands and goes into the right direction.
See https://alpopkes.com/posts/python/packaging_tools/

@return42 @Bnyro @unixfox @mrpaulblack and everyone else, what are your opinions on a tool like rye?

return42:
> @return42 @Bnyro @unixfox @mrpaulblack and everyone else, what are your opinions on a tool like rye?

Looks promising .. Most python build, env and package tools are based on the python itself installed on a system ... since different python versions are installed on all systems this usually does not work well (e.g. pyenv).

I still have to take a closer look at Rye, but at first glance it does not seem to be programmed in python itself / does not have the described problems ..

image

I found all previous solutions I have seen to complicated or too fragile, which is why I have always oriented myself to my shell scripts ... "bash" .. somewhat archaic but it always works.  With rye we could have a real alternative in our hands ... we should definitely pursue this further ..

And it's also from Armin Ronacher ... then it can only be good :)

dalf:
> Looks promising .. Most python build, env and package tools are based on the python itself installed on a system ... since different python versions are installed on all systems this usually does not work well (e.g. pyenv).

It uses prebuilt python version from https://github.com/indygreg/python-build-standalone
See https://gregoryszorc.com/docs/python-build-standalone/main/

> I still have to take a closer look at Rye, but at first glance it does not seem to be programmed in python itself / does not have the described problems ..

It is written in Rust and some Python: the first thing rye does is to download a Python version for its Python scripts. This is seamless, just a line of code.

> I found all previous solutions I have seen to complicated or too fragile

The tool is new and hacky according Armin (I'm not sure he uses exactly this word, but that's the idea), but it gets good comments and reviews. The fact that astral endorse the project is another good point.

Bnyro:
This means that everybody willing to work on SearXNG needs to install `rye`, right? Due to the fact that `rye` is not really a standard package and probably not part of many Linux repositories, I personally am not sure about this PR.

Manually installing such a package before contributing is possible has the downside that some contributors might fail setting things up properly, and also cause inconvenience? (just my opinion)

dalf:
> This means that everybody willing to work on SearXNG needs to install rye, right?

Yes

> Due to the fact that rye is not really a standard package and probably not part of many Linux repositories, I personally am not sure about this PR.

There is an issue about that: https://github.com/astral-sh/rye/issues/452

Also, SearXNG already has many unusual setup:

* nvm is downloaded directly from GitHub.
* go built chain was downloaded directly from golang.org
* redis source coded is downloaded and then compiled outside of any package management.

> Manually installing such a package before contributing is possible has the downside that some contributors might fail setting things up properly, and also cause inconvenience?

I have not checked, may be it is possible to install it inside the `./local` directory, so rye would be installed like other tools. In this case, would it be a problem?

Bnyro:
> I have not checked, may be it is possible to install it inside the ./local directory, so rye would be installed like other tools. In this case, would it be a problem?
>
That sounds great :+1:

dalf:
Installation without user interaction in a custom directory:

1/ Download the binary from https://github.com/astral-sh/rye/releases/tag/

2/ Run

```sh
$ RYE_HOME="$(pwd)/local" rye self install --toolchain /usr/bin/python3.10 --no-modify-path --yes
```

-->

```
Welcome to Rye!

This installer will install rye to /home/aflament/code/searxng/local
This path can be changed by exporting the RYE_HOME environment variable.

Details:
  Rye Version: 0.27.0
  Platform: linux (x86_64)
  Internal Toolchain Path: /usr/bin/python3.10

Installed binary to /home/aflament/code/searxng/local/shims/rye
Registering toolchain at /usr/bin/python3.10
Registered toolchain as cpython@3.10.12
Bootstrapping rye internals
Found a compatible Python version: cpython@3.10.12
Updated self-python installation at /home/aflament/code/searxng/local/self
Skipping PATH modification. You will need to add /home/aflament/code/searxng/local/shims to your PATH manually.

All done!
```

As described in the log, the PATH must contains `local/shims`, and it is all set.

I will try to add that to the PR, so rye won't be a prior requirement.

dalf:
Done: rye is automatically downloaded and installed in the `.local` directory.

Actually now, you don't even need Python: rye is going to download the version defined in `.python-version` (It is possible to ask rye to use `/usr/bin/python3` with `rye toolchain register `).

PR #3269, opened 2024-03-01
Docker: use Alpine 3.19

dalf:
## Author's checklist

I have not checked `make docker.buildx`.

May be we can consider to add the `riscv64` to the list of supported architecture by the docker image? ping @unixfox

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.18
+FROM alpine:3.19
 ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
 EXPOSE 8080
 VOLUME /etc/searxng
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN apk add --no-cache -t build-dependencies \
     uwsgi \
     uwsgi-python3 \
     brotli \
- && pip3 install --no-cache -r requirements.txt \
+ && pip3 install --break-system-packages --no-cache -r requirements.txt \
```
unixfox, line 48: can you explain why this new arg?

unixfox (changes requested):

Dockerfile
dalf: Because the new pip refuses to install the packages without a virtualenv.
Since this is docker image, there is no issue to break the system packages ( = install Python packages not from Alpine).

unixfox (approved):

PR #2981, opened 2023-11-01
[mod] isolation of botdetection from SearXNG core

return42:
In PR-2894[1] we isolated botdetection from the limiter, this PR isolates the botdetection from the SearXNG core code.

This PR also fixes the issue [2] that the ``server.public_instance`` option needs to activate the limiter.

- [1] https://github.com/searxng/searxng/pull/2894
- [2] https://github.com/searxng/searxng/issues/2975

## Resources of `botdetection`

Package `botdetection` at PyPi:

- https://pypi.org/project/botdetection/

Repository of `botdetection` at SearXNG org:

- https://github.com/searxng/botdetection

PR #3272, opened 2024-03-02
Drop pytomlpp dependency for Python >= 3.11

dalf:
## What does this PR do?

Rely on tomllib for Python >= 3.11

## Related issues

- https://github.com/searxng/searxng/discussions/3267
- #2981

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,4 +16,4 @@ redis==5.0.2
 markdown-it-py==3.0.0
 typing_extensions==4.9.0
 fasttext-predict==0.9.2.2
-pytomlpp==1.0.13
+pytomlpp==1.0.13; python_version < '3.11'
--- a/searx/botdetection/config.py
+++ b/searx/botdetection/config.py
@@ -14,7 +14,18 @@ import copy
 import typing
 import logging
 import pathlib
-import pytomlpp as toml
+
+try:
+    import tomllib
+
+    pytomlpp = None
+    USE_TOMLLIB = True
+except ImportError:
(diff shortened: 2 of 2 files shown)


return42:
Thanks for this PR / I will review!  .. ~but first I want to get #2981 reviewed and merged / if it is OK for you.~

return42:
As long we do not see any drawbacks in using pytomlpp I'm not sure we should remove it .. its [4 times faster](https://github.com/hukkin/tomli?tab=readme-ov-file#performance) compared to the pure python implementation in py3.11.

However if we want less dependencies we should replace pytomlpp by tomli in py < 3.11 / tomli is the tomllib module that has been added to the standard library in Python 3.11 and the [compatibility layer](https://github.com/hukkin/tomli?tab=readme-ov-file#building-a-tomlitomllib-compatibility-layer) is simpler, compared the one we need for pytomlpp & standard library.

dalf:
>  its 4 times faster compared to the pure python implementation in py3.11.

I don't think config loading is the SearXNG startup time bottleneck.

> if we want less dependencies we should replace pytomlpp by tomli in py < 3.11

I agree: tomli is a written in pure Python, so there is no additional build time for "uncommon" architectures.

unixfox:
Real pleasure. Building pytomlpp for ARM on alpine Linux takes ages!

unixfox:
Is this PR ready :)?

unixfox (approved):

Bnyro (approved):

PR #3283, opened 2024-03-03
Checker: tests for bahnhof and wikispecies

dalf:
## What does this PR do?

Add a dedicated checker test for the bahnhof and wikispecies engines.

On the weekly checker tests, avoid a false negative for the bahnhof and wikispecies engines.

`make search.checker.bahnhof`

Expected result:

```
Engine bahnhof                       Checking

== Results ======================================================================
Engine bahnhof                       OK
    found languages: de
```

Same for `make search.checker.wikispecies`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -565,6 +565,16 @@ engines:
       require_api_key: false
       results: JSON
       language: de
+    tests:
+      bahnhof:
+        matrix:
+          query: berlin
+          lang: en
+        result_container:
+          - not_empty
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -1860,6 +1870,16 @@ engines:
     about:
       website: https://species.wikimedia.org/
       wikidata_id: Q13679
+    tests:
+      bahnhof:
```
return42: `bahnhof:` C&P error? --> `wikispecies` ?

dalf: Yes, fixed

PR #3288, opened 2024-03-03
internetarchivescholar engine: set timeout to 15 seconds

dalf:
## What does this PR do?

5 seconds is not enough

In the master branch, the internetarchivescholar engine is broken (timeout)

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1570,7 +1570,7 @@ engines:
   - name: internetarchivescholar
     engine: internet_archive_scholar
     shortcut: ias
-    timeout: 5.0
+    timeout: 15.0
 
   - name: superuser
     engine: stackexchange


searx/settings.yml
```diff
@@ -1570,7 +1570,7 @@ engines:
   - name: internetarchivescholar
     engine: internet_archive_scholar
     shortcut: ias
-    timeout: 5.0
+    timeout: 15.0
```
return42, line 1574: On common science term I have a response time of 2-3 sec .. the word `test` is in nearly all papers .. (I have response time 6-8sec) .. I would suggest to increase the response time to 8 and to use a search term other than `test` in the checker.

dalf: Why not 15, so it can return results whatever the query is?

return42: Why not 1h? .. since it is to long, when we have at maximum a response time of 8sec?

Did you had 15sec response time in your sample tests .. or why did you think 15sec is the best value?

dalf: I have a 11 seconds response time for `time`.

return42: > I have a 11 seconds response time for `time`.

Hm .. haven't thought a request took longer than 8sec .. but you are right: `time` has a response time > 11sec

PR #3297, opened 2024-03-06
[fix] engine searchcode.com - paging is broken in searchcode.com's API

### LINKED ISSUE #3287: Bug: searchcode code engine: paging is broken
Paging is supported according to the code:
https://github.com/searxng/searxng/blob/b1431e1670771481961a9e7cf6816d30f6afcfaa/searx/engines/searchcode_code.py#L21

Search for `!searchcode_code time` always returns the same page (page 1, page 2, etc... are identical)

return42:
paging is broken in searchcode.com's API .. not sure it will ever been fixed / this commit disables paging in the engine and BTW pylint `searchcode_code.py`.

Closes: https://github.com/searxng/searxng/issues/3287

### DIFF
--- a/searx/engines/searchcode_code.py
+++ b/searx/engines/searchcode_code.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""
- Searchcode (IT)
+# lint: pylint
+"""Searchcode (IT)
+
 """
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Looks good to me, I can confirm that the search API doesn't currently work the way they describe it in their docs :+1:

PR #3298, opened 2024-03-06
[drop] engine framalibre - sarch URL is no longer supported

### LINKED ISSUE #3286: Bug: framalibre engine
On the master branch, the framalibre engine does not work.

The requested URL returns an 404 error.

return42:
Closes: https://github.com/searxng/searxng/issues/3286

## What does this PR do?

removes engine framalibre / the previous method no longer works and I haven't found an alternative / may its because everything on framalibre is written in French and I don't understand French :o


dalf:
Actually, there is a big JSON file with everything in it: https://framalibre.org/notices.json
Then the search field does not send any request, it "just" look into the content of `notices.json`

dalf:
A possible implementation:
* the initialize function load the .json file and create in a memory SQLite database
* the engine is "offline" and make a SQL query to the database using `MATCH` to search into all the fields.

return42:
> make a SQL query to the database using `MATCH` to search into all the fields.

Such a change would be tantamount to a deletion and new implementation.

Since the engine is not functional in its current implementation, it will be deleted ... I will also post a engine request ..

- https://github.com/searxng/searxng/issues/3315

Issue #3268, opened 2024-03-01
Use the template packages.html


dalf:
Update engines to use packages.html :

* [x] cargo: #3487
* [x] codeberg: https://github.com/searxng/searxng/pull/3468
* [x] gitea : https://github.com/searxng/searxng/pull/3468
* [x] hex: https://github.com/searxng/searxng/pull/3464
* [x] lib.rs: https://github.com/searxng/searxng/pull/3299
* [x] npm: https://github.com/searxng/searxng/pull/3263
* [x] pkg.go.dev: https://github.com/searxng/searxng/pull/3301
* [x] pypi : https://github.com/searxng/searxng/pull/3249
* [x] docker hub
* [ ] anaconda (which belong to the packages category)
* [ ] hoogle
* [ ] metacpan
* [ ] packagist
* [ ] pub.dev
* [ ] rubygems

Bnyro:
I would like to look into lib.rs and pkg.go.dev some day when I have more time again.

PR #3301, opened 2024-03-06
[feat] pkg.go.dev: use packages.html template

Bnyro:
see https://github.com/searxng/searxng/issues/3268


return42 (approved):
FYI: I squashed l10n of the field `popularity` into your commit ([diff](https://github.com/searxng/searxng/compare/5ae15395da9ec2fed8845371ed75ea1199e05559..34641294388d54c9df01ceea5e751c7c8f0af4e4)).

You can test it by changing the UI language from/to EN (15,000) -> DE (15.000)

PR #3302, opened 2024-03-06
[fix] XPath engine azlyrics - new parameter `x` is needed.

### LINKED ISSUE #3280: Bug: azlyrics engine
Engine broken on the master branch

A new `x` parameter is required.

return42:
This commit adds a static `x` argument to the GET URL.  Its not clear how long this static value will work.  The alternative is to implement a python modul for this engine to get the `x` argument from the HTML form of

https://search.azlyrics.com/search.php / the XPath selector to get the value is::

        //input[@name='x']@value

Closes: https://github.com/searxng/searxng/issues/3280

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1093,7 +1093,8 @@ engines:
     disabled: true
     categories: [music, lyrics]
     paging: true
-    search_url: https://search.azlyrics.com/search.php?q={query}&w=lyrics&p={pageno}
+    search_url:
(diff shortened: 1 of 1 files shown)


Bnyro:
It unfortunately seems like the access token `x` you used in this PR is no longer valid, I'm always getting an empty result page in the browser and SearXNG with that token -> I think we should create a Python module instead, as you supposed above

return42:
- https://github.com/searxng/searxng/pull/3433

PR #3308, opened 2024-03-08
Italian translation of infopages

centralscrutinizer21:
Added italian translation of about.md and search-syntax infopages


return42 (approved):
Thanks a lot for IT translation :rocket:

PR #3316, opened 2024-03-09
Implement  typed models for searx/answerers

glanham-jr:
## What does this PR do?

- Introduces typing to the `answers` modules
- in statitics answerer module, switch `if-else` chain with a dictionary lookup
- Introduce a `BaseQuery` class that `SearchQuery` and `RawTextQuery` inherit from.  Needed for the `answerers` module for strong typing.
- Update unit tests for `answers` based on new typings
- Remove unnecessary `str` cast in webapp

I asked a while back if introducing typing to more places would be acceptable, and the feedback seemed to be positive. With this, I wanted to introduce a small typing change on how I want to try and go about introducing these changes. Specifically, work on a lower-risk happy path, then continue to introduce linting.

## Why is this change important?
Improve code maintainability, catch errors before runtime, and a sort of "code as documentation" to assist newer developers.

## How to test this PR locally?
n/a - ensure typings match as expected

## Author's checklist
 - [x] Review using abstract classes for typing dynamically loaded modules

## Related issues
n/a

PR #3330, opened 2024-03-13
[mod] activate pyright checks (in CI)

return42:
We have been using a static type checker (pyright) for a long time, but its check was not yet a prerequisite for passing the quality gate.  It was checked in the CI, but the error messages were only logged.

As is always the case in life, with checks that you have to do but which have no consequences; you neglect them :-)

We didn't activate the checks back then because we (even today) have too much monkey patching in our code (not only in the engines, httpx and others objects are also affected).

We want to replace monkey patching with clear interfaces for a long time, the basis for this is increased typing and we can only achieve this if we make type checking an integral part of the quality gate.

**This PR activates the type check;** in order to pass the check, a few typing were corrected in the code, but most type inconsistencies were deactivated via inline comments.

This was particularly necessary in places where the code uses properties that stick to the objects (monkey patching).  The sticking of properties only happens in a few places, but the access to these properties extends over the entire code, which is why there are many `# type: ignore` markers in the code ... which we will hopefully be able to remove again successively in the future.


return42:
### Some practical hints about the usage of "type annotations":

#### a. make use of `from __future__ import annotations`

In py < v3.11 we use `from __future__ import annotations` / manly to get in use of the modern notation of complex/compound types ..

```python
import typing

foo: typing.Optional[str] = None
```

versus [Shorthand Syntax](https://typing.readthedocs.io/en/latest/source/best_practices.html#shorthand-syntax)

```python
from __future__ import annotations

foo: str | None = None
```

#### b. deprecated names from `typing` until py3.8 EOL

We use type hints from the [`typing`](https://docs.python.org/3/library/typing.html) module as long we support py3.8 / note: most of the type hints in `typing` are deprecated since py3.9

```python
$ python3.8
>>> from typing import Callable
```
versus

```python
$ python3.9
>>> from collections.abc import Callable
```

------

Further of interest:
 
- Python's [Static Typing with Python](https://typing.readthedocs.io/en/latest/index.html)
- my[py]'s [Type hints cheat sheet](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html)
- [unsave Union / AnyOf thread](https://mail.python.org/archives/list/typing-sig@python.org/thread/HC6ZLDMCBX6ES4ORRIDMRXYW5BTUBDB4)

return42:
## dynamic versus static typing

Python is a dynamically typing language, typing it afterwards can have advantages, but sometimes (often?) typing is rather disadvantageous.

An exemplary implementation / that could be solved concretely by overloading, but that is not the solution, the example is merely intended to illustrate the problem ..

As soon as the return value of a call can have different types (aka a `Union`), the advantages of static typing clash with those of dynamic typing.

```python
def foo(a: int|str) -> int|str:
    return a

# A major advantage of dynamic typing .. the caller knows the return type
x = foo(2) / 2
y = foo("hello").upper()
```

In the upper example, the caller knows the return type of a call .. what is typical for a dynamic typing system! 

But the static type checker doesn't know the returned type:

```
  anyof.py:4:5 - error: Operator "/" not supported for types "int | str" and "Literal[2]"
    Operator "/" not supported for types "str" and "Literal[2]" (reportOperatorIssue)
  anyof.py:5:18 - error: Cannot access member "upper" for type "int"
    Member "upper" is unknown (reportAttributeAccessIssue)
2 errors, 0 warnings, 0 informations 
```

In a static typing system additional code that is unnecessarily executed at runtime is needed to calm down the type check: 

```python
x = foo(2)
if isinstance(x, int):  # can never be False
    x = x / 2
y = foo("hello")
if isinstance(y, str):  # can never be False
    y = y.upper()
```

This issue is discussed in the [unsave Union / AnyOf thread](https://mail.python.org/archives/list/typing-sig@python.org/thread/HC6ZLDMCBX6ES4ORRIDMRXYW5BTUBDB4) thread / in this thread Guido van Rossum described the issue in more details --> [post](https://mail.python.org/archives/list/typing-sig@python.org/message/TTPVTIKZ6BFVWZBUYR2FN2SPGB63Z7PH/)

Another alternative is to use [typing.cast](https://adamj.eu/tech/2021/07/06/python-type-hints-how-to-use-typing-cast/). But both methods (`isinstance` & `cast`) are conceptually wrong in my opinion, they try to solve a static type checking problem with methods that are executed at runtime ... the additional load at runtime (no matter how minimal it may seem) is disproportionate to a solution that is only executed once -- as part of the type check.


#### I am not aware of a sensible solution to this problem where the concepts get in each other's way.

Since with Python I always assume that I am working with a dynamically typing language, I would dispense with static typing in these cases ... but that would also mean that I would only use the advantages of static typing in very isolated cases ... which for me then calls the whole static typing of a dynamically typing language into question again.

:shrug: .. meanings?

searx/enginelib/__init__.py
```diff
@@ -76,7 +76,7 @@ class Engine:  # pylint: disable=too-few-public-methods
 
     # settings.yml
 
-    categories: List[str]
+    categories: list[str]
```
glanham-jr, line 79: This changed form `List` to `list` - I believe this breaks Python 3.8? I just had this issue in my PR.

searx/engines/__builtins__.pyi
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
+
+# Ugly hack to avoid errors from pyright when checking the engiens / sadly this
```
glanham-jr, line 4: nitpick: engine typo along with builtins typo

searx/engines/presearch.py
```diff
@@ -128,7 +128,7 @@ def _get_request_id(query, params):
         # performs an IP-based geolocation of the user, we don't want that in
         # SearXNG ;-)
 
-        if l.territory:
```
glanham-jr, line 131: If we are adding `if l.territory`, do we also need to include `if l.language`?

searx/exceptions.py
```diff
@@ -61,7 +61,7 @@ class SearxEngineAccessDeniedException(SearxEngineResponseException):
     """This settings contains the default suspended time (default 86400 sec / 1
     day)."""
 
-    def __init__(self, suspended_time: int = None, message: str = 'Access denied'):
+    def __init__(self, suspended_time: int = 0, message: str = 'Access denied'):
```
glanham-jr, line 64: Changing the default frome None to 0 would print `suspended_time=0`. Is this change going to result in a confusing message? If so, consider using `typing.Optional[int]`.

searx/engines/ask.py
```diff
@@ -2,7 +2,7 @@
 """Ask.com"""
 
 from urllib.parse import urlencode
-import dateutil
+import dateutil.parser
```
glanham-jr, line 5: I've seen a few cases where the imports changed (`import x` to `import x.y`) - just want to double check they aren't accidental.

searx/search/checker/impl.py
```diff
@@ -263,7 +263,7 @@ def _check_infoboxes(self, infoboxes):
     def check_basic(self):
         if len(self.result_container.unresponsive_engines) > 0:
             for message in self.result_container.unresponsive_engines:
-                self._record_error(message[1] + ' ' + (message[2] or ''))
+                self._record_error(message.error_type + ' ' + (str(message.suspended) if message.suspended else ''))
```
glanham-jr, line 266: Just double checking - is it safe to switch from indexer to a field/property access for this object? If so, just resolve this comment.

searx/engines/__builtins__.pyi
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
```
glanham-jr, line 2: Should we add the below comments as a docstring?

glanham-jr (commented):
PR looks good from what I can see, just left some clarifying comments.

glanham-jr:
> I am not aware of a sensible solution to this problem where the concepts get in each other's way.

1. Code would have to be re-structured in a way such that dynamic types aren't accepted/returned. This isn't always easily done, and honestly this is where OOP concepts starts becoming more heavy-handed into the code base. Some parts of the codebase could benefit from this, which is what I'd like to do for specific modules (i.e. the typing added to the [answerers  typing PR](https://github.com/searxng/searxng/pull/3316)). Its not perfect due to forcing a Module to type to a class and hope they match, but that is what we are doing anyway without typing - might as well make it safer outside the module + have unit tests to ensure it matches.

2. Work with unions that don't need to have the runtime checks based on the implementation. Lets say you have a method that takes either an `int` or a `float`, and all you need to do is multiply by 2. Doesn't matter if its an int or float - both support multiplication. On the other hand, having two entirely different classes attempt to run in the same method based on assumed similarities would require they share a common interface, have no checks and hope for the best, or have different paths for execution, results in the `isinstance` code being littered around at runtime.

Not to say the code-base should be entirely re-written for the above two cases. But in some cases we can add some level of static typing to the existing code, and the linter can now help out during development more.

searx/enginelib/__init__.py
return42: Hi @glanham-jr, you can add `from __future__ import annotations` to get [Shorthand Syntax](https://typing.readthedocs.io/en/latest/source/best_practices.html#shorthand-syntax) in py3.8 / see my comment above: https://github.com/searxng/searxng/pull/3330#issuecomment-1993824738

searx/engines/presearch.py
return42: I don't think so .. SXNG only knows tags like `de-DE` or `de` .. but there is no *region-only* tag like `?-DE` .. so we always have a `l.language` when locale is not `None`.

searx/engines/ask.py
return42: Not sure why, but when we `import dateutil` pyright reports:

     searx/engines/ask.py:59:41 - error: "parser" is not a known \
        member of module "dateutil" (reportAttributeAccessIssue)

This might on of the many cases where static typing fails in a dynamic typing system.

searx/search/checker/impl.py
return42: > .. for this object?

this object is a named tuple / if we already give names to the items of a tuple, we should also use them :)

https://github.com/searxng/searxng/blob/74f6dfb576e66e17d3f968b252dfeee0a20eaf28/searx/results.py#L151-L154

searx/engines/presearch.py
glanham-jr: Makes sense to me. If it becomes a problem for any reason, it can be dealt with then.

return42:
> Code would have to be re-structured in a way such that dynamic types aren't accepted/returned.

If we don't want dynamic typing, Python is the wrong language.  One of the reasons why python was chosen at the time was dynamic typing / dynamic typing is intentional .

 .. static typing should be used sensibly without fundamentally reversing the concept of a language.  I see significant advantages in the type annotations, but there are also disadvantages. 

In my opinion, choosing only one and not the other is always the wrong decision :) 

> Work with unions that don't need to have the runtime checks based on the implementation. 

This does not fit for all cases / I moved from C to Python 25 years ago because of its dynamic typing. It has helped me for more than two decades to create small expressive code for my clients in a very short time. 

I have been supporting some of the projects for more than 20 years and thanks to duck-typing the further developments have always been very cost-effective.  Static typing does not offer this flexibility in further development at the same price / changes to static models are most often more expensive

 ... giving all that up just to type everything seems questionable to me.

----

I think when implementing libraries, you should strongly typing ... but when writing applications, you should only use typing in places where delimitation is desired and makes sense .. or with other words: for me typing makes sense on interfaces / typing every single function and object is questionable.

In my experience, one of the biggest challenges in software development is to correctly identify the boundaries of modularization at an early stage and to create interfaces that also meet future requirements (you do not know today) without having to completely throw the model overboard.

searx/engines/ask.py
glanham-jr: Does from x import y work in these scenarios?

glanham-jr:
I think we are in agreement. Typing absolutely everything seems like a weird choice for Python. But in this case, Searxng is already battle tested with its current implementation, so unless a re-write occurs, I do not foresee that the project will easily switch to a new language just for enforced static type checking. So the choice for typing / not typing should be based on the other points you made - flexibility.

I'd also agree on when it is most important to add it - for the interfaces on services which have multiple implementations, typing would be nice. This application has a bunch of "mini libraries" which may benefit to some degree of typing.

> In my experience, one of the biggest challenges in software development is to correctly identify the boundaries of modularization at an early stage and to create interfaces that also meet future requirements (you do not know today) without having to completely throw the model overboard.

💯

return42:
FYI: last force push was just a rebase on branch master / to solve a conflict.

mrpaulblack:
In our last maintainer meeting we decided to not add pyright checks to the CI; CLosing.

PR #3347, opened 2024-03-21
add keyboard shortcuts for copying search results

thunder-coding:
- **[enh] add editorconfig rules for javascript files**
- **[enh] add keybindings for copying URLs**

## What does this PR do?

adds functionality to copy search results with keyboard

helpful for keyboard power users

(c)opy in searxng keybindinngs, (y)ank for vim-like

## Author's checklist

- [X] It works
PS: this PR template is annoyingly large, might make sense to strip down some of the fields

## Related issues

None


searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -435,6 +447,13 @@ searxng.ready(function () {
     }
   }
 
+  function copyURLToClipboard () {
+    var current = document.querySelector('.result[data-vim-selected] h3 a');
+    if (current === null) return;
+    var url = current.getAttribute('href');
+    navigator.clipboard.writeText(url);
+  }
+
```
Bnyro:
```suggestion
  function copyURLToClipboard() {
    const currentUrlElement = document.querySelector('.result[data-vim-selected] h3 a');
    if (currentUrlElement === null) return;
    
    const url = currentUrlElement.getAttribute('href');
    navigator.clipboard.writeText(url);
  }

```

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -174,6 +180,12 @@ searxng.ready(function () {
           des: 'select next search result',
           cat: 'Results'
         },
+        'y': {
+          key: 'y',
+          fun: copyURLToClipboard,
+          des: 'copy URL of the selected result to the clipboard',
```
Bnyro:
```suggestion
          des: 'copy url of the selected result to the clipboard',
```
for consistency, lowercase is probably the better choice here too.

searx/static/themes/simple/src/js/main/keyboard.js
```diff
@@ -63,6 +63,12 @@ searxng.ready(function () {
       des: 'remove focus from the focused input',
       cat: 'Control'
     },
+    'c': {
+      key: 'c',
+      fun: copyURLToClipboard,
+      des: 'copy URL of the selected result to the clipboard',
```
Bnyro:
```suggestion
      des: 'copy url of the selected result to the clipboard',
```
for consistency, lowercase is probably the better choice here too.

Bnyro:
Looks cool, thanks for the pull request!

Small note: I would prefer to move the `[enh] add editorconfig rules for javascript files` commit to a different pull request since it's not related to this pull request.

searx/static/themes/simple/src/js/main/keyboard.js
thunder-coding: I did use `function name ()` as in same file all functions are declared as such including the `toggleHelp` which too takes no other argument so that my changes are consistent with the other code. I am not commiting this part, but will surely rename the variable. Let me know if you want this change to happen.

thunder-coding:
> Small note: I would prefer to move the [enh] add editorconfig rules for javascript files commit to a different pull request since it's not related to this pull request.

I too would have created a separate PR, but I found the change is needed while working on this PR, and both changes are quite small. So I decided to send them together

thunder-coding:
I've applied the changes and also rebased my branch against the master branch. Let me know if you want me to squash the last two commits into one.

Also a suggestion, maybe do not store the generated `*.min.js` files in repository but instead just build them with GH Actions and publish the repository as archives. This should reduce the chances of a mallicious contributor bundling unwanted things in the `*.min.js` files.

return42:
> but instead just build them with GH Actions and publish the repository as archives.

We use the [`make static.build.commit`](https://docs.searxng.org/dev/quickstart.html) workflow to build a commit ..

> Let me know if you want me to squash the last two commits into one.

I can reorder the commits in my review / if it is OK for you.

----

> this PR template is annoyingly large, might make sense to strip down some of the fields

Delete what you think is not needed.

thunder-coding:
> I can reorder the commits in my review / if it is OK for you

that will not preserve my commit signatures, but totally fine!


> Delete what you think is not needed.

 will look into creating a new discussion for that when I get some time

return42:
> that will not preserve my commit signatures, but totally fine!

It will, `Author:` and `AuthorDate` of the commit will be unchanged .. only `Commit:` and `CommitDate` will be modified to the modified commit.

return42 (approved):
@thunder-coding thanks a lot for this PR / very nice addition :+1: 

> > that will not preserve my commit signatures, but totally fine!
> 
> It will, `Author:` and `AuthorDate` of the commit will be unchanged .. only `Commit:` and `CommitDate` will be modified to the modified commit.

sorry for the delay: I cleaned up your patch series .. the author is unchanged / see https://github.com/searxng/searxng/pull/3347/commits

.. if it is OK for you I'm going to merge .. otherwise we have to restore 4fac705 and you have to clean up the patch series by yourself.

thunder-coding:
Feel free to go ahead and merge

PR #3350, opened 2024-03-22
Add user-agent to stract requests

mikkeldenker:
Hi,

I am the core maintainer of [Stract](https://stract.com/). We got a sudden huge spike in search requests/hr (~4x-5x) yesterday which almost took down the site and I am pretty sure it came from some searxng instances. It seems to be much better now, but it would be nice if we could add a user agent to the requests from searxng so we can better handle such spikes in the future and make sure the site doesn't crash.

Cheers!

## What does this PR do?
Adds the user-agent from `searx_useragent()` for search requests to stract.

Almost took down the search engine yesterday.

### DIFF
--- a/searx/engines/stract.py
+++ b/searx/engines/stract.py
@@ -6,6 +6,7 @@ ends.
 """
 
 from json import dumps
+from searx.utils import searx_useragent
 
 about = {
     "website": "https://stract.com/",
@@ -23,7 +24,11 @@ search_url = "https://stract.com/beta/api/search"
 def request(query, params):
     params['url'] = search_url
     params['method'] = "POST"
-    params['headers'] = {'Accept': 'application/json', 'Content-Type': 'application/json'}
+    params['headers'] = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'User-Agent': searx_useragent(),
+    }
     params['data'] = dumps({'query': query, 'page': params['pageno'] - 1})
 
     return params


unixfox:
hello,

the point of searxng is to keep as much privacy as possible, if we start to include searxng (or searx) in the user agent then you may easily distinguish searxng users from browser users.

That is why we have "mozilla firefox" user agent by default for all the engines.

mikkeldenker:
I get that, and I am not trying to diminish anyones privacy. You already use our api and we don't log anything from the requests except date+ip+useragent in ram from nginx. Nothing is stored.

I should probably have added, that when [searxng initially implemented stract](https://github.com/searxng/searxng/pull/3099), @return42 very nicely said that
> if you (stract) have any special needs (UA header or whatever) please let us know.

I want to serve as many search requests from searxng as possible, but if it crashes stract then no requests can be served at all.

return42:
@unixfox we send UA already in other search engines / https://github.com/search?q=repo%3Asearxng%2Fsearxng%20searx_useragent&type=code

If it helps @mikkeldenker to distinguish bots from serious SearXNG instances I think, we should add the UA .. even we all know that a UA header can easily faked.

return42 (approved):
LGTM / if you want headers in the `api/autosuggest` path you should add the following to your patch ..

```diff
diff --git a/searx/autocomplete.py b/searx/autocomplete.py
index e277c6631..30cd77556 100644
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -17,6 +17,7 @@ from searx.engines import (
 )
 from searx.network import get as http_get, post as http_post
 from searx.exceptions import SearxEngineResponseException
+from searx.utils import searx_useragent
 
 
 def update_kwargs(**kwargs):
@@ -157,7 +158,12 @@ def stract(query, _lang):
     # stract autocompleter (beta)
     url = f"https://stract.com/beta/api/autosuggest?q={quote_plus(query)}"
 
-    resp = post(url)
+    headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'User-Agent': searx_useragent(),
+    }
+    resp = post(url, headers=headers)
 
     if not resp.ok:
         return []
```

Bnyro (approved):

Bnyro:
> We got a sudden huge spike in search requests/hr (~4x-5x) yesterday which almost took down the site and I am pretty sure it came from some searxng instances.
>
Sorry to hear that, I really hate how bots continuously try to attack SearXNG instances :/

unixfox:
I fail to understand how adding an user agent will help.
Are you going to block searxng as a whole when there are too many requests?

I mean if you receive too many requests apply some limits based on IP adresses not user agents.

mikkeldenker:
> LGTM / if you want headers in the api/autosuggest path you should add the following to your patch 

That's totally fine. Autosuggest is very fast so we might as well just serve all the requests.

> Sorry to hear that, I really hate how bots continuously try to attack SearXNG instances :/

Yea I can imagine, especially since it makes it more annoying for actual users.

> Are you going to block searxng as a whole when there are too many requests?

The problem is that the requests yesterday came from many different ip addresses. I think some bots scraped legitimate searxng instances very aggresively, which then sent a lot of requests back to us. We already have an ip rate limit, but this is set as high as possible not to disturb users that uses a vpn. The idea is to try a rate limit on a `(useragent, ip)` combination and see if that is enough without having to lower the ip rate limit which I really want to avoid.

Thank you everyone!

unixfox:
> The problem is that the requests yesterday came from many different ip addresses. I think some bots scraped legitimate searxng instances very aggresively, which then sent a lot of requests back to us. We already have an ip rate limit, but this is set as high as possible not to disturb users that uses a vpn. The idea is to try a rate limit on a `(useragent, ip)` combination and see if that is enough without having to lower the ip rate limit which I really want to avoid.

If your service can only handle up to X amount of requests then limiting based on the user agent will not help in the long run. Outside of searxng, one can always send requests with different user agents and with different IP addresses.

To me adding the user agent of searxng sounds like a temporary fix until you get "attacks" from other sides like outside of searxng.

I know very well the struggle of running a public service, I run a public searxng instance and invidious instance that are getting blasted by bots all days.
But if your service can only handle a certain amount of requests maximum then you should look into really implementing hard limits that you know if you receive a lot of traffic your service will continue to be accessible for the majority of users.

You have to remember that you are not Google or Amazon or Microsoft, you do with what you have. And it's ok if some people may get a degraded experience when in high spike of traffic, people can't take for granted that your service will always be available if your service is free to use.

You can only do a limited amount of things with pure rate limits. If you really want to effectively block bad bots then you have to dig into the rabbit hole of fingerprinting, javascript based verifications and so on. Most of which are not great for privacy.

Have you also looked at [bucket rate limit or leaky bucket rate limit](https://medium.com/@vaibhav1403/token-bucket-algorithm-97bf6b2b67e3)? This may solve the issue of disturbing some of your users due to the rate limits.

Have you considered maybe adding some cache? Sometimes the attacker may just well be doing the same request over and over. The cache will help in reducing the ressources used for serving the same requests.

PR #3354, opened 2024-03-25
Update BASE.HTML To Use Title Case In Links

wormuths:
Simply makes cosmetic changes so that the links on the search page use Title Case.

## What does this PR do?

Changes the display of the footer links to use the internationally recognized "Title Case" to display text not intended to be a complete sentence with punctuation.

This makes the links more visually appealing, and more professional looking.  It also improves consistency of the visually displayed information.

This change is cosmetic in nature, and does not change the core functionality in any way.

## Author's Checklist

N/A

## Related Issues

N/A

### DIFF
--- a/searx/templates/simple/base.html
+++ b/searx/templates/simple/base.html
@@ -64,18 +64,18 @@
   </main>
   <footer>
     <p>
-    {{ _('Powered by') }} <a href="{{ url_for('info', pagename='about') }}">searxng</a> - {{ searx_version }} — {{ _('a privacy-respecting, open metasearch engine') }}<br/>
-        <a href="{{ searx_git_url }}">{{ _('Source code') }}</a>
-        | <a href="{{ get_setting('brand.issue_url') }}">{{ _('Issue tracker') }}</a>
-        {% if enable_metrics %}| <a href="{{ url_for('stats') }}">{{ _('Engine stats') }}</a>{% endif %}
+    {{ _('Powered by') }} <a href="{{ url_for('info', pagename='about') }}">SearXNG</a> - {{ searx_version }} — {{ _('A Privacy-Respecting, Open Metasearch Engine') }}<br/>
+        <a href="{{ searx_git_url }}">{{ _('Source Code') }}</a>
+        | <a href="{{ get_setting('brand.issue_url') }}">{{ _('Issue Tracker') }}</a>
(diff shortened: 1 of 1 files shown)


return42:
Hi @wormuths thanks a lot for your PR ..

> This makes the links more visually appealing, and more professional looking.

is that the case? IMO it is not .. most services do use a notation where the first letter is upper case and all other words are in lower case (if it is English) / by example here on github the footer is:

image

wormuths:
> Hi @wormuths thanks a lot for your PR ..
> 
> > This makes the links more visually appealing, and more professional looking.
> 
> IMO it is not .. most services do use a notation where the first letter is upper case and all other words are in lower case (if it is English) / by example here on github the footer is:
> 
> image

There will certainly be people who don't find it "better", but it certainly doesn't hurt anything.  I have found that most things I see have a few things in common.  For example...

Amazon

Amazon

CNN

CNN

Home Assistant

Home Assistant

In fact, even most software menus (browsers seem to be an exception) follow similar rules.

LibreOffice

LibreOffice

Even menus in KDE seem to follow my suggestions.  In fact, the only entry that stands out is the one that doesn't follow the convention, and it appears as though it was an afterthought rather than a fully developed feature.

KDE

That should be "Create Data Project with K3b".  Menus that deviate from the capitalization tend to add punctuation such as "Create data project with K3b..." 

This is my first pull request.  I'm not a developer, and want to support a project I believe in the only way I know how.  I hope this pull request will add some polish to (really) the only part of the user interface presented by SearXNG.  In fact, the only other interface component SearXNG provides are the search categories, and that already follows this format.  It says "Social Media", not "Social media"...

SearXNG

It's obviously up to the developers, but I think my suggestion is more consistent.

Have a great day...

return42:
>  but it certainly doesn't hurt anything.

That is my point .. it hurts a lot:

These strings are translated --> every string we modify has an impact on the 58 translations!

.. and to be consistent we would have to modify most of the option names in the preferences --> translations are gone for most options and names.

Conversely, if we were to fix only the few strings that do not currently follow the currently applied rule, the impact would be limited to a few strings.

> There will certainly be people who don't find it "better", 

Everyone may have their own preferences and I don't want to contradict them.

But since we have a current rule that has already been applied 90% of the time and changing this rule would destroy the translation efforts of the community, I prefer to refrain from changing the rule.

I ask for your understanding / I'm sorry, .. going to close this PR.

Nevertheless, I would like to thank you for your efforts and your contribution.   :+1:  The project lives from your contributions, even if we sometimes have to reject a contribution.

PR #3375, opened 2024-04-06
[feat] preferences: button to enable/disable all engines

### LINKED ISSUE #660: Allow all/Disable all buttons also toggle engines in other tabs
I don't have it installed, I'm just a user of other people's searx and searxng instances.

I see a lot of different instances because I have my default search in Firefox set to the Random Searx Redirector at searx.neocities.org.

Whether a searx or searxng instance, any version, when you click the "Allow all" or "Disable all" buttons in the Engines tab of Preferences, it not only flips all the toggle switches under the Engines tab, but all the toggle switches in the General, Plugins, User Interface, Privacy, and Special Queries tabs too.

I don't think it should do that, as the "Allow all" and "Disable all" buttons are only present under the Engines tab which seems to indicate that they should only apply to the engines in the Engines tab.

Bnyro:
## What does this PR do?
- add two buttons to the engines preferences page: `disable all`, and `enable all`
- clicking these buttons enables/disables all engines of the current category

- it's convenient to for example disable all engines at the beginning and then only select the ones you really wanna use - or enable all for testing which results are the best

- make run
- go to the engines tab in the preferences

## Author's checklist
UI inspired by https://github.com/searx/searx/pull/1979

## Related issues
closes #660


return42 (approved):
Wow cool .. a long awaited feature :rocket: .. works like a charm :)

@unixfox you asked for "oscar style" in the past .. enable/disable all engines is a oscar feature we did not had in simple theme .. what do you think, is PR ready to merge? 

---

@Bnyro I rebased your branch to solve conflict with current master

PR #3377, opened 2024-04-07
[feat] plugins: new calculator plugin

Bnyro:
## What does this PR do?
- add a new calculator plugin that can calculate simple math

- users no longer rely on 3rd party services for basic math equations

- !make run
- search `(2^3+3-5) x 2 * 4` -> `48`

## Author's checklist
- the calculating part is adapted from https://ao.chatoyer.de/questions/2371436/evaluating-a-mathematical-expression-in-a-string

## Related issues
closes https://github.com/searxng/searxng/pull/1022 (which tried building a calculator in a different way with JavaScript)

### DIFF
--- a/searx/plugins/calculator.py
+++ b/searx/plugins/calculator.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Calculate mathematical expressions using ack#eval
+"""
+
+import ast
+import operator
+
+from flask_babel import gettext
+from searx import settings
+
+name = "Basic Calculator"
+description = gettext("Calculate mathematical expressions via the search bar")
+default_on = False
+
+preference_section = 'general'
+plugin_id = 'calculator'
+
+operators = {
+    ast.Add: operator.add,
+    ast.Sub: operator.sub,
+    ast.Mult: operator.mul,
+    ast.Div: operator.truediv,
+    ast.Pow: operator.pow,
+    ast.BitXor: operator.xor,
+    ast.USub: operator.neg,
+}
+
+
+def _eval_expr(expr):
+    """
+    >>> _eval_expr('2^6')
+    4
+    >>> _eval_expr('2**6')
+    64
+    >>> _eval_expr('1 + 2*3**(4^5) / (6 + -7)')
+    -5.0
+    """
+    return _eval(ast.parse(expr, mode='eval').body)
+
+
+def _eval(node):
+    if isinstance(node, ast.Constant) and isinstance(node.value, int):
+        return node.value
+
+    if isinstance(node, ast.BinOp):
+        return operators[type(node.op)](_eval(node.left), _eval(node.right))
+
+    if isinstance(node, ast.UnaryOp):
+        return operators[type(node.op)](_eval(node.operand))
+
+    raise TypeError(node)
+
+
+def post_search(_request, search):
+    # don't run on public instances due to possible attack surfaces
+    if settings['server']['public_instance']:
+        return True
+
+    # only show the result of the expression on the first page
+    if search.search_query.pageno > 1:
+        return True
+
+    query = search.search_query.query
+    # in order to avoid DoS attacks with long expressions, ignore long expressions
+    if len(query) > 100:
+        return True
+
+    # replace commonly used math operators with their proper Python operator
+    query = query.replace("x", "*").replace(":", "/")
+
+    # only numbers and math operators are accepted
+    if any(str.isalpha(c) for c in query):
+        return True
+
+    # in python, powers are calculated via **
+    query_py_formatted = query.replace("^", "**")
+    try:
+        result = str(_eval_expr(query_py_formatted))
+        if result != query:
+            search.result_container.answers['calculate'] = {'answer': f"{query} = {result}"}
+    except (TypeError, SyntaxError, ArithmeticError):
+        pass
+
+    return True
+
+
+def is_allowed():
+    return not settings['server']['public_instance']
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -220,6 +220,7 @@ outgoing:
 #   - 'Ahmia blacklist'  # activation depends on outgoing.using_tor_proxy
 #   # these plugins are disabled if nothing is configured ..
 #   - 'Hostname replace'  # see hostname_replace configuration below
+#   - 'Calculator plugin'
 #   - 'Open Access DOI rewrite'
 #   - 'Tor check plugin'
 #   # Read the docs before activate: auto-detection of the language could be
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -38,7 +38,7 @@
 
 {%- macro plugin_preferences(section) -%}
 {%- for plugin in plugins -%}
-{%- if plugin.preference_section == section -%}
+{%- if plugin.preference_section == section and (plugin.is_allowed() if plugin.is_allowed else True) -%}
 <fieldset>{{- '' -}}
     <legend>{{ _(plugin.name) }}</legend>{{- '' -}}
     <div class="value">


dalf:
I don't know the security risk regarding `ast` module, and the fact that it is access publicly.

Bnyro:
> I don't know the security risk regarding `ast` module, and the fact that it is access publicly.
>
It's possible to DOS it with large requests in theory. Maybe we just want to only parse the input if it's less than 100 chars or something similar? That should probably mitigate all possible issues.

Arbitrary code execution is not an attack surface when using ast, since we parse (and don't execute) everything first before it's evaluated.

Source: https://ao.chatoyer.de/questions/4710247/python-3-are-there-any-known-security-holes-in-ast-literal-evalnode-or-string

dalf:
> Arbitrary code execution is not an attack surface when using ast, since we parse (and don't execute) everything first before it's evaluated.

My thinking was not clear: not carefully craft parser are subject to buffer overflow or similar attacks.

This is not buffer overflow, but see https://github.com/python/cpython/issues/95185 : 
* python 3.11.1 raises an `RecursionError` exception
* python 3.10.11 crashes

If we merge this PR, the minimal version is 3.11.

---

[lark](https://lark-parser.readthedocs.io/en/stable/) has an [example](https://github.com/lark-parser/lark/blob/master/examples/calc.py) which implements a calculator, but this is an additional dependency...

Bnyro:
> My thinking was not clear: not carefully craft parser are subject to buffer overflow or similar attacks.
> 
> This is not buffer overflow, but see [python/cpython#95185](https://github.com/python/cpython/issues/95185) :
> 
Yes, that's why I proposed to limit the input length that triggers the calculator plugin to `100`, so that things like that can't happen.
> If we merge this PR, the minimal version is 3.11.
> 
Well, in my opinion we should have stopped supporting very outdated Python versions a while ago, I still don't get why we still support Python "3.8", "3.9", "3.10". Debian Bookworm already supports Python 3.11, and so does any other popular distro.
> [lark](https://lark-parser.readthedocs.io/en/stable/) has an [example](https://github.com/lark-parser/lark/blob/master/examples/calc.py) which implements a calculator, but this is an additional dependency...
>
I'm not really a fan of a new dependency for a rather small plugin, but if there's no better way it would be fine with me too.

dalf:
[searxng/fasttext-predict](https://github.com/searxng/fasttext-predict) has a [python package](https://pypi.org/project/fasttext-predict/), but I guess it is nearly exclusively used by searxng --> it can provides more or less accurate stats about searxng.

https://pypistats.org/packages/fasttext-predict

Looking at the last graph, 3.11 is more or less 2/3 of the download.

image

If this is accurate, I don't think this is reasonable to ask 1/3 of the users to upgrade their Python versions.

return42 (approved):
Nice simple calculator / I will be happy to use it :)

> This is not buffer overflow, but see [python/cpython#95185](https://github.com/python/cpython/issues/95185) :

I tested with py3.8.0:

```
$ python --version
Python 3.8.0
$ python -c 'import ast; ast.parse("+0"*209644)'
Segmentation fault (core dumped)
```

note: the query term `"+0"*209644` includes a type cast `"+0"` .. but this PR will only query terms with numbers and math operators ..

```python
   # only numbers and math operators are accepted
    if any(str.isalpha(c) for c in query):
        return True
```

> If we merge this PR, the minimal version is 3.11.

I don't think so, since such expressions contain non alphanumeric characters and the operators are limited to:

```python

operators = {
    ast.Add: operator.add,
    ast.Sub: operator.sub,
    ast.Mult: operator.mul,
    ast.Div: operator.truediv,
    ast.Pow: operator.pow,
    ast.BitXor: operator.xor,
    ast.USub: operator.neg,
}
```

> Yes, that's why I proposed to limit the input length that triggers the calculator plugin to 100, so that things like that can't happen.

I think this is a good suggestion.

----

Further I tested (with python 3.8.0) terms like `1/0` which results in a traceback in the log:

```
  File "searxng/searx/plugins/__init__.py", line 203, in call
  File "searxng/searx/plugins/calculator.py", line 66, in post_search
    result = str(_eval_expr(query_py_formatted))
  File "searxng/searx/plugins/calculator.py", line 35, in _eval_expr
    return _eval(ast.parse(expr, mode='eval').body)
  File "searxng/searx/plugins/calculator.py", line 43, in _eval
    return operators[type(node.op)](_eval(node.left), _eval(node.right))
ZeroDivisionError: division by zero
```

Not sure .. may we wan't to catch such exceptions also?

dalf:
> note: the query term "+0"*209644 includes a type cast "+0" .. but this PR will only query terms with numbers and math operators 

The code does not match the comment: the code accept any query as long there is no letters (a-z, A-Z).

* `"+0"*209644` is evaluated by `_eval`.
* once parsed by ast, "+0" is an instance `ast.Constant`, however `node.value` is not an instance int --> this raise the exception `TypeError(node)`.

==> the plugin does not crash Python with expression `"+0"*209644` on Python < 3.11
==> there is no requirement to have Python >= 3.11 as far I know.

I'm nitpicking because it is difficult to think about all the possible exploits.
Example suggested by ChatGPT: `2**1000000000` can easily creates an DOS (CPU usage at 100% for a time).

For an personal / small group / family instance, that's not an issue, the current implementation is fine.
For public instance, that's a different story for my point of view.

> may we wan't to catch such exceptions also

`except (TypeError, SyntaxError) as e:` does not catch `ZeroDivisionError` --> it should catch `ArithmeticError`.

return42:
> > may we wan't to catch such exceptions also
> 
> `except (TypeError, SyntaxError) as e:` does not catch `ZeroDivisionError` --> it should catch `ArithmeticError`.

:white_check_mark: 

> The code does not match the comment: the code accept any query as long there is no letters (a-z, A-Z).

Argh .. sorry ..

> ==> the plugin does not crash Python with expression `"+0"*209644` on Python < 3.11 ==> there is no requirement to have Python >= 3.11 as far I know.

:white_check_mark: 

> I'm nitpicking because it is difficult to think about all the possible exploits. Example suggested by ChatGPT: `2**1000000000` can easily creates an DOS (CPU usage at 100% for a time).

It's not nitpicking ... I didn't see the problems at first glance / thanks for the heads up :+1: 

I tested the query `2**1000000000`; in python 3.8.0 (please note, this is patch level 0) several CPU cores are utilized at 100% and in python 3.12.0 I got:

```
  File "searxng/searx/plugins/__init__.py", line 203, in call
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "searxng/searx/plugins/calculator.py", line 66, in post_search
    result = str(_eval_expr(query_py_formatted))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Exceeds the limit (4300 digits) for integer string conversion; use sys.set_int_max_str_digits() to increase the limit
```

About `sys.set_int_max_str_digits()` please read [Integer string conversion length limitation](https://docs.python.org/3/library/stdtypes.html#integer-string-conversion-length-limitation)

- Limiting conversion size offers a practical way to avoid [CVE-2020-10735](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-10735).

With vulnerability in mind I installed python 3.8.19 (please note, this is patch level 19) CPU cores are **no longer** utilized at 100% .. and in 3.8.19 the same exception is raised as shown from 3.12.0 above.

> For an personal / small group / family instance, that's not an issue, the current implementation is fine. For public instance, that's a different story for my point of view.

I think you are right with your concerns and basically we will not be able to catch all potentially harmful terms as long we gave the AST *"to the outside"*.

But I also have a doubt that we will ever be able to implement a calculator that catches all vulnerabilities .. not with [lark](https://lark-parser.readthedocs.io/en/stable/) (which is just about grammar parsing / alternative we can limit to 100 characters) nor with any other new dependency / personally I expect more vulnerabilities with more dependencies.

Personally, I would like to have such a calculator, but I don't know how to decide ... There will always be vulnerabilities (no matter which tools) ...  any suggestions how to go further?

Bnyro:
https://github.com/searxng/searxng/compare/453ac043a000a8de6ed652695c29b3f7c53d2e37..c1d009309ca4f72a02ee183f5cd4979cb60559e4:
- catch `ArithmeticException`
- limit input length that triggers the plugin to 100 characters
- disable plugin by default

My ideas:
- We currently don't have a way to allow instance administrators to completely forbid the usage of a plugin at instance level, do we? That would be useful for just de-activing the plugin by default and only run it on private instances.
- Or we could automatically forbid using it when `server.public_instance` is set to true?

dalf:
The safe way is to have additional process with limited access:
* capabilities reduced to nearly zero
* process priority at 19 (or equivalent for MacOS and other OS)
* perhaps sandboxing inside WASM

==> since Python is what it is, it can be an Flask servers.

Another approach: allow/implement satellite projects: if this is easy to create and use an additional repo, then a dedicate repo can can contain some unsafe plugins.

Currently, we have to put a warning, but how:
* if the plugin is disabled, someone can enable it and bam.
* if the plugin has to be copy to the plugin, nobody is going to do that.
* forbid on a public, but search for `site:Powered by searxng` and it is possible to find open access instances but not public.

> `sys.set_int_max_str_digits()`

It set the number of digits for the whole process ==> it should be done in a dedicated process for the calculator.

Bnyro:
As discussed, this can now only be invoked on public instances. In order to not display a preference for it on public instances I added a function `is_allowed` which determines if the plugin may or may not be used - if somebody has a better idea for an implementation I'm open for that.

PR #3378, opened 2024-04-07
[feat] plugins: new unit converter plugin

### LINKED ISSUE #3015: Add unit conversion like google.
no

Unit conversion like google when performing searched like "10 miles in km"

use google.com

**Additional context**
Currency conversion already works, however distance, weight, speed would be nice to have it implemented.

Bnyro:
## What does this PR do?
- implement a new unit converter plugin

- to not rely on 3rd party search engines for basic unit conversion, see #3015

- make run
- 5m to cm
- 5kg to g

## Related issues
closes #3015

Would be great if you could have a look at these changes @dalf, there might be a better and safer way to implement `_parse_text_and_convert()`.


searx/plugins/unit_converter.py
```diff
@@ -0,0 +1,76 @@
...
+        if unit['symbol'] == to_unit_key:
+            to_unit = unit
+
+        if from_unit and to_unit:
+            break
+
+    if from_unit is None or to_unit is None or to_unit.get('si_name') != from_unit.get('si_name'):
+        return
+
+    result = _convert(float(from_value), from_unit['to_si_factor'], to_unit['to_si_factor'])
```
dalf, line 54: [`Decimal`](https://docs.python.org/3/library/decimal.html) would be better because Decimal numbers can be represented exactly.

searx/plugins/unit_converter.py
```diff
@@ -0,0 +1,76 @@
...
+    search.result_container.answers['conversion'] = {'answer': f"{result:g} {to_unit['symbol']}"}
+
+
+def post_search(_request, search):
+    # only convert between units on the first page
+    if search.search_query.pageno > 1:
+        return True
+
+    query = search.search_query.query
+    query_parts = query.split(" ")
```
dalf, line 64: why not use `search.search_query.user_query_parts`?

especially looking `len(query_parts) < 3` below: a query like `:fr 2km in cm`.

searx/plugins/unit_converter.py
```diff
@@ -0,0 +1,76 @@
...
+    if len(splitted_query) != 2 or splitted_query[0].strip() == "" or splitted_query[1].strip() == "":
+        return
+
+    from_value = ""
+    from_unit_key = ""
+
+    # only parse digits as value that belong together
+    read_alpha = False
+    for c in splitted_query[0]:
+        if not read_alpha and (c in ("-", ".") or str.isdigit(c)):
```
dalf, line 30: Nitpicking: we don't deal with localization:
* `3.1415` <=> `3,1415`
* `1000km` <=> `1 000 km`

No need to do that now, the currency engine has the same issue.

searx/plugins/unit_converter.py
Bnyro: The `SearchQuery` class doesn't have `user_query_parts`, see https://github.com/searxng/searxng/blob/26a92c1a8d0420c0254653f40d107a877b882a17/searx/search/models.py#L27. Only `RawTextQuery` does, but that's not accessible within the plugin (except for creating a new instance of `RawTextQuery`).

searx/plugins/unit_converter.py
Bnyro: This makes `1m in cm` result in `99.99999999999999791833182883 cm`

dalf (approved):
Note: `42 F in C` does not work

searx/sxng_locales.py
```diff
@@ -56,6 +56,7 @@
     ('fr-CA', 'Français', 'Canada', 'French', '\U0001f1e8\U0001f1e6'),
     ('fr-CH', 'Français', 'Suisse', 'French', '\U0001f1e8\U0001f1ed'),
     ('fr-FR', 'Français', 'France', 'French', '\U0001f1eb\U0001f1f7'),
+    ('gl', 'Galego', '', 'Galician', '\U0001f310'),
```
return42: can you remove this hunk from the patch / I assume its a leftover when you updated the `searx/data/wikidata_units.json`

Bnyro: Oh yeah sorry, I didn't notice `make data.all` also modifies `sxng_locales.py`, I've only had reverted everything in `searxng/data`.

return42:
> Note: `42 F in C` does not work

I think this is due to the ambiguity of symbol `F` .. and `C` / `°C` & `F` / `°F`

However: `42 °F to °C` did also not work 

```json 
   "Q1396128": {
       "si_name": "Q25406",      -- Coulomb
       "symbol": "F",
       "to_si_factor": 96485.33212331001
   }
   "Q131255": {
       "si_name": "Q131255",     -- Farad
       "symbol": "F",
       "to_si_factor": 1.0
   }
   "Q25406": {
       "si_name": "Q25406",      -- Coulomb
       "symbol": "C",
       "to_si_factor": 1.0
   },
   "Q69362731": {
      "si_name": "Q69363953",    -- kelvin difference
      "symbol": "°C",
       "to_si_factor": 1.0
   },
   "Q42289": {
       "si_name": null,
       "symbol": "°F",
       "to_si_factor": null
   },
```
----------

The issues are addressed by:

- https://github.com/searxng/searxng/pull/3435

searx/plugins/unit_converter.py
```diff
@@ -0,0 +1,76 @@
...
+            from_unit = unit
+
+        if unit['symbol'] == to_unit_key:
+            to_unit = unit
+
+        if from_unit and to_unit:
+            break
+
+    if from_unit is None or to_unit is None or to_unit.get('si_name') != from_unit.get('si_name'):
+        return
```
return42, line 52: IDK if possible but may we can improve unit detection where symbols are ambiguous or `si_name` is `null` .. see https://github.com/searxng/searxng/pull/3378#issuecomment-2080974863

----

- https://github.com/searxng/searxng/pull/3435

return42 (approved):
Thanks for unit conversion :+1: / I think this is a very good initial version .. I'm going to merge, may we can improve unit handling in a follow up PR ..

Bnyro:
Oh yes, we only filter for letters here (and no ° symbol): https://github.com/searxng/searxng/pull/3378/files#diff-1a8835e42547f530fd3a8267f9ba44b660670c9cd0b0a7b0c15bd77c53e9a43eR30

We can probably just allow any kind of symbols that are not numbers for the unit name.

return42:
Seems we have an issue in the CI:

- https://github.com/searxng/searxng/pull/3437#issuecomment-2082121730

PR #3384, opened 2024-04-08
[mod] engine 'mwmbl': Update API URL

haexwise:
## What does this PR do?

Update the API for the Mwmbl engine to the latest one from last year. Also remove a stale link to API documentation.
To my knowledge there is no new API documentation to replace the stale one. As there is no change in functionality this is not important.

The current URL is deprecated as per the TODO comment in  https://github.com/mwmbl/mwmbl/blob/main/mwmbl/urls.py and may be removed in the future.

### DIFF
--- a/searx/engines/mwmbl.py
+++ b/searx/engines/mwmbl.py
@@ -17,7 +17,6 @@ from urllib.parse import urlencode
 
 about = {
     "website": 'https://github.com/mwmbl/mwmbl',
-    "official_api_documentation": 'https://api.mwmbl.org/docs',
     "use_official_api": True,
     "require_api_key": False,
     "results": 'JSON',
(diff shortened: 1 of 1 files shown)


searx/engines/mwmbl.py
```diff
@@ -17,15 +17,14 @@
...
-    "official_api_documentation": 'https://api.mwmbl.org/docs',
     "use_official_api": True,
     "require_api_key": False,
     "results": 'JSON',
 }
 paging = False
 categories = ['general']
 
-api_url = "https://api.mwmbl.org"
+api_url = "https://api.mwmbl.org/api/v1/search/"
```
Bnyro: Please remove the trailing `/`. In line 31 there would now be two of them right next to wach other. Also, `search` should probably be removed here because it's already in line 31. Have you even tested this?

Bnyro (changes requested):
See above

haexwise:
I did test it with curl. The https://api.mwmbl.org/api/v1/search/?s=<query> is correct.

Bnyro:
Again, `search/` shouldn't be part of the api url.

It should only be appended in line 31.

haexwise:
> Again, `search/` shouldn't be part of the api url.
> 
> It should only be appended in line 31.

Better ? Note that /search?s= without the / in between does redirect to /search/.
I am not doing this that often so please excuse if my style/correctness are way off.

Bnyro (approved):
Lgtm :+1:

return42 (approved):

PR #3398, opened 2024-04-14
[feat] engine: implementation of LibreTranslate

### LINKED ISSUE #3387: Libtretranslate
https://libretranslate.com/

This is a translation engine like google translate or deepl

... and it is free and self hostable!

**Features of this engine**

Web site creates the api calls when typing in a request!

... the web page is self describing

It is - like all other translation engines "general"

**Additional context**

The engine is written in python.

Bnyro:
## What does this PR do?
- implement a new engine, LibreTranslate

- make run
- !lt en-de hello world

## Related issues
closes https://github.com/searxng/searxng/issues/3387

### DIFF
--- a/searx/engines/deepl.py
+++ b/searx/engines/deepl.py
@@ -13,7 +13,7 @@ about = {
 }
 
 engine_type = 'online_dictionary'
-categories = ['general']
+categories = ['general', 'translate']
 
 url = 'https://api-free.deepl.com/v2/translate'
 api_key = None
@@ -51,11 +51,6 @@ def response(resp):
 
     infobox += "</dl>"
 
-    results.append(
-        {
-            'infobox': 'Deepl',
-            'content': infobox,
-        }
-    )
+    results.append({'answer': infobox})
 
     return results
--- a/searx/engines/libretranslate.py
+++ b/searx/engines/libretranslate.py
@@ -1,0 +1,46 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""LibreTranslate (Free and Open Source Machine Translation API)"""
+
+import random
+from json import dumps
+
+about = {
+    "website": 'https://libretranslate.com',
+    "wikidata_id": None,
+    "official_api_documentation": 'https://libretranslate.com/docs/',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+engine_type = 'online_dictionary'
+categories = ['general', 'translate']
+
+base_url = "https://translate.terraprint.co"
+api_key = ''
+
+
+def request(_query, params):
+    request_url = random.choice(base_url) if isinstance(base_url, list) else base_url
+    params['url'] = f"{request_url}/translate"
+
+    args = {'source': params['from_lang'][1], 'target': params['to_lang'][1], 'q': params['query']}
+    if api_key:
+        args['api_key'] = api_key
+    params['data'] = dumps(args)
+
+    params['method'] = 'POST'
+    params['headers'] = {'Content-Type': 'application/json'}
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    json_resp = resp.json()
+    text = json_resp.get('translatedText')
+    if text:
+        results.append({'answer': text})
+
+    return results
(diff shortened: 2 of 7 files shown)


cy8aer:
Tested it and works like a charm.

searx/engines/deepl.py
```diff
@@ -13,7 +13,7 @@
 }
 
 engine_type = 'online_dictionary'
-categories = ['general']
+categories = ['general', 'translate']
```
return42, line 16: To call all translators in one, I added a second category `!translate de-en hallo`

searx/engines/libretranslate.py
```diff
@@ -31,13 +31,19 @@ def request(_query, params):
 
     params['method'] = 'POST'
     params['headers'] = {'Content-Type': 'application/json'}
+    params["raise_for_httperror"] = False
```
return42: On some queries I get sometimes HTTP exception . here is the traceback of a 502

```
  File "searxng/local/py3/lib/python3.12/site-packages/httpx/_models.py", line 749, in raise_for_status
httpx.HTTPStatusError: Server error '502 Proxy Error' for url 'https://trans.zillyhuhn.com/translate'
```

To suppress those issues we should handle HTTP issues in the response method ...

searx/engines/libretranslate.py
```diff
@@ -31,13 +31,19 @@ def request(_query, params):
...
 
     return params
 
 
 def response(resp):
-    json_resp = resp.json()
 
-    result = {'infobox': json_resp['translatedText'], 'content': ''}
+    results = []
+    if resp.ok:
```
return42: ... ignore http issues

searx/engines/libretranslate.py
```diff
@@ -31,13 +31,19 @@ def request(_query, params):
...
+        json_resp = resp.json()
+        text = json_resp.get('translatedText')
+        if text:
```
return42: Sometimes translation returns an empty string.

searx/engines/libretranslate.py
```diff
@@ -31,13 +31,19 @@ def request(_query, params):
...
+            # results.append({'infobox': text, 'content': ''})
+            results.append({'answer': text})
```
return42: Personally I prefer to have the translation in the *answers box* .. but I'm not sure if this is a really good idea / please try it out and rate it

return42 (commented):
I added a WIP commit with some suggestions .. see my comments below.

Feel free to take over what you think is good and drop what is not good ..

searx/engines/libretranslate.py
Bnyro: That's probably been because of some instance related issues at zillyhuhn.com, I don't think that it generally throws such exceptions when it's hosted properly.

searx/engines/libretranslate.py
Bnyro: Good point :+1:

searx/engines/libretranslate.py
Bnyro: Yes, I agree that this is a good idea. We should however do the same for the other translation engines :+1:

searx/engines/libretranslate.py
return42: Yes, we can drop this .. 

but if we know zillyhuhn.com  is not hosted properly, we may replace zillyhuhn.com  by a different [mirrow](https://translate.terraprint.co/) .. by [libretranslate.eownerdead.dedyn.io](https://libretranslate.eownerdead.dedyn.io/)?

Bnyro: That mirror doesn't seem to load at all for me :thinking: 

There probably are more people self-hosting public instances, they're just not all at that list.

return42: > That mirror doesn't seem to load at all for me

Oops I haven't tested :-o / sorry.

searx/engines/libretranslate.py
Bnyro: I've applied this for the engines that don't use HTML now - the ones using HTML display more detailed information that are like a dictionary, so they probably fit well in the infobox :+1:

return42 (approved):

PR #3401, opened 2024-04-17
[feat] engine: implementation of 4get

Austin-Olacsi:
## What does this PR do?

adds support for 4get

supporting this api means we can use any scrapers they have, including a few that are missing from searxng such as: mojeek news, findthatmeme, and yandex. other scrapers might have some improvements, for example 4get's youtube scraper shows the video descriptions.

you need to use an instance with api access and no captchas, i made this instance you can use it to test:
https://4get.lol

there is also a list of public instances at https://4get.ca/instances

to use this in searxng you must edit the settings.yml, change the base_url and remove the inactive: true.
then, you can use: make run

### DIFF
--- a/searx/engines/4get.py
+++ b/searx/engines/4get.py
@@ -1,0 +1,184 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=too-many-branches, invalid-name
+
+"""4get (web, images, videos, music, news)
+
+.. hint::
+   Make sure the name of the scraper you want to use is set correctly!
+   duckduckgo is ddg, findthatmeme is ftm, souncloud is sc, and youtube
+   is yt.
+"""
+
+import time
+from urllib.parse import urlencode, urlparse, parse_qs
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
+
+# Engine metadata
+about = {
+    "website": 'https://4get.ca/',
+    "official_api_documentation": 'https://4get.ca/api.txt',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+# Engine configuration
+paging = True
+base_url: list = []
+scraper: list = []
+search_type: list = []
+safesearch = True
+time_range_support = True
+safesearch_map = {0: 'yes', 1: 'maybe', 2: 'no'}
+
+
+def request(query, params):
+    key = params['engine_data'].get('npt')
+
+    query_params = {
+        "s": query,
+        "scraper": scraper,
+        "country": "any",
+        "nsfw": safesearch_map[params['safesearch']],
+        "lang": "any",
+    }
+
+    if params['time_range']:
+        date = (datetime.now() - relativedelta(**{f"{params['time_range']}s": 1})).strftime("%Y-%m-%d")
+        query_params["newer"] = date
+
+    params['url'] = f"{base_url}/api/v1/{search_type}?{urlencode(query_params)}"
+
+    if params['pageno'] > 1:
+        params['url'] += f"&npt={key}"
+
+    return params
+
+
+# Format the video duration
+def format_duration(duration):
+    seconds = int(duration)
+    length = time.gmtime(seconds)
+    if length.tm_hour:
+        return time.strftime("%H:%M:%S", length)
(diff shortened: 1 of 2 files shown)


unixfox:
I would prefer to implement the actual interesting engines in searxng instead of using another service for that.

For me I vote against this engine into searxng core.

On top of that this might harm public 4get instances because they would suddenly have a new wave of traffic due to searxng.

Austin-Olacsi:
this engine is inactive i don't think it would really harm the 4get instances. it might not even get used on any public searxng instances.

p.s. don't we have a searx engine? so that would be using a metasearch engine to search another metasearch engine. just like this.

unixfox:
searx engine is a relic of searx. it was not implemented by us, I don't think it's a good example for why we would like to have 4get.

anyway adding more traffic to the 4get instances is just one of the reasons I disagree to add this engine, it's not the only reason.

spectrapulse:
I don't see why querying other metasearch engines would be a good idea as that would just add yet another level of additional latency which I don't think is desirable.

And also limit the engine's control on how to deal with upstream results and ranking them locally and potential future extensions to upstream engines such as ddg definitions...

If we're going to add support for other upstream engines we should do that within searxng and not rely on yet another metasearch engine. Also what if 4get ends up becoming unmaintained? Then if we want to keep compatibility with the engines backing 4get we'll have to implement these engines anyways.

Austin-Olacsi:
it is fast and we have much slower engines than this in searxng already; the added latency doesn't really matter. 

it may not be ideal but i don't think it is bad to have support for 4get. 

we wouldn't have to maintain these engines ourselves for as long as 4get is still updated.

plus if 4get adds another search engine we could already use it in searxng without needing to write our own implementation of it.

unixfox:
> plus if 4get adds another search engine we could already use it in searxng without needing to write our own implementation of it.

that's not the goal of searxng. we rather implement the engine into searxng than rely on an external service to have it.

@return42 @mrpaulblack @dalf whats your opinion on this?

return42:
I had already looked at the sources some time ago and realized that we had already implemented most of the scrappers ourselves (in better quality).

> for example 4get's youtube scraper shows the video descriptions.

then we should learn from there and improve our code.

We also have other engines that basically don't have their own index and therefore deliver similar results to the originals.   Here with this metasearch engine, I feel the overlap with SearXNG is so great that I think it would be better if we implemented (or improve) our own engines.  This would also give us greater operational reliability and allow us to act directly in the event of problems // I don't really like the dependence on a third party for things that we can do (better) ourselves.

> @return42 @mrpaulblack @dalf whats your opinion on this?

Even if there are reasons to merge this PR, I would rather agree with @unixfox to not merge this PR.

Austin-Olacsi:
alright, due to the unpopularity of this PR i will close it

no hurt feelings 👍 i also have other stuff i want to make for searxng in the future.

return42:
@Austin-Olacsi Thank you for your understanding. If you would like other priorities, please let me know and I will try to respond more positively than I have done here. However, SearXNG needs contributions from the community and we are very grateful for your commitment :+1:

lolcatw:
@spectrapulse 
>Also what if 4get ends up becoming unmaintained?

Won't happen unless I die.

I agree with @unixfox 's decision here. Implementing 4get in SearxNG is quite hilarious.

spectrapulse:
> @spectrapulse
> 
> > Also what if 4get ends up becoming unmaintained?
> 
> Won't happen unless I die.
> 
> I agree with @unixfox 's decision here. Implementing 4get in SearxNG is quite hilarious.
A strong feel for commitment which I'm happy to hear lol

but tbf in my defense for as far as I've seen you seem to be and overwhelming majority of all commits in 4get and if you ever dropped the project or desired to stop working on it which I can't vouch for it's pretty much dead as you're kind of carrying your own project lol.

Additionally my other points still stand and yes implementing 4get would be hilarious and incredibly cursed but shouldn't be how we implement upstream engines nor justified by it.

Austin-Olacsi:
I just wanted to say here that recently i did a lot of work to implement these engines into searxng the "proper way" (without depending on another service) and as a result we have yandex and findthatmeme in searxng now.

PR #3413, opened 2024-04-21
[fix] add missing localizable (gettext) messages to searxng.msg

### LINKED ISSUE #3412: Missing localizable (gettext) messages from some files in translation workflow
I noticed that we use terms such as 

https://github.com/searxng/searxng/blob/f5bb64cca8a63a10c470b4f9bff647702b6de74c/searx/engines/wttr.py#L55

https://github.com/searxng/searxng/blob/f5bb64cca8a63a10c470b4f9bff647702b6de74c/searx/engines/wttr.py#L68

But for some reason they seem to be missing in our [translation workflow](https://docs.searxng.org/dev/translation.html) ... [search for "Temperature" on Weblate](https://translate.codeberg.org/projects/searxng/searxng/en/?q=temperature#search) ..

May we have a bug in our translation workflow ..

_Originally posted by @return42 in https://github.com/searxng/searxng/pull/3359#pullrequestreview-1985132187_
            
-----

I'm not sure, but it seems to me that pybabel doesn't find (all?) `gettext` passages in the engines?

https://github.com/searxng/searxng/blob/af3be2b12b3bfd03af319e0901a19121281c1d11/utils/lib_sxng_weblate.sh#L143-L147

https://github.com/searxng/searxng/blob/af3be2b12b3bfd03af319e0901a19121281c1d11/babel.cfg#L1-L6

return42:
To test this patch I used .. and checked the diff of the `messages.pot` file::

    $ ./manage pyenv.cmd pybabel extract -F babel.cfg \
              -o ./searx/translations/messages.pot searx/
    $ git diff ./searx/translations/messages.pot

----

[hint from @dalf](https://github.com/searxng/searxng/issues/3412#issuecomment-2068010674): f-string are not supported [1] but there is no error [2].

[1] python-babel/babel#594
[2] python-babel/babel#715

Closes: https://github.com/searxng/searxng/issues/3412

### DIFF
--- a/searx/engines/yummly.py
+++ b/searx/engines/yummly.py
@@ -68,7 +68,7 @@ def response(resp):
                 'title': result['display']['displayName'],
                 'content': content,
                 'img_src': img_src,
-                'metadata': f"{gettext('Language')}: {result['locale'].split('-')[0]}",
+                'metadata': gettext('Language') + f": {result['locale'].split('-')[0]}",
             }
         )
 
--- a/searx/searxng.msg
+++ b/searx/searxng.msg
@@ -12,6 +12,8 @@ __all__ = [
     'CATEGORY_GROUPS',
     'STYLE_NAMES',
     'BRAND_CUSTOM_LINKS',
+    'WEATHER_TERMS',
+    'SOCIAL_MEDIA_TERMS',
 ]
 
 CONSTANT_NAMES = {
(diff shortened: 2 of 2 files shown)


Bnyro:
Just for my understanding, we need to add all strings that should be translated and are part of an f-string manually in the future? Or not use f-strings at all for things that should be translated?

dalf:
From my understanding, we can't combine `gettext` and f-string: it won't be detected by babel

return42:
> Just for my understanding, 

I didn't want to believe it at first either, but it's actually true :-(

Bnyro (approved):

Issue #3358, opened 2024-03-28
Time range in bing engine not working


Mena489:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
b21aaa89075b325bd89d2f9f5dc82bbb66855d12

**How did you install SearXNG?**
just built the docker image using Dockerfile

**What happened?**
When choosing time range like last year and last month it gives results outside of the range

Mena489:
It's likely related to this bug
https://github.com/searxng/searxng/issues/2641

Mena489:
Edit : It works but gets irrelevant results and when try the same bing api url on my browser it gets accurate results

Mena489:
Edit 2 : it happens when using double qoutes in the query

return42:
Yeah, bing is a source of constant errors / I tagged some of them: https://github.com/searxng/searxng/labels/bing-engine

Personally I gave up to fix this engine, we will probably never find a stable solution for this, because bing itself is already full of quirks https://github.com/searxng/searxng/issues/2641#issuecomment-1742526819 (Seems bing does not work in a VPN).

return42:
strong related:

- https://github.com/searxng/searxng/issues/3402

- [information on the parameters](https://github.com/searxng/searxng/pull/3416#issuecomment-2344841069)

PR #3416, opened 2024-04-24
fix bing returning same results, page numbering, minor refactor

### LINKED ISSUE #3402: bing engine paging always return same result
**What happened?**
only use bing engine, paging not working.

**How To Reproduce**
only use bing engine and search query, then paging, the result is always same.

**Expected behavior**
return different results.

**Screenshots & Logs**
990cdebf-55b5-41e5-a738-1aad416aca5d

138a9bbb-b08c-4231-8774-c1d177bc0d04

glanham-jr:
## What does this PR do?
The  'sc' parameter, whatever it means, needs to be present in order to not return the same results.

Bing page numbering doesn't increase by 10 each time. The first page returns 10 results, and all pages thereafter return 14 results. This means we need to update the page numbering to account for this. This also seems to be the case when running on searxng.

Finally, the code to check the page had some duplicate checks, so I refactored the code in this section which seems low-risk. I can undo this if we want a dedicated PR for this.

Fixes 3402

1. Run searxng
2. Use `!bing {...}`
3. Validate that each pages results are unique

## Author's checklist
n/a

## Related issues
Closes #3402 

### Page 1
image

### Page 2
image

### Page 3
image

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -72,16 +72,7 @@ base_url = 'https://www.bing.com/search'
 
 
 def _page_offset(pageno):
-    """
-    Page 1 => 1 (returns 10 results)
-    Page 2 => 11 (returns 14 results)
-    Page 3 => 25 (returns 14 results)
-    Page 4 => 39 (returns 14 results)
-    """
-    pageno_int = int(pageno)
-    if pageno_int <= 1:
-        return 1
-    return 11 + (pageno_int - 2) * 14
+    return (int(pageno) - 1) * 10 + 1
 
 
 def set_bing_cookies(params, engine_language, engine_region):
@@ -104,8 +95,14 @@ def request(query, params):
         # don't ask why it is only sometimes / its M$ and they have never been
         # deterministic ;)
         'pq': query,
-        # Unsure meaning of sc, but breaks bing if not present
+        # TODO: Figure out how below parameters are populated
         'sc': '0-0',
+		"sp": "-1",
+		"lq": "0",
+		"qs": "n",
+		"ghsh": "0",
+		"ghacc": "0",
+		"ghpl": "",
     }
 
     # To get correct page, arg first and this arg FORM is needed, the value PERE
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -72,7 +72,16 @@ base_url = 'https://www.bing.com/search'
 
 
 def _page_offset(pageno):
-    return (int(pageno) - 1) * 10 + 1
+    """
+    Page 1 => 1 (returns 10 results)
+    Page 2 => 11 (returns 14 results)
+    Page 3 => 25 (returns 14 results)
+    Page 4 => 39 (returns 14 results)
+    """
+    pageno_int = int(pageno)
+    if pageno_int <= 1:
+        return 1
+    return 11 + (pageno_int - 2) * 14
 
 
 def set_bing_cookies(params, engine_language, engine_region):
@@ -95,18 +104,19 @@ def request(query, params):
         # don't ask why it is only sometimes / its M$ and they have never been
         # deterministic ;)
         'pq': query,
+        # Unsure meaning of sc, but breaks bing if not present
+        'sc': '0-0',
     }
 
     # To get correct page, arg first and this arg FORM is needed, the value PERE
     # is on page 2, on page 3 its PERE1 and on page 4 its PERE2 .. and so forth.
     # The 'first' arg should never send on page 1.
-
     if page > 1:
         query_params['first'] = _page_offset(page)  # see also arg FORM
-    if page == 2:
-        query_params['FORM'] = 'PERE'
-    elif page > 2:
-        query_params['FORM'] = 'PERE%s' % (page - 2)
+        if page == 2:
+            query_params['FORM'] = 'PERE'
+        else:  # page > 2:
+            query_params['FORM'] = 'PERE%s' % (page - 2)
 
     params['url'] = f'{base_url}?{urlencode(query_params)}'
 


Bnyro (commented):
It's Bing time again :cry:

- The changes don't work for me unfortunately
- Page `start`s are `1`, `11`, `21`, `31`, ... for me, not as you changed it here
- The `sc` parameter is `11-3` for me - I tried changing it in SearXNG but pagination still doesn't work oddly

It pretty much seems like Bing behaves different a lot depending on your location (mine is Germany) or other circumstances - which makes it pretty hard to get things to work properly...

return42:
original posted in https://github.com/searxng/searxng/issues/3358#issuecomment-2026813958

> Yeah, bing is a source of constant errors / I tagged some of them: [bing-engine](https://github.com/searxng/searxng/issues?q=label%3Abing-engine)
> 
> Personally I gave up to fix this engine, we will probably never find a stable solution for this, because bing itself is already full of quirks [#2641 (comment)](https://github.com/searxng/searxng/issues/2641#issuecomment-1742526819) (Seems bing does not work in a VPN).

Anyway, I'm very happy to see that the community is addressing the issue ... the way the bing engine is currently running, any improvement (no matter how small) is welcome :+1:

glanham-jr:
> It's Bing time again :cry:
> 
> 
> 
> - The changes don't work for me unfortunately
> 
> - Page `start`s are `1`, `11`, `21`, `31`, ... for me, not as you changed it here
> 
> - The `sc` parameter is `11-3` for me - I tried changing it in SearXNG but pagination still doesn't work oddly
> 
> 
> 
> It pretty much seems like Bing behaves different a lot depending on your location (mine is Germany) or other circumstances - which makes it pretty hard to get things to work properly...

Hmmmmm, there is definitely some strange non deterministic behavior occurring then. I'll try setting my VPN to Germany to see what's working.

As for the SC, parameter, it seemed to vary in value, so I set it to a default which solved the issue for me.

glanham-jr:
@Bnyro Good news :partying_face:  I was able to reproduce the issue by tunneling requests through Germany via a VPN. So I can investigate at least Germany soon.

But now that I'm seeing issues depending on region.. I'm hoping I can make a patch that doesn't cause actual regressions. I'll try to test various different regions and see what sticks. 

My current plan of action is to...
1.  include all HTTP Parameters. If that fails then...
2. Start investigating differences of cookies between regions.

For anyone elses info, here is the current HTTP Parameters I'm investigating. 

```json
{
	"GET": {
		"scheme": "https",
		"host": "www.bing.com",
		"filename": "/search",
		"query": {
			"q": "{query}",
			"sp": "-1",
			"lq": "0",
			"pq": "",
			"sc": "0-0",
			"qs": "n",
			"sk": "",
			"cvid": "{random string 32 chars long}",
			"ghsh": "0",
			"ghacc": "0",
			"ghpl": "",
			"FPIG": "{random string 32 chars long}",
			"first": "{page offset}",
			"FORM": "{weird bing specific page offset}"
		},
	}
}
```
Seems like [this Stack Overflow answer](https://stackoverflow.com/questions/37389192/what-does-the-parameter-cvid-in-bing-search-engine-stands-for#60359248) may have some insight.

> In the Bing search context, cvid represents the JavaScript parameter ConversationId. Bing uses this key to identify your search result collection as its reply to your query, q. Similarly, pq is PartialQuery. These and other parameters may also apply to different kinds of searches, such as image or video searches.
>
> Next, qs is your query's SuggestionType, sc shows your SuggestionCount, and from the suggestion list (dropped down, if enabled), sp shows the SuggestionPosition you chose. In your case, you did not select a suggestion, so &sp=-1. Toward the end of your string, sk is the SkipValue, because you might skip through your result pages, first tells the issuer how many results belong on the first page, and I'll let you figure out what FORM means. ;) 

Lastly, while we aren't using the Bing API, there may be some information we can cross reference? Here is a link to the [REST Documentation](https://learn.microsoft.com/en-us/bing/search-apis/bing-custom-search/reference/endpoints).

---

Also, now the page numbering is back to 1, 11, 21, ..., so who knows whats happening there. I'm going to revert those changes and assume Bing was having a moment. I don't want to cause regressions either, so less is more in that respect. However, the page numbering may be a strange combination of these other parameters, so I'll dig into this as well.

Going to put this in draft - anyone is welcome to add their inputs or testing, but I'll need to do more extensive testing by region.

glanham-jr:
Given Bing seems to be causing region-specific issues now, we should probably add documentation (whether in code or somewhere else) with major regions to check when making changes to this engine. But let me validate this first with extensive testing.

return42:
@glanham-jr Thank you for your research, your https://github.com/searxng/searxng/pull/3416#issuecomment-2078541809 is very interesting :+1: 

> Going to put this in draft - anyone is welcome to add their inputs or testing, but I'll need to do more extensive testing by region.

May I have time next weekend / but to answer your question ...

> we should probably add documentation (whether in code or somewhere else) with major regions to check when making changes to this engine.

Lets document in comments and doc-strings in the engine .. the doc-strings are used here in the online documentation --> https://docs.searxng.org/dev/engines/online/bing.html

glanham-jr:
Tried testing this last night again. Germany is still having this issue despite supplying all HTTP Parameters. From what I saw, it appears to be an issue with the headers or the cookies? 

I copied the exact same URL for page 2 that searxng emitted (that returned page 1 results) and tried it on a regular browser, where I got actual page 2 results.

It seems there are some slightly different behaviors even from HTTP parameters / cookies / headers. I didn't take exact notes, but I can try documenting the deltas of what I am seeing between the US and Germany.

glanham-jr:
I'm also thinking this PR partially fixes the problem for the US region, so maybe let's acknowledge this PR is a partial patch and let's dedicate a new issue specifically for Germany.

0xhtml:
I dug through the JS with the debugger and was able to compile the following information on the parameters.

- sc: suggestion count
  e.g. `1-2`
  - first number: number of suggestions/autocompletions displayed to the user
  -  second number: number of characters of the partial query
- sp: suggestion position
  e.g. `-1`
  the index of the suggestion the user selected; -1 for no suggestion
- qs: suggestion type
  e.g. `n`
  the type of suggestion the user clicked on; `n` for no suggestion
- cvid: conversation id
  e.g. `F8805B579DEF4D868ACE97068D1DA696`
  ~likely for AI integration~ the existence of this parameter predates LLMs, so this must be some more generic user identification
- ghc: ghosting
  e.g. `1`
  ghosting is apparently a feature where a suggestion is shown directly in the search box while typing, which can be completed by pressing tab; the value in this param is some number in relation to ghosting
- ghsh: ghosting shown
  `0` or `1`
  whether ghosting was shown to the user
- ghacc: ghosting accepted
  `0` or `1`
  wheter ghosting was accepted/selected by the user
- ghpl: ghosting prefix length
  e.g. empty
  the amount of characters the user typed
- pq: partial query
  e.g. `abc`
  query w/o the suggestion if the user clicked on a suggestion
- lq: long query param
  `0` or `1`
  whether the query is long; if so the parameters pq, sc, qs, sk, cvid, ghsh, ghacc and ghpl are dropped
- sk: skip value
  e.g. empty

0xhtml:
some more

- FORM
  referes to the form on the page that was used; e.g. QBLH for the start page; PERE, PERE1-3 for the page numbers at the bottom of the page; or PORE for the next page button; as far as I can tell using PORE for all pages above 1 and QBLH for the initial search should be fine
  edit: to clarify; the number at the end of PERE is just to indicate which button the user clicked; if the user is on page 2 then page 1 would be PERE, page 3 would be PERE1 and page 4 would be PERE2; the number has nothing to do with the actual page number

- FPIG or IG
  FP is probably for first page; ig seems to be somehow connected to conversation id; seems to be generated serverside after searching for the first page and is then used to load all other pages

dsmith2001:
> I'm also thinking this PR partially fixes the problem for the US region, so maybe let's acknowledge this PR is a partial patch and let's dedicate a new issue specifically for Germany.

Would love to see those changes as an experimental setting you can enable. Bing hasn't returned any result on my US instance for a long time now.

glanham-jr:
@dsmith2001 I'll revisit this PR at some point in the future. I'm very curious about trying to tackle Bing and its many mysteries.

return42:
Not sure, do we really need the `_EDGE_CD` and `_EDGE_S` cookies any longer? [used here ...](https://github.com/search?q=repo%3Asearxng%2Fsearxng+set_bing_cookies&type=code)

https://github.com/searxng/searxng/blob/81aaca8f44f435f81585eea4f77a61d15749ae61/searx/engines/bing.py#L78-L82

https://github.com/searxng/searxng/blob/81aaca8f44f435f81585eea4f77a61d15749ae61/searx/engines/bing.py#L89

https://github.com/searxng/searxng/blob/81aaca8f44f435f81585eea4f77a61d15749ae61/searx/engines/bing_videos.py#L49

https://github.com/searxng/searxng/blob/81aaca8f44f435f81585eea4f77a61d15749ae61/searx/engines/bing_images.py#L57

https://github.com/searxng/searxng/blob/81aaca8f44f435f81585eea4f77a61d15749ae61/searx/engines/bing_news.py#L64

0xhtml:
I did some more digging, and it seems that the issue rather lies with the http implementation used. (At least for a German residential IP.) In firefox I can usually get the second page results simply with the `q` and `first` parameter. With searx from the same IP I can never get the second page, even when trying to send all correct parameters and all 20 or something cookies.

Some more stuff I found (region Germany):
- The `cvid` is omitted by extensions like uBlock Origin and everything works fine
- MS Edge generates the `cvid` locally and offline, meaning it could be generated by searx as well (Been reverse engineering MS Edge, but no results so far)
- the `FORM` parameter is passed as `form` (lowecase) when using the page buttons
- if no suggestion is used, the whole query is considered the partial query
- right now all queries are considered short queries
- w/o cookies the region and language is autodetected based on something, I get Chinese results for English queries sometimes

-----

Related:

- https://github.com/searxng/searxng/pull/4546#issuecomment-2754721344
- https://github.com/searxng/searxng/commit/772f1e86f8fe26e759e289dbc15fb2308e85588f

PR #3420, opened 2024-04-25
[feat] media.ccc.de: implement module with pagination and iframe

Bnyro:
## What does this PR do?
- implement media.ccc.de as a dedicated python module

- the CCC hosts some interesting talks about cyber security and the impact of CS on the society, partially in German and partially in English

- !c3tv linux


return42 (approved):
> media.ccc.de: implement module with pagination and iframe

Didn't aware we have so lovely hackish vids available under `!c3tv` :+1: --> I tend to enable this engine by default :)

PR #3430, opened 2024-04-26
[feat] engine: implementation of searchmysite

Austin-Olacsi:
## What does this PR do?

Adds searchmysite engine

it was requested on the old project searx: https://github.com/searx/searx/issues/2817
the content is mostly personal blogs that don't show up in the results of other search engines. 

make run
!sms android

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1563,6 +1563,20 @@ engines:
   #    engine: scanr_structures
   #    disabled: true
 
+  - name: searchmysite
+    engine: xpath
+    shortcut: sms
+    categories: general
+    paging: true
+    search_url: https://searchmysite.net/search/?q={query}&page={pageno}
+    results_xpath: //div[contains(@class,'search-result')]
+    url_xpath: .//a[contains(@class,'result-link')]/@href
+    title_xpath: .//span[contains(@class,'result-title-txt')]/text()
+    content_xpath: ./p[@id='result-hightlight']
+    disabled: true
+    about:
+      website: https://searchmysite.net
+
   - name: sepiasearch
     engine: sepiasearch
     shortcut: sep


return42 (approved):
thanks :+1:

searx/settings.yml
```diff
@@ -1563,6 +1563,19 @@ engines:
...
   #    disabled: true
 
+  - name: searchmysite
+    engine: xpath
+    shortcut: sms
+    categories: general
+    paging: true
+    search_url: https://searchmysite.net/search/?q={query}&page={pageno}
+    url_xpath: //p[@class='sms-p3']/a[@class='result-link']/text()
+    title_xpath: //span[@class='result-title-txt']/text()
```
```suggestion
    title_xpath: //span[contains(@class,'result-title-txt')]/text()
```
makes it more robust against changes at their side

I would do the same for the classes in the url and content xpath

Bnyro:
I think that it would be more robust to use a `results_xpath` (you'll find enough other engines in SearXNG for reference) - it just lowers chances of getting mixed results (like the url from one result but the title of another)

searx/settings.yml
```diff
@@ -1563,6 +1563,19 @@ engines:
...
+    content_xpath: //*[@id='result-hightlight']
```
Bnyro: Why do you use a wildcard selector here?

Bnyro:
> I think that it would be more robust to use a `results_xpath` (you'll find enough other engines in SearXNG for reference) - it just lowers chances of getting mixed results (like the url from one result but the title of another)
>
@Austin-Olacsi I've quickly pushed my suggestions, you can find these changes here: https://github.com/searxng/searxng/compare/07bdea1d95d8c87c9c4959c4a83ab4febb54a5a6..d7693e65d271c9a4f1b8c64b0f0e4df9faa55952.

Bnyro (approved):
Looks good to me, thanks for the PR :+1:

PR #3435, opened 2024-04-28
[mod] improve unit converter plugin

return42:
- l10n support: parse and format decimal numbers by babel
- ability to add additional units
- improved unit detection (symbols are not unique)
- support for alias units (0,010C to F --> 32,018 °F)

### DIFF
--- a/searx/plugins/unit_converter.py
+++ b/searx/plugins/unit_converter.py
@@ -1,10 +1,29 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Calculate mathematical expressions using ack#eval
+"""A plugin for converting measured values from one unit to another unit (a
+unit converter).
+
+The plugin looks up the symbols (given in the query term) in a list of
+converters, each converter is one item in the list (compare
+:py:obj:`ADDITIONAL_UNITS`).  If the symbols are ambiguous, the matching units
+of measurement are evaluated.  The weighting in the evaluation results from the
+sorting of the :py:obj:`list of unit converters<symbol_to_si>`.
+
+Enable in ``settings.yml``:
+
+.. code:: yaml
+
+  enabled_plugins:
+    ..
+    - 'Unit converter plugin'
+
 """
 
-from flask_babel import gettext
+import re
+import babel.numbers
+from flask_babel import gettext, get_locale
 
-from searx.data import WIKIDATA_UNITS
+from searx import data
+
 
 name = "Unit converter plugin"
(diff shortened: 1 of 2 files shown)


searx/plugins/unit_converter.py
```diff
@@ -2,57 +2,167 @@
...
+            alias_items.append(
+                (
+                    alias,
+                    item[1],
+                    item[2],  # from_si
+                    item[3],  # to_si
+                    item[0],  # origin unit
+                )
+            )
+    symbol_to_si = symbol_to_si + alias_items
```
Bnyro: It would make sense to move this to an `init` function so that it doesn't run each time you search something often - don't plugins have something like that?

Or we could move it to the `fetch_wikidata_units` function in `searx.data`.

Bnyro:
Thanks for this followup PR @return42, works as I expect it to from my test :+1:

searx/plugins/unit_converter.py
return42: > don't plugins have something like that?

we can define a `def init(..)` in the plugin module:

https://github.com/searxng/searxng/blob/c8d0b6529b616aa33678afa10edc414b460ea017/searx/plugins/__init__.py#L177-L181

but `plugin.init(*init_args)` is executed when the app is initialized:

https://github.com/searxng/searxng/blob/c8d0b6529b616aa33678afa10edc414b460ea017/searx/plugins/__init__.py#L228-L230

> Or we could move it to the `fetch_wikidata_units` function in `searx.data`.

Yes, makes more sense / especially if we will have lazy evaluation for it (#3443).

return42: > Or we could move it to the `fetch_wikidata_units` function in `searx.data`.

FYI last force push was just a rebase to get https://github.com/searxng/searxng/pull/3445

return42: > Or we could move it to the `fetch_wikidata_units` function in `searx.data`.

I forgot we have functions which can't stored in a JSON string:

```python
ADDITIONAL_UNITS = [
    {
        "to_si": lambda val: val + 273.15,
        "from_si": lambda val: val - 273.15,
        ...
```

I modified the code to generate the list only once and store it in a global / BTW I added documentation with more details to the module.

return42: is PR ready to merge?

Bnyro: Yes :+1:

Bnyro (approved):

PR #3443, opened 2024-04-29
[mod] lower memory footprint by lazy loading JSON data

return42:
This patch implements lazy loading of the JSON data.

Motivation: in most requests not all JSON data is needed, but loaded.  By example these four JSON files:

- currencies.json ~550KB
- engine_descriptions.json ~1,3MB
- external_bangs.json ~1,3MB
- osm_keys_tags.json ~ 2,2MB

most often not used and consume a lot of memory and BTW they also extend the time required to instantiate a walker.

### DIFF
--- a/searx/data/__init__.py
+++ b/searx/data/__init__.py
@@ -20,13 +20,20 @@ __all__ = [
 
 import json
 from pathlib import Path
+from searx import logger
 
 data_dir = Path(__file__).parent
+logger = logger.getChild('data')
 
-
-def _load(filename):
-    with open(data_dir / filename, encoding='utf-8') as f:
-        return json.load(f)
+CURRENCIES: dict
+USER_AGENTS: dict
+EXTERNAL_URLS: dict
+WIKIDATA_UNITS: dict
+EXTERNAL_BANGS: dict
+OSM_KEYS_TAGS: dict
+ENGINE_DESCRIPTIONS: dict
+ENGINE_TRAITS: dict
+LOCALES: dict
 
 
 def ahmia_blacklist_loader():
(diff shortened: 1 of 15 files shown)


Bnyro (approved):
Works as expected from testing :+1:

dalf:
There are two points here:
* speed up the start of app
* lower the memory footprint

The speed up is clear.

The memory footprint is different: an long running instance will have the same memory footprint as now. It requires one request on:
* the currency engine
* one on OSM engine
* one on ddg definitions info or unit conversion
* on engines tab of the preferences pages (okay, the mouse has to be over an engine name)

==> it won't reduce the memory footprint of darmarit.org/searx/ , paulgo.io, searx.be for example (according the stats)

IMO, the solution is sqlite: https://github.com/searxng/searxng/discussions/2633

return42:
> The memory footprint is different: an long running instance

Yes, this is known .. I assumed uWSGI worker processes do not live long .. but TBH I don't really know how many request processed before a new process is spawned --> [`max-requests`](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#max-requests)

dalf:
As far I understand, this require a change of the uwsgi configuration in all instances to make this PR relevant.

return42:
> As far I understand, this require a change of the uwsgi configuration in all instances to make this PR relevant.

I would have expected the processes to be restarted regularly ... I would be surprised if the processes should live indefinitely and process millions of requests (memory leaks?) ... but I can't find any reasonable documentation either ... we could customize our wsgi config and add max-config next to lazy-apps

https://github.com/searxng/searxng/blob/dbed8da284f0e91bef0fb42606ae10241505b364/dockerfiles/uwsgi.ini#L21

https://github.com/searxng/searxng/blob/dbed8da284f0e91bef0fb42606ae10241505b364/utils/templates/etc/uwsgi/apps-available/searxng.ini#L41

dalf:
Same I can't find for sure the default of max_requests.
Even the source code is not clear for me:
https://github.com/unbit/uwsgi/blob/353b7dd19c9af762f3874ed46a604766e1d7c6d5/core/uwsgi.c#L279

When Paul has tested some PR on his instance, we could clearly see a memory leak over a week: the memory never dropped to the initial value. Same for my instance using Docker.

> we could customize our wsgi config

The docker image can be updated, but what about all the instances using the installation script or something else (like the Arch package) ?

return42:
> When Paul has tested some PR on his instance, we could clearly see a memory leak over a week: the memory never dropped to the initial value. Same for my instance using Docker.

Wow, never thought the defaults of uWSGI are not suitable for a productive environment / we can discuss here in this PR: 

Since you have now sent PR 

- https://github.com/searxng/searxng/pull/3458

we should focus on the SQL solution .. I change this PR to DRAFT.

----

> The docker image can be updated, but what about all the instances using the installation script or something else (like the Arch package) ?

Not related to this PR but in general we should not weight deployment questions over improvements of SearXNG core.

return42:
Superseded by https://github.com/searxng/searxng/pull/3458

dalf:
> Superseded by https://github.com/searxng/searxng/pull/3458

Doppelgänger of this PR merged: #4834

PR #3446, opened 2024-05-02
[fix] respect user preferences to open links in new tab for short answers

thunder-coding:
## What does this PR do?

*Same as in commit description*

Previously only result urls were set to open in new tab by default, this
should make the behaviour consistent.

Also adds the missing rel="noreferrer" to the anchor tag. Although this
should not be needed as long as the `referrer-policy: no-referrer`
header is set, it's always nice to play safer than to have to say sorry.
For example some reverse proxy configurations might strip off
unwhitelisted headers in which case it's nice to have this set.

*Pretty much explained in commit message and above as well

```bash
make run
```

## Author's checklist

- [X] It works

## Related issues

*None*

### DIFF
--- a/searx/templates/simple/results.html
+++ b/searx/templates/simple/results.html
@@ -24,8 +24,12 @@
         {%- for answer in answers.values() -%}
         <div class="answer">
         <span>{{ answer.answer }}</span>
-          {% if answer.url -%}
-          <a href="{{ answer.url }}" class="answer-url">{{ urlparse(answer.url).hostname }}</a>
+          {%- if answer.url -%}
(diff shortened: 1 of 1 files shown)


Bnyro (approved):

return42 (approved):
Thanks a lot for your fix :+1: / I modified your patch a little bit to get less whitespace in the output ([diff](https://github.com/searxng/searxng/compare/e7a9d1c914601284b2d0918ffb3698e163a27bb7..6f6f1ba7ad445aa65792467cae8007c0ecb0f888))

PR #3457, opened 2024-05-04
[fix] startpage engine: XPath expressions adapted for new HTML layout

return42:
Startpage has changed its HTML layout, classes like ``w-gl__result__main`` do no longer exists and the result items have been slightly changed in their structure.


Bnyro (approved):
Works as expected, thanks for quickly working on a fix :+1:

PR #3460, opened 2024-05-05
[mod] uWSGI config: configuring uwsgi for production

return42:
As stated in .. and other posts, the defaults of uWSGI not suitable for a productive environment.  To give just one example, the workers run indefinitely and the memory leaks aggregate.

- "Configuring uWSGI for Production: The defaults are all wrong" EuroPython 2019 [1]
- "Configuring uWSGI for Production Deployment" [2]
- ~"When Paul has tested some PR on his instance, we could clearly see a memory leak over a week: the memory never dropped to the initial value. Same for my instance using Docker." [3]~

[1] https://av.tib.eu/media/44810
[2] https://www.bloomberg.com/company/stories/configuring-uwsgi-production-deployment/
[3] https://github.com/searxng/searxng/pull/3443#issuecomment-2094347004

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -1,3 +1,4 @@
+# -*- mode: conf-unix; coding: utf-8  -*-
 [uwsgi]
 # Who will run the code
 uid = searxng
@@ -6,6 +7,13 @@ gid = searxng
 # Number of workers (usually CPU count)
 # default value: %k (= number of CPU core, see Dockerfile)
 workers = $(UWSGI_WORKERS)
+harakiri = 60
+# max-requests = 1000                  # Restart workers after this many requests
+# max-worker-lifetime = 3600           # Restart workers after this many seconds
+reload-on-rss = 4096                 # Restart workers after this much resident memory
+worker-reload-mercy = 60             # How long to wait before forcefully killing workers
+die-on-term = true                   # Shutdown when receiving SIGTERM (default is respawn)
+py-callos-afterfork = true           # allow workers to trap signals
 
 # Number of threads per worker
 # default value: 4 (see Dockerfile)
@@ -17,6 +25,9 @@ chmod-socket = 666
 # Plugin to use and interpreter config
 single-interpreter = true
 master = true
+strict = true
+vacuum = true                        # Delete sockets during shutdown
+need-app = true
 plugin = python3
 lazy-apps = true
 enable-threads = true
@@ -33,7 +44,8 @@ auto-procname = true
 
 # Disable request logging for privacy
 disable-logging = true
-log-5xx = true
+log-4xx = true                       # but log 4xx's anyway
+log-5xx = true                       # and 5xx's
 
 # Set the max size of a request (request-body excluded)
 buffer-size = 8192
--- a/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
+++ b/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
@@ -1,4 +1,4 @@
-# -*- mode: conf; coding: utf-8  -*-
+# -*- mode: conf-unix; coding: utf-8  -*-
 [uwsgi]
 
 # uWSGI core
@@ -24,6 +24,8 @@ env = SEARXNG_SETTINGS_PATH=${SEARXNG_SETTINGS_PATH}
 # disable logging for privacy
 logger = systemd
 disable-logging = true
+log-4xx = true                       # but log 4xx's anyway
+log-5xx = true                       # and 5xx's
 
 # The right granted on the created socket
 chmod-socket = 666
@@ -33,6 +35,9 @@ single-interpreter = true
 
 # enable master process
 master = true
+strict = true
+vacuum = true                        # Delete sockets during shutdown
+need-app = true
 
 # load apps in each worker instead of the master
 lazy-apps = true
@@ -50,6 +55,13 @@ enable-threads = true
 # Number of workers (usually CPU count)
 workers = ${UWSGI_WORKERS:-%k}
 threads = ${UWSGI_THREADS:-4}
+harakiri = 60
+# max-requests = 1000                  # Restart workers after this many requests
+# max-worker-lifetime = 3600           # Restart workers after this many seconds
+reload-on-rss = 4096                 # Restart workers after this much resident memory
+worker-reload-mercy = 60             # How long to wait before forcefully killing workers
+die-on-term = true                   # Shutdown when receiving SIGTERM (default is respawn)
+py-callos-afterfork = true           # allow workers to trap signals
 
 # plugin: python
 # --------------
--- a/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini:socket
+++ b/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini:socket
@@ -1,4 +1,4 @@
-# -*- mode: conf; coding: utf-8  -*-
+# -*- mode: conf-unix; coding: utf-8  -*-
 [uwsgi]
 
 # uWSGI core
@@ -24,6 +24,8 @@ env = SEARXNG_SETTINGS_PATH=${SEARXNG_SETTINGS_PATH}
 # disable logging for privacy
 logger = systemd
 disable-logging = true
+log-4xx = true                       # but log 4xx's anyway
+log-5xx = true                       # and 5xx's
 
 # The right granted on the created socket
 chmod-socket = 666
@@ -33,6 +35,9 @@ single-interpreter = true
 
 # enable master process
 master = true
+strict = true
+vacuum = true                        # Delete sockets during shutdown
+need-app = true
 
 # load apps in each worker instead of the master
 lazy-apps = true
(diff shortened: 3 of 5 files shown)


dockerfiles/uwsgi.ini
```diff
@@ -6,6 +7,13 @@ gid = searxng
 # Number of workers (usually CPU count)
 # default value: %k (= number of CPU core, see Dockerfile)
 workers = $(UWSGI_WORKERS)
+harakiri = 60
+max-requests = 1000                  # Restart workers after this many requests
+max-worker-lifetime = 3600           # Restart workers after this many seconds
```
inetol: There are several open issues about requests being lost because of workers respawning at the same time. AFAIK, an undocumented parameter was added some time ago that allows smoother restarts https://github.com/unbit/uwsgi/issues/2377#issuecomment-1002560594

dalf:
> "When Paul has tested some PR on his instance, we could clearly see a memory leak over a week: the memory never dropped to the initial value. Same for my instance using Docker."

For global communication and as discussed yesterday, there is no memory leak on the master branch, only in some PR. The Paul's public instance allowed to discover the memory leak and make sure to update the code before merge (like the searx.network PR...)

dockerfiles/uwsgi.ini
return42: @inetol thanks for pointing out, do you recommend to add "delta" and disable "strict"?


```ini
max-worker-lifetime-delta = 60
strict = false
```

return42:
> there is no memory leak on the master branch, only in some PR

Thanks for reminder .. I modified the initial comment :+1: 

Even if we do not have a memory leak, we should discuss this PR, as it generally makes sense to run WEB services in a controlled process environment (hardening).

You also pointed out in the meeting that we lose the statistics with every restart ... and that it would be better if we first looked for a solution in which we persist the statistics :+1:

dockerfiles/uwsgi.ini
inetol: ~..Or removing `max-worker-lifetime*` to avoid disabling `strict`, `max-requests` *should* prevent potential memory leakage issues on its own~

It seems that the "issue" with workers respawning at the same time also applies to `max-requests` which is worse because `max-requests-delta` was introduced for uWSGI 2.1 which means that this flag will never be available. Best approach is to add `max-worker-lifetime-delta`, disable `strict` and remove `max-requests` yup.

Also had issues with `reload-on-rss` making the workers reload almost 20 times per minute, perhaps raising the value to 4096 or even 8192 would give more room for bursts of requests.

dalf:
I wonder if the workers should really be restarted:
* the current configuration exist for more than a decade over multiple instances without issue
* we have made sure there is no memory leak very carefully until now.

If the worker are restarted, we won't notice the initial memory leaks: it will be under radar.
When we will notice an issue, it will be a mess to fix because it will be multiple memory leaks.

---
Slightly related: https://twitter.com/Carnage4Life/status/1556374093938626560

return42:
> I wonder if the workers should really be restarted:
> - If the worker are restarted, we won't notice the initial memory leaks: it will be under radar.

So our users are our test field?

> - the current configuration exist for more than a decade over multiple instances without issue

Do we know "without issue" .. in the past? .. and what does it forecast about the future? .. I have to restart uWSGI from time to time .. not sure why .. 

For me, shorten livetime of a process that serves public request is also some kind of hardening against attacks (its not only about memory).

--> I wonder why a WEB server should not be hardened

-----

> Slightly related: https://twitter.com/Carnage4Life/status/1556374093938626560

.. in this picture we let explode our users :)

dalf:
> So our users are our test field?

Nope, the Paul's instance is the lab under monitoring to test PR with potential memory issue.
The last change of the network took many weeks before merge.

Another example, Paul has deployed #3458 since Monday. We will see at the end of the week if the memory is still as low as on Monday. Then, Paul can deploy #3443 and see how long it takes for the workers to load all the .json. This can provide an insight on the worker lifespan using this PR.

But there is another point: a small memory won't be notice if the worker restart often. Actually, we can accumulate unnoticed memory leaks without seeing them. When it will starts to mater, it will be a mess to fix all the issues.

> For me, shorten livetime of a process that serves public request is also some kind of hardening against attacks (its not only about memory).

Can you provide an example? like a third-party taking control of the worker through a buffer overflow or something similar?

dalf:
Only max-requests and max-worker-lifetime are debated.

For example:

* `reload-on-rss = 2048` provides a safety net regarding the memory.
* `harakiri = 60` is useful especially with the calculator PR.
* `strict = true` might break existing configuration but it makes sense.
* `vacuum = true` seems good regarding the documentation

return42:
@inetol and @dalf I patched according to your remarks --> [diff](https://github.com/searxng/searxng/compare/9733cf4db6cb1545e6b2e3dbb33f10254e48d6b5..e28a69515ce3e0b28e5ebea5cf58a2dc75467313)

> It seems that the "issue" with workers respawning at the same time also applies to `max-requests` which is worse because `max-requests-delta` was introduced for uWSGI 2.1 which means that this flag will never be available. Best approach is to add `max-worker-lifetime-delta`, disable `strict` and remove `max-requests` yup.
> 

commented out `max-requests` & `max-worker-lifetime` .. and left `strict` enabled

> Also had issues with `reload-on-rss` making the workers reload almost 20 times per minute, perhaps raising the value to 4096 or even 8192 would give more room for bursts of requests.

set to 4096

> Only max-requests and max-worker-lifetime are debated.

both are now commented out

----

Not sure how to progress with this PR .. I tend to merge the PR to get more practical experience ..

inetol (approved):
Looks good to me

inetol:
Let me shed some light on this PR, after a few weeks testing and evaluating several settings against my instance, I *may* have found a sweet spot for the vast majority of environments.

This comparison is made from commit e28a695 ...

I have removed:
* harakiri: In slow instances this could trigger easily (also may be vulnerable to Slowloris).
* max-requests: Vulnerable to DoS attacks.
* py-callos-afterfork: There are many issues related to this feature.
* add-header: No longer needed, uWSGI by default does not enable keep-alive: https://uwsgi-docs.readthedocs.io/en/latest/HTTP.html#http-keep-alive
* static-expires: No header is applied to static files.

I have modified:
* strict: A weird error appears, "unknown config directive: settings-path".
* workers: Unless you have an instance with lots and lots of concurrent users this value is not going to have any performance improvement, it will double the amount of memory used (I'm not sure, but it seems to me that "botdetection" works better by keeping it at 1 worker maybe it is placebo).
* offload-threads: No more than 4 offload-threads should be necessary in any situation.

I have added:
* thunder-lock: Depends on each system (and uWSGI config) it might bring some performance gain: https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html
* max-worker-lifetime: I run my instance without this, but in most cases this option is advisable (replacement for "max-requests").

.ini config:
```ini
# -*- mode: conf-unix; coding: utf-8  -*-
[uwsgi]
uid = searxng
gid = searxng

# uWSGI
master = true
vacuum = true			# Delete sockets during shutdown
need-app = true
lazy-apps = true
die-on-term = true		# https://github.com/searxng/searxng/issues/3427
thunder-lock = true		# https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html
auto-procname = true		# Automatically set processes name to something meaningful
chmod-socket = 666
single-interpreter = true

# Module to import
plugin = python3
module = searx.webapp

# Workers
workers = 1			# Better increase the value of "threads"
max-worker-lifetime = 3600

# Threads (per worker)
threads = $(UWSGI_THREADS)
offload-threads = 4
enable-threads = true

# Virtual env and Python path
pythonpath = /usr/local/searxng/
chdir = /usr/local/searxng/searx/

# Logging
disable-logging = true
log-4xx = true			# Overrides "disable-logging" for 4xx HTTP codes
log-5xx = true			# Overrides "disable-logging" for 5xx HTTP codes

# Set the max size of a request (request-body excluded)
buffer-size = 8192

# Static files
static-map = /static=/usr/local/searxng/searx/static
static-gzip-all = true
```

return42:
> I have removed:
>   * ... 
>   * static-expires: No header is applied to static files.

:+1: .. yeah cache busting:
 
https://github.com/searxng/searxng/blob/150b2e21fdcc06492b7d78bbcb91fbcc037a7831/docs/admin/update-searxng.rst?plain=1#L64-L66

.. in its current state causes a lot of trouble (see [label:"clear browser cache"](https://github.com/search?q=repo%3Asearxng%2Fsearxng%20label%3A%22clear%20browser%20cache%22%20))

We removed it in PR:

- https://github.com/searxng/searxng/pull/4436

inetol:
As we are going to replace uWSGI soon, this is no longer relevant.

virtadpt:
You could always just do it short and sweet.  Here's my production uwsgi.ini file:

```
[uwsgi]
workers = 12
single-interpreter = true
master = true
plugin = python
lazy-apps = true
enable-threads = true
module = searx.webapp
virtualenv = /home/searx/searxng/env/
pythonpath = /home/searx/searxng/
chdir = /home/searx/searxng/searx/
http = 127.0.0.1:8888
disable-logging = true
route-run = fixpathinfo:
cache2 = name=searxcache,items=2000,blocks=2000,blocksize=4096,bitmap=1
```

PR #3463, opened 2024-05-05
[BREAKING CHANGE] hostname replace plugin rewrite

### LINKED ISSUE #1086: A feature that prioritizes certain websites for higher placement in search results
I love this search engine. But sometimes it is less accurate than other search engines.

It would be nice to have a feature (plugin) that prioritizes certain sites (such as stackoverflow) at the higher end of the search results.

### LINKED ISSUE #2619: [FEATURE REQUEST] external list file for hostname_replace

Bnyro:
## What does this PR do?
- refactors the hostname replace plugin to a `hostnames` plugin with additional features:
    - `high_priority` hostnames: these websites will get a higher score/ranking
    - `low_priority` hostnames: these websites will get a lower score/ranking
- basically an implementation of https://github.com/searxng/searxng/issues/1086#issuecomment-1097049011
- additionally this adds support for external list files as requested in #2619. Usage explanation: https://github.com/searxng/searxng/pull/3463/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR246-R254

- uncomment the `hostnames` configuration and enable the hostnames plugin
- search for `google`, `youtube`, `forgejo` (to see codeberg up in the results), ...

## Related issues

- closes #1086
- closes #2619


return42:
Some thoughts of mine:

- Do we need backward compatibility for a while? .. reading `hostname_replace` from settings.yml and prompt a deprecation message?

- When we are implementing a replacement for the old `hostname_replace` plugin, we should also implement a solution where the old fails:
  - https://github.com/searxng/searxng/issues/1348

Bnyro:
>     * Do we need backward compatibility for a while? .. reading `hostname_replace` from settings.yml and prompt a deprecation message?
> 
Yes, we should probably continue supporting the old `hostname_replace` format for now, I've actually had that in my initial draft of this but then removed it because I refactored some code. 
>     * When we are implementing a replacement for the old `hostname_replace` plugin, we should also implement a solution where the old fails:
>       
>       * [Links in the infobox are not modified by hostname_replace #1348](https://github.com/searxng/searxng/issues/1348)
>
If you know how, that can surely be part of this PR, but I think that could also be made later in an other PR.

Austin-Olacsi:
Are you interested in adding this to your PR?: https://github.com/searxng/searxng/pull/3051

return42:
> Yes, we should probably continue supporting the old `hostname_replace` format for now, I've actually had that in my initial draft of this but then removed it because I refactored some code.

Can you implement backward compatibility in this PR?

> > - [Links in the infobox are not modified by hostname_replace #1348](https://github.com/searxng/searxng/issues/1348)
> 
> If you know how, that can surely be part of this PR, but I think that could also be made later in an other PR.

After taking a closer look at the problem with the info box (["With the current architecture, the plugin must parse the infobox results."](https://github.com/searxng/searxng/issues/1348#issuecomment-1159484471)), I also think that this is not part of this PR.

Bnyro:
> Can you implement backward compatibility in this PR?
>
Yes, done now and tested :+1:

Bnyro:
Diff to test the fallback to the old hostname_replace plugin:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 57b3dba86..12654cd18 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -212,7 +212,8 @@ outgoing:
 
 # Comment or un-comment plugin to activate / deactivate by default.
 #
-# enabled_plugins:
+enabled_plugins:
+  - 'Hostnames plugin'
 #   # these plugins are enabled if nothing is configured ..
 #   - 'Hash plugin'
 #   - 'Self Information'
@@ -252,6 +253,15 @@ outgoing:
 # '(.*\.)?youtube\.com$': 'invidious.example.com'
 # '(.*\.)?youtu\.be$': 'invidious.example.com'
 #
+hostname_replace:
+  '(.*\.)?youtube\.com$': 'invidious.example.com'
+  '(.*\.)?youtu\.be$': 'invidious.example.com'
+  '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
+  '(.*\.)?reddit\.com$': 'teddit.example.com'
+  '(.*\.)?redd\.it$': 'teddit.example.com'
+  '(www\.)?twitter\.com$': 'nitter.example.com'
+  # to remove matching host names from result list, set value to false
+  '(.*\.)?google\.com$': false
 
 checker:
   # disable checker when in debug mode
```

Bnyro:
> Are you interested in adding this to your PR?: #3051
>
I think that this looks like a nice change as well, but I think that we should focus on the basic changes this PR tries to implement for now and look at such other improvements later.

searx/results.py
```diff
@@ -355,7 +355,7 @@ def close(self):
 
         for result in self._merged_results:
             score = result_score(result)
-            result['score'] = score
+            result['score'] = result.get('score', 0) + score
```
return42: The score is calculated using the `weight` factor of the engine, I'm not sure if an constant value subtracted (or added) from (to) the score is the best choice.

The factor `weight` has an exponential effect, e.g. if several engines have this result item ...

https://github.com/searxng/searxng/blob/75e4b65127668a1d3bfe559ecca38687b6433a7c/searx/results.py#L133-L142

The deduction of a constant from the score only has a linear effect ... and with increasing `weight` the effect of this constant on the result becomes less and less.

But since the calculation is encapsulated behind `low_priority` and `high_priority` we can also think about the calculation and weighting in another PR / for me this solution would be acceptable for now.

return42 (changes requested):
FYI I rebased an placed a WIP commit to improve the scoring (as I mentioned in https://github.com/searxng/searxng/pull/3463#pullrequestreview-2080957408) .. can you please have a look at the WIP commit .. if you think its OK please squash accepted modifications into your commits.

----

> Diff to test the fallback to the old hostname_replace plugin:

Thanks for the patch but I think its not really downward compatible since the old name in the existing setups is `'Hostname replace'` ..  I don't have a clue right now how we can solve this issue .. ideas?

> ```diff
>  #
> -# enabled_plugins:
> +enabled_plugins:
> +  - 'Hostnames plugin'
> ```

return42:
I don't know why this wasn't noticed before, but the pylint has found errors in one of your commits ... maybe it's because the rebase to the master now has a new pylint version in it.

searx/plugins/hostnames.py
```diff
@@ -0,0 +1,107 @@
...
+    if expressions:
+        return expressions
+
+    # fallback to the old `hostname_replace` settings format
+    hostname_replace_config = settings.get('hostname_replace', {})
+
+    if settings_key == 'replace':
+        return {re.compile(p): r for (p, r) in hostname_replace_config.items() if r != False}
+    else:
+        return {re.compile(p) for (p, r) in hostname_replace_config.items() if r == False}
```
Bnyro: Following pylint here and changing that to `if not r` will also remove things like
- `www\.google\.de`
instead of what it should look like with the old `hostnames_replace` plugin
- `www.\google\.de: False`
but I guess that's okay

Bnyro:
> FYI I rebased an placed a WIP commit to improve the scoring (as I mentioned in [#3463 (review)](https://github.com/searxng/searxng/pull/3463#pullrequestreview-2080957408)) .. can you please have a look at the WIP commit .. if you think its OK please squash accepted modifications into your commits.
> 
Done, thanks :+1:
> Thanks for the patch but I think its not really downward compatible since the old name in the existing setups is `'Hostname replace'` .. I don't have a clue right now how we can solve this issue .. ideas?
> 
Nope, not really. The easiest solution would probably be to keep the old name for the plugin - but it's doing more than just basic hostname replaces now, so I don't think it's a good idea. I'll think about if there's another way.

> I don't know why this wasn't noticed before, but the pylint has found errors in one of your commits ... maybe it's because the rebase to the master now has a new pylint version in it.
>
Done :+1: (all checks should pass now hopefully)

return42:
> I'll think about if there's another way.

I pushed a WIP commit .. not very sexy but I think it is sufficient for a transitional period.

Can you please check my WIP / if OK squash it into your commits .. additional to this PR we need an announcement .. 

------------

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index d5d311956..80b5aba22 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -212,14 +212,14 @@ outgoing:
 
 # Comment or un-comment plugin to activate / deactivate by default.
 #
-# enabled_plugins:
+enabled_plugins:
 #   # these plugins are enabled if nothing is configured ..
 #   - 'Hash plugin'
 #   - 'Self Information'
 #   - 'Tracker URL remover'
 #   - 'Ahmia blacklist'  # activation depends on outgoing.using_tor_proxy
 #   # these plugins are disabled if nothing is configured ..
-#   - 'Hostnames plugin'  # see 'hostnames' configuration below
+  - 'Hostname replace'  # see 'hostnames' configuration below
 #   - 'Calculator plugin'
 #   - 'Open Access DOI rewrite'
 #   - 'Tor check plugin'
@@ -228,6 +228,17 @@ outgoing:
 #   # preferences if they want.
 #   - 'Autodetect search language'
 
+hostname_replace:
+  '(.*\.)?youtube\.com$': 'invidious.example.com'
+  '(.*\.)?youtu\.be$': 'invidious.example.com'
+  '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
+  '(.*\.)?reddit\.com$': 'teddit.example.com'
+  '(.*\.)?redd\.it$': 'teddit.example.com'
+  '(www\.)?twitter\.com$': 'nitter.example.com'
+  # to remove matching host names from result list, set value to false
+  'spam\.example\.com': false
+
+
 # Configuration of the "Hostnames plugin":
 #
 # hostnames:
```

return42:
FYI: I added [4489388](https://github.com/searxng/searxng/pull/3463/commits/4489388c069a3501b360378d1b81d766ebbd2e07) to your commit series .. you can build the documentation in this commit by `make docs.live` and then visit http://0.0.0.0:8000/src/searx.plugins.hostnames.html

Bnyro:
> FYI: I added [4489388](https://github.com/searxng/searxng/pull/3463/commits/4489388c069a3501b360378d1b81d766ebbd2e07) to your commit series .. you can build the documentation in this commit by `make docs.live` and then visit http://0.0.0.0:8000/src/searx.plugins.hostnames.html
>
Great, thank you!

The documentation on this change looks very clear and simple to understand to me, so it should be easy for instance maintainers to follow it :+1:

Though I wondered if we should maybe also add that link `https://docs.searxng.org/src/searx.plugins.hostnames.html` to the `deprecated_msg` to make it easier for instance maintainers to find out where to start?

return42:
> Though I wondered if we should maybe also add that link `https://docs.searxng.org/src/searx.plugins.hostnames.html` to the `deprecated_msg` to make it easier for instance maintainers to find out where to start?

Good point .. done --> [diff](https://github.com/searxng/searxng/compare/1ba8b53d605e945fad2fdd0056d77e4ea799b043..a6b408d71956c5bd5326e7d944087e2147f7b235)

From my side PR is ready to merge ..

Bnyro:
Ready to merge from my side too :+1:

sjehuda:
Congratulations!

I want to add a note following my post at https://github.com/searxng/searxng/issues/2619#issuecomment-2060349202

While for my project I use proxies.toml which is only a statis list, I was thinking of writing a separate package for hostname switcher and using a config.py file which would be located in a configuration directory at user-level, similarly to project @getnikola

1) A package would mean that there would be more consistency and, better sharing of URLs and easier fashion to add new rules.
2) A config.py would allow customization of rules (not only change of hostname).

mrusme:
Pardon for resurrecting this, but I cannot seem to find an answer anywhere in the documentation. 

Is there a way to specify a yaml file for the priority options, similar to how `rewrite-hosts.yml` is specified in the example given here? If so, how would that yaml file look like?

Thank you kindly!

return42:
@mrusme you can find a full example in the docs --> https://docs.searxng.org/dev/plugins/hostnames.html

mrusme:
@return42 thank you, I had found and checked that page but it doesn't give an indication on whether a yaml file could be used for the priority fields unfortunately.

//edit: I went through the source code and found out that every of those settings is using [this function](https://github.com/searxng/searxng/pull/3463/files#diff-4198c1a8aa9a1d3cbff921013996b3c5ec70dcd3ce1550539f16e469d16e2620R80), which checks whether it's a list or a string, in which case it tries to load the (yaml) file located at the given path. This means that the documentation is simply lacking/not specific enough but the functionality is there.

PR #3468, opened 2024-05-07
[feat] gitea: implement dedicated module

Bnyro:
## What does this PR do?
- implement a module for gitea
- by default we only have a codeberg engine available (as before)

- it provides more details than before
- see #3268

- !cb foot
- !cb gadgetbridge

## Related issues
see #3268

### DIFF
--- a/searx/engines/gitea.py
+++ b/searx/engines/gitea.py
@@ -1,0 +1,109 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine to search in collaborative software platforms based on Gitea_.
+
+.. _Gitea: https://about.gitea.com/
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`base_url`
+
+Optional settings are:
+
+- :py:obj:`sort`
+- :py:obj:`order`
+- :py:obj:`page_size`
+
+.. code:: yaml
+
+  - name: gitea.com
+    engine: gitea
+    base_url: https://gitea.com
+    shortcut: gitea
+
+If you would like to use additional instances, just configure new engines in the
+:ref:`settings <settings engine>` and set the ``base_url``.
+
+
+Implementation
+==============
+
+"""
+
+from urllib.parse import urlencode
+from dateutil import parser
+
+about = {
+    "website": 'https://about.gitea.com',
+    "wikidata_id": None,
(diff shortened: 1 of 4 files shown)


searx/engines/gitea.py
```diff
@@ -0,0 +1,57 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Github (IT)
```
return42: C&P issue :)

searx/engines/gitea.py
```diff
@@ -0,0 +1,57 @@
...
+    "official_api_documentation": 'https://docs.gitea.com/next/development/api-usage',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = ['it', 'repos']
+paging = True
+
+base_url = "https://codeberg.org"
```
return42: Lets set default `base_url` to `None`  and add a [`init(..)`](https://docs.searxng.org/dev/engines/demo/demo_online.html#searx.engines.demo_online.init)  to check if base_url is configured in the engine settings.

return42 (changes requested):
> implement a module for gitea

yeah having a gitea module is much better: developer can now search their local gitea instance .. in best quality .. :rocket:

To make it more clear in the commit history, we should split the commit into two; in the first add the python module and in the second move codeberg from XPath engine to gitea engine.

Can you please split the commit? .. afterwards I can squash some more documentation into the gitea module / thanks a lot!

searx/settings.yml
```diff
@@ -800,26 +800,18 @@ engines:
...
-      official_api_documentation: https://try.gitea.io/api/swagger
-      use_official_api: false
-      require_api_key: false
-      results: JSON
+
+  - name: gitea.com
+    engine: gitea
+    base_url: https://gitea.com
+    shortcut: gitea
+    disabled: true
```
return42, line 814: FYI: I added gitea.com to the first commit.

return42 (approved):
FYI: I added the engine to the online docs / when merged into master available at: https://docs.searxng.org/dev/engines/online/gitea.html

Bnyro:
> FYI: I added the engine to the online docs / when merged into master available at: https://docs.searxng.org/dev/engines/online/gitea.html
>
Thanks, looks good to me :+1:

Issue #3428, opened 2024-04-26
Bug: yacy's used instance seem broken


TheFrenchGhosty:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2024.4.24+e0214412f

**How did you install SearXNG?**

Docker

**What happened?**

Yacy hasn't worked in a while on my instance and looks like it was never reported.
After looking into it, it looks like the used instance stopped showing any results (and is really slow to load the (empty) results page): 
![image](https://github.com/searxng/searxng/assets/47571719/d11c6a23-aadd-4040-a747-bb127a6832d8)

**Additional context**

Semi-private instance with dozens of users... (so IP ban is possible, Mojeek has done it recently)

**Technical report**

Error
  * Error: httpx.ReadTimeout
  * Percentage: 50
  * Parameters: `(None, None, 'yacy.searchlab.eu')`
  * File name: `searx/search/processors/online.py:116`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

return42:
Do you know alternatives for ..?

https://github.com/searxng/searxng/blob/0e09014df56d1d3b6011c8022f88790aa5cb37bf/searx/settings.yml#L2074

https://github.com/searxng/searxng/blob/0e09014df56d1d3b6011c8022f88790aa5cb37bf/searx/settings.yml#L2087

ping @Bnyro @pointhi

Bnyro:
There unfortunately is no list of public instances, see https://github.com/yacy/yacy_search_server/issues/326.

It seems to be possinle seeing the public instances when temporary self-hosting an instance (described in the linked issue). If anyone wants to do that and share the results here, that'd be helpful.

lolcatw:
The best way to pull results from YaCy is to host your own local instance and pulling results from the API, or re-implement the YaCy p2p protocol to handle it within Searx

Bnyro:
https://search.webproject.link/Network.html?page=2&maxCount=1000

ThunderHen:
search.webproject.link/
s.menghuqiangwei.com/
yacy.excloo.com/
yacy.applefritter.com/
search.open.us.org/
yacy/s.cd0.eu/
search.lomig.me/
yacy.ecosys.eu/
yacy.daj.pw/
yacy.p4u.xyz/
search.blume-it.com/
yayc.fxlqwq.link/
[...]

PR #3472, opened 2024-05-09
[mod] yacy engine: pick base_url randomly from a list of instances

return42:
Inspired by post [1] in the discussion we had, while yacy.searchlab.eu was broken.

[1] https://github.com/searxng/searxng/issues/3428#issuecomment-2101080101

### DIFF
--- a/searx/engines/yacy.py
+++ b/searx/engines/yacy.py
@@ -22,20 +22,26 @@ The engine has the following (additional) settings:
 - :py:obj:`search_mode`
 - :py:obj:`search_type`
 
+The :py:obj:`base_url` has to be set in the engine named `yacy` and is used by
+all yacy engines.
+
 .. code:: yaml
 
   - name: yacy
     engine: yacy
     categories: general
     search_type: text
-    base_url: https://yacy.searchlab.eu
     shortcut: ya
+    base_url:
+      - https://yacy.searchlab.eu
+      - https://search.lomig.me
+      - https://yacy.ecosys.eu
+      - https://search.webproject.link
 
   - name: yacy images
     engine: yacy
     categories: images
     search_type: image
-    base_url: https://yacy.searchlab.eu
     shortcut: yai
     disabled: true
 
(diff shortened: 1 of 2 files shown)


searx/settings.yml
```diff
@@ -2094,7 +2098,6 @@ engines:
     engine: yacy
     categories: images
     search_type: image
-    base_url: https://yacy.searchlab.eu
```
Bnyro, line 2097: Are we intentonally using only `https://yacy.searchlab.eu` (defined in the python module) for the images category by default?

Bnyro (approved):

searx/engines/yacy.py
```diff
@@ -22,20 +22,26 @@
 - :py:obj:`search_mode`
 - :py:obj:`search_type`
 
+The :py:obj:`base_url` has to be set in the engine named `yacy` and is used by
+all yacy engines.
+
```
return42, line 27: > Are we intentonally using only `https://yacy.searchlab.eu` (defined in the python module) for the images category by default?

Nope .. see doc above .. its implemented in `_base_url()` function from line 115 on.

Bnyro: Oh okay. But what happens if somebody wants to use different engines for general and image searches because one might be better than the other for images? (I've no idea if that makes sense though, I don't know how yacy works internally)

And if somebody renames the engines to something else than yacy, it won't work either?

return42: > Oh okay. But what happens if somebody wants to use different engines for general and image searches because one might be better than the other for images? (I've no idea if that makes sense though, I don't know how yacy works internally)

With current implementation its not possible .. but can be implemented in a future PR as a new feature.

> And if somebody renames the engines to something else than yacy, it won't work either?

No, it won't work anymore / the documentation is clear at this point "has to be set in the engine named `yacy`" .. I assume nobody will rename the yayc WEB engine, IDK a reason for.

In the end, it is a compromise to reduce the maintenance of settings.yml, albeit with little loss in flexibility / we do similar in the [piped configuration](https://docs.searxng.org/dev/engines/online/piped.html#configuration)

PR #3475, opened 2024-05-09
[fix] chefkoch: use German words in content field

Bnyro:
## What does this PR do?
- chefkoch engine: use German words in content field

- Chefkoch is a German engine, so it doesn't make sense to have English fields in the results

- !chefkoch torte


return42 (approved):
> Chefkoch is a German engine, so it doesn't make sense to have English fields in the results

.. and we do not need l10n (gettext) for it.

PR #3483, opened 2024-05-12
[mod] simple theme: drop img_src from default results

return42:
The use of `img_src` AND thumbnail in the default results makes no sense (only a thumbnail is needed).  In the current state this is rather confusing, because `img_src` is displayed like a thumbnail (small) and `thumbnail` is displayed like an image (large).


searx/static/themes/simple/src/less/style.less
```diff
@@ -281,24 +281,12 @@ article[data-vim-selected].category-social {
...
-      max-height: 7rem;
-      object-fit: scale-down;
-      object-position: right top;
-    }
+  img.thumbnail {
+    .ltr-float-left();
+    padding-top: 0.6rem;
+    .ltr-padding-right(1rem);
+    width: 7rem;
+    height: unset; // remove heigth value that was needed for lazy loading
```
dalf, line 289: I remember to have see very narrowed images with a long height.
Without `max-height: 7rem` (or similar), one result may "break" the whole page (break = the user has to scroll a lot to reach the next result).

`max-height: 7rem` implies `object-fit: scale-down;`. 
And `object-fit: scale-down;` implies `object-position`.

However, I don't remember why the value is `scale-down`, [`contains`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#contain) seems good too.

[EDIT] see my first comment on the PR.

searx/static/themes/simple/src/less/style.less
```diff
@@ -281,24 +281,12 @@ article[data-vim-selected].category-social {
...
-      .ltr-float-left();
-      padding-top: 0.6rem;
-      .ltr-padding-right(1rem);
-      width: 7rem;
-      max-height: 7rem;
-      object-fit: scale-down;
-      object-position: right top;
-    }
+  img.thumbnail {
+    .ltr-float-left();
```
dalf, line 285: This is slightly out of topic: more than once, it was asked to put the image of the right rather than of the left (at least for LTR languages). With images on the right, all the tittles are aligned on the left side --> it is easier to read them.

dalf:
I agree the names are confusing, however I don't think the `thumbnail` should be removed.
Video engines uses them to display thumbnails of the videos.

image

various API from the video engines use the term "thumbnails".

I'm not sure how to resolve this confusion.

searx/static/themes/simple/src/less/style.less
dalf: see my first comment on the PR --> Closing.

searx/static/themes/simple/src/less/style.less
return42: This hunk just removes the `img.image` CSS and leaves the `img.thumbnail` untouched .. and I tested all engines I modified in this PR without an issue.  When I tested I compared the old rendering with the new one .. rare, but if there was a difference, the new rendering was always better and more consistent in terms of layout.

Further as the existing comment already indicates, height specifications have had a negative effect on lazy load / which currently works with `img.thumbnail` (tested).  I would be reluctant to change anything about the rendering without seeing an example that illustrates that this rendering (which we've always had) is detrimental in some way.

----

> [EDIT] see my first comment on the PR.

It is not recognizable to me what the first comment was, this was the first comment that reached me via mail.

searx/static/themes/simple/src/less/style.less
```diff
@@ -394,6 +382,16 @@ article[data-vim-selected].category-social {
...
 
+.result-videos {
+  img.thumbnail {
+    .ltr-float-left();
+    padding-top: 0.6rem;
+    .ltr-padding-right(1rem);
+    width: 20rem;
+    height: unset; // remove heigth value that was needed for lazy loading
+  }
+}
```
return42, line 393: > various API from the video engines use the term "thumbnails".
> I'm not sure how to resolve this confusion.

The intention of this PR is to remove the confusion by removing `img.img_src` .. to show video thumbnail in the old size (as shown in your screenshot) this CSS was added for results of the `videos.html` template.  

Not sure, could it be that you mixed `img.img_src` and `img.thumbnail` .. sorry if my commit message was not clear, but this PR removes `img.img_src` from the results of the `default.html` template.

searx/static/themes/simple/src/less/style.less
return42: > This is slightly out of topic: more than once, it was asked to put the image of the right rather than of the left (at least for LTR languages). With images on the right, all the tittles are aligned on the left side --> it is easier to read them.

In the current layout (in current master) the thumbnails are on the left side (at least for LTR languages) / here is a screenshot from a `default.html` result, how it is rendered in branch master:

grafik

We can change the layout in a follow up PR / as you said "slightly out of topic" in this PR to change the layout (substantially).

searx/static/themes/simple/src/less/style.less
dalf: I clearly remember some thumbnails with an strange aspect ratio like 1:5. In this case, the layout is "broken" from the user point of view.
I haven't checked the lazy loading (it surprises me that an unknown height helps the lazy loading).
But anyway... okay.

searx/static/themes/simple/src/less/style.less
dalf: Ah sorry, I have not seen this one. It makes sense.

dalf:
I have not test the PR, but the whole thing makes sense.

searx/templates/simple/macros.html
```diff
@@ -25,8 +25,7 @@
   <span class="url_o{{loop.index}}"><span class="url_i{{loop.index}}">{{- part -}}</span></span>
   {%- endfor %}
   {{- result_close_link() -}}
-  {%- if result.img_src %}{{ result_open_link(result.url) }}<img class="image" src="{{ image_proxify(result.img_src) }}" title="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{ result_close_link() }}{% endif -%}
-  {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}" loading="lazy" width="200" height="200">{{ result_close_link() }}{% endif -%}
+  {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}" loading="lazy">{{ result_close_link() }}{% endif -%}
```
return42, line 28: > it surprises me that an unknown height helps the lazy loading

We had some fight with lazy load in the past (2021) --> https://github.com/searxng/searxng/pull/573

As far as I can remember it was a problem when height was unclear and the browser needed a value to calculate the layout .. may be browser have changed since .. in this PR I removed the fixed size of 200x200 in the template.

----

The details seem to be very complex and I don't know if the browser behavior has changed in the meantime:

To be sure, I just tested it again: in the current Firefox and Chrome the lazy load works, both for the results from the `image.html` and for those from the `default.html` template

Bnyro (approved):
Works as expected with all engines that I tested and I remember some confusion earlier already when we were discussing whether to use `thumbnail` or `img_src` for some engines, so I think that this PR is a very good idea :+1:

PR #3464, opened 2024-05-06
[mod] package.html template: additional links (a python dict)

dkuku:
## What does this PR do?
Refactor package template to include custom amount of links
Rename publishedDate variable to published_date
change engines that use this variable

image
image

PR #3487, opened 2024-05-14
[Feat] Add cargo package search

dkuku:
## What does this PR do?

Add cargo package search
this depends on changes from #3464 
image

### DIFF
--- a/searx/engines/crates.py
+++ b/searx/engines/crates.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Cargo search on crates.io"""
+
+from collections import OrderedDict
+from urllib.parse import urlencode
+
+from dateutil import parser
+
+about = {
+    "website": "https://crates.io/",
+    "wikidata_id": None,
+    "official_api_documentation": "https://crates.io/data-access",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["it", "packages", "cargo"]
+
+
+# engine dependent config
+paging = True
+page_size = 10
+search_url = "https://crates.io/api/v1/crates"
+
+linked_terms = OrderedDict(
+    [
+        ("homepage", "Project homepage"),
+        ("documentation", "Documentation"),
+        ("repository", "Source code"),
+    ]
+)
+
+
+def request(query: str, params):
+
+    args = urlencode({"page": params["pageno"], "q": query, "per_page": page_size})
+    params["url"] = f"{search_url}?{args}"
+    return params
+
+
+def response(resp):
+    results = []
+
+    for package in resp.json()["crates"]:
+
+        published_date = package.get("updated_at")
+        published_date = parser.parse(published_date)
+
+        links = {}
+        for k, v in linked_terms.items():
+            l = package.get(k)
+            if l:
+                links[v] = l
+
+        results.append(
+            {
+                "template": "packages.html",
+                "url": f'https://crates.io/crates/{package["name"]}',
+                "title": package["name"],
+                "package_name": package["name"],
+                "tags": package["keywords"],
+                "content": package["description"],
+                "version": package["newest_version"] or package["max_version"] or package["max_stable_version"],
+                "publishedDate": published_date,
+                "links": links,
+            }
+        )
+
+    return results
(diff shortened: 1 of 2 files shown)


searx/engines/cargo.py
```diff
@@ -0,0 +1,69 @@
...
+    docs_key = gettext("Documentation")
+    homepage_key = gettext("Project homepage")
+    source_code_key = gettext("Source code")
+    links = {}
+    if package.get("documentation"):
+        links[docs_key] = package["documentation"]
+    if package.get("homepage"):
+        links[homepage_key] = package["homepage"]
+    if package.get("repository"):
+        links[source_code_key] = package["repository"]
```
Bnyro:
```suggestion
    if "documentation" in package:
        links[docs_key] = package["documentation"]
    if "homepage" in package:
        links[homepage_key] = package["homepage"]
    if "repository" in package:
        links[source_code_key] = package["repository"]
```

searx/engines/cargo.py
```diff
@@ -0,0 +1,69 @@
...
+    if links:
+        return links
```
Bnyro: This should probably return `links`, no matter if it's empty or not (not sure how Python handles functions without return values)

searx/engines/cargo.py
```diff
@@ -0,0 +1,69 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""hex.pm"""
+
+from urllib.parse import urlencode
+from dateutil import parser
+from flask_babel import gettext
+
+
+about = {
+    # pylint: disable=line-too-long
```
Bnyro: nit: not needed here because the api docs line is short enough

searx/engines/cargo.py
```diff
@@ -0,0 +1,67 @@
...
+categories = ["it", "packages"]
+
+
+# engine dependent config
+paging = True
+search_url = "https://crates.io/api/v1/crates"
+
+
+def request(query: str, params):
+    args = urlencode({"page": params["pageno"], "q": query, "per_page": 20})
```
Bnyro: per_page could be made configurable in `settings.yml` by making it a global variable called `page_size`, but that's not mandatory

Bnyro (approved):

searx/engines/crates.py
```diff
@@ -0,0 +1,70 @@
...
+about = {
+    "website": "https://crates.io/",
+    "wikidata_id": None,
+    "official_api_documentation": "https://crates.io/data-access",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["it", "packages", "cargo"]
```
return42, line 18: You can further use `!cargo` .. when we implement other cargo reposetories we will add this engines also to group `cargo` .. so with `!cargo` we are able to search through all cargo managers in once.

searx/engines/crates.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+# engine dependent config
+paging = True
+page_size = 10
```
return42, line 23: As @Bnyro mentioned .. the page_size is configurable and I reduced from 20 to 10 .. 10 results per page is the default what we have in nearly all engines.

return42 (approved):
I renamed the module from `cargo.py` to `crates.io` .. "cargo" is the package manager while "crates.io" is just one package repository / may we want to implement for example https://lib.rs/ at one day.

Due to the rebase on

- https://github.com/searxng/searxng/pull/3464

and the rename to `crates.py` this PR has changed significantly .. can you please review again / thanks!

searx/settings.yml
```diff
@@ -924,6 +924,12 @@ engines:
     shortcut: hex
     disabled: true
 
+  - name: crates.io
+    engine: crates
+    shortcut: crates
+    disabled: true
+    timeout: 6.0
```
return42, line 931: Forgot to say: I had timeout issues on search terms like `!cargo test` or `!cargo lib` .. so I increased timeout from the default 3sec to 6sec .. not sure this is a good value.

dkuku:
Can this be merged? I don't have permissions.

return42:
> Can this be merged? I don't have permissions.

sure .. I was just waiting for the OK from you for my modifications :-)

PR #3490, opened 2024-05-16
[mod] sort hex packages by popularity

dkuku:
## What does this PR do?
Last modification to hex search:
Put popular packages on top of search result, instead of sorting alphabetically.
The search results are much more relevant.

image
image

### DIFF
--- a/searx/engines/hex.py
+++ b/searx/engines/hex.py
@@ -21,6 +21,9 @@ categories = ["it", "packages"]
 # engine dependent config
 paging = True
 search_url = "https://hex.pm/api/packages/"
+# Valid values: name inserted_at updated_at total_downloads recent_downloads
+sort_criteria = "recent_downloads"
+page_size = 10
 
 linked_terms = {
     # lower-case : replacement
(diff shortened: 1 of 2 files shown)


return42:
Please define a global variable `sort_criteria` with comment what values are valid:  

https://github.com/hexpm/hexpm/blob/25ab4dc62b51520df05585dba9626f01bea7c381/lib/hexpm_web/views/package_view.ex#L4-L10

dkuku:
I added a page_size param in hex https://github.com/hexpm/hexpm/blob/0ffc66feb3284ff8ad19b951571be285fb818a29/lib/hexpm_web/controllers/api/package_controller.ex#L17 and updated the pr to use it also.

searx/engines/hex.py
```diff
@@ -21,6 +21,9 @@
 # engine dependent config
 paging = True
 search_url = "https://hex.pm/api/packages/"
+# Valid values: name inserted_at updated_at total_downloads recent_downloads
+sort_criteria = "recent_downloads"
```
Bnyro, line 25: Please also add that option to the hex engine at `searx/settings.yml`, most instance maintainers probably only use the configuration file and options that are described in it, and don't read the engine implementation code.

dkuku: I copied the variables to settings.

Bnyro (approved):

return42 (approved):
Hi @dkuku, sorry for the delay .. FYI; I squashed your 5 commits into one and modified the commit message / thanks a lot for you contribution :+1:

PR #3498, opened 2024-05-19
[fix] fix elixir docs search

dkuku:
## What does this PR do?
The elixir-docs search currently does not work.
The documentation was [moved to hexdocs](https://github.com/elixir-lang/elixir/issues/12747)
This pr changes the behaviour:
`!!elixir-docs` searches directly in elixir documentation
`!!elixir` returns results from all the packages including elixir

### DIFF
--- a/searx/data/external_bangs.json
+++ b/searx/data/external_bangs.json
@@ -4888,8 +4888,8 @@
             "lix": {
                 "\u0010": "//www.elix-lsf.fr/spip.php?page=recherche_definitions&recherche=\u0002&lang=fr\u00010",
                 "ir": {
-                    "\u0010": "//hexdocs.pm/elixir/search.html?q=\u0002\u000164",
-                    "-docs": "//duckduckgo.com/?q=site%3Aelixir-lang.org%2Fdocs+\u0002\u00010",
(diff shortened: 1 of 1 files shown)


return42:
The [`!!<bang>` external bangs](https://docs.searxng.org/user/search-syntax.html#bang-external-bangs) are from ddg and update from the [`update-external-bangs.py`](https://docs.searxng.org/dev/searxng_extra/update.html#update-external-bangs-py) script.  This task is started every four weeks by the CI:

https://github.com/searxng/searxng/blob/ec41b535876c7906871ea25638837db3dbc5e4ab/.github/workflows/data-update.yml#L3-L4

This means we have to wait its fixed at [ddg bangs](https://duckduckgo.com/bangs) .. I have tried:

```
./manage pyenv.cmd python searxng_extra/update/update_external_bangs.py
```

.. but seems there are no modifications at ddg bangs since last sync from CI in PR #3324.

return42:
I tried at [ddg bangs](https://duckduckgo.com/bangs) ...

image

> `!!elixir-docs` searches directly in elixir documentation

seems broken / compare screenshot.

> `!!elixir` returns results from all the packages including elixir

nope, it searches in the documentation / compare screenshot .. I tried `!elixir` at ddg and redirects to https://hexdocs.pm/elixir/search.html?q=%s

PR #3501, opened 2024-05-20
[enh] Add engine for discourse.org forums

allendema:
## What does this PR do?
Add engine for forums based on [https://discourse.org](https://discourse.org).

A lot of forums can be searched using this engine.

[https://trends.builtwith.com/websitelist/Discourse](https://trends.builtwith.com/websitelist/Discourse)

Search for something (by default on [discuss.python.org](discuss.python.org)) ```!dpy lru_cache```
## Author's checklist

N/A
## Related issues

N/A

### DIFF
--- a/searx/engines/discourse.py
+++ b/searx/engines/discourse.py
@@ -1,0 +1,153 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+""".. sidebar:: info
+
+  - `builtwith.com Discourse <https://trends.builtwith.com/websitelist/Discourse>`_
+
+Discourse is an open source Internet forum system.  To search in a forum this
+engine offers some additional settings:
+
+- :py:obj:`base_url`
+- :py:obj:`api_order`
+- :py:obj:`search_endpoint`
+- :py:obj:`show_avatar`
+
+Example
+=======
+
+To search in your favorite Discourse forum, add a configuration like shown here
+for the ``paddling.com`` forum:
+
+.. code:: yaml
+
+   - name: paddling
+     engine: discourse
+     shortcut: paddle
+     base_url: 'https://forums.paddling.com/'
+     api_order: views
+     categories: ['social media', 'sports']
+     show_avatar: true
+
+
+Implementations
+===============
+
+"""
+
+from urllib.parse import urlencode
+from datetime import datetime, timedelta
+import html
+
+from dateutil import parser
+
+from flask_babel import gettext
+
+about = {
+    "website": "https://discourse.org/",
+    "wikidata_id": "Q15054354",
+    "official_api_documentation": "https://docs.discourse.org/",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url: str = None  # type: ignore
+"""URL of the Discourse forum."""
+
+search_endpoint = '/search.json'
+"""URL path of the `search endpoint`_.
+
+.. _search endpoint: https://docs.discourse.org/#tag/Search
+"""
+
+api_order = 'likes'
+"""Order method, valid values are: ``latest``, ``likes``, ``views``, ``latest_topic``"""
+
+show_avatar = False
+"""Show avatar of the user who send the post."""
+
+paging = True
+time_range_support = True
+
+AGO_TIMEDELTA = {
+    'day': timedelta(days=1),
+    'week': timedelta(days=7),
+    'month': timedelta(days=31),
+    'year': timedelta(days=365),
+}
+
+
+def request(query, params):
+
+    if len(query) <= 2:
+        return None
+
+    q = [query, f'order:{api_order}']
+    time_range = params.get('time_range')
+    if time_range:
+        after_date = datetime.now() - AGO_TIMEDELTA[time_range]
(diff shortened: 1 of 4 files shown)


searx/engines/discourse.py
```diff
@@ -0,0 +1,95 @@
...
+        }
+    )
+
+    params['url'] = base_url + search_endpoint + args
+    params['headers'].update(
+        {
+            'Accept': 'application/json, text/javascript, */*; q=0.01',
+            'X-Requested-With': 'XMLHttpRequest',
+        }
+    )
```
Bnyro:
```suggestion
    params['headers'] = {
        'Accept': 'application/json, text/javascript, */*; q=0.01',
        'X-Requested-With': 'XMLHttpRequest',
    )
```
params['headers'] should be empty by default, shouldn't it?

searx/engines/discourse.py
```diff
@@ -0,0 +1,95 @@
...
+        content = '@' + post.get('username', '')
+
+        if int(comments) > 1:
+            content += f' | Comments: {comments}'
+
+        if result.get('has_accepted_answer', ''):
+            content += ' | Answered'
+        else:
+            if int(comments) > 1:
+                content += ' | No Accepted Answer'
```
Bnyro:
```suggestion
        status = "Closed" if result.get('closed', '') else "Open"
        comments = result.get('posts_count', 1)

        url = f"{base_url}/t/{result['id']}"
        publishedDate = parser.parse(result['created_at'])

        content = ['@' + post.get('username', '')]

        if int(comments) > 1:
            content += f'Comments: {comments}'

        if result.get('has_accepted_answer', ''):
            content += 'Answered'
        elif int(comments) > 1:
            content += 'No Accepted Answer'
            
        content = " | ".join(conent)
```
(not better performance-wise, but much easier to read)

searx/engines/discourse.py
```diff
@@ -0,0 +1,95 @@
...
+
+        results.append(
+            {
+                'url': url,
+                'title': html.unescape(str(result['title']) + status),
```
Bnyro:
```suggestion
                'title': html.unescape(str(result['title']) + ' | ' + status),
```
(when following the suggestion above)

searx/engines/discourse.py
```diff
@@ -0,0 +1,95 @@
...
+                'content': html.unescape(content),
```
Bnyro: Why are we calling `html.unescape` here? All parameters that were added to `content` are self-created fields and not taken directly from the response?

searx/engines/discourse.py
allendema: This joined every char, so I now used a list to append.
Thanks.

searx/engines/discourse.py
allendema: Undone. :)

Bnyro (approved):

searx/settings.yml
```diff
@@ -1581,6 +1581,13 @@ engines:
     api_site: 'superuser'
     categories: [it, q&a]
 
+  - name: discusspython
+    engine: discourse
+    shortcut: dpy
+    base_url: 'https://discuss.python.org'
+    categories: [it, q&a]
```
return42, line 1588: I think we can better set these two values as defaults in the python module / can be modified in custom settings anyway .. but there should not be the need to set the defaults here in the settings.yml

@allendema can you patch it .. or if it is OK; I can patch it in my review ..

allendema: Maybe leave it so the admins can see an example of what is possible to do i.e use another forum like:
[https://caddy.community](https://caddy.community) or [https://discourse.pi-hole.net](https://discourse.pi-hole.net).

But you can patch it how you see fit.

searx/engines/discourse.py
```diff
@@ -0,0 +1,96 @@
...
+
+def response(resp):
+
+    results = []
+    json_data = resp.json()
+
+    if ('topics' or 'posts') not in json_data.keys():
+        return []
+
+    for post, result in zip(json_data['posts'], json_data['topics']):
```
return42: python's `zip` function won't work here .. does not what we expect:

- `topics` and `posts` list do not have the same number of items
- the sorting of both lists is not the same / in the example below, `posts[4]` refers to topic id found in `topics[1]`

.. I will fix it in my review / give me some time, its in work ..

image

return42 (approved):
@allendema @Bnyro can you please review: I made some modifications mainly to

- fix issue reported [here](https://github.com/searxng/searxng/pull/3501#pullrequestreview-2082651373)
- add time-range support
- add online documentation
- separated commit for engine config (settings.yml) from commit with engine implementation.

Here is the [diff](https://github.com/searxng/searxng/compare/7f5146b0526c772901a1e83b847be528f4af207b..95fd8284c15cf321081f3070a01f883de49ed435) of my modification to what you have already been reviewed.

Sorry for the significant changes, but I think the engine benefits from the revision.

@allendema thank you for your contribution, it was a very interesting task for me / I didn't know Discourse yet.

searx/engines/discourse.py
allendema: Thats right. Good catch.

allendema:
@return42 Thanks for you fixes and enhancements. Tested it and everything works.

PR #3504, opened 2024-05-23
[enh] add re-usable func to filter text

allendema:
## What does this PR do?

Add a function to extract text; for example from a response.text.
From: [https://github.com/mikf/gallery-dl/gallery_dl/text.py#L129](https://github.com/mikf/gallery-dl/blob/ea434963aed0cfb9c9a6fd412c42631cf5091e70/gallery_dl/text.py#L129)

Fix updating engine traits exiting because of babel error.

Also remove a slash in vimeo engine because it was using two for requests.

When having to extract text every file had to create its own way to extract needed string with `.find()`. 

Now you can simply `from searx.utils import extr`.

- ~`python3 update_engine_traits.py`~ `make data.traits`
- `make format.python`
- `make run` --> `!yt !vimeo james acaster`

## Author's checklist

Check compat with: [gallery_dl/text.py#L1-L7](https://github.com/mikf/gallery-dl/blob/ea434963aed0cfb9c9a6fd412c42631cf5091e70/gallery_dl/text.py#L1-L7)

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -2,6 +2,9 @@
 """Utility functions for the engines
 
 """
+
+from __future__ import annotations
+
 import re
 import importlib
 import importlib.util
@@ -369,6 +372,35 @@ def convert_str_to_int(number_str: str) -> int:
     if number_str.isdigit():
         return int(number_str)
     return 0
+
+
+def extr(txt: str, begin: str, end: str, default: str = ""):
+    """Extract the string between ``begin`` and ``end`` from ``txt``
+
+    :param txt:     String to search in
+    :param begin:   First string to be searched for
+    :param end:     Second string to be searched for after ``begin``
+    :param default: Default value if one of ``begin`` or ``end`` is not
+                    found.  Defaults to an empty string.
+    :return: The string between the two search-strings ``begin`` and ``end``.
+             If at least one of ``begin`` or ``end`` is not found, the value of
+             ``default`` is returned.
+
+    Examples:
+      >>> extr("abcde", "a", "e")
+      "bcd"
+      >>> extr("abcde", "a", "z", deafult="nothing")
+      "nothing"
+
+    """
+
+    # From https://github.com/mikf/gallery-dl/blob/master/gallery_dl/text.py#L129
+
+    try:
+        first = txt.index(begin) + len(begin)
+        return txt[first : txt.index(end, first)]
+    except ValueError:
+        return default
 
 
 def int_or_zero(num: Union[List[str], str]) -> int:
--- a/searx/engines/brave.py
+++ b/searx/engines/brave.py
@@ -132,6 +132,7 @@ from lxml import html
 from searx import locales
 from searx.utils import (
     extract_text,
+    extr,
     eval_xpath,
     eval_xpath_list,
     eval_xpath_getindex,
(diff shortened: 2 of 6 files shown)


searx/engines/startpage.py
```diff
@@ -442,7 +442,12 @@ def fetch_traits(engine_traits: EngineTraits):
     for lang_code in filter(
         lambda lang_code: lang_code.find('_') == -1, babel.localedata.locale_identifiers()  # type: ignore
     ):
-        native_name = babel.Locale(lang_code).get_language_name().lower()  # type: ignore
+        try:
+            native_name = babel.Locale(lang_code).get_language_name().lower()  # type: ignore
+        except (TypeError, AttributeError) as e:
+            print(f"ERROR: {lang_code} is unknown by babel {e}")
+            continue
```
Bnyro: These changes probably make sense but they're not mentioned in your commit message and are probably better off in a different PR?

Bnyro:
It probably makes sense to re-use this with https://github.com/searxng/searxng/blob/master/searx/engines/brave.py#L257-L259 as well (maybe making the end string optional)?

searx/utils.py
```diff
@@ -371,6 +371,35 @@ def convert_str_to_int(number_str: str) -> int:
     return 0
 
 
+def extr(txt, begin, end, default=""):
```
Bnyro: I'm not really sure about an empty string per default, I think that `None` would be easier to handle / more convenient

searx/engines/startpage.py
return42: I think @Bnyro  is right / its in the fetch function and not related to this PR .. @allendema can you please split this hunk off into a separate PR / thanks :+1:

allendema: Had to test my changes to ddg, startpage and qwant but `update_engine_traits.py` would fail without this patch. Supposedly the github checks would fail then too. Thats why I did not create a seperate pull request.

The fix for the babel error is indeed not mentioned in the commit but the pull request itself.

searx/utils.py
allendema: `bool` of extracted will still be False i.e
```python
if not extracted:
    pass
```
If None is needed then pass it as default=None.

This is also tested at [https://github.com/mikf/gallery-dl/test/test_text.py#L208](https://github.com/mikf/gallery-dl/blob/ea434963aed0cfb9c9a6fd412c42631cf5091e70/test/test_text.py#L208)

allendema (commented):
See my replies/reasons below.

searx/engines/startpage.py
return42: I had to fix this issue in a dedicated PR / for the reasons see:

- https://github.com/searxng/searxng/pull/3525

I merged this PR #3525 into master and after, I rebased you branch on master.

searx/utils.py
return42: Python is a dynamically typing language, so I would also prefer the `None` data type as default.  This changes with typing.  As soon as a function has several return types, the problem of typing lies with the caller again:

image

I would therefore prefer an empty string in the case of typing.

Since the module utils.py is a “library”, all names in it should also be typed / I have now typed the function (and slightly adapted the doc-string) ... see [diff](https://github.com/searxng/searxng/compare/bda3561d61d0a9768a0d721ab8979bde28ee801b..af61b9b6b53db92d72cd268ea339860cc946f86a)

--------

Related comments:

- [Some practical hints about the usage of "type annotations":](https://github.com/searxng/searxng/pull/3330#issuecomment-1993824738)
- [dynamic versus static typing](https://github.com/searxng/searxng/pull/3330#issuecomment-1996659677)

return42 (approved):

return42:
> It probably makes sense to re-use this with https://github.com/searxng/searxng/blob/master/searx/engines/brave.py#L257-L259 as well (maybe making the end string optional)?

I successfully tested ([see diff from force push](https://github.com/searxng/searxng/compare/af61b9b6b53db92d72cd268ea339860cc946f86a..17716136c13058a1d9a68a639110d0599fd34e6c)):

```python
    datastr = extr(resp.text, "const data = ", ";\n").strip()
```

I tested `!brave.images !brave.videos !brave.news foo`

Bnyro (approved):

PR #3505, opened 2024-05-24
[feature] add engine google-2000 / a "disables AI support" google

return42:
AI Overviews has been rolling out to all US based searchers in the mid/end of May 2024.  AI Overviews will launch in more countries soon and be available to more than a billion users by the end of the year. [3]

By adding extra argument ``&udm=14`` the AI overviews [2] from google are disabled. Suggested by @cocoonkid in [1].

[1] https://github.com/searxng/searxng/discussions/3503
[2] https://tedium.co/2024/05/17/google-web-search-make-default/
[3] https://searchengineland.com/google-rolls-out-ai-overviews-in-us-with-more-countries-coming-soon-440418

----

This PR is a POC .. both engines (google & google-2000) are enabled by default to compare the results .. Since the rollout of the AI-Overviews is currently only for US users, we need testers with IPs from different countries (I only have IPs in DE)

While testing, I would remind what @unixfox mentioned in https://github.com/searxng/searxng/discussions/3503#discussioncomment-9531130 ..

> .. but the issue is that you *(google-2000)* won't get suggestions and answers anymore from google.

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -56,6 +56,8 @@ time_range_dict = {'day': 'd', 'week': 'w', 'month': 'm', 'year': 'y'}
 # Filter results. 0: None, 1: Moderate, 2: Strict
 filter_mapping = {0: 'off', 1: 'medium', 2: 'high'}
 
+extra_args = ""
+
 # specific xpath variables
 # ------------------------
 
@@ -292,6 +294,8 @@ def request(query, params):
             }
         )
     )
+    if extra_args:
+        query_url += extra_args
 
     if params['time_range'] in time_range_dict:
         query_url += '&' + urlencode({'tbs': 'qdr:' + time_range_dict[params['time_range']]})
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -820,6 +820,11 @@ engines:
     timeout: 4.0
     disabled: true
 
+  - name: google-2000
+    engine: google
+    shortcut: go2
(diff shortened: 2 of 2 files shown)


unixfox:
Thanks for the PR.

I was thinking about a way to enable or disable the usage of WEB category, not being able to add more "args". Mainly because one may mess up with this config and may render their Google engine unworkable.

On top of that, this may make it sounds like that we support other parameters from google, but we likely do not as we are relying on fixed xpath for gathering the results.

Finally, the parameter for activating the WEB category may evolve overtime, we are unsure if it's going to stay "udm=14". We know better than the user if the parameter for activating the WEB category has changed or not. Or could require more parameters in order to have it working.

return42:
@unixfox I fully agree with your words .. this PR -- this kind of implementation -- is just a POC (marked PR as DRAFT) .. I think we need testers with IPs from US / I have only IPs in DE [3].

And we will have to see how the global rollout affects other IPs and XPath expression ..

> as we are relying on fixed xpath for gathering the results.

> Finally, the parameter for activating the WEB category may evolve overtime, 

The final implementation will be different than in this PR ... the way it was implemented here in this PR was just a simple and is good enough to test and compare the result lists (and the answers as well as suggestions).

[3] https://searchengineland.com/google-rolls-out-ai-overviews-in-us-with-more-countries-coming-soon-440418

return42:
Due to lack of interest, I am now closing this DRAFT.

PR #3513, opened 2024-05-25
Adding one-click deploy link for RepoCloud to README.md

cosark:
## What does this PR do?

Adding link enabling community access to one-click deploy template for self-hosting on RepoCloud.io

Gives community a quick easy and way to deploy searxng with one click.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,11 +84,13 @@ Setup
 - Alternatively there are *up to date* `installation scripts`_.
 - For individual setup consult our detailed `Step by step`_ instructions.
 - To fine-tune your instance, take a look at the `Administrator documentation`_.
+- One-click deployment available via `RepoCloud`_.
 
 .. _Administrator documentation: https://docs.searxng.org/admin/index.html
 .. _Step by step: https://docs.searxng.org/admin/installation-searxng.html
 .. _installation scripts: https://docs.searxng.org/admin/installation-scripts.html
 .. _Docker image: https://github.com/searxng/searxng-docker
+.. _RepoCloud: https://repocloud.io/details/?app_id=53
 
 Translations
 ============


Bnyro:
I think it's a pretty nice idea to publish the project to different hosting platforms to make it easier to install :)

However, I'm not really sure if we want to advertise them the same way as the currently existing installation methods.

The install scripts and Docker image are maintained by the SearXNG team on GitHub with a publicly available source code and history of contributions.
From what I've seen, RepoCloud does not appear to be open source, and neither the service you made there is - which means we as SearXNG team can't verify the SearXNG deployment on RepoCloud works and that it's still using 100% of the source code here.

Don't get me wrong, I think we should mention such projects in the README because it's great if anyone with few experience can just oneclick install SearXNG in the environment they're used to - I just would add a dedicated section about community maintained ways of installation with an additional note that these are not maintained by the SearXNG team and not verified, hence used on the user's own risk.

That's only my two cents, I'm curious about the other team members' opinions :)

unixfox:
Yeah I'm against it. I would be ok if it was on a "community" page like invidious have: https://docs.invidious.io/applications/
But not on the official installation guide, it's like saying we officially manage this project which is not true.

return42:
> I think we should mention such projects in the README because it's great if anyone with few experience can just oneclick install SearXNG in the environment they're used to 

> I would be ok if it was on a "community" page like invidious have

I would suggest that we first collect a few known hosters and applications in a discussion thread and then add an "Applications" page to the documentation (as suggested by @unifox).

I will close this PR ... @cosark if you want, please start a discussion thread and ask for more applications and hosters .. I hope that's OK for you .. anyway, thank you very much for your contribution :+1:

Issue #2033, opened 2022-12-21
Increase in image size and use of resources


ngosang:
In recent PR the langdetect library was changes with fasttest for language detection. https://github.com/searxng/searxng/commit/9925a209503cc41e37a1c032548a5fd8fd8ea362 & https://github.com/searxng/searxng/commit/ed901ab18edddcfd566ac37ddb1b429969ccc2ad

I pretty sure that library is a bit better in many ways but it increases the image size because it uses several data scientist libraries. 
* 22 MB compressed: 62.26 MB => 84.94 MB https://hub.docker.com/r/searxng/searxng/tags
* 103 MB uncompressed: 168 MB => 271 MB

The memory usage of searxng is almost the same but I also notice more CPU usage in each search.

I believe that this increase in resources is not justified. It would be necessary to know how much better that model is. I have used langdetect for many years and it works very very well. Other libraries could also be evaluated.

dalf:
Here an experimental drop-in replacement without the numpy dependency:
* https://github.com/dalf/fasttext-predict 
* there are wheels for various architectures which avoid to deal with two packages fasttext-wheel and fasttext.
* published for now here: https://test.pypi.org/project/fasttext-predict/

jamesoff:
numpy seems to be hard to build on Raspbian (https://github.com/numpy/numpy/issues/21038) and the version being pulled in isn't on piwheels for Python 3.7, so I'm in favour of this. Instead of going down the rabbit hole of upgrading/building Python on my Pi, I just pulled out the `fasttext` changes enough for the install to complete and the service to work - for my personal instance I don't need search language detection since I'm about the only user, and I only search in one language ;)

Or, could this be expressed as an "extra" for pip so that those who don't need it don't have it pulled in? If that's acceptable, happy to look at the changes.

If anyone else ends up here looking for it, this was my hack (although I probably could have just `git revert`ed the two commits in the initial post:

```diff
diff --git a/requirements.txt b/requirements.txt
index 5c45fff6..dde4727a 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,3 @@ setproctitle==1.3.2
 redis==4.4.0
 markdown-it-py==2.1.0
 typing_extensions==4.4.0
-fasttext-wheel==0.9.2
diff --git a/searx/utils.py b/searx/utils.py
index 2157a4ce..df96106f 100644
--- a/searx/utils.py
+++ b/searx/utils.py
[...]

PR #508, opened 2021-11-16
[mod] Tools to install and maintain golang binaries & packages.

return42:
## What does this PR do?

[mod] Tools to install and maintain golang [1] binaries & packages.

[1] https://golang.org/doc/devel/release#policy

## How to test this PR locally?

```
./manage go.help
go.:
  ls        : list golang binary archives (stable)
  golang    : (re-) install golang binary in user's $HOME/local folder
  install   : install go package in user's $HOME/go-apps folder
  bash      : start bash interpreter with golang environment sourced
```

check the results for `os` and `arch`:

    ./manage go.os
    ./manage go.arch

A simple *helloworld* test with user 'my_user' :

    sudo -H adduser my_user
    ./manage go.golang go1.17.3 my_user
    ./manage go.install github.com/go-training/helloworld@latest my_user
    ./manage go.bash my_user
    $ helloword
    Hello World!!

Don't forget to remove 'my_user':  

    sudo -H deluser --remove-home my_user

If test above success try to [install](https://searxng.github.io/searxng/admin/installation.html) filtron and morty:

    sudo -H ./utils/filtron.sh install all
    sudo -H ./utils/morty.sh install all

## Author's checklist

I need some tester for the various `os` and `arch` (see above)

## Related issues

Closes: https://github.com/searxng/searxng/issues/507
Related: https://github.com/searxng/searxng/issues/53

PR #3515, opened 2024-05-26
Add raspbian distribution

lannic:
## What does this PR do?
Added feature : 
Use script for updating the raspbian distribution

Add raspbian distribution like debian or ubuntu
Can use the script  for the installation or update of the raspbian distribution

Update with cron the src git directory 
Install searxng in the raspbian with the script
Update the src directory

sudo -H ./utils/searxng.sh install all
sudo -H ./utils/searxng.sh instance update

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -943,7 +943,7 @@ nginx_distro_setup() {
     NGINX_APPS_AVAILABLE="/etc/nginx/default.apps-available"
 
     case $DIST_ID-$DIST_VERS in
-        ubuntu-*|debian-*)
+        ubuntu-*|debian-*|raspbian-*)
             NGINX_PACKAGES="nginx"
             NGINX_DEFAULT_SERVER=/etc/nginx/sites-available/default
             ;;
@@ -1101,7 +1101,7 @@ nginx_dissable_app() {
 apache_distro_setup() {
     # shellcheck disable=SC2034
     case $DIST_ID-$DIST_VERS in
-        ubuntu-*|debian-*)
+        ubuntu-*|debian-*|raspbian-*)
             # debian uses the /etc/apache2 path, while other distros use
             # the apache default at /etc/httpd
             APACHE_SITES_AVAILABLE="/etc/apache2/sites-available"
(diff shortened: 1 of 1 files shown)


utils/lib.sh
```diff
@@ -1686,7 +1686,7 @@ LXC_BASE_PACKAGES_fedora="bash git @development-tools python"
 LXC_BASE_PACKAGES_centos="bash git python3"
 
 case $DIST_ID in
-    ubuntu|debian) LXC_BASE_PACKAGES="${LXC_BASE_PACKAGES_debian}" ;;
+    ubuntu|debian|raspbian-*) LXC_BASE_PACKAGES="${LXC_BASE_PACKAGES_debian}" ;;
```
Bnyro, line 1689:
```suggestion
    ubuntu|debian|raspbian) LXC_BASE_PACKAGES="${LXC_BASE_PACKAGES_debian}" ;;
```

Bnyro: We're only inspecting $DIST_ID here, not its version.

lannic: Yes i missed that

return42:
Unfortunately, we cannot carry out any (automated) tests for raspbian, as we currently have no infrastructure for this.

Even if raspbian is based on debian, the packages are partly assembled differently; especially the uWSGI but also developer packages come to mind (the changes from this PR will most likely not be sufficient for all functions of the scripts) .. as long as we can't do any tests, there is always the danger that the distros will be affected by changes to the shell scripts as part of unwanted side effects.

Our long-term goal is to completely remove the installation scripts from the SearXNG core (I am already working on an alternative, but as I rarely have time for this, it may be a long time before I can present something to you).

When we have outsourced the installation one day, I can also imagine that we will support other, additional distributions.  Until then, I don't want to increase the complexity and the associated maintenance effort of the shell scripts here in the SearXNG core any further.

I know that this is not a satisfactory solution, but unfortunately I can't offer more at the moment ... so please understand if I close this PR.

-----

Related: 

- https://github.com/searxng/searxng/pull/508#issuecomment-971562552
- https://github.com/searxng/searxng/discussions/2897
- https://github.com/searxng/searxng/issues/2033#issuecomment-1363930129

PR #3517, opened 2024-05-27
[feat] engine: implementation of findthatmeme

Austin-Olacsi:
## What does this PR do?

adds findthatmeme engine

for the memes

make run
!ftm sonic

## Author's checklist

for reference: https://git.lolcat.ca/lolcat/4get/src/branch/master/scraper/ftm.php

### DIFF
--- a/searx/engines/findthatmeme.py
+++ b/searx/engines/findthatmeme.py
@@ -1,0 +1,54 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""FindThatMeme (Images)"""
+
+from json import dumps
+from datetime import datetime
+from searx.utils import humanize_bytes
+
+about = {
+    "website": 'https://findthatmeme.com',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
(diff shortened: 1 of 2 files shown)


searx/engines/findthatmeme.py
```diff
@@ -0,0 +1,54 @@
...
+    return params
+
+
+def response(resp):
+    search_res = resp.json()
+    results = []
+
+    for item in search_res:
+        img = 'https://findthatmeme.us-southeast-1.linodeobjects.com/' + item['image_path']
+        thumb = 'https://findthatmeme.us-southeast-1.linodeobjects.com/thumb/' + item.get('thumbnail', '')
```
Bnyro, line 39: This should be a variable maybe, so that people hosting it can use different mirrors (there probably is a european one too ...)?

Austin-Olacsi: This seems to be the only one... i just tried to use the website with a european IP and it still uses us-southeast

Bnyro (approved):

return42 (approved):
very funny results :rofl: .. works like a charm / thanks :+1:

PR #3519, opened 2024-05-29
[fix] Do not show DDG IP zero click

JeffAlyanak:
## What does this PR do?

Currently, the DDG zero click results are being displayed if the user search includes `IP`.

This will return the IP of the searxng server—not the user's IP—and conflicts with the `self_info` plugin.

For example, if the searxng server's IP is `1.2.3.4` and the user's ip is `4.3.2.1`—and if the `self_info` plugin is enabled—the results will look like this:

Screenshot_20240529_004145

_Note: the full zero click answer from DDG includes a link to the user's location as well, but this gets dropped by searxng:_

Screenshot_20240529_010102

This issue affects end-user usability, but could also represent a minor security issue as it will result in the searxng server IP being leaked to the user. Instances behind a proxy that are not proxying their outbound traffic may not want their server's IP to be easily obtained this way.

This is a minor UI fix, it was tested visually in the browser:
Screenshot_20240529_004302

## Author's checklist

I have not been able to find an exhaustive list of zero click result types. If IP is the only one which needs to be excluded then this simple fix is probably sufficient, but it's possible there are others that would need to be excluded.

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -331,7 +331,7 @@ def response(resp):
     zero_click_info_xpath = '//html/body/form/div/table[2]/tr[2]/td/text()'
     zero_click = extract_text(eval_xpath(doc, zero_click_info_xpath)).strip()
 
-    if zero_click:
+    if zero_click and "Your IP address is" not in zero_click:
         current_query = resp.search_params["data"].get("q")
 
         results.append(


return42 (approved):
Truly a good point, thanks for the patch :+1: .. I slightly modified your commit to place both conditions in just `if` statement / see [diff](https://github.com/searxng/searxng/compare/c15cfe03b5edfafabc0c01348f8ed47cf16b6c5b..cb29f8f7b606b8f88c2591ad60bf6f77be653d53)

JeffAlyanak:
> Truly a good point, thanks for the patch 👍 .. I slightly modified your commit to place both conditions in just if statement / see [diff](https://github.com/searxng/searxng/compare/c15cfe03b5edfafabc0c01348f8ed47cf16b6c5b..cb29f8f7b606b8f88c2591ad60bf6f77be653d53)

Makes sense, I only had a separate if statement in case an array of exceptions was needed.

return42:
> Makes sense, I only had a separate if statement in case an array of exceptions was needed.
> I have not been able to find an exhaustive list of zero click result types. If IP is the only one which needs to be excluded then this simple fix is probably sufficient

.. I don't know whether we need any more .. practice will tell, until then we'll leave it at that :-)

PR #3000, opened 2023-11-11
[feat] engine: implementation of brave goggles

Austin-Olacsi:
## What does this PR do?

Adds the brave goggles engine. Additionally this fixes the broken site descriptions in brave. For more details see this deleted PR (which no longer works anyway): https://github.com/searxng/searxng/pull/2983

## Why is this change important?

brave goggles is a killer feature of brave search. You can manipulate brave's search results to get a very different experience. You can increase or decrease the rankings of websites, and filter unwanted sites.
I explained how it works here: https://github.com/searxng/searxng/issues/2939

## How to test this PR locally?

Uncomment this engine in the settings.yml, and add the desired goggles URL. For testing puposes i use this one:
https://raw.githubusercontent.com/brave/goggles-quickstart/main/goggles/no_pinterest.goggle
Then, you can use: make run

## Related issues

closes https://github.com/searxng/searxng/issues/2939

PR #3534, opened 2024-06-01
[feat] mojeek: implement dedicated module

### LINKED ISSUE #3526: [Documentation - Enhancement] Mojeek Safe Search
Mojeek now provides safe search: https://www.mojeek.com/support/api/search/request_parameters.html

This can be used with ```safe=1``` i.e.

https://www.mojeek.com/search?q=mojeek&safe=1

This has been introduced recently but could be reflected on: https://docs.searxng.org/user/configured_engines.html

Bnyro:
## What does this PR do?
- This implements a dedicated module for Mojeek
- This adds safesearch support as well as support for images and news
- Mojeek supports 4 or 5 different languages, so support for them could be added in the future optionally, but it doesn't seem that necessary currently

- !mjk earth
- !mjkimg earth
- !mjknews earth

## Author's checklist
- I got ratelimited multiple times for a few minutes when testing, this one probably won't work well for large instances with many users

## Related issues
closes #3526

### DIFF
--- a/searx/engines/mojeek.py
+++ b/searx/engines/mojeek.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Mojeek (general, images, news)"""
+
+from datetime import datetime
+from urllib.parse import urlencode
+from lxml import html
+
+from dateutil.relativedelta import relativedelta
+from searx.utils import eval_xpath, eval_xpath_list, extract_text, gen_useragent
+
+about = {
+    'website': 'https://mojeek.com',
+    'wikidata_id': 'Q60747299',
+    'official_api_documentation': 'https://www.mojeek.com/support/api/search/request_parameters.html',
+    'use_official_api': False,
+    'require_api_key': False,
+    'results': 'HTML',
+}
+paging = True  # paging is only supported for general search
+safesearch = True
+time_range_support = True  # time range search is supported for general and news
+max_page = 10
+
+base_url = "https://www.mojeek.com"
+
+categories = ["general", "web"]
+search_type = ""  # leave blank for general, other possible values: images, news
+
+results_xpath = '//ul[@class="results-standard"]/li/a[@class="ob"]'
+url_xpath = './@href'
+title_xpath = '../h2/a'
+content_xpath = '..//p[@class="s"]'
+suggestion_xpath = '//div[@class="top-info"]/p[@class="top-info spell"]/em/a'
+
+image_results_xpath = '//div[@id="results"]/div[contains(@class, "image")]'
+image_url_xpath = './a/@href'
+image_title_xpath = './a/@data-title'
+image_img_src_xpath = './a/img/@src'
+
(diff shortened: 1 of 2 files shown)


Austin-Olacsi:
Time range support is possible also, by adding another parameter:
`&since=YYYYMMDD`
This will show results from after the specified date.

but that can be added later if you don't feel like it.

searx/engines/mojeek.py
```diff
@@ -0,0 +1,127 @@
...
+def request(query, params):
+    args = {
+        'q': query,
+        'safe': min(params['safesearch'], 1),
+        'fmt': search_type,
+    }
+
+    if search_type == '':
+        args['s'] = 10 * (params['pageno'] - 1)
+
```
Austin-Olacsi, line 60: we would add this (after import what we need and enable time range support):

    if params['time_range']:
        date = (datetime.now() - relativedelta(**{f"{params['time_range']}s": 1})).strftime("%Y%m%d")
        args["since"] = date

Bnyro:
> Time range support is possible also, by adding another parameter: `&since=YYYYMMDD` This will show results from after the specified date.
>
Thank you, that's a nice addition :+1:

allendema:
Added the focus/filter feature to include and exclude certain domains.
If you think this is worth it, you can use or modify the changes.

https://github.com/searxng/searxng/commit/e3fb4c93be3f9ffba5dbcb86792bc9964ccefe03

Bnyro:
> Added the focus/filter feature to include and exclude certain domains. If you think this is worth it, you can use or modify the changes.
> 
> [e3fb4c9](https://github.com/searxng/searxng/commit/e3fb4c93be3f9ffba5dbcb86792bc9964ccefe03)
>
To be honest I'm not really sure how well this fits into SearXNG because users can't set the `search focus` themselves, and it probably doesn't make sense if instance maintainers need to add a lot of focuses themselvers ...

allendema:
> because users can't set the `search focus` themselves, and it probably doesn't make sense if instance maintainers need to add a lot of focuses themselvers ...

Yeah, this is more for the private instances because many admins want to include/exclude domains https://github.com/searxng/searxng/issues/2351,  https://github.com/searx/searx/issues/2001, https://github.com/searxng/searxng/issues/1086

Basically goggles implemented in https://github.com/searxng/searxng/pull/3000  but from mojeek.

And no need for plugins when admins modify the engine directly https://github.com/searxng/searxng/issues/2351#issuecomment-1789468720.

return42 (approved):
> Yeah, this is more for the private instances because many admins want to include/exclude domains 

@allendema lets add/discuss this feature in a dedicated PR / the intention of this PR is just: "mojeek: implement dedicated module" and with the addition of the *Image* & *News* search, the PR has already overstepped the mark.

Beside focus I think language (`&lb=en`) and probably location `&arc=us` support are more important (BTW: *News* seems only available for `&arc=us` & `&arc=en`).

- https://www.mojeek.com/preferences?updated=true&tab=location-language

PR #3567, opened 2024-06-12
[feat] docker: add env vars for common public instance settings

holysoles:
## What does this PR do?

This PR adds support for recommended configuration settings from `settings.yml` when setting up public instances:
- image_proxy
- limiter
- static_use_hash
- public_instance

When possible, its expected that docker images can be configured primarily through environment variables. Lessening the reliance upon manually configuring the settings.yml file helps with making this more deployable.

`limiter.toml` generation isn't in scope for this PR, but will be needed to fully realize the benefits of supporting enabling the limiter explicitly. However, the goal of this PR is to support the common public instance options, and setting `SEARXNG_PUBLIC_INSTANCE` still achieves this.

[Build the docker image](https://docs.searxng.org/admin/installation-docker.html#build-the-image): `make docker.build`

Then run it with the new environment variables, but with the public instance one false, so we test it separate from the limiter variable.

```bash
docker run --rm -ti \
             -e SEARXNG_IMAGE_PROXY=true  \
             -e SEARXNG_LIMITER=true \
             -e SEARXNG_PUBLIC_INSTANCE=false \
             -e SEARXNG_STATIC_USE_HASH=true \
             -d -p 8080:8080 \
             ${GITHUB_LOGIN}/searxng
```

Validate the contents of the `settings.yml` file have had their default values overridden. Validate logs as well as contents of the `/config` endpoint.

Do the same, but with the inverse case
```bash
docker run --rm -ti \
             -e SEARXNG_IMAGE_PROXY=true  \
             -e SEARXNG_LIMITER=false \
             -e SEARXNG_PUBLIC_INSTANCE=true \
             -e SEARXNG_STATIC_USE_HASH=true \
             -d -p 8080:8080 \
             ${GITHUB_LOGIN}/searxng
```

## Author's checklist

Any feedback on the way I implemented the documentation for these new environment variables would be welcomed.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -78,14 +78,18 @@ server:
   # public URL of the instance, to ensure correct inbound links. Is overwritten
   # by ${SEARXNG_URL}.
   base_url: false  # "http://example.com/location"
-  limiter: false  # rate limit the number of request on the instance, block some bots
-  public_instance: false  # enable features designed only for public instances
+  # rate limit the number of request on the instance, block some bots.
+  # Is overwritten by ${SEARXNG_LIMITER}
+  limiter: false
+  # enable features designed only for public instances.
+  # Is overwritten by ${SEARXNG_PUBLIC_INSTANCE}
+  public_instance: false
 
   # If your instance owns a /etc/searxng/settings.yml file, then set the following
   # values there.
 
   secret_key: "ultrasecretkey"  # Is overwritten by ${SEARXNG_SECRET}
-  # Proxying image results through searx
+  # Proxy image results through SearXNG. Is overwritten by ${SEARXNG_IMAGE_PROXY}
   image_proxy: false
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
@@ -106,6 +110,7 @@ redis:
 ui:
   # Custom static path - leave it blank if you didn't change
   static_path: ""
+  # Is overwritten by ${SEARXNG_STATIC_USE_HASH}.
   static_use_hash: false
   # Custom templates path - leave it blank if you didn't change
   templates_path: ""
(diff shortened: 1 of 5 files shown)


Bnyro (commented):
We do have an environment variable property for `SettingsValue`s, see https://github.com/searxng/searxng/blob/f5eb56b63f250c7804e5e1cf4426e550bc933906/searx/settings_defaults.py#L175 for example.

So in theory we shouldn't need to do all that creepy `sed` stuff I think.

unixfox:
Ha yes don't use sed please :)

holysoles:
appreciate the guidance! I didnt see that when reviewing how some of the existing values were set. I'll re-implement this

holysoles:
Should be updated, once again appreciate being steered in the right direction!

searx/settings.yml
```diff
@@ -78,14 +78,18 @@ server:
...
+  # enable features designed only for public instances.
+  # Is overwritten by ${SEARXNG_PUBLIC_INSTANCE}
+  public_instance: false
 
   # If your instance owns a /etc/searxng/settings.yml file, then set the following
   # values there.
 
   secret_key: "ultrasecretkey"  # Is overwritten by ${SEARXNG_SECRET}
-  # Proxying image results through searx
+  # Proxy image results through searx. Is overwritten by ${SEARXNG_IMAGE_PROXY}
```
```suggestion
  # Proxy image results through SearXNG. Is overwritten by ${SEARXNG_IMAGE_PROXY}
```
Not related to your changes, but since you're already on it :)

Bnyro (approved):
Looks good to me :+1:

searx/settings.yml
holysoles: updated!

PR #3570, opened 2024-06-12
[perf] torrents.html, files.html: don't parse and re-format filesize

### LINKED ISSUE #3419: Duplicate logic for formatting human readable file sizes
@return42 added a `humanize_bytes` function in https://github.com/searxng/searxng/commit/e76ab1a4b3a99b0ec2ef90c3aadc92ffbf6889b1.

However I figured out when working on a PR recently, that we already had a similar logic before https://github.com/searxng/searxng/blob/master/searx/templates/simple/result_templates/files.html#L38 used in the `files.html` template - it however covers less cases that the `humanize_bytes` function.

We should probably replace the logic in `files.html` with using `humanize_bytes`.

Bnyro:
## What does this PR do?
- remove the `get_torrent_size` method, that's just redundant

- it doesn't make sense to parse file sizes in string format like `2.5 GB` to an integer (byte size) first, and then format it again
- before this PR, there were multiple different ways how filesizes were formatted. This pr aims to fix that

- search with the engines modified in this PR

## Author's checklist
- I didn't test the btdig engine after this change because it requires an API key. Everything is tested and confirmed to work fine.

## Related issues
closes #3419

### DIFF
--- a/searx/engines/1337x.py
+++ b/searx/engines/1337x.py
@@ -6,7 +6,7 @@
 
 from urllib.parse import quote, urljoin
 from lxml import html
-from searx.utils import extract_text, get_torrent_size, eval_xpath, eval_xpath_list, eval_xpath_getindex
+from searx.utils import extract_text, eval_xpath, eval_xpath_list, eval_xpath_getindex
 
 # about
 about = {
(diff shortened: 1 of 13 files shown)


return42 (approved):
> search with the engines modified in this PR

Tested all engines without issues :+1: .. except:

- https://digbt.org/ .. has an SSL issue (not related to the modifications of this PR
- [Torznab EZTV](https://docs.searxng.org/dev/engines/online/) .. not sure how to setup .. and I don't have a API .. but the patch LGTM

Many thanks for the clearing action :rocket:

PR #3575, opened 2024-06-14
[feat] mozhi: fix crash, support synonyms and definition

Bnyro:
## What does this PR do?
- fix a crash when searching, reproducible with queries like `hey`, `test`, ...
- show the definition of the inserted and translated word if available
- show synonyms if available

- !mozhi en-de bird


return42 (approved):
I can reproduce the bug .. and this is the fix :+1:

PR #1792, opened 2022-09-20
use the internal API for google images

unixfox:
## What does this PR do?

This PR switch the parsing for Google Images engine to the internal API used by the Google Go Android application.

Everything is explained in https://github.com/searxng/searxng/issues/1641

The custom user-agent is needed for having more than 10 results, this user-agent is a modified version of the original user-agent with a more generic way.

## Why is this change important?

This brings back the pagination, and now we use a more stable way to get the results.

## How to test this PR locally?

1. Run SearXNG
2. Query: `!goi ok`

## Author's checklist

Not sure if I entered the correct parameters for each item in the results list, see https://github.com/searxng/searxng/pull/1792/files#diff-7f888885ad7eb66947d95abd3ffa0dabe8a1faa6b1a4b5d4ea0b40908bf0afa6R104

Here is an example JSON object if you want to be sure that I entered the correct parameters:
```json
{
    "background_color": "rgb(88,152,30)",
    "gsa": {
        "file_size": "41KB",
        "referrer_host": "icon-icons.com"
    },
    "is_animated": false,
    "layout": {
        "crop_bottom": 6,
        "crop_left": 6,
        "crop_right": 6,
        "crop_top": 6,
        "strict_cropping": true
    },
    "original_image": {
        "height": 512,
        "url": "https://cdn.icon-icons.com/icons2/1506/PNG/512/emblemok_103757.png",
        "width": 512
    },
    "result": {
        "collected": false,
        "image_source_url": "icon-icons.com",
        "is_from_teragoogle": false,
        "is_from_visual_dictionary": false,
        "page_title": "Ok Pictogram in Super Flat Remix V1.08 Emblems",
        "referrer_doc_id": "SdzLj61ig76tjM",
        "referrer_url": "https://icon-icons.com/nl/pictogram/ok/103757",
        "rendered_result_preview_size": {
            "max_thumbnail_shown": "STANDARD"
        },
        "site_title": "Icon-Icons"
    },
    "result_type": "REGULAR",
    "rich_metadata_type": "NONE",
    "text_in_grid": {
        "snippet": "Ok Pictogram in Super Flat Remix V1.08 Emblems"
    },
    "thumbnail": {
        "height": 225,
        "url": "https://encrypted-tbn0.gstatic.com/images?q\u003dtbn:ANd9GcQO1e8SbPXH5NNMX9c5WRUS9bDQ-EoIu84lk0SRCUZ48w_rDRG2\u0026s",
        "width": 225
    },
}
```

## Related issues

[...]

Issue #1641, opened 2022-08-09
Google images internal API with JSON results


unixfox:
Warning: I discovered two internal API for Google images, this one and another one in #1642!

----

Is it known that Google images has a pagination issue on SearXNG: https://github.com/searxng/searxng/issues/175

I just found an internal API in the [Google Go android](https://play.google.com/store/apps/details?id=com.google.android.apps.searchlite) application where you can search on Google API and have a JSON result returned! And it support the pagination with just a `start=` parameter.

Here is how to do the requests.

You need to send a GET request to this URL: `https://www.google.com/search?q=ok&hl=fr&start=0&asearch=isch&tbm=isch&async=_fmt:json,p:1` and with the header `accept: */*`.

If you want more results in one page, you need to use this user-agent: `NSTN/3.58.464445598.release Dalvik/2.1.0 (Linux; U; Android 12; US) gzip`. It's a modified one from the stock in order to avoid any special fingerprinting: `NSTN/3.58.464445598.release Dalvik/2.1.0 (Linux; U; Android 10; SM-G930F) Mobile`

Here is an explanation of the parameters:
- `asearch=isch` to images "category"
- `tbm=isch` optional but related to images "category"
- `q=ok` the search query, similar parameter to the one used in the google engine
- `hl=fr` the language, similar parameter to the one used in the google engine
- `async=_fmt:json,p:1,ijn:1`:
  - `_fmt:json`: ask Google for JSON results, if you set `_fmt:pb` you get the results in protobuf
  - `p:1`: no idea what is this parameter but if you remove it the final JSON is encoded inside a JSON object called `results`, I don't recommend removing it
  - `ijn:1`: that's the page number, increment it for each next page

All the other parameters used in the google engine should work too: gl, ie, oe and so on.

Krando:
Wonderful. Hopefully this gets added soon, is this method compatible with safesearch?

unixfox:
> is this method compatible with safesearch?

Yes all the parameters used in the current google images implementation work.

Krando:
> > is this method compatible with safesearch?
> 
> Yes all the parameters used in the current google images implementation work.

Does this method work for videos as well? Pagination for Google videos hasn't been implemented on searxng either. It would be great if pagination is enabled for both image & videos with this method.

unixfox:
> Does this method work for videos as well? Pagination for Google videos hasn't been implemented on searxng either. It would be great if pagination is enabled for both image & videos with this method.

Yes please see #1642

PR #3579, opened 2024-06-15
[fix] !goi irrelevant results AND display more results

### LINKED ISSUE #1864: Google images engine returning poorly sorted results with no simmilarity to google images or startpage images

Latest release on dockerhub, "2022.10.14-1a5b0965"

searxng-docker

**What happened?**
With only google images enabled, image search returns oddly sorted results that, while funny, are quite useless.

**How To Reproduce**

- Enable only google images for image search
- Search for "first time meme"
- Do the same search on Google or Startpage and compare

**Expected behavior**

Results should be roughly the same as regular google images or startpage images.

**Screenshots & Logs**

Google images natively:
image
Startpage images (just an example because they retain roughly the same order)
image
SearXNG google images:
image

**Additional context**

I am currently using qwant as a workaround and other image search engines work flawlessly too.

allendema:
## What does this PR do?

Fixes results from second page displaying on first page.
Adds ua from android app to get more results instead of just 10.

Better and more results.

make format.python
make run
`!goi your query`

## Author's checklist

The custom headers have been removed from https://github.com/searxng/searxng/pull/1792 / commit cf7ee67f71f6c6b0c2e357ffef12a79d88fea73b but could not see why.
## Related issues

Closes https://github.com/searxng/searxng/issues/1864

### DIFF
--- a/searx/engines/google_images.py
+++ b/searx/engines/google_images.py
@@ -63,16 +63,11 @@ def request(query, params):
         'https://'
         + google_info['subdomain']
         + '/search'
-        + "?"
-        + urlencode(
-            {
-                'q': query,
-                'tbm': "isch",
-                **google_info['params'],
-                'asearch': 'isch',
-                'async': '_fmt:json,p:1,ijn:' + str(params['pageno']),
-            }
-        )
+        + '?'
+        + urlencode({'q': query, 'tbm': "isch", **google_info['params'], 'asearch': 'isch'})
+        # don't urlencode this because wildly different AND bad results
+        # pagination uses Zero-based numbering
+        + f'&async=_fmt:json,p:1,ijn:{params["pageno"] - 1}'
     )
 
     if params['time_range'] in time_range_dict:
@@ -80,9 +75,13 @@ def request(query, params):
     if params['safesearch']:
         query_url += '&' + urlencode({'safe': filter_mapping[params['safesearch']]})
     params['url'] = query_url
-
     params['cookies'] = google_info['cookies']
     params['headers'].update(google_info['headers'])
+    # this ua will allow getting ~50 results instead of 10. #1641
+    params['headers']['User-Agent'] = (
+        'NSTN/3.60.474802233.release Dalvik/2.1.0 (Linux; U; Android 12;' f' {google_info.get("country", "US")}) gzip'
+    )
+
     return params
 
 
@@ -96,7 +95,6 @@ def response(resp):
     json_data = loads(resp.text[json_start:])
 
     for item in json_data["ischj"].get("metadata", []):
-
         result_item = {
             'url': item["result"]["referrer_url"],
             'title': item["result"]["page_title"],


searx/engines/google_images.py
```diff
@@ -64,25 +64,29 @@ def request(query, params):
...
-                'asearch': 'isch',
-                'async': '_fmt:json,p:1,ijn:' + str(params['pageno']),
-            }
-        )
+        + urlencode({'q': query, 'tbm': "isch", **google_info['params'], 'asearch': 'isch'})
+        # don't urlencode this because wildly different AND bad results
+        # pagination uses Zero-based numbering
+        + '&'
+        + 'async=_fmt:json,p:1,ijn:'
+        + str(int(params['pageno'] - 1))
```
Bnyro:
```suggestion
        + f'&async=_fmt:json,p:1,ijn:{params["pageno"] - 1}'
```

searx/engines/google_images.py
```diff
@@ -64,25 +64,29 @@ def request(query, params):
...
     if params['safesearch']:
         query_url += '&' + urlencode({'safe': filter_mapping[params['safesearch']]})
     params['url'] = query_url
-
     params['cookies'] = google_info['cookies']
     params['headers'].update(google_info['headers'])
+    # this ua will allow getting ~50 results instead of 10. #1641
+    params['headers'].update(
+        {
+            "User-Agent": f"NSTN/3.60.474802233.release Dalvik/2.1.0 (Linux; U; Android 12; \
```
Bnyro: It seems like we had already used that user agent earlier but @return42 changed it in https://github.com/searxng/searxng/commit/249989955497cd048fa3312d115971282983b269#diff-7f888885ad7eb66947d95abd3ffa0dabe8a1faa6b1a4b5d4ea0b40908bf0afa6.

@return42 can you still remember why?

Bnyro:
Works pretty well from my tests, so generally I think this change is a good idea.

searx/engines/google_images.py
return42: > @return42 can you still remember why?

nope .. hm, may it was accidentally removed when I forgot to read ..

- https://github.com/searxng/searxng/issues/1641#issue-1333642470

searx/engines/google_images.py
```diff
@@ -63,26 +63,28 @@ def request(query, params):
...
```
return42: Please note, this is a string with a lot of whitepaces in .. I recommend something like this (untested)

        params['headers']['User-Agent'] = f'NSTN/3.60.474802233.release Dalvik/2.1.0 (Linux; U; Android 12; {google_info.get('country', 'US')}) gzip'

searx/engines/google_images.py
allendema: Done.

searx/engines/google_images.py
allendema: ###### Had different variations but `make test` / `make format.python` would complain or make it even worse.
Changed it. :+1:

Bnyro (approved):

Issue #3337, opened 2024-03-15
autocomplete has a bug


return42:
As #3316 pointed out to me, we have the [answerers](https://github.com/searxng/searxng/tree/master/searx/answerers) I didn't aware before / may since it has never been worked (AFAIK).

It was introduced in 

- commit 63f17d2e4c from https://github.com/searx/searx/pull/2593

The OpenSearch part was fixed in

- commit 169438137f from https://github.com/searx/searx/pull/2716

the main issue is the `raw_text_query.changeQuery(result)` ..

https://github.com/searxng/searxng/blob/e2af3e49702f6fb40e1614f826544dc3b03bca2f/searx/webapp.py#L843-L846

.. which modifies the result query for all following usages .. I tried to understand the why this side effect is needed here .. but the autocompleter implementation is to complex and I gave up after 4h of investigation.

----

I have a doubt we really want to have *answers* in the autocompletion, a simple fix will look like this:

```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index a6cadcf6c..3c645e7a6 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -831,6 +831,10 @@ def autocompleter():
     raw_text_query = RawTextQuery(request.form.get('q', ''), disabled_engines)
     sug_prefix = raw_text_query.getQuery()
 
+    for answers in ask(raw_text_query):
+        for answer in answers:
+            results.append(str(answer['answer']))
+
     # normal autocompletion results only appear if no inner results returned
     # and there is a query part
     if len(raw_text_query.autocomplete_list) == 0 and len(sug_prefix) > 0:
@@ -849,10 +853,6 @@ def autocompleter():
         for autocomplete_text in raw_text_query.autocomplete_list:
             results.append(raw_text_query.get_autocomplete_full_query(autocomplete_text))
 
-    for answers in ask(raw_text_query):
-        for answer in answers:
-            results.append(str(answer['answer']))
-
     if request.headers.get('X-Requested-With') == 'XMLHttpRequest':
         # the suggestion request comes from the searx search form
         suggestions = json.dumps(results)
```

Here by example the random string from 

https://github.com/searxng/searxng/blob/e2af3e49702f6fb40e1614f826544dc3b03bca2f/searx/answerers/random/answerer.py#L21-L22

is shown in the autocompletion ..

![image](https://github.com/searxng/searxng/assets/554536/93bc8c1c-eb49-4538-91be-1b8b1ca48402)

[...]

PR #3580, opened 2024-06-16
[refactor] Update getFullQuery to optionally override the query, remove changeQuery

glanham-jr:
## What does this PR do?

- Introduce typing in RawTextQuery
- Remove an outdated description based on the change
- Refactor setting the autocomplete_location, its more clear to set the
variable in the construct instead of within the method
- Update unit tests to reflect the new method

Mutating the RawTextQuery object may lead to unexpected results or different call chains updating the query without careful consideration.  Looking at the usage of `changeQuery`, it was simply a proxy to call `getFullQuery` with a new `query`. We can reduce the number of risky operations by removing the mutable `changeQuery` in favor of passing an override query into `getFullQuery` itself.

Run SearxNG locally, search, and ensure there are no issues searching the happy path.

Ensure unit tests run as expected.

## Author's checklist

n/a

## Related issues

Suggestion provided by Dalf on #3337 , modified after realizing usage of `changeQuery`

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -1,10 +1,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=invalid-name, missing-module-docstring, missing-class-docstring
-
 from __future__ import annotations
+
 from abc import abstractmethod, ABC
 import re
-
 from searx import settings
 from searx.sxng_locales import sxng_locales
 from searx.engines import categories, engines, engine_shortcuts
@@ -274,14 +273,17 @@ class RawTextQuery:
         # internal properties
         self.query_parts = []  # use self.getFullQuery()
         self.user_query_parts = []  # use self.getQuery()
-        self.autocomplete_location = None
         self.redirect_to_first_result = False
-        self._parse_query()
+        self.autocomplete_location = self._parse_query()
 
     def _parse_query(self):
         """
         parse self.query, if tags are set, which
         change the search engine or search-language
+
+        Returns a tuple:
+            [0] The query parts as a list
+            [1] The indexor into the above list
         """
 
         # split query, including whitespaces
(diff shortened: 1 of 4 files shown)


searx/query.py
```diff
@@ -306,28 +310,30 @@ def _parse_query(self):
...
         qlist[position] = text
         return self.getFullQuery()
 
-    def changeQuery(self, query):
-        self.user_query_parts = query.strip().split()
-        self.query = self.getFullQuery()
-        self.autocomplete_location = (self.user_query_parts, len(self.user_query_parts) - 1)
-        self.autocomplete_list = []
-        return self
+    def changeQuery(self, query: str) -> "RawTextQuery":
```
return42: > Suggestion provided by Dalf on https://github.com/searxng/searxng/issues/3337

Its not clear, why this method is named `changedQuery`:

- what does it change? .. and why should the caller change (a "immutable") query
- seems to be a factory method?  .. what does it produce?

My point is, when someone (like me) reads these lines of code its unclear what the intention of the code is and for what it is good .. I have to look at the blocks of code where this code is used to get a vague idea of what the intent of this code was

Unfortunately, this PR does nothing to change this situation, even if it may remedy a current issue.

- https://github.com/searxng/searxng/issues/3337#issuecomment-1999553069

glanham-jr: Happy to rename the method to something like 'copyWithNewQuery', and should solve the ambiguity.

The goal here is incremental development. Making this method immutable has its benefits, and we can have a subsequent PR to handle answerers. Perhaps this method will be deleted if no longer used then, but it can be immutable for the time being.

glanham-jr: On reading the implementations and a fresh night of sleep, I saw the same pattern `changeQuery` was being used for. Its only current use it to create a new FullQuery with a fresh set of parameters. However, since `getFullQuery` only requires state from `query_parts` and `user_query_parts`, the entire modification isn't necessary for an updated `getFullQuery`. Check out the new commits - it shows the current usages + what the code is good for now :smile: should reduce risk with this clearer usage.

return42:
> Suggestion provided by @dalf on #3337 , modified after realizing usage of `changeQuery`

@dalf what do you think about this PR? .. ready to merge?

From my side: LGTM :+1:

searx/query.py
```diff
@@ -306,28 +309,33 @@ def _parse_query(self):
...
         return ' '.join(self.user_query_parts)
 
-    def getFullQuery(self):
+    def getFullQuery(self) -> str:
         """
-        get full query including whitespaces
+        Get full query including whitespaces
         """
-        return '{0} {1}'.format(' '.join(self.query_parts), self.getQuery()).strip()
+        return RawTextQuery._getFullQuery(self.query_parts, self.getQuery())
```
return42: I think these are to much methods for more or less the same purpose: getNewFullQuery, getFullQuery, _getFullQuery, getQuery .. if we need a copy of the object, we should implement a [__copy__ method](https://docs.python.org/3/library/copy.html#object.__copy__).

glanham-jr: Looking back on this old PR, you are definitely right.

I've simplified the change by just removing `changeQuery` and just adding a new parameter to `getNewFullQuery` to keep it simple. Let me know if the new changes are good.

Issue #3558, opened 2024-06-09
[Bug] New "Hostnames plugin" does not seem to work at all


saloniamatteo:
TL;DR; [the "unhashable type" error is a result of mixing up YAML's lists and mappings.](https://github.com/searxng/searxng/issues/3558#issuecomment-2175058128) .. online docs has been improved .. read [online doc](https://docs.searxng.org/src/searx.plugins.hostnames.html) carefully  ..

------------

SearXNG version: 2024.6.7+f5eb56b63

**What happened?**
After the switch from "hostname_replace" to the new "Hostnames plugin", it doesn't seem to work anymore.

Modifying settings.yml, and searching for a relevant query, shows us that the plugin does not seem to be working. (Replacing does not work either, nor do priorities.)

**How To Reproduce**
Edit `settings.yml`, and add the following:

```yaml
enabled_plugins:
 - 'Hostnames plugin'
```

```yaml
hostnames:
 remove:
 - '(.*\.)?example.com$'
```

**Expected behavior**
The results for `example.com` and its subdomains should not appear at all.

**Screenshots & Logs**
![Screenshot_20240609_183858_Termux](https://github.com/searxng/searxng/assets/28765699/dc80b564-48bb-445a-8dfc-4837d4f1a0ae)

(Just to be sure, I added 'Hostname replace' as well, however, it didn't make a difference)

![Screenshot_20240609_183947_Mull](https://github.com/searxng/searxng/assets/28765699/90ae8ff6-6098-49b3-87f0-8b09a510d719)

saloniamatteo:
Upon further inspection, I got the following error:

![Screenshot_20240609_190351_Termux](https://github.com/searxng/searxng/assets/28765699/74ad1289-4261-42a2-b582-2db592a4bb88)

I do not know why the plugin appears to be disabled, even though I added both names (old & new).

Edit: I want to make it clear this error happens even when removing the old deprecated name.

return42:
Its disabled because the setting has an issue .. check indentations:

```yaml
hostnames:
 remove:
 - '(.*\.)?example.com$'
```
fixed by indentaion of the list items .. (the last line):

```yaml
hostnames:
 remove:
   - '(.*\.)?example.com$'
```

saloniamatteo:
@return42 I'm afraid this did not fix the issue.

return42:
I'm pretty sure you have problems with the indentations or something like that ...

saloniamatteo:
[...]

PR #3585, opened 2024-06-18
[doc] hostname plugin: improve online documentation

return42:
The data types (list & map) should be made clearer, as these sometimes lead to misunderstandings.

[1] https://github.com/searxng/searxng/issues/3558#issuecomment-2175058128

### DIFF
--- a/searx/plugins/hostnames.py
+++ b/searx/plugins/hostnames.py
@@ -13,21 +13,54 @@ other features.
    **To maintainers of SearXNG instances, please modify your old plugin config
    to the new.**
 
-- ``hostnames.replace``: A mapping of regular expressions to hostnames to be
+- ``hostnames.replace``: A **mapping** of regular expressions to hostnames to be
   replaced by other hostnames.
 
-- ``hostnames.remove``: A list of regular expressions of the hostnames whose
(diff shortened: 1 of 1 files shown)


searx/plugins/hostnames.py
```diff
@@ -13,35 +13,68 @@
...
+     hostnames:
+       low_priority:
+         - '(.*\\.)?google(\\..*)?$'
+         - ...
+
+If the URL matches the pattern of ``high_priority`` AND ``low_priority``, the
+higher priority wins over the lower priority.
+
+Alternatively, a file name can also be specified for the **mappings** or
+**lists**:
```
Bnyro:
```suggestion
Alternatively, you can also specify a file name for the **mappings** or
**lists** to load these from an external file:
```
Just to make it easier to understand when reading for the first time

return42: yeah, my English is poor, I am grateful for any improvement :-) .. fixed [diff](https://github.com/searxng/searxng/compare/c062190a71d09ed55d5f90f6628ad4a8be703382..ee8fefd92a59c366b7e8942cb8551d20aea2dde1)

Issue #2034, opened 2022-12-21
Bug: elasticsearch engine


ErezAmihud:
Using searxng-docker
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2022.12.16-b9274821

I set the engine in the settings to the following:
```
- name: elasticsearch
    shortcut: es
    engine: elasticsearch
    base_url: http://localhost:9200
    username: elastic
    password: changeme
    index: crawl-data
    # available options: match, simple_query_string, term, terms, custom
    query_type: match
    # if query_type is set to custom, provide your query here
    #custom_query_json: {"query":{"match_all": {}}}
    #show_metadata: false
    disabled: false
```

and with the docker-compose:
```
version: '2'
services:
  ache:
    image: vidanyu/ache
    command: startCrawl -c /config/ -s /config/docker.seeds -o /data -e crawl-data
    ports:
    - "8080:8080"
    volumes:
    - ./data-ache/:/data
    - ./:/config
    links:
    - elasticsearch
  elasticsearch:
    image: docker.elastic.co/elasticsearch/elasticsearch:6.8.22
    environment:
      - discovery.type=single-node
      - cluster.name=docker-cluster
      - bootstrap.memory_lock=true
    ulimits:
      memlock:
        soft: -1
        hard: -1
    volumes:
      - ./data-es/:/usr/share/elasticsearch/data # elasticsearch data will be stored at ./data-es/
    ports:
      - 9200:9200
  searx:
    container_name: searx
    image: searxng/searxng:latest
    restart: always
    command: ${SEARX_COMMAND:-}
    ports:
      - 8090:8090
    volumes:
      - ./searxng:/etc/searxng:rw
    environment:
      - BIND_ADDRESS=0.0.0.0:8090
      - BASE_URL=http://${SEARX_HOSTNAME:-localhost}/
    links:
    - elasticsearch
```
**What happened?**
After populating the elastic search I tried to search anything, and there was an error message about the engine, including the following traceback in the docker-compose logs:
```
2022-12-21 20:24:01,974 WARNING:searx.engines.elasticsearch: ErrorContext('searx/engines/elasticsearch.py', 54, 'raise ValueError(\'query format must be "key:value"\') from e', 'ValueError', None, ()) False
searx  | 2022-12-21 20:24:01,974 ERROR:searx.engines.elasticsearch: exception : query format must be "key:value"
searx  | Traceback (most recent call last):
[...]

PR #3588, opened 2024-06-19
Fix search_url construction for elasticsearch/opensearch.

rick-github:
## What does this PR do?

Fixes the `search_url` construction to take the values of `base_url` and `index` into account.

It's impossible to connect to an elasticsearch/opensearch instance not running on localhost:9200.

1.  Uncomment elasticsearch in settings.yml.
2. Change `base_url` to http://localhost:9201, delete `disabled` line, add `enable_http: true`.
3.  Run a fake elasticsearch instance: `echo -e 'HTTP/1.0 200 OK\n\n{"hits":{"hits":[{"_source":{"content":"stuff"}}]}}' | nc -l -p 9201`.
4. Start searxng and search for `!es key:value`.

## Author's checklist

Not sure this ever worked because `index` would always have been empty, resulting in a call to localhost:9200//_search.

## Related issues

Issue https://github.com/searxng/searxng/issues/2034 raised this but was closed without an update.

### DIFF
--- a/searx/engines/elasticsearch.py
+++ b/searx/engines/elasticsearch.py
@@ -49,7 +49,7 @@ base_url = 'http://localhost:9200'
 username = ''
 password = ''
 index = ''
-search_url = base_url + '/' + index + '/_search'
+search_url = '{base_url}/{index}/_search'
 query_type = 'match'
 custom_query_json = {}
 show_metadata = False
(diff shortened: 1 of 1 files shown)


return42 (approved):
Thanks a lot for the fix :+1:

PR #3592, opened 2024-06-20
[fix] simple theme: in URLs don't truncate descenders (typography)

### LINKED ISSUE #3550: URL is truncated in some cases

2024.5.31+18fb701be

searxng-docker

**What happened?**

The lower part of the URL is truncated in the desktop version. Can't see that in mobile version. Probably because of the `overflow: hidden` in `.url_wrapper`.

**How To Reproduce**

1. Conduct any searches.
2. Can observe that the lower part of some characters will be truncated.

**Expected behavior**

No parts should be truncated, except for URLs that are too long.

**Screenshots & Logs**

Note that the "g" in the screenshot. 
image

**Additional context**

https://github.com/searxng/searxng/pull/510

return42:
Descenders [1] in the URL are truncated, caused by the `overflow: hidden;`
because part of link overflow the flex box [2].

[1] https://en.wikipedia.org/wiki/Descender
[2] https://github.com/searxng/searxng/issues/3550

Closes: https://github.com/searxng/searxng/issues/3550
Suggested-by: @coxde


return42:
@coxde I copy your feedback to this PR from https://github.com/searxng/searxng/issues/3550#issuecomment-2180876631 to here in the thread of the PR .. (lets discuss the PR here in the PR).

> Thanks for the quick fix, I've tested it and I found some issues. As in the screenshot, the `https://www.techtarget.com` works well but the `g` in `searchnetworking › definition › URL` is still be truncated.

> image

You have good eyes ;-) .. I forgot to remove the `overflow: hidden` also from the right side (the path part of the url) .. its now fixed, look for `.url_o2` in the [diff](https://github.com/searxng/searxng/compare/bcdf731141b44cf58c2782efe148c5013825a686..b358c1ace7f3108e174a0204f7135e5cbf2b34c3).

Can you please give it another try / thanks :+1:

coxde:
Thanks again. Removing `overflow: hidden` from .`url_o2` did fix the problem, but also broke the truncation of too long URL.

image

return42:
> but also broke the truncation of too long URL.

:facepalm: .. I'm a fool .. re-added y-axis again, see [fix](https://github.com/searxng/searxng/compare/b358c1ace7f3108e174a0204f7135e5cbf2b34c3..575df56dd2e64669ad1826066b77d7633139df69)

coxde:
I should've told you that I've tried to mess around with `overflow-y: hidden;` but it didn't work as expected, it still looks like this...very strange

image

Edit: it should be `overflow-x: hidden;` right? Because we only want to hide the URL that's too long.

return42:
> Edit: it should be `overflow-x: hidden;` right? Because we only want to hide the URL that's too long.

Sure .. I mixed x & y -axis :-)

I tried several settings without success, I ended in set overflow on x and y axis on the flex box .. and on every element in the flex box .. just to be explicit in the settings .. 0cfde2e82 .. and I got this:

grafik

The descender of the `g` in the wikipedia `.org` is shown .. but the `g` from hostinger.com is truncated .. now I'm really confused.

Since this only happens to me in my FFox and only when I set zoom factor to 80%  ..

--> is it a bug in the rendering engine of FFox?

I don't understand why there should be an overflow in the y axis at all .. is something wrong with the FFox in the calculation of font sizes & descender?

TBH: I don't have a clue ..

coxde:
It is very frustrating...after some [research](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue), it turns out that this behavior is a W3C spec (very bad one tho)

> The computed values of ‘overflow-x’ and ‘overflow-y’ are the same as their specified values, except that some combinations with ‘visible’ are not possible: if one is specified as ‘visible’ and the other is ‘scroll’ or ‘auto’, then ‘visible’ is set to ‘auto’. The computed value of ‘overflow’ is equal to the computed value of ‘overflow-x’ if ‘overflow-y’ is the same; otherwise it is the pair of computed values of ‘overflow-x’ and ‘overflow-y’.

And here's my hacky solution:

https://github.com/coxde/searxng/commit/d2a64b253adac40f4aa564aff47b05dbb40ec87f

image

return42:
> .after some [research](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue),

Thanks a lot for your research .. what I do not understand: 

- why is there a overflow on the y-axis, descender should not cause the overflow on y-axis .. or I'm wrong? .. do you know why there is an overflow? .. 
- I see this overflow only in FFox (at a zoom of 80%) .. did you have ever seen this issue in other rendering engines, Chrome or Safari?

coxde:
> why is there a overflow on the y-axis, descender should not cause the overflow on y-axis .. or I'm wrong? .. do you know why there is an overflow? ..

I think so, I found [this](https://developer.mozilla.org/en-US/docs/Glossary/Ink_overflow). So probably still a FF bug or [decision](https://github.com/w3c/csswg-drafts/issues/8649)?

> I see this overflow only in FFox (at a zoom of 80%) .. did you have ever seen this issue in other rendering engines, Chrome or Safari?

I can't test on it Safari, but it didn't happen on Chrome and Firefox Android, and for me, it always happens on Firefox Desktop (no need to zoom), possibly because of font (it didn't happen if I change from Noto Sans to another font).

return42:
> I think so, I found [this](https://developer.mozilla.org/en-US/docs/Glossary/Ink_overflow). So probably still a FF bug or [decision](https://github.com/w3c/csswg-drafts/issues/8649)?

Thanks for sharing / to be honest, I can't rate it, my CSS knowledge is too poor for that :-)

> I can't test on it Safari, but it didn't happen on Chrome and Firefox Android, and for me, it always happens on Firefox Desktop (no need to zoom), possibly because of font (it didn't happen if I change from Noto Sans to another font).

For me, its a bug in FFox and the only workaround I found is to add a minimal padding at the bottom of the elemnts https://github.com/searxng/searxng/commit/3f4c69444a8127c443064e64d8fca36b30908ecd

Does the workaround work for you?. If so, we can merge this PR .. otherwise I don't have any better suggestion .. and would give up at this point :-)

coxde:
Hi, thanks again and it do work for me. Feel free to merge this PR if you like!

PR #3599, opened 2024-06-21
Fix unit tests for engines.

rick-github:
## What does this PR do?

Enables unit tests in the engines directory by adding `__init__.py`, and fixups for the enabled tests.

Unit tests.

`make test`

## Related issues

### DIFF
--- a/tests/unit/engines/__init__.py
+++ b/tests/unit/engines/__init__.py
@@ -1,0 +1,2 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
--- a/tests/unit/engines/test_command.py
+++ b/tests/unit/engines/test_command.py
@@ -21,7 +21,7 @@ from searx.engines import command as command_engine
 from tests import SearxTestCase
 
 
-class TestCommandEngine(SearxTestCase):
+class TestCommandEngine(SearxTestCase):  # pylint: disable=missing-class-docstring
     def test_basic_seq_command_engine(self):
         ls_engine = command_engine
         ls_engine.command = ['seq', '{{QUERY}}']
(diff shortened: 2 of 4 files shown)


searx/engines/xpath.py
```diff
@@ -72,6 +72,7 @@
 from lxml import html
 from searx.utils import extract_text, extract_url, eval_xpath, eval_xpath_list
 from searx.network import raise_for_httperror
+from searx import logger
```
return42: No need to import logger.  In the engine modules, logger is already added by some monkey patching of the SearXNG core / compare https://github.com/searxng/searxng/commit/73a6da4dd9ab95555969d42aa64bbc1c78bb0a71

https://github.com/searxng/searxng/blob/f6f622f7e5627cc5afc8919a8fbc451b6003c9d7/searx/engines/__init__.py#L155-L171

searx/engines/xpath.py
```diff
@@ -208,11 +209,14 @@
...
 
+categories = []
+'''engine dependent config'''
+
 
 def request(query, params):
     '''Build request parameters (see :ref:`engine request`).'''
     lang = lang_all
-    if params['language'] != 'all':
+    if params.get('language', 'all') != 'all':
```
return42: These param keys are set with defaults .. no need to safety *get*.

https://github.com/searxng/searxng/blob/f6f622f7e5627cc5afc8919a8fbc451b6003c9d7/searx/engines/__init__.py#L29C1-L48

return42 (changes requested):
Thanks for fixing a long unseen bug :+1:  .. the modifications of the XPath engine is not needed, can you revert the patch .. the rest of this PR LGTM.

return42 (approved):
FYI: I squashed your two commits into one and modified the commit message.

searx/engines/xpath.py
return42: > No need to import logger. In the engine modules ...

OK, now I understand why you have been added the logger to the xpath.py module .. today I had this (randomly) error in the unit tests .. its fixed in https://github.com/searxng/searxng/pull/3609

PR #3607, opened 2024-06-25
GitHub Actions: Upgrade data-update.yml infrastructure

cclauss:
## What does this PR do?

Upgrade infrastructure.

Modern versions are more performant (esp. Python).


return42 (approved):
> Modern versions are more performant (esp. Python).

:-) .. not really needed in the "Update data" matrix .. but .. yes, why not :+1:

Issue #3590, opened 2024-06-20
Dragging highlighted text to a new tab does not work


kchousos:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

Docker version: 2024.6.20+f195d98bf

**How did you install SearXNG?**

From the official docker compose

**What happened?**

When highlighting text and dragging it in a new tab in Firefox 127, SearXNG does not perform a search, it just loads the starting page.

**How To Reproduce**

1. Make SearXNG the default search engine.
2. Highlight some text in a tab.
3. Drag it to create a new tab.

**Expected behavior**

There should be a new SearXNG tab with the dragged text as a search query.

return42:
Works for me; FF 126 and [OpenSearch](https://docs.searxng.org/user/about.html#how-do-i-set-it-as-the-default-search-engine) .. not sure its related but my instance https://darmarit.org/searx/ uses HTTP GET method by default.

kchousos:
I tried both POST and GET, same issue persists.

return42:
> 1. Make SearXNG the default search engine.

How? .. did you used [OpenSearch](https://docs.searxng.org/user/about.html#how-do-i-set-it-as-the-default-search-engine)?

kchousos:
> did you used [OpenSearch](https://docs.searxng.org/user/about.html#how-do-i-set-it-as-the-default-search-engine)?

Yes. I removed it and added it again just now and the problem persists.

return42:
> in Firefox 127

on what OS & desktop?

kchousos:
Fedora Linux 40, Gnome 46 under Wayland.

Bnyro:
My SearXNG instance says the following in the debug logs:
`
OK ::/48: /search || X-Forwarded-For: my.ip.address || X-Real-IP: my.ip.address || form: {} || Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8 || Accept-Language: en-US,en;q=0.5 || Accept-Encoding: gzip, deflate, br, zstd || Content-Type: None || Content-Length: None || Connection: None
`
when dragging text to a new tab as describe above.

So it seems that the search query is not actually forwarded anywhere (at least SearXNG doesn't see any query), I can't tell why as it works for DDG & co.

return42:
> My SearXNG instance says the following in the debug logs:

[...]

Issue #711, opened 2022-01-06
Method POST harms UX without providing a tangible privacy benefit


not-my-profile:
In the preferences users can choose which HTTP method is used by the UI to perform searches: POST or GET. The setting defaults to POST.

## Privacy implications

POST has two privacy benefits:

1. the search query does not end up in the browser history
2. the search query does not end up in the web server log

I'd like to argue that both of these benefits are mostly superficial, because:

1. If you don't want your search queries to end up in your browser history, you should be using a private/incognito browser window. Because otherwise the search results you click on will end up in the browser history anyway, revealing what you have searched for.
2. There is nothing stopping a malicious server from logging search queries in POST request bodies. Either you trust the server admin or you don't. If you trust them, you should also be able to trust them to disable / periodically delete their web server logs. If you don't trust them, then they might log your every query even when you submit them using POST.

I think we should rather educate our users about the importance of private/incognito browser mode & trusting the server admin than giving them a false sense of security by using POST.

## UX implications

* **Open Link in New Tab** does not work for our search result tabs because they aren't actual links.
* **Bookmarking** a search becomes more difficult with POST. You cannot just press a browser keyboard shortcut `Ctrl+D` ... no you have to copy the Search URL from the sidebar. This can pose a real struggle for less technically-minded users.
* **Sharing** a search with somebody else becomes more difficult with POST (for the same reason).

## Conclusion

I think by using POST we are sacrificing our UX without actually getting any tangible privacy benefit.

I therefore propose that we should switch to GET by default. Since POST does not provide any real privacy benefit I think that we should remove the option.

return42:
> I therefore propose that we should switch to GET by default. Since POST does not provide any real privacy benefit I think that we should remove the option.

additional note: I remember about some issues in the context of [browsers that support OpenSearch](https://github.com/searxng/searxng/issues/431)  .. if I'm not wrong POST is not supported by Chrome.

unixfox:
While I agree that if a user wants to hide their search history they should use the incognito mode, and if a malicious instance owner really wants to log the search of their users he can log the POST body.

But the benefit of having POST method by default is that we conveniently offer a protection for the searx users without having them doing anything.
- By default, the search queries are not logged in the history.
- It's more time-consuming for instance owners to log the search queries.
- Whatever the web server configuration, the search queries keyword are not logged in the logs of the web server. This also offer the benefit of leaving no trace of the search queries done if an investigation would be done by external parties on the web server.

About the UX arguments:
[...]

PR #3619, opened 2024-06-28
[doc] adds the missing documentation of the server.method settings

return42:
TL;DR; For all the issues that comes with HTTP POST I recommend instance maintainers to switch to GET and lock the property in the preferences:

```yaml
server:
  method: GET

preferences:
  lock:
    - method
```

We don't want this in the defaults of the SearXNG distributions for the pro vs cons reasons named in the discussion below: 

-----------

HTTP GET vs POST
================

Already diskussed in "Method POST harms UX without providing a tangible privacy benefit" [1].  At that time we agreed to POST.  However, as we are having more and more drawbacks with POST, I suggest that we reconsider our previous decision. The latest cause was [2], but we also have other problems that negatively affect the UI [1]:

* Open Link in New Tab: does not work for our search result tabs because they aren't actual links.

* Bookmarking: a search becomes more difficult with POST. You cannot just press a browser keyboard shortcut `Ctrl+D` ... no you have to copy the Search URL from the sidebar. This can pose a real struggle for less technically-minded users.

* Sharing: a search with somebody else becomes more difficult with POST (for the same reason).

* [While autocompletion in the search field of SearXNG sends a cookie, the autocompletion in the URL bar does not send a cookie.](https://github.com/searxng/searxng/issues/1889#issuecomment-1565366772)
* https://github.com/searxng/searxng/pull/2333#issuecomment-1565392120
* [POST vs GET --> some broswers don't like HTTP POST requests in a OpenSearch query ](https://github.com/searxng/searxng/issues/431)

* https://github.com/searxng/searxng/issues/3590#issuecomment-2196891552

Lock HTTP method in the preferences
===================================

If the user changes the HTTP method in his settings, e.g. from GET to POST, but has not removed the SearXNG instance from the WEB browser and added it again, the WEB browser will continue to work with the old setting (GET), while entries in the HTML form use the newly set method (POST). Not realted to this commit, but this complication is also known from autocomplete[3].

Only very few maintainers are aware of this fact and probably none of the users know about it.  We should provide a setup in our defaults that is manageable in its entirety and comprehensible for the user.  For this reason, the option to select the HTTP method in the preferences is also disabled in this commit.

- [1] https://github.com/searxng/searxng/issues/711
- [2] https://github.com/searxng/searxng/issues/3590
- [3] https://github.com/searxng/searxng/pull/2333#issuecomment-1565392120

### DIFF
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -72,4 +72,4 @@
 .. _HTTP headers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
 
 ``default_http_headers`` :
-  Set additional HTTP headers, see `#755 <https://github.com/searx/searx/issues/715>`__
+  Set additional `HTTP headers`_, see `#755 <https://github.com/searx/searx/issues/715>`__
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -55,11 +55,21 @@
 
 .. _method:
 
-``method`` : ``$SEARXNG_METHOD``
-  Whether to use ``GET`` or ``POST`` HTTP method when searching.
+``method`` : ``GET`` | ``POST``
+  HTTP method.  By defaults ``POST`` is used / The ``POST`` method has the
+  advantage with some WEB browsers that the history is not easy to read, but
+  there are also various disadvantages that sometimes **severely restrict the
+  ease of use for the end user** (e.g. back button to jump back to the previous
+  search page and drag & drop of search term to new tabs do not work as
+  expected .. and several more).  We had a lot of long discussions about the
+  *pros v2 cons*:
+
+  - `set HTTP GET method by default
+    <https://github.com/searxng/searxng/pull/3619>`__
+  - `http methods GET & POST
+    <https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22http+methods+GET+%26+POST%22>`__
 
 .. _HTTP headers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
 
 ``default_http_headers`` :
   Set additional HTTP headers, see `#755 <https://github.com/searx/searx/issues/715>`__
-
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -100,8 +100,9 @@ server:
   image_proxy: false
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
-  # POST queries are more secure as they don't show up in history but may cause
-  # problems when using Firefox containers.
+  # POST queries are "more secure!" but are also the source of hard-to-locate
+  # annoyances, which is why GET may be better for end users and their browsers.
+  # see https://github.com/searxng/searxng/pull/3619
   # Is overwritten by ${SEARXNG_METHOD}
   method: "POST"
   default_http_headers:


unixfox:
No, I'm against it. Already explained why here: https://github.com/searxng/searxng/issues/711#issuecomment-1008331605

> Sharing: a search with somebody else becomes more difficult with POST (for the same reason).

`Search URL` section already exist for that. But since you hide it a while ago, people may not find it...

> Bookmarking: a search becomes more difficult with POST. You cannot just press a browser keyboard shortcut Ctrl+D ... no you have to copy the Search URL from the sidebar. This can pose a real struggle for less technically-minded users.

We could have a button to easily add the page as a bookmark: https://buffernow.com/create-bookmark-link-using-javascript/

> Open Link in New Tab: does not work for our search result tabs because they aren't actual links.

What's that? Can you elaborate a bit more?

Bnyro:
@unixfox see https://github.com/searxng/searxng/issues/3590#issue-2364139446

unixfox:
> @unixfox see [#3590 (comment)](https://github.com/searxng/searxng/issues/3590#issue-2364139446)

Ok. It's a very niche feature. That's not a valid reason for migrating EVERYONE to get method.

return42:
> No, I'm against it. Already explained why here: [#711 (comment)](https://github.com/searxng/searxng/issues/711#issuecomment-1008331605)

I think the arguments @unixfox give in the linked comment are valid.  On the other hand, I think the UX arguments are also valid.  I'm torn; 

I think we should leave this PR open for now and not merge it.  The open PR can invite further readers to discuss ... or serve maintainers as a template for a setup of their instance / I switched my instance to GET years ago for the reasons mentioned above and looked the HTTP method in the preferences.  Issue https://github.com/searxng/searxng/issues/3590 has also confirmed that the settings are ideal .. at least for me :-)

ACR-Jeff:
Why not leave it the way it is and more secure for the users, Isn't that what Searxng is all about being secure? I know why some may want GET over POST, But that brings a security issue, I have defaults set for that very reason, To keep it secure for the users, I also have one system that I have GET set, Due to needing the bookmarks, Etc, That can easily be done in the user preferences, Still giving them security and the choice otherwise.

dalf:
Since the project became more popular, the struggle has always been the same: privacy or UX.

The moto says "a privacy-respecting, open metasearch engine", so I would stick to the POST method for the reasons explained by @unixfox 

An idea: a ~wizard~ two buttons to choose between privacy and UX in the preferences (in the first visible tab):
* Enable privacy preferences
* Enable user friendly preferences

With a tooltip, a label or something else to explain the role of each buttons.

They can be displayed and implemented in JS: user without JS won't bother about these buttons especially if the defaults are set to preserve privacy.

An alternative is to implement a real wizard with different steps and details description of each choice. Same it can be implemented in JS. A button on the top right corner can be displayed in the front page when the user don't have any preferences (no cookies).

Quick mock-up:
image

ACR-Jeff:
Privacy is what Searxng has always been about, Yet they still have the option in preferences to change to the alternative, I use it when I need it. But IMO the alternative shouldn't be the default. Stay with what searxng has always stood for.

Aetherinox:
I like @dalf's proposal. POST is honoring the promise of ensuring privacy. Yes, it has drawbacks, but simply introducing a way to educate the user between the two options allows them to choose if they want the privacy benefits, or if they're willing to give up a bit of privacy over a bit more of a user-friendly experience.

Personally, I stick to POST.

omentic:
I strongly prefer GET. Aside from being how conventional search engines operate, I am not convinced that there are any tangible privacy drawbacks: private tabs + browser history are a widely understood concept, and any implications that it does not log to servers are misleading since an instance is under full control of the instance owner (and adding POST logging is trivial).

But I also don't have a stake in the matter! It is well-advertised as a configurable default, and the descriptions of each option's behavior is clear enough, IMO. People checking out the project will wind up on instances not run by the SearXNG team so I don't think that changing the default matters all too much.

return42:
Related: 

- https://github.com/searxng/searxng/discussions/4262#discussioncomment-12005746

return42:
> I like @dalf's proposal. POST is honoring the promise of ensuring privacy. Yes, it has drawbacks, but simply introducing a way to educate the user between the two options allows them to choose if they want the privacy benefits, or if they're willing to give up a bit of privacy over a bit more of a user-friendly experience.

I have now revised the PR, the defaults will not be changed ... but documentation will be added (f4250b6d40) .. is that OK for you all? ... can we merge the PR? .. we need a documentation of the `server.method` settings in any case :-)

unixfox:
@return42 Can you change the PR title accordingly then?

return42:
> @return42 Can you change the PR title accordingly then?

Thanks for reminder / done :+1:

PR #3627, opened 2024-06-29
[mod] libretranslate: add direct link to translation (engine)

cy8aer:
This may help for further translations.

## What does this PR do?

This adds a link to the used libretranslate instance. It uses `from_lang`, `to_lang`, and `query` to transfer the translation parameters.

This can be helpful to do further translations on the libretranslate instance.

Search string: `!lt en-de this is a nice test` - will output the search string in the answer field (as before) and in addition the link to the instance.

## Author's checklist

-

## Related issues

-

### DIFF
--- a/searx/engines/libretranslate.py
+++ b/searx/engines/libretranslate.py
@@ -31,6 +31,7 @@ def request(_query, params):
 
     params['method'] = 'POST'
     params['headers'] = {'Content-Type': 'application/json'}
+    params['req_url'] = request_url
 
     return params
 
@@ -40,7 +41,13 @@ def response(resp):
 
     json_resp = resp.json()
     text = json_resp.get('translatedText')
+
+    from_lang = resp.search_params["from_lang"][1]
+    to_lang = resp.search_params["to_lang"][1]
+    query = resp.search_params["query"]
+    req_url = resp.search_params["req_url"]
+
     if text:
-        results.append({'answer': text})
+        results.append({"answer": text, "url": f"{req_url}/?source={from_lang}&target={to_lang}&q={query}"})
 
     return results


Bnyro (changes requested):
Please undo the formatting changes, they're not related to this PR.

cy8aer:
Rebased.

cy8aer:
Ok, pure patch now (without formatting) though I do not understand why the python CIs do not make problems now :-/

searx/engines/libretranslate.py
```diff
@@ -40,7 +40,17 @@ def response(resp):
...
+    from_lang = resp.search_params["from_lang"][1]
+    to_lang = resp.search_params["to_lang"][1]
+    origin = resp.search_params["query"]
+
     if text:
-        results.append({'answer': text})
+        results.append(
+            {
+                "answer": text,
+                "url": "%s/?source=%s&target=%s&q=%s" % (base_url[0], from_lang, to_lang, origin),
```
return42: `base_url` can be a list or a string / see line 24 in the request function.

cy8aer:
I now use the same mechanism like in the request: Get some random `base_url`  if it is a list, `base_url` else. But this may use another server of the list than in the request of course. Probably better too to distribute load.

cy8aer:
Oups typo in description: random.choice!

Bnyro:
I think it'd be better to set `params['base_url']` in the `request` method and use that base url via `resp.search_params['base_url']` in the `response` method instead of picking an other instance, as the other instance might not be working. It's only safe to say that the instance used for the request did return a proper response and thus probably works.

searx/engines/libretranslate.py
```diff
@@ -40,7 +40,18 @@ def response(resp):
...
+    to_lang = resp.search_params["to_lang"][1]
+    origin = resp.search_params["query"]
+    r_url = random.choice(base_url) if isinstance(base_url, list) else base_url
+
     if text:
-        results.append({'answer': text})
+        results.append(
+            {
+                "answer": text,
+                "url": "%s/?source=%s&target=%s&q=%s" % (r_url, from_lang, to_lang, origin),
```
Bnyro:
```suggestion
                "url": f"{req_url}/?source={from_lang}&target={to_lang}&q={query}",
```

cy8aer:
> I think it'd be better to set `params['base_url']` in the `request` method and use that base url via `resp.search_params['base_url']` in the `response` method instead of picking an other instance, as the other instance might not be working. It's only safe to say that the instance used for the request did return a proper response and thus probably works.

Oh, that makes more sense. I did not know that this persistence is possible.

cy8aer:
I had a typo in the new url and now rebased the whole stuff to clean up.

Bnyro (approved):
Thank you!

PR #3629, opened 2024-07-01
[enh] Allow using multiple autocompleters

allendema:
## What does this PR do?

Allows setting multiple autocompleters.
Get different suggestions at the same time.

1. make test
2. make run
3. enable multiple autocomplete engines in your settings, e.x
```yaml
autocomplete:
  - duckduckgo
  - stract
```
####
Note: the existing string value will still continue to work and will be comma seperated to set more multiple ones.
  

## Author's checklist

Are there any edge cases?
## Related issues
https://github.com/searxng/searxng/issues/1889

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -11,6 +11,7 @@ import lxml
 from httpx import HTTPError
 
 from searx import settings
+from searx.utils import unique
 from searx.engines import (
     engines,
     google,
@@ -252,11 +253,31 @@ backends = {
 }
 
 
-def search_autocomplete(backend_name, query, sxng_locale):
-    backend = backends.get(backend_name)
-    if backend is None:
-        return []
+def search_autocomplete(backend_names, query, sxng_locale):
+
+    enabled_backends = list(unique(backend_names))
+
+    len_enabled_backends = len(enabled_backends)
+
     try:
-        return backend(query, sxng_locale)
+        results = []
+
+        for backend_name in enabled_backends:
+            backend = backends.get(backend_name)
+            if backend is None:
+                # if somehow 'searx.preferences.ValidationException' was not raised
+                continue
+
+            backend_results = backend(query, sxng_locale)
+            if (len_enabled_backends > 2) and (len(backend_results) > 3):
+                # if more than 2 autocompleters: only get the first 3 results from each
+
+                results.extend(backend_results[:3])
+
+            else:
+                results.extend(backend_results)
+
+        return list(unique(results))
+
     except (HTTPError, SearxEngineResponseException):
         return []
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -96,6 +96,10 @@ class MultipleChoiceSetting(Setting):
     """Setting of values which can only come from the given choices"""
 
     def __init__(self, default_value: List[str], choices: Iterable[str], locked=False):
+        # backwards compat for autocomplete setting (was string, now is a list of strings)
+        if isinstance(default_value, str):
+            default_value = [str(val) for val in default_value.split(",")]
+
         super().__init__(default_value, locked)
         self.choices = choices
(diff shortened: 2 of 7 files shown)


searx/autocomplete.py
```diff
@@ -252,11 +253,36 @@ def yandex(query, _lang):
...
+            if (len_enabled_backends > 1) and (len(backend_results) > 3):
+                # if more than 3 autocompleters: only get the first 3 results from each
+
+                results.extend(backend_results[:3:])
+
+            else:
+                results.extend(backend_results)
+
+        random.seed(1)
+        random.shuffle(results)
```
Bnyro: We didn't shuffle results before that change either, I don't think we should start doing that now.

searx/autocomplete.py
```diff
@@ -252,11 +253,36 @@ def yandex(query, _lang):
...
+
```
Bnyro: We should filter out duplicated results here (or use a `set` instead of a `list`).

searx/autocomplete.py
```diff
@@ -252,11 +253,36 @@ def yandex(query, _lang):
...
     try:
-        return backend(query, sxng_locale)
+        results = []
+
+        for backend in enabled_backends:
+            backend_results = backend(query, sxng_locale)
+            if (len_enabled_backends > 1) and (len(backend_results) > 3):
+                # if more than 3 autocompleters: only get the first 3 results from each
+
+                results.extend(backend_results[:3:])
```
Bnyro: Shouldn't this be
```suggestion
                results.extend(backend_results[:3])
```
?

searx/autocomplete.py
allendema: Removed the shuffle but in the future the results must be somehow mixed.

Now the ordering in the settings matters as there can be more engines producing the following suggestions: 
- 3 from autocompleter1, then
- 3 from autocompleter2, then
- 3 from autocompleter3

searx/autocomplete.py
allendema: Yes, the result is the same but thats better.

searx/autocomplete.py
allendema: Yes, I changed some things.

searx/autocomplete.py
Bnyro: In the end, we should probably do
- result 1 from autocompleter 1
- result 1 from autocompleter 2
- result 1 from autocompleter 3
- ...
- result 2 from autocompler 1
- result 2 from autocompleter 2
- ...

with no more than 10 results/suggestions in total returned to the user.

At least that's the behavior I'd personally like the most as a user.

0xhtml:
Some code to merge the results in a better order. e.g. [A1, A2, A3], [B1, B2, B3] -> [A1, B1, A2, B2, A3, B3].
```python
from collections import deque
from typing import Sequence, Iterable, Generator, TypeVar
T = TypeVar("T")
def merge(iterables: Sequence[Iterable[T]]) -> Generator[T, None, None]:
    fifo = deque(iter(iterable) for iterable in iterables)
    while fifo:
        iterator = fifo.pop()
        try:
            yield next(iterator)
        except StopIteration:
            continue
        fifo.appendleft(iterator)
```

return42:
I'm not sure its a good idea to enable more than one auto-completer.  It will spread what the user types into the search field to more than really needed origin auto-completer / beside data spread we should consider the additional response times.  Further every autocompleter service has its own ranking rules/concept .. if we enable by example "google" and "duckduckgo", we mix the concept from both providers. If one provider provides e.g. a "language" like "de" or "fr" and the other provider did not support languages (e.g. returns "en" completions) .. not sure how a ranking should work in those scenarios.

-----------

I'll turn the PR into a DRAFT for now ... if we find strong arguments for this PR we can merge this PR

PR #3630, opened 2024-07-01
[fix] html.unescape stract autocomplete suggestions

allendema:
## What does this PR do?

Unescape html entities for better readability of autocomplete suggestions.

1. `make format.python`
2. `make test`
3. `make run`
4. set stract as autocompleter and search for ```who is``` or ```what is```.

## Author's checklist

Is more unescaping needed?
## Related issues
N/A

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -5,6 +5,7 @@
 # pylint: disable=use-dict-literal
 
 import json
+import html
 from urllib.parse import urlencode, quote_plus
 
 import lxml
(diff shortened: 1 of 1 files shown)


searx/autocomplete.py
```diff
@@ -162,7 +163,7 @@ def stract(query, _lang):
     if not resp.ok:
         return []
 
-    return [suggestion['raw'] for suggestion in resp.json()]
+    return [html.unescape(suggestion['raw']) for suggestion in resp.json()]
```
Bnyro, line 166: We already have a method for that:
```suggestion
    return [html_to_text(suggestion['raw']) for suggestion in resp.json()]
```
see https://github.com/searxng/searxng/blob/39aaac40d63d53555b7fc12d56f0825c85ecf567/searx/utils.py#L149

Bnyro (approved):

return42 (approved):
> 4\. set stract as autocompleter and search for `who is` or `what is`.

I used `count` ..

before:

image

and after your patch:

image

Thanks a lot for this patch :+1:

PR #3631, opened 2024-07-02
[fix] engine yacy: update list of base URLs

return42:
https://search.lomig.me
  Poor results / tested `!yacy :en hello` and got zero results

https://yacy.ecosys.eu
  Slow response (> 6sec for trivial search terms)

https://search.webproject.link
  Dead instance / URL offline

## we need more stable yacy instances 

.. suggestions are welcome

-----

Related:

- https://github.com/searxng/searxng/issues/3428
- https://github.com/searxng/searxng/pull/3472

### DIFF
--- a/searx/engines/yacy.py
+++ b/searx/engines/yacy.py
@@ -118,6 +118,8 @@ def _base_url() -> str:
     url = engines['yacy'].base_url  # type: ignore
     if isinstance(url, list):
         url = random.choice(url)
+    if url.endswith("/"):
+        url = url[:-1]
     return url
 
 
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -2125,21 +2125,25 @@ engines:
     disabled: true
 
   - name: yacy
+    # https://docs.searxng.org/dev/engines/online/yacy.html
     engine: yacy
     categories: general
     search_type: text
     base_url:
       - https://yacy.searchlab.eu
-      - https://search.lomig.me
-      - https://yacy.ecosys.eu
-      - https://search.webproject.link
+      # see https://github.com/searxng/searxng/pull/3631#issuecomment-2240903027
+      # - https://search.kyun.li
+      # - https://yacy.securecomcorp.eu
+      # - https://yacy.myserv.ca
+      # - https://yacy.nsupdate.info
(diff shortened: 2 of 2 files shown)


Bnyro (approved):
I can confirm what you described above about the 3 instances here.

As yacy.searchlab.eu is the "official instance", it's probably the easiest to keep it the instance chosen by default (even though it had issues with downtimes as well previously).

It should probably be up to instance maintainers to properly configure Yacy in case they have their own or better instances available.

toyboatcash:
https://search.kyun.li/
https://yacy.securecomcorp.eu/
https://yacy.myserv.ca/
https://yacy.nsupdate.info/
https://yacy.electroncash.de/
http://sokrates.homeunix.net:6060/
http://4h8.de:8090/

As for getting more instances can we use https://yacy.nsupdate.info/Network.html?page=1&maxCount=1000  just  add ?page=1&maxCount=1000  to the instance. Look under Info, ipv4 https://yacy.nsupdate.info/env/grafics/NodeDisqualifiedIPv6.gif and ipv6 https://yacy.nsupdate.info/env/grafics/NodeDisqualifiedIPv6.gif you will get a ip and port for different instances. Some of the IPs are also on domains.  I hope this helps here are some examples below.

ipv6 
http://[2a01:e34:ec79:ad90:808f:bdff:fe17:c41e]:8090/
http://[2a02:8071:6072:9f20:44f5:6732:2e68:264c]:8092/
ipv4
http://87.182.195.104:8090/
http://93.232.220.112:8090/
http://138.201.222.58:8096/
http://178.63.23.24:8092/
http://23.94.86.140:8090/

return42:
@toyboatcash thanks a lot for your helping comment .. I squashed your suggestions and a link to your comment into the patch / see [diff](https://github.com/searxng/searxng/compare/a8133a5f05aaf46a20fefc6f7ced4678fbcd24a5..1d1e039741857bdccb7a3615fc4bce14fefdc5e7)

> It should probably be up to instance maintainers to properly configure Yacy in case they have their own or better instances available.

I think you're right about that ... with the comments that have now been added, every operator should be able to maintain their own list ... that's all we can do here in the defaults / I'm going to merge PR.

PR #3634, opened 2024-07-02
[feat] engine: implementation of alpine linux packages

Bnyro:
## What does this PR do?
- Implement a new engine, alpine linux package search

- Alpine Linux is very commonly used for servers and docker images
- using the package search directly on their website is not that intuitive - you need to put `*` before and after the query in order to do proper "contains" search

- !alp firefox

### DIFF
--- a/docs/dev/engines/online/alpinelinux.rst
+++ b/docs/dev/engines/online/alpinelinux.rst
@@ -1,0 +1,13 @@
+.. _alpinelinux engine:
+
+=====================
+Alpine Linux Packages
+=====================
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.alpinelinux
+   :members:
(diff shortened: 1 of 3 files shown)


return42:
.. the engine will find many friends among us nerds :-)

I had an issue when there are no packages listed (e.g. search for `!alp searxng`) .. I patched the response function to ignore table rows that do not have the needed columns.  Further I made some marginal modifications to the doc-strings and added them to the online documentation. --> [diff](https://github.com/searxng/searxng/compare/afcff48656d37722e772b51a1406719d927ff928..e42d0ef71d379ed8142781696df120ee8541c3b8)

What do you think about kernel architecture, should we add a regexp to grep `arch:xyz` from the search term?

Bnyro:
> What do you think about kernel architecture, should we add a regexp to grep arch:xyz from the search term?
>
I think that's a nice idea, yeah :+1:

return42 (approved):
Sorry, haven't realized that you have implemented *arch* feature in search query two weeks ago .. works like a charm / thanks :+1:

PR #3635, opened 2024-07-02
[fix] gentoo: use mediawiki engine

Bnyro:
## What does this PR do?
- modifies the gentoo engine config to use the mediawiki base engine

- Gentoo is currently broken and the engine hasn't really been maintained in the past, so I think this is the best and simplest fix

- !ge chrony
- !ge busybox


return42:
I don't know the relevance of translations .. should we consider to support languages?

I found some pages that have translations:

- https://wiki.gentoo.org/wiki/Main_Page --> https://wiki.gentoo.org/wiki/Handbook:Main_Page/de
- https://wiki.gentoo.org/wiki/FAQ --> https://wiki.gentoo.org/wiki/FAQ/de

Further there seems some activity:

- https://wiki.gentoo.org/wiki/Special:ActiveLanguages

but TBH I don't know how a user can search in languages nor is it clear to me how a user can navigate e.g. from https://wiki.gentoo.org/wiki/Main_Page  to the DE translation https://wiki.gentoo.org/wiki/Handbook:Main_Page/de (note: there is no language menu in the origin EN and in the URL for the DE translation there is an additional `Handbook:`.

Bnyro:
> I don't know the relevance of translations .. should we consider to support languages?
>
I don't think translation support is necessary in this case - Gentoo is mainly for "advanced" users anyways, so that's probably not their first Linux distro and thus they should already be quite familiar with reading English documentation (most troubleshooting guides you find in the internet about Linux are English anyways).

Apart from that I don't think many people use that engine, otherwise someone would have reported that it's broken way earlier.

return42 (approved):

PR #3647, opened 2024-07-10
Flush flag icon to the right in the search_language select option values

### LINKED ISSUE #1885: Cycle through countries when pressing the key of the first letter

Nearly all websites i visit allow me to press the first letter of what i'm looking for in the dropout and jump to it. I frequently need to change between English and Swedish results and having to use the scroll wheal to find and change the country search location is annoying

As an example when you have the country search dropdown on, pressing S will jump to countries in that list starting with S, containing to press S scrolls down the list until the desired country is highlighted

Using google since they search by the language your type in or brave search where they have a search box to find a country easier (does not in brave too but they have a search bar at least)
**Additional context**

[Screen recording 2022-10-25 10.41.20 AM.webm](https://user-images.githubusercontent.com/34938708/197726965-51bebea8-2f3b-4cbc-bd14-0817567081dc.webm)
An example in another unrelated project uptime kuma, pressing the first letter cycles through those options. Optionally would allow you to select a language if you are hovering your mouse over the dropdown

### LINKED ISSUE #3645: Do not prefix language select with icons.
When selecting the language to search, do not prefix each select option with an icon.

Currently, if I click the select menu and start typing e.g. "sv" for "Svenska", the select list won't be able to select sweden because the icon is in the way. If you get rid of the icons, people could use the keyboard to select what language they want used in the search by typing the keyboard. As it is now, the user has to use their mouse and/or arrow keys to scroll down the list of every country in the world. It doesn't make any sense.

glanham-jr:
## What does this PR do?

Move the flag icon next to the language to the right in the option value for languages.

If the flag icon is first, it prevents easily searching the select list on the keyboard. By moving the icon fully to the right, this will enable a user to search the select list.

1. Run SearXNG locally
2. Navigate to the preferences page
3. Click on the select option for `Search language`
4. Validate the icons are to the right instead of to the left

## Author's checklist
n/a

## Related issues

- closes: #3645
- closes: #1885

## Screenshots

image

### DIFF
--- a/searx/templates/simple/preferences/language.html
+++ b/searx/templates/simple/preferences/language.html
@@ -13,9 +13,9 @@
       {% for sxng_tag,lang_name,country_name,english_name,flag in sxng_locales | sort(attribute=1) -%}
         <option value="{{ sxng_tag }}"
                 {%- if sxng_tag == current_language %} selected="selected" {%- endif -%}>
-                {%- if flag -%}{{ flag }} {% endif -%}
                 {{ lang_name }}{%- if country_name -%}-{{ country_name }}{%- endif -%}
-                {{- ' ' -}}[{{sxng_tag}}]{{- '' -}}
+                {{- ' ' -}}[{{sxng_tag}}]{{- ' ' -}}
+                {%- if flag -%}{{ flag }}{%- endif -%}
         </option>
       {%- endfor -%}
     </select>{{- '' -}}
--- a/searx/templates/simple/filters/languages.html
+++ b/searx/templates/simple/filters/languages.html
@@ -1,12 +1,19 @@
 <select class="language" id="language" name="language" aria-label="{{ _('Search language') }}">{{- '' -}}
-	<option value="all" {% if current_language == 'all' %}selected="selected"{% endif %}>{{ _('Default language') }} [all]</option>
-	<option value="auto" {% if current_language == 'auto' %}selected="selected"{% endif %}>
-		{{- _('Auto-detect') -}}
-		{%- if current_language == 'auto' %} ({{ search_language }}){%- endif -%}
-	</option>
-	{%- for sxng_tag,lang_name,country_name,english_name,flag in sxng_locales | sort(attribute=1) -%}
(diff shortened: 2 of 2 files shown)


searx/templates/simple/preferences/language.html
```diff
@@ -13,9 +13,9 @@
       {% for sxng_tag,lang_name,country_name,english_name,flag in sxng_locales | sort(attribute=1) -%}
         <option value="{{ sxng_tag }}"
                 {%- if sxng_tag == current_language %} selected="selected" {%- endif -%}>
-                {%- if flag -%}{{ flag }} {% endif -%}
                 {{ lang_name }}{%- if country_name -%}-{{ country_name }}{%- endif -%}
-                {{- ' ' -}}[{{sxng_tag}}]{{- '' -}}
+                {{- ' ' -}}[{{sxng_tag}}]{{- ' ' -}}
```
glanham-jr, line 17: Currently if the flag doesn't exist, an unecessary extra space will be added at the end. I can fix this, but also I doubt it will really affect UI/UX at all.

return42 (approved):
The language menu from the search page was missed in this PR .. I added it and rebased this branch (see [diff](https://github.com/searxng/searxng/compare/5112805b3b40a097390aa0a53c58cb61dcca1100..88482c34a12138fea625809a1bcb01948a1b89b5)).

One question resits on my side: how can I select the language from them options by keyboard .. what ever I have tried, I first have to klick into the language with the mouse first and then I can type to select my preferred language .. is the mouse click really needed? .. what trick can I use to avoid the mouse click?

Bnyro (approved):
Neat, I've also linked issue https://github.com/searxng/searxng/issues/1885 since it's kinda related

Bnyro:
> One question resits on my side: how can I select the language from them options by keyboard .. what ever I have tried, I first have to klick into the language with the mouse first and then I can type to select my preferred language .. is the mouse click really needed? .. what trick can I use to avoid the mouse click?
>
All I can come up with is pressing tab until the language selector `select` is highlighted and then press enter to open the language selection menu - we could also capture the keyboard events directly and open the select menu at a specific index - but I remember I tried to implement that and it didn't really work well.

PR #3656, opened 2024-07-14
Dockerfile: use Alpine 3.20 & Python 3.12

dalf:
## What does this PR do?

Use Alpine 3.20 and Python 3.12

EOL of 3.19 is [2025-11-01](https://alpinelinux.org/releases/), so there is time (only security issues are fixed, but we never has specific issues related to Alpine).

Up to date dependency.
Slightly faster Python 3.12.

## Author's checklist

We might want to add a `linux/riscv64` docker image ?

Ref: https://github.com/searxng/searxng/blob/5468d97d39d47701e652db88920eb40068312152/manage#L190

Note: I know the branch should have been in my fork.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.19
+FROM alpine:3.20
 ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
 EXPOSE 8080
 VOLUME /etc/searxng


Bnyro:
> We might want to add a linux/riscv64 docker image ?
>
I personally think that's a good idea as Riscv64 is gaining some popularity recently.

Bnyro (approved):

return42 (approved):
> Note: I know the branch should have been in my fork.

:smile:

return42:
> I personally think that's a good idea as Riscv64 is gaining some popularity recently.

I knew its OT here, but while we re talking about:  which SBC can you recommend?

I have read debian officially supports RISC-V 64 since mid 2023 ... what are your experiences with Debian or other distributions? 

I also read that alpine 3.20 is the first version that officially supports RISC-V 64 .. any experience?

... and alpine 3.2 has replaced Redis with Valkey :-)

Bnyro:
> I knew its OT here, but while we re talking about: which SBC can you recommend?
>
I personally would probably go with something like https://pine64.org/devices/star64/ for experimenting a bit, but I haven't yet used any RISC-V machine, so I can't say anything about how good it works with Linux in reality. I don't know what hardware specs you are looking for though, there are probably many more decent devices.

> I have read debian officially supports RISC-V 64 since mid 2023 ... what are your experiences with Debian or other distributions?
>
I personally don't really like Debian's release scheme - a large update every two years causes that you're usually one to two years behind with updates, so you can't benefit from new features or bug fixes that well (security fixes are usually backported so that you don't miss them, but that's all of an update you get with Debian).

> I also read that alpine 3.20 is the first version that officially supports RISC-V 64 .. any experience?
>
I'm running Alpine Linux on all my servers and SBCs, so I'm quite a fan of it, especially for machines that are not connected to a monitor and just used via SSH only :)
It's very stable in one hand, but provides new updates frequently at the same time. Apart from that the repositories contain any command line utility you could dream of - so you don't really have to add extra repositories like in Debian for example.
I wouldn't recommend using it on a desktop though, from my experience some desktop apps don't go well with musl libc.

PR #3663, opened 2024-07-19
[upd] upgrade: github actions / to ubuntu: 24.04

return42:
```
- ubuntu-20.04            --> ubuntu-24.04
```

-----

Related:

- [GitHub Actions Runner Images](https://github.com/actions/runner-images)

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -7,7 +7,7 @@ on:  # yamllint disable-line rule:truthy
 jobs:
   checker:
     name: Checker
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,10 +12,10 @@ permissions:
 jobs:
   python:
     name: Python ${{ matrix.python-version }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
-        os: [ubuntu-20.04]
+        os: [ubuntu-24.04]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - name: Checkout
@@ -48,7 +48,7 @@ jobs:
 
   themes:
     name: Themes
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -67,7 +67,7 @@ jobs:
             ./local
             ./.nvm
             ./node_modules
-          key: python-ubuntu-20.04-3.12-${{ hashFiles('requirements*.txt', 'setup.py','.nvmrc', 'package.json') }}
+          key: python-ubuntu-24.04-3.12-${{ hashFiles('requirements*.txt', 'setup.py','.nvmrc', 'package.json') }}
       - name: Install node dependencies
         run: make V=1 node.env
       - name: Build themes
@@ -75,7 +75,7 @@ jobs:
 
   documentation:
     name: Documentation
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     permissions:
       contents: write  # for JamesIves/github-pages-deploy-action to push changes in repo
     steps:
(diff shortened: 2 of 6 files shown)


return42:
In ubu-24.04 the `apt install firefox`:

https://github.com/searxng/searxng/blob/ffde2563645f074c52e4a1bf9fb66948e74e7797/.github/workflows/integration.yml#L20-L26

does no longer install the firefox apt package with binaries (`/usr/lib/firefox/firefox`), it installs the transitional package "firefox -> firefox snap" (`/snap/bin/firefox`).

The CI now fails at ...

https://github.com/searxng/searxng/blob/ffde2563645f074c52e4a1bf9fb66948e74e7797/tests/robot/__main__.py#L58

with error message:

     selenium.common.exceptions.InvalidArgumentException: Message: binary is not a Firefox executable

I suspect that the Firefox from the snap cannot be started in batch from the CI for some reason .. 

    /snap/bin/firefox: symbolic link to /usr/bin/snap

but I also have no idea how we can solve the problem .. probably we need some quirks again .. maybe we should also get away from this Ubuntu in general .. it smells more and more worse.

allendema:
Can we point to installed geckdriver [directly](https://github.com/searxng/searxng/blob/98c73010f1f6333d37cd75a75ea924704ba121a2/manage#L251)?
https://splinter.readthedocs.io/en/latest/drivers/firefox.html#custom-executable-path

Maybe pass the snap executable like [here](https://github.com/noyainrain/room/commit/9cef7c22ded9733a54fa7233d98f2e3e9d49a701):
https://splinter.readthedocs.io/en/latest/drivers/firefox.html#custom-binary-path

_All because canonical installs snap packages when snap is not explicitly used :clown_face:._

return42:
> Can we point to installed geckdriver [directly](https://github.com/searxng/searxng/blob/98c73010f1f6333d37cd75a75ea924704ba121a2/manage#L251)?

@allendema thanks for pointing out .. I tried it in 555d344201709 but without luck .. now I read: ["Do not use container-packaged Firefox builds with geckodriver."](https://firefox-source-docs.mozilla.org/testing/geckodriver/Usage.html#running-firefox-in-a-container-based-package)  :thinking:

dalf:
I guess the issue is firefox is not installed in a "default" path.

That's for GitHub workflow, no need to make an generic installer. 

Why not replace `apt install -y firefox` by :

```sh
sudo install -d -m 0755 /etc/apt/keyrings
wget -q https://packages.mozilla.org/apt/repo-signing-key.gpg -O- | sudo tee /etc/apt/keyrings/packages.mozilla.org.asc > /dev/null
gpg -n -q --import --import-options import-show /etc/apt/keyrings/packages.mozilla.org.asc | awk '/pub/{getline; gsub(/^ +| +$/,""); if($0 == "35BAA0B33E9EB396F59CA838C0BA5CE6DC6315A3") print "\nThe key fingerprint matches ("$0").\n"; else print "\nVerification failed: the fingerprint ("$0") does not match the expected one.\n"}'
echo "deb [signed-by=/etc/apt/keyrings/packages.mozilla.org.asc] https://packages.mozilla.org/apt mozilla main" | sudo tee -a /etc/apt/sources.list.d/mozilla.list > /dev/null
echo '
Package: *
Pin: origin packages.mozilla.org
Pin-Priority: 1000
' | sudo tee /etc/apt/preferences.d/mozilla 
sudo apt-get update && sudo apt-get install -y firefox
```

Copied / paste from https://support.mozilla.org/en-US/kb/install-firefox-linux#w_install-firefox-deb-package-for-debian-based-distributions-recommended

Or if we want to keep the GH workflow minimal, put these command in ./manage with a command named `github-prepare-ci` (or similar) ?

return42:
> Why not replace `apt install -y firefox` by :

I will give it a try / thanks a lot for suggestion :+1:

return42:
That was quite a mess, I built an ubu2404 image and tested in it:

1. if gecko is installed, you don't need firefox
2. on ubuntu you are not allowed to install firefox because it comes from the snap packages and since the "snap-firefox" is in the PATH variable, it will break with what selenium expects.

Then there is the fact that the github action runner has already installed firefox and the geckodriver --> https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#browsers-and-drivers

So the solution is quite simple: we must not install firefox via apt and the generic installation of the geckodriver, as we have done so far, is superfluous / at least on the image from the github runner.

--------

I removed the cache action .. this led to further confusion, because sometimes there was something old in there ... I also don't have the impression that the cache really brings great advantages (on the contrary, we often have segmentation faults in the gh actions and I suspect they result from the caches )... or have I missed something?

PR #3669, opened 2024-07-23
[feat] engine: implementation of geizhals.de

Bnyro:
## What does this PR do?
- add support for geizhals.de

- because geizhals.de can basically "replace" all other shopping search engines in Germany

- `!geizhals sort:asc monitor`
- `!geizhals tastatur`

### DIFF
--- a/searx/engines/geizhals.py
+++ b/searx/engines/geizhals.py
@@ -1,0 +1,97 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Geizhals is a German website to compare the price of a product on the
+most common German shopping sites and find the lowest price.
+
+The sorting of the search results can be influenced by the following additions
+to the search term:
+
+``asc`` or ``price``
+  To sort by price in ascending order.
+
+``desc``
+  To sort by price in descending order.
+
+"""
+
+import re
+
+from urllib.parse import urlencode
+from lxml import html
+
+from searx.utils import eval_xpath, eval_xpath_list, extract_text
+
+about = {
+    'website': 'https://geizhals.de',
+    'wikidata_id': 'Q15977657',
+    'use_official_api': False,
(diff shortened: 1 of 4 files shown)


allendema:
Nice.
Used to have an xpath engine for this but it was very simple and [not ported over](https://github.com/allendema/searx/commit/eb9f70b86823b51dccfb888509c848b567885c56).

searx/engines/geizhals.py
```diff
@@ -0,0 +1,86 @@
...
+    'results': 'HTML',
+    'language': 'de',
+}
+paging = True
+categories = ['shopping']
+
+base_url = "https://geizhals.de"
+sort_order = 'relevance'
+
+SORT_RE = re.compile("sort:(\w+)")
```
return42: String constant might be missing an `r` prefix. (anomalous-backslash-in-string) --> `r"sort:(\w+)`

searx/engines/geizhals.py
```diff
@@ -0,0 +1,87 @@
...
+base_url = "https://geizhals.de"
+sort_order = 'relevance'
+
+SORT_RE = re.compile(r"sort:(\w+)")
+sort_order_map = {
+    'relevance': None,
+    'price': 'p',
+    'asc': 'p',
+    'desc': '-p',
+}
```
return42, line 44: Whats the difference of `price` and `asc` in a search term?

Bnyro: It's the same, it's just for convenience so thay users can use both

return42 (approved):
Thanks a lot for adding geizhals to SearXNG / I often use this page while I'm on shopping :-)

I slightly modified your commit to add the doc strings to the online documentation and BTW: I fixed an issue in the products.html in c3cc660ea.

PR #3672, opened 2024-07-24
[fix] Switch values of binary and decimal prefixes

basxto:
## What does this PR do?

This switches the values of binary and decimal byte prefixes. They were the wrong way around unless I’m misunderstanding the purpose of that dictionary.

Ti**bi**byte (TiB) has a **bi**nary prefix.

I also added KB, which is unambiguously binary due to SI prefix kilo using a lower case k. I added kB as well, which is theoretically possible and likely not used irl.

MB, GB and TB are generally ambiguous, they are the normal decimal prefixes but were customarily used with a binary meaning. The best value depends on the main use case:
* storage devices: decimal is safer due to free space being underestimated in the worst case
* file sizes: binary is safer due to required space being overestimated in the worst case

It could misinterpret values.


return42 (changes requested):
Thanks a lot for pointing out .. I haven't seen it before your patch but this dictionary is a left over from: 

- https://github.com/searxng/searxng/pull/3570

in this PR we removed the old implementations but forgot to delete this `_STORAGE_UNIT_VALUE` .. can you please delete this dict .. it is superfluous and as you already wrote also wrong.

PR #3697, opened 2024-07-29
fix[google]: Strip bubble text from answers

0xhtml:
## What does this PR do?

Google underlines words inside of answers that can be clicked to show additional definitions (bubbles). These definitions inside the answer were not correctly handled and ended up in the middle of the answer text. With this fix, the extra definitions are stripped from the answer shown by the frontend.

The answer text contains undesired text in the middle of the sentences.

Example query “distance earth to sun” with language set to en.

## Author's checklist

N/A

## Related issues

N/A

### After

~(Don't mind the different URL in the corner. That's just because I used a public instance for the before screenshot and the URL for the answer happened to be slightly different.)~

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -338,7 +338,7 @@ def response(resp):
             bubble.drop_tree()
         results.append(
             {
-                'answer': item.xpath("normalize-space()"),
+                'answer': extract_text(item),
                 'url': (eval_xpath(item, '../..//a/@href') + [None])[0],
             }
         )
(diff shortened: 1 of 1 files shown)


0xhtml:
Update: The URL provided before this fix is actually wrong! The code takes the first link, which is the one inside the bubble (science.nasa.gov/neptune…). Instead of the correct link below the answer text. As the fix removes the bubbles from the DOM this is actually also fixed by this PR.

return42 (approved):
Perfect work, thanks a lot for your PR fixing a long standing issue :+1:

Issue #3724, opened 2024-08-10
[bug] right to left url spacing is broken


vojkovic:
While working on https://github.com/privau/searxng/issues/36, I noticed something strange in the spacing of urls in right-to-left languages.

See screenshot below:

![bug](https://github.com/user-attachments/assets/b81f2958-7a1e-4fad-a3f9-6dd885951c2c)

This is due to: https://github.com/searxng/searxng/blob/f1c05e7c16e59a8d5d98640bc057958ea268cddf/searx/static/themes/simple/src/less/style.less#L251-L255

This only happens when the second element in the url, (.url_i2) is empty. (i.e. there is no path attached to the result).

I've experimented with a few different changes to the css without luck.

This css can't be simply removed as it has an important UI purpose in separating the host and path.

return42:
I have no idea what it feels like for natural RTL speakers ... which might be better, but may we should order the `url_o1` and `url_o2` from RTL .. at least it is what google does:

![image](https://github.com/user-attachments/assets/11490f23-1f13-4235-9f3d-8b2639102cec)

to compare, same result item in SearXNG with LTR order of the URL parts:

![image](https://github.com/user-attachments/assets/5246e8ed-078b-44b8-8b6d-98357fb2adbd)

return42:
OT: I really like your Macchiato theme style / good contrast, clean design ..

![image](https://github.com/user-attachments/assets/f484ca67-bc65-4bba-9451-ec5afb8f893a)

suggest that one day we can integrate this into the SearXNG core .. and also the theme from @mrpaulblack :-)

alr86:
> I have no idea what it feels like for natural RTL speakers ... which might be better, but may we should order the `url_o1` and `url_o2` from RTL .. at least it is what google does:

This is done according to the "Visual Hierarchy", and in my own experience as a user with the local RTL language, it is not an interesting experience to see the logo of the site first, then to check the URL on the left, and finally to see the title and description.

PR #3727, opened 2024-08-11
[feat] add favicons to result urls

### LINKED ISSUE #1014: [Theme] Display the favicons of the result links.
The one thing I miss in SearxNG is showing a site's favicon next to its link in the results page. It's a tremendously helpful feature when you're quickly scanning the results and trying to determine what's what.

I'd like it if the site's favicon were displayed next to each result.

Instead of making this behavior the default, perhaps a toggle in the settings could be implemented.

**Additional context**
AFAICT, this was briefly discussed in #504 (in the OP: "..._Also, I believe someone proposed adding the website icons next to the results, and in this case, we may add them to the left of the URL like DDG does._") but was seemingly forgotten/fallen out of view as the discussion in that thread shifted to how the links themselves are displayed, which resulted in #510.

vojkovic:
## What does this PR do?

Adds favicons near the urls in results.

image
image
image
image
(Spacing is broken on RTL mode due to https://github.com/searxng/searxng/issues/3724)

This works very similar to the implementation of the autocomplete backends.

You can also turn off favicons completely in results by setting it to none. I think some optimisations to my css and templates can be made for rtl mode.

After my tests, I believe that caching on the client-side is decent enough, so I've not implemented any cache to SearXNG. Let me know your feedback.

You can see the changes live already on https://priv.au, although this also has my theme patches.

## Related issues

Closes: https://github.com/searxng/searxng/issues/1014 (after 2 years!)

- https://github.com/searxng/searxng/issues/1014#issuecomment-1079044776


return42:
@vojkovic thanks a lot for your work on this task :+1:  .. the CI reports some issues you can fix by `make format.python` further its recommended to run `make test`before commit / [Development Quickstart](https://docs.searxng.org/dev/quickstart.html)

vojkovic:
> @vojkovic thanks a lot for your work on this task 👍 .. the CI reports some issues you can fix by `make format.python` further its recommended to run `make test`before commit / [Development Quickstart](https://docs.searxng.org/dev/quickstart.html)

Should be good now 👍

return42:
Yeah, thanks .. we separate the source code modifications of the static/themes from the build of the static files .. for this we use `make static.build.commit`, the workflow is described in the [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) .. can you separate the commits .. if you have a doubt about how, I can do it for you in my review .. if it's okay with you if I do a force push on your feature branch .. which in this case is your master branch.

For a PR it would be better to set up a feature branch ... but now this PR is already in your master and I hope you don't need the master for other purposes than this PR.

vojkovic:
Should have read the guide... - Should be alright now. Not using the master branch for anything else, but will keep that in mind for future commits. :)

return42:
Today I had my first look at your PR and it's a pleasure to see how you've immersed yourself in the topic and the development of SearXNG .. many thanks for this really excellent work :+1: 

We will be able to simplify a few details and I already have solutions in mind on how we can implement the whole thing in a more resource efficient way (we should reduce the incoming and outgoing request/responses cycles on the server side).

But before I start I will have one or the other question ... please do not change the implementations after my questions, we can do that together in the end ... or I will then make concrete suggestions.

searx/webapp.py
```diff
@@ -866,6 +888,42 @@ def autocompleter():
...
+
+    # return 404 if the favicon is not found
+    if not resp:
+        theme = request.preferences.get_value("theme")
+        # return favicon from /static/themes/simple/img/empty_favicon.svg
+        # we can't rely on an onerror event in the img tag to display a default favicon as this violates the CSP.
+        # using redirect to save network bandwidth (user will have this location cached).
+        return redirect(url_for('static', filename='themes/' + theme + '/img/empty_favicon.svg'))
+
+    # will always return a PNG image
```
return42: OK, here is my first question: I don't think we should assume that all favicons are PNGs .. when I try https://f1.allesedv.com/32/google.com for example, I see that it is an `image/x-icon` mime-type ... or do you see it differently than I do?

vojkovic: Yup - it's an x-icon on allesedv. It still works so it slipped by me. Maybe we can implement a way of just copying the media type of the backend.

vojkovic:
I realised - I've forgot to add any HTTP headers to cache the image on the client side.

searx/static/themes/simple/img/empty_favicon.svg
return42, line 1: Next question :-) .. from where does this symbol (SVG) comes from / who is the creator, do we have the rights to use it or are there any copyrights? 

Does this symbol have a special meaning?

vojkovic: Not really - it's just one of the first globe icons I saw. It's MIT licenced, from https://www.svgrepo.com/svg/507325/globe-alt

searx/static/themes/simple/src/less/search.less
```diff
@@ -378,3 +378,12 @@ html.no-js #clear_search.hide_if_nojs {
 #categories_container {
   position: relative;
 }
+
+.favicon img {
+  height: 1.8rem;
+  width: 1.8rem;
+  border-radius: 20%;
+  background-color: #ddd;
```
return42: Did you have any special reason to set the background color? Some favicons have a transparency; I personally would prefer to leave the background of the `` transparent so that the background of the theme shines through the favicon. We already have a border that provides a visual demarcation.  In your opinion, would there be anything against leaving the background of the `` transparent and deleting the background-color?

vojkovic: Yeah, there's a good reason to have a white background colour.

When first testing the feature out, I had your idea as well. It's just some icons aren't made for no background, see:
nobackground
vs:
withbackground

return42: Yeah, makes sense .. thanks for clarify :+1:

searx/webapp.py
```diff
@@ -297,6 +298,24 @@ def morty_proxify(url: str):
...
+    if not resolver:
+        return ""
+
+    # check resolver is valid
+    if resolver not in favicon_backends:
+        return ""
+
+    h = new_hmac(settings['server']['secret_key'], url.encode())
+
+    return '{0}?{1}'.format(url_for('favicon_proxy'), urlencode(dict(q=url.encode(), h=h)))
```
Bnyro:
```suggestion
    return f"{url_for('favicon_proxy')}?{urlencode(dict(q=url.encode(), h=h)))}"
```

return42: @Bnyro thanks for your review, but please wait a little bit with this PR / I'm in review with this PR .. I modified a lot .. I will need some more time to finish my rework .. Then you can make a review.

searx/webapp.py
```diff
@@ -866,6 +888,42 @@ def autocompleter():
...
+        return '', 400
+
+    resolver = request.preferences.get_value('favicon_resolver')
+
+    # check if the favicon resolver is valid
+    if not resolver or resolver not in favicon_backends:
+        return '', 400
+
+    # parse query
+    raw_text_query = RawTextQuery(url, [])
```
Bnyro: I don't think using a `RawTextQuery` makes sense here since we're dealing with an fqdn, not a user query input?

searx/templates/simple/macros.html
```diff
@@ -21,9 +21,29 @@
 {% macro result_header(result, favicons, image_proxify) -%}
 <article class="result {% if result['template'] %}result-{{ result.template|replace('.html', '') }}{% else %}result-default{% endif %} {% if result['category'] %}category-{{ result['category'] }}{% endif %}{% for e in result.engines %} {{ e }}{% endfor %}">
   {{- result_open_link(result.url, "url_wrapper") -}}
+  {% if not rtl %}
```
Bnyro: I don't really like the duplicated code for rtl/ltr. Can't we just do that via CSS by changing the flex direction?

Bnyro:
Thanks for this PR @vojkovic, this looks awesome!

PS: I'm already thinking about writing my own small self-hostable favicon cache/proxy that could be used with these changes here to not rely on any of the big providers (or include one that's opt-in in SearXNG, although I think a separated project would be better for this).

searx/static/themes/simple/img/empty_favicon.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24">
+  <path fill="#fff" d="M0 0h24v24H0z"/>
```
return42: Here we can delete the white background in the SVG icon (transparent). As mentioned in https://github.com/searxng/searxng/pull/3727#discussion_r1718513261, the `` background in the theme is already designed for transparent icons.

vojkovic: This was an intentional design decision by me, although it's up to your preference, but I prefer this icon as a bright white (#fff) rather than the dull white of the regular background (#ddd).

return42:
I can't edit (git push to) the PR .. Did you blocked this branch or is it because it is the master branch? For git push I missed this option:

image

I did a lot of work the last days and my intention was to force push the additional commits from my WIP-3727 branch:

- https://github.com/return42/searxng/commits/WIP-3727/

vojkovic:
Nice changes - Weird you can't push to it, maybe because it's from my org... I've sent an invite so you can make changes to that repo.

I like your changes, and I think it's the correct direction for this feature.

Just a quick thing if we're going to start doing caching - my searxng docker image is 263MB on disk, there should be some sort of warning if searxng is going to potentially x4 in disk size (max favicon storage is 1gb).

return42:
> sent an invite so you can make changes t

:+1: 

> Just a quick thing if we're going to start doing caching 

Both consume resources; if the icons come from the cache, the cache needs disk space, if the cache is turned off, the additional requests from the client and the additional requests from the resolver use significantly more bandwidth.

Here are a few thoughts from me:

- The activation of the favicons must therefore be a decision of the maintainer of the instance (Currently the client decides).

- The maintainer must be able to configure which resolvers are offered in the client (additional resolvers require additional resources)

- The maintainer must be able to determine the location of the CACHE / this is not currently possible, but could be done with a small change.


----

>  I'm already thinking about writing my own small self-hostable favicon cache/proxy that could be used with these changes here to not rely on any of the big providers (or include one that's opt-in in SearXNG, although I think a separated project would be better for this).

@Bnyro I had this in mind when revising the favicons; the current revision is modularized to the extent that it can easily be moved to an external project. However, in order to work in a resource-efficient manner, it will always be necessary for SearXNG to have access to the cache.

vojkovic:
@return42 I'm getting a 500 Internal Server Error when trying to search.

`
AttributeError: 'SQLiteProperties' object has no attribute '_DB'
` 

This is running on a fresh Github codespace doing `make run` as usual. Is this something on my end?

return42:
> I'm getting a 500 Internal Server Error when trying to search.

Sorry did a mistake when I implemented some compatibility stuff .. should now fixed / please try again.

searx/favicons/resolvers.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def google(domain, req_args):
+    """Favicon Resolver from google.com"""
+    data, mime = (None, None)
+
+    # URL https://www.google.com/s2/favicons?sz=32&domain={domain}" will be
+    # redirected (HTTP 301 Moved Permanently) to t1.gstatic.com/faviconV2:
+    response = network.get(
+        f"https://t1.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&url=https://{domain}&size=32",
```
vojkovic: Looks like this only works with some websites.

Doesn't work:
https://t1.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&url=https://reddit.com&size=32

For it to load needs the `fallback_opts` param:
https://t0.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=http://reddit.com&size=32


But some (actually most) sites work fine as:
https://t1.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&url=https://priv.au&size=32

and from my brief testing, they also do with the `fallback_opts` param, so maybe this should be included:

https://t1.gstatic.com/faviconV2?client=SOCIAL&type=FAVICON&fallback_opts=TYPE,SIZE,URL&url=https://priv.au&size=32

return42: > For it to load needs the `fallback_opts` param:

Yeah this is also what the redirect https://www.google.com/s2/favicons?sz=32&domain=reddit.com does .. thought I can shorten the URL a little bit .. but I can't :-)

Fixed and squashed into the WIP commit on top of the branch [diff](https://github.com/searxng/searxng/compare/e63e83b1d1688201f338b01a0187c5a3c162ef7f..f3acc085b709a9aba45ddb801a85466d56d1fbe4).

vojkovic:
So the favicon proxy has been in production on my instance for a few days now, and it's been working flawlessly. There's a noticeable speedup in having the favicons cached on the server. Really great work on this @return42.

> The activation of the favicons must therefore be a decision of the maintainer of the instance (Currently the client decides).

I agree. No matter what we do, this feature will take up more system resources. I suggest having a `favicons_enabled:` option in settings.yml (default: off) which will turn off favicons completely. Then, instance maintainers can make their own decision.

It also might be worth considering more maintainer changeable options notably for:
- Hold time of a cached favicon. 
- Limit of the favicon cached size.
- Location of faviconcache.db (touched more on later)

> The maintainer must be able to configure which resolvers are offered in the client (additional resolvers require additional resources)

I don't think this is a feature that's needed assuming the maintainer can choose the maximum amount of space that the database will take up. I don't see many people changing an option like this.

> * The maintainer must be able to determine the location of the CACHE / this is not currently possible, but could be done with a small change.

Yes, I agree. I'm not really sure about keeping the database in /tmp, I feel like this should be more persistent.

Also, a lot of instances are running inside docker containers with a tool like [watchtower](https://github.com/containrrr/watchtower), that automatically updates to the latest image. Every time the container restarts, the entire database gets wiped.

I suggest we recommend users to create a docker volume for the faviconcache.db so the data isn't purged. I can send in a PR to https://github.com/searxng/searxng-docker to make the changes to the default compose config once this gets merged.

> I had this in mind when revising the favicons; the current revision is modularized to the extent that it can easily be moved to an external project. However, in order to work in a resource-efficient manner, it will always be necessary for SearXNG to have access to the cache.

My 2c is to keep it as one project - it's easily maintainable as is.

searx/static/themes/simple/src/less/definitions.less
```diff
@@ -227,6 +230,9 @@
   --color-toolkit-loader-borderleft: rgba(0, 0, 0, 0);
   --color-doc-code: #ddd;
   --color-doc-code-background: #4d5a6f;
+  // Favicons Colors
+  --color-favicon-background-color: #bbb;
+  --color-favicon-border-color: #333;
```
vojkovic: Is there a reason as to why we're changing the background colours in light and dark mode?

realies:
I was wondering what I'm missing in SearXNG, and it was favicons - thanks for working on making this happen!

I'm unsure what the overhead is in enabling this feature, but I think it would be great if it is turned on by default, using the domain.tld/favicon.ico type of pull. Also, integrating the proxy within the same codebase would make it easier to use.

vojkovic:
> I'm unsure what the overhead is in enabling this feature, but I think it would be great if it is turned on by default, using the domain.tld/favicon.ico type of pull. Also, integrating the proxy within the same codebase would make it easier to use.

Sending requests to each individual search result compromises the instance privacy, so we've chosen to use public favicon resolvers instead. Keep in mind that not all domains host their favicon at `/favicon.ico`; some specify it explicitly in the HTML `<head>`, while others include it in `manifest.json`.

My most active instance location has cached 209.8MB of favicons, exceeding the 184MB size of the Docker image of searxng. Since we can't always count on instance maintainers to monitor new updates closely, this feature should require the explicit opt-in from the maintainer to prevent any unexpected surprises.

realies:
Thanks for the answers! I hope I'm not bloating the PR with additional thoughts and comments:

> Sending requests to each individual search result compromises the instance privacy, so we've chosen to use public favicon resolvers instead.

If favicon caching is a thing, a single request per instance per a reasonable unit of time (until it expires/time for a re-check) may not be particularly compromising, but I agree that using a public service to provide those acts as a privacy layer between the site and the search engine. Is caching using a public resolver possible?

> My most active instance location has cached 209.8MB of favicons, exceeding the 184MB size of the Docker image of searxng. Since we can't always count on instance maintainers to monitor new updates closely, this feature should require the explicit opt-in from the maintainer to prevent any unexpected surprises.

I probably did not get what is meant with this - why does it matter how big the Docker image for SearXNG is relative to an active instance favicons data size? Downloading more files to a container started from a Docker image does not limit its filesystem size to the Docker image size. I'd expect the favicons containing folder to be a named volume and would personally use a mounted volume to store any cache outside of the SearXNG container to make updating the image more efficient.

vojkovic:
> Is caching using a public resolver possible?

Yes, that's what is happening currently. Caching on both the client and server side.

> Why does it matter how big the Docker image for SearXNG is relative to an active instance favicons data size?

Maybe I didn't quite get my point across, it's not a technical issue, but if you download a 200mb container, it would be surprising for it to double in size.

Keep in mind, there's not really any other SearXNG feature that caches like this.

> would personally use a mounted volume to store any cache outside of the SearXNG container to make updating the image more efficient.

Yes, that's exactly what I recommended here: https://github.com/searxng/searxng/pull/3727#issuecomment-2302223861

realies:
Hope I don't come across as too stubborn about this, but this feature makes the most frequent type of SearXNG use at least twice better for me due to the additional visual queue information, and it would probably work the same for most people. 

>  but if you download a 200mb container, it would be surprising for it to double in size

I was surprised I did not have to mount any folders when I first tried SearXNG...

What if favicons and caching are enabled by default, and the default max storage space for favicons is much smaller? That would minimise the disk use surprise factor (for people who have not read the docs 👀 ) and enhance the default UX.

searx/static/themes/simple/src/less/definitions.less
return42: I found the background of transparent favicons (and the border) a bit too bright in dark mode ... so I thought it would be good to include this background and the border into the `definitions.less`.  I then made the two colors less bright in the dark theme.  If you think the latter is not good we can use the same colors in light and dark theme .. but we should not remove theses both two colors from the `definitions.less`.

RMI78:
I was about to open an issue for that, but that is really what's missing me in SearXNG. Seems like @realies made a good point, it would just make sens to let the user set the cache size himself

Aetherinox:
I just tried these changes out on a built container, and favicons are really what this app was missing. It enhances the overall app x2 just on its own. This addition works nice.

The space consumption would be a bit of a concern though. Granted, I'm hosting on a 4TB machine, and I'd have to visit an exorbitant amount of sites to deal any damage, but as much as I use searching for various development reasons throughout the day; I could really rack it up.

---

Something I thought of, obviously if it's a lot of work, could be looked at later. But I just wanted to toss the idea out there.

Would there be any interest in a custom host option for favicons? In short, I've written my own favicon grabber, it takes a lot of different scenarios into account because there's so many different ways of grabbing a favicon, and I host that on a Cloudflare worker, so cloudflare does the hosting for it. I Just call the query via a domain URL, and it returns my favicon which I can then do whatever I want with.

Out of the current options the developer of this PR has provided, I haven't looked at what the query limits are. But if there are limits, having a backup option may be nice.

Only env var I see being needed is just a host URL to the favicon fetcher. My current one is queried using `fav.domain.com/?domain=google.com`

And it just returns the URL to the favicon for that domain (or overrides, if I input them into my worker).

If you're asking "Why a backup", take a look at these three website's favicons:

oihDvKwHSM

It appears I can't fetch that favicon for that domain, but if I use my custom favicon grabber, I can get it.

return42:
Dear followers of this favicon PR, I'm sorry for my long absent on this topic.  FYI: To continue I squashed the patches from the [WIP commit](https://github.com/searxng/searxng/commit/e63e83b1d1688201f338b01a0187c5a3c162ef7f) into their related commits.  Further I did a rebase on SearXNG's current master branch (to fix a conflict with the static files).

return42:
As @Aetherinox already stated out, there is a bug .. when the cache is empty and a favicon provider does not serve a icon for a domain, we get this:

image

what is related to the 500 response:

image


Not sure, but I think the issue is; when the client sends a icon request to the icon proxy and the icon resolver of the icon proxy is not able to find an icon for the domain, the default icon should be in the response (and not a HTTP 500) .. at the moment (in the current implementation), the HTTP 500 is passed from the icon resolver directly to the client .. and the client will show the image alternative text (`foofighters.com I will study the case again in more detail and develop a solution.

return42:
> Not sure, but I think the issue is; when the client sends a icon request to the icon proxy and the icon resolver of the icon proxy is not able to find an icon for the domain, the default icon should be in the response (and not a HTTP 500)

My assumption was wrong, the default icon is already returned.  The real error was that the transaction was not completed when the tables were created, which resulted in the 500 error on the proxy.  The transactions of a connection must be explicitly closed (commit) ... the `with` context apparently does not do `commit` / see [diff](https://github.com/searxng/searxng/compare/307c07c4d8c33a9fd64cfc4a090d53d53c72793b..1bf883bfcb367fd43ea6889e93d394957642c41f)

----

> If you're asking "Why a backup", take a look at these three website's favicons ...

@Aetherinox this issue should be fixed now / thanks for pointing out :+1:

return42:
> > The maintainer must be able to configure which resolvers are offered in the client (additional resolvers require additional resources)
> 
> I don't think this is a feature that's needed assuming the maintainer can choose the maximum amount of space that the database will take up. I don't see many people changing an option like this.

As long as the cache size is a critical factor, it should not be left to the users to decide which favicon resolver they want to use ... it may be enough when just one user choose a resolver other than the default resolver .. and the cache is permanently at or over the limit.

Furthermore, the normal user has no criteria according to which he makes his selection.

I would therefore recommend defining a sensible default with the introduction of the favicons and this may only be changed by the instance operator for the time being ... when we have gained experience after the introduction of the favicons, then we can still consider whether and how we allow the user to make a selection.

return42:
FYI I pushed commit [33afea6](https://github.com/searxng/searxng/pull/3727/commits/33afea6f28baf2b72e691e551dc42ced9f6dd238) on top of this branch / read commit message to get the intention of this commit.

The commit is still WIP and suffers some more tests .. and detailed documentation about "How to setup a favicon service" is also missing .. I will continue my work the next days ..

------

> Just a quick thing if we're going to start doing caching - my searxng docker image is 263MB on disk, there should be some sort of warning if searxng is going to potentially x4 in disk size (max favicon storage is 1gb).

max favicon storage by default is now reduced to 50MB

> Would there be any interest in a custom host option for favicons? 

Custom favicon resolvers can now be integrated by a configuration.

--> things become more clear when I have written down a "How to" ..

Aetherinox:
> Custom favicon resolvers can now be integrated by a configuration.

Dang, you got that implemented already? Friggen nice. I'll have to test it out later.

Like with my Cloudflare worker, all the worker does is return the URL to the favicon, and then I use that image in whatever code to actually display the favicon.

If it works out well. I can also offer up a working example script for this repo, if some users want to run their own self-hosted favicon grabber through Cloudflare. They'll just have to set up the worker.

return42:
> The transactions of a connection must be explicitly closed (commit) ... the `with` context apparently does not do `commit` / see [diff](https://github.com/searxng/searxng/compare/307c07c4d8c33a9fd64cfc4a090d53d53c72793b..1bf883bfcb367fd43ea6889e93d394957642c41f)

Now I found out that the `with` context does a `commit` .. The actual error lies in the Cpython implementation as of 3.12+ 

- https://github.com/python/cpython/issues/118172

The workaround is to set caching in the DB connection to zero .. I will apply the workaround in my next push .. coming soon ..

return42:
FYI: We cannot rely on the resolver to return a valid favicon! ... here is one example wich returns a defective PNG: https://f1.allesedv.com/32/test-english.com

Unfortunately, we also have no way of checking whether a favicon is defective.  The browser only notices that this icon is defective when it is displayed in the client :-(

vojkovic:
> FYI: We cannot rely on the resolver to return a valid favicon! ... here is one example wich returns a defective PNG: https://f1.allesedv.com/32/test-english.com

Yeah I included allesedv as it wasn't hosted by a mainstream search engine - but if it's causing more trouble than it's worth, removing it might be the best option.

searx/favicons/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # pylint: disable=import-outside-toplevel
+
+    from . import config, cache, proxy
+
+    cfg_file = pathlib.Path("/etc/searxng/favicons.toml")
+    if not cfg_file.exists():
+        if is_active():
+            logger.error(f"missing favicon config: {cfg_file}")
+        cfg_file = config.DEFAULT_CFG_TOML
```
return42: Config file is now read from `/etc/searxng/favicons.toml`

searx/templates/simple/macros.html
```diff
@@ -21,9 +21,19 @@
...
+    {%- if favicon_resolver != "" %}
+    <div class="favicon"><img loading="lazy" src="{{ favicon_url(result.parsed_url.netloc) }}"></div>
```
return42: favicons are now loaded lazy (will save some bandwidth)

searx/sqlitedb.py
```diff
@@ -0,0 +1,322 @@
...
+        # with no restriction.
+        #
+        # if self.SQLITE_THREADING_MODE != "serialized":
+        #     if getattr(self.thread_local, 'DB', None) is None:
+        #         self.thread_local.DB = self.connect()
+        #     return self.thread_local.DB
+        #
+        # if self._DB is None:
+        #     self._DB = self.connect()  # pylint: disable=attribute-defined-outside-init
+        # return self._DB
```
return42: I had a lot of trouble with the optimization of DB connections in Python 3.12 ..  https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety says

> 3. Serialized: In serialized mode, SQLite can be safely used by multiple threads with no restriction.
> .. Threads may share the module, connections and cursors

But I have seen lots of issues when sharing the DB cursor over multiple threads.

to avoid this problem, one DB connection per thread is now set up for “read-only accesses” / a new connection had to be created for SQL updates, as the SQL updates in the different threads would otherwise block each other (as long as a transaction is running in one thread, the other thread may not make any updates, these competing updates could only be synchronized with separate DB connections)

return42 (commented):
I did some more tests today and fixed some bugs and made some minor improvements ... I will create the documentation in the next few days.

Here are a few notes on my work from today ...

searx/templates/simple/macros.html
vojkovic: Looks good - I was never able to get it working, but do you think we can avoid the rtl formatting in the html?

searx/static/themes/simple/src/less/style-ltr.less
```diff
@@ -82,4 +82,8 @@
   transform: scale(1, 1);
 }
 
+.favicon {
+  margin: 0 8px 0 0;
+}
+
```
vojkovic: When I wrote this, I was sure there must be a better way - but nothing I did worked, so I just kept it to discuss later. Can we use flex to maybe fix this?

searx/static/themes/simple/src/less/definitions.less
vojkovic: I'll leave this at your discretion - but I think the same colors should be used on both themes. There's not enough contrast between the favicon and background on the dark theme in my opinion.

searx/templates/simple/macros.html
return42: I have not yet found a solution (TBH I have not yet dealt it) .. @Bnyro suggests CSS in https://github.com/searxng/searxng/pull/3727#pullrequestreview-2240912152 .. I'm not sure how.

searx/static/themes/simple/src/less/definitions.less
return42: No problem .. now we have same colors in dark & light theme / [diff](https://github.com/searxng/searxng/compare/bbfd9249ef47e5afc7ebd7aca5d1300606959f69..388cdcad2c94ad0d05fb6757125ffc15d9c7272f)

return42:
FYI: last force push was a rebase on SearXNG's master branch to get https://github.com/searxng/searxng/pull/3836

return42:
FYI: today I added some documentation / to read, run ``make docs.live`` and visit http://0.0.0.0:8000/admin/searx.favicons.html

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,154 @@
...
+   # MAINTENANCE_PERIOD = 600                       # 10min / default: 1h
+
+:py:obj:`cfg_schema <.FaviconConfig.cfg_schema>`:
+  Is required to trigger any processes required for future upgrades / don't
+  change it.
+
+:py:obj:`cache.db_url <.FaviconCacheConfig.db_url>`:
+  The path to the (SQLite_) database file.  The default path is in the `/tmp`_
+  folder, which is deleted on every reboot and is therefore unsuitable for a
+  productive environment.  The FHS_ provides the folder  for the
```
vojkovic: 'productive environment' -> 'production environment'

return42:
Today I finished the docs / before review or test this PR I recommend to read the docs first / run `make docs.live` and visit http://0.0.0.0:8000/admin/searx.favicons.html

With the last force push I have also finished my local tests ... next I will patch this PR to my SearXNG instance to test a favicon setup in production extensively.

-----

If I have not forgotten anything from the above comments, only these points should be open, for which I do not know a better solution (maybe someone of you has suggestions):


- [ ] https://github.com/searxng/searxng/pull/3727#discussion_r1718732672
- [ ] https://github.com/searxng/searxng/pull/3727#discussion_r1765372691
- [ ] https://github.com/searxng/searxng/pull/3727#pullrequestreview-2313194469

Aetherinox:
> Thanks for this PR @vojkovic, this looks awesome!
> 
> PS: I'm already thinking about writing my own small self-hostable favicon cache/proxy that could be used with these changes here to not rely on any of the big providers (or include one that's opt-in in SearXNG, although I think a separated project would be better for this).

I highly recommend. I wrote my own, but I'm using a Cloudflare worker to run it.

It uses numerous methods to fetch a favicon from a site, and if for some unknown reason, a favicon is not found, I set a default one. Out of the thousands of URLs I've fetched, I've so far not found a single domain it can't pull.

I also hooked it up to a Github repository repo I run, and I can upload custom favicons to that repo if I want to override the favicon a site hosts. (Because some favicons suck, bad coloring, poor quality, etc). So I can push my own to my github repo, and it prioritizes mine first.

mrpaulblack:
Hi,
As requested by Markus in Matrix I deployed this PR to my prod instance for integration testing. I will keep you posted in the next days how resource usage on my VPS changes :+1:  (See -> https://paulgo.dev/infra/paulgoio/searxng/-/commit/0fbd92308388b84df8246ffe9f7caed28872a886)

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
+.. _favicons:
+
+========
+Favicons
+========
+
+.. sidebar:: warning
+
+   Don't activate the favacions before reading the documentation.
```
vojkovic: favacions -> favicons

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+be discussed later, first we have to setup the favicons cache*.
+
+Infrastructure
+==============
+
+The infrastructure for providing the favicons essentially consists of three
+parts:
+
+- :py:obj:`Favicons-Proxy <.favicons.proxy>` (aka *proxy*)
+- :py:obj:`Favicons-Resolvers <.favicons.resolvers>` (aka *reolver*)
```
vojkovic: reolver -> resolver

docs/admin/settings/settings_search.rst
```diff
@@ -41,6 +42,11 @@
   - ``qwant``
   - ``wikipedia``
 
+``favicon_resolver``:
+  To activate favicons in SearXNG's result list select a default
+  favicon-resolver, leave blank to turn off the feature.  Don't activate the
+  favacions before reading the :ref:`Favicons documentation <favicons>`.
```
vojkovic: favacions -> favicons

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+  write permission to this folder.
+
+  .. code:: bash
+
+     $ sudo mkdir /var/cache/searxng
+     $ sudo chown root:searxng /var/cache/searxng/
+     $ sudo chmod g+w /var/cache/searxng/
+
+:py:obj:`cache.LIMIT_TOTAL_BYTES <.FaviconCacheConfig.LIMIT_TOTAL_BYTES>`:
+  Maximum of bytes (default) stored in the cache of all blobs. The limit is only
```
vojkovic: Maximum of bytes (default)?

Did you forget to include the figure?

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+
+Activating the favicons in SearXNG is very easy, but this **generates a
+significantly higher load** in the communication between client and server and
+additional resources are used on the server.
+
+To mitigate these disadvantages, various methods have been implemented,
+including a *cache*.  The cache must be parameterized according to your own
+requirements and maintained regularly.
+
+To activate favicons in SearXNG's result list set a default favicons-resolver in
```
vojkovic: Add comma:

To activate favicons in SearXNG's result list set a default favicons-resolver in

To activate favicons in SearXNG's result list, set a default favicons-resolver in

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+the :ref:`search <settings search>` settings:
+
+.. code:: yaml
+
+   search:
+     favicon_resolver: "duckduckgo"
+
+By default and without any extensions SearXNG serves these resolvers:
```
vojkovic: Add comma:

By default and without any extensions SearXNG serves these resolvers:

By default and without any extensions, SearXNG serves these resolvers:

return42:
@vojkovic thanks a lot for your review :+1: .. I fixed the issues [diff](https://github.com/searxng/searxng/compare/90e3a8c760ecc7ad9d129348bb782fd2f2522d95..932650d1ed9ea7f7b0fe0d1636ca8a37e0664890) .. and rebased on master to solve a conflict (with master).

vojkovic:
@return42 Cheers for that. That'll conclude my review - fantastic work on your implementation, I'm very happy with the direction you took this feature.

LGTM after Bnyro's review, potentially he may come up with a solution to the RTL formatting, but I have nothing further to add. 👍

return42:
> @return42 Cheers for that. That'll conclude my review - fantastic work on your implementation, I'm very happy with the direction you took this feature.

Thank you very much, the pleasure was on my side. You have provided an [excellent submission with this PR](https://github.com/searxng/searxng/pull/3727#issuecomment-2284469305) .. I just had to optimize the number of requests :smile: 

Today (24st of September) I patched this PR on my instance ..

```
$ date
Thu Oct  3 02:58:28 PM CEST 2024

$ sudo -H -u searxng /usr/local/searxng/searx-pyenv/bin/python -m searx.favicons cache state
...
number of favicons in cache: 42K
total size (approx. bytes) of cache: 228.66 MB
total number of domains in cache: 60K
number of resolvers: 2

ls -lh /var/cache/searxng/
total 1.5G                                                             <--- !!
-rw-r--r-- 1 searxng searxng 266M Oct  3 14:54 faviconcache.db
-rw-r--r-- 1 searxng searxng 2.4M Oct  3 14:57 faviconcache.db-shm
-rw-r--r-- 1 searxng searxng 1.2G Oct  3 14:56 faviconcache.db-wal     <--- !!
```

Seems SQLite's WAL takes much more space compared the counted size.

----

```toml
[favicons]

cfg_schema = 1   # config's schema version no.

[favicons.proxy]

max_age = 5184000             # 60 days / default: 7 days (604800 sec)

[favicons.proxy.resolver_map]

"duckduckgo" = "searx.favicons.resolvers.duckduckgo"
"allesedv" = "searx.favicons.resolvers.allesedv"
"google" = "searx.favicons.resolvers.google"
"yandex" = "searx.favicons.resolvers.yandex"

[favicons.cache]

db_url = "/var/cache/searxng/faviconcache.db"  # default: "/tmp/faviconcache.db"
HOLD_TIME = 5184000                            # 60 days / default: 30 days
LIMIT_TOTAL_BYTES = 2147483648                 # 2 GB / default: 50 MB
BLOB_MAX_BYTES = 81920                         # 80 KB / default 20 KB
```

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+   Don't activate the favicons before reading the documentation.
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+Activating the favicons in SearXNG is very easy, but this **generates a
+significantly higher load** in the communication between client and server and
+additional resources are used on the server.
```
Bnyro:
```suggestion
significantly higher load** in the communication between client and server and
increases resource usage on the server.
```
nit: two `and`s, grammar

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+
+To mitigate these disadvantages, various methods have been implemented,
+including a *cache*.  The cache must be parameterized according to your own
+requirements and maintained regularly.
+
+To activate favicons in SearXNG's result list, set a default favicons-resolver
```
Bnyro:
```suggestion
To activate favicons in SearXNG's result list, set a default ``favicon_resolver``
```
named favicon_resolver without s below

docs/admin/searx.favicons.rst
```diff
@@ -0,0 +1,249 @@
...
+
+- ``duckduckgo``
+- ``allesedv``
+- ``google``
+- ``yandex``
+
+With the above setting favicons are displayed, the user has the option to
+deactivate this feature in his settings.  If the user is to have the option of
+selecting from several *resolvers*, a further setting is required / *but this will
+be discussed later, first we have to setup the favicons cache*.
```
Bnyro:
```suggestion
With the above setting, favicons are displayed by default. Users have the option to
deactivate this feature in their settings..
```
Forgot to remove? / TODO?

Bnyro:
The patch for the RTL thing with the favicons:
```
diff --git a/searx/static/themes/simple/src/less/style-ltr.less b/searx/static/themes/simple/src/less/style-ltr.less
index 5d8c5dbe5..6f7218b02 100644
--- a/searx/static/themes/simple/src/less/style-ltr.less
+++ b/searx/static/themes/simple/src/less/style-ltr.less
@@ -82,8 +82,4 @@
   transform: scale(1, 1);
 }
 
-.favicon {
-  margin: 0 8px 0 0;
-}
-
 @import "style.less";
diff --git a/searx/static/themes/simple/src/less/style-rtl.less b/searx/static/themes/simple/src/less/style-rtl.less
index aa663436f..8e2634d34 100644
--- a/searx/static/themes/simple/src/less/style-rtl.less
+++ b/searx/static/themes/simple/src/less/style-rtl.less
@@ -94,12 +94,12 @@
   direction: initial;
   text-align: right;
 
+  .result .url_header {
+    direction: rtl;
+  }
+
   .result .url_wrapper {
     justify-content: end;
-
-    .favicon {
-      margin: 0 0 0 8px;
-    }
   }
 }
 
diff --git a/searx/static/themes/simple/src/less/style.less b/searx/static/themes/simple/src/less/style.less
index 5c9d7a07a..8df6980c6 100644
--- a/searx/static/themes/simple/src/less/style.less
+++ b/searx/static/themes/simple/src/less/style.less
@@ -232,6 +232,11 @@ article[data-vim-selected].category-social {
     }
   }
 
+  .url_header {
+    display: flex;
+    gap: 0.5rem;
+  }
+
   .url_wrapper {
     display: flex;
     align-items: center;
diff --git a/searx/templates/simple/macros.html b/searx/templates/simple/macros.html
index 858ee2d7e..a6c1f98ab 100644
--- a/searx/templates/simple/macros.html
+++ b/searx/templates/simple/macros.html
@@ -20,20 +20,15 @@
 <!-- Draw result header -->
 {% macro result_header(result, favicons, image_proxify) -%}
 <article class="result {% if result['template'] %}result-{{ result.template|replace('.html', '') }}{% else %}result-default{% endif %} {% if result['category'] %}category-{{ result['category'] }}{% endif %}{% for e in result.engines %} {{ e }}{% endfor %}">
-  {{- result_open_link(result.url, "url_wrapper") -}}
-  {% if not rtl %}
-    {%- if favicon_resolver != "" %}
-    <div class="favicon"><img loading="lazy" src="{{ favicon_url(result.parsed_url.netloc) }}"></div>
-    {%- endif -%}
-  {%- endif -%}
-  {%- for part in get_pretty_url(result.parsed_url) -%}
-  <span class="url_o{{loop.index}}"><span class="url_i{{loop.index}}">{{- part -}}</span></span>
-  {%- endfor %}
-  {% if rtl %}
-    {%- if favicon_resolver != "" %}
-    <div class="favicon"><img loading="lazy" src="{{ favicon_url(result.parsed_url.netloc) }}"></div>
-    {%- endif -%}
+  {{- result_open_link(result.url, "url_header") -}}
+  {%- if favicon_resolver != "" %}
+  <div class="favicon"><img loading="lazy" src="{{ favicon_url(result.parsed_url.netloc) }}"></div>
   {%- endif -%}
+  <div class="url_wrapper">
+    {%- for part in get_pretty_url(result.parsed_url) -%}
+    <span class="url_o{{loop.index}}"><span class="url_i{{loop.index}}">{{- part -}}</span></span>
+    {%- endfor %}
+  </div>
   {{- result_close_link() -}}
   {%- if result.thumbnail %}{{ result_open_link(result.url) }}<img class="thumbnail" src="{{ image_proxify(result.thumbnail) }}" title="{{ result.title|striptags }}" loading="lazy">{{ result_close_link() }}{% endif -%}
```

searx/favicons/config.py
```diff
@@ -0,0 +1,62 @@
...
+from pydantic import BaseModel
+
+from searx.compat import tomllib
+from .cache import FaviconCacheConfig
+from .proxy import FaviconProxyConfig
+
+CONFIG_SCHEMA: int = 1
+"""Version of the configuration schema."""
+
+TOML_CACHE: dict[str, "FaviconConfig"] = {}
```
Bnyro:
```suggestion
TOML_CACHE_CFG: dict[str, "FaviconConfig"] = {}
```

searx/favicons/config.py
```diff
@@ -0,0 +1,62 @@
...
+"""Cache config objects by TOML's filename."""
+
+DEFAULT_CFG_TOML = pathlib.Path(__file__).parent / "favicons.toml"
```
Bnyro:
```suggestion
DEFAULT_CFG_TOML_PATH = pathlib.Path(__file__).parent / "favicons.toml"
```

Bnyro (commented):
Everything looks fine to me generally, just left some small comments above.

I didn't deeply review the database part though as I'm not that familiar with SQL in Python.

docs/admin/searx.favicons.rst
return42: my intention was to say: *"higher load in the client/server communication and increases resource usage on the server."* --> [fixed](https://github.com/searxng/searxng/compare/a76b37fe75f50c8281fbec24499e74211ba2b67e..7c7482fc107484e883803df65176647d58aff989)

docs/admin/searx.favicons.rst
return42: Nope .. "but this will be discussed later" means later in this article .. but I [fixed](https://github.com/searxng/searxng/compare/a76b37fe75f50c8281fbec24499e74211ba2b67e..7c7482fc107484e883803df65176647d58aff989) this part to be more clear.

searx/favicons/config.py
return42: [fixed](https://github.com/searxng/searxng/compare/7c7482fc107484e883803df65176647d58aff989..2021ebc593f7fea2c2c893bb0361f8ff860d3f13)

return42:
> The patch for the RTL thing with the favicons:

Yeah, thanks a lot :+1:  .. I assembled [433e5f8](https://github.com/searxng/searxng/pull/3727/commits/433e5f8f35e8888928cd2a321f50b9eeee313e74) from your patch.

Bnyro (approved):

return42:
FYI: I had to rebase the branch to solve a conflict in ``definitions.less`` and in the `/static` files build by `make static.build.commit`.

return42:
What do you guys think, is this PR ready to release?

I test this PR for more than a week on my production instance ..

The favicons work perfectly, only the size of the WAL bothers me (see [ls -lh /var/cache/searxng/](https://github.com/searxng/searxng/pull/3727#issuecomment-2370980820)) ... I guess the WAL file grows because the server processes run endlessly and the DB-Connector is therefore never closed.

That reminds me that we have to continue with this PR ;-)

For those of you testing this PR in production .. what can you say about your WAL file?

mrpaulblack:
From my point of view I also tested this for over a week in prod without any problems. RAM and CPU usage seem to be about the same as before this PR :+1:


The database faviconcache.db is 150MB on my instance. Seems ok IMO. The `faviconcache.db-wal` file is 4.2MB on my instance.

vojkovic:
Here's the statistics from each of my nodes from the last two months of use:

Singapore
526.7M	/tmp/faviconcache.db
152.8M	/tmp/faviconcache.db-wal

Frankfurt
762.8M	/tmp/faviconcache.db
231.2M	/tmp/faviconcache.db-wal

London
425.8M	/tmp/faviconcache.db
119.5M	/tmp/faviconcache.db-wal

Chicago
670.5M	/tmp/faviconcache.db
208.3M	/tmp/faviconcache.db-wal

Sydney
172.9M	/tmp/faviconcache.db
50.8M	/tmp/faviconcache.db-wal

Bangalore
87.1M	/tmp/faviconcache.db
19.8M	/tmp/faviconcache.db-wal

San Francisco
255.1M	/tmp/faviconcache.db
81.3M	/tmp/faviconcache.db-wal

I'm pretty happy with how this looks, obviously the more that gets cached, the less favicons that will need to be requested in the future. LGTM 🚀

return42:
Many thanks for the cooperation on the favicons, thanks to @vojkovic SearXNG is once again a bit more up-to-date and user-friendly ... everything seems to be fine so far ... lets merge :rocket:

Nocifer:
A big, big THANK YOU to everyone involved in this PR!

Aetherinox:
Thanks everyone for working on this. Can't wait to try the latest out. But first I need to compare the CSS changes. I re-built the entire Searxng theme from the ground up, and I've stuck myself in a situation where I must manually do everything. Joy....

Aetherinox:
@return42 Just got to say.

There's a crap-ton of settings in there. 

> In the following we list the resolvers available in the core of SearXNG, but via the [FQN](https://en.wikipedia.org/wiki/Fully_qualified_name) it is also possible to implement your own resolvers and integrate them into the proxy:

You busted your tail on this favicon functionality, and it's about everything a person could dream of. The amount of detail in here is amazing.

Thanks again. I'm going to go read the docs and see how I can implement the custom favicon resolver. This'll be the first big thing I've done with the app.

omentic:
Hey, this is very cool work, thank you! Excited to pull it down and test it out :-)

return42:
Dear contributors to the favicon PR, can you please give 

- https://github.com/searxng/searxng/pull/3938

a try and give me your feedback? /thanks :+1:

PR #3201, opened 2024-02-10
draft: yandex engine (web, images, videos)

Austin-Olacsi:
## What does this PR do?

this draft shows my current progress on a yandex engine. it does not use xpath selectors yet and yes i am aware it needs some cleanup.

## Why is this change important?

once it is ready it will be a valuable addition

## How to test this PR locally?
make run

## Related issues

https://github.com/searxng/searxng/issues/961
https://github.com/searxng/searxng/issues/2181

web
images
videos

PR #3733, opened 2024-08-12
[feat] engine: implementation of yandex (web, images)

### LINKED ISSUE #961: Yandex engine request
**Working URL to the engine**

https://yandex.ru/
https://yandex.eu/

(and more country specific TLD)

Yandex is the only big company backed search engine that isn't owned by a western company. 

It's the only one that doesn't filter anything (yes, this includes things about the current war) - Edit 2024-06-04: at least it hasn't be proven to be filtering stuff.

It's also only the only engine that doesn't "biases" its results for the sake of being politically correct/inclusive. (Example, I heard about recently: [DDG](https://duckduckgo.com/?q=blonde+braids&iax=images&ia=images) / [Yandex](https://yandex.eu/images/search?text=blonde%20braids)) - Edit 2024-06-04: some testing was done, Yandex.eu is now biased (though a lot less than DDG and western engines), the "problem" isn't present on Yandex.ru ([see here](https://yandex.ru/images/search?text=blonde%20braids)) (so both will have to be added)

**Features of this engine**

See previous points.

General, Images,Maps, Videos

**Additional context**

Support for it was removed in #709 (but it basically was broken at this point)

Austin-Olacsi:
## What does this PR do?

adds yandex engine for web and image results.

it's about time we have yandex in searxng again.

it's set to inactive in settings.yml because of captcha. you need to remove that from the settings.yml.
then you can use: make run

## Author's checklist

my old draft: https://github.com/searxng/searxng/pull/3201
engine request: https://github.com/searxng/searxng/issues/961

## Related issues
Closes #961

### DIFF
--- a/searx/engines/yandex.py
+++ b/searx/engines/yandex.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Yandex (Web, images)"""
+
+from json import loads
+from urllib.parse import urlencode
+from html import unescape
+from lxml import html
+from searx.exceptions import SearxEngineCaptchaException
+from searx.utils import humanize_bytes, eval_xpath, eval_xpath_list, extract_text, extr
+
+
+# Engine metadata
+about = {
+    "website": 'https://yandex.com/',
+    "wikidata_id": 'Q5281',
+    "official_api_documentation": "?",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+# Engine configuration
+categories = []
+paging = True
+search_type = ""
+
+# Search URL
+base_url_web = 'https://yandex.com/search/site/'
+base_url_images = 'https://yandex.com/images/search'
+
+results_xpath = '//li[contains(@class, "serp-item")]'
+url_xpath = './/a[@class="b-serp-item__title-link"]/@href'
+title_xpath = './/h3[@class="b-serp-item__title"]/a[@class="b-serp-item__title-link"]/span'
+content_xpath = './/div[@class="b-serp-item__content"]//div[@class="b-serp-item__text"]'
+
+
+def catch_bad_response(resp):
+    if resp.url.path.startswith('/showcaptcha'):
+        raise SearxEngineCaptchaException()
+
+
+def request(query, params):
+    query_params_web = {
+        "tmpl_version": "releases",
+        "text": query,
+        "web": "1",
+        "frame": "1",
+        "searchid": "3131712",
+    }
+
+    query_params_images = {
+        "text": query,
+        "uinfo": "sw-1920-sh-1080-ww-1125-wh-999",
+    }
+
+    if params['pageno'] > 1:
+        query_params_web.update({"p": params["pageno"] - 1})
+        query_params_images.update({"p": params["pageno"] - 1})
+
+    params["cookies"] = {'cookie': "yp=1716337604.sp.family%3A0#1685406411.szm.1:1920x1080:1920x999"}
+
+    if search_type == 'web':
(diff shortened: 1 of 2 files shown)


searx/engines/yandex.py
```diff
@@ -0,0 +1,137 @@
...
+    if params['pageno'] > 1:
+        query_params_web.update({"p": params["pageno"] - 1})
+        query_params_images.update({"p": params["pageno"] - 1})
+
+    params["cookies"] = {'cookie': "yp=1716337604.sp.family%3A0#1685406411.szm.1:1920x1080:1920x999"}
+
+    if search_type == 'web':
+        params['url'] = f"{base_url_web}?{urlencode(query_params_web)}"
+    elif search_type == 'images':
+        params['url'] = f"{base_url_images}?tmpl_version=releases&{urlencode(query_params_images)}"
```
Bnyro: `tmpl_version` should likely be part of `query_params_images` as well to keep things cleaner

searx/engines/yandex.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return params
+
+
+def response(resp):  # pylint: disable=too-many-locals
+    if search_type == 'web':
+        if (resp.url).path.startswith('/showcaptcha'):
```
Bnyro: repeated in line 85, should be outside the if condition.

the brackets around `resp.url` are not needed either here

searx/engines/yandex.py
```diff
@@ -0,0 +1,137 @@
...
+            raise SearxEngineCaptchaException()
+
+        html_data = html.fromstring(resp.text)
+        html_sample = unescape(html.tostring(html_data, encoding='unicode'))
+
+        start_tag = '{"location":"/images/search/'
+        end_tag = 'advRsyaSearchColumn":null}}'
+
+        start_index = html_sample.find(start_tag)
+        start_index = start_index if start_index != -1 else -1
```
Bnyro: what it does is `start_index = start_index`, if that's the intention the line can be dropped

searx/engines/yandex.py
```diff
@@ -0,0 +1,137 @@
...
+        if resp.url.path.startswith('/showcaptcha'):
+            raise SearxEngineCaptchaException()
+
+        html_data = html.fromstring(resp.text)
+        html_sample = unescape(html.tostring(html_data, encoding='unicode'))
+
+        start_tag = '{"location":"/images/search/'
+        end_tag = 'advRsyaSearchColumn":null}}'
+
+        start_index = html_sample.find(start_tag)
```
Bnyro: This could be simplified using https://github.com/searxng/searxng/blob/master/searx/utils.py#L356

return42 (approved):
Thank you for devoting yourself to the topic again. Even if we can't prevent the CAPTCHA yet, I think it's OK if we add the engines as "inactive".  Maybe someone will see this code and find a solution to avoid the CAPTCHA.

I squashed your two commits into one and fixed a nit / [diff](
https://github.com/searxng/searxng/compare/6014340a1e5b4a7714ea27430f39f289b07ab0b4..e04ff2c8d0f3d8f2f2720e95c6d31e31956af91f)

ikkqqstudio:
Any updates in 2025? And Why is it taking so longer than 4 months to add Yandex into The open-source search engine?

return42:
> And Why is it taking so longer than 4 months to add Yandex into The open-source search engine?

It is in .. but

> it's set to inactive in settings.yml because of captcha.

Maintainers can activate the engine in their settings.

@ikkqqstudio Once you show us how to avoid the CAPTCHA, we can enable Yandex by default ... We are just waiting for your cooperation ;-)

ikkqqstudio:
There is a Yandex search API, but it requires Paid Cloud subscription

Issue #3722, opened 2024-08-09
hostname_replace has stopped working in latest 2024.8.8+dcf95644c


825i:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

**How did you install SearXNG?**
Through UNRAID
**What happened?**
I updated without problems but hostname_replace plugin isn't working in the latest version.

**How To Reproduce**
```
enabled_plugins:
#   # these plugins are enabled if nothing is configured ..
#   - 'Hash plugin'
#   - 'Self Information'
#   - 'Tracker URL remover'
#   - 'Ahmia blacklist'  # activation depends on outgoing.using_tor_proxy
#   # these plugins are disabled if nothing is configured ..
    - 'Hostname replace'  # see hostname_replace configuration below
#   - 'Open Access DOI rewrite'
#   - 'Tor check plugin'
#   # Read the docs before activate: auto-detection of the language could be
#   # detrimental to users expectations / users can activate the plugin in the
#   # preferences if they want.
#   - 'Autodetect search language'

# Configuration of the "Hostname replace" plugin:
#

hostname_replace:
#  '(.*\.)?facebook\.com$': false
#  '(.*\.)?instagram\.com$': false
#  '(.*\.)?quora\.com$': false
#  '(.*\.)?twitter\.com$': false
#  '(.*\.)?x\.com$': false
#   '(.*\.)?youtube\.com$': 'invidious.example.com'
#   '(.*\.)?youtu\.be$': 'invidious.example.com'
#   '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
#   '(.*\.)?reddit\.com$': 'teddit.example.com'
#   '(.*\.)?redd\.it$': 'teddit.example.com'
#   '(www\.)?twitter\.com$': 'nitter.example.com'
#   # to remove matching host names from result list, set value to false
  '(.*\.)?facebook\.com$': false
  '(.*\.)?instagram\.com$': false
  '(.*\.)?quora\.com$': false
  '(.*\.)?twitter\.com$': false
  '(.*\.)?x\.com$': false
  '(.*\.)?reddit\.com$': false
  '(.*\.)?youtube\.com$': false
  '(.*\.)?google\.com$': false
  '(.*\.)?apple\.com$': false
  '(.*\.)?tripadvisor\.com$': false
  '(.*\.)?tripadvisor\.com.au$': false
  '(.*\.)?tripadvisor\.in$': false
[...]

PR #3739, opened 2024-08-16
[docs] improve Hostname plugin documentation

dennis1248:
## What does this PR do?

It clearly defines in the docs that `Hostname replace` has been deprecated and replaced by `Hostnames plugin`. It also provides some additional references material to make it easier on admins to port to the new plugin. And on top of all that as an exiting bonus it also fixes a few typos.

It makes it easier for people who don't actively follow Searx development channels to port after this breaking change on their own. This should reduce questions in the eg. Matrix and issue reports on Git. 

I build the docs and validated it all works and looks properly by running it in a local Podman Apache webserver container.

## Related issues

Lack of proper docs about this breaking change was discussed here; https://github.com/searxng/searxng/issues/3722

----

update: see new docs: https://docs.searxng.org/dev/plugins/hostnames.html

### DIFF
--- a/searx/plugins/hostnames.py
+++ b/searx/plugins/hostnames.py
@@ -1,7 +1,19 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=too-many-branches
-"""In addition to rewriting/replace reslut URLs, the *hoostnames* plugin offers
-other features.
+"""
+.. attention::
+
+    The **"Hostname replace"** plugin has been replace by **"Hostnames
+    plugin"**, see :pull:`3463` & :pull:`3552`.
+
+The **Hostnames plugin** can be enabled by adding it to the
+``enabled_plugins`` **list** in the ``setting.yml`` like so.
+
+  .. code:: yaml
+
+     enabled_plugins:
+       - 'Hostnames plugin'
+       ...
 
 - ``hostnames.replace``: A **mapping** of regular expressions to hostnames to be
   replaced by other hostnames.


Bnyro (approved):

searx/plugins/hostnames.py
```diff
@@ -1,7 +1,21 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=too-many-branches
-"""In addition to rewriting/replace reslut URLs, the *hoostnames* plugin offers
-other features.
+""" .. note::
+
+    The **Hostname replace** plugin has been deprecated and replace by
+    **Hostnames plugin**.
```
return42: "deprecated" is the wrong word here:

    The **Hostname replace** plugin has been replace by the **Hostnames plugin**.

searx/plugins/hostnames.py
```diff
@@ -1,7 +1,21 @@
...
+``enabled_plugins`` **list** in the ``setting.yml`` like so.
+
+  .. code:: yaml
+
+     enabled_plugins:
+       - 'Hostnames plugin'
+       ...
+
+In addition to rewriting/replacing result URLs, the **Hostnames plugin** offers
+additional features.
```
return42: "In addition" .. to what? .. implies a reference to the old plugin... but in my opinion we don't need that / the sentence can be deleted.

return42 (changes requested):

searx/plugins/hostnames.py
dennis1248: That is what the original text said. It is also unclear to me what is intended here. I will drop it.

searx/plugins/hostnames.py
dennis1248: Changed it.

Issue #3728, opened 2024-08-11
Bug: tineye engine


return42:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2024.8.9

**How did you install SearXNG?**

`make run`

**How To Reproduce**

reverse search for an image `!tin https://hekint.org/wp-content/uploads/2019/11/1989.1_Fountain-of-Milk_web-768x667.jpg`

**What happened?**

Seems the response from tineye has changed in the past and we need to adapt the new response:

```
Traceback (most recent call last):
  File "searx/search/processors/online.py", line 163, in search
    search_results = self._search_basic(query, params)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "searx/search/processors/online.py", line 151, in _search_basic
    return self.engine.response(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "searx/engines/tineye.py", line 196, in response
    for match_json in json_data['matches']:
                      ~~~~~~~~~^^^^^^^^^^^
KeyError: 'matches'
```
**Technical report**

Error
  * Error: KeyError
  * Percentage: 100
  * Parameters: `()`
  * File name: `searx/engines/tineye.py:196`
  * Function: `response`
  * Code: `for match_json in json_data['matches']:`

glanham-jr:
Going to dump my investigation results below:

Post URL: https://tineye.com/api/v1/result_json/?sort=score&order=desc

I just tried to use TinEye directly for that image, and received an error.

![image](https://github.com/user-attachments/assets/a6640607-7e95-47c6-903a-4c14ae98dd61)

What is strange that after a few tries, it seems to be working now and I cannot reproduce the error. Regardless, to replicate an error, I just entered an invalid URL (I just chopped off the extension in the image you provided).

The error JSON doesn't have a `matches` result, but is formatted as follows:

```json
{
  "page": 1,
  "sort_selector": null,
  "limit": 10,
  "domain_name": "",
  "no_cache": false,
  "image_server": "https://img.tineye.com/",
  "load_query_summary": false,
  "show_unavailable_domains": false,
  "sort": "score",
  "order": "desc",
  "domain": "",
  "tags": "",
  "offset": 0,
  "suggestions": {
    "key": "Download Error",
    "title": "Oops, something didn't work!",
    "suggestions": [],
    "description": [
[...]

PR #3654, opened 2024-07-13
[fix] tear down TEST_ENGINES after TestBang is proceeded

return42:
Engines are loaded into global name `searx.engines.engines` other applications such as statistics or the histogram use this global variable to search for values in their own memories, which can lead to key errors as described in

- https://github.com/searxng/searxng/issues/2988

## Related issues

Closes: https://github.com/searxng/searxng/issues/2988

Issue #2988, opened 2023-11-04
`make test.unit` fails with `KeyError: ('engine', 'dummy engine', 'search', 'count', 'sent')`


dalf:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

Last commit of the master branch 96ccec1aa6507ef4d3dc05a08e01187f80080972

**How did you install SearXNG?**
N/A

**How To Reproduce**
`make test.unit`

Output:
```
  File "/home/alexandre/code/searxng/searx/metrics/models.py", line 146, in get
    return self.counters[args]
           ~~~~~~~~~~~~~^^^^^^
KeyError: ('engine', 'dummy engine', 'search', 'count', 'sent')
```

<details>

<summary>full log</summary>

```
BUILDENV  INFO    searx                         : load the default settings from /home/alexandre/code/searxng/searx/settings.yml
BUILDENV  INFO    searx                         : max_request_timeout=None
BUILDENV  INFO    searx                         : version: 2023.5.10+fe0cce2ad+dirty
BUILDENV  build utils/brand.env (settings from: /home/alexandre/code/searxng/etc/settings.yml)
....WARNING searx.botdetection.config     : missing config file: /etc/searxng/limiter.toml
.........................................................................................EE..........E.E..........E..............................................
======================================================================
ERROR: test_browser_empty_locale (tests.unit.test_webapp.ViewsTestCase.test_browser_empty_locale)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/alexandre/code/searxng/tests/unit/test_webapp.py", line 219, in test_browser_empty_locale
    result = self.app.get('/preferences', headers={'Accept-Language': ''})
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alexandre/code/searxng/local/py3/lib/python3.11/site-packages/werkzeug/test.py", line 1160, in get
    return self.open(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alexandre/code/searxng/local/py3/lib/python3.11/site-packages/flask/testing.py", line 232, in open
    response = super().open(
               ^^^^^^^^^^^^^
  File "/home/alexandre/code/searxng/local/py3/lib/python3.11/site-packages/werkzeug/test.py", line 1114, in open
    response = self.run_wsgi_app(request.environ, buffered=buffered)
[...]

PR #3746, opened 2024-08-20
Fix tineye engine url, datetime parsing, and minor refactor

glanham-jr:
## What does this PR do?

Fixes the tineye engine to point to the new API url.

Changes made to tineye engine:
1. Importing logging if TYPE_CHECKING is enabled
2. Remove unnecessary try-catch around json parsing the response, as this masked the original error and had no immediate benefit
3. Improve error handling explicitly for status code 422 and 400 upfront, deferring json_parsing only for these status codes and successful status codes
4. Unit test all new applicable changes to ensure compatibility

I was able to force a `400` response by editing and sending a malformed POST request. With this, I was able to get a specific JSON result back (see unit test on status code 400) which I can parse the description of the error and log it. However, I imagine this is most likely a genuine error, but for now I am opting for backwards compatibility in handling the 400 status code.

Fixes the tineye engine that is currently broken.

1. search a valid request: `!tin https://hekint.org/wp-content/uploads/2019/11/1989.1_Fountain-of-Milk_web-768x667.jpg`
2. search a bad URL request: `!tin https://hekint.org/wp-content/foobaretc`

## Author's checklist

 - n/a

## Related issues

 - #3728

## Screenshots

image

image

### DIFF
--- a/searx/engines/tineye.py
+++ b/searx/engines/tineye.py
@@ -14,9 +14,15 @@ billion images `[tineye.com] <https://tineye.com/how>`_.
 
 """
 
+from typing import TYPE_CHECKING
 from urllib.parse import urlencode
 from datetime import datetime
 from flask_babel import gettext
+
+if TYPE_CHECKING:
+    import logging
+
+    logger = logging.getLogger()
 
 about = {
     "website": 'https://tineye.com',
@@ -34,7 +40,7 @@ categories = ['general']
 paging = True
 safesearch = False
 base_url = 'https://tineye.com'
-search_string = '/result_json/?page={page}&{query}'
+search_string = '/api/v1/result_json/?page={page}&{query}'
 
 FORMAT_NOT_SUPPORTED = gettext(
     "Could not read that image url. This may be due to an unsupported file"
@@ -120,7 +126,7 @@ def parse_tineye_match(match_json):
 
             crawl_date = backlink_json.get("crawl_date")
             if crawl_date:
-                crawl_date = datetime.fromisoformat(crawl_date[:-3])
+                crawl_date = datetime.strptime(crawl_date, '%Y-%m-%d')
             else:
                 crawl_date = datetime.min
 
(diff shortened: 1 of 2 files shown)


glanham-jr:
Will format later - bulk of work is still relevant.

glanham-jr:
My tests regarding the logging work locally fine, but are breaking in CI. Seems like `TYPE_CHECKING` is returning False in CI. Might need some pointers on addressing this.

return42:
> My tests regarding the logging work locally fine, but are breaking in CI. Seems like `TYPE_CHECKING` is returning False in CI. Might need some pointers on addressing this.

Don't worry, we've had a similar problem before:

- https://github.com/searxng/searxng/issues/2988#issuecomment-2226929084) fixed in
- https://github.com/searxng/searxng/pull/3654

Our tests are poorly programmed and there are many global variables & monkey patching in SearXNG. Depending on the order in which the tests are called, parts of SearXNG may or may not be initialized, and global variables may or may not contain values ​​left over from previous tests.

Below you can see the error; there is no logger in tineye.py. The logger comes in via monkey patching when the engines are initialized. If you run the tests locally, then a test that initializes the engines seems to run before this test. If the CI runs the test, the order of the tests may be different and no initialization has taken place yet.

Since the order of the unit tests is not deterministic, the error may occur sometimes and not sometimes.

The solution is that we need to set up our tests more cleanly. I'll have to look into it and get back to you as soon as I have a solution.


-----
```
======================================================================
ERROR: test_empty_list_for_400 (tests.unit.test_tineye.TinEyeTests.test_empty_list_for_400)
  File "/home/runner/work/searxng/searxng/tests/unit/test_tineye.py", line 69, in test_empty_list_for_400
    results = tineye.response(response)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/searxng/searxng/searx/engines/tineye.py", line 186, in response
    logger.error(message)
    ^^^^^^
NameError: name 'logger' is not defined
```

return42:
> The solution is that we need to set up our tests more cleanly. I'll have to look into it and get back to you as soon as I have a solution.

FYI: I merged https://github.com/searxng/searxng/pull/3750 and rebased your branch .. then I fixed the unit test of your tineye --> [diff](https://github.com/searxng/searxng/compare/47aff473ba1ef833b58173aecb81f683cd32dc1e..57d970f7711f7f0c34728f44902f4d23a9b249ef)

return42 (approved):
Works like a charm / thanks a lot for your investigation :+1:

PR #3751, opened 2024-08-21
[mod] hardening "calculator plugin" / limit execution time to 50 ms 

return42:
### [mod] hardening "calculator plugin" / limit execution time to 50 ms

The execution of the function for the calculation is outsourced to a process whose runtime is limited to 50 milliseconds.

### [mod] enable calculator and allow plugin on public instances

Remove quirks that prevented the Calculator from being used on public instances.

### Related:

- https://github.com/searxng/searxng/pull/3377#issuecomment-2067977375

### DIFF
--- a/searx/plugins/calculator.py
+++ b/searx/plugins/calculator.py
@@ -4,9 +4,12 @@
 
 import ast
 import operator
+from multiprocessing import Process, Queue
 
 from flask_babel import gettext
 from searx import settings
+
+from searx.plugins import logger
 
 name = "Basic Calculator"
 description = gettext("Calculate mathematical expressions via the search bar")
@@ -14,6 +17,8 @@ default_on = False
 
 preference_section = 'general'
 plugin_id = 'calculator'
+
+logger = logger.getChild(plugin_id)
 
 operators = {
     ast.Add: operator.add,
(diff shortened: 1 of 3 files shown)


searx/plugins/calculator.py
```diff
@@ -51,10 +55,31 @@ def _eval(node):
     raise TypeError(node)
 
 
+def timeout_func(timeout, func, *args, **kwargs):
```
glanham-jr, line 58: This seems like a useful utility method that could be used for any general func. Do you think this has the possibility to be used in other places in the project, and if so for it to belong in a separate shared file?

return42: I had already thought about it, but then decided against it: Forking a new process is a resource "expensive" operation and should remain the exception. Should we one day have the need elsewhere, we could then consider moving this function to another module such as `searx.utils`.

searx/plugins/calculator.py
```diff
@@ -51,10 +55,31 @@ def _eval(node):
...
+
+    def handler(q: Queue, func, args, **kwargs):  # pylint:disable=invalid-name)
```
Bnyro:
```suggestion
    def handler(q: Queue, func, args, **kwargs):  # pylint:disable=invalid-name
```
small typo

Bnyro (approved):
This works as expected and feels like a much better solution than limiting it to public instances only.

Great job!

searx/plugins/calculator.py
return42: Thanks! .. fixed / [diff](https://github.com/searxng/searxng/compare/719edaf0daaa31625fc1abf868530bcf63729579..5c1d4710d5d768e67081781a748af3c6fb863c58)

PR #3102, opened 2024-01-06
Add loading time information

Myzel394:
## What does this PR do?

To see the durations it took each engine to load, this PR adds this information to the sidebar.
**This is still a WIP**, I opened this draft to see whether this is heading into the right direction :D.

## Why is this change important?

All major search engines show their duration. Having easy access also makes it more transparent to users what engine may be causing long loading times.

## How to test this PR locally?

Run like normally

## Related issues

Closes #3092

PR #3755, opened 2024-08-22
[feat] show duration per engine

### LINKED ISSUE #3092: Show duration per engine / and in total
Google and other engine show the duration it took them to load the results:

image

Add a same hint for the duration of the engine that took the longest to respond beneath the search input. Show a detailed view about the duration of each engine on the right side.

**Additional context**
I can try to open a PR if you're okay with this :)

GenericMale:
## What does this PR do?

Changes the engine-message sidebar entry into a table which also shows the response times of each engine similar to the engine stats page.
Errors from engines are shown on top and the list is sorted by response time (slowest to fastest).
The sidebar entry is always shown but collapsed by default.

See the related issue.
Also serves as a good general information which engines are enabled.

Just search. Also try some engines which run into errors.

## Related issues

Closes #3092 

Conflicts with #3102 which solves the same issue but takes a different approach (work has apparently stopped on it).

## Screenshots

Default Engines:
Default Engines

All Engines with some errors:
Errors


glanham-jr:
Love the design of this, conveys information in a great way.

1. How does this look for Mobile SearXNG? Could you provide a screenshot for smaller widths?
2. Could you give the tldr version how how you determined the length of the progress bar? Is it relative to the longest request length? Relative to the average request duration of each engine for an aggregate value of the server (i.e. avg 1 day google response time from the instance)?
3. This may be a silly question, but I'm not sure how themes work as of today: can the progress bar color be changed by other themes?

GenericMale:
Thanks for looking into this, great questions.

1. Looks like this on mobile:
image
This can get quite long with a lot of engines enabled but it's collapsed by default, so should be fine imho.

2. Everything has been "stolen" from the Engine Stats page, so works exactly the same as the Response Times shown there. So yeah it's relative to the highest response time (the bar for the slowest engine has 100% width).
3. tbh i also have no idea what the current concept of themes is. There is only the simple theme and afaik a new theme has to come with a complete new set of templates and styles (so the entire UI). But the bar color can very easily be changed using a userstyle as it uses the same classes as the Engine Stats page. So theres a good chance existing styles already change this accordingly.  Something like this should do it:
`.bar-chart-bar {
  border: 3px solid #00ff00;
}`

glanham-jr:
Thanks for the answers.

I think as you said the number of response times can be **massive** if we have large list of engines, and the expanded result can be big. For this first design, I think the placement should be at the bottom of the list for this reason, what do you think?

For a subsequent PR after this is merged, I have an idea: In the `Response Time` header, what if we rename it to `Response Time: {{ MAX TIME || UNRESPONSIVE_MESSAGE }`? This way, if we just want to know what the overall response time was, I don't need to expand the header to see the longest response time engine. Also if all the engines crash, then we can probably just have that say `all engines failed`, or nothing at all. This would be nice, as we if we bang `!` search via a single engine, I get all the information I need and don't need to click anything. But this isn't needed for the first design round, and we can discuss after this.

GenericMale:
great idea to show the response time on the section header. just pushed this change. in case all the engines fail the collapsible is expanded by default anyway, so i don't think we have to show the error again.

looks like this now:
image

if everything fails it looks like this right away without having to expand it:
image

tbh i'm not a huge fan of moving the section below the search results on mobile. this way you would have to scroll super far to see engine errors etc and i feel most users wouldnt have an insane amount of engines enabled. guess most likely you just wanna see the errors and slowest engines anyway. so you could open the details, check the top engines (what fits on the screen is most likely all you are interested in) and close it again and enjoy your search results without scrolling.

GenericMale:
makes no sense to show "Response time" on the label when everything failed 😅 it's the old text now in this case

glanham-jr:
I'm fine with not moving the label, and the new changes look good!

searx/templates/simple/elements/engines_msg.html
```diff
@@ -1,27 +1,34 @@
...
+        <td class="response-error">{{- error_type -}}</td>
+      </tr>
+      {%- endfor -%}
+      {%- for timing in timings | reverse -%}
+      <tr>
+        <td class="engine-name"><a href="{{ url_for('stats', engine=timing.engine|e) }}">{{ timing.engine }}</a></td>
+        <td class="response-time">
+          <div class="bar-chart-value">{{- timing.total | round(1) -}}</div>
+          <div class="bar-chart-graph" aria-labelledby="{{timing.engine}}_time" aria-hidden="true">
+            <div class="bar-chart-bar bar{{ (100 * timing.total / timings[-1].total) | round | int }}"></div>
```
glanham-jr: With how its written, the UI is now some business logic to calculate the bar-chart length. For this iteration, I think this approach is simplest, as doing it in the backend will involve mapping the `request.timings` to a new array object with this calculation. I also don't suspect we will be changing this anytime to worry about it too much.

searx/webapp.py
```diff
@@ -789,7 +789,8 @@ def search():
             settings['search']['languages'],
             fallback=request.preferences.get_value("language")
         ),
-        timeout_limit = request.form.get('timeout_limit', None)
+        timeout_limit = request.form.get('timeout_limit', None),
+        timings = request.timings
```
glanham-jr: Is there ever a chance `request.timings` is null, and we may need to do a null check and return an empty list? I don't believe this scenario will happen, but wanted to clarify.

searx/templates/simple/elements/engines_msg.html
```diff
@@ -1,27 +1,34 @@
...
+        <td class="engine-name">
+          <a href="{{ url_for('stats', engine=engine_name|e) }}"
+             title="{{ _('View error logs and submit a bug report') }}">
+             {{- engine_name -}}
+          </a>
+        </td>
+        <td class="response-error">{{- error_type -}}</td>
+      </tr>
+      {%- endfor -%}
+      {%- for timing in timings | reverse -%}
```
glanham-jr: Instead of `reverse` is there a way we could to `sort by desc`? This would ensure we always display in descending time, and if `request.timings` ever decided to change its ordering, it wouldn't affect this UI.

glanham-jr (commented):
A few clarifying questions and a suggestion. Overall looks good.

searx/webapp.py
return42: The problem is, that the internal `timing` objects are now an official API.  Even this list of `timings` is now a part of the API, including properties such as the order of the list

Since there is unfortunately no documentation for this API, developers now have to look at the code of the templates for every further development ... this massively inhibits any further development.

Unfortunately, we have many such quirks from the history of missing delimitation (API) between active code and templating.

It is also a niche feature that is only used by nerds and DevOps in exceptional cases (analysis); we would now pack analysis data into the HTML response with every response, which unnecessarily sizes the HTML response in 99% of requests.

In my opinion, the runtimes of the engines are analysis data that should be published in a different context, if at all ... I personally think they should not be published at all.

return42:
@GenericMale please understand that I am now closing this PR [for the reasons mentioned above](https://github.com/searxng/searxng/pull/3755#discussion_r1729777038). I would still like to thank you very much for this PR.

GenericMale:
@return42 Thank you for taking the time to look at this but it's honestly confusing to me why this was immediately rejected. There is an open feature request from january where you personally asked for a PR to be made and you assigned the "New Feature" label to it 2 days ago. There is also an older PR which basically does the same thing, uses the same objects etc where members of the core maintainer team where involved.

In my personal opinion response times is valuable information for users because it helps them to decide which engines to enable/disable. We could also add a switch in the settings to show them or not (maybe disabled by default).

Can we solve the issue with the internal Timings object by mapping it (and sorting the mapped list) similar to how the UnresponsiveEngine object is mapped?

glanham-jr:
Or to get it merged, ditch the bar code in this version and let's do it later?

GenericMale:
i just pushed a change (which i guess isn't shown here because the PR is closed) which addresses the internal object issue. it's now mapped the same as the UnresponsiveEngines. this is also sorted explicitly server side including figuring out which one is the highest response time.

we can of course also remove the bar code and just show a plain list if this is too much overhead (doesn't look great imho, see #3102). also as i mentioned before, the bar code isn't something i came up with but from [stats.html](https://github.com/searxng/searxng/blob/4f7dd05d99f159111db152f86a583a5e2964d4a4/searx/templates/simple/stats.html#L52)

glanham-jr:
return42 linked to a different comment, I just read [the related feedback now](https://github.com/searxng/searxng/pull/3755#discussion_r1729777038). I think I'm seeing two issues.

1. Business logic tied to UX. This is an easy to solve problem - map the `request.timings` to new objects that will return very limited information: only engine timings, not the whole object.
2. Whether we should even return this information to begin with, as what problem is the actually solving? How does this benefit the user searches?

Regarding point 2, the arguments against it is that this information is only relevant for tech-savvy people who know how to consume this information and maintainers may not even want to allow being returned. Arguments for it are assuming that people know what to do with this information, and to possibly disable engines that take too long. As a developer, this may help to diagnose some basic information on which engines are taking longer (but in reality the `stats` page is likely a better use of analysis rather than individual requests, which we can timeout these requests in settings).

With return42's arguments, I mostly agree. I personally think this has some potential, but I haven't thought about longer term maintenance of the feature and if this is something maintainers even want exposed by default. 

However, I still like the feature, and have personal use cases as a private instance runner that would help me diagnose some VPN issues regarding individual requests.

My proposal is: what if we get this behind a settings flag disabled by default? By default, the API will not return this information and stick to the current experience, and only when enabled will you see full engine timing information. We have the ability to [enable/disable metrics](https://docs.searxng.org/admin/settings/settings_general.html) - I could see this as an additional settings flag.

---

If the above proposal is rejected, then the smallest iteration on this is to just return the total response duration. This is something even Google returns, which we could return as the longest engine response time. This wouldn't bloat the HTML at all.

image

GenericMale:
Point 1 should hopefully be fixed in my branch. Point 2 is for sure up to discussion but I'm just confused why this wasn't already discussed as part of the feature request. Reading this i got the impression that work on it would be welcomed but i guess i misunderstood. If this isn't of value to most users, than it's fine to close this PR obviously.

Just a side note: The timing stats are already exposed to the clients.
So it is also possible to solve this in pure javascript without touching the backend at all and an alternative is to offer this as a userscript.

You can try to execute the following in the browser console when you are on the result page of a search on any instance and you should get the same result (i just hacked this together quickly, plz don't judge):
```js
function showServerTimings (entryList, observer) {
  observer.disconnect();

  const serverTiming = entryList.getEntries().reduce((a, e) => a.concat(e.serverTiming), []);
  serverTiming.sort((a, b) => b.duration - a.duration)

  const total = serverTiming.find(e => e.name === 'total');
  const sidebar = document.getElementById('sidebar');
  sidebar?.insertAdjacentHTML('afterbegin', `
    <div>
      <details class="sidebar-collapsable">
        <summary class="title">Response time: ${(total.duration / 1000).toFixed(1)} seconds</summary>
        <table class="engine-stats">
          <tbody id="timings_container"></tbody>
        </table>
      </details>
    </div>  
  `)

  const container = document.getElementById('timings_container');
  serverTiming.filter(t => t.name.startsWith('total_')).forEach(timing => {
    const name = timing.name.split('_').pop();
    const bar = Math.round(100 * timing.duration / total.duration);
    container?.insertAdjacentHTML('beforeend', `
    <tr>
        <td class="engine-name" style="width: 10rem">
            <a href="./stats?engine=${name}">${name}</a>
        </td>
        <td class="response-time">
            <div class="bar-chart-value">${(timing.duration / 1000).toFixed(1)}</div>
            <div class="bar-chart-graph">
                <div class="bar-chart-serie1 bar${bar}"></div>
            </div>
        </td>
    </tr>`);
  });
}

new PerformanceObserver(showServerTimings).observe({type: "navigation", buffered: true});
```

glanham-jr:
> Just a side note: The timing stats are already exposed to the clients.

Didn't know that - looks like its returned here under the [Server-Timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) header: 
https://github.com/searxng/searxng/blob/4f7dd05d99f159111db152f86a583a5e2964d4a4/searx/webapp.py#L552

`PerformanceObserver` is indeed a valid way to do this without any back-end changes. Your quick script works well on a few public instances I tried. However, it failed on my private instance because access control thinks I'm not the "localhost" and doesn't want to expose it to me. With this, you would need to do update the script to include like `.filter((e) => e.serverTiming).reduce(...)`, and if its an empty list than don't create the header.

This approach has the following downsides:

1. Users who disable JS won't get to see it
2. Presentation is partially generated from back-end, partially front-end

However, what caused me issues, is what I think is a good benefit: access control helps limit how this data is ingested, and doesn't return any new information than what is already publicly available.

return42:
> @return42 Thank you for taking the time to look at this but it's honestly confusing to me why this was immediately rejected. There is an open feature request from january where you personally asked for a PR to be made and you assigned the "New Feature" label to it 2 days ago.

I assume you ref to 

-  https://github.com/searxng/searxng/issues/3092

and I thought about total time

grafik

I reopen this issue / but don't have time to review or further discuss the next time.

searx/webapp.py
```diff
@@ -761,6 +761,11 @@ def search():
         )
     )
 
+    # timings: get engine response times sorted from slowest to fastest
+    timings = webutils.get_engine_timings(result_container.get_timings())
+    # first entry in list is the slowest engine. [0] on object is engine name, [1] is total time
+    max_response_time = timings[0][1] if len(timings) > 0 else None
```
Bnyro: To avoid this additional explaination here, it might be easier to just move the sorting code here and map it into pairs afterwards, i.e.
```suggestion
    # engine_timings: get engine response times sorted from slowest to fastest
    engine_timings =  sorted(result_container.get_timings(), reverse=True, key=lambda e: e.total)
    max_response_time = timings[0].total if len(timings) > 0 else None
    engine_timings_pairs = [(timing.engine, timing.total) for timing in timings]
```

Bnyro:
```
timings[0][1] if len(timings) > 0 else None
```
could be simplified to
```
timings[0][1] if timings else None
```

glanham-jr: Didn't know an empty list returned false, the more you know.

glanham-jr:
@GenericMale Based on return42's feedback, I propose we do the following.

1. Create a new branch from this one (`engine-timings-longest-only` or something)
2. Undo all changes related to passing all the engine information, and only pass the longest engine time + update the UI to only display this longest time
3. After that PR is merged, we can rebase it back into this branch and revisit duration/engine at a later date

This should unblock your work so we can at least see the total response time sooner, which is still useful

GenericMale:
> @GenericMale Based on return42's feedback, I propose we do the following.
> 
>     1. Create a new branch from this one (`engine-timings-longest-only` or something)
> 
>     2. Undo all changes related to passing all the engine information, and only pass the longest engine time + update the UI to only display this longest time
> 
>     3. After that PR is merged, we can rebase it back into this branch and revisit duration/engine at a later date
> 
> 
> This should unblock your work so we can at least see the total response time sooner, which is still useful

Good idea but the problem is that we would have to display the longest time somewhere else. This pull request merges the engine error messages neatly with the engine response times, therefore it makes sense to show the longest time on the section header.
Without that, the overall response time has to be somewhere else i guess. Most search engines show this next to the number of results but searxng doesn't always display that and i don't really want to add an own line somewhere just for the response time.

Bnyro:
I personally think that this PR (as it is now) would be quite useful and very nice to have (for me at least), so we should certainly keep the discussion about it going here.

Bnyro (changes requested):
Almost good to be merged, I just have some minor notes before:
- If there's a crashing engine, it's not properly aligned to the start of the progress bars at the left
- this should probably be placed above the search url in the results
- I'm not sure about having the title named `Response time: x.y seconds`, because this also includes engine errors. But I guess that's okay.

Bnyro (approved):

Myzel394:
This looks much better! Thank you for implementing this @GenericMale!

GenericMale:
> Almost good to be merged, I just have some minor notes before:
> 
>     * If there's a crashing engine, it's not properly aligned to the start of the progress bars at the left
> 
>     * this should probably be placed above the search url in the results
> 
>     * I'm not sure about having the title named `Response time: x.y seconds`, because this also includes engine errors. But I guess that's okay.

thx!!! for the last point, we could maybe change the title to something like `Response time: 1.9s (2 failed)` if at least one engine failed

PR #3756, opened 2024-08-22
[fix] Don't show categories without an active engine

GenericMale:
## What does this PR do?

When all engines in a category are disabled, that category shouldn't be shown.
This seems to be a regression introduced with #2740 

A search will always result in an error if no engine is enabled in a category and it's just unnecessary UI clutter to show them.

Disable all engines in any of the categories and check if it's gone on the result page.
Test with "Search on category select" both enabled and disabled because this was working before changes related to that features where made.

## Related

* [Search on category select without JavaScript #2740](https://github.com/searxng/searxng/pull/2740)
* [[fix] don't show a category if there is no active engine in #2311](https://github.com/searxng/searxng/pull/2311)

### DIFF
--- a/searx/templates/simple/categories.html
+++ b/searx/templates/simple/categories.html
@@ -14,7 +14,7 @@
 <div id="categories" class="search_categories">{{- '' -}}
     <div id="categories_container">
         {%- if not search_on_category_select or not display_tooltip -%}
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Works as expected, thank you!

PR #3761, opened 2024-08-24
[feat] Create a Git pre-commit hook which formats changed Python files

### LINKED ISSUE #3760: [Feat] Add opt-in Git Hook to Format on Commit

As I've been developing features for SearXNG, I keep forgetting to format my changes before I commit a PR. Then I get tazed by the CI to say I need to do these changes. This slows down my development speed, causes CI to fail before I get to more meaningful checks like unit testing, and maintainers need to remind developers to format their PR. 

I see that [Git Hooks](https://www.atlassian.com/git/tutorials/git-hooks) has a `Pre-Commit` hook which can alter the files changes before git commits them. With this, I'd like to enable a `Git-Commit` for formatting code on commit. I also don't need to remember how to run the formatter - I enable the hook and it does the work for me. Given that this is a CI breaking issue and not an optional one, I would much rather have this automated for me.

However, this style can be annoying for some developers, as now commits need to run the formatter for every single commit, increasing the time it takes. For those who do a bunch of tiny commits and squash them later, it can become very annoying. My proposal is to disable this feature by default. To enable the pre-commit hook, we can set some environment variable like `SEARXNG_PRECOMMIT=1`, or some other means. The changes will be included in the developer documentation.

If we get feedback from developers that this is great, we can always have it be `opt-out` at a future point of time.

I could continue formatting without the pre-commit hook, but the provided solution is `opt-in`, so by default everyone will continue to not execute the behavior even after implementing this feature.

**Additional context**

This is for improved Developer Experience, and I believe will help onboard new developers who get tazed by the CI for formatting.

glanham-jr:
## What does this PR do?

1. Update `manage format.python` to accept arguments to only format specific files (if no arguments, will go back to default behavior of formatting everything).
2. Update `manage format.python` to include all information in the `build_msg`.
3. A `.githooks/pre-commit` file is created containing the python formatting hook, passing in only the files to format for python
4. Update `manage` to install/uninstall these git hook updates
5. Update the Makefile to install/uninstall these hooks at the same time as  `pyenv`.
6. Update documentation to add a tip for this new behavior

The only downside to the current implementation is that if anyone else uses pre-commit hooks and we don't know about it today, this will override it. Not sure how much of an issue this really is.

I would have also updated `make format.python` to allow propagating arguments, but I'm not familiar enough with Makefiles to even know how to make this change. Directly calling `manage` works for now. If someone else wants to assist with this, let me know.

With this, I'm also thinking just add a new make option "git-hook.install" and "git-hook.uninstall", and just skip the env enable method and entirely. I'll probably do this and update the PR if the concept is OK to the reviewer.

Improved Developer Experience

Context is provided in related issue.

1. Ensure you have exported `SEARXNG_PRECOMMIT="true"` to your current environment
2. Modify any 2 Python files with a whitespace change
3. Stage the file and commit it
7. Ensure the formatter is ran

## Author's checklist

 - n/a

## Related issues

Closes #3760 

## Screenshots

After running `make install`. This screenshot is outdated - now the git hook only formats the staged files.

image

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+if [ "$SEARXNG_PRECOMMIT" = "true" ]; then
+    # inspired from: https://prettier.io/docs/en/precommit.html#option-4-shell-script
+    staged_files=$(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g')
+    python_files=$(echo "$staged_files" | grep '\.py$')
+    # Only attempt to run the formatter if Python files were changes
+    if [ -n "$python_files" ]; then
+        echo "$python_files" | xargs ./manage format.python
+        echo "$python_files" | xargs git add
+    fi
+fi
--- a/docs/dev/quickstart.rst
+++ b/docs/dev/quickstart.rst
@@ -28,6 +28,13 @@ Here is how a minimal workflow looks like:
 2. *run* your code: :ref:`make run`
 3. *format & test* your code: :ref:`make format.python` and :ref:`make test`
 
+.. tip::
+
+   If you run `make install`, and you export the environment variable
+   `SEARXNG_PRECOMMIT="true"`, a git pre-commit hook will run which will auto format
+   all python files you check into git for you. If you find this would be
+   better to enable by default, let us know.
+
 If you think at some point something fails, go back to *start*.  Otherwise,
 choose a meaningful commit message and we are happy to receive your pull
 request. To not end in *wild west* we have some directives, please pay attention
--- a/manage
+++ b/manage
@@ -94,6 +94,9 @@ pyenv.:
   OK        : test if virtualenv is OK
 format.:
   python    : format Python code source using black
+git.:
+  install   : Installs developer git utilities, such as hooks
+  uninstall : Removes developer git utilities previously installed
 pygments.:
   less      : build LESS files for pygments
 EOF
@@ -316,9 +319,27 @@ pyenv.uninstall() {
 }
 
 format.python() {
-    build_msg TEST "[format.python] black \$BLACK_TARGETS"
-    pyenv.cmd black "${BLACK_OPTIONS[@]}" "${BLACK_TARGETS[@]}"
-    dump_return $?
+    if [ -z "$1" ]; then
+        build_msg TEST "[format.python] black \$BLACK_OPTIONS \$BLACK_TARGETS"
+        pyenv.cmd black "${BLACK_OPTIONS[@]}" "${BLACK_TARGETS[@]}"
+    else
+        build_msg TEST "[format.python] black \$BLACK_OPTIONS $@"
+        pyenv.cmd black "${BLACK_OPTIONS[@]}" "$@"
+    fi
+}
+
+git.install() {
+    if [ -d ".git/hooks" ]; then
+        cp .githooks/pre-commit .git/hooks/pre-commit
+    else
+        build_msg INSTALL "Could not find '.git/hooks', hooks were not installed"
+    fi
+}
+
+git.uninstall() {
+    if [ -f ".git/hooks/pre-commit" ]; then
+        rm .git/hooks/pre-commit
+    fi
 }
 
 # shellcheck disable=SC2119
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ run:  install
 PHONY += install uninstall
 install uninstall:
 	$(Q)./manage pyenv.$@
+	$(Q)./manage git.$@
 
 PHONY += clean
 clean: py.clean docs.clean node.clean nvm.clean test.clean


.githooks/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+if [ "$SEARXNG_PRECOMMIT" = "true" ]; then
+    staged_files=$(git diff --cached --name-only)
+    python_files=$(echo "$staged_files" | grep '\.py$')
+    # Only attempt to run the formatter if Python files were changes
+    if [ -n "$python_files" ]; then
+        make format.python
```
glanham-jr, line 8: I'd like to make this more performant by passing in the list of generated files, but this is not an option at this point of time. I can definitely include this in a follow up PR.

docs/dev/quickstart.rst
```diff
@@ -28,6 +28,13 @@ Here is how a minimal workflow looks like:
 2. *run* your code: :ref:`make run`
 3. *format & test* your code: :ref:`make format.python` and :ref:`make test`
 
+.. tip::
+
+   If you run `make install`, and you export the environment variable
+   `SEARXNG_PRECOMMIT="true"`, a git pre-commit hook will run which will auto format
+   all python files you check into git for you. If you find this would be
+   better to enable by default, let us know.
+
```
return42, line 37: In ReST markup *monospaced* needs to be quoted by ``` `` ```
```
`make install`
`SEARXNG_PRECOMMIT="true"`
```

```
``make install``
``SEARXNG_PRECOMMIT="true"``
```

manage
```diff
@@ -316,9 +319,27 @@ pyenv.uninstall() {
 }
 
 format.python() {
-    build_msg TEST "[format.python] black \$BLACK_TARGETS"
-    pyenv.cmd black "${BLACK_OPTIONS[@]}" "${BLACK_TARGETS[@]}"
-    dump_return $?
```
return42, line 321: The `dump_return $?` should not be removed .. add it right below the both `pyenv.cmd black "${BLACK_OPTIONS[@]}" ...` commands

return42 (commented):
Not sure I like pre-commits; 

- If you implement pre-commits as an opt-in (as here in the PR), then they will not be used.

Lets hear what other developers think

----

I left some comments below to fix the implementation.

- If you implement pre-commits as opt-out, then they can significantly influence the workflow, potentially conflicts in processes such as ``make static.build.commit`` cannot be ruled out

return42:
> Lets hear what other developers think

We had a discussion at [#searxng:matrix.org](https://matrix.to/#/#searxng:matrix.org) --> comment ["It seems like the hooks aren’t quite desirable by default"](https://matrix.to/#/!vxScbLNEAmRvOraXBn:matrix.org/$nKJ3H3Eb-5LeH9M12OKhr1w0uL72lxTf4mogS_4E0z8?via=matrix.org&via=tchncs.de&via=envs.net) .. 

I will therefore close the PR for now ... if this is not OK please contact me again ...

knguyen1:
I disagree with the consensus decision but will respect it.  Having hooks like this is a must, especially with the proliferation of AI coding tools.

Anyone else interested, I run these hooks locally and they seem to work well.  (Add to `.pre-commmit-config.yaml` to your `./.git/info/exclude` file to ignore it from git).  Then run `uv pip install pre-commit` and then `uv run pre-commit install`.

```bash
# .pre-commit-config.yaml
fail_fast: true
default_install_hook_types: [pre-commit, pre-push]

repos:
  # General hooks
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v6.0.0
    hooks:
      - id: trailing-whitespace
      - id: end-of-file-fixer
      - id: check-yaml
      - id: check-json
      - id: check-merge-conflict
      - id: check-added-large-files
        args: ['--maxkb=5000']
      - id: check-case-conflict
      - id: check-merge-conflict
      - id: mixed-line-ending
        args: ['--fix=lf']
  - repo: local
    hooks:
      - id: format-python
        name: format Python code
        entry: ./manage format.python
        language: system
        types: [python]
        pass_filenames: false
        always_run: true
      - id: test-pylint
        name: run pylint
        entry: ./manage test.pylint
        language: system
        types: [python]
        pass_filenames: false
        always_run: true
        stages: [pre-push]
      - id: test-black
        name: check Python code format
        entry: ./manage test.black
        language: system
        types: [python]
        pass_filenames: false
        always_run: true
        stages: [pre-push]
      - id: test-pyright
        name: check Python types
        entry: ./manage test.pyright
        language: system
        types: [python]
        pass_filenames: false
        always_run: true
        stages: [pre-push]
  - repo: local
    hooks:
      - id: nose2
        name: run nose2 tests
        entry: uv run nose2 -F tests.unit
        # tests donn't work with mp plugin
        # entry: uv run --frozen nose2 -qq -F --plugin=nose2.plugins.mp --processes=$(nproc) tests.unit
        language: python
        types: [python]
        pass_filenames: false
        always_run: true
        stages: [pre-push]  # takes a long time, only run on pre-push
```

return42:
> I run these hooks locally and they seem to work well. 

Developers are free to do what helps .. but I can't recommend to implement individual hooks for `./manage test.xyz` .. these are sub-commands that might bee change in the future .. further; there is no guarantee that the number and type of tests (sub-commands) will be identical to the number of tests in the CI.

In the [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) we find the official and *future stable* commands to run test & builds, its very simple ..

    make format test

and for modifications of the client, additionally do a:

    make static.build.commit

These tests & builds can take a very long time to run, and it makes no sense to build the client if no changes have been made to it.   This example shows how difficult it is to make assumptions about the development process of a patch. 

To implement a pre-commit hook, all tests would usually have to be run, which in 90% of cases leads to unnecessary builds and tests (which the developer then has to wait for before they can complete their commit).

A pre-commit would also require a developer to install a variety of tools in the environment in which they developed the patch, which are necessary for performing the tests.  This may still be the case on the desktop workstation, but if a patch has to be developed in a remote session on a remote server, these tools are not available and should not be installed there (on the remote server):

As can be seen from our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide, **we have made every effort to give developers an easy start and to make as few assumptions as possible about the developer's context** .. 

For example, we do not want to specify that a developer must install certain tools... we only recommend it .. and, with the `make` targets, also provide a (automated) installation of these tools.

----

> Having hooks like this is a must, 

[I often work in the commit history, modifing the latest n-commits (e.g. to have a clean history befrore merge).. not sure how pre-commit hooks will in the way ..](https://app.element.io/#/room/#searxng:matrix.org/$nKJ3H3Eb-5LeH9M12OKhr1w0uL72lxTf4mogS_4E0z8)

----

knguyen1:
> These tests & builds can take a very long time to run, and it makes no sense to build the client if no changes have been made to it.

`pre-commit` has a feature called `stages`.  So tests/lints that take a long time has the `stages: [pre-push]` setting.  See `stages: [pre-push]  # takes a long time, only run on pre-push` above.

> these are sub-commands that might bee change in the future

It's kind of trivial to sync pre-commit scripts with CI pipelines, like any other PR review.

> there is no guarantee that the number and type of tests

Now this I agree with you.  And there's no guarantee the testing environment closely matches with the developer's environment.  But package manager does let you target specific python versions, so... e.g. `uv run --frozen --python=3.10 pytest` etc.

```bash
# Correct format - use dots for version
uv run --frozen --python=3.10 pytest

# Other version examples
uv run --frozen --python=3.11 pytest
uv run --frozen --python=3.12 pytest

# You can also use more specific versions
uv run --frozen --python=3.10.13 pytest

# Or just major version
uv run --frozen --python=3 pytest
```

> To implement a pre-commit hook, all tests would usually have to be run, which in 90% of cases leads to unnecessary builds and tests 

`stages: [pre-push]`

> we do not want to specify that a developer must install certain tools

You  can't enforce `pre-commit install` anyway--it's a manual command.  But it does give those who are familiar with the platform another indicator that it is recommended.

return42:
> And there's no guarantee the testing environment closely matches with the developer's environment. 

This statement is incorrect; we have all made every effort to ensure that the test environments in CI and locally on the developers' workstations are identical.

>  But it does give those who are familiar with the platform another indicator that it is recommended.

To become familiar, all need is the look at the our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) guide. 

**ALL WHAT IS NEEDED**:  `make test`

Is it really such hard?

I already mentioned this in my review above: if we force developers to run git hooks, we are forcing them to install the entire development environment and perform tasks that may not even be possible in their development context and may not even be necessary.

-----

SearXNG is an application, not a Python package as typically found in Python projects. The SearXNG environment includes client and server development, and SearXNG is a rolling release (no package revisions!).

As far as possible, we want to apply current standards, which is not always trivial in a development project that has been ongoing for more than 10 years, and sometimes unnecessary.

One thing is clear, however: with such an evolutionarily developed application, the patterns learned in school cannot always be applied.

PR #3770, opened 2024-08-28
[fix] search: titles and content including html brackets are not shown properly

### LINKED ISSUE #3767: Page titles in search results do not show text in angle brackets
**What happened?**
When browsing search results, if a page title has opening and closing angle brackets (e.g. `C Library - <string.h>`), the text in between will not show up (`C Library -`).

**How To Reproduce**
Search up something that would have angle brackets around it (e.g. `string.h`).

**Expected behavior**
Page titles are shown in full (e.g. `C Library - <string.h>`).

Bnyro:
## What does this PR do?
- don't try to convert to convert the title and content from html to text by default
- that never made much sense in my opinion for the following reasons
  - I'm not aware of any search engines that use html in titles
  - The specific engine can handle this perfectly fine by itself in its own module
  - The same applies for the content: I'm aware of some engines that manually call html_to_text already, I don't think that this should be generally done for all results
  - This unnecessarily decreases the performance because the html_to_text algorithm is not the fastest and called a lot unnecessarily.
  - If the algorithm is called twice (which is currently the case for some engines that calls the functions too), it's possible that things get really broken

- currently, result titles are missing text that are between `<` and `>`

- search for `C Library - <string.h>`

## Related issues
closes #3767

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -9,7 +9,6 @@ from typing import List, NamedTuple, Set
 from urllib.parse import urlparse, unquote
 
 from searx import logger
-from searx import utils
 from searx.engines import engines
 from searx.metrics import histogram_observe, counter_add, count_error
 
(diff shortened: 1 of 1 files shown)


return42:
Seems we are not alone with this issue, when I use google origin [https://www.google.de/search?q=C+Library+-+<string.h>](https://www.google.de/search?q=C+Library+-+<string.h>) the third result looks like this .. [Cstring> (string.h)](https://cplusplus.com/reference/cstring/)

grafik

and links to 

grafik

return42 (approved):
> * The same applies for the content: I'm aware of some engines that manually call html_to_text already, I don't think that this should be generally done for all results

I tested all XPath and JSON engines to be sure that we don't have HTML in the results there.

BTW if a JSON engine has HTML in its response, the options `content_html_to_text` and `title_html_to_text` can be set to `True`:

https://github.com/searxng/searxng/blob/cbd86473aadb31c0a12694f56a94c41ab2581142/searx/engines/json_engine.py#L25-L26

Bnyro:
Thank you Markus!

PR #3771, opened 2024-08-28
[fix] self_info: request.user_agent is not a str

AlexanderS:
The `user_agent` attribute of the Flask request object is an instance of the [`werkzeug.user_agent.UserAgent`](https://werkzeug.palletsprojects.com/en/3.0.x/utils/#werkzeug.user_agent.UserAgent) class.

This will fix the following error of the self_info plugin:

```
ERROR:searx.plugins.self_info: Exception while calling post_search
  File "searx/plugins/__init__.py", line 203, in call
  File "searx/plugins/self_info.py", line 31, in post_search
    search.result_container.answers['user-agent'] = {'answer': gettext('Your user-agent is: ') + ua}
TypeError: can only concatenate str (not "UserAgent") to str
```


Bnyro (approved):
Thanks!

PR #3772, opened 2024-08-28
[fix] Do not show DDG user-agent from zero click

AlexanderS:
We do not want to show the user-agent information from the duckduckgo zero click info. This is the user-agent used by searxng and not the user-agent used by the user.

This was already done for the IP address in: 0fb3f0e4aeecf62612cb6568910cf0f97c98cab9


Bnyro (approved):
That's reasonable imo :+1:

PR #3788, opened 2024-08-31
[fix] correct typo in API_WONDOW to API_WINDOW

vojkovic:
## What does this PR do?

Fixes typo, just noticed it while reading.

Should be a quick change.


return42 (approved):
Thanks :+1:

PR #3789, opened 2024-08-31
[feat] add paramaterized package for unit testing

glanham-jr:
## What does this PR do?

 - Install `paramaterized` as part of the dev-requirements.txt
 - Refactor some existing unit tests to showcase how this would be used (also removed the `assertLogs` as it wasn't used).

When I was writing unit tests in prior PRs, I was sad to see that paramaterization of test cases was not supported by default. I've used this style extensively in other languages and projects, and thought it would be nice to include in here. I'm hoping this aids others in reducing boilerplace for writing tests.

 - run `make.unit`, ensure both test cases run

## Author's checklist

 - n/a

## Related issues

 - n/a

## Additional Context

See below for compatibility outlined in the [PyPi docs](https://pypi.org/project/parameterized/). searxng currently uses `unittest`, so `parameterized.expand` works fine for all versions of Python. If we want to switch to `nose2`, that works fine. If we want to switch to `pytest`, well they have their own paramaterization framework that `paramaterized` doesn't try to target, which we would have to do migration for anyway. [unittest2](https://pypi.org/project/unittest2/) is a backport of the latest `unittest` features for Python 3.4 and earlier + PyPy, so likely not our focus. With this, I see no issues with this compatibility chart and the projects current focus.

image

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,4 +22,4 @@ wlc==1.14
 coloredlogs==15.0.1
 docutils<=0.21; python_version == '3.8'
 docutils>=0.21.2; python_version > '3.8'
-
+parameterized==0.9.0
--- a/tests/unit/test_tineye.py
+++ b/tests/unit/test_tineye.py
@@ -5,6 +5,7 @@
 from datetime import datetime
 from unittest.mock import Mock
 from requests import HTTPError
+from parameterized import parameterized
 from searx.engines import load_engines, tineye
 from tests import SearxTestCase
 
@@ -23,13 +24,13 @@ class TinEyeTests(SearxTestCase):  # pylint: disable=missing-class-docstring
         response.raise_for_status.side_effect = HTTPError()
         self.assertRaises(HTTPError, lambda: tineye.response(response))
 
-    def test_returns_empty_list_for_422(self):
+    @parameterized.expand([(400), (422)])
+    def test_returns_empty_list(self, status_code):
         response = Mock()
         response.json.return_value = {}
-        response.status_code = 422
+        response.status_code = status_code
         response.raise_for_status.side_effect = HTTPError()
-        with self.assertLogs(tineye.logger) as _dev_null:
-            results = tineye.response(response)
+        results = tineye.response(response)
         self.assertEqual(0, len(results))
 
     def test_logs_format_for_422(self):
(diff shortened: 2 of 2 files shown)


Bnyro (commented):
From a code perspective this definitely cleans things up, by I am wondering whether it's really worth adding a new dev dependency, since only few engines have tests that require such a use of parameters and a for loop could do the same job to be honest.

Nevertheless `parameterized` could also be useful for testing different search queries with a single function, which is probably already done somewhere else in our engine tests.

In the end, I'm neutral about this.

glanham-jr:
> From a code perspective this definitely cleans things up, by I am wondering whether it's really worth adding a new dev dependency, since only few engines have tests that require such a use of parameters and a for loop could do the same job to be honest.
> 
> Nevertheless `parameterized` could also be useful for testing different search queries with a single function, which is probably already done somewhere else in our engine tests.
> 
> In the end, I'm neutral about this.

I also thought about adding another dependency. Looking at the [history of releases](https://pypi.org/project/parameterized/#history), it seems like the release cadence is quite slow. Further, the source code also seems pretty concise and focused. My guess is once its installed, we likely don't need to upgrade this for a long time unless there is some new feature / fix we want from a new release.

With regard to its usage, I only provided one quick example to show how it would be used. Here are some other areas I can see refactored, just randomly grokking the code:
 - [exceptions](https://github.com/searxng/searxng/blob/master/tests/unit/test_exceptions.py)
 - [locales should definitely be split up for focused tests](https://github.com/searxng/searxng/blob/master/tests/unit/test_locales.py)
 - [hash plugins should be split up +  paramaterized too](https://github.com/searxng/searxng/blob/master/tests/unit/test_plugins.py#L104)
 - [This autocomplete test](https://github.com/searxng/searxng/blob/master/tests/unit/test_query.py#L132)
 - [This webutils test kind of does it already, but with the for loop you described](https://github.com/searxng/searxng/blob/master/tests/unit/test_webutils.py#L38) - but the issue here is that some of the test is in this `data`, and other parts aren't, so now the test is a weird combination of both.

I didn't want to conflate this PR to be too big for a proof of concept, but if it makes more sense to try and refactor tests that seem viable to work with the `paramaterized` library in a single PR here I'd be happy to do so. Let me know your thoughts.

return42:
>  I only provided one quick example to show how it would be used. Here are some other areas I can see refactored, just randomly grokking the code:
> ...
> I didn't want to conflate this PR to be too big for a proof of concept, but if it makes more sense to try and refactor tests that seem viable to work with the `paramaterized` library in a single PR here I'd be happy to do so. Let me know your thoughts.

I agree mostly with you have written :+1: 

> by I am wondering whether it's really worth adding a new dev dependency,

Its just a dev dependency .. not needed in a production environment / from there I think its OK to add this dependency when it helps to clean up the test code.

PR #3792, opened 2024-09-01
[doc] Update own-instance.rst

dwitterer:
Improved English grammar and punctuation for professionalism.

## What does this PR do?

Improves the English grammar and sentence structure of this documentation page.

This PR enhances the professionalism and readability of the modified page and makes this page more clear for new users.

N/A, documentation change.

## Author's checklist

N/A

## Related issues

N/A, will check for duplicates now.
UPDATE: No duplicates, I'm the first to update this page :)

### DIFF
--- a/docs/own-instance.rst
+++ b/docs/own-instance.rst
@@ -2,9 +2,9 @@
 Why use a private instance?
 ===========================
 
-.. sidebar:: Is it worth to run my own instance?
+.. sidebar:: Is running my own instance worth it?
 
-  \.\. is a common question among SearXNG users.  Before answering this
+  \.\.\.is a common question among SearXNG users.  Before answering this
   question, see what options a SearXNG user has.
 
 .. contents::
@@ -12,13 +12,13 @@ Why use a private instance?
    :local:
    :backlinks: entry
 
-Public instances are open to everyone who has access to its URL.  Usually, these
+Public instances are open to everyone who has access to their URL.  Usually, they
 are operated by unknown parties (from the users' point of view).  Private
-instances can be used by a select group of people.  It is for example a SearXNG of
-group of friends or a company which can be accessed through VPN.  Also it can be
-single user one which runs on the user's laptop.
+instances can be used by a select group of people, such as a SearXNG instance for a
+group of friends, or a company which can be accessed through a VPN.  Instances can also be
+single-user instances, which run locally on the user's machine.
 
-To gain more insight on how these instances work let's dive into how SearXNG
+To gain more insight on how these instances work, let's dive into how SearXNG
 protects its users.
 
 .. _SearXNG protect privacy:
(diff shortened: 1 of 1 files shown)


unixfox (approved):

docs/own-instance.rst
```diff
@@ -80,7 +80,7 @@ Always use an instance which is operated by people you trust.  The privacy
 features of SearXNG are available to users no matter what kind of instance they
 use.
 
-If someone is on the go or just wants to try SearXNG for the first time public
-instances are the best choices.  Additionally, public instance are making a
-world a better place, because those who cannot or do not want to run an
-instance, have access to a privacy respecting search service.
+For those on the go, or just wanting to try SearXNG for the first time, public
+instances are the best choice.  Additionally, public instance are making the
+world a better place by giving those who cannot, or do not, want to run an
```
return42: > "those who cannot, or do not, want to run"

I can't judge, but is this punctuation really correct in English? Why not..

> "those who can't or don't want to run"

docs/own-instance.rst
```diff
@@ -2,62 +2,62 @@
...
 instance.  This means that the user of the public instance does not know whether
-their requests are logged, aggregated and sent or sold to a third party.
+their requests are logged, aggregated, and sent or sold to a third party.
 
-Also, public instances without proper protection are more vulnerable to abusing
-the search service, In this case the external service in exchange returns
-CAPTCHAs or bans the IP of the instance.  Thus, search requests return less
+Also, public instances without proper protection are more vulnerable to abuse of
+the search service which may cause the external service to enforce
+CAPTCHAs, or to ban the IP address of the instance.  Thus, search requests would return less
```
return42: > "to enforce CAPTCHAs, or to ban the IP address"

is the additional comma before the “or” necessary?

return42 (commented):
Thank you for improving SearXNG's documentation / my English is not good enough to correct a professional text, but I noticed two things that I would have expected differently .. see my comments below:

Bnyro (approved):

return42:
@dwitterer I like to merge this PR / but first please give short feedback to my questions from my review / thanks :+1:

docs/own-instance.rst
dwitterer: Hi, sorry, was a bit busy. I'm not a linguist but I'm pretty sure both are correct/acceptable :) I'd keep it with the commas though, as I think it's a subjunctive clause and it flows better. I **_did_** make an oopsie though as it should be:

"Those who cannot, or do not want to, run" - moving the comma :)

docs/own-instance.rst
dwitterer: I think you're right in this case actually. The comma before "or" can be omitted, but we should add one between "service" and "which" as it's a "non-restrictive" clause. I'll see if I can edit the PR.

docs/own-instance.rst
dwitterer: I'd keep it with the commas though, as I think it's a subjunctive clause and it flows better.

dwitterer:
@return42 See my responses above. I've also pushed some changes to my fork to fix a couple things, including what you said, so hopefully the PR should have updated to reflect this. If you click on 'commits' then it shows up for me :) Let me know if you are happy with this new version.

Issue #3793, opened 2024-09-02
Bug: Nu HTML Checker Failures across most pages


glanham-jr:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

Repository: https://github.com/searxng/searxng
Branch: master
Version: 2024.8.31+

**How did you install SearXNG?**

```
make run
```

**How To Reproduce**

Validate the results of HTML pages using the [Nu HTML Checker](https://validator.w3.org/nu/) (you can read more details in the [about page](https://validator.nu/about.html)). This can be done one of two ways:
1. Use the online tool linked above. You can upload the HTML directly as a file/text, or you can paste the URL of the site.
2. Download the `validatornu` package yourself and test directly that way. However, I needed to save the files as HTML files from my browser to test. You can try running this against the raw templates, but it will throw a lot of false positives.

**Expected behavior**

The Nu HTML Checker reports back 0 **error** results. While it is nice to clean up the info alerts, it is less of an issue.

Not sure if any of the errors are expected behavior in some of these cases either, and is something that may need investigating.

**Screenshots & Logs**

I've pasted the general reports of running `validatornu` package against local files. These are what I ran at the time of creating this report, and may have changed. I also haven't checked each and every possible template, but provided some of the easier to find pages to show what the errors may look like.

**Common across all pages**

```
"file:/tmp/":1.16-2.64: error: Bad value “en-EN” for attribute “lang” on element “html”: Bad region subtag.
"file:/tmp/":4.3-4.26: info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
"file:/tmp/":12.3-12.65: error: A “meta” element with an “http-equiv” attribute whose value is “X-UA-Compatible” must have a “content” attribute with the value “IE=edge”.
"file:/tmp/":14.3-14.148: info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
"file:/tmp/":15.3-15.78: info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
"file:/tmp/":17.3-17.569: error: Attribute “client_settings” not allowed on element “script” at this point.
"file:/tmp/":19.3-19.136: info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
"file:/tmp/":22.3-22.118: info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.
```

I've only uploaded an image of what this looks like if you use the online tool, the rest I didn't use.

[...]

PR #3797, opened 2024-09-04
Update html code to properly set closing tag on void elements

glanham-jr:
## What does this PR do?

- Removes `/>` ending tags for [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements) and replaces them with `>`.

 - Part of the larger cleanup to cleanup invalid HTML throughout the codebase.

 - Ensure unit tests run
 - `make run` to ensure none of the affected pages display invalid
 - Check the rendered pages against the NU Html Checker

## Author's checklist

 - [ ] Manually check all void elements for proper ending tags. I did a quick regex sweep to get the simplest cases, and now need to validate each file manually.

## Related issues

#3793

### DIFF
--- a/docs/admin/api.rst
+++ b/docs/admin/api.rst
@@ -84,9 +84,9 @@ HTML of the site.  URL of the SearXNG instance and values are customizable.
 .. code:: html
 
    <form method="post" action="https://example.org/">
-     <!-- search      --> <input type="text" name="q" />
-     <!-- categories  --> <input type="hidden" name="categories" value="general,social media" />
-     <!-- language    --> <input type="hidden" name="lang" value="all" />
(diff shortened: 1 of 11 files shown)


return42 (approved):
Thanks a lot for this patch .. in any case its an real improvement of the generated HTML markup :+1: .. I just squashed a nit into your commit / see [diff](https://github.com/searxng/searxng/compare/c290dec741fc51bb32832ee953610dd91d102f70..3fa56ca079e66a6ccf04a94a3d8adca466155b1b)

glanham-jr:
@return42 Thanks for merging. I only searched for the easiest cases (used a regex search for invalid HTML), but we can patch as we go.

PR #3798, opened 2024-09-04
[feat] engine: add right dao

0xhtml:
## What does this PR do?

Add Right Dao as an engine.
https://rightdao.com

Adding new engine.
Information on engine: https://seirdy.one/posts/2021/03/10/search-engines-with-own-indexes/ (ctrl-f for right dao)

query `!rd …`

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1553,9 +1553,9 @@ engines:
     paging: true
     page_size: 12
     search_url: https://rightdao.com/search?q={query}&start={pageno}
-    results_xpath: //div[@class="description"]
-    url_xpath: ../div[@class="title"]/a/@href
-    title_xpath: ../div[@class="title"]
+    results_xpath: //div[contains(@class, "description")]
+    url_xpath: ../div[contains(@class, "title")]/a/@href
+    title_xpath: ../div[contains(@class, "title")]
     content_xpath: .
     categories: general
     shortcut: rd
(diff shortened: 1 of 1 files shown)


searx/settings.yml
```diff
@@ -1548,6 +1548,24 @@ engines:
...
     disabled: true
 
+  - name: right dao
+    engine: xpath
+    paging: true
+    page_size: 12
+    search_url: https://rightdao.com/search?q={query}&start={pageno}
+    results_xpath: //div[@class="description"]
+    url_xpath: ../div[@class="title"]/a/@href
+    title_xpath: ../div[@class="title"]
```
```suggestion
    results_xpath: //div[contains(@class, "description")]
    url_xpath: ../div[contains(@class, "title")]/a/@href
    title_xpath: ../div[contains(@class, "title")]
```
prevents in from breaking in case they add additional classes to these html elements.

searx/settings.yml
```diff
@@ -1548,6 +1548,24 @@ engines:
     page_size: 25
     disabled: true
 
+  - name: right dao
+    engine: xpath
+    paging: true
+    page_size: 12
+    search_url: https://rightdao.com/search?q={query}&start={pageno}
+    results_xpath: //div[@class="description"]
```
Bnyro: Why don't you use `//div[contains(@class, "item")]` here to directly match the parent element? That way you wouldn't need to use `../` below, would you?

0xhtml: The class item is also used for headers and news stories in the results. Only genuine results have a description.

Bnyro (approved):

Issue #3589, opened 2024-06-20
Yep and Yep images are broken


Austin-Olacsi:
the api works in a real browser, but without a real browser we get cloudflared.

below i try to get the search results using curl. it does NOT work if using regular curl, however, it does work when we use curl-impersonate-chrome instead:

`curl-impersonate-chrome -H "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8" -H "Accept-Language: en-US,en;q=0.5" -H "Connection: keep-alive" -H "Host: api.yep.com" -H "Sec-Fetch-Dest: document" -H "Sec-Fetch-Mode: navigate" -H "Sec-Fetch-Site: none" -H "Sec-Fetch-User: ?1" -H "TE: trailers" -H "Upgrade-Insecure-Requests: 1" -H "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36" "https://api.yep.com/fs/2/search?client=web&gl=CA&no_correct=false&q=undefined+variable+javascript&safeSearch=off&type=web"`

Austin-Olacsi:
This probably isn't possible to fix without adding a new dependency. requests and httpx can't impersonate a real browser fingerprint.

@dalf @return42 would it be acceptable to add this program as a new dependency for searxng: https://pypi.org/project/curl-cffi/

That should also make it possible to add other services that would otherwise detect us as a bot.

return42:
Sadly that is not such easy .. I'm not the expert, but as far I remember:

About fingerprint: 

there are various methods to fingerprint and we do not know what kind of fingerprint method is implemented on the side of the origin search engine. Here are some links where we already talked about:  

- https://github.com/searxng/searxng/issues/2246#issuecomment-1467895556
- https://github.com/searxng/searxng/pull/2266

About curl-impersonate:

- fingerprint isn't limited on cipher list, but we had already some discussions about (see links below)
- curl-cffi needs to be built into the searx.network and replace the httpx there.  There is currently a revision of searx.network (see link below)
- If my memory serves me right, @dalf had already thought about using a curl implementation in the searx.network ... but this is currently not the case in branch master nor the PR-2685

- https://github.com/searxng/searxng/pull/2685
- https://github.com/searxng/searxng/issues/3068#issuecomment-1873000693
[...]

PR #3813, opened 2024-09-11
[fix] yep engine: remove links to other engines

0xhtml:
## What does this PR do?

This PR removes links to other search engines from the Yep results.

Yep includes links to search for the same query on Google and other search engines as a result in the search result.

Search with `!yep` and compare the results w/ and w/o this PR

### DIFF
--- a/searx/engines/yep.py
+++ b/searx/engines/yep.py
@@ -67,6 +67,8 @@ def response(resp):
 
     for result in resp.json()[1]['results']:
         if search_type == "web":
+            if result['type'] != 'Organic':
+                continue
             results.append(_web_result(result))
         elif search_type == "images":
             results.append(_images_result(result))


Bnyro (approved):
Can confirm that by investigating their API responses at yep.com.

I generally can't get the engine to work via SearXNG due to Captchas, is there any trick how to get it run?

0xhtml:
I'm using my custom network implementation with curl-impersonate/curl_cffi as discussed in #3589. Take a look at [0xhtml/search-engine](https://github.com/0xhtml/search-engine). Though, to run my implementation w/ only one engine you have to patch the relevant [code](https://github.com/0xhtml/search-engine/blob/f04da5fb4b97a01fd5c76926e6d58598fedafe85/searchengine/engines.py#L442-L456).

PR #3814, opened 2024-09-11
[feat] gitlab: implement dedicated module

### LINKED ISSUE #3810: GitLab engine mixes up titles, URLs and produces invalid results
Latest Docker image: `2024.9.6+153a28ccd`

Official Docker image (https://hub.docker.com/r/searxng/searxng)

**What happened?**
When searching with the default Gitlab engine, the titles and URLs don't match up. Additionally, some results shouldn't be showing up at all.

For most or all search results, the title doesn't match the URL of the same result, but rather belongs to some other result. There is no clear pattern, as sometimes it's off by one, sometimes it's a title that belongs to a result all the way down the page.

Additionally, some results don't seem to match the query, mostly users or groups.

Example: `!gitlab searx` (see screenshot)

The group `gugal` (https://gitlab.com/groups/gugal) owns a repository that matches the query `searx`. That's technically an indirect match, but it should rather return that repository, rather than the group. And it should certainly rank the repo much higher than the group/

Additionally, a result for https://gitlab.com/uhidontkno doesn't match at all. Nowhere on that page do I see a reference to the search term.

Looking at the raw API results from Gitlab (`curl -s 'https://gitlab.com/api/v4/projects?search=searx&page=1'`), those groups and users don't even show up in the 4 pages of results.

**How To Reproduce**
- Search for `!gitlab searx`
- I first noticed it on my own instance, but the same happens at e.g. https://metasearx.com/?q=%21gitlab%20searx&categories=none&language=en or https://searx.bndkt.io/search?q=%21gitlab%20searx&language=all&time_range=&safesearch=0&categories=none

**Expected behavior**

- Title and URL should match up correctly
- Users and groups that have no direct (name, description, etc.) match with the query, and instead only match indirectly through owning a repository should not be ranked as high.

**Screenshots & Logs**
image

Bnyro:
## What does this PR do?
- Implement a dedicated module for Gitlab using the `packages.html` template

## Related issues
closes #3810
see https://github.com/searxng/searxng/issues/3268

### DIFF
--- a/docs/dev/engines/online/gitlab.rst
+++ b/docs/dev/engines/online/gitlab.rst
@@ -1,0 +1,8 @@
+.. _gitlab engine:
+
+======
+GitLab
+======
+
+.. automodule:: searx.engines.gitlab
+   :members:
--- a/searx/engines/gitlab.py
+++ b/searx/engines/gitlab.py
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine to search in collaborative software platforms based on GitLab_ with
+the `GitLab REST API`_.
+
+.. _GitLab: https://about.gitlab.com/install/
+.. _GitLab REST API: https://docs.gitlab.com/ee/api/
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`base_url`
+
+Optional settings are:
+
+- :py:obj:`api_path`
+
+.. code:: yaml
+
+  - name: gitlab
+    engine: gitlab
+    base_url: https://gitlab.com
+    shortcut: gl
+    about:
+      website: https://gitlab.com/
+      wikidata_id: Q16639197
+
+  - name: gnome
+    engine: gitlab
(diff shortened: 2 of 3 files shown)


Bnyro:
Oh no, my auto-formatter broke the yaml code ...

return42 (changes requested):
LGTM / thanks :+1: .. I pushed a WIP commit with some suggestions .. if you are fine with, squash them  into your initial commit.

I removed the timeout of 10 sec (longest time I had with gnome and gitlab hosters was 1,5sec) and I added the engine to the online docs.

Bnyro:
> I pushed a WIP commit with some suggestions .. if you are fine with, squash them into your initial commit.
>
Nice, thank you!

> I removed the timeout of 10 sec (longest time I had with gnome and gitlab hosters was 1,5sec) and I added the engine to the online docs.
>
Yes, I agree with that, I wondered about the 10 second timeout too. If somebody uses a very slow Gitlab instance, the timeout can still be changed, but for the official and Gnome's instance I can confirm fast responses.

PR #3823, opened 2024-09-14
add get_embeded_stream_url to searx.utils

Austin-Olacsi:
## What does this PR do?

adds get_embeded_stream_url to searx.utils

used in multiple engines, so makes sense to have it in searx.utils.
We can also have embeds for more sites than just youtube (they do appear in search results sometimes.)

test embeds for each site:
!ddv cats site:tiktok.com

### DIFF
--- a/searx/engines/duckduckgo_extra.py
+++ b/searx/engines/duckduckgo_extra.py
@@ -7,6 +7,7 @@ DuckDuckGo Extra (images, videos, news)
 from datetime import datetime
 from typing import TYPE_CHECKING
 from urllib.parse import urlencode
+from searx.utils import get_embeded_stream_url
 
 from searx.engines.duckduckgo import fetch_traits  # pylint: disable=unused-import
 from searx.engines.duckduckgo import (
@@ -108,7 +109,7 @@ def _video_result(result):
         'title': result['title'],
         'content': result['description'],
         'thumbnail': result['images'].get('small') or result['images'].get('medium'),
-        'iframe_src': result['embed_url'],
+        'iframe_src': get_embeded_stream_url(result['content']),
         'source': result['provider'],
         'length': result['duration'],
         'metadata': result.get('uploader'),
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -17,7 +17,7 @@ from os.path import splitext, join
 from random import choice
 from html.parser import HTMLParser
 from html import escape
-from urllib.parse import urljoin, urlparse
+from urllib.parse import urljoin, urlparse, parse_qs, urlencode
 from markdown_it import MarkdownIt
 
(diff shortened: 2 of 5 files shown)


searx/utils.py
```diff
@@ -615,6 +615,48 @@ def _get_fasttext_model() -> "fasttext.FastText._FastText":  # type: ignore
     return _FASTTEXT_MODEL
 
 
+def _get_iframe_src(url):
+    parsed_url = urlparse(url)
+    iframe_src = None
```
return42: I think the function should be renamed to something similar to ``def get_embeded_stream_url(url)`` .. and a doc-string is issing that describes what distinguishes the input URL from the returned one .. what streamning services supported (or what types of URLs).

searx/engines/duckduckgo_extra.py
```diff
@@ -108,7 +109,7 @@ def _video_result(result):
         'title': result['title'],
         'content': result['description'],
         'thumbnail': result['images'].get('small') or result['images'].get('medium'),
-        'iframe_src': result['embed_url'],
+        'iframe_src': _get_iframe_src(result['content']),
```
return42: Is it really needed to call the function when the `result['embed_url']` already holds the url? .. Or is `result['embed_url']` sometimes failed?

return42 (changes requested):
@Austin-Olacsi thanks a lot for digging in the embedded URLs .. I left some comments below ..

searx/engines/duckduckgo_extra.py
Austin-Olacsi: It's because the result['embed_url'] only works with youtube videos, whereas the function works with more services.

return42: May I oversee something, but here we are in the ddg-video results and when parsing the result from ddg the `result['embed_url']` is already a *embedable* URL .. while `get_embeded_stream_url` was made to *"Converts a standard video URL into its embed format. Supported services include Youtube, Facebook, Instagram, TikTok, and Dailymotion."* .. and this is not needed for ddg-video results since they already serve a embedable URL / or I'm wrong?

Austin-Olacsi: what i'm saying:
result['embed_url'] is already a embedable URL, but ONLY for youtube videos.
So, if i search !ddv cats site:tiktok.com the result['embed_url'] does not work.
the function does work in this case.

return42: > So, if i search !ddv cats site:tiktok.com the result['embed_url'] does not work.

It was not clear to me, that we fix an issue in the ddg-video engine .. thanks :+1:

return42 (approved):

PR #3830, opened 2024-09-15
fix HTML generated by the simple templates

return42:
This is another PR that improves the quality of the generated HTML markup from the ``simple`` templates.  The discussion on this topic is conducted under the incident https://github.com/searxng/searxng/issues/3793

> ``/preferences``
> 
> There are too many to list out, most of which point to aria-labelledby.

In its previous implementation, the macro ``checkbox_onoff_reversed`` always created an ``aria-labelledby`` attribute, even if there was no descriptive tag with the generated ID (used as the value of the ``aria-labelledby``).

Before this patch, the Nu-HTML-Checker [1] reported 255 issues of this type::

    The aria-labelledby attribute must point to an element in the same document. (255)

[1] https://validator.w3.org/nu/

---

For more details read the commit messages in this patch series

### DIFF
--- a/searx/templates/simple/preferences/answerers.html
+++ b/searx/templates/simple/preferences/answerers.html
@@ -33,10 +33,10 @@
     {%- for plugin in plugins -%}
       {%- if plugin.preference_section == 'query' -%}
         <tr>{{- '' -}}
-          <td class="checkbox-col">{{- checkbox_onoff_reversed('plugin_' + plugin.id, plugin.id not in allowed_plugins) -}}</td>{{- '' -}}
+          <td class="checkbox-col">{{- checkbox_onoff_reversed('plugin_' + plugin.id, plugin.id not in allowed_plugins, 'plugin_labelledby' + plugin.id) -}}</td>{{- '' -}}
           <td>{{ plugin.query_keywords|join(', ') }}</td>{{- '' -}}
           <td>{{ _(plugin.name) }}</td>{{- '' -}}
-          <td>{{ _(plugin.description) }}</td>{{- '' -}}
+          <td id="{{ 'plugin_labelledby' + plugin.id }}">{{ _(plugin.description) }}</td>{{- '' -}}
           <td>{{ plugin.query_examples }}</td>{{- '' -}}
         </tr>
       {%- endif -%}
--- a/searx/templates/simple/preferences.html
+++ b/searx/templates/simple/preferences.html
@@ -27,11 +27,11 @@
   {%- endif -%}
 {%- endmacro -%}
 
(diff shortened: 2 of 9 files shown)


searx/templates/simple/preferences/answerers.html
```diff
@@ -33,10 +33,10 @@
     {%- for plugin in plugins -%}
       {%- if plugin.preference_section == 'query' -%}
         <tr>{{- '' -}}
-          <td class="checkbox-col">{{- checkbox_onoff_reversed('plugin_' + plugin.id, plugin.id not in allowed_plugins) -}}</td>{{- '' -}}
+          <td class="checkbox-col">{{- checkbox_onoff_reversed('plugin_' + plugin.id, plugin.id not in allowed_plugins, 'plugin_labelledby' + plugin.id) -}}</td>{{- '' -}}
           <td>{{ plugin.query_keywords|join(', ') }}</td>{{- '' -}}
           <td>{{ _(plugin.name) }}</td>{{- '' -}}
-          <td>{{ _(plugin.description) }}</td>{{- '' -}}
+          <td id="{{ 'plugin_labelledby' + plugin.id }}">{{ _(plugin.description) }}</td>{{- '' -}}
```
glanham-jr, line 39: We have the same ID used in the preferences template. Is there any risk in ID collision in answers and preferences?

glanham-jr (approved):
Just one clarification comment, but otherwise looks good to me.

searx/templates/simple/preferences/answerers.html
return42: The `id` is build up from the prefix `plugin_labelledby` plus the value from the `plugin.id` .. since IDs of plugins are unique, I don't expect a ID collision from those of *answers* and *preferences*.

return42:
- from https://github.com/searxng/searxng/issues/3793#issuecomment-2351689483

>  also looked at the remaining aria-labelledby issues from the preferences. They are created from the following sections of code: ...

@glanham-jr thanks for pointing out, I added commit [e3083f4](https://github.com/searxng/searxng/pull/3830/commits/e3083f464330f440b53f631b9efa906d6660f6c1) to this PR.  And i pushed this branch on my instance --> https://validator.w3.org/nu/?doc=https%3A%2F%2Fdarmarit.org%2Fsearx%2Fpreferences

glanham-jr:
No other comments, the PR looks good along with the validations. As a nit, I'd suggest the PR is titled `Fix aria-labelledby HTML generation by the simple templates`, as the PR specifically addresses these issues

Bnyro (approved):

PR #3841, opened 2024-09-20
[feat] simple theme: pure black theme style

### LINKED ISSUE #1461: OLED/blackout dark mode?
The current dark mode is rather a grey mode. Would be cool to have an extra black/oled option

A theme that uses black(rgb code: #000000) as its background

**Additional context**

Bnyro:
## What does this PR do?
- This adds a pure black theme (based on the dark theme)
- The theme looks quite nice with OLED devices

## Related issues
- closes https://github.com/searxng/searxng/issues/1461


return42:
What comes into my mind about *"theme styles"*:

- the existing themes are named not by color name (``light``, ``dark``) .. should we name this theme ``oled dark``?
- https://github.com/searxng/searxng/issues/3842  .. should we solve first?
- @mrpaulblack and @vojkovic also have very beautiful themes ... does it makes sense if we coordinate the selection of themes that we already know today. [comment](https://github.com/searxng/searxng/issues/3724#issuecomment-2282183579)?

.. about *"theme"*:

- @mrpaulblack has customized [search.less](https://github.com/paulgoio/searxng/blob/main/src/less/search.less) .. 

My thoughts on this certainly go far beyond this PR, but I would like to make sure that we have an approach to new themes & styles that scales with requirements that go beyond this PR. 

I think this PR still scales quite well because it only works with definitions.less.  Nevertheless, we should think about more options, including “themes” and not just “styles” ... do any of you have any thoughts on this?

BTW: the needed modifications to:

- [`searx/preferences.py`](https://github.com/searxng/searxng/pull/3841/files#diff-c68b1e3201d656b33513a919fceae6b6fb776aa31f2b8f5aa80ac8eb4da64c0d)
- [`searx/searxng.msg`](https://github.com/searxng/searxng/pull/3841/files#diff-c48645ae25e0b43d291e66cf32069cccb7308805d8f1c5e3c9b5e13386a2facb)
- [`searx/settings_defaults.py`](https://github.com/searxng/searxng/pull/3841/files#diff-6a82f4161887c10ef8e86cf63056ae5832896914bd0240628ec076cd075c4fff)
- [`searx/templates/simple/preferences/theme.html`](https://github.com/searxng/searxng/pull/3841/files#diff-b53ee577a752a7a000be859c6560f0c95308ba8cd69f5e112e1108b53ec4b456)

.. that even the “styles” still don't scale well today.

Bnyro:
>  the existing themes are named not by color name (`light`, `dark`) .. should we name this theme `oled dark`?
> 
I've named it `black` to keep all theme names one word, but oled dark would be fine with me too.

> [Remove hardcoded colors from Less files #3842](https://github.com/searxng/searxng/issues/3842)  .. should we solve first?
> 
I think it doesn't matter in which order we do these since the black theme is just inheriting every color from the dark theme (except the background). If we plan to integrate any other color theme style, we should however solve #3842 first.

> @mrpaulblack and @vojkovic also have very beautiful themes ... does it makes sense if we coordinate the selection of themes that we already know today. [comment](https://github.com/searxng/searxng/issues/3724#issuecomment-2282183579)?
> 
I think it's simple to port vojkovic's themes since they only change the colors (as far as I can see), whereas Paul's theme also changes CSS like the border radius of the search bar, which is going to be way more difficult to implement.

return42 (approved):

PR #3846, opened 2024-09-26
refactor unit tests to utilize paramaterized and break down monolithic tests

glanham-jr:
## What does this PR do?

After the initial work done in #3789, this PR fully realizes a rework of the unit tests to utilize the parameterize. I've avoided trying to move tests to paramaterize which test very specific behavior, even if some of the setup is the same.

 - For tests which perform the same arrange/act/assert pattern but with different data, the data portion has been moved to the `paramaterized.expand` fields
 - For monolithic tests which performed multiple arrange/act/asserts, they have been broken up into different unit tests
 - When possible, change generic `assertEqual` to more concise asserts (i.e. `assertIsNone`)

This work ultimately is focused on creating smaller and more concise tests. While paramaterized may make adding new configurations for existing tests easier, that is just a beneficial side effect. The main benefit that I see is that smaller tests are easier to reason about, meaning they are easier to debug when they start failing. This improves the developer experience in debugging what went wrong when refactoring the project.

Total number of tests went from 192 -> 259; or, broke apart larger tests into 69 more concise ones.

See below for when a paramaterized test fails - it shows exactly the index of data which failed, and appended some of the information to the parameter. This pinpoints the exact issue.

image

 - `make test.unit`

## Author's checklist

 - n/a

## Related issues

 - n/a

### DIFF
--- a/searx/query.py
+++ b/searx/query.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=invalid-name, missing-module-docstring, missing-class-docstring
 
+from __future__ import annotations
 from abc import abstractmethod, ABC
 import re
 
@@ -258,7 +259,7 @@ class RawTextQuery:
         FeelingLuckyParser,  # redirect to the first link in the results list
     ]
 
-    def __init__(self, query, disabled_engines):
+    def __init__(self, query: str, disabled_engines: list):
         assert isinstance(query, str)
         # input parameters
         self.query = query
--- a/tests/unit/test_exceptions.py
+++ b/tests/unit/test_exceptions.py
@@ -1,42 +1,36 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=missing-module-docstring
 
+from parameterized import parameterized
 from tests import SearxTestCase
 import searx.exceptions
 from searx import get_setting
 
 
 class TestExceptions(SearxTestCase):  # pylint: disable=missing-class-docstring
-    def test_default_suspend_time(self):
-        with self.assertRaises(searx.exceptions.SearxEngineAccessDeniedException) as e:
-            raise searx.exceptions.SearxEngineAccessDeniedException()
+    @parameterized.expand(
+        [
+            searx.exceptions.SearxEngineAccessDeniedException,
+            searx.exceptions.SearxEngineCaptchaException,
+            searx.exceptions.SearxEngineTooManyRequestsException,
+        ]
+    )
+    def test_default_suspend_time(self, exception):
+        with self.assertRaises(exception) as e:
+            raise exception()
         self.assertEqual(
             e.exception.suspended_time,
-            get_setting(searx.exceptions.SearxEngineAccessDeniedException.SUSPEND_TIME_SETTING),
+            get_setting(exception.SUSPEND_TIME_SETTING),
         )
 
-        with self.assertRaises(searx.exceptions.SearxEngineCaptchaException) as e:
-            raise searx.exceptions.SearxEngineCaptchaException()
-        self.assertEqual(
-            e.exception.suspended_time, get_setting(searx.exceptions.SearxEngineCaptchaException.SUSPEND_TIME_SETTING)
-        )
-
-        with self.assertRaises(searx.exceptions.SearxEngineTooManyRequestsException) as e:
-            raise searx.exceptions.SearxEngineTooManyRequestsException()
-        self.assertEqual(
-            e.exception.suspended_time,
-            get_setting(searx.exceptions.SearxEngineTooManyRequestsException.SUSPEND_TIME_SETTING),
-        )
-
-    def test_custom_suspend_time(self):
-        with self.assertRaises(searx.exceptions.SearxEngineAccessDeniedException) as e:
-            raise searx.exceptions.SearxEngineAccessDeniedException(suspended_time=1337)
+    @parameterized.expand(
+        [
+            searx.exceptions.SearxEngineAccessDeniedException,
+            searx.exceptions.SearxEngineCaptchaException,
+            searx.exceptions.SearxEngineTooManyRequestsException,
+        ]
+    )
+    def test_custom_suspend_time(self, exception):
+        with self.assertRaises(exception) as e:
+            raise exception(suspended_time=1337)
         self.assertEqual(e.exception.suspended_time, 1337)
-
-        with self.assertRaises(searx.exceptions.SearxEngineCaptchaException) as e:
-            raise searx.exceptions.SearxEngineCaptchaException(suspended_time=1409)
-        self.assertEqual(e.exception.suspended_time, 1409)
-
-        with self.assertRaises(searx.exceptions.SearxEngineTooManyRequestsException) as e:
-            raise searx.exceptions.SearxEngineTooManyRequestsException(suspended_time=1543)
-        self.assertEqual(e.exception.suspended_time, 1543)
--- a/tests/unit/test_locales.py
+++ b/tests/unit/test_locales.py
@@ -2,6 +2,8 @@
 # pylint: disable=missing-module-docstring
 """Test some code from module :py:obj:`searx.locales`"""
 
+from __future__ import annotations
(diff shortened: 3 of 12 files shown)


tests/unit/test_exceptions.py
```diff
@@ -1,42 +1,36 @@
...
+    @parameterized.expand(
+        [
+            searx.exceptions.SearxEngineAccessDeniedException,
+            searx.exceptions.SearxEngineCaptchaException,
+            searx.exceptions.SearxEngineTooManyRequestsException,
+        ]
+    )
+    def test_custom_suspend_time(self, exception):
+        with self.assertRaises(exception) as e:
+            raise exception(suspended_time=1337)
```
glanham-jr, line 35: I don't think the actual number is important in this test, so I didn't bother trying for the other random values

tests/unit/test_locales.py
```diff
@@ -13,98 +15,104 @@ class TestLocales(SearxTestCase):
     - :py:obj:`searx.locales.match_locale`
     """
 
-    def test_match_locale(self):
-
-        locale_tag_list = [x[0] for x in sxng_locales]
-
+    @classmethod
+    def setUpClass(cls):
+        cls.locale_tag_list = [x[0] for x in sxng_locales]
```
glanham-jr, line 20: This is needed for every test and is static, so I set it up to the class itself

tests/unit/test_locales.py
```diff
@@ -13,98 +15,104 @@ class TestLocales(SearxTestCase):
...
-        self.assertEqual(locales.match_locale('zh-hans', locale_tag_list), 'zh-CN')
-        self.assertEqual(locales.match_locale('zh-hans-cn', locale_tag_list), 'zh-CN')
-        self.assertEqual(locales.match_locale('zh-hant', locale_tag_list), 'zh-TW')
-        self.assertEqual(locales.match_locale('zh-hant-tw', locale_tag_list), 'zh-TW')
-
-        # Test individual locale lists
-
-        self.assertEqual(locales.match_locale('es', [], fallback='fallback'), 'fallback')
-
+    def test_locale_de(self):
```
glanham-jr, line 64: The `test_local_xxx` for specific regions seems to be specific to that region, so I combined them by that name.

tests/unit/test_locales.py
```diff
@@ -13,98 +15,104 @@ class TestLocales(SearxTestCase):
...
+            # Territory tests
+            ('en', ['en-GB'], 'en-GB'),  # A user selects only a language
+            (
+                'fr',
+                ['fr-FR', 'fr-CA'],
+                'fr-FR',
+            ),  # the engine supports fr_FR and fr_CA since no territory is given, fr_FR takes priority
+        ]
+    )
+    def test_locale_optimized_selected(self, locale: str, locale_list: list[str], expected_locale: str):
```
glanham-jr, line 92: All these tests are the same, but I left the comments for context inline to each test case

tests/unit/test_plugins.py
```diff
@@ -23,143 +24,125 @@ class PluginMock:  # pylint: disable=missing-class-docstring, too-few-public-met
 
 
 class PluginStoreTest(SearxTestCase):  # pylint: disable=missing-class-docstring
-    def test_PluginStore_init(self):
-        store = plugins.PluginStore()
-        self.assertTrue(isinstance(store.plugins, list) and len(store.plugins) == 0)
+    def setUp(self):
+        self.store = plugins.PluginStore()
```
glanham-jr, line 28: The store is needed for every test, so its easier to place in the setup.

tests/unit/test_plugins.py
```diff
@@ -23,143 +24,125 @@ class PluginMock:  # pylint: disable=missing-class-docstring, too-few-public-met
...
-        self.assertFalse(testplugin.asdf.called)  # pylint: disable=E1101
-
         store.call([testplugin], 'asdf', request, Mock())
-        self.assertTrue(testplugin.asdf.called)  # pylint: disable=E1101
+        self.assertTrue(getattr(testplugin, 'asdf').called)  # pylint: disable=E1101
 
 
-class SelfIPTest(SearxTestCase):  # pylint: disable=missing-class-docstring
-    def test_PluginStore_init(self):
+class PluginIPSelfInfo(SearxTestCase):  # pylint: disable=missing-class-docstring
```
glanham-jr, line 57: Renamed this to be closer to `self_info`, the actual plugin name

tests/unit/test_plugins.py
```diff
@@ -23,143 +24,125 @@ class PluginMock:  # pylint: disable=missing-class-docstring, too-few-public-met
...
+        search = get_search_mock(query=query, pageno=2)
+        self.store.call(self.store.plugins, 'post_search', request, search)
+        self.assertNotIn('user-agent', search.result_container.answers)
 
 
-class HashPluginTest(SearxTestCase):  # pylint: disable=missing-class-docstring
-    def test_PluginStore_init(self):
-        store = plugins.PluginStore()
+class PluginHashTest(SearxTestCase):  # pylint: disable=missing-class-docstring
+    def setUp(self):
```
glanham-jr, line 110: Same here - the setup is shared for all tests, and I moved the assert to its own test.

tests/unit/test_preferences.py
```diff
@@ -10,12 +11,12 @@
...
     ValidationException,
 )
-from tests import SearxTestCase
+from searx.plugins import Plugin
 
 locales_initialize()
 
 
-class PluginStub:  # pylint: disable=missing-class-docstring, too-few-public-methods
+class PluginStub(Plugin):  # pylint: disable=missing-class-docstring, too-few-public-methods
```
glanham-jr, line 19: Inheriting the `Plugin` reduces linting noise, and doesn't break anything. Not sure why `Plugin` wasn't used directly, but this keeps the change simple for now.

tests/unit/test_preferences.py
```diff
@@ -47,22 +48,22 @@ def test_map_setting_valid_choice(self):
 
     def test_enum_setting_invalid_default_value(self):
         with self.assertRaises(ValidationException):
-            EnumStringSetting(3, choices=[0, 1, 2])
+            EnumStringSetting('3', choices=['0', '1', '2'])
```
glanham-jr, line 51: `EnumStringSetting` isn't supposed to take integers (its typed arguments enforce `str`...), and none of its usage uses integers - converted these all to strings to represent actual usage.

searx/query.py
```diff
@@ -258,7 +259,7 @@ class RawTextQuery:
         FeelingLuckyParser,  # redirect to the first link in the results list
     ]
 
-    def __init__(self, query, disabled_engines):
+    def __init__(self, query: str, disabled_engines: list):
```
glanham-jr, line 262: Had some weird typing issues with this, so I provided very basic typing for now.

tests/unit/test_settings_loader.py
allendema, line 1: While you are there, do you think it is better to directly use assertIsInstance() in L36-40?

glanham-jr: Good catch. I'll update these, and do another sweep of assertTrue and assertFalse.

glanham-jr: Fixed - didn't see any other generic asserts to update

Bnyro (approved):
Looks good to me, nice cleanup :+1:

Bnyro:
One question:
We sometimes use
`assertEqual(len(array), 0)` (e.g. https://github.com/glanham-jr/searxng/blob/gl/refactor-unit-tests/tests/unit/test_webadapter.py#L36)
and sometimes `assertEqual(array, [])`.
Are you aware of an `assertIsEmpty` function or do you think it would make sense to create one (mustn't be done in this PR though)

glanham-jr:
I was thinking of making one, but I wanted to stick to strictly refactoring in this PR. For new additions, I'll make a new PR.

return42 (approved):
> This improves the developer experience in debugging what went wrong when refactoring the project.

Yeah in deed, that is true .. thanks a lot for your work on this topic :+1: 

---

FYI: I rebased and slightly modified the commit message.

PR #3869, opened 2024-09-29
[enh] engine: stract - add language/region support

0xhtml:
## What does this PR do?

Add region support to Stract. Stract supports selecting regions in the UI as well as the API with the selectedRegion field. Supported regions can be retrieved through the openapi specification.

Region support enhances the experience.

Test for example with an ambiguous query like `!stract test` and select different languages.

### DIFF
--- a/searx/data/engine_traits.json
+++ b/searx/data/engine_traits.json
@@ -6678,6 +6678,19 @@
       "zh-TW": "zh-TW_TW"
     }
   },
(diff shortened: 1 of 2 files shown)


return42 (approved):
Wow .. perfect .. thanks a lot :+1: .. I just seperated the `engine_traits.json` in its own commit.

PR #3871, opened 2024-09-29
[enh] engine: mojeek - add language support

0xhtml:
## What does this PR do?

Add language support to the Mojeek engines through the language restriction (`lr`) parameter. The parameter is marked as beta in the API specification (https://www.mojeek.com/support/api/search/request_parameters.html). However, in my experience the feature is working fine even for less common languages that I tested with.

Language support enhances the experience.

As per #3869 use `!mojeek test` or other ambiguous queries and select different languages.


searx/engines/mojeek.py
```diff
@@ -130,3 +134,18 @@ def response(resp):
...
+    from babel import Locale, UnknownLocaleError
+    import contextlib
+
+    resp = network.get(base_url + "/preferences")
+    dom = html.fromstring(resp.text)
+
+    for code in eval_xpath_list(dom, '//select[@name="lb"]/option/@value'):
+        with contextlib.suppress(UnknownLocaleError):
+            Locale(code)
+            engine_traits.languages[code] = code
```
return42: When we iterate over `//select[@name="arc"]/option` we get the items of the region menu.  

-----

The item where `value="none"` is `engine_traits.all_locale = 'none'` in SearXNG's locale abstarction.

About `value=""`: I don't know, but I would guess mojeek handles this value like we know from startpage (query based on [HTTP Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language):

https://github.com/searxng/searxng/blob/f07ab6deb0f43a2d08f4f12335481825c6aa77ac/searx/engines/startpage.py#L52-L60

https://github.com/searxng/searxng/blob/f07ab6deb0f43a2d08f4f12335481825c6aa77ac/searx/engines/startpage.py#L121-L125

searx/engines/mojeek.py
```diff
@@ -42,6 +43,8 @@
 news_title_xpath = './/h2/a'
 news_content_xpath = './/p[@class="s"]'
 
+traits: EngineTraits
```
return42, line 58: we should consider to set property `send_accept_language_header`, see [ref](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.send_accept_language_header) & [in the engine settings](https://docs.searxng.org/admin/settings/settings_engine.html):

```python
send_accept_language_header = True
```

searx/engines/mojeek.py
```diff
@@ -130,3 +134,18 @@ def response(resp):
...
     raise ValueError(f"Invalid search type {search_type}")
+
+
+def fetch_traits(engine_traits: EngineTraits):
+    # pylint: disable=import-outside-toplevel
+    from searx import network
+    from babel import Locale, UnknownLocaleError
+    import contextlib
+
+    resp = network.get(base_url + "/preferences")
```
return42: I guess its not really needed, but I would recommend to add HTTP header `'Accept-Language': "en-US,en;q=0.5"` (see comment below).

searx/engines/mojeek.py
```diff
@@ -53,6 +56,7 @@ def request(query, params):
         'q': query,
         'safe': min(params['safesearch'], 1),
         'fmt': search_type,
+        'lr': traits.get_language(params['searxng_locale'], ''),
```
return42: Not documented in the API but used in the https://www.mojeek.com/preferences is the parameter `arc` for the region .. see my comments in fetch_traits function about `//select[@name="arc"]/option` from the preferences.

return42 (commented):
Thanks for adding language support to mojeek :+1: .. may we should also have a look at the region (`arc`) parameter .. see my comments below / what do you think?

----

> Add language support to the Mojeek engines through the language restriction (`lr`) parameter.  The parameter is marked as beta in the API specification 

:thinking: we do not use the official API, we use the HTML web page and in the https://www.mojeek.com/preferences page (on the bottom) the URL for something like de-DE is:

     https://www.mojeek.com/?lb=de&arc=de

0xhtml:
Even though the `lr` parameter is only documented for the official API, which we don't use, it seems like its working fine to me. Though using the `lb` parameter yields the same results in my testing. I will therefore switch to that parameter.

I'm also trying to implement the region `arc` parameter, which to me seems equivalent to the `rb` parameter of the official API. I am however uncertain as to how to select which regions to support. Searxng only supports regions through regional languages while Mojeek supports every region freely combined with every selectable language. Simply using the official languages of the regions and filtering by languages supported by Mojeek yields very niche regional languages like: Swedish (Åland Islands), English (American Samoa), German (Liechtenstein), French (Luxembourg) or German (Belgium) and many more. I'm uncertain how to proceed.

return42:
> Even though the `lr` parameter is only documented for the official API, which we don't use, it seems like its working fine to me. Though using the `lb` parameter yields the same results in my testing. I will therefore switch to that parameter.

:+1: 

> I'm also trying to implement the region `arc` parameter, which to me seems equivalent to the `rb` parameter of the official API. I am however uncertain as to how to select which regions to support.

IMO we should use the parameters shown in the right bottom of the preferences (see "*Cookieless: copy the Preferences URL ...*")

grafik

> Searxng only supports regions through regional languages while Mojeek supports every region freely combined with every selectable language. Simply using the official languages of the regions and filtering by languages supported by Mojeek yields very niche regional languages like: Swedish (Åland Islands), English (American Samoa), German (Liechtenstein), French (Luxembourg) or German (Belgium) and many more. I'm uncertain how to proceed.

Haven't looked into the details, but if I'm not wrong, we can reduce them to the *"official languages in a region"* .. I have done similar in the bing engine ...
 
https://github.com/searxng/searxng/blob/f07ab6deb0f43a2d08f4f12335481825c6aa77ac/searx/engines/bing.py#L249-L255

If it is not clear (or I'm wrong) I can implement the regions in my final review.

0xhtml:
Welp, the code you suggest is pretty much the same I tried:

```python
from searx.locales import get_official_locales, region_tag
for code in eval_xpath_list(dom, '//select[@name="arc"]/option/@value'):
    for locale in get_official_locales(code):
        if locale.language in engine_traits.languages:
            print('region:', code, locale.english_name)
            engine_traits.regions[region_tag(locale)] = code
```

And this does produce the likely undesirable regional languages I mentioned.

0xhtml:
Is there a way to somehow say that Mojeeks regional languages should not extend the languages available for selection in searxng? We could store the regions as extra data in the engine traits, not as regional languages but rather as regions/territories the way Mojeek understands them.

searxng_extra/update/update_engine_traits.py
```diff
@@ -101,7 +101,7 @@ def log(msg):
 def filter_locales(traits_map: EngineTraitsMap):
     """Filter language & region tags by a threshold."""
 
-    min_eng_per_region = 15
+    min_eng_per_region = 18
```
return42, line 104: > Is there a way to somehow say that Mojeeks regional languages should not extend the languages available for selection in searxng?

Not directly to an engine,  but the default catalog of locales is generated by thresholds and when we increase the amount of engines with a language/region support we have to consider to increase the thresholds.

searx/engines/mojeek.py
```diff
@@ -48,6 +50,13 @@
 language_param = 'lb'
 region_param = 'arc'
 
+_delta_kwargs = {'day': 'days', 'week': 'weeks', 'month': 'months', 'year': 'years'}
```
return42, line 51: Not really needed but I thought for better readability it could help.

return42 (commented):
> We could store the regions as extra data in the engine traits, not as regional languages but rather as regions/territories the way Mojeek understands them.

Nope, we should not / the regions of an engine should be integrated into the locale concept that exists in SearXNG .. 

- https://github.com/searxng/searxng/pull/2269

----

I added a WIP commit (mainly to fix some type hint reports) and some comments below .. If you fine with the modifications from my WIP commit, we can squash this commit into your initial commit and merge the PR.

searx/engines/mojeek.py
```diff
@@ -150,15 +160,13 @@ def fetch_traits(engine_traits: EngineTraits):
...
         with contextlib.suppress(UnknownLocaleError):
             Locale(code)
             engine_traits.languages[code] = code
 
     for code in eval_xpath_list(dom, f'//select[@name="{region_param}"]/option/@value'):
-        for locale in get_official_locales(code):
-            if locale.language in engine_traits.languages:
-                print('region:', code, locale.english_name)
-                engine_traits.regions[region_tag(locale)] = code
+        for locale in get_official_locales(code, engine_traits.languages, de_facto=False):
```
0xhtml: Setting `de_facto=False` excludes regions such as `en-US`. So, I'm not sure if that is desired.

searxng_extra/update/update_engine_traits.py
0xhtml: That seems like a reasonable way to handle the uncommon language regions.

searx/engines/mojeek.py
```diff
@@ -159,14 +159,23 @@ def fetch_traits(engine_traits: EngineTraits):
...
         with contextlib.suppress(UnknownLocaleError):
-            Locale(code)
-            engine_traits.languages[code] = code
+            locale = Locale(code)
+            engine_traits.languages[locale.language] = code
+
+    regions = eval_xpath_list(dom, f'//select[@name="{region_param}"]/option/@value')
+
+    engine_traits.custom['region_default'] = regions[0]
+    engine_traits.custom['region_all'] = regions[1]
```
0xhtml, line 174: Can't use `traits.all_locale` here 'cause that would also be used by `traits.get_language`, which is undesirable as that value is only supported for regions.

searx/engines/mojeek.py
0xhtml: I'm no maintainer, therefore my words may not matter, but my workflow involves looking at the git blame quite regularly. Therefore, I would vouch for this to be a separate commit to the other changes. Otherwise, I agree w/ the improved readability.

searx/engines/mojeek.py
0xhtml: I also would assume the empty values mean that Mojeek detects the language based on accept-language and the region based on IP address.

0xhtml:
From my testing it seems that Mojeek uses the IP for region detection regardless of accept language headers if region is set to empty. Mojeek also seems to not prefer any language or detect any language when setting language to empty. Therefore, the `send_accept_language_header = True` likely doesn't matter, and we should use the `none` option for region if the searx user hasn't selected a region. Never using the region detection as that would result in regionalization to the searx server location.

searx/engines/mojeek.py
return42: > excludes regions such as `en-US`.

Oops .. thanks for pointing out, I fixed the issue --> https://github.com/searxng/searxng/compare/cc7dfbd2b0b1204687611dedc457caadf879408a..c98e83153c9e03b72bad9cd92404d97c201d8c3a

searx/engines/mojeek.py
return42: I squashed the 7 commits into two:

1. squashed patches of the code: 0262d56d0
2. squashed patches generated by `make data.traits`: cc17f80775

return42 (approved):
@0xhtml I fully agree with the results of your reverse engineering .. thanks for taking the time :+1: 

I just fixed the `de_facto=False (e.g. en-US)` issue, squashed the commits / see https://github.com/searxng/searxng/pull/3871#discussion_r1799158315

Further I had to rebase on current `master` branch to solve a conflict in the `tests/unit/test_locales.py`.

From my side this PR is ready to merge / if you give me your final *"GO"* I can merge the PR.

0xhtml:
Looks good to merge to me! :+1:

PR #3876, opened 2024-10-02
[mod] py3.8 EOL, upgrade to actions/setup-python@v5

return42:
## What does this PR do?

- drop py 3.8 support
- add py 3.13 to CI
- lift py3.9 based builds to builds based on py3.12
- upgrade CI to actions/setup-python@v5

Related:

- https://github.com/searxng/searxng/pull/3663

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version:  ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version:  ["3.9", "3.10", "3.11", "3.12",]
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


Bnyro:
Apparently `uvloop` is not yet supporting Python 3.13.0?

I agree with dropping Python 3.8 though, it's been discontinued May 3, 2021, 3 years ago ...

return42:
> Apparently `uvloop` is not yet supporting Python 3.13.0?

At this time 3.13 is a RC [1] / I expect uvloop support after 3.13 has final released.

I just added 3.13rc here .. just to see what will happen in the CI .. I will remove 3.13 from this PR

> I agree with dropping Python 3.8 though, it's been discontinued May 3, 2021, 3 years ago ...

.. sure, but the EOL has been reached this month [1] .. since yesterday :)

[1] https://devguide.python.org/versions/#supported-versions

return42:
I'm going to merge this PR .. it helps to solve issues with other PRs we have in the loop ..

PR #3877, opened 2024-10-02
[feat] engine: support for openlibrary

Bnyro:
## What does this PR do?
- add support for https://openlibrary.org

- open library provides a full list of (almost) all books ever published

- !ol dan brown

## Notes
- The engine is by the Internet Archive guys, so responses can be very slow
- ~This doesn't use the documented `search.json` API, it uses `search/inside.json` due to some additional features~

### DIFF
--- a/searx/engines/openlibrary.py
+++ b/searx/engines/openlibrary.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Open library (books)
+"""
+from urllib.parse import urlencode
+import re
+
+from dateutil import parser
+
+about = {
+    'website': 'https://openlibrary.org',
+    'wikidata_id': 'Q1201876',
+    'require_api_key': False,
+    'use_official_api': False,
+    'official_api_documentation': 'https://openlibrary.org/developers/api',
+}
+
+paging = True
+categories = []
+
+base_url = "https://openlibrary.org"
+results_per_page = 10
+
+
+def request(query, params):
+    args = {
+        'q': query,
+        'page': params['pageno'],
+        'limit': results_per_page,
+    }
+    params['url'] = f"{base_url}/search.json?{urlencode(args)}"
+    return params
+
+
+def _parse_date(date):
+    try:
+        return parser.parse(date)
+    except parser.ParserError:
+        return None
+
+
+def response(resp):
+    results = []
+
+    for item in resp.json().get("docs", []):
+        cover = None
+        if 'lending_identifier_s' in item:
+            cover = f"https://archive.org/services/img/{item['lending_identifier_s']}"
+
+        published = item.get('publish_date')
+        if published:
+            published_dates = [date for date in map(_parse_date, published) if date]
+            if published_dates:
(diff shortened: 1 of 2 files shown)


return42:
> * This doesn't use the documented `search.json` API, it uses `search/inside.json` due to some additional features

Is there a documentation available .. I miss a language pref by example.

----

See below, I forced pushed a fix / here is the [diff](https://github.com/searxng/searxng/compare/aa313b20943a0d2aeaa65212e367d7f2105e018b..26054e3702b8b703cc5c5a26a17ee7b044f36985)

Bnyro:
> Is there a documentation available .. I miss a language pref by example.
>
There's https://openlibrary.org/dev/docs/api/search, but the language pref doesn't seem to work from my tests.

It turned out the search inside API only searches within the content of books which gives a lot of unexpected results (no clue how I missed that), so I rewrote the module to use the normal API.

return42:
> so I rewrote the module to use the normal API.

:+1: 

> There's https://openlibrary.org/dev/docs/api/search, but the language pref doesn't seem to work from my tests.

Sadly this was also my experience .. I also tried `sherlock holmes language:fre` ..

image

but without luck: 

- In the WEB it works: https://openlibrary.org/search?q=sherlock+holmes+language%3Afre&mode=everything
- but not as JSON: https://openlibrary.org/search.json?q=sherlock+holmes+language%3Afre&mode=everything&limit=10

searx/utils.py
```diff
@@ -754,3 +754,17 @@ def js_variable_to_python(js_variable):
...
+    Get the first n items of a list.
+
+    :param input_list: The list to select the first n items from.
+    :param limit: The maximum amount of items to return.
+
+    :returns The first n items of the list.
+    """
+    if len(input_list) > limit:
+        return input_list[:limit]
+    return input_list
```
return42: Do we really need this function? Isn't it the same when using `[:limit]`:

```python
>>> [1,2,3,4,5][:4]
[1, 2, 3, 4]
>>> [1,2,3,4,5][:12]
[1, 2, 3, 4, 5]
```

searx/engines/openlibrary.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if not published:
+            published = parser.parse(str(item.get('first_published_year')))
+
+        result = {
+            'template': 'paper.html',
+            'url': f"{base_url}{item['key']}",
+            'title': item['title'],
+            'content': re.sub(r"\{|\}", "", item['first_sentence'][0]) if item.get('first_sentence') else '',
+            'isbn': take_n(item.get('isbn', []), 5),
```
return42: Alternatively we can use:

```python
item.get('isbn', [])[:5]
```

return42 (changes requested):
See my comments about `take_n(..)`

searx/utils.py
Bnyro: Indeed, I thought there would be an IndexOutOfBoundsError if the list is too short.

return42 (approved):

Issue #3854, opened 2024-09-24
Calculator Plugin Errors When Using Decimal Values In Search Bar


fostro:
Container Image:  docker.io/searxng/searxng:latest

When the calculator plugin is enabled, trying to pass in any equation that contains decimal values (floating point math) throws the following error

```
searxng        | 2024-09-24T20:20:16.381773929Z Process Process-6:                                                                                                                                                                           
searxng        | 2024-09-24T20:20:16.383401418Z Traceback (most recent call last):                                                                                                                                                           
searxng        | 2024-09-24T20:20:16.384448893Z   File "/usr/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap                                                                                                             
searxng        | 2024-09-24T20:20:16.384466676Z     self.run()                                                                                                                                                                               
searxng        | 2024-09-24T20:20:16.384507243Z   File "/usr/lib/python3.12/multiprocessing/process.py", line 108, in run                                                                                                                    
searxng        | 2024-09-24T20:20:16.384518744Z     self._target(*self._args, **self._kwargs)                                                                                                                                                
searxng        | 2024-09-24T20:20:16.384640545Z   File "/usr/local/searxng/searx/plugins/calculator.py", line 62, in handler                                                                                                                 
searxng        | 2024-09-24T20:20:16.384652938Z     q.put(func(*args, **kwargs))                                                                                                                                                             
searxng        | 2024-09-24T20:20:16.384658509Z           ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                              
searxng        | 2024-09-24T20:20:16.385008178Z   File "/usr/local/searxng/searx/plugins/calculator.py", line 42, in _eval_expr                                                                                                              
searxng        | 2024-09-24T20:20:16.385034477Z     return _eval(ast.parse(expr, mode='eval').body)                                                                                                                                          
searxng        | 2024-09-24T20:20:16.385041711Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                          
searxng        | 2024-09-24T20:20:16.385439661Z   File "/usr/local/searxng/searx/plugins/calculator.py", line 50, in _eval                                                                                                                   
searxng        | 2024-09-24T20:20:16.385458727Z     return operators[type(node.op)](_eval(node.left), _eval(node.right))                                                                                                                     
[...]

PR #3885, opened 2024-10-05
[fix] fix float operations in calculator plugin

glanham-jr:
## What does this PR do?

This PR adds an additional isinstance check within the ast parser to check for float along with int, fixing the underlying issue.

This commit also includes some additional refactoring in unit tests. As we test more plugins, it seems unwieldy to include every test class in the test_plugins.py file. I split apart all of the test plugins to their own respective files, including the new test_plugin_calculator.py file.

**NOTE**

This PR does NOT intend to address localization issues, and was not explicitly tested.

Fixes an issue that decimal numbers don't work for the calculator.

I was able to recreate the issue entirely through unit tests, so all that is needed to assert this works is to run the unit tests.

 - `make test.unit`

## Author's checklist

 - n/a

## Related issues

- #3854

### DIFF
--- a/tests/unit/test_plugin_calculator.py
+++ b/tests/unit/test_plugin_calculator.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
+
+from mock import Mock
+from parameterized.parameterized import parameterized
+from searx import plugins
+from tests import SearxTestCase
+from .test_utils import random_string
+
+from .test_plugins import get_search_mock
+
+
+class PluginCalculator(SearxTestCase):  # pylint: disable=missing-class-docstring
+    def setUp(self):
+        self.store = plugins.PluginStore()
+        plugin = plugins.load_and_initialize_plugin('searx.plugins.calculator', False, (None, {}))
+        self.store.register(plugin)
+
+    def test_plugin_store_init(self):
+        self.assertEqual(1, len(self.store.plugins))
+
+    def test_single_page_number_true(self):
+        request = Mock(remote_addr='127.0.0.1')
+        search = get_search_mock(query=random_string(10), pageno=2)
+        result = self.store.call(self.store.plugins, 'post_search', request, search)
+        self.assertTrue(result)
+        self.assertNotIn('calculate', search.result_container.answers)
+
+    def test_long_query_true(self):
+        request = Mock(remote_addr='127.0.0.1')
+        search = get_search_mock(query=random_string(101), pageno=1)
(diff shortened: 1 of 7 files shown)


return42:
> This commit also includes some additional refactoring in unit tests. 

@glanham-jr I did split the single commit into two commits:

- `[refactor] unit tests (continued) - plugins` and ..
- `[fix] float operations in calculator plugin`

> This PR does NOT intend to address localization issues, and was not explicitly tested.

I'm working on a solution ..

glanham-jr:
> I did split the single commit into two commits

I'm used to working with squash commits - good to know moving forward.

return42:
I placed a WIP commit on top .. it solve the lacale issue .. but sadly, since non of the tests has an `search.search_query.locale` defined most of the tests will fail (not only these from the calculator).

glanham-jr:
@return42 I pushed a commit to fix the unit tests + catch the DivideByZero exception to reduce logging. I also added a check that if a locale isn't present, it will return early. Feel free to modify the changes, but its in a close working state now :smile: 

EDIT: Thanks for the help and the teamwork!

return42:
@glanham-jr I force pushed last week / see [diff](https://github.com/searxng/searxng/compare/2f02a0f57fe978cc17307ec88378f142d63c9f51..f3541cd4dc8e15783cf9685fbdbdac1d8b0a1bf1) .. manly to dense the test functions into one (with a huge parameter block in front :-) 

Further I added a `search_query.locale` to pass the tests:

grafik

grafik

Can you please have a final look / thanks :+1:

tests/unit/test_plugin_calculator.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @parameterized.expand(
+        [
+            ("1+1", "2"),
+            ("1-1", "0"),
+            ("1*1", "1"),
+            ("1/1", "1"),
+            ("1**1", "1"),
+            ("1^1", "1"),
+            # en-US
```
glanham-jr: nit: although it's more verbose, we could just remove the default parameter assignment and have en-US for each rest. Then we have no need for comments. But this is non blocking and we can merge anyway.

glanham-jr (commented):
@return42 looks good to me, nothing blocking from merging. The calculator is in a much better state now!

tests/unit/test_plugin_calculator.py
return42: Good point :+1:  .. I modified the test sources [diff](https://github.com/searxng/searxng/compare/f3541cd4dc8e15783cf9685fbdbdac1d8b0a1bf1..2c20f41e6077335bc9f0dda70c2bada16fe15fb5)

PR #3906, opened 2024-10-09
[feat] search: support for filtering image licenses

### LINKED ISSUE #3903: Add Public Domain \ CC0 Image Search
**Working URL to the engine**
https://duckduckgo.com/

This engine searches for images with the legal status of CC0 Public Domain. It is based on Microsoft Edge and is not free or open source.

**Features of this engine**
This engine searches for stock images that have been released into the public domain by their authors. These images are free and can be legally used in any multimedia project without attribution or compensation to the authors. For some reason, SearXNG does not have a search for CC0 images. I tried adding the words "Public Domain" or "CC0" to the search bar, but SearXNG still searched for images everywhere, including commercial sites with copyrights, where you have to be very careful about choosing stocks for projects.

**How ​​can SearXNG fetch the information from this engine?**

I would like to see the image usage license search feature available not only for image searches, but also for audio, video, and 3D.

**Additional context**
I am not comfortable using DuckDuckGo to search for stock images, so I would like to see usage license search in SearXNG.

Bnyro:
## What does this PR do?
- this PR adds support for filtering for copyleft image licenses to use in your own project
- supported engines: duckduckgo, bing, google images

- see #3903

- search for an image and select one of the license filters

## Author's checklist
- there's some strange bug in SearXNG that if you select a filter in the images page, the search is repeated with the new filter in the general category and not the images category
- ddg images ratelimited me so not sure if this the duckduckgo changes work

## Related issues
- closes #3903


searxng_extra/standalone_searx.py
```diff
@@ -160,6 +162,9 @@ def parse_argument(
     parser.add_argument(
         '--timerange', type=str, nargs='?', choices=['day', 'week', 'month', 'year'], help='Filter by time range'
     )
+    parser.add_argument(
+        '--license', type=str, nargs='?', choices=['any', 'public', 'freetouse', 'commercial'], help='Filter by license'
```
return42: I think we should define what the choices mean, its not clear to me e.g what "any" stands for .. does it mean *"don't filter by license"* .. engine developers should have a rough idea of what we want to filter with these categories: 'public', 'freetouse', 'commercial'

Bnyro: I agree, but we probably should document it somewhere else than in `standalone_searx.py`, right? Possibly at the [Passed arguments documentation](https://docs.searxng.org/dev/engines/engine_overview.html#passed-arguments-request)?

return42: > Possibly at the [Passed arguments documentation](https://docs.searxng.org/dev/engines/engine_overview.html#passed-arguments-request)?

Seems a quite good place to document the meaning of the *license categories*

- [searx/templates/simple/filters/license.html](https://github.com/searxng/searxng/pull/3906/files#diff-e258700dc86b956c917ed2fd84b2188bd21fa2e4742ce0f0e4158d1227867045)

.. since it is more a "category of licenses to filter" what the user will select, we should name the field `filter_lic-categ` or similar.  The lic-filter named *all* should not exists, the user selects a license *category* `filter_lic-categ=[public|freetouse|commercial]` or `None`.

return42:
@Bnyro I completely lost sight of this PR (sorry) ... can you please do a rebase / thanks!

Bnyro:
I decided that it's not really feasible/worth it having a license filter in SearXNG because most engines simply don't support it or have very different licenses, and it takes a lot of efforts per engine to implement this.

PR #3925, opened 2024-10-15
[refactor] translation engines: common interface

### LINKED ISSUE #3576: Common interface for translation engines
All translation engines currently do their own thing when it comes to showing the results to the user.

The results however should be
- placed at the same location (currently some engines use the infobox, some use answerers, and some use normal results)
- presented with a similar interface to make it easy for users to use multiple translation engines at a time
to the user.

Affected engines (6 as of now):
https://github.com/search?q=repo%3Asearxng%2Fsearxng+%27translate%27&type=code

Bnyro:
## What does this PR do?
- add a common answer template for translation engine results

- see https://github.com/searxng/searxng/issues/3576

- use mozhi, lingva, LibreTranslate, Dictzone, MyMemory Translated or DeepL

## Author's checklist
- this is just a basic idea, more engines need to be converted if the approach seems okay
- something similar should also be done to the weather engines

## Related issues
closes https://github.com/searxng/searxng/issues/3576

### DIFF
--- a/searx/engines/deepl.py
+++ b/searx/engines/deepl.py
@@ -1,8 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Deepl translation engine"""
 
-from json import loads
-
 about = {
     "website": 'https://deepl.com',
     "wikidata_id": 'Q43968444',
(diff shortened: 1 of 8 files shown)


searx/utils.py
```diff
@@ -800,3 +803,12 @@ def js_variable_to_python(js_variable):
     s = s.replace(chr(1), ':')
     # load the JSON and return the result
     return json.loads(s)
+
+
+def render_template(template_path, **args):
+    templates_path = settings['ui']['templates_path']
+    with open(os.path.join(templates_path, "simple/" + template_path)) as f:
```
Bnyro: The `simple` theme is just hardcoded here, but we should use the theme from the request cookies - not sure how to get that inside a search engine module though.

Bnyro:
As a side effect, this PR also improves some of the translation engines (e.g. LibreTranslate and Lingva) a lot. I should probably split these into dedicated engine module refactor commits in the future.

return42:
FYI: last force push was just a rebase on current master / before I go into the review.

return42:
FYI: I included the commit from this PR in https://github.com/searxng/searxng/pull/4183

return42:
closed by: https://github.com/searxng/searxng/pull/4183

PR #3932, opened 2024-10-17
[enh] use longest title and fix get_ordered_results()

allendema:
## What does this PR do?

Make sure that when multiple engines get the same result, the longest title will be used.
This was already done for 'description'.

Some engines might return empty titles (mojeek). 
Not have the exact query at hand right now.

`make test`
`make run`
## Author's checklist

Without adding the initialize() there are a bunch of errors when using get_ordered_results().
Feel free to change anything.
## Related issues

N/A

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -12,7 +12,6 @@ from searx import logger
 from searx.engines import engines
 from searx.metrics import histogram_observe, counter_add, count_error
 
-
 CONTENT_LEN_IGNORED_CHARS_REGEX = re.compile(r'[,;:!?\./\\\\ ()-_]', re.M | re.U)
 WHITESPACE_REGEX = re.compile('( |\t|\n)+', re.M | re.U)
 
@@ -133,7 +132,7 @@ def result_score(result, priority):
     weight = 1.0
 
     for result_engine in result['engines']:
-        if hasattr(engines[result_engine], 'weight'):
+        if hasattr(engines.get(result_engine), 'weight'):
             weight *= float(engines[result_engine].weight)
 
     weight *= len(result['positions'])
@@ -332,9 +331,13 @@ class ResultContainer:
         return None
 
     def __merge_duplicated_http_result(self, duplicated, result, position):
-        # using content with more text
+        # use content with more text
         if result_content_len(result.get('content', '')) > result_content_len(duplicated.get('content', '')):
             duplicated['content'] = result['content']
+
+        # use title with more text
+        if result_content_len(result.get('title', '')) > len(duplicated.get('title', '')):
+            duplicated['title'] = result['title']
 
         # merge all result's parameters not found in duplicate
         for key in result.keys():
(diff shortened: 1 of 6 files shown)


searx/results.py
```diff
@@ -184,6 +183,7 @@ class ResultContainer:
 
     def __init__(self):
         super().__init__()
+        metrics_initialize()
```
return42: Why is the init needed here .. as far I can say metrics should be initialized by the `initialize_metrics` call here in the `searx.search` package

https://github.com/searxng/searxng/blob/cf7627557ab29e31efbf4a0719d1083592e22242/searx/search/__init__.py#L36

allendema: get_ordered_results() -> close() -> counter_add() -> counter_storage not defined (or similiar) would somehow fail.

 So I just made sure that counter_storage was set by initializing it it explicitly. I'm sure there is another way to solve it.

return42: > get_ordered_results() -> close() -> counter_add() -> counter_storage not defined (or similiar) would somehow fail.

Sure? .. I can't reproduce it .. I commented out `metrics_initialize` and made some tests `make test run ..` without an issue.

allendema: Something is flaky.
Here it's reproducible and this is one example traceback.

```ERROR: test_result_merge_by_title (tests.unit.test_results.ResultContainerTestCase.test_result_merge_by_title)
  File "/workspaces/searxng/tests/unit/test_results.py", line 72, in test_result_merge_by_title
    self.container.extend('stract', [fake_result(engine='stract', title='short title')])
  File "/workspaces/searxng/searx/results.py", line 243, in extend
    histogram_observe(standard_result_count, 'engine', engine_name, 'result', 'count')
  File "/workspaces/searxng/searx/metrics/__init__.py", line 49, in histogram_observe
    histogram_storage.get(*args).observe(duration)
    ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

return42: This may be due to the unit tests ... they are not always executed in the same order ... for me metrics seems to be already initialized when the test_results.py is executed ... for you the order is different and then the initialization does not seem to have taken place yet ... our tests do not build and unbuild their environments cleanly ... I will have to take a closer look at the problem ...

return42: @allendema I removed the `metrics_initialize()`  from your commit and I placed a commit on top of your branch to fix the issues you have seen in the unit tests, further I rebased you branch on current master.

Can you please give it another try .. thanks :+1:

return42: Argh .. unit tests fails in the CI .. seems I have to do more work / comming soon ..

return42:
FYI: last force push was just a rebase on branch master.

return42 (approved):
Longer title is shown in the result & CI has now been passed ;-) .. Thanks for this PR :+1:

allendema:
That's nice.

PR #3936, opened 2024-10-17
[fix] ci: docker builds broken on armv7l (hack)

rhee876527:
Add temporary fix to broken docker builds in anticipation of yet to be released pydantic version v2.24.1

## What does this PR do?

Temporary fix for broken docker builds 

Closes #3887

PR #3938, opened 2024-10-17
[refactor] replace pydantic by msgspec

return42:
## What does this PR do?

replace pydantic by msgspec and revert:

- https://github.com/searxng/searxng/pull/3936
- https://github.com/searxng/searxng/pull/3937

After we had the problems with pydantic in the docker build, we took a closer look at pydantic and are no longer convinced that this is the right solution for SearXNG ... @dalf suggested that we could also use msgspec for our purposes.
The core of msgspec is written in C and msgspec is generally considered to be more performant than pydantic, which also has a rust implementation that makes compiling the wheels (which are often required on platforms like armv7) easier.

Activate the favicons:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index b37134b88..b120378b6 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -37,7 +37,7 @@ search:
   autocomplete_min: 4
   # backend for the favicon near URL in search results.
   # Available resolvers: "allesedv", "duckduckgo", "google", "yandex" - leave blank to turn it off by default.
-  favicon_resolver: ""
+  favicon_resolver: "duckduckgo"
   # Default search language - leave blank to detect from browser information or
   # use codes from 'languages.py'
   default_lang: "auto"
```

Start `make run` and in the result of a query you should see the favicons (and no ERROR in the log on stdout).

Stop instance by `CTRL-C`.

To test type checking of msgspec change a config setting to an invalid type / here an example for "str" where "int" is expected:

```diff
diff --git a/searx/favicons/favicons.toml b/searx/favicons/favicons.toml
index 0e433d3aa..5f664e72d 100644
--- a/searx/favicons/favicons.toml
+++ b/searx/favicons/favicons.toml
@@ -1,6 +1,6 @@
 [favicons]
 
-cfg_schema = 1   # config's schema version no.
+cfg_schema = "1"   # config's schema version no.
 
 [favicons.proxy]
```

Again try to start the instance `make run` / it should end up in an error:

```
msgspec.ValidationError: Expected `int`, got `str` - at `$.favicons.cfg_schema`
make: *** [Makefile:28: run] Error 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,8 +43,7 @@ RUN apk add --no-cache -t build-dependencies \
     tini \
     uwsgi \
     uwsgi-python3 \
-    brotli
-
+    brotli \
 # For 32bit arm architecture install pydantic from the alpine repos instead of requirements.txt
 ARG TARGETARCH
 RUN if [ "$TARGETARCH" = "arm" ]; then \
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,14 +44,8 @@ RUN apk add --no-cache -t build-dependencies \
     uwsgi \
     uwsgi-python3 \
     brotli \
-# For 32bit arm architecture install pydantic from the alpine repos instead of requirements.txt
-ARG TARGETARCH
-RUN if [ "$TARGETARCH" = "arm" ]; then \
-        apk add --no-cache py3-pydantic && pip install --no-cache --break-system-packages -r <(grep -v '^pydantic' requirements.txt); \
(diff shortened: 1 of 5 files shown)


Bnyro (approved):
Tested as described by @return42, it works!

I've no personal opinion on msgspec vs pydantic, but since msgspec would likely fix the armv7l build, that's a "killer" argument for using msgspec.

PR #3945, opened 2024-10-18
Reduce new favicon size in search results

omentic:
## What does this PR do?

This PR is for minor stylistic changes to the theming of the new favicons system.

Before: 

After:

This change is not important. I am making these changes to my local instance, and thought they looked a lot better, so figured I might as well send them upstream if desired. Feel free to do whatever with it!


searx/static/themes/simple/src/less/search.less
```diff
@@ -380,10 +380,8 @@ html.no-js #clear_search.hide_if_nojs {
 }
 
 .favicon img {
-  height: 1.8rem;
-  width: 1.8rem;
+  height: 1.25rem;
+  width: 1.25rem;
```
return42: I like both, but ``1.25rem`` fits the line height much better

searx/static/themes/simple/src/less/search.less
```diff
@@ -380,10 +380,8 @@ html.no-js #clear_search.hide_if_nojs {
...
   border-radius: 20%;
```
return42: I would suggest to reduce the border-radius to a value of 10%.

searx/static/themes/simple/src/less/search.less
```diff
@@ -380,10 +380,8 @@ html.no-js #clear_search.hide_if_nojs {
...
-  background-color: var(--color-favicon-background-color);
-  border: 1px solid var(--color-favicon-border-color);
```
return42: Not sure its good to remove background color: most of the transparent favicons are made for a light (white) background .. e.g. a black symbol with a transparent background is ok on a light theme but not on a dark theme where the base color in the background is dark (what results in a black symbol on a dark background).

return42 (approved):
I slightly modified the commit message and placed `make static.build.commit` on top.

It's hard for me to decide whether we should make these changes ... I've also added a few comments below.

Lets hear what others say .. ping: @vojkovic  @mrpaulblack @Bnyro

return42:
Here is an example for the `` background white vs transparent: 

before:

grafik

after:

grafik

searx/static/themes/simple/src/less/search.less
vojkovic: The purpose of the favicon is to quickly be able to quickly see what you're clicking on. At 1.25rem it becomes much harder to make out the favicon, and even more so on mobile devices.

An issue is that SearXNG doesn't have website 'short-names'. For example, make a search on Google and you'll see how the short name is above the link which balances out the design.

Obviously, I'm no expert on UX/UI design but I think usability should be prioritised. I would be in favour of maybe lowering the size of the favicons just a little if people want to do that but this is just too much in my opinion.

searx/static/themes/simple/src/less/search.less
vojkovic: My thoughts mirror Markus' here - https://github.com/searxng/searxng/pull/3727#discussion_r1718513261

searx/static/themes/simple/src/less/search.less
perennialtech: Just popping in with my thoughts.

At `1.8rem`, the favicons are far too large and distract from reading the names for the search results; they're meant to be auxiliary content, not the main focus.

On the other hand, `1.25rem` *is* a bit small, and could use a slight bump (maybe `1.5rem`?)

return42: > An issue is that SearXNG doesn't have website 'short-names'. For example, make a search on Google and you'll see how the short name is above the link which balances out the design.

I don't think a "short-name" is needed in SearXNG / most origin search sites (e.g. ddg) do not have a short name we can fetch in our engines.

Here is how it looks on google (incl. "short-name") ..

grafik

and here a screenshot from ddg:

grafik

Bnyro (approved):
I personally agree with reducing the favicon size since it feels very big, it's even larger than the height of the url.

However, as Markus said, I think we should keep the background.

searx/static/themes/simple/src/less/search.less
omentic: Ah, yeah, I see... seems like there isn't any one perfect solution. I prefer no background slightly more personally but wrt. defaults a white background makes sense.

searx/static/themes/simple/src/less/search.less
omentic: Current status (1.25rem width/height, 10% rounding):

omentic: Alternatively, there is this (1.5rem width/height, 10% rounding):


Which one do you all prefer?

return42: > The purpose of the favicon is to quickly be able to quickly see what you're clicking on. At 1.25rem it becomes much harder to make out the favicon, and even more so on mobile devices.

> At `1.8rem`, the favicons are far too large  .. On the other hand, `1.25rem` *is* a bit small, and could use a slight bump (maybe `1.5rem`?)

.. both arguments count, its hard for me to decide for one or the other .. (may we see some day a PR that places the favicon left to the title?) .. 

> Which one do you all prefer?

I'm not an UI expert to judge, but since I have to make a decision I will choose the compromise, which seems to be 1.5rem .. hope thats OK for all of you.

Issue #3927, opened 2024-10-16
duckduckgo blocked by CAPTCHA


return42:
Not often, but at some point ddg might block the SearXNG requests by a CAPTCHA,

The CAPTCHA goes away when opening duckduckgo.com in a browser / from the SearXNG IP .. as described here: [Answer CAPTCHA from server’s IP](https://docs.searxng.org/admin/answer-captcha.html#answer-captcha-from-server-s-ip)

I have created this ticket with the intention that we (I) re-engineer the ddg engine ... especially the determination of the [vqd value](https://docs.searxng.org/dev/engines/online/duckduckgo.html#searx.engines.duckduckgo.get_vqd) should be paid attention to again, this should actually avoid CAPTCHAs

----

May we consider to switch to https://html.duckduckgo.com/html/ the URL where ddg routes your browser, when JS is disabled.

https://github.com/searxng/searxng/blob/058a072404a4e9ec4c5a0a957fe06ce66e652938/searx/engines/duckduckgo.py#L56-L56

MansurEsm:
Hi @return42,

I'm still experiencing DuckDuckGo CAPTCHA issues despite running **searxng/searxng:2025.9.1-b93cc2f** which should contain the fix from PR #3955. It's happening occasionally.

The error from log is:
ERROR searx.engines.duckduckgo: CAPTCHA searx.exceptions.SearxEngineCaptchaException: CAPTCHA (de-de), suspended_time=0 DEBUG searx.network.duckduckgo: HTTP Request: POST [https://html.duckduckgo.com/html/](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) "HTTP/2 202 Accepted"

I run it as a pod in-cluster (one replica).

My config:
'''
search:
   [...]
   formats:
        - html
        - json
 [....]
 - name: duckduckgo
        engine: duckduckgo
        shortcut: ddg
        disabled: false
        weight: 0.8
        categories: [general]
        timeout: 5.0
'''

If you have further questions to assist you in solving, then let me know.

return42:
@MansurEsm this issue is obsolete (from oct. 2024) .. please follow https://github.com/searxng/searxng/issues/4824

odellus:
This needs revisiting. It's an arms race.

return42:
I lock this already closed issue from 2024, please follow: https://github.com/searxng/searxng/issues/6596

PR #3947, opened 2024-10-19
[fix] engine: duckduckgo - CAPTCHA detection

return42:
## What does this PR do?

Implements CAPTCHA detection in the duckduckgo engine

The previous implementation could not distinguish a CAPTCHA response from an ordinary result list.  In the previous implementation a CAPTCHA was taken as a result list where no items are in.

DDG does not block IPs.  Instead, a CAPTCHA wall is placed in front of request on a dubious request.

## Related issues

- #3927


return42:
Since https://github.com/searxng/searxng/pull/3947 is merge we can see the CAPTCHA issues of the ddg engine:

grafik

The reliability is low .. in particular, when the second or third results page is called up, ddg responds with a CAPTCHA.

PR #3948, opened 2024-10-19
[feat] rss: add xsl style to view rss in browser

### LINKED ISSUE #897: Printable and shareable and readable feeds
This is a great (and cheap) addition to add to SearxNG.

Not only it makes RSS pages to look better, it also makes them printable, and also exportable to PDF with hyperlinks too.  See attachments set

Test link: https://archlinuxarm.org/forum/feed.php
a) [screenshot](https://github.com/OtterBrowser/otter-browser/files/8014673/Arch.Linux.ARM.png.pdf)
b) [shareable in pdf and hyperlinks](https://github.com/OtterBrowser/otter-browser/files/8014680/Arch.Linux.ARM.pdf)
c) [without xsl](https://github.com/OtterBrowser/otter-browser/files/8014724/Arch.Linux.ARM.otter.pdf)

https://github.com/QupZilla/qupzilla/issues/1361#issuecomment-262802841

Searx https://github.com/searx/searx/issues/2051 @dalf @gloschtla @return42 @sjehuda @unixfox

Bnyro:
- this allows viewing the rss in the browser without downloading it (to test it for example)
- see #897

- enable the rss output format and use it

## Author's checklist
- the styling is very basic, but it's sufficient for what it's for imo

## Related issues
closes #897


searx/static/themes/simple/xsl/rss.xsl
```diff
@@ -1,59 +1,39 @@
...
-			
-                a {
-					color: #334999;
-					text-decoration: none;
-				}
-
-				a:hover {
-					text-decoration: underline;
-				}
-            </style>
```
return42: I think there is no need to style the generated HTML .. the default rendering for HTML elements is good enough.

searx/templates/simple/opensearch_response_rss.xml
```diff
@@ -21,34 +21,11 @@
...
-      {% endfor %}
-    {% endif %}
-    {% if suggestions %}
-      {% for a in suggestions %}
-      <item>
-        <title>{{ a }}</title>
-        <type>suggestion</type>
-      </item>
-      {% endfor %}
-    {% endif %}
```
return42, line 51: Thanks to this PR its the first time I see we had answers, corrections and suggestions in the result list.  Theses had been marked by a `<type>` tag .. but in the RSS 2.0 spec there does not exists such a tag .. instead of producing invallid XML, the RSS feed should only contain real results, nothing more.

Bnyro: True, I agree on that.

searx/static/themes/simple/xsl/rss.xsl
Bnyro: I personally prefer styling this a bit as I hate the default `<a>` styling of HTML, and that small CSS part here is not much of an overhead (just 10 lines).

return42: OK, but then CSS should be implemented in our LESS workflow.

I pushed a commit ``[mod] add CSS (LESS) to xsl style to view rss in browser`` on top of your branch .. may you might have a look at / thanks :+1:

Bnyro:
> OK, but then CSS should be implemented in our LESS workflow.
>
> I pushed a commit [mod] add CSS (LESS) to xsl style to view rss in browser on top of your branch .. may you might have a look at / thanks 👍
>
That's much better now as we can use the CSS variables from the simple theme now, thank you for the commit!

Tested and looks almost perfect to me! I noticed that there is no horizontal space anymore now, I would suggest adding `margin-inline: 1rem` to `body`.

I'm doing so and squashing everything.

sjehuda:
Thank you @Bnyro and @return42

PR #3951, opened 2024-10-20
chore: Chore/docker

lfelipeapo:
Alterado configs do dockerfile e docker-compose


unixfox:
Don't push your secret_key publicly. You should change it now.

PR #3957, opened 2024-10-24
Update calculator.py

### LINKED ISSUE #3956: calculator reports: AttributeError: 'NoneType' object has no attribute 'number_symbols'
In my server logs I see a lot of:

```
ERROR   searx.plugins.calculator      : Exception while calling post_search
  File "/usr/local/searxng/searxng-src/searx/plugins/__init__.py", line 203, in call
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searxng-src/searx/plugins/calculator.py", line 109, in post_search
    decimal = search.search_query.locale.number_symbols["latn"]["decimal"]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'number_symbols'
```

yeger00:
Add default locale in case it doesn't configured (as it can be None)

## What does this PR do?

Fix issue https://github.com/searxng/searxng/issues/3956

instead of crashing, it uses a default locale. I choose to change it in the plugin since this is the only such usage, but it will make sense to change in the query class itself. Let me know what you think. Thanks

## Related issues

Closes #3956


yeger00:
Fixed lint after https://github.com/searxng/searxng/actions/runs/11506185347

return42:
@yeger00 thanks a lot for your PR :+1: .. while I was in review with your PR I realized; we use the search language where we should better use the UI language / I Implemented a PR for:

- https://github.com/searxng/searxng/pull/3964

Issue #3763, opened 2024-08-24
Preferences show and overwrite unknown cookies


GenericMale:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
2024.8.23+4f7dd05d9

**How did you install SearXNG?**
searxng-docker

**What happened?**
The preferences cookies tab shows **all** cookies which are sent by the client, even those which are not from searxng.
Furthermore when the preferences are saved, **all** the cookies will be set again with a new Max-Age and without HttpOnly, Secure and other flags.
This could potentially cause security issues if for example an authentication service is used on the same domain (e.g. authelia).
Returning all cookies in the HTML effectively bypasses the HttpOnly flag and storing them back without HttpOnly, Secure and a very long Max-Age can be very problematic for security critical cookies. Furthermore setting a Max-Age converts all session cookies into permanent cookies.

As a workaround, the problematic cookies (e.g. the authelia session cookie) can be removed in a reverse proxy via a rewrite rule before the request is forwarded to searxng.

**How To Reproduce**

1. Run another service on the same domain which sets some cookies or manually set a cookie via javascript.
2. Open the preferences, navigate to the cookies tab and click save. All the cookies get the same flags and Max-Age, even those not from searxng.

**Expected behavior**
Only the cookies known to searxng are displayed and saved in the preferences.

glanham-jr:
I'd need to investigate more, but I'll post my current code findings.

Start of the call-chain is in webapp.py `preferences` endpoint: https://github.com/searxng/searxng/blob/master/searx/webapp.py#L881-L900

This makes a redirection call to 'index', which I'm not quite sure how thats process or what is returned. Will need to investigate this more. This might be made just so the information on the "save' is set to the response 

Cookies are saved in the preferences.py in the `save` method and dependent chains: https://github.com/searxng/searxng/blob/master/searx/preferences.py#L570-L582

I imagine the update needs to be somewhere in the `save` method, and possibly dependent save methods, to check that these were valid to begin with. Will need to run locally and log to get more results.

glanham-jr:
@return42 Would you happen to have any context into the `unknown_params` dict? See the below commit.

https://github.com/searxng/searxng/commit/237d251d14e50d937805fcf030c4fbdb9d3918c9

[...]

PR #3966, opened 2024-10-28
[cleanup] remove unknown_params usage in preferences

glanham-jr:
## What does this PR do?

remove `unknown_params` usage in preferences - removing obsolete code. Also add some more unit tests to cover some basic cases for Preferences.

Possible security issues in allowing arbitrary cookie manipulation originating not from SearXNG. Doesn't completely fix the issue, but a step towards mitigation.

More context from my comment here:

https://github.com/searxng/searxng/issues/3763#issuecomment-2424230198

1. `make run`
2. Edit your browser cookies to add a random cookie not set by SearXNG
3. Verify that this does not show in Preferences

## Author's checklist

 - n/a

## Related issues

 - #3763

### DIFF
--- a/tests/unit/test_preferences.py
+++ b/tests/unit/test_preferences.py
@@ -1,10 +1,13 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=missing-module-docstring, invalid-name
 
+import flask
+from mock import Mock
 from tests import SearxTestCase
 from searx import favicons
 from searx.locales import locales_initialize
 from searx.preferences import (
+    Setting,
     EnumStringSetting,
     MapSetting,
     SearchLanguageSetting,
@@ -13,6 +16,7 @@ from searx.preferences import (
     ValidationException,
 )
 from searx.plugins import Plugin
+from searx.preferences import Preferences
 
 locales_initialize()
 favicons.init()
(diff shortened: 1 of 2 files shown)


tests/unit/test_preferences.py
```diff
@@ -155,8 +159,48 @@ def test_encode(self):
...
+    def test_false_key_value_setting(self):
+        setting_key = 'foo'
+
+        cookie_callback = {}
+
+        def set_cookie_callback(name, value, max_age):  # pylint: disable=unused-argument
+            cookie_callback[name] = value
+
+        response_mock = Mock(flask.Response)
+        response_mock.set_cookie = set_cookie_callback
```
Bnyro, line 198: I might be wrong, but wouldn't it be easier/cleaner to use a lambda function here?

glanham-jr: I was debugging with the lambda at one point and needed a multi line method - just didn't refactor it to a lambda. I'm indifferent if we want to in-line the function.

return42: In my opinion it is a callback function, and I prefer to read a clear callback definition (multi-line) over a more hidden inline (lambda) function.

@Bnyro, @glanham-jr: I haven't looked at the details of the PR yet, but is there anything else you think this PR shouldn't be merged yet?

Bnyro: Looks good to me.

Bnyro (approved):

PR #3980, opened 2024-10-31
[fix] google: display every result when keyword is contained in conte…

uply23333:
## What does this PR do?

- Display all results that contain keyword in Google search results, ensuring users have access to relevant content matching their search terms.

In certain scenarios, using `!go` to search specific keywords may only yield a few results per page, while many relevant results are incorrectly filtered out. 

For example, searching for “Marvel Rivals !go” currently returns only two results, even though there are more that match the criteria. 

This change is important to ensure that all relevant results are displayed, improving the accuracy and comprehensiveness of search results for users.

- !go searxng stats


return42 (approved):
That's the kind of patch we all love; small change, big improvement :-) ... works like a charm / thank you :+1:

Issue #3983, opened 2024-11-01
CI often fails for no reason


return42:
In the CI log in 9 / 10 fails we see that the unit test is passed (OK) and the coverage process ends with an exit code (2). --> after a "Re-Run all failed jobs" the jobs are passed without any fail (in a second or third attempt !?!). 

```
./manage test.unit
INFO:  sourced NVM environment from /home/runner/.nvm
TEST      tests/unit

Ran 286 tests in 2.483s
... 
OK
...
./manage test.coverage
INFO:  sourced NVM environment from /home/runner/.nvm
TEST      unit test coverage
...
======================================================================
FAIL: test_localized_query_07_1_0_1_0 (tests.unit.test_plugin_calculator.PluginCalculator)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/searxng/searxng/local/py3/lib/python3.10/site-packages/parameterized/parameterized.py", line 620, in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
  File "/home/runner/work/searxng/searxng/tests/unit/test_plugin_calculator.py", line 88, in test_localized_query
    self.assertIn('calculate', search.result_container.answers)
AssertionError: 'calculate' not found in {}

...
ERROR: test.coverage exit with error (1)
make: *** [Makefile:94: test.coverage] Error 1
Error: Process completed with exit code 2.
```

When building make target `test.coverage` on a local machine, the last line in the LOG is:

```
Wrote HTML report to file:///share/searxng/coverage/index.htmlcoverage/index.html
```

The absence of this message suggests that the coverage test in the CI has problems writing the file, which in turn suggests that this problem is related to the upload-artifact@v3  action (BTW v3 is deprecated):

https://github.com/searxng/searxng/blob/cc148a76b0abb91986db5b1183458d636face5e3/.github/workflows/integration.yml#L48-L55

--------------

Another issue often seen in the CI (but never seen in a local run) is a segmentation fault:

```
./manage test.unit
INFO:  sourced NVM environment from /home/runner/.nvm
TEST      tests/unit
PYENV     [check] import yaml --> OK
PYENV     OK
+ local/py3/bin/python -c 'import sys; sys.path.pop(0); import  searx ;'
+ python -m nose2 -s tests/unit
ERROR:searx:Error while getting the git URL & branch: fatal: no such branch: 'master'

[...]

PR #3984, opened 2024-11-01
[fix] CI: remove target test.coverage from python's test matrix

return42:
The test.coverage cause a lot of failed CI jobs for reasons that cannot be explained.  As we do not monitor the coverage anyway, it is superfluous to run this job, especially as it only has a disruptive effect on the CI.

BTW and the CI action upload-artifact@v3 is depricated [1]

[1] https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact

Related: https://github.com/searxng/searxng/issues/3983

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -45,14 +45,6 @@ jobs:
         make V=1 gecko.driver
     - name: Run tests
       run: make V=1 ci.test
-    - name: Test coverage
-      run: make V=1 test.coverage
-    - name: Store coverage result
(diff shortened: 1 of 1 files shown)


return42 (commented):
Since we are currently having massive problems running the tests and building the master branch, I will merge this hotfix / this PR (without further review).

If someone has better suggestions or knows reasons how and why we should continue the coverage; this decision can also be withdrawn. 

It is just necessary to ensure that the operation of further development is guaranteed.

PR #3985, opened 2024-11-01
[fix] annas archive: crash when no thumbnail, differing results, paging

Bnyro:
## What does this PR do?
- fixes a crash when there's no thumbnail/image available
- fixes that results differ from annas archive due to bad filter handling (we sent filters that were None)
- support for paging


return42 (approved):
Much better than before / thanks :+1:

PR #4001, opened 2024-11-09
[feat] engine: add adobe stock photos

Bnyro:
## What does this PR do?
- this PR adds support for Adobe Stock photos
- the full image (when expanded by the user) is usually watermarked

- !as tree
- try some other queries and test pagination

### DIFF
--- a/docs/dev/engines/online/adobe_stock.rst
+++ b/docs/dev/engines/online/adobe_stock.rst
@@ -1,0 +1,13 @@
+.. _adobe stock engine:
+
+===========
+Adobe Stock
+===========
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.adobe_stock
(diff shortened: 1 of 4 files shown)


return42 (approved):
@Bnyro sorry for the delay .. it took me a while, but now I got it: I placed a commit on top to add adobe stock video and audio engines.

There is now the option ``adobe_content_types`` with which it is possible to configure engines for video and audio from the adobe stock.  BTW this patch adds documentation to the engine.

To test all three engines in one use a search term like::

    !asi !asv !asa sound

Bnyro:
> @Bnyro sorry for the delay .. it took me a while, but now I got it: I placed a commit on top to add adobe stock video and audio engines.
>
No worries about the delay, thank you for taking a look at this PR!

Your changes look good to me, I didn't consider adding these different media types because I personally always used adobe stock for pictures only, but that's definitely nice to have too!

PR #3578, opened 2024-06-15
[mod] revision of the settings_loader

return42:
The intention of this PR is to modernize the settings_loader implementations. The concept is old (remember, this is partly from 2014), back then we only had one config file, meanwhile we have had a folder with config files for a very long time.  Callers can now load a YAML configuration from this folder as follows ::

    settings_loader.get_yaml_cfg('my-config.yml')

- BTW this is a fix of #3557.

- Further the `existing_filename_or_none` construct dates back to times when there was not yet a `pathlib.Path` in all Python versions we supported in the past.

- Typehints & documentation have been added wherever appropriate

At the same time, this patch should also be downward compatible and not introduce a new environment variable. The localization of the folder with the configurations is further based on:

    SEARXNG_SETTINGS_PATH (wich defaults to /etc/searxng/settings.yml)

Which means, the default config folder is `/etc/searxng/`.

> [!IMPORTANT] 
> **intended functional changes!**
>
> If environment `SEARXNG_SETTINGS_PATH` was set and pointed to a not existing file, the previous implementation silently loaded the default configuration.  This behavior has been changed: if the file or folder does not exist, an EnvironmentError exception will be thrown in future.

Closes: https://github.com/searxng/searxng/issues/3557

PR #4002, opened 2024-11-10
Support SEARXNG_SETTINGS_PATH folder in docker

silver886:
## What does this PR do?

Support `SEARXNG_SETTINGS_PATH` folder in Docker in order use https://github.com/searxng/searxng/blob/master/searx/settings_loader.py#L64-L88.

Without this change, it seems no way to config `settings.yml`, `limiter.toml`, and `favicons.toml` at the same time without mounting volume to `/etc/searxng`.

Build docker image and set `SEARXNG_SETTINGS_PATH` to a folder.

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -135,6 +135,10 @@ update_conf() {
     fi
 }
 
+if [ -d "${SEARXNG_SETTINGS_PATH}" ]; then
+    $SEARXNG_SETTINGS_PATH="${SEARXNG_SETTINGS_PATH}/settings.yml"
+fi
+
 # searx compatibility: copy /etc/searx/* to /etc/searxng/*
 SEARX_CONF=0
 if [ -f "/etc/searx/settings.yml" ]; then


return42:
The evaluation of the environmental variables `SEARXNG_SETTINGS_PATH` has been implemented in: 

- https://github.com/searxng/searxng/pull/3578
- https://github.com/searxng/searxng/pull/3962
- https://github.com/searxng/searxng/pull/3954

Setting ``$SEARXNG_SETTINGS_PATH="${SEARXNG_SETTINGS_PATH}/settings.yml"`` should not change anything, if ``get_user_cfg_folder(..)`` does not work as described below, we have a bug that needs to be fixed (but this error would then not be limited to docker).

https://github.com/searxng/searxng/blob/2fbf15eccbbc3af70f8b93257c9a39d26603f097/searx/settings_loader.py#L64-L88

silver886:
Ah, okay.
I was about to close this pull request since I figured out a solution.

I set `SEARXNG_SETTINGS_PATH` to a folder and mount it with `ro`, and it works.

PR #4003, opened 2024-11-10
Update enabled_plugins in settings

silver886:
## What does this PR do?

Remove 'Autodetect search language', which is no longer valid, from settings, and add 'Unit converter plugin', which is now default enabled, to settings.

Reflect current, or default, settings in `settings.yml`.

It's default commented.


return42 (approved):
> Reflect current, or default, settings in settings.yml

Thanks a lot for your patch .. I just modified the commit message (and did a force push).

PR #3955, opened 2024-10-22
[refactor] engine: duckduckgo - https://html.duckduckgo.com/html

return42:
The entire source code of the duckduckgo engine has been reengineered and purified.

1. DDG used the URL https://html.duckduckgo.com/html for no-JS requests whose response is also easier to parse than the previous https://lite.duckduckgo.com/lite/ URL

2. the bot detection of DDG has so far caused problems and often led to a CAPTCHA, this can be circumvented using `'Sec-Fetch-Mode'] = “navigate”`

Closes: https://github.com/searxng/searxng/issues/3927

PR #4011, opened 2024-11-14
[fix] engine: duckduckgo - only uses first word of the search terms 

### LINKED ISSUE #4009: search string pulls up odd results for multiple different searches

Powered by [searxng](https://searx.bndkt.io/info/en/about) - 2024.11.10+dfaf5868e

no I'm using the search at this URL:
 
https://searx.bndkt.io/

**What happened?**

I search for this exact string
"what types of shares are there in a company"

and it produces one result at the top that explains what a share is from some indian site (e.g. Different types of shares include ordinary, preference, redeemable preference, convertible preference and treasury shares. Shares represent ownership in a company and are an essential aspect of the corporate world.
[www.bajajfinserv.in](https://www.bajajfinserv.in/types-of-shares) ) then every other site search (minus the first one which is a link for whatsapp) result is the definition of the word 'what' (e.g. https://www.merriam-webster.com/dictionary/what ). then it goes back to share pages for a few iterations and then it goes back to definitions of 'what' and then goes to looking up pages that contain the word 'what' in them not related to anything at all (e.g. https://www.nytimes.com/2024/10/23/opinion/election-polls-results-trump-harris.html)

moving to page 2 of results then shows normal expected results of share definitions but moving back to page 1 then shows correct share definitions and is pulling resources from a different cache. 

image
not sure if this is related to the search string being removed after navigating the search results or something?

**How To Reproduce**

search for the string what types of shares are there in a company 
URL result is
https://searx.bndkt.io/search?q=what+types+of+shares+are+there+in+a+company
**Expected behavior**

not to get so many mixed search results in a clear type of search function.

**Screenshots & Logs**

image

image

image

**Additional context**

nicolasdato:
Suddenly I'm getting odd results from DuckDuckGo, like if only searches for the first word in the query.

This change improved the results and looks normal and expected to me.

## What does this PR do?

I'm getting the same issue as #4009, DuckDuckGo results are only for the first word in the query.
With this change, I'm getting normal results now.

It's very annoying to have nonsense results, if I search for "history of Egypt", the only results from DuckDuckGo are definition of "history", wikipedia's page of "history", and so on. This is reported in #4009

So a fix to this issue is important.

I don't know

## Author's checklist

I don't know why `query` was a list. Maybe it should work that way too, and instead of changing duckduckgo.py this should be changed somewhere else, like the module making the request.

## Related issues

Closes #4009

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -6,7 +6,7 @@ DuckDuckGo Lite
 
 from typing import TYPE_CHECKING
 import re
-from urllib.parse import urlencode
+from urllib.parse import urlencode, quote_plus
 import json
 import babel
 import lxml.html
@@ -245,10 +245,12 @@ def request(query, params):
 
     # Advanced search syntax ends in CAPTCHA
     # https://duckduckgo.com/duckduckgo-help-pages/results/syntax/
-    query = [
-        x.removeprefix("site:").removeprefix("intitle:").removeprefix("inurl:").removeprefix("filetype:")
-        for x in query.split()
-    ]
+    query = " ".join(
+        [
+            x.removeprefix("site:").removeprefix("intitle:").removeprefix("inurl:").removeprefix("filetype:")
+            for x in query.split()
+        ]
+    )
     eng_region = traits.get_region(params['searxng_locale'], traits.all_locale)
     if eng_region == "wt-wt":
         # https://html.duckduckgo.com/html sets an empty value for "all".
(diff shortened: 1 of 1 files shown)


return42 (approved):
> I don't know why `query` was a list.

during the revision in PR https://github.com/searxng/searxng/pull/3955 the query string was accidentally converted into
a list of words, further the query must be quoted before POSTed in the ``data``
field.. [``urllib.parse.quote_plus``](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote_plus)

------------

Thanks a lot for your fix, I modified the commit:

1. join ``query`` list to a space separated *query string*
2. from ``urllib.parse` use ``quote_plus(query)`` to quote *query string*
3. modified the commit message / see 3539228

Issue #4020, opened 2024-11-16
Problem with DDG and Hebrew search


ShlomiD83:
Searxng is running inside a docker container on a raspberry pi 4.
Since version  - 2024.11.15+4b57bc3db I'm having issues with Hebrew searching, the results from duckduckgo are in gibberish. 
![Screenshot_20241116_221012_Chrome](https://github.com/user-attachments/assets/b022b828-973a-4580-9eca-52919acfc6cf)
![Screenshot_20241116_221019_Chrome](https://github.com/user-attachments/assets/d36d1111-b90e-46b3-a156-0d427f356268)
When Reverting to the 10.11 version, the issue doesn't happen. 
Please assist.

ShlomiD83:
https://github.com/searxng/searxng/issues/4019

This seems to be the same issue.

ShlomiD83:
@unixfox what does that mean? 🤔

return42:
> @unixfox what does that mean? 🤔

The issue you opened is a duplicate of https://github.com/searxng/searxng/issues/4019 ..
in order not to fray the discussion we have closed your issue, the discussion will be continued in the already existing issue.

Issue #4019, opened 2024-11-16
When using Chinese search terms, the results from DuckDuckGo (DDG) always "URL-decode" related URLs


Aliang-code:
Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG 
2024.11.15+4b57bc3db 

How did you install SearXNG? 
searxng-docker 

What happened? 
When using Chinese search terms, the results from DuckDuckGo (DDG) always URL-decode related URLs. 

How To Reproduce 
Use Chinese search terms to query, regardless of language setting being "all" or "zh". 

Expected behavior 
DDG results should be relevant to the search query without URL encoding related URLs. 

Screenshots & Logs 
![图片](https://github.com/user-attachments/assets/b817346b-ad99-4c6a-b221-eadcf0540817)
![图片](https://github.com/user-attachments/assets/66745f4a-b8f4-49c2-963e-7a4a4d49b54d)

Additional context 
https://github.com/searxng/searxng/pull/4011
maybe this fix is not completion

return42:
> Additional context
> #4011
> maybe this fix is not completion

Yeah, I url-qouted the query string in #4011 .. but the URL-qouted query string result in unexpected *URL decoded* and other garbish results .. to fix the issue I implemented this PR:

- #4021

please give it a try :+1:

ShlomiD83:
> please give it a try 👍

My Hebrew problem seems to be resolved with this fix.

return42:
Closed by https://github.com/searxng/searxng/pull/4021

PR #4021, opened 2024-11-17
[fix] engine: duckduckgo - don't quote query string

return42:
The query string send to DDG must not be qouted.

The query string was URL-qouted in #4011, but the URL-qouted query string result in unexpected *URL decoded* and other garbish results as reported in #4019 and #4020

To test compare the results of a query like / I deployed this patch on my [instance darmarit.org](https://darmarit.org/searx/)::

    !ddg Häuser und Straßen :de
    !ddg Häuser und Straßen :all
    !ddg 房屋和街道 :all
    !ddg 房屋和街道 :zh

Closed:

- [#4019]
- [#4020]

Related:

- [#4011] https://github.com/searxng/searxng/pull/4011

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -6,7 +6,7 @@ DuckDuckGo Lite
 
 from typing import TYPE_CHECKING
 import re
-from urllib.parse import urlencode, quote_plus
+from urllib.parse import urlencode
 import json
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
I can confirm your findings from my tests, the results are much better with this PR.

PR #4034, opened 2024-11-23
[fix] engine Library of Congress: fix API URL loc.gov -> www.loc.gov

return42:
Avoid HTTP 404 and redirects. Requests to the JSON/YAML API use the base url [1]

    https://www.loc.gov/{endpoint}/?fo=json

[1] https://www.loc.gov/apis/json-and-yaml/requests/

### DIFF
--- a/searx/engines/loc.py
+++ b/searx/engines/loc.py
@@ -27,7 +27,7 @@ categories = ['images']
 paging = True
 
 endpoint = 'photos'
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Can confirm this fixes the engine :+1:

(gives a parsing error without the change because it's not getting a proper API response)

PR #4039, opened 2024-11-23
[doc] update quantities on the intro page

return42:
The quantities on the intro page were partly out of date / example; we already have 210 engines and not just 70. To avoid having to change the quantities manually in the future, they are now calculated from the jinja context.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,26 +4,31 @@ Welcome to SearXNG
 
   *Search without being tracked.*
 
-SearXNG is a free internet metasearch engine which aggregates results from more
-than 70 search services.  Users are neither tracked nor profiled.  Additionally,
-SearXNG can be used over Tor for online anonymity.
+.. jinja:: searx
+
+   SearXNG is a free internet metasearch engine which aggregates results from up
+   to {{engines | length}} :ref:`search services <configured engines>`.  Users
+   are neither tracked nor profiled.  Additionally, SearXNG can be used over Tor
+   for online anonymity.
 
 Get started with SearXNG by using one of the instances listed at searx.space_.
 If you don't trust anyone, you can set up your own, see :ref:`installation`.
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
That's cool, I didn't yet quite get how accessing the Jinja context works here, but it works fine locally :+1:

return42:
> I didn't yet quite get how accessing the Jinja context works here,

Its quite simple, you need the sphinx extension:

https://github.com/searxng/searxng/blob/c4b874e9b0cf2b54b0152d95aa5f5b00543383c6/docs/conf.py#L122

and define your contexts (plural form) .. its a dict of key/value pairs. Here we define (the only) context we have `searx`:

https://github.com/searxng/searxng/blob/c4b874e9b0cf2b54b0152d95aa5f5b00543383c6/docs/conf.py#L54-L65

To evaluate a reST markup with ninja replacements, place the reST markup in a `.. jinja:: ` directive and give the name of the context (`searx`) as the first argument:

https://github.com/searxng/searxng/blob/c4b874e9b0cf2b54b0152d95aa5f5b00543383c6/docs/user/configured_engines.rst?plain=1#L14-L17

Jinja tags like ``{{engines | length}}`` will be evaluated in the context and inserted at this place.

These contexts can be used to create entire documentations. Here is a more complex (and already slightly cryptic) example that generates all the tables we see on https://docs.searxng.org/user/configured_engines.html.

https://github.com/searxng/searxng/blob/c4b874e9b0cf2b54b0152d95aa5f5b00543383c6/docs/user/configured_engines.rst?plain=1#L30-L89

BTW: thanks for your review :+1:

Bnyro:
That's really cool, thank you for the detailed explanation!

Looking at this extension, I definitely see why you've always told me that Sphinx is actually quite cool and worth learning the syntax, even though it's not as commonly known as Markdown.

PR #4041, opened 2024-11-23
[fix] ddg extra crashes and returns no results (images, videos, ...)

Bnyro:
## What does this PR do?
- this PR tries to get DuckDuckGo extra (especially images) to work again
- it's been broken since the recent switch at https://github.com/searxng/searxng/pull/3955

Previously, these engines are just crashing as we pass the query to `_get_vqd()` and not a data `dict` object at https://github.com/searxng/searxng/blob/master/searx/engines/duckduckgo_extra.py#L53.

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """
-DuckDuckGo Lite
-~~~~~~~~~~~~~~~
+DuckDuckGo WEB
+~~~~~~~~~~~~~~
 """
+
+from __future__ import annotations
 
 from typing import TYPE_CHECKING
 import re
-from urllib.parse import urlencode
+from urllib.parse import urlencode, quote_plus
 import json
 import babel
 import lxml.html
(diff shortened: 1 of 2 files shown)


searx/engines/duckduckgo.py
```diff
@@ -60,42 +62,30 @@
...
-        SearXNG's IP on a block list.
-
-        Requests from IPs in this block list run into timeouts.
-
-    Not sure, but it seems the block list is a sliding window: to get my IP rid
-    from the bot list I had to cool down my IP for 1h (send no requests from
-    that IP to DDG).
+    :param query: The query term
+    :param region: DDG's region code
+    :param force_request: force a request to get a vqd value from DDG
```
return42, line 88: Added force_request argument to the function -> so the caching is encapsulated in duckduckgo.py and calls from duckduckgo_extra.py do not need to implement anything for caching.

return42 (approved):
> - it's been broken since the recent switch at https://github.com/searxng/searxng/pull/3955

Shame on me .. did not had these DDG engines in the scope :facepalm: 

Thanks for the patch :+1: .. I slightly modified in hope to simplify the implementation a little bit.

searx/engines/duckduckgo.py
Bnyro: That's a good idea indeed, I thought about doing something similar too, but didn't want to make that function more complex than it is already. Your solution is a nice one though.

PR #4055, opened 2024-11-27
[feat] search: shift/ctrl click a category to select multiple categories

### LINKED ISSUE #2527: [Feature] Shift/Ctrl Click a Category To Select Multiple Categories
Add the ability to shift/ctrl click another category to select it. This would work similar to how file selection works in GUI file managers (or just any kind of multi-selection functionality found in GUIs).
Clicking a category would deselect all other categories and search only for that category (similar to what the `Search on category select` plugin does).
Clicking a category while holding shift/ctrl would select that category and retain all the other selected categories (similar to the default behavior).

This would remove the need for the `Search on category select` plugin. Perhaps this request could be made into a new plugin which replaces it.

Bnyro:
## What does this PR do?
- This PR allows people that have `search on category select` enabled to select multiple categories by pressing `Shift` while selecting new categories

## Related issues
closes #2527


searx/templates/simple/categories.html
```diff
@@ -12,7 +12,7 @@
     'social media': 'people-outline',
 }  -%}
 <div id="categories" class="search_categories">{{- '' -}}
-    <div id="categories_container">
+    <div id="categories_container" class="nojs">
```
dalf: I'm not sure if it can replace this line change, but there is `ns-js` or `js` class on the `<html>` element:

https://github.com/searxng/searxng/blob/342d321196a84437d51c75da348c533e01626361/searx/templates/simple/base.html#L2

https://github.com/searxng/searxng/blob/342d321196a84437d51c75da348c533e01626361/searx/static/themes/simple/src/js/head/00_init.js#L17-L18

Bnyro: Indeed we can, thank you Alexandre!

return42 (approved):

searx/static/themes/simple/src/js/main/search.js
```diff
@@ -166,21 +166,46 @@
...
-          const selected = d.querySelector("button.category_button.selected");
-          console.log(selected);
-          selected.classList.remove("selected");
-        })
-      }
+    // remove CSS class that's only needed when the browser supports no JavaScript
+    const categoriesContainer = d.querySelector("#categories_container");
+    if (categoriesContainer) {
+      categoriesContainer.classList.remove("nojs");
+    }
```
dalf, line 173: nitpicking: is it a leftover?

Bnyro: Indeed, good catch

PR #4056, opened 2024-11-27
[fix] google engine: remove <script> tags from result items

return42:
In some results, Google returns a <script> tag that must be removed before extracting the content.

I tested the patch with search term: `!go :de the fog of war deutsch`

Before:

grafik

After this patch:

grafik


Bnyro (approved):
Tested with the example you provided, looks good to me :+1:

PR #4057, opened 2024-11-27
[feat] search: add url formatting preference

### LINKED ISSUE #3376: Ability to customise result URL formatting
I personally find it weird that the result URLs are formatted using spaces and `>` characters, instead of `/` like one expects to see in a URL. This formatting looks like a [breadcrumb trail](https://en.wikipedia.org/wiki/Breadcrumb_navigation), which doesn't always match up with the URL of a webpage, so it's off-putting.

I understand this was requested in https://github.com/searxng/searxng/issues/504 as a way to make the results page cleaner, but I'd like a way to disable it.

Would it be possible to add a setting to toggle this behaviour on and off? That way everyone can have the formatting they prefer. DuckDuckGo has a similar setting:

image

Leaving things as they are, or reverting them to how they were, are options; but they don't allow all users to have the formatting they like.

**Additional context**
image

Bnyro:
## What does this PR do?
There are 3 possible values for URL formatting:
- Pretty: as before, e.g. `https://foo.bar > patha > pathb`
- Full: full url, e.g. `https://foo.bar/patha/pathb`
- Host: host of website, e.g. `foo.bar`

## Related issues
closes #3376


return42 (changes requested):
Nice addition :+1: .. I tested with RTL language (hebrew) and got a slash on some result items on the left side:

grafik

return42:
@vojkovic, @alr86 would you please have a look at this PR / thanks a lot :+1: 

This PR does not fix, but is related to:

- https://github.com/searxng/searxng/issues/3724

Bnyro:
> Nice addition 👍 .. I tested with RTL language (hebrew) and got a slash on some result items on the left side:
> 
> grafik
>
Fixed by setting
```css
.url_i1 {
        unicode-bidi: plaintext;
}
```
see https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi.

PR #4058, opened 2024-11-27
[feat] json/xpath engine: config option for method and body

### LINKED ISSUE #701: Make request parameters configurable for `json_engine` and `xpath`
Somebody on IRC recounted that they tried to add a [Manticore](https://manticoresearch.com/) engine to their SearXNG instance using the `json_engine` but failed to do so because Manticore requires POST requests and `json_engine` apparently only supports GET requests.

Unless I'm missing something they're correct.  Looking into this I think the most elegant way to solve this is to allow overriding the [request `params`](https://docs.searxng.org/dev/engine_overview.html#specify-request) in the [engine settings](https://docs.searxng.org/admin/engines/settings.html#settings-engine), so e.g. something like:

```yml
- name: example engine
  engine: json_engine
  request_params:
    method: POST
```

To clarify this would work for any engine module: the specified request parameters would simply be applied as overrides just before sending the request (and could e.g. also be used to specify a specific HTTP header for an engine). The main appeal for this of course however is to allow `json_engine` and `xpath` engines that use `POST`.

(The exact config format is of course up to debate). What do you think?

Bnyro:
## What does this PR do?
- This PR adds support for the `method` and `request_body` request parameters
- We currently don't have any engine using it, but it might be useful for people that want to implement their own json/xpath engines quickly, and having these configuration options could probably have replaced some simple engine modules

## Related issues
closes  #701

### DIFF
--- a/searx/engines/json_engine.py
+++ b/searx/engines/json_engine.py
@@ -16,23 +16,17 @@ from json import loads
 from urllib.parse import urlencode
 from searx.utils import to_string, html_to_text
 
-
+# parameters for generating a request
 search_url = None
-url_query = None
-url_prefix = ""
-content_query = None
-title_query = None
-content_html_to_text = False
-title_html_to_text = False
-paging = False
-suggestion_query = ''
-results_query = ''
+method = 'GET'
+request_body = ''
 
 cookies = {}
 headers = {}
 '''Some engines might offer different result based on cookies or headers.
 Possible use-case: To set safesearch cookie or header to moderate.'''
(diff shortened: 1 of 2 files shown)


return42 (approved):
Thanks, LGTM :+1:  .. I patched the doc a little bit --> [diff](https://github.com/searxng/searxng/compare/542b5db62e363a034700a3bba017eb146420354d..6885e85639397cd6399ce5b5985c882653722b46)

PR #4062, opened 2024-11-27
[fix] wikicommons engine: remove HTML tags from result items

return42:
BTW: humanize filesize (Bytes) to KB, MB, GB ..

Search term for a test: `!wikicommons.files :en The Fog of War`

Before this patch ..

grafik

After this patch:

grafik


Bnyro (approved):
Tested and can confirm the previous bugs and the fix with this PR, thank you for looking into it @return42 :+1:

PR #4063, opened 2024-11-27
[chore] drop lingva engine / lingva-translate no onger maintained

return42:
The lingva-translate seems to be dead. Even the instances specified in the README [1] have not been updated for 12 months and no longer correspond to the current status, for example https://translate.projectsegfau.lt/ has already switched to Mozhi.

The official instance lingva.ml keeps failing [2] and seems to me to have a different interface than we have implemented in the engine.  Somehow it all seems to be a bit of a mess, I can't think of any other solution than to take the engine out of SearXNG for the time being (unfortunately).

[1] https://github.com/thedaviddelta/lingva-translate/blob/main/README.md#instances
[2] https://github.com/thedaviddelta/lingva-translate/issues/184#issuecomment-2461363898


Bnyro:
Even though it's no longer maintained, it's still working properly if you are self-hosting an instance of Lingva Translate.

Thus I'd prefer keeping the module, chances are high someone has set up the engine to use their own custom instance's Lingva Translate app that works just fine.

return42:
> Thus I'd prefer keeping the module, chances are high someone has set up the engine to use their own custom instance's Lingva Translate app that works just fine.

makes sense, but since we don't know of any instance that works, we should at least set it to `inactive=true`?

Bnyro:
Yes, that makes sense.

return42:
https://lingva.thedaviddelta.com/ seems back again .. going to close this PR.

PR #4081, opened 2024-11-29
Add an engine for ads, the astrophysical data system

DanielMowitz:
## What does this PR do?

The [astrophysics data system](https://ui.adsabs.harvard.edu/) is a solr instance providing a large collection of astrophysical data, as the name suggests.
As it uses solr, but not with the standard api paths, I modified the solr engine to successfully retrieve data from ads.

It enables searching for astrophysical data in the ads, which was not easily possible with the existing engines (xpath could have worked, but solr didn't do the trick).

- Get an API token as described in https://ui.adsabs.harvard.edu/help/api/
- Add the ads engine, i.e.

```
  - name: ads
    engine: ads
    sort: asc
    weight: 5
    categories: [science]
    api_key: your-new-key
    shortcut: ads
```

test search with `!ads`

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -174,3 +174,4 @@ features or generally made searx better:
 - @micsthepick
 - Daniel Kukula `<https://github.com/dkuku>`
 - Patrick Evans `https://github.com/holysoles`
+- Daniel Mowitz `<https://daniel.mowitz.rocks>`
--- a/searx/engines/astrophysics_data_system.py
+++ b/searx/engines/astrophysics_data_system.py
@@ -1,0 +1,93 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+""".. sidebar:: info
+
+The Astrophysics Data System (ADS) is a digital library portal for researchers in astronomy and physics,
+operated by the Smithsonian Astrophysical Observatory (SAO) under a NASA grant.
+The engine is adapted from the solr engine.
+
+"""
+
+# pylint: disable=global-statement
+
+from datetime import datetime
+from json import loads
+from urllib.parse import urlencode
+from searx.exceptions import SearxEngineAPIException
+
+about = {
+    "website": 'https://ui.adsabs.harvard.edu/',
+    "wikidata_id": 'Q752099',
+    "official_api_documentation": 'https://ui.adsabs.harvard.edu/help/api/api-docs.html',
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": 'JSON',
+}
+
+base_url = 'https://api.adsabs.harvard.edu/v1/search'
+result_base_url = 'https://ui.adsabs.harvard.edu/abs/'
+rows = 10
+sort = ''  # sorting: asc or desc
+field_list = ['bibcode', 'author', 'title', 'abstract', 'doi', 'date']  # list of field names to display on the UI
+default_fields = ''  # default field to query
+query_fields = ''  # query fields
+paging = True
+api_key = 'unset'
+
+
+def init(_):
+    if api_key == 'unset':
+        raise SearxEngineAPIException('missing ADS API key')
+
+
+def request(query, params):
+    query_params = {'q': query, 'rows': rows}
+    if field_list:
+        query_params['fl'] = ','.join(field_list)
+    if query_fields:
+        query_params['qf'] = ','.join(query_fields)
+    if default_fields:
+        query_params['df'] = default_fields
+    if sort:
+        query_params['sort'] = sort
+
+    query_params['start'] = rows * (params['pageno'] - 1)
+
+    params['headers']['Authorization'] = f'Bearer {api_key}'
+    params['url'] = f"{base_url}/query?{urlencode(query_params)}"
+
+    return params
+
+
+def response(resp):
+    try:
+        resp_json = loads(resp.text)
+    except Exception as e:
+        raise SearxEngineAPIException("failed to parse response") from e
+
+    if 'error' in resp_json:
+        raise SearxEngineAPIException(resp_json['error']['msg'])
+
+    resp_json = resp_json["response"]
+    result_len = resp_json["numFound"]
+    results = []
+
+    for res in resp_json["docs"]:
+        author = res.get("author")
+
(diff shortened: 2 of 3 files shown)


Bnyro:
Please also add the yaml code you wrote above to `searx/settings.yml` and comment it out with `#` at the front of the line.

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+# pylint: disable=global-statement
+
+from json import loads
+from urllib.parse import urlencode
+from searx.exceptions import SearxEngineAPIException
+
+about = {
+    "website": 'https://ui.adsabs.harvard.edu/',
+    "wikidata_id": 'Q752099',
+    "official_api_documentation": 'https://github.com/adsabs/adsabs-dev-api/tree/master',
```
Bnyro: nit: `tree/master` is not necessary

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+    if field_list != '':
+        query_params['fl'] = ','.join(field_list)
+    if query_fields != '':
+        query_params['qf'] = ','.join(query_fields)
+    if default_fields != '':
+        query_params['df'] = default_fields
+    if sort != '':
+        query_params['sort'] = sort
+
+    if 'pageno' in params:
```
Bnyro: That's always `True`.

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+        query_params['start'] = rows * (params['pageno'] - 1)
+
+    params['headers']['Authorization'] = 'Bearer ' + api_key
```
Bnyro:
```suggestion
    params['headers']['Authorization'] = f'Bearer {api_key}'
```

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+    params['url'] = _search_url.format(params=urlencode(query_params))
```
Bnyro:
```suggestion
    params['url'] = f"{base_url}/query?{urlencode(query_params)}"
```
Thus `_search_url` should also be dropped above as it's not needed.

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+def request(query, params):
+    query_params = {'q': query, 'rows': rows}
+    if field_list != '':
+        query_params['fl'] = ','.join(field_list)
+    if query_fields != '':
+        query_params['qf'] = ','.join(query_fields)
+    if default_fields != '':
+        query_params['df'] = default_fields
+    if sort != '':
+        query_params['sort'] = sort
```
Bnyro:
```suggestion
    if field_list
        query_params['fl'] = ','.join(field_list)
    if query_fields:
        query_params['qf'] = ','.join(query_fields)
    if default_fields':
        query_params['df'] = default_fields
    if sort:
        query_params['sort'] = sort
```
That's just simpler syntax we use generally, as it's also false when e.g. `sort = None` when users set it to none in their yaml config.

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+def __get_response(resp):
+    try:
+        resp_json = loads(resp.text)
+    except Exception as e:
+        raise SearxEngineAPIException("failed to parse response") from e
+
+    if 'error' in resp_json:
+        raise SearxEngineAPIException(resp_json['error']['msg'])
+
+    return resp_json
```
Bnyro: This should be inlined to where it's called in my opinion, as error handing is done in the main response function for most if not all other engine modules too.

searx/engines/ads.py
```diff
@@ -0,0 +1,98 @@
...
+    resp_json = resp_json["response"]
+    result_len = resp_json["numFound"]
+    results = []
+
+    for res in resp_json["docs"]:
+        url = result_base_url + res.get("bibcode") + "/"
+        title = res.get("title")[0]
+        author = res.get("author")
+        abstract = res.get("abstract")
+        doi = res.get("doi")
```
Bnyro: title, abstract and doi could be inlined to where they are used instead of creating extra variables

allendema:
I would consider changing the python filanem to something more descriptive, 
e.x avoid confusion when admins during review/installation see `ads.py`.

DanielMowitz:
> I would consider changing the python filanem to something more descriptive, e.x avoid confusion when admins during review/installation see `ads.py`.

So `astrophysics-data-system.py`?

allendema:
> So `astrophysics-data-system.py`?

Yes but in snake_case.

searx/engines/ads.py
DanielMowitz: No worries, but I think I should note that this code is taken directly from `solr.py` ;)

searx/engines/ads.py
DanielMowitz: This is also taken verbatim from `solr.py`… Fixed it.

searx/engines/ads.py
DanielMowitz: Changed to a different url altogether.

searx/engines/ads.py
DanielMowitz: Also done.

searx/engines/ads.py
DanielMowitz: Done, now the response function looks better :)

return42:
@DanielMowitz thanks a lot for your PR  .. the CI reported an issue .. FYI: you can run all the tests needed to pass for review by `make test` .. in our ["Development Quickstart"](https://docs.searxng.org/dev/quickstart.html#development-quickstart) you will find some more hints :+1:

DanielMowitz:
> @DanielMowitz thanks a lot for your PR .. the CI reported an issue .. FYI: you can run all the tests needed to pass for review by `make test` .. in our ["Development Quickstart"](https://docs.searxng.org/dev/quickstart.html#development-quickstart) you will find some more hints 👍

I did that and it seems like the unused variable was the only error. There were some other s but they seemed to be due to inactive modules.

Bnyro (approved):
Thank you!

Fyi, I squashed your commits into one and rebased the PR.

Issue #4066, opened 2024-11-28
When image load fails on client side, show default image (JS)


return42:
When image load fails on client side, the layout is more or less violated:

![grafik](https://github.com/user-attachments/assets/bda37942-e37c-4d6f-bc0c-8b3059b560e7)

In the past we had a feature that loads a default image in case of errors

- https://github.com/searxng/searxng/pull/562

but when the simple theme was revised, this feature was missed in our scope:

- https://github.com/searxng/searxng/pull/689

Would be good to get this feature back:

The image is here 

We need to add a event handler for `error`, something like:

```JS
 img.addEventListener('error', img_load_error, {once: true});
```

may be here .. 

https://github.com/searxng/searxng/blob/5a9c1c6b5bc1963482e9c8142dffe8aebaed82b8/searx/static/themes/simple/src/js/main/results.js#L12-L15

```JS
    function img_load_error (event) {
      // console.log("ERROR can't load: " + event.originalTarget.src);
      event.originalTarget.src = w.searxng.static_path + w.searxng.theme.img_load_error;
    }
```

We used `w.searxng.theme.img_load_error` in the past, like shown above .. this `theme` object seems no longer exists in the client settings, may we need to similar back to / would take place here:

https://github.com/searxng/searxng/blob/5a9c1c6b5bc1963482e9c8142dffe8aebaed82b8/searx/webapp.py#L349-L360

return42:
Closed by PR 

- https://github.com/searxng/searxng/pull/4083

PR #4083, opened 2024-11-29
[fix] if image load fails on client side, show default image

return42:
BTW: change icon color from red to gray

Closes:

- https://github.com/searxng/searxng/issues/4066

-------

Before:

grafik

After:

grafik


Bnyro (approved):
Tested and works as expected, the color change is definitely reasonable :+1:

return42:
FYI: last force push was just to solve conflict with master branch.

PR #4084, opened 2024-12-01
[refactor] webapp.py: reuse get_client_settings() method to unify pref loading on render

Bnyro:
## What does this PR do?
- the `client_settings` object in JavaScript basically needs the same preferences that are needed for preference handling when rendering the template
- thus they are merged into the same function with this PR, in order to make it unnecessary to have the duplicated code for assigning these client settings into a dictionary


return42:
Not sure why and if related to this PR (but haven't seen before): When opening the preferences I see a uncaught exception in the JS debugger: 

grafik

In one of the (thousands of) `searxng.on` calls the `obj` argument is `null`.

Bnyro:
> Not sure why and if related to this PR (but haven't seen before): When opening the preferences I see a uncaught exception in the JS debugger:
>
It also happens without this PR (tested locally), not sure what causes it though, I'll try to figure it out.

return42 (approved):
FYI rebased to get https://github.com/searxng/searxng/pull/4115 .. and tested again / works as intended / thanks :+1:

PR #4090, opened 2024-12-04
[enh] Add plugin to allow regex rewriting of URLs

joestump:
## What does this PR do?

This PR is a followup to an exchange I had with @return42 in https://github.com/searxng/searxng/discussions/4027 – It introduces a new plugin called `url_rewrite`, which is similar to the Hostnames plugin, but offers two notable enhancements:

1. It allows the use of regular expressions to manipulate any and all parts of the `result["url"]`. 
2. It allows you to optionally disable mutating `parsed_url`. I added this because some of my content proxies use GET arguments so 100% of the result URLs were just `proxy.example.com`. 

I have been running this plugin on my local SearXNG for a few days now without issues.

I use two content proxies locally: 

* https://github.com/everywall/ladder 
* https://github.com/joestump/reader

In some cases, e.g. NYT, I wanted to use both together. However, this would require mutating GET arguments for the `result["url"]`, which the existing hostnames plugin did not support. 

1. Build a container locally [with the plugin installed](https://github.com/joestump/searxng-privacy-plugins?tab=readme-ov-file#create-a-docker-container).
2. Run either Ladder or Reader locally – they're both very easy to run. (not strictly required)
4. Update [`settings.yml`](https://github.com/joestump/searxng-privacy-plugins/blob/main/config/settings.yml#L2491-L2508) to include some rewrite rules.
5. Run SearXNG locally and search the news tab. 

## Author's checklist

TODO:

- [ ] Enable the ability to wash out results by setting `repl` to `false`.

### DIFF
--- a/searx/plugins/url_rewrite.py
+++ b/searx/plugins/url_rewrite.py
@@ -1,0 +1,109 @@
+"""
+The **URL Rewrite plugin** allows modification of search result URLs based on regular expression patterns.
+URLs can be rewritten to different destinations, blocked entirely, or have their priority adjusted in the
+results list.
+
+The plugin can be enabled by adding it to the `enabled_plugins` **list** in the `settings.yml`:
+
+.. code:: yaml
+
+    enabled_plugins:
+        - 'URL Rewrite'
+        ...
+
+Configuration options in `settings.yml`:
+
+- ``url_rewrite.rules``: A **list** of rules that define how URLs should be handled. Each rule
+  can contain the following keys:
+
+  - ``pattern``: Regular expression pattern to match against result URLs
+  - ``repl``: Replacement string for matched URLs, or `false` to remove matching results
+  - ``priority``: Optional priority adjustment ('high' or 'low') for matching results
+
+Example configuration:
+
+.. code:: yaml
+
+    url_rewrite:
+        rules:
+            - pattern: '^(?P<url>https://(.*\.)?cnn\.com/.*)$'
+              repl: 'https://reader.example.com/?url=\g<url>'
+            - pattern: '^https?://(?:www\.)?facebook\.com/.*'
+              repl: false
+            - pattern: '^(?P<url>https://news\.example\.com/.*)$'
+              repl: 'https://proxy.example.com/?url=\g<url>'
+              priority: high
+
+In this example:
+- CNN articles are redirected through a reader service
+- Facebook URLs are removed from results
+- News site URLs are proxied and given higher priority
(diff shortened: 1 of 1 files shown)


Bnyro:
Isn't that already possible by using the [Hostnames plugin](https://docs.searxng.org/src/searx.plugins.hostnames.html)?

joestump:
@Bnyro it's not currently possible to mutate URLs beyond _just_ the hostname with the Hostnames plugin. Hostnames only allows you to replace the hostname of the `parsed_url`. It also doesn't allow you to preserve the original URL for display purposes.

Bnyro:
> @Bnyro it's not currently possible to mutate URLs beyond _just_ the hostname with the Hostnames plugin. Hostnames only allows you to replace the hostname of the `parsed_url`. It also doesn't allow you to preserve the original URL for display purposes.
>
Nevertheless both plugins are quite similar, wouldn't it make sense to modify the existing hostname plugin to support these features instead?

joestump:
@Bnyro I can do that. Went this route per @return42's guidance in the linked discussion. I'll take a look at refactoring.

joestump:
This would, I suspect, necessitate a refactor of the `settings.yml` for that plugin. Are the maintainers comfortable with PRs that break backwards compatibility? Or should I introduce a new "mode" to the plugin?

Bnyro:
> This would, I suspect, necessitate a refactor of the settings.yml for that plugin. Are the maintainers comfortable with PRs that break backwards compatibility? Or should I introduce a new "mode" to the plugin?
>
It might makes sense to add a `replace_url` option as you already did, and another one called `replace_full_url` for not only replacing the regex in the hostname.

I'd rather wait for @return42's comment though, I'm not too sure if we should merge the two plugins when rethinking it.

The use cases for the hostnames plugin and the url rewrite plugin are quite different as it seems, but they share a lot of similar code.

return42:
> I'd rather wait for @return42's comment though, I'm not too sure if we should merge the two plugins when rethinking it.

I lost sight of this PR .. I'm very sorry.

> The use cases for the hostnames plugin and the url rewrite plugin are quite different as it seems, but they share a lot of similar code.

I'm not sure myself, but it feels better to keep the two plugins separate. Even though they both have overlapping features, hostname-replace is something different to me. If we mix them, we'd have to take too much account of what already exists. If we don't mix them, replace-URLs would be able to evolve unhindered until perhaps one day it can completely replace hostname-replace.

PR #4108, opened 2024-12-14
[fix]: fix engine name may break some css style

### LINKED ISSUE #4107: The favicon icons overlap when the returned URL is empty.
Branch: master
Version: 2024.12.13-c60fe999c

Docker

**What happened?**

some favicon will display not correctly

**How To Reproduce**

Search some keywords, ~~if the result no contains url~~, some favicon will display not in right place. Using mobile view is more noticeable. The two results block combine into one view.

**Expected behavior**

**Screenshots & Logs**

|Mobile|Desktop|
|-|-|
|image|image|

**Additional context**

Seems related to this CSS (.right)

|Checked|Unchecked|
|-|-|
|image|image|

hezhijie0327:
## What does this PR do?

If engine name contains some keyword like "right", "left", "center" will break css sytle.

Eg: `right dao` will add `.right` to css
https://github.com/searxng/searxng/blob/c60fe999cf6dd77599992918b0cdb6f6c5df4baf/searx/settings.yml#L1588-L1604

Trim white space, if engine name contains. eg: `right dao` -> `rightdao`

|Before|After|
|-|-|
|image|image|

## Related issues

Closes #4107

### DIFF
--- a/searx/templates/simple/macros.html
+++ b/searx/templates/simple/macros.html
@@ -19,7 +19,7 @@
 
 <!-- Draw result header -->
 {% macro result_header(result, favicons, image_proxify) -%}
-<article class="result {% if result['template'] %}result-{{ result.template|replace('.html', '') }}{% else %}result-default{% endif %} {% if result['category'] %}category-{{ result['category'] }}{% endif %}{% for e in result.engines %} {{ e|replace(' ', '') }}{% endfor %}">
+<article class="result {% if result['template'] %}result-{{ result.template|replace('.html', '') }}{% else %}result-default{% endif %} {% if result['category'] %}category-{{ result['category'] }}{% endif %}{% for e in result.engines %} engine-{{ e|replace(' ', '-') }}{% endfor %}">
   {{- result_open_link(result.url, "url_header") -}}
   {%- if favicon_resolver != "" %}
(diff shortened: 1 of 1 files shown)


vojkovic:
Wow - I noticed this too, what a funny bug, very nice catch!

return42:
Oh yes, what a funny bug / good catch :+1: 

> Trim white space, if engine name contains. eg: right dao -> rightdao

Sadly this will not solve the main issue, engine names can't be used as CSS class names (what when the name of an engine is `right`)

This issue was introduced in commit f0819e88 in the old searx project / the intention was to give a solution to this request: https://github.com/searx/searx/issues/670

I can't remember this feature (*..permit restyling following source engine..*) has ever been used.

I think its better we revert the additions from f0819e88 entirely .. what do you think?

hezhijie0327:
You are correct! `right` still got the issue...

How about like category, add a prefix? `engine-right`, `engine-right-dao`

image

---

BTW, if this feature is not in used, then revert it should no risk.

return42:
> BTW, if this feature is not in used, then revert it should no risk.

Yes, we should remove the loop over the engines in the class attribute :+1: 

----

Its very ugly code .. name conflicts and uncontrolled side effects, as you have found, are inevitable.

hezhijie0327:
removed.

image

return42 (approved):
@hezhijie0327 thanks a lot for your analysis & cleanup :+1:

PR #4115, opened 2024-12-15
[fix] settings: javascript crash when entering due to missing search form

Bnyro:
## What does this PR do?
- this fixes a regression from https://github.com/searxng/searxng/pull/4055

- navigate to the settings and observe the JavaScript logs

## Related issues
see https://github.com/searxng/searxng/pull/4084#issuecomment-2542988190


return42 (approved):
> [It also happens without this PR (tested locally), not sure what causes it though, I'll try to figure it out.](https://github.com/searxng/searxng/pull/4084#issuecomment-2543838265)

I can confirm that the error no longer occurs / thanks for the patch :+1:

PR #4140, opened 2024-12-27
[fix] dockerhub: switch to new api path

### LINKED ISSUE #4126: engine Docker-Hub  does not work / API v1 no longer supported
Docker-Hub API v1 ..

https://github.com/searxng/searxng/blob/19ee529b78bd7840488fd36005270f7ae25e566b/searx/engines/docker_hub.py#L23

does no longer work and the linked API docs no longer available

https://github.com/searxng/searxng/blob/19ee529b78bd7840488fd36005270f7ae25e566b/searx/engines/docker_hub.py#L13

Seems we need a response function like 

```python
base_url = "https://hub.docker.com/"
search_api = base_url + "api/search/v3/catalog/search?"
page_size = 10


def request(query, params):

    args = urlencode(
        {
            "query": query,
            "from": page_size * (params['pageno'] -1),
            "size": page_size,
        }
    )
    params['url'] = search_api + args
    params["headers"]["Search-Version"] = "v3"

    return params
```

.. the url is https://hub.docker.com/api/search/v3/catalog/search?query=searx&from=0&size=10

but this needs a new implementation of the response function and I don't know if there is a better solution ..

https://github.com/searxng/searxng/blob/19ee529b78bd7840488fd36005270f7ae25e566b/searx/engines/docker_hub.py#L34-L35

Bnyro:
## What does this PR do?
- migrate to the new dockerhub search api

- the previous docker hub search api has been removed

- `!dh searx`
- `!dh alpine`

## Related issues
closes #4126

### DIFF
--- a/searx/engines/docker_hub.py
+++ b/searx/engines/docker_hub.py
@@ -19,14 +19,17 @@ about = {
 categories = ['it', 'packages']  # optional
 paging = True
 
-base_url = "https://hub.docker.com/"
-search_url = base_url + "api/content/v1/products/search?{query}&type=image&page_size=25"
+base_url = "https://hub.docker.com"
+page_size = 10
 
 
 def request(query, params):
-
-    params['url'] = search_url.format(query=urlencode(dict(q=query, page=params["pageno"])))
-    params["headers"]["Search-Version"] = "v3"
+    args = {
+        "query": query,
+        "from": page_size * (params['pageno'] - 1),
+        "size": page_size,
+    }
+    params['url'] = f"{base_url}/api/search/v3/catalog/search?{urlencode(args)}"
 
     return params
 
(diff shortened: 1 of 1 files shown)


return42:
FYI: I had to rebase to avoid an old dependency:

    ERROR: Could not find a version that satisfies the requirement msgspec-python313-pre (from versions: none)
    ERROR: No matching distribution found for msgspec-python313-pre

Related:

- https://github.com/searxng/searxng/pull/4142
- https://github.com/jcrist/msgspec/issues/764#issuecomment-2564999865

searx/engines/docker_hub.py
```diff
@@ -19,14 +19,17 @@
 categories = ['it', 'packages']  # optional
 paging = True
 
-base_url = "https://hub.docker.com/"
-search_url = base_url + "api/content/v1/products/search?{query}&type=image&page_size=25"
+base_url = "https://hub.docker.com"
+page_size = 10
```
return42, line 23: removing the slash here, needs to add the slash there ..

```diff
diff --git a/searx/engines/docker_hub.py b/searx/engines/docker_hub.py
index 1aa25e39f..021c05827 100644
--- a/searx/engines/docker_hub.py
+++ b/searx/engines/docker_hub.py
@@ -56,7 +56,7 @@ def response(resp):
 
         result = {
             'template': 'packages.html',
-            'url': base_url + ("_/" if is_official else "r/") + item.get("slug", ""),
+            'url': base_url + ("/_/" if is_official else "/r/") + item.get("slug", ""),
             'title': item.get("name"),
             'content': item.get("short_description"),
             'thumbnail': item["logo_url"].get("large") or item["logo_url"].get("small"),
```

searx/engines/docker_hub.py
```diff
@@ -19,14 +19,17 @@
...
 def request(query, params):
-
-    params['url'] = search_url.format(query=urlencode(dict(q=query, page=params["pageno"])))
-    params["headers"]["Search-Version"] = "v3"
+    args = {
+        "query": query,
+        "from": page_size * (params['pageno'] - 1),
+        "size": page_size,
+    }
+    params['url'] = f"{base_url}/api/search/v3/catalog/search?{urlencode(args)}"
```
return42, line 32: The link from line 3 ..

```
"official_api_documentation": 'https://docs.docker.com/registry/spec/api/',
```
.. seems no longer valid .. from where did you got the `/api/search/v3/catalog/search`?

return42 (changes requested):
Thanks a lot for fixing docker hub engine .. I have left comments in the source code below ..

searx/engines/docker_hub.py
Bnyro: True, I changed the `base_url` at the end of my changes but forgot to change the links, good catch :+1:

searx/engines/docker_hub.py
Bnyro: Via their Web UI, just as you probably, I can't seem to find any documentation about the catalog API.

Nevertheless it might make sense to keep the documentation link there in case they update their documentation? Their website claims that the API is still in beta, so that might be why it's not yet fully documented.

return42 (approved):

PR #4141, opened 2024-12-27
[enh] add pyproject.toml listing build requirements

0xhtml:
## What does this PR do?

This PR adds a pyproject.toml with the basic requirements of setuptools and pyyaml required to run the current setup.py.

To run the setup.py pyyaml needs to already be installed. PEP 518 'Specifying Minimum Build System Requirements for Python Projects' introduced the pyproject.toml file to specify the required dependencies in a format that build tooling like pip can parse prior to executing the setup.py.

Test e.g., `pip install .` in a clean environment. Prior to this PR pip would fail with a subprocess error of `ModuleNotFoundError: No module named 'yaml'` with this PR pip installs searxng nicely.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "pyyaml"]


return42:
> Test e.g., `pip install .` in a clean environment. 

I understand the intention, but I am unsure whether this is a good idea.  SearXNG is not a Python package (!) that users can install in their Python installation / it will never be offered as a package on pypi.  

As a rolling release it will only ever be installed in developer mode (`pip install -e .` in a virtual env) --> I want to avoid inexperienced users starting to install SearXNG into their Python environment thinking it is a Python package and then inevitably running into problems.

We can only provide support for the installation methods described under in the chapter [Installation](https://docs.searxng.org/admin/installation.html)

I mark this PR as DRAFT.

0xhtml:
I understand that installation through pip with e.g., `pip install searxng@git+https://github.com/searxng/searxng` is unsupported and users should be discouraged from / warned about using it. I personally don't believe that a missing dependency is the best way to accomplish that goal, though my opinion may not be your concern.

Furthermore, I have to add that even the step by step installation explicitly contains `pip install -U pyyaml` and `pip install  --use-pep517 --no-build-isolation -e .` to circumvent the missing build dependency. With a proper `pyproject.toml` installing pyyaml and the extra flags to pip could be dropped.

I would very much prefer an explicit warning or error when running the setup.py through pip aka installing searx through pip.

return42:
> I personally don't believe that a missing dependency is the best way to accomplish that goal,

Surely true. The short answer: SearXNG is not ready yet / or at least I'm not ready yet.

In the long term, I would like to develop SearXNG so that it is no longer an application but a library that can also be installed via pypi.   But we are still a long way from that.

If we introduce a pyproject.toml, then this should also simplify the complexity of our environment; we are currently still using various shell scripts to build up a developer environment.  So it is not enough to create a pyproject.toml, the entire developer environment must be migrated from the shell scripts to a development environment such as [hatch](https://github.com/pypa/hatch).  @dalf implemented a prototype based on [rye](https://rye.astral.sh/)

- https://github.com/searxng/searxng/pull/3266

Personally, however, I think we would be better off relying on [hatch](https://hatch.pypa.io/latest/). 

For such decisions some experience is needed which we (I) still have to build up.  Quite apart from the fact that the migration is also a big task ... (take in mind, it will also touch the installation methods we support today).

However, we still have other major tasks, such as the typing of the engines and the [result types](https://github.com/searxng/searxng/pull/4183), which are also a prerequisite for turning SearXNG into a library ..

> To run the setup.py pyyaml needs to already be installed

This is currently our smallest problem for which we have a solution (shell script and `make`) ... once we have migrated the developer environment to a python framework like hutch or rye, the problem will be solved anyway.

I also want to avoid contributors starting to put something in pyproject.toml that we (also) need in the shell scripts ... and then we have duplicate implementations for the same task ... (this PR adds something that we already have in the shell scripts, when we need one requirement more in the developer environment we have to add it in two places from now on).

I just ask for your understanding that we don't  start things halfway without having a concept for the complete solution ... this approach of adding something “here and there” makes the project error-prone and the development more and more difficult.

We have already built in too many “quick fixes” in the past ... the dismantling later is all the more painful ... we should better leave it, even if we can't satisfy everyone for the moment.

PR #4149, opened 2024-12-29
[fix] only show and clear known cookies

### LINKED ISSUE #3763: Preferences show and overwrite unknown cookies

GenericMale:
## What does this PR do?

This PR modifies the `clear_cookies` function to now only delete cookies related to user preferences. It also adjusts the cookie display logic to only show these relevant preference cookies. This ensures that session cookies and other non-preference cookies are preserved, preventing unintended side effects (see #3763).

Previously, the `clear_cookies` function was deleting *all* cookies, including cookies from other services on the same domain, session cookies etc.

1.  **Navigate to the preferences page.** 
2.  **Set and save some user preferences.**
3.  **Navigate to the cookies tab.**
4.  **Verify all searxng cookies are shown but no other cookies.** Foreign cookies can be simulated using Javascript, e.g. `document.cookie = "username=John Doe"` in the Browser console
5.  **Click the "Reset defaults" button.**
6.  **Verify all searxng cookies are deleted but other cookies are preserved.**

## Related issues

Closes #3763

### DIFF
--- a/searx/templates/simple/preferences/cookies.html
+++ b/searx/templates/simple/preferences/cookies.html
@@ -10,10 +10,12 @@
       <th>{{ _('Value') }}</th>{{- '' -}}
     </tr>
     {%- for cookie in cookies -%}
-      <tr>{{- '' -}}
-        <td>{{ cookie }}</td>{{- '' -}}
-        <td>{{ cookies[cookie] }}</td>{{- '' -}}
-      </tr>
+      {% if cookie in preferences.key_value_settings %}
+        <tr>{{- '' -}}
+          <td>{{ cookie }}</td>{{- '' -}}
+          <td>{{ cookies[cookie] }}</td>{{- '' -}}
+        </tr>
+      {% endif %}
     {%- endfor -%}
   </table>
 {%- else -%}
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1289,7 +1289,8 @@ def favicon():
 def clear_cookies():
     resp = make_response(redirect(url_for('index', _external=True)))
     for cookie_name in request.cookies:
-        resp.delete_cookie(cookie_name)
+        if cookie_name in request.preferences.key_value_settings:
+            resp.delete_cookie(cookie_name)
     return resp
 
 


return42:
I still hesitate to merge this PR since I have a few doubts:

Name collisions with other apps on the same domain are not solved (e.g. `language` and `theme` are very common) ~and problematic code mentioned by @glanham-jr in https://github.com/searxng/searxng/issues/3763#issuecomment-2408775920 is not addressed~ -> I think the preference code needs a complete review .. meanings?

GenericMale:
> I still hesitate to merge this PR since I have a few doubts:
> 
> Name collisions with other apps on the same domain are not solved (e.g. `language` and `theme` are very common) and problematic code mentioned by @glanham-jr in [#3763 (comment)](https://github.com/searxng/searxng/issues/3763#issuecomment-2408775920) is not addressed -> I think the preference code needs a complete review .. meanings?

hasn't the mentioned "problematic" code been removed in #3966 or is this something else?

in my opinion the very generic names of the cookies used by SearXNG is a related but different issue which will obviously require a lot more work  (and some sort of migration) than this small improvement which (hopefully) has minimal impact but ensures SearXNG will not touch any cookies it doesn't know about at all (which imho is required in any case).

searx/templates/simple/preferences/cookies.html
```diff
@@ -10,10 +10,12 @@
...
     </tr>
     {%- for cookie in cookies -%}
-      <tr>{{- '' -}}
-        <td>{{ cookie }}</td>{{- '' -}}
-        <td>{{ cookies[cookie] }}</td>{{- '' -}}
-      </tr>
+      {% if cookie in preferences.key_value_settings %}
+        <tr>{{- '' -}}
+          <td>{{ cookie }}</td>{{- '' -}}
+          <td>{{ cookies[cookie] }}</td>{{- '' -}}
```
glanham-jr, line 16: This may result in a key not found exception, now that we are enumerating a different object. Consider adding a "if key in cookies" check.

GenericMale: sorry, i can't follow. are you talking about the cookies[cookie]?
i haven't changed whats being enumerated and only added the if to filter the cookie names.

glanham-jr: I misread, yes this is good.

PR #4160, opened 2025-01-05
[Wallhaven] Fix usage of `api_key` engine setting

Lucki:
## What does this PR do?

It ensures the `api_key` engine setting is actually used somewhere by setting the `X_API_Key` header.

> [Users can authenticate by including their API key either in a request URL by appending `?apikey=<API KEY>`, or by including the `X-API-Key: <API KEY>` header with the request.](https://wallhaven.cc/help/api)

Currently, the value of `params['api_key']` isn't read anywhere, making the presence of the `api_key` setting rather pointless.

1. Add the engine settings `api_key` to your `settings.yml`.
    ~~~ yaml
    engines:
      - name: wallhaven
        api_key: asd123asd
    ~~~
1. Search something using the engine: `!wh test`
2. Observe the request headers and find the `X-API-Key` header set to the settings value
    OR
    Search for images with safe mode off, the key has to be valid for this.

### DIFF
--- a/searx/engines/wallhaven.py
+++ b/searx/engines/wallhaven.py
@@ -57,7 +57,7 @@ def request(query, params):
     }
 
     if api_key:
-        params['api_key'] = api_key
+        params['headers']['X-API-Key'] = api_key
 
     params['url'] = f"{base_url}/api/v1/search?{urlencode(args)}"
     return params


return42:
> Currently, the value of `params['api_key']` isn't read anywhere, making the presence of the `api_key` setting rather pointless.

I think it was an typo from #2818 and the fix would be:

```diff
diff --git a/searx/engines/wallhaven.py b/searx/engines/wallhaven.py
index 127873d5d..b5fee23e4 100644
--- a/searx/engines/wallhaven.py
+++ b/searx/engines/wallhaven.py
@@ -57,7 +57,7 @@ def request(query, params):
     }
 
     if api_key:
-        params['api_key'] = api_key
+        args['api_key'] = api_key
 
     params['url'] = f"{base_url}/api/v1/search?{urlencode(args)}"
     return params
```

@Bnyro should we switch to HTTP headers like this PR does or should we fix the URL arguments?

... having the key in the header could protect it from appearing unintentionally in log files.

Bnyro:
> I think it was an typo from #2818 and the fix would be:
> 
> ```diff
> diff --git a/searx/engines/wallhaven.py b/searx/engines/wallhaven.py
> index 127873d5d..b5fee23e4 100644
> --- a/searx/engines/wallhaven.py
> +++ b/searx/engines/wallhaven.py
> @@ -57,7 +57,7 @@ def request(query, params):
>      }
>  
>      if api_key:
> -        params['api_key'] = api_key
> +        args['api_key'] = api_key
>  
>      params['url'] = f"{base_url}/api/v1/search?{urlencode(args)}"
>      return params
> ```
> 
Indeed, good catch 👍
> @Bnyro should we switch to HTTP headers like this PR does or should we fix the URL arguments?
> 
> ... having the key in the header could protect it from appearing unintentionally in log files.
>
I agree, using it as header is probably the more secure way for authentication generally, and if it's tested and works, that should be all good.

Bnyro (approved):

Lucki:
Works after pulling the new version with `podman auto-update` :+1:

PR #4170, opened 2025-01-09
[feat] Added shell.nix (Nix development support)

Denperidge:
## What does this PR do?

Add a `shell.nix` file. This is *not* for use in the Nix version of searchx, but instead allows developers to get a compatible development environment in one command. 

Some developers are NixOS based, which cannot run arbitrary executables (e.g. the nvm-based node) by default. The nix-shell wraps it in an FHS environment, which fixes that thanks to Nix magic I do not understand.

On a device with nix installed, navigate to the repository and run `make test`. This will probably not work. Then, run `nix-shell` and try `make test` again. It should work!

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -10,7 +10,7 @@ Done!
 (pkgs.buildFHSEnv {
   name = "searxng";
   multiPkgs = pkgs: (with pkgs; [
-    bash
+    bashInteractive
     wget
     gnumake
     git
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,23 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+/*
+nix shell for development usage
+1. run `nix-shell`
+2. run `make run` (or any other make scripts)
+Done!
+*/
+
+(pkgs.buildFHSEnv {
+  name = "searxng";
+  multiPkgs = pkgs: (with pkgs; [
+    bash
+    wget
+    gnumake
+    git
+    python3
+    geckodriver
+    shellcheck
+  ]);
+  runScript = "bash";
+}).env
+


return42:
@SuperSandro2000 can you review this PR? .. thanks :+1:

shell.nix
```diff
@@ -0,0 +1,23 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+/*
+nix shell for development usage
+1. run `nix-shell`
+2. run `make run` (or any other make scripts)
+Done!
+*/
+
+(pkgs.buildFHSEnv {
```
SuperSandro2000, line 10: Why didn't you use mkShell?

shell.nix
```diff
@@ -0,0 +1,23 @@
...
+  name = "searxng";
+  multiPkgs = pkgs: (with pkgs; [
+    bash
```
SuperSandro2000:
```suggestion
    bashInteractive
```

shell.nix
Denperidge: Didn't let me run the node nvm stuff from the makefile otherwise! I'm on NixOS 24.11 but in case this was a my-config-issue then we can switch it

SuperSandro2000: Do you still have the error?

SuperSandro2000: Why are you using nvm? NixOS has [nodejs_23](https://search.nixos.org/packages?channel=24.11&from=0&size=50&sort=relevance&type=packages&query=npm) and we should probably use that because then we avoid issues with certs and don't need any kind of patching.

inetol:
I'm afraid we'll have to close this PR. None of the maintainers actively use Nix tooling, and we won't be able to keep this file in sync over time.

PR #4179, opened 2025-01-12
[json_engine] Documentation and overhaul

Lucki:
## What does this PR do?

This PR overhauls the JSON engine, bringing it on par to the XPath engine with (almost) the same functionality.
I am not entirely sure what the caching option in the XPath engine is doing, so I ignored it.

Every step is (or should be) cleanly separated in own commits:
* First, the existing documentation is properly added to the sphinx generator.
* After that, the functionality from the XPath engine is mirrored to the JSON engine.
* I then added unit tests to catch errors, again, mirroring the XPath engine unit tests.
* ~The next commit might be a bit opinionated, but I exposed `publishedDate` from within the settings. The reason is that it's available in every single template. I also exposed the `template` parameter setting, allowing to choose a different template style which might be better fitting for the content.~
  Dropped it because of different python versions failing the tests.
* Now `make test` was complaining with `R0912 (too-many-branches)` which I tackled in the last commit. This combines the two possible branches (with and without a `request_query`) into a function.

If any of the commits don't appeal to you, they can be easily dropped.

I played around, looking how to implement a JSON engine using the settings, but the information and functionality was rather lacking.
Crucially, the information about the `request_body` having to double the curly braces and the handling of arrays and paths wasn't easily available.

Run 
* `make test`
* `make test.unit`
* `make docs.live` and visit http://0.0.0.0:8000/dev/engines/json_engine.html
* a search that uses the `json_engine`

### DIFF
--- a/docs/dev/engines/index.rst
+++ b/docs/dev/engines/index.rst
@@ -45,6 +45,7 @@ Online Engines
    demo/demo_online
    xpath
    mediawiki
+   json_engine
 
 .. toctree::
    :maxdepth: 1
--- a/docs/dev/engines/json_engine.rst
+++ b/docs/dev/engines/json_engine.rst
@@ -1,0 +1,13 @@
+.. _json_engine engine:
+
+============
+JSON Engine
+============
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.engines.json_engine
+  :members:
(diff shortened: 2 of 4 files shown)


return42 (approved):
Perfect work :+1:  .. and doc will help engine developer a lot .. thanks!

-----

> I am not entirely sure what the caching option in the XPath engine is doing, so I ignored it.

Yes, lets ignore .. was added in PR https://github.com/searx/searx/pull/565 opened back in 2016 .. (and finally merged in 2020) from the commit message of c3daa0853:

> Xpath engine and results template changed to account for the fact that
archive.org doesn't cache .onions, though some onion engines migth have their own cache.

PR #4180, opened 2025-01-12
Public domain image archive

### LINKED ISSUE #4167: New engine: Public Domain Image Archive
**Working URL to the engine**

https://pdimagearchive.org/about/

The Public Domain Image Archive (PDIA) — brought to you by [The Public Domain Review](https://publicdomainreview.org/) (PDR) — is a curated collection of more than 10,000 out-of-copyright historical images, free for all to explore and reuse.

**Features of this engine**

Not sure how we can make use of 

grafik

grafik

images

Denperidge:
## What does this PR do?

Adds the public domain image archive engine & settings.yml entry. It is based on the unsplash engine!

This would close #4167! Also, public domain image sources are incredible to have

`make run`, enable public domain image archive, search `!pdia dog`, check the next page of results

## Author's checklist

I don’t know whether it should be default enabled or disabled, so feel free to switch that (or let me know if I should change it)

## Related issues

Closes #4167

### DIFF
--- a/searx/engines/public_domain_image_archive.py
+++ b/searx/engines/public_domain_image_archive.py
@@ -1,15 +1,12 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Public domain image archive, based on the unsplash engine
-
-Meow meow
-
-"""
+"""Public domain image archive"""
 
 from urllib.parse import urlencode, urlparse, urlunparse, parse_qsl
 from json import dumps
 
-algolia_api_key = "153d2a10ce67a0be5484de130a132050"
-"""Algolia API key. See engine documentation """
+from searx.network import get
+from searx.utils import extr
+from searx.exceptions import SearxEngineAccessDeniedException, SearxEngineException
 
 THUMBNAIL_SUFFIX = "?fit=max&h=360&w=360"
 """
@@ -42,23 +39,58 @@ about = {
     "results": 'JSON',
 }
 
-base_url = 'https://oqi2j6v4iz-dsn.algolia.net/'
-search_url = base_url + f'1/indexes/*/queries?x-algolia-api-key={algolia_api_key}&x-algolia-application-id=OQI2J6V4IZ'
+base_url = 'https://oqi2j6v4iz-dsn.algolia.net'
+pdia_config_url = 'https://pdimagearchive.org/_astro/config.BiNvrvzG.js'
 categories = ['images']
 page_size = 20
 paging = True
 
 
-def clean_url(url):
+__CACHED_API_KEY = None
+
+
+def _clean_url(url):
     parsed = urlparse(url)
     query = [(k, v) for (k, v) in parse_qsl(parsed.query) if k not in ['ixid', 's']]
 
     return urlunparse((parsed.scheme, parsed.netloc, parsed.path, parsed.params, urlencode(query), parsed.fragment))
 
 
+def _get_algolia_api_key():
+    global __CACHED_API_KEY  # pylint:disable=global-statement
+
+    if __CACHED_API_KEY:
+        return __CACHED_API_KEY
+
+    resp = get(pdia_config_url)
+    if resp.status_code != 200:
+        raise LookupError("Failed to obtain Algolia API key for PDImageArchive")
+
+    api_key = extr(resp.text, 'r="', '"', default=None)
+
+    if api_key is None:
+        raise LookupError("Couldn't obtain Algolia API key for PDImageArchive")
+
+    __CACHED_API_KEY = api_key
+    return api_key
+
+
+def _clear_cached_api_key():
+    global __CACHED_API_KEY  # pylint:disable=global-statement
+
+    __CACHED_API_KEY = None
+
+
 def request(query, params):
-    params['url'] = search_url
+    api_key = _get_algolia_api_key()
+
+    args = {
+        'x-algolia-api-key': api_key,
+        'x-algolia-application-id': 'OQI2J6V4IZ',
+    }
+    params['url'] = f"{base_url}/1/indexes/*/queries?{urlencode(args)}"
     params["method"] = "POST"
+
     request_params = {
         "page": params["pageno"] - 1,
         "query": query,
@@ -71,13 +103,22 @@ def request(query, params):
         ]
     }
     params["data"] = dumps(data)
-    logger.debug("query_url --> %s", params['url'])
+
+    # http errors are handled manually to be able to reset the api key
+    params['raise_for_httperror'] = False
     return params
 
 
 def response(resp):
     results = []
     json_data = resp.json()
+
+    if resp.status_code == 403:
+        _clear_cached_api_key()
+        raise SearxEngineAccessDeniedException()
+
+    if resp.status_code != 200:
+        raise SearxEngineException()
 
     if 'results' not in json_data:
         return []
@@ -97,9 +138,9 @@ def response(resp):
         results.append(
             {
                 'template': 'images.html',
-                'url': clean_url(f"{about['website']}/images/{result['objectID']}"),
-                'img_src': clean_url(base_image_url),
-                'thumbnail_src': clean_url(base_image_url + THUMBNAIL_SUFFIX),
+                'url': _clean_url(f"{about['website']}/images/{result['objectID']}"),
+                'img_src': _clean_url(base_image_url),
+                'thumbnail_src': _clean_url(base_image_url + THUMBNAIL_SUFFIX),
                 'title': f"{result['title'].strip()} by {result['artist']} {result.get('displayYear', '')}",
                 'content': content,
             }
--- a/searx/engines/public_domain_image_archive.py
+++ b/searx/engines/public_domain_image_archive.py
@@ -1,0 +1,108 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Public domain image archive, based on the unsplash engine
+
+Meow meow
+
+"""
+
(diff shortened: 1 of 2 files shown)


searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+    "website": 'https://pdimagearchive.org',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+base_url = 'https://oqi2j6v4iz-dsn.algolia.net/'
+search_url = (
+    base_url
+    + '1/indexes/*/queries?x-algolia-api-key=153d2a10ce67a0be5484de130a132050&x-algolia-application-id=OQI2J6V4IZ'
```
Bnyro: That's basically a hardcoded API key, we should probably make that configurable, since that key probably differs per user?

Denperidge: I've only confirmed using private windows, but I think the x-algolia-api-key might actually just be hardcoded like that in the original website.

I can make it configurable either way, but would you mind checking if going to gives the same API key for you when capturing the request when searching at https://pdimagearchive.org/search/?q=dog?
A screenshot of the Firefox dev console. The network tab shows the selected queries request, highlighting the x-algolia-api-key in the parameters below

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+    return urlunparse((parsed.scheme, parsed.netloc, parsed.path, parsed.params, urlencode(query), parsed.fragment))
+
+
+def request(query, params):
+    params['url'] = search_url
+    params["method"] = "POST"
+    params["data"] = (
+        """{"requests":[{"indexName":"prod_all-images","params":"highlightPostTag=__%2Fais-highlight__&highlightPreTag=__ais-highlight__&"""
+    )
+    params["data"] += f'page={params["pageno"] - 1}&query={query}' + '"}]}'
```
Bnyro:
```suggestion
    params = {
        "page": params["pageno"] - 1,
        "query": query,
        "highlightPostTag":  "__ais-highlight__",
        "highlightPreTag": "__ais-highlight__",
    }
    data = {
        "requests": [
            {"indexName": "prod_all-images", "params": urlencode(params),
        ]
    }
    params["data"] = dumps(data)
```

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+    logger.debug("query_url --> %s", params['url'])
+    return params
+
+
+def response(resp):
+    results = []
+    json_data = loads(resp.text)
```
Bnyro:
```suggestion
    json_data = resp.json()
```

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+
+    if 'results' in json_data:
```
Bnyro:
```suggestion
    if 'results' not in json_data:
        return []
```

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+        for result in json_data['results'][0]['hits']:
+            content = ""
+            if "themes" in result:
+                content += "Themes: " + result['themes']
+            if "encompassingWork" in result:
+                if content != "":
+                    content += "\n"
+                content += "Encompassing work: " + result['encompassingWork']
```
Bnyro:
```suggestion
            content = []
            if "themes" in result:
                content.append("Themes: " + result['themes'])
            if "encompassingWork" in result:
                content.append("Encompassing work: " + result['encompassingWork'])
            content = "\n".join(content)
```

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+
+            base_image_url = result['thumbnail'].split("?")[0] if "?" in result['thumbnail'] else result['thumbnail']
```
Bnyro:
```suggestion
            base_image_url = result['thumbnail'].split("?")[0]
```
does exactly the same

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+            
+            results.append(
+                {
+                    'template': 'images.html',
+                    'url': clean_url(f"{about['website']}/images/{result['objectID']}"),
+                    'img_src': clean_url(base_image_url),
+                    'thumbnail_src': clean_url(base_image_url + THUMBNAIL_SUFFIX),
+                    'title': f"{result['title'].strip()} by {result['artist']} {result.get('displayYear') or ''}",
+                    'content': f"Themes: {result['themes']}",
```
Bnyro:
```suggestion
                    'content': content,
```
Otherwise creating the `content` variable wouldn't make any sense

searx/engines/public_domain_image_archive.py
```diff
@@ -0,0 +1,86 @@
...
+
+            base_image_url = result['thumbnail'].split("?")[0] if "?" in result['thumbnail'] else result['thumbnail']
+            
+            results.append(
+                {
+                    'template': 'images.html',
+                    'url': clean_url(f"{about['website']}/images/{result['objectID']}"),
+                    'img_src': clean_url(base_image_url),
+                    'thumbnail_src': clean_url(base_image_url + THUMBNAIL_SUFFIX),
+                    'title': f"{result['title'].strip()} by {result['artist']} {result.get('displayYear') or ''}",
```
Bnyro:
```suggestion
                    'title': f"{result['title'].strip()} by {result['artist']} {result.get('displayYear', '')}",
```

searx/engines/public_domain_image_archive.py
Bnyro: You're right, same API key for me, that's funny.

We can't be sure if it'll always stay the same or change at some point, so putting it into an `api_key` variable won't hurt :)

searx/engines/public_domain_image_archive.py
Denperidge: I didn't trust in Python's user-friendliness, that is on me!

searx/engines/public_domain_image_archive.py
Denperidge: The embarassment I'm currently feeling. That's why the encompassing work wasn't showing. I need to sleep more.

Bnyro:
The API key changed to 23b97dfd03aaa998c7055f832d4022cf today, looks like we need to figure out where it's coming from.

Bnyro:
The API key is set in https://pdimagearchive.org/_astro/config.BiNvrvzG.js, should be easy enough to extract.

Bnyro (commented):
The API key is set in https://pdimagearchive.org/_astro/config.BiNvrvzG.js, should be easy enough to extract. It's possible that the file name changes every day too, so we should first fetch https://pdimagearchive.org/, check which linked javascript element includes `config` in its name and extract the api code from there. Afterwards we should cache the api key via redis.

I would have pushed the necessary changes if you had 

enabled, see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork.

Denperidge:
I want to enable the option 😭 somethings awry that im not getting the option

Denperidge:
Found the issue: https://github.com/orgs/community/discussions/5634 . Let me see to shuffle some things

Denperidge:
@Bnyro enabled!

Bnyro (approved):
I've added automation to obtain the api key in https://github.com/searxng/searxng/pull/4180/commits/2ebe89bf717435cc95361ff045d37bf691983073 and rebased the PR.

The URL to the JavaScript file containing the API key hasn't changed the past week, so I guess it should be fine to hardcode it as it apparently doesn't change.

Looks good to me now!

return42:
@Denperidge any reason why you have two identities on github (@Denperidge-Contrib)? I'm (and my tools) a bit confused .. further you forked searxng with a different name `searxng-contrib` while you also have https://github.com/Denperidge-Contrib .. `Denperidge/public-domain-image-archive` and `Denperidge-Contrib/public-domain-image-archive`.

Not needed for this PR but for the future some general rules: 

- It is not good to have more than one identity
- forks of repos should have the same name as the origin .. 

these basic rules make it much easier to collaborate, and the tools like git (or gihub) work better if these rules are met.

Denperidge:
> @Denperidge any reason why you have two identities on github (@Denperidge-Contrib)? I'm (and my tools) a bit confused .. further you forked searxng with a different name `searxng-contrib` while you also have https://github.com/Denperidge-Contrib ..
> 
> Not needed for this PR but for the future some general rules:
> 
> * It is not good to have more than one identity
> * forks of repos should have the same name as the origin ..
> 
> these basic rules make it much easier to collaborate, and the tools like git (or gihub) work better if these rules are met.

Very fair question! I created the org for my contributions, because I prefer to keep my repositories overview cleaner. I’m not super fond of the idea of having a repo named Denperidge/searxng like the fork is meant to be a functional; while the truth is that I simply wish to contribute upstream. It artificially inflates my number of repositories too! Hence the contrib- addition for searxng, as I couldn’t keep it in Denperidge-Contrib due to GitHub not allowing edits from maintainers in org-owned repos.

I’ll keep in mind what you said though; maybe once I fix up my website I won’t have need for GitHub repo view to give me an overview of my work anymore, and I can move every contribution fork to just Denperidge/originalname

(Yes, GitHub has an is-fork filter for repo search, but there are also things I’ve forked that I do use as a de facto version. Things get messy)

return42 (approved):
Thanks a lot for this engine implementation :+1:

Issue #4189, opened 2025-01-15
add Fediverse/Mastodon in Wikidata side panel


Popolon:
**Is your feature request related to a problem? Please describe.**

There are currently several social media links in side panel, but in at least some cases, not the mastodon (or more generally fediverse instance).

This can already be described in:
- [Property:P4033 (Mastodon address)](https://www.wikidata.org/wiki/Property:P4033)
- [Property:P12622 (PeerTube address)](https://www.wikidata.org/wiki/Property:P12622)

As example for testing, there is the [property:P1855 (Wikidata property example)](https://www.wikidata.org/wiki/Property:P1855) in the two properties

Another interesting property in the same subject is [Property:P553 (website account on)](https://www.wikidata.org/wiki/Property:P553) that contain a list of service, and for each service, at least:
- An element (Qxxxx) of the corresponding social network
- [Property:P554 (website username or ID)](https://www.wikidata.org/wiki/Property:P554) as address.
[Property:P1659 (related property)](https://www.wikidata.org/wiki/Property:P553) of Property:P553 definition, contain a list of possible values.

**Describe the solution you'd like**
Test if these fields exist and then display them.

**Describe alternatives you've considered**
Add other Fediverse medias 

**Additional context**

return42:
Closed by https://github.com/searxng/searxng/pull/4209

PR #4191, opened 2025-01-15
add mastodon and peertube accounts from wikidata search

Popolon:
## What does this PR do?

It add in wikidata side box links to mastodon and peertube accounts of users if they are defined. Resolve the first part of #4189

get the following properties:
- [Property:P4033 (Mastodon address)](https://www.wikidata.org/wiki/Property:P4033)
- [Property:P12622 (PeerTube address)](https://www.wikidata.org/wiki/Property:P12622)

and add links using https://fedirect.toolforge.org/?id=$1 resolver, as wikidata does.

Several authors use Fediverse as social media, they should be available in searxNG.

See  #4189

Install searxNG as usual, search for some people with Mastodon account. I searched "David Revoy", that has both Mastodon and Peertube accounts in wikidata.


Bnyro:
The IDs are generally in the format `thelinuxexperiment_channel@tilvids.com` for PeerTube and `ccc@social.bau-ha.us` for Mastodon Links. We could actually just take these links, split them at the '@' and rewrite to `https://tilvids.com/@thelinuxexperiment_channel` or `https://social.bau-ha.us/@ccc`.

Otherwise toolforge.org can track / count clicks of that redirect link, and since SearXNG is a privacy-first search engine we probably don't want that :)

Bnyro (commented):
If you're already on that, would you mind adding support for Lemmy community IDs too? Otherwise I'd open a follow-up PR for that.

They have the property `P11947`, see https://www.wikidata.org/wiki/Property:P11947 and are used by Android's wikidata entry for example, see https://www.wikidata.org/wiki/Q94#P11947.

Popolon:
Ooops, I don't have the habbit  of pull request on github, sorry, I believe I muse open a new pull request? I believe I found where to push the hook, not sure it will be the most elegant way

Bnyro:
> Ooops, I don't have the habbit of pull request on github, sorry, I believe I muse open a new pull request?
>
No, you could have restored that branch and re-opened the pull request via GitHub's UI.

PR #4198, opened 2025-01-16
[mod] add swipe events to the image gallery (gesture control)

### LINKED ISSUE #4151: Improved Mobile Image Viewer
Not Really, more like an annoyance.
Im daily driving my Instance and I also use the image search a lot on my phone, and when trying to page the images, when one is focussed, I allways miss the paging buttons, because they are so small and ergonomicly annoyingly placed.
The Other reason is because they overlap with the image on some images

On Mobile the Theme should:
- hide the forward and backwards buttons
- either
  - allow swiping left or right on the image for paging
  - allowing pressing the left or right side of the image for paging
- disable the click on image, which redirects to  just the image

allot bigger buttons for mobile paging

**Additional context**

browser devtools
mobile image viewer with ergonomic zone overlay

reality:
Screenshot_20241230-130306_Fennec

return42:
## What does this PR do?

Adds [1] to the searxng.min.js and horizontal swipe events to the image gallery.

[1] https://www.npmjs.com/package/swiped-events

## Author's checklist

There is a critical issue https://github.com/john-doherty/swiped-events/issues/38 .. may we copy the [JS script](https://github.com/john-doherty/swiped-events/blob/master/src/swiped-events.js) under the SearXNG umbrella .. but for this POC its OK to use the last release.

If needed, we can [configure](https://github.com/john-doherty/swiped-events/tree/master#configure) how swiped-events works.

## Related issues

Closes: https://github.com/searxng/searxng/issues/4151


return42:
@snoweuph @WreckingBANG you can test gesture control from this PR [on my instance](https://darmarit.org/searx/search?q=!images+superman+lover+johnny+guitar+watson).

snoweuph:
> @snoweuph @WreckingBANG you can test gesture control from this PR [on my instance](https://darmarit.org/searx/search?q=!images+superman+lover+johnny+guitar+watson).

Swipe works like a charm.

Also the close button on your intance seems to work perfectly even when the image is partially under it :)

return42:
One drawback I see when swiping: compared to thumb previews, the gallery view loads the full images .. this can be big images .. when a user swipes in the gallery view (which is the new ui model on small devices) every "big" image is loaded.

**Swiping is very resource-intensive and "expensive" when a user swipes over a series of images very quickly!**

Perhaps we could delay the loading of the full size images by 3-5 seconds ... if the user has already sipped further, then not every image would be loaded in its maximum size.

What suggestions do you have so that we don't load full-size images that the user doesn't even look at?

Bnyro (commented):
> Perhaps we could delay the loading of the full size images by 3-5 seconds ... if the user has already sipped further, then not every image would be loaded in its maximum size.
>
When users swipe fast, we create a new `imgLoader` for each new image the user swiped to.
This means, even when the user swiped to a new image, the loading of the previous image (that'd no longer be needed in theory) continues in the background, because we always create new image loaders but never destroy or cancel them (if I read the code correctly), ref https://github.com/searxng/searxng/pull/4198/files#diff-886bad7a631bc7afd8895a66a1406281cf3b69927c19ce5c6e80480e6517260aR71-R78.

In my opinion, it would make sense here to refactor the code to only use one image loader and cancel the loading when the user swipes to the next image, to avoid loading images the user swiped away after seeing the thumbnail.

A short delay (maybe 1 or 2 seconds) before loading the image make sense in my opinion, but it shouldn't be more than 2 because otherwise at least I would personally feel discomfy with waiting so long.

Bnyro:
Very cool change btw :+1:

WreckingBANG:
> @snoweuph @WreckingBANG you can test gesture control from this PR [on my instance](https://darmarit.org/searx/search?q=!images+superman+lover+johnny+guitar+watson).

This is also really cool but not what i meant. I wrote it a bit misleading. I meant when you do a back press (or swipe back) on your phone. So basically when onbackpressed gets called. The behavior right now is that a back press sends you back to the previous page, so for example to the results page. But the normal expected behavior from other search engines is that you just leave the image view and are back to the overview of all images.

I hope this is more understandable then my previous bad comment.

snoweuph:
> > @snoweuph @WreckingBANG you can test gesture control from this PR [on my instance](https://darmarit.org/searx/search?q=!images+superman+lover+johnny+guitar+watson).
> 
> This is also really cool but not what i meant. I wrote it a bit misleading. I meant when you do a back press (or swipe back) on your phone. So basically when onbackpressed gets called. The behavior right now is that a back press sends you back to the previous page, so for example to the results page. But the normal expected behavior from other search engines is that you just leave the image view and are back to the overview of all images.
> 
> I hope this is more understandable then my previous bad comment.

yea, both would be nice

Bnyro:
> This is also really cool but not what i meant. I wrote it a bit misleading. I meant when you do a back press (or swipe back) on your phone. So basically when onbackpressed gets called. The behavior right now is that a back press sends you back to the previous page, so for example to the results page. But the normal expected behavior from other search engines is that you just leave the image view and are back to the overview of all images.
>
https://github.com/searxng/searxng/pull/4201 adds that, I agree on that (also found myself accidentally pressing the back button there sometimes).

The issue linked by @return42 is matching this PR though.

return42:
> When users swipe fast, we create a new `imgLoader` for each new image the user swiped to.
> This means, even when the user swiped to a new image, the loading of the previous image (that'd no longer be needed in theory) continues in the background, because we always create new image loaders but never destroy or cancel them

:+1:

I moved this to a new issue:

- https://github.com/searxng/searxng/issues/4214

and I'm going to merge this PR now ..

snoweuph:
im on `2025.4.1+e6308b816` and the swiping is broken, might #4217 have completly broke it, or was there some other change were it was disabled?

snoweuph:
looks like the swiped-events dont fire. using firefox mobile (fennec) all extensions disabled

return42:
> im on `2025.4.1+e6308b816` and the swiping is broken

@snoweuph thanks for report this issue.  For real bugs, like this one don't hesitate to open a bug-ticket .. bug-comments in old PRs can be overlooked :-)

> might https://github.com/searxng/searxng/pull/4217 have completly broke it,

Nope, it was simply forgotten in the migration to vite [PR-4282](https://github.com/searxng/searxng/pull/4284) .. bug is fixed in:

- https://github.com/searxng/searxng/pull/4586

.. could you please give it a try / thanks  :+1:

snoweuph:
in `2025.4.1-e6308b816`  it doesn't works as expected as well.

PR #4201, opened 2025-01-16
[feat] image results: dismiss image viewer on back button press on mobile devices

Bnyro:
## What does this PR do?
- When pressing the system back button (or the browser back button) on mobile devices, this now closes the image details view if opened instead of returning to the previous page.
- This is only applied for mobile devices, because on desktop devices, the image details are shown side by side to the image results.

## Related issues
see https://github.com/searxng/searxng/issues/4151#issuecomment-2570850613


searx/static/themes/simple/src/js/main/results.js
```diff
@@ -52,6 +52,7 @@
       this.innerText = this.dataset.copiedText;
     });
 
+    const isMobile = screen.orientation.type.startsWith('portrait');
```
return42, line 55: I don't think we should limit the back button to mobile devices, I'd like to have this feature on desktop as well.  / AFAIK “portrait” is not a good criterion for mobiles though.

return42 (approved):
I remove the `isMobile` condition and fixed some other typos (not related to this PR .. but while we are here).

see [diff](https://github.com/searxng/searxng/compare/eb1a34252242c14a90b1faa5f341e9a21606b5f9..1f86f69ccd8fa854f62ba43050579093fa270eee)

PR #4210, opened 2025-01-19
[feat] plugins: new rerank results plugin

GenericMale:
## What does this PR do?

- Implements a "Rerank plugin" using the [Okapi BM25](https://en.wikipedia.org/wiki/Okapi_BM25) algorithm to re-order search results based on relevance.
- Uses the [bm25s](https://github.com/xhluca/bm25s) library for BM25 calculations.
- Supports two modes:
  - By default the engines are kept in the result but all positions are overwritten with the new rank. This will give results returned from multiple engines a boost as usual.
  - Alternatively only the first engine is retained.
- Plugin is disabled by default and needs to be added explicitly.
- Performance is good in my tests (usually below 10ms to rank >100 results) but still might not be suitable for public instances.

- Improves result relevance, especially when using more than a few engines and particularly for videos, images and when using more focused search terms.

Make sure bm25s is installed (`pip install bm25s` or add to requirements.txt)

Enable Plugin in settings.yml:
```
  enabled_plugins:
    - 'Rerank plugin'
```

To configure the two modes:

```
  rerank:
    remove_extra_engines: true
```

## Author's checklist

- [x] Documentation (not sure whats required?)
- [ ] Unit Tests
- [ ] Performance Tests including resource usage (CPU & Memory)


Bnyro:
I personally think that this is a quite interesting plugin, definitely worth looking into :+1: 

> Documentation (not sure whats required?)
>
First, you should probably add a comment for it to [`settings.yml`](https://github.com/searxng/searxng/blob/master/searx/settings.yml#L229) and mention that it requires the additional bm25s dependency there.

Second, it'd probably make sense to also include the documentation you wrote in `rerank.py` in the SearXNG web docs, so you should create a file at `docs/src/searx.plugins.rerank.rst` with the content:
```rst
.. _rerank plugin:

================
Rerank plugin
================

.. automodule:: searx.plugins.rerank
  :members:
```
That'd make it available at https://docs.searxng.org/src/searx.plugins.rerank.html.

searx/plugins/rerank.py
```diff
@@ -0,0 +1,77 @@
...
+try:
+    import bm25s
+except ImportError:
+    # Import error is ignored because the admin has to install bm25s manually to use the engine
+    pass
+
+name = 'Rerank plugin'
+description = 'Rerank search results, ignoring original engine ranking'
+default_on = False
+preference_section = 'general'
```
Bnyro: It's also visible in the settings ui when `bm25s` is not installed. We should however hide the plugin in that case.

Could be implemented like here: https://github.com/searxng/searxng/pull/3377/files#diff-554a7ca4c7612fcaa88c3e01928cdea683ae55d3a7c35cbc2677539a1001710aR41 and https://github.com/searxng/searxng/pull/3377/files#diff-340f925dc468a2e8194cc4059868c642ad96dadac281836d0969074b944e85e3R87.
(We removed that `is_allowed` method recently since it's not been used anywhere, so it'd need to be re-added)

searx/plugins/rerank.py
```diff
@@ -0,0 +1,77 @@
...
+def post_search(_request, search):
+    # pylint: disable=protected-access
+    results = search.result_container._merged_results
+    query = search.search_query.query
+    locale = search.search_query.locale
+
+    # Determine the stopwords based on the selected locale
+    stopwords = locale.language if locale and locale.language in stopword_langs else True
+
+    retriever = bm25s.BM25()
```
Bnyro: That's the point where it crashes when the user enables the plugin with bm25s enabled. Alternatively to the above suggestion, we should just `return` here.

Bnyro: Or in fact, we should always return here if `bm25s` is missing, since it's possible that the user enabled the plugin via the UI and it's still set to enabled in the user cookies but the instance doesn't have `bm25s` anymore.

GenericMale:
tbh i'm not expecting this to get merged soon. it's more like a proof of concept/idea to play around with.

but i'm using this locally since last week and it does work pretty well. has the obvious drawback that with short, generic queries and a small list of engines (e.g. if only google and bing is used), the more relevant sites might get ranked down.

an alternative idea i was exploring is to have the plugin add a dummy engine to the engine list of each result with the position according to the algorithm. this way you can customize the weight of it and have it influence the existing algorithm more or less.

Issue #4211, opened 2025-01-19
Date displayed for search results is in Greek and not English despite SearXNG preferences set to English and browser set to en-US


metronidazole:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
2025.1.10+94a0b415e

**How did you install SearXNG?**
searxng-docker

**What happened?**
The date displayed for results that have a date are in Greek despite having English set at the language in searxng preferences. My browser is sending Accept-Language as "en-US" so it's not a misconfiguration on my end.

**How To Reproduce**
1. Install SearXNG
2. Search for anything

**Expected behavior**
Dates should be in English e.g. "1 January 2024"

**Screenshots & Logs**

![Image](https://github.com/user-attachments/assets/3bc4b0cb-4f5c-4034-bdb9-9f0364aca2ac)

**Additional context**
N/A

return42:
This date comes from brave engine (try: `!brave searxng :en`) .. it is a part of the *content*:

![Image](https://github.com/user-attachments/assets/e9b609dc-10d0-44d0-a4ad-b425d39c7934)

I will see how we can fix it ..

return42:
I am closing this PR in favor of 

- https://github.com/searxng/searxng/issues/4006

PR #4212, opened 2025-01-20
[fix] engine brave: remove date from the content string

### LINKED ISSUE #4006: Dates in results are not shown in the correct language

2024.10.4+3e747d049 

searxng-docker

**What happened?**

Dates are in some other language even though the auto-detected lang is en-US

2024-11-11-183102_grim

**How To Reproduce**

Search for anything where results have dates

**Expected behavior**

Dates displayed for results should be in the auto-detected language

return42:
Related: https://github.com/searxng/searxng/issues/4211#issuecomment-2601941440
Closes: https://github.com/searxng/searxng/issues/4006

### DIFF
--- a/searx/engines/brave.py
+++ b/searx/engines/brave.py
@@ -291,15 +291,21 @@ def _parse_search(resp):
         if url is None or title_tag is None or not urlparse(url).netloc:  # partial url likely means it's an ad
             continue
 
-        content_tag = eval_xpath_getindex(result, './/div[contains(@class, "snippet-description")]', 0, default='')
+        content: str = extract_text(
+            eval_xpath_getindex(result, './/div[contains(@class, "snippet-description")]', 0, default='')
+        )  # type: ignore
         pub_date_raw = eval_xpath(result, 'substring-before(.//div[contains(@class, "snippet-description")], "-")')
+        pub_date = _extract_published_date(pub_date_raw)
+        if pub_date and content.startswith(pub_date_raw):
+            content = content.lstrip(pub_date_raw).strip("- \n\t")
+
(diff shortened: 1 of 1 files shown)


return42:
@craftyguy @metronidazole

This PR fixes the problems you have reported

- https://github.com/searxng/searxng/issues/4006
- https://github.com/searxng/searxng/issues/4211

To test, check out this PR branch, start an instance `make run` and do a query like `!brave :en searxng`

searx/engines/brave.py
```diff
@@ -291,15 +291,22 @@ def _parse_search(resp):
         if url is None or title_tag is None or not urlparse(url).netloc:  # partial url likely means it's an ad
             continue
 
-        content_tag = eval_xpath_getindex(result, './/div[contains(@class, "snippet-description")]', 0, default='')
+        content = (
+            extract_text(eval_xpath_getindex(result, './/div[contains(@class, "snippet-description")]', 0, default=''))
+            or ''
```
Bnyro: `extract_text()` always returns a string when its argument is an `ElementBase` (or a string in the default case), so this `or ''` is redundant. Reference: https://github.com/searxng/searxng/blob/master/searx/utils.py#L208

return42: Yeah is redundant .. it was because of the type checker reports a type error .. fixed it ([diff](https://github.com/searxng/searxng/compare/190bfa390c45ef31008cef688375378772be2dde..76e40474c192edddb7a731bee728d44d2b5e1815)) by adding a type to `content: str` and `#type: ignore`.

craftyguy:
Thank you for fixing this 😁

PR #4215, opened 2025-01-20
[fix] LESS sourcemaps broken in less-plugin-clean-css v1.6.0

### LINKED ISSUE #4143: less (CSS) source maps compiled but not used or shown any longer in the browser dev tools
The less (CSS) map files are used to debug and analyze CSS / less .. here in this example we see `style.less:109`

grafik

when I recently compile the less file, all we see is `$stdin`:

grafik

I don't think this problem is related to our source code. 

When I checkout an old commit e.g b42ce34ca the map is still loaded, but when I compile b42ce34ca these days again the source map is broken as shown above

I think this issue is related to the build tools .. but I don't have a clue where to look.

-----

https://github.com/searxng/searxng/blob/9e32cd2047b054f510aeb5e216f5ac1aba3f61b0/searx/static/themes/simple/gruntfile.js#L115-L146

return42:
The ``less-plugin-clean-css`` lacks some maintenance: the sourcemaps are broken since v1.6.0 (08/2024) [1]

- [1] https://github.com/less/less-plugin-clean-css/issues/42

Closes: https://github.com/searxng/searxng/issues/4143


Bnyro:
This means that CSS will no longer be minified and thus possibly our CSS files are larger in size?

I don't think that the change in size would really matter, we don't have that much CSS anyways in SearXNG (only some kB), which loads fast enough anyways.

Bnyro (approved):

return42:
> This means that CSS will no longer be minified and thus possibly our CSS files are larger in size?

Sadly "yes" .. I hope the issue will be fixed 

- https://github.com/less/less-plugin-clean-css/issues/42#issuecomment-2602524014

We can implement some quirks here .. but I think its better to wait for a fix from upstream.

Issue #4214, opened 2025-01-20
Improve resources of the gallery


return42:
A request that arose during the work on PR:

- https://github.com/searxng/searxng/pull/4198#pullrequestreview-2556793637

> Perhaps we could delay the loading of the full size images by 3-5 seconds ... if the user has already sipped further, then not every image would be loaded in its maximum size.

When users swipe fast, we create a new `imgLoader` for each new image the user swiped to.
This means, even when the user swiped to a new image, the loading of the previous image (that'd no longer be needed in theory) continues in the background, because we always create new image loaders but never destroy or cancel them (if I read the code correctly), ref https://github.com/searxng/searxng/pull/4198/files#diff-886bad7a631bc7afd8895a66a1406281cf3b69927c19ce5c6e80480e6517260aR71-R78.

In my opinion, it would make sense here to refactor the code to only use one image loader and cancel the loading when the user swipes to the next image, to avoid loading images the user swiped away after seeing the thumbnail.

A short delay (maybe 1 or 2 seconds) before loading the image make sense in my opinion, but it shouldn't be more than 2 because otherwise at least I would personally feel discomfy with waiting so long.

_Originally posted by @Bnyro in https://github.com/searxng/searxng/pull/4198#pullrequestreview-2556793637_

return42:
closed by: https://github.com/searxng/searxng/pull/4217

PR #4217, opened 2025-01-20
[refactor] results.js: cancel image loading after next one selected

Bnyro:
## What does this PR do?
- this PR refactors the image details dialog to only load one image at once, and cancel loading the previous ones when they are skipped before loading finished

- to decrease network load, see #4214 and https://github.com/searxng/searxng/pull/4198#pullrequestreview-2556793637

- search in the images category
- quickly navigate through images
- navigate back and observer that previous, skipped images were not fully loaded but cancelled and thus reloaded

## Related issues
- addresses #4214 (doesn't include any delay for loading the image as of now)


searx/static/themes/simple/src/js/main/results.js
```diff
@@ -52,43 +52,80 @@
...
+    const imgLoader = new Image();
+
+    const loadImage = (imgSrc, onSuccess) => {
+      // if defered image load exists, stop defered task.
+      if (imgTimeoutID) clearTimeout(imgTimeoutID);
+
+      // defer load of the detail image for 1 sec
+      imgTimeoutID = setTimeout(() => {
+        imgLoader.src = imgSrc;
+      }, 1000);
```
return42, line 76: > doesn't include any delay for loading the image as of now)

I implemented a `setTimeout` (for 1sec).

searx/static/themes/simple/src/js/main/results.js
```diff
@@ -52,43 +52,80 @@
...
-      // this allows us to dismiss the image details on pressing the back button on mobile devices
+      // add a hash to the browser history so that pressing back doesn't return
+      // to the previous page this allows us to dismiss the image details on
+      // pressing the back button on mobile devices
       window.location.hash = '#image-viewer';
 
       searxng.scrollPageToSelected();
+
+      // if there is none element given by the caller, stop here
+      if (!resultElement) return;
```
return42, line 102: I have moved the cancel conditions (`return`) to the top.

searx/static/themes/simple/src/js/main/results.js
```diff
@@ -52,43 +52,80 @@
...
+      img.src = thumbnail.src;
+
+      // show a progress spinner
+      const detailElement = resultElement.querySelector('.detail');
+      detailElement.appendChild(imgLoaderSpinner);
+
+      // load full size image in background
+      loadImage(src, () => {
+        // after the singelton loadImage has loaded the detail image into the
+        // cache, it can be used in the origin <img> as src property.
```
return42, line 125: I have added comments for explanation (I hope my interpretation of the source code was correct).

return42 (approved):
I placed a WIP commit on top of your initial commit .. 

I have tried to make the code a little more readable ... but I have almost no practical experience with JS ... I leave it to you whether we should take something from the WIP commit or not :-)

Bnyro:
> I placed a WIP commit on top of your initial commit ..

I have tried to make the code a little more readable ... but I have almost no practical experience with JS ... I leave it to you whether we should take something from the WIP commit or not :-)
>
That's awesome, thank you! I didn't think adding the delay would be as simple as adding a `timeout` to the `loadImage` function, that's a really great solution :+1:

All your assumptions in the comments are right, and the comments are really helpful to make the code more readable :+1: I was also struggling to understand the old code when working on this PR because it was much too complicated for what it actually did.

PS: Fyi, I just reworded your commit message, anything else stayed the same.

Bnyro:
PS: We need to see how well the timeout of one second is perceived by users, maybe some think that it's a bit too long. But until somebody complains, I personally think that one second is a good solution.

Issue #2247, opened 2023-03-10
Add kagi.com


toyboatcash:
**Working URL to the engine**
(https://kagi.com/search?q=)

**Why do you want to add this engine?**
kagi has their own index of the finest results, augmented by the results from the best search engines on the market. 

**Features of this engine**
kagi has their own index, Doesn't track its users, fast, clutter-free search experience

**How can SearXNG fetch the information from this engine?**
Session Link This URL can be used to automatically log you into Kagi, as long as you remain logged in to this session. Do NOT share this link with anyone else! https://kagi.com/search?token= or https://help.kagi.com/kagi/search-api/overview.html 

**Applicable category of this engine**
general, Images, Videos, News, Maps 

**Additional context**
https://github.com/kagisearch/kagi-docs/discussions/24

-----

Related:

- https://github.com/searxng/searxng/pull/6182
- https://github.com/searxng/searxng/pull/4220

gatunet:
I was looking into making a PR for adding Kagi, but to access the search api you need an [special account and you pay $25 for 1000 queries](https://help.kagi.com/kagi/api/search.html), so in theory it could be done easily, but I don't think people would use it(or maybe even be allowed to, given that you have to ask for the special team account).

Also I don't think we should implement this engine without following their guidelines, for example by keep using the token and parsing the results, because this in a way hurts them, and they at least [seam to care about privacy](https://kagi.com/privacy), so I think if we do this without using their API and paying the proper amount we'll be hurting a privacy respecting search engine.

They do [seam open to feedback and requests](https://github.com/kagisearch/kagi-docs/discussions/categories/kagi-search-api-feature-requests-ideas), so maybe try to talk about just lifting the special account thing, but then there's the price which is a problem for heavy users and I don't think they would be open to change that.

lolcatw:
I looked at their API myself, they piss me off. They load all of the page's data through websockets (json).

unixfox:
@lolcatw if you add `html/` before `/search` then you get the javascriptless version which is probably easier to parse since it's just HTML.

https://kagi.com/html/search?q=ok

lolcatw:
@unixfox oh sweet, I'll look into it...

gmemstr:
> They load all of the page's data through websockets (json).

Not quite sure where you got this from, as their own API docs[1] says you can just GET an endpoint and recieve a reasonable looking JSON response. Should be trivial to implement.

[1]: https://help.kagi.com/kagi/api/search.html

[...]

PR #4220, opened 2025-01-21
Initial Kagi search support

SentientTapeDrive:
## What does this PR do?

Adds Kagi general web search via no-JS scraping when using a GET API key of a paid account.

SearXNG lacks any kind of Kagi support.

Some unit tests were made, and search using the Kagi engine after setting GET API key in settings config file.

## Author's checklist

Working:
- General search
- Pagination of general search
- Capture and reuse of cookies after initial GET request with API key

Not setup yet:
- Safe search adjustment
  - Generic account-wide setting; no obvious GET/POST param to change per search instance
    - "Safe Search" and "Image Safe Search" are on/off toggles under Kagi settings -> search -> general
- Regional results by country / territory
- Time range
  - Breaks page results if not "Anytime"
- Image search
- Video search

## Related issues

- https://github.com/searxng/searxng/issues/2247
- https://github.com/searxng/searxng/pull/6182

### DIFF
--- a/searx/engines/kagi.py
+++ b/searx/engines/kagi.py
@@ -1,0 +1,148 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Kagi Search
+Scrapes Kagi's HTML search results.
+"""
+
+from urllib.parse import urlencode
+from lxml import html
+
+from searx.utils import extract_text, eval_xpath, eval_xpath_list
+from searx.exceptions import SearxEngineAPIException
+from searx import logger
+
+logger = logger.getChild('kagi')
+
+about = {
+    "website": 'https://kagi.com',
+    "wikidata_id": None,
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": True,
+    "results": 'HTML',
+}
+
+categories = ['general', 'web']
+paging = True
+time_range_support = False
+safesearch = False
+
+base_url = 'https://kagi.com/html/search'
+
+api_key = None  # Set in settings.yml
+
+# Global cookie storage for Kagi authentication
+kagi_cookies = {'kagi_session': None, '_kagi_search_': None}
+
+
+def request(query, params):
+    if not api_key:
+        raise SearxEngineAPIException('missing Kagi API key')
+
+    page = params['pageno']
+
+    if 'cookies' not in params:
+        params['cookies'] = {}
+    params['cookies'].update(kagi_cookies)
+
+    if kagi_cookies['kagi_session'] and kagi_cookies['_kagi_search_']:
+        logger.debug(
+            "Using Kagi cookies for authentication - session: %s, search: %s",
+            kagi_cookies['kagi_session'],
+            kagi_cookies['_kagi_search_'],
+        )
+        search_url = base_url + '?' + urlencode({'q': query, 'batch': page})
+    else:
+        missing = []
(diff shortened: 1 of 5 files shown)


searx/engines/kagi.py
```diff
@@ -0,0 +1,148 @@
...
+
+def request(query, params):
+    if not api_key:
+        raise SearxEngineAPIException('missing Kagi API key')
+
+    page = params['pageno']
+
+    if 'cookies' not in params:
+        params['cookies'] = {}
+    params['cookies'].update(kagi_cookies)
```
Bnyro, line 45:
```suggestion
      params['cookies'] = kagi_cookies
```
`params['cookies']` is always an empty dict before the request.

searx/settings.yml
```diff
@@ -2466,6 +2466,19 @@ engines:
...
+    shortcut: kg
+    categories: [general, web]
+    disabled: true
+    timeout: 4.0
+    api_key: ""
+    about:
+      website: https://kagi.com
+      use_official_api: false
+      require_api_key: true
+      results: HTML
```
Bnyro, line 2480: This should be commented out by default since the api key is required. Apart from that the `about` section here is redundant, it's already set in `kagi.py`.

tests/unit/test_engine_kagi.py
```diff
@@ -0,0 +1,152 @@
...
+                    </h3>
+                </div>
+                <div class="__sri-url-box">
+                    <span class="host">example2.com</span>
+                </div>
+                <div class="__sri-body">
+                    <div class="__sri-desc">Content 2</div>
+                </div>
+            </div>
+        </div>
```
Bnyro, line 44: I don't think we should include all that HTML in SearXNG, it doesn't really help us if Kagi changes its layout, because this test would still pass although it's not related to the real Kagi response anymore. It'd be better to test actual responses from Kagi if an api key is available.

searx/engines/kagi.py
```diff
@@ -0,0 +1,148 @@
...
+paging = True
+time_range_support = False
+safesearch = False
+
+base_url = 'https://kagi.com/html/search'
+
+api_key = None  # Set in settings.yml
+
+# Global cookie storage for Kagi authentication
+kagi_cookies = {'kagi_session': None, '_kagi_search_': None}
```
Bnyro, line 34: I'm noting here that SearXNG is multi-threaded and it theoretically would be better to store these in Redis if available, but I don't think that'd really be worth changing.

searx/engines/kagi.py
```diff
@@ -0,0 +1,148 @@
...
+    params['max_redirects'] = 1
+
+    return params
+
+
+def response(resp):
+    results = []
+
+    if 'set-cookie' in resp.headers:
+        cookies = resp.headers.get_list('set-cookie')
```
Bnyro, line 85: You should be able to access `resp.cookies` here I think? (not 100% sure)

Bnyro:
I noticed they also have a JSON API, but that one seems to be more expensive than the normal plan?

jcrabapple:
I would love to see Kagi support in SearXNG!

return42:
@SentientTapeDrive since there is no response from your side over a longer time to this PR, I mark this PR as DRAFT.

return42:
Closed due to inactivity.

PR #4277, opened 2025-02-03
[fix] public domain image archive: change name to lowercase

allixx:
Otherwise logs are flooded with warnings:
WARNING:searx.engines: Engine name is not lowercase: "Public Domain Image Archive", converting to lowercase


Bnyro (approved):
Thanks :+1:

PR #4282, opened 2025-02-05
[refactor] plugins: allow setting default enabled state via settings.yml

### LINKED ISSUE #4263: Hostnames plugin defaults to off even though it is enabled in the settings.yml

2025.1.29+738906358 

Docker-compose

**What happened?**

I haven't changed anything in the settings, but suddenly after some recent update the hostnames plugin is disabled even though it has been enabled in the settings.yml this whole time.  I can enable it again by going to settings and toggle the Hostnames plugin on and click save. But then after I click the button to get back the default settings, the hostnames plugin is disabled again.

**How To Reproduce**

Enable the hostnames plugin in settings.yml. Then check the settings page of Searxng and see if the Hostnames plugin is enabled.

**Expected behavior**

Hostnames plugin should be enabled when I enable it in settings.yml

**Screenshots & Logs**

Settings.yml:

```yaml
enabled_plugins:
#   # these plugins are enabled if nothing is configured ..
  - 'Hash plugin'
  - 'Search on category select'
  - 'Self Information'
  - 'Tracker URL remover'
#   - 'Ahmia blacklist'  # activation depends on outgoing.using_tor_proxy
#   # these plugins are disabled if nothing is configured ..
#   - 'Hostname replace'  # see hostname_replace configuration below
  - 'Hostnames plugin'  # see 'hostnames' configuration below
  - 'Basic Calculator'


hostnames:
  replace: # 'rewrite-hosts.yml'
    '(.*\.)?youtube\.com$': 'invidious.example.com'
    '(.*\.)?youtu\.be$': 'invidious.example.com'
    #   '(.*\.)?youtube-noocookie\.com$': 'yotter.example.com'
    '(.*\.)?reddit\.com$': 'farside.link/libreddit'
    '(.*\.)?redd\.it$': 'farside.link/libreddit'
    '(www\.)?twitter\.com$': 'farside.link/nitter'

``` 

Settings in Searxng:

Image

**Additional context**

Bnyro:
## What does this PR do?
- this PR changes the syntax of enabling plugins:
```yaml
   plugins:
     - id: 'calculator'
       default_on: true
     - id: 'hash_plugin'
       default_on: false
```
- the old syntax is still valid though
- any plugins that are not listed in this configuration option are **disabled** completely

- this PR allows instance maintainers to set whether a plugin is either
  - on by default: users can manually opt out in the settings
  - off by default: users can still enable it in the settings 
- Previously this has been hardcoded in the plugin implementation, which is bad for instances that want to provide a plugin, but not have it enabled by default (e.g. for performance reasons)

- enable and disable plugins

## Related issues
closes #4263
closes #759

### DIFF
--- a/docs/admin/plugins.rst
+++ b/docs/admin/plugins.rst
@@ -30,7 +30,6 @@ Configuration defaults (at built time):
       {% for plg in plugins %}
 
       * - {{plg.info.name}}
-        - {{(plg.default_on and "y") or ""}}
         - {{plg.info.description}}
 
       {% endfor %}
--- a/docs/admin/settings/settings_plugins.rst
+++ b/docs/admin/settings/settings_plugins.rst
@@ -12,13 +12,13 @@ Plugins
 
 
 The built-in plugins can be activated or deactivated via the settings
-(:ref:`settings enabled_plugins`) and external plugins can be integrated into
+(:ref:`settings plugins`) and external plugins can be integrated into
 SearXNG (:ref:`settings external_plugins`).
 
 
-.. _settings enabled_plugins:
+.. _settings plugins:
 
-``enabled_plugins:`` (internal)
+``plugins:`` (internal)
 ===============================
 
 In :ref:`plugins admin` you find a complete list of all plugins, the default
@@ -26,14 +26,23 @@ configuration looks like:
 
 .. code:: yaml
 
-   enabled_plugins:
-     - 'Basic Calculator'
-     - 'Hash plugin'
-     - 'Self Information'
-     - 'Tracker URL remover'
-     - 'Unit converter plugin'
-     - 'Ahmia blacklist'
+  plugins:
+    - id: 'calculator'
+      default_on: true
+    - id: 'hash_plugin'
+      default_on: true
+    - id: 'self_info'
+      default_on: true
+    - id: 'tracker_url_remover'
+      default_on: true
+    - id: 'unit_converter'
+      default_on: true
+    - id: 'ahmia_filter'  # activation depends on outgoing.using_tor_proxy
+      default_on: true
 
+In order to disable a plugin by default, but still allow users to use it by enabling
+it in their user settings, set ``default_on`` to ``false``. To completely disable a
+plugin, you can set `inactive` to `true`.
 
 .. _settings external_plugins:
 
@@ -56,9 +65,11 @@ In the :ref:`settings.yml` activate the ``plugins:`` section and add module
 .. code:: yaml
 
    plugins:
-     - only_show_green_results
-     # - mypackage.mymodule.MyPlugin
-     # - mypackage.mymodule.MyOtherPlugin
+     - id: only_show_green_results
+       default_on: true
+     # - fqn: mypackage.mymodule.MyPlugin
+     # - fqn: mypackage.mymodule.MyOtherPlugin
+     #   default_on: false
 
 .. hint::
 
--- a/searx/plugins/__init__.py
+++ b/searx/plugins/__init__.py
@@ -18,7 +18,6 @@ area:
    class MyPlugin(Plugin):
 
        id = "self_info"
-       default_on = True
 
        def __init__(self):
            super().__init__()
--- a/searx/plugins/_core.py
+++ b/searx/plugins/_core.py
@@ -75,9 +75,6 @@ class Plugin(abc.ABC):
     id: typing.ClassVar[str]
     """The ID (suffix) in the HTML form."""
 
-    default_on: typing.ClassVar[bool]
-    """Plugin is enabled/disabled by default."""
-
     keywords: list[str] = []
     """Keywords in the search query that activate the plugin.  The *keyword* is
     the first word in a search query.  If a plugin should be executed regardless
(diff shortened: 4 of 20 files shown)


Bnyro:
I'm a bit confused about the current way we identify plugins:
- In `settings.yml`, we reference them by its `name`, although that is defined via `gettext(...)` in some plugins, so that it's translatable
  - This probably will break if the instance uses a different default language? Why are we even doing this at all?
- In the user settings (via cookies), we identify them by their `id`, which should probably be the **only** way that it's possible to reference them

dalf:
> I'm a bit confused about the current way we identify plugins

Since a long time, it is possible to reference plugin by id (python module name), the reference by user name was there as a retrocompatibility ; however the switch was never done. 

This PR is good opportunity to do it.

[EDIT] the code has changed compare to what I remember, the above statement might be wrong.

Bnyro:
> Since a long time, it is possible to reference plugin by id (python module name), the reference by user name was there as a retrocompatibility ; however the switch was never done.
>
> This PR is good opportunity to do it.
>
`plugin_id` is a variable that can be configured in each plugin's python file.

Many plugins don't have an ID set yet, so I assume we need to add IDs to them? Or does the ID default to the filename of the module? Isn't that ID property useless anyways, shouldn't we always use the filename as ID?

@return42 As you have recently modified the plugins logic, do you have any opinion on that?

return42:
Sorry for my late response: I will look for answers to your questions later, but first I want to introduce the model behind:

TL;DR; Plugins should be identified by ID and they are identified by the ID .. there is just some old quirk around the settings option `enabled_plugins` / @dalf knows this better than I and I agree with his https://github.com/searxng/searxng/pull/4282#issuecomment-2637396064

---

Plugins & Answerers were revised in the course of typification.  These can now be implemented as classes.

- https://docs.searxng.org/dev/plugins/development.html
- https://docs.searxng.org/dev/answerers/development.html

Plugins have an `id`:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L72-L79

One Plugin has implemented as a class:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/hash_plugin.py#L19-L26

I haven't touched all plugins .. for those plugins that have not been migrated to the class model, there is a legacy wrapper:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L165-L175

> Many plugins don't have an ID set yet, so I assume we need to add IDs to them? Or does the ID default to the filename of the module?

It is the module name for legacy builtin plugins:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L186


Plugins are manged by the `PluginStorage`

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L236-L240

When SearXNG starts `PluginStorage.load_builtins` is called. 

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L267-L271

The hash plugin shown above is loaded by this part:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L279-L281

the other internal plugins are loaded with the legacy wrapper `ModulePlugin` (shown above)
 
https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L283-L285

external plugins are loaded by this part of the `load_builtins` method:

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L287-L288 

----

The `plugin` option for external plugins has been revised ... those plugins are enabled by the full qualified class name 

- https://docs.searxng.org/admin/settings/settings_plugins.html

The `enabled_plugins` hasn't been touched and I agree with @dalf "This PR is good opportunity to do it."

return42:
> The `enabled_plugins` hasn't been touched and I agree with @dalf "This PR is good opportunity to do it."

From an admin's point of view, but also from a technical point of view, it makes no sense to distinguish between external and internal plugins / the switch `enabled_plugins` can actually be omitted ..

All plugins are inheritors of the `Plugin` class ..

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L72-L73

And all plugins are managed in the `PluginStorage` ..

https://github.com/searxng/searxng/blob/147bda894e380b066f0d57dd7d3b1cb0974a0307/searx/plugins/_core.py#L236-L237

In the configuration, external & internal plugins should be identified by there full qualified name (ATM this is only true for externel ones):

```
plugins:
  - mypackage.mymodule.MyPlugin
  - mypackage.mymodule.MyOtherPlugin
```

We can migrate this list to a list of dict:

```
plugins:
     - fqn: mypackage.mymodule.MyPlugin
       default_on: true
```

The `plugin` option hasn't been set by any instance owner / we wont have downward compatibilities ..

We just have to think about a migration path for `enabled_plugins` and the internal plugins that are still modules today should be migrated to the plugin class ... I was planning to do that anyway to get the legacy wrapper out of the code at some point.

825i:
In re: of my outburst in another issue. Please thanks and hugs and kisses if you can merge this asap and please consider never changing the hostnames plugin again in such a way that it breaks entirely lol.

return42:
Superseded by:

... The ``enabled_plugins:`` section in SearXNG's settings do no longer exists. There is no longer a distinction between built-in and external plugin, all plugins are registered via the settings in the ``plugins:`` section.

The implementation is similar to what has been described in this comment: https://github.com/searxng/searxng/pull/4282#issuecomment-2642231325

> I'm a bit confused about the current way we identify plugins:

With #4538, plugins can be registered via a fully qualified class name.  A configuration (`PluginCfg`) can be transferred to the plugin, e.g. to activate it by default / *opt-in* or *opt-out* from user's point of view.

```YAML

    plugins:
      ...
        active: true
      ...
```

PR #4283, opened 2025-02-05
[chore] engines: replace datetime.utcfromtimestamp with datetime.fromtimestamp

### LINKED ISSUE #4257: Deprecation Warning: datetime.utcfromtimestamp()
The following deprecation warning was encountered during execution of `hackernews.py`:

```
hackernews.py:90: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC)
```

Also, to search comments instead of the story, how do I use the "comment" tag?

Bnyro:
## What does this PR do?
- replace `datetime.utcfromtimestamp` (deprecated) with `datetime.fromtimestamp`
- `datetime.fromtimestamp` seems to default to utc, so the optional timezone parameter can be omitted

## Related issues
closes #4257

### DIFF
--- a/searx/engines/9gag.py
+++ b/searx/engines/9gag.py
@@ -51,7 +51,7 @@ def response(resp):
                     'url': result['url'],
                     'title': result['title'],
                     'content': result['description'],
-                    'publishedDate': datetime.utcfromtimestamp(result['creationTs']),
+                    'publishedDate': datetime.fromtimestamp(result['creationTs']),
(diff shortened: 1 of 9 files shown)


return42 (approved):
I tested the engines and it works .. thanks a lot :+1:

Issue #559, opened 2021-11-29
Document JS source code


return42:
## Is your feature request related to a problem?

We need a markup and a toolchain to extract documentation from the JS sources.

[1] https://jsdoc.app/
[2] https://pypi.org/project/sphinx-js/

## Describe the solution you'd like

Markup: `@use JSDoc` [[1]](https://jsdoc.app) seems widely used and common.

Toolchain: Sphinx-Plugin https://pypi.org/project/sphinx-js/

_"""It provides a handful of directives, patterned after the Python-centric autodoc ones, for pulling JSDoc-formatted documentation into reStructuredText pages. And, because you can keep using JSDoc in your code, you remain compatible with the rest of your JS tooling,
"""_[[2]](https://pypi.org/project/sphinx-js/)

## Additional context

We already have markup & toolchain for two languages (worth to document by time):

### Python

The Python source code documentation is generated by the autodoc module::

https://github.com/searxng/searxng/blob/d0e21a01b4e14df905f25445a3f6bcbec481671c/docs/conf.py#L83

In the reST documentation the entry points are by example used in the [docs/src](https://github.com/searxng/searxng/tree/master/docs/src) folder ...

```reST

.. automodule:: searx.engines.demo_online
  :members:

```

### Shell Scipts

In a pre-build step, the Shell scripts generate some *include* files in the build directory:

https://github.com/searxng/searxng/blob/d0e21a01b4e14df905f25445a3f6bcbec481671c/manage#L383-L394

In the reST documentation the entry points are by example used in the [docs/admin/installation-searxng.rst](https://github.com/searxng/searxng/blob/master/docs/admin/installation-searxng.rst#install-packages) file, see [`kernel-include` directive](https://return42.github.io/linuxdoc/linuxdoc-howto/kernel-include-directive.html#kernel-include-directive) / [literal-blocks](https://searxng.github.io/searxng/dev/reST.html#literal-blocks) ...

```reST
.. kernel-include:: $DOCS_BUILD/includes/searx.rst
   :start-after: START distro-packages
   :end-before: END distro-packages
```

In the `./utils/searx.sh doc` generates this START/END tags from a template file:

https://github.com/searxng/searxng/blob/d0e21a01b4e14df905f25445a3f6bcbec481671c/utils/searx.sh#L1026

See `docs/build-templates/searx.rst` where the "[START|END] distro-packages" are generated:

[...]

PR #4284, opened 2025-02-05
migrate web-client development from grunt to vite

### LINKED ISSUE #141: Static files: move source to a dedicated directory.

return42:
## What does this PR do?

Migrate web client development (simple theme) from grunt to [Vite](https://vite.dev/).

Vite consists of a builder and a server, this PR only implements the build, what grunt has done so far. We will integrate the vite server into our development processes in a followup PR.

1. grunt and its plugins are no longer maintained  
2. we need a modern developer environment for front-end (FE) development

Start a server (`make run`) and see if there are no changes in the frontend.  Apart from a few small bugs that have been fixed around the icons, this PR should not bring any changes from the user's point of view.

To test FE development jump to folder `client/simple` and have a look at the README there.

## Notes

The switch to vite in this PR will enable us to develop our FE faster and more securely in future (e.g. we do no longer rely on unmaintained components and can benefit from state of the art technologies).

This PR deliberately avoids changes to the JS code and adopts it 1:1 for the time being. In further PRs we should then restructure the JS code base to JS modules (and imports/exports).

The migration to vite currently still has one catch: I have not been able to generate the source maps for the CSS that are generated from the LESS ... perhaps we can find a solution for this in the course of the review ..

- https://github.com/vitejs/vite/discussions/13845#discussioncomment-11992084

## Related issues

- https://github.com/searxng/searxng/issues/559
- https://github.com/searxng/searxng/issues/141

-----

Closes: #141


return42:
@vojkovic I remember you have custom LESS for the theme .. this PR will move the sources from `searx/static/themes/simple/src` to `client/simple/src` .. after merging this PR, you may have to adjust the workflow with which you apply your themes .. I have also already informed @mrpaulblack ... I don't know who else has customizations to the LESS files ...

vojkovic:
@return42 I appreciate you letting me know - I'll follow this PR and make the changes once the merge happens. :-)

Bnyro (commented):
It looks like the JavaScript on the results page is currently not being loaded properly - I can't tell why exactly yet.

To reproduce: Try to `Shift`-select additional categories with the left mouse button - works on the latest master but not with this PR.

client/simple/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "simple",
+  "version": "1.0.0",
+  "type": "module",
+  "scripts": {
+    "clean": "rm -Rf node_modules",
+    "build": "node theme_icons.js && vite build",
+    "fix": "eslint --fix && stylelint --fix strict 'src/**/*.{css,scss,sass,less,styl,vue,svelte}'",
```
Bnyro, line 8: `sass,scss,styl,vue,svelte` are probably not needed since we don't use these anywhere? 

We only use `less` (and maybe also `css`).

client/simple/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "stylelint-prettier": "^5.0.3",
+    "svgo": "^3.3.2",
+    "vite": "^6.0.7",
+    "vite-plugin-static-copy": "^2.2.0",
+    "vite-plugin-stylelint": "^6.0.0",
+    "webpack": "^5.97.1",
+    "webpack-cli": "^6.0.1"
+  },
+  "dependencies": {
+    "autocomplete-js": "^2.7.1"
```
Bnyro, line 38: This dependency is unmaintained since 6 years, we shouldn't use this anymore (not related to this PR) ...
(source: https://www.npmjs.com/package/autocomplete-js)

I'll make a PR to replace it in some time...

return42: Thats true .. we have such old unmaintained libs .. 

> I'll make a PR to replace it in some time...

:rocket: .. and we should take in mind there are open tasks --> https://github.com/searxng/searxng/issues?q=label%3A%22area%3A%20autocomplete%22%20state%3Aopen

But first lets migrate the code "as is" into the new environment for our FE development (vite), than we can migrate the JS code base to JS modules .. and when we have JS modules we can implement a new autocompleter and improve other client stuff.

Bnyro: That warning was too late, already implemented in https://github.com/searxng/searxng/pull/4318.

I can rebase it after this PR is merged though.

return42: > I can rebase it after this PR is merged though.

I will see what will be easier .. but not today :-)

client/simple/package.json
return42: By default stylelint checks all files matching js glob pattern in `./`

    ./**/*.{css,scss,sass,less,styl,vue,svelte}

I left it at default, with the only difference that we have our sources in `/src`

    src/**/*.{css,scss,sass,less,styl,vue,svelte}

If you insist, we change it, but in my opinion, the advantage is that if a PR adds new file types in the future, they will be checked automatically (without any changes to the command here) by stylelint ... and there is no disadvantage.

vojkovic:
Running this now in production on my instance. https://github.com/privau/searxng/commit/e64d2b9f07caee677b2062530d3f39b286be88f4

Haven't noticed any issues just yet. But did notice a considerable speedup in the frontend build time. :)

As mentioned, the main issue currently looks to be that the css sourcemaps not being built.

vojkovic:
@return42 Noticed something! - the new svg optimiser seems to be breaking a few svgs like the information-circle and news category icon. There are internal shapes not being coloured.

return42:
> @return42 Noticed something! - the new svg optimizer seems to be breaking a few svgs

It's not the optimizer ... it's the icons themselves:

>  There are internal shapes not being colored.

News icon haven't changed (was also broken before) .. but in the old client we had a manually fixed info icon.

I now fixed both icons --> [see diff](https://github.com/searxng/searxng/compare/032aac7d87e64cd4cbdf641ee169badefb3ea4c4..166426234601bd7fc74d312844abb804128e8fe0)

```
 // some of the ionicons are not suitable for a dark theme, we fixed the svg
 // manually in src/svg/ionicons
```

client/simple/theme_icons.js
```diff
@@ -0,0 +1,84 @@
...
+      "people": "people-outline.svg",
+      "play": "play-outline.svg",
+      "radio": "radio-outline.svg",
+      "save": "save-outline.svg",
+      "school": "school-outline.svg",
+      "search": "search-outline.svg",
+      "seeder": "swap-vertical.svg",
+      "settings": "settings-outline.svg",
+      "tv": "tv-outline.svg",
+    },
```
return42, line 68: @vojkovic: while we are just talking about icons .. this PR also fix some very old issues with some missing SVGs: e.g. SVGs for `seeder`, `lecher` and *file size* (`save`) had been missed in the past ..


grafik


grafik

vojkovic: Very nice - great work with this mate.

vojkovic (approved):

vojkovic:
Looks good - I had a look at the issue with vite not generating the css source maps. I read over the github discussion and even after checking the source of vite, I can't get them to build either.

utils/lib_sxng_themes.sh
```diff
@@ -1,100 +1,36 @@
...
-    )
-    dump_return $?
-}
-
-themes.live() {
-    local LIVE_THEME="${LIVE_THEME:-${1}}"
-    case "${LIVE_THEME}" in
-        simple)
-            theme="searx/static/themes/${LIVE_THEME}"
-            ;;
```
return42, line 50: > I read over the github discussion and even after checking the source of vite, I can't get them to build either.

@vojkovic many thanks for your research on this .. I guess the issue is related to the *minification* of the CSS .. but for the **build phase** I prefer to to *minify* ..

> Vite consists of a builder and a server, this PR only implements the build,

in a follow up PR I will integrate vite's development server in a (new) `themes.live` command .. then, in the **development phase** we don't need to *minify* and have CSS source maps back .. at least I hope so ;-)

----

BTW: we already have a similar issue with LESS & CSS

- https://github.com/searxng/searxng/pull/4215
- https://github.com/less/less-plugin-clean-css/issues/42#issuecomment-2602524014

The problem may lie in a [sub-sub-sub dependency](https://images.ctfassets.net/vkoe68wv76dt/2zWABwW9gb33FGGWBflZ3y/143994d9368dd33f1dbe31aba17ec517/node-modules-app-performance_.png) and therefore affects both (vite and less-plugin-clean-css)... but I can only speculate about that ..

Bnyro (approved):
Tested again, looks good to me now! :+1:

PR #4286, opened 2025-02-06
[fix] Internet links disappeared from wikidata side box

return42:
Closes: https://github.com/searxng/searxng/issues/4285

Reported and tested by: @Popolon
Suggested-by: @dalf

PR #4287, opened 2025-02-06
[fix] Internet links disappeared from wikidata side box (second try)

return42:
Related:

- https://github.com/searxng/searxng/pull/4286#issuecomment-2639960013

@dalf @Popolon can you please have a look this patch works for you .. thanks :+1:

### DIFF
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -641,7 +641,7 @@ class WDURLAttribute(WDAttribute):
             return get_external_url(url_id, value)
 
         if self.url_path_prefix:
-            [account, domain] = value.split('@', 1)
+            [account, domain] = [x.strip("@ ") for x in value.rsplit('@', 1)]
             return f"https://{domain}{self.url_path_prefix}{account}"
 
         return value


searx/engines/wikidata.py
```diff
@@ -641,7 +641,7 @@ def get_str(self, result, language):
             return get_external_url(url_id, value)
 
         if self.url_path_prefix:
-            [account, domain] = value.split('@', 1)
+            [account, domain] = [x.strip("@ ") for x in value.rsplit('@', 1)]
```
Bnyro, line 644:
```suggestion
            [account, domain] = [x.lstrip("@") for x in value.rsplit('@', 1)]
```
would be more applicable I think? We probably don't want to strip `@` from the end as we only add an `@` to the start.

return42: :thinking: sure? .. I think we don't want to have an `@` in `account` nor in `domain` .. this was at least my intention.

Bnyro: We certainly don't want it in domain, but that's already assured by using `rsplit(...)`, since it splits at the last `@`.

`@` in `account` might be possible though if some service allow usernames like `foo@`. I don't know if there's any that does so though ...

Bnyro: Though you're probably right that most services don't allow any `@` in the `account`, so it doesn't matter which solution we use in that case.

PR #4294, opened 2025-02-07
[mod] improve description of the prefernces URL in the "COOKIES" tab

### LINKED ISSUE #4272: String/Term Suggestion: Specifying custom settings in the preferences URL can be used to sync preferences across devices.
Location:

https://github.com/searxng/searxng/blob/6324a9752a2b21a1dcb709c7fb643e361189163e/searx/templates/simple/preferences/cookies.html#L43

I think

### A URL with preference settings allows you to sync your preferences across devices.

Is more clear, any ideas?

return42:
@unixfox what do you think .. is this a better description?

Closes: https://github.com/searxng/searxng/issues/4272

### DIFF
--- a/searx/templates/simple/preferences/cookies.html
+++ b/searx/templates/simple/preferences/cookies.html
@@ -40,7 +40,7 @@
   </pre>{{- '' -}}
 </div>{{- '' -}}
 <p class="small_font">
-  {{- _('Specifying custom settings in the preferences URL can be used to sync preferences across devices.') -}}
+  {{- _('A URL containing your preferences. This URL can be used to restore your settings on a different device.') -}}
 </p>
 <h4>
   {{- _('Copy preferences hash') -}}:{{- '' -}}


searx/templates/simple/preferences/cookies.html
```diff
@@ -40,7 +40,7 @@ <h4>
   </pre>{{- '' -}}
 </div>{{- '' -}}
 <p class="small_font">
-  {{- _('Specifying custom settings in the preferences URL can be used to sync preferences across devices.') -}}
+  {{- _('A URL with preference settings allows you to sync your preferences across devices.') -}}
```
Bnyro:
```suggestion
  {{- _('A URL containing your preferences. This URL can be used to restore your settings on a different device.') -}}
```
would be my take since "preferences" is often used as a synonym to "settings", that'd almost be twice the same word.

unixfox (approved):

Issue #3191, opened 2024-02-08
SearXNG sometimes redirects to main page without query


mooreye:
update from the thread maintainers: Please read the summary in https://github.com/searxng/searxng/issues/3191#issuecomment-2670632164

----

See: https://github.com/libredirect/browser_extension/issues/880

Problem encountered in Firefox extension Libredirect, but they believe this is SearXNG's problem.

unixfox:
It's possible that something does not work great with the anti bot protection.

return42:
Can't reproduce the issue with my instance https://darmarit.org/searx/ or https://paulgo.io/  you mentioned in  https://github.com/libredirect/browser_extension/issues/880#issuecomment-1899420751 .. not sure what the issue is you have.

Austin-Olacsi:
I can say for sure that this is not an issue with libredirect. It has happened to me on some occasions, and i do not have this extension.

i don't know how to reproduce it unfortunately. it just happened to me on my private instance a few moments ago. it usually does not happen. seems very random.

gabeklavans:
Happens to me with enough frequency to be annoying (about once a week). Both when I search using Firefox opensearch (no extensions), and when I use a redirect on iOS safari. It seems to happen with about the same frequency with both access methods. 

The only other variables for me are my DNS provider (cloudflare, on which I have proxy turned off), and nginx being my reverse proxy. But I'm not sure if the problem lies with either of those.

I will try searching directly from the search home page for a while to see if it still occurs.

MadAim123:
I have the same problem. I testes a lot of instances from https://searx.space/ and it happens randomly. I tried it with different browsers and devices and can't resolve it. I don't have any special plugins or extensions installed.

- https://github.com/searxng/searxng/discussions/3219

return42:
@MadAim123 do you have the same issue on my instance? https://darmarit.org/searx/search?q=test&language=de&safesearch=0&categories=general&time_range=month

MadAim123:
@return42 when I tested it now (Firefox, Ege, Chrome on Windows), I can't reproduce it. 
But it happens randomly. So maybe it would happen tomorrow or in 2 weeks - or maybe not. Did you changed something in routing? Or do you use vanilla?

return42:
Vanilla installed by the [scripts](https://docs.searxng.org/admin/installation-scripts.html#installation-scripts) ..

```
./utils/searxng.sh install all
./utils/searxng.sh install apache
```
[...]

PR #4299, opened 2025-02-08
[fix] botdetection: return error, do not fail silently

bearz314:
## What does this PR do?

Presently, if `SUSPICIOUS_IP_MAX` is exceeded, it fails silently to the user (though there is debug log) and returns to homepage with HTTP 302 Found.

This PR changes its behavior to match the other 5 bot detection. They all return HTTP 429 Too Many Requests, which is clear to the user. Prior to b8c7c2c, all of them return HTTP 429.

Without this change, if users are in blocked state and type in a search term, they see the query disappears and nothing happens, causing confusion. This is also important during the initial setup where proxies incorrectly set up can trigger the bot limits. This change makes it obvious that some rate limiting applied.

## Related issues

Refer #3191 for confusion.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -175,3 +175,4 @@ features or generally made searx better:
 - Daniel Kukula `<https://github.com/dkuku>`
 - Patrick Evans `https://github.com/holysoles`
 - Daniel Mowitz `<https://daniel.mowitz.rocks>`
+- `Bearz314 <https://github.com/bearz314>`_
--- a/searx/botdetection/ip_limit.py
+++ b/searx/botdetection/ip_limit.py
@@ -122,8 +122,7 @@ def filter_request(
             redis_client, 'ip_limit.SUSPICIOUS_IP_WINDOW' + network.compressed, SUSPICIOUS_IP_WINDOW
         )
         if c > SUSPICIOUS_IP_MAX:
-            logger.error("BLOCK: too many request from %s in SUSPICIOUS_IP_WINDOW (redirect to /)", network)
-            return flask.redirect(flask.url_for('index'), code=302)
+            return too_many_requests(network, "too many request in SUSPICIOUS_IP_WINDOW (SUSPICIOUS_IP_MAX)")
 
         c = incr_sliding_window(redis_client, 'ip_limit.BURST_WINDOW' + network.compressed, BURST_WINDOW)
         if c > BURST_MAX_SUSPICIOUS:


return42:
> Prior to [b8c7c2c](https://github.com/searxng/searxng/commit/b8c7c2c9aa604fd1fb7be5559c9ad025ceb17aa4), all of them return HTTP 429.

Yes, the redirect to index is intended: 

1. if a boot has triggered the request, the bot should not be informed that it has exceeded a limit // if we tell a bot the reasons why its request was rejected, then it could train itself with this information

2. Take in mind this redirect is a part of the [link_token method](https://docs.searxng.org/src/searx.botdetection.html#method-link-token): If a user has triggered the request (if it is a real browser), then the browser will follow this redirect and subsequently also trigger the [ping request](https://docs.searxng.org/src/searx.botdetection.html#searx.botdetection.link_token.ping) of the link_token method, which takes the IP of the request from the block list.

> This change makes it obvious that some rate limiting applied.

Yes, but with the drawback that the user has to manually navigate to the index page to trigger the ping, which takes his IP from the block list.

>  They all return HTTP 429 Too Many Requests, which is clear to the user.

In sense of 1. it would better they redirect to index, but since the other bot detection methods do not offer a procedure to remove a IP from the block list, I think its better to return a 429.

return42:
Bot detection is a complex thing and always a compromise between various aspects such as user-friendliness, security, privacy ...

From #3191's point of view, the current implementation of the link_token method may lead to confusion under certain circumstances, but in the vast majority of use cases this approach is the most user-friendly.  

The possibilities of bot detection are limited as long as we don't want to use fingerprints or similar ... which we don't want because we don't want personalization ... so our bot detection can only  work at IP level and that can be irritating in some usage scenarios. (e.g. VPNs and subnets).   In #3191 I would like to capture the specifics of such scenarios in order to then (if possible) take these special scenarios into account in the implementation.

unixfox:
I'm going to share the same reasoning as https://github.com/searxng/searxng/pull/3965#issuecomment-2443787588

Hiding blocks/rate limits is just pointless because:

- people can read how our bot detection works and circumvent it. It's just easy and doesn't take much time compared to defeat a solution like cloudflare.
- nowadays it's easy to rotate through many ip addresses. Further more if the instance accept IPv6.

I think at the current state, I expect 80% of all bots to successfully bypass the current bot protections.

Now about the fact that redirecting to the home page will unblock the user because the browser will request the ping resource.

Why not having this ping resource on the 429 page too then? This way the user is informed of the block and at the same time their browser sends the request to the ping resource.

In conclusion, I value more the user experience especially for a bot protection that is now blocking just newbies bot developers. That's why I said it's just a bad idea to implement by default https://github.com/searxng/searxng/pull/3965#issuecomment-2441426436

Public instances whatever the project (invidious, searxng, piped and so on) have always gave a subpart experience of what you can have if you install the tool locally. We should focus more on the ease of installing searxng locally than trying to fix the eternal issues with bots and public instances.

This is especially going to be more the case with all the popular social networks as they are actively trying to block bots from their platform. Youtube, twitter, instagram, reddit and so on.

return42:
> Why not having this ping resource on the 429 page too then?

The bot defense provides a set of methods, one of these methods is the [`link_token` method](https://docs.searxng.org/src/searx.botdetection.html#method-link-token) and only if this method classifies a request as a bot request, it is redirected to the index page to remove the IP-block from the `link_token` method.  This unblocking function exists to revise an incorrect classification without user interaction --> user experience

The other methods do not have such a function for unblocking .. for good reason, e.g. when the Uer-Agent (see [`http_user_agent` methode](https://docs.searxng.org/src/searx.botdetection.html#method-http-user-agent)) matches to known bots, then there is no need to unblock the IP that sends such requests.

> people can read how our bot detection works and circumvent it.

It may change in the future, but right now I can see that most bots are pretty basic and rely more on rotating IPs than reengineering a bot defense.

> I think at the current state, I expect 80% of all bots to successfully bypass the current bot protections.

My public instance is massively fueled by bot requests and recognizes about 90-95% of all requests as bot requests, this is a very good rate .. TBH I have https://github.com/searxng/searxng/pull/3965 merged on my instance which does a good work for me.

bearz314:
Thank you both for the discussions.

My thought process is along the lines of it being really hard to fool bots that are specifically designed to target searxng instances, or even, those with a bit of intelligence. If a bot spams the search feature, then any replies other than the expected list of result is going to tip the bot off. For example, if the bot is redirected back to homepage:
- It can know it is being blocked.
- The search appears broken, and can invite more denial-of-service attack.

Then again, if the bot was just using a wordlist to bruteforce/scan, then they are neither encouraged nor discouraged to spam. In this sense, the error code 302 or 429 does not make too much a difference.

That said, if @return42 reports a practical, immediate benefit, the present method may still be useful, not to also mention the ability to redirect real users to load the CSS. What if we redirect to index with 302 but display a message using this in the template? The bots already know something is broken anyway, because it is not giving them the intended result.

https://github.com/searxng/searxng/blob/a1e2b254677a22f1f8968a06564661ac6203c162/searx/templates/simple/base.html#L35-L44

Many thanks.

unixfox:
@bearz314 

> My thought process is along the lines of it being really hard to fool bots that are specifically designed to target searxng instances, or even, those with a bit of intelligence. If a bot spams the search feature, then any replies other than the expected list of result is going to tip the bot off. For example, if the bot is redirected back to homepage:
> 
>     * It can know it is being blocked.
> 
>     * The search appears broken, and can invite more denial-of-service attack.
> 
> 
> Then again, if the bot was just using a wordlist to bruteforce/scan, then they are neither encouraged nor discouraged to spam. In this sense, the error code 302 or 429 does not make too much a difference.

Thank you for this, I had the same thoughts! :+1: 

And it just takes a couple of intelligent bots to ruin a whole public instance!

----------

@return42  

> My public instance is massively fueled by bot requests and recognizes about 90-95% of all requests as bot requests, this is a very good rate .. TBH I have #3965 merged on my instance which does a good work for me.

You take a lot of care with your instance because you are experienced in searxng, ip blocking, bot blocking and more. That's why you are able to block a lot of bots.

But be aware that not everyone has the same capabilities as you, so you should expect a way lower blocking percentage on other public instances. Try to think outside of your context.

If things works for you because you are more experienced in the field, then you should implement these kinds of blocks on your side. But don't try to push them for the default users without balancing the potential harm to the general user experience. Or make them configurable, exactly like I said here: https://github.com/searxng/searxng/pull/3965#issuecomment-2443877890

----------

In conclusion, I'm all for trying to be more upfront to the user in case of rate limits. If this is an issue for certain people, then let's make this configurable. Default behavior always display a clear message, and have a setting that allow to change this behavior to be vague about the error (user or bot).

Installing SearXNG wasn't meant to be a CTF challenge :smile:

return42:
> You take a lot of care with your instance because you are experienced in searxng, ip blocking, bot blocking and more. That's why you are able to block a lot of bots.

The reason I don't have as many problems is that my server is IPv4 only, which makes bot blocking much more robust.

> Installing SearXNG wasn't meant to be a CTF challenge 😄

To be clear I have a very simple installation ... only IPv4 incoming .. default SearXNG installation (via shell scripts) .. no proxy .. no caching in the network infrastructure --> 

The installation of SearXNG is simple, most problems result from the fact that some installations are done behind complex networks that make changes to IPs and requests that make a robust implementation of an IP based bot defense difficult to impossible.

@bearz314 did an excellent analysis: [Key takeaways: Check caching policies and check if configurations allow client IPs to pass in correctly --> **Link token method of bot detection requires user to request a CSS file. Depending on server/proxy/VPN/client setup, this CSS link with empty content may be cached.**](https://github.com/searxng/searxng/issues/3191#issuecomment-2648536441)

I think we can close this PR, the solution of #3191 is :

- https://github.com/searxng/searxng/pull/4303

@bearz314 thank you for your in-depth analysis and the report you have made available to the community :rocket:

bearz314:
Thank you @return42 @unixfox both as well!

Since SearXNG calls itself privacy-respecting, there will be a not insignificant number of users further hardening it with VPNs and proxies. I use CloudFlare because I have seen a drastic drop in bots and probings on my own sites.

As pointed out earlier this PR, as is, causes a regression because it breaks this:

> If a user has triggered the request (if it is a real browser), then the browser will follow this redirect and subsequently also trigger the [ping request](https://docs.searxng.org/src/searx.botdetection.html#searx.botdetection.link_token.ping) of the link_token method, which takes the IP of the request from the block list.

So I'd like to ask you the maintainers what should happen next.

- Close this PR without further changes?
- 302 redirect to index [with error messaging](https://github.com/searxng/searxng/pull/4299#issuecomment-2648257901)?
- 429 redirect to index with error messaging?

Personally I also value user interactions because it is a cat-and-mouse game with the bots. For web scraping I would use Selenium which would defeat the link_token method. For simpler Python HTTP requests, I would look at return HTTP code, page length and page content/hash; any one of those will tip me off. What it can protect well against are generic non-targeting probing bots(?) In that case, if the same homepage-plus-error is returned all the time, nothing stands out from the point of view of the bots anyway (HTTP code, length, hash).

On that note, is the adjacent PR #4303 being considered for merging? It compounds towards the UX and I argue is a different facet compared this user-facing error messaging, since in specific circumstances, #4303 will stop this problem in the first place. Thanks!

return42:
> My thought process is along the lines of it being really hard to fool bots that are specifically designed to target searxng instances, 
> I use CloudFlare because I have seen a drastic drop in bots and probings on my own sites.

The intention of the limiter is to block requests that could be identified as a bot by the engines (the services behind) ... to avoid malfunction of the SearXNG instance.  Bots that are neither recognized as such by our limiter nor by the original engine do not interfere with the operation of a SearXNG instance and do not have to be blocked.

> So I'd like to ask you the maintainers what should happen next.

I want to close *this* PR .. and review/merge #4303 ..  

> - 302 redirect to index 

.. is what we have

> - 429 redirect to index with error messaging?

.. this could be a useful change, but I would rather evaluate this in another PR ... we need to see how we can give the error message to the user without scaring him off or otherwise affecting his comfort.

> On that note, is the adjacent PR #4303 being considered for merging? It compounds towards the UX and I argue is a different facet compared this user-facing error messaging, since in specific circumstances, #4303 will stop this problem in the first place. Thanks!

Yes, unfortunately I haven't had time to gain experience with #4303 yet ... even if the code changes are marginal, I first have to test it in practice ... which is a bit more time-consuming ... maybe I can think of something else that we could use to improve it ... but I have to take a little more time for that and unfortunately I haven't had a larger time window for it yet .. be patient :-)

However, your analysis was very valuable and I thank you very much for it :+1:

return42:
> I use CloudFlare because I have seen a drastic drop in bots and probings on my own sites.

We advise against Cloudflare .. and similar services:

- Cloudflare is the source of many malfunctions: We always have bug reports here, which cost us a lot of time until it turns out that a service like Cloudflare is used and this service is responsible for the malfunction ... the error https://github.com/searxng/searxng/issues/3191 is the best example of this.

- Privacy & data protection: all of your users' data traffic runs through Cloudflare's network (and may be read there unencrypted) .. raising considerations about data privacy and compliance with regulations like GDPR or HIPAA.


- Cloudflare‘s massive scale means that its infrastructure handles a staggering amount of Internet traffic – over 25% by most estimates.

- ...

- is damn slow 

---------

### Cloudflare & Co. contradicts the expectations of SearXNG end users and counteracts all efforts we make to protect privacy.

-----------

bearz314:
Thank you very much for your detailed updates @return42 :smiley:

> I want to close this PR

Please do whenever appropriate. Thanks!!

unixfox:
I'm all for better user experience, but since @return42 has been taking the reins of the project, I'll let him decide.

return42:
> I'm all for better user experience, but since @return42 has been taking the reins of the project, I'll let him decide.

Not sure we have the same understanding, to be clear: all my development decisions are in the interest of the users! To explain: this PR is more of a disadvantage for the majority of users and possibly an advantage for a small number of users --> I have decided in favor of the majority.

We have discussed above how the intention of this PR could possibly be better implemented in a new PR to meet the needs of the smaller group of users also .. 

and what counts most: @bearz314 has implemented #4303 which solves the cause of this PR here and thus probably makes this PR superfluous.

I have rarely had such a productive discussion that serves the users in every way.

I absolutely don't understand how one can come to the conclusion that I have decided against the usability of the user!?!?

Issue #4245, opened 2025-01-28
AttributeError: 'NoneType' object has no attribute 'netloc' in hostnames plugin


qymab:
Encountered an AttributeError: 'NoneType' object has no attribute 'netloc' when using the hostnames plugin in SearxNG

```
AttributeError: 'NoneType' object has no attribute 'netloc'
ERROR:hostnames: Exception while calling on_result
Traceback (most recent call last):
  File "/usr/local/searxng/searx/plugins/_core.py", line 360, in on_result
    ret = bool(plugin.on_result(request=request, search=search, result=result))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/plugins/_core.py", line 228, in on_result
    return self.module.on_result(request, search, result)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/plugins/hostnames.py", line 147, in on_result
    if _matches_parsed_url(result, pattern):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/plugins/hostnames.py", line 142, in _matches_parsed_url
    return parsed in result and pattern.search(result[parsed].netloc)
                                               ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'netloc'
```

return42:
@qymab can you give us your settings of the plugin / thanks!

qymab:
@return42 
```

enabled_plugins:
#   # these plugins are enabled if nothing is configured ..
  - 'Basic Calculator'
  - 'Hash plugin'
  - 'Self Information'
  - 'Tracker URL remover'
  - 'Unit converter plugin'
#   - 'Ahmia blacklist'  # activation depends on outgoing.using_tor_proxy
#   # these plugins are disabled if nothing is configured ..
  - 'Hostnames plugin'  # see 'hostnames' configuration below
  - 'Open Access DOI rewrite'
#   - 'Tor check plugin'

# Configuration of the "Hostnames plugin":
#
hostnames:
  replace:
    '(.*\.)?reddit\.com$': 'old.reddit.com'
  remove:
    - '(.*\.)?facebook.com$'
  low_priority:
    - '(.*\.)?google(\..*)?$'
[...]

PR #4300, opened 2025-02-08
[fix] results.py: crashes when 'parsed_url' is None

Bnyro:
## What does this PR do?
- this fixes two possible crashes that can happen while results are being merged, that happen when the `url` of a result is `None`
- I'm not entirely sure if that's been the right place to do these checks, maybe we can already catch that somewhere earlier?

- `!tagesschau demo`

## Author's checklist
- this bug causes SearXNG to crash reproducibly and not return any results

## Related issues
related to https://github.com/searxng/searxng/issues/4245

### DIFF
--- a/searx/results.py
+++ b/searx/results.py
@@ -318,8 +318,9 @@ class ResultContainer:
     def __find_duplicated_http_result(self, result):
         result_template = result.get('template')
         for merged_result in self._merged_results:
-            if 'parsed_url' not in merged_result:
+            if not merged_result.get('parsed_url'):
                 continue
+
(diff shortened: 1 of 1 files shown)


return42 (approved):
> I'm not entirely sure if that's been the right place to do these checks, maybe we can already catch that somewhere earlier?

TLD;DR I can't think of a better place than this at the moment.
  
There are result types that must have a url and those that don't (e.g. answers don't have to have a url).  At the moment almost all results are of type LegacyResult (which nothing more than a wrapper of the dicts), where you don't know if they have to have a URL ... once we have implemented all result types that can exist and we no longer need a wrapper like LegacyResult, then we can implement the merge of the results a little more intelligently based on the typing.

PR #4305, opened 2025-02-10
Update locales.json

alexgabi:
updated Euskara (Basque) ID


return42:
This file is auto generated by https://github.com/searxng/searxng/blob/master/searxng_extra/update/update_locales.py

The `eu` translation is here: https://translate.codeberg.org/projects/searxng/searxng/eu/

PR #4314, opened 2025-02-12
[refactor] engines: outsource cache implementation from duckduckgo engine source

Bnyro:
## What does this PR do?
- This PR moves the caching logic from Redis into a separate file
- That allows us to re-use it for other engines in the future, i.e. for the `Public domain image archive` engine

- It probably makes sense to provide an easy to use api to cache values to engine developers

- use the duckduckgo engine, once with Redis configured and once without

## Author's checklist
- there's currently some issue in `duckduckgo.py` that initializing the `__CACHE` doesn't work properly - it should certainly be done in the `init(_)` method because `get_or_create_cache()` needs to be called **after** the Redis database has been initialized (if it's available)

### DIFF
--- a/searx/engine_cache.py
+++ b/searx/engine_cache.py
@@ -1,0 +1,69 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""This provides an easy to use interface for engine implementations to store and read key-value pairs.
+
+For instance, this can be used to remember programmatically extracted API keys or other kinds of secret tokens.
+"""
+
+from typing import Optional
+from searx import redisdb, redislib
+
+
+class EngineCache:
+    def store(self, key: str, value: str):
+        pass
+
+    def get(self, key: str) -> Optional[str]:
+        pass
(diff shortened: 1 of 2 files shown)


return42:
I had been planning for some time to develop a simple, reusable solution for exactly this problem. However, I wanted to avoid storing values in memory (here, for example, if redis or valkey is not installed). If values are stored in memory, they are not always available in all threads and processes and have to be generated several times (depending on the WEB configuration of the workers and threads).  This can have negative effects, e.g. if the token for a search engine has to be determined several times .. which may be considered untipical behavior and puts your SearXNG's IP on the block list of the origin list.

We already have a convenient integration for DB applications based on SQLite that is available as a builtin in all Python installations.  Based on this, I have now developed an alternative to your PR:

- https://github.com/searxng/searxng/pull/4650

Bnyro:
replaced by https://github.com/searxng/searxng/pull/4974

PR #4318, opened 2025-02-13
[refactor] search.js: use custom auto completion implementation

Bnyro:
## What does this PR do?
- refactor: use custom auto completion implementation

- the previously used library is unmaintained for 6 years now, see https://github.com/searxng/searxng/pull/4284#discussion_r1954493434

- enable an autocompletion provider
- test different HTTP methods from the privacy settings
- test selecting suggestions with the arrow keys
- test pressing enter on the search form
- test inserting garbage queries that don't return any autocompletion suggest

## Related issues
see https://github.com/searxng/searxng/pull/4284#discussion_r1954493434


return42:
Old autocomplete can be removed:

https://github.com/searxng/searxng/blob/d37dc307831ef0f16a6cb2fbb0c5afe9bb5e4258/searx/static/themes/simple/gruntfile.js#L112

https://github.com/searxng/searxng/blob/d37dc307831ef0f16a6cb2fbb0c5afe9bb5e4258/searx/static/themes/simple/package.json#L24

But first we should merge:

- https://github.com/searxng/searxng/pull/4284

.. with the new environment (based on vite) we will have more/better methods for FE development.

.. and we should take in mind there are open tasks --> https://github.com/searxng/searxng/issues?q=label%3A%22area%3A%20autocomplete%22%20state%3Aopen

searx/static/themes/simple/src/js/main/search.js
```diff
@@ -35,118 +32,105 @@
...
-            var form = item.closest('form');
-            if (form) {
-              form.submit();
+        searxng.on(qinput, 'input', () => {
+          const query = qinput.value;
+          if (query.length < searxng.settings.autocomplete_min) return;
+
+          setTimeout(() => {
+            if (query == qinput.value) fetchResults(query);
+          }, 300);
```
vojkovic: If you don't mind me asking, how come you are adding a delay to the autocomplete? It seems to make the user experience of using autocomplete much worse.

With autocomplete.js, there was a bug where if the previous request hadn't finished, the next request would cancel and never get sent (so a delay of 300ms was added which would make the issue less apparent).

Let me know your thoughts.

Bnyro: > If you don't mind me asking, how come you are adding a delay to the autocomplete? It seems to make the user experience of using autocomplete much worse.
>
The goal of this PR is to rebuild the autocomplete behavior of autocomplete.js without any changes, thus I kept the delay.

> With autocomplete.js, there was a bug where if the previous request hadn't finished, the next request would cancel and never get sent (so a delay of 300ms was added which would make the issue less apparent).
>
Interesting, I wasn't aware of that.

Nevertheless I think such behavior changes should be discussed in an other issue, as I said the only goal of this PR is to remove the very outdated, unmaintained `autocomplete.js` dependency without changing any behavior.

vojkovic:
Had an issue reported to me about a few bugs in the mobile version of the autocompleter.

First is a bug where the autocomplete result needs to be selected twice before it runs the search query, and the second is a ui bug where the autocomplete result box width goes past the screen border.

A few screenshots to demonstrate:





To replicate, run the autocomplete on the main page of searxng in the mobile view with browser devtools.

vojkovic:
Also was reported by a user in https://github.com/privau/searxng/issues/56 that the search box isn't being focused when the page loads.

return42:
@Bnyro should we proceed with this PR?

Bnyro:
> @Bnyro should we proceed with this PR?
>
Yes, but I'm on vacation over the weekend and probably won't have much time during next week, so it can take a while until I can look into the issues mentioned.

Bnyro:
I've addressed the overflow issues from screenshot 2, the double click issue on mobile and the automatic search bar focus on large screens.

>  
> 
This issue happens on the current master branch too, so it's unrelated to this PR.

vojkovic:
Looks like the search box is now in the middle D:

middle

Also ensure sure that the searchbox auto-focusing only happens on the frontpage so that it doesn't autofocus on the results pages.

return42:
FYI: last force push was just a rebase on current master ..

return42 (approved):
I slightly modified the commit message / added ... to the body:

```
The previously used library is unmaintained for 6 years now [1] and the solution
had know issues [2][3]

[1] https://github.com/searxng/searxng/pull/4284#discussion_r1954493434
[2] https://github.com/searxng/searxng/pull/4318#issuecomment-2731576657
[3] https://github.com/privau/searxng/issues/56
```

and I pushed the commit from `make static.build.commit` on top.

@vojkovic @Bnyro from my side this PR is ready to merge / do you have any open questions?

vojkovic:
Looks good on my end - no questions from me.

PR #4371, opened 2025-02-23
[fix] various issues in the documentation

### LINKED ISSUE #4370: Documenttion says key is called `engine` while default configuration file uses `engines`
2025.2.17+cc9dbde2e

Installation using the official docker image.

**What happened?**
The initial configurtion was generated with a key called `engines:` under which all the engines are configured. However [the documentation](https://docs.searxng.org/admin/settings/settings_engine.html) says it is called `engine:` (without the s), which is slightly confusing.

return42:
- closes: https://github.com/searxng/searxng/issues/4370

### DIFF
--- a/docs/admin/settings/index.rst
+++ b/docs/admin/settings/index.rst
@@ -13,7 +13,7 @@ Settings
    :maxdepth: 2
 
    settings
-   settings_engine
+   settings_engines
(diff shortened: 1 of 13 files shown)


slatian:
Huge Thank You!

PR #4378, opened 2025-02-26
Updated health check in Dockerfile

thegabriele97:
## What does this PR do?

Updated health check in Dockerfile. It now points to 127.0.0.1 instead of localhost.

It doesn't always work. For example, localhost may not be defined.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,7 @@ RUN su searxng -c "/usr/bin/python3 -m compileall -q searx" \
     -o -name '*.svg' -o -name '*.ttf' -o -name '*.eot' \) \
     -type f -exec gzip -9 -k {} \+ -exec brotli --best {} \+
 
-HEALTHCHECK CMD wget --quiet --tries=1 --spider http://localhost:8080/healthz || exit 1
+HEALTHCHECK CMD wget --quiet --tries=1 --spider http://127.0.0.1:8080/healthz || exit 1
 
 # Keep these arguments at the end to prevent redundant layer rebuilds
 ARG LABEL_DATE=


unixfox:
How comes "localhost" does not exist?

Yes it exists. The reason it didn't work is that, wget tries to initiate the connection in IPv6 only. But searxng probably does not listen to ipv6 but instead to ipv4 only.

It is easily reproducible by entering the command in a running container:
```
/usr/local/searxng # wget --quiet --tries=1 --spider http://127.0.0.1:8080/healthz
/usr/local/searxng # echo $?
0
/usr/local/searxng # wget --quiet --tries=1 --spider http://localhost:8080/healthz
wget: can't connect to remote host: Connection refused
/usr/local/searxng # echo $?
1
```

I think the PR was merged a bit too quickly and does not solve the original issue. Is that searxng should listen to "::" and not "0.0.0.0". This way it listen to both ipv4 and ipv6.

unixfox:
This: https://github.com/searxng/searxng/blob/1fd4dab3b41bd45da12a66677ed94ae438d27912/dockerfiles/docker-entrypoint.sh#L26

Should be renamed from `0.0.0.0:8080` to `[::]:8080`

If I override BIND_ADDRESS like this: `docker run -e BIND_ADDRESS="[::]:8080" searxng/searxng` then this works:
```
/usr/local/searxng # wget --quiet --tries=1 --spider http://localhost:8080/healthz
/usr/local/searxng # echo $?
0
/usr/local/searxng # wget --quiet --tries=1 --spider http://127.0.0.1:8080/healthz
/usr/local/searxng # echo $?
0
```

return42:
> I think the PR was merged a bit too quickly and does not solve the original issue. 

Oh yes, I'm very sorry for the mess .. I send:

- https://github.com/searxng/searxng/pull/4380

to revert both PRs .. @unixfox if you think its good to revert and give it a new try in a new PR then feel free to merge my #4380 .. 

I think it's better if I keep my hands off the docker stuff in the future ... I have no idea about docker and can only merge blindly .. again: I'm very sorry, thanks for understanding.

unixfox:
> I think it's better if I keep my hands off the docker stuff in the future ... I have no idea about docker and can only merge blindly .. again: I'm very sorry, thanks for understanding.

I don't think it's a good idea to avoid Docker. Our main user base uses Docker, and it's important to understand what most of our users are using. Sometimes it's possible to make changes that impact Docker without even noticing.

What was wrong here is to immediately merge the PR without at least ask for the original author ([vojkovic](https://github.com/vojkovic)) of the change to review it.

vojkovic:
@unixfox @return42 

Ah interesting, this case didn't come up during my testing in the devcontainer so I guessed it would failover to ipv4 if there wasn't any v6 connectivity - Apparently not!

The issue can be solved by failing over to v4 if v6 fails, wget supports the `-4` and `-6` flags to only use one connectivity option at a time. The command we're looking for is:

```HEALTHCHECK CMD wget --quiet --tries=1 --spider -6 http://localhost:8080/healthz || wget --quiet --tries=1 --spider -4 http://localhost:8080/healthz || exit 1```

Not sure if you want to revert the commit first because I see https://github.com/searxng/searxng/pull/4380 is open. Let me know if you want another PR from me or if you can do it yourself.

return42:
> Not sure if you want to revert the commit first because I see https://github.com/searxng/searxng/pull/4380 is open.

I now merged the reverts .. This should restore the original functionality of the docker tasks.

> Let me know if you want another PR from me or if you can do it yourself.

I can't help on this topic .. ping @unixfox .. question goes to you.

unixfox:
> Ah interesting, this case didn't come up during my testing in the devcontainer so I guessed it would failover to ipv4 if there wasn't any v6 connectivity - Apparently not!

`curl` do follow [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs) but not wget.

> The issue can be solved by failing over to v4 if v6 fails, wget supports the `-4` and `-6` flags to only use one connectivity option at a time. The command we're looking for is:

IMO this is the same as writing "http://127.0.0.1", just a bit more complicated :smile:.

> Not sure if you want to revert the commit first because I see #4380 is open. Let me know if you want another PR from me or if you can do it yourself.

What I would suggest is to do adapt all cases where `0.0.0.0` is set and instead replace it with `[::]`. It's 2025 and we should support IPv6.

Python has support for IPv6 since most likely its creation, so there is no point in listening to ipv4 only in the container.

Then we can add the new change in order to do a healthcheck on http://localhost.

PR #4382, opened 2025-02-26
[engine] elasticsearch: add pagination support

Bnyro:
## What does this PR do?
- add support for pagination
- add some missing about information

## Why is this change important?
- this improves the user experience when using elasticsearch

## How to test this PR locally?
```
services:
  elasticsearch:
    image: elasticsearch:8.8.0
    ports:
      - 9200:9200
      - 9300:9300
    environment:
      - discovery.type=single-node
      - xpack.security.enabled=false
```
- Create an index: `curl -X PUT http://localhost:9200/my-index`
- Create items: `curl -X POST -H 'Content-Type: application/json' -d '{ "name": "Awesome T-Shirt", "description": "This is an awesome t-shirt for casual wear.", "price": 19.99, "category": "Clothing", "brand": "Example Brand" }' http://localhost:9200/my-index/_doc`
- Enable elasticsearch in `settings.yml` by uncommenting it

## Related issues
closes https://github.com/searxng/searxng/issues/4346

PR #4383, opened 2025-02-26
[fix] engines: key-value result templates show no results and badly formatted

Bnyro:
## What does this PR do?
- this fixes that templates without url don't work

- key value templates currently don't work because they don't have a URL

- see https://github.com/searxng/searxng/pull/4382

## Author's checklist
- I introduced this bug recently at https://github.com/searxng/searxng/pull/4300/files, so we will likely need another solution to fix that

### DIFF
--- a/client/simple/src/less/style.less
+++ b/client/simple/src/less/style.less
@@ -1050,6 +1050,12 @@ summary.title {
   }
 }
 
+#main_results .key-value-table {
+  td {
+    padding: 0.3rem 0.5rem;
+  }
+}
+
 /*
   phone layout
 */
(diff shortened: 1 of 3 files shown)


return42:
FYI: last force push was just to rebase on the new vite environment, nothing has changed from my side.  I will now review this PR ..

return42:
> so we will likely need another solution to fix that

Yes, we are currently implementing the typing of the results for this purpose.

I have now implemented the second part of the typing in PR ...

- https://github.com/searxng/searxng/pull/4424

The PR also includes (in separate commits) the result type `KeyValue` and a migration of the engines that previously used the key-value.html template

return42:
Superseeded by https://github.com/searxng/searxng/pull/4424

PR #4384, opened 2025-02-26
[feat] engines: add openclipart.org

Bnyro:
## What does this PR do?
- this PR adds support for openclipart.org

- !ocl europe
- !ocl linux

## Author's checklist
- the engine sometimes doesn't return results for the first few pages, that's however the same at openclipart.org

### DIFF
--- a/searx/engines/openclipart.py
+++ b/searx/engines/openclipart.py
@@ -1,0 +1,47 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""OpenClipArt (images)"""
+
+from urllib.parse import urlencode
+from lxml import html
+from searx.utils import extract_text, eval_xpath, eval_xpath_list
+
+about = {
+    "website": 'https://openclipart.org/',
+    "wikidata_id": 'Q979593',
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'HTML',
+}
+
+categories = ['images']
+paging = True
+
+base_url = "https://openclipart.org"
+
+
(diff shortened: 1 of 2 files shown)


return42 (approved):
> the engine sometimes doesn't return results for the first few pages, that's however the same at openclipart.org

It returns, but very very slow .. I had to increase the timeout to 30sec and since openclipart.org is such slow in responding (and the images itself are taking minutes to load) I also set this engine to inactive --> [diff](https://github.com/searxng/searxng/compare/a43af91fde2c32db7e23c7e5e3358179f2d5ffc3..bc0811dbccaa01b3d2b7213f47018d8d7a9ee87f)

Bnyro:
> It returns, but very very slow .. I had to increase the timeout to 30sec and since openclipart.org is such slow in responding (and the images itself are taking minutes to load) I also set this engine to inactive --> [diff](https://github.com/searxng/searxng/compare/a43af91fde2c32db7e23c7e5e3358179f2d5ffc3..bc0811dbccaa01b3d2b7213f47018d8d7a9ee87f)
>
Weird, I got some very decent results within a second when I tested it :shrug:

return42:
> Weird, I got some very decent results within a second when I tested it 🤷

Maybe it was just a snapshot of me where the response times were extremely long.  What I noticed: if you keep the search term you get a faster response (cache?) ... when I changed the search terms, the response times became extremely long.  You can observe this in your instance, if you have good experiences with shorter timeouts, then you can create another PR and adjust the values.

Bnyro:
Sure, will test it a bit 👍

PR #4385, opened 2025-02-26
[feat] Keyboard/tab-navigation accessible preferences and index

Denperidge:
## What does this PR do?

The comments on the commits are descriptive, but a quick rundown:
- No more outline: none, but a stylised outline instead
- The clear search button is now hidden from keyboard/tab navigation when search is empty
- You can now set preferences using the tab & enter keys, yay!

Improved user experience for people using keyboard navigation

Clone, run, and try using tab navigation & the enter button in search & preferences. Interactive elements are now clearly in focus, invisible elements (like the clear button in an empty search) are now skipped. Also, you can now do preferences/default categories/choose the tab/use the checkboxes using tab & enter, which wasn't possible before!

## Author's checklist

Let me know if there's any issues!

## Related issues

Progresses #818


return42:
@mrpaulblack, @dalf, @Bnyro .. if you have time for, could you have a look at this PR and share your opinion .. 

---------

@Denperidge thanks a lot for your work on this topic :+1: .. I rebased PR on current master to solve conflicts ..


> No more outline: none, but a stylised outline instead

Perhaps they were there for aesthetic reasons. Here's an example where it's a bit of a hitch:

grafik

Although I prefer keyboard navigation myself, I never use it on web pages, there are simply too many elements to select ... and when the desired field is reached after a hundred TABs, it often cannot be activated ... 

Whatever my personal opinion is, I think the PR basically enables keyboard navigation.

But keyboard navigation and aesthetics are also partly contradictory and I remember that @mrpaulblack and @dalf paid a lot of attention to aesthetics.

Bnyro (commented):
I'm not really sure about the outline in general, especially when the search bar is focused.

The outline doesn't only appear when you navigate via keyboard which would be fine, it also appears when the search field is selected via mouse and used normally. I think it's quite useless and feels a bit awkward in that case.

I haven't tested it, but maybe the [`focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) pseudo-class could be useful for that?

return42:
Closed due to lack of feedback .. we can reopen when you have time / maybe the keyboard navigation can be improved a bit with @Bnyro's suggestion, but I have doubts that keyboard navigation can be implemented in the preferences in such a way that it creates added value for users.

However .. if you have time and interest we can reopen the PR.

PR #4413, opened 2025-03-03
[feat] add tls support to the backend searxng flask application

AeshEmi1:
## What does this PR do?

- Adds support for TLS on the SearXNG flask application in webapp.py.
- Enables https-socket support for uwsgi in docker-entrypoint.sh.
- Adds three new configuration options to searx's settings.yml (enable_tls, certificate_path, and certificate_key_path).
- Updates the settings_server documentation to provide information on enable_tls, certificate_path, and certificate_key_path.

Currently, the SearXNG Flask application and uWSGI implementation do not support HTTPS. This presents a significant security and privacy vulnerability, particularly in potentially hostile environments like cloud platforms. While a reverse proxy (like Caddy) might handle SSL termination for external users, the internal traffic between the proxy and the SearXNG backend remains unencrypted.

In the event that an adversary obtains access to the machine where SearXNG is hosted, either through a vulnerability or insider attack (particularly in cloud environments), they will be able to read and intercept every search made to SearXNG in plaintext. One might argue that if an adversary gets access to the machine where SearXNG is hosted, it's game over regardless, since they could just change the configuration to use HTTP for the backend again. However, it's much easier to detect and audit for configuration changes than it is to detect passive traffic sniffing.

This change is not forcing the use of TLS in the backend (in fact it is disabled by default); it simply adds support for those who wish to enable it.

Additionally, this change will make searches more private, reinforcing SearXNG's privacy-by-design approach.

### To test HTTPS support
1. Generate an openssl private key and certificate.
2. Copy the private key to /etc/searxng/certs/searxng.key
3. Copy the generated certificate to /etc/searxng/certs/searxng.crt
4. Set enable_tls to true in the searx settings.yml file.
5. Run SearXNG.

### To test HTTP support
1. Set enable_tls to false in the searx settings.yml file.
2. Run SearXNG.

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -54,7 +54,26 @@ get_searxng_version(){
        2>/dev/null
 }
 
+# For TLS support
+get_searxng_tls_status(){
+    su searxng -c \
+       "python3 -c \"import six; from searx import settings; six.print_(settings['server']['enable_tls'])\"" \
+       2>/dev/null
+}
+
+get_searxng_tls_cert(){
+    su searxng -c \
+       "python3 -c \"import six; from searx import settings; from os.path import join; six.print_(join('/etc/searxng/', settings['server']['certificate_path']))\"" 
+}
+
+get_searxng_tls_key(){
+    su searxng -c \
+       "python3 -c \"import six; from searx import settings; from os.path import join; six.print_(join('/etc/searxng/', settings['server']['certificate_key_path']))\"" \
+       2>/dev/null
+}
+
 SEARXNG_VERSION="$(get_searxng_version)"
+SEARXNG_TLS_STATUS="$(get_searxng_tls_status)"
 export SEARXNG_VERSION
 echo "SearXNG version ${SEARXNG_VERSION}"
 
@@ -175,4 +194,11 @@ unset MORTY_KEY
 
 # Start uwsgi
 printf 'Listen on %s\n' "${BIND_ADDRESS}"
-exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+# If server.enable_tls is True, enable TLS on searxng
+if [ "${SEARXNG_TLS_STATUS}" = "True" ]; then
+    SEARXNG_TLS_CERT="$(get_searxng_tls_cert)"
+    SEARXNG_TLS_KEY="$(get_searxng_tls_key)"
+    exec uwsgi --master --uid searxng --gid searxng --https-socket "${BIND_ADDRESS}","${SEARXNG_TLS_CERT}","${SEARXNG_TLS_KEY}" "${UWSGI_SETTINGS_PATH}"
+else
+    exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+fi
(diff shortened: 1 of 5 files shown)


unixfox:
I'm against this. Let's not complicate the HTTP part of the backend.

If you want to secure a connection between searxng with a native way, you can use unix socket for that. No traffic will go through a network card, just directly to the program.

Second if you can't do that, there are many others solutions like service mesh or having an encrypted tunnel like wireguard if searxng is not on the same machine as your reverse proxy.

Or if it's a separate machine just put a reverse proxy with TLS on the same machine as searxng and sends the traffic over an unix socket. Have your main reverse proxy sending the traffic to that reverse proxy using TLS.

In my opinion, if TLS is a requirement for you for the traffic between searxng and the reverse proxy, there are many ways to achieve that without bloating the HTTP part of searxng.

AeshEmi1:
I disagree that using a UNIX socket will secure the connection between the reverse proxy and searxng. Even if we switch over to using unix sockets for the backend, the traffic will still be sent in HTTP. Tools already exist to passively sniff sockets (https://github.com/mechpen/sockdump). This would be no different than passively sniffing on lo0, which is the issue I'm trying to solve.

Additionally, sticking with the socket idea, a TLS enabled socket will secure the communication between the rprxy and backend. However, this will create more bloat than enabling TLS directly on the flask application and uWSGI implementations themselves (https://en.ittrip.xyz/python/secure-socket-with-ssl-tls-python). We'd have to write code to create the socket, wrap it with TLS support, and then have both the flask application and uWSGI implementations interface with that socket. The current solution I have proposed directly enables TLS on the flask application and uWSGI, creating less bloat.

I'm confused by "if TLS is a requirement for you for the traffic between searxng and the reverse proxy, there are many ways to achieve that without bloating the HTTP part of searxng." The commits that I have requested to be pushed are adding less than 70 lines of code. Could you please elaborate on what is bloated about how my change is designed?

Also, I'm open to more ideas, but strongly disagree that using UNIX sockets are a more secure or efficient way of securing the connection between the rprxy and backend searxng application.

return42:
> The commits that I have requested to be pushed are adding less than 70 lines of code.

To incorporate this, a little more needs to be taken into account than what the patch does here.

> However, this will create more bloat than enabling TLS directly on the flask application 

Furthermore, I think that we are here in the SearXNG core and topics such as certification belong to the "deployment" and not here in the core development.

In short: I understand the motivation, but I do more agree with @unixfox's  https://github.com/searxng/searxng/pull/4413#issuecomment-2693491362

I close this issue for now .. I think we won't this in SearXNG .. but this does not mean I want to stop the discussion ..

unixfox:
If your unix socket is readable/writable by only one dedicated user (for your reverse proxy and searxng), in order to use a tool like sockdump, you need to either:
- run sockdump as the same user
  - you have to exploit either searxng or the reverse proxy.
    if someone exploit searxng or the reverse proxy, they can easily inject some code into the running process and inspect the traffic.
    for the reverse proxy, they could hot reload some changes in order to inspect the traffic
- run sockdump using root privileges
  - in which case, it's also possible to transparently decrypt the TLS connections.  
    one example tool: https://github.com/gojue/ecapture

IMO if you want to restrict the ability to inspect the traffic without exploiting searxng and reverse proxy, make the system read only where searxng is running. Remove as much attack surface like having no shell in the system (can be done in container).
The point is to make it hard for an attacker to try to inspect the traffic.

AeshEmi1:
TLS support is a requirement for me so it's unfortunate to see this PR be closed. I'm disappointed to see such basic functionality be dismissed.

For those who are looking for this feature and don't want to overcomplicate securing your backend connections, I'm maintaining a [fork](https://github.com/aeshemi1/searxng-se/) of SearXNG that implements this functionality.

PR #4420, opened 2025-03-04
[feat] engines: add www.chinaso.com

hezhijie0327:
## What does this PR do?

1. add ChinaSo (中国搜索) engine for SearXNG (News, Videos, Images)

ref: https://www.chinaso.com/

Wikidata: https://www.wikidata.org/entity/Q17164385

|Function|Screenshot|
|-|-|
|News|image|
|Videos|image|
|Images|image|

### DIFF
--- a/searx/engines/chinaso.py
+++ b/searx/engines/chinaso.py
@@ -1,0 +1,142 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""ChinaSo: A search engine from ChinaSo."""
+
+from urllib.parse import urlencode
+from datetime import datetime
+
+from searx.exceptions import SearxEngineAPIException
+from searx.utils import html_to_text
+
+about = {
+    "website": "https://www.chinaso.com/",
+    "wikidata_id": "Q10846064",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+paging = True
+time_range_support = True
+results_per_page = 10
+categories = []
+chinaso_category = 'news'
+"""ChinaSo supports news, videos, images search.
+
+- ``news``: search for news
+- ``videos``: search for videos
+- ``images``: search for images
+"""
+
+time_range_dict = {'day': '24h', 'week': '1w', 'month': '1m', 'year': '1y'}
+
+base_url = "https://www.chinaso.com"
+
+
+def init(_):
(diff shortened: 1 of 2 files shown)


searx/engines/chinaso.py
```diff
@@ -0,0 +1,143 @@
...
+    category_config = {
+        'news': {'endpoint': '/v5/general/v1/web/search', 'params': {'pn': params["pageno"], 'ps': 10}},
+        'images': {
+            'endpoint': '/v5/general/v1/search/image',
+            'params': {'start_index': (params["pageno"] - 1) * 10, 'rn': 10},
+        },
+        'videos': {
+            'endpoint': '/v5/general/v1/search/video',
+            'params': {'start_index': (params["pageno"] - 1) * 10, 'rn': 10},
+        },
```
Bnyro: You should probably use the `results_per_page` variable instead of `10` here?

searx/engines/chinaso.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if chinaso_category not in parsers:
+        raise SearxEngineAPIException(f"Unsupported category: {chinaso_category}")
+
+    return parsers[chinaso_category](data)
+
+
+def parse_news(data):
+    results = []
+    if "data" not in data or "data" not in data["data"]:
```
Bnyro:
```suggestion
    if not data.get("data", {}).get("data"):
```

searx/engines/chinaso.py
```diff
@@ -0,0 +1,143 @@
...
+            'endpoint': '/v5/general/v1/search/image',
+            'params': {'start_index': (params["pageno"] - 1) * 10, 'rn': 10},
+        },
+        'videos': {
+            'endpoint': '/v5/general/v1/search/video',
+            'params': {'start_index': (params["pageno"] - 1) * 10, 'rn': 10},
+        },
+    }
+
+    if chinaso_category not in category_config:
```
Bnyro: The check here and the one in line 72 should be moved to the `init` function which runs when SearXNG is started:
```python
def init(_):
    if chinaso_category not in parsers:
        raise SearxEngineAPIException(f"Unsupported category: {chinaso_category}")
` ``

return42 (approved):
Nice work .. thanks a lot :+1:

PR #4423, opened 2025-03-05
[feat] engines: add www.acfun.cn

hezhijie0327:
## What does this PR do?

1. add Acfun engine for SearXNG (Videos)

ref: https://www.acfun.cn/
Wikidata: https://www.wikidata.org/wiki/Q3077675

|Function|Screenshot|
|-|-|
|Videos|image|

### DIFF
--- a/searx/engines/acfun.py
+++ b/searx/engines/acfun.py
@@ -1,0 +1,108 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Acfun search engine for searxng"""
+
+from urllib.parse import urlencode
+import re
+import json
+from datetime import datetime, timedelta
+from lxml import html
+
+from searx.utils import extract_text
+
+# Metadata
+about = {
+    "website": "https://www.acfun.cn/",
+    "wikidata_id": "Q3077675",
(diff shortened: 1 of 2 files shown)


searx/engines/acfun.py
```diff
@@ -0,0 +1,108 @@
...
+        video_data = json.loads(data_exposure_log)
+
+        content_id = video_data.get("content_id", "")
+        title = video_data.get("title", "")
+
+        url = f"{base_url}/v/ac{content_id}"
+        iframe_src = f"{base_url}/player/ac{content_id}"
+
+        create_time = extract_text(video_block.xpath('.//span[contains(@class, "info__create-time")]'))
+        video_cover = extract_text(video_block.xpath('.//div[contains(@class, "video__cover")]/a/img/@src')[0])
```
return42, line 78: I was hoping there would be a real thumb ... but acfun itself uses large JPG to scale them to the size of a thumb ... that also explains why the pages from acfun are so slow and resource consuming :-)

return42 (approved):
perfect .. thanks :+1:

PR #4432, opened 2025-03-06
[feat] add bilibili support to get_embeded_stream_url

Austin-Olacsi:
## What does this PR do?

Explained in the title.

Improvement for users.

Two types of URLS are handled. Test for both:

!360sov shanghai teahouse
!ddv site:bilibili.com shanghai teahouse

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -633,7 +633,7 @@ def _get_fasttext_model() -> "fasttext.FastText._FastText":  # type: ignore
 def get_embeded_stream_url(url):
     """
     Converts a standard video URL into its embed format. Supported services include Youtube,
-    Facebook, Instagram, TikTok, and Dailymotion.
+    Facebook, Instagram, TikTok, Dailymotion, and Bilibili.
     """
     parsed_url = urlparse(url)
     iframe_src = None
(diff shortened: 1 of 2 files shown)


searx/utils.py
```diff
@@ -673,6 +673,22 @@ def get_embeded_stream_url(url):
             video_id = path_parts[2]
             iframe_src = 'https://www.dailymotion.com/embed/video/' + video_id
 
+    # Bilibili
+    elif parsed_url.netloc in ['www.bilibili.com', 'bilibili.com'] and parsed_url.path.startswith('/video/'):
+        path_parts = parsed_url.path.rstrip('/').split('/')
```
Bnyro: `rstrip('/')` is completely unnecessary here, `split` starts from the left anyways and removes all occurrences of `/`, so what's at the end doesn't matter

searx/utils.py
```diff
@@ -673,6 +673,22 @@ def get_embeded_stream_url(url):
...
+
+        video_id = path_parts[2]
+        param_key = None
+        if video_id.startswith('av'):
+            video_id = video_id[2:]
+            param_key = 'aid'
+        elif video_id.startswith('BV'):
+            param_key = 'bvid'
```
Bnyro, line 686: I haven't found any video with a `BV` videoId, thus the question: Is it intentional to not remove the `BV` prefix from the videoId here?

Austin-Olacsi: Yes that was intentional, we need to keep the BV prefix. If you try the duckduckgo videos engine with site:bilibili.com, then you should find a lot of videos that use a BV videoid.

Bnyro (approved):

Issue #4415, opened 2025-03-03
The latest version shows an error


chung1912:
![Image](https://github.com/user-attachments/assets/458e5f19-f5c3-45f3-bdee-62696b933976)

return42:
Delete Browser's cache.. see [label:"clear browser cache"](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

dalf:
A long time ago, there was this idea to fix this issue once for all: https://github.com/searxng/searxng/pull/932#issuecomment-1067216426

return42:
> A long time ago, there was this idea to fix this issue once for all: [#932 (comment)](https://github.com/searxng/searxng/pull/932#issuecomment-1067216426)

Thanks for reminder, I send PR:

- https://github.com/searxng/searxng/pull/4433

PR #4433, opened 2025-03-06
[mod] enable CSS & JS cache busting by default

return42:
The static files (CSS & JS) are cached in the browser for one year.  If we have changes to the CSS or JS files, these are not yet transported to the client by default, which causes problems and irritations on the client side:

- see issues & discussions with [`label:"clear browser cache"`](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

Related:

- https://github.com/searxng/searxng/issues/4415#issuecomment-2702917901
- https://github.com/searxng/searxng/pull/932

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -118,7 +118,7 @@ ui:
   # Custom static path - leave it blank if you didn't change
   static_path: ""
   # Is overwritten by ${SEARXNG_STATIC_USE_HASH}.
-  static_use_hash: false
+  static_use_hash: true
   # Custom templates path - leave it blank if you didn't change
   templates_path: ""
   # query_in_title: When true, the result page's titles contains the query


dalf:
As far I remember, and I can't remember why, this can't be enabled outside the docker image.

return42:
> As far I remember, and I can't remember why, this can't be enabled outside the docker image.

Its a long time ago .. can't remember last state of this topic, but may you refer to this:

- https://github.com/searxng/searxng/issues/964

return42:
For a interim solution I send PR:

- https://github.com/searxng/searxng/pull/4436

return42:
Obsolete since:

- https://github.com/searxng/searxng/pull/5004

Issue #4425, opened 2025-03-05
Category sizes are inconsistent and engines are formatted vertically


martyfuhry:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

It says "Powered by searxng 2025.3.2+8984d7ae0"

**How did you install SearXNG?**
docker compose

**What happened?**
The category sizes are different in the search results page:

![Image](https://github.com/user-attachments/assets/bbd675ea-fc9b-4f17-bcfd-cb134a15c06c)

And underneath each article, the engines are replaced by vertically stacked circles. This makes the search results look oddly tall with fewer per-page than usual.

![Image](https://github.com/user-attachments/assets/ab65b124-1eb4-4c17-8ada-8f47be6ba37f)

In mobile:

![Image](https://github.com/user-attachments/assets/128d4b64-d177-46d8-9773-d0ef658da0ea)

**How To Reproduce**
Search for anything.

**Expected behavior**
I would prefer if the categories were all the same size. I do not know what the three little circles are supposed to be, so I would prefer if they were hidden and the search results to look less tall.

**Screenshots & Logs**
See above.

**Additional context**
Thank you for all your hard work! I use searxng every single day.

pixelpassport:
I was also having the 3 circle issue in librewolf after updating my docker container, it seems to be fixed by restarting the browser. I had the same issue on fennec for android and force closing and clearing the cache fixed the issue for me.

return42:
It would be helpful if you looked around before creating new error entries .. I don't know how often I have to  answer this question again:

-> duplicate https://github.com/searxng/searxng/issues/4415

delete browser's cache ..

Issue #4419, opened 2025-03-04
Several Icons broken in Search Results


Haepaxlog:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
2025.3.2+8984d7ae0

**How did you install SearXNG?**
searxng-docker

**What happened?**
On the search results page it appears that several icons are not used correctly or sized wrongly.

**Expected behavior**
Correctly sized icons and no ellipsis-vertical.

**Screenshots & Logs**

![Image](https://github.com/user-attachments/assets/58a2a70b-de8e-474c-8426-afc9de315196)

https://github.com/user-attachments/assets/f38116e9-f5cf-4665-a397-35153dda6acf

**Additional context**
This seems to be the cause of it https://github.com/searxng/searxng/commit/189dd0155df60c98b5d7de73b873c8ad444ff931, because this line appears to trigger this behaviour for at least the ellipsis-vertical.  https://github.com/searxng/searxng/blob/8984d7ae022e4e28a9ecb6032c5c535d6ebee520/searx/templates/simple/macros.html#L50

return42:
duplicate https://github.com/searxng/searxng/issues/4415 .. see [label:"clear browser cache"](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

delete browser's cache..

PR #4436, opened 2025-03-06
[fix] uwsgi: don't set static-expires

return42:
As long we don't have a working solution for cache busting [3][4] we should not set an expire time in the uWSGI config.

If we have changes to the CSS or JS files, these are not yet transported to the client by default, which causes problems and irritations on the client side [3]:

- see issues & discussions with [`label:"clear browser cache"`](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

The default procedure in every web browser is the "304 Not Modified" [2] and this default procedure should also be sufficient for us as long as we have not implemented a complete alternative (cache busting) / form [1]

> By default uWSGI will add a Last-Modified [2] header to all static responses,
> and will honor the If-Modified-Since [2] request header.

- [1] https://uwsgi-docs.readthedocs.io/en/latest/StaticFiles.html#setting-the-expires-headers
- [2] https://developer.mozilla.org/de/docs/Web/HTTP/Status/304
- [3] https://github.com/searxng/searxng/pull/4433
- [4] https://github.com/searxng/searxng/issues/964

### DIFF
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -49,6 +49,5 @@ die-on-term
 # uwsgi serves the static files
 static-map = /static=/usr/local/searxng/searx/static
 # expires set to one day
-static-expires = /* 86400
 static-gzip-all = True
 offload-threads = %k
--- a/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
+++ b/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini
@@ -81,7 +81,5 @@ buffer-size = 8192
 #     static_use_hash: true
 #
 static-map = /static=${SEARXNG_STATIC}
-# expires set to one day
-static-expires = /* 86400
 static-gzip-all = True
 offload-threads = %k
--- a/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini:socket
+++ b/utils/templates/etc/uwsgi/apps-archlinux/searxng.ini:socket
@@ -78,7 +78,5 @@ buffer-size = 8192
 #     static_use_hash: true
 #
 static-map = /static=${SEARXNG_STATIC}
-# expires set to one day
-static-expires = /* 86400
 static-gzip-all = True
 offload-threads = %k
--- a/utils/templates/etc/uwsgi/apps-available/searxng.ini
+++ b/utils/templates/etc/uwsgi/apps-available/searxng.ini
@@ -84,7 +84,5 @@ buffer-size = 8192
 #     static_use_hash: true
 #
 static-map = /static=${SEARXNG_STATIC}
-# expires set to one day
-static-expires = /* 86400
 static-gzip-all = True
 offload-threads = %k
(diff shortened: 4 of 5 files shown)


dalf:
Please wait before merging this PR, some context is required.

dalf:
As far I remember, I might be wrong

### With the docker image

The static files are served by UWSGI.
UWSGI ignores the query parameter so `https://searx.example.com/static/simple/logo.svg?gijreghiregjioergjeriogrjio` reads the file `/static/simple/logo.svg`.

==> 

https://github.com/searxng/searxng/blob/8984d7ae022e4e28a9ecb6032c5c535d6ebee520/dockerfiles/uwsgi.ini#L52

and 

https://github.com/searxng/searxng-docker/blob/4aa1d1487869d13160a5bf14c79eaf95d28def15/searxng/settings.yml#L9

works together to provide a nice speed boost.

### No docker, no configuration about the static file

In this scenario, there is a reverse proxy but the static files are served the Flask application.

In this case `https://searx.example.com/static/simple/logo.svg?gijreghiregjioergjeriogrjio` does not work.

Markus, you may think ok this is not the standard installation, so this is not supported. However, I think a lot of existing installations are configured this way (static files served by the Flask application).

Perhaps it is possible to "fix" the Flask application, but I'm not sure this is right way. The recommended way I see is URL like this `https://searx.example.com/static/simple/logo.gijreghiregjioergjeriogrjio.svg`.

return42:
> UWSGI ignores the query parameter so `https://searx.example.com/static/simple/logo.svg?gijreghiregjioergjeriogrjio` reads the file `/static/simple/logo.svg`.

Sure? .. As far I remember the `?` issue was related to the (nginx) proxy .. if I understand your comment correctly:

- https://github.com/searxng/searxng/issues/964#issue-1168788003

However .. as long we set: 

https://github.com/searxng/searxng/blob/8984d7ae022e4e28a9ecb6032c5c535d6ebee520/dockerfiles/uwsgi.ini#L52

in our installation methods, we have a **bug** when  `ui.static_use_hash` is set to `false` (what the default in SearXNG core is) .. this is what I mean by:

>> As long we don't have a working solution for cache busting [3][4] we should not set an expire time in the uWSGI config.

I fully agree with ..

> The recommended way I see is URL like this `https://searx.example.com/static/simple/logo.gijreghiregjioergjeriogrjio.svg`.

If we have fixed the URL, then we can activate `ui.static_use_hash: true` by default and we can set `static-expires = /* 86400` in our uWSGI installation methods.

But before it isn't fixed and cache busting is not the default we should not set `static-expires` in uWSGI .. otherwise we will see issue reports like these from the last days:

- https://github.com/searxng/searxng/issues/4415
- https://github.com/searxng/searxng/issues/4419
- https://github.com/searxng/searxng/issues/4425

> works together to provide a nice speed boost.

This PR is just a interim solution to get static files updated on our clients, which is currently not the case on some installations that using the SearXNG defaults from `searx/settings.yml` and this [uWSGI setup](https://docs.searxng.org/admin/installation-uwsgi.html#uwsgi-setup) .. which is also used by the [installation scripts](https://docs.searxng.org/admin/installation-scripts.html#installation-scripts).

return42:
I merge this PR / it gives us time to look for:

- https://github.com/searxng/searxng/pull/4466

Issue #4435, opened 2025-03-06
Google suspended: too many requests


AdminGodZ:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

**How did you install SearXNG?**
I use my own selfhosted [SearXNG](https://searxng.admingod.ch/) instance, which is forked from [priv.au](https://priv.au/) (I also opened an issue there).
Source code: [priv.au github](https://github.com/privau/searxng)

**What happened?**
Currently when having the google search engine activated under preferences and you are searching something, then google gives back the "Suspended: too many requests" error. Seems like google did something there, idk

**How To Reproduce**
Just have google in preferences activated and then search something, you should then get the same error

**Expected behavior**
Google results are showing

**Screenshots & Logs**
Example:
![Image](https://github.com/user-attachments/assets/8b41ee51-db05-449c-8d06-04605d186932)

**Additional context**
As far as I know this problem is with every SearXNG instance right now, at least for some I tried out at [searx.space](https://searx.space/)

fr0st-xyz:
got the same issue it just happened to me

ark-yh:
+1

return42:
Please guys don't post every minute the same issue:

- https://github.com/searxng/searxng/discussions/4429
- https://github.com/searxng/searxng/discussions/4434
- https://github.com/searxng/searxng/issues/4431

**I really won't to work on the issue** but every two minutes I have to give an answer here that you could have found yourself ..

return42:
Starting around ~12h ago Google does no longer accept requests without enabling JS:

![Image](https://github.com/user-attachments/assets/bacef24b-d668-4219-baef-3626deeec27d)

unixfox:
@return42 it's not new: https://github.com/benbusby/whoogle-search/issues/1211 but keep in mind we are using a special endpoint for sending to google: https://github.com/searxng/searxng/issues/159

unixfox:
Found the issue there is a new "arc_id" which probably comes from actually going a request from the real www.google.com. If you don't pass it then you get a CAPTCHA.

`https://www.google.com/search?q=ok&hl=en-US&lr=lang_en&cr=countryUS&ie=utf8&oe=utf8&filter=0&start=0&asearch=arc&async=arc_id:srp_zF_JZ_ehGr-L7M8Po-jhgQE_120,use_ac:true,_fmt:prog`

VS what's sending searxng:

[...]

Issue #4431, opened 2025-03-06
Search for "abc1" yields no results


zb960303:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

**How did you install SearXNG?**

**What happened?**
After adding numbers to the content I searched for, such as "abc1", I couldn't find any results.

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

![Image](https://github.com/user-attachments/assets/4ff9f51d-f094-485d-a09d-3090edcf8759)

**Additional context**

seirdotexe:
Same here.
```
searxng  | 2025-03-06 07:33:44,808 WARNING:searx.network.google: HTTP Request failed: GET https://www.google.com/sorry/index?continue=https://www.google.com/search%3Fq%3Dtest%26hl%3DZZ-ZZ%26lr%3D%26cr%3D%26ie%3Dutf8%26oe%3Dutf8%26filter%3D0%26start%3D0%26asearch%3Darc%26async%3Duse_ac%253Atrue%252C_fmt%253Aprog&hl=ZZ-ZZ&q=EgRQOVLqGNifpb4GIjB08Jv84Ftv_yVhp7e2Tby9FLuVLH01_IQqVeuDS6-psBhJkCplNXtqJXfQwoYX6_IyAXJaAUM
searxng  | 2025-03-06 07:33:44,840 WARNING:searx.engines.google: ErrorContext('searx/search/processors/online.py', 116, "response = req(params['url'], **request_args)", 'searx.exceptions.SearxEngineTooManyRequestsException', None, ('Too many request',)) False
searxng  | 2025-03-06 07:33:44,840 ERROR:searx.engines.google: Too many requests
searxng  | Traceback (most recent call last):
searxng  |   File "/usr/local/searxng/searx/search/processors/online.py", line 160, in search
searxng  |     search_results = self._search_basic(query, params)
searxng  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng  |   File "/usr/local/searxng/searx/search/processors/online.py", line 144, in _search_basic
searxng  |     response = self._send_http_request(params)
searxng  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng  |   File "/usr/local/searxng/searx/search/processors/online.py", line 116, in _send_http_request
searxng  |     response = req(params['url'], **request_args)
[...]

PR #4442, opened 2025-03-06
[fix] engines: Google-Web & Google-Video (random arc_id)

return42:
Both enghines have been reported ``TooManyRequests``, additionaly Google-Video's thumbnails needed a review.

Based on the research from @unixfox [1] this patch generates every hour a new random ``arc_id``.

[1] https://github.com/searxng/searxng/issues/4435#issuecomment-2703279522

Closes:

- https://github.com/searxng/searxng/issues/4435
- https://github.com/searxng/searxng/issues/4431

Related:

- https://github.com/searxng/searxng/discussions/4434
- https://github.com/searxng/searxng/discussions/4429

-----

ping @fatwang2 @AdminGodZ @fr0st-iwnl @UchihaArk @foosel @merit-mediumdigest @andersnordh @alyx161 @psychic69 @StLyn4 @zb960303 @seirdotexe

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -10,10 +10,14 @@ engines:
 - :ref:`google autocomplete`
 
 """
+from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
 import re
+import random
+import string
+import time
 from urllib.parse import urlencode
 from lxml import html
 import babel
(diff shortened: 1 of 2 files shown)


foosel (approved):
Can't easily test on the exact same host right now, but could test from the same network with the same public IP and can confirm that `2025.3.6+cf90f6ea9` works whereas `2025.3.2+8984d7ae0` doesn't.

Also went through the changes and from my very limited understanding of the searxng codebase but high understanding of Python, they look good.

PR #4443, opened 2025-03-06
[feat] engines: add news.qq.com

hezhijie0327:
## What does this PR do?

1. add Tencent News engine for SearXNG (News)

ref: https://news.qq.com/
Wikidata: https://www.wikidata.org/wiki/Q47524073

*Note: Very High Ratelimit*

|Item|Screenshot|
|-|-|
|News|image|
|Paging|image|

### DIFF
--- a/searx/engines/tencent_news.py
+++ b/searx/engines/tencent_news.py
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Tencent-News: A search engine for retrieving news from Tencent."""
+
+from urllib.parse import urlencode
+import json
+from datetime import datetime
+
+from searx.exceptions import SearxEngineAPIException, SearxEngineTooManyRequestsException
+
+# Metadata
+about = {
+    "website": "https://news.qq.com/",
+    "wikidata_id": "Q47524073",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+# Engine Configuration
+paging = True
+results_per_page = 10
+categories = ["news"]
+
+base_url = "https://i.news.qq.com"
+
+
+def request(query, params):
+    query_params = {
+        "page": params["pageno"],
+        "query": query,
+        "search_type": "all",
+        "search_count_limit": results_per_page,
+        "is_pc": 1,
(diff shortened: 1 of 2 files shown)


searx/engines/tencent_news.py
```diff
@@ -0,0 +1,80 @@
...
+    try:
+        data = resp.json()
+    except json.JSONDecodeError as e:
+        raise SearxEngineAPIException(f"Invalid JSON response: {e}") from e
+
+    sec_list = data.get("secList")
+    if not sec_list:
+        raise SearxEngineTooManyRequestsException(
+            suspended_time=0, message=f"Response is empty or rate-limited by {base_url}, secList not found in response."
+        )
```
return42, line 52: No matter what search term I use, I always get 

```json
{
  "ret":400,
  "words":[{"word":"foo","col":"2e74e6","colDark":"2660bf"}],
  "queryid":"1741337514907372136","real_search_word":"foo","transparam":"{\"trace_id\":\"1741337514907372136\"}","search_id":"1741337514907372136",
  "isEmpty":1,  <-- ??
  "hasMore":0,
  "search_tabs":null,
  "searchTabs":null,
  "secInfo":null,
  "secList":null,  <-- and here
  "topWordsSetting":{"queryWordsColor":"#999999","queryWordsNightColor":"#696969"},
  "total_num":0
}
```

return42 (changes requested):
I always get `SearxEngineTooManyRequestsException` / see my comment in the code ..

searx/engines/tencent_news.py
hezhijie0327: this endpoint has IP risk detection... so may set this PR as draft or close

```
curl 'https://i.news.qq.com/gw/pc_search/result?page=0&query=china&is_pc=1&search_type=all&search_count_limit=10'
```

after 2 request, the ip will be banned about 5min and return 400 (at this time if you access main site also return empty)...
|Scenario|Screenshot|
|-|-|
|Empty Result|image|
|CAPTCHA|image|



---

Work Around:
Use browser visit `https://news.qq.com/search?query=china&page=1` to unblock the IP, then the results back again

PR #4424, opened 2025-03-05
[refactor] typification of SearXNG (MainResult) / result items (part 2)

return42:
Typification of SearXNG
-----------------------

This PR is the continuation of the typification of results introduced in PR:

- https://github.com/searxng/searxng/pull/4183

#### ``[refactor] typification of SearXNG (MainResult) / result items (part 2)``

The class ReslutContainer has been revised, it can now handle the typed Result
items of classes:

- MainResult
- LegacyResult (a dict wrapper for backward compatibility)

Due to the now complete typing of theses three clases, instead of the *getitem*
accesses, the fields can now be accessed directly via attributes (which is also
supported by the IDE).

#### ``[mod] typification of SearXNG: add new result type KeyValue``

This patch adds a new result type: KeyValue

- Python class:   `searx/result_types/keyvalue.py`
- Jinja template: `searx/templates/simple/result_templates/keyvalue.html`
- CSS (less)      `client/simple/src/less/result_types/keyvalue.less`

#### ``[mod] migrate all key-value.html templates to KeyValue type``

The engines now all use KeyValue results and return the results in a
EngineResults object.

The sqlite engine can return MainResult results in addition to KeyValue
results (based on engine's config in settings.yml)

--------

To test `KeyValue` result types add these two engines and use search term ``!demo1 !demo2 *.less``::

```yaml
  - name: find
    engine: command
    shortcut: demo1
    categories: [general, demo]
    command: ['find', '.', '-name', '{{QUERY}}']
    query_type: path
    delimiter:
      chars: ' '
      keys: ['line']
    disabled: false

  - name: demo offline
    engine: demo_offline
    shortcut: demo2
    categories: [general, demo]
    disabled: false
```

You can use the ``sqlite``engine to test `KeyValue` and `MainResult` types:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index 4016d5755..534fa8892 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1827,19 +1827,19 @@ engines:
   # and unpack into searx/data/filmliste-v2.db
   # Query to test: "!mediathekview concert"
   #
-  # - name: mediathekview
-  #   engine: sqlite
-  #   shortcut: mediathekview
-  #   categories: [general, videos]
-  #   result_type: MainResult
-  #   database: searx/data/filmliste-v2.db
-  #   query_str: >-
-  #     SELECT title || ' (' || time(duration, 'unixepoch') || ')' AS title,
[...]

PR #4448, opened 2025-03-06
container: add default support for IPv6

### LINKED ISSUE #965: Docker image: the default_bind_address in docker-entrypoint.sh takes over and forces ipv4.
Version: 2022.03.13-cd92a7ea

From matrix:
> no combination of "::/0", "[::/0]", the full hostname, nor the actual ipv6 address works (*). the default_bind_address in [docker-entrypoint.sh](http://docker-entrypoint.sh/) takes over and forces ipv4.

(*) either in settings.yml or the docker compose yaml file.

For reference:
https://gitlab.com/andrew-gitlab/searxng-iceland/-/issues/3#note_874116212

inetol:
Related https://github.com/searxng/searxng/pull/4378

Closes https://github.com/searxng/searxng/issues/965

To be merged after https://github.com/searxng/searxng/pull/4424

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -23,7 +23,7 @@ Volume:
 EOF
 }
 
-export DEFAULT_BIND_ADDRESS="0.0.0.0:8080"
+export DEFAULT_BIND_ADDRESS="[::]:8080"
 export BIND_ADDRESS="${BIND_ADDRESS:-${DEFAULT_BIND_ADDRESS}}"
 
 # Parse command line
--- a/docs/admin/settings/settings.rst
+++ b/docs/admin/settings/settings.rst
@@ -59,7 +59,7 @@ and can relied on the default configuration :origin:`searx/settings.yml` using:
     use_default_settings: true
     server:
         secret_key: "ultrasecretkey"   # change this!
-        bind_address: "0.0.0.0"
+        bind_address: "[::]"
 
 ``engines:``
   With ``use_default_settings: true``, each settings can be override in a
--- a/tests/unit/settings/user_settings.yml
+++ b/tests/unit/settings/user_settings.yml
@@ -12,7 +12,7 @@ search:
 
 server:
   port: 9000
-  bind_address: "0.0.0.0"
+  bind_address: "[::]"
   secret_key: "user_settings_secret"
   base_url: false
   image_proxy: false
--- a/tests/unit/settings/user_settings_keep_only.yml
+++ b/tests/unit/settings/user_settings_keep_only.yml
@@ -5,7 +5,7 @@ use_default_settings:
       - wikinews
 server:
   secret_key: "user_secret_key"
-  bind_address: "0.0.0.0"
+  bind_address: "[::]"
   default_http_headers:
     Custom-Header: Custom-Value
 engines:
--- a/tests/unit/settings/user_settings_remove.yml
+++ b/tests/unit/settings/user_settings_remove.yml
@@ -5,6 +5,6 @@ use_default_settings:
       - wikinews
 server:
   secret_key: "user_secret_key"
-  bind_address: "0.0.0.0"
+  bind_address: "[::]"
   default_http_headers:
     Custom-Header: Custom-Value
--- a/tests/unit/settings/user_settings_remove2.yml
+++ b/tests/unit/settings/user_settings_remove2.yml
@@ -5,7 +5,7 @@ use_default_settings:
       - wikinews
 server:
   secret_key: "user_secret_key"
-  bind_address: "0.0.0.0"
+  bind_address: "[::]"
   default_http_headers:
     Custom-Header: Custom-Value
 engines:
--- a/tests/unit/settings/user_settings_simple.yml
+++ b/tests/unit/settings/user_settings_simple.yml
@@ -1,7 +1,7 @@
 use_default_settings: true
 server:
   secret_key: "user_secret_key"
-  bind_address: "0.0.0.0"
+  bind_address: "[::]"
   default_http_headers:
     Custom-Header: Custom-Value
 result_proxy:


vojkovic:
How does it listen to ipv4 as well? You've only made it bind to the v6 wildcard address. This will probably confuse people when debugging they can't reach anything over v4.

inetol:
> How does it listen to ipv4 as well? You've only made it bind to the v6 wildcard address. This will probably confuse people when debugging they can't reach anything over v4.

```
Applications may use AF_INET6 sockets to open TCP connections to IPv4
nodes, or send UDP packets to IPv4 nodes, by simply encoding the
destination's IPv4 address as an IPv4-mapped IPv6 address, and
passing that address, within a sockaddr_in6 structure, in the
connect() or sendto() call.
```
https://www.rfc-editor.org/rfc/rfc3493#section-3.7

..basically, every operating system or thing that supports IPv6 will automatically convert from IPv4 to IPv6 space for you.

And if someone explicitly defines the BIND_ADDRESS and the dual stack screwed it is not our problem, all applications go and will go through this transition process someday and users need to learn how to deal with it

unixfox:
In most programming language like python, [::] means that the app will open a socket on both IPv4 and IPv6.

vojkovic:
Ah interesting - in golang it would just listen to v6. Learned something new about python today :)

Ok, so after reading the RFC, IPv4 clients should appear as ::ffff:192.0.2.1 in the logs for example?

I'll take a in depth look later tonight.

unixfox:
Ok well maybe not in python about the sockets listening on both ipv4/ipv6. But indeed it's going to do what the RFC say.

If do I:
```
python -m http.server -b :: 8080
curl 127.0.0.1:8080
```

Then I get:
```
::ffff:127.0.0.1 - - [07/Mar/2025 12:04:48] "GET / HTTP/1.1" 200 -
```

Now I that I think about it, I wonder if this could mess up with the rate limiter that doesn't expect IPv4 IP addresses into an IPv6 address.

According to stackoverflow, if we want to have a dual stack socket then we need to use `dualstack_ipv6=True`: https://stackoverflow.com/questions/65877630/how-to-bind-both-ipv4-and-ipv6-source-address-to-python-socket

In Docker I thought it would "remove" the `::ffff` but seems like it is not:
```
docker run -p 8080:8080 python:3-slim python3 -m http.server -b :: 8080
curl 127.0.0.1:8080
```

```
::ffff:172.17.0.1 - - [07/Mar/2025 11:11:43] "GET / HTTP/1.1" 200 -
```

inetol:
> In Docker I thought it would "remove" the `::ffff` but seems like it is not:

In fact, this is expected to identify that an IPv6 address comes from IPv4 (RFC3493; section 3.7), although now that you mention about ratelimit this may be an issue

unixfox:
@return42 would you be ok to add a specific case about `::ffff` IPv6 address in the limiter?

I guess we would only need to catch this specific `:ffff:` and then remove the `:ffff:` part in order to get the wanted IPv4 address.

return42:
> @return42 would you be ok to add a specific case about `::ffff` IPv6 address in the limiter?

By default [IPv4Address.is_link_local](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_link_local) are not filtered:

https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/botdetection/ip_limit.py#L102-L104

Or with other words, by default [botdetection.ip_limit.filter_link_local](https://docs.searxng.org/admin/searx.limiter.html#limiter-toml) is set to `false`.

unixfox:
@return42 if we merge this PR, all the IPv4 address space will be like this for Docker.

So my question is not only about the localhost IP addresses but all the IPv4 address space.

return42:
> So my question is not only about the localhost IP addresses but all the IPv4 address space.

Python has an abstraction named [`ip_address`](https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_address).

A `::ffff:` prefixed IP is handled as (any other) IPv6 address:

```python
>>> from ipaddress import ip_address
>>> ip_address("::ffff:192.0.2.1")
IPv6Address('::ffff:192.0.2.1')
```

while this is a IPv4 address ..

```python
>>> ip_address("192.0.2.1")
IPv4Address('192.0.2.1')
```

vojkovic:
> By default [IPv4Address.is_link_local](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_link_local) are not filtered:

Keep in mind that IPv4-mapped addresses aren't link local, they are from the block `::ffff:0:0/96` whilst link local are from `fe80::/10` - they won't be exempt from the limiter by default.

Either way the limiter must be able to differentiate between v4 and v6 so that IP ranges can be properly blocked. Because this range is treated like any other v6 address, blocking a /48 [(default)](https://docs.searxng.org/admin/searx.limiter.html#limiter-toml) would ban all of IPv4 address space.

----

https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/limiter.toml#L11

return42:
The limiter uses this function to get the IP address ..

https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/botdetection/_helpers.py#L71-L74

> Either way the limiter must be able to differentiate between v4 and v6 so that IP ranges can be properly blocked. Because this range is treated like any other v6 address, blocking a /48 [(default)](https://docs.searxng.org/admin/searx.limiter.html#limiter-toml) would ban all of IPv4 address space.

May we can (as @unixfox suggested) remove a `:ffff:` prefix before returning the `request_ip`?

https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/botdetection/_helpers.py#L128-L130

vojkovic:
> May we can (as @unixfox suggested) remove a `:ffff:` prefix before returning the `request_ip`?

Yes, this will be the best place to do it (I recommend using [ipv6_mapped](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.ipv6_mapped) though).

We should also consider where else we're using remote_addr - such as in the [Tor Check plugin](https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/plugins/tor_check.py#L78) - which this PR would break for v4.

Actually after looking at this, I think the tor check plugin should use get_real_ip() instead - would you be happy if I send through a PR for this done?

https://github.com/searxng/searxng/blob/67a8b66f750c2ab8e98124951b2401dcbf6634e6/searx/plugins/tor_check.py#L75-L79

return42:
> would you be happy if I send through a PR for this done?

FYI I incorporated the [patch](https://github.com/searxng/searxng/pull/4424/commits/615a92feb4a4eda56dd23842d2182fab5c28fc54) in PR:

- https://github.com/searxng/searxng/pull/4424

unixfox:
I vouch on changing the bind address for the settings.yml too: https://github.com/searxng/searxng/blob/02b76c8389ed0f53be75f686f6063c8c188c1921/docs/admin/settings/settings.rst#L62

Is everyone ok?

unixfox:
@inetol could you add that too as a modification in this PR :)?

you can change the title like "add default support for IPv6"

unixfox:
> > would you be happy if I send through a PR for this done?
> 
> FYI I incorporated the [patch](https://github.com/searxng/searxng/pull/4424/commits/615a92feb4a4eda56dd23842d2182fab5c28fc54) in PR:
> 
>     * [[refactor] typification of SearXNG (MainResult) / result items (part 2) #4424](https://github.com/searxng/searxng/pull/4424)

@return42 May I understand that what I have asked you is resolved in the PR above?

> @return42 would you be ok to add a specific case about `::ffff` IPv6 address in the limiter?
> 
> I guess we would only need to catch this specific `:ffff:` and then remove the `:ffff:` part in order to get the wanted IPv4 address.

Thank you in advance :)

return42:
> @return42 May I understand that what I have asked you is resolved in the PR above?

yes / sorry if it was not clear :-)

unixfox:
Great, I guess we can wait for https://github.com/searxng/searxng/pull/4424 to be merged in order to merge this PR (#4448).

return42:
> > @return42 May I understand that what I have asked you is resolved in the PR above?
> 
> yes / sorry if it was not clear :-)

FYI: PR is now merged .. https://github.com/searxng/searxng/pull/4424

PR #4439, opened 2025-03-06
[feat] engines add images & kaifa from baidu.com

hezhijie0327:
## What does this PR do?

1. add Baidu Images engine (Images)
ref: https://image.baidu.com/
Wikidata: https://www.wikidata.org/wiki/Q109342769

2. add Baidu Kaifa (Developer) engine (IT)
ref: https://kaifa.baidu.com/

3. add Baidu Web time range support

|Engine|Screenshot|
|-|-|
|Image|image|
|Kaifa|image|

PR #4449, opened 2025-03-07
[feat] engines: add kaifa.baidu.com

hezhijie0327:
## What does this PR do?

1. add Baidu Kaifa (Developer) engine for SearXNG (IT)

Note: this engine is used to search more result from IT tech resources

ref: https://kaifa.baidu.com/

|Screenshot|
|-|
|image|


hezhijie0327:
ref #4439

PR #4458, opened 2025-03-07
[fix] presearch engine: Title showing <em> html code

Aadniz:
## What does this PR do?

Fixes an issue where Presearch would show HTML code in the title as shown in the screenshot below. To trigger the bug, go to page 2 or more.

Screenshot_2025-03-07_10:26:27

There is actually 2 issues here, but only 1 issue is related to searxng. I do have another branch ready if you want to fix *their* bug: https://github.com/searxng/searxng/compare/master...Aadniz:searxng:presearch-fix

The fix is simple, wrap `html_to_text` around `item['title']`

It is not pleasant to look at, and nobody wants to see HTML code when reading normal text. It also messes with the title if multiple engines find the same results.

```shell
$ git switch presearch-title-fix
$ python3 setup.py build
$ searxng-run
```

### DIFF
--- a/searx/engines/presearch.py
+++ b/searx/engines/presearch.py
@@ -161,7 +161,7 @@ def parse_search_query(json_results):
     for item in json_results.get('specialSections', {}).get('topStoriesCompact', {}).get('data', []):
         result = {
             'url': item['link'],
-            'title': item['title'],
+            'title': html_to_text(item['title']),
             'thumbnail': item['image'],
             'content': '',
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thanks :+1:

Bnyro:
Please feel free to send a PR for the other fix, I'd however simply use `title.removeprefix(url.netloc)` instead of doing string splitting in this case.

Issue #3888, opened 2024-10-05
Remove morty and filtron leftovers


return42:
Morty and filtron are unmaintained components and had been replaced by an image-proxy and the bot-detection in SearXNG since years .. we should remove any leftovers:

Related:

- [The infrastructure of SearXNG has changed dramatically over the past year. The installation scripts and documentation have not been adapted so far.](https://github.com/searxng/searxng/pull/1332)
- filtron https://github.com/search?q=repo%3Asearxng%2Fsearxng+filtron+&type=code
- morty https://github.com/search?q=repo%3Asearxng%2Fsearxng+morty+&type=code

allixx:
what is the alternative for morty as result_proxy service? occasionally I like to open search results via morty. image_proxy does not provide this functionality.

for what is worth, morty works just fine on private instances.

Bakr-Ali:
> what is the alternative for morty as result_proxy service? occasionally I like to open search results via morty. image_proxy does not provide this functionality.
> 
> for what is worth, morty works just fine on private instances.

Did you find any alternative? Do you know where to find a list of searxng instances that still support morty proxy?

I don't understand why anyone would remove such a useful feature

return42:
> I don't understand why anyone would remove such a useful feature

Did you ever tested .. it does not work with most modern WEB sites and morty is not maintained .. last release was [in 2018](https://github.com/asciimoo/morty/tags) .. users gave up in 2022 to report issues .. --> nobody really uses it anymore.

Bakr-Ali:
> Did you ever tested .. it does not work with most modern WEB sites 

Yes, I did (on the only working instance i know of), and it works perfectly as far as I have tested on multiple websites and without any errors.
Example: https://morty.ononoki.org/?mortyurl=https://www.euronews.com/next/2025/02/26/asteroid-2024-yr4-is-no-longer-a-threat-to-earth-scientists-say

> and morty is not maintained .. last release was [in 2018](https://github.com/asciimoo/morty/tags) .. users gave up in 2022 to report issues .. --> nobody really uses it anymore.

I didn't notice that until after I wrote my comment here.

Do you know of any alternatives / similar services?

loorisr:
maybe https://www.privoxy.org/ or https://mitmproxy.org can be a replacement?

Issue #3887, opened 2024-10-05
Docker build fails on installing pydantic-core [linux/arm/v7]


return42:
Docker build fails on installing pydantic-core / see report from CI:

- https://github.com/searxng/searxng/actions/runs/11191050608/job/31113678686#step:8:981

```
#24 [linux/arm/v7 5/8] RUN apk add --no-cache -t build-dependencies     build-base     py3-setuptools     python3-dev     libffi-dev     libxslt-dev     libxml2-dev     openssl-dev     tar     git  && apk add --no-cache     ca-certificates     python3     py3-pip     libxml2     libxslt     openssl     tini     uwsgi     uwsgi-python3     brotli  && pip3 install --break-system-packages --no-cache -r requirements.txt  && apk del build-dependencies  && rm -rf /root/.cache
...
#24 221.9 Collecting pydantic-core==2.20.1 (from pydantic==2.8.2->-r requirements.txt (line 19))
#24 221.9   Downloading pydantic_core-2.20.1.tar.gz (388 kB)
#24 222.0      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 388.4/388.4 kB 18.9 MB/s eta 0:00:00
#24 222.3   Installing build dependencies: started
#24 230.0   Installing build dependencies: finished with status 'done'
#24 230.0   Getting requirements to build wheel: started
#24 230.6   Getting requirements to build wheel: finished with status 'done'
#24 230.6   Preparing metadata (pyproject.toml): started
#24 231.2   Preparing metadata (pyproject.toml): finished with status 'error'
#24 231.2   error: subprocess-exited-with-error
#24 231.2   
#24 231.2   × Preparing metadata (pyproject.toml) did not run successfully.
#24 231.2   │ exit code: 1
#24 231.2   ╰─> [6 lines of output]
#24 231.2       
#24 231.2       Cargo, the Rust package manager, is not installed or is not on PATH.
#24 231.2       This package requires Rust and Cargo to compile extensions. Install it through
#24 231.2       the system's package manager or via https://rustup.rs/
#24 231.2       
#24 231.2       Checking for Rust toolchain....
#24 231.2       [end of output]
#24 231.2   
#24 231.2   note: This error originates from a subprocess, and is likely not a problem with pip.
#24 231.2 error: metadata-generation-failed
#24 231.2 
#24 231.2 × Encountered error while generating package metadata.
#24 231.2 ╰─> See above for output.
#24 231.2 
[...]

Issue #3626, opened 2024-06-29
all engines running into timeout


0x5247:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2024.6.29+adaec6872

**How did you install SearXNG?**
docker

**What happened?**
When I searched for something I got this error
Error! Engines cannot retrieve results:

**Screenshots & Logs**
![Screenshot (61)](https://github.com/searxng/searxng/assets/143602102/d75c8003-964a-441f-8df4-45f98b1e8aea)

return42:
This is not a google issue (I change the title) .. seems you have a problem with the internet connection since all engines have timeouts.  For more help, please ask community at [#searxng:matrix.org](https://matrix.to/#/#searxng:matrix.org) .. and open an issue only for a real issue / thanks for understanding.

sommerfelddev:
I'm running into the same issue occasionally, nothing seems odd about my internet connection.

Huddeij:
Yes my inet connection works fine as well while running into this issue occasionally. Self-hosting via docker container 

```yml
services:
  redis:
    container_name: searxng_redis
    image: docker.io/library/redis:alpine
    command: redis-server --save 30 1 --loglevel warning
    networks:
      - searxng
    volumes:
      - redis-data:/data
    cap_drop:
      - ALL
    cap_add:
      - SETGID
      - SETUID
      - DAC_OVERRIDE
    restart: unless-stopped

  searxng:
    container_name: searxng
    image: docker.io/searxng/searxng:latest
    networks:
      - searxng
    ports:
      - 8080:8080
    volumes:
      - ./searxng:/etc/searxng:rw
    environment:
      - SEARXNG_BASE_URL=https://${SEARXNG_HOSTNAME:-localhost}/
    cap_drop:
      - ALL
    cap_add:
      - CHOWN
      - SETGID
      - SETUID
    logging:
      driver: "json-file"
      options:
        max-size: "1m"
        max-file: "1"
    restart: unless-stopped

networks:
  searxng:
    ipam:
      driver: default

volumes:
  redis-data:
```

return42:
@Huddeij you want get any feedback here in the SearXNG core repository, its the wrong place. I recommend to discuss your local network setup in the community at [#searxng:matrix.org](https://matrix.to/#/#searxng:matrix.org).

PR #4463, opened 2025-03-07
Dockerfile simpler

### LINKED ISSUE #4460: Simpler Dockerfile
The current Dockerfile is using an alpine Base image and install python3.
A python3 alpine image exists.

We could use the python-alpine image as a base image to have a simplier Dockerfile. The build time will be faster.

Here is the required command line to install the packages.

```yaml
RUN apk add brotli tini openssl && rm -rf /root/.cache

COPY requirements.txt ./requirements.txt

RUN apk add --no-cache -t build-dependencies gcc libc-dev linux-headers && pip install --no-cache -r requirements.txt \
&& apk del build-dependencies \
&& rm -rf /root/.cache
````

The complete Dockerfile is available here: https://github.com/loorisr/searxng/blob/master/Dockerfile.simple

**Additional context**
uwsgi is built and installed using pip

If you agree I will make a PR

### PR DESCRIPTION
**Pull request by loorisr**
## What does this PR do?

This PR installs uwsgi with pip and use a python-alpine base image instead of alpine.

It reduces the complexity of the docker file.

I think that further arg/variables could be removed from the dockerfile but I let them as in the current DOckerfile in this PR.

Build the image with docker and run it.

## Author's checklist

loorisr

## Related issues

Closes #4460

### CODE REVIEW

**Comment by loorisr** [comment]
On my dev machine it is working fine but on my prod machine I have a problem with serving static files. I'm investigating

**Comment by loorisr** [comment]
/etc/mime.types was missing. I added mailcap to fix it.

git is not installed, I don't known if you really need it. it is to get the current version, but on the main it is installed and removed later, that's why I haven't install it.

**Comment by unixfox** [comment]
Could you check if building the Dockerfile works on ARM64 and ARMv7 too? You can find all the architectures here: https://hub.docker.com/r/searxng/searxng/tags

docker buildx command: https://github.com/searxng/searxng/blob/02b76c8389ed0f53be75f686f6063c8c188c1921/manage#L194

**Comment by unixfox** [comment]
> /etc/mime.types was missing. I added mailcap to fix it.
> 
> git is not installed, I don't known if you really need it. it is to get the current version, but on the main it is installed and removed later, that's why I haven't install it.

I guess @return42 knows more about that?

File: Dockerfile [hunk 1]
```diff
@@ -21,32 +20,13 @@ ENV INSTANCE_NAME=searxng \
...
-    tini \
-    uwsgi \
-    uwsgi-python3 \
-    brotli \
- && pip3 install --break-system-packages --no-cache -r requirements.txt \
- && apk del build-dependencies \
- && rm -rf /root/.cache
+RUN apk add --no-cache -t build-dependencies gcc libc-dev linux-headers && pip install --no-cache -r requirements.txt \
+&& apk del build-dependencies \
+&& rm -rf /root/.cache
```
inetol: Screenshot

We should probably keep git installed until finishes building and then removing `build-dependencies`? Or what is this message from, building directly from the Dockerfile?

Dockerfile
```diff
@@ -21,32 +20,13 @@ ENV INSTANCE_NAME=searxng \
 
 WORKDIR /usr/local/searxng
 
+RUN apk add brotli tini openssl mailcap && rm -rf /root/.cache
```
inetol: `--no-cache` flag is missing

inetol (changes requested):
I've looked at it quite diagonally, since I can't instance containers now to see if everything is running OK. These changes look pretty good at first glance

Please check my comments.

Dockerfile
loorisr: git seems to be used only to get the version in `version.py`

it returns git_version, tag_version, docker_tag but when I build the docker on my environment, even with git installed it doesn't return anything.

if someones can try on another environment with git that would be great!

Dockerfile
loorisr: you are right.

I've updated it

Dockerfile
inetol: Weird, I found where those arguments are set:

https://github.com/searxng/searxng/blob/eb3633629a8e91ece7f75ab58226de2d2df1b698/manage#L202-L213

I don't understand exactly why this is, maybe the rest have a better idea than I do right now.

Bnyro (commented):
The environment variables in line 15 and 16 are no longer needed / working, so we should probably remove them too, see https://github.com/searxng/searxng/issues/3888.

Dockerfile
```diff
@@ -21,32 +20,13 @@ ENV INSTANCE_NAME=searxng \
...
-    libxslt-dev \
-    libxml2-dev \
-    openssl-dev \
-    tar \
-    git \
- && apk add --no-cache \
-    ca-certificates \
-    python3 \
-    py3-pip \
-    libxml2 \
```
Bnyro, line 40: https://pypi.org/project/lxml/ requires `libxml2` and `libxslt`.
If we install `lxml` via `pip` now, they're probably built using wheel which increases build time I think? (haven't tested it yet)

requirements.txt
```diff
@@ -19,3 +19,4 @@ tomli==2.0.2; python_version < '3.11'
 msgspec==0.19.0
 typer-slim==0.15.2
 isodate==0.7.2
+uwsgi==2.0.28
```
Bnyro: I don't think that's reasonable to move `uwsgi` here, uwsgi is only needed when running the docker image, it's not needed for running SearXNG for testing or deploying it with other methods than Docker.

Dockerfile
unixfox: yes indeed, hence why I said to test on arm too

requirements.txt
inetol: There is no problem installing directly from pip, right? I prefer installing uWSGI directly from pip rather than Alpine Linux repos

```diff
-RUN apk add --no-cache -t build-dependencies gcc libc-dev linux-headers && pip install --no-cache -r requirements.txt \
-&& apk del build-dependencies \
-&& rm -rf /root/.cache
+RUN apk add --no-cache -t build-dependencies gcc libc-dev linux-headers \
+ && pip install --no-cache -r requirements.txt \
+ && pip install --no-cache "uwsgi~=2.0.0" \
+ && apk del build-dependencies \
+ && rm -rf /root/.cache
```

inetol:
> On my dev machine it is working fine but on my prod machine I have a problem with serving static files. I'm investigating

Probably unrelated, but It seems that if installing uwsgi from pip, `pcre-dev` package *should* be installed on build time and the “plugin” parameter must be removed from uwsgi.ini..:

<details>
<summary>diff+</summary>

```diff
diff --git a/Dockerfile b/Dockerfile
index c91744474..d05bb67cb 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,12 +20,11 @@ ENV INSTANCE_NAME=searxng \
 
 WORKDIR /usr/local/searxng
 
-RUN apk add --no-cache brotli tini openssl mailcap && rm -rf /root/.cache
+RUN apk add --no-cache brotli tini openssl mailcap pcre-dev && rm -rf /root/.cache
 
 COPY requirements.txt ./requirements.txt
 
 RUN apk add --no-cache -t build-dependencies gcc libc-dev linux-headers && pip install --no-cache -r requirements.txt \
-&& apk del build-dependencies \
 && rm -rf /root/.cache
 
 COPY --chown=searxng:searxng dockerfiles ./dockerfiles
@@ -40,7 +39,8 @@ RUN su searxng -c "/usr/local/bin/python3 -m compileall -q searx" \
  && touch -c --date=@${TIMESTAMP_UWSGI} dockerfiles/uwsgi.ini \
  && find /usr/local/searxng/searx/static -a \( -name '*.html' -o -name '*.css' -o -name '*.js' \
     -o -name '*.svg' -o -name '*.ttf' -o -name '*.eot' \) \
-    -type f -exec gzip -9 -k {} \+ -exec brotli --best {} \+
+    -type f -exec gzip -9 -k {} \+ -exec brotli --best {} \+ \
+ && apk del build-dependencies
 
 # Keep these arguments at the end to prevent redundant layer rebuilds
 ARG LABEL_DATE=
diff --git a/dockerfiles/uwsgi.ini b/dockerfiles/uwsgi.ini
index 812716853..f29253558 100644
--- a/dockerfiles/uwsgi.ini
+++ b/dockerfiles/uwsgi.ini
@@ -17,7 +17,6 @@ chmod-socket = 666
 # Plugin to use and interpreter config
 single-interpreter = true
 master = true
-plugin = python3
 lazy-apps = true
 enable-threads = true
 
 ```

</details>

**Comment by return42** [comment]
> > git is not installed, I don't known if you really need it. it is to get the current version, but on the main it is installed and removed later, that's why I haven't install it.
> 
> I guess @return42 knows more about that?

not really .. I can only guess: git is installed to generate the [searx.version_frozen](https://github.com/searxng/searxng/issues/728) number, after the generation git is no longer needed.

File: requirements.txt (hunk 4 above)
Comment (return42) [reply requirements.txt]: IDK much about the docker build of SearXNG, but installing uwsgi from package manager of OS and installing a pip package are fundamentally different things .. while pip installs not more than a python package, the package manager from OS install a services.

Can't speak for Alpine but here I documented it for some other distros: https://docs.searxng.org/admin/installation-uwsgi.html

File: Dockerfile (hunk 3 above)
Comment (return42) [reply Dockerfile:40]: > they're probably built using wheel which increases build time I think?

@dalf invested a lot of time optimizing builds (build time) on the different architectures, I can't believe there is even one apk package we can drop ..

Comment (loorisr) [reply Dockerfile:40]: yes you are right libxslt are missing.

the -dev packages are used at build time and are not necessary if pip can install a binary. I will test on an arm64 today.

File: requirements.txt (hunk 4 above)
Comment (loorisr) [reply requirements.txt]: IMHO we should move away from uWSGI that is in maintenance only.

We can install uWSGI using apk but I struggled to make it works. with pip with have the latest version and it works fine.

**Comment by loorisr** [comment]
On a raspberrypi4, arm64,  it is working fine.
I've added : 
* pcre
* libxml2
* libxslt 

I've haven't add git as in the current main it is not used in the dockerfile.

If you want we could replace tini with --init?
https://docs.docker.com/reference/cli/docker/container/run/#init

**Comment by unixfox** [comment]
Make sure to test on armv7 too. With the buildx command above you can test using virtualization on a x86 processor.

File: Dockerfile (hunk 3 above)
Comment (return42) [reply Dockerfile:40]: > if pip can install a binary. 

Thats the point, we support [amd64, arm64 and arm/v7](https://github.com/searxng/searxng-docker?tab=readme-ov-file#multi-architecture-docker-images) .. while binaries for amd64 often exists, most often a binary for arm/v7 (and arm64) is missed .. as far as my memory serves me I remember about a lot of problems in the past related to, pygments, lxml, Brotli, fasttext-predict, tomli and msgspec packages.

Further: **Alpine is *musl* while most python packages are compiled for *gnu***

**I can only recommend relying on the expertise of @dalf, who knows every package and is in the best position to assess problems and risks.**

File: requirements.txt (hunk 4 above)
Comment (return42) [reply requirements.txt]: ["has been merged (on Oct. 2014) to the master branch of uWSGI but had never been released; the last major release is from Dec. 2013, since the there had been only bugfix releases."](https://docs.searxng.org/admin/installation-uwsgi.html#pitfalls-of-the-tyrant-mode) 

We have discussed leaving uwsgi many times ... and we need to, but it should not be part of this PR as it is a fundamental decision and not part of the “Dockerfile simpler” topic of this PR.

File: dockerfiles/uwsgi.ini [hunk 5]
```diff
@@ -17,7 +17,7 @@ chmod-socket = 666
 # Plugin to use and interpreter config
 single-interpreter = true
 master = true
-plugin = python3
+#plugin = python3
```
return42, line 20: The plugin is needed, in this context I want to remember a discussion we had about uwsgi setup:

Dockerfile
loorisr: Yes while making compilation with arm/v7 I needed to add libxml2-dev libxslt-dev 

libxml doesn't provide packages for arm/v7 on alpine :(

requirements.txt
loorisr: I agree, it is a different topic

dockerfiles/uwsgi.ini
inetol: It is not necessary when installing uwsgi from pip, in fact it throws an error that it cannot find the python3 plugin

return42: > when installing uwsgi from pip,

and this is the first issue / read my comment from above: https://github.com/searxng/searxng/pull/4463#discussion_r1986016740

However I don't use docker and cant help when there are issues  ..  I have expressed all my concerns about it, that's all I can do .. I'm out of this topic.

loorisr:
For arm/v7, build-base is needed to build uvloop, lxml, ...

I propose to use a condition in the Dockerfile to install it only for arm/v7? so builds for arm64 & amd64 will be faster?

loorisr:
The current PR compile on amd64, arm64 and arm/v7.
Several packages are needed for arm/v7 only

Further simplification proposal:
* replace tini with docker built in `--init` command
* we could replace openssl, that is needed to create a secret, by `xxd  -l 32 -p /dev/urandom`
* I don't see the purpose of 
```
ARG TIMESTAMP_SETTINGS=0
ARG TIMESTAMP_UWSGI=0

--date=@${TIMESTAMP_SETTINGS}
--date=@${TIMESTAMP_UWSGI
```

inetol:
> * replace tini with docker built in `--init` command

If it is used at the time of instantiating the container it will need a later PR updating also the compose file in https://github.com/searxng/searxng-docker. Although honestly I have not noticed any changes by using this "init" in several containers that I use so we could try to remove it directly.

> * I don't see the purpose of
> 
> ```
> ARG TIMESTAMP_SETTINGS=0
> ARG TIMESTAMP_UWSGI=0
> 
> --date=@${TIMESTAMP_SETTINGS}
> --date=@${TIMESTAMP_UWSGI
> ```

It is to check for settings.yml and uwsgi.ini file updates

https://github.com/searxng/searxng/blob/3fe602a468a7ac3992ba2f4df6ce41b46b3559bc/dockerfiles/docker-entrypoint.sh#L102-L136

inetol:
> I propose to use a condition in the Dockerfile to install it only for arm/v7? so builds for arm64 & amd64 will be faster?

I will assume that there is already a functional layer cache system in the CI, so the impact on consecutive rebuilds should be non-existent

loorisr:
> > * replace tini with docker built in `--init` command
> 
> If it is used at the time of instantiating the container it will need a later PR updating also the compose file in https://github.com/searxng/searxng-docker. Although honestly I have not noticed any changes by using this "init" in several containers that I use so we could try to remove it directly.
> 
> > * I don't see the purpose of
> > 
> > ```
> > ARG TIMESTAMP_SETTINGS=0
> > ARG TIMESTAMP_UWSGI=0
> > 
> > --date=@${TIMESTAMP_SETTINGS}
> > --date=@${TIMESTAMP_UWSGI
> > ```
> 
> It is to check for settings.yml and uwsgi.ini file updates
> 
> https://github.com/searxng/searxng/blob/3fe602a468a7ac3992ba2f4df6ce41b46b3559bc/dockerfiles/docker-entrypoint.sh#L102-L136

--init set the PID of uwsgi to 7 instead of 1. it works exactly as tini. Yes we need to remove init in apk, change the entrypoint and update the docker-compose. I will update the PR.

If we don't change `requirements.txt`, the caching system is working in local deployment. I don't known with the CI used. 

Ok thanks, TIMESTAMP_SETTINGS and TIMESTAMP_UWSGI are updated in the `manage` script. I used to build the docker image using github actions or docker build and so I doesn't set properly these variables (but it is working fine anyway :))

`VERSION_GITCOMMIT` doesn't seem to be used in the code

unixfox:
I was wondering, if we switch to debian based `python:3.13`, do we get a way higher docker image size compared to alpine? Because we could benefit from all the wheels that are compiled for libc (instead of musl)

inetol:
At best, it would only increase 16 MB from what we have now (using `-slim` variant)

loorisr:
docker image (amd64) with 3.13-alpine is 171mo
docker image (amd64) with 3.13-slim is 337mo

Maybe it's possible to optimize the 3.13-slim build but using a similar Dockerfile than with alpine, image size is *2

loorisr:
I've added a PR https://github.com/searxng/searxng-docker/pull/362 to replace tini with Docker --init parameter

return42:
> image size is *2

As a developer, I would clearly opt for glibc, because we will have less problems using Python libraries than with musl ... and as a developer I would like to choose a library and I am less interested if it is available on 32bit musl.

Even if the image size (more than) doubles, I want to remind:   We already have databases like valkey and sqlite that also consume resources: if an image twice the size is is not running on an old Pi, then you can no longer use SearXNG properly on it anyway.

unixfox:
> docker image (amd64) with 3.13-alpine is 171mo docker image (amd64) with 3.13-slim is 337mo
> 
> Maybe it's possible to optimize the 3.13-slim build but using a similar Dockerfile than with alpine, image size is *2

What's important is to check the compressed size, what you are comparing if the decompressed size: https://stackoverflow.com/questions/33352901/get-the-size-of-a-docker-image-before-a-pull and https://stackoverflow.com/questions/74314081/how-to-estimate-compressed-docker-image-size-before-the-push

I recently tried to build a docker image for armv7 using AMD64 processor (emulation) and that took like 45 minutes on a 4 cores CPU. We can probably cut down on that time when using wheel already built.

> As a developer, I would clearly opt for glibc, because we will have less problems using Python libraries than with musl ... and as a developer I would like to choose a library and I am less interested if it is available on 32bit musl.

True, musl is known to have odd behavior like the famous DNS issue: https://news.ycombinator.com/item?id=36933028 (which is now fixed).

loorisr:
Compressed sizes are (amd64):
* python:3.13-alpine: 58 mo
* python:3.13-slim: 110 mo

Yes compiling for arm/v7 takes ages on my amd64 computer! But then with docker caching system, new built are almost instant.

It's true that alpine variant can cause troubles. As SearxNG is working with alpine linux, I'd recommend to keep it, as long as we don’t add an unsupported library. Except for arm/v7, and uwsgi, wheels already exist for alpine linux on amd64/arm64.

inetol:
> Yes compiling for arm/v7 takes ages on my amd64 computer! But then with docker caching system, new built are almost instant.

The thing is that the layer caching has never worked in this repo so this is going to be overkill, I would like to remedy this in another PR when this is merged.

I have no preference and it depends more on what you are going to run, one or another may be better (I remember reading that running Python programs under musl could impact on its performance unlike glibc, I don't know if that is currently fixed). Due to the issue above, I see more convenient to use the `-slim` variants atm

unixfox:
> As SearxNG is working with alpine linux

There has been so many issues about "timeouts" and I wonder if it's not due to musl and DNS: https://github.com/searxng/searxng/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aclosed%20timeout and https://github.com/searxng/searxng/issues/3626

return42:
> I see more convenient to use the `-slim` variants atm

Me too .. I remember we often had issues in the past when using modern python libs, to name just two

- https://github.com/searxng/searxng/issues/3887
- https://github.com/searxng/searxng/issues/2028

With lot of efforts @dalf fixed those issue or filled the gap otherwise in the past .. from a developer point of view, we want/need to use libs .. but we don't have the resources to fix issues pooping up in the deployment .. This is especially annoying when we do a major refactoring based on a new library and then realize during deployment that the library is not available or cannot be compiled / application developers are not always familiar with what is needed to compile this or that lib ... and binaries from the package manager are more resource efficient anyway .. and shorten the time to build and test the application.

unixfox:
>     * python:3.13-slim: 110 mo

Odd for me, it's 42MB (amd64): https://hub.docker.com/_/python/tags?name=3.13-slim

loorisr:
> > ```
> > * python:3.13-slim: 110 mo
> > ```
> 
> Odd for me, it's 42MB (amd64): https://hub.docker.com/_/python/tags?name=3.13-slim

yes but I was talking about SearxNG compiled using the slim image. Not just the python slim image.

unixfox:
> > > ```
> > > * python:3.13-slim: 110 mo
> > > ```
> > 
> > 
> > Odd for me, it's 42MB (amd64): [hub.docker.com/_/python/tags?name=3.13-slim](https://hub.docker.com/_/python/tags?name=3.13-slim)
> 
> yes but I was talking about SearxNG compiled using the slim image. Not just the python slim image.

Do you have an example of a Dockerfile to review?

loorisr:
I've added pip cache for faster build (--mount=type=cache,target=/root/.cache/pip)

Compilation for arm/v7 is still running so I'm waiting for it

I'm also making test with a 2 stage build to reduce a little bit more the image size and have a better caching handling

````yaml
FROM python:3.13-slim
EXPOSE 8080
VOLUME /etc/searxng

ARG SEARXNG_GID=977
ARG SEARXNG_UID=977

RUN groupadd -g ${SEARXNG_GID} searxng && \
    useradd -u ${SEARXNG_UID} -d /usr/local/searxng -s /bin/sh -g searxng searxng

ENV INSTANCE_NAME=searxng \
    AUTOCOMPLETE= \
    BASE_URL= \
    SEARXNG_SETTINGS_PATH=/etc/searxng/settings.yml \
    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini \
    UWSGI_WORKERS=%k \
    UWSGI_THREADS=4

WORKDIR /usr/local/searxng

# install necessary runtime packages
RUN apt-get update && apt-get install -y --no-install-recommends \
      brotli \
      openssl \
      mailcap \
      libxml2 \
      libxslt1.1 \
      libpcre3 

COPY requirements.txt ./requirements.txt

# build and install uwsgi and necessary python packages
RUN --mount=type=cache,target=/root/.cache/pip apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    libffi-dev \
    libxml2-dev \
    libxslt1-dev \
    libpcre3-dev && \
    pip install "uwsgi~=2.0.0" && \
    pip install -r requirements.txt && \
    apt-get purge -y build-essential libffi-dev libxml2-dev libxslt1-dev libpcre3-dev && \
    apt-get autoremove -y && \
    rm -rf /var/lib/apt/lists/*

COPY --chown=searxng:searxng dockerfiles ./dockerfiles
COPY --chown=searxng:searxng searx ./searx

ARG TIMESTAMP_SETTINGS=0
ARG TIMESTAMP_UWSGI=0
ARG VERSION_GITCOMMIT=unknown

USER searxng 

RUN python3 -m compileall -q searx \
 && touch -c --date=@${TIMESTAMP_SETTINGS} searx/settings.yml \
 && touch -c --date=@${TIMESTAMP_UWSGI} dockerfiles/uwsgi.ini \
 && find /usr/local/searxng/searx/static  \( -name '*.html' -o -name '*.css' -o -name '*.js' \
    -o -name '*.svg' -o -name '*.ttf' -o -name '*.eot' \) \
    -type f -exec gzip -9 -k {} \+ -exec brotli --best {} \+

# Keep these arguments at the end to prevent redundant layer rebuilds
ARG LABEL_DATE=
ARG GIT_URL=unknown
ARG SEARXNG_GIT_VERSION=unknown
ARG SEARXNG_DOCKER_TAG=unknown
ARG LABEL_VCS_REF=
ARG LABEL_VCS_URL=
LABEL maintainer="searxng <${GIT_URL}>" \
      description="A privacy-respecting, hackable metasearch engine." \
      version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.schema-version="1.0" \
      org.label-schema.name="searxng" \
      org.label-schema.version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.url="${LABEL_VCS_URL}" \
      org.label-schema.vcs-ref=${LABEL_VCS_REF} \
      org.label-schema.vcs-url=${LABEL_VCS_URL} \
      org.label-schema.build-date="${LABEL_DATE}" \
      org.label-schema.usage="https://github.com/searxng/searxng-docker" \
      org.opencontainers.image.title="searxng" \
      org.opencontainers.image.version="${SEARXNG_DOCKER_TAG}" \
      org.opencontainers.image.url="${LABEL_VCS_URL}" \
      org.opencontainers.image.revision=${LABEL_VCS_REF} \
      org.opencontainers.image.source=${LABEL_VCS_URL} \
      org.opencontainers.image.created="${LABEL_DATE}" \
      org.opencontainers.image.documentation="https://github.com/searxng/searxng-docker"
      
ENTRYPOINT ["/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
````

**Comment by loorisr** [comment]
I have 2 proposals, with 1 stage or 2 stages that compiles on amd64, arm64, arm/v7
To run without error, alpine needs to be disable from settings.yml

The 2 stages allows a better use of docker caching system without having a larger image.

 I've replaced openssl with `head -c 32 /dev/urandom | od -An -tx1 | tr -d ' \n'` in docker-entrypoint.sh

<details>
  <summary>1 stage, with python:slim</summary>

  ```yaml
FROM python:3.13-slim
EXPOSE 8080
VOLUME /etc/searxng

ARG SEARXNG_GID=977
ARG SEARXNG_UID=977

RUN groupadd -g ${SEARXNG_GID} searxng && \
    useradd -u ${SEARXNG_UID} -d /usr/local/searxng -s /bin/sh -g searxng searxng

ENV INSTANCE_NAME=searxng \
    AUTOCOMPLETE= \
    BASE_URL= \
    SEARXNG_SETTINGS_PATH=/etc/searxng/settings.yml \
    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini \
    UWSGI_WORKERS=%k \
    UWSGI_THREADS=4

WORKDIR /usr/local/searxng

# install necessary runtime packages
RUN apt-get update && apt-get install -y --no-install-recommends \
      brotli \
      openssl \
      mailcap \
      libxml2 \
      libxslt1.1 \
      libpcre3 

COPY requirements.txt ./requirements.txt

# build and install uwsgi and necessary python packages
RUN --mount=type=cache,target=/root/.cache/pip apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    zlib1g-dev \
    libffi-dev \
    libxml2-dev \
    libxslt-dev \
    libpcre3-dev && \
    pip install "uwsgi~=2.0.0" && \
    pip install -r requirements.txt && \
    apt-get purge -y build-essential libffi-dev libxml2-dev libxslt1-dev libpcre3-dev && \
    apt-get autoremove -y && \
    rm -rf /var/lib/apt/lists/*

COPY --chown=searxng:searxng dockerfiles ./dockerfiles
COPY --chown=searxng:searxng searx ./searx

ARG TIMESTAMP_SETTINGS=0
ARG TIMESTAMP_UWSGI=0
ARG VERSION_GITCOMMIT=unknown

USER searxng 

RUN python3 -m compileall -q searx \
 && touch -c --date=@${TIMESTAMP_SETTINGS} searx/settings.yml \
 && touch -c --date=@${TIMESTAMP_UWSGI} dockerfiles/uwsgi.ini \
 && find /usr/local/searxng/searx/static  \( -name '*.html' -o -name '*.css' -o -name '*.js' \
    -o -name '*.svg' -o -name '*.ttf' -o -name '*.eot' \) \
    -type f -exec gzip -9 -k {} \+ -exec brotli --best {} \+

# Keep these arguments at the end to prevent redundant layer rebuilds
ARG LABEL_DATE=
ARG GIT_URL=unknown
ARG SEARXNG_GIT_VERSION=unknown
ARG SEARXNG_DOCKER_TAG=unknown
ARG LABEL_VCS_REF=
ARG LABEL_VCS_URL=
LABEL maintainer="searxng <${GIT_URL}>" \
      description="A privacy-respecting, hackable metasearch engine." \
      version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.schema-version="1.0" \
      org.label-schema.name="searxng" \
      org.label-schema.version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.url="${LABEL_VCS_URL}" \
      org.label-schema.vcs-ref=${LABEL_VCS_REF} \
      org.label-schema.vcs-url=${LABEL_VCS_URL} \
      org.label-schema.build-date="${LABEL_DATE}" \
      org.label-schema.usage="https://github.com/searxng/searxng-docker" \
      org.opencontainers.image.title="searxng" \
      org.opencontainers.image.version="${SEARXNG_DOCKER_TAG}" \
      org.opencontainers.image.url="${LABEL_VCS_URL}" \
      org.opencontainers.image.revision=${LABEL_VCS_REF} \
      org.opencontainers.image.source=${LABEL_VCS_URL} \
      org.opencontainers.image.created="${LABEL_DATE}" \
      org.opencontainers.image.documentation="https://github.com/searxng/searxng-docker"
      
ENTRYPOINT ["/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
  ```
</details>


<details>
  <summary>2 stages, with python:slim</summary>

  ```yaml
FROM python:3.13-slim AS builder

RUN mkdir /usr/local/searxng
WORKDIR /usr/local/searxng

RUN python3 -m venv /venv
ENV PATH=/venv/bin:$PATH

COPY requirements.txt ./requirements.txt

# Install build dependencies and Python packages
RUN apt-get update && apt-get install -y --no-install-recommends build-essential libpcre3-dev libxml2-dev libxslt1-dev  zlib1g-dev 
RUN --mount=type=cache,target=/root/.cache/pip  pip install "uwsgi~=2.0.0"
RUN --mount=type=cache,target=/root/.cache/pip  pip install -r requirements.txt 

COPY dockerfiles ./dockerfiles
COPY searx ./searx

ARG TIMESTAMP_SETTINGS=0
ARG TIMESTAMP_UWSGI=0
ARG VERSION_GITCOMMIT=unknown

RUN python3 -m compileall -q searx \
    && touch -c --date=@${TIMESTAMP_SETTINGS} searx/settings.yml \
    && touch -c --date=@${TIMESTAMP_UWSGI} dockerfiles/uwsgi.ini

# Final image stage
FROM python:3.13-slim

EXPOSE 8080
VOLUME /etc/searxng

ARG SEARXNG_GID=977
ARG SEARXNG_UID=977

RUN groupadd -g ${SEARXNG_GID} searxng && \
    useradd -u ${SEARXNG_UID} -d /usr/local/searxng -s /bin/sh -g searxng searxng

ENV INSTANCE_NAME=searxng \
    AUTOCOMPLETE= \
    BASE_URL= \
    SEARXNG_SETTINGS_PATH=/etc/searxng/settings.yml \
    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini \
    UWSGI_WORKERS=%k \
    UWSGI_THREADS=4

WORKDIR /usr/local/searxng

# Install necessary runtime packages
RUN apt-get update && apt-get install -y --no-install-recommends \
    brotli \
#    openssl \
    mailcap \
    libxml2 \
    libxslt1.1 \
    libpcre3 && \
    rm -rf /var/lib/apt/lists/*

# Copy only the necessary files from the builder stage
COPY --from=builder --chown=searxng:searxng /usr/local/searxng /usr/local/searxng
COPY --from=builder --chown=searxng:searxng /venv /venv

ENV PATH=/venv/bin:$PATH

USER searxng

RUN find /usr/local/searxng/searx/static \( -name '*.html' -o -name '*.css' -o -name '*.js' \
    -o -name '*.svg' -o -name '*.ttf' -o -name '*.eot' \) \
    -type f -exec gzip -9 -k {} + -exec brotli --best {} +

# Keep these arguments at the end to prevent redundant layer rebuilds
ARG LABEL_DATE=
ARG GIT_URL=unknown
ARG SEARXNG_GIT_VERSION=unknown
ARG SEARXNG_DOCKER_TAG=unknown
ARG LABEL_VCS_REF=
ARG LABEL_VCS_URL=
LABEL maintainer="searxng <${GIT_URL}>" \
      description="A privacy-respecting, hackable metasearch engine." \
      version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.schema-version="1.0" \
      org.label-schema.name="searxng" \
      org.label-schema.version="${SEARXNG_GIT_VERSION}" \
      org.label-schema.url="${LABEL_VCS_URL}" \
      org.label-schema.vcs-ref=${LABEL_VCS_REF} \
      org.label-schema.vcs-url=${LABEL_VCS_URL} \
      org.label-schema.build-date="${LABEL_DATE}" \
      org.label-schema.usage="https://github.com/searxng/searxng-docker" \
      org.opencontainers.image.title="searxng" \
      org.opencontainers.image.version="${SEARXNG_DOCKER_TAG}" \
      org.opencontainers.image.url="${LABEL_VCS_URL}" \
      org.opencontainers.image.revision=${LABEL_VCS_REF} \
      org.opencontainers.image.source=${LABEL_VCS_URL} \
      org.opencontainers.image.created="${LABEL_DATE}" \
      org.opencontainers.image.documentation="https://github.com/searxng/searxng-docker"
      
ENTRYPOINT ["/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
  ```  
</details>

**Comment by unixfox** [comment]
Would prefer to switch to debian/glibc based with this Dockerfile refactor. Especially if it allows to cut on the build time for ARM architectures.

**Comment by loorisr** [comment]
I've added the files in the PR with the different options.

The multistage dockerfile allows a better caching (and it can be tweaked more)

`RUN --mount=type=cache,target=/root/.cache/pip` allows to cache the built wheels

on arm64 and amd64, only uwsgi needs to be built a first time. Then with caching, build time is very fast.

File: Dockerfile.multistage [hunk 6]
```diff
@@ -0,0 +1,96 @@
+FROM python:3.13-slim AS builder
```
inetol: Let's make the multistage build the default, the "usual" (the actual Dockerfile build path) doesn't give us any advantages over this

Dockerfile.multistage
```diff
@@ -0,0 +1,96 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends build-essential libpcre3-dev libxml2-dev libxslt1-dev  zlib1g-dev 
+RUN --mount=type=cache,target=/root/.cache/pip  pip install "uwsgi~=2.0.0"
+RUN --mount=type=cache,target=/root/.cache/pip  pip install -r requirements.txt 
+
+COPY dockerfiles ./dockerfiles
+COPY searx ./searx
+
+ARG TIMESTAMP_SETTINGS=0
+ARG TIMESTAMP_UWSGI=0
+ARG VERSION_GITCOMMIT=unknown
```
inetol: This argument is unused (and not set), should be probably SEARXNG_GIT_VERSION instead?

Dockerfile.multistage
```diff
@@ -0,0 +1,96 @@
...
+  libxslt1.1 \
+  libpcre3 && \
+  rm -rf /var/lib/apt/lists/*
+
+# Copy only the necessary files from the builder stage
+COPY --from=builder --chown=searxng:searxng /usr/local/searxng /usr/local/searxng
+COPY --from=builder --chown=searxng:searxng /venv /venv
+
+ENV PATH=/venv/bin:$PATH
+
```
inetol: Missing `RUN mkdir /etc/searxng && chown searxng:searxng /etc/searxng`

Dockerfile.alpine
```diff
@@ -0,0 +1,73 @@
+FROM python:3.13-alpine
```
inetol: We can remove this file now (see Dockerfile.multistage top comment)

inetol (changes requested):
OK looking good, please review my comments

Dockerfile.multistage
loorisr: These arguments are set in https://github.com/searxng/searxng/blob/master/manage

`VERSION_GITCOMMIT` can be deleted you are right

inetol:
Surprisingly, I can't run the container now. It seems to be the function to get the SearXNG version. As the USER keyword is now set in the Dockerfile, there is no need to modify the user with `su` anymore (in fact, removing that solved the issue)

<details>
<summary>diff+</summary>

```diff
diff --git a/dockerfiles/docker-entrypoint.sh b/dockerfiles/docker-entrypoint.sh
index 0bad97b15..ae4af98c2 100755
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -49,9 +49,7 @@ do
 done
 
 get_searxng_version(){
-    su searxng -c \
-       'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' \
-       2>/dev/null
+    python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)" 2>/dev/null
 }
 
 SEARXNG_VERSION="$(get_searxng_version)"
```

</details>

loorisr:
good catch! Strange that on my setup with docker compose it is working fine (I guess it doesn't call get_searxng_version)

I've put USER in the dockerfile for security best practices.

inetol (approved):
LGTM, needs rebasing

Dockerfile
```diff
@@ -1,66 +1,71 @@
...
-RUN addgroup -g ${SEARXNG_GID} searxng && \
-    adduser -u ${SEARXNG_UID} -D -h /usr/local/searxng -s /bin/sh -G searxng searxng
+RUN groupadd -g ${SEARXNG_GID} searxng && \
+  useradd -u ${SEARXNG_UID} -d /usr/local/searxng -s /bin/sh -g searxng searxng
 
 ENV INSTANCE_NAME=searxng \
-    AUTOCOMPLETE= \
-    BASE_URL= \
-    MORTY_KEY= \
-    MORTY_URL= \
```
return42, line 16: Removing morty should not be a part of this PR .. for morty we have PR:

- https://github.com/searxng/searxng/pull/4594

Futher, in this PR here, the MORTY_KEY/URL are removed here but not in the [dockerfiles/docker-entrypoint.sh L87](https://github.com/searxng/searxng/pull/4463/files#diff-d939447b81599c5be18711c71019c48364c56a77dcbc39ebc50f40a540c5be1bL87)

loorisr: I put back MORTY vars

unixfox:
Could you please rebase? Thanks

PR #4464, opened 2025-03-07
Fix Qwant engine

### LINKED ISSUE #3929: qwant blocked by CAPTCHA
Its well known that qwant blocks very fast requests from a SearXNG instance:

We already discussed this issue more than a year ago in https://github.com/searxng/searxng/issues/2011#issuecomment-1553317619 and the documentation was a result of the issue:

- [Answer CAPTCHA from server’s IP](https://docs.searxng.org/admin/answer-captcha.html)

This method no longer works; the captcha dialog is not displayed in the WEB browser session.  This leads me to believe that qwant has changed its procedure for bot recognition and we should re-engineer the engine again.

----

I don't know if it might be related to bot-detection; I noticed that qwant makes requests to the server in the browser that look like a ping to me.  The returned response has a size of 0 Bytes

grafik

In this request a JSON structure is send to the qwant server:

grafik

This `search_id` comes from the previously sent query / e.g.

grafik

In the response to our query, beside the result items, we find this `search_id` 

grafik

To be clear, first the query is send (GET https://api.qwant.com/v3/search/web?), in the response there is a `search_id` and this search ID is pinged back to https://www.qwant.com/action/display (POST)

Since https://www.qwant.com/action/display is not a part of the API (api.qwant.com/v3) and the response is always 0 Byte I have the assumption that this could be related to the bot detection of qwant ... would have to be investigated further to validate this assumption.

-----

Related: 

- https://github.com/searxng/searxng/issues/2454

loorisr:
## What does this PR do?

This PR add tgp and llm arguments in the Qwant API request, to avoid beeing detected as a bot.

With current main, Qwant search engine is not useable.

make a search with !qwant and check that the results are not blocked by a captcha

## Author's checklist

loorisr

## Related issues

Related  #3929

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -135,6 +135,7 @@ def request(query, params):
         args['locale'] = q_locale
         args['safesearch'] = params['safesearch']
         args['count'] = 50
+        args['tgp'] = 3
         args['offset'] = (params['pageno'] - 1) * args['count']
 
     else:  # web, news, videos
@@ -142,6 +143,8 @@ def request(query, params):
         args['locale'] = q_locale
         args['safesearch'] = params['safesearch']
         args['count'] = 10
+        args['llm'] = 'false'
+        args['tgp'] = 3
         args['offset'] = (params['pageno'] - 1) * args['count']
 
     params['url'] = url + urlencode(args)


return42 (approved):
I think it will not really avoid the CAPTCHA from qwant .. but why not, lets give it a try.

loorisr:
on my computer if I go to 
* https://api.qwant.com/v3/search/web?q=fgh&locale=fr_FR&tgp=3  it works
* https://api.qwant.com/v3/search/web?q=fgh&locale=fr_FR  it generates a captcha url

loorisr:
Thanks for the merge

return42:
> * https://api.qwant.com/v3/search/web?q=fgh&locale=fr_FR&tgp=3  it works

Don't use a browser use the search in SearXNG and repeat it 30 times .. then your IP is blocked .. 

grafik

PR #4466, opened 2025-03-07
[enh] Static file with hashes

dalf:
## What does this PR do?

Look at the commit cc0ff4650b60f37766666373019d7b100b46bfa9

The code takes core of:

* map files
  ```json
  {
    "js/leaflet.js": "js/leaflet.db49d009.js",
    "js/leaflet.js.map": "js/leaflet.db49d009.js.map",
  }
  ```
* link inside JS and CSS:
  ```js
      searxng.loadStyle('css/leaflet.SEARXNG_HASH.css');
      searxng.loadScript('js/leaflet.SEARXNG_HASH.js', function () {
  ```

  `leaflet.SEARXNG_HASH.css` is replaced by `leaflet.db49d009.js`

   WARNING: the code does not handle the fact that we can have two files with the same name, for example `img/big/searxng.png` and `img/small/searxng.png`.

TODO: 

* [ ] Is `basename.SEARX_HASH.extension` the right way ?
* [x] `make static.build.restore` needs some care to manage files with hashes
* [x] Leaftlet icons are not found (the .less code of leaftlet does not add the SEARXNG_HASH)

* Avoid the browser cache issue
* Allow to cache the static URL for a long time for any installation (with/without docker, with/without reverse proxy)

* `make run`
* make some changes on the static files of the theme, rebuild the theme, check the hashes are updated

## Author's checklist

Feel free to close / update / whatever this PR


searx/webapp.py
```diff
@@ -244,24 +244,23 @@ def get_result_template(theme_name: str, template_name: str):
...
-            filename_with_theme = "themes/{}/{}".format(theme_name, values['filename'])
-            file_hash = static_files.get(filename_with_theme)
-            if file_hash:
-                values['filename'] = filename_with_theme
-        if get_setting('ui.static_use_hash') and file_hash:
-            suffix = "?" + file_hash
+            logical_filename = "themes/{}/{}".format(theme_name, values['filename'])
+            actual_filename = static_files.get(logical_filename)
+            if actual_filename:
+                values['filename'] = actual_filename
```
dalf: lines 254 and 255 are useless --> to remove

useless because lines 256 and 257 do exactly the same thing

dalf: done in a3b8cee

dalf:
There are hashes for the images in Leaflet. The leaflet.css file references these images, and if their URLs change, the .map file will break. One solution is to recompile leaflet.css from the .less files; the other solution is to avoid adding the hashes. Commit 
2761f78768600c7dc3feef395c462f1a60afae1a implements the second choice (no hash).

The code is not very clean:
* the variable names might neither matched their usages nor being harmonized
* there are multiple new imports for example in webapp.py, not sure this is the right place

but this issue aside, it seems to be good enough.

searx/webutils.py
```diff
@@ -178,30 +178,45 @@ def get_themes(templates_path):
...
+        if assets_file.is_file():
+            # assets.json exist
+            with assets_file.open("r", encoding="utf-8") as f:
+                assets = json.load(f)
+                for rel_logical_filename, rel_actual_filename in assets.items():
+                    logical_filename = f"themes/{theme_dir.name}/{rel_logical_filename}"
+                    actual_filename = f"themes/{theme_dir.name}/{rel_actual_filename}"
+                    results[logical_filename] = actual_filename
+        else:
+            # assets.json does not exist
```
dalf, line 211: The simple theme is the only theme, but the code still supports multiple themes --> the existing themes won't have the assets.json files, so the following code is here for backward compatibility (= for custom themes in the different forks).

return42:
Isn't it simpler/better to use `[hash]` from [roolup.js](https://rollupjs.org/configuration-options/#output-assetfilenames) / e.g.:

```diff
index f8d3e6757..45ae8a574 100644
--- a/client/simple/vite.config.js
+++ b/client/simple/vite.config.js
@@ -107,9 +107,9 @@ export default defineConfig({
 
       // file naming conventions / pathnames are relative to outDir (PATH.dist)
       output: {
-        entryFileNames: "[name].js",
-        chunkFileNames: "[name].js",
-        assetFileNames: "[name].[ext]",
+        entryFileNames: "[name].[hash].js",
+        chunkFileNames: "[name].[hash].js",
+        assetFileNames: "[name].[hash].[ext]",
         // Vite does not support "rollupOptions.output.sourcemap".
         // Please use "build.sourcemap" instead.
         // sourcemap: true,
```

with `./manage vite.simple.build` we get a hash values in the file names of JS and CSS files and the manifest (``searxng/searx/static/themes/simple/manifest.json``) is like this:

```json
{
  "js/searxng.head.js": {
    "file": "js/searxng.head.min.5vbFT7cW.js",
    "name": "js/searxng.head.min",
    "src": "js/searxng.head.js",
    "isEntry": true
  },
  "js/searxng.js": {
    "file": "js/searxng.min.CoHvYy0-.js",
    "name": "js/searxng.min",
    "src": "js/searxng.js",
    "isEntry": true
  },
  "less/rss.less": {
    "file": "css/rss.min.aJKEdM9t.css",
    "src": "less/rss.less",
    "isEntry": true
  },
  "less/style-ltr.less": {
    "file": "css/searxng.min.vfsuff-7.css",
    "src": "less/style-ltr.less",
    "isEntry": true
  },
  "less/style-rtl.less": {
    "file": "css/searxng-rtl.min.DI7zUmR_.css",
    "src": "less/style-rtl.less",
    "isEntry": true
  }
}
```

dalf:
The custom code in this PR add a hash for the all the files (with explicite exceptions)
Can roolup do that, including for example leaflet.js, img/searxng.svg ?
How does it handle references between the ressources ? For example https://github.com/searxng/searxng/blob/8b1d73c5b96e9059c397b62f084aa8cee378218b/client/simple/src/less/index.less#L9

return42:
At the moment, this is not possible because we are copying the assets (images).

With the PR:

- https://github.com/searxng/searxng/pull/4284

we moved to vite, but at the moment we are not using it properly yet, we are currently only doing with it what we did before with grunt.

For example, we currently only use the build process and assets such as images are copied .. the path to real client development is not yet complete (the migration to vite is not finished). As soon all objects are managed by vite at some point, then all this is possible, but in the current state the framework is not aware of the connection between CSS, JS and assets.

dalf:
Ok, I don't know what you want to prioritize.
The PR is here, feel free to merge / put aside / close / whatever.

PR #4467, opened 2025-03-07
[fix] presearch engine: domain sometimes included in beginning of titles

Aadniz:
## What does this PR do?

Fixes an issue where Presearch would sometimes show the domain before the actual title. This is not an issue from searxng per say, but rather from Presearch. This issue could be fixed some time in the future by their end.

You can trigger this bug by finding/starting up an instance, and search for `!ps google`. The result should be as shown below.

image

Here, I created a new function called `_fix_title(title, url)` that takes in the title and url. The long if statement is due to edge cases, such as the title being equal the domain, the title starting with the url as intended by the website owner (Maybe something like `example.com - an example website`), and the slash is for good measurement, it seems like this bug will never show the domain + path.
With this chain of if statements, it should match this bug pretty well.

The PR also takes care of the previous issue https://github.com/searxng/searxng/pull/4458

It looks ascetically bad and could misrepresent what the website actually wants to have in their title.

```shell
$ cd searxng
$ git switch presearch-fix-2
$ python3 setup.py build
$ searxng-run
```

## Author's checklist

Previous discussion with @Bnyro in https://github.com/searxng/searxng/pull/4458 (note it is slightly changed from what was previous discussed https://github.com/searxng/searxng/compare/master...Aadniz:searxng:presearch-fix)

### DIFF
--- a/searx/engines/presearch.py
+++ b/searx/engines/presearch.py
@@ -64,7 +64,7 @@ Implementations
 
 """
 
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse
 from searx import locales
 from searx.network import get
 from searx.utils import gen_useragent, html_to_text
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
LGTM, thanks!

PR #4476, opened 2025-03-10
[web-client] simple theme: dynamic vertical height

bearz314:
On mobile devices with browser chrome (UI) that recedes on scroll, 'vh' includes the height covered by chrome, while 'dvh' refers to the visible screen. By using 'dvh' the footer is immediately visible on the homepage without scrollbar on mobile devices.

## What does this PR do?

Use 'dvh' for body if browser supports 'dvh', otherwise fallback to use 'vh'.
'dvh' is [widely supported by browsers](https://caniuse.com/?search=dvh).

On desktop devices, the footer is immediately visible.
On mobile devices, the footer is hidden behind the browser chrome, and scroll bars are presented.
See comparison below (iPhone and Samsung) via BrowserStack.

Following the readme file, first:
https://github.com/searxng/searxng/blob/8b1d73c5b96e9059c397b62f084aa8cee378218b/client/simple/README.rst?plain=1#L5-L8

Then:
https://github.com/searxng/searxng/blob/8b1d73c5b96e9059c397b62f084aa8cee378218b/client/simple/README.rst?plain=1#L22-L24

## Related issues

Closes #4410

edited


return42 (approved):
Many thanks for your research and the detailed and helpful description .. :+1:

Issue #4190, opened 2025-01-15
Gallery view overlaps category bar


return42:
![grafik](https://github.com/user-attachments/assets/c18b0f07-a3b5-4be7-a780-41026977b52e)

--- 

update .. related to comment:

- https://github.com/searxng/searxng/issues/3509#issuecomment-2228396871

igorzhilin:
![Screenshot_20250116_015643_Iceraven](https://github.com/user-attachments/assets/202676c8-643a-4ef6-896e-f21d254c4261)
Is this issue related to the problem as in this screenshot on mobile version? The right sidebar shifts the left result column away from the screen beyond the left screen edge.

return42:
> Is this issue related to the problem as in this screenshot on mobile version? 

Nope .. what you address is covered by 

- https://github.com/searxng/searxng/issues/4187#event-15954737988

Issue #3509, opened 2024-05-24
(theme) (area: ui) Image viewer hides autocomplete suggestions


allendema:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
2024.5.10+5f29bf66b Public instance

**How did you install SearXNG?**
N/A

**What happened?**
See printscreen below.

**How To Reproduce**
Under preferences tick the option to center results.
Search for something with !images
Select an image to view then start entering a new query.

**Expected behavior**
To be able to read autocomplete suggestions even when an image is clicked.

**Screenshots & Logs**

![searxNG_Bildschirmfoto vom 2024-05-24 23-38-27](https://github.com/searxng/searxng/assets/64094914/ffbd85e3-4cdb-460e-babf-30fd6dd96f91)

**Additional context**

return42:
Not only is the auto-completion hidden, the preview with the image details also hides the selection menus and there is a small overlap on the category tabs ...

![image](https://github.com/user-attachments/assets/1275fba2-f6ab-45b4-a004-d3dee35ed28d)

dalf:
> Not only is the auto-completion hidden, the preview with the image details also hides the selection menus and there is a small overlap on the category tabs ...

This variable must be adjusted to fix this issue (not the autocompletion issue) :
https://github.com/searxng/searxng/blob/9a4fa7cc4f0339a1ea696f6fb7068fd6156f03e2/searx/static/themes/simple/src/less/definitions.less#L256

return42:
> This variable must be adjusted to fix this issue (not the autocompletion issue) :

That was my first thought too ... but unfortunately that's not enough ... I'm afraid we'll have to rearrange the search header ... I'm still working on a solution ([WIP](https://github.com/return42/searxng/commits/fix-3509)) and will get back to you with a PR ...

PR #4501, opened 2025-03-15
[fix] image deatils view: vertical overflow in results.image-detail-open

return42:
If the information on the image exceeds the visible area in the detail view,
there must be a way to access this information via scroll.

grafik

grafik


dalf:
The original design is to avoid scroll and to reduce the image height to fit into the available browser window = there is no scroll, so the user can see all the information easily. This is especially useful when the user move from one image to another.

I remember to have spent some time to make it work, may be a previous commit has broken that.

Bnyro:
> I remember to have spent some time to make it work, may be a previous commit has broken that.
>
We probably just added some new info fields without adapting the size measurements.

return42:
> We probably just added some new info fields without adapting the size measurements.

And I remember we also had to fix some issues related to the image detail view. An uncompleted list:

- https://github.com/searxng/searxng/issues/3509
- https://github.com/searxng/searxng/issues/4190

It is also possible that we have broken the layout for some time; it is only noticeable if you have a hit list like the one shown above which only contains a few images and the displayed image is in portrait format .. and the viewport has an unfavorable size ... I had difficulty finding examples of this.

However I think its worth to *"spent some time to make it work"* again.

return42:
> However I think its worth to _"spent some time to make it work"_ again.

There are new fields in the lower area, so the height of the image (top) must be reduced, see [diff](https://github.com/searxng/searxng/compare/ba4c0ca7162c3a31da1559c29c97fd516e119c8a..ceacfbcb7831275e140e81305cb26797ff4a8ae7)  or commit: https://github.com/searxng/searxng/pull/4501/commits/e6eeaf266adb5163d2d6ad9c87ac2e7e009a4e83

PR #4502, opened 2025-03-16
[feat] engines: add Quark engine

hezhijie0327:
## What does this PR do?

1. add Quark (夸克搜索, Formerly Shenma) engine (General)

ref: https://m.quark.cn/
Wikidata: https://www.wikidata.org/wiki/Q48816502

---

Note:
1. Quark is a search engine from Alibaba Group
2. Quark may return a CAPTCHA challenge after 9 requests in a short period. Typically, the ban duration is around 15 minutes.
3. Quark returns a variety of different `sc` values on a single page, depending on the query type.
4. Quark has 2 domains, `m.quark.cn` and `quark.sm.cn`
5. Quark's data struct is very weird, sometime the same `sc` type will return different JSON struct. eg: `ss_doc`, `ss_kv`

---

|Type|Screenshot|
|-|-|
|Results|image|
|Images|image|
|CAPTCHA|image|

### DIFF
--- a/searx/engines/quark.py
+++ b/searx/engines/quark.py
@@ -1,0 +1,383 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Quark (Shenma) search engine for searxng"""
+
+from urllib.parse import urlencode
+from datetime import datetime
+import re
+import json
+
+from searx.utils import html_to_text, gen_useragent
+from searx.exceptions import SearxEngineAPIException, SearxEngineCaptchaException
+
+# Metadata
+about = {
+    "website": "https://m.quark.cn/",
+    "wikidata_id": "Q48816502",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "HTML",
+    "language": "zh",
+}
+
+# Engine Configuration
+categories = []
+paging = True
+results_per_page = 10
+
+quark_category = 'general'
+
+time_range_support = True
+time_range_dict = {'day': '4', 'week': '3', 'month': '2', 'year': '1'}
+
+CAPTCHA_PATTERN = r'\{[^{]*?"action"\s*:\s*"captcha"\s*,\s*"url"\s*:\s*"([^"]+)"[^{]*?\}'
+
+
+def is_alibaba_captcha(html):
+    """
+    Detects if the response contains an Alibaba X5SEC CAPTCHA page.
+
+    Quark may return a CAPTCHA challenge after 9 requests in a short period.
+
+    Typically, the ban duration is around 15 minutes.
+    """
+    return bool(re.search(CAPTCHA_PATTERN, html))
+
+
+def init(_):
+    if quark_category not in ('general', 'images'):
+        raise SearxEngineAPIException(f"Unsupported category: {quark_category}")
+
+
+def request(query, params):
+    page_num = params["pageno"]
+
+    category_config = {
+        'general': {
+            'endpoint': 'https://m.quark.cn/s',
+            'params': {
+                "q": query,
+                "layout": "html",
+                "page": page_num,
+            },
+        },
+        'images': {
+            'endpoint': 'https://vt.sm.cn/api/pic/list',
+            'params': {
+                "query": query,
+                "limit": results_per_page,
+                "start": (page_num - 1) * results_per_page,
+            },
+        },
+    }
+
+    query_params = category_config[quark_category]['params']
+    query_url = category_config[quark_category]['endpoint']
+
+    if time_range_dict.get(params['time_range']) and quark_category == 'general':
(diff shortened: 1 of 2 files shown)


searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+
+            results.append(
+                {
+                    "template": "images.html",
+                    "url": item.get("imgUrl"),
+                    "thumbnail_src": item.get("img"),
+                    "img_src": item.get("bigPicUrl"),
+                    "title": item.get("title"),
+                    "source": item.get("site"),
+                    "resolution": f"{width} x {height}",
```
Bnyro:
```suggestion
                    "resolution": f"{item['width']} x {item['height']}",
```
inlining the width and height makes sense here

searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+            [
+                data.get('establish_time'),
+                data.get('company_status'),
+                data.get('controled_type'),
+                data.get('company_type'),
+                data.get('capital'),
+                data.get('address'),
+                data.get('business_scope'),
+            ],
+        )
```
Bnyro:
```suggestion
      info for info in [
            data.get('establish_time'),
            data.get('company_status'),
            data.get('controled_type'),
            data.get('company_type'),
            data.get('capital'),
            data.get('address'),
            data.get('business_scope'),
        ] if info,
    )
```
I think that's easier to read because not everybody knows what `filter(None, ...)` does.

searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+
+    query_params = category_config[quark_category]['params']
+    query_url = category_config[quark_category]['endpoint']
+
+    if time_range_dict.get(params['time_range']) and quark_category == 'general':
+        query_params["tl_request"] = time_range_dict.get(params['time_range'])
+
+    params["url"] = f"{query_url}?{urlencode(query_params)}"
+    params["headers"] = {
+        "User-Agent": searx_useragent(),
```
Bnyro: We should probably use https://github.com/searxng/searxng/blob/84636ef49acc6e74421c23a80f6f63aa4ccc93c7/searx/utils.py#L72 instead? Otherwise they could just block SearXNGs specific user agent.

searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+paging = True
+results_per_page = 10
+
+quark_category = 'general'
+
+time_range_support = True
+time_range_dict = {'day': '4', 'week': '3', 'month': '2', 'year': '1'}
+
+
+def is_alibaba_captcha(html):
```
Bnyro, line 35: This should probably return the captcha url or None if there's no captcha and let the caller function throw the Exception, I don't really like throwing exceptions as a side effect of a normal-looking function call.

searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+        "url": data.get('title_url'),
+        "content": html_to_text(content),
+    }
+
+
+def parse_life_show_general_image(data):
+    results = []
+    for item in data.get('image', []):
+        width = item.get("width")
+        height = item.get("height")
```
Bnyro: Again, this should be inlined

searx/engines/quark.py
```diff
@@ -0,0 +1,293 @@
...
+
+def is_alibaba_captcha(html):
+    """
+    Detects if the response contains an Alibaba X5SEC CAPTCHA page.
+
+    Quark may return a CAPTCHA challenge after 9 requests in a short period.
+
+    Typically, the ban duration is around 15 minutes.
+    """
+    pattern = r'\{[^{]*?"action"\s*:\s*"captcha"\s*,\s*"url"\s*:\s*"([^"]+)"[^{]*?\}'
```
Bnyro: This regex should probably be defined globally as a constant, e.g. `CAPTCHA_PATTERN`, similar to how we do it at other engines.

hezhijie0327:
Done.

For the CAPTCHA part, after some testing, found that `captcha_url` is unnecessary for unbanning the session. So, change it to Bool.

searx/engines/quark.py
```diff
@@ -0,0 +1,380 @@
...
+        for match in matches:
+            data = json.loads(match)
+            initial_data = data.get('data', {}).get('initialData', {})
+            extra_data = data.get('extraData', {})
+
+            source_category = extra_data.get('sc')
+            source_category_parsers = {
+                'addition': parse_addition,
+                'ai_page': parse_ai_page,
+                'baike_sc': parse_baike_sc,
```
return42: I think this catalog not needs to be build again on every loop cycle .. place it above, outside o the loop.

hezhijie0327: done

return42 (approved):
Thanks :+1:

PR #4508, opened 2025-03-17
[fix] duckduckgo: show proper source url of answers

Bnyro:
## What does this PR do?
- this PR extracts the real source url of duckduckgo's answers

- the quick answer should directly redirect you to the source of the answer, not to duckduckgo's search results.
- that enforces the same behavior as with all the other search engine modules

- `!ddg mixed raster content`


return42 (approved):
> the quick answer should directly redirect you to the source of the answer, not to duckduckgo's search results.

Yeah, much better :-)  .. thanks :+1:

PR #4509, opened 2025-03-17
[feat] engines: add reuters news engine

Bnyro:
## What does this PR do?
- add reuters.com news engine

- reuters is one of the worlds largest news agencies

- search `!reu query`, optionally with time ranges and pagination

### DIFF
--- a/searx/engines/reuters.py
+++ b/searx/engines/reuters.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Reuters_ (news) is an international news agency.
+
+.. _Reuters: https://www.reuters.com
+
+Configuration
+=============
+
+The engine has the following additional settings:
+
+- :py:obj:`sort_order`
+
+.. code:: yaml
+
+   - name: reuters
+     engine: reuters
+     shortcut: reu
+     sort_order: "relevance"
+
+
+Implementations
+===============
+
+"""
+
+from json import dumps
+from urllib.parse import quote_plus
+from datetime import datetime, timedelta
+
+from searx.result_types import EngineResults
+
+about = {
+    "website": "https://www.reuters.com",
+    "wikidata_id": "Q130879",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["news"]
+time_range_support = True
+paging = True
+
+base_url = "https://www.reuters.com"
+
+results_per_page = 20
+sort_order = "relevance"
+"""Sort order, one of ``relevance``, ``display_date:desc`` or ``display_data:asc``."""
+
+time_range_duration_map = {
+    "day": 1,
+    "week": 7,
+    "month": 30,
+    "year": 365,
+}
+
+
+def request(query, params):
+    args = {
+        "keyword": query,
+        "offset": (params["pageno"] - 1) * results_per_page,
+        "orderby": sort_order,
+        "size": results_per_page,
+        "website": "reuters",
+    }
+    if params["time_range"]:
+        time_diff_days = time_range_duration_map[params["time_range"]]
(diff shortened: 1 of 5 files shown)


searx/engines/reuters.py
```diff
@@ -0,0 +1,90 @@
...
+    return params
+
+
+def response(resp) -> EngineResults:
+    res = EngineResults()
+
+    for result in resp.json().get("result", {}).get("articles", []):
+        res.add(
+            res.types.MainResult(
+                url=base_url + result["canonical_url"],
```
return42, line 82: I implemented returned list by instance of [EngineResults](https://docs.searxng.org/dev/engines/index.html#searx.result_types.EngineResults) with [MainResult](https://docs.searxng.org/dev/result_types/main/mainresult.html) items.

But first we should review https://github.com/searxng/searxng/pull/4538 .. and then rebase this PR after #4538 has been merged.

searx/engines/reuters.py
```diff
@@ -0,0 +1,90 @@
...
+                title=result["web"],
+                content=result["description"],
+                thumbnail=result.get("thumbnail", {}).get("url", ""),
```
return42, line 85: Not sure we should use Reuter's *thumbnails* .. The URLs point to full (desktop) size images .. here is one example for such an "*thumbnail*" : https://cloudfront-us-east-2.images.arcpublishing.com/reuters/7NDGEYHE4BJZBGMT4GYQ5IP4GI.jpg

searx/result_types/_base.py
```diff
@@ -347,6 +358,24 @@ class MainResult(Result):  # pylint: disable=missing-class-docstring
...
+
+    views: str = ""
+    """View count in humanized number format."""
+
+    author: str = ""
+    """Author of the title."""
+
+    metadata: str = ""
+    """Miscellaneous metadata."""
+
```
return42, line 378: > I implemented returned list by instance of [EngineResults](https://docs.searxng.org/dev/engines/index.html#searx.result_types.EngineResults) with [MainResult](https://docs.searxng.org/dev/result_types/main/mainresult.html) items.

.. to do so, the missing properties of a MainResult had been added here.

searx/webapp.py
```diff
@@ -694,14 +694,6 @@ def search():
...
                 result['title'] = highlight_content(escape(result['title'] or ''), search_query.query)
 
-        if getattr(result, 'publishedDate', None):  # do not try to get a date from an empty string or a None type
-            try:  # test if publishedDate >= 1900 (datetime module bug)
-                result['pubdate'] = result['publishedDate'].strftime('%Y-%m-%d %H:%M:%S%z')
-            except ValueError:
-                result['publishedDate'] = None
-            else:
-                result['publishedDate'] = webutils.searxng_l10n_timespan(result['publishedDate'])
-
```
return42, line 704: Normalize the *date* fields of a Result is not a job related to `webapp.py` .. this task has been moved to the `Result.normalize_result_fields()` method.

return42 (approved):
> .. and then rebase this PR after https://github.com/searxng/searxng/pull/4538 has been merged.

FYI: I rebased on current master (solve conflicts) and left some comments below.

Beyond that, engine is ready to merge from my side. :+1:

searx/engines/reuters.py
return42: @Bnyro after using Reuter's engine a while I think we should remove the thumbnails (the thumbnail size is a multiple of MB / I have seen up to 5MB image sizes and none of them is below 1 MB) .. what do you think?

Bnyro: Yeah, I experienced slow loading times too, so that's reasonable. Or we figure out if they have thumbnails with lower size somewhere, but I couldn't think anything about it.

return42: > but I couldn't think anything about it.

As far I remember I have looked for real thumbs in my review, but there wasn't a field in the returned JSON.

PR #4514, opened 2025-03-19
[feat] engines: add Niconico videos engine

hezhijie0327:
## What does this PR do?

add Niconico (ニコニコ) engine (Videos)

ref: https://www.nicovideo.jp/
lang: `ja`
Wikidata: https://www.wikidata.org/wiki/Q697233

|Screenshot|
|-|
|image|

[Niconico](https://en.wikipedia.org/wiki/Niconico) is a Japanese video sharing service

search `!nico query`, optionally with time ranges and pagination

### DIFF
--- a/searx/engines/niconico.py
+++ b/searx/engines/niconico.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Niconico search engine for searxng"""
+
+from urllib.parse import urlencode
+from datetime import datetime, timedelta
+from lxml import html
+
+from searx.utils import eval_xpath_getindex, eval_xpath_list, eval_xpath, extract_text
+
+about = {
+    "website": "https://www.nicovideo.jp/",
+    "wikidata_id": "Q697233",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "HTML",
+    "language": "ja",
+}
+
+categories = ["videos"]
+paging = True
+
+time_range_support = True
+time_range_dict = {"day": 1, "week": 7, "month": 30, "year": 365}
+
+base_url = "https://www.nicovideo.jp"
+embed_url = "https://embed.nicovideo.jp"
+
+results_xpath = '//li[@data-video-item]'
+url_xpath = './/a[@class="itemThumbWrap"]/@href'
+video_length_xpath = './/span[@class="videoLength"]'
+upload_time_xpath = './/p[@class="itemTime"]//span[@class="time"]/text()'
+title_xpath = './/p[@class="itemTitle"]/a'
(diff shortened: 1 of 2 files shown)


searx/engines/niconico.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def request(query, params):
+    query_params = {"page": params['pageno']}
+
+    if time_range_dict.get(params['time_range']):
+        now = int(time.time())
+        past = now - time_range_dict[params["time_range"]]
+
+        query_params['end'] = time.strftime('%Y-%m-%d', time.gmtime(now))
```
Bnyro: The `end` parameter isn't required actually, I think it makes sense to let the server set the current time automatically.

searx/engines/niconico.py
```diff
@@ -0,0 +1,85 @@
...
+
+base_url = "https://www.nicovideo.jp"
+
+
+def request(query, params):
+    query_params = {"page": params['pageno']}
+
+    if time_range_dict.get(params['time_range']):
+        now = int(time.time())
+        past = now - time_range_dict[params["time_range"]]
```
Bnyro: I think using `datetime` and `timedelta` would be cleaner here than calculating with integers, see https://github.com/searxng/searxng/pull/4509/files#diff-704c60f5e4ecffba5ee315ed9698f013f905b5923d232a313632dee21e5df528R46 for example.

searx/engines/niconico.py
```diff
@@ -0,0 +1,85 @@
...
+    params['url'] = f"{base_url}/search/{query}?{urlencode(query_params)}"
+    return params
+
+
+def response(resp):
+    results = []
+    dom = html.fromstring(resp.text)
+
+    for item in dom.xpath('//li[@data-video-item]'):
+        relative_url = extract_text(item.xpath('.//a[@class="itemThumbWrap"]/@href')[0])
```
Bnyro: Please use `eval_xpath` from `searx.utils` instead of `item.xpath` everywhere here.

searx/engines/niconico.py
```diff
@@ -0,0 +1,85 @@
...
+            try:
+                published_date = datetime.strptime(upload_time, "%Y/%m/%d %H:%M")
+            except ValueError:
+                pass
+
+        results.append(
+            {
+                'template': 'videos.html',
+                'title': extract_text(item.xpath('.//p[@class="itemTitle"]/a')),
+                'content': extract_text(item.xpath('.//p[@class="itemDescription"]/@title')[0]),
```
Bnyro: Please use `eval_xpath_index` here from `searx.utils`.

hezhijie0327:
Done.

return42 (approved):
Thanks :+1:

PR #4517, opened 2025-03-19
[fix] hostname replace plugin doesn't work in the infobox

Bnyro:
## What does this PR do?
- remove and replace urls from the infobox
- do some small cleanup to remove duplicated url parsing logic
- search through the blocked patterns first before replacing things in the url, because the url rewriting is unnecessary if the result matches a blocked pattern and thus removed anyways

- I've been annoyed by the fact that infobox links are not rewritten properly, thus the PR.

- Test the hostnames plugin with engines using infobox `urls`, e.g. wikidata

## Related issues
- partially addresses https://github.com/searxng/searxng/issues/1348

### DIFF
--- a/searx/plugins/hostnames.py
+++ b/searx/plugins/hostnames.py
@@ -110,7 +110,8 @@ preference_section = 'general'
 plugin_id = 'hostnames'
 
 parsed = 'parsed_url'
-_url_fields = ['iframe_src', 'audio_src']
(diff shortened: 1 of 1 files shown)


return42:
Can we first review:

return42:
Superseded by 

- https://github.com/searxng/searxng/pull/4538#issuecomment-2751500506

PR #4518, opened 2025-03-20
[fix] presearch engine: News and Videos formatted incorrectly

Aadniz:
## What does this PR do?

Fixes an issue where titles and descriptions of Presearch news and video results would sometimes give incorrectly formatted text showing text such as `&quot;`, `&apos;`, `&quot;` etc.. It also fixes the issue where it should show HTML code in the results, showing `<em></em>` (same issue as https://github.com/searxng/searxng/pull/4458)

image

image

Lastly, it fixes a bug on their end where the time on Presearch news page would simply return `</a>`, resulting in a slash at the end of the metadata section.

image

To fix these issues, I put `html_to_text` around the affected parts, and checking if `html_to_text(item.get('time')).strip()` is not an empty string.

I'm unsure if the comment I added regarding time formats at the top are OK to include in this PR, let me know otherwise.

Looks ascetically bad.

```shell
$ cd searxng
$ git switch presearch-fix-3
$ python3 setup.py build
$ searxng-run
```

### DIFF
--- a/searx/engines/presearch.py
+++ b/searx/engines/presearch.py
@@ -58,6 +58,12 @@ have to set these values in both requests we send to Presearch; in the first
 request to get the request-ID from Presearch and in the final request to get the
 result list (see ``send_accept_language_header``).
 
+The time format returned by Presearch varies depending on the language set.
+Multiple different formats can be supported by using ``dateutil`` parser, but
+it doesn't support formats such as "N time ago", "vor N time" (German),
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
I can confirm the issue, thank you!

You don't need to paste the full setup instructions here in the `How to test this PR locally?` part, a simple example query to compare the difference before and after these PR's changes is more useful.

PR #4523, opened 2025-03-20
[refactor] duration strings: move parsing logic to utils.py

Bnyro:
## What does this PR do?
- this PR creates a new helper function `parse_duration_string` and applies it where needed

- we currently have duplicated logic for parsing time duration strings

- `!bib !psvid !ptb test`

## Author's checklist
- we should probably use `timedelta` for all duration objects instead of using strings sometimes, in order to allow us to properly format video lengths consistently across different search engines

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Utility functions for the engines
-
-"""
+"""Utility functions for the engines"""
 
 from __future__ import annotations
 
@@ -18,6 +16,7 @@ from random import choice
 from html.parser import HTMLParser
 from html import escape
 from urllib.parse import urljoin, urlparse, parse_qs, urlencode
+from datetime import timedelta
 from markdown_it import MarkdownIt
 
 from lxml import html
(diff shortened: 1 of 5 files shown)


searx/utils.py
```diff
@@ -831,3 +830,27 @@ def js_variable_to_python(js_variable):
     s = s.replace(chr(1), ':')
     # load the JSON and return the result
     return json.loads(s)
+
+
+def parse_timediff_string(timediff_str: str) -> timedelta | None:
+    """Parse a time string in format MM:SS or HH:MM:SS and convert it to a `timedelta` object.
+
+    Returns None if the provided string doesn't match any of the formats.
+    """
```
return42: Suggestion to simplify "time_parts":

```python
>>> duration = "12:42"
>>> hour, min, sec = [int(x) for x in (duration.split(":") + ["00"])[:3]]

>>> timedelta(hours=hour, minutes=min, seconds=sec)
datetime.timedelta(seconds=45720)
```

Bnyro: The real value would be `datetime.timedelta(seconds=762)`, `["00"]` has to be appended to the start of the list, not the end :)

Bnyro: Seems like a good suggestion though :+1:

return42 (approved):
:+1:

PR #4525, opened 2025-03-21
[fix] Results.url: don't normalize www.example.com to example.com

return42:
## What does this PR do?

Hostname "www" in URL results can't be normalized to an empty string:

- https://www.tu-darmstadt.de/
- https://tu-darmstadt.de/

Try to search for "TU Darmstadt"


Bnyro (approved):
Thank you for looking into this so quickly!

PR #4530, opened 2025-03-21
[fix] make data.traits - partial revert of commit 30a8204

### LINKED ISSUE #4519: Many language filters missing
Branch: master
Version: 2025.3.19+40feede51

**What happened?**
A lot of previously available language filters are suddenly missing on the search results page.

For comparison you can see the available language filters on [searx.foss.family](https://searx.foss.family/) (that as of writing is on version 2025.2.28+355883f36) or [sx.catgirl.cloud](https://sx.catgirl.cloud/) (on version 2025.02.09+a1e2b254) and compare them to a more up-to-date instance like [search.ononoki.org](https://search.ononoki.org/) (on version 2025.3.19+40feede51) to see that a lot of language filter options are missing

**How To Reproduce**

1. Search something on an up-to-date instance like [search.ononoki.org](https://search.ononoki.org/)
2. Click on the language filter option on the top left
3. See that in the dropdown menu that a lot of language options are missing

**Expected behavior**

That previously available language filters would still be there like they were on older versions of SearxNG

**Screenshots & Logs**

Screenshots of the language filters: 
[searx.foss.family](https://i.ibb.co/qYTtC5jk/searx-foss-family-language-filters.png) , [sx.catgirl.cloud](https://i.ibb.co/2GVv0Yq/sx-catgirl-cloud-language-filters.png), 
[search.ononoki.org](https://i.ibb.co/990v3CQn/search-ononoki-org-language-filters.png) (up-to-date instance)

**Additional context**

Lack of these language filters hinders the ability to easily search for results in those specific languages

return42:
The entries in the catalog of search languages are build up from the "Engine Traits" [1] and which entries are included in the catalog is controlled qby two threshold values 

https://github.com/searxng/searxng/blob/96a6e3dcb2283fa7ad9db4172a00582073a166d7/searxng_extra/update/update_engine_traits.py#L104-L105

If possible, the values should ensure that no languages or regions disappear from the catalog of search languages [3].

The threshold values should have been adjusted in commit 30a8204:

- ``min_eng_per_region = 18``
- ``min_eng_per_lang = 22``

Because the threshold values were not adjusted, many entries were missing in the search language catalog.  This bug has been fixed with this patch: the threshold values have been adjusted and the catalog of search languages has been completed again.

[1] https://docs.searxng.org/dev/engines/enginelib.html#module-searx.enginelib.traits
[3] https://github.com/searxng/searxng/blob/master/searx/sxng_locales.py

----

- Closes: https://github.com/searxng/searxng/issues/4519


Bnyro:
If you're free, could you please explain how we are currently determining the threshold values? And why have they been incremented by 2 (for lang) and 4 (for region) with https://github.com/searxng/searxng/pull/4401? Shouldn't they have been increased by 1 each, because one single engine has been fixed (bing in this case)?

Bnyro (approved):
Looks good to me, the comment above doesn't block this PR from being merged 👍

return42:
> could you please explain how we are currently determining the threshold values?

The number of languages from the original engines changes from time to time, for example when new languages are added to an engine.   The threshold values are then adjusted so that the catalog of search languages does not change if possible.

> And why have they been incremented by 2 (for lang) and 4 (for region) with https://github.com/searxng/searxng/pull/4401? 

I can't remember why I did that back then, it was a mistake.  The threshold values are now set here with this PR so that the old catalog of search languages is displayed again.

> Shouldn't they have been increased by 1 each, because one single engine has been fixed (bing in this case)?

If only the languages of one engine change, then you could assume that .. but usually additional languages from other engines also change.

However, the factor would not always be 1, it can also be a multiple of 1 if the fetch method is imported in other modules or if there are several configured engines for a module.

For example, `searx.engines.bing.fetch_traits` is used by four engines (bing-web,-images,-news,-videos).

PR #4531, opened 2025-03-21
[feat] engine: add bitchute

naughtymommy42069:
## What does this PR do?

adds bitchute

1. platform supports free expression
2. requested on searx: https://github.com/searx/searx/issues/1322
3. completes the trinity: odysee, rumble, and bitchute.

make run
and use the bang !bit to test

### DIFF
--- a/searx/engines/bitchute.py
+++ b/searx/engines/bitchute.py
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""bitchute (Videos)"""
+
+from json import dumps
+from datetime import datetime
+from searx.utils import html_to_text
+
+about = {
+    "website": 'https://bitchute.com',
+    "wikidata_id": "Q45287179",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://api.bitchute.com/api/beta/search/videos"
+categories = ['videos']
+paging = True
+results_per_page = 20
+
+
+def request(query, params):
+
+    start_index = (params["pageno"] - 1) * results_per_page
+    data = {"offset": start_index, "limit": results_per_page, "query": query, "sensitivity_id": "normal", "sort": "new"}
+    params["url"] = base_url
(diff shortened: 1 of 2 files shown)


searx/engines/bitchute.py
```diff
@@ -0,0 +1,58 @@
...
+
+base_url = "https://api.bitchute.com/api/beta/search/videos"
+categories = ['videos']
+paging = True
+
+
+def request(query, params):
+
+    start_index = (params["pageno"] - 1) * 50
+    data = {"offset": start_index, "limit": 50, "query": query, "sensitivity_id": "normal", "sort": "new"}
```
Bnyro: Please move 50 to a global `results_per_page` variable.

Also, 50 seems quite high, we usually set it to `10` or `20` per engine.

searx/engines/bitchute.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def response(resp):
+    search_res = resp.json()
+    results = []
+
+    for item in search_res.get('videos', []):
+
+        date = datetime.strptime(item["date_published"].split("T")[0], "%Y-%m-%d")
+        formatted_date = datetime.fromtimestamp(date.timestamp())
```
Bnyro: That's absolutely unnecessary, `formatted_date` is always the same as `date`.

That seems like something that only an AI could come up with... If you used AI to work with the code, please mention that in the pull request description.

searx/engines/bitchute.py
```diff
@@ -0,0 +1,58 @@
...
+    return params
+
+
+def response(resp):
+    search_res = resp.json()
+    results = []
+
+    for item in search_res.get('videos', []):
+
+        date = datetime.strptime(item["date_published"].split("T")[0], "%Y-%m-%d")
```
Bnyro: The date may also contain a time.
```suggestion
        date = datetime.strptime(item["date_published"], "%Y-%m-%dT%H:%M:%S.%fZ")
```
Example date returned by the bitchute API I used to build this format: `2025-03-19T11:53:13.648557Z`.

return42 (approved):
Thanks :+1:

Issue #4472, opened 2025-03-08
[poll] do we need plugin "Tracker URL remover" any longer?


return42:
The plugin has not been maintained for a long time, the number of signatures is low and probably no longer has any significance today.

In general, the engines should already remove tracker URLs, a plugin would make sense to remove commonly known signatures (URL arguments), but then the list of rules for this should be updated regularly or obtained from a source that is maintained.

Opinions / suggestions?  

Related:

- ClearURLs rules: https://github.com/ClearURLs/Rules/blob/master/data.min.json
- Adguard rules: https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/filter_17_TrackParam/filter.txt

unixfox:
would have been better in the poll section of discussions though: https://github.com/searxng/searxng/discussions/categories/polls

unixfox:
and discussions links are automatically sent on matrix so you get way more views. I would suggest to also post the link on reddit for more discovery

Bnyro:
> Opinions / suggestions?
>
I think it makes sense to source the rules from a third party project like `ClearURLs`, it's however not that easy to implement their format because their format uses regular expressions for tracker query params and has rules for exceptions, redirects, ...
So we'd need somebody with enough interest to rewrite the plugin completely.

The other possibility would be to make the plugin just remove all query parameters - nowadays 99% of search results don't contain any necessary query parameters for the site to work, most routing on websites is done via paths, e.g. `/products/garden/...`, instead of `products?category=garden&...`

PR #4538, opened 2025-03-24
[refactor] migrate plugins from "module" to class SXNGPlugin

### LINKED ISSUE #1348: Links in the infobox are not modified by hostname_replace

return42:
The ``enabled_plugins:`` section in SearXNG's settings do no longer exists. There is no longer a distinction between built-in and external plugin, all plugins are registered via the settings in the ``plugins:`` section.

To understand plugin development compile documentation:

    $ make docs.clean docs.live

and read http://0.0.0.0:8000/dev/plugins/development.html

In SearXNG, plugins can be registered via a fully qualified class name.  A configuration (`PluginCfg`) can be transferred to the plugin, e.g. to activate it by default / *opt-in* or *opt-out* from user's point of view.

built-in plugins
================

The built-in plugins are all located in the namespace `searx.plugins`.

```YAML

    plugins:

      searx.plugins.calculator.SXNGPlugin:
        active: true

      searx.plugins.hash_plugin.SXNGPlugin:
        active: true

      searx.plugins.self_info.SXNGPlugin:
        active: true

      searx.plugins.tracker_url_remover.SXNGPlugin:
        active: true

      searx.plugins.unit_converter.SXNGPlugin:
        active: true

      searx.plugins.ahmia_filter.SXNGPlugin:
        active: true

        active: true

      searx.plugins.oa_doi_rewrite.SXNGPlugin:
        active: false

      searx.plugins.tor_check.SXNGPlugin:
        active: false
```

external plugins
================

SearXNG supports *external plugins* / there is no need to install one, SearXNG runs out of the box.

- Only show green hosted results: https://github.com/return42/tgwf-searx-plugins/

To get a developer installation in a SearXNG developer environment:

```sh

   $ git clone git@github.com:return42/tgwf-searx-plugins.git
   $ ./manage pyenv.cmd python -m \
         pip install -e tgwf-searx-plugins
```

To register the plugin in SearXNG add `only_show_green_results.SXNGPlugin` to the `plugins:` ...

```yaml

    plugins:
      # ...
      only_show_green_results.SXNGPlugin:
        active: false
```

Result.filter_urls(..)
======================

The ``Result.filter_urls(..)`` can be used to filter and/or modify URL fields. In the following example, the filter function ``my_url_filter``:

```python

   def my_url_filter(result, field_name, url_src) -> bool | str:
       if "google" in url_src:
           return False              # remove URL field from result
       if "facebook" in url_src:
           new_url = url_src.replace("facebook", "fb-dummy")
           return new_url            # return modified URL
       return True                   # leave URL in field unchanged
```

is applied to all URL fields in the `Plugin.on_result` hook:

```python

   class MyUrlFilter(Plugin):
       ...
       def on_result(self, request, search, result) -> bool:
           result.filter_urls(my_url_filter)
           return True
```

Related
========

- closes: https://github.com/searxng/searxng/issues/1348
- closes: https://github.com/searxng/searxng/pull/4517

### DIFF
--- a/docs/admin/settings/settings_plugins.rst
+++ b/docs/admin/settings/settings_plugins.rst
@@ -1,67 +1,77 @@
 .. _settings plugins:
 
-=======
-Plugins
-=======
+============
+``plugins:``
+============
+
+.. attention::
+
+   The ``enabled_plugins:`` section in SearXNG's settings no longer exists.
+   There is no longer a distinction between built-in and external plugin, all
+   plugins are registered via the settings in the ``plugins:`` section.
 
 .. sidebar:: Further reading ..
 
    - :ref:`plugins admin`
    - :ref:`dev plugin`
-   - :ref:`builtin plugins`
 
+In SearXNG, plugins can be registered in the :py:obj:`PluginStore
+<searx.plugins.PluginStorage>` via a fully qualified class name.
 
-The built-in plugins can be activated or deactivated via the settings
-(:ref:`settings enabled_plugins`) and external plugins can be integrated into
-SearXNG (:ref:`settings external_plugins`).
+A configuration (:py:obj:`PluginCfg <searx.plugins.PluginCfg>`) can be
+transferred to the plugin, e.g. to activate it by default / *opt-in* or
+*opt-out* from user's point of view.
 
+Please note that some plugins, such as the :ref:`hostnames plugin` plugin,
+require further configuration before they can be made available for selection.
 
-.. _settings enabled_plugins:
+built-in plugins
+================
 
-``enabled_plugins:`` (internal)
-===============================
-
-In :ref:`plugins admin` you find a complete list of all plugins, the default
-configuration looks like:
+The built-in plugins are all located in the namespace `searx.plugins`.
 
 .. code:: yaml
 
-   enabled_plugins:
-     - 'Basic Calculator'
-     - 'Hash plugin'
-     - 'Self Information'
-     - 'Tracker URL remover'
-     - 'Unit converter plugin'
-     - 'Ahmia blacklist'
+    plugins:
+
+      searx.plugins.calculator.SXNGPlugin:
+        active: true
+
+      searx.plugins.hash_plugin.SXNGPlugin:
+        active: true
+
+      searx.plugins.self_info.SXNGPlugin:
+        active: true
+
+      searx.plugins.tracker_url_remover.SXNGPlugin:
+        active: true
+
+      searx.plugins.unit_converter.SXNGPlugin:
+        active: true
+
+      searx.plugins.ahmia_filter.SXNGPlugin:
+        active: true
+
+      searx.plugins.hostnames.SXNGPlugin:
+        active: true
+
+      searx.plugins.oa_doi_rewrite.SXNGPlugin:
+        active: false
+
+      searx.plugins.tor_check.SXNGPlugin:
+        active: false
 
 
 .. _settings external_plugins:
 
-``plugins:`` (external)
-=======================
+external plugins
+================
+
+.. _Only show green hosted results:
+   https://github.com/return42/tgwf-searx-plugins/
 
 SearXNG supports *external plugins* / there is no need to install one, SearXNG
-runs out of the box.  But to demonstrate; in the example below we install the
-SearXNG plugins from *The Green Web Foundation* `[ref]
-<https://www.thegreenwebfoundation.org/news/searching-the-green-web-with-searx/>`__:
+runs out of the box.
 
-.. code:: bash
-
-   $ sudo utils/searxng.sh instance cmd bash -c
-   (searxng-pyenv)$ pip install git+https://github.com/return42/tgwf-searx-plugins
-
-In the :ref:`settings.yml` activate the ``plugins:`` section and add module
-``only_show_green_results`` from ``tgwf-searx-plugins``.
-
-.. code:: yaml
-
-   plugins:
-     - only_show_green_results
-     # - mypackage.mymodule.MyPlugin
-     # - mypackage.mymodule.MyOtherPlugin
-
-.. hint::
-
-   ``only_show_green_results`` is an old plugin that was still implemented in
-   the old style.  There is a legacy treatment for backward compatibility, but
-   new plugins should be implemented as a :py:obj:`searx.plugins.Plugin` class.
+- `Only show green hosted results`_
+- ..
--- a/searx/plugins/_core.py
+++ b/searx/plugins/_core.py
@@ -3,31 +3,24 @@
 
 from __future__ import annotations
 
-__all__ = ["PluginInfo", "Plugin", "PluginStorage"]
+__all__ = ["PluginInfo", "Plugin", "PluginCfg", "PluginStorage"]
 
 import abc
 import importlib
+import inspect
 import logging
-import pathlib
-import types
+import re
 import typing
-import warnings
 
 from dataclasses import dataclass, field
 
-import flask
-
-import searx
-from searx.utils import load_module
 from searx.extended_types import SXNG_Request
 from searx.result_types import Result
 
-
 if typing.TYPE_CHECKING:
     from searx.search import SearchWithPlugins
+    import flask
 
-
-_default = pathlib.Path(__file__).parent
 log: logging.Logger = logging.getLogger("searx.plugins")
 
 
@@ -69,14 +62,17 @@ class PluginInfo:
     """See :py:obj:`Plugin.keywords`"""
 
 
+ID_REGXP = re.compile("[a-z][a-z0-9].*")
+
+
 class Plugin(abc.ABC):
     """Abstract base class of all Plugins."""
 
     id: str = ""
     """The ID (suffix) in the HTML form."""
 
-    default_on: bool = False
-    """Plugin is enabled/disabled by default."""
+    active: typing.ClassVar[bool]
+    """Plugin is enabled/disabled by default (:py:obj:`PluginCfg.active`)."""
 
     keywords: list[str] = []
     """Keywords in the search query that activate the plugin.  The *keyword* is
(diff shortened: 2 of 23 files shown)


tests/unit/test_plugins.py
```diff
@@ -47,9 +47,9 @@ def do_post_search(query, storage, **kwargs) -> Mock:
 
 class PluginMock(searx.plugins.Plugin):
 
-    def __init__(self, _id: str, name: str, default_on: bool):
+    def __init__(self, _id: str, name: str, active: bool):
         self.id = _id
-        self.default_on = default_on
+        self.activen = active
```
Bnyro:
```suggestion
        self.active = active
```

return42: Sorry, this PR seems not ready to review .. some more work needs to be done .. I mark this PR as DRAFT.

Bnyro: No worries :)

return42:
@Bnyro finished my development phase .. 

To understand plugin development compile documentation:

    $ make docs.clean docs.live

and read http://0.0.0.0:8000/dev/plugins/development.html

Function `searx.plugins.hostnames.filter_url_field(..)` is used to filter URL fields by the hostnames plugin.

docs/admin/settings/settings_plugins.rst
```diff
@@ -1,67 +1,77 @@
...
-=======
-Plugins
-=======
+============
+``plugins:``
+============
+
+.. attention::
+
+   The ``enabled_plugins:`` section in SearXNG's settings do no longer exists.
```
Bnyro:
```suggestion
   The ``enabled_plugins:`` section in SearXNG's settings no longer exists.
```
nit

searx/plugins/_core.py
```diff
@@ -121,7 +126,7 @@ def __eq__(self, other):
 
         return hash(self) == hash(other)
 
-    def init(self, app: flask.Flask) -> bool:  # pylint: disable=unused-argument
+    def init(self, app: "flask.Flask") -> bool:  # pylint: disable=unused-argument
```
Bnyro, line 129: That's probably not an intended change?

searx/plugins/ahmia_filter.py
```diff
@@ -1,34 +1,51 @@
...
+    from searx.extended_types import SXNG_Request
+    from searx.result_types import Result
+    from searx.plugins import PluginCfg
 
 ahmia_blacklist: list = []
 
 
-def on_result(_request, _search, result) -> bool:
-    if not getattr(result, 'is_onion', None) or not getattr(result, 'parsed_url', None):
+class SXNGPlugin(Plugin):
```
Bnyro, line 23: I'm not sure how convenient it is that plugins always need to be called `SXNGPlugin` and it's not possible to create two similar plugins in the same file or use inheritance.

We could possibly change this by doing some introspection, just wanted to leave that as a side note for the future.

searx/plugins/ahmia_filter.py
```diff
@@ -1,34 +1,51 @@
...
+            id=self.id,
+            name=gettext("Ahmia blacklist"),
+            description=gettext("Filter out onion results that appear in Ahmia's blacklist."),
+            preference_section="general",
+        )
+
+    def on_result(
+        self, request: "SXNG_Request", search: "SearchWithPlugins", result: Result
+    ) -> bool:  # pylint: disable=unused-argument
+        if not getattr(result, "is_onion", None) or not getattr(result, "parsed_url", None):
```
Bnyro:
```suggestion
        if not getattr(result, "is_onion", False) or not getattr(result, "parsed_url", None):
```
This might help to understand it better that the default is not onion.

searx/plugins/hostnames.py
```diff
@@ -92,91 +83,132 @@
...
-    if isinstance(setting_value, dict):
-        return {re.compile(p): r for (p, r) in setting_value.items()}
+        for pattern, replacement in REPLACE.items():
 
-    return None
+            if result.parsed_url and pattern.search(result.parsed_url.netloc):
+                result.parsed_url = result.parsed_url._replace(
+                    netloc=pattern.sub(replacement, result.parsed_url.netloc)
+                )
+                result.url = urlunparse(result.parsed_url)
```
Bnyro: That's already covered by `result.filter_urls`, isn't it?

searx/plugins/hostnames.py
```diff
@@ -92,91 +83,132 @@
...
 
-replacements: dict = _load_regular_expressions('replace') or {}  # type: ignore
-removables: set = _load_regular_expressions('remove') or set()  # type: ignore
-high_priority: set = _load_regular_expressions('high_priority') or set()  # type: ignore
-low_priority: set = _load_regular_expressions('low_priority') or set()  # type: ignore
+            for pattern in HIGH:
+                if result.parsed_url and pattern.search(result.parsed_url.netloc):
+                    result.priority = "high"
 
+        result.filter_urls(filter_url_field)
```
Bnyro: We should probably do the filtering / rewriting before setting low and high priority results, because the URL rewrite can result in URLs that have higher priority.

E.g. if I rewrite `stackoverflow.com` to `ao.chatoyer.de`, `ao.chatoyer.de` could be set to a higher priority (because I'm self-hosting it). That way we keep the previous behavior.

searx/plugins/hostnames.py
```diff
@@ -92,91 +83,132 @@
...
+
+    if not url_src:
+        log.debug("missing a URL in field %s", field_name)
+        return True
+
+    url_src_parsed = urlparse(url=url_src)
+
+    for pattern in REMOVE:
+        if pattern.search(url_src_parsed.netloc):
+            # log.debug("FIXME: remove [%s] %s", field_name, url_src_parsed.netloc)
```
Bnyro: are the FIXMEs still relevant?

searx/plugins/oa_doi_rewrite.py
```diff
@@ -1,54 +1,91 @@
...
-    doi_resolvers = settings['doi_resolvers']
-    selected_resolver = preferences.get_value('doi_resolver')[0]
-    if selected_resolver not in doi_resolvers:
-        selected_resolver = settings['default_doi_resolver']
-    return doi_resolvers[selected_resolver]
+    doi = extract_doi(result.parsed_url)
+    if doi and len(doi) < 50:
+        for suffix in ("/", ".pdf", ".xml", "/full", "/meta", "/abstract"):
+            if doi.endswith(suffix):
+                doi = doi[: -len(suffix)]
```
Bnyro:
```suggestion
            doi = doi.removesuffix(suffix)
```

searx/plugins/tracker_url_remover.py
```diff
@@ -2,43 +2,58 @@
...
+        if not result.parsed_url:
+            return True
+
+        parsed_query = parse_qsl(result.parsed_url.query)
+        changes = 0
+        for i, (param_name, _) in enumerate(list(parsed_query)):
+            for reg in regexes:
+                if reg.match(param_name):
+                    parsed_query.pop(i - changes)
+                    changes += 1
```
Bnyro:
```suggestion
                    parsed_query.remove(param_name)
```
That's really terrifying code, although that's unrelated to this PR this plugin should probably be reworked (or removed, as discussed in https://github.com/searxng/searxng/issues/4472.

Bnyro:
The `filter_urls` method seems like a good addition, I've been thinking about how to remove the duplicated logic in `hostnames.py` and that seems like a good solution.

searx/plugins/_core.py
return42: > That's probably not an intended change?

It is intended .. this is the declaration notation by a string, it can be used for forward definitions or for names not imported at runtime.  Note; `flask` is not imported at run-time: 

```python
if typing.TYPE_CHECKING:
    import flask
```

There are discussions about the `if typing.TYPE_CHECKING .. import ..` construct (some say its an anti pattern), I find it quite useful because you don't have to import names at runtime  that are only used for declarations which only active in the type checking prozess..

searx/plugins/ahmia_filter.py
return42: > I'm not sure how convenient it is that plugins always need to be called `SXNGPlugin`

A plugin class does not necessarily have to have the name `SXNGPlugin`.  It is also possible to implement several plugin classes in one python module.

A plugin (class) is registered by its full qualified name:

```yaml
    plugins:
      # ...
      my_py_pkg.my_module.MyFirstPlugin:
        active: false
      my_py_pkg.my_module.MySecondPlugin:
        active: false
```

searx/plugins/hostnames.py
return42: Yes, the `REPLACE` should be done by the `result.filter_urls(filter_url_field)` .. we can drop this `REPLACE` of `result.parsed_url` here.

searx/plugins/hostnames.py
return42: > are the FIXMEs still relevant?

nope .. I remove the out-commented lines :+1:

searx/plugins/tracker_url_remover.py
return42: I have adopted your suggestion ... however, the code needs to be revised ;-)

searx/plugins/hostnames.py
return42: > We should probably do the filtering / rewriting before setting low and high priority results,

Good catch .. I modified the code to first rewrite and after rewrite set priority, this and the other changes are shown in this [diff](https://github.com/searxng/searxng/compare/dfefcf5a944e02f8c7cf3b99d52b7a3ed50814c7..56c829f92761f10c2e17502dab0f99ece6f31679)

From my side PR is ready .. do we have any more open question / doubt?

return42:
> The `filter_urls` method seems like a good addition, I've been thinking about how to remove the duplicated logic in `hostnames.py` and that seems like a good solution.

I think it is the most flexible solution ... if there are additional URL fields at some point, additional URL fields only have to be implemented in the loop of the Result class (the caller does not have to worry about the inner logic of the Result class and what are URL-fields and what fields not).

return42:
I fixed an issue; the `Result.parsed_url` hasn't been synced with the `new_url` in `Result.url` / I fixed that, see [diff](https://github.com/searxng/searxng/compare/56c829f92761f10c2e17502dab0f99ece6f31679..c1ba85ac9db59fdfd657e04783608868949d2488) ..

Bnyro (approved):
LGTM!

return42:
> LGTM!

Thanks for review / FYI I patched my [instance](https://darmarit.org/searx/preferences) with this PR .. just to see if there are any issues in production .. none seen so far ..

PR #4546, opened 2025-03-26
[feat] bing: raise error upon receiving wrong results page

0xhtml:
## What does this PR do?

With this change bing raises a `SearxEngineAPIException` if the received results page is not the page requested.

Bing seems to return the first result page instead of the requested page as a rate limiting and bot detection mechanism. This behavior has not yet been fixed (see #3402 and #3416) but with this change the user no longer has to realize themselves that the results on the second page are the same and the pagination of bing can quickly and automatically be tested.

Just try loading page two results of bing.

## Related issues

#3402, #3416

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -38,6 +38,7 @@ import babel.languages
 from searx.utils import eval_xpath, extract_text, eval_xpath_list, eval_xpath_getindex
 from searx.locales import language_tag, region_tag
 from searx.enginelib.traits import EngineTraits
+from searx.exceptions import SearxEngineAPIException
 
 if TYPE_CHECKING:
     import logging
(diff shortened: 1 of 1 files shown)


return42:
> Bing seems to return the first result page instead of the requested page

FYI: in 2023 I tried to improve bing .. https://github.com/searxng/searxng/commit/772f1e86f8fe26e759e289dbc15fb2308e85588f .. not sure what the relevance is today, what works or not today .. in this commit in the request function, the arguments to get correct page had been modified (fixed?) --> [ref](https://github.com/searxng/searxng/commit/772f1e86f8fe26e759e289dbc15fb2308e85588f#diff-a6e94d38fef17f6ae1d7de46ca6a58c8da464f0bb5dafec0c785c281dca0557cR109-R113)

*Someone should test whether the correct page selection really (still) works.*

searx/engines/bing.py
Bnyro, line 1: It looks like if we remove line 97: `'pq': query,`, pagination actually works again at least for me. Could anyone please test that from a different location?

return42:
FYI: last force push  was just to rebase this branch on current master (was 89 commits behind).

searx/engines/bing.py
```diff
@@ -161,27 +162,31 @@ def response(resp):
...
-        # For example, if there is 100 results from some search and we try to get results from 120 to 130,
-        # Bing will send back the results from 0 to 10 and no error.
-        # If we compare results count with the first parameter of the request we can avoid this "invalid" results.
-        return []
+            start_str, result_len_container = re.split(r'-\d+', result_len_container)
+            start = int(start_str)
+        else:
+            start = 1
+
+        result_len = int(re.sub('[^0-9]', '', result_len_container))
```
return42: Got exception on search term `!bing ybco`:

```
  File "/share/searxng/searx/search/processors/online.py", line 160, in search
  File "/share/searxng/searx/search/processors/online.py", line 148, in _search_basic
  File "/share/searxng/searx/engines/bing.py", line 173, in response
    result_len = int(re.sub('[^0-9]', '', result_len_container))
ValueError: invalid literal for int() with base 10: ''
```

searx/engines/bing.py
0xhtml: Makes no difference for me in Germany.

Bnyro (approved):
Works as expected, thank you! (and sorry for the delay)

PR #4570, opened 2025-03-29
[feat] engine: add engine for italian online newspaper "il post"

gioleppe:
## What does this PR do?

This PR adds an engine for an Italian online newspaper, [Il Post](https://www.ilpost.it/).
I've added the engine and included configuration in settings.yml (disabled, of course).
The engine supports "month" and "year" range queries as well.

Il Post is a great resource for Italian people as it's usually one the less biased newspapers.
This would greatly benefit the Italian userbase

`make run`
Then enable the engine from settings 
Finally navigate to the news tab and search something

### DIFF
--- a/searx/engines/il_post.py
+++ b/searx/engines/il_post.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine for Il Post, a largely independent online Italian newspaper.
+
+To use this engine add the following entry to your engines
+list in ``settings.yml``:
+
+.. code:: yaml
+
+  - name: il post
+    engine: il_post
+    shortcut: pst
+    disabled: false
+
+"""
+
+from urllib.parse import urlencode
+from searx.result_types import EngineResults
+
+engine_type = "online"
+language_support = False
+categories = ["news"]
+paging = True
+page_size = 10
+
+time_range_support = True
+time_range_args = {"month": "pub_date:ultimi_30_giorni", "year": "pub_date:ultimo_anno"}
+
+search_api = "https://api.ilpost.org/search/api/site_search/?"
+
+about = {
+    "website": "https://www.ilpost.it",
+    "wikidata_id": "Q3792882",
+    "official_api_documentation": None,
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+    "language": "it",
+}
+
+
+def request(query, params):
+    query_params = {
+        "qs": query,
+        "pg": params["pageno"],
+        "sort": "date_d",
+        "filters": "ctype:articoli",
+    }
+
+    if params["time_range"]:
+        if params["time_range"] not in time_range_args:
+            return None
+        query_params["filters"] += f";{time_range_args.get(params['time_range'], 'pub_date:da_sempre')}"
+    params["url"] = search_api + urlencode(query_params)
+    return params
+
+
+def response(resp) -> EngineResults:
+    res = EngineResults()
+    json_data = resp.json()
+
+    for result in json_data["docs"]:
+        res.add(
+            res.types.MainResult(
+                url=result["link"],
+                title=result["title"],
+                content=result.get("summary", ""),
+                thumbnail=result.get("image"),
+            )
+        )
+
+    return res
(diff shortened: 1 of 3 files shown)


searx/engines/il_post.py
```diff
@@ -0,0 +1,76 @@
...
+    encoded_querystring = urlencode(query_params)
+
+
+    params['url'] = search_api + encoded_querystring
+    return params
+
+
+def response(resp) -> EngineResults:
+    res = EngineResults()
+    json_data = loads(resp.text)
```
Bnyro:
```suggestion
    json_data = resp.json()
```

searx/engines/il_post.py
```diff
@@ -0,0 +1,76 @@
...
+
+def request(query, params):
+    query_params = {
+        'qs': query,
+        'pg': params['pageno'],
+        'sort': "date_d",
+        'filters': 'ctype:articoli',
+    }
+
+    if params['time_range']:
```
Bnyro:
```suggestion
    if params['time_range']:
        if params['time_range'] not in time_range_args:
            return None
```
Since this engine doesn't support all time ranges, it shouldn't return any results for unsupported time ranges.

searx/engines/il_post.py
```diff
@@ -0,0 +1,76 @@
...
+
+search_api = 'https://api.ilpost.org/search/api/site_search/?'
+
+about = {
+    "website": 'https://www.ilpost.it',
+    "wikidata_id": 'Q3792882',
+    "official_api_documentation": None,
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
```
Bnyro:
```suggestion
    "results": 'JSON',
    "language": "it",
```

searx/engines/il_post.py
```diff
@@ -0,0 +1,76 @@
...
+    res = EngineResults()
+    json_data = loads(resp.text)
+
+    for result in json_data['docs']:
+
+        res.append(
+            {
+                'url': result['link'],
+                'title': result['title'],
+                'content': result['summary'] or None,
```
Bnyro:
```suggestion
                'content': result.get('summary', ''),
```

searx/engines/il_post.py
```diff
@@ -0,0 +1,76 @@
...
+                'thumbnail': result['image'] or None,
```
Bnyro: Is the `or None` for the case that it's an empty string? Otherwise, if `result['image']` is None, this expression is `None or None` and thus redundant.

searx/engines/il_post.py
gioleppe: thank you for pointing this out

searx/engines/il_post.py
gioleppe: I agree, better for clarity, thanks

searx/engines/il_post.py
gioleppe: I do remember a couple of cases where the content and/or the thumbnail was absent, but I couldn't find those results again to test, I don't know if the field was empty.
I thought about using an empty string for both fields as you proposed, but I couldn't decide and ended up doing like this. Is there a recommended/idiomatic way of doing this in the project's conventions? I also thought Python's "or" is short-circuited so the possible redundancy wouldn't be a problem.

Please let me know what you think about this and thank you for your comments and tips, my Python is a little bit rusty :)

Bnyro: For SearXNGs `content` field: This must be set in order for the result to be displayed. If it's none, the result will be hidden, thus the empty string here. (That's just because of some SearXNG internal stuff, I remember we had plans to change that behavior in the future)

Bnyro: For the thumbnail, we should probably just use `result.get('image')`, which defaults to None if the field is unset.

If using `result['image']` although the result doesn't contain an image field, this call would raise an exception and crash the engine.

gioleppe: done! :)

searx/engines/il_post.py
```diff
@@ -0,0 +1,75 @@
...
+    return params
+
+
+def response(resp) -> EngineResults:
+    res = EngineResults()
+    json_data = resp.json()
+
+    for result in json_data['docs']:
+
+        res.append(
```
return42: `EngineResults` has a `add` method which expects a `Result` type:

https://github.com/searxng/searxng/blob/b231cb4b5941c62310fc08d8307f28ce5d04a91c/searx/result_types/__init__.py#L46-L48

I have fixed this in my review.

return42 (approved):

PR #4575, opened 2025-03-29
[feat] engine: add engine for italian press agency ansa

gioleppe:
## What does this PR do?

This PR adds an engine for an Italian press agency, [ANSA](https://www.ansa.it/).
I've added the engine and included configuration in settings.yml (disabled by default)
The engine supports all range queries.

ANSA is a Italy's oldest press agency (still active). It is largely unbiased and provides news about both national and international topics. 
This addition would greatly benefit the Italian userbase

make run
Then enable the engine from settings
Finally navigate to the news tab and search something

### DIFF
--- a/searx/engines/ansa.py
+++ b/searx/engines/ansa.py
@@ -1,0 +1,81 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine for Ansa, Italy's oldest news agency.
+
+To use this engine add the following entry to your engines
+list in ``settings.yml``:
+
+.. code:: yaml
+
+  - name: ansa
+    engine: ansa
+    shortcut: ans
+    disabled: false
+
+"""
+
+from urllib.parse import urlencode
+from lxml import html
+from searx.result_types import EngineResults, MainResult
+from searx.utils import eval_xpath, eval_xpath_list, extract_text
+
+engine_type = 'online'
+language_support = False
+categories = ['news']
+paging = True
+page_size = 12
+base_url = 'https://www.ansa.it'
+
+time_range_support = True
+time_range_args = {
+    'day': 1,
+    'week': 7,
+    'month': 31,
+    'year': 365,
+}
+# https://www.ansa.it/ricerca/ansait/search.shtml?start=0&any=houthi&periodo=&sort=data%3Adesc
+search_api = 'https://www.ansa.it/ricerca/ansait/search.shtml?'
+
+about = {
+    'website': 'https://www.ansa.it',
+    'wikidata_id': 'Q392934',
+    'official_api_documentation': None,
+    'use_official_api': False,
+    'require_api_key': False,
+    'results': 'HTML',
+    'language': 'it',
+}
+
+
+def request(query, params):
+    query_params = {
+        'any': query,
+        'start': (params['pageno'] - 1) * page_size,
+        'sort': "data:desc",
+    }
+
+    if params['time_range']:
+        query_params['periodo'] = time_range_args.get(params['time_range'])
+
+    params['url'] = search_api + urlencode(query_params)
+    return params
+
+
+def response(resp) -> EngineResults:
+    res = EngineResults()
+    doc = html.fromstring(resp.text)
+
+    for result in eval_xpath_list(doc, "//div[@class='article']"):
+
+        res_obj = MainResult(
+            title=extract_text(eval_xpath(result, "./div[@class='content']/h2[@class='title']/a")),
+            content=extract_text(eval_xpath(result, "./div[@class='content']/div[@class='text']")),
+            url=base_url + extract_text(eval_xpath(result, "./div[@class='content']/h2[@class='title']/a/@href")),
+        )
+
+        thumbnail = extract_text(eval_xpath(result, "./div[@class='image']/a/img/@src"))
+        if thumbnail:
+            res_obj.thumbnail = base_url + thumbnail
+
+        res.append(res_obj)
+
+    return res
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -433,6 +433,11 @@ engines:
     disabled: true
     shortcut: aa
 
+  - name: ansa
+    engine: ansa
+    shortcut: ans
+    disabled: false
+
   # - name: annas articles
   #   engine: annas_archive
   #   shortcut: aaa


searx/engines/ansa.py
```diff
@@ -0,0 +1,83 @@
...
+        'start': (params['pageno'] - 1) * page_size,
+        'sort': "data:desc",
+    }
+
+    if params['time_range']:
+        query_params['periodo'] = time_range_args.get(params['time_range'])
+
+    encoded_querystring = urlencode(query_params)
+
+    params['url'] = search_api + encoded_querystring
```
Bnyro:
```suggestion
    params['url'] = search_api + urlencode(query_params)
```

searx/engines/ansa.py
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+
+from urllib.parse import urlencode
+from lxml import html
+from searx.result_types import EngineResults
+from searx.utils import eval_xpath, eval_xpath_list, extract_text
+
+engine_type = 'online'
+language_support = False
```
Bnyro, line 22: That doesn't need to be set explicitely, but it also won't hurt :)

searx/engines/ansa.py
```diff
@@ -0,0 +1,83 @@
...
+
+    for result in eval_xpath_list(doc, "//div[@class='article']"):
+
+        res_obj = {
+            'url': base_url + extract_text(eval_xpath(result, "./div[@class='content']/h2[@class='title']/a/@href")),
+            'title': extract_text(eval_xpath(result, "./div[@class='content']/h2[@class='title']/a")),
+            'content': extract_text(eval_xpath(result, "./div[@class='content']/div[@class='text']")),
+        }
+
+        thumbnail = extract_text(eval_xpath(result, "./div[@class='image']/a/img/@src"))
```
Bnyro: It's usually better to use xpaths like
```py
eval_xpath(result, "./div[contains(@class, 'content')]/div[contains(@class, 'text')]")
```
because that won't break if they add additional classes (e.g. `class="text paragraph"` would still work properly).

searx/engines/ansa.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine for Ansa, Italy's oldest news agency.
+
+To use this engine add the following entry to your engines
+list in ``settings.yml``:
+
+.. code:: yaml
+
+  - name: ansa
```
Bnyro, line 9: This is already part of `settings.yml`, so I don't think we need to explain that explicitely here.

The engine can be enabled like any other engine, so people should read the SearXNG documentation about it at https://docs.searxng.org/admin/settings/settings_engines.html instead.

searx/engines/ansa.py
gioleppe: it that the "default" I guess?
Either way I won't add it again in the future, or if you prefer I will remove it for consistency! :)

searx/engines/ansa.py
gioleppe: better, I agree. solved with my latest commit

searx/engines/ansa.py
gioleppe: Sorry @Bnyro I don't think I properly get what you mean here:
are you suggesting we remove this?

.. code:: yaml
  - name: ansa
    engine: ansa
    shortcut: ans
    disabled: false

Bnyro: Yeah, exactly.

searx/engines/ansa.py
gioleppe: solved, thanks for the tip

gioleppe:
@Bnyro I checked all your suggestion, thanks! 
I have a question: I've seen that news without a thumbnail are moved to the end of the results page by default, even if the publishedDate response field is correctly set (I tried with some unpublished branch). 
Is this documented somewhere? I guess some kind of stable sort is used somewhere, since I can see that the relative order of the documents is retained

Bnyro:
I don't think so, @return42 knows that part of SearXNG better than me.

gioleppe:
> @Bnyro I checked all your suggestion, thanks! I have a question: I've seen that news without a thumbnail are moved to the end of the results page by default, even if the publishedDate response field is correctly set (I tried with some unpublished branch). Is this documented somewhere? I guess some kind of stable sort is used somewhere, since I can see that the relative order of the documents is retained

@return42 do you have any insight on why this may happen? 
I would be happy to merge this branch nonetheless, if that's ok with you, since this doesn't seem to be related with my implementation, but rather a default behavior by SearXNG

return42:
> I've seen that news without a thumbnail are moved to the end of the results page by default,

the observation is not entirely correct ...

> @return42 do you have any insight on why this may happen?

Results are sorted by scores ..

https://github.com/searxng/searxng/blob/b146b745a7e6b89f6e70ac6f98384a816edd67b2/searx/results.py#L190-L207

and about "pass 2": we do not (yet) have a special class (template) for *news* .. this means a  news result item is a ordinary [MainResult](https://docs.searxng.org/dev/result_types/main_result.html#main-search-results) .. and therefore news result items are also placed (sorted) into the other ordinary MainResults, based on their score.

@Bnyro @gioleppe if you think PR is ready to merge .. then  please merge ;-)

gioleppe:
@return42 thank you for the code snippet! that's what I was looking for! 
Would this line:
https://github.com/searxng/searxng/blob/b146b745a7e6b89f6e70ac6f98384a816edd67b2/searx/results.py#L216
be the culprit?

Would there be a simple way to disable it (maybe on a engine-basis)? Would this make sense?
Maybe we could find a better way:  I'd argue that for press agencies / news the score should be computed on the "freshness" of the article (i.e., ordered descending by date), what do you think?

OR: we may as well design a News template, but I'd still argue that the thumbnail should not influence the final outcome :)

---

I'd say we merge this, and probably we should talk through the underlying "issue" separately, what do you think @Bnyro  @return42 ?

return42:
> So I guess presence (or absence) of the thumbnail affects score.

No, it didn't .. 

https://github.com/searxng/searxng/blob/b146b745a7e6b89f6e70ac6f98384a816edd67b2/searx/results.py#L17-L35

Thats what I mean by *"the observation is not entirely correct"* in my last comment.

> Maybe we could find a better way: I'd argue that for press agencies / news the score should be computed on the "freshness" of the article (i.e., ordered descending by date), what do you think?

Thats not trivial, please take in mind: in the main result list we combine [*news*, *torrent*, *images* and ..](https://docs.searxng.org/dev/result_types/main_result.html) .. and not every result item has a *publish date*.

That would require a complete overhaul of the results list and I wouldn't even know how to sort by date if some items in the list don't have a date... Please let's not discuss this in this PR.

gioleppe:
> > So I guess presence (or absence) of the thumbnail affects score.
> 
> No, it didn't ..
> 

Yeah, I edited the comment after reading through the code, sorry I was hasty lol

---

Then I guess we can merge for now, and discuss this separately! :)

gioleppe:
@Bnyro @return42 is it ok if we merge this?
I don't want to push you, I just want to know if there's anything else I need to do with this specific PR.

thank you for all your work! :)

return42 (approved):
FYI I squashed the commits into one and modified the dict items into typed results ([MainResult](https://docs.searxng.org/dev/result_types/main_result.html))  / see [diff](https://github.com/searxng/searxng/compare/3faaee970b2559ccc5b069f684d17fd2982acca8..0c0df8ae6e0d5e50c29562ff7150949f51dd0fa1)

gioleppe:
> FYI I squashed the commits into one and modified the dict items into typed results ([MainResult](https://docs.searxng.org/dev/result_types/main_result.html)) / see [diff](https://github.com/searxng/searxng/compare/3faaee970b2559ccc5b069f684d17fd2982acca8..0c0df8ae6e0d5e50c29562ff7150949f51dd0fa1)

It makes perfect sense to use typed results, thanks for the correction! :)

gioleppe:
@return42 I just spotted a small error: the ansa engine should probably be disabled by default, I wrongly enabled it in the settings.yml, maybe you could commit directly a simple fix by setting the disabled to true in the yaml!

Sorry, my fault

return42:
> Sorry, my fault

You're not alone, I also fall into the “double negative is positive” trap time and time again / thanks for reporting :+1: 

Fixed in master branch, see [777ba6d](https://github.com/searxng/searxng/commit/777ba6ddf822be87ee31609779fc1a8cb4cf13f1).

PR #4578, opened 2025-03-30
Don't hardcode uWSGI parameters

inetol:
## What does this PR do?

Moves all the config to `/etc/searxng/uwsgi.ini`

We have special requirements and we must modify params that the flags overwrite. It is also confusing to have the configuration scattered in different files.

`podman build .`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN addgroup -g ${SEARXNG_GID} searxng && \
 ENV INSTANCE_NAME=searxng \
     AUTOCOMPLETE= \
     BASE_URL= \
+    BIND_ADDRESS=[::]:8080 \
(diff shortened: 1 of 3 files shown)


unixfox (approved):
I guess why not but I don't know anything about uwsgi, I don't use uwsgi on my instance. @return42

inetol:
Now I realize, the instances will stop working if this change is made as it requires updating the uwsgi.ini file. We should leave http-socket flag set for the moment.

Bnyro:
You should probably add a comment to the entrypoint that we want to remove `bind-adress` from the arguments there inthe future.

And we could do an announcement to instance maintainers to update their uwsgi.ini if they use it.

Bnyro (approved):

PR #4583, opened 2025-04-01
Added commented-out JSON option to search formats block

xalgia:
## What does this PR do?

Added commented-out JSON option to the search formats block.

Users can now know how to enable json without needing to search 'how-to' online.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -77,6 +77,7 @@ search:
   # formats: [html, csv, json, rss]
   formats:
     - html
+    # - json
 
 server:
   # Is overwritten by ${SEARXNG_PORT} and ${SEARXNG_BIND_ADDRESS}


Bnyro (approved):

searx/settings.yml
```diff
@@ -77,6 +77,7 @@ search:
   # formats: [html, csv, json, rss]
   formats:
     - html
+    # - json
```
return42, line 80: > Users can now know how to enable json without needing to search 'how-to' online.

Pew .. json, csv and rss are already listed two lines above .. in line 77 .. and why are searching for a "how-to"? .. I will never understand why people ask for outdated and faulty HowTo on the internet when there is up-to-date documentation -> https://docs.searxng.org/admin/settings/settings_search.html ..

PR #4584, opened 2025-04-01
add Microsoft Learn engine

gioleppe:
What does this PR do?
This PR adds an engine for [Microsoft Learn](https://learn.microsoft.com).
I've added the engine and included configuration in settings.yml (disabled by default)

Why is this change important?
MS Learn is a very useful resource for all developers using a MS stack, as well as IT professionals in general
.

How to test this PR locally?
make run

then **either** enable the engine from settings then navigate to the IT tab and search something
**or** type !msl <query> in the search bar

--
BIG thanks to the community in Matrix for providing useful suggestions and especially to https://github.com/comral-86 for providing a first yml version using the JSON engine (I ditched it to make pagination work, since a -1 is needed to initial page number)

### DIFF
--- a/searx/engines/microsoft_learn.py
+++ b/searx/engines/microsoft_learn.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine for Microsoft Learn, Microsoft's technical knowledge base.
+
+To use this engine add the following entry to your engines list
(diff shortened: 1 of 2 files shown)


return42 (approved):
Thanks for adding the engine, good work :+1: 

--- 

FYI: I squashed the two commits into one and rebased on current master.

PR #4596, opened 2025-04-06
[PoC] Granian

loorisr:
## What does this PR do?

This PR replace uWSGI with granian.

It is a Proof of Concept, it works but Granian configuration has to be tuned correctly.
Also all uWSGI files need to be cleaned (uwsgi.ini as well as docker entrypoint)

uWSGI is in bug fix only. A more modern solution should be used.

`The project is in maintenance mode (only bugfixes and updates for new languages apis). Do not expect quick answers on github issues and/or pull requests (sorry for that) A big thanks to all of the users and contributors since 2009.`

## Related issues

#4479

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -175,4 +175,5 @@ unset MORTY_KEY
 
 # Start uwsgi
 printf 'Listen on %s\n' "${BIND_ADDRESS}"
-exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+#exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+exec granian --interface wsgi searx.webapp:app  --workers 3 --host 0.0.0.0 --port 8080
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,10 +41,9 @@ RUN apk add --no-cache -t build-dependencies \
     libxslt \
     openssl \
     tini \
-    uwsgi \
-    uwsgi-python3 \
     brotli \
  && pip3 install --break-system-packages --no-cache -r requirements.txt \
+ && pip3 install --break-system-packages --no-cache granian \
  && apk del build-dependencies \
  && rm -rf /root/.cache
 


Bnyro:
> uWSGI is not maintained any more.
>
Source? Their git repo seems to be very active.

Apparently it doesn't push new large features anymore, but why should it if it works perfectly fine with all its features now? ...

Bnyro (commented):
Please convert the PR to a draft for now.

loorisr:
> > uWSGI is not maintained any more.
> 
> Source? Their git repo seems to be very active.

Sorry I will modify by "bugfix only", according to their website: https://uwsgi-docs.readthedocs.io/

> The project is in maintenance mode (only bugfixes and updates for new languages apis). Do not expect quick answers on github issues and/or pull requests (sorry for that) A big thanks to all of the users and contributors since 2009.

dockerfiles/docker-entrypoint.sh
```diff
@@ -175,4 +175,5 @@ unset MORTY_KEY
 
 # Start uwsgi
 printf 'Listen on %s\n' "${BIND_ADDRESS}"
-exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+#exec uwsgi --master --uid searxng --gid searxng --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+exec granian --interface wsgi searx.webapp:app  --workers 3 --host 0.0.0.0 --port 8080
```
Bnyro, line 179: As Granian doesn't support config files, we would also need to port all the config here: https://github.com/searxng/searxng/blob/master/dockerfiles/uwsgi.ini to CLI arguments

Bnyro:
I personally think it's generally a good idea to move away from Uwsgi and Granian looks like a good choice, however it's crucial that it still remains possible to run SearXNG with other wsgis (if not used via docker) so that we don't break existing installations.

inetol:
Lack of config files is kind of a problem, as this PR seems to be focused on just Docker I'm going to pretend that everyone hasn't done weird hacks to run other WSGI servers with the default container image.

I've thought about allowing configuration through the [CMD](https://docs.docker.com/reference/dockerfile/#cmd) keyword in the Dockerfile itself, and then having the script passthrough all the parameters to granian. This would allow the default values to be easily modified if necessary.

loorisr:
On docker, we have `docker-entrypoint.sh` that calls granian so we could keep a config file if needed and the entrypoint call granian with the arguments from the config file?

inetol:
I would prefer not to create new abstractions to configure other programs. I don't mind configuring granian with flags (as long as they are not hardcoded), but if really necessary, we could request a new feature for config file support.

dalf:
The current docker image is built for [linux/arm/v7](https://github.com/searxng/searxng/blob/9424bac031db78ab23d1e5f21ec2906e3d603433/manage#L194) (Raspberry Pi 3 and some other SBC). Granian does [not have wheels for arm/v7](https://pypi.org/project/granian/#files). We need to
* either compile Granian for this architecture in the docker image (cumbersome and slow)
* or make a PR to Granian to support this architecture, if this is okay. I'm [not sure how](https://github.com/emmett-framework/granian/blob/master/.github/workflows/release.yml) (messense/manylinux_2_24-cross does not support armv7).
* or stop support for this architecture.

 [EDIT] no more relevant: there are now wheels for armv7

return42:
> Granian does [not have wheels for arm/v7](https://pypi.org/project/granian/#files). We need to

@gi0baro do you see a chance to support wheels for arm/v7 in Granian?

gi0baro:
> > Granian does [not have wheels for arm/v7](https://pypi.org/project/granian/#files). We need to
> 
> @gi0baro do you see a chance to support wheels for arm/v7 in Granian?

I see the [maturin action](https://github.com/PyO3/maturin-action) supports `armv7l`, I can try to add that to the build process. A couple of *very big unknowns* though:
- I have no idea if `armv7l` covers all your cases
- I have no idea how Granian behaves on 32bit architectures

In addition, I have no access to such kind of hardware to actually test the builds, so even if a build step can be added and it actually compiles, sounds quite far from "supporting armv7" tbh.
If any of you have access to such kind of hardware, we might try to coordinate some tests in this regard.

return42:
> Apparently it doesn't push new large features anymore, but why should it if it works perfectly fine with all its features now? ...

:thinking: 

.. the software is not in a state where everything works fine, one example can be found in [Pitfalls of the Tyrant mode](https://docs.searxng.org/admin/installation-uwsgi.html#pitfalls-of-the-tyrant-mode) .. and what distributors do with uWSGI is little mess / distributors have their own concepts .. see: [Distributors](https://docs.searxng.org/admin/installation-uwsgi.html#distributors)

If we get rid of "Distributor's concepts" and install (with Granain) directly from PyPi, I think the setup would be much easier.

All in all, with uWSGI we have a poorly maintained product that will not be improved any further and that has to be set up on the various distros in different ways. On the other hand, with Granian we have a product that is actively maintained and the maintainer (thanks goes to @gi0baro) who is already actively supporting us here in our thread :-)

gi0baro:
Out of curiosity, I successfully run a build workflow with `armv7` builds here https://github.com/emmett-framework/granian/actions/runs/14295016210

Again, I have no idea if those wheels actually work.

dalf:
There are now (experimental) wheels for armv7, thank you @gi0baro  : https://github.com/emmett-framework/granian/releases/tag/v2.2.3

inetol:
I have done a production deploy with this change on https://search.inetol.net instance to pick up performance telemetry and properly configure granian. All the errors that are coming out will be published on https://github.com/inetol-infrastructure/searxng-container/issues to avoid polluting this PR.

loorisr:
> I have done a production deploy with this change on https://search.inetol.net instance to pick up performance telemetry and properly configure granian. All the errors that are coming out will be published on https://github.com/inetol-infrastructure/searxng-container/issues to avoid polluting this PR.

are you using arm/v7 on the host?

inetol:
> are you using arm/v7 on the host?

No, only x86. I'm not building the images for arm/v7 there, as I don't have the hardware to test it.

return42:
> are you using arm/v7 on the host?


The test of @inetol should be limited to the SearXNG's migration from uWSGI to Granian ... a functional test of Granian on 32bit should be done by someone else (not really in the scope of the SearXNG project) .. a 32bit hardware is not required for this, it can be emulated with [Qemu](https://www.qemu.org/) on a 64bit architecture (Pi2 Model has a [Cortex-A7 CPU](https://en.wikipedia.org/wiki/ARM_Cortex-A7)).

inetol:
There don't seem to be any other problems with Granian apart from the [issue with static files](https://github.com/inetol-infrastructure/searxng-container/issues/7). Obviously, this PR needs to be updated with the Granian env settings from my repo.

As I'm using a somewhat tight SXNG settings there might be something I missed, so if anyone wants to quickly spin up a new instance for testing this I'm now publishing the container images to [the registry](https://quay.io/repository/inetol-infrastructure/searxng?tab=tags).

Bnyro:
superseded by https://github.com/searxng/searxng/pull/4820

PR #4597, opened 2025-04-06
[fix] engine radio browser: get servers from DNS api.radio-browser.info

### LINKED ISSUE #4576: [bug] engine radiobrowser: "https://de1.api.radio-browser.info"  is dead
The hard coded API URL is dead:

https://github.com/searxng/searxng/blob/50f92779bd9f2762d92a24a63851f4069f88297d/searx/engines/radio_browser.py#L30

Instead of a hard coded URL, the [API doc](https://api.radio-browser.info/) recommends:

>These steps should be done in your APP or program.
>
>    Get a list of available servers>
>
>  Do a DNS-lookup of 'all.api.radio-browser.info'. This gives you a list of all available servers. To get the nice names you can do a reverse DNS-Lookup of the entries from the first request and display them to the user. There are some examples of how to do this in different languages: [NodeJS (prefered)](https://api.radio-browser.info/examples/serverlist_fast.js), [NodeJS](https://api.radio-browser.info/examples/serverlist.js), [Python 2](https://api.radio-browser.info/examples/serverlist_python2.py), [Python 3](https://api.radio-browser.info/examples/serverlist_python3.py), [Java (Android)](https://api.radio-browser.info/examples/serverlist.java), [C#](https://api.radio-browser.info/examples/serverlist.cs), [Javascript(Browser)](https://api.radio-browser.info/examples/serverlist-browser.js)
    Now you have a list of multiple names of servers which you can use to directly connect to with HTTP and preferably HTTPS.
    Example: https://de2.api.radio-browser.info, https://fi1.api.radio-browser.info, ..

Here is an example code to do a dns + reverse lookup and select randomly a URL from available servers:

https://github.com/andreztz/pyradios/blob/main/pyradios/base_url.py

return42:
Do a DNS-lookup of `all.api.radio-browser.info`, add reverse lookup and select randomly a URL from available servers.

Closes: https://github.com/searxng/searxng/issues/4576

### DIFF
--- a/searx/data/engine_traits.json
+++ b/searx/data/engine_traits.json
@@ -7196,6 +7196,7 @@
       "km": "khmer",
       "kn": "kannada",
       "ko": "korean",
+      "ks": "kashmiri",
       "ku": "kurdish",
       "kw": "cornish",
       "la": "latin",
--- a/searx/engines/radio_browser.py
+++ b/searx/engines/radio_browser.py
@@ -6,6 +6,8 @@
 
 """
 
+import random
+import socket
 from urllib.parse import urlencode
 import babel
 from flask_babel import gettext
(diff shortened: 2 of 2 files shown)


Bnyro:
@return42 Thank you for looking into this, I didn't have the time for it recently unfortunately.

PR #4603, opened 2025-04-09
[fix] issues when launching a local development server

return42:
A local development server can be launched by one of these command lines::

    $ flask --app searx.webapp run
    $ python -m searx.webapp

The different ways of starting the server should lead to the same result, which is generally the case.  However, if the modules are reloaded after code changes (reload option), it must be avoided that the application is initialized twice at startup.  We have already discussed this in 2022 [1][2].

Further information on this topic can be found in [3][4][5].

To test, a bash in the ./local environment was started and the follwing commands had been executed::

    $ ./manage pyenv.cmd bash --norc --noprofile
    (py3) SEARXNG_DEBUG=1 flask --app searx.webapp run --reload
    (py3) SEARXNG_DEBUG=1 python -m searx.webapp

Since the generic parts of the docs also initialize the app to generate doc from it, the build of the docs was also tested::

    $ make docs.clean docs.live

- [1] https://github.com/searxng/searxng/pull/1656#issuecomment-1214198941
- [2] https://github.com/searxng/searxng/pull/1616#issuecomment-1206137468
- [3] https://flask.palletsprojects.com/en/stable/api/#flask.Flask.run
- [4] https://github.com/pallets/flask/issues/5307#issuecomment-1774646119
- [5] https://stackoverflow.com/a/25504196

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -6,6 +6,7 @@
 # pylint: disable=use-dict-literal
 from __future__ import annotations
 
+import inspect
 import hashlib
 import hmac
 import json
@@ -29,6 +30,8 @@ from pygments import highlight
 from pygments.lexers import get_lexer_by_name
 from pygments.formatters import HtmlFormatter  # pylint: disable=no-name-in-module
 
+from werkzeug.serving import is_running_from_reloader
+
 import flask
 
 from flask import (
@@ -48,12 +51,12 @@ from flask_babel import (
     format_decimal,
 )
 
+import searx
 from searx.extended_types import sxng_request
 from searx import (
     logger,
     get_setting,
     settings,
-    searx_debug,
 )
 
 from searx import infopage
@@ -81,7 +84,6 @@ from searx.webutils import (
     exception_classname_to_text,
     new_hmac,
     is_hmac_of,
-    is_flask_run_cmdline,
     group_engines_in_tab,
 )
 from searx.webadapter import (
@@ -127,11 +129,6 @@ from searx.search.checker import get_result as checker_get_result
 logger = logger.getChild('webapp')
 
 warnings.simplefilter("always")
-
-# check secret_key
-if not searx_debug and settings['server']['secret_key'] == 'ultrasecretkey':
-    logger.error('server.secret_key is not changed. Please use something else instead of ultrasecretkey.')
-    sys.exit(1)
 
 # about static
 logger.debug('static directory is %s', settings['ui']['static_path'])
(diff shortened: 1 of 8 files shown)


searx/webapp.py
```diff
@@ -1329,45 +1326,108 @@ def page_not_found(_e):
...
+        sys.exit(1)
+
+    # When automatic reloading is activated stop Flask from initialising twice.
+    # - https://github.com/pallets/flask/issues/5307#issuecomment-1774646119
+    # - https://stackoverflow.com/a/25504196
+
+    reloader_active = is_werkzeug_reload_active()
+    werkzeug_run_main = is_running_from_reloader()
+
+    if reloader_active and not werkzeug_run_main:
```
Bnyro, line 1413: 1. Do we really need `reloader_active` here? I think it's enough to check `is_running_from_reloader()`, because `is_running_from_reloader()` can only be true if we use Werkzeug's reload mode?
2. Shouldn't this be `... and is_running_from_reloader()`? If I understand this correctly, `is_running_from_reloader()` is `True` if the process has already been started once before, looking at https://github.com/pallets/werkzeug/blob/cbd049d88727936173386d2e80bb5ffa51fedd6e/werkzeug/_reloader.py#L129 and https://github.com/pallets/werkzeug/blob/cbd049d88727936173386d2e80bb5ffa51fedd6e/werkzeug/_reloader.py#L129

Pardon my confusion if this doesn't make much sense :)

tests/__init__.py
```diff
@@ -5,13 +5,8 @@
 import os
 import aiounittest
 
-# Before import from the searx package, we need to set up the (debug)
-# environment.  The import of the searx package initialize the searx.settings
-# and this in turn takes the defaults from the environment!
 
 os.environ.pop('SEARXNG_SETTINGS_PATH', None)
-os.environ['SEARXNG_DEBUG'] = '1'
```
Bnyro, line 13: Why is this no longer needed? Is that because it's already set in `settings.yml`?

searx/webapp.py
```diff
@@ -1329,45 +1326,108 @@ def page_not_found(_e):
...
+
+    if searx.sxng_debug:
+        logger.debug("run local development server (DEBUG) on %s:%s", host, port)
+        app.run(
+            debug=True,
+            port=port,
+            host=host,
+            threaded=True,
+            extra_files=[DEFAULT_SETTINGS_FILE],
+        )
```
Bnyro, line 1360: We previously also set `use_debugger` here. I'm not 100% what that changes though :shrug:

return42: `debug` implies `use_debugger` :-)

tests/__init__.py
return42: > Why is this no longer needed? I

Its an leftover (see chapter "Unit Tests" in https://github.com/searxng/searxng/pull/4183)

>  Is that because it's already set in `settings.yml`?

Nope, the settings for the unit test are loaded from [tests/unit/settings](https://github.com/searxng/searxng/blob/master/tests/unit/__init__.py)

searx/webapp.py
return42: > Pardon my confusion if this doesn't make much sense :)

Its complex .. I was confused .. and still am :)

Its a know "bug/limitation" of the Flask reloader; the Flask app is initialized twice when using the reload option.  The Process is

A. Start with reload option:

A.1. The Watchdock process is started: in the watchdog process, the Flask app is initialized (unnecessarily) first time and WERKZEUG_RUN_MAIN is unset

A.2. The watchdog start the Flask-App in a Subprocess: Flask app is initialized second time and WERKZEUG_RUN_MAIN is set to `true`)

B. Start without reload option

B.1. Flask app is initialized once and WERKZEUG_RUN_MAIN is unset

We need to init the app only in A.2 (WERKZEUG_RUN_MAIN=true) and B.1 (WERKZEUG_RUN_MAIN unset).

The fact that we can't clearly distinguish A.1 from B.1 also means that we have to look at the command line ... ~I don't know why "Werkzeug" doesn't set `WERKZEUG_RUN_MAIN=true` in B.1 as well~ .. it took me days to understand this (and I am always confused when I look at it)

---  

update:

> I don't know why "Werkzeug" doesn't set `WERKZEUG_RUN_MAIN=true` in B.1 as well

Because any other WSGI server (case B.1) won't set such an environment.

Bnyro: Thanks for the explaination, that makes sense now 👍 

Still, I'm not very happy with what the `werkzeug` people did here, but we can't change it ...

Bnyro (approved):

tests/__init__.py
Bnyro: > Nope, the settings for the unit test are loaded from [tests/unit/settings](https://github.com/searxng/searxng/blob/master/tests/unit/__init__.py)
>
Oh yes, that's what I meant, pardon :)

PR #4606, opened 2025-04-10
Remove brackets before passing to ip_address

inetol:
## What does this PR do?

I have added a small condition to remove the brackets from the IPv6 coming from the upstream, [#ip_address](https://docs.python.org/3.13/library/ipaddress.html#ipaddress.ip_address) does not read the brackets (I don't know why). I also removed the port section too, leaving a clean address

Progress on replacing uWSGI

This error appears when using Granian as WSGI server (or any server that sends a spec compliant IPv6 structure downstream), needs enabling the `public_instance` param on settings.yml and accessing `/healthz` over IPv6.

## Related issues

Closes https://github.com/inetol-infrastructure/searxng-container/issues/6

### DIFF
--- a/searx/botdetection/_helpers.py
+++ b/searx/botdetection/_helpers.py
@@ -126,7 +126,15 @@ def get_real_ip(request: SXNG_Request) -> str:
     if real_ip and remote_addr and real_ip != remote_addr:
         logger.warning("IP from WSGI environment (%s) is not equal to IP from X-Real-IP (%s)", remote_addr, real_ip)
 
-    request_ip = ip_address(forwarded_for or real_ip or remote_addr or '0.0.0.0')
+    raw_ip = forwarded_for or real_ip or remote_addr or '0.0.0.0'
+
+    if raw_ip.startswith('['):
+        raw_ip = raw_ip.split(']')[0][1:]
+    elif ':' in raw_ip and raw_ip.count(':') > 1:
+        if raw_ip.rsplit(':', 1)[1].isdigit():
+            raw_ip = raw_ip.rsplit(':', 1)[0]
+
+    request_ip = ip_address(raw_ip)
     if request_ip.version == 6 and request_ip.ipv4_mapped:
         request_ip = request_ip.ipv4_mapped
 


return42:
The `[::ffff:10.89.0.25]:56446` is the `IP:port` notation of an URL, but the value of a the HTTP header [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) (for example) should be a list of IPs (or a single IP).  And the `get_real_ip()` function in SearXNG looks for IPs [in the HTTP headers](https://docs.searxng.org/src/searx.botdetection.html#searx.botdetection.get_real_ip) (not a `IP:port` notation).

> Closes https://github.com/inetol-infrastructure/searxng-container/issues/6

Do you know which component in your setup sets a HTTP header to `[::ffff:10.89.0.25]:56446`? 

The header is generally set by the proxy (like shown [here](https://docs.searxng.org/admin/installation-nginx.html#nginx-s-searxng-site) for NGINX).

inetol:
> The `[::ffff:10.89.0.25]:56446` is the `IP:port` notation of an URL, but the value of a the HTTP header [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) (for example) should be a list of IPs (or a single IP). And the `get_real_ip()` function in SearXNG looks for IPs [in the HTTP headers](https://docs.searxng.org/src/searx.botdetection.html#searx.botdetection.get_real_ip) (not a `IP:port` notation).

Damn, I didn't know that.

> Do you know which component in your setup sets a HTTP header to `[::ffff:10.89.0.25]:56446`?
> 
> The header is generally set by the proxy (like shown [here](https://docs.searxng.org/admin/installation-nginx.html#nginx-s-searxng-site) for NGINX).

This is almost certainly coming from `request.remote_addr`, our only web server that sits between SearXNG and WAN does not set the `X-Forwarded-For` and `X-Real-IP` headers.

```console
2025-04-10 15:52:08,162 WARNING:searx.botdetection: X-Forwarded-For: None || X-Real-IP: None || request.remote_addr: [::ffff:10.89.0.26]:37898
```

I know that this is a web server issue, but it has nothing to do with this PR

return42:
> This is almost certainly coming from `request.remote_addr`, our only web server that sits between SearXNG and WAN does not set the `X-Forwarded-For` and `X-Real-IP` headers.

FYI SearXNG uses [ProxyFix](https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/) --> "X-Forwarded-For sets REMOTE_ADDR".

> I know that this is a web server issue, but it has nothing to do with this PR

Not sure I fully understand this last sentence .. this PR tries to make an IP from a IPv6-URL notation, which is mistakenly set by a WEB server in some headers where an IP is expected .. so I think this patch (PR) is not needed, since the bug is in the WEB server .. or I'm wrong?

inetol:
> Not sure I fully understand this last sentence .. this PR tries to make an IP from a IPv6-URL notation, which is mistakenly set by a WEB server in some headers where an IP is expected .. so I think this patch (PR) is not needed, since the bug is in the WEB server .. or I'm wrong?

Sorry, I meant that the web server by not passing the `X-Forwarded-For` and `X-Real-IP` headers makes SearXNG only know the internal IP and not the real IP of the client.

Let me start again... The problem I want to fix in this PR is between the Granian server and the application, I have checked further and enabled a debug log to see what differences there are between uWSGI and Granian for this error to occur.

Inside the container (applicable also locally) with uWSGI I have installed curl and run `curl -v http://[::1]:8080/healthz`..:

```console
2025-04-10 17:08:07,953 WARNING:searx.botdetection: X-Forwarded-For: None || X-Real-IP: None || request.remote_addr: ::1
```

OK, now the same with Granian..:

```console
2025-04-10 17:17:04,722 WARNING:searx.botdetection: X-Forwarded-For: None || X-Real-IP: None || request.remote_addr: [::1]:58236
```

...

I've also tested with IPv4. With uWSGI and running `curl -v http://127.0.0.1:8080/healthz`..:

```console
2025-04-10 17:29:45,458 WARNING:searx.botdetection: X-Forwarded-For: None || X-Real-IP: None || request.remote_addr: 127.0.0.1
```

The same with Granian..:

```console
2025-04-10 17:27:18,426 WARNING:searx.botdetection: X-Forwarded-For: None || X-Real-IP: None || request.remote_addr: [::ffff:127.0.0.1]:50994
```

You can see that `request.remote_addr` has changed "a little bit" between uWSGI and Granian. Honestly, I don't know anything about web development with Python, so I don't know how it handles requests between Python and the WSGI server.

return42:
@inetol thanks for your detail report .. I implemented a minimal test suite and opened a bug report in the granian project:

- https://github.com/emmett-framework/granian/issues/560

return42:
Fixed in Granian, see:

- https://github.com/emmett-framework/granian/pull/562

PR #4628, opened 2025-04-13
[feat] engine: add Steam engine

hezhijie0327:
## What does this PR do?

add Steam engine

ref: https://store.steampowered.com/
Wikidata: https://www.wikidata.org/wiki/Q337535

|`steam_app_details`|Screenshot|
|-|-|
|true|image|
|false|image|

1. search games via `!stm foo`

Note:
set `steam_app_details` to `true`, will fetch the app detail from separate API endpoint.

### DIFF
--- a/searx/engines/steam.py
+++ b/searx/engines/steam.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Steam (store) for SearXNG."""
+
+from urllib.parse import urlencode
+
+from searx.utils import html_to_text
+from searx.result_types import EngineResults, MainResult
+
+about = {
+    "website": 'https://store.steampowered.com/',
+    "wikidata_id": 'Q337535',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = []
+
+base_url = "https://store.steampowered.com"
+
+
+def request(query, params):
+    query_params = {"term": query, "cc": "us", "l": "en"}
+    params['url'] = f'{base_url}/api/storesearch/?{urlencode(query_params)}'
+    return params
+
+
+def response(resp) -> EngineResults:
+    results = EngineResults()
+    search_results = resp.json()
+
+    for item in search_results.get('items', []):
+        app_id = item.get('id')
+
+        currency = item.get('price', {}).get('currency', 'USD')
+        price = item.get('price', {}).get('final', 0) / 100
+
+        platforms = ', '.join([platform for platform, supported in item.get('platforms', {}).items() if supported])
+
+        content = [f'Price: {price:.2f} {currency}', f'Platforms: {platforms}']
+
+        results.add(
+            MainResult(
+                title=item.get('name'),
+                content=html_to_text(' | '.join(content)),
+                url=f'{base_url}/app/{app_id}',
+                thumbnail=item.get('tiny_image', ''),
+            )
+        )
+
+    return results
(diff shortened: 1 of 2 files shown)


searx/engines/steam.py
```diff
@@ -0,0 +1,62 @@
...
+
+def request(query, params):
+    query_params = {"term": query, "cc": "us", "l": "en"}
+    params['url'] = f'{base_url}/api/storesearch/?{urlencode(query_params)}'
+    return params
+
+
+def response(resp):
+    results = []
+    search_results = json.loads(resp.text)
```
Bnyro:
```suggestion
    search_results = resp.json()
```

searx/engines/steam.py
```diff
@@ -0,0 +1,62 @@
...
+
+    for item in search_results.get('items', []):
+        app_id = item.get('id')
+
+        currency = item.get('price', {}).get('currency', 'USD')
+        price = item.get('price', {}).get('final', 0) / 100
+
+        platforms = ', '.join([platform for platform, supported in item.get('platforms', {}).items() if supported])
+
+        content = f'Price: {price:.2f} {currency} | Platforms: {platforms}'
```
Bnyro: You should make `content` a list and join it together in the end with `'content': ' | '.join(content)`

searx/engines/steam.py
```diff
@@ -0,0 +1,62 @@
...
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+categories = []
+
+base_url = "https://store.steampowered.com"
+
+steam_app_details = False
```
Bnyro: Please add some explanatory comment here about what this settings does

hezhijie0327: Added.

searx/engines/steam.py
```diff
@@ -0,0 +1,71 @@
...
+        platforms = ', '.join([platform for platform, supported in item.get('platforms', {}).items() if supported])
+
+        content = [f'Price: {price:.2f} {currency}', f'Platforms: {platforms}']
+
+        if steam_app_details:
+            app_data = get(f'{base_url}/api/appdetails/?appids={app_id}').json().get(str(app_id), {})
+
+            description = app_data.get('data', {}).get('short_description') if app_data.get('success') else None
+            if description:
+                content.append(f'Description: {description}')
```
return42: I think we should not offer such an option .. it will fire at least 10 additional HTTP request, can we please remove this option?

searx/engines/steam.py
```diff
@@ -0,0 +1,71 @@
...
+
+from urllib.parse import urlencode
+
+from searx.network import get
+from searx.utils import html_to_text
+
+about = {
+    "website": 'https://store.steampowered.com/',
+    "wikidata_id": 'Q337535',
+    "use_official_api": True,
```
return42: Do you have link to the official API docs?

return42 (changes requested):

searx/engines/steam.py
hezhijie0327: Removed.

searx/engines/steam.py
hezhijie0327: no... it should be `False`, clone from HuggingFace engine forgot to change it.

return42 (approved):
I squashed the commits into one and modified the dict results into typed results ([EngineResults](https://docs.searxng.org/dev/engines/index.html#searx.result_types.EngineResults) and [MainResult](https://docs.searxng.org/dev/result_types/main_result.html)). For details see [diff](https://github.com/searxng/searxng/compare/099341b5b1ddf64332380b2e6b289bfa6c57f7e1..1caee5e675c6b2dedb0005d3d12d566d18e151bd).

Bnyro (approved):

Issue #4405, opened 2025-03-01
favicon_proxy: sqlite3.OperationalError: attempt to write a readonly database


inetol:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

`2025.3.1+1d16b9427`

**How did you install SearXNG?**

`Dockerfile`

**What happened?**

I noticed that the instance was writing lots of logs, and I was only seeing errors like the ones below

**How To Reproduce**

I don't know, favicons were enabled a month ago and nothing weird happened until now

**Expected behavior**

No errors

**Screenshots & Logs**

```console
[pid: 8|app: 0|req: 1670/1939] 10.201.3.188 () {58 vars in 1936 bytes} [Sat Mar  1 13:27:54 2025] GET /favicon_proxy?authority=REDACTED => generated 265 bytes in 4 msecs (HTTP/1.1 500) 7 headers in 270 bytes (1 switches on core 0)

2025-03-01 13:27:54,602 ERROR:searx.webapp: Exception on /favicon_proxy [GET]

Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/searxng/searx/favicons/proxy.py", line 147, in favicon_proxy
    data, mime = search_favicon(resolver, authority)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/searxng/searx/favicons/proxy.py", line 192, in search_favicon
[...]

PR #4650, opened 2025-04-22
[mod] persistent key/value pairs with expiry date (based on SQLite-DB)

return42:
## What does this PR do?

Implements a key/value store in class `searx.cache.ExpireCacheSQLite`. The key/value pairs are automatically deleted after a specified period of time (expire) .. similar to the expire already known from redis or valkey.

The class `searx.cache.ExpireCacheSQLite` is based on the `searx.sqlitedb.SQLiteApp`, which is already used in the caching of the favicons. However, the `searx.sqlitedb.SQLiteApp` had to be revised again to handle known errors/problems (see below).

--> the favicons should also be tested

#### 3. [mod] engines: migration of the individual cache solutions to EngineCache

The `EngineCache` class replaces all previously individual solutions for caches in the context of the engines.

- demo_offline.py
- duckduckgo.py
- radio_browser.py
- soundcloud.py
- startpage.py
- wolframalpha_api.py
- wolframalpha_noapi.py

Search term to test most of the modified engines::

    !ddg !rb !sc !sp !wa test
    !ddg !rb !sc !sp !wa foo

For introspection of the DB, jump into developer environment and run command to show cache state::

    $ ./manage pyenv.cmd bash --norc --noprofile
    (py3) python -m searx.enginelib cache state

For more details read doc-strings and commit message.

#### 2. `[mod] ExpireCache - sqlite based key/value cache with expire time`

Implements a key/value store in class `searx.cache.ExpireCacheSQLite`. 

#### 1. `[fix] ResourceWarning: unclosed database in sqlite3`

Reported:

- https://github.com/inetol-infrastructure/searxng-container/issues/5

Related:

- https://github.com/searxng/searxng/issues/4405#issuecomment-2692352352

### DIFF
--- a/searx/cache.py
+++ b/searx/cache.py
@@ -16,20 +16,13 @@ import hashlib
 import hmac
 import os
 import pickle
-import secrets
 import sqlite3
 import string
 import tempfile
 import time
 import typing
 
-from base64 import urlsafe_b64encode, urlsafe_b64decode
-
 import msgspec
-
-from cryptography.fernet import Fernet
-from cryptography.hazmat.primitives import hashes
-from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
 
 from searx import sqlitedb
 from searx import logger
@@ -70,21 +63,14 @@ class ExpireCacheCfg(msgspec.Struct):  # pylint: disable=too-few-public-methods
       if required.
     """
 
-    # encryption of the values stored in the DB
-
     password: bytes = get_setting("server.secret_key").encode()  # type: ignore
-    """Password used in case of :py:obj:`ExpireCacheCfg.ENCRYPT_VALUE` is
-    ``True``.
+    """Password used by :py:obj:`ExpireCache.secret_hash`.
 
     The default password is taken from :ref:`secret_key <server.secret_key>`.
-    When the password is changed, the values in the cache can no longer be
-    decrypted, which is why all values in the cache are deleted when the
-    password is changed.
+    When the password is changed, the hashed keys in the cache can no longer be
+    used, which is why all values in the cache are deleted when the password is
+    changed.
     """
-
-    ENCRYPT_VALUE: bool = True
-    """Encrypting the values before they are written to the DB (see:
-    :py:obj:`ExpireCacheCfg.password`)."""
 
     def __post_init__(self):
         # if db_url is unset, use a default DB in /tmp/sxng_cache_{name}.db
@@ -138,8 +124,7 @@ class ExpireCache(abc.ABC):
 
     cfg: ExpireCacheCfg
 
-    hmac_iterations: int = 10_000
-    crypt_hash_property = "crypt_hash"
+    hash_token = "hash_token"
 
     @abc.abstractmethod
     def set(self, key: str, value: typing.Any, expire: int | None) -> bool:
@@ -154,16 +139,16 @@ class ExpireCache(abc.ABC):
         """Return *value* of *key*.  If key is unset, ``None`` is returned."""
 
     @abc.abstractmethod
-    def maintenance(self, force: bool = False, drop_crypted: bool = False) -> bool:
+    def maintenance(self, force: bool = False, truncate: bool = False) -> bool:
         """Performs maintenance on the cache.
 
         ``force``:
           Maintenance should be carried out even if the maintenance interval has
           not yet been reached.
 
-        ``drop_crypted``:
-           The encrypted values can no longer be decrypted (if the password is
-           changed), they must be removed from the cache.
+        ``truncate``:
+          Truncate the entire cache, which is necessary, for example, if the
+          password has changed.
         """
 
     @abc.abstractmethod
@@ -191,67 +176,13 @@ class ExpireCache(abc.ABC):
         _valid = "-_." + string.ascii_letters + string.digits
         return "".join([c for c in name if c in _valid])
 
-    def derive_key(self, password: bytes, salt: bytes, iterations: int) -> bytes:
-        """Derive a secret-key from a given password and salt."""
-        kdf = PBKDF2HMAC(
-            algorithm=hashes.SHA256(),
-            length=32,
-            salt=salt,
-            iterations=iterations,
-        )
-        return urlsafe_b64encode(kdf.derive(password))
-
     def serialize(self, value: typing.Any) -> bytes:
         dump: bytes = pickle.dumps(value)
-        if self.cfg.ENCRYPT_VALUE:
-            dump = self.encrypt(dump)
         return dump
 
     def deserialize(self, value: bytes) -> typing.Any:
-        if self.cfg.ENCRYPT_VALUE:
-            value = self.decrypt(value)
         obj = pickle.loads(value)
         return obj
-
-    def encrypt(self, message: bytes) -> bytes:
-        """Encode and decode values by a method using `Fernet with password`_ where
-        the key is derived from the password (PBKDF2HMAC_).  The *password* for
-        encryption is taken from the :ref:`server.secret_key`
-
-        .. _Fernet with password:  https://stackoverflow.com/a/55147077
-        .. _PBKDF2HMAC: https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/#pbkdf2
-        """
-
-        # Including the salt in the output makes it possible to use a random
-        # salt value, which in turn ensures the encrypted output is guaranteed
-        # to be fully random regardless of password reuse or message
-        # repetition.
-        salt = secrets.token_bytes(16)  # randomly generated salt
-
-        # Including the iteration count ensures that you can adjust
-        # for CPU performance increases over time without losing the ability to
-        # decrypt older messages.
-        iterations = int(self.hmac_iterations)
-
-        key = self.derive_key(self.cfg.password, salt, iterations)
-        crypted_msg = Fernet(key).encrypt(message)
-
-        # Put salt and iteration count on the beginning of the binary
-        token = b"%b%b%b" % (salt, iterations.to_bytes(4, "big"), urlsafe_b64encode(crypted_msg))
-        return urlsafe_b64encode(token)
-
-    def decrypt(self, token: bytes) -> bytes:
-        token = urlsafe_b64decode(token)
-
-        # Strip salt and iteration count from the beginning of the binary
-        salt = token[:16]
-        iterations = int.from_bytes(token[16:20], "big")
-
-        key = self.derive_key(self.cfg.password, salt, iterations)
-        crypted_msg = urlsafe_b64decode(token[20:])
-
-        message = Fernet(key).decrypt(crypted_msg)
-        return message
 
     def secret_hash(self, name: str | bytes) -> str:
         """Creates a hash of the argument ``name``.  The hash value is formed
@@ -276,7 +207,6 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
     - :py:obj:`ExpireCacheCfg.MAXHOLD_TIME`
     - :py:obj:`ExpireCacheCfg.MAINTENANCE_PERIOD`
     - :py:obj:`ExpireCacheCfg.MAINTENANCE_MODE`
-    - :py:obj:`ExpireCacheCfg.ENCRYPT_VALUE`
     """
 
     DB_SCHEMA = 1
@@ -300,18 +230,17 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
         if not ret_val:
             return False
 
-        if self.cfg.ENCRYPT_VALUE:
-            new = hashlib.sha256(self.cfg.password).hexdigest()
-            old = self.properties(self.crypt_hash_property)
-            if old != new:
-                if old is not None:
-                    log.warning("[%s] crypt token changed: drop all cache tables", self.cfg.name)
-                self.maintenance(force=True, drop_crypted=True)
-                self.properties.set(self.crypt_hash_property, new)
+        new = hashlib.sha256(self.cfg.password).hexdigest()
+        old = self.properties(self.hash_token)
+        if old != new:
+            if old is not None:
+                log.warning("[%s] hash token changed: truncate all cache tables", self.cfg.name)
+            self.maintenance(force=True, truncate=True)
+            self.properties.set(self.hash_token, new)
 
         return True
 
-    def maintenance(self, force: bool = False, drop_crypted: bool = False) -> bool:
+    def maintenance(self, force: bool = False, truncate: bool = False) -> bool:
 
         if not force and int(time.time()) < self.next_maintenance_time:
             # log.debug("no maintenance required yet, next maintenance interval is in the future")
@@ -321,7 +250,7 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
         # (e.g. in multi thread or process environments).
         self.properties.set("LAST_MAINTENANCE", "")  # hint: this (also) sets the m_time of the property!
 
-        if drop_crypted:
+        if truncate:
             self.truncate_tables(self.table_names)
             return True
 
--- a/searx/cache.py
+++ b/searx/cache.py
@@ -83,12 +83,12 @@ class ExpireCacheStats:
     """Dataclass wich provides information on the status of the cache."""
 
     cached_items: dict[str, list[tuple[str, typing.Any, int]]]
-    """Values in the cache mapped by table name.
+    """Values in the cache mapped by context name.
 
     .. code: python
 
        {
-           "table name": [
+           "context name": [
                ("foo key": "foo value", <expire>),
                ("bar key": "bar value", <expire>),
                # ...
@@ -98,22 +98,22 @@ class ExpireCacheStats:
     """
 
     def report(self):
-        c_tables = 0
+        c_ctx = 0
         c_kv = 0
         lines = []
 
-        for table_name, kv_list in self.cached_items.items():
-            c_tables += 1
+        for ctx_name, kv_list in self.cached_items.items():
+            c_ctx += 1
             if not kv_list:
-                lines.append(f"[{table_name:20s}] empty")
+                lines.append(f"[{ctx_name:20s}] empty")
                 continue
 
             for key, value, expire in kv_list:
                 valid_until = datetime.datetime.fromtimestamp(expire).strftime("%Y-%m-%d %H:%M:%S")
                 c_kv += 1
-                lines.append(f"[{table_name:20s}] {valid_until} {key:12}" f" --> ({type(value).__name__}) {value} ")
+                lines.append(f"[{ctx_name:20s}] {valid_until} {key:12}" f" --> ({type(value).__name__}) {value} ")
 
-        lines.append(f"number of tables: {c_tables}")
+        lines.append(f"Number of contexts: {c_ctx}")
         lines.append(f"number of key/value pairs: {c_kv}")
         return "\n".join(lines)
 
@@ -127,15 +127,27 @@ class ExpireCache(abc.ABC):
     hash_token = "hash_token"
 
     @abc.abstractmethod
-    def set(self, key: str, value: typing.Any, expire: int | None) -> bool:
+    def set(self, key: str, value: typing.Any, expire: int | None, ctx: str | None = None) -> bool:
         """Set *key* to *value*.  To set a timeout on key use argument
         ``expire`` (in sec.).  If expire is unset the default is taken from
         :py:obj:`ExpireCacheCfg.MAXHOLD_TIME`.  After the timeout has expired,
         the key will automatically be deleted.
+
+        The ``ctx`` argument specifies the context of the ``key``.  A key is
+        only unique in its context.
+
+        The concrete implementations of this abstraction determine how the
+        context is mapped in the connected database.  In SQL databases, for
+        example, the context is a DB table or in a Key/Value DB it could be
+        a prefix for the key.
+
+        If the context is not specified (the default is ``None``) then a
+        default context should be used, e.g. a default table for SQL databases
+        or a default prefix in a Key/Value DB.
         """
 
     @abc.abstractmethod
-    def get(self, key: str, default=None) -> typing.Any:
+    def get(self, key: str, default=None, ctx: str | None = None) -> typing.Any:
         """Return *value* of *key*.  If key is unset, ``None`` is returned."""
 
     @abc.abstractmethod
@@ -171,7 +183,7 @@ class ExpireCache(abc.ABC):
     @staticmethod
     def normalize_name(name: str) -> str:
         """Returns a normalized name that can be used as a file name or as a SQL
-        table name."""
+        table name (is used, for example, to normalize the context name)."""
 
         _valid = "-_." + string.ascii_letters + string.digits
         return "".join([c for c in name if c in _valid])
@@ -320,14 +332,15 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
 
     # implement ABC methods of ExpireCache
 
-    def set(self, key: str, value: typing.Any, expire: int | None, table: str | None = None) -> bool:
-        """Set key/value in ``table``.  If expire is unset the default is taken
-        from :py:obj:`ExpireCacheCfg.MAXHOLD_TIME`.  If ``table`` argument is
-        ``None`` (the default), a table name is generated from the
-        :py:obj:`ExpireCacheCfg.name`.  If DB ``table`` does not exists, it will be
-        created (on demand) by :py:obj:`self.create_table
+    def set(self, key: str, value: typing.Any, expire: int | None, ctx: str | None = None) -> bool:
+        """Set key/value in DB table given by argument ``ctx``.  If expire is
+        unset the default is taken from :py:obj:`ExpireCacheCfg.MAXHOLD_TIME`.
+        If ``ctx`` argument is ``None`` (the default), a table name is
+        generated from the :py:obj:`ExpireCacheCfg.name`.  If DB table does not
+        exists, it will be created (on demand) by :py:obj:`self.create_table
         <ExpireCacheSQLite.create_table>`.
         """
+        table = ctx
         self.maintenance()
 
         value = self.serialize(value=value)
@@ -360,12 +373,14 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
 
         return True
 
-    def get(self, key: str, default=None, table: str | None = None) -> typing.Any:
-        """Get value of ``key`` from ``table``.  If ``table`` argument is
-        ``None`` (the default), a table name is generated from the
-        :py:obj:`ExpireCacheCfg.name`.  If ``key`` not exists (in table), the
-        ``default`` value is returned.
+    def get(self, key: str, default=None, ctx: str | None = None) -> typing.Any:
+        """Get value of ``key`` from table given by argument ``ctx``.  If
+        ``ctx`` argument is ``None`` (the default), a table name is generated
+        from the :py:obj:`ExpireCacheCfg.name`.  If ``key`` not exists (in
+        table), the ``default`` value is returned.
+
         """
+        table = ctx
         self.maintenance()
 
         if not table:
--- a/searx/cache.py
+++ b/searx/cache.py
@@ -1,0 +1,461 @@
+"""Implementation of caching solutions.
+
+- :py:obj:`searx.cache.ExpireCache` and its :py:obj:`searx.cache.ExpireCacheCfg`
+
+----
+"""
+
+from __future__ import annotations
+
+__all__ = ["ExpireCacheCfg", "ExpireCacheStats", "ExpireCache", "ExpireCacheSQLite"]
+
+import abc
+import dataclasses
+import datetime
+import hashlib
+import hmac
+import os
+import pickle
+import secrets
+import sqlite3
+import string
+import tempfile
+import time
+import typing
+
+from base64 import urlsafe_b64encode, urlsafe_b64decode
+
+import msgspec
+
+from cryptography.fernet import Fernet
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+
+from searx import sqlitedb
+from searx import logger
+from searx import get_setting
+
+log = logger.getChild("cache")
+
+
+class ExpireCacheCfg(msgspec.Struct):  # pylint: disable=too-few-public-methods
+    """Configuration of a :py:obj:`ExpireCache` cache."""
+
+    name: str
+    """Name of the cache."""
+
+    db_url: str = ""
+    """URL of the SQLite DB, the path to the database file.  If unset a default
+    DB will be created in `/tmp/sxng_cache_{self.name}.db`"""
+
+    MAX_VALUE_LEN: int = 1024 * 10
+    """Max lenght of a *serialized* value."""
+
+    MAXHOLD_TIME: int = 60 * 60 * 24 * 7  # 7 days
+    """Hold time (default in sec.), after which a value is removed from the cache."""
+
+    MAINTENANCE_PERIOD: int = 60 * 60  # 2h
+    """Maintenance period in seconds / when :py:obj:`MAINTENANCE_MODE` is set to
+    ``auto``."""
+
+    MAINTENANCE_MODE: typing.Literal["auto", "off"] = "auto"
+    """Type of maintenance mode
+
+    ``auto``:
+      Maintenance is carried out automatically as part of the maintenance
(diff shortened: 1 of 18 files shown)


searx/enginelib/__init__.py
```diff
@@ -9,16 +19,158 @@
...
+    print(str(ENGINES_CACHE.properties))
+
+
+@app.command()
+def maintenance(force: bool = True):
+    """Carry out maintenance on cache of the engines."""
+    ENGINES_CACHE.maintenance(force=force)
+
+
+class EngineCache:
```
Bnyro, line 74: > Implements a key/value store in class searx.cache.ExpireCacheSQLite. The key/value pairs are automatically deleted after a specified period of time (expire) .. similar to the expire already known from redis or valkey.
>
As far as I know, SQLite is not really intended to be used as a key value store. I think it's okay to have it as a fallback, but we should always use Redis if it's available because its performance will be much better as it's actually just made for that single purpose. So I think this should be similar to what I started to implement in https://github.com/searxng/searxng/pull/4314 with different engine cache implementations depending on redis is available or not (use sqlite then as fallback).

searx/engines/radio_browser.py
```diff
@@ -52,18 +54,33 @@
...
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache("radio_browser")
+    server_list()
+
+
+def server_list() -> list[str]:
+
+    servers = CACHE.get("servers", [])
+    if servers:
+        return servers
```
Bnyro, line 72: Isn't this just the same behavior as
```suggestion
    return CACHE.get("servers", [])
```
?

searx/engines/radio_browser.py
```diff
@@ -52,18 +54,33 @@
...
     ips = socket.getaddrinfo("all.api.radio-browser.info", 80, 0, 0, socket.IPPROTO_TCP)
     for ip_tuple in ips:
         _ip: str = ip_tuple[4][0]  # type: ignore
         url = socket.gethostbyaddr(_ip)[0]
         srv = "https://" + url
         if srv not in servers:
             servers.append(srv)
+    # update server list once in 24h
+    CACHE.set(key="servers", value=servers, expire=60 * 60 * 24)
+    return servers
```
Bnyro, line 92: Maybe we should add a check here to throw/show a proper error message if `servers` is empty? Otherwise `random.choice` will just crash where this method is called.

searx/enginelib/__init__.py
```diff
@@ -9,16 +19,158 @@
...
+    the last modification date of the LAST_MAINTENANCE property and the hash value
+    of the password can be used to detect whether the password has been changed (in
+    this case the DB entries can no longer be decrypted and the entire cache must be
+    discarded).
+    """
+
+    def __init__(self, engine_name: str, expire: int | None = None):
+        self.expire = expire or ENGINES_CACHE.cfg.MAXHOLD_TIME
+        _valid = "-_." + string.ascii_letters + string.digits
+        self.table_name = "".join([c if c in _valid else "_" for c in engine_name])
```
Bnyro, line 157: This should probably be replaced with `normalize_name` from `searx.cache`?

searx/enginelib/__init__.py
dalf: If the SQLite solution is kept, here some links that might help to avoid some already seen problems:
* https://github.com/piskvorky/sqlitedict
* https://stackoverflow.com/questions/47233562/key-value-store-in-python-for-possibly-100-gb-of-data-without-client-server

(there is [lmdb](https://lmdb.readthedocs.io/en/release/) too, but there is no wheels for the ARMv7 architecture)

searx/cache.py
```diff
@@ -0,0 +1,461 @@
...
+    def build_cache(cfg: ExpireCacheCfg) -> ExpireCache:
+        """Factory to build a caching instance.
+
+        .. note::
+
+           Currently, only the SQLite adapter is available, but other database
+           types could be implemented in the future, e.g. a Valkey (Redis)
+           adapter.
+        """
+        return ExpireCacheSQLite(cfg)
```
return42, line 181: > As far as I know, SQLite is not really intended to be used as a key value store.

We can implement a Redis/Valkey adapter based on the abstract class `ExpireCache(abc.ABC)`.

> but we should always use Redis if it's available because its performance will be much better as it's actually just made for that single purpose.

I wouldn't expect too much:

1. We only have a few values ​​in the cache (hundreds or less). Key/value databases only show their performance advantages with very large data sets (millions).

2. The SQLite3 DB system runs in-process, meaning the DB query is processed in-process. A key/value DB requires a request/response cycle, which must run over IP stack or at least over sockets.

Bnyro: > We can implement a Redis/Valkey adapter based on the abstract class ExpireCache(abc.ABC)
>
I think that'd be a good idea. Some people are probably more familiar with Redis and already have the skills to work/debug with it.

> We only have a few values ​​in the cache (hundreds or less). Key/value databases only show their performance advantages with very large data sets (millions).
>
That's indeed true looking at the current usages of `EngineCache`.

> The SQLite3 DB system runs in-process, meaning the DB query is processed in-process. A key/value DB requires a request/response cycle, which must run over IP stack or at least over sockets.
>
Nevertheless it's easier for instance admins to maintain and scale (e.g. limit resources) of database accesses when using Redis. I think we should at least consider the option to allow instance admins to use their existing Redis installation (which they already have because of the limiter anyways).

searx/enginelib/__init__.py
return42: @dalf thanks for the links .. just a short feedback after my brief research:

- https://github.com/piskvorky/sqlitedict I can't recommend, not even for research ..

They make similar mistakes to the ones I'm correcting in this PR... I suspect this is also one of the reasons why the project hasn't been continued for three years (with Python 3.12 and also with Python 3.13, the SQLite3 integration has changed; there are other features, e.g., the "serialized" threading model... which doesn't work smoothly yet, and a lot also depends on the options set at the time of compilation of the Python interpreter... it gets ugly when you delve deeper into it).

Threads and the meaning of "Autocommit" / The latter has a different meaning in Python's SQLite3 module than the "autocommit" familiar from DB systems... In DB systems, "autocommit" stands for transaction-free SQL update/insert, and in the Python module, the "Autocommit" option determines whether the `with` context is used. Python closes a transaction or not when leaving the `with` context of a courser (that was also my mistake in my first implementation) .. BTW: the `with` context of a DB-connection never closes the DB-connection  (some knowledge I had to learn).

- [lmdb](https://lmdb.readthedocs.io/en/release/) 

I didn't know about it (thanks for the tip), it seems to be extremely efficient, but I think we've already decided on Valkey (Redis) and I would be reluctant to integrate a third DB system into SearXNG at the moment.

dalf:
cryptography relies on Rust

As far I understand the values are encrypted with a code similar to this function:

```python
from base64 import urlsafe_b64encode, urlsafe_b64decode

from cryptography.fernet import Fernet
from cryptography.hazmat.primitives import hashes
from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
from searx import get_setting

def serialize(value):
  # pickle.dumps : fine because Fernet has authentication (as far I understand...), 
  # otherwise can lead to execution of unknown code (if the database is not trusted)
  # msgpack can be alternative (supported by msgspec).
  dump = pickle.dumps(value)  
  salt = secrets.token_bytes(16)
  kdf = PBKDF2HMAC(algorithm=hashes.SHA256(), length=32, salt=salt, iterations=10000)
  secret_key = get_setting("server.secret_key").encode()
  derived_key =  urlsafe_b64encode(kdf.derive(secret_key))
  crypted_msg = Fernet.encrypt(derived_key).encrypt(dump)
  token = b"%b%b%b" % (salt, iterations.to_bytes(4, "big"), urlsafe_b64encode(crypted_msg))
  return urlsafe_b64encode(token)
```

As far I know, the values are not encrypted with Redis, why on SQLite?

Everything relies on secret_key which is stored next to the SQLite file. If the SQLite database is accessed by a third-party, most probably this third-party can access the secret_key.

In the database, if the purpose is to protected the secret_key value:
* 10000 iterations is [too low](https://en.wikipedia.org/wiki/PBKDF2#Purpose_and_operation), however the entropy is already high (from what I can remember, it just 32 random chars), so a simple sha256 should be good enough if this is the case.
* a SHA256 of secret_key is [stored in the database](https://github.com/searxng/searxng/blob/5adbad5bd32557a4ba834af156c9bb4d602b0a5f/searx/cache.py#L303-L310), I can't say if it reduce the protection given by PBKDF2HMAC (even if more iteration).

searx/cache.py
```diff
@@ -0,0 +1,461 @@
...
+                lines.append(f"[{table_name:20s}] {valid_until} {key:12}" f" --> ({type(value).__name__}) {value} ")
+
+        lines.append(f"number of tables: {c_tables}")
+        lines.append(f"number of key/value pairs: {c_kv}")
+        return "\n".join(lines)
+
+
+class ExpireCache(abc.ABC):
+    """Abstract base class for the implementation of a key/value cache
+    with expire date."""
```
dalf, line 123: If the purpose is to have multiple implementation of this class (SQLite and Redis), I prefer to have a dedicated module `searx.cache`

* `searx/cache/__init__.py`
* `searx/cache/sqlitecache.py`
* `searx/cache/rediscache.py`

(or similar names)

searx/cache.py
```diff
@@ -0,0 +1,461 @@
...
+        """Performs maintenance on the cache.
+
+        ``force``:
+          Maintenance should be carried out even if the maintenance interval has
+          not yet been reached.
+
+        ``drop_crypted``:
+           The encrypted values can no longer be decrypted (if the password is
+           changed), they must be removed from the cache.
+        """
```
dalf, line 164: Related to:
https://github.com/dalf/searxng/blob/e92755d358df5b34b0181f48f8ba02c7f2939e8f/searx/search/checker/scheduler.py#L1-L13

The initial version was able to schedule multiple functions, and was making sure to run the function once for the whole server (not once per worker):

https://github.com/dalf/searxng/blob/e92755d358df5b34b0181f48f8ba02c7f2939e8f/searx/search/checker/scheduler.py

Another related PR : https://github.com/searxng/searxng/pull/1513

searx/cache.py
```diff
@@ -0,0 +1,461 @@
...
+    def set(self, key: str, value: typing.Any, expire: int | None) -> bool:
+        """Set *key* to *value*.  To set a timeout on key use argument
+        ``expire`` (in sec.).  If expire is unset the default is taken from
+        :py:obj:`ExpireCacheCfg.MAXHOLD_TIME`.  After the timeout has expired,
+        the key will automatically be deleted.
+        """
+
+    @abc.abstractmethod
+    def get(self, key: str, default=None) -> typing.Any:
+        """Return *value* of *key*.  If key is unset, ``None`` is returned."""
```
dalf, line 151: It seems the engine implementations use the factory pattern:
* read a value from the cache
* if there is no value, create and store the value
* return the value

Would it make sense to add this pattern here so each cache implementation could use its native implementation:
* SQL transaction
* redis transaction
?

Also, it would be simple to cache a value in memory (not sure about this)

return42:
> cryptography relies on Rust

Why do you mention that? .. because of the wheels? If so, they are all available--> https://pypi.org/project/cryptography/#files

> As far I understand the values are encrypted with a code similar to this function:
> As far I know, the values are not encrypted with Redis, why on SQLite?

The idea is to have an abstraction for an "ExpireCache", which form it takes (Redis, SQLite or ..) is of secondary importance. 

This abstraction can be used for a variety of applications, here in this PR we use it (initial) for the global `searx.enginelib.ENGINES_CACHE` (and for convenience there exists a *table-name* wrapper `searx.enginelib.EngineCache`)

An instance of the "ExpireCache" abstraction is created/parameterized using a configuration object of type `searx.cache.ExpireCacheCfg`.

Not required in this PR, but possible for other use cases: an `ExpireCacheCfg` object can be loaded via a configuration file (TOML, YAML, ... / like we did for the favicons config).

The bool `ExpireCacheCfg.ENCRYPT_VALUE` is the option to encrypt the values or not.

I decided to set the default of `ENCRYPT_VALUE` to value `True` --> encrypt values by default

> Everything relies on secret_key which is stored next to the SQLite file. If the SQLite database is accessed by a third-party, most probably this third-party can access the secret_key.

If we only have `searx.enginelib.ENGINES_CACHE` in mind .. yes, but this is an implementation which can be used for other use cases also.

This is a cache, and these values ​​usually disappear after a certain period of time. The maximum retention period for objects in the `searx.enginelib.ENGINES_CACHE` is set to 7 days.

ExpireCache encryption doesn't claim to be the hardest encryption to crack... it's not Fortnox .. it's a compromise of 

1. encryption and decryption infrastructure *by zero config* 
2. performance.

This compromise is good enough for most tasks, and certainly good enough for the values ​​stored in the `searx.enginelib.ENGINES_CACHE` .. one might consider whether we even need encryption for the `ENGINES_CACHE`.

However, since we don't know whether there will be engines that need to cache search terms at some point, I thought it would be a good idea to have a simple but high-performance encryption right from the start that doesn't require any configuration or additional infrastructure (*zero config*).

If one day there are applications that have higher encryption requirements, then these applications will have to implement their own encryption and decryption infrastructure.

> In the database, if the purpose is to protected the secret_key value:
> 
> * 10000 iterations is [too low](https://en.wikipedia.org/wiki/PBKDF2#Purpose_and_operation), however the entropy is already high (from what I can remember, it just 32 random chars), so a simple sha256 should be good enough if this is the case.
> 
> * a SHA256 of secret_key is [stored in the database](https://github.com/searxng/searxng/blob/5adbad5bd32557a4ba834af156c9bb4d602b0a5f/searx/cache.py#L303-L310), I can't say if it reduce the protection given by PBKDF2HMAC (even if more iteration).

I don't think so, but I can't rule it out either...

One could dwell on this point for a very long time ;-)... but in the end, what matters is how high the "need" is... and that is very low for the use cases known so far and more than adequately covered by the current implementation.

searx/cache.py
return42: I had already thought about that, which is why I export the following names from `searx.cache`:

```py
__all__ = ["ExpireCacheCfg", "ExpireCacheStats", "ExpireCache", "ExpireCacheSQLite"]
```
I can do it here in this PR, but it would only be one module in the package. We can do that later, too. Either way is fine with me ...

searx/cache.py
return42: The maintenance method is implemented in the inheritance and the heirs will know how to perform the maintenance interval only once across all processes (and not in each of the n-workers).

For example, in the `ExpireCacheSQLite.maintenance()` a value is set in a property table. This property (stored in the SQL-Database) prevents a second process from performing the same maintenance again.

You can basically do this with any DB application... only if you try to keep the values ​​only in the memory of the processes, then this is not possible across processes (logically).

dalf:
> Why do you mention that? .. because of the wheels? If so, they are all available--> https://pypi.org/project/cryptography/#files

Not anymore for the armv7 architecture.

https://github.com/pyca/cryptography/issues/6286#issuecomment-922563400

> ExpireCache encryption doesn't claim to be the hardest encryption to crack... it's not Fortnox .. it's a compromise of encryption and decryption infrastructure by zero config [and] performance.
> ...
> I don't think so, but I can't rule it out either...

Sorry to ask again, why encryption of the values is needed?

| engine               | key                 | value            |
| -- | -- | -- |
| duckduckgo | contains the hashed user query |  `vqd` : a per query unique identifier | 
| radio_browser | `servers` | public data | 
| soundcloud | `guest_client_id` | a global unique identifier | 
| startpage | `SC_CODE` | a global unique identifier | 
| wolframalpha_api | `token` | a global unique identifier |
| wolframalpha_noapi | `token` | a global unique identifier |

Currently, the only sensitive data is the cache key of the duckduckgo engine where a hash function is enough (there is no need to decrypt the key).

What data will be encrypted in future?
What are the threats?
A third-party gets access to the database but not settings.yml?

searx/cache.py
return42: > It seems the engine implementations use the factory pattern:

No .. `get()` is simply a *getter* and `set()` is simply a *setter* .. nothing more.

We should not mutate the *getter* to a  *"get and if not exists set to default .. and get default"*

> Would it make sense to add this pattern here so each cache implementation could use its native implementation:

When the *setter* method returns the transaction should be commited / whatever DB system .. otherwise the new seted value is not available in other DB connections.


> Also, it would be simple to cache a value in memory (not sure about this)

Makes absolutely no sense to me .. What we need is a cross-process memory. The simplest basis for implementing such a thing is a SQLite3 database, and this is available in every Python ecosystem.

It is imperative that we make values ​​such as a volatile API key available across processes. As long as we only have such values ​​in the memory of one process, each process (each walker) must regenerate this value.

Example: startpage engine requires a *vqd code*, which the engine retrieves via a bot-suspicious request to startpage.com. If 16 or 32 workers are running on a server, then up to 32 requests to startpage.com are necessary. This in turn contributes to the server IP heating up and possibly being blocked by startpage.

dalf: > No .. get() is simply a getter and set() is simply a setter .. nothing more.

I'm not saying get and set should be removed but suggest to implement a factory pattern.

radio_browser:
```python
def server_list() -> list[str]:
    servers = CACHE.get("servers", [])
    if servers:
        return servers
    ...
    CACHE.set(key="servers", value=servers, expire=60 * 60 * 24)

    return servers
```

startpage:
```python
def get_sc_code(searxng_locale, params):
    sc_code = CACHE.get("SC_CODE", "")
    if sc_code:
        return sc_code
    ...
    CACHE.set(key="SC_CODE", value=sc_code, expire=sc_code_cache_sec)
    return sc_code
```

same for the other engines except duckduckgo.py which use the params (actually the HTTP headers).

Related: when SearXNG starts, without lock, multiple workers might create and store a value for the same key at the same time (not really an issue).

return42: > but suggest to implement a factory pattern.

And what should this pattern look like? ... Even if we currently have similar code blocks in the engines, the example with DDG already shows that the processes for obtaining values can be very different.

> Related: when SearXNG starts, without lock, multiple workers might create and store a value for the same key at the same time (not really an issue).

There are still several open issues at this point, on the one hand with the initialization, on the other hand with the retrieval of updated values ... for example, the `radio_browser.server_list()` function is also executed in the request and can take up to 40sec ... but these problems are not new and should be solved in another PR (finding a solution for this will probably not be easy with the existing infrastructure / I don't want to solve it in this PR).

dalf: Here a draft where a bunch of edges case are not fixed: https://gist.github.com/dalf/d277af159c19030b6da6287bd04e9846#file-cache_with_factory-py-L44

The idea is to wrap EngineCache with ExpireCacheExtension which provide one additional method:

```python
def register_factory(self, key: str, factory: Callable, expired: int | None, autoupdate: bool):
```

* If `autoupdate` is False, a call `get(key)` automatically updates the value if it has expired using the factory.
* If `autoupdate` is True, a scheduler automatically updates the values before expiration, so for example `radio_browser.server_list()` is always fast during the user query. 
 
Since there is a scheduler, this is related to the comment on line 164. It is implemented using asyncio because it does exactly what is required: be able to schedule multiple tasks. All run inside one dedicated thread to minimize the threads, but it's easy to call a thread pool instead.

This require more thinking, this is just a draft, it can be done (or not) in another PR, feel free to close that.

return42:
> Not anymore for the armv7 architecture.

I was not aware of that... thank you very much for the tip... with this knowledge the prophylactic encryption no longer makes sense... I have now removed it :+1:

searx/cache.py
return42: > This require more thinking, this is just a draft, it can be done (or not) in another PR, feel free to close that.

I think your POC is heading in the right direction, but as you yourself have already noted, it won't cover all the scenarios required today.

factory: Yes, it could be done that way, but it wouldn't cover, for example, the process flow of the ddg engine (which currently generates 99% of all key/value pairs in the cache).

scheduler: In general, we need a infrastructure (a second server process) for recurring and/or long-running activities... but these activities shouldn't run in the process of the outgoing requests, regardless of whether it's isolated in a subprocess, thread, or processed async --> The engine requests/response handles are running in threads, and these threads remain in memory until they terminate themselves. An engine timeout of, say, 3 seconds doesn't guarantee that the thread will be removed from memory after 3 seconds. It can remain there until the thread terminates itself / in the worst case scenario of an error, this can be forever.

At least, those are my initial thoughts on both topics. However, we should discuss this further in two new PRs (or two new discussion posts).

dalf:
Related to https://github.com/searxng/searxng/issues/4676#issuecomment-2837443209

>  I would like to use the cryptography

See my questions above: what data will be encrypted in future? what are the threats?

return42:
@dalf @Bnyro Is there anything that stands in the way of merging this PR? ... or have I missed something else from the review comments?

searx/cache.py
```diff
@@ -0,0 +1,390 @@
...
+        table_name = table
+        if not table_name:
+            table_name = self.normalize_name(self.cfg.name)
+        self.create_table(table_name)
+
+        sql = (
+            f"INSERT INTO {table_name} (key, value, expire) VALUES (?, ?, ?)"
+            f"    ON CONFLICT DO "
+            f"UPDATE SET value=?, expire=?"
+        )
```
dalf, line 364: Nit: Why not UPSERT ?

https://sqlite.org/lang_upsert.html
> UPSERT syntax was added to SQLite with version 3.24.0 (2018-06-04).

searx/cache.py
```diff
@@ -0,0 +1,390 @@
...
+
+        if table:
+            with self.DB:
+                self.DB.execute(sql, (key, value, expire, value, expire))
+        else:
+            with self.connect() as conn:
```
dalf, line 370: The usage of with is misleading: it won't close the connection at the end of the block. Why [`SQLiteAppl.connect`](https://github.com/searxng/searxng/blob/fd33559cfb623c1e9edef7ab08c1e9fbf43824f8/searx/sqlitedb.py#L116-L129) is not [contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) ? If there is an exception on line 358, the connection won't be closed by the line 359.

This comment applies for all the usages of `with self.connect()` in this file.

---

As far I understand, each call to the set method:
* create a new connection
* run one query
* close the connection

Why not to use [`self.DB`](https://github.com/searxng/searxng/blob/fd33559cfb623c1e9edef7ab08c1e9fbf43824f8/searx/sqlitedb.py#L155-L179) like it is done in [`table_names`](https://github.com/searxng/searxng/blob/db6241267cca9d4548b8b2280e415637a3bc2bb3/searx/cache.py#L300-L305) property definition? Why when the `table` parameter is specify this is different?

dalf (commented):
I'm lost by the number of abstractions.

However here a small benchmark. At the end of `cache.py` I have added : 

```python
##########################################

def main():
    conf = ExpireCache.build_cache(
        ExpireCacheCfg(
            name="ENGINES_CACHE",
            MAXHOLD_TIME=60 * 60 * 24 * 7,  # 7 days
            MAINTENANCE_PERIOD=60 * 60,  # 2h
        )
    )
    from time import perf_counter
    count = 1000
    t_before = perf_counter()
    for _ in range(count):
        conf.set("key", 12, expire=1000)
    t_after = perf_counter()
    total_time = t_after - t_before
    print(total_time / count * 1000, "ms")

if __name__ == "__main__":
    main()
```

then run that `python -m searx.cache`, so **one call** to the set method cost

|||
| -: | -- |
| 0.7 ms | server with SSD |
| 34 ms | server with HDD 5400RPM, ZFS, a lot of RAM (only IO, not CPU bound) |

searx/cache.py
```diff
@@ -0,0 +1,390 @@
...
+
+    @property
+    def next_maintenance_time(self) -> int:
+        """Returns (unix epoch) time of the next maintenance."""
+
+        return self.cfg.MAINTENANCE_PERIOD + self.properties.m_time("LAST_MAINTENANCE", int(time.time()))
+
+    # implement ABC methods of ExpireCache
+
+    def set(self, key: str, value: typing.Any, expire: int | None, table: str | None = None) -> bool:
```
dalf: The parameter `table` is not in the abstract method `ExpireCache.set(self, key: str, value: typing.Any, expire: int | None) -> bool`.

This parameter is never used (?), what is the intent? Is it for debugging purpose?

searx/cache.py
return42: > Why not UPSERT ?
> UPSERT syntax was added to SQLite with version 3.24.0 (2018-06-04).

:-) read the second paragraph from the article you linked ..

> An UPSERT is an ordinary [INSERT](https://sqlite.org/lang_insert.html) statement that is followed by one or more ON CONFLICT clauses, as shown in the syntax diagram above.

searx/cache.py
return42: From the doc-string:

*"Set key/value in ``table`` ... If ``table`` argument is ``None`` (the default), a table name is generated from the :py:obj:`ExpireCacheCfg.name`"*

> what is the intent? 

Instead of saving all `keys` in one table, it is possible to save the `key`s in different tables.  This has the advantage that the table represents the context and identical `key`s can be used in different contexts 

> This parameter is never used (?),

It is used, see e.g. doc-string for the wrapper `searx.enginelib.EngineCache`: 

1. The context is the name of the engine --> it is therefore possible that two different engines can use an identical `key` (e.g. more than one engine can have a key like `api_key` or `token`).

2. Another advantage is that not all engines write to the same DB table, so the probability of a table lock is limited to one engine ... all engines can write to the DB at the same time without blocking each other!

dalf: Then the signature of `ExpireCache.set` should include the `table` parameter:

```python
def set(self, key: str, value: typing.Any, expire: int | None, table: str | None = None) -> bool
```

EngineCache:

image

An alternative:
* add a parameter to the constructor of ExpireCache to provide a context name: 
    * a table name in the case of SQLite
    * a prefix for the keys in the case of Redis (or whatever)
* remove EngineCache abstraction --> use directly ExpireCache

return42:
> I'm lost by the number of abstractions.

Really? .. Its very simple, there are only two abstractions:

1. `searx.sqlitedb.SQLiteAppl(abc.ABC)` is an abstract class for the development of DB applications. It was introduced when we needed a cache for the favicons, its first concrete implementation at that time was `searx.favicons.cache.FaviconCacheSQLite`.

I have been developing (relational) DB applications (mostly Oracle or MS-SQL / rarely sqlite3) for 25 years, but the requirements are always the same (reduced to a minimum): DB connectors are needed, there are transaction spaces etc. ... and it is further helpful if there is something like “properties” in which the properties of an application are stored (e.g. the date of the last maintenance).

We also have to bear in mind that, for example, transactions and connectors can run concurrently in processes or threads ... and how the sqlite in python deals with connectors and transactions in threads depends on the Python version on the one hand and on the other hand on which sqlite options the sqlite was compiled with when the Python interpreter was compiled ...

So it was obvious for me to “hide” this complexity in an abstract class `searx.sqlitedb.SQLiteAppl(abc.ABC)` when implementing the cache for the favicons.


2. `searx.cache.ExpireCache(abc.ABC)` (implemented in this PR) is the abstraction of a cache that loses its values after a specified **expire** time, similar to what we know from the Redis or ValKey databases.

The intention of this abstraction is that we can connect different DBs behind it, e.g. Redis or sqlite.

In this PR a sqlite connection is implemented, see `class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):` .  You can already **see from the inheritance that it is a combination of 1. and 2.**

Later we can implement a class `ExpireCacheRedis` (@Bnyro has already mentioned it), which would also inherit from `ExpireCache` / but not from `SQLiteAppl` (logically).

I hope to have given a better orientation to the class model .. / otherwise please ask again.

searx/sqlitedb.py
```diff
@@ -168,57 +252,66 @@ def DB(self) -> sqlite3.Connection:
...
         else:
             ver = int(ver)
             if ver != self.DB_SCHEMA:
                 raise sqlite3.DatabaseError("Expected DB schema v%s, DB schema is v%s" % (self.DB_SCHEMA, ver))
             logger.debug("DB_SCHEMA = %s", ver)
 
-    def create_schema(self, conn):
+        return True
+
+    def create_schema(self, conn: sqlite3.Connection):
```
dalf, line 306: Which code use this method?

The code is a way similar to the `init` method of `searx.cache.ExpireCacheSQLite` except the schema version is the secret_key.

searx/cache.py
return42: > Why [`SQLiteAppl.connect`](https://github.com/searxng/searxng/blob/fd33559cfb623c1e9edef7ab08c1e9fbf43824f8/searx/sqlitedb.py#L116-L129) is not [contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) ?

You linked the old implementation, look at the doc-string in this PR:

```
    def connect(self) -> sqlite3.Connection:
        """Creates a new DB connection (:py:obj:`SQLITE_CONNECT_ARGS`).  If not
        already done, the DB schema is set up.  The caller must take care of
        closing the resource.  Alternatively, :py:obj:`SQLiteAppl.DB` can also
        be used (the resource behind `self.DB` is automatically closed when the
        process or thread is terminated).
        """
```

> If there is an exception on line 358, the connection won't be closed by the line 359.

Implement transaction spaces, fetching (SQL) exceptions and closing transaction us up to the caller.  

BTW: with the same argument the context manager (`with`) of the `sqlite3.Connection` does not close the connection .. the context manager only closes a transaction / never a connection --> https://docs.python.org/3/library/sqlite3.html#sqlite3-connection-context-manager

-----

```python
        if table:
            with self.DB:
                self.DB.execute(sql, (key, value, expire, value, expire))
        else:
            with self.connect() as conn:
                conn.execute(sql, (key, value, expire, value, expire))
            conn.close()
```

the establishment of connections is expensive ... this implementation assumes that if values are stored in different tables, the risk of a table lock is very low ... if all values are stored in a single table, the risk of a table lock is higher, which is why a new connection is established in these cases ... since we operate the DB in WAL mode, a table lock is excluded for new connections ... I have already discussed the competing write operations in a previous comment.

dalf: > Implement transaction spaces, fetching (SQL) exceptions and closing transaction us up to the caller.

In this case what is the point of the with statement if SQLiteAppl.connect is not a contextmanager? 

From my view point and if closing / exception etc are up to the caller, instead

```python
            with self.connect() as conn:
                conn.execute(sql, (key, value, expire, value, expire))
            conn.close()
```

it should be

```python
            conn = self.connect()
            conn.execute(sql, (key, value, expire, value, expire))
            conn.close()
```

or more robust

```python
            conn = self.connect()
            try:
                conn.execute(sql, (key, value, expire, value, expire))
            finally:
                conn.close()
```

or more usual SQLiteAppl.connect should be a contextmanager and take care of closing the connection at the end of the with statement. [SQLAchemy](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#opening-and-closing-a-session) or [peewee](https://docs.peewee-orm.com/en/latest/peewee/api.html#Database.__enter__) do that.

return42: > In this case what is the point of the with statement if `SQLiteAppl.connect` is not a contextmanager?

The `SQLiteAppl.connect()` returns a `sqlite3.Connection` / means the context object of the `with` block is an instance of `sqlite3.Connection` .. when the context is closed the exit-method of context object is called -> the DB transaction is finalized (commit).  Since this connection is no longer used, it has to be closed.

> From my view point and if closing / exception etc are up to the caller,

The caller here is the `SQLiteAppl.set(..)` method itself .. not the one who calls `SQLiteAppl.set(..)` .. and this setter method cares about the transaction (by implementing the `with` block) and finally closes the transaction .. so, yes "exception **etc** are up to the caller" .. In case of an exception while the setter method processes, the caller of the setter method tries to set invalid values .. and it is OK to pass this exception to caller of the setter method.

> or more robust

You can implement it in different ways, but your example for "robust" closes the DB connection without a DB commit .. the implicit SQL `BEGIN TRANSACTION` (from the `conn.execute(sql, ..)`) is not closed by a SQL `COMMIT` ..  

----------

Even for me, despite my long experience, it was not easy to understand how SQL is implemented in python in the sqlite3 module, for example I had misunderstood the `autocommit` option when opening a connection or when a transaction begins .. and there were corresponding misassumptions about the transaction spaces in the first implementation, which I am now correcting in this PR.

The sqlite3 documentation is complete and correct in its sense, but I only really understood it after implementing small examples and experimenting with “autocommit”, explicit SQL `commit` and the context managers.

searx/cache.py
return42: > add a parameter to the constructor of ExpireCache to provide a context name

yeah , thats true :+1:  .. is fixed in bbaf3c3dd

return42:
> However here a small benchmark.

On my laptop it varies from 3.5ms up to 4.5ms .. I think values of ms should be OK.

return42:
For those who followed this review, there is an discussion in parallel related to favicons and commit 4a594f1 from this PR:
 
- https://github.com/inetol-infrastructure/searxng-container/issues/5#issuecomment-2848553051

PR #4651, opened 2025-04-22
[mod] include SEARXNG_METHOD environment variable

Aadniz:
## What does this PR do?

This PR includes `SEARXNG_METHOD` environment variable as a way to change the default HTTP search method.

Documentation is included.

It might come in handy for people to be able to specify the environment variable to set the setting, just like the other options that are under the server settings section.

There was also a person on Matrix wondering about documentation for this option, and figured it might be nice to include that.

Run SearXNG with the `SEARXNG_METHOD` set to either `GET` or `POST`. Anything else should result in an error (ValueError: Invalid settings.yml)

### DIFF
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -14,6 +14,7 @@
        limiter: false
        public_instance: false
        image_proxy: false
+       method: "POST"
        default_http_headers:
          X-Content-Type-Options : nosniff
          X-Download-Options : noopen
@@ -49,6 +50,11 @@
 
 ``image_proxy`` : ``$SEARXNG_IMAGE_PROXY``
   Allow your instance of SearXNG of being able to proxy images.  Uses memory space.
+
+.. _method:
+
+``method`` : ``$SEARXNG_METHOD``
+  Whether to use ``GET`` or ``POST`` HTTP method when searching.
 
 .. _HTTP headers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
 
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -101,7 +101,8 @@ server:
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
   # POST queries are more secure as they don't show up in history but may cause
-  # problems when using Firefox containers
+  # problems when using Firefox containers.
+  # Is overwritten by ${SEARXNG_METHOD}
   method: "POST"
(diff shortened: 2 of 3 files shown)


return42 (approved):
Thanks, I hadn't realized that the HTTP method was not documented yet :+1: 

Since all other `server` settings are already controlled via environment variables, there is no reason from my point of view why the HTTP method should not also be controlled via an environment variable.

.. lets hear what other reviewers say ..

Bnyro (approved):
I agree with both of you, this can definitely be handy for quick deployment.

unixfox (approved):

dalf:
Out of topic but related: at point it might be interesting to implement something similar to
https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values

return42:
> interesting to implement something similar to https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values

Please not, its brings a jungle of environment variables and it makes it hard to say at some point *this or that* environment variable does not longer exits or changed it name.

Environment variables are official switches; like an API, they must be version stable.  Simply assigning all switches (additionally) as environment variables to the user restricts the freedom of movement during development and in the long term causes frustration on the part of the developers and on the part of the users, because sooner or later they will change.

Example the switch "foo" is renamed "bar" ... now you not only have to keep the switch "foo" in mind for downward compatibility, but also the environment variable "FOO" ... have fun!

Users usually only need a few basic switches as environment variables (usually those required to run the service).

If we — as often in the past — always make sure to expose everything possible via switches, this complicates development (trapped in the cage of options that must not be violated). 

--> **I advocate for significantly fewer options in favor of a stable product.**

Also OT, but since we're talking about `settings_defaults.py`: the hand-crafted solution for the config [SCHEMA](https://github.com/searxng/searxng/blob/master/searx/settings_defaults.py#L138) should be switched to [msgspec.Struct](https://jcristharif.com/msgspec/structs.html).

Bnyro:
> Please not, its brings a jungle of environment variables and it makes it hard to say at some point this or that environment variable does not longer exits or changed it name.
>
I disagree with this. Automatic naming of environment variables could make it much simpler to configure SEARXNG via env variables and could save us some headaches if we forgot to add an environment variable for a setting.

E.g. if you deploy via docker, it's usually more convenient to configure things via env variables (at least the basics) in order to have everything at a single place, the docker compose file and not spread across multiple different files.

PR #3256, opened 2024-02-29
Implement displaying of answer fields, title, and sources

secondtruth:
## What does this PR do?

This PR implements the ability for plugins to display additional information in answers: fields, title, and sources.

**Visual Example:**
image

**Plugin Code for Example:** https://gist.github.com/secondtruth/ed32c9d3cb898513dc2c6ecaea57bb25

## Why is this change important?

This change allows plugins to display additional information and structured data in a beautiful way, without letting them inject arbitrary HTML into the frontend.

## How to test this PR locally?

as usual

## Author's checklist

nothing

## Related issues

none

Issue #4542, opened 2025-03-25
Norwegian Meteorological Institute


Secret-chest:
**Working URL to the engine**

https://api.met.no

**Why do you want to add this engine?**

Good way to provide weather info as the data is freely licensed (I don't think the server software is fully libre, though). Even though it's Norwegian, it provides data for the whole world.

**Features of this engine**

Has a very extensive API with English documentation.

**How can SearXNG fetch the information from this engine?**

https://api.met.no

**Applicable category of this engine**

A new category, I think, for weather, or in a definition panel.

**Additional context**

- https://github.com/searxng/searxng/pull/1800

Bnyro:
I think we should work on adding a result template for weather engines (as done in https://github.com/searxng/searxng/pull/1800) first before adding any new weather engines.

The more weather engines we add, the harder it will get to migrate all of them to the proper weather result answer type and the matching template.

PR #1800, opened 2022-09-23
Implement weather forcast data types and a rudimentary template

return42:
## What does this PR do?

Implement weather forecast data types and a rudimentary template.

This is a DRAFT in a very early stage .. read along the commit message and read the doc-string in the code.

## Related issues

inspired by #1750

-  [netcdf python library](https://github.com/Unidata/netcdf4-python)

Issue #1352, opened 2022-06-20
Remove hard coded HTML from engine result list.


return42:
- [ ] recoll engine:

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/recoll.py#L31

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/recoll.py#L87-L89

## infobox

Most of the HTML in the *infobox* shown below can be handled by adding a new type to the infobox, a [Description List](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) like `infobox.descriptions` .. but first we need to merge:

- https://github.com/searxng/searxng/pull/1412

- [x] lingva engine / infobox:

  https://github.com/searxng/searxng/blob/5c1c0817c3996c5670a545d05831d234d21e6217/searx/engines/lingva.py#L50-L59

- [ ] sjp engine / infobox:

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/wordnik.py#L59-L67

- [ ] jisho engine: infobox

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/jisho.py#L103-L137

- [ ] sjp engine / infobox:

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/sjp.py#L69

  https://github.com/searxng/searxng/blob/8f7cb4ad6fa3f9dec4b89807a32fa69e0e72bb84/searx/engines/sjp.py#L77-L89

Bnyro:
https://github.com/searxng/searxng/pull/4183 already started with removing hardcoded HTML from all translation engines, so we're on a good way :+1:

PR #4663, opened 2025-04-24
[feat] result types: add weather result answerer and template

Bnyro:
## What does this PR do?
- This PR adds result types and a template for weather engines

- currently every engine hardcode its html

- `!om Berlin`

## Author's checklist
- Duckduckgo Weather and Wttr.in are not migrated yet - they will be migrated as well after the answer type and template are finalized (if there any suggestions for improvements), I don't want to take the efforts to migrate and change them multiple times

## Related issues
superseeds/closes #3256
related: https://github.com/searxng/searxng/issues/4542, https://github.com/searxng/searxng/pull/1800, https://github.com/searxng/searxng/issues/1352

2025-04-24T17:35:37,110507410+02:00
2025-04-24T17:35:45,160382185+02:00


return42:
FYI: last force push was just a rebase on master branch.

return42:
I pushed a WIP commit .. but I haven't yet finished my work .. I set this PR to DRAFT and will inform you, when I have finished my work .. *he must have patience, the young jedi* ;-)

return42:
Haven't yet finished my work, but if you want to have a look at current state .. I added types, i18n/l10n, symbols (icons) and unit conversions (based on what we had in the converter plugin) ..

image

and when selecting search language `en-US` temperature is converted into °F:

image

To see forecast, click on the table (or weather symbol) ..

image

.pylintrc
```diff
@@ -206,7 +206,7 @@ logging-modules=logging
 [MISCELLANEOUS]
 
 # List of note tags to take in consideration, separated by a comma.
-notes=FIXME,XXX,TODO
+notes=FIXME,XXX
```
Bnyro: The pylint changes are probably not related to this PR and should be separated?

searx/weather.py
```diff
@@ -0,0 +1,610 @@
...
+            ExpireCacheCfg(
+                name="WEATHER_DATA_CACHE",
+                MAX_VALUE_LEN=1024 * 200,  # max. 200kB per icon (icons have most often 10-20kB)
+                MAXHOLD_TIME=60 * 60 * 24 * 7 * 4,  # 4 weeks
+            )
+        )
+    return WEATHER_DATA_CACHE
+
+
+def _get_sxng_locale_tag() -> str:
```
Bnyro, line 57: Maybe this should be moved to https://github.com/searxng/searxng/blob/master/searx/locales.py? It's not necessarily coupled to the specific weather engine request and could be useful in other places too.

searx/templates/simple/answer/weather.html
```diff
@@ -0,0 +1,63 @@
...
+        {{ answer.service }}
+      {% endif -%}
+    </div>
+    {{ show_weather_data(answer, answer.current) }}
+  </summary>
+  <div class="weather-forecast">
+    {%- if answer.forecasts -%}
+      <div class="answer-weather-forecasts">
+        {%- for forecast in answer.forecasts -%}
+          <div class="summary">{{ forecast.datetime.l10n(locale=answer.current.location,format="short") }} {{ forecast.summary }}</div>
```
Bnyro:
```suggestion
          <div class="summary">{{ forecast.datetime.l10n(locale=answer.current.location, fmt="short") }} {{ forecast.summary }}</div>
```
This currently leads to a crash, typo :)

searx/templates/simple/answer/weather.html
```diff
@@ -0,0 +1,63 @@
...
+        <td>{{ data.wind_from.l10n(locale=data.location) }}: {{ data.wind_speed.l10n(locale=data.location) }}</td>
+        <td>{{ _("Pressure") }}:</td>
+        <td>{{ data.pressure.l10n(locale=data.location) }}</td>
+      </tr>
+      <tr>
+        <td>{{_("Humidity")}}:</td>
+        <td>{{ data.humidity.l10n(locale=data.location) }}</td>
+        <td></td>
+        <td></td>
+      </tr>
```
Bnyro, line 33: Do we really want to make all these mandatory? I don't think all weather engines have pressure data (e.g. wttr.in I think).

searx/weather.py
```diff
@@ -0,0 +1,610 @@
...
+        self,
+        unit: Units | None = None,
+        locale: babel.Locale | GeoLocation | None = None,
+        template: str = "{value} {unit}",
+        num_pattern: str = "#,##0",
+    ) -> str:
+        if isinstance(locale, GeoLocation):
+            locale = locale.locale()
+        elif locale is None:
+            locale = babel.Locale.parse(_get_sxng_locale_tag(), sep='-')
```
Bnyro, line 324: This could probably be moved to a helper function (or `_get_sxng_locale_tag()`)? It's the same in all `l10n` methods.

client/simple/src/less/weather.less
```diff
@@ -0,0 +1,34 @@
+#answers .weather {
+  summary {
+    display: block;
+  }
+
+  div.summary {
+    margin: 0;
+    padding: 0.5rem 1rem;
+    background-color: var(--color-header-background);
```
Bnyro, line 10: Maybe we should add some rounded borders to the summary to make it fit in better?

```suggestion
    background-color: var(--color-header-background);
    .rounded-corners-tiny;
```
2025-05-25T11:18:27,893244402+02:00

vs

2025-05-25T11:26:27,171415336+02:00

searx/weather.py
```diff
@@ -0,0 +1,610 @@
...
+        location can be determined for the search term, a :py:obj:`ValueError`
+        is thrown.
+        """
+
+        ctx = "weather_geolocation_by_query"
+        cache = get_WEATHER_DATA_CACHE()
+        geo_props = cache.get(search_term, ctx=ctx)
+
+        if not geo_props:
+            geo_props = cls._query_open_meteo(search_term=search_term)
```
Bnyro, line 177: I'm not too sure if we really want to rely on OpenMeteo here, but I don't have a better idea either.

searx/templates/simple/answer/weather.html
```diff
@@ -0,0 +1,63 @@
...
+      <col span="1" class="title">
+      <col span="1" class="measured">
+      <col span="1" class="title">
+      <col span="1" class="measured">
+      <col>
+    </colgroup>
+    <tbody>
+      <tr>
+        <td rowspan="4">
+          {%- if data.url %}<img class="symbol" src="{{ data.url }}" title="{{ data.summary }}">{% endif -%}
```
Bnyro, line 13: I think this should be in the bottom right of the result like for the other answerers and not in the summary line. Also, shouldn't this point you to the web ui of the weather engine with the location already entered instead of just to their docs? I think it's not really helpful for the user this way.

searx/weather.py
```diff
@@ -0,0 +1,610 @@
...
+    ctx = "weather_symbol_url"
+    cache = get_WEATHER_DATA_CACHE()
+    origin_url = f"{YR_WEATHER_SYMBOL_URL}/{fname}.svg"
+
+    data_url = cache.get(origin_url, ctx=ctx)
+    if data_url is not None:
+        return data_url
+
+    response = network.get(origin_url, timeout=3)
+    if response and response.status_code == 200:
```
Bnyro:
```suggestion
    if response.status_code == 200:
```
I don't think `response` can ever be `None`?

Bnyro (commented):
Looks great already, thanks for your work!

I've left some comments / suggestions / questions.

searx/weather.py
return42: We have several places in the code where we try to determine the "ideal" language selection for the client. All of these implementations still differ in detail today. This mess will have to be cleaned up one day, but until then, it's more reliable to implement this method here rather than, for example, this one (which does something similar).

https://github.com/searxng/searxng/blob/848c8d0544eb0138da4414a750f79b22883f91b2/searx/webapp.py#L161-L164

I'm working on a parallel branch that cleans up the properties and also takes this problem into account to find a general solution.

searx/templates/simple/answer/weather.html
return42: > Do we really want to make all these mandatory? 

I would leave the fields mandatory for now. If we switch to the other weather engines, we can always decide to relax this requirement.

However, every weather engine will have to meet a few minimum requirements, and I wouldn't want to decide which ones at this point (without having looked at all the other engines in depth).

searx/weather.py
return42: In the currently implemented six classes, the default locale is determined from the query. However, we may eventually want to implement classes that determine their default locale solely from the UI property (where the language of the query is irrelevant). We can implement an additional helper function, as you suggested, but that would only be a wrapper that would then have to call [_get_sxng_locale_tag](https://github.com/searxng/searxng/pull/4663#discussion_r2106152038). That would be a few fewer lines of code, but I wasn't sure if it was worth it. After all, it would be another function.

If you find an additional helper function helpful, we can implement it.

searx/weather.py
return42: This isn't set in stone either: if we have alternative sources, we can implement them alongside the `_query_open_meteo` method. Currently, however, I don't know of any alternatives, and open-meto responds quickly and robustly. Since we also cache the responses, this should provide us with a good initial solution.

return42:
> I've left some comments / suggestions / questions.

I have adopted some changes, see [diff](https://github.com/searxng/searxng/compare/c64636fba53c1f49270147eb08a075f59d7bbe06..91cbe67c14509b3b3635511f58409e1191540d13) .. and I have left a comment with the others.

Switched PR from DRAFT to "ready for review" ..

searx/weather.py
```diff
@@ -0,0 +1,607 @@
...
+
+
+def symbol_url(condition: WeatherConditionType) -> str | None:
+    """Returns ``data:`` URL for the weather condition symbol or ``None`` if
+    the condition is not of type :py:obj:`WeatherConditionType`.
+
+    If symbol (SVG) is not already in the :py:obj:`WEATHER_DATA_CACHE` its
+    fetched from https://github.com/nrkno/yr-weather-symbols
+    """
+    # Symbols for darkmode/lightmode? .. and day/night symnbols? .. for latter
```
Bnyro:
```suggestion
    # Symbols for darkmode/lightmode? .. and day/night symnbols? .. for later
```

searx/weather.py
```diff
@@ -0,0 +1,607 @@
...
+    # we need a geopint (critical in sense of privacy)
```
Bnyro: What's a "geopint"?

return42: its a pint of beer waiting for you at a geopoint :-)

typo: geopint --> geopoint .. :facepalm: .. will fix it.

client/simple/src/less/weather.less
```diff
@@ -0,0 +1,37 @@
...
+    .rounded-corners-tiny;
```
Bnyro, line 11: Currently the arrow to open the weather forecast is misaligned.

2025-05-31T19:13:52,585033174+02:00

searx/weather.py
Bnyro: > If you find an additional helper function helpful, we can implement it.
>
You're right, it probably doesn't make too much sense looking at the code again. I've only seen that there's duplicated code and didn't really look deeper into why it's duplicated here :)

Bnyro:
Looks awesome to me after all, thanks for all the efforts you've put into this!
My initial work on this has been about 200 lines maybe, it's been improving a lot since then :)

I already thought about porting the other weather engines to using the new API but I think that would be better suited for a new PR, so we should merge this here first.

client/simple/src/less/weather.less
Bnyro: Okay, we can fix it by replacing the `<div class="summary">` with `<span class="summary">`, that's the simplest fix :+1:

searx/weather.py
return42: Its not the word "later" in sense of *time* ... its the word "latter" in sense of *the latest* in the list given before

Bnyro:
```suggestion
    # Symbols for darkmode/lightmode? .. and day/night symbols? .. for the latter
```
:)

client/simple/src/less/weather.less
return42: Your first suggestion "list-style: none" works better .. for now .. the `<summary>` element is the simplest solution... it's a bit impractical because you can only expand/collapse it in the header. We should perhaps turn it into a horizontal (left/right) slider in a future PR.

Bnyro:
There it is, finally :rocket:

PR #4678, opened 2025-04-28
[feat] tracker url plugin: use ClearURL tracking param list

### LINKED ISSUE #4472: [poll] do we need plugin "Tracker URL remover" any longer?

Bnyro:
## What does this PR do?
- this PR implements a (for us) automated way of having up to date tracking rules using ClearURLs rules

- see #4472

- enable the tracker url remover plugin and do a search for something with tracker params
- `dan brown amazon`  has some for example
- compare result urls to the urls when the plugin is disabled

## Author's checklist
- we could also add other sources to `update_tracker_patterns.py` and merge the tracking rules, e.g. from AdGuard, with the existing ones
- The format used is quite simple and thus should be usable with other sources too (if we parse them to this format) 

## Related issues
closes #4472


return42:
Cool work / can you rebase on master branch to solve the conflict .. thanks!

searx/plugins/tracker_url_remover.py
```diff
@@ -39,20 +36,53 @@ def __init__(self, plg_cfg: "PluginCfg") -> None:
...
         )
 
-    def on_result(
-        self, request: "SXNG_Request", search: "SearchWithPlugins", result: Result
-    ) -> bool:  # pylint: disable=unused-argument
-        if not result.parsed_url:
+    def on_result(self, request: "SXNG_Request", search: "SearchWithPlugins", result: Result) -> bool:
+
+        result.filter_urls(self.filter_url_field)
+        return True
```
return42: To remove tracker arguments from *all* URLs (including links in the infobox and other areas) we have to use the [`Result.filter_urls(..)`](https://docs.searxng.org/dev/result_types/base_result.html#searx.result_types._base.Result.filter_urls) method / its the same procedure we have in the hostname-replace plugin.

searxng_extra/update/update_tracker_patterns.py
```diff
@@ -0,0 +1,36 @@
...
+    providers = resp.json()["providers"]
+    rules = []
+    for rule in providers.values():
+        rules.append(
+            {
+                "urlPattern": rule["urlPattern"].replace("\\\\", "\\"),  # fix javascript regex syntax
+                "exceptions": [exc.replace("\\\\", "\\") for exc in rule["exceptions"]],
+                "trackerParams": rule["rules"],
+            }
+        )
```
return42: When PR https://github.com/searxng/searxng/pull/4836 has been merged we have a boilerplate to manage `data` in a *ExpireCache*.

.. and instead of the json file which needs to be created today, we can directly read the rules from `CLEAR_LIST_URL` in the *ExpireCahe* .. 

ping @dalf: Here's an example where SearXNG uses `data` from external projects. To ensure efficient access to the `data` objects, these objects are loaded into an (expiring) cache.  This cache is not yet implemented in this PR, but its related to what we discuss in #4836.

return42:
> When PR https://github.com/searxng/searxng/pull/4836 has been merged we have a boilerplate to manage data in a ExpireCache.

I merged PR 4836, rebased your branch and pushed a commit on top.  In this commit the tracker data is loaded directly into the cache, eliminating the need to maintain the JSON file and the implementation requires less lines of code / is easier to read ...

- https://github.com/searxng/searxng/pull/4678/files

Bnyro:
> I merged PR 4836, rebased your branch and pushed a commit on top. In this commit the tracker data is loaded directly into the cache, eliminating the need to maintain the JSON file and the implementation requires less lines of code / is easier to read ...
>
Isn't it significantly slower to load all the data from the sql database on the disk for each result instead of just loading the whole db into memory once and re-using it? Or are we doing memory caching somewhere? I find copying the JSON files to a database a bit confusing to be honest :thinking:

return42:
> Isn't it significantly slower to load all the data from the sql database on the disk for each result instead of just loading the whole db into memory once and re-using it?

Yes in memory is always faster compared to query a DB .. but loading all datasets from `searx/data/*.json` does consume a lot of the memory today.

- https://github.com/searxng/searxng/discussions/1892#discussioncomment-13237671

Remember: when you have 4 workers the datasets will be loaded 4 times in you memory.

When caching datasets in a (SQL) DB .. each of the 4 workers has one DB connect to one and the same DB .. a DB management system does load the file / only the result of a query is temporarily consuming MEM.

Bnyro:
> When caching datasets in a (SQL) DB .. each of the 4 workers has one DB connect to one and the same DB .. a DB management system does load the file / only the result of a query is temporarily consuming MEM.
>
Yes, but even though we only temporarily load the data now, we load the whole pattern database for each single result that we need to check? So in worst case, this could be more than 100 disk loads of the pattern database for a single user request (if many engines are enabled)?

When I enable the plugin and do multiple subsequent request by spamming the search button, the process loads and loads stuff into memory (ram usage increases), but the memory doesn't actually seem to become de-allocated. (can't tell if that's because of this PR or an other sxng issue though)

searx/settings.yml
```diff
@@ -245,7 +245,7 @@ plugins:
     active: false
 
   searx.plugins.tracker_url_remover.SXNGPlugin:
-    active: false
```
unixfox, line 248: I think this shouldn't be by default because this may have potential side effects that we do not know.

And we do not maintain the list in "CLEAR_LIST_URL".

It's like hosting a pihole or adguard home, you start to add multiple sources but then some websites starts to break or "half break".

The user should be aware of such feature by turning it on by themselves.

unixfox (changes requested):

searx/settings.yml
```diff
@@ -245,7 +245,7 @@ plugins:
...
+    active: true
```
unixfox, line 248: I think this shouldn't be by default because this may have potential side effects that we do not know.

And we do not maintain the list in "CLEAR_LIST_URL".

It's like hosting a pihole or adguard home, you start to add multiple sources but then some websites starts to break or "half break".

The user should be aware of such feature by turning it on by themselves.

unixfox (changes requested):

searx/settings.yml
return42: > The user should be aware of such a feature by turning it on by themselves.

Under our promise "Users are neither tracked nor profiled," I, as a user, would assume that all links offered to me are also cleared of trackers.

> And we do not maintain the list in "CLEAR_LIST_URL."

If problems arise, the user can disable URL cleanup.

The lists come from this project: https://docs.clearurls.xyz/1.27.3/ . It can be assumed that they are designed to be conservative; otherwise, there would be more issue reports in the project.

unixfox: Yes, this plugin is great for privacy. But I would be in more confidence to enable it by default if that was a list maintained by us or a fixed algorithm written by us.

But that's not the case. I would rather by default present to the user the original unmodified link than having some weird edge cases where the link is broken due to this plugin. Privacy is important, but what's more important is usability. The majority of people use a software because it works great for them and privacy is a nice perk for them. If the software partially works, they are not going to use it, even though it is great for privacy.

Since, this plugin can be easily activated in the settings page, I do not see the point of enabling it by default.

unixfox: > It can be assumed that they are designed to be conservative; otherwise, there would be more issue reports in the project.

Look at the issues in https://github.com/ClearURLs/Rules/issues. I found one issue about those "edge cases" that could happen in SearXNG:
- https://github.com/ClearURLs/Rules/issues/129

Not saying that we are going to encounter a lot of issues like these. But on a large scale of users, the likeliness is going to be higher. That's why it's better to be opt-in, so that the user is aware of such possible breakage.

And this list was primarily used for the ClearURLs extension, not for SearXNG. I already had some bad experience in the past by trying to use an adblock list URLs in pihole, that lead to some website not functioning anymore because the regex was designed for adblock not pihole. Adblock acts on the browser level, whereas pihole acts on the DNS level.

return42: It is only logical that external data can also contain errors. There are errors here and in many other places where we get data from external sources (and as an aggregator, all we do is collect and process external data).

There's no right or wrong answer to the question of whether this plugin should be enabled by default or not. I'd just like to see in practice how it behaves overall if we enable it by default.

If we later realize that this was a bad decision because there are too many errors, then it's easy to set the default to "off"... at least then we know for sure why "off" is the better choice by default.

#### TLDR: *Let's just try it out and make fewer assumptions.*  ..

unixfox: The right path is to let people try a new feature in a testing phase and then enable it after if the general public wants to have it enabled by default.

Not the reverse. Otherwise, you impact everyone, which is not a good choice when testing a new feature.

return42: I give up asking for a little more relaxation and hope that my conversation partner is willing to follow the recommendations or requests of others.

unixfox:
Good for me, except for the fact that the plugin is enabled by default like explained in https://github.com/searxng/searxng/pull/4678/files#r2106208778.

I can advertise the improved tracking URL removal feature on reddit and matrix, and asking people to test it. Thanks.

searx/settings.yml
Bnyro: > And this list was primarily used for the ClearURLs extension, not for SearXNG. I already had some bad experience in the past by trying to use an adblock list URLs in pihole, that lead to some website not functioning anymore because the regex was designed for adblock not pihole. Adblock acts on the browser level, whereas pihole acts on the DNS level.
>
ClearURL is web based, so we don't have such an issue.

> There's no right or wrong answer to the question of whether this plugin should be enabled by default or not. I'd just like to see in practice how it behaves overall if we enable it by default.
>
I think it's a good compromise to enable it by default and disable it if there are many complaints, we'll do it like that 👍

unixfox: Marking my veto on this. I can provide some efforts in publicly sharing this improvement for asking people to test it. But I do not want this feature to be on by default, for all the reasons explained above.

Bnyro: Well, we have a 2:1 vote here and you didn't seem to care about our vetos in other PRs either, so we'll do it as we said.

unixfox:
Converting this PR in draft until all the team can find an agreement.

return42:
> Converting this PR in draft until all the team can find an agreement.

What's this nonsense for? Come off your dictator trip.

dalf:
What happen if I say this feature should be off by default?
The universe breaks down?
We decide using chifoumi? Street Fighter?  DOTA? a coin? the Schrödingers' cat?

Bnyro:
> What happen if I say this feature should be off by default?
The universe breaks down?
We decide using chifoumi? Street Fighter? DOTA? a coin? the Schrödingers' cat?
>
You could say the same about turning it on by default.

Markus and I spent the time implementing this feature, so we should ultimately be the most qualified persons to tell if the feature is reliable enough to turn it on by default or not. And the answer is: yes, it's reliable enough.

dalf:
> so we should ultimately be the most qualified persons to tell if the feature is reliable enough to turn it on by default or not. And the answer is: yes, it's reliable enough.

I don't know what is happening with this boolean value, what is really at stake? That's really tense.

It reminds me an issue five years ago: https://github.com/searx/searx/issues/2025

unixfox:
@dalf all the discussions are here: https://github.com/searxng/searxng/pull/4678#discussion_r2106208778

return42:
> I don't know what is happening with this boolean value, what is really at stake? That's really tense.

The URLs are cleaned of trackers... in the process a URL could be broken... there is no evidence that a URL is broken, all tests so far have been positive.

> The universe breaks down?

Yes, that exactly the risk .. in the eyes of one.

dalf:
> What's this nonsense for? Come off your dictator trip.

> Yes, that exactly the risk .. in the eyes of one.

> Well, we have a 2:1 vote here and you didn't seem to care about our vetos in other PRs either, so we'll do it as we said.

Looking at just this PR, from outside eyes, all sound like as you said : one maintainer strongly disagree on a default settings. 

Regarding the number of things that has been accepted from your contribution @return42 that are way off [this comment](https://github.com/searx/searx/issues/2025#issuecomment-649572285), the number of time @mrpaulblack made the diplomat to the point he don't want to do anymore, the number of time I tried to round the corner and find compromise, your comments that's sound like more about assertion of power and role reversal. Your tone is way off the line.

Bnyro:
We haven't agreed on https://github.com/searxng/searxng-docker/pull/384 either and it was merged anyways (2 vetos).

Here it's one person claiming that the plugin would cause issues without any proof, so I don't see why that shouls block anything seeing that two vetos are not enough either.

SuperSandro2000:
How can I prevent the plugin from initing? This broke our some tests in NixOS because we can no longer run searx without an active internet connection.

Bnyro:
> How can I prevent the plugin from initing? This broke our some tests in NixOS because we can no longer run searx without an active internet connection.
>
I think you have to patch out https://github.com/searxng/searxng/blob/master/searx/settings.yml#L246-L247.

@return42 Didn't we want to add a way to disable plugins globally? That's probably gotten lost during the typification refactor? Because as we're merging the default `settings.yml` plugins config with the instance admin's config, there's no way to remove a plugin once it's been added to the base `settings.yml`, like setting `disabled: true`?

return42:
> Didn't we want to add a way to disable plugins globally? .. Because as we're merging the default ..

Sure .. I missed it :facepalm: .. fix is on the way:

- https://github.com/searxng/searxng/pull/5019

PR #4696, opened 2025-04-30
[fix] botdetection: check if the browser supports Sec-Fetch headers

unixfox:
Related to https://github.com/searxng/searxng/pull/3965.

Check that the browsers checked do support the feature Sec-Fetch before blocking them based on this factor.

Avoiding blocking legitimate users using ancient browsers that do not support these headers.

Reasoning explained: https://github.com/searxng/searxng/pull/3965#issuecomment-2441426436

### DIFF
--- a/searx/botdetection/http_sec_fetch.py
+++ b/searx/botdetection/http_sec_fetch.py
@@ -28,6 +28,7 @@ from ipaddress import (
     IPv6Network,
 )
 
+import re
 import flask
 import werkzeug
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
I think that some bots use Firefox user agents, so this should be helpful for such.

I'm not sure if the general impact is as good as with @return42's PR because many bots will probably use other user agents than the common ones (or rotate different ones) - we'll have to see how it performs in production.

unixfox:
@return42 could you try on your searxng instance please :)?

unixfox:
@return42 did you see any change in efficiency between PR https://github.com/searxng/searxng/pull/3965 and https://github.com/searxng/searxng/pull/4696 on your server?

return42:
@unixfox please give me some more time .. 

> I'm not sure if the general impact is as good as with @return42's PR because many bots will probably use other user agents than the common ones (or rotate different ones) - we'll have to see how it performs in production.

I want to implement an option for the SearXNG admins to switch additional browser tests from 399a31f473 on or off ...

unixfox:
What do you mean? I'm not following you.

All the browsers in https://github.com/searxng/searxng/pull/4696/files#diff-ab6d5589d48207e72a940dbaa37d671093e2e1e7bc9849faed4e3163ed72a1e7R41-R76 match all the browsers that exist and support this Sec-Fetch headers. Firefox, Chrome, Safari.

return42:
Sorry, I quoted the wrong paragraph :facepalm: .. I fixed my comment above to:

> I'm not sure if the general impact is as good as with @return42's PR because many bots will probably use other user agents than the common ones (or rotate different ones) - we'll have to see how it performs in production.

unixfox:
> I want to implement an option for the SearXNG admins to switch additional browser tests from [399a31f](https://github.com/searxng/searxng/commit/399a31f47376f2cf236346b2233bb0e30b52bf00) on or off ...

Again I'm not following what you want to do. This is not clear whenever you want to implement an option to enable the feature "HTTP Fetch Metadata Request Headers" from PR #4696 or #3965. Or give the ability to add more browsers to check against "HTTP Fetch Metadata Request Headers".

If it's the latter one: why? Only the browsers checked in https://github.com/searxng/searxng/pull/4696/files#diff-ab6d5589d48207e72a940dbaa37d671093e2e1e7bc9849faed4e3163ed72a1e7R41-R76, do support Sec-Fetch. Why giving the ability to configure more? If there are more later on, then we can add them in the code.

return42:
> many bots will probably use other user agents

The PR https://github.com/searxng/searxng/pull/3965 doesn't care UA header and blocks when Sec header is missed, while this PR looks at UA header and blocks only when UA header and Sec header do not fit together .. I want to implement an option, so admin can decide if he wants to care UA header or not.

unixfox:
> I want to implement an option, so admin can decide if he wants to care UA header or not.

I'm sorry I do not want this option. Here is why:

I do not want all the public instances of SearXNG to only work on these browsers:

- Chrome >= 80
- Firefox >= 90
- Safari >= 16.4

Feel free to do that in your fork, on your instance, but I do not want such thing in the general SearXNG core.

Even if this is going to not be a default option, public instances owners will enable it because they want the best protection, and they will not understand the damages this is going to do.

My PR as it is a good balance to block based on one more factor (that remember, can always be found by reading the source code) and avoid blocking legitimate users.

So please, test it on your instance, give me some feedback for the efficiency in blocking bots (not legitimate users, if that can be analyzed).

return42:
> Even if this is going to not be a default option, public instances owners will enable it because they want the best protection, and they will not understand the damages this is going to do.

There will be documentation that clearly describes what the disadvantages can be.

unixfox:
I'm reiterating what I said, I'm not adding such feature. The built-in bot protection was created to reduce bots while fully knowing that anyone can bypass it by reading the source code, but I'm not turning it into a tool for restricting the access to a class of browsers family.

There are a myriad of other factors to catch bots, this PR is not the only ultimate factor to block bots or reduce the effects from bots:
- http2: https://github.com/searxng/searxng/issues/4709
- ja4h: https://github.com/searxng/searxng/issues/4710
- captcha solver: https://github.com/searxng/searxng/issues/2844
- other ideas for blocking based on non-js usage: https://github.com/TecharoHQ/anubis/issues/95

By the end of next week, I'm merging this PR as is, feel free to give me some feedback in case you test it.

return42 (commented):
> Improved version of https://github.com/searxng/searxng/pull/3965 which check for the browser version before checking the Sec-Fetch headers.

In my opinion, this PR is a weakening of the protection we got through PR #3965, if you think this weakening is necessary, then (rebase and) merge this PR.

return42:
> I think that some bots use Firefox user agents, so this should be helpful for such.
> 
> I'm not sure if the general impact is as good as with @return42's PR because many bots will probably use other user agents than the common ones (or rotate different ones) - we'll have to see how it performs in production.

See my DRAFT:

- https://github.com/searxng/searxng/pull/4807

PR #4699, opened 2025-05-01
Rework Dockerfile

### LINKED ISSUE #4460: Simpler Dockerfile

inetol:
## What does this PR do?

This is one of various PR to refactor the entire SearXNG Docker workflow.

Switches to Python glibc based images, all dependencies are installed via pip and not from system repositories, and several minor changes.

~~By default, the container will now run under root, this is to prevent weird issues when using a rootless Docker/Podman setup (you can revert to how it was by specifying the `searxng` user on container instancing, just remember to update the ownership recursively in the already created volumes).~~

This PR will increase the image size from 194.9 MB to 345.47 MB (amd64), this is due to ARMv7 images (needs dependencies for wheels compilation and runtime (?)) and uWSGI webserver. Later PR will reduce the final image size.

It's being talked about in various issues and PRs as Docker images are getting in the way and hindering development.

#4676
#4596
#4463

`docker build .` or `podman build .`

## Related issues

Supersedes / Close #4463

Closes https://github.com/searxng/searxng/issues/4460

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -43,15 +43,7 @@ do
     esac
 done
 
-get_searxng_version(){
-    su searxng -c \
-       'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' \
-       2>/dev/null
-}
-
-SEARXNG_VERSION="$(get_searxng_version)"
-export SEARXNG_VERSION
-echo "SearXNG version ${SEARXNG_VERSION}"
+echo "SearXNG version $SEARXNG_VERSION"
 
 # helpers to update the configuration files
 patch_uwsgi_settings() {
@@ -76,7 +68,7 @@ patch_searxng_settings() {
         -e "s|base_url: false|base_url: ${BASE_URL}|g" \
         -e "s/instance_name: \"SearXNG\"/instance_name: \"${INSTANCE_NAME}\"/g" \
         -e "s/autocomplete: \"\"/autocomplete: \"${AUTOCOMPLETE}\"/g" \
-        -e "s/ultrasecretkey/$(openssl rand -hex 32)/g" \
+        -e "s/ultrasecretkey/$(head -c 24 /dev/urandom | base64 | tr -dc 'a-zA-Z0-9')/g" \
         "${CONF}"
 
     # Morty configuration
@@ -172,4 +164,4 @@ printf 'Listen on %s\n' "${BIND_ADDRESS}"
 
 # Start uwsgi
 # TODO: "--http-socket" will be removed in the future (see uwsgi.ini.new config file): https://github.com/searxng/searxng/pull/4578
-exec uwsgi --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
+exec /usr/local/searxng/venv/bin/uwsgi --http-socket "${BIND_ADDRESS}" "${UWSGI_SETTINGS_PATH}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,92 +1,100 @@
-FROM alpine:3.20
-ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
-EXPOSE 8080
-VOLUME /etc/searxng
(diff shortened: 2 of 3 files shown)


unixfox:
> By default, the container will now run under root

hum no, we should run a non-root container by default.

inetol:
I have reverted the change, there is no need to specify uid/gid in uWSGI as it will already run under the user that is configured to run in the container (in this case now `searxng`)

unixfox:
> This PR will increase the image size from 194.9 MB to 345.47 MB (amd64), this is due to ARMv7 images (needs dependencies for wheels compilation and runtime (?)) and uWSGI webserver. Later PR will reduce the final image size.

Maybe we could do a multi-stage Docker image? This way we don't have to include the whole build-essential.

Or separate Dockerfiles like I proposed in https://github.com/searxng/searxng/issues/4676#issuecomment-2840903079

And have you looked into adding more source for the wheels like https://www.piwheels.org/? I just found out that this wheels repo is configured in my raspberry pi (I don't use searxng on it but I saw that [when installing cryptography](https://github.com/searxng/searxng/issues/4676#issuecomment-2840237949) on armv7)

inetol:
> Maybe we could do a multi-stage Docker image? This way we don't have to include the whole build-essential.

It's already being done, the issue is that uWSGI relies on some *heavy* dependencies, I hope to be able to remove that once we use Granian:

https://github.com/inetol/searxng/blob/4aaa97a1d80d6267685e0291ad270875393cdb35/Dockerfile#L44

Also in https://github.com/searxng/searxng/pull/4707 I will separate the "modern" archs and "armv7", so less dependencies for amd64 and arm64 images.


> Or separate Dockerfiles like I proposed in [#4676 (comment)](https://github.com/searxng/searxng/issues/4676#issuecomment-2840903079)

I want to do this here, as this requires modifying the whole CD process: https://github.com/searxng/searxng/pull/4707


> And have you looked into adding more source for the wheels like https://www.piwheels.org/? I just found out that this wheels repo is configured in my raspberry pi (I don't use searxng on it but I saw that [when installing cryptography](https://github.com/searxng/searxng/issues/4676#issuecomment-2840237949) on armv7)

I find it interesting, unfortunately it follows the mainline Python versions of RPiOS. The container images use Python 3.13 and piwheels are built for Python 3.11 backwards, I'm not going to stick to using Python 3.11 on the container image if it only affects one architecture (although once the whole Docker process is refactored, I don't rule out that the arch armv7 will be keyed to use the same mainline Python versions of RPiOS, we'll see...).

unixfox (approved):
Ok good for me.

We can continue the work in reducing the image size in https://github.com/searxng/searxng/pull/4707. Thanks

dockerfiles/docker-entrypoint.sh
```diff
@@ -43,15 +43,7 @@ do
...
-get_searxng_version(){
-    su searxng -c \
-       'python3 -c "import six; import searx.version; six.print_(searx.version.VERSION_STRING)"' \
-       2>/dev/null
-}
-
-SEARXNG_VERSION="$(get_searxng_version)"
-export SEARXNG_VERSION
-echo "SearXNG version ${SEARXNG_VERSION}"
+echo "SearXNG version $SEARXNG_VERSION"
```
unixfox, line 46: How does it work now that get_searxng_version is gone?

inetol: It is set directly in the ENV now, much more practical.

https://github.com/searxng/searxng/pull/4699/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R83

PR #4707, opened 2025-05-02
[mod] rework container CI/CD

inetol:
## What does this PR do?

This is one of various PR to refactor the entire SearXNG Docker workflow.

Refactors `integration.yml` and moves container related things into `container.yml`, also touches needed updates into script files.

* Podman is now supported to build the container images (Docker also received a refactor, merging both build and buildx)
* Bringing massive reduction on runtime across all CI/CD workflows.
* Container images are being built by Buildah instead of Docker BuildKit.
* Container images are being tested before release.
* Splitting "modern" (amd64 & arm64) and "legacy" (armv7) arches on different Dockerfiles to reduce dependency burden and allowing future image optimizations.

Same as #4699

`make container.build`

## Author's checklist

We can't verify locally that the CI process works as expected, especially the container image release process. Due to the changes being relatively large, it's likely that some workflow will not work correctly the first time.

I try as much as possible to verify that everything works individually, but no promises, CI stuff.


inetol:
Pending on #4699 merge to continue testing

return42:
> Due to my limited access to the org .. I try as much as possible to verify that everything works individually, but no promises, CI stuff.

Thank you very much for the work you put in here ... if you get stuck at some point, don't hesitate to ask the core team for support.  I myself am weak on the topics of GH actions and docker ... but there are others in our team and as far as I know they are very familiar with them ... and can also decide whether we should grant you additional rights on the GH repository.

unixfox:
Just would like to point out that every commit creates a notification on our side. It would maybe be interesting to test in a fork on your side. Thanks

inetol:
> Just would like to point out that every commit creates a notification on our side. It would maybe be interesting to test in a fork on your side. Thanks

Sorry about that, I will try to move the tests directly to the fork, but I don't know if the CI will work as expected.

inetol:
There should be no more major changes to this PR unless requested. I have updated the PR information with the most significant changes https://github.com/searxng/searxng/pull/4707#issue-3035843677

Take your time to read the diff and request changes.

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+                echo "git_url=$GIT_URL"
+                echo "git_branch=$GIT_BRANCH"
+            } >>"$GITHUB_OUTPUT"
+        fi
+    )
+    dump_return $?
+}
+
+container.test() {
+    if ! "$GITHUB_ACTIONS"; then
```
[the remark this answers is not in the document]
inetol: 27fc1ea08808e988f5ff8a942742f4079c71fb7a

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+            variants+=("")
+            platforms+=("linux/${archs[-1]}")
+            ;;
+        "ARMV7" | "armv7" | "armhf" | "arm")
+            archs+=("arm")
+            variants+=("v7")
+            platforms+=("linux/${archs[-1]}/${variants[-1]}")
+            ;;
+        *)
+            err_msg "Unsupported architecture; (ARCH=\"${arch[-1]}\")"
```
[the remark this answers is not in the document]
inetol: 87d5000c9463611e99b93718bc09e6f92724a17a

manage
```diff
@@ -339,22 +262,25 @@ main() {
...
+        ;;
+    --help) help ;;
+    --*)
+        help
+        err_msg "unknown option $cmd"
+        return 42
+        ;;
+    *)
+        _type="$(type -t "$cmd")"
+        if [ "$_type" != 'function' ]; then
```
return42, line 277: I think these modifications here are only whitespaces .. may related to your editor settings.  Some editors interpret one tab as 4 spaces (a tab is by definition 8 spaces). To avoid tab problem, we don't use tab for indentation (except in Makefiles, where a tab is needed).

For shell files we use 4 spaces for one indentation:

https://github.com/searxng/searxng/blob/bc06b1aece43c50fa0dae7e6d90389814bebbc91/.editorconfig#L5-L7

Can you please check your editor settings.

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+container.help() {
+    cat <<EOF
+container.:
```
return42, line 6: This new shell script should be added to the `shellcheck` in the Makefile. You can test your modification on shell scripts by `make test.shell`

https://github.com/searxng/searxng/blob/bc06b1aece43c50fa0dae7e6d90389814bebbc91/Makefile#L58-L71

manage
```diff
@@ -226,24 +147,25 @@ gecko.driver() {
...
-            "windows 64 bit") ARCH="win64";;
-            "mac 64bit") ARCH="macos";;
+        "linux 32bit" | "linux2 32bit") ARCH="linux32" ;;
+        "linux 64bit" | "linux2 64bit") ARCH="linux64" ;;
+        "windows 32 bit") ARCH="win32" ;;
+        "windows 64 bit") ARCH="win64" ;;
+        "mac 64bit") ARCH="macos" ;;
         esac
-        GECKODRIVER_URL="https://github.com/mozilla/geckodriver/releases/download/$GECKODRIVER_VERSION/geckodriver-$GECKODRIVER_VERSION-$ARCH.tar.gz";
+        GECKODRIVER_URL="https://github.com/mozilla/geckodriver/releases/download/$GECKODRIVER_VERSION/geckodriver-$GECKODRIVER_VERSION-$ARCH.tar.gz"
```
return42, line 169: :thinking: do you use a code optimizer tool / code formating tool ? --> except the removed docker function all other modifications in this script are not needed.

If you know a code formating tool for shell scripts, and you think its worth to add to our tool chain, then please open a new PR for such a task.

return42:
The PR brings many valuable changes, but it also combines too many changes.  It will therefore be difficult for any reviewer to approve this PR in its entirety.

I would suggest creating smaller PRs that cherry pick topics from this DRAFT.  I would therefore suggest to create 

- one PR for the build of the online doc (`documentation.yml`)
- one PR for translation tasks (`l10n.yml`)
- one PR for docker build tasks (the name `container.yml` does not fit,   since it is not about containers in general but docker in particular)
- ...

return42:
TL;DR / the follwing rules should be in mind, when creating a commit:

- **commit history should be read like a history book**
- **commit messages are for the reader not for the author of the commit**

----

> There should be no more major changes to this PR unless requested.

This PR has 57 commits .. but not a single one is related to a separate topic ... some commit messages have titles like `.` or `..` ... strictly speaking they are not commits but saves of files (like: *"save my files to the repo"*).

A commit should always have a context and the commit message describes what is to be changed in that context, just as a function description should describe what the intention and the goal of the function is, a commit message should describe what the intention and the goal of that commit is.

- Choose meaningful commit messages](https://docs.searxng.org/dev/contribution_guide.html#code)

The commits are not used to save files.

The commit messages form the history and are the first and therefore most important information a developer has when he has to research when and why a change had to be made and how it was made (what the goal was). 

Like any text, a commit message should be written for the reader and not from the perspective of the author.

When scrolling through the history, the first thing one see is the title of the commit message. Therefore the title should describe the change as briefly and precisely as possible ... followed by a blank line and then a somewhat detailed description of the change.

----

Since the 57 commits are of no further significance, they could be combined into one commit ... or you could take the trouble to divide them into topics, but since we probably have to make new (individual) PRs for the topics anyway, we can improve the commit history even then.

manage
inetol: The editor forces me to use the style defined in .editorconfig and indeed those are spaces and not tabs.

What I do use is https://github.com/mvdan/sh and it seems that by default it flattens the case branches, that's why the diff

utils/lib_sxng_container.sh
inetol: It is already covered by the glob, isn't it?  `utils/lib_sxng*.sh`

manage
inetol: Should I revert those style change for now? I'll open a PR later with the tool.

inetol:
> This PR has 57 commits .. but not a single one is related to a separate topic ... some commit messages have titles like `.` or `..` ... strictly speaking they are not commits but saves of files (like: _"save my files to the repo"_).

Yeah, most of them come from another branch of this fork. I'll squash and document them now.

manage
return42: > Should I revert those style change for now? 

Yes, it is better to not have this modification in this PR / they are not needed to achieve the goal of this PR.

> I'll open a PR later with the tool.

:+1:

.github/workflows/checker.yml
```diff
@@ -1,31 +1,46 @@
...
+  contents: read
+
+env:
+  PYTHON_VERSION: "3.13"
 
 jobs:
-  checker:
-    name: Checker
-    runs-on: ubuntu-24.04
+  search:
```
dalf, line 21: The name of the internal is checker, search is going to be confusing

dalf, line 21: Note: 
* the Python module name should be `searx.checker` rather than `searx.search.checker`.
* for a time it made sense to embed this tool in the Flask app, I think now it could move to external tool.

Dockerfile.compat
```diff
@@ -0,0 +1,106 @@
+FROM docker.io/library/python:3.13-slim AS builder
```
dalf, line 1: This might be common knowledge, but adding a comment at the top of the file to indicate that this Dockerfile is intended for ARMv7 would be helpful for new comers.

If we want to avoid to duplicate information (ARMv7 here in comment, ARMv7 in code elsewhere), we can add a comment stating that `utils/lib_sxng_container.sh` pick this file for some architectures.

Dockerfile
edevosc2c, line 1: Would it be possible to change from 
```
FROM docker.io/library/python:3.13-slim
```
to:
```
FROM docker.io/library/python:3.12-24.04_stable
```

Debian has a lot of CVE that ubuntu does not have:
- https://github.com/searxng/searxng/security/code-scanning
- https://github.com/netbox-community/netbox-docker/issues/800
- https://www.chainguard.dev/unchained/the-zero-cve-challenge-can-official-docker-hub-images-pass-the-test

edevosc2c (changes requested):

Dockerfile
edevosc2c: Or maybe just from ubuntu:24.04 then install python

inetol: > FROM docker.io/library/python:3.12-24.04_stable

The official images are only served under Debian or Alpine, there is no Ubuntu distro tagged.

> Debian has a lot of CVE that ubuntu does not have:

I was also thinking about the vulnerabilities that there are a lot of them right now, I already have a workaround for the amd64 and arm64 images, but it might not be accepted and depends on adding all the new workflows first.

unixfox, line 1: (OOps I commented with my work account)

---
@unixfox:

Would it be possible to change from 
```

FROM docker.io/library/python:3.13-slim
```
to:
```
FROM docker.io/library/python:3.12-24.04_stable
```

Debian has a lot of CVE that ubuntu does not have:
- https://github.com/searxng/searxng/security/code-scanning
- https://github.com/netbox-community/netbox-docker/issues/800
- https://www.chainguard.dev/unchained/the-zero-cve-challenge-can-official-docker-hub-images-pass-the-test

Only for AMD64 and ARM64. armv7 we keep it like this since we said it's not officially supported anymore.

----

@unixfox:

Or maybe just from ubuntu:24.04 then install python
Then we can also use ubuntu on armv7.

---

@inetol:

> FROM docker.io/library/python:3.12-24.04_stable

The official images are only served under Debian or Alpine, there is no Ubuntu distro tagged.

> Debian has a lot of CVE that ubuntu does not have:

I was also thinking about the vulnerabilities that there are a lot of them right now, I already have a workaround for the amd64 and arm64 images, but it might not be accepted and depends on adding all the new workflows first.

unixfox (changes requested):

Dockerfile
unixfox: @inetol we found that with @dalf  that ubuntu has a much better track record for fixing the vulnerabilities.

Hence why, it would make sense to migrate to ubuntu base.

> FROM docker.io/library/python:3.12-24.04_stable

I was referering to `ubuntu/python:3.12-24.04_stable` instead but it is not a good candidate because it doesn't have bash nor apt installed.

So `FROM ubuntu:24.04` then install python is probably much better.

inetol: Please move this thread to a new issue

unixfox: ok done: https://github.com/searxng/searxng/issues/4753

inetol:
I'll keep the commit devel history here.

This PR is superseded by #4764

Dockerfile.compat
inetol: I've renamed the `Dockerfile.compat` to `armv7.Dockerfile` here https://github.com/searxng/searxng/pull/4764/files#diff-92d87db7221875202a920aefc797462a7edac78c8ee2c172540cb2876811be75

Issue #4718, opened 2025-05-05
Bug: Latest container not working


Drikanis:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

Docker container tag:
`searxng/searxng:latest@sha256:46745cb5e69796fc601c17b2c4dc34351341203e15ada46ecae100bfe050e42a`

Cannot determine application commit id because it will not load.

**How did you install SearXNG?**

Docker deployed in kubernetes cluster, issue reproducible with only docker.

**What happened?**

On upgrade to the image listed above, the container would not start up at all because uwsgi.ini could not be found. After finding the new location of the file in the container and setting `UWSGI_SETTINGS_PATH=/usr/local/searxng/dockerfiles/uwsgi.ini`, the container would start and appear to start listening but would not respond to any requests, timing out the gateway instead.

**How To Reproduce**

You can reproduce this just with docker:
```
docker run --rm -p 8080:8080 -e UWSGI_SETTINGS_PATH=/usr/local/searxng/dockerfiles/uwsgi.ini searxng/searxng:latest@sha256:46745cb5e69796fc601c17b2c4dc34351341203e15ada46ecae100bfe050e42a
```
Browsing to http://localhost:8080 eventually just times out as the application doesn't respond.

**Expected behavior**

Application should function.

**Screenshots & Logs**

```
SearXNG version 2025.5.4+2e74d8632
Use existing /usr/local/searxng/dockerfiles/uwsgi.ini
Use existing /etc/searxng/settings.yml
Listen on [::]:8080
[uWSGI] getting INI configuration from /usr/local/searxng/dockerfiles/uwsgi.ini
[uwsgi-static] added mapping for /static => /usr/local/searxng/searx/static
*** Starting uWSGI 2.0.29 (64bit) on [Mon May  5 02:02:23 2025] ***
compiled with version: 12.2.0 on 04 May 2025 20:34:43
os: Linux-6.1.0-31-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.128-1 (2025-02-07)
nodename: searx-698fcbfd7f-689jg
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 12
current working directory: /usr/local/searxng
detected binary path: /usr/local/searxng/venv/bin/uwsgi
chdir() to /usr/local/searxng/searx/
your memory page size is 4096 bytes
detected max file descriptor number: 1048576
[...]

PR #4721, opened 2025-05-05
[mod] Rework Dockerfile - migrate to glibc (debian)

inetol:
Same as #4699

I've looked over it (I can't replicate it as usual) and I deduce that the problem is this here https://github.com/searxng/searxng/pull/4699/commits/4aaa97a1d80d6267685e0291ad270875393cdb35

I will keep the same execution routine as before, entrypoint will run under root and then uWSGI under the `searxng` user.

Related https://github.com/searxng/searxng/issues/4718

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,8 @@ WORKDIR /usr/local/searxng/
 
 COPY ./requirements.txt ./requirements.txt
 
-RUN --mount=type=cache,id=pip,target=$HOME/.cache/pip python -m venv ./venv \
+# Readd on #4707 "--mount=type=cache,id=pip,target=/root/.cache/pip"
+RUN python -m venv ./venv \
  && . ./venv/bin/activate \
  && pip install -r requirements.txt \
  && pip install "uwsgi~=2.0"
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,8 @@ FROM docker.io/library/python:3.13-slim
 
 RUN apt-get update \
  && apt-get install -y --no-install-recommends \
+    # healthcheck
+    wget \
     # uwsgi
     libpcre3 \
     libxml2 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,92 +1,102 @@
-FROM alpine:3.20
-ENTRYPOINT ["/sbin/tini","--","/usr/local/searxng/dockerfiles/docker-entrypoint.sh"]
-EXPOSE 8080
-VOLUME /etc/searxng
+FROM docker.io/library/python:3.13-slim AS builder
 
-ARG SEARXNG_GID=977
-ARG SEARXNG_UID=977
-
-RUN addgroup -g ${SEARXNG_GID} searxng && \
-    adduser -u ${SEARXNG_UID} -D -h /usr/local/searxng -s /bin/sh -G searxng searxng
-
-ENV INSTANCE_NAME=searxng \
-    AUTOCOMPLETE= \
-    BASE_URL= \
-    BIND_ADDRESS=[::]:8080 \
-    MORTY_KEY= \
-    MORTY_URL= \
-    SEARXNG_SETTINGS_PATH=/etc/searxng/settings.yml \
-    UWSGI_SETTINGS_PATH=/etc/searxng/uwsgi.ini \
-    UWSGI_WORKERS=%k \
-    UWSGI_THREADS=4
-
-WORKDIR /usr/local/searxng
-
-COPY requirements.txt ./requirements.txt
-
-RUN apk add --no-cache -t build-dependencies \
-    build-base \
-    py3-setuptools \
-    python3-dev \
-    libffi-dev \
-    libxslt-dev \
-    libxml2-dev \
-    openssl-dev \
-    tar \
-    git \
- && apk add --no-cache \
-    ca-certificates \
-    python3 \
-    py3-pip \
-    libxml2 \
-    libxslt \
-    openssl \
-    tini \
-    uwsgi \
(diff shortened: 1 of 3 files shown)


unixfox:
Most likely also here: https://github.com/searxng/searxng-docker/blob/21bca5efa84c04890b13ae8f3cab6d2aa87d2a73/docker-compose.yaml#L56.

Because it's not a named volume, the permissions are not created with the UID and GID of searxng user.

I would vote to create a migration guide asking to set the new permissions. Because running the docker as root is definitively harmful to the user.

We can for example have a message printed in the logs when permissions denied issue is found.

What do you think?

unixfox:
> Most likely also here: [searxng/searxng-docker@`21bca5e`/docker-compose.yaml#L56](https://github.com/searxng/searxng-docker/blob/21bca5efa84c04890b13ae8f3cab6d2aa87d2a73/docker-compose.yaml#L56).
> 
> Because it's not a named volume, the permissions are not created with the UID and GID of searxng user.
> 
> I would vote to create a migration guide asking to set the new permissions. Because running the docker as root is definitively harmful to the user.
> 
> We can for example have a message printed in the logs when permissions denied issue is found.
> 
> What do you think?

I guess we can do this take in another PR and just focus for the moment on migrating to glibc.

The fact that https://github.com/searxng/searxng-docker/blob/21bca5efa84c04890b13ae8f3cab6d2aa87d2a73/docker-compose.yaml#L56 uses a local folder because it's mounted from https://github.com/searxng/searxng-docker/tree/21bca5efa84c04890b13ae8f3cab6d2aa87d2a73/searxng and the need to migrate to a named volume is going to be tricky to solve.

If we run uWSGI as searxng that's fine.

unixfox:
Last comment, why do we let uwsgi.ini to be modified though? Why not having it outside of /etc/searxng and directly only in the container? We could just let the ability configure this file only through environment variables. That would make things easier.

inetol:
> I would vote to create a migration guide asking to set the new permissions. Because running the docker as root is definitively harmful to the user.
> 
> We can for example have a message printed in the logs when permissions denied issue is found.
> 
> What do you think?

We can automate part of the migration, in the entrypoint under root we modify the ownership of the two volumes recursively to the `searxng` user and warn afterwards (uWSGI will keep running as `searxng`).

What I think might be a better idea, though I don't know if it works this way, is if moving the USER keyword on top of VOLUME will cause them to be created *(or updated)* under the USER ownership. It's something I'd have to look into further, but if it works, we'd have this issue solved.

Running the container under one user and then switching to another is horrible, it's messy for anyone who needs to set up the container to run under a specific user, and I'd like to avoid this behavior.

inetol:
> Last comment, why do we let uwsgi.ini to be modified though? Why not having it outside of /etc/searxng and directly only in the container? We could just let the ability configure this file only through environment variables. That would make things easier.

No idea, uWSGI I believe allows configuration via env, but for how close we are to replacing it with Granian (at least in the container) I think I will refrain from touching or seeing anything else from uWSGI. With Granian we can configure via env, so will be one less file to track.

unixfox:
OK thanks.

I propose to tackle the issues with the user permissions after https://github.com/searxng/searxng/pull/4596 is merged. Let's just tackle glibc support in this PR.

unixfox:
It works after migrating from an existing install with searxng-docker. But the searxng version is incorrect because it needs "git" to be installed:
```
searxng  | 2025-05-05 17:11:18,513 ERROR:searx: git is not found, fallback to the default version
searxng  | 2025-05-05 17:11:18,514 ERROR:searx: git is not found, fallback to the default version
searxng  | 2025-05-05 17:11:18,514 ERROR:searx: git is not found, fallback to the default version
searxng  | 2025-05-05 17:11:18,609 ERROR:searx: git is not found, fallback to the default version
```

inetol:
I can't replicate to get that error on clean builds on this branch. I remember getting that same message every time before, *so probably some black magic is happening which needs a defined argument on container build* (yes).

https://github.com/searxng/searxng/blob/0315988f5ab9d5d55693bc9a0eca045bf9220506/searx/version.py#L79:L99

Try building the container with `make docker.buildx` or `make docker.build`

unixfox:
@inetol Great that was it. I used `docker build` directly (didn't want to wait for armv7 to finish) and that was causing the issues with the version.

I have added @Bnyro @tiekoetter as reviewers, this way, nobody will slap me for merging this PR because I won't be the only one reviewing it :smile: 

@inetol Feel free to merge this PR once the reviewers have reviewed your PR. Since you have rights to do it now. Thanks!

unixfox (approved):
1. I launched searxng-docker on a public VPS with a public IP.
2. Built this docker image.
3. Switched to this docker image on the previous searxng-docker install.
4. That worked without errors.

tiekoetter (approved):
Using `make docker.build` and using this created image with the default `docker-compose.yml` yielded a working SearXNG instance.

Host: New install of Debian 12 on a KVM instance, with necessary tools (vim, git, build-essential, python3, etc.) installed.

Procedure:
1. Cloning inetol/searxng as well as searxng/searxng-docker
2. Building the SearXNG docker image with `make docker.build`.
3. Editing `docker-compose.yml` to use the local built SearXNG image.
4. Setting up the environment variables to use the correct domain.
5. `docker compose up`
6. Using the local install.

-> No errors.

unixfox:
@inetol you can merge this PR with "squash and merge" button. @Bnyro approved the message of @tiekoetter so I think this is good for everyone.

Thank you!

Bnyro:
Thanks for working on that 👍

PR #4728, opened 2025-05-07
[fix] sec-fetch-* blocking infinite scroll

vojkovic:
## What does this PR do?

Immediately fixes the infinite scroll by including 'cors' as allowed for Sec-Fetch-Mode and 'empty' for Sec-Fetch-Dest.

The infinite scroll has been broken by https://github.com/searxng/searxng/commit/fe08bb1d909cb6cef57ce91211c2cbed63300c9e

Enable the botdetection and test the infinite scroll. Patch has already been applied on https://priv.au

https://github.com/privau/searxng/commit/4c814a2e798a99d3fa3914a966ef6ad8c5d51a5c

Requesting review by @return42

### DIFF
--- a/searx/botdetection/http_sec_fetch.py
+++ b/searx/botdetection/http_sec_fetch.py
@@ -86,7 +86,7 @@ def filter_request(
     user_agent = request.headers.get('User-Agent', '')
     if is_browser_supported(user_agent):
         val = request.headers.get("Sec-Fetch-Mode", "")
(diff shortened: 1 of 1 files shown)


unixfox:
Very good catch @vojkovic. The infinite scroll sends additional AJAX requests, so it makes sense to whitelist those headers.

unixfox (approved):

PR #4729, opened 2025-05-08
[doc] add chapter on the purpose of (git) commits

return42:
The commits and their messages are elementary for the traceability of changes and are unfortunately still too often given too little attention.

It therefore seems necessary to dedicate a chapter to this topic in the context of development.

### DIFF
--- a/docs/dev/commits.rst
+++ b/docs/dev/commits.rst
@@ -1,0 +1,108 @@
+.. _create commit:
+
+===============================
+Git Commits & Change Management
+===============================
+
+.. sidebar:: Create good commits!
+
+   - `Conventional Commits`_
+   - `Structural split of changes`_
+   - `Git Commit Good Practice`_
+
+A commit and its commit message are among the most important information
+available to a developer for bug fixing and further development.  A commit is a
+change and changes have a context (a change request).
+
+In a SCM system (git), the change history is derived from the commit history. A
+commit message is therefore part of the documentation for change management and
+thus elementary for the traceability of changes.
+
+**What a commit is not**: *A commit to an SCM system is not used to save files!*
+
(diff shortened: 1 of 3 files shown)


docs/dev/commits.rst
```diff
@@ -0,0 +1,62 @@
...
+- **Commit messages are for the reader not for the author of the commit.**
+- **A commit is the atomic code-modification of a change in change management.**
+- **Think about which descriptions from your PR might belong in the commit message.**
+- **The maximum line length in a commit message is 80 characters.**
+
+Choose meaningful commit messages, read `Conventional Commits`_
+
+  .. code::
+
+     <type>[optional scope]: <description>
```
Bnyro: At SearXNG we actually use `[type] optional scope: description` and not the style mentioned here.

Bnyro: I think that it makes sense to add some examples here for easier understandability.

docs/dev/contribution_guide.rst
```diff
@@ -88,15 +80,7 @@ In order to submit a patch, please follow the steps below:
...
-- Choose meaningful commit messages, read `Conventional Commits`_
-
-  .. code::
-
-     <type>[optional scope]: <description>
-
-     [optional body]
-
-     [optional footer(s)]
+- Choose meaningful commit messages, read :ref:`create commit`
```
Bnyro:
```suggestion
- Choose meaningful commit messages, see :ref:`create commit`
```
just a minor idea to make it sound slightly better

docs/dev/commits.rst
return42: Yeah, thats true .. I have revised the text again, please take another look at it.

PR #3765, opened 2024-08-25
[fix] reduce traceback logging in search checker

glanham-jr:
## What does this PR do?

 - Implement `EngineProcessor.log_engine_exc_info` to allow removing exception logging in Checker. Enabled to its default value `True`.
 - In the Checker, disable this value
 - In the ErrorContext, we currently log all ErrorContext as `warning`, including more noise. Here, I propagate the log_level by allowing it to be injected (defaulting to warning). Then, it is set to NOTSET in the online processor if `enging_exc_info` is False (this may not be the best approach, but works for this scenario).
 - With the greatly reduced log lines, we can now log which test is occurring, which may single out which test failed for better investigation. This doesn't have to be done, but thought it may aid in debugging (see screenshots)

All changes are backwards compatible. Most of the changes largely sum up to injecting parameters, such that injected parameters default to the original implementation.

## Why is this change important?

See related issue.

## How to test this PR locally?

 - run `make search.checker`
 - ensure noisy logs are not generated

## Author's checklist

 - n/a

## Related issues

closes #3310 

## Screenshots

Removed exception details, added logging for which test is occurring.

image

PR #3312, opened 2024-03-09
Checker: change image check

dalf:
## What does this PR do?

In the master branch, the checker starts to stream the response and cut the connection. However it creates a lot of read error, which are false negative. I don't know how to fix the issue.

This commit change the checker to download the whole image. The error reporting is also changed to report only one line, instead of the whole stacktrace.

Do note that I have not test the checker running in background. This feature seems forgotten and lack of interrest despite the initial move few years ago.

## How to test this PR locally?

* `make search.checker.brave.images`
* `make search.checker.google_images` (`_` not `.`).
 
## Related issues

Close https://github.com/searxng/searxng/issues/3311

PR #4730, opened 2025-05-08
[enh] refactor checker.yml workflow

inetol:
This PR is part of #4707

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -1,31 +1,46 @@
-name: "Checker"
-on:  # yamllint disable-line rule:truthy
+---
+name: Checker
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
   schedule:
     - cron: "0 4 * * 5"
-  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  PYTHON_VERSION: "3.13"
 
 jobs:
-  checker:
-    name: Checker
-    runs-on: ubuntu-24.04
+  search:
+    name: Search
+    runs-on: ubuntu-24.04-arm
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Install Ubuntu packages
-        run: |
-          sudo ./utils/searxng.sh install packages
-
-      - name: Set up Python
+      - name: Setup Python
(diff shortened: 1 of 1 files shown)


return42 (approved):
LGTM / slightly improved the commit title.

> shared cache between .. not functional until all workflows have been refactored

But the PR itself (the CI job) will also work beforehand, just without the cache ... right?

If so, I would like to merge the PR right away .. tomorrow we would then also see whether the cron task starts.

.github/workflows/checker.yml
```diff
@@ -1,31 +1,46 @@
...
+  PYTHON_VERSION: "3.13"
 
 jobs:
-  checker:
-    name: Checker
-    runs-on: ubuntu-24.04
+  search:
+    name: Search
+    runs-on: ubuntu-24.04-arm
     steps:
```
return42, line 24: I think that's a very good idea to use an ARM image for the CI jobs, so we discover any problems early on that we don't have on our developer desktop systems :+1:

inetol:
> But the PR itself (the CI job) will also work beforehand, just without the cache ... right?

If no cache is found, then everything is set up from scratch (like running `make install` from a clean `git clone`) and the “local/” folder will be saved at the end of the workflow. That cache can be reused as many times as necessary, and it will be automatically selected/rotated depending on the cache key values

inetol:
> If so, I would like to merge the PR right away .. tomorrow we would then also see whether the cron task starts.

You can manually run the workflow with `workflow_dispatch`, I put this "just in case" it fails the crontab. See here: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#running-a-workflow

return42:
> You can manually run the workflow 

Thanks for pointing out .. FYI: to have a test in the master branch, I merged this PR and started https://github.com/searxng/searxng/actions/runs/14906115807

----

The Checker ends with an error, but this issue is not related to this PR (its an old known issue). Related:

- https://github.com/searxng/searxng/pull/3765
- https://github.com/searxng/searxng/pull/3312

.github/workflows/checker.yml
```diff
@@ -1,31 +1,46 @@
-name: "Checker"
-on:  # yamllint disable-line rule:truthy
+---
+name: Checker
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
```
return42, line 6: > You can manually run the workflow with `workflow_dispatch`

This option is very helpful :+1: :+1:

PR #4732, opened 2025-05-08
[enh] refactor data-update.yml workflow

inetol:
This PR is part of https://github.com/searxng/searxng/pull/4707

### DIFF
--- a/.github/workflows/data-update.yml
+++ b/.github/workflows/data-update.yml
@@ -1,14 +1,27 @@
-name: "Update searx.data"
-on:  # yamllint disable-line rule:truthy
+---
+name: Update searx.data
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
   schedule:
     - cron: "59 23 28 * *"
-  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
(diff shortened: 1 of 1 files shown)


return42:
I manually started the "Update searx.data" action, but it fails when creating a PR

- https://github.com/searxng/searxng/actions/runs/14907353305

grafik

I can't explain why the PRs could still be created before this change, but now the permission is missing. Does it possibly have something to do with the fact that the “comitter” has changed?

Before: `searxng-bot <noreply@github.com>`

grafik


Now: `searxng-bot <searxng-bot@users.noreply.github.com>`

grafik

----

It has nothing to do with this change, but what I've been wondering for a long time is that my `return42` account is used there as an author ... shouldn't that also be a neutral account?

https://github.com/searxng/searxng/blob/c6a70782b221917f58907ba43aba505f02fd6b3f/.github/workflows/data-update.yml#L65-L66

inetol:
> It has nothing to do with this change, but what I've been wondering for a long time is that my return42 account is used there as an author ... shouldn't that also be a neutral account?

Well, you manually executed the workflow, so I think that's the expected thing to happen. Although it has nothing to do with the permission denied issue, I'll push a PR now.

return42:
> Well, you manually executed the workflow, so I think that's the expected thing to happen. 

Nope .. `github.actor` seems to the author from the commit that patches the workflow file. 

- https://github.com/searxng/searxng/pull/4738

.. now its you :-) 

>  ... shouldn't that also be a neutral account?

PR #4733, opened 2025-05-08
[mod] new documentation.yml workflow

inetol:
This PR is part of https://github.com/searxng/searxng/pull/4707

PR #4734, opened 2025-05-08
[mod] new l10n.yml workflow

inetol:
This PR is part of https://github.com/searxng/searxng/pull/4707


return42:
The process of synchronizing the translations of weblate and the master branch takes place via an inter-branch [(branch `translations`)](https://github.com/searxng/searxng/tree/translations). 

The process is therefore anything but a standard workflow and is described here: https://docs.searxng.org/dev/translation.html

@dalf knows this best, which is why I would ask him for support.

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,134 @@
...
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - Integration
+    types:
+      - completed
+    branches:
+      - master
+  schedule:
+    - cron: "05 07 * * 5"
```
dalf, line 15: As far I understand, the jobs `update` and `pr` have two different triggers. Is this the modern approach - to merge everything related to one topic into a single workflow, even if it effectively consists of two separate ones?

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,134 @@
...
+  contents: read
+
+env:
+  PYTHON_VERSION: "3.13"
+
+jobs:
+  update:
+    if: github.repository_owner == 'searxng' && github.event.workflow_run.conclusion == 'success'
+    name: Update
+    runs-on: ubuntu-24.04-arm
```
dalf, line 31: Perhaps this has already been answered: why arm?
I don't mind, just curious.

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,134 @@
...
+        uses: peter-evans/create-pull-request@v7
+        with:
+          author: "${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>"
+          committer: "searxng-bot <searxng-bot@users.noreply.github.com>"
+          title: "[l10n] update translations from Weblate"
+          commit-message: "[l10n] update translations from Weblate"
+          branch: "translations_update"
+          delete-branch: "true"
+          draft: "false"
+          signoff: "false"
```
dalf, line 129: Nitpicking and curiosity: why `"false"` rather than `false` as in the [documentation](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#reference-example)?

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,134 @@
...
+name: Translation
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - Integration
+    types:
+      - completed
```
dalf, line 11: The fact the translations are not extracted in the "Integration" workflow can lead be problematic (*) in this scenario:
* a commit makes the extraction of the translations impossible for whatever reason
* the workflow "Integration" works as expected --> there is a green check
* this workflow fails

From my point of view, actually the commit should be rejected with a red cross (no green mark)

(*) there is a notification that this workflow fails, but this is less clear which commit has an issue.

Perhaps I miss something about GitHub and/or workflows ?

.github/workflows/l10n.yml
inetol: It is a bit confusing to have two related things in different workflows, I would like to target each workflow to a specific thing.

It is now a single workflow that is executed once integration.yml is finished, whether it is executed manually or by crontab (this is defined on the `on` on top of the file), but depending on each event it will execute one or another job. You can see under what conditions in the `if` at the beginning of each job.

.github/workflows/l10n.yml
inetol: To me it is for efficiency reasons, but it might also help to catch an unexpected error under arm64 which doesn't appear under amd64 on development

.github/workflows/l10n.yml
inetol: This is for consistency and ease of distinguishing the `with` section under the IDE.

I think it was it the action to setup Python that the versions must be in quotes because if not it might not get all the value correctly. To avoid having *that* in quotes and the rest of the params not, I put it all under quotes. In the end, all these are strings for yaml.

.github/workflows/l10n.yml
inetol: Ok I understand, I don't know if there is any way to show its status in *the status* while keeping it chained with integration.yml. If there is no way, I could make l10n.yml standalone and have it run like integration.yml, every push and pull request will also trigger l10n.yml.

.github/workflows/l10n.yml
dalf: Still nitpicking: it's only one parameter in another GH action.
But anyway GH provide supports reading boolean from string.

* https://github.com/peter-evans/create-pull-request/blob/889dce9eaba7900ce30494f5e1ac7220b27e5c81/src/main.ts#L24
* https://github.com/actions/toolkit/blob/main/packages/core/README.md#inputsoutputs

.github/workflows/l10n.yml
dalf: Actually this is already for documentation with https://github.com/searxng/searxng/pull/4733
Let's see how it goes.

.github/workflows/l10n.yml
return42: my 5cent: to avoid YAML's implicit typing, for non keywords I recommend to use quotation marks: [The Norway problem](https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell) .. and other pitfalls.

.github/workflows/l10n.yml
inetol: I have a doubt, this workflow (the `update` job) should be executed after the integration.yml result is successful? Or that doesn't matter and it should be executed independently of the workflow status of integration.yml?

*Originally, this (the `update` job) ran under the master branch depending on `python` and `themes` jobs in integration.yml finishing successfully.*

dalf: It updates the branch `translations` which is read by weblate to manage the translations, technically this workflow can as soon as it can. However, it usually does not make sense to ask for new translations on weblate if `python` or `themes` fail, that's why `update` depends on them.

If the workflow fail, GH issue aside, it usually mean the new commit has an issue that must be fixed to provide the new translation.

.github/workflows/l10n.yml
dalf: From my point of view, I find it confusing to have multiple triggers within a single file, each performing different functions, but let's try it.
A comment would be helpful to have a quick overview:
* cron an workflow_dispatch --> pr
* workflow_run --> update

dalf:
From my point view, the PR is ready to be merged except the `on` section.

In documentation.yml there is 

```yaml
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

to trigger the workflow, I'm not sure how to do that here while waiting for the Integration workflow?

inetol:
In the case of documentation.yml as we want to test if it will build correctly before merging then we trigger on push or PR without depending on other workflows. Here instead we wait for integration.yml to finish, then it will automatically run this workflow (workflow_run trigger) and check if integration.yml finished correctly. (on `update` job with github.event.workflow_run.conclusion)

dalf (approved):

PR #4731, opened 2025-05-08
[enh] refactor security.yml workflow

inetol:
This PR is part of https://github.com/searxng/searxng/pull/4707

PR #4735, opened 2025-05-08
[fix] missing permission in security.yml workflow

inetol:
Uploading SARIFs needs to write into the repository GitHub security tab.

Reported in #4731


return42:
Perfect, the security.yml workflow now runs without issues ..

- https://github.com/searxng/searxng/actions/runs/14909847906/job/41881223678

PR #4736, opened 2025-05-08
[fix] missing permission in data-update.yml workflow

inetol:
We need to keep the token on checkout and allow writing into the repository to create the branch

Reported on #4732


return42:
Thanks .. permission is now given and we are one step further .. but there is a followup  error :-)

grafik

-----

As we can see, the feature branches are created .. only the PR is missing:

grafik

PR #4764, opened 2025-05-10
[mod] rework container deployment

### LINKED ISSUE #484: GitHub workflow: docker steps: check the image is actually working.
Replace 
https://github.com/searxng/searxng/blob/082d55e6c5c2f576396d4790179e13ad627f7253/.github/workflows/integration.yml#L161-L163

By:
* build the image
* start the image
* make some tests
* push

inetol:
This PR supersedes and is part of https://github.com/searxng/searxng/pull/4707

Closes #484


utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+            "ghcr.io/$CONTAINER_IMAGE_ORGANIZATION/cache:$CONTAINER_IMAGE_NAME-$arch$variant" >/dev/null
+
+        podman start "$name" >/dev/null
+        podman logs -f "$name" &
+        pid_logs=$!
+
+        # Wait until container is ready
+        sleep 5
+
+        curl -vf --max-time 5 "http://localhost:8080/healthz"
```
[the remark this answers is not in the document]
inetol: No

return42 (commented):
> @inetol requested your review on: https://github.com/searxng/searxng/pull/4764 [mod] rework container deployment.

I'm sorry, I can't review since I don't have any docker knowledge.

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+        platform="linux/$arch/$variant"
+        ;;
+    *)
+        err_msg "Unsupported architecture; $parch"
+        exit 1
+        ;;
+    esac
+    build_msg CONTAINER "Selected platform: $platform"
+
+    # Check if podman is installed
```
dalf: Nitpicking: I would move all requirements at one top of the function, just after

```sh
    if [ "$GITHUB_ACTIONS" != "true" ]; then
        die 1 "This command is intended to be run in GitHub Actions"
    fi
```

(same for `container.push`)

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+        platform="linux/$arch"
+        ;;
+    "ARM64" | "aarch64" | "arm64")
+        dockerfile="Dockerfile"
+        arch="arm64"
+        variant=""
+        platform="linux/$arch"
+        ;;
+    "ARMV7" | "armhf" | "armv7l" | "armv7")
+        dockerfile="armv7.dockerfile"
```
dalf: Nitpicking: why not `Dockerfile.armv7`, so this is next to `Dockerfile` when the file list is sorted?

dalf (approved):
LGTM

I haven't test `container.push`, but it makes sense.
I'm not sure why this is draft?

Thank you.

utils/lib_sxng_container.sh
inetol: I put it as `.dockerfile` because the IDE highlights it correctly with that extension, which doesn't happen with `dockerfile.`

I will try to move the Dockerfiles to dedicated folders instead of having it in the project root.

utils/lib_sxng_container.sh
inetol: 392a033913ca9496935ebe66bac061badda79d01

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+    local container_engine
+    local dockerfile
+    local arch
+    local variant
+    local platform
+
+    # Check if git is installed
+    if ! command -v git &>/dev/null; then
+        die 1 "Git is not installed"
+    fi
```
return42, line 25: No need for, but if it fits to your needs and you want: there is a `required_commands git` ..

https://github.com/searxng/searxng/blob/1b787ed35e9c51e335c42faee1f76695780ba4cb/utils/lib.sh#L60-L74

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+        else
+            die 1 "Podman/Docker is not installed"
+        fi
+    fi
+    info_msg "Selected engine: $container_engine"
+
+    # Setup arch specific
+    case $parch in
+    "X64" | "x86_64" | "amd64")
+        dockerfile="Dockerfile"
```
return42, line 55: nitpicking: indentation can be fixed to 8 spaces .. related to https://github.com/searxng/searxng/pull/4707#discussion_r2079256799 .. but we can also standardize indentations later when you create the PR for a shell formater.

inetol:
Note;

Clean build: https://github.com/searxng/searxng/actions/runs/14957565596
Cached build: https://github.com/searxng/searxng/actions/runs/14958140293

utils/lib_sxng_container.sh
```diff
@@ -0,0 +1,319 @@
...
+            "$container_engine" push "ghcr.io/$CONTAINER_IMAGE_ORGANIZATION/cache:$CONTAINER_IMAGE_NAME-$arch$variant"
+
+            # Output to GHA
+            {
+                echo "version_string=$VERSION_STRING"
+                echo "version_tag=$VERSION_TAG"
+                echo "docker_tag=$DOCKER_TAG"
+                echo "git_url=$GIT_URL"
+                echo "git_branch=$GIT_BRANCH"
+            } >>"$GITHUB_OUTPUT"
```
return42, line 155: No need for, just a hind .. for multiline outputs I often use *heredoc format*

```bash
cat << EOF > /path/to/your/file
This line will write to the file.
${THIS} will also write to the file, with the variable contents substituted.
EOF
```

*heredoc* is quite good to *read* and comfortable to *write*  .. it can be used in a variety of ways ..

If you want to pipe a series of commands e.g. to a bash .. and these commands should be displayed simultaneously on the terminal (to show the progress of the script)

https://github.com/searxng/searxng/blob/1b787ed35e9c51e335c42faee1f76695780ba4cb/utils/lib.sh#L893-L896

- https://linuxize.com/post/bash-heredoc/
- https://tldp.org/LDP/abs/html/here-docs.html

inetol: b43428830cb845dd752a41a6bd910e60dc23253d

utils/lib_sxng_container.sh
inetol: a22df071107a67604f1d881cc145af0f27246f2c

PR #4767, opened 2025-05-11
[fix] container not releasing

inetol:
`env.DOCKERHUB_USERNAME` shouldn't be an empty string as it's defined and set (I think, I can't see this), but it's failing the condition. Even if wasn't defined, GitHub Org/Repo wide envs/secrets should return empty strings?

I reverted the change back to use "null" keyword because that's what worked before, just don't understand why now doesn't works.

See https://github.com/searxng/searxng/actions/runs/14957565596/job/42015464191

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -153,13 +153,13 @@ jobs:
       - test
 
     steps:
-      - if: env.DOCKERHUB_USERNAME != ''
+      - if: env.DOCKERHUB_USERNAME != null
         name: Checkout
         uses: actions/checkout@v4
         with:
           persist-credentials: "false"
 
-      - if: env.DOCKERHUB_USERNAME != ''
+      - if: env.DOCKERHUB_USERNAME != null
         name: Login to GHCR
         uses: docker/login-action@v3
         with:
@@ -167,7 +167,7 @@ jobs:
           username: "${{ github.repository_owner }}"
           password: "${{ secrets.GITHUB_TOKEN }}"
 
-      - if: env.DOCKERHUB_USERNAME != ''
+      - if: env.DOCKERHUB_USERNAME != null
         name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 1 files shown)


inetol:
Still not working.

@return42 Are the envs/secrets revoked?

dalf:
I can see the DOCKERHUB_USERNAME variable is set.

inetol:
I don't know what to say, those are the logs:

```console
##[debug]Evaluating condition for step: 'Checkout'
##[debug]Evaluating: (success() && (env.DOCKERHUB_USERNAME != null))
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating NotEqual:
##[debug]....Evaluating Index:
##[debug]......Evaluating env:
##[debug]......=> Object
##[debug]......Evaluating String:
##[debug]......=> 'DOCKERHUB_USERNAME'
##[debug]....=> null
##[debug]....Evaluating Null:
##[debug]....=> null
##[debug]..=> false
##[debug]=> false
##[debug]Expanded: (true && (null != null))
##[debug]Result: false
```

You can see that `env.DOCKERHUB_USERNAME` returns null (also empty string), maybe there is some protection enforced related to this?

return42:
@inetol sorry for nitpicking: 

FYI: I have written down the previously unspoken recommendations for commit messages.

- https://docs.searxng.org/dev/commits.html

I wanted to ask you whether we can also specify a "CI" as the scope for the CI work.

e.g. da3c640cef

```
revert using null keyword

env.DOCKERHUB_USERNAME shouldn't be an empty string as it's defined and set (I think, I can't see this). Even if wasn't defined, GitHub Org/Repo wide envs/secrets should return an empty string (?)
```

```
[fix] CI: rework container deployment

env.DOCKERHUB_USERNAME shouldn't be an empty string as it's defined and set (I
think, I can't see this). Even if wasn't defined, GitHub Org/Repo wide
envs/secrets should return an empty string (?)

This patch reverted the change back to use "null" keyword because that's what
worked before, just don't understand why now doesn't works.

Bug comes from / patch fixes:

- https://github.com/searxng/searxng/pull/4764
```

Issue #4753, opened 2025-05-09
[Docker] Migrate from Debian to Ubuntu


unixfox:
As suggested in https://github.com/searxng/searxng/pull/4707#discussion_r2081283018, it would be better to migrate to another glibc docker base like ubuntu.

Ubuntu seems to have a good track record regarding fixing the vulnerabilities.

We could use the `ubuntu:24.04` base and install python ourselves. `ubuntu` docker image has support for ARMv7 too.

Debian has a lot of CVE that ubuntu does not have:
- https://github.com/searxng/searxng/security/code-scanning
- https://github.com/netbox-community/netbox-docker/issues/800
- https://www.chainguard.dev/unchained/the-zero-cve-challenge-can-official-docker-hub-images-pass-the-test

dalf:
Using

```diff
diff --git a/Dockerfile b/Dockerfile
index 9aeb28214..c32efc9d1 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,19 @@
-FROM docker.io/library/python:3.13-slim AS builder
+
+FROM docker.io/ubuntu:noble AS builder

 RUN apt-get update \
  && apt-get install -y --no-install-recommends \
     build-essential \
     brotli \
     # lxml
+    python3-dev \
     libxml2-dev \
     libxslt1-dev \
     zlib1g-dev \
     # uwsgi
     libpcre3-dev \
+    python3.12-venv \
+    python-is-python3 \
  && rm -rf /var/lib/apt/lists/*

 WORKDIR /usr/local/searxng/
@@ -42,7 +46,7 @@ RUN grep -m1 root /etc/group > /tmp/.searxng.group \
  && echo "searxng:x:$SEARXNG_GID:" >> /tmp/.searxng.group \
  && echo "searxng:x:$SEARXNG_UID:$SEARXNG_GID:searxng:/usr/local/searxng:/bin/bash" >> /tmp/.searxng.passwd

-FROM docker.io/library/python:3.13-slim
+FROM docker.io/ubuntu:noble

 RUN apt-get update \
  && apt-get install -y --no-install-recommends \
```

```sh
make docker.build
docker run -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/Library/Caches:/root/.cache/ aquasec/trivy:0.62.1 image searxng/searxng
```

The number of issues:
* 28 for Ubuntu
* 1 for searxng ( https://github.com/searxng/searxng/pull/4674 )

Note: 
[...]

PR #4768, opened 2025-05-11
[enh] container: use Wolfi OS as base image

inetol:
Using a specially designed base image for containers not only reduces maintenance burdens, but improves overall experience for developers (fewer packages we have to track) and end users (smaller images).

```console
REPOSITORY                 TAG                 SIZE
docker.io/searxng/searxng  2025.5.4-19b116f1d  195 MB
localhost/searxng/searxng  wolfi               214 MB
docker.io/searxng/searxng  2025.5.10-1b787ed35 346 MB
```
*(Comparison between the current images and the ones that were based on Alpine under amd64. Ordered from smallest to largest size)*

This change applies to both amd64 and arm64. armv7 device owners are excluded since it is not supported upstream.

CI latest test: https://github.com/inetol/searxng/actions/runs/15054787930

Related https://github.com/searxng/searxng/issues/4753

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -1,12 +1,10 @@
-FROM docker.io/library/python:3.13-slim AS builder
+FROM cgr.dev/chainguard/wolfi-base:latest AS builder
 
-RUN apt-get update \
- && apt-get install -y --no-install-recommends \
-    build-essential \
-    brotli \
-    # uwsgi
-    libpcre3-dev \
- && rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache \
+    build-base \
+    python-3.13-dev \
+    py3-pip \
+    brotli
 
 WORKDIR /usr/local/searxng/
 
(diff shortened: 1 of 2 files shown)


inetol:
`pcre` is not available on Wolfi repos, uWSGI will complain about this:
```console
!!! no internal routing support, rebuild with pcre support !!!
```

We can ignore this warning, as no issues appear present. Hopefully we will be able to get rid of uWSGI soon.

container/Dockerfile
```diff
@@ -37,20 +35,22 @@ RUN grep -m1 root /etc/group > /tmp/.searxng.group \
  && echo "searxng:x:$SEARXNG_GID:" >> /tmp/.searxng.group \
  && echo "searxng:x:$SEARXNG_UID:$SEARXNG_GID:searxng:/usr/local/searxng:/bin/bash" >> /tmp/.searxng.passwd
 
-FROM docker.io/library/python:3.13-slim
+FROM scratch
+
+# Patch base image
+COPY --from=builder /tmp/.searxng.passwd /etc/passwd
+COPY --from=builder /tmp/.searxng.group /etc/group
+COPY --chown=root:root --from=cgr.dev/chainguard/wolfi-base:latest / /
```
dalf, line 43: Why not use `FROM cgr.dev/chainguard/wolfi-base:latest` or `FROM cgr.dev/chainguard/python:3.13`?

inetol: `cgr.dev/chainguard/python` images are locked to the latest Python version under latest tag (paywall basically, like bitnami images recently), imagine if a new major Python version releases and some forked Dockerfile of sxng depending on a concrete version is disrupted because it uses that image.

`cgr.dev/chainguard/wolfi-base` has lots of garbage on labels, user `nonroot` by default which is good for simple builds, but for us it breaks root by default (you can't unset USER keyword without polluting final image). I prefer to start with a clean slate and copy all fs on top, shouldn't be any issues as the image is very lightweight (in the end everything is flattened to a single layer)

dalf: > cgr.dev/chainguard/python images are locked to the latest Python version under latest tag 

https://images.chainguard.dev/directory/image/python/versions

For example, the version with Python 3.10 has been updated few days ago, same for the other versions.
What I am missing?
Does that mean we can use these tags?

inetol: Nope, we can't. Only latest and latest-dev tags

inetol (commented):
CI reports OK. *(updated link)*
Exec under Docker/Podman OK.

In another PR I will look up a solution for armv7.

dalf:
There is an issue:
https://github.com/searxng/searxng/actions/runs/15055086429

In the tests:

> Error: no container with ID fc64445efc17e2a2fcb77dcbdc9d37130aeb293fbb1174e7200ecf31fdc9f4cc found in database: no such container

inetol:
I have absolutely no idea where this error comes from, I assume that with all the image and cache changes in the last days some layer has broken. I can purge the entire runner cache, but the container cache requires the cleanup workflow and the images need to be older than 1 week.

I will create a new branch in the repo modifying this to clear all the registry cache and restart the process. If I still can't get it to work I will revert this commit until I figure out what's happening.

inetol:
Magic https://github.com/searxng/searxng/actions/runs/15055806613

Anyway, I will take a closer look to see if it was due to what I said before or something else.

Issue #4374, opened 2025-02-25
Push a container to ghcr.io because of new DockerHub rate limits


ich777:
EDIT by @unixfox: SearXNG Docker images benefit from unlimited pulls: https://github.com/searxng/searxng/issues/4374#issuecomment-2704322138

----------

**Is your feature request related to a problem? Please describe.**
Since DockerHub changes it's rate limit with 1. April 2025 it would probably help many users to push the Docker container from SearXNG to ghcr.io to circumvent the really strict limit for 10 pulls per hour for unauthenticated users.

**Describe the solution you'd like**
Would be super cool if you could push a container to ghcr.io too.

**Describe alternatives you've considered**
None

**Additional context**
Here is the link to the new rate limit:
https://docs.docker.com/docker-hub/usage/

return42:
> Would be super cool if you could push a container to ghcr.io too.

Limits and pricing?

- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
- https://github.com/pricing

unixfox:
https://quay.io is also a good candidate. Has been free for a long time and is reliable.

> > Would be super cool if you could push a container to ghcr.io too.
> 
> Limits and pricing?
> 
>     * [docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)
> 
>     * [github.com/pricing](https://github.com/pricing?rgh-link-date=2025-02-25T09%3A00%3A43.000Z)

No, there is no issue in storage limitation for public repositories.

500MB is for private repositories.

ich777:
> Limits and pricing?

GitHub packages should come with no cost at all and no limits since it falls under the same rules as GitHub Releases where they only limit the size per package to 5GB.

I have about 25GB Packages (90 Packages) in my repository for Docker containers that I maintain.

return42:
> https://quay.io/ is also a good candidate. Has been free for a long time and is reliable.

👍 

> it falls under the same rules as GitHub Releases where they only limit the size per package to 5GB.

👍  .. just for reference -> https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas

unixfox:
[...]

PR #4769, opened 2025-05-11
[enh] container: support multiple registries

inetol:
docker.io applies a ratelimit to unauthenticated users, we should mirror to ghcr.io.

Images will be pushed to both `ghcr.io/searxng/searxng` and `docker.io/searxng/searxng`

Closes? https://github.com/searxng/searxng/issues/4374

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -117,10 +117,6 @@ jobs:
             os: ubuntu-24.04-arm
             emulation: true
 
-    permissions:
-      # Organization GHCR
-      packages: write
-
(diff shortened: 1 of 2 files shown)


inetol (commented):
Let's try this and see how it goes

inetol:
Images listed at https://github.com/searxng/searxng/pkgs/container/searxng

dalf:
The container is shown as private, I guess we can switch it to public?

inetol:
> I guess we can switch it to public?

Yes

PR #4772, opened 2025-05-12
[fix] CI: handle old cache images from registry

inetol:
Currently, we have 1100~ cache images uploaded to GHCR that weigh more than 300 MB each (most of them are layers from the second phase of the Dockerfile that were uploaded by mistake, read below). To avoid problems, I have set up a new job in a new workflow to be run weekly purging all images older than 1 week, but leaving always the 100 most recent ones.

Only the builder images should be uploaded to cache, the actual behaviour not only slows down the time for building the container, but also wastes lots of space by saving large and useless layers to GHCR that will never be used again.

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -1,0 +1,34 @@
+---
+name: Cleanup
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "4 4 * * 0"
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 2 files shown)


return42:
@inetol, this job is running on my clone https://github.com/return42/searxng/actions/runs/15223456127/job/42822507547

Can we restrict this job to run only on the repo searxng/searxng branch master?

inetol:
Yes. I'm seeing that error too, I don't know where it throws but needs to be fixed.

*Fixed in https://github.com/searxng/searxng/pull/4849*

unixfox:
@return42 

Tip: On the top right, in the “hamburger” menu, you can manually disable workflows like for example here: https://github.com/return42/searxng/actions/workflows/cleanup.yml

PR #4778, opened 2025-05-12
[fix] CI: container workflow env reference

### LINKED ISSUE #4777: CI: conditions for container release are not met
org envs/secrets for docker.io registry are missing, failing the conditions for the `container` workflow `release` job steps.

Runner logs:
```console
##[debug]Evaluating condition for step: 'Checkout'
##[debug]Evaluating: (success() && (env.DOCKERHUB_USERNAME != null))
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating NotEqual:
##[debug]....Evaluating Index:
##[debug]......Evaluating env:
##[debug]......=> Object
##[debug]......Evaluating String:
##[debug]......=> 'DOCKERHUB_USERNAME'
##[debug]....=> null
##[debug]....Evaluating Null:
##[debug]....=> null
##[debug]..=> false
##[debug]=> false
##[debug]Expanded: (true && (null != null))
##[debug]Result: false
```

See https://github.com/searxng/searxng/actions/runs/14975986186/job/42068697702

Reported from https://github.com/searxng/searxng/pull/4767

inetol:
When making the container rework, I unknowingly deleted the section where an env with the same name as the secret was defined on the job scope, making it look like it was originally defined as an organization env.

Since we can't validate the secrets in a condition directly, it's better to let `docker/login-action` take care of failing the entire job if the credentials are invalid.

Thanks @unixfox for noticing this.

Closes https://github.com/searxng/searxng/issues/4777

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -153,30 +153,26 @@ jobs:
       - test
 
     steps:
-      - if: env.DOCKERHUB_USERNAME != null
-        name: Checkout
+      - name: Checkout
         uses: actions/checkout@v4
         with:
           persist-credentials: "false"
 
-      - if: env.DOCKERHUB_USERNAME != null
-        name: Login to GHCR
+      - name: Login to GHCR
         uses: docker/login-action@v3
         with:
           registry: "ghcr.io"
           username: "${{ github.repository_owner }}"
           password: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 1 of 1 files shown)


unixfox (approved):

return42:
I missed the scope in the commit title: 5d99373b .. / while the PR title has one: ``[fix] CI: container workflow env reference``

- https://docs.searxng.org/dev/commits.html

PR #4784, opened 2025-05-14
[fix] CI: post refactor fixes

inetol:
I group these commits together because it's pointless to break them into separate PRs, they are all related to the CI.


return42 (approved):
LGTM :+1:

PR #4793, opened 2025-05-16
[mod] container: refactor entrypoint script

inetol:
That entrypoint is prone to screw things up, especially with permission handling. The new script handles initialization better and fixes some issues like delayed settings update via ENVs and timestamp overwriting, also adjusts what should be copied into the container.

This is the first step on making `searxng` user exec by default.

Latest CI: https://github.com/inetol/searxng/actions/runs/15134094471

Related https://github.com/searxng/searxng/pull/4721#issuecomment-2850272129


container/entrypoint.sh
```diff
@@ -0,0 +1,166 @@
+#!/bin/sh
```
return42, line 1: Of what type needs an docker-entry script to be? In the past (if I remember correctly) we used dash from alpine.

https://github.com/searxng/searxng/blob/86373e7c87aeca9fb0d8889a12cfe09c0ffd1df9/docs/admin/installation-docker.rst?plain=1#L24

Since most distros today install bash 

https://github.com/searxng/searxng/blob/86373e7c87aeca9fb0d8889a12cfe09c0ffd1df9/docs/admin/installation-docker.rst?plain=1#L142-L144

and having bash and sh (dash) scripts in this repo makes writing scripts unneeded complicated ..

https://github.com/searxng/searxng/blob/86373e7c87aeca9fb0d8889a12cfe09c0ffd1df9/docs/admin/installation-docker.rst?plain=1#L145-L146

I would recommend to use bash whenever possible.

-------------

I suspect `bash` can be added here (?) ..

https://github.com/searxng/searxng/blob/86373e7c87aeca9fb0d8889a12cfe09c0ffd1df9/container/base.yml#L6-L14

inetol: Here is ash, which is the shell provided by busybox. We don't require the use of bash, since we don't fiddle around inside the entrypoint. As it's a script that is only going to run in a container, we can save on interoperability.

inetol (commented):
CI reports OK. (updated link)
Exec under Docker/Podman OK.

unixfox:
hello @inetol, I would recommend you to try your future changes on https://github.com/searxng/searxng-docker too because it was obvious that this PR broke the project.

I am not against changing how things work, but maybe we could have done it a bit less harsh. Maybe by giving a head up/warning to everyone like 2-3 weeks in advance to remove the cap_drop and cap_add.

Thanks :).

PR #4795, opened 2025-05-16
redis to valkey migration

### LINKED ISSUE #3348: Investigating alternatives to using Redis
Redis will publish all their future code under a different, non-free license, see https://github.com/redis/redis/pull/13157.

Obviously this means that SearXNG would depend on a project that's not FLOSS and might only distribute binaries instead of keeping all new changes public - which is critical.

There is a fork https://codeberg.org/redict/redict which looks promising but still in its beginning, we'll see how it goes.

We could also consider using a completely different database from now on, but I'm certain that'd cause a lot of trouble with migrations, making all instance maintainers switch, ...

ping @unixfox, @return42, @dalf, @mrpaulblack

HLFH:
Take note:

- https://github.com/searxng/searxng/discussions/4986

----

## What does this PR do?

Fixes #3348

Completes migration from redis to valkey

## Related issues

Closes #3348


return42:
Wow .. very cool :rocket: .. thanks a lot .. related to https://github.com/searxng/searxng/pull/4790#issuecomment-2885899704, I'm back in the office next week and will make some tests and my review...

HLFH:
@return42 Hey, it fully works thanks to my second commit.

I think you can merge ;)

The ArchLinux AUR package `searxng-git` is currently available to test the `redis-to-valkey-migration` branch: https://aur.archlinux.org/packages/searxng-git.

HLFH:
@Bnyro Hello :) Any help for the merge / review? Thanks!

return42:
@HLFH I'm very sorry for the delay .. I haven't forget this PR ;-)

Merging this PR requires a bit more work. For example, a migration path needs to be documented (uninstalling the old Redis DB) and communicated with the community .. I have to test it on all the distros we offer .. and all the things that don't spontaneously occur to me at the moment.

That's a lot of work. I'd have to take a whole day off and do nothing else. And I haven't had that time yet (plus I first have to set up a virtualization environment where I can test everything; the old LXD scripts no longer work).

Please be patient with me a little longer... I'm very grateful for this PR, and my guilty conscience reminds me every day that I need to take action here.

searx/settings_defaults.py
```diff
@@ -185,8 +185,8 @@ def apply_schema(settings, schema, path_list):
         'method': SettingsValue(('POST', 'GET'), 'POST', 'SEARXNG_METHOD'),
         'default_http_headers': SettingsValue(dict, {}),
     },
-    'redis': {
-        'url': SettingsValue((None, False, str), False, 'SEARXNG_REDIS_URL'),
+    'valkey': {
+        'url': SettingsValue((None, False, str), False, 'SEARXNG_VALKEY_URL'),
     },
```
inetol: We should alias `SEARXNG_REDIS_URL` ENV internally so that if `SEARXNG_VALKEY_URL` is not defined, it fallbacks to `SEARXNG_REDIS_URL` instead.

docs/admin/settings/settings_redis.rst
inetol, line 1: It is possible to make an HTTP 308 redirect from `settings_redis.html` to `settings_valkey.html` on docs? Search engines keep this page indexed and might be problematic until every engine flushes this route. (maybe should we keep the `settings_redis` file name)

inetol (commented):
Since valkey lib has transparent compatibility with Redis, this won't (probably) break any existing installations on update. Someone can still use a Redis server without issues (for now) keeping the same URI as "redis*://..." in their config.

I've tried to build/test a container directly from this branch, using both Redis and Valkey servers and everything works OK on runtime, but I have some concerns...

searx/settings_defaults.py
virtadpt: If they speak the same wire protocol, there probably isn't any harm in continuing to use the string /redis/i.  Folks who know are just going to point Searx at their Valkey instance and be done with it.  Folks who don't know are going to be confused by the noticeable specificity (especially if they're using the original Redis for some reason (I'm thinking institutions)) and it'll trip them up.

utils/lib_valkey.sh
```diff
@@ -0,0 +1,355 @@
...
+#    "bar"
+#    [CTRL-D]
+
+
+# shellcheck disable=SC2091
+# shellcheck source=utils/lib.sh
+. /dev/null
+
+VALKEY_GIT_URL="https://github.com/valkey/valkey.git"
+VALKEY_GIT_TAG="${VALKEY_GIT_TAG:-6.2.6}"
```
return42: @HLFH sure it works for you? .. the URL seems to be https://github.com/valkey-io/valkey.git and a release 6.2.6 does not exists (initial it was 7.2.5 and current release is 8.1.2) .. 

Please don't modify your branch .. I will fix all issues in my review.

HLFH: You are right, it should be:

```
VALKEY_GIT_URL="https://github.com/valkey-io/valkey.git"
VALKEY_GIT_TAG="${VALKEY_GIT_TAG:-8.1.2}"
```

I fixed it locally but I won't push it to my upstream branch, as you said you will fix all issues in your review.

docs/admin/settings/settings_redis.rst
HLFH: Certainly. The webmaster managing `docs.searxng.org` should add an HTTP 308 redirect.
From: https://docs.searxng.org/admin/settings/settings_redis.html
To: https://docs.searxng.org/admin/settings/settings_valkey.html

File name `settings_valkey.rst` is perfect though as we are migrating away from Redis to Valkey.

utils/lib_valkey.sh
return42: > but I won't push it to my upstream branch, as you said you will fix all issues in your review.

yeah, thanks :+1: .. I think it will be better we do no longer compile valkey (redis), instead we can install from distro's package managers as described here https://valkey.io/topics/installation/

.. at least, I will give it a try in my review ... coming soon ..

searx/settings_defaults.py
HLFH: I think it'd be best to make `valkey:` the default as we are migrating away from `redis` and we are not the only one doing that.

Valkey is supported in diverse ways by Alibaba Cloud, AWS, Ericsson, Google Cloud, Heroku, Oracle, Percona, Verizon, Huawei, Tencent, OVH, Oracle, ByteDance, etc.
GNU/Linux distributions such as Fedora and ArchLinux have replaced Redis with Valkey.

I don't think we need a `SEARXNG_REDIS_URL` ENV fallback.
`valkey-py` supports legacy `redis:` URL scheme within the `settings.yml` of `searxng`.
I believe this is enough.

return42: > I don't think we need a `SEARXNG_REDIS_URL` ENV fallback.

In my review I implement a migration path for existing installations .. very simple .. for a *"deprecated"* phase `SEARXNG_REDIS_URL` will be valid .. in a few month then we can remove the 3 lines of code to end the phase. I will need some more time for my review, please give me the time, we can then discuss these topics further / if necessary.

utils/lib_valkey.sh
HLFH: As an example, my ArchLinux AUR package `searxng-git` does not compile `valkey`, it is using the `valkey` package provided by ArchLinux. 

If `lib_valkey.sh` is no longer useful, we could drop it and support `searxng` package maintainers for the mainstream distributions (ArchLinux, Fedora, Debian, Ubuntu).

docs/admin/settings/settings_redis.rst
inetol: The thing is that the subdomain docs.searxng.org is managed by GitHub Pages, I don't know if we can create redirects there.

HLFH: I believe we can create redirects. As we are using `Sphinx`. As a start, we could add this extension in `docs/conf.py`: https://github.com/sphinx-contrib/redirects.

return42: > It is possible to make an HTTP 308 redirect from `settings_redis.html` to `settings_valkey.html` on docs?

Thanks for pointing out :+1: ..

> I believe we can create redirects.

No need for .. in my review I added `settings_redis.html` back, it now contains notes about a migration path from Redis to Valkey.

utils/lib_valkey.sh
return42: > If `lib_valkey.sh` is no longer useful, we could drop it and support `searxng` package maintainers for the mainstream distributions (ArchLinux, Fedora, Debian, Ubuntu).

Yeah, is exactly what I did in `lib_valkey.sh` in my review .. [here](https://github.com/searxng/searxng/pull/4795/commits/71c5ba1a6c79c12c5d5756a47ab78edf5e72acc1#diff-88b0103fb95a8f95b9ef6a599d670f577d099a058ee2253b8ab67c50eb395bc1) ;-)

return42 (commented):
FYI I rebased your branch and pushed a WIP commit on top.  In the WIP commit (we can squash later) I implemented a migration path and the Valkey DB is now installed from distro's package manager.

I haven't finished all my tests, but my modifications from the WIP are in a good state .. you can review/comment the WIP commit ...

-----

To review first, read the docs::

    $ make docs.live

Follow the instructions to remove redis:

- http://0.0.0.0:8000/admin/settings/settings_redis.html

Config and install a local valkey DB:

- http://0.0.0.0:8000/admin/settings/settings_valkey.html

return42:
solved `Script attempted to access nonexistent global variable 'server'` in 

- https://github.com/searxng/searxng/pull/4795#issuecomment-3021954158

----

When I perform the update for this PR in an instance (master) with an old Redis DB, the instance no longer works. I don't yet know exactly why this is happening or how to avoid it.

The traceback is shown below. Since the usual get/set methods work and the following error only occurs with LUA scripts ~that are already registered in an old Redis DB~, it is likely that the LUA scripts are not downward compatible (global variable `redis`, `valkey` or `server`)? ~.. can we unregister / re-register LUA scripts? .. or should we force instance maingtainers to migrate to a Valkey-DB?~

Traceback::

      File "/usr/local/searxng/searx-pyenv/lib/python3.11/site-packages/flask/app.py", line 1536, in __call__
        return self.wsgi_app(environ, start_response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...
      File "/usr/local/searxng/searxng-src/searx/valkeylib.py", line 239, in incr_sliding_window
        c = script(args=[duration], keys=[name])
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/searxng/searx-pyenv/lib/python3.11/site-packages/valkey/commands/core.py", line 5959, in __call__
        return client.evalsha(self.sha, len(keys), *args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...
      File "/usr/local/searxng/searx-pyenv/lib/python3.11/site-packages/valkey/client.py", line 590, in parse_response
        response = connection.read_response()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/searxng/searx-pyenv/lib/python3.11/site-packages/valkey/connection.py", line 584, in read_response
        raise response
    valkey.exceptions.ResponseError: Error running script (call to f_d9e2a0cd9d7d0f3c6f01ebd3d82cdb84ba358786): \
      @enable_strict_lua:15: user_script:4: Script attempted to access nonexistent global variable 'server'

inetol:
> The traceback is shown below. Since the usual get/set methods work and the following error only occurs with LUA scripts that are already registered in an old Redis DB, it is likely that the LUA scripts in the old Redis DB are not upward compatible? .. can we unregister / re-register LUA scripts? .. or should we force instance maingtainers to migrate to a Valkey-DB?

What version of Valkey are you running? `server` var was introduced in [7.2.4-rc1](https://github.com/valkey-io/valkey/releases/tag/7.2.4-rc1).

Maybe we should modify 6fe65db4b502af41ca5269ff7d177abb94566f09 and use `redis` var until major distros bump to a "supported" version.

return42:
> What version of Valkey are you running? 

Not Valkey, its an old (existing) Redis-DB, version is `6.2.6`

> `server` var was introduced in [7.2.4-rc1](https://github.com/valkey-io/valkey/releases/tag/7.2.4-rc1).

In this use-case I test the *"deprecated phase"*, I have an old existing Redis DB:

```yaml
redis:
  url: unix:///usr/local/searxng-redis/run/redis.sock?db=0
```

The old Redis-DB was build by `./utils/searxng.sh install redis` .. The Redis-DB version is `6.2.6` ..

https://github.com/searxng/searxng/blob/39c50dc013944a0a27b4354c23f406956ac45971/utils/lib_redis.sh#L41-L42
 
----

> Maybe we should modify [6fe65db](https://github.com/searxng/searxng/commit/6fe65db4b502af41ca5269ff7d177abb94566f09) and use `redis` var until major distros bump to a "supported" version.

You hit the nail :+1: I changed 6fe65db to 23b2b3f50 ... In the *"deprecated phase"*, as long as we maintain backward compatibility with older Redis DB instances, we must use the global variable `redis` in LUA scripts.

return42:
I tested the migration path on ubu2504, fedora 42, arch linux and debian 12.

The tests resulted in some smaller commits; in particular, I had problems with the valkey server on Ubuntu and Debian (I had to comment out `PrivateUsers=true` / see https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PrivateUsers=).

I've finished all my tests and would now like to start cleaning up the commit history a bit. Is that okay with you?

return42:
FYI last force push was a rebase on master branch to solve a conflict ..

> I've finished all my tests and would now like to start cleaning up the commit history a bit. Is that okay with you?

due lack of feedback I assume its OK ;-) .. coming soon ..

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM mcr.microsoft.com/devcontainers/base:debian
 
 RUN apt-get update && \
-    apt-get -y install python3 python3-venv redis firefox-esr graphviz imagemagick librsvg2-bin fonts-dejavu shellcheck
+    apt-get -y install python3 python3-venv valkey firefox-esr graphviz imagemagick librsvg2-bin fonts-dejavu shellcheck
```
return42, line 4: @inetol can you please validate the move from Redis to Valkey in the docker environment / thanks :+1:

return42 (approved):
I squashed all the commits into one and modified the commit message.  In my final review I removed some leftovers -> [diff](https://github.com/searxng/searxng/compare/553d66069c1bd598cc1b1dbfbfce7873936f900e..36538e6a20d51a716f2a2df9cf266d3cc3d7af32)

inetol (approved):
I haven't faced any issues upgrading from the sxng image with Redis lib to Valkey. The old ENVs are respected, so as far as I'm concerned the container part works correctly.

Bnyro (approved):

PR #4799, opened 2025-05-17
[mod] container: build custom base images

inetol:
Instead of using Wolfi base images from `cgr.dev` and making that mess on the Dockerfile, why don't we build the base images ourselves from Wolfi repos with [`apko`](https://github.com/chainguard-dev/apko)? The intention of this is to simplify the main Dockerfile and avoid having to patch the base image every time, it also simplifies some steps like image ownership management and provides extremely fast builds.

Latest CI: https://github.com/inetol/searxng/actions/runs/15084778493

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -5,7 +5,7 @@ name: Cleanup
 on:
   workflow_dispatch:
   schedule:
-    - cron: "4 4 * * 0"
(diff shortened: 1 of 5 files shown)


inetol (commented):
This PR is ready, after the container.yml workflow finishes, the `ghcr.io/searxng/base` image visibility needs to be public, otherwise people won't be able to build their own container.

PR #4674, opened 2025-04-26
[upd] Upgrade to httpx 0.28.1

dalf:
## What does this PR do?

Upgrade httpx to the last version 0.28.1

## Why is this change important?

Use the last version of httpx which is a core dependency of SearXNG

## How to test this PR locally?

* [x] check `enable_http: false` works 
* [x] check `enable_http2: true` works 
* [ ] check `verify: false` works
* [ ] check `verify: "/path/to/ca.pem"` works
* [ ] check `local_addresses` works
* [x] check HTTP proxy
* [x] check HTTPS proxy
* [ ] check SOCKS proxy
* [ ] Cipher list and rotation

PR #4801, opened 2025-05-17
httpx impersonate

dalf:
## What does this PR do?

Use [httpx-curl-cffi](https://github.com/vgavro/httpx-curl-cffi) to add a network parameter: `impersonate`.

See https://github.com/lexiforest/curl_cffi/blob/8b4ee6d4db0982a3d93191b698c56936dfdfdcf0/curl_cffi/requests/impersonate.py#L9-L77 for the possible values

See https://github.com/searxng/searxng/issues/3929

* [ ] check `enable_http2: true` works 
* [ ] check `enable_http2: false` works 
* [ ] check `verify: false` works
* [ ] check `verify: "/path/to/ca.pem"` works
* [ ] check `local_addresses` works
* [ ] check HTTP proxy
* [ ] check HTTPS proxy
* [ ] check SOCKS proxy
* [ ] check the qwant engine

## Author's checklist

Based on https://github.com/searxng/searxng/pull/4674

curl-cffi requires Python 3.10 or above and ARM64 or AMD64. With curl-cffi, the engines using the impersonate parameter crash with an explanation.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ python-dateutil==2.9.0.post0
 pyyaml==6.0.2
 httpx[http2]==0.28.1
 httpx-socks[asyncio]==0.10.0
+httpx_curl_cffi==0.1.3 ; python_version >= "3.10" and (platform_machine == "aarch64" or platform_machine == "arm64"  or platform_machine == "x86_64")
 Brotli==1.1.0
 uvloop==0.21.0
 setproctitle==1.3.6
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -148,6 +148,12 @@ def request(query, params):
         args['offset'] = (params['pageno'] - 1) * args['count']
 
     params['url'] = url + urlencode(args)
+    params['headers'] = {
+        "Accept": "application/json",
+        "Accept-Language": q_locale.replace("_", "-"),
+        "Referer": "https://www.qwant.com/",
+        "Origin": "https://www.qwant.com",
+    }
 
     return params
 
--- a/searx/network/client.py
+++ b/searx/network/client.py
@@ -11,7 +11,15 @@ from typing import Any, Dict
 import httpx
 from httpx_socks import AsyncProxyTransport
 from python_socks import parse_proxy_url, ProxyConnectionError, ProxyTimeoutError, ProxyError
+
 import uvloop
+
+try:
+    from httpx_curl_cffi import AsyncCurlTransport, CurlOpt, CurlHttpVersion
+except ImportError:
+    AsyncCurlTransport = None
+    CurlOpt = None
+    CurlHttpVersion = None
 
 from searx import logger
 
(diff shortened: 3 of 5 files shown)


unixfox:
I would use httpx_curl_cffi globally:
- Better for privacy, blends in with the mass of Chome users.
- Many engines probably already use JA3 identification for bot usage, at least all the ones using Cloudflare do. Specifying "network: chrome" on all the affected engines would be very time-consuming, at least you get it by default. This might solve the rate limits or CAPTCHA issues for some existing engines.

dalf:
> I would use httpx_curl_cffi globally

The package is not available on ARMv7 and Python 3.9.


[EDIT] it should compile on ARMv7 : https://github.com/lexiforest/curl_cffi/issues/304

unixfox:
> > I would use httpx_curl_cffi globally
> 
> The package is not available on ARMv7 and Python 3.9.

Yes but only if it's available. I'm talking about the customization done like these: https://github.com/searxng/searxng/pull/4801/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR1713-R1714

dalf:
Notes:
* If we replace the user agent by Chrome, some other engines might need some updates since the default user agent is FF.
* I've tried to use "impersonate: firefox", it didn't worked.

(I'm saying we should or should make it global)

dalf:
Looking at the benchmark, httpx-async is one of the slowest:
* [multiple_workers.csv](https://github.com/lexiforest/curl_cffi/blob/main/benchmark/multiple_workers.csv). Look for `httpx_async` vs `curl_cffi_async`.
* [single_worker.csv](https://github.com/lexiforest/curl_cffi/blob/main/benchmark/single_worker.csv). There is no async version.

Should we be worried that a single, commercially backed company is driving the project and could one day switch off the updates, requiring SearXNG to migrate to a pure httpx client?

dsmith2001:
> Notes:
> 
> * If we replace the user agent by Chrome, some other engines might need some updates since the default user agent is FF.
> 
> * I've tried to use "impersonate: firefox", it didn't worked.
> 
> 
> (I'm saying we should or should make it global)

Just a heads up if you're using e.g. **impersonate=chrome**: curl_cffi sets the user agent and the impersonated browser's headers for you. You can override them but it's something to keep in mind.

You can use something like httpbin's "anything" endpoint to see what's being sent: 
- https://httpbin.org/#/Anything for headers
- https://tls.browserleaks.com/json for server side fingerprints.

> Should we be worried that a single, commercially backed company is driving the project and could one day switch off the updates, requiring SearXNG to migrate to a pure httpx client?

The project is open source and can be forked, it's not like it's some super obscure project either so I am not too worried.

return42:
Thanks a lot, PR .. 

- https://github.com/searxng/searxng/pull/5476

is based on the first commit of this PR.

Closing: superseded by #5476

PR #4807, opened 2025-05-20
[fix] Revert "fix: check if the browser supports Sec-Fetch headers (#4696)"

return42:
Many bots will probably use other user agents than the common ones (or rotate different ones)[1].

On my instance I can observe how bots with other UA headers are no longer stopped with the patch of PR #4696.

This reverts PR #4696 commit 19b116f1d7aaea2ab2b00868003e9dd852eabbe0.

[1] https://github.com/searxng/searxng/pull/4696#pullrequestreview-2806695481

----

To be clear, I understand the intention of #4696 and think the reasoning is correct.  However, I only see disadvantages for me in practice, which is why I am applying this patch to my instance and making it available to the community with this PR.

The PR is rebased in the event of conflicts with the master branch / it can simply be patched on top of your own instance via cherry picking.

### DIFF
--- a/searx/botdetection/http_sec_fetch.py
+++ b/searx/botdetection/http_sec_fetch.py
@@ -28,7 +28,6 @@ from ipaddress import (
     IPv6Network,
 )
 
-import re
 import flask
 import werkzeug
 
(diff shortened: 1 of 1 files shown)


unixfox:
How do you know that these are not legitimate users instead:
> Many bots will probably use other user agents than the common ones

?

Any insights to share? Screenshots or explanation about real data?

return42:
FYI last force push was just a rebase on current master to solve a conflict / related:

- https://github.com/searxng/searxng/pull/4962

Issue #4536, opened 2025-03-23
Yahoo News in different languages


anakinrpi:
**Working URL to the engine**

https://es-us.noticias.yahoo.com/

**Why do you want to add this engine?**

The yahoo news search engine provided only search english news
**Features of this engine**

Search yahoo news in spanish
**How can SearXNG fetch the information from this engine?**

**Applicable category of this engine**

News

**Additional context**

Bnyro:
In fact there are many more languages available, e.g. https://de.nachrichten.yahoo.com/.

It would help if somebody finds a list of all available news sites.

If they share the same layout, it should be easy to allow instance maintainers to configure the base url used as a workaround, that's not possible at the moment either looking at https://github.com/searxng/searxng/blob/master/searx/engines/yahoo_news.py#L41.

anakinrpi:
Thank you for the prompt response. If necessary, I can compile a list of the different Yahoo News sites in various languages, but I have noticed that regardless of the local news site being searched, if you prepend the country code to the URL of the results, it generates localized responses.

I found these subdomains working: co, th, ve, cl, hk, pe, ca, de, fr, tw, uk, br, in, espanol, ph, ar, mx, sg.

Examples: 

![Image](https://github.com/user-attachments/assets/63be692a-5144-46c9-856a-0e3a58d2b763)

![Image](https://github.com/user-attachments/assets/cda89c11-2fe6-439c-9ca7-f19c60df3bbb)

return42:
> I found these subdomains working: co, th, ve, cl, hk, pe, ca, de, fr, tw, uk, br, in, espanol, ph, ar, mx, sg.

And how does `es-us.noticias.yahoo.com` fits in?

anakinrpi:
I don't think it fits; if you search using the search box on https://es-us.noticias.yahoo.com/ or https://news.yahoo.com/ in your browser, it takes you to Yahoo's general search.
To see the news results, you need to click on the News tab.
That is, https://es-us.noticias.yahoo.com/ or https://news.yahoo.com/ is just a news portal—I don't find a search box exclusively for news.

My original approach is flawed; it wouldn't be necessary to incorporate https://es-us.noticias.yahoo.com/ but rather to explore the possibility of incorporating the alternatives I mentioned in the previous message.

return42:
> but rather to explore the possibility of incorporating the alternatives I mentioned in the previous message.

@anakinrpi PR

- #4826 

is ready for review.

PR #4814, opened 2025-05-20
[fix] yahoo: url and title xpath

useralias:
## What does this PR do?
Updates the url and title XPath for the Yahoo engine.

There were no results before because the URL XPath didn't match anything, this fixes it.

1. make run
2. !yh test

## Author's checklist
Tested on US IP. I don't have access to IPs from different regions so I would appreciate feedback in case it doesn't work in other regions.

## Related issues
None, as far as I'm aware.

### DIFF
--- a/searx/engines/yahoo.py
+++ b/searx/engines/yahoo.py
@@ -159,12 +159,12 @@ def response(resp):
 
     # parse results
     for result in eval_xpath_list(dom, '//div[contains(@class,"algo-sr")]'):
-        url = eval_xpath_getindex(result, './/h3/a/@href', 0, default=None)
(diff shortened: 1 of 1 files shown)


Bnyro:
Thanks for the PR, I noticed this issue too and planned to work on it in the next days, but seems like you've been faster 🎉

Bnyro (approved):
This does the job, thank you!

return42:
@useralias @Bnyro This PR works for the English language (`!yh test :en`) but breaks e.g. a german search query (`!yh test :de`).

- related: https://github.com/searxng/searxng/issues/4536#issuecomment-2754070816

PR #4815, opened 2025-05-20
[fix] typos

ddogfoodd:
## What does this PR do?

fix some typos

it's not that important, just keeps the code base clean


Bnyro (approved):
Thanks!

PR #4817, opened 2025-05-20
[refactor] icon engines: add new icon category

Bnyro:
- I think that this category makes sense because it allows to quickly search for free svg icons to use for websites / other designs with a quick `!icons` query
- they don't seem to fit into the normal images category that well because icons are quite a special type of images

## Author notes
- I probably forgot to add the category at some other places where it's needed

### DIFF
--- a/searx/engines/material_icons.py
+++ b/searx/engines/material_icons.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Material Icons (images)
-
-"""
+"""Material Icons (icons)"""
 
 import re
 from json import loads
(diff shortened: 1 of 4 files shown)


return42:
Should we add `icons` to .. ?

https://github.com/searxng/searxng/blob/7420706a5074803a78f3c795cf79427746e504e0/searx/settings.yml#L325-L335

If so, we should modify the category of the engines from `categories = ['images', 'icons']` to `['icons']` (or place icon in the first position `['icons', 'images']`)

Further we need an icon for ..

https://github.com/searxng/searxng/blob/7420706a5074803a78f3c795cf79427746e504e0/client/simple/theme_icons.js#L38-L67

https://github.com/searxng/searxng/blob/7420706a5074803a78f3c795cf79427746e504e0/searx/templates/simple/categories.html#L2-L17

----

However, we should add the new categories to i18n/l10n:

https://github.com/searxng/searxng/blob/7420706a5074803a78f3c795cf79427746e504e0/searx/searxng.msg#L25-L39

Bnyro:
> Should we add icons to .. ?
>
We only have 3 engines at the moment (4 with https://github.com/searxng/searxng/pull/4819), so I think that's not enough for adding it to the search bar.

It's like with the weather engines, and we don't have any additional category for it either in the search bar.

Or are there other reasons why we should add it as its own category to the category tabs?

return42 (approved):
> Or are there other reasons why we should add it as its own category to the category tabs?

Nope .. forget my comment :-)

PR #4819, opened 2025-05-20
[feat] engines: add uxwing engine for icons

Bnyro:
## What does this PR do?
- add https://uxwing.com as a new engine

- it provides attribution-free icons to use for design projects
- svgrepo was my go-to before, but it's ratelimiting a lot recently

- !ux tree

## Author's checklist
- this should also be added to the icons category if https://github.com/searxng/searxng/pull/4817 is merged


return42 (approved):
> this should also be added to the icons category

Done / last force push from my side.

PR #4820, opened 2025-05-20
[mod] container: replace uWSGI with Granian

### LINKED ISSUE #4802: Review of SearXNG's docker documentation
A lot of new things have happened in the area of ​​Docker builds in the last few days and weeks. Once all the relevant changes have been consolidated, the documentation should be revised or, if necessary, completely rewritten.

- https://docs.searxng.org/admin/installation-docker.html
- https://github.com/searxng/searxng/blob/master/docs/admin/installation-docker.rst

I created the documentation back then because there wasn't any, and there wasn't any sign of anyone being willing to write it. The problem then was, and still is, that I have no knowledge of the topic.

This time, however, it should be written by someone who does. If he/she have problems with the reST markup ([reST primer](https://docs.searxng.org/dev/reST.html#rest-primer)), I'm happy to correct them in the PR review. It would just be important that the content be written down in some form for the general public (like me).

### LINKED ISSUE #4818: change path configuration

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

Image

**Additional context**

sed: bad option in substitution expression
SearXNG 2025.5.20+7420706
...
... INFORMATION
... Update available for "/etc/searxng/uwsgi.ini"
... It is recommended to update the configuration file to ensure proper functionality
...
... New version placed at "/etc/searxng/uwsgi.ini.new"
... Please review and merge changes
...
...
... INFORMATION
... Update available for "/etc/searxng/settings.yml"
... It is recommended to update the configuration file to ensure proper functionality
...
... New version placed at "/etc/searxng/settings.yml.new"
... Please review and merge changes
...

### LINKED ISSUE #4894: 404 error Dockerfile  compose
I'm trying to found the Dockerfile  compose. But when I click the link I'm getting 404

inetol:
The configuration in Granian is handled with ENVs, much more convenient and practical for updating. The settings have been tested for over two months in a production instance, being usable on small to somewhat large instances without having to modify anything.

It also removes the patch functions and ENVs abstraction from the entrypoint, this makes it possible to run the container with immutable configuration.

In some setups, It may be desired to have the volumes/files under a specific uid/gid (other than searxng:searxng), if the entrypoint has root permissions it will chown automatically on every start, which may not be desired. Explicitly setting the new ENV `FORCE_OWNERSHIP=false` will prevent ownership from being modified.

Those changes are reflected in the documentation, which also received a rework.

No manual migration is necessary **unless** the user has changed the default uWSGI configuration or has a very specific setup.

Closes https://github.com/searxng/searxng/issues/4894
Closes https://github.com/searxng/searxng/issues/4818
Closes https://github.com/searxng/searxng/issues/4802

Supersedes #4596

Related https://github.com/searxng/searxng/discussions/4479

---

Latest CI: https://github.com/inetol/searxng/actions/runs/15582123713


return42:
FYI: I changed `Closes #4479` to `Related #4479` .. 

Docker users often forget: *"there is life outside the container"* ;-) .. we have shell scripts and documentation ..

---

sorry if I insist: in commit messages, I often miss the reference which is only given in the PR intro message. If a commit, for example, has a reference to a proposal or an issue, then this reference should also be included in the commit message.

Please read the description of [`[optional trailers]`](https://docs.searxng.org/dev/commits.html) again carefully; there are links to a few commits below as examples .. 

Basically, the commit message should contain all the information that the PR intro comment also has.

Since the commit message always exists before the PR, I always write everything in the commit message and then simply copy this message into the PR (GH already copies the commit message when creating the PR).

inetol:
> Docker users often forget: _"there is life outside the container"_ ;-) .. we have shell scripts and documentation ..

I'm not saying no, I'm now updating the docs and adding the new Granian section, but the scripts? Wouldn't it be more prudent to first *test* migrating the container users and then providing in another PR the migration for the rest of those who decide to do the *normal* installation route?

I can convert this PR to migrate the whole stack to Granian including all installation routes, but I'm afraid it's going to be quite a big change.

return42:
> I can convert this PR to migrate the whole stack to Granian including all installation routes, but I'm afraid it's going to be quite a big change.

No, this was not my intention .. these are different tasks and the scope of this PR is docker .. nothing more .. I just wanted to say "we can't close the feature request https://github.com/searxng/searxng/discussions/4479 before all tasks are done" .. sorry if I was unclear.

inetol:
Regarding https://github.com/searxng/searxng/pull/4891#discussion_r2123150872 ... This is the first revision of the documentation, can you check if it's OK, or should I modify something?

inetol:
Rebases on master to fix merge conflicts, no changes were made.

container/Dockerfile
```diff
@@ -46,14 +42,23 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
       org.opencontainers.image.version="$SEARXNG_GIT_VERSION"
 
 ENV SEARXNG_VERSION="$SEARXNG_GIT_VERSION" \
-    INSTANCE_NAME="SearXNG" \
-    AUTOCOMPLETE="" \
-    BASE_URL="" \
```
Bnyro, line 51: I assume that these variables should just be provided as runtime envs, e.g. `SEARXNG_BASE_URL`? I think we should also mention that in the `Environment variables` section of the newly added documentation, i.e. all env variables from https://github.com/searxng/searxng/blob/master/searx/settings_defaults.py#L179-L185.

I don't think we should copy all of them because we'd always need to keep the documentation in sync, but maybe we can reference that file somehow?

container/Dockerfile
```diff
@@ -46,14 +42,23 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
...
+    GRANIAN_PORT="8080" \
+    # https://github.com/emmett-framework/granian/issues/568
+    GRANIAN_HTTP1_KEEP_ALIVE="false" \
+    GRANIAN_WEBSOCKETS="false" \
+    GRANIAN_LOOP="uvloop" \
+    GRANIAN_BLOCKING_THREADS="2" \
+    GRANIAN_RUNTIME_THREADS="3" \
+    GRANIAN_RUNTIME_BLOCKING_THREADS="2" \
+    GRANIAN_RESPAWN_INTERVAL="0" \
+    GRANIAN_WORKERS_KILL_TIMEOUT="30" \
```
Bnyro, line 53:
```suggestion
    GRANIAN_WORKERS="1" \
    GRANIAN_WORKERS_KILL_TIMEOUT="30" \
```
I think it makes sense to add this for clarity, so that users can look through the envs here and can simply see how to set up more workers.

container/Dockerfile
```diff
@@ -46,14 +42,23 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
...
+    GRANIAN_PROCESS_NAME="searxng" \
+    GRANIAN_INTERFACE="wsgi" \
+    GRANIAN_HOST="::" \
+    GRANIAN_PORT="8080" \
+    # https://github.com/emmett-framework/granian/issues/568
+    GRANIAN_HTTP1_KEEP_ALIVE="false" \
+    GRANIAN_WEBSOCKETS="false" \
+    GRANIAN_LOOP="uvloop" \
+    GRANIAN_BLOCKING_THREADS="2" \
+    GRANIAN_RUNTIME_THREADS="3" \
```
Bnyro: I think this had previously set to `4` by default when we used uwsgi?

container/Dockerfile
```diff
@@ -46,14 +42,23 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
...
+    GRANIAN_RUNTIME_BLOCKING_THREADS="2" \
```
Bnyro: > blocking threads: the number of threads per worker interacting with the Python interpreter
>
> the default number of runtime threads and runtime blocking threads is fine for the vast majority of applications out there
>
The default seems to be 512 (https://github.com/emmett-framework/granian/blob/df0a44493488406b234838ae44dd8fc3ab1cc062/src/wsgi/serve.rs#L579). Can you explain why we set it to 2, please?

docs/admin/installation-granian.rst
```diff
@@ -0,0 +1,50 @@
...
+   pip install granian
+
+.. _Granian configuration:
+
+Configuration
+=============
+
+Granian can be configured via option parameters and environment variables.
+
+We provide sane defaults that should fit all use cases,
```
Bnyro:
```suggestion
We provide sane defaults that should fit most use cases,
```
I think that it makes sense to add that maintainers of large instances might need to adjust the settings a bit, I assume?

Bnyro (commented):
Looks awesome, can't wait to use it on my instance :)

I've left some small comments about the configuration, the general changes structure looks good to me though :)

container/Dockerfile
inetol: I would prefer not to make workers option visible, as it induces increasing that value to solve the pains of life by following the bad practices we have now with uWSGI. If the server runs short, increase threads, not workers.

I said it in another issue, but increasing workers doubles the resource usage and reduces the effectiveness of botprotection (placebo?). The option is there documented, you can crank it up whenever you want, but I see this as the *not the best solution* for improving performance.

container/Dockerfile
inetol: I have used the smallest possible values without affecting server performance. Architecturally, Granian is *faster* than uWSGI, so this change makes sense.

container/Dockerfile
inetol: I didn't see that it defaults to that many threads, I guess it's because of the synchronous nature of WSGI? I don't remember if there was another reason why I left this value so low, but I'll test this with 512 threads and see how it goes.

container/Dockerfile
Bnyro: > I would prefer not to make workers option visible, as it induces increasing that value to solve the pains of life by following the bad practices we have now with uWSGI. If the server runs short, increase threads, not workers.
>
In that case, please just mention that in the documentation as well, that's probably a good hint for people running large instances :)

docs/admin/installation-granian.rst
inetol: 3def38b22e0ff2b2614e30fb4c31412768767765

container/Dockerfile
```diff
@@ -46,14 +42,23 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
...
+    GRANIAN_RESPAWN_INTERVAL="0" \
+    GRANIAN_WORKERS_KILL_TIMEOUT="30" \
+    GRANIAN_BACKPRESSURE="1024" \
```
gi0baro: Granian maintainer here. After taking a look at the configuration, I'd rather suggest the following change:

```suggestion
    GRANIAN_BLOCKING_THREADS="4" \
    GRANIAN_WORKERS_KILL_TIMEOUT="30" \
```

Explanation:
- Given the app is WSGI, there's no need to change the Rust runtime threads configuration. Granian is fast enough to just use 1 runtime thread for serving http requests, as the bottleneck will always be the Python code in that regards.
- As you're using the static file serving, I won't change the number of the runtime blocking threads, as those threads will be used to read from disk and are managed in a dynamic pool by the runtime depending on the requests. Limiting those threads to such a low number will quite limit the throughput.
- With 1 worker, the respawn interval is noop, so you don't really need that. And, if you ever change the config to use more than 1 worker, I would still keep the default, as between spawning up the new one and killing the old one Granian would wait that amount of time. Having that delay to 0 means killing the previous worker even if the new one is not ready, and thus increasing the risk of dropping requests
- If you set the blocking threads, then I would leave the backpressure to Granian. Also, if you have 1 worker, that value is exactly the same of the backlog.

Also, keep in mind the number of blocking threads is as well affecting the maximum size of a dynamic pool, that will grow or shrink depending on the traffic. Given that, you might also want to tune the default `GRANIAN_BLOCKING_THREADS_IDLE_TIMEOUT` from 30s to a value that suits your needs more ;)

inetol: Thanks for the feedback. Regarding `GRANIAN_BLOCKING_THREADS_IDLE_TIMEOUT`, I understand that there might be an overhead when spawning a new thread, but is there anything wrong with keeping it alive for a long time? (e.g. 300 sec)

gi0baro: @inetol I'd say, since the load across the thread pool is usually quite balanced, such a high value basically mean Granian will ~never shrink the pool. Which can be a valid strategy, but really depends on the traffic shape. If you expect the traffic to stay at 0 for minutes and then spike up, it can be a good strategy. Otherwise, sounds a bit wasty to me :)

inetol: a7e1d5dd92ddc1e7b56b5b280d20f370ffefa500

container/Dockerfile
inetol: Resolved in https://github.com/searxng/searxng/pull/4820#discussion_r2134827964

container/Dockerfile
inetol: ca70b86c075330598f4aaf31c6fcf8a4459a15c7

container/Dockerfile
inetol: 8ae618f9949787cf9b7ad8ea023a708bf5ef0a4c

container/Dockerfile
```diff
@@ -46,14 +42,20 @@ LABEL org.opencontainers.image.authors="searxng <$GIT_URL>" \
...
     SEARXNG_SETTINGS_PATH="$CONFIG_PATH/settings.yml" \
-    UWSGI_SETTINGS_PATH="$CONFIG_PATH/uwsgi.ini" \
-    UWSGI_WORKERS="%k" \
-    UWSGI_THREADS="4"
+    GRANIAN_PROCESS_NAME="searxng" \
+    GRANIAN_INTERFACE="wsgi" \
+    GRANIAN_HOST="::" \
+    GRANIAN_PORT="8080" \
+    # https://github.com/emmett-framework/granian/issues/568
+    GRANIAN_HTTP1_KEEP_ALIVE="false" \
```
gi0baro: *and just one more thing:* I would not disable keep alive, as it might improve latency especially in browsers. You already set the kill timeout, so you should be fine on shutdowns. You might even lower down that value to match what's the maximum response time you expect plus some margin (I'd say if your typical response time is in ms order, forcefully closing open connections after 5 seconds of inactivity is perfectly fine).

inetol: I would prefer not to defer shutdown, what I don't know is whether setting `GRANIAN_WORKERS_KILL_TIMEOUT` would kill or gracefully stop the Python interpreter on timeout finish?

We don't care if we force close HTTP connections breaking active connections, but we don't want the interpreter to be SIGKILLed because other issues could arise.

gi0baro: Understood.
fyi, 2.4 *might* improve keepalived connections shutdown.

inetol:
Rebases on master to fix merge conflicts, no changes were made.

container/Dockerfile
inetol: https://github.com/searxng/searxng/pull/4820/commits/cb15e70d20b42b3d9150f702798f34a476adfa7a

return42 (approved):
looks awesome :+1: .. I placed a WIP commit on top, we can squash before merge this PR .. In the WIP commit there are only minor changes to the documentation .. no significant changes.

return42:
@inetol your last force push removes some (small) modifications from my WIP commit, see [docs/admin/installation-granian.rst](https://github.com/searxng/searxng/compare/ed60969333190764f61a2002276d18e0257c0bbc..be60bc9f436112aebbdd26a6964fbeb7c4d0077f#diff-8da272bbd8581cb9da2989b03d68d611e8f8adc17b85040752a16c67737cdcb6) .. these are only nits, but why did you removed them?

inetol:
> your last force push removes some (small) modifications from my WIP commit, see [docs/admin/installation-granian.rst](https://github.com/searxng/searxng/compare/ed60969333190764f61a2002276d18e0257c0bbc..be60bc9f436112aebbdd26a6964fbeb7c4d0077f#diff-8da272bbd8581cb9da2989b03d68d611e8f8adc17b85040752a16c67737cdcb6) .. these are only nits, but why did you removed them?

It looks redundant, since we are already on Granian page. When making a reference from somewhere else, it is seen with the anchor name and not the header title, isn't it?

https://github.com/searxng/searxng/pull/4820/files#diff-36f30ffc7390620fd192a0b7d0eef5de8cf30c04e160d0da73c4d6b68d0c742aR158-R159

Bnyro (approved):
I've been using this on my instance for the past week and didn't encounter any issues, so lgtm in that regard :+1:

return42:
> It looks redundant, since we are already on Granian page. When making a reference from somewhere else, it is seen with the anchor name and not the header title, isn't it?

```rest
lorem ipsum `Granian options`_

.. _Granian options: https://github.com/emmett-framework/granian/blob/master/README.md#options
```

will be rendered to:

lorem ipsum  [Granian options](https://github.com/emmett-framework/granian/blob/master/README.md#options)

----

```rest

lorem ipsum :ref:`Granian configuration`

.. _Granian configuration:

Config
======
...

```

will be rendered to header title ..

lorem ipsum [Config](example.org)

----

To verify your modifications, use `make docs.live` .. it will build the docs and open them in your WEB browser.

----

If you don't want to see the header title you can use a notation like this ..

```rest
lorem ipsum :ref:`Granian config <Granian configuration>`
```

lorem ipsum [Granian config](example.org)


A short introduce can be found in our primer: [Anchors & Links](https://docs.searxng.org/dev/reST.html#anchors-links)

return42:
@inetol  after a cleanup of the git history (squash the last 3 commits into the 2 first commits) this PR is ready to merge I think .. will you squash and merge .. or should I do it?

inetol:
~I'll do it, we need to make an announcement in Matrix room and https://github.com/searxng/searxng/discussions/categories/announcements~

GH: https://github.com/searxng/searxng/discussions/4972
Matrix: https://matrix.to/#/!vxScbLNEAmRvOraXBn:matrix.org/$bWCBghIP7vUUC_XPi2HGB70H3_-zCFWiEoNVs4o6T0E?via=matrix.org&via=tchncs.de&via=envs.net

PR #4826, opened 2025-05-21
[mod] engines: Yahoo in different languages

return42:
BTW fix issue reported in [1]

[1] https://github.com/searxng/searxng/pull/4814#issuecomment-2896948787

-----

This patch supports different languages / regions .. by example try `!yh bmw :es-CL` and `!yh bmw :es-MX` .. the BMW branches in Chile (CL) and Mexico (MX).  Compare with `!yh bmw :es`. Try `!yh bmw :zh-TW` ..

### DIFF
--- a/searx/engines/yahoo.py
+++ b/searx/engines/yahoo.py
@@ -44,6 +44,29 @@ time_range_dict = {
     'month': ('1m', 'm'),
 }
 
+region2domain = {
+    "CO": "co.search.yahoo.com",  # Colombia
+    "TH": "th.search.yahoo.com",  # Thailand
+    "VE": "ve.search.yahoo.com",  # Venezuela
+    "CL": "cl.search.yahoo.com",  # Chile
+    "HK": "hk.search.yahoo.com",  # Hong Kong
+    "PE": "pe.search.yahoo.com",  # Peru
+    "CA": "ca.search.yahoo.com",  # Canada
+    "DE": "de.search.yahoo.com",  # Germany
+    "FR": "fr.search.yahoo.com",  # France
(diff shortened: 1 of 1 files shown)


searx/engines/yahoo.py
```diff
@@ -157,14 +183,22 @@ def response(resp):
     results = []
     dom = html.fromstring(resp.text)
 
+    url_xpath = './/h3/a/@href'
+    title_xpath = './/h3//a/@aria-label'
```
Bnyro, line 186: I was just about to create a PR with the exact same yahoo engine (not news) changes :/

searx/engines/yahoo.py
```diff
@@ -157,14 +183,22 @@ def response(resp):
     results = []
     dom = html.fromstring(resp.text)
 
+    url_xpath = './/h3/a/@href'
```
Bnyro:
```suggestion
    url_xpath = './/div[contains(@class,"compTitle")]/h3/a/@href'
```
The `compTitle` class is also used everywhere for the localized pages from my tests, so it might be more robust to add it here too. The thing that's mainly different is the ordering of `h3 > a` vs `a > h3`

Bnyro (approved):

Bnyro:
This PR is not related to Yahoo News though, it's only about the general Yahoo search, right (contrary to the PR title)?

return42:
> This PR is not related to Yahoo News though, it's only about the general Yahoo search, right (contrary to the PR title)?

truly, you are right .. I mixed "News" & "Search" :facepalm: 

I'll clean up the PR when I get a chance / sorry.

PR #4828, opened 2025-05-21
[mod] container: remove -e flag

inetol:
Temporarily remove the -e flag from `set` to prevent entrypoint.sh from stopping execution if any command returns a non-zero status. This doesn't solve anything but relaxes the script checks.

Related https://github.com/searxng/searxng/issues/4818

### DIFF
--- a/container/entrypoint.sh
+++ b/container/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 # shellcheck shell=dash
-set -eu
+set -u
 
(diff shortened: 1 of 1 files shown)


return42 (approved):
LGTM

Bnyro (approved):
I think that the "merge new incoming changes" method shouldn't prevent the container startup at all if it's not able to create the `.new` file, but that's another discussion.

Fine as a temporary fix.

Issue #4093, opened 2024-12-05
Bug: pypi engine


dalf:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2024.11.22+67f754857

**How did you install SearXNG?**

Docker image

**What happened?**

Any search leads to a parsing error ( `!pypi lxml` for example).

**How To Reproduce**

see above

**Expected behavior**

engine works

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineXPathException
  * Percentage: 35
  * Parameters: `('./h3/span[@class="package-snippet__version"]', 'index 0 not found')`
  * File name: `searx/engines/pypi.py:50`
  * Function: `response`
  * Code: `version = extract_text(eval_xpath_getindex(entry, './h3/span[@class="package-snippet__version"]', 0))`

Bnyro:
When you initially open the website without a session cookie set, you get to a Fastly page with some JavaScript that generates a session cookie, sets it and reloads the page (JavaScript must be disabled to test it).

It's unlikely that we can figure out how to generate session cookies ourselves...

dalf:
I've seen that, but pip/uv work.

* <https://pypi.org/simple/>
* <https://pypi.org/pypi/fasttext-predict/json> (for example)

I'm not sure how to make an engine from that.

Perhaps using https://github.com/searxng/searxng/blob/3742d558ac5fe756a62632fdbdf84f0d3ddb7de2/searx/network/__init__.py#L100C5-L100C19 ?

return42:
Thats really annoying ..  XML-RPC search is no longer supported and all query tools are broken:

```
$ pypi-search flask
[INFO] :: Searching for package `flask`...
Traceback (most recent call last):
  File "/tmp/env/bin/pypi-search", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/tmp/env/lib/python3.13/site-packages/pypi_search/main.py", line 35, in main
    version_info = pypi.get_version_info()
  File "/tmp/env/lib/python3.13/site-packages/pypi_search/utils.py", line 41, in get_version_info
    version_no = package_name_header.text.split()[1]
                 ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'text'
```

[...]

PR #4836, opened 2025-05-22
[mod] data: implement a simple currencies (SQL) database

return42:
To reduce the memory footprint, this patch no longer loads the JSON data completely into memory.  Instead, there is an SQL database based on `ExpireCacheSQLite`.

The class CurrenciesDB is a simple DB application that encapsulates the DB (queries and initialization) and provides convenient methods like `name_to_iso4217` and `iso4217_to_name`.

Related:

- https://github.com/searxng/searxng/discussions/1892
- https://github.com/searxng/searxng/pull/3458#issuecomment-2900807671
- https://github.com/searxng/searxng/pull/4650

### DIFF
--- a/searx/data/__init__.py
+++ b/searx/data/__init__.py
@@ -4,21 +4,17 @@
   make data.all
 
 """
-
+from __future__ import annotations
 
 __all__ = ["ahmia_blacklist_loader"]
 
 import json
-from pathlib import Path
 import typing
 
-from searx import logger
+from .core import log, data_dir
+from .currencies import CurrenciesDB
 
-log = logger.getChild("data")
-
-data_dir = Path(__file__).parent
-
-CURRENCIES: dict[str, typing.Any]
+CURRENCIES: CurrenciesDB
 USER_AGENTS: dict[str, typing.Any]
 EXTERNAL_URLS: dict[str, typing.Any]
 WIKIDATA_UNITS: dict[str, typing.Any]
@@ -29,7 +25,7 @@ ENGINE_TRAITS: dict[str, typing.Any]
 LOCALES: dict[str, typing.Any]
 
 lazy_globals = {
-    "CURRENCIES": None,
+    "CURRENCIES": CurrenciesDB(),
     "USER_AGENTS": None,
     "EXTERNAL_URLS": None,
     "WIKIDATA_UNITS": None,
@@ -41,7 +37,6 @@ lazy_globals = {
 }
 
 data_json_files = {
-    "CURRENCIES": "currencies.json",
     "USER_AGENTS": "useragents.json",
     "EXTERNAL_URLS": "external_urls.json",
     "WIKIDATA_UNITS": "wikidata_units.json",
@@ -63,6 +58,7 @@ def __getattr__(name):
         return data
 
     log.debug("init searx.data.%s", name)
+
     with open(data_dir / data_json_files[name], encoding='utf-8') as f:
         lazy_globals[name] = json.load(f)
 
--- a/searx/data/core.py
+++ b/searx/data/core.py
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
+from __future__ import annotations
+
+import pathlib
+
+from searx import logger
+from searx.cache import ExpireCacheCfg, ExpireCacheSQLite
+
+log = logger.getChild("data")
+
+data_dir = pathlib.Path(__file__).parent
+
+_DATA_CACHE: ExpireCacheSQLite = None  # type: ignore
+
+
+def get_cache():
+
+    global _DATA_CACHE  # pylint: disable=global-statement
+
+    if _DATA_CACHE is None:
+        _DATA_CACHE = ExpireCacheSQLite.build_cache(
+            ExpireCacheCfg(
+                name="DATA_CACHE",
+                # MAX_VALUE_LEN=1024 * 200,  # max. 200kB length for a *serialized* value.
+                # MAXHOLD_TIME=60 * 60 * 24 * 7 * 4,  # 4 weeks
+            )
+        )
+    return _DATA_CACHE
--- a/searx/data/currencies.py
+++ b/searx/data/currencies.py
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Simple implementation to store currencies data in a SQL database."""
+
+from __future__ import annotations
+
+__all__ = ["CurrenciesDB"]
+
+import json
+import pathlib
+
+from .core import get_cache, log
+
+
+class CurrenciesDB:
+    # pylint: disable=missing-class-docstring
+
+    ctx_names = "data_currencies_names"
+    ctx_iso4217 = "data_currencies_iso4217"
+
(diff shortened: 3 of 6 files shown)


dalf:
If I understand correctly the workflow is:

* keep the .json on the disk
* when searxng starts:
	* create CurrenciesDB() without any data
* when the first currency info is requested:
	* load the data from the .json to save them as SQLite temporary file, for all currencies --> the data are seen in memory and then garbage collected.
	* read one entry from the temporary SQLite file
* when further currency infos are requested:
	* read one entry from the temporary SQLite file


Why not store directly the SQLite files in the git repository, and then read directly these files:
* no issue with temporary files
* the DB files can be opened in read only mode
?

To be able to read the automatic updates, the update scripts can write a .csv files in addition to the SQLite file.

See #3458 for example implementation. This PR 3458 missed a check in production to make sure the memory footprint does not grow over time, then it went into limbo.

return42:
> If I understand correctly the workflow is:

:+1: yes, this is the "**current** workflow".

> Why not store directly the SQLite files in the git repository,

Thinkable .. *why not* ..  

- drawback: SQLite's `.db` file is a binary --> we no longer have a [diff](https://github.com/searxng/searxng/commit/1319b250af3e7049cb3e60495877630f99832ca4) which I often read before merge the [PR](https://github.com/searxng/searxng/pull/4740). 

----

On the long term I had a different solution in mind .. 

:thinking: .. the SQLite's `.db` file is from a **DB application** which was intended to be a key/value cache with *expire date* .. the name of the DB app is "ExpireCache" --> *from SearXNG's POV it is cache(!)*

- separate all data (the json files) from SearXNG core (repo `searxng/searxng`) into its own repository `searxng/data` .. include the [CI (data-update.yml)](https://github.com/searxng/searxng/blob/master/.github/workflows/data-update.yml) that updates theses JSON files

- In SearXNG core, instead of *read from a file* we have a *read from URL* e.g.:
 `https://raw.githubusercontent.com/searxng/data/refs/heads/master/<schema-version>/currencies.json`

We need to versioning the DB schema (`<schema-version>`) and the workflow will be slightly different to the "**current** workflow":  the tables in the DB will be maintained when the *expire date* has been reached (ExpireCache).

---

OT but to give an example: I implemented similar for the SVG weather icons in PR https://github.com/searxng/searxng/pull/4663#issuecomment-2889035367 .. 

- there is a [WEATHER_DATA_CACHE](https://github.com/searxng/searxng/blob/c64636fba53c1f49270147eb08a075f59d7bbe06/searx/weather.py#L36) with a [`MAXHOLD_TIME`](https://github.com/searxng/searxng/blob/c64636fba53c1f49270147eb08a075f59d7bbe06/searx/weather.py#L51)
- SVG files are loaded from [`YR_WEATHER_SYMBOL_URL`](https://github.com/searxng/searxng/blob/c64636fba53c1f49270147eb08a075f59d7bbe06/searx/weather.py#L39)

dalf:
> drawback: SQLite's .db file is a binary --> we no longer have a [diff](https://github.com/searxng/searxng/commit/1319b250af3e7049cb3e60495877630f99832ca4) which I often read before merge the https://github.com/searxng/searxng/pull/4740.

Look at #3458 which creates .csv files in addition to the .db files.
For example [currencies.csv](https://github.com/dalf/searxng/blob/c0a96d8ef8547bc7876cc75e409a0b4b8c656f82/searx/data/dumps/currencies.csv).

> separate all data (the json files) from SearXNG core (repo searxng/searxng) into its own repository searxng/data 

What are the benefits of a different repository?
What are the problem(s) with the current implementation which embed the data?

The data size is 8.6M, pretty much nothing if we look at the any web page.
No need for database versioning: code and data are always synchronized.
Also no error trying to load external data: there are already here locally.

>  from SearXNG's POV it is cache(!)

Is there an urge to merge this PR?

return42:
> Look at https://github.com/searxng/searxng/pull/3458 which creates .csv files in addition to the .db files.

Above you asked ..

> Why not store directly the SQLite file

When we replace .json by .cvs .. what have we won? / except a csv diff is hard to read, compared to json file with sorted keys ..

> What are the benefits of a different repository?

The JSON files are dumps of databases; in my opinion, maintaining these databases shouldn't be part of the code.

For example, take a look at the https://github.com/ClearURLs project, whose rules are also made available to the application via a database (a JSON dump https://github.com/ClearURLs/Rules ).

By separating them, the database could be maintained independently of the code.

But we don't have to separate it, if you prefer to maintain the DB (the JSON file) here in the code, then we'll leave it here (for now / we can decide it later).

> No need for database versioning: code and data are always synchronized.

As long we hold the DB here .. but when we separate DB from code, we need a versioning of the DB schema.

dalf:
>  except a csv diff is hard to read

Because this is simplest way to make a human readable SQL dump : mapping from one SQL row to one CSV line using `SELECT * FROM ... ORDER BY ...` so things don't jump around at each updates.

But why not .json, however without foreign keys, everything is flatten to rows.

> For example, take a look at the https://github.com/ClearURLs project

ClearURL is a browser extension where the users can pick different rules from different URLs.
There is no point in SearXNG to allow different URLs, as far I know?

In addition, all SearXNG instances are going to ping one more URL when they starts.

>  ... maintaining these databases shouldn't be part of the code.
> By separating them, the database could be maintained independently of the code.

I read again the messages, and feel sorry to insist: why we need to do that?

Currently if we change the format of a data file, it's simple:
* update the `searxng_extra/update` script
* run the script to update `searx/data`
* update the code in `searx/*` and `tests/*`
* one commit
* done

Everyone is going to pick the updates [1], no connectivity issue, no GH token, no CloudFlare protection, it's a simple life.
[1] one git pull or container pull, and everything is here, up to date.

Most probably I miss your point.

return42:
> I probably don't understand your point.

... we can stick with the current procedure, which is the *simplest* solution for us developers, as you already said.

As I already said, we can consider separation later if we ever have compelling reasons to do so. There's no need to decide that now.

> Because this is the simplest way to make a human-readable SQL dump:

A dump, yes, but the diff is hard to read, and a diff is what we have to read with every PR review.

Anyway, I'd like to stick with JSON files for now... we can decide later whether it might be better to switch to a CVS file for one or another data source.


> however without foreign keys, everything is flattened to rows.

For most cases, that's sufficient, and the ExpireCache we currently have in our code is sufficient for that. Regarding external bangs, I need to reconsider whether we need a different DB schema than a simple key/value cache (a DB table with more than just one primary key)... but I can't judge that at this point... perhaps this data is unsuitable for a relational DB... we'll see.

dalf:
> perhaps this data is unsuitable for a relational DB... we'll see.

See https://github.com/searxng/searxng/issues/4093#issuecomment-2815336107

Can we close this PR?

return42:
> > perhaps this data is unsuitable for a relational DB... we'll see.
> 
> See [#4093 (comment)](https://github.com/searxng/searxng/issues/4093#issuecomment-2815336107)
> 
> Can we close this PR?

No! ..why? and how is it related to a bug in the pipy engine .. I'm confused.

More PRs will follow for the other data objects.  This is just one to demonstrate.

return42:
> More PRs will follow for the other data objects. This is just one to demonstrate.

I need this patch for further development, since no technical defects could be found I will now merge this PR.

Issue #4824, opened 2025-05-21
Bug: duckduckgo engine, Error: SearxEngineCaptchaException


Finkregh:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2025.5.20+7420706

**How did you install SearXNG?**

searxng-docker / docker-compose

**What happened?**

duckduckgo search broken

**How To Reproduce**

not sure

**Screenshots & Logs**

```text
2025-05-21 09:39:40,840 WARNING:searx.engines.duckduckgo: ErrorContext('searx/engines/duckduckgo.py', 367, 'raise SearxEngineCaptchaException(suspended_time=0, message=f"CAPTCHA ({resp.search_params[\'data\'].get(\'kl\')})")', 'searx.exceptions.SearxEngineCaptchaException', None, ('CAPTCHA (us-en)',)) False
2025-05-21 09:39:40,840 ERROR:searx.engines.duckduckgo: CAPTCHA
Traceback (most recent call last):
  File "/usr/local/searxng/searx/search/processors/online.py", line 160, in search
    search_results = self._search_basic(query, params)
  File "/usr/local/searxng/searx/search/processors/online.py", line 148, in _search_basic
    return self.engine.response(response)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/local/searxng/searx/engines/duckduckgo.py", line 367, in response
    raise SearxEngineCaptchaException(suspended_time=0, message=f"CAPTCHA ({resp.search_params['data'].get('kl')})")
searx.exceptions.SearxEngineCaptchaException: CAPTCHA (us-en), suspended_time=0
```

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineCaptchaException
  * Percentage: 100
  * Parameters: `('CAPTCHA (us-en)',)`
  * File name: `searx/engines/duckduckgo.py:367`
  * Function: `response`
  * Code: `raise SearxEngineCaptchaException(suspended_time=0, message=f"CAPTCHA ({resp.search_params['data'].get('kl')})")`

return42:
From time to time ddg blocks the server IP by a CAPTCHA challenge / I can't reproduce the issue ..  please try to [answer CAPTCHA from server’s IP](https://docs.searxng.org/admin/answer-captcha.html) .. and report your experience here / thanks.

useralias:
[...]

PR #4837, opened 2025-05-22
[refactor] duckduckgo engine: improve request logic and code structure

useralias:
## What does this PR do?
Mainly reworks the engine's request parameters and cookies to better match a real request and avoid CAPTCHAs.

Changes:
- Adds trailing slash to base URL to prevent potential redirects
- Removes advanced search syntax filtering (no longer guarantees a CAPTCHA)
- Corrects pagination offset calculation: Page 2 now starts at offset 10, subsequent pages use 10 + (n-2)*15 formula instead of the previous broken 20 + (n-2)*50 calculation that caused CAPTCHAs
- Restructures request parameter building to better match a real request
- "kt" cookie is no longer an empty string if the language/region is "all"
- Groups related parameter assignments together
- Adds header logging to debugging output

Requesting the next page or searching with advanced syntax no longer instantly triggers a CAPTCHA.

1. make run
2. !ddg test
3. Test languages (all, english, etc)
4. Test pagination (page 1, page 2, page 3+)
5. Test time filter (day, week, month, year)
6. Test advanced syntax

## Author's checklist
Worked fine in my testing, but may need to be battle tested.

## Related issues
https://github.com/searxng/searxng/issues/4824

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -58,7 +58,7 @@ paging = True
 time_range_support = True
 safesearch = True  # user can't select but the results are filtered
 
-url = "https://html.duckduckgo.com/html"
+url = "https://html.duckduckgo.com/html/"
 
 time_range_dict = {'day': 'd', 'week': 'w', 'month': 'm', 'year': 'y'}
 form_data = {'v': 'l', 'api': 'd.js', 'o': 'json'}
@@ -248,7 +248,6 @@ def quote_ddg_bangs(query):
 
 
 def request(query, params):
-
     query = quote_ddg_bangs(query)
 
     if len(query) >= 500:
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thanks for this PR!

I tested
- prefixes like `site:...`: works fine and no captcha
- paging (and compared results with ddg web, offset matches as you described)

So no complaints here. I also agree with moving the related params together, so overall looks like a good improvement, good to merge from my side :)

return42 (approved):
> Mainly reworks the engine's request parameters and cookies to better match a real request and avoid CAPTCHAs.

Tested and have to say *"well done"* .. thanks a lot :+1:

> Worked fine in my testing, but may need to be battle tested.

battle test is in production ;-) .. lets merge!

PR #4850, opened 2025-05-25
[fix] search: autocomplete focus on results page

### LINKED ISSUE #4846: Bug: UI/UX issues with new autocomplete
2025.5.23+230215c

Kubernetes with Docker container (minus entrypoint script, because it doesn't work for me). Likely not relevant for this report in any case.

**What happened?**

The flow of searching, then navigating results with hotkeys is impeded due to the search field being selected by default after a search is done.

Additionally, the autocomplete list now cannot be closed without reloading.

**How To Reproduce**

Do a search, and attempt to navigate to the results with the 'j' key (if using vim-like hotkeys). It instead types into the beginning of the search box. Autocomplete then pops up over the first result, and cannot be closed without reloading and/or searching again.

Similar issue happens with SearxNG hotkeys (left/right arrows) which also do not work because the search box is focused by default on the search results screen.

**Expected behavior**

- Search input field on the search results page should not be selected by default, so search result navigation hotkeys will work.
- The autocomplete list should close if both the autocomplete list and the search field lose focus, and/or the esc key is pressed.

**Additional context**

These problems appear to have been introduced in https://github.com/searxng/searxng/commit/32823ecb69b115a6726475d6421f0a1c0327fafa but I have not bisected to confirm.

Bnyro:
## What does this PR do?

This has been a regression introduced with the removal of the unmaintained autocomplete.js library.

We should only focus the search bar on the main search page at `/` and not at the results page located at `/search`.

I'm not sure if there's a better way to figure out if we're on the results page than checking if the id of the main element is `#main_results`, checking the path obviously isn't a better solution because it can differ depending on the instance / reverse proxy / ....

## Related issues
- related to https://github.com/searxng/searxng/commit/32823ecb69b115a6726475d6421f0a1c0327fafa
- closes https://github.com/searxng/searxng/issues/4846


return42 (approved):
> a better way to figure out if we're on the results page than checking if the id of the main element

I don't know a different solution .. this one does work -> OK!

PR #4851, opened 2025-05-25
[mod] container: allow immutable config

inetol:
Remove the patch functions and ENVs abstraction from the entrypoint, this makes it possible to run the container with immutable configuration required in some setups.

In other setups, it may be required to have the volumes/files under a specific uid/gid (other than searxng:searxng), if the entrypoint has root permissions will correct this, which may not be desired. Explicitly setting the new ENV `FORCE_OWNERSHIP=false` will prevent ownership from being modified.

There are also some minor consistency changes.

Related: https://github.com/searxng/searxng/issues/4818

---

Latest CI: https://github.com/inetol/searxng/actions/runs/15237554045


unixfox:
Thanks inetol, but isn't that doing double work with this PR: https://github.com/searxng/searxng/pull/4820? With this PR #4820, aren't you going to achieve immutable config too?

inetol:
But that PR is not ready yet and it might take a couple more weeks to be, I'll rebase the changes there. For now, let's focus on what we have now.

unixfox:
I would prefer to not change around the GHA every day, it's becoming quite difficult to follow.

We have a temporary solution from https://github.com/searxng/searxng-docker/pull/375, we know it's not great, hence why I created a PR to revert https://github.com/searxng/searxng-docker/pull/385. But it works for now.

I would rather spend more time in testing granian (https://github.com/searxng/searxng/pull/4820) than introducing changes that will get obsolete at some point with granian.

inetol:
All right, I'll merge the changes there https://github.com/searxng/searxng/pull/4820

PR #4857, opened 2025-05-26
[chore] il post: disable by default

Bnyro:
- the engine only provides italian results and thus shouldn't be part of the news tab by default


Bnyro:
Sorry for requesting reviews from everybody, but apparently I require two approvals for this game breaking change...

tiekoetter (approved):

inetol (approved):

Bnyro:
Thanks for the approvals 👍

PR #4858, opened 2025-05-27
[fix] searx.network: cookies being reset when more than one requests

### LINKED ISSUE #4854: Presearch returns nothing: json.decoder.JSONDecodeError

9ed9a9aa53a17df476437fb16c24c44e7c345f56

Installed using the following commands:
```bash
$ git clone https://github.com/searxng/searxng.git
$ cd searxng
$ python3 setup.py build
$ searxng-run
```
**What happened?**

I noticed Presearch engine was returning this error: `json.decoder.JSONDecodeError`. At first, I thought something changed at their end, but with the help of `git bisect` was able to track down when it stopped working: 9ed9a9aa53a17df476437fb16c24c44e7c345f56

I tested a bit, but couldn't exactly wrap my head around why it is showing nothing. It might have something to do with the headers mismatch upon getting the result id, and retrieving the result from the id. Because unlike many other engines, Presearch requires 2 HTTP requests to do a complete search.

https://github.com/searxng/searxng/blob/9ed9a9aa53a17df476437fb16c24c44e7c345f56/searx/engines/presearch.py#L104-L146

I have tested replacing presearch `base_url` to webhook.site, checking if there is any difference commenting out the two lines in `network.py`:

https://github.com/searxng/searxng/blob/9ed9a9aa53a17df476437fb16c24c44e7c345f56/searx/network/network.py#L272-L273

I did not find any difference. Maybe it cares about the order in which the headers are stored in? Maybe wireshark might reveal some useful information? I don't know.

**How To Reproduce**

1. Be on commit 9ed9a9aa53a17df476437fb16c24c44e7c345f56 or later
2. Install and run SearXNG
3. Search for `!ps <you search here>`

Aadniz:
## What does this PR do?

This PR fixes an issue where cookies would be reset when an engine does more than 1 requests.

Implementation-wise, it checks if the cookie is not None, and only then sets the httpx cookie.

Search engines such like Presearch requires 2 requests to do a search, and also requires cookies from request 1 to be present during the 2nd request. Otherwise it gets blocked by Amazon WAF.

## Related issues

Closes #4854

### DIFF
--- a/searx/network/network.py
+++ b/searx/network/network.py
@@ -270,7 +270,8 @@ class Network:
         while retries >= 0:  # pragma: no cover
             client = await self.get_client(**kwargs_clients)
             cookies = kwargs.pop("cookies", None)
-            client.cookies = httpx.Cookies(cookies)
+            if cookies:
+                client.cookies = httpx.Cookies(cookies)
             try:
                 if stream:
                     response = client.stream(method, url, **kwargs)


Bnyro:
Thanks for looking into this!

I don't think that we should generally be keeping the client cookies from the request because this way cookies would also be kept across different search requests of the same engine which could be
- suboptimal in terms of privacy and 
- could have some negative impacts on some other engine implementations that spoof the be a completely new request / session, e.g. to avoid auth

I've opened an alternative PR at https://github.com/searxng/searxng/pull/4860.

Aadniz:
Alright, I did not think of that. Thank you

Bnyro:
Please test/review https://github.com/searxng/searxng/pull/4860 if you have some time :)

PR #4873, opened 2025-05-29
add Ukrainian flag instead of globe in  sxng_locales.py

zodmy:


return42:
Its a language not a region.

Issue #4847, opened 2025-05-24
CI: automated tasks (gh-actions, dependabot., ..) should not run on clones


return42:
The automated tasks should only run on the origin, not on other GitHub clones.

TL;DR; Currently, the jobs are designed as opt-out: a clone must edit its master branch to disable them. It should be an opt-in, where the clone actively chooses to perform certain automated tasks.

---

The jobs/workflows/bots are only implemented and tested for the master branch of the searxng/searxng repository. On other clones, they encounter errors or perform actions that are generally not desired on the clone.

For example, the dependabot is activated on the master branch of a clone and generates unwanted messages there. A well-maintained master branch usually follows the master branch of its origin and receives its updates from it --> On a clone, dependabot creates various branches and PRs that you don't want there.

Building the documentation and deploying it on GH-docs of a clone is generally not desirable either --> We have > 2k clones, but we only need one up-to-date documentation and that is always only in the master branch of the searxng/searxng repo.

----

*By the way, all the GitHub Actions running on all the clones GitHub hosts is a gigantic waste of energy... as far as it's in our power, we should limit the use of resources to what we think is helpful.*

*I know, in light of AI, saving GH actions doesn't seem to be important... but we can all make our small contribution and use resources carefully.*

unixfox:
Github actions does not run by default on forks: https://github.com/orgs/community/discussions/31889#discussioncomment-3566755

It will however run on clones but that's probably a good thing if the user that cloned the repo wants to maintain a separate code from searxng.

Forks = when you click on fork button on github
Clones = when you create a new repo and pull/push manually the changes from another repo.

In insights section of the repo, you see 2k forks not 2k clones. You can not see clones from this part, except by actively searching for them using github search.

I do not see any changes needed, GitHub already did the job at preventing the actions to run on many forks.

inetol:
Some workflows and steps will not run if they are not in this organisation, see [data-update.yml](https://github.com/searxng/searxng/blob/743f24d8c6b277191a915ef6f1558d6c33ee72c7/.github/workflows/data-update.yml) for example, but it is true that we can cut down more on what runs in the forks.

In fact, we should make all workflows in forks run only if they are triggered by "workflow_dispatch". Not only is it annoying, but as you say, it's a huge waste of resources since most of the fork authors won't bother about GHA.

inetol:
> Github actions does not run by default on forks: https://github.com/orgs/community/discussions/31889#discussioncomment-3566755

[...]

PR #4875, opened 2025-05-30
[mod] CI: exclude SearXNG checker and build & deploy of online docs

return42:
checker.yml

1. The checker is not yet of sufficient quality to allow the results of the check to be evaluated / we do not evaluate them ourselves.

2. The checker sends hundreds of requests to the search engines and causes problems there / we either overload small providers or we train their bot defenses to use the SearXNG signature.

documentation.yml

Building the documentation and deploying it on GH-docs of a clones (GH forks) is generally not desirable either --> We have >2k clones, but we only need one up-to-date documentation and that is the one from the master branch of the searxng/searxng repo.

If search engines like Google start linking to the documentation in the clones, SearXNG users may no longer find the original documentation or be lost in the flood of options.

Related:

- https://github.com/searxng/searxng/issues/4847

-------

About checker: 

I like the idea of ​​an automated engine check, but the current implementation and framework still raise questions that need to be clarified:

1. There is currently no reasonable evaluation of the tests.

2. It needs to be examined whether the GH-CI network is suitable for such checks: many service providers block requests from the GH-CI network, which would make the test results unrepresentative.

--> So it would be worth considering whether we should deactivate this checker on our repo as well until these questions/problems are resolved

.. meanings?

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   search:
+    if: github.repository_owner == 'searxng' || github.event_name == 'workflow_dispatch'
     name: Search
     runs-on: ubuntu-24.04-arm
     steps:
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -23,6 +23,7 @@ env:
 
 jobs:
   release:
+    if: github.repository_owner == 'searxng' || github.event_name == 'workflow_dispatch'
     name: Release
     runs-on: ubuntu-24.04-arm
     permissions:


inetol:
> If search engines like Google start linking to the documentation in the clones, SearXNG users may no longer find the original documentation or be lost in the flood of options.

I didn't see any other documentation page except ours on any search engine, as you need to explicitly configure the branch and enable the option to publish the site on GitHub.

> --> So it would be worth considering whether we should deactivate this checker on our repo as well until these questions/problems are resolved

Honestly I don't understand the use of the checker, since it also fails running locally. I don't think it is a good idea to rely on network requests to other engines to fulfil our tests.

return42:
> I didn't see any other documentation page except ours on any search engine, as you need to explicitly configure the branch and enable the option to publish the site on GitHub.

Sure? .. I see a lot of GitHub pages are build in the last days ..

- https://github.com/tiekoetter/searxng/tree/gh-pages
- https://github.com/kp-forks/searxng/tree/gh-pages
- https://github.com/LovelyGuYiMeng/SearXNG/tree/gh-pages
- ... 
- more can be found in this list: https://github.com/searxng/searxng/forks?include=active&page=1&period=2y&sort_by=last_updated

> Honestly I don't understand the use of the checker, since it also fails running locally. I don't think it is a good idea to rely on network requests to other engines to fulfil our tests.

Don't mix "code tests" and "engine checker" .. 

The "code tests" are done in the integration workflows:

https://github.com/searxng/searxng/blob/37d851fb23935166ac7714a83c7273ce712bece0/.github/workflows/integration.yml#L57-L58

The intention of the "engine checker" ...

https://github.com/searxng/searxng/blob/37d851fb23935166ac7714a83c7273ce712bece0/.github/workflows/checker.yml#L45-L46

.. is to check (test) real requests from the engines. This is intended to ensure that the requests to the original search engine and the parsing of the results lists it returns are still working. In the event of an error, a bug fix could then be proactively implemented. Currently, we can only implement a bug fix if we receive a bug report from users.

At the current time, however, the implementation is insufficient, and there is also no overarching concept (who monitors the processes, who triggers a bug fix, there is no report, we can't compare test results from a past commit with the test results of the current code ..etc.). 

The "engine checker" was included in SearXNG years ago, but unfortunately, no one has had the time to fully develop this concept.

inetol:
> Sure? .. I see a lot of GitHub pages are build in the last days ..

Those are branches only, but nothing is published on GitHub Pages.

inetol (approved):
For the time being, we can lock this until we find a better solution.

return42:
> For the time being, we can lock this until we find a better solution.

OK, FYI I disabled the workflow --> https://github.com/searxng/searxng/actions/workflows/checker.yml

----

About checker issues see label [search.checker](https://github.com/searxng/searxng/issues?q=state%3Aopen%20label%3A%22search.checker%22)

PR #4884, opened 2025-05-31
[feat] engines: add pixabay for royalty free images/videos

Bnyro:
## What's changed?
- this PR adds Pixabay, a collection of royalty free images
- additionaly it seems to have some videos, so there's an engine for it too

## Author Notes
- when using SearXNG's transport, all our requests will get blocked, probably due to fingerprinting
- we should find an alternative solution because this is just a hacky change to make things work for now, but idk how ...

### DIFF
--- a/searx/engines/pixabay.py
+++ b/searx/engines/pixabay.py
@@ -1,0 +1,98 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Pixabay provides royalty-free media (images, videos)"""
+
+from datetime import timedelta
+from urllib.parse import quote_plus, urlencode
+from dateutil import parser
+from searx.utils import gen_useragent
+
+# about
+about = {
+    "website": 'https://pixabay.com',
+    "wikidata_id": 'Q1746538',
+    "official_api_documentation": 'https://pixabay.com/api/docs/',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+base_url = 'https://pixabay.com'
+categories = ['images']
+pixabay_type = "images"  # alternative: 'videos'
+
+paging = True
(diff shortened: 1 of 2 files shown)


searx/network/client.py
```diff
@@ -191,7 +191,7 @@ def new_client(
         event_hooks = {'response': [hook_log_response]}
 
     return httpx.AsyncClient(
-        transport=transport,
+        # transport=transport,
```
Bnyro: When using SearXNG's transport, all our requests will get blocked, probably due to fingerprinting. We should find an alternative solution because this is just a hacky change to make things work for now, but idk how ...
I've already spent a lot of time debugging to figure out that the 403 request forbidden responses are caused by the transport used here.

Bnyro: Okay, we have to disable http2 in order for this to work apparently. `params['enable_http2'] = False` has no impact on whether it's enabled in the end ...

Bnyro: One has to set `enable_http2 = False` in the engine as global variable, wow...

searx/engines/pixabay.py
```diff
@@ -0,0 +1,88 @@
...
+time_range_map = {'day': '1d', 'week': '1w', 'month': '1m', 'year': '1y'}
+
+# using http2 returns forbidden errors
+enable_http2 = False
+
+
+def request(query, params):
+    args = {
+        'pagi': params['pageno'],
+    }
```
return42, line 37: Pagination did not work (not in image search nor in video search)

return42 (changes requested):

searx/engines/pixabay.py
Bnyro: Fixed, thanks for the catch :+1:

Issue #4885, opened 2025-05-31
Migrate other weather engines (wttr, duckduckgo) to new weather engines template


Bnyro:
For reference, see 
- https://github.com/searxng/searxng/pull/4663
- https://github.com/searxng/searxng/blob/master/searx/engines/wttr.py
- https://github.com/searxng/searxng/blob/master/searx/engines/duckduckgo_weather.py

## Current progress
- [X] OpenMeteo
- [X] Wttr
- [x] DuckduckGo Weather

PR #4888, opened 2025-06-01
[feat] wttr.in: migrate to new weather engine template

Bnyro:
## Author Notes
- wttr.in provides 8 hourly time forecasts per day, I assumed that they're always describing the weather for 3 hours each, starting at 1 o'clock in the morning

## Related issues
- https://github.com/searxng/searxng/pull/4663
- https://github.com/searxng/searxng/issues/4885

### DIFF
--- a/searx/engines/wttr.py
+++ b/searx/engines/wttr.py
@@ -1,9 +1,11 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """wttr.in (weather forecast service)"""
 
-from json import loads
 from urllib.parse import quote
-from flask_babel import gettext
+from datetime import datetime
+
+from searx.result_types import EngineResults, WeatherAnswer
+from searx import weather
 
 about = {
     "website": "https://wttr.in",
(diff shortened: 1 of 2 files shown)


searx/engines/wttr.py
```diff
@@ -18,118 +20,117 @@
...
 
-        infobox += "</tbody></table>"
+    location = json_data["nearest_area"][0]
+    geoloc = weather.GeoLocation(
+        name=location["areaName"][0]["value"],
+        longitude=float(location["longitude"]),
+        latitude=float(location["latitude"]),
+        elevation="",
+        country_code="",
+        timezone="",
```
Bnyro: These parameters are not provided by Wttr.in. Do we actually need them though, I think they can be made optional, right?

searx/engines/wttr.py
```diff
@@ -18,118 +20,117 @@
...
+    "386": "rain showers and thunder",
+    "389": "heavy rain showers and thunder",
+    "392": "snow showers and thunder",
+    "395": "heavy snow showers",
+}
 
 
 def request(query, params):
     if query.replace('/', '') in [":help", ":bash.function", ":translation"]:
         return None
```
return42: I don't know what this is for. I suppose we can remove it in this revision.

searx/engines/wttr.py
```diff
@@ -18,118 +20,114 @@
...
+    "284": "light sleet",
+    "293": "light rain",
+    "296": "light rain",
+    "299": "heavy showers",
+    "302": "heavy rain",
+    "305": "heavy showers",
+    "308": "heavy rain",
+    "311": "light sleet",
+    "314": "light sleet",
+    "317": "light sleet",
```
return42: wttr.in has only a small set of symbols, a more detailed mapping is given in ..

- https://www.worldweatheronline.com/feed/wwoConditionCodes.txt
- https://www.worldweatheronline.com/weather-api/api/docs/weather-icons.aspx

searx/engines/wttr.py
return42: wttr.in internally uses geopy's [geocoders.Nominatim.geocode(..)](https://geopy.readthedocs.io/en/latest/#geopy.geocoders.Nominatim.geocode) which returns a [location.Location](https://geopy.readthedocs.io/en/latest/#geopy.location.Location) with the needed data .. unfortunately, wttr.in does not pass through this data :-(

The simple solution is to use the factory method [`searx.weather.GeoLocation.by_query()`](https://docs.searxng.org/src/searx.weather.html#searx.weather.GeoLocation.by_query).

On the long term we can consider to use [geopy](https://geopy.readthedocs.io/en/latest/#) as an alternative to our ...

https://github.com/searxng/searxng/blob/93751b63798bc9331987cef0e2dff17123a55325/searx/weather.py#L182-L184

searx/engines/wttr.py
Bnyro: Yeah, but wttr probably has a reason to only use these, the others probably never occur or are handled differently by wttr? :)

My motivation to rewrite the list again is equal to zero...

searx/engines/wttr.py
Bnyro: > The simple solution is to use the factory method [searx.weather.GeoLocation.by_query()](https://docs.searxng.org/src/searx.weather.html#searx.weather.GeoLocation.by_query).
>
That's an additional api call -> longer response time.

Do we actually need the elevation, country code and time zone of the location? I don't think we use them anywhere, right? So they can be removed?

> On the long term we can consider to use [geopy](https://geopy.readthedocs.io/en/latest/#) as an alternative to our ...
>
I agree, it definitely makes sense to become independent of OpenMeteo for that task.

return42:
FYI last force push was just a rebase on current master

searx/engines/wttr.py
return42: > That's an additional api call -> longer response time.

< 0.3 sec and the geo-location of the search term is cached .. I added the call and fixed some of the icon codes .. [diff](https://github.com/searxng/searxng/compare/master...return42:searxng:redis-to-valkey-migration)

return42 (approved):

PR #4891, opened 2025-06-03
[doc] fix typo and add missing help of the docker entrypoint script

return42:
### container/entrypoint.sh

before https://github.com/searxng/searxng/pull/4793 there was a -h option which is used to generate documentation from ..

https://github.com/searxng/searxng/blob/d63bdcd773b05d3a119cc0c710ba424d87172f93/docs/admin/installation-docker.rst?plain=1#L188-L190

### docs/dev/result_types/main_result.rst

Fix typo: MainResult is the type for the default.html template

### DIFF
--- a/container/entrypoint.sh
+++ b/container/entrypoint.sh
@@ -2,6 +2,46 @@
 # shellcheck shell=dash
 set -u
 
+help() {
+    cat <<EOF
+Command line:
+  -h  Display this help
+
+Environment variables:
+  SEARXNG_SETTINGS_PATH   path of the local settings.yml file
+  UWSGI_SETTINGS_PATH     path of the uwsgi.ini wfile
+  DATA_PATH:              default /var/cache/searxng
+  BIND_ADDRESS            default [::]:8080
+
+Environment variables used for settings.yml:
+  BASE_URL        -> server.base_url
+  INSTANCE_NAME   -> general.instance_name
+  AUTOCOMPLETE    -> search.autocomplete
+
+Environment variables used for uwsgi.ini:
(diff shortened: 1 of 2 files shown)


container/entrypoint.sh
```diff
@@ -2,6 +2,46 @@
...
+Environment variables used for settings.yml:
+  BASE_URL        -> server.base_url
+  INSTANCE_NAME   -> general.instance_name
+  AUTOCOMPLETE    -> search.autocomplete
+
+Environment variables used for uwsgi.ini:
+  UWSGI_WORKERS   -> workers
+  UWSGI_THREADS   -> threads
+EOF
+}
```
inetol, line 25: Why do we need a help menu in the entrypoint? I would prefer to document the available options (ENVs) in the documentation itself to avoid fragmentation.

return42: Whatever you prefer, I just *recreated* what existed before.

I wrote the Docker documentation [(installation-docker.rst)](https://github.com/searxng/searxng/blob/master/docs/admin/installation-docker.rst?plain=1) back then without much knowledge, because no one else had written it. As you know, I don't have any in-depth Docker knowledge... Would you mind revising the entire documentation? I'll help you with the [reST markup](https://docs.searxng.org/dev/reST.html) in your PR... and we'll close this PR?

inetol: Follow https://github.com/searxng/searxng/pull/4820#issuecomment-2935831273

return42:
Close this PR .. follow https://github.com/searxng/searxng/pull/4820#issuecomment-2935831273

PR #4893, opened 2025-06-03
[mod] container: revert to alpine

### LINKED ISSUE #4753: [Docker] Migrate from Debian to Ubuntu

### LINKED ISSUE #4792: [Docker] New image based on Wolfi OS breaks compatibility with x86_64-v1 CPUs
As reported [here](https://github.com/searxng/searxng/issues/4753#issuecomment-2885688182) the [new base image for Docker based on Wolfi OS](https://github.com/searxng/searxng/pull/4768) does not seem to be compatible with "old" x64 CPUs (< x86_64-v2) like Intel Core 2.

The following error is displayed in the logs when starting the container:
```
searxng-1  | Fatal glibc error: CPU does not support x86-64-v2
```

The first image to have this problem seems to be [2025.5.15-03083f0](https://hub.docker.com/layers/searxng/searxng/2025.5.15-03083f0/images/sha256-0863d2a2e352d1b4fe8ded34595572b93cfa42fbb28e79761932909f056f3d9a).

inetol:
I'm not too pleased to reverse this, but issues like https://github.com/searxng/searxng/issues/4792 have not been foreseen, and we can't just turn away. It has become apparent over the last weeks that there are still quite a few people with an incompatible CPU or having SearXNG on some random VM provider who can't (or won't) modify the configuration of their machines to expose the features needed for x86_64v2 march.

As I don't want to trash the work with apko and base images, I thought about trying building Alpine again now that we have all the container related workflow refactored.

There will still be the discussion of whether to use musl and its drawbacks, but right now I don't know any other alternatives.

The nice part of this is that both Dockerfiles (mainline and legacy) can now be unified under the same umbrella again.

Closes https://github.com/searxng/searxng/issues/4792
Closes https://github.com/searxng/searxng/issues/4753

---

Latest CI https://github.com/inetol/searxng/actions/runs/15423045981/job/43403194848

*Note: Build time has increased 2 minutes more for ARMv7 (uncached build: https://github.com/inetol/searxng/actions/runs/15422810016/job/43402274963)*


Bnyro (approved):
I haven't tested it (I believe you did, so that's fine), but I agree that it makes sense to switch back to Alpine given the mentioned issues.

PR #4895, opened 2025-06-05
[feat] Add ability to display a message of the day

benpiano800:
## What does this PR do?

This PR adds the ability to display a message of the day (MOTD) below the search bar on the homepage. 

Instance maintainers can use a MOTD to notify users about important information such as upcoming maintenance or new features. 

* Replace simple_search.html with the updated file
* Add `motd: ""` to the `brand` section of settings.yml
* Write your desired MOTD inside the quotation marks

### DIFF
--- a/searx/templates/simple/simple_search.html
+++ b/searx/templates/simple/simple_search.html
@@ -18,3 +18,5 @@
   <input type="hidden" name="theme" value="{{ theme }}" >
   {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
 </form>
+<br>
+{{ get_setting('brand.motd') }}
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -29,6 +29,8 @@ brand:
   #   links:
   #     Uptime: https://uptime.searxng.org/history/darmarit-org
   #     About: "https://searxng.org"
+  # Display a message of the day (MOTD) below the search bar
+  motd: ""
 
 search:
   # Filter results. 0: None, 1: Moderate, 2: Strict


searx/templates/simple/simple_search.html
```diff
@@ -18,3 +18,5 @@
   <input type="hidden" name="theme" value="{{ theme }}" >
   {% if timeout_limit %}<input type="hidden" name="timeout_limit" value="{{ timeout_limit|e }}" >{% endif %}
 </form>
+<br>
+{{ get_setting('brand.motd') }}
```
Bnyro, line 22: If we consider adding this, the motd should be wrapped inside its own HTML element (e.g. a `span`) and have an id/class, so that it's possible for users to override its appearance with custom css.

Bnyro (commented):
I'm not really sure if we actually need a MOTD, did you have some actual use case for it yet?

It's already possible to put custom text inside the footer.

benpiano800:
> I'm not really sure if we actually need a MOTD, did you have some actual use case for it yet?
> 
> It's already possible to put custom text inside the footer.

Right, I forgot about that. This may not be necessary then.

useralias:
image

searx.be added a thing for their homepage here, for me personally I think it would be useful for my own instance when settings are being disabled/enabled.

benpiano800:
> image
> 
> searx.be added a thing for their homepage here, for me personally I think it would be useful for my own instance when settings are being disabled/enabled.

That looks better than what my PR would do.

return42:
FYI searx.be is a fork of SearXNG with some patches on top ..

benpiano800:
> FYI searx.be is a fork of SearXNG with some patches on top ..

It is? Searx.space lists it as "vanilla."

return42:
> It is? Searx.space lists it as "vanilla."

The possibilities that searx.space has to verify an instance are very limited... therefore the assumptions that searx.space makes are often misleading, just like the misleading ranking of instances.

searx.be has always been a customized version of SearXNG.

> Right, I forgot about that. This may not be necessary then.

I close this PR / thanks anyway for your help in improving SearXNG :+1:

Issue #4907, opened 2025-06-11
`get_real_ip()` picks proxy IP


inetol:
It seems to be catching the last proxy before reaching SearXNG instance (at least in my case). I don't understand the reason for `x_for`, as far as I have seen it is only used to pick the client IP and this will always be the first pos of X-Forwarded-For header (if present).

https://github.com/searxng/searxng/blob/8888d71ab9391a8865959aa125cc7a1ae537f0b8/searx/botdetection/_helpers.py#L116

Example logs

![Image](https://github.com/user-attachments/assets/f7d650e8-992b-4c75-86e7-c818121d0968)

Reported https://github.com/inetol-infrastructure/searxng-container/issues/81

return42:
> this will always be the first pos of X-Forwarded-For header

Sadly not, the list of IPs in this HTTP header can be faked (pre set) by the client .. all other proxies usually just add their own IP at the end without validating the previous ones.

[Any security-related use of X-Forwarded-For (such as for rate limiting or IP-based access control) must only use IP addresses added by a trusted proxy. Using untrustworthy values can result in rate-limiter avoidance, access-control bypass, memory exhaustion, or other negative security or availability consequences.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#security_and_privacy_concerns)

Related:

- `limiert.toml`: [x_for --> Number of values to trust for X-Forwarded-For](https://docs.searxng.org/admin/searx.limiter.html#limiter-toml)
- SearXNG uses Werkzeug’s [ProxyFix](https://werkzeug.palletsprojects.com/middleware/proxy_fix/) (with it default x_for=1).
- [`searx.botdetection.get_real_ip()`](https://docs.searxng.org/src/searx.botdetection.html#searx.botdetection.get_real_ip)
- [The prerequisite for IP based methods is the correct determination of the IP of the client.](https://docs.searxng.org/admin/searx.limiter.html)
- [Make sure that X-Forwarded-For is passed correctly from the proxy, it must be a list of IPs, where the last IP is the proxy's IP and the second last (the first) is the client's IP. Some proxies transmit an X-Real-IP, this should then be trustworthy / please check whether this is the real IP of the client.](https://github.com/searxng/searxng/issues/3191#issuecomment-2202729527)

Open issues:

- https://github.com/searxng/searxng/issues/3632

inetol:
[...]

Issue #3632, opened 2024-07-02
Configurable `x_for` for both the WSGI app and for botdetection


rwjack:
Currently, only botdetection has a configurable `x_for` in `limiter.toml`.

In my use case of `client->proxy1->proxy2->searx`, I have to configure `x_for = 2`, for the limiter to work, which then creates an endless log spam of:
```
2024-07-02 14:58:00,377 WARNING:searx.botdetection: IP from WSGI environment (proxy2) is not equal to IP from X-Real-IP (client)
2024-07-02 14:58:02,093 WARNING:searx.botdetection: IP from WSGI environment (proxy2) is not equal to IP from X-Forwarded-For (client)
```
Warning found here: https://github.com/searxng/searxng/blob/39aaac40d63d53555b7fc12d56f0825c85ecf567/searx/botdetection/_helpers.py#L123-L124

---

As per @return42's recommendation:

> SearXNG uses the [ProxyFix](https://werkzeug.palletsprojects.com/en/3.0.x/middleware/proxy_fix/) with it default x_for=1 .. its implemented here in this line: https://github.com/searxng/searxng/blob/39aaac40d63d53555b7fc12d56f0825c85ecf567/searx/flaskfix.py#L77

Meaning the ideal solution would be to make the `x_for` parameter configurable in `settings.yml` for Searx, instead of `limiter.toml`, which would set the parameter for both the UWSGI app and for the limiter.

nowheretobefound:
I have the same issue with the log spam after setting limiter.toml x_for=2

```
WARNING:searx.botdetection: IP from WSGI environment (proxy2) is not equal to IP from X-Forwarded-For (client)
```

But I didnt understand  @return42's recommendation: Is there a workaround for this issue ( app.wsgi_app = ReverseProxyPathFix(ProxyFix(app.wsgi_app))  with default x_for=1) or  do we have to wait for a fix to get x_for=2 for the wsgi_app ?

return42:
> or do we have to wait for a fix to get x_for=2 for the wsgi_app 

Yes, sadly this is not implemented yet .. and this issue here is the request to implement it.

return42:
closed by https://github.com/searxng/searxng/pull/4911

@nowheretobefound @rwjack please have a look at the PR .. the `x_for` option is no longer needed/supported. If needed, you can add your proxies to the `botdetection.trusted_proxies` list in the [limiter.toml](https://docs.searxng.org/admin/searx.limiter.html#limiter-toml)

[...]

PR #4911, opened 2025-06-12
[mod] limiter: trusted proxies

inetol:
Replaces `x_for` functionality with `trusted_proxies`. This allows defining which IP / ranges to trust extracting the client IP address from X-Forwarded-For and X-Real-IP headers.

We don't know if the proxy chain will give us the proper client address, so we rely on reading the headers of the proxy before SearXNG (if there is one, in that case it must be added to trusted_proxies) hoping it has done the proper checks. In case a proxy in the chain does not check the client address correctly, integrity is compromised and this should be fixed by whoever manages the proxy, not us.

Closes:

- https://github.com/searxng/searxng/issues/4940
- https://github.com/searxng/searxng/issues/4939
- https://github.com/searxng/searxng/issues/4907
- https://github.com/searxng/searxng/issues/3632
- https://github.com/searxng/searxng/issues/3191
- https://github.com/searxng/searxng/issues/1237

Related:

- https://github.com/searxng/searxng-docker/issues/386
- https://github.com/inetol-infrastructure/searxng-container/issues/81

### DIFF
--- a/searx/botdetection/_helpers.py
+++ b/searx/botdetection/_helpers.py
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 # pylint: disable=missing-module-docstring, invalid-name
 from __future__ import annotations
+import typing as t
+
+__all__ = ["log_error_only_once", "dump_request", "get_network", "logger", "too_many_requests"]
 
 from ipaddress import (
     IPv4Network,
@@ -8,20 +11,19 @@ from ipaddress import (
     IPv4Address,
     IPv6Address,
     ip_network,
-    ip_address,
 )
 import flask
 import werkzeug
 
 from searx import logger
-from searx.extended_types import SXNG_Request
 
-from . import config
+if t.TYPE_CHECKING:
+    from . import config
 
 logger = logger.getChild('botdetection')
 
 
-def dump_request(request: SXNG_Request):
+def dump_request(request: flask.Request):
     return (
         request.path
         + " || X-Forwarded-For: %s" % request.headers.get('X-Forwarded-For')
@@ -52,86 +54,33 @@ def too_many_requests(network: IPv4Network | IPv6Network, log_msg: str) -> werkz
 
 
 def get_network(real_ip: IPv4Address | IPv6Address, cfg: config.Config) -> IPv4Network | IPv6Network:
-    """Returns the (client) network of whether the real_ip is part of."""
+    """Returns the (client) network of whether the ``real_ip`` is part of.
 
+    The ``ipv4_prefix`` and ``ipv6_prefix`` define the number of leading bits in
+    an address that are compared to determine whether or not an address is part
+    of a (client) network.
+
+    .. code:: toml
+
+       [botdetection]
+
+       ipv4_prefix = 32
+       ipv6_prefix = 48
+
+    """
+
+    prefix: int = cfg["botdetection.ipv4_prefix"]
     if real_ip.version == 6:
-        prefix = cfg['real_ip.ipv6_prefix']
-    else:
-        prefix = cfg['real_ip.ipv4_prefix']
+        prefix: int = cfg["botdetection.ipv6_prefix"]
     network = ip_network(f"{real_ip}/{prefix}", strict=False)
     # logger.debug("get_network(): %s", network.compressed)
     return network
 
 
-_logged_errors = []
+_logged_errors: list[str] = []
 
 
-def _log_error_only_once(err_msg):
+def log_error_only_once(err_msg: str):
     if err_msg not in _logged_errors:
         logger.error(err_msg)
         _logged_errors.append(err_msg)
-
-
-def get_real_ip(request: SXNG_Request) -> str:
-    """Returns real IP of the request.  Since not all proxies set all the HTTP
-    headers and incoming headers can be faked it may happen that the IP cannot
-    be determined correctly.
-
-    .. sidebar:: :py:obj:`flask.Request.remote_addr`
-
-       SearXNG uses Werkzeug's ProxyFix_ (with it default ``x_for=1``).
-
-    This function tries to get the remote IP in the order listed below,
-    additional some tests are done and if inconsistencies or errors are
-    detected, they are logged.
-
-    The remote IP of the request is taken from (first match):
-
-    - X-Forwarded-For_ header
-    - `X-real-IP header <https://github.com/searxng/searxng/issues/1237#issuecomment-1147564516>`__
-    - :py:obj:`flask.Request.remote_addr`
-
-    .. _ProxyFix:
-       https://werkzeug.palletsprojects.com/middleware/proxy_fix/
-
-    .. _X-Forwarded-For:
-      https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
-
-    """
-
-    forwarded_for = request.headers.get("X-Forwarded-For")
-    real_ip = request.headers.get('X-Real-IP')
-    remote_addr = request.remote_addr
-    # logger.debug(
-    #     "X-Forwarded-For: %s || X-Real-IP: %s || request.remote_addr: %s", forwarded_for, real_ip, remote_addr
-    # )
-
-    if not forwarded_for:
-        _log_error_only_once("X-Forwarded-For header is not set!")
-    else:
-        from . import cfg  # pylint: disable=import-outside-toplevel, cyclic-import
-
-        forwarded_for = [x.strip() for x in forwarded_for.split(',')]
-        x_for: int = cfg['real_ip.x_for']  # type: ignore
-        forwarded_for = forwarded_for[-min(len(forwarded_for), x_for)]
-
-    if not real_ip:
-        _log_error_only_once("X-Real-IP header is not set!")
-
-    if forwarded_for and real_ip and forwarded_for != real_ip:
-        logger.warning("IP from X-Real-IP (%s) is not equal to IP from X-Forwarded-For (%s)", real_ip, forwarded_for)
-
-    if forwarded_for and remote_addr and forwarded_for != remote_addr:
-        logger.warning(
-            "IP from WSGI environment (%s) is not equal to IP from X-Forwarded-For (%s)", remote_addr, forwarded_for
-        )
-
-    if real_ip and remote_addr and real_ip != remote_addr:
-        logger.warning("IP from WSGI environment (%s) is not equal to IP from X-Real-IP (%s)", remote_addr, real_ip)
-
-    request_ip = ip_address(forwarded_for or real_ip or remote_addr or '0.0.0.0')
-    if request_ip.version == 6 and request_ip.ipv4_mapped:
-        request_ip = request_ip.ipv4_mapped
-
-    # logger.debug("get_real_ip() -> %s", request_ip)
-    return str(request_ip)
--- a/searx/botdetection/config.py
+++ b/searx/botdetection/config.py
@@ -7,18 +7,31 @@ structured dictionaries.  The configuration schema is defined in a dictionary
 structure and the configuration data is given in a dictionary structure.
 """
 from __future__ import annotations
-from typing import Any
+import typing
 
 import copy
-import typing
 import logging
 import pathlib
 
 from ..compat import tomllib
 
-__all__ = ['Config', 'UNSET', 'SchemaIssue']
+__all__ = ['Config', 'UNSET', 'SchemaIssue', 'set_global_cfg', 'get_global_cfg']
 
 log = logging.getLogger(__name__)
+
+CFG: Config | None = None
+"""Global config of the botdetection."""
+
+
+def set_global_cfg(cfg: Config):
+    global CFG  # pylint: disable=global-statement
+    CFG = cfg
+
+
+def get_global_cfg() -> Config:
+    if CFG is None:
+        raise ValueError("Botdetection's config is not yet initialized.")
+    return CFG
 
 
 class FALSE:
@@ -57,7 +70,7 @@ class Config:
     UNSET = UNSET
 
     @classmethod
-    def from_toml(cls, schema_file: pathlib.Path, cfg_file: pathlib.Path, deprecated: dict) -> Config:
+    def from_toml(cls, schema_file: pathlib.Path, cfg_file: pathlib.Path, deprecated: dict[str, str]) -> Config:
 
         # init schema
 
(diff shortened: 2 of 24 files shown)


Bnyro (commented):
I agree that it only makes sense to check these headers if we're using a proxy.
- Otherwise, if somebody runs an instance without reverse proxy, attackers could just spoof faulty `X-Real-IP` headers. (I seriously hope that everybody uses a reverse proxy though)

searx/botdetection/ip_lists.py
```diff
@@ -4,21 +4,29 @@
...
 
 .. code:: toml
 
    [botdetection.ip_lists]
 
+   trusted_proxies = [
+     '127.0.0.1/32',    # IPv4 localhost
+     '::1',             # IPv6 localhost
+     '192.168.0.0/16',  # IPv4 private network
+   ]
```
Bnyro: https://datatracker.ietf.org/doc/html/rfc1918#section-3 declares 10.0.0.0/8 and 172.16.0.0/12 as well, they could probably be added here too?

Bnyro: Oh nvm, they're already added in `limiter.toml`.

Bnyro (dismissed):

searx/botdetection/ip_lists.py
inetol: Only localhost by default, If needed you can uncomment the rest of private ranges (just in case)

searx/botdetection/ip_lists.py
```diff
@@ -45,6 +53,18 @@
 """Passlist of IPs from the SearXNG organization, e.g. `check.searx.space`."""
 
 
+def trusted_proxies(remote_addr: IPv4Address | IPv6Address, cfg: config.Config) -> bool:
+    """Checks if the remote IP is in one of the members of the
+    ``botdetection.ip_lists.trusted_proxies`` list.
+    """
+
+    for net in cfg.get("botdetection.ip_lists.trusted_proxies", default=["127.0.0.1/32", "::1"]):
```
vojkovic: Loopback is 127.0.0.1/8 as per RFC 1122

searx/limiter.toml
```diff
@@ -21,6 +17,19 @@ link_token = false
 
 [botdetection.ip_lists]
 
+# If the request IP is in trusted_proxies list, the client IP address is
+# extracted from the X-Forwarded-For and X-Real-IP headers. This should be
+# used if SearXNG is behind a reverse proxy or load balancer.
+
+trusted_proxies = [
+  '127.0.0.1/32',
```
vojkovic: Same here

searx/botdetection/ip_lists.py
inetol: I think it would be 127.0.0.0/8 instead, but really someone uses any address other than 127.0.0.1 (any special use case)? It looks too broad to me to allow the full range by default?

vojkovic: Yes that's right - a typo from me. Sometimes I'll have two services on different IPs within the loopback subnet because I don't need to worry about port conflicts.

Everything in that /8 is unroutable and doesn't leave the host so there's no security risk.

inetol: 8b21fc6a11b8f661e27a1ed0e2d31ebb483db189

vojkovic (dismissed):

inetol:
Rebase to latest master, no changes made.

BTW this PR might require a migration for anyone using `x_for`:
> We have replaced `x_for` with `trusted_proxies` in [limiter.toml](https://github.com/searxng/searxng/blob/master/searx/limiter.toml). You can now configure which proxies to trust to get the client IP, which is necessary for botdetection and plugins like "self info".
> 
> If you want to know if everything is working after updating, with the "self info" plugin enabled, search "ip" in your instance, if your public IP appears then you are all set.
> 
> https://docs.searxng.org/admin/searx.limiter.html#limiter-toml
> https://docs.searxng.org/src/searx.botdetection.html#method-ip-lists

return42:
I think this PR will not solve the main issue ..

> Closes:
> - https://github.com/searxng/searxng/issues/3632

SearXNG uses the [ProxyFix](https://werkzeug.palletsprojects.com/en/3.0.x/middleware/proxy_fix/) with it default `x_for=1` .. its implemented here in this line: 

https://github.com/searxng/searxng/blob/d5743398539263f2c0cac676e073a08cb3eb6eb8/searx/flaskfix.py#L77

Meaning the ideal solution would be to make the `x_for` parameter configurable in `settings.yml` for SearXNG, instead of `limiter.toml`, which would set the parameter for both the uWSGI app and for the limiter.

inetol:
`trusted_proxies` is a full replacement of `x_for` and friends, not a thing that goes along with it.

I have added a commit cleaning the leftover.

searx/botdetection/config.py
```diff
@@ -182,6 +192,17 @@ def toml_load(file_name):
         raise
 
 
+def get_cfg() -> Config:
+    global CFG  # pylint: disable=global-statement
+
+    if CFG is None:
+        from searx import settings_loader  # pylint: disable=import-outside-toplevel
+
+        cfg_file = (settings_loader.get_user_cfg_folder() or pathlib.Path("/etc/searxng")) / "limiter.toml"
```
return42: Moving the `get_cfg` here seems obvious, but the limiter implementation shouldn't be tied directly to the `botdetection` package.

Background:

A long time ago, there was an intention to move [`botdetection`](https://github.com/searxng/botdetection) into a standalone Python package. Due to lack of time, this project was abandoned.

In the past, we therefore avoided binding botdetection directly to the limiter (as far as I remember, this worked, except for one minor detail).

That's for your information. I'll consider how this situation could possibly be improved in my review... but the simplest approach will probably be to leave the `get_cfg` in the limiter.

return42:
I pushed a `WIP` commit on top, with the suggestions from mine:

- Improve separation of limiter (cfg & Valkey) from botdetection.
- The `trusted_proxies` config is needed to determine the real IP --> moved option `botdetection.ip_limit.trusted_proxies` to `real_ip.trusted_proxies`
- Added obsolete `real_ip.x_for` to deprecated list.

The second commit `[fix] botdetection types: SXNG_Request -> flask.Request` is a little bit OT in context of this PR / However, it improves the separation of bot detection from the limiter.

searx/botdetection/_helpers.py
```diff
@@ -140,3 +158,21 @@ def get_real_ip(request: SXNG_Request) -> IPv4Address | IPv6Address:
 
     logger.debug("get_real_ip() -> %s", request_ip.compressed)
     return request_ip
+
+
+def is_trusted_proxy(remote_ip: IPv4Address | IPv6Address, cfg: config.Config) -> bool:
+    """Checks if the ``remote_ip`` is a member of one of the networks in the
+    ``real_ip.trusted_proxies`` list."""
+
+    trusted_proxy = cfg.get("real_ip.trusted_proxies", default=None)
```
inetol: I was looking at this and I noticed that the behavior had changed, I don't know exactly why.

In `limiter.toml`:
```toml
[real_ip]

trusted_proxies = [
  '10.0.0.0/8',
]
```

Making a search with "ip":
```console
DEBUG   searx.botdetection            : real_ip.trusted_proxies: ['127.0.0.0/8', '::1', '10.0.0.0/8']
DEBUG   searx.botdetection            : get_real_ip() -> 10.88.0.16
```

The debug response includes my range, but also loopback (?)

What's odd is that it didn't accept the proxy IP as trusted even it's inside the allowed range (?)

return42: > The debug response includes my range, but also loopback (?)

Yes, the configs are merged, but this not related to this PR, 

https://github.com/searxng/searxng/blob/b95a3e905d05695a04c424764b8c58020ca38b5c/searx/botdetection/config.py#L64-L65

https://github.com/searxng/searxng/blob/b95a3e905d05695a04c424764b8c58020ca38b5c/searx/botdetection/config.py#L72-L73

https://github.com/searxng/searxng/blob/b95a3e905d05695a04c424764b8c58020ca38b5c/searx/botdetection/config.py#L80

https://github.com/searxng/searxng/blob/b95a3e905d05695a04c424764b8c58020ca38b5c/searx/botdetection/config.py#L314-L315

This implementation of `config.py` dates back to a time before we had [msgspec](https://jcristharif.com/msgspec/structs.html); today I would probably implement something like this with `msgspec.Struct`.

return42: > What's odd is that it didn't accept my actual IP as trusted even it's inside the allowed range (?)

Not sure what you mean by / the ranges (the networks) in `trusted_proxies` are used for the proxies, not for the client IP.

inetol: I was referring to the range where the proxy is. I tested it again, and found that `is_trusted_proxy` remote_ip was `::ffff:10.88.0.135` and not `10.88.0.135`.

inetol: I've added a "fix", but I'm not entirely convinced that this needs to be done this way

https://github.com/searxng/searxng/pull/4911/commits/052bf0656db7d2b3ad70f41cd6c2a60e2c8f7a6f

return42: > I've added a "fix", but I'm not entirely convinced that this needs to be done this way  https://github.com/searxng/searxng/pull/4911/commits/052bf0656db7d2b3ad70f41cd6c2a60e2c8f7a6f

:thinking: may it needs more work .. look at line 157, https://github.com/searxng/searxng/commit/052bf0656db7d2b3ad70f41cd6c2a60e2c8f7a6f#diff-90808102c32e438beac1f927387d464c4ba15d70ebbff79ec0e3aa6b826dff19R157-R158 there is the same code for `request_ip` ..

If you have `::ffff:10.88.0.135` in the `request.remote_addr` .. I assume you have IPv4-Mapped Addresses also in the `X-Forwarded-For` and `X-Real-IP` headers / if so, we need to handle these IPv4-Mapped Addresses in the same way ..

return42:
> `trusted_proxies` is a full replacement of `x_for` and friends, not a thing that goes along with it.

Sure? 

> I have added a commit cleaning the leftover.

The old Werkzeug's `ProxyFix` class sets  `environ["REMOTE_ADDR"]` -->

https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/middleware/proxy_fix.py#L150-L152

By setting the environment, the `flask.Request.request.remote_addr` depends on the value of `x_for` .. as far I can see, this PR does not influence the value of `flask.Request.request.remote_addr` .. or I'm wrong?

inetol:
> By setting the environment, the `flask.Request.request.remote_addr` depends on the value of `x_for` .. as far I can see, this PR does not influence the value of `flask.Request.request.remote_addr` .. or I'm wrong?

Correct, but as I see that this will once again cause issues in the future, like https://github.com/searxng/searxng/pull/4448, I created a middleware to normalize the incoming `REMOTE_ADDR`. https://github.com/searxng/searxng/pull/4911/files#diff-75819cdbd4611ce01c94e44e6f99b8a0110b1cb91c9e1d20143f4d088926c189

I've added additional tests in case someone connects to WSGI server via socket, since the implementation was not handled correctly in those cases. https://github.com/searxng/searxng/pull/4911/files#diff-4ff2a4139508d0036ded86f1086e92adadfb74ba33bb9939038bcb9bc3fe1729

For the rest, I've cleaned up some verbose logging, squash everything and rebase to master. From what I have been able to test, I have not seen further issues.

searx/botdetection/_helpers.py
inetol: I assume that WSGI server or Flask does not modify the headers without explicit setup, so we should not alter them

return42:
I rebased and pushed a WIP commit on top ..

In the WIP I implemented a middleware `searx.botdetection.trusted_proxies.ProxyFix` which sets the correct client IP in the WSGI environment (`flask.Request.remote_addr`) .. by this there is no longer a need for a `get_real_ip` function .. 

I implemented a handful of tests / hope I have covered all variations .. Since we do not know whether the IPs in the environments are valid IPs (the WSGI servers usually only pass through strings), these are also checked by the middleware.

inetol:
I like the idea of moving this to the middleware, but wouldn't that mean it would no longer be specific to "botdetection"? Maybe we should move this to settings.yml instead as being webserver specific.

I added a commit fixing a couple of things. The addr mapping normalizations for origin and `X-Real-IP` header are missing (I know I made [this comment](https://github.com/searxng/searxng/pull/4911#discussion_r2230830074) a while ago, but I've changed my mind about this aproach while reviewing again).
There is also a case where the request failed because `x_forwarded_for` was out of range (if all addr are trusted and whitelisted), if everything is trusted, should return the first address *as expected* in https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#syntax

return42:
>  but wouldn't that mean it would no longer be specific to "botdetection"? 

From my POV it is specific to botdetection: Just as Flask offers a ProxyFix class (based on x_for), the botdetection implementation offers `botdetection.trusted_proxy.ProxyFix` class, which is based on a configuration named:

```toml
[botdetection]
trusted_proxies = [
  '127.0.0.0/8',
  '::1'
]
``` 

> Maybe we should move this to settings.yml instead as being webserver specific.

Sorry, not sure what you mean by "this" .. 

> I added a commit fixing a couple of things.

Thanks a lot .. much better than my first implementation :+1: 

> if everything is trusted, should return the first address

Good catch .. I added a test for this use case (and some other minor changes) in my WIP commit on top of your branch.

inetol:
> Sorry, not sure what you mean by "this" ..

The `trusted_proxies` setting, it's not specific to `limiter.toml` as is used in the "self information" plugin too (and it's a webserver setting). We could move it into `server` section inside `settings.yml`.

return42:
The "self information" is no longer based on the botdetection .. it now uses the [`flask.Request.remote_addr`](https://flask.palletsprojects.com/en/stable/api/#flask.Request.remote_addr) / [see](https://github.com/searxng/searxng/pull/4911/files#diff-c1ada1edd75a419a69ffd5835a3d07aae922a635ef09a0ffdf32e2c501ca3306R52-R55)

> The `trusted_proxies` setting, it's not specific to `limiter.toml`

The name of the config file for SearXNG's limiter is `limiter.toml` but in this file you find configuration to a tool named *botdetection* .. What matters is the namespace / not the filename .. here is the namespace:

```toml 
[botdetection]
ipv4_prefix =
ipv6_prefix = 
trusted_proxies =
[botdetection.ip_limit]
[botdetection.ip_lists]
```

Imagine if one day we integrate a tool called `foo`, then its namespace in the `limiter.toml` would be as follows:

```toml 
[foo]
opt = "this is option foo.opt"
[foo.bar]
opt = "this is option foo.bar.opt"
```

inetol:
OK, looks good to me now.

return42 (approved):

PR #4913, opened 2025-06-12
[mod] Hash plugin: List the available hash functions in the description

benpiano800:
## What does this PR do?

This PR lists the available hash functions/algorithms in the description of the hash plugin

Previously, it was not clear which hash functions are available.

Replace hash_plugin.py in your instance with the updated file.

### DIFF
--- a/searx/plugins/hash_plugin.py
+++ b/searx/plugins/hash_plugin.py
@@ -32,7 +32,9 @@ class SXNGPlugin(Plugin):
         self.info = PluginInfo(
             id=self.id,
             name=gettext("Hash plugin"),
-            description=gettext("Converts strings to different hash digests."),
+            description=gettext(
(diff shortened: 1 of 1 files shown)


searx/plugins/hash_plugin.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, plg_cfg: "PluginCfg") -> None:
         self.info = PluginInfo(
             id=self.id,
             name=gettext("Hash plugin"),
-            description=gettext("Converts strings to different hash digests."),
+            description=gettext("Converts strings to different hash digests. Available functions: md5, sha1, sha224, sha256, sha384, sha512."),
```
Bnyro:
```suggestion
            description=gettext(
                "Converts strings to different hash digests. Available functions: md5, sha1, sha224, sha256, sha384, sha512."
            ),
```
Please wrap this into a new line to fix the CI checks.

Bnyro (approved):
Thanks!

PR #4916, opened 2025-06-14
Update search.html

nandhana04cb:


tiekoetter:
Please fill out the form. JS and CSS changes should be done in their corresponding files and not inline.

PR #4918, opened 2025-06-15
[enh] sh: add shell formatter

### LINKED ISSUE #4803: Add shell (bash) formatter to SearXNG's tool chain
Like we have for python code `make format.python` we need a code formatter for shell scripts (bash) ..  (`make format.shell`).

> What I do use is https://github.com/mvdan/sh and it seems that by default it flattens the case branches, that's why the diff

_Originally posted by @inetol in https://github.com/searxng/searxng/pull/4707#discussion_r2079256799_
            
- https://github.com/mvdan/sh?tab=readme-ov-file#shfmt

inetol:
Related https://github.com/searxng/searxng/issues/4803

---

Closes https://github.com/searxng/searxng/issues/4803

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ search.checker.%: install
 	$(Q)./manage pyenv.cmd searxng-checker -v "$(subst _, ,$(patsubst search.checker.%,%,$@))"
 
 PHONY += test ci.test test.shell
-ci.test: test.yamllint test.black test.types.ci  test.pylint test.unit test.robot test.rst test.shell test.pybabel
-test:    test.yamllint test.black test.types.dev test.pylint test.unit test.robot test.rst test.shell
+ci.test: test.yamllint test.shfmt test.black test.types.ci  test.pylint test.unit test.robot test.rst test.shell test.pybabel
+test:    test.yamllint test.shfmt test.black test.types.dev test.pylint test.unit test.robot test.rst test.shell
 test.shell:
 	$(Q)shellcheck -x -s dash \
 		container/entrypoint.sh
@@ -82,8 +82,8 @@ MANAGE += gecko.driver
 MANAGE += node.env node.env.dev node.clean
 MANAGE += py.build py.clean
 MANAGE += pyenv pyenv.install pyenv.uninstall
-MANAGE += format.python
-MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.types.dev test.types.ci
+MANAGE += format.python format.shell
+MANAGE += test.yamllint test.pylint test.shfmt test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.types.dev test.types.ci
 MANAGE += themes.all themes.simple themes.fix themes.lint themes.test
 MANAGE += static.build.commit static.build.drop static.build.restore
 MANAGE += nvm.install nvm.clean nvm.status nvm.nodejs
--- a/manage
+++ b/manage
@@ -65,7 +65,7 @@ while IFS= read -r line; do
     if [ "$line" != "tests/unit/settings/syntaxerror_settings.yml" ]; then
         YAMLLINT_FILES+=("$line")
     fi
-done <<< "$(git ls-files './tests/*.yml' './searx/*.yml' './utils/templates/etc/searxng/*.yml' '.github/*.yml' '.github/*/*.yml')"
+done <<<"$(git ls-files './tests/*.yml' './searx/*.yml' './utils/templates/etc/searxng/*.yml' '.github/*.yml' '.github/*/*.yml')"
 
 RST_FILES=(
     'README.rst'
@@ -113,7 +113,6 @@ environment ...
 EOF
 }
 
-
 if [ "$VERBOSE" = "1" ]; then
     SPHINX_VERBOSE="-v"
     PYLINT_VERBOSE="-v"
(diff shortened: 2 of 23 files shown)


return42:
Thanks a lot :+1: /  I will review this PR hopefully soon. However, I (we) should review https://github.com/searxng/searxng/pull/4795 first.

utils/lib_sxng_test.sh
```diff
@@ -100,11 +101,17 @@ test.types.ci() {
 }
 
 test.black() {
-    build_msg TEST "[black] \$BLACK_TARGETS"
+    build_msg TEST "[black] ${BLACK_TARGETS[*]}"
```
return42, line 105: The dollar sign was deliberately escaped  here, output should be:

    TEST      [black] $BLACK_TARGETS

.. not the value of the environment variable.

----

update: now I see modification does not come from code formater .. may its OK what you did here (as `$BLACK_TARGETS` gives not really an information to the log).

manage
```diff
@@ -51,6 +51,14 @@ GECKODRIVER_VERSION="v0.35.0"
 # SPHINXOPTS=
 BLACK_OPTIONS=("--target-version" "py311" "--line-length" "120" "--skip-string-normalization")
 BLACK_TARGETS=("--exclude" "(searx/static|searx/languages.py)" "--include" 'searxng.msg|\.pyi?$' "searx" "searxng_extra" "tests")
+SHFMT_TARGETS=(
+    "./container/"
+    "./docs/"
+    "./searx/"
+    "./searxng_extra/"
+    "./tests/"
```
Bnyro, line 59: I don't think these directories contain any shell scripts as they're just python modules?

manage
```diff
@@ -226,11 +234,17 @@ pyenv.uninstall() {
...
 format.python() {
-    build_msg TEST "[format.python] black \$BLACK_TARGETS"
+    build_msg TEST "[format.python] black ${BLACK_TARGETS[*]}"
     pyenv.cmd black "${BLACK_OPTIONS[@]}" "${BLACK_TARGETS[@]}"
     dump_return $?
 }
 
+format.shell() {
+    build_msg TEST "[format.shell] shfmt ${SHFMT_TARGETS[*]}"
+    shfmt --list --write --apply-ignore --simplify "${SHFMT_TARGETS[@]}"
```
Bnyro, line 238: We could additionally pass `--indent 4` here for 4 spaces as indentation, because apparently it defaults to tabs and we use 4 spaces generally in SearXNG, as per https://github.com/searxng/searxng/blob/master/.editorconfig#L7

inetol: It applies the rules from .editorconfig when specifying [`--apply-ignore`](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#generic-flags) flag, so this is unneeded

manage
inetol: But what if it changes in the future and we place some shell script there? I would have preferred to exclude directories instead of including, but shfmt hasn't implemented this option.

manage
return42: @Bnyro @inetol 

> It applies the rules from .editorconfig when specifying [--apply-ignore](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#generic-flags) flag, so this is unneeded

this is a misunderstanding .. the docs say *"Always apply EditorConfig ignore rules."*

> I would have preferred to exclude directories instead of including, but shfmt hasn't implemented this option.

For this you can use `--apply-ignore` .. see example below:

Related:

- https://editorconfig.org/#file-format-details

----

from `man shfmt` ..

>  If any EditorConfig files are found, they will be used to apply formatting options. If any parser or printer flags are given to the tool, no EditorConfig formatting options will be used. A  default  like `-i=0` can be used for this purpose.

------

From: https://github.com/patrickvane/shfmt/blob/master/README.md

```
[*.sh]
# like -i=4
indent_style = space
indent_size = 4

shell_variant      = posix # like -ln=posix
binary_next_line   = true  # like -bn
switch_case_indent = true  # like -ci
space_redirects    = true  # like -sr
keep_padding       = true  # like -kp
function_next_line = true  # like -fn
never_split        = true  # like -ns

# Ignore the entire "third_party" directory.
[third_party/**]
ignore = true
```

----

We should try to put all definitions in the editorconfig file .. and as a call we should only use a simple `shfmt` (I have not been able to evaluate how far this is possible, but we should try to implement it roughly)

Makefile
```diff
@@ -50,8 +50,8 @@ search.checker.%: install
 	$(Q)./manage pyenv.cmd searxng-checker -v "$(subst _, ,$(patsubst search.checker.%,%,$@))"
 
 PHONY += test ci.test test.shell
-ci.test: test.yamllint test.black test.types.ci  test.pylint test.unit test.robot test.rst test.shell test.pybabel
-test:    test.yamllint test.black test.types.dev test.pylint test.unit test.robot test.rst test.shell
+ci.test: test.yamllint test.shfmt test.black test.types.ci  test.pylint test.unit test.robot test.rst test.shell test.pybabel
+test:    test.yamllint test.shfmt test.black test.types.dev test.pylint test.unit test.robot test.rst test.shell
```
return42, line 54: Most often we modify py files and the sh files are rarely modified, so their checking should be moved to the end (shortens the runtime when we change py files / in the most cases).

@inetol do not modify this branch / I'm in review and will leave a WIP commit covering my comments here on GH.

return42:
While reviewing this PR, I realized that our toolchain needed an overhaul beyond shfmt. Therefore, I implemented a completely new PR:

- https://github.com/searxng/searxng/pull/5018

inetol:
Superseded by https://github.com/searxng/searxng/pull/5018

PR #4930, opened 2025-06-21
remove dissem.in due to its closure

NanashiTheNameless:
## What does this PR do?

remove dissem.in as an option as it has shut down
https://association.dissem.in/dissemin-closure.html
https://web.archive.org/web/20250621200318/https://association.dissem.in/dissemin-closure.html

dissem.in is shut down and to avoid confusion it should be removed as an option.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -2771,7 +2771,6 @@ engines:
 doi_resolvers:
   oadoi.org: 'https://oadoi.org/'
   doi.org: 'https://doi.org/'
-  doai.io: 'https://dissem.in/'
   sci-hub.se: 'https://sci-hub.se/'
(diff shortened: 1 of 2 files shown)


inetol (changes requested):
Needs to be removed also from test suite

https://github.com/searxng/searxng/blob/e52e9bb4b699e39d9ce51874ea339d4773717389/tests/robot/settings_robot.yml#L54

NanashiTheNameless:
> Needs to be removed also from test suite
> 
> https://github.com/searxng/searxng/blob/e52e9bb4b699e39d9ce51874ea339d4773717389/tests/robot/settings_robot.yml#L54

Thanks for pointing that out! done!

inetol (approved):

Issue #4334, opened 2025-02-15
startpage autocompleter is broken


return42:
The URL (``https://startpage.com/suggestions``):
https://github.com/searxng/searxng/blob/d37dc307831ef0f16a6cb2fbb0c5afe9bb5e4258/searx/autocomplete.py#L186-L189

results in an json exception (response or URL might have changed):

```
INFO    werkzeug                      : 127.0.0.1 - - [15/Feb/2025 10:13:17] "POST /autocompleter HTTP/1.1" 500 -
Traceback (most recent call last):
...
  File "/share/searxng/searx/webapp.py", line 834, in autocompleter
    for result in search_autocomplete(backend_name, sug_prefix, sxng_locale):
                  ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/share/searxng/searx/autocomplete.py", line 279, in search_autocomplete
    return backend(query, sxng_locale)
  File "/share/searxng/searx/autocomplete.py", line 191, in startpage
    data = resp.json()
  File "/share/searxng/local/py3/lib/python3.13/site-packages/httpx/_models.py", line 756, in json
    return jsonlib.loads(self.text, **kwargs)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/home/markus/.asdf/installs/python/3.13.1/lib/python3.13/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/home/markus/.asdf/installs/python/3.13.1/lib/python3.13/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/markus/.asdf/installs/python/3.13.1/lib/python3.13/json/decoder.py", line 363, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Bnyro:
They added a mandatory `sc` query parameter which is embedded in the HTML page: (can be found with a regex like `/\?sc=(.*)\&` on the HTML page).

This would however require use to do two requests per fetched suggestion, so it'd be quite slow compared to others.

return42:
[...]

PR #4932, opened 2025-06-23
[fix] restore startpage autocompleter

useralias:
## What does this PR do?
Restores the deleted Startpage autocompleter.

Changes:
- Undo deletions of the autocompleter in settings and logic
- Add fixed autocomplete function in autocomplete.py

We can have Startpage autocomplete results again.

1. make run
2. Set autocomplete to **startpage**
3. Test supported languages (and check that request params match the proper **lui**)
4. Test unsupported language (and check that **lui=english**)

## Author's checklist
Uses languages from the supported languages found in the Firefox extension's files.

## Related issues
- https://github.com/searxng/searxng/issues/4334

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -20,7 +20,7 @@ from searx.engines import (
 )
 from searx.network import get as http_get, post as http_post
 from searx.exceptions import SearxEngineResponseException
-from searx.utils import extr
+from searx.utils import extr, gen_useragent
 
 
 def update_kwargs(**kwargs):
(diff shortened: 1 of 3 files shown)


searx/autocomplete.py
```diff
@@ -233,6 +234,55 @@ def sogou(query, _lang):
...
+        'sv': 'svenska',
+    }
+
+    base_lang = sxng_locale.split('-')[0]
+    lui = lang_map.get(base_lang, 'english')
+
+    # Needs user agent, returns a 204 otherwise
+    linux_os = 'X11; Linux x86_64'
+    latest_version = max(USER_AGENTS['versions'])
+    user_agent = USER_AGENTS['ua'].format(os=linux_os, version=latest_version)
```
Bnyro: using `gen_useragent` from `searx.utils` didn't work?

Bnyro:
Works as expected from my location too, awesome! :)

searx/autocomplete.py
useralias: It's not so much that it doesn't work, it's that rotating user agents while typing out a query isn't what a normal browser would do and we're trying to avoid bot-like behaviors.

Bnyro: I'm not sure if this is really needed though, we've been rotating user agents for years now without any specific issues. Normal networks have different clients too, so Startpage can't expect to always get the same user agent by the same IP (especially since CGNAT is a thing at many places).

If you have reason to believe this increases reliability, maybe we could move this to a helper method `gen_fixed_useragent` in `searx.utils` then as it might be useful to other engines too? (you can probably find a better method name :) )

useralias: > If you have reason to believe this increases reliability, maybe we could move this to a helper method [...] as it might be useful to other engines too?

It's really just for this case (to keep user agent consistency without tracking the user's query).

For engine search requests, it's fine to rotate user agents, but I think we should also make sure that the params belong to the browser we're spoofing (e.g. an engine making a token for X browser). Right now, out of the box the agent can be different when changing pages.

Bnyro: Yeah, I'm just not comfortable with hardcoding user agent parts in an engine specifically, I guess we'll wait for somebody else's opinion here.

return42 (changes requested):
> It's not so much that it doesn't work, it's that rotating user agents while typing out a query isn't what a normal browser would do and we're trying to avoid bot-like behaviors.

Unlike a search, autocompleters generally don't have a session... and any bot detection can't establish context via the IP, since it usually represents a subnet containing hundreds or thousands of browsers. --> Please remove the hard coding of the UA and replace it with `gen_useragent`... thanks :+1:

Bnyro (approved):
Looks good now, thank you!

PR #4941, opened 2025-06-26
[fix] container: remove HEALTHCHECK

### LINKED ISSUE #4722: OCI image format on containers
On PR https://github.com/searxng/searxng/pull/4707, Buildah will start building the future images for SearXNG and not Docker. Docker specific flags and keywords like [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) are excluded, this is because it's using the OCI image format to build the images. From our side it is not something to worry about as everything is handled on the container engine, Docker/Podman and friends can run OCI images without problems, in fact Docker is the founder of this initiative.

However, as I said before, the keyword [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) becomes unusable if you use the OCI format...
```console
WARN[0035] HEALTHCHECK is not supported for OCI image format and will be ignored. Must use `docker` format
```

**For the time being, the images will continue to be built with the “docker” format**, but we should see if we want to switch to the OCI image format. For the sake of standardization I prefer OCI, I think the health checks can be configured independently on container instantiating with Docker (or directly on Compose files, Kube...)

Let me know your opinion about this.

The spec is defined here: https://github.com/opencontainers/image-spec/blob/main/spec.md

### LINKED ISSUE #4906: Healthcheck port is hardcoded to 8080, not the server port or bind address

2025.6.9+d0bb977

using the docker image `docker.io/searxng/searxng:latest`

**What happened?**

Healthcheck is unhealthy, unless changed with

```yaml
  healthcheck:
    test: ["CMD", "wget", "--quiet", "--tries=1", "--spider", "http://localhost:9003/healthz"]
    interval: 30s
    timeout: 10s
    retries: 3
```

9003 is my server port.

**How To Reproduce**

use a custom port with

```yaml
environment:
  - BIND_ADDRESS=0.0.0.0:9003
```

or 

```yml
server:
  port: 9003
```

**Expected behavior**

Healthcheck should be on the same port as the configured port or bind address.

inetol:
This is a poorly designed instruction, which is hardcoded and cannot be easily modified or maintained on a rolling release sw like ours. This *should* be set in the SearXNG Docker Compose template, not in the image itself.

The OCI format is now used since we no longer have the [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) on the Dockerfile.

Closes https://github.com/searxng/searxng/issues/4906
Closes https://github.com/searxng/searxng/issues/4722

---

Latest CI https://github.com/inetol/searxng/actions/runs/15900692051

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -60,6 +60,4 @@ VOLUME $DATA_PATH
 
 EXPOSE 8080
 
-HEALTHCHECK CMD wget --quiet --tries=1 --spider http://localhost:8080/healthz || exit 1
-
 ENTRYPOINT ["/usr/local/searxng/entrypoint.sh"]
--- a/utils/lib_sxng_container.sh
+++ b/utils/lib_sxng_container.sh
@@ -93,8 +93,8 @@ container.build() {
         info_msg "Set \$GIT_BRANCH: $GIT_BRANCH"
 
         if [ "$container_engine" = "podman" ]; then
-            params_build_builder="build --format=docker --platform=$platform --target=builder --layers --identity-label=false"
-            params_build="build --format=docker --platform=$platform --layers --squash-all --omit-history --identity-label=false"
+            params_build_builder="build --format=oci --platform=$platform --target=builder --layers --identity-label=false"
+            params_build="build --format=oci --platform=$platform --layers --squash-all --omit-history --identity-label=false"
         else
             params_build_builder="build --platform=$platform --target=builder"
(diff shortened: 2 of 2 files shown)


vojkovic:
I agree, I've come to change my mind on this after reading https://github.com/opencontainers/image-spec/issues/749 - healthchecks should be for the orchestration layer. Thanks Ivan

vojkovic (approved):

PR #4952, opened 2025-06-28
[mod] theme/simple: improve fmt/lint tech

inetol:
This is one of various PR to refactor the simple theme internally.

Replace eslint tool with Biome. I have been using this for quite some time, and it will help us to have more consistent and *valid* code without extending on other third party plugins.

Removes unused dependencies.

---

Tested:
- [x] infinite_scroll
- [x] keyboard
- [x] mapresult
- [x] preferences
- [x] results
- [x] search
- [x] ...utils

Visual from Vite 6 baseline browsers:

<details>
<summary>Chromium 87</summary>

Search:
chrome87-search

Images:
chrome87-images

</details>

<details>
<summary>Firefox 78</summary>

Search:
firefox78-search

Images:
firefox78-images

</details>

<details>
<summary>Safari 14</summary>

Search:
safari14-search

Images:
safari14-images

</details>


Bnyro (commented):
`biome` only seems to support JavaScript, but we probably also want to have a formatter/linter for our style `.less` files. It doesn't seem to depend on `eslint`, so why did you remove it?

Bnyro:
I wasn't aware that there's something faster than eslint now, that's quite cool.

I'm still using eslint in some projects for JavaScript linting, so it's probably worth migrating seeing the performance benefits :)

We'll probably need some testing to see if the lints broke something unexpectedly, but looks good overall :+1:

client/simple/src/js/head/00_init.js
```diff
@@ -1,20 +1,19 @@
 /* SPDX-License-Identifier: AGPL-3.0-or-later */
-(function (w, d) {
-  'use strict';
-
+((w, d) => {
```
Bnyro, line 2: I'm not really a fan of using the arrow syntax instead of `function` here at the file declarations. Apparently after that, we can no longer reference the variable inside the method with `this` if I understand it correctly, possibly that could break something?

For reference: https://biomejs.dev/linter/rules/use-arrow-function/

Bnyro:
We'll probably need some testing to see if the lints broke something unexpectedly, but looks good overall :+1:

inetol:
> `biome` only seems to support JavaScript, but we probably also want to have a formatter/linter for our style `.less` files. It doesn't seem to depend on `eslint`, so why did you remove it?

Oops, I forgot about those files (thought were CSS compatible with Biome). I'll add back Stylelint.

client/simple/src/js/head/00_init.js
inetol: It could be problematic, but we don't use `this` in the entire codebase, I don't have an opinion about using functions or arrows so if you wish to keep the regular ones I'll override the rule.

Note that those IIFE will be converted into pure ESM in a *future* PR.

Bnyro (approved):

client/simple/src/js/main/keyboard.js
```diff
@@ -18,205 +17,218 @@ searxng.ready(function () {
...
     if (
-      Object.prototype.hasOwnProperty.call(keyBindings, e.key)
-        && !e.ctrlKey && !e.altKey
-        && !e.shiftKey && !e.metaKey
+      // biome-ignore lint/suspicious/noPrototypeBuiltins: FIXME: support for Chromium 93-87, Firefox 92-78, Safari 15.4-14
+      Object.prototype.hasOwnProperty.call(keyBindings, e.key) &&
+      !e.ctrlKey &&
+      !e.altKey &&
+      !e.shiftKey &&
+      !e.metaKey
```
return42, line 211: > FIXME: support for Chromium 93-87, Firefox 92-78, Safari 15.4-14

just for my knowledge .. what does "FIXME" here means? .. does it means we have to fix the code for the browser versions .. or does it mean, "this itself" is a "workaround" for those browser versions?

Makefile
```diff
@@ -83,7 +83,7 @@ MANAGE += py.build py.clean
 MANAGE += pyenv pyenv.install pyenv.uninstall
 MANAGE += format.python
 MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.types.dev test.types.ci
-MANAGE += themes.all themes.fix themes.test
+MANAGE += themes.all themes.simple themes.fix themes.lint themes.test
```
return42, line 86: FYI I added missing targets to the Makefile

return42 (approved):
Great work, I didn't know Biome before, but it looks promising... we should definitely try it out / thanks for the PR :+1:

client/simple/src/js/main/keyboard.js
inetol: Right now, we don't have a defined baseline of what minimum browser version is 100% compatible with snxg. I've put as FIXME as using hasOwnProperty brings doom and for compatibility reasons (listed versions) we are forced to use it.

I'm working on a better way to *transpile* things down which I hope to have ready this week.

PR #4962, opened 2025-06-30
[fix] sec-fetch-* headers cannot be verified for non-secure requests

return42:
Sec-fetch-* headers cannot be verified for non-secure requests


return42:
I tested this fix on my instance / https://github.com/return42/searxng/commits/darmarit.org/

PR #4988, opened 2025-07-08
[mod] theme/simple: migrate codebase to TypeScript

inetol:
Superseded by https://github.com/searxng/searxng/pull/5073


inetol:
I think I can now consider this ready for review, I list what is most important to see:

- The manifest that Vite generates (includes final bundles and metadata):

https://github.com/searxng/searxng/pull/4988/files#diff-d6634068e2ad39fc45bf65845d7028dd885fe7e49351fac767afbc9e4ee084e2

- The browser baseline:

https://github.com/searxng/searxng/pull/4988/files#diff-b9d144211e016923a961d806f7d1949bf8ca671e18f2b9872b5f6bef121fe014R20-R25

- The minimum supported version of Node.js, should not be a problem as it is handled by `nvm` but still:

https://github.com/searxng/searxng/pull/4988/files#diff-7ea01bc4f29b753f673d5d07ccc988162643e4e27d33e2e4113ea76081d89800R7

- The `mapresult`, I've replicated the features that Leaflet had (except map style switch):

https://github.com/searxng/searxng/pull/4988/files#diff-473131774f5a2c2c0df21ea7eec37068500025662182619b7d1e46135c157096

- `base.html` (the only HTML file I've touched):

https://github.com/searxng/searxng/pull/4988/files#diff-636e8c10c74c7cbe0b81b98aadf5174bfb6e580ce68099344f03f67c4b029bfd

---

As soon as you test the theme, open the console and keep the logs, if you see any errors or weird behaviour report back.

inetol:
Rebase to latest master, no changes made.

return42:
> LightningCSS now handles minification and prefix creation for CSS.

for my understanding .. does it mean, we can rename the `.less` files to `.css`?

inetol:
> for my understanding .. does it mean, we can rename the `.less` files to `.css`?

No, what I mean by prefix is, based on the CSS properties spec, add its equivalent for Mozilla (-moz) and WebKit (-webkit) depending on what browsers we support in the baseline.

To avoid extending more this PR, I haven't touched the CSS, but all the vendor specific added manually in the `.less` files can be removed, and it will be managed instead by LightningCSS.

return42:
> No, what I mean by prefix is, based on the CSS properties spec

Sorry, my question was not clear: With this PR we have transpilation done by lightningcss .. does it fully replace the LESS preprocessor (`css.preprocessorOptions.less`)?

inetol:
> Sorry, my question was not clear: With this PR we have transpilation done by lightningcss .. does it fully replace the LESS preprocessor (`css.preprocessorOptions.less`)?

Nop

```console
[plugin vite:css] /searxng/client/simple/src/less/style-rtl.less                                                                                                                                                     
Error: Preprocessor dependency "less" not found. Did you install it? Try `npm install -D less`.
```

return42:
> To avoid extending more this PR, I haven't touched the CSS, but all the vendor specific added manually in the `.less` files can be removed, and it will be managed instead by LightningCSS.

When we touch the less files .. should we convert to CSS? .. / Apparently we will never get a source map for the LESS files... and as far as I can see, less is not used that often anymore, at least not in the vite environment.

inetol:
> When we touch the less files .. should we convert to CSS? .. / Apparently we will never get a source map for the LESS files... and as far as I can see, less is not used that often anymore, at least not in the vite environment.

Take my opinion with a pinch of salt, but depending on how complex our styles are, we *might* prefer to move to [SCSS](https://sass-lang.com/documentation/syntax/#scss) instead of pure CSS. AFAIK, there are any issues generating sourcemaps there.

inetol:
Superseded by https://github.com/searxng/searxng/pull/5073

Issue #4481, opened 2025-03-11
Three dots/circles appearing in search results


FinallyCanuck:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
2025.3.11+3fe602a46

**How did you install SearXNG?**
Docker

**What happened?**
After updating container to latest image, search results return a series of three dots or circles at the end of each result.

**How To Reproduce**
Install the container image, use default preferences, perform any search.

**Expected behaviour**
Search results returned with no dots.

**Screenshots & Logs**
I have a screenshot but having trouble adding it here. I'll try adding as a comment to this issue.

**Additional context**
Also notice for the Social Media tab is treated. It has an icon and different font size compared to the others.

FinallyCanuck:
![Image](https://github.com/user-attachments/assets/8969701a-b32e-487d-bc14-fc567764491b)

ivoarch:
I have same issue after pull my container to latest release (2025.3.11+3fe602a46) .

vojkovic:
clear browser cache and reload.


duplicate of:
https://github.com/searxng/searxng/issues/4425
https://github.com/searxng/searxng/issues/4419
https://github.com/searxng/searxng/issues/4415

see [label:"clear browser cache"](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22clear+browser+cache%22)

ivoarch:
Thanks, fixed for me.

PR #5004, opened 2025-07-11
[mod] remove option ui.static_use_hash (cache busting)

return42:
Cache busting has caused serious problems for end users in the past, here are two examples:

- https://github.com/searxng/searxng/issues/4419
- https://github.com/searxng/searxng/issues/4481

And it makes development and deployment significantly more complex because it binds the client side to the server side:

- https://github.com/searxng/searxng/pull/4466

In the light of a decoupled development of the WEB clients from the server side:

- https://github.com/searxng/searxng/pull/4988

is it appropriate to abandon this feature. In fact,  it has been ineffective since #4436 anyway.

However, the benefit has always been questionable, since at best only a few kB of data are saved (at least in the context of an image_proxy, the effect is below the detection limit). Ultimately, the client is responsible for caching.

Related: https://github.com/searxng/searxng/issues?q=label%3A%22clear%20browser%20cache%22

Closes: 

- https://github.com/searxng/searxng/pull/4466
- https://github.com/searxng/searxng/issues/1326
- https://github.com/searxng/searxng/issues/964

### DIFF
--- a/searx/settings_defaults.py
+++ b/searx/settings_defaults.py
@@ -194,7 +194,6 @@ SCHEMA = {
     },
     'ui': {
         'static_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'static')),
-        'static_use_hash': SettingsValue(bool, False, 'SEARXNG_STATIC_USE_HASH'),
         'templates_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'templates')),
         'default_theme': SettingsValue(str, 'simple'),
         'default_locale': SettingsValue(str, ''),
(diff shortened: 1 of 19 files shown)


searx/settings_defaults.py
```diff
@@ -194,7 +194,6 @@ def apply_schema(settings, schema, path_list):
     },
     'ui': {
         'static_path': SettingsDirectoryValue(str, os.path.join(searx_dir, 'static')),
-        'static_use_hash': SettingsValue(bool, False, 'SEARXNG_STATIC_USE_HASH'),
```
inetol, line 197: Will the instance crash on start if someone still has this option defined in `settings.yml`?

return42: no

inetol (approved):

PR #5006, opened 2025-07-11
[fix] container: remove static path mount

inetol:
I've seen that by default Granian sets a `Cache-Control` header for 86400 seconds **IF `GRANIAN_STATIC_PATH_MOUNT` is set**. This option is not a hard requirement, but it's set because I found to be faster when serving the static files.

Another thing is that by removing `GRANIAN_STATIC_PATH_MOUNT`, the headers set by the application are present again in the static files, which was not the case before 🤦

Having the `GRANIAN_STATIC_PATH_MOUNT` set:
Screenshot From 2025-07-11 17-43-20

Without `GRANIAN_STATIC_PATH_MOUNT`:
Screenshot From 2025-07-11 17-43-48

Related https://github.com/searxng/searxng/pull/5004

---

Related with issue in https://github.com/searxng/searxng-docker/pull/410#discussion_r2200955441


inetol:
If there are no objections I will merge this PR in a few hours

return42 (approved):
> If there are no objections I will merge this PR in a few hours

Sorry for being such unresponsive .. LGTM, thanks for your deep analyses :+1:

gi0baro:
@inetol fyi, Granian 2.5.x allows to disable cache headers on static files.
If the headers thing was the only reason for this, re-enabling direct static files serving in Granian will for sure have better perf.

PR #5013, opened 2025-07-14
Fix/theme image phone portrait

mrpaulblack:
## What does this PR do?

Fix an issue where images are way to small on portrait phones when opening the image detail view. This is done by reducing the margin on the `max-hight` of the image for layouts > 50 rem display width (phones).

Fixes a display issue and is a minor change.

`make run`

open a image via phone portrait view.

## Author's checklist
Before:
image

With this PR:
image
image


Bnyro (approved):
Tested (on Firefox), I can confirm this resolves the issue :+1:

inetol (commented):
This change is only visible if the screen width is <=800px. On a 2340x1080 screen I see no change:

Image

Not a problem of this PR but I wanted to put it on record.

return42:
> On a 2340x1080 screen I see no change:

@inetol in the developer mode of your WEB-Browser, you can activate a "device toolbar" ..  in chrome it is the second icon on the left side.

grafik

inetol:
> inetol in the developer mode of your WEB-Browser, you can activate a "device toolbar" .. in chrome it is the second icon on the left side.

I know, but for me at least it breaks the display and doesn't show it accurately as I would see it from a real/virtualized device.

mrpaulblack:
ah well phone screens have gotten so big that 50rem width is maybe not enough anymore. Maybe we can up the value a bit in the future...

PR #5016, opened 2025-07-16
[fix] calculator plugin: filtering real calculation tasks

return42:
Whether the query is a real calculation tasks is currently only detected in the AST, resulting in unnecessary creatins of subprocesses. This problem is mitigated with this patch: if the query contains letters, it is obviously not a math problem, and the plugin can return without further action.

### DIFF
--- a/searx/plugins/calculator.py
+++ b/searx/plugins/calculator.py
@@ -71,6 +71,22 @@ class SXNGPlugin(Plugin):
         # replace commonly used math operators with their proper Python operator
         query = query.replace("x", "*").replace(":", "/")
 
+        # Is this a term that can be calculated?
+        word, constants = "", set()
+        for x in query:
+            # Alphabetic characters are defined as "Letters" in the Unicode
+            # character database and are the constants in an equation.
+            if x.isalpha():
+                word += x.strip()
+            elif word:
+                constants.add(word)
+                word = ""
+
+        # In the term of an arithmetic operation there should be no other
+        # alphabetic characters besides the constants
+        if constants - set(math_constants):
+            return results
+
         # use UI language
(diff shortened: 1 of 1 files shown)


inetol (commented):
What about operations with notations only? Or simpler still, obtaining the value of "pi" will no longer be possible:

Image

searx/plugins/calculator.py
```diff
@@ -71,6 +71,13 @@ def post_search(self, request: "SXNG_Request", search: "SearchWithPlugins") -> E
         # replace commonly used math operators with their proper Python operator
         query = query.replace("x", "*").replace(":", "/")
 
+        # Is this a term that can be calculated?
+        for x in query:
+            # Alphabetic characters are defined as "Letters" in the Unicode
+            # character database and cannot be passed to the calculator.
+            if x.isalpha():
```
Bnyro, line 79: Since recently, we support math constants like `e`, `pi`, see https://github.com/searxng/searxng/pull/4942 (that's why I removed that check).

Hence we have to allow the pattern `(?!<\w)e|pi(?!\w)`, even though they contain letters. Possibly we also have to support other patterns depending on which further `math_constants` we will add, so we should directly build the pattern from `math_constants`.

return42:
> so we should directly build the pattern from `math_constants`.

Sorry, I totally ignored the constants .. should be fixed by [diff](https://github.com/searxng/searxng/compare/bfd7f1212f3ce455982d214c89e1c877e91b1a5f..193f9926454687d01c41467d3b06cbf33be4dc8e).

> What about operations with notations only? 

Have I missed something else?

inetol (approved):
LGTM

Bnyro (approved):
Works as expected, thank you for fixing this!

PR #5018, opened 2025-07-16
Rework of the developer environment

### LINKED ISSUE #4803: Add shell (bash) formatter to SearXNG's tool chain

return42:
## What does this PR do?

This PR cleans up the developer environment and enables the use of tools from the Go ecosystem.

The versions of the Node.js and Go tools are pinned and updated by the Dependantbot.

The old LXC scripts have been removed (they have been unusable for some time).

The shell scripts are formatted with `shfmt`.

... for further details, please read the commit messages.

- [chore] overdue maintenance of shell scripts
- [mod] add Golang ecosystem to the SearXNG toolchain
- [mod] create a Go project with the developer tools from Go
- [mod] dependabot: add searxng.org/devtools from Go and Node.js
- [enh] CI - add shell script formatter
- [fix] don't install pyright twice and do some clean up
- [fix] apply shell formating / shfmt (make format.shell)
- [upd] update to gecko driver v36

## Related issues

- https://github.com/searxng/searxng/pull/4918


utils/lib_govm.sh
```diff
@@ -0,0 +1,222 @@
...
+}
+
+go.install() {
+    go.vm.ensure
+    GOENV="${GOENV}" "${GOVM_EXE}" install "$@"
+    # not sure why, but go installs some files without setting the write access
+    # for the file owner
+    chmod -R u+w "${GOMODCACHE}"
+}
+
```
inetol, line 75: We should add tidy as module manifest will get messy over time
```sh
go.tidy() {
    go.vm.ensure
    GOENV="${GOENV}" "${GOVM_EXE}" mod tidy
}
```

https://go.dev/ref/mod#go-mod-tidy

Makefile
```diff
@@ -82,11 +79,12 @@ MANAGE += gecko.driver
...
 MANAGE += py.build py.clean
 MANAGE += pyenv pyenv.install pyenv.uninstall
-MANAGE += format.python
-MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.types.dev test.types.ci
+MANAGE += format.python format.shell
+MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.pyright test.shfmt
 MANAGE += themes.all themes.simple themes.fix themes.lint themes.test
 MANAGE += static.build.commit static.build.drop static.build.restore
 MANAGE += nvm.install nvm.clean nvm.status nvm.nodejs
+MANAGE += go.env.dev go.clean
```
inetol, line 87: Expose `go.tidy`

inetol (commented):
Just a small detail, other than that the changes look good.

utils/lib_govm.sh
return42: We can implement a `go.tidy()` if you want .. but I have two  comments before:

1. The tools installed by `go mod tidy` are a part of the developer environment, thats why I think its enough when `go mod tidy` is called (only) once, when `go.env.dev()` is called (see line 47).

2. For reasons unknown to me, go installs some files without write permission for the owner. This means that the owner who installed these files can no longer delete them... not without doing a chmod and giving themselves write permission for the files.  That's why I added a chmod to `go.install()` on line 68 and also to `go.env.dev()` on line 48. We would then have to add this chmod here as well.

What do you think, based on my explanation? Do we need a `go.tidy()`?

inetol: > 1. The tools installed by `go mod tidy` are a part of the developer environment, thats why I think its enough when `go mod tidy` is called (only) once, when `go.env.dev()` is called (see line 47).

Fair enough. Closing https://github.com/searxng/searxng/pull/5018#discussion_r2213094467

> 2. For reasons unknown to me, go installs some files without write permission for the owner. This means that the owner who installed these files can no longer delete them... not without doing a chmod and giving themselves write permission for the files.  That's why I added a chmod to `go.install()` on line 68 and also to `go.env.dev()` on line 48. We would then have to add this chmod here as well.

Weird, in my case I have not seen that the ownership differs, but for the moment we can also place the chmod there.

utils/lib_govm.sh
```diff
@@ -0,0 +1,222 @@
...
+        # prevent GOBIN (etc.) from being obtained from the environment
+        unset -v GOBIN GOPATH GOTOOLCHAIN GOCACHE GOMODCACHE
+
+        "${GOVM_EXE}" env -w \
+            GOBIN="$GOBIN" \
+            GOTOOLCHAIN="$GOTOOLCHAIN" \
+            GOCACHE="$GOCACHE" \
+            GOPATH="$GOPATH" \
+            GOMODCACHE="$GOMODCACHE"
+    )
```
inetol: Looking at the modules folder, I realized that they were not installed in the expected `$GOPATH`, but in my home folder (~/go/). As I have Go installed system wide, it defaults to that path.

Removing the `unset` got it to work as expected, as it's also unsetting the ENVs from the script:

Image

return42: Argh, sorry .. the `unset` was an leftover from one of my first implementations :facepalm: .. thanks for giving heads up :+1:

utils/lib_govm.sh
return42: >> 2. For reasons unknown to me, go installs some files without write permission for the owner. 

> Weird, in my case I have not seen that the ownership differs

not the ownership differs, the `w` (write) permisson is not set for the owner.  Apart from this PR, when I try to fully remove all go installations from my $HOME folder I use this `rm`

```
$ rm -rf ~/go/ ~/.config/go ~/.cache/go-build/
```

But there are a lot of messages "Permission denied" from files of the `go/pkg/mod/` folder (`$GOMODCACHE`), here is one for example:

```
rm: cannot remove '/home/markus/go/pkg/mod/mvdan.cc/sh/v3@v3.12.0/cmd/shfmt/main_test.go': Permission denied
```

And here is the `ls` for this file:

```
$ ls -la /home/markus/go/pkg/mod/mvdan.cc/sh/v3@v3.12.0/cmd/shfmt/main_test.go
-r--r--r-- 1 markus markus 579 13. Jul 11:20 /home/markus/go/pkg/mod/mvdan.cc/sh/v3@v3.12.0/cmd/shfmt/main_test.go
```

This means, Go has installed this file without write permission and I have to give myself write permission back before I can delete the files (a file in my own HOME!?!?) .. not sure why Go does this.

```
chmod -R u+w /home/markus/go/pkg/mod/
```

utils/lib_govm.sh
return42: I fixed the *unset* issue .. further I added the python virtualenv to PATH .. and the Go URL has moved from `https://golang.org/dl` to `https://go.dev/dl`

utils/lib_govm.sh
```diff
@@ -0,0 +1,224 @@
...
+
+# shellcheck source=utils/lib.sh
+. /dev/null
+
+# configure golang environment for go.vm
+# --------------------------------------
+
+_GO_DL_URL="https://go.dev/dl"
+
+GOVERSION="${GOVERSION:-go1.24.5}"
```
inetol: We should probably pick the version defined in `go.mod`

```suggestion
GOVERSION="${GOVERSION:-go$(awk '/^go /{print $2}' "${REPO_ROOT}/go.mod")}"
```

I don't know about Dependabot (probably the same thing), but [Renovate](https://github.com/renovatebot/renovate) bumps the minimum version when some module requires a higher version than the one defined in `go.mod`

return42: Good catch .. [diff](https://github.com/searxng/searxng/compare/cbaef5ab8e8bbc89dc396e3386db7705970ac10b..b26c7d9f5afc66fe6652bdbb0ab43bde89f0acec)

inetol (approved):
LGTM

go.mod
```diff
@@ -0,0 +1,14 @@
+module searxng.org/devtools
+
+go 1.24.5
+
+tool mvdan.cc/sh/v3/cmd/shfmt
```
Bnyro, line 5: The `tool` thing is quite a neat feature, I wasn't aware of it even though I use Go regularly. Seems to have been released just some months ago.

utils/lib_govm.sh
```diff
@@ -0,0 +1,224 @@
...
+go.arch() {
+    local ARCH
+    case "$(command uname -m)" in
+        "x86_64") ARCH=amd64 ;;
+        "aarch64") ARCH=arm64 ;;
+        "armv6" | "armv7l") ARCH=armv6l ;;
+        "armv8") ARCH=arm64 ;;
+        .*386.*) ARCH=386 ;;
+        ppc64*) ARCH=ppc64le ;;
+        *) die 42 "ARCH is unknown: $(command uname -m)" ;;
```
Bnyro, line 97: Is the 42 here a canonical error code or is it just because the answer to everything is 42? :)

utils/lib_govm.sh
```diff
@@ -0,0 +1,224 @@
...
+    "${PY_ENV_BIN}/python" <<EOF
+import sys, json, requests
+resp = requests.get("${_GO_DL_URL}/?mode=json&include=all")
+for ver in json.loads(resp.text):
+    if not ver['stable']:
+        continue
+    for f in ver['files']:
+        if f['kind'] != 'archive' or not f['size'] or not f['sha256'] or len(f['os']) < 2:
+            continue
+        print(" %(version)-10s|%(os)-8s|%(arch)-8s|%(filename)-30s|%(size)-10s|%(sha256)s" % f)
```
Bnyro, line 193: It'd probably be cleaner to either
- use `curl` and `jq` (or a similar JSON parser)
- put the python scripts in a separate file to benefit from code formatting and other LSP hints

go.mod
return42: > Seems to have been released just some months ago.

Yes, implemented in go 1.24 .. However, I am not sure if it is a good idea to implement our own solutions for each ecosystem for a developer environment. Perhaps in the long term we should use a tool like https://mise.jdx.dev/demo.html (?)

utils/lib_govm.sh
return42: > * use `curl` and `jq` (or a similar JSON parser)

The code from above is executed in the pythen dev-env we already have (`"${PY_ENV_BIN}/python"`) .. while `jq` is a dependency that needs to be installed first from distro's package manager ..


> * put the python scripts in a separate file to benefit from code formatting and other LSP hints

:thinking: thats true  .. but involves more files  (laying arround).. I think we will never touch this code again, and its easier to have all the "go" code in one file.  But if you insist, I can separate the python code into its own file.

---

On the long run we should to enforce:

- https://github.com/searxng/searxng/issues/5056

.. which could make these hacky solutions unnecessary and simplify our devtools environment considerably.

go.mod
inetol: [asdf](https://docs.searxng.org/dev/rtm_asdf.html) is an already documented solution to manage toolchain versions, I don't know if it still works or if someone is using this. There's also [nvm](https://docs.searxng.org/dev/makefile.html#nvm-make-nvm-install-nvm-status) tracking Node.js/npm used in scripts. We should really unify how we handle devtools.

return42: > [asdf](https://docs.searxng.org/dev/rtm_asdf.html) is an already documented solution to manage toolchain versions,

yeah, thanks for reminding me :) .. I wrote this article at a time when I used asdf .. but nowadays I started to get experience with [mise](https://mise.jdx.dev/demo.html) --> 

I'm currently still in the testing phase, but mise-en-place is already convincing me: while asdf compiles the binaries (e.g., for the Python interpreter), mise-en-place installs the binaries from a repository [(backend)](https://mise.jdx.dev/architecture.html#backend-system-src-backend). The latter is not only faster, it also has other advantages:

- no build chain needs to be installed
- the binaries are identical

For example, with asdf, I had the problem that the Python interpreter was built .. and the result depends heavily on the system on which the interpreter is built. For example, I discovered that the [threading properties](https://github.com/searxng/searxng/blob/54a2b553f46df6dc98a4b8410788e35165e0fa3e/searx/sqlitedb.py#L101-L114) of the [sqlite3](https://docs.python.org/3/library/sqlite3.html) API in the self-built interpreter depended heavily on which sqlite was used during the build time.

It took me a while to realize that even a newly built Python 3.13 (on an old LTS) is not identical to the official Python interpreter in its SQLite3 API.

> We should really unify how we handle devtools.

Good to hear that you agree .. may we should open an issue for this task.

inetol: Opened an issue in https://github.com/searxng/searxng/issues/5056

Bnyro (approved):

return42:
FYI last force push was just to solve a conflict with master branch / I'm going to merge this PR now ..

PR #5019, opened 2025-07-16
[fix] custom `plugins:` settings must not be merged.

return42:
In customizing it should be decided which plugin modules should be loaded and which should not.

Related:

- https://github.com/searxng/searxng/pull/4678#issuecomment-3079109035


Bnyro (approved):
Thanks for the quick fix 👍

SuperSandro2000:
It is not clear for me right now how to disable a plugin. Should I just set active to false in the server config?

return42:
@SuperSandro2000 sorry for the inconvenience you had .. I added some documentation to this PR: https://docs.searxng.org/admin/settings/settings_plugins.html

grafik

PR #5020, opened 2025-07-17
[fix] py: overwrite version_frozen on explicit freeze

inetol:
Once version_frozen.py has been created, it will never be updated again unless the file is manually deleted.


inetol:
This will need git installed while it is running. I will look another way to handle this

return42 (approved):

PR #5037, opened 2025-07-23
[mod] make run: start granian server and versioning by Dependabot 

return42:
The new ``requirements-server.txt`` (granian) is installed into the virtualenv
of Dockerfile.

When ``make run`` is called, a granian server is started with auto reload on
application's files changes / requires granian[reload] extra, see
``requirements-dev.txt``.

Dependabot supports updates to any ``.txt`` file [1].

[1] https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#pip-and-pip-compile


container/Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY ./requirements.txt ./requirements.txt
 RUN --mount=type=cache,id=pip,target=/root/.cache/pip python -m venv ./venv \
  && . ./venv/bin/activate \
  && pip install -r requirements.txt \
- && pip install "granian~=2.0"
+ && pip install -r requirements-granian.txt
```
return42: with this patch, we have the version updates back in the dependantbot: [Dependabot supports updates to any ``.txt`` file.](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#pip-and-pip-compile)

manage
```diff
@@ -126,7 +127,17 @@ webapp.run() {
         sleep 3
         xdg-open http://127.0.0.1:8888/
     )&
-    SEARXNG_DEBUG=1 pyenv.cmd python -m searx.webapp
```
inetol, line 129: If I understand this change correctly, we will no longer be using `werkzeug.serving`?

Granian has autoreload with `GRANIAN_RELOAD` (there are other ENVs related), it would be interesting to replace this missing functionality.

https://github.com/emmett-framework/granian?tab=readme-ov-file#options

container/Dockerfile
inetol: https://github.com/searxng/searxng/pull/5037/files#diff-cc3b3db023efc81750e416e31d7a4b949db47a02683968e53761662b6138bcedL3

```diff
-COPY ./requirements.txt ./requirements.txt
+COPY ./requirements*.txt ./
```

We copy all the `requirements*.txt` to builder image.

inetol (changes requested):

manage
return42: Thanks for pointing out / using granian's auto-reload option will simplify the code a lot .. we do not longer have to stop Flask from initializing twice:

https://github.com/searxng/searxng/blob/83adda8567909b6077706195d8695aa58f5e9700/searx/webapp.py#L1411-L1420

manage
```diff
@@ -126,7 +126,19 @@ webapp.run() {
...
+    SEARXNG_DEBUG=1 \
+        GRANIAN_RELOAD=1 \
+        GRANIAN_RELOAD_PATHS="./searx" \
```
inetol: We should add `GRANIAN_RELOAD_IGNORE_WORKER_FAILURE="true"` to prevent Granian from exiting on Python errors while editing source.

inetol (approved):
Looks good!

manage
return42: > We should add `GRANIAN_RELOAD_IGNORE_WORKER_FAILURE="true"`

Another good catch .. [done](https://github.com/searxng/searxng/compare/fc8dffdf772f6be416893af598bfcef66d5e312c..23c103f7b095340b6d4c75b092bea45fc630eacb).

manage
return42: > If I understand this change correctly, we will no longer be using `werkzeug.serving`?

Yes, I think we are closer to the deployment, when we use granian already here in the developer environment .. further we can test version upgrades of granian in the developer environment ..

return42:
For those of you missing *the good old* [Python Debugger `pdb`](https://docs.python.org/3/library/pdb.html#module-pdb) in the developer environment:

Alternatively to `make run` which starts granian WSGI server (since this PR has been merged) you can jump into the developer environment by:

```bash
$ ./manage dev.env
...
(dev.env)$ 
```

and start a python based HTTP server by:

```
SEARXNG_DEBUG=1 python -m searx.webapp
```

since this is a pure python solution, you can set `pdb.set_trace()` in your code, as previously known from `make run`.

PR #5051, opened 2025-07-26
[fix] webapp.py: info (and other) page(s) don't load properly

### LINKED ISSUE #5050: About page not loading correctly
 2025.7.25+168fa9b

using searxng-docker

**What happened?**
About page does not load properly 

**How To Reproduce**
Click on the about page

**Expected behavior**
for it to load correctly

**Screenshots & Logs**
Image

Image

**Additional context**

Bnyro:
## Author notes
- I'm not 100% sure if this works if hosting SearXNG on a subfolder (e.g. `/searx/`), as I didn't test it. But looking at how we handled it previously, I guess it should.

## Related issues
closes #5050

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""WebbApp
-
-"""
+"""WebApp"""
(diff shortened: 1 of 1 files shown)


inetol (approved):
Thanks for looking into this, working OK now.

PR #5067, opened 2025-07-29
Windows support

### LINKED ISSUE #2463: [Question] Remove colons in the files located in the git repository

Vizonex:
## What does this PR do?

This pull request will add support for windows operating systems which has been one of searxing's weakspots for as long as I can remember. I desire to hopefully put an end to this since I am one of the main maintainers behind winloop which has been used to overcome some os barriers that uvloop has. 

This change will allow windows users to host searxng servers of their own locally my goal is to hopefully refactor searx a bit while I go about adding support to windows. I'll be editing things if I notice anything that my pyright extension doesn't like.

To get caught up with my changes here's what you can do. 

Step 1. Get a windows operating system of your choice for me I am using windows 10 and I refuse to move to 11 even after EOL.
Step 2. I'm using `uv` for the sake of simplicity and saving as much diskspace as I can since I make & maintain multiple libraries including contributing to many different parts of __aiohttp__ to name a few of the places I've been.
```
uv venv
```
Step 3.  activate the VM
```
.venv/scripts/activate 
```
Step 4. Install everything needed to run the server.
```
uv pip install -e .
```
Step 5. To run, performing the following will get you to where I am currently stuck which is sqlite3 issues (I am using 3.10.18 for backwards compatability so that I can attempt to support all versions of python that still get security updates)
```
py -m searx.webapp
```

## Author's checklist

- [ ] ensure windows runs
- [ ] remove or add libraries that do or don't support windows by performing a `sys.platform != 'win32' or sys.platform == 'win32'` check.

## Related issues

Feel free to list the issues that we can close and I will edit and put them all in here, sorry If I'm a bit on the lazy side. I've had a long day.

- 'fixes #2463'
- 'fixes #2465'


Vizonex:
image
So Apparently I got it to run but now comes the annoying part, figuring out why the CSS won't load.

Vizonex:
I'll come back to this a little bit tomorrow. Glad it runs but now it's just stuck at loading the Jinja2 templates nothing I do gets it to run beyond that point.

return42:
> my goal is to hopefully refactor searx a bit while I go about adding support to windows.

Thank you for your commitment, but unfortunately, I have to inform you [MS-Windows is not in the scope of this project](https://github.com/searxng/searxng/issues?q=label%3ANT)

**The SearXNG project is based on the concepts of open source and user privacy.**

Microsoft's recent and past history has been characterized by its efforts to combat these very concepts.

> .. company's potential approaches to stopping open source momentum. One strategy was "[embrace-extend-extinguish](https://en.wikipedia.org/wiki/Embrace-extend-extinguish)", in which Microsoft would adopt standard technology, add proprietary extensions, and upon establishing a customer base, would lock consumers into the proprietary extension to assert a monopoly of the space. The memos also acknowledged open source as a methodology capable of meeting or exceeding proprietary development methodology. Microsoft downplayed these memos as the opinions of an individual employee and not Microsoft's official position [[1]](https://en.wikipedia.org/wiki/Microsoft_and_open_source#Initial_stance_on_open_source)

But even if we ignore these aspects, once we support MS Windows, we face several problems that do not exist in the Linux/Unix world, including porting the developer environment and quality assurance.

We might consider taking on these efforts (some of which are already evident in this PR), but we would only encourage people to stick with their decision to use MS Windows ...

> This change will allow windows users to host searxng servers of their own locally 

Let's be clear: MS Windows is anything but a server operating system.

Ultimately, however, it is everyone's own decision as to which operating system they choose... and even for MS Windows, a solution is available with the Docker image.

searx/plugins/calculator.py
```diff
@@ -176,7 +177,11 @@ def _compare(ops: list[ast.cmpop], values: list[int | float]) -> int:
 # the old behavior "fork") but it will not solve the core problem of fork, nor
 # will it remove the deprecation warnings in py3.12 & py3.13.  Issue is
 # ddiscussed here: https://github.com/searxng/searxng/issues/4159
-mp_fork = multiprocessing.get_context("fork")
+
+# XXX: WINDOWS CAN'T FORK!!!
+
+
+mp_fork = multiprocessing.get_context("fork" if sys.platform != "win32" else "spawn")
```
return42, line 184: Here we can see what I mean by:

> But even if we ignore these aspects, once we support MS Windows, we face several problems that do not exist in the Linux/Unix world ..

and here's a paper listing the problems with fork() and suggesting it should be removed from other OSs. 

https://www.microsoft.com/en-us/research/uploads/prod/2019/04/fork-hotos19.pdf 

Needless to point out this Paper is from MS and these problems only exist on Windows because MS (no longer) implements fork(). And MS's logical suggestion is that all other operating systems should also abandon fork... just like any other POSIX standard that MS cannot or will not implement.

[Fork is an interface which is required for compliance with the POSIX and Single UNIX Specification standards. It is usually implemented as a C standard library wrapper to the fork, clone, or other system calls of the kernel.](https://en.wikipedia.org/wiki/Fork_(system_call))

kalle07:
dont know if this is comparable but 
https://github.com/yacy/yacy_search_server
based on java, dont know if that is more unsecure than WSL ^^
but its an easy exe to install

PR #5071, opened 2025-07-29
[chore] favicon.html: slightly improve wording

benpiano800:
## What does this PR do?

This PR slightly improves the wording of the description of the "Favicon" option.

"Display favicons next to search results" is better wording than "display favicons near search results"

Replace favicon.html in your instance with the updated file.

### DIFF
--- a/searx/templates/simple/preferences/favicon.html
+++ b/searx/templates/simple/preferences/favicon.html
@@ -12,6 +12,6 @@
     </select>{{- '' -}}
   </div>{{- '' -}}
   <div class="description">
-    {{- _('Display favicons near search results') -}}
+    {{- _('Display favicons next to search results') -}}
   </div>{{- '' -}}
 </fieldset>{{- '' -}}


vojkovic:
Thanks for your PR. I agree that your wording is probably better.

However, think the preference descriptions should be all be looked at as a complete whole. Just looking, I can see many more cases of poor wording and I would like to fix this as part of a larger commit as to not pollute the git log.

Do you agree @return42? Happy to quickly draft this commit or Ben can if he chooses.

return42:
> Do you agree @return42? 

Cleaning up several strings in one commit makes sense in my opinion... BUT:

Please note that with every change to a string for which there is a translation, all previous translations are lost. Since we essentially translate all strings, this affects every string.

- https://translate.codeberg.org/projects/searxng/searxng/

We recently had several changes to various strings... the translation rate is slowly recovering, but it will take time.

We should therefore only change strings if the wording is really bad (even horrible). Since my native language isn't English and I'm also dyslexic, I usually can't properly assess whether a change should be made.

In any case, we should avoid changing the same string multiple times, because different people/cultural groups have different ideas about what constitutes "good pronunciation." ..

.. our translators don't like playing ping-pong games like that ;-)

------------

If the changes to the strings are minor, the "auto suggestion" usually suggests translating the old string. After a PR with major changes to strings, I occasionally dig through the translations to see which of the "auto suggestions" I can adopt. But this is tedious work. I would be happy if the opener of a PR also had a look at the translations and supported me in this process :+1:

benpiano800:
> I would be happy if the opener of a PR also had a look at the translations

The only other language I understand is Spanish (roughly C1 level) so I'm not sure I'd be much help in most languages.

vojkovic:
> The only other language I understand is Spanish (roughly C1 level) so I'm not sure I'd be much help in most languages.

I don't think this is what Markus was saying... more so that every time we change a string in the codebase, all the translations need to be redone for that string, in every language. A change in English needs to be redone for 50+ other languages.

So, to avoid annoying our translators, we should only update each description once, and only if it is necessary / meaningful.

I'm voting to close this PR in favour of a larger commit to rework the preference descriptions.

benpiano800:
Oh okay. Yeah, let's do that then.

PR #5072, opened 2025-07-29
[feat] statistics answerer: add the ability to calculate range

benpiano800:
## What does this PR do?

This adds to the statistics answerer the ability to calculate the range of a given set by subtracting the minimum from the maximum.

Quickly showing the range of a set can be useful.

Replace statistics.py in your instance with the updated file.

### DIFF
--- a/searx/answerers/statistics.py
+++ b/searx/answerers/statistics.py
@@ -21,6 +21,7 @@ kw2func = [
     ("avg", lambda args: sum(args) / len(args)),
     ("sum", sum),
     ("prod", lambda args: reduce(mul, args, 1)),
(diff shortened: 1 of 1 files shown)


Bnyro (commented):
Works as expected.

However this does exactly the same as just subtracting the minimum from the maximum value, so what's an actual use case for using it?

benpiano800:
Less typing I guess? It just felt kind of weird that the statistics answerer didn't have a "range" function.

Bnyro:
The change is small enough to not worry about it, so fine with me :)

Thanks for the PR!

PR #5074, opened 2025-07-30
[feat] engines: add Github Code Search engine

fiffeek:
## What does this PR do?

Adds an engine to query Github repositories using the [Github Code Search API](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-code).  For motivation see `Why is this change important?`.

Since the security is important for the project I tried to make sure to adhere to the standards:

1. Auth Token

The API can be used without authenticating:
image

and the defaults in this implementation is not to require a token, however, github heavily limits such API calls (I've never managed to properly query it this way, it seems to depend on the public IP that the API is presented with):
```json
{
  "message": "API rate limit exceeded for REDACTED. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) If you reach out to GitHub Support for help, please include the request ID REDACTED and timestamp REDACTED UTC.",
  "documentation_url": "https://docs.github.com/rest/overview/rate-limits-for-the-rest-api",
  "status": "403"
}
``` 

[So realistically one needs to either generate a personal token, or an app bearer token](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28).

2. The engine is disabled by default.

### Screenshots

A few examples:
image
image

I use searxng as a single pane for multiple search results, as a coder, I spend considerable amount of time searching through Github repos to find a better solution or just poke around peoples' ideas how to accomplish something specific. In most cases the returned context doesn't matter that much for me, I just want to be able to type a query and get significant matching results, then I skim through the visible context of the file and decide to open the repo or not. IMO it would be fantastic to have github code search natively in searxng so that the context switching is minimized. 

I feel like there is a lot of non-public smaller searxng instances that would benefit from the ability to simply surface github's code search (myself and my household included). 

More than likely you need a personal token since unauth queries are too heavily throttled.

1. Generate a token in [developer settings](https://github.com/settings/tokens/new), the easiest way is to generate the classic token, the only scope needed really is the public repos read access:
image
2. Adjust the searxng `settings.yaml`, e.g.:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index c03e3fa67..1d4aae5c4 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -986,15 +986,15 @@ engines:
   - name: github code
     engine: github_code
     shortcut: ghc
-    disabled: true
+    disabled: false
     ghc_auth:
       # type is one of:
       # * none
       # * personal_access_token
       # * bearer
       # When none is passed, the token is not requried.
-      type: "none"
-      token: "token"
+      type: "personal_access_token"
+      token: "your token here"
     # specify whether to highlight the matching lines to the query
     ghc_highlight_matching_lines: true
     ghc_strip_new_lines: true
```
3. `make run` then open searxng and type your query with `ghc` bang, e.g. `!ghc lazyvim`, you can use github's query language and that will be passthrough'd, e.g. `!ghc lazyvim language:Lua`

unit tests are added for parsing and highlighting logic, can be run in isolation:
```bash
source local/py3/bin/activate
python -m nose2 -vvvv tests.unit.test_engine_github_code
```

#### Limitations

The API does not return the line numbers alongside the response, it sends the matching code fragments and tells you what words can be highlighted on the ui, e.g. query for `lazivim` returns (this is stripped to showcase the payloads):

```json
    {
      "name": "TODO.md",
      "path": "TODO.md",
      "sha": "e9a138103a10ab2be63f5e1d155bc4e55415feb6",
      "url": "https://api.github.com/repositories/214174020/contents/TODO.md?ref=3140f4f5720c3cc6b5034c624eb7706f8533a82c",
      "git_url": "https://api.github.com/repositories/214174020/git/blobs/e9a138103a10ab2be63f5e1d155bc4e55415feb6",
      "html_url": "https://github.com/folke/dot/blob/3140f4f5720c3cc6b5034c624eb7706f8533a82c/TODO.md",
      "repository": {
        "id": 214174020,
        "node_id": "MDEwOlJlcG9zaXRvcnkyMTQxNzQwMjA=",
        "name": "dot",
        "full_name": "folke/dot",
        }
        "text_matches": [
        {
          "object_url": "https://api.github.com/repositories/214174020/contents/TODO.md?ref=3140f4f5720c3cc6b5034c624eb7706f8533a82c",
          "object_type": "FileContent",
          "property": "content",
          "fragment": "- [x] windows picker\n- [x] toggle cwd / root (LazyVim)\n- [x] dynamic workspace symbol",
          "matches": [
            {
              "indices": [
                46,
                53
              ],
              "text": "LazyVim"
            }
          ]
        },
        {
          "object_url": "https://api.github.com/repositories/214174020/contents/TODO.md?ref=3140f4f5720c3cc6b5034c624eb7706f8533a82c",
          "object_type": "FileContent",
          "property": "content",
          "fragment": "- [x] smart stops working after custom\n- [x] edit in empty buffer\n- [x] support toggling line nr for preview",
          "matches": [
            {
              "indices": [
                59,
                65
              ],
              "text": "buffer"
            },
            {
              "indices": [
                89,
                93
              ],
              "text": "line"
            }
          ]
        }
      ]
```

And this is how github  shows this match:
image

Whereas, this is how searxng implementation looks like:

image

The following decisions were made:
* since there is no line indices, just relabel them starting from 1 (pygments "0")
* concatenate the fragments of code one after another
* allow for line (instead of word) highlighting (as far as I can see pygments just does not natively support word highlights, I can work on that later if it would be of use?)

An alternative would be to query for each file contents in the search result and then locate the text, I did not go into that direction (yet?) to see whether there is any interest first, since this approach does not really play well in the searx framework and more importantly generates loads of query calls that are just (IMO) unnecessary.

**Another option is to hide the line numbers altogether but still display highlighting, I'd imagine it's just some css change, I'd like to hear the maintainers opinions on this.**

#### code.html and webapp changes

The method used previously to highlight code in searchcode relied on file extensions and was mapping them heuristically (+ pygments internal logic), this can be simplified by just the other pygments method:
```python
from pygments.lexers import get_lexer_for_filename
````
no need to do the work if pygments does it for us, the callers have been adjusted.

I've exposed stripping arguments from pygments parsers to the user, I think it should be a choice whether they want to keep the leading and trailing whitespace/newlines. Particularly, the previous implementation in searchcode explicitly relies on stripping the whitespace which just breaks the first line of code indentation (this behavior is left intact) and the engine works as expected:
image

#### EngineResults

Since the types for `Code` are not implemented yet I used the legacy typing, I can output the type addition and change it in another PR if it's of value.

## Related issues

If it's of interest, I'd be down to implement something similar for Gitlab, although a token is definitely required to code search (let me know, since I'll be implementing this for myself anyway, just depends whether I'd output the PR in the end). Open to discussing this in an issue as well.

### DIFF
--- a/searx/engines/github_code.py
+++ b/searx/engines/github_code.py
@@ -1,0 +1,272 @@
+# SPDX-License-Identifier: AGPL-3.0-or-lat_er
+"""GitHub code search with `search syntax`_ as described in `Constructing a
+search query`_ in the documentation of GitHub's REST API.
+
+.. _search syntax:
+    https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax
+.. _Constructing a search query:
+    https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#constructing-a-search-query
+.. _Github REST API for code search:
+    https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-code
+.. _Github REST API auth for code search:
+    https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-code--fine-grained-access-tokens
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`ghc_auth`
+  Change the authentication method used when using the API, defaults to none.
+
+Optional settings are:
+
+- :py:obj:`ghc_highlight_matching_lines`
+   Control the highlighting of the matched text (turns off/on).
+- :py:obj:`ghc_strip_new_lines`
+   Strip new lines at the start or end of each code fragment.
+- :py:obj:`ghc_strip_whitespace`
+   Strip any whitespace at the start or end of each code fragment.
+- :py:obj:`ghc_insert_block_separator`
+   Add a `...` between each code fragment before merging them.
+
+.. code:: yaml
+
+  - name: github code
+    engine: github_code
+    shortcut: ghc
+    ghc_auth:
+      type: "none"
+
+  - name: github code
+    engine: github_code
+    shortcut: ghc
+    ghc_auth:
+      type: "personal_access_token"
+      token: "<token>"
+    ghc_highlight_matching_lines: true
+    ghc_strip_whitespace: true
+    ghc_strip_new_lines: true
+
+
+  - name: github code
+    engine: github_code
+    shortcut: ghc
+    ghc_auth:
+      type: "bearer"
+      token: "<token>"
+
+Implementation
+===============
+
+GitHub does not return the code line indices alongside the code fragment in the
+search API. Since these are not super important for the user experience all the
+code lines are just relabeled (starting from 1) and appended (a disjoint set of
+code blocks in a single file might be returned from the API).
+"""
+
+from __future__ import annotations
+
+import typing as t
+from urllib.parse import urlencode, urlparse
+
+from pygments.lexers import guess_lexer_for_filename
+from pygments.util import ClassNotFound
+from searx.result_types import EngineResults
+from searx.extended_types import SXNG_Response
+from searx.network import raise_for_httperror
+
+# about
+about = {
+    "website": 'https://github.com/',
+    "wikidata_id": 'Q364',
+    "official_api_documentation": 'https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-code',
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+# engine dependent config
+categories = ['code']
+
+
+search_url = 'https://api.github.com/search/code?sort=indexed&{query}&{page}'
+# https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#text-match-metadata
+accept_header = 'application/vnd.github.text-match+json'
+paging = True
+
+ghc_auth = {
+    "type": "none",
+    "token": "",
+}
+"""Change the method of authenticating to the github API.
+
+``type`` needs to be one of ``none``, ``personal_access_token``, or ``bearer``.
+When type is not `none` a token is expected to be passed as well in
+``auth.token``.
+
+If there is any privacy concerns about generating a token, one can use the API
+without authentication.  The calls will be heavily rate limited, this is what the
+API returns on such calls::
+
+    API rate limit exceeded for <redacted ip>.
+    (But here's the good news: Authenticated requests get a higher rate limit)
+
+The personal access token or a bearer for an org or a group can be generated [in
+the `GitHub settings`_.
+
+.. _GitHub settings:
+   https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-code--fine-grained-access-tokens
+"""
+
+ghc_highlight_matching_lines = True
+"""Highlight the matching code lines."""
+
+ghc_strip_new_lines = True
+"""Strip leading and trailing newlines for each returned fragment.
+Single file might return multiple code fragments.
+"""
+
+ghc_strip_whitespace = False
+"""Strip all leading and trailing whitespace for each returned fragment.
+Single file might return multiple code fragments. Enabling this might break
+code indentation.
+"""
+
+ghc_api_version = "2022-11-28"
+"""The version of the GitHub REST API.
+"""
+
+ghc_insert_block_separator = False
+"""Each file possibly consists of more than one code block that matches the
+search, if this is set to true, the blocks will be separated with ``...`` line.
+This might break the lexer and thus result in the lack of code highlighting.
+"""
+
+
+def request(query: str, params: dict[str, t.Any]) -> None:
+
+    params['url'] = search_url.format(query=urlencode({'q': query}), page=urlencode({'page': params['pageno']}))
+    params['headers']['Accept'] = accept_header
+    params['headers']['X-GitHub-Api-Version'] = ghc_api_version
+
+    if ghc_auth['type'] == "none":
+        # Without the auth header the query fails, so add a dummy instead.
+        # Queries without auth are heavily rate limited.
+        params['headers']['Authorization'] = "placeholder"
+    if ghc_auth['type'] == "personal_access_token":
+        params['headers']['Authorization'] = f"token {ghc_auth['token']}"
+    if ghc_auth['type'] == "bearer":
+        params['headers']['Authorization'] = f"Bearer {ghc_auth['token']}"
+
+    params['raise_for_httperror'] = False
+
+
+def get_code_language_name(filename: str, code_snippet: str) -> str | None:
+    """Returns a code language name by pulling information from the filename if
+    possible otherwise by scanning the passed code snippet. In case there is any
+    parsing error just default to no syntax highlighting."""
+    try:
+        lexer = guess_lexer_for_filename(filename, _text=code_snippet)
+        if lexer is None:
+            return None
+        code_name_aliases = lexer.aliases
+        if len(code_name_aliases) == 0:
+            return None
+        return code_name_aliases[0]
+    except ClassNotFound:
+        return None
+
+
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by fiffeek** [comment]
Pushed [a small update](https://github.com/searxng/searxng/compare/74e48b4501e7d613f6d08881d885fd98851362fa..d8c5cf9b9b273b02ecc122411d128f901cdd710b) to strip from the start then calculate the stripped offset and then strip from the end so that the highlighting does not break (since gh returns the indices of the words in the text, stripping leading characters results in effectively offsetting all indices).

I'd appreciate thoughts/reviews, will ping in couple of days if it's left untouched -- I'm fine with not having this merged if it breaks the project's privacy guidelines (due to relying on the user token for auth to have this working reliably).

File: searx/webapp.py [hunk 1]
```diff
@@ -181,24 +181,32 @@ def _get_locale_rfc5646(locale):
...
-    if not language:
-        language = 'text'
+def code_highlighter(codelines, filename=None, hl_lines=None, strip_whitespace=True, strip_new_lines=True):
+    if not filename:
+        filename = '.txt'
 
     try:
-        # find lexer by programming language
-        lexer = get_lexer_by_name(language, stripall=True)
+        lexer = get_lexer_for_filename(filename, stripall=strip_whitespace, stripnl=strip_new_lines)
```
return42: > no need to do the work if pygments does it for us, the callers have been adjusted.

I understand the intention, but its problematic when we are not *explicit* .. by example when I search for `!ghc searxng locales python` the log says:

```
WARNING searx.webapp          : pygments lexer: no lexer for filename 'settings.yml.new' found
WARNING searx.webapp          : pygments lexer: no lexer for filename 'settings.yml.new' found
WARNING searx.webapp          : pygments lexer: no lexer for filename 'settings.yml.example' found
```

return42 (changes requested):
Wow, it's an amazing work / thanks a lot for sharing :+1: 

I pushed a WIP (commit on top) with some nits I suggest, further see my comment below in the code.

If you agree with my suggestions, please squash WIP into your commit.

> Since the types for `Code` are not implemented yet I 

Yeah, would be cool if you can implement a result type for the `Code`  results :rocket: 

----

About reStructured markup: We have a handy introduction with the most important markups .. you usually don't need more --> [reST primer](https://docs.searxng.org/dev/reST.html)

searx/webapp.py
fiffeek: If I understand you correctly you'd like to have these highlighted as well (do correct me if I am wrong), but I do not think that's necessary (and is not a regression from the previous code block). 

Github does not highlight these files either:
image

and treats them as text (which is consistent with this PR; all custom extensions are treated are such [`.new`, `.example` etc.]). I think for the user experience it would make sense that we don't heuristically highlight in the engine as github also ignores the syntax highlighting (hence the no lexer debug log).

Unless I completely misunderstood your point I think this is a noop but do please let me know.

---

This is how it looks like in searxng (for completeness of this comment lol):
image

So tldr; this is a deliberate decision to match github's behavior but happy to hear your thoughts.

---

As far as your changes are concerned LGTM, squashed and updated. Any comment on the line numbers? I can get a PR out probably to hide them (and will work on the Code `EngineResults` as well) so that it does not confuse users (atm these are arbitrary somewhat per the description).

return42: > If I understand you correctly you'd like to have these highlighted as well (do correct me if I am wrong), 

When we see its e.g. a YAML file, we can highlight the file .. but when there is no chance to detect the language we should not highlight and not emit a WARNING.

```
def code_highlighter(codelines, language=None, ...):
    if not language:
        language = 'text'
```

>> but its problematic when we are not *explicit*

By *"explicit"* I mean: in the previous implementation the caller (the engine implementation) decides which language the code block has .. now we *implicit*  that the file suffix promise us the correct language .. 

grafik


Sidenote "Explicit is better than implicit." is the second rule in [The Zen of Python](https://peps.python.org/pep-0020/#the-zen-of-python) .. and the WARNING we see confirms this rule :-)

> Any comment on the line numbers? I can get a PR out probably to hide them (and will work on the Code `EngineResults` as well) .. 

Yes, would be cool if you send follow up PR .. but first we can finalize the review of this PR (and merge).

fiffeek: Gotcha, busy weekend but thank you for this, will get back to you shortly with something reasonable, thanks lots for the pointers!

fiffeek: Heya, updated to what I think is a reasonable solution here:
* left out the code in scc engine as is, left the `code_language` param to be defined from the client side instead
* updated ghc logic to pull the syntax highlighting from the filename + code using the lexers lib and ignore all errors to pass `None` highlighter if we can't find the language (no warn will be produced along the way here)
  * this can be extended to some other heuristics later if necessary; i wanted to leverage the mapping that pygments already gives us for the file extension to lexer name here so that it's more flexible 
* for now not unify these two engines (in terms of the highlighting), it's up to the engine to explicitly define how the code block is highlighted

lmk if this is acceptable now; couple of things left we agreed on above that I'll do a follow up on, one other thing I noticed that I'll tackle as a part of the CSS on the line numbers is that syntax highlighting conflicts with code line highlights for a few languages (this seems to be a simple change to adjust `.hll` css class for pygments to be darker/lighter):
image

as that touches the css part I'll just do a follow up so we can have this rolling.

---

Everything works as expected after the changes above:
image

no warns produced on the query you highlighted:
image


and scc:
image

fiffeek:
hey @return42, would it be possible to get it looked at again? Thanks.

return42:
> hey @return42, would it be possible to get it looked at again? Thanks.

Of course, I will do that... but I must ask you for a little patience.

return42:
FYI last force push was just a rebase on current master branch, to get in use of:

- https://github.com/searxng/searxng/pull/5075

return42:
@fiffeek sorry for the delay .. I finally had time today to get back to your PR.

I've put two WIP commits on your branch... if the changes are OK with you, please squash them into your commit.

> couple of things left we agreed on above that I'll do a follow up on

I'm looking forward to your next PRs :smile:  ..   great work so far :rocket:  .. thank you very much!

searx/engines/github_code.py
```diff
@@ -0,0 +1,275 @@
...
+    try:
+        lexer = find_lexer_class_for_filename(filename, code=code_snippet)
+        if lexer is None:
+            return None
+        code_name_aliases = lexer.aliases
+        if len(code_name_aliases) == 0:
+            return None
+        return code_name_aliases[0]
+    except Exception:  # pylint: disable=broad-except
+        return None
```
return42: Sorry, one final request I have: catching broad exceptions is an anti-pattern we should avoid .. AFAIK: `find_lexer_class_for_filename` is not a part of the official API (name is not exported by `pygments/lexers/__init__.py`) / see offical API docs: https://pygments.org/docs/api 

The [`guess_lexer_for_filename`](https://pygments.org/docs/api/#pygments.lexers.guess_lexer_for_filename) functions accepts filename as first argument and code-lines in the second argument and might be what we are looking for.

Plus: the only exception raised by this function could be [`pygments.util.ClassNotFound`](https://pygments.org/docs/api/#pygments.util.ClassNotFound).

return42 (changes requested):

searx/engines/github_code.py
return42: Forget last comment from above .. I already did it in my PR:

- https://github.com/searxng/searxng/pull/5131

.. the PR is based on your branch here ..

fiffeek: Thanks for all the work, updated with your commits and changed the API call + exception handling, very nice find!

return42 (approved):

return42:
> Since the types for `Code` are not implemented yet I used the legacy typing, I can output the type addition and change it in another PR if it's of value.

@fiffeek I implemented a type for `Code` .. can you please review the PR:

- https://github.com/searxng/searxng/pull/5131

PR #5075, opened 2025-07-31
[mod] switching from pyright to basedpyright (plus first rules)

return42:
I tried various type checkers, focusing primarily on the features of the LSP. `basedyright`, or rather, the LSP `basedpyright-langserver`, convinced me.

For integration into your IDE (editor) use the basedpyright-langserver (LSP) installed by `pipx basedpyright` and read:

- https://docs.basedpyright.com/latest/installation/ides/

The `$REPO_ROOT/pyrightconfig.json` uses the virtualenv found in `$REPO_ROOT/local/py3` and can be created by a `make pyenv`

-----------

- `pyrightconfig.json` :
  for the paths searx, searxng_extra and tests, individual rules were defined (for example, in test fewer / different rules are needed than in the searx package

- `searx/engines/__builtins__.pyi` :
  The builtin types that are added to the global namespace of a module by the intended monkey patching of the engine modules / replaces the previous filtering of the stdout using grep.

- `test.pyright_modified (utils/lib_sxng_test.sh)` :
  static type check of local modified files not yet commited

- `make test` :
  prerequisite `test.pyright` has been replaced by `test.pyright_modified`

- `searx/engines/__init__.py`, `searx/enginelib/__init__.py` :
  First, minimal typifications that were considered necessary.

-------------

*I think we need to put more effort into typing, but we shouldn't try to type everything either; that just creates unnecessary work and is frustrating. Python is and remains a dynamically typed language, but types can improve quality and, with the LSP (IDE), positively impact productivity. In this sense, we should also align our goals with regard to typing.*

### DIFF
--- a/searx/engines/__builtins__.pyi
+++ b/searx/engines/__builtins__.pyi
@@ -1,0 +1,43 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""The builtin types that are added to the global namespace of a module by the
+intended monkey patching of the engine modules.
+
+.. attention::
+
+   Monkey-patching modules is a practice from the past that shouldn't be
+   expanded upon.  In the long run, there should be an engine class that can be
+   inherited.  However, as long as this class doesn't exist, and as long as all
+   engine modules aren't converted to an engine class, these builtin types will
+   still be needed.
+"""
+from __future__ import annotations
+
+import logging
+
+logger: logging.Logger
+supported_languages: str
+language_aliases: str
+
+# from searx.engines.ENGINE_DEFAULT_ARGS
+about: dict[str, dict[str, str | None | bool]]
+categories: list[str]
+disabled: bool
+display_error_messages: bool
+enable_http: bool
+engine_type: str
+inactive: bool
+max_page: int
+paging: int
+safesearch: int
+send_accept_language_header: bool
+shortcut: str
+time_range_support: int
+timeout: int
+tokens: list[str]
+using_tor_proxy: bool
+
+# from searx.engines.check_engine_module
+network: str
+
+# from searx.engines.update_attributes_for_tor
+search_url: str
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ search.checker.%: install
 	$(Q)./manage pyenv.cmd searxng-checker -v "$(subst _, ,$(patsubst search.checker.%,%,$@))"
 
 PHONY += test ci.test test.shell
-test:    test.yamllint test.black test.pyright test.pylint test.unit test.robot test.rst test.shell test.shfmt
+test:    test.yamllint test.black test.pyright_modified test.pylint test.unit test.robot test.rst test.shell test.shfmt
 ci.test: test test.pybabel
 test.shell:
 	$(Q)shellcheck -x -s dash \
@@ -80,7 +80,7 @@ MANAGE += node.env node.env.dev node.clean
 MANAGE += py.build py.clean
 MANAGE += pyenv pyenv.install pyenv.uninstall
 MANAGE += format.python format.shell
-MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.pyright test.shfmt
+MANAGE += test.yamllint test.pylint test.black test.pybabel test.unit test.coverage test.robot test.rst test.clean test.themes test.pyright test.pyright_modified test.shfmt
 MANAGE += themes.all themes.simple themes.simple.analyze themes.fix themes.lint themes.test
 MANAGE += static.build.commit static.build.drop static.build.restore
 MANAGE += nvm.install nvm.clean nvm.status nvm.nodejs
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "searxng.org/devtools",
   "type": "module",
   "dependencies": {
-    "pyright": "^1.1.403"
   },
   "scripts": {
     "clean": "rm -Rf node_modules package-lock.json"
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,9 +1,48 @@
 {
-    "venvPath": "local",
-    "venv": "py3",
-    "include": [
-        "searx",
-        "searxng_extra",
-        "tests"
-    ]
+  "venvPath": "local",
+  "venv": "py3",
+  "include": [
+    "searx",
+    "searxng_extra",
+    "tests"
+  ],
+  "reportAny" : "information",
+  "enableTypeIgnoreComments": true,
+  "reportIgnoreCommentWithoutRule": true,
+  "reportImplicitStringConcatenation": false,
+  "reportUninitializedInstanceVariable": false,
+  "reportUnnecessaryIsInstance": false,
+  "reportUnreachable": "information",
+  "reportUnusedCallResult": false,
+  "executionEnvironments": [
+    {
+      "root": "searx",
+      "extraPaths": ["."]
+    },
+    {
+      "root": "searxng_extra",
+      "extraPaths": ["."],
+      "reportAny" : false,
+      "reportUnknownMemberType": false,
+      "reportUnknownVariableType": false
+    },
+    {
+      "root": "tests",
+      "extraPaths": ["."],
+      "reportAny" : false,
+      "reportImplicitOverride": false,
+      "reportMissingParameterType": false,
+      "reportMissingTypeArgument": false,
+      "reportMissingTypeStubs": false,
+      "reportPrivateLocalImportUsage": false,
+      "reportPrivateUsage": false,
+      "reportUnannotatedClassAttribute": false,
+      "reportUnknownArgumentType": false,
+      "reportUnknownLambdaType": false,
+      "reportUnknownMemberType": false,
+      "reportUnknownParameterType": false,
+      "reportUnknownVariableType": false,
+      "reportUnusedParameter": false
+    }
+  ]
 }
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,3 +22,4 @@ coloredlogs==15.0.1
 docutils>=0.21.2
 parameterized==0.9.0
 granian[reload]==2.5.0
+basedpyright==1.31.0
--- a/searx/enginelib/__init__.py
+++ b/searx/enginelib/__init__.py
@@ -252,7 +252,7 @@ class Engine:  # pylint: disable=too-few-public-methods
     display_error_messages: bool
     """Display error messages on the web UI."""
 
-    proxies: dict
+    proxies: dict[str, dict[str, str]]
     """Set proxies for a specific engine (YAML):
 
     .. code:: yaml
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -9,24 +9,26 @@ usage::
 """
 
 from __future__ import annotations
+import typing as t
 
 import sys
 import copy
 from os.path import realpath, dirname
 
-from typing import TYPE_CHECKING, Dict
 import types
 import inspect
 
 from searx import logger, settings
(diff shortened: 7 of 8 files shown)


searx/engines/__builtins__.pyi
```diff
@@ -0,0 +1,43 @@
...
+logger: logging.Logger
+supported_languages: str
+language_aliases: str
+
+# from searx.engines.ENGINE_DEFAULT_ARGS
+about: dict[str, dict[str, str | None | bool]]
+categories: list[str]
+disabled: bool
+display_error_messages: bool
+enable_http: bool
```
Bnyro, line 26: It's also possible to override values from the `outgoing:` yaml settings as mentioned in https://github.com/searxng/searxng/blob/4942c9b914d45dae9aadf193c97ddeb4acc8a466/docs/admin/settings/settings_engines.rst#L59-L74, e.g. `enable_http2`. I'm not sure if it makes sense to add them here?

inetol:
> I tried various type checkers, focusing primarily on the features of the LSP. `basedyright`, or rather, the LSP `basedpyright-langserver`, convinced me.

Have we considered [Ruff](https://docs.astral.sh/ruff/)? I came across this now while looking into uv, at first glance, looks like it replaces every existing linters and formatters, like Biome in JS stack does.

return42:
> Have we considered [Ruff](https://docs.astral.sh/ruff/)? I came across this now while looking into uv, at first glance, looks like it replaces every existing linters and formatters

I had a look at ruff in some other projects I had / [hatch](https://hatch.pypa.io/latest/config/internal/static-analysis/) for example uses ruff by default.

However, I was not convinced by the quality... compared to the trio consisting of pylint, basedpyright & black.

I'm aware that many people prefer Ruff because it's generally considered the super-fast, all-purpose tool. Personally, I prefer basedpyright's LSP (I haven't encountered any speed issues yet), it's easy to integrate into the IDE, and (as far as I know) has more options.

I'd also like to keep pylint in our toolchain, as I've had good experiences with it as well. Please keep in mind that we only have a very small amount of static types. The vast majority of the code is and probably always will be dynamically typed.

For formatting, we have a third tool (black). I'm aware that it certainly seems easier to have one tool for all three tasks (the all-purpose tool). But true to the Linux principle of "do one thing and do it well," I think it's justified to have three tools in use here.

searx/engines/__builtins__.pyi
return42: >  I'm not sure if it makes sense to add them here?

:thinking: are these options really related to the engine .. for me these options are related to the `outgoing:` settings, since it overwrites those settings .. before we cover these types in the engine config, we should first port the type definitions in `settings_defaults.py` to [`msgspec.Struct`](https://jcristharif.com/msgspec/structs.html).

When we have a `msgspec.Struct` for ...
  
https://github.com/searxng/searxng/blob/5e7109cd261dcf3e8634aa011c51a134346cead2/searx/settings_defaults.py#L216-L235

we can rethink how we add this type to the engine builtins .. but I hope we will implement a Engine class before we add more and more typing for all these monkey patched stuff into the builtins.

---

BTW we should not mix all these names of config options into a flat namespace / on the long term we should replace it by `outgoing` section in the engine config.  This will simplify the coding a lot and prevents name collisions.

```yaml
- name: example
  engine: example
  shortcut: demo
  outgoing:
    enable_http2: false
      proxies:
        http:
         - http://proxy1:8080
         - http://proxy2:8080
       https:
         - socks5://user:password@proxy3:1080
         - socks5h://user:password@proxy4:1080
```

searx/engines/__builtins__.pyi
```diff
@@ -0,0 +1,43 @@
...
+max_page: int
+paging: int
+safesearch: int
+send_accept_language_header: bool
+shortcut: str
+time_range_support: int
+timeout: int
+tokens: list[str]
+using_tor_proxy: bool
+
```
return42, line 38: `EngineTraits` type at module name `traits` is missed ..

https://github.com/searxng/searxng/blob/5e7109cd261dcf3e8634aa011c51a134346cead2/searx/engines/__init__.py#L138-L139

https://github.com/searxng/searxng/blob/5e7109cd261dcf3e8634aa011c51a134346cead2/searx/enginelib/traits.py#L192

searx/engines/__builtins__.pyi
Bnyro: > BTW we should not mix all these names of config options into a flat namespace / on the long term we should replace it by outgoing section in the engine config. This will simplify the coding a lot and prevents name collisions.
>
Totally agree on that, I found it very confusing when I used it once to change `enable_http2`.

return42:
> Have we considered [Ruff](https://docs.astral.sh/ruff/)?

While implementing https://github.com/searxng/searxng/pull/5101 I looked once more into Ruff & LSP .. they had an rewrite  of the ruff-lsp last year:

- https://astral.sh/blog/ruff-v0.4.5

With the new `ruff server` (LSP) integrated in my Emacs I realized that Ruff only does linting and formatting right now.  Ruff has no implementation for a Static Type Checker and the LSP can therefore only implement a fraction of the functionality one would expect.  Even finding a (function) definition is not possible.

----

TL;DR;

basedpyright is a Static Type Checker for Python and the LSP implements (all) function one would expect, here by example we have a [Literal](https://typing.python.org/en/latest/spec/literal.html) with a catalog of names (`weather.WeatherConditionType`):

grafik

A literal not from the catalog is detected as a type issue ..

grafik

The LSP from Ruff is not able to follow type hints .. no type issue is reported (`"clear"` isn't a valid value for this Literal) and no completion catalog:

grafik

All issues Ruff LSP has found are trivials, like unknown names (e.g. `traits`):

grafik

---------

The announcement of ASTRAL last year was still...

grafik

.. but they are still far from a minimal feature set.

ASTRAL repeatedly emphasizes (with a pinch of arrogance) how fast their tools are... which should surely be a given if these tools just implement a fraction of the features.

*"Faster" cannot be equated with "more accurate."*

inetol:
That's weird. I tried `ruff` following *standard installation procedure* and it doesn't provide completion for dictionary values: (idk what is this "for")

sh20250808211512

...But it does ~resolve~ replace the type with all the literals on the warning msg (messy but working?):

sh20250808212135

...I tried setting up a simple constant with the bare type and STC works there:

sh20250808211217

Probably an LSP server bug (`ruff-lsp` isn't working either), I haven't found any issues related on their repo.

Bnyro:
Just to add my two cents:

I can confirm your experiences with `ruff` so far: It's definitely fine for formatting code, but it's certainly not a fully featured LSP that could replace Pylsp or Pyright as of now.

We could consider replacing `black` with `ruff` for formatting, but I think that there's no real benefit in switching to `ruff`. It currently runs within a second (depending on the hardware), see screenshot below

Screenshot from 2025-08-09 10-55-45

return42:
> but it's certainly not a fully featured LSP that could replace Pylsp or Pyright as of now.

I recommend to use [basedpyright](https://docs.basedpyright.com/latest/installation/ides/) (not pyright) and I install all the needed tools into my HOME by pipx:

    $ pipx install basedpyright ruff uv hatch 'python-lsp-server[all]'

Don't be confused by the *PyLSP* names:

- [`python-lsp-server`](https://github.com/python-lsp/python-lsp-server) is a fork of
- [`python-language-server`](https://github.com/palantir/python-language-server) (from palantir)
- `pylsp` is the name of the command

inetol:
> We could consider replacing `black` with `ruff` for formatting, but I think that there's no real benefit in switching to `ruff`. It currently runs within a second (depending on the hardware), see screenshot below

From what's listed in ~https://docs.astral.sh/ruff/formatter/#style-guide and comparing with Black~ https://docs.astral.sh/ruff/formatter/black/, I noticed that we have a little more magic involved in comments, we could try this in another PR and see if the changes are acceptable.

return42:
FWIW I also tested the language server from https://zubanls.com/ (alpha) .. When it comes to speed, the emphasis is also on “speed” (I'm tired of hearing it) ... Zuban compares itself to mypy in terms of speed ... but unfortunately, it doesn't come close to having the same range of functions.

I feel like I've tried every Python LSP on the planet :-) ... Most are reimplementations of linterns, but they can't be compared to pylint either. Others integrate pylint, but still don't have strong type checking.

In the IDE, pyright and its successor basedpyright work best. They show the programmer most of the problems and offer good support for autocompletion.

In the build chain, pylint provides us with additional checks that a static type checker does not have in its scope, and we have black for code formatting... I think we are in a very good position and should not change our Python toolchain any further.

-----

If there are no objections, I would like to merge the PR. 

// There will still be a lot of fine-tuning possible in the configuration in `pyrightconfig.json`, but we should first gather some experience, which will then also be possible in the IDE with the merge of this PR.

return42:
A little OT, but for the sake of completeness, here is an excerpt from my scrapbook of what I have installed for my SearXNG development tasks:

```
;; mise-en-place: https://mise.jdx.dev/installing-mise.html
;;
;;   $ mise settings set python.compile false
;;   $ mise use -g usage go node python shellcheck shfmt
;;
;;   Configure npm to install global executables to ~/.local/bin and the
;;   libraries to ~/.local/lib/node_modules/
;;
;;   $ npm config set prefix ~/.local
;;
;;   $ mise use -g jq # https://jqlang.org/
;;   $ mise use -g rust  # rustup command --> https://www.rust-lang.org/tools/install
;;   $ mise use -g github:kristoff-it/superhtml  # https://github.com/kristoff-it/superhtml
;;   ...
;;   $ mise self-update
;;   $ mise upgrade --bump  # Upgrades to latest version, bumping the version in mise.toml
;;   ...
;;   $ pipx upgrade-all
;;   ...
;;   $ npm update -g
;;   ...
;;   $ rustup update
;;
;; Python development:
;;
;;   $ pipx install \
;;       uv hatch black \
;;       basedpyright ruff  'python-lsp-server[all]' pylsp-mypy jedi-language-server zuban
;;
;; JavaScript development:
;;
;;   Typescript LSP alternative: vtsls --> https://github.com/yioneko/vtsls
;;
;;   $ npm install -g typescript-language-server typescript
;;   $ npm install -g @vtsls/language-server
;;   $ npm install -g eslint eslint-plugin-json
;;
;; Rust Development
;;
;;   $ rustup component add rust-analyzer
;;
;; Bash scripting:
;;
;;   $ mise use -g shellcheck shfmt
;;   $ npm install -g bash-language-server
;;
;; Language Servers:
;;
;;   $ npm install -g yaml-language-server
;;   $ npm install -g vscode-langservers-extracted  # html css json eslint
```

inetol:
Shouldn't the root [package.json](https://github.com/searxng/searxng/blob/6cccb46f2b2839daee13c5d12e86a7987a22e6cd/package.json) be removed, as it was only used for pyright?

return42:
FYI: last force push was a just a rebase on branch master to solve a conflict.

return42:
> Shouldn't the root [package.json](https://github.com/searxng/searxng/blob/6cccb46f2b2839daee13c5d12e86a7987a22e6cd/package.json) be removed, as it was only used for pyright?

I removed the pyright dependency [diff](https://github.com/searxng/searxng/compare/e16e931bebec7fcb0943ddcd9c114c84d9a79bbe..e78daf9320f1da81524b3320db0d7e9262251ca4) but not the entire file: The `package.json` is used to trace (dependantbot) our developer tools (NodeJS), even if we currently do not need tools from the NodeJS eco system (at least not on this level of the project), we should leave the file there so that we can add any tools we may need later .. similar to how we do it with the developer tools from the Go ecosystem:

- https://github.com/searxng/searxng/pull/5018 and its commit [`[mod] create a Go project with the developer tools from Go`](https://github.com/searxng/searxng/commit/cf2cc87594e1d91779b36dca72d8dc7e34723132)

- https://github.com/searxng/searxng/pull/2183

inetol (approved):

PR #5087, opened 2025-08-05
[feat] engines: add Marginalia

### LINKED ISSUE #3034: Engine request: Marginalia

Austin-Olacsi:
## What does this PR do?

adds Marginalia

Interesting search results. 

uncomment the engine in settings.yml
add your API key, and then use make run

## Related issues

closes https://github.com/searxng/searxng/issues/3034

### DIFF
--- a/searx/engines/marginalia.py
+++ b/searx/engines/marginalia.py
@@ -1,13 +1,43 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Marginalia Search"""
+"""`Marginalia Search`_ is an independent open source Internet search engine
+operating out of Sweden. It is principally developed and operated by Viktor
+Lofgren .
 
+.. _Marginalia Search:
+   https://about.marginalia-search.com/
+
+Configuration
+=============
+
+The engine has the following required settings:
+
+- :py:obj:`api_key`
+
+You can configure a Marginalia engine by:
+
+.. code:: yaml
+
+   - name: marginalia
+     engine: marginalia
+     shortcut: mar
+     api_key: ...
+
+Implementations
+===============
+
+"""
+from __future__ import annotations
+
+import typing as t
 from urllib.parse import urlencode, quote_plus
 from searx.utils import searxng_useragent
+from searx.result_types import EngineResults
+from searx.extended_types import SXNG_Response
 
 about = {
-    "website": 'https://marginalia.nu',
+    "website": "https://marginalia.nu",
     "wikidata_id": None,
-    "official_api_documentation": 'https://about.marginalia-search.com/article/api/',
+    "official_api_documentation": "https://about.marginalia-search.com/article/api/",
     "use_official_api": True,
     "require_api_key": True,
     "results": "JSON",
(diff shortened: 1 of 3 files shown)


Bnyro (commented):
We intentionally removed the engine previously, see https://github.com/searxng/searxng/issues/1673.

You're trying to solve the overload issue by commenting out the engine by default if I see that correctly? Could you please clarify why we should re-add it given the removal after https://github.com/searxng/searxng/issues/1673?

Austin-Olacsi:
We should re-add Marginalia because recently, the site owner @vlofgren said that he is okay with it in https://github.com/searxng/searxng/issues/3034

It's commented-out in the code, that means only a small number of people will about know about it. Additionaly this sends a User-Agent header to identify searxng traffic. 

But maybe the python module should include code that prevents the public API key from being used at all? Bassicly give an error instead of allowing it to be used?

searx/settings.yml
```diff
@@ -1277,6 +1277,13 @@ engines:
       require_api_key: false
       results: HTML
 
+  # - name: marginalia
+  #   engine: marginalia
+  #   # contact for an API key: contact@marginalia-search.com
+  #   # rate-limited API key, for test & debug: public
```
Bnyro: > But maybe the python module should include code that prevents the public API key from being used at all? Bassicly give an error instead of allowing it to be used?
>
I'd just remove this line here so that people contact for an API key instead.

Bnyro (approved):

searx/engines/marginalia.py
```diff
@@ -0,0 +1,54 @@
...
+    query_params = {
+        "count": results_per_page,
+        "nsfw": min(params['safesearch'], 1),
+    }
+
+    params["url"] = f"{base_url}/{api_key}/search/{quote_plus(query)}?{urlencode(query_params)}"
+
+    params['headers'] = {
+        'User-Agent': searxng_useragent(),
+    }
```
return42: By this, you will overwrite all headers already set / better use `params["headers"]["User-Agent"] = searxng_useragent()`  / I fixed it in my WIP commit.

searx/settings.yml
return42: Fixed in WIP / [see](https://github.com/searxng/searxng/pull/5087/commits/dda696aea4ae95c48232e629e1023c81b513523d#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR1280-R1286)

return42 (approved):
I pushed a WIP commit on top in which I 

- fixed some minor issue (see comments below),
- added online doc for the engine,
- added a `init()` routine 
- and I added some typing stuff.

@Austin-Olacsi can you please check everything works as expected after my modifications / thanks a lot :+1:

Bnyro (approved):

Austin-Olacsi:
ok we forgot to add `disabled: true`
just fixed, the new engine works good.

PR #5073, opened 2025-07-30
[mod] theme/simple: refactor codebase

inetol:
This is one of various PR to refactor the simple theme internally.

TypeScript is a superset of JavaScript, converting the entire theme to TypeScript allows us to receive much more feedback on possible issues made in package updates or our own typos, furthermore, it allows to transpile properly to lower specs. This PR couldn't be done in smaller commits, a lot of work needed to make everything *work properly*:

From https://github.com/searxng/searxng/pull/4988:
- A browser baseline has been set that requires minimum **Chromium 93, Firefox 92 and Safari 15** (proper visuals/operation on older browser versions is not guaranteed)
- LightningCSS now handles minification and prefix creation for CSS.
- All hardcoded polyfills and support for previous browser baseline versions have been removed.
- Convert codebase to TypeScript.
- Convert IIFE to ESM, handling globals with IIFE is cumbersome, ESM is the standard for virtually any use of JS nowadays.
- Vite now builds the theme without the need for `vite-plugin-static-copy`.
- `searxng.ready` now accepts an array of conditions for the callback to be executed.
- Replace `leaflet` with `ol` as there were some issues with proper Vite bundling.
- Merged `head` with `main` script, as head was too small now.
- Add `assertElement` to properly check the existence of critical DOM elements.
- `searxng.on` renamed to `searxng.listen` with some handling improvements.

From https://github.com/searxng/searxng/pull/5073:
- Lay the foundation for loading scripts granularly depending on the endpoint it's on.
- Remove vendor specific prefixes as there are now managed by browserslist and LightningCSS.
- Enabled quite a few rules in Biome that don't come in recommended to better catch issues and improve consistency.

---

Visual from baseline browsers:

<details>
<summary>Chromium 93</summary>

Search:
chrome93-search

Images: (2nd review; image fallback was not working properly)
chrome93-images

Maps:
chrome93-maps

Preferences:
chrome93-preferences

</details>

<details>
<summary>Firefox 92</summary>

Search:
firefox92-search

Images:
firefox92-images

Maps:
firefox92-maps

Preferences:
firefox92-preferences

</details>

<details>
<summary>Safari 15</summary>

Search:
safari15-search

Images:
safari15-images

Maps:
safari15-maps

Preferences:
safari15-preferences

</details>

PR #5092, opened 2025-08-07
[fix] tagesschau: crash if there's no video stream available

Bnyro:
Sometimes, there's only an `adaptivestreaming` field in `streams`, which
is usually an m3u8 file. That's however not supported by the video player
of any browser, so we can't use and must build a different url instead.

### DIFF
--- a/searx/engines/tagesschau.py
+++ b/searx/engines/tagesschau.py
@@ -101,6 +101,9 @@ def _video(item):
         title = title.replace("_vapp.mxf", "")
         title = re.sub(r"APP\d+ (FC-)?", "", title, count=1)
 
+    # sometimes, only adaptive m3u8 streams are available, so video_url is None
+    url = video_url or f"{base_url}/multimedia/video/{item['sophoraId']}.html"
+
     return {
         'template': 'videos.html',
         'title': title,
@@ -108,5 +111,5 @@ def _video(item):
         'publishedDate': datetime.strptime(item['date'][:19], '%Y-%m-%dT%H:%M:%S'),
         'content': item.get('firstSentence', ''),
         'iframe_src': video_url,
-        'url': video_url,
+        'url': url,
     }


return42:
> is usually an m3u8 file. That's however not supported by the video player
> of any browser, so we can't use and must build a different url instead.

OT for the review of this PR but related [`.m3u` and `.m3u8`](https://en.wikipedia.org/wiki/M3U) files:

In the `mediathekviewweb.py` engine we often have `m3u8` (link results) and these will not shown in the iframe we have (clicking on the "show video" button will download the  useless `m3u8` file instead of open the video).

The issue lays in the iframe we have ..

https://github.com/searxng/searxng/blob/612b76b75e9128201ff4e9042e1f5a2829f7d3cd/searx/templates/simple/result_templates/videos.html#L20-L22

I haven't tested but a video tag with a proper mime-type  might help

```html
<video>
  <source src="https://example.org/test.m3u8" type="application/vnd.apple.mpegurl">
</video>
```

Related:

- [HTTP Live Streaming](https://www.rfc-editor.org/rfc/rfc8216.html#section-4)

return42 (approved):

Bnyro:
Good point, I didn't know browsers support m3u8 without an extra JavaScript library like video.js.

However, Firefox Desktop doesn't support that either as of now :/

See https://caniuse.com/?search=M3U8

return42:
> However, Firefox Desktop doesn't support that either as of now :/

Ah .. OK .. I could play the m3u8 with the help of [Video.js](https://videojs.com/getting-started/#videojs-cdn) ..


```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <link href="https://vjs.zencdn.net/8.23.3/video-js.css" rel="stylesheet" />
    <title>m3u8 via video-js</title>
  </head>

  <body>
    <video
      id="player"
      class="video-js vjs-16-9 vjs-big-play-centered"
      controls
      preload="auto"
      data-setup="{}"
    >
      <source
        src="https://srf-vod-amd.akamaized.net/world/hls/ts20/2025/08/ts20_20250808_124500_12636258_v_webcast_h264_,q40,q10,q20,q30,q50,q60,.mp4.csmil/index-f6-v1-a1.m3u8"
        type="application/x-mpegURL"
      />
      <p class="vjs-no-js">
        To view this video please enable JavaScript, and consider upgrading to a
        web browser that
        <a href="https://videojs.com/html5-video-support/" target="_blank"
        >supports HTML5 video</a
			     >
      </p>
    </video>

    <script src="https://vjs.zencdn.net/7.7.6/video.js"></script>
  </body>
</html>
```

For simplicity, I have used the CDN in the example above / it might make sense to integrate the libraries via our client?

.. as @inetol already did for [OpenLayers](https://www.npmjs.com/package/ol) in his PR [here](https://github.com/searxng/searxng/pull/5073/files#diff-473131774f5a2c2c0df21ea7eec37068500025662182619b7d1e46135c157096R22-R23)

Bnyro:
> For simplicity, I have used the CDN in the example above / it might make sense to integrate the libraries via our client?
> 
> .. as @inetol already did for [OpenLayers](https://www.npmjs.com/package/ol) in his PR [here](https://github.com/searxng/searxng/pull/5073/files#diff-473131774f5a2c2c0df21ea7eec37068500025662182619b7d1e46135c157096R22-R23)
>
I'm not sure if it's worth the additional dependency and load times.

I don't actually think that many users play the video directly via SearXNG, I think most people click the link to get to the website to play the video and play it form there.

PR #5093, opened 2025-08-07
[fix] legacy results: published date missing

Bnyro:
The `publishedDate` has always been `None` before that change, which
causes that there are no `publishedDate`s visible for any result.

### DIFF
--- a/searx/result_types/_base.py
+++ b/searx/result_types/_base.py
@@ -446,7 +446,7 @@ class LegacyResult(dict):
     positions: list[int]
     score: float
     category: str
-    publishedDate: datetime.datetime | None = None
+    publishedDate: datetime.datetime | None
     pubdate: str = ""
 
     # infobox result
@@ -476,6 +476,7 @@ class LegacyResult(dict):
         self["positions"] = self.get("positions", "")
         self["score"] = self.get("score", 0)
         self["category"] = self.get("category", "")
+        self["publishedDate"] = self.get("publishedDate")
 
         if "infobox" in self:
             self["urls"] = self.get("urls", [])


searx/result_types/_base.py
```diff
@@ -446,7 +446,7 @@ class LegacyResult(dict):
     positions: list[int]
     score: float
     category: str
-    publishedDate: datetime.datetime | None = None
+    publishedDate: datetime.datetime | None
```
return42, line 449: I guess the default here is not the problem / this modification is not needed (compare with `MainResult.publishedDate`).

searx/result_types/_base.py
```diff
@@ -476,6 +476,7 @@ def __init__(self, *args, **kwargs):
         self["positions"] = self.get("positions", "")
         self["score"] = self.get("score", 0)
         self["category"] = self.get("category", "")
+        self["publishedDate"] = self.get("publishedDate", "")
```
return42: The default is `None` (not an empty string) .. `self.get("publishedDate")`.

return42 (changes requested):
> .. which causes that there are no `publishedDate`s visible for any result.

Oops thanks for the fix / I left some comments below.

searx/result_types/_base.py
Bnyro: If I add the default value back, the fix doesn't work anymore, i.e. dates are missing again.

I assume that if we add a default value `self.get("publishedDate")` will return that default `None` value instead of the actual dict entry we want.

return42: your are right / forget my comment, I mixed how `MainResult` (a class) and `LegacyResult` ( a `dict`) do work .. :facepalm:

return42 (approved):

PR #5099, opened 2025-08-08
[mod] ci: strict(er) exec

inetol:
All actions are pulled using the version hash, versions are handled by dependabot, and we'll have control over which actions get updated.

Replaces Trivy scanner with Docker Scout, we have recently begun analyzing the images there, and the action will keep us in sync about the problems on GHCS dashboard.

### DIFF
--- a/.github/workflows/checker.yml
+++ b/.github/workflows/checker.yml
@@ -24,17 +24,17 @@ jobs:
     runs-on: ubuntu-24.04-arm
     steps:
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
         with:
           python-version: "${{ env.PYTHON_VERSION }}"
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
         with:
           persist-credentials: "false"
 
       - name: Setup cache Python
-        uses: actions/cache@v4
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809  # v4.2.4
         with:
           key: "python-${{ env.PYTHON_VERSION }}-${{ runner.arch }}-${{ hashFiles('./requirements*.txt') }}"
           restore-keys: "python-${{ env.PYTHON_VERSION }}-${{ runner.arch }}-"
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
       - name: Prune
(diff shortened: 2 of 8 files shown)


.github/workflows/checker.yml
```diff
@@ -24,17 +24,17 @@ jobs:
     runs-on: ubuntu-24.04-arm
     steps:
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
```
return42, line 27: Why not using the tag 

    actions/setup-python@v5.6.0

should checkout https://github.com/actions/setup-python/tree/v5.6.0 which is commit https://github.com/actions/setup-python/commit/a26af69be951a213d495a4c3e4e4022e16d87065

inetol: Because those can change, a hash *not*.

> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

https://docs.github.com/en/actions/reference/security/secure-use?learn=getting_started#using-third-party-actions

return42: > Because those can change, a hash not.

OK, I understand :+1: 

>  versions are handled by dependabot, and we'll have control over which actions get updated.

When we use a hash, are you sure that the dependabot can then assign a version to the hash in order to perform sematic versioning?

I looked in the documentation but could only find this: [Dependabot determines if there is a new version of a dependency by looking at the semantic versioning of the dependency to decide whether it should update to that version.](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#about-dependabot-version-updates) ... which makes me doubt that the dependabot can handle the hashes the way we imply.

inetol: > When we use a hash, are you sure that the dependabot can then assign a version to the hash in order to perform sematic versioning?

Yes, kinda, that's why the comment on the right of the hash is for (or maybe the hash is directly connected to the tag, I don't know how this works). You can see an example commit how dependabot updates an action https://github.com/JSPaste/Library/commit/a61bd9fb1f76297f044cb03e94015aface6844e0

return42: > I don't know how this works

*After further research* .. Seems the hash is directly connected to the tag (semantic version number) and the *version* in the comment is just for better readability:

- [Dependabot now updates comments in GitHub Actions workflows referencing action versions](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/)
- https://github.com/dependabot/dependabot-core/pull/5951
- https://github.com/dependabot/dependabot-core/issues/4691

PR #5100, opened 2025-08-08
[doc] Development Quickstart: debug with the Python Debugger pdb

return42:
Related:

- https://github.com/searxng/searxng/pull/5037#issuecomment-3166851578

grafik


Bnyro (approved):
Interestingly I never tried debugging tools with Python before, can confirm that it works fine :+1:

PR #5101, opened 2025-08-08
[fix] duckduckgo weather: add type hints and fix `WEATHERKIT_TO_CONDITION`

return42:
## What does this PR do?

Follow Up of PR #4979 adding some types and fixes issues reported by the type checker (basedpyright).


Bnyro (approved):
Good catch, thank you for the follow-up!

PR #5103, opened 2025-08-09
[feat] engines: add repology.org engine for linux packages

Bnyro:
`Repology`_. `Repology`_ monitors a huge number of package repositories and
other sources comparing packages versions across them and gathering other
information. Repology shows you in which repositories a given project is
packaged, which version is the latest and which needs updating, who maintains
the package, and other related information.

.. _About Repology: https://repology.org/docs/about

The engine is set to inactive by default, because it has quite a strict rate-limit and we probably don't want our SearXNG user-agent to get blocked.

### DIFF
--- a/docs/dev/engines/online/repology.rst
+++ b/docs/dev/engines/online/repology.rst
@@ -1,0 +1,8 @@
+.. _repology engine:
+
+========
+Repology
+========
+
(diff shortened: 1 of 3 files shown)


return42 (approved):
Sorry for the delay / I just improved some type hints and some nits in the docs.

Issue #5105, opened 2025-08-09
`X-Forwarded-Proto` not handled anymore


inetol:
Seems that [werkzeug ProxyFix](https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/) did other things, like replacing `X-Forwarded-Proto` header from WSGI ENV.

I think this issue is *not that urgent*, but some interaction like saving preferences is somewhat broken and warnings appear in console in each request:

```console
WARNING:searx.botdetection: Sec-Fetch cannot be verified for non-secure requests (HTTP headers are not set/sent by the client).
```

This issue doesn't affect instances having `base_url` setting explicitly set.

Related to https://github.com/searxng/searxng/pull/4911

return42:
🤔 

> Seems that [werkzeug ProxyFix](https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/) did other things, like replacing `X-Forwarded-Proto` header from WSGI ENV.

Since the default of `x_proto=1` was not adapted to the implementation before https://github.com/searxng/searxng/pull/4911/ (could not be adapted by configuration), the value was always `1`

- [The HTTP X-Forwarded-Proto (XFP) request header is a de-facto standard header for identifying the protocol (HTTP or HTTPS) that a client used to connect to a proxy or load balancer.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Proto)

- In our [documentation](https://docs.searxng.org/admin/installation-apache.html#apache-s-searxng-site) we recommend to set the `X-Scheme` header.

https://github.com/searxng/searxng/blob/935f3fe332551482d6c8a3d5b690f12d5ad67e90/searx/flaskfix.py#L24

if `base_url` is set ..

https://github.com/searxng/searxng/blob/935f3fe332551482d6c8a3d5b690f12d5ad67e90/searx/flaskfix.py#L51

if not ..

https://github.com/searxng/searxng/blob/935f3fe332551482d6c8a3d5b690f12d5ad67e90/searx/flaskfix.py#L62-L64

--> IMO if we think this is an real issue, we should fix it in the flaskfix.py from above and implement `X-Forwarded-Proto` as the third alternative to `bas_url` and `X-Scheme`.

inetol:
> ... and `X-Scheme`.

[...]

PR #5107, opened 2025-08-10
[fix] replace ``X-Scheme`` by ``X-Forwarded-Proto`` header

return42:
The HTTP X-Forwarded-Proto (XFP) request header is a *de-facto* standard header for identifying the protocol (HTTP or HTTPS) that a client used to connect to a proxy or load balancer.[1]

The ``X-Scheme`` header was added 10 years ago, why ``X-Scheme`` was used back then and not ``X-Forwarded-Proto``, nobody knows today / possibly because ``X-Forwarded-Proto`` wasn't a *de-facto* standard back then.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Proto
[2] https://github.com/searx/searx/commit/6ef7c3276

----

Related: https://github.com/searxng/searxng/issues/5105#issuecomment-3172447170

### DIFF
--- a/utils/templates/etc/nginx/default.apps-available/searxng.conf:socket
+++ b/utils/templates/etc/nginx/default.apps-available/searxng.conf:socket
@@ -8,10 +8,10 @@ location ${SEARXNG_URL_PATH} {
     uwsgi_param    HTTP_CONNECTION       \$http_connection;
 
     # see flaskfix.py
-    uwsgi_param    HTTP_X_SCHEME         \$scheme;
+    uwsgi_param    HTTP_X_FORWARDED_PROTO  \$scheme;
     uwsgi_param    HTTP_X_SCRIPT_NAME    ${SEARXNG_URL_PATH};
 
-    # see limiter.py
+    # see botdetection/trusted_proxies.py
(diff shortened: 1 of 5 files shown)


utils/templates/etc/nginx/default.apps-available/searxng.conf:socket
```diff
@@ -8,10 +8,10 @@ location ${SEARXNG_URL_PATH} {
     uwsgi_param    HTTP_CONNECTION       \$http_connection;
 
     # see flaskfix.py
-    uwsgi_param    HTTP_X_SCHEME         \$scheme;
+    uwsgi_param    HTTP_X_FORWARDED-PROTO  \$scheme;
```
inetol:
```suggestion
    uwsgi_param    HTTP_X_FORWARDED_PROTO  \$scheme;
```

return42: [fixed](https://github.com/searxng/searxng/compare/95e9d98e9961a89d368302707efa4ffa87bd56a6..71603a69d26c3e8ce5d55f1e066a06b32d2c77b1)

inetol (approved):

PR #5115, opened 2025-08-14
Fix publishedDate format in reuters engine

ishbir:
## What does this PR do?

FIxes publishedDate format in reuters engine to encompass ISO 8601 times both with and without milliseconds.

Previously, the engine would sometimes fail saying:

```
2025-08-12 21:13:23,091 ERROR:searx.engines.reuters: exception : time data '2024-04-15T19:08:30.833Z' does not match format '%Y-%m-%dT%H:%M:%SZ'

...
  File "/usr/local/searxng/searx/engines/reuters.py", line 87, in response

    publishedDate=datetime.strptime(result["display_time"], "%Y-%m-%dT%H:%M:%SZ"),

                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...

```

Note that most queries seem to work with Reuters, but there are some results that have the additional milliseconds and fail. Regardless, the change is backwards compatible as both the formats (with and without the ms) should now parse correctly.

## Author's checklist

I tested this locally and search results have entries from Reuters.

### DIFF
--- a/searx/engines/reuters.py
+++ b/searx/engines/reuters.py
@@ -84,7 +84,7 @@ def response(resp) -> EngineResults:
                 content=result["description"],
                 thumbnail=result.get("thumbnail", {}).get("url", ""),
                 metadata=result.get("kicker", {}).get("name"),
-                publishedDate=datetime.strptime(result["display_time"], "%Y-%m-%dT%H:%M:%SZ"),
+                publishedDate=datetime.fromisoformat(result["display_time"]),
             )
         )
     return res


searx/engines/reuters.py
```diff
@@ -84,7 +84,7 @@ def response(resp) -> EngineResults:
                 content=result["description"],
                 thumbnail=result.get("thumbnail", {}).get("url", ""),
                 metadata=result.get("kicker", {}).get("name"),
-                publishedDate=datetime.strptime(result["display_time"], "%Y-%m-%dT%H:%M:%SZ"),
+                publishedDate=datetime.fromisoformat(result["display_time"]),
```
Bnyro, line 87: After this change, the result date sometimes looks like `2021-05-12 08:10:47.323000+00:00`. I think we should remove the microseconds and time zone info.
```suggestion

publishedDate=datetime.fromisoformat(result["display_time"]).replace(tzinfo=None, microsecond=0),
```

Or possibly that should be done for all templates in a preprocessing step before rendering the results, opinion @return42 ?

ishbir: Maybe the format used for rendering the datetime should not include TZ and microseconds (since it seems like a UI/UX thing)

Bnyro: > Maybe the format used for rendering the datetime should not include TZ and microseconds (since it seems like a UI/UX thing)
>
Yeah, I agree. Perhaps we should open a new issue about that.

Bnyro (approved):

Bnyro:
Thanks 👍

searx/engines/reuters.py
return42: > Maybe the format used for rendering the datetime should not include TZ and microseconds (since it seems like a UI/UX thing)

@ishbir please open a new issue .. so we have it on the list :+1:

return42 (approved):

searx/engines/reuters.py
Bnyro: I've opened an issue:

- https://github.com/searxng/searxng/issues/5121

PR #5125, opened 2025-08-18
[fix] revision of utils.HTMLTextExtractor

return42:
## What does this PR do?

Analyze an issue reported in CI only (can't be reproduced on local systems)

## Related issues

- https://github.com/searxng/searxng/pull/5073#issuecomment-3196282632

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -74,11 +74,7 @@ def gen_useragent(os_string: Optional[str] = None) -> str:
     return USER_AGENTS['ua'].format(os=os_string or choice(USER_AGENTS['os']), version=choice(USER_AGENTS['versions']))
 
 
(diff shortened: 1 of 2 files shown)


searx/utils.py
```diff
@@ -149,23 +146,28 @@ def html_to_text(html_str: str) -> str:
         >>> html_to_text('<style>.span { color: red; }</style><span>Example</span>')
         'Example'
 
-        >>> html_to_text(r'regexp: (?<![a-zA-Z]')
+        >>> html_to_text(r'regexp: (?&lt;![a-zA-Z]')
```
return42, line 149: I had to modify the test .. versions of the python interpreter implement a different handling for an `<` which opens a tag markup but did not close it, before the string ends.

PR #5129, opened 2025-08-19
[fix] move initial "JS is enabled?" (no-js) to client side

return42:
To avoid an `unsafe-inline` in the CSP header, the JS code must be moved to the
client side [1].

The `<script>` tag at the end of the HTML originates from the old implementation
of the JS client. Since PR-5073 [2] was merged, the `type` is now `module`, and
the tag must be moved to the beginning of the HTML.

> We need to inline this "JS is enabled?" thing to prevent layout shifts and
> temporary "no JS enabled" visuals as ESM scripts loads and evals everything
> deferred from initial DOM render [3]

That's true in theory, but in practice, this effect is unnoticeable because it's
masked by another effect (which we can't avoid): If we load the page with a
severely throttled connection, the HTML (result list) takes a long time to
load. Then the CSS is loaded, which also takes longer. Until the CSS has loaded,
there's no layout. A layout shift is therefore largely determined by the loading
of the HTML and CSS itself.

The running times of the ESM script can be neglected compared to the loading
times of HTML & CSS.

[1] https://github.com/searxng/searxng-docker/pull/424#issuecomment-3199494256
[2] https://github.com/searxng/searxng/pull/5073
[3] https://github.com/searxng/searxng-docker/pull/424#issuecomment-3199622504


client/simple/src/js/core/nojs.ts
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import { ready } from "./toolkit.ts";
+
+ready(() => {
+  const htmlElement = <HTMLHtmlElement>document.getElementsByTagName("html").item(0);
+  htmlElement.classList.remove("no-js");
+  htmlElement.classList.add("js");
```
inetol: Why not `document.documentElement` and we avoid a query?

```diff
-  const htmlElement = <HTMLHtmlElement>document.getElementsByTagName("html").item(0);
-  htmlElement.classList.remove("no-js");
-  htmlElement.classList.add("js");
+  document.documentElement.classList.remove("no-js");
+  document.documentElement.classList.add("js");
```

The root element will be always be `html` in HTML, so we can directly use [documentElement](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement)

Bnyro (approved):

client/simple/src/js/core/nojs.ts
return42: > The root element will be always be `html` in HTML,

Ah .. thanks for pointing out .. [fixed it](https://github.com/searxng/searxng/compare/ca5c7a5b6c9fb244c067112463fe82188811357d..d5390a0b2be5e85ede9f4b567c90598645f97a75)

PR #5131, opened 2025-08-19
[mod] typification of SearXNG: add new result type Code

return42:
This patch adds a new result type: Code

- Python class:   searx/result_types/code.py
- Jinja template: searx/templates/simple/result_templates/code.html
- CSS (less)      client/simple/src/less/result_types/code.less

----

This PR is based on https://github.com/searxng/searxng/pull/5074


searx/result_types/code.py
```diff
@@ -0,0 +1,190 @@
...
+                ("linenostart", code_block_start),
+                ("cssclass", "code-highlight"),
+                ("hl_lines", [hl - code_block_start + 1 for hl in self.hl_lines]),
+            ]:
+                kwargs[k] = kwargs.get(k, default)  # pyright: ignore[reportUnknownMemberType]
+
+            # Wrap the code inside <pre> blocks using <code>, as recommended by
+            # the HTML5 specification (default is False).  Do we need this?
+            #
+            # kwargs["wrapcode"] = kwargs.get("wrapcode", True)
```
Bnyro: I think we should always set `wrapcode` to `True`.

It doesn't change the HTML formatting when looking at the result, but it semantically declares this paragraph as a `code` block. If we only used `pre`(formatted), it's not clear that this is actually source code.

And according to https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/HTML
> Use semantic HTML, aka "The right element for the right job", because the browser provides so many built-in accessibility hooks.
>
Possibly, screen readers could also benefit from knowing its actual type?

searx/result_types/code.py
```diff
@@ -0,0 +1,190 @@
...
+                code_block_start = line_no
+
+            if last_line_no is not None and last_line_no + 1 != line_no:
+                # new code block is detected, render current code block
+                _render(**options)  # pyright: ignore[reportUnknownArgumentType]
+                # reset conditions for next code block, which first line is the
+                # current code line
+                code_block = [code_line]
+                code_block_start = line_no
+                last_line_no = line_no
```
Bnyro:
```suggestion
                previous_line_no = line_no
```
I've struggled a bit with understanding this code initially because I thought that `last_line_no` is meant to represent the last number of all codelines, e.g. (`6` if we have codelines `1,2,...,6`.

searx/result_types/code.py
```diff
@@ -0,0 +1,190 @@
...
+                lexer = guess_lexer_for_filename(self.filename, src_code)
+            except ClassNotFound:
+                lexer = None
+
+        if not lexer:
+            try:
+                lexer = guess_lexer(src_code)
+            except ClassNotFound:
+                lexer = get_lexer_by_name("text")
+        return lexer
```
Bnyro:
```suggestion
        src_code = "\n".join([l[1] for l in self.codelines])

        if self.filename:
            try:
                return guess_lexer_for_filename(self.filename, src_code)
            except ClassNotFound:
                pass

        try:
             return guess_lexer(src_code)
        except ClassNotFound:
             pass

        return get_lexer_by_name("text")
```
Perhaps this could be simplified by returning the found `lexer` early.

Bnyro (commented):
Looks great all in all!

I've just been wondering if it wouldn't make more sense to display the file path / name directly above the formatted source code instead of the title, e.g. 
Screenshot from 2025-08-30 11-47-18

Perhaps that's not related to this PR though?

Bnyro:
And I noticed that `searchcode.com` was taken down 2 days ago, so we'll have to remove it :/

See https://boyter.org/posts/searchcode-is-being-rebooted/.

return42:
> And I noticed that `searchcode.com` was taken down 2 days ago

I send a PR to set engine to inactive for now / read:

- https://github.com/searxng/searxng/pull/5170

return42:
> I've just been wondering if it wouldn't make more sense to display the file path

I added the "Filename" and in the title link, now there is the *basename* (no longer the full path name): 

grafik

searx/result_types/code.py
return42: Yes, the name is miss leading .. this is a pointer to the end of the *current* code block, I renamed the variable to: 

```
 code_block_end: int | None = None  # line where the current code ends
```

searx/result_types/code.py
return42: Yes. makes sense / [done](https://github.com/searxng/searxng/compare/9c62d45688cb3253f8b7f8b88157e48314887483..77b5822990091f414076d3bb40336ff34fe7b94f)

searx/result_types/code.py
return42: > I think we should always set `wrapcode` to `True`.

Thank you for keeping an eye on this; first I was unsure about my decision, but with you comment -> [done](https://github.com/searxng/searxng/pull/5131/files#diff-89850fe51ec4b76e79b5a89bb362f7080b339e5855ddf4a76be693f7267839a6R154)

PR #5132, opened 2025-08-20
[fix] selfhst icons: icon list url invalid, set to active

Bnyro:
- the previous CDN icon list url no longer works
- a list of all icons is mirrored to the JSDelivr CDN however
- there's no reason to set the engine to inactive now that we use public CDNs

### DIFF
--- a/searx/engines/selfhst.py
+++ b/searx/engines/selfhst.py
@@ -14,12 +14,11 @@ about = {
 categories = ['images', 'icons']
 
 
-icons_list_url = 'https://cdn.selfh.st/directory/icons.json'
-icons_cdn_base_url = 'https://cdn.jsdelivr.net'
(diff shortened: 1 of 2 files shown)


inetol:
FYI raw endpoints have an undocumented ratelimit of 5k/req hour https://github.com/github/docs/issues/8031#issuecomment-881427112

If this is an issue we should get the list directly from jsDelivr

Bnyro:
> FYI raw endpoints have an undocumented ratelimit of 5k/req hour [github/docs#8031 (comment)](https://github.com/github/docs/issues/8031#issuecomment-881427112)
> 
> If this is an issue we should get the list directly from jsDelivr
>
True, that's a better solution as we've already been using the CDN for fetching the image resources anyways.

inetol (approved):

PR #5133, opened 2025-08-20
[feat] client/simple: move cursor to end of search input on mobile

### LINKED ISSUE #5112: feat: Cursor placement at end of search input for easier editing

Using SearXNG on mobile, the cursor in the search input does not automatically go to the end of search query. This means that the user have to scroll horizontally to add to or edit the end of search query which is inconvenient and impacts usability

Automatically place the cursor at the end of the existing text within the search bar when clicked in mobile view. This would allow user to immediately type and add to the search query without any additional scrolling or interaction

I have observed that other search engines, such as Brave and many browser's address bars, already implement this behavior. They automatically position the cursor at the end of the text, providing a smoother user experience

**Additional context**

https://github.com/user-attachments/assets/8c3248ee-22db-40c3-b7f2-85db4f996555

amit-katyal:
## What does this PR do?

This PR introduces a new behavior on mobile devices: when a user focuses on the search input box, the cursor is automatically moved to the end of any existing text in the input.

This is achieved by adding a `focus` event listener to the search input element. The listener is only active on mobile screen sizes and uses `setTimeout` to ensure it executes after the browser's native focus handling.

This change improves the user experience and accessibility on mobile devices by making it easier to edit or add to an existing search query without manually repositioning the cursor.

1.  Open your browser's developer tools and switch to a mobile view
2.  Type text into the search bar and make a search
3.  Click outside the search bar, then click back in
4.  Verify the cursor is at the end of the text

## Author's checklist

This change is purely client-side and only affects the search bar on mobile devices.

## Related issues

Closes #5112


Bnyro (approved):

client/simple/src/js/main/search.ts
```diff
@@ -40,6 +40,18 @@ if (!(isMobile || isResultsPage)) {
...
 
+// On mobile, move cursor to the end of the input on focus
+if (isMobile) {
+  listen("focus", qInput, () => {
+    // Use a timeout to run this after the browser's own focus handling
+    setTimeout(() => {
+      const end = qInput.value.length;
+      qInput.setSelectionRange(end, end);
+      qInput.scrollLeft = qInput.scrollWidth;
+    }, 0);
```
inetol: Instead of racing against the event loop, we can directly override the cursor before the next frame is rendered, so the user won't see the cursor travelling from one position to another:

```diff
-    // Use a timeout to run this after the browser's own focus handling
-    setTimeout(() => {
-      const end = qInput.value.length;
-      qInput.setSelectionRange(end, end);
-      qInput.scrollLeft = qInput.scrollWidth;
-    }, 0);
+    // before rendering the next frame
+    requestAnimationFrame(() => {
+      const end = qInput.value.length;
+      qInput.setSelectionRange(end, end);
+      qInput.scrollLeft = qInput.scrollWidth;
+    });
```

inetol (changes requested):
Thanks for looking at this. I've noticed some issues though, sometimes the cursor doesn't move to the end or the cursor appears to jump around when clicking on the input. See my recommendations below:

client/simple/src/js/main/search.ts
amit-katyal: Thanks for the feedback! I’ve made the requested changes

inetol (approved):

return42 (approved):
I had to rebase to solve a conflict with current master .. BTW I squashed the commits and moved the modifications of the files in the `/static` folder to a separated commit .. we do this to have diff-friendly commits .. @amit-katyal  please see `make static.build.commit` in our [Development Quickstart](https://docs.searxng.org/dev/quickstart.html#development-quickstart) guide .. additional, for next PR please create a feature branch (don't use your master branch for PRs).

Nevertheless, thank you very much for your contribution to SearXNG :+1:

-----

> This change improves the user experience and accessibility on mobile devices by making it easier to edit or add to an existing search query without manually repositioning the cursor.

True as long the query term is longer as the input field .. but on small search terms, the user can't no longer positioning the courser somewhere in the search term  .. however, I'm not often use mobile devices therefore I can't judge on such question.  

Long story told short: lets merge and see what the majority says about.

amit-katyal:
Thanks @return42 for merging and for the feedback! I'll use feature branches and `make static.build.commit` going forward.

On the cursor positioning, you're right about short queries, but this matches how other mobile search engines work (Brave, Google, etc.). Users typically want to add to their search rather than edit the middle on mobile. They can still tap to reposition if needed, after initial interaction.

Let's see what the community thinks! Thanks for the merge 👍

PR #5147, opened 2025-08-23
[mod] addition of various type hints / tbc

return42:
- pyright configuration [1]_
- stub files: types-lxml [2]_
- addition of various type hints
- enable use of new type system features on older Python versions [3]_
- ``.tool-versions`` - set python to lowest version we support (3.10.18) [4]_:
  Older versions typically lack some typing features found in newer Python
  versions.  Therefore, for local type checking (before commit), it is necessary
  to use the older Python interpreter.

.. [1] https://docs.basedpyright.com/v1.20.0/configuration/config-files/
.. [2] https://pypi.org/project/types-lxml/
.. [3] https://typing-extensions.readthedocs.io/en/latest/#
.. [4] https://mise.jdx.dev/configuration.html#tool-versions

These changes collectively enhance code maintainability, ensure better compatibility with type checkers and development tools, and improve the overall robustness of the project.

----

To avoid further complicating static typing due to the quirks that still exist in Python 3.9, it was also necessary to remove Python 3.9 from the support matrix. This PR is based on:

- https://github.com/searxng/searxng/pull/5148

----

This is intended to form a first basis for the typification of our source code.  Much more will be required, but I already mentioned it in [PR-5075](https://github.com/searxng/searxng/pull/5075):

> *I think we need to put more effort into typing, but we shouldn't try to type everything either; that just creates unnecessary work and is frustrating. Python is and remains a dynamically typed language, but types can improve quality and, with the LSP (IDE), positively impact productivity. In this sense, we should also align our goals with regard to typing.*

The changes in this [PR-5147](https://github.com/searxng/searxng/pull/5147) are also designed in this sense.

Some people might think, what could be so difficult about retroactively typing code?  If you work with static types in a project from the beginning, it is still easy, but if you retroactively typing the code there are some challenges.

Basically, you have to re-engineer the code to find the correct types.  Wich is not feasible with reasonable effort for modules such as searx.network or searx.metrics. Here, one should consider whether these modules might be replaced by new modules in the future (network, for example, based on *curl impoersonate* / metrics replaced by container classes).

However, there will always be a lot of code in SearXNG that cannot be (reasonably) typed; for example, the JSON and DOM trees that are examined by the engines. Requirements like pygments also have generic interfaces (classes) that make static type checking impossible.

I/we are still at the very beginning. With a little more experience, we'll be able to provide (write down) somewhat helpful recommendations for action, and as the quality of our source code improves, we'll also be increasingly able to implement some criteria in CI as quality gates. But we still have a long way to go.

But all the complaining doesn't help, we have to start with the topic... this PR takes us a first step in the right direction and the benefits can already be felt with the merge of this PR in the developers' IDEs.

----

We use basedpyright, which can be easily installed via pipx, for example:

     pipx install basedpyright

Its recommended to have a look at the following links:

- https://docs.basedpyright.com/latest/installation/command-line-and-language-server/
- https://docs.basedpyright.com/latest/installation/ides/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,7 @@ trim_trailing_whitespace = true
 end_of_line = lf
 charset = utf-8
 
-[*.py]
+[{*.py,*.pyi}]
 # code formatter accepts length of 120, but editor should prefer 80
 max_line_length = 80
 
--- a/.pylintrc
+++ b/.pylintrc
@@ -311,7 +311,7 @@ dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_
 ignored-argument-names=_.*|^ignored_|^unused_
 
 # Tells whether we should check for unused import in __init__ files.
-init-import=no
+init-import=yes
 
 # List of qualified module names which can have objects that can redefine
 # builtins.
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
-nodejs 24.3.0
-python 3.13.1
+nodejs     24.3.0
+python     3.10.18
 shellcheck 0.10.0
-sqlite 3.47.2
+sqlite     3.47.2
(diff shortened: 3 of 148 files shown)


return42:
FYI: This PR modifies a lot of files and therefore has high conflict potential .. further force pushes (like the last ones) are just to solve conflicts with current master branch.

return42:
@inetol I'm very sorry, I have seen to late, that you had self-requested a review for this PR.  Sadly matrix.org is down at this time I can't connect you that way.

PR #5149, opened 2025-08-25
[mod] optimize Baidu Captcha suspended time

hezhijie0327:
## What does this PR do?

Set custom suspended time for Baidu Captcha.

---

Note:
1. As far as investigating, seems the suspended time is around 5min. 

Current behavior is follow the default `SearxEngineCaptcha: 86400`.

`!baidu test`


return42:
Closed due to inactivity.

PR #5172, opened 2025-08-31
[feat] naver engine: add video embeds

Austin-Olacsi:
## What does this PR do?

add video embeds for naver engine

it was missing

make run
!nvrv test


Bnyro (approved):
Thanks!

PR #5173, opened 2025-09-01
[fix] chinaoso: add random `uid` to cookie

hubutui:
## What does this PR do?

add random uid to cookie

without this cookie setting, chinaso do not return the search results.


return42 (approved):
> add random uid to cookie

Works like a charm :-) / thanks a lot :+1:

PR #5185, opened 2025-09-05
[enh] docs: update README/CONTRIBUTING

inetol:
README should be concise, everything else goes to CONTRIBUTING, docs page...

---

https://github.com/inetol/searxng/blob/enh-readme/README.rst
https://github.com/inetol/searxng/blob/enh-readme/CONTRIBUTING.rst


CONTRIBUTING.rst
```diff
@@ -0,0 +1,92 @@
+.. SPDX-License-Identifier: AGPL-3.0-or-later
+
+.. _Quickstart guide: https://docs.searxng.org/dev/quickstart.html
+.. _Commits guide: https://docs.searxng.org/dev/commits.html
+.. _Weblate: https://translate.codeberg.org/projects/searxng/searxng/
+.. _GitHub Codespaces: https://docs.github.com/en/codespaces/overview
+.. _120 hours per month: https://github.com/settings/billing
+.. _list of existing Codespaces: https://github.com/codespaces
```
Bnyro, line 8: Perhaps these links should be moved down to the Codespaces section?

return42 (approved):

CONTRIBUTING.rst
inetol: rest anchors can be referenced everywhere in the document, including on top of the anchor itself. As a general rule, global defined *things* should always be at the top of everything else.

Bnyro (approved):

CONTRIBUTING.rst
return42: > As a general rule, global defined things should always be at the top of everything else.

Yeah, its a good rule .. However, in (plain text) documents, it's not uncommon to resolve references into footnotes. Since there are no pages in a file, the footer could be far away (and the reader needs to scroll). I usually insert the links in the chapter in which I use them, but that's not really a better rule. I sometimes wonder what's better: "close" to the use or perhaps right at the top/bottom, but then "far away" from the use. It will probably be decided on a case-by-case basis; either is fine.

----

Related: [reST primer](https://docs.searxng.org/dev/reST.html#rest-primer)

PR #5191, opened 2025-09-09
[upd] disable public domain image archive engine by default

vojkovic:
## What does this PR do?

Disables the Public Domain Image Archive engine by default. 

Will massively reduce the amount of bot and spam traffic directed toward the PDIA engine.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1733,6 +1733,7 @@ engines:
   - name: public domain image archive
     engine: public_domain_image_archive
     shortcut: pdia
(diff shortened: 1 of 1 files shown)


return42 (approved):
@Denperidge please take note of this PR, we disable the engine by default.  The pdimagearchive.org is a small non-profit organization, which suffer significantly from requests from the SearXNG instances. We are disabling the engine by default to improve the situation for their maintainers / If that isn't sufficient, we may have to consider other options in a subsequent step.

Issue #4869, opened 2025-05-29
Feature: Add support for multiple YaCy servers to SearXNG engines.


smokingwheels:
Working URL to the engine:
I am running a YaCy instance at http://58.96.112.6:8091 on a volatile RAM drive.
After every reboot, the YaCy data is automatically restored from Nextcloud account.

I have access to many peers, but to avoid overload, I currently limit SearXNG to querying only one peer on the server at a time. For ongoing monitoring and testing, I occasionally may replace that peer with a different instance, same address and port to evaluate performance and stability under real-world conditions. Peer may go offline or be restored independently.

Why do you want to add this engine?
YaCy is a decentralized, open-source search engine that doesn’t track its users. Running multiple peers increases resiliency and diversity of results, and offers a unique privacy-respecting alternative for meta-search.

Features of this engine:

Open source and privacy-focused.

Decentralized, federated search.

Can be run in volatile (RAM) mode for speed; my instance is restored from Nextcloud after reboot for fast recovery

Users can add their own YaCy peers for results.

Handles cases where the peer returns few or no results (I patched the engine definition for this) 

How can SearXNG fetch the information from this engine?
Via the YaCy HTTP API, e.g.:

http://58.96.112.6:8091/yacysearch.json?query=test

My working SearXNG engine definition is set up to query 3 single specified peers.
The engine gracefully handles if the peer returns partial results.

Applicable category of this engine:

general

Additional context:

My YaCy peer is running in RAM, so crashes may occur; the engine is designed to be robust against failures and restores automatically via Nextcloud.

Allowing SearXNG users to specify their own YaCy peers increases decentralization and privacy.

Thanks

Bnyro:
Please remove that copied code from the issue description, or create a pull request to propose your changes.

The current way it's hard to understand what you're actually proposing here. SearXNG already supports Yacy, so I don't really get whay you're asking for here?

smokingwheels:
I would like it to be easy for users to add their own YaCy peer(s) to SearXNG, simply by specifying the HTTP URL (e.g., http://myyacy.example.com:8090). Ideally, this could be done via the SearXNG preferences UI or a simple config field, so users can connect to their own decentralized search peer without manually editing engine YML files.

Use Case:

Many YaCy users run their own instance, sometimes on localhost or a home server.

Making it straightforward to add your own YaCy peer would encourage more privacy-respecting, decentralized search and community experimentation.

Suggestions:

[...]

Issue #3953, opened 2024-10-21
KeyError when YaCy config is not named "yacy"


LillySchramm:
> Version: `2024.10.19+050451347` (Docker)

It is currently not possible to name a YaCy engine config anything other that `yacy`. It should be possible to do so.

The following config causes said behavior:
```yaml
engines:
  - name: a_name
    engine: yacy
    categories: general
    search_type: text
    base_url:
      - https://search.example.com
    shortcut: ya
    timeout: 4.0
```
```
searxng-1  | 2024-10-21 12:06:23,796 WARNING:searx.engines.ysrv-yacy: ErrorContext('searx/engines/yacy.py', 118, "url = engines['yacy'].base_url  # type: ignore", 'KeyError', None, ()) False
searxng-1  | 2024-10-21 12:06:23,796 ERROR:searx.engines.ysrv-yacy: exception : 'yacy'
searxng-1  | Traceback (most recent call last):
searxng-1  |   File "/usr/local/searxng/searx/search/processors/online.py", line 163, in search
searxng-1  |     search_results = self._search_basic(query, params)
searxng-1  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng-1  |   File "/usr/local/searxng/searx/search/processors/online.py", line 137, in _search_basic
searxng-1  |     self.engine.request(query, params)
searxng-1  |   File "/usr/local/searxng/searx/engines/yacy.py", line 141, in request
searxng-1  |     params["url"] = f"{_base_url()}/yacysearch.json?{urlencode(args)}"
searxng-1  |                        ^^^^^^^^^^^
searxng-1  |   File "/usr/local/searxng/searx/engines/yacy.py", line 118, in _base_url
searxng-1  |     url = engines['yacy'].base_url  # type: ignore
searxng-1  |           ~~~~~~~^^^^^^^^
searxng-1  | KeyError: 'yacy'
```
When setting `name: yacy`, this issue does not exist.

*I might fix this issue myself at some point*

return42:
The KeyError comes from the `yacy images` engine:

https://github.com/searxng/searxng/blob/050451347b021d05d26c7a0797c790bbd83442e4/searx/settings.yml#L2182-L2184

As you can see above, the `yacy images` engine did use the `network` from an engine named `yacy`  but this engine does no longer exists, when you rename the engine to `a_name`.

LillySchramm:
[...]

PR #5193, opened 2025-09-10
Fix base_url conflict with YaCy engines

xylobol:
## What does this PR do?

This is a quick bugfix for an issue where you cannot have multiple YaCy engines with different base_urls. See #4869 for more details.

This allows administrators to add multiple YaCy engines. Personally, I use this to keep an index of web results and an index of results from my intranet separate.

Run searxng with multiple YaCy engines configured, each with a different `base_url`, and with none named `yacy`. 

## Author's checklist

I don't know what to put in this section. I've tested it locally after building in Docker, and I'm running it in my homelab right now. I've also made sure the default configuration works. 

## Related issues

Would help with #4869 and #3953.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -179,3 +179,4 @@ features or generally made SearXNG better:
 - Tommaso Colella `<https://github.com/gioleppe>`
 - @AgentScrubbles
 - Filip Mikina `<https://github.com/fiffek>`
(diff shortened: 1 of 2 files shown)


return42:
Thanks a lot for your work on this, please follow my comment in the issue: https://github.com/searxng/searxng/issues/4869#issuecomment-3273359285

PR #5199, opened 2025-09-11
[enh] container: build with uv

inetol:
The cache improvements in https://github.com/searxng/searxng/pull/5086 didn't work as expected because pip doesn't seem to offer reproducible builds (?). Alas, until we migrate to native `uv` management in some future PR, we will not be able to start using lockfiles which will limit us to provide truly reproducible installs. Instead, we can try to use cached wheels as we do now to have venv reproducibility (I hope).

---

CI (outdated): https://github.com/inetol/searxng/actions/runs/17650638050

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -104,8 +104,6 @@ jobs:
     needs: build-base
     strategy:
       fail-fast: false
-      # Faster runners first to cache arch independent wheels
-      max-parallel: 1
       matrix:
         include:
           - arch: amd64
@@ -121,8 +119,6 @@ jobs:
     permissions:
       # Organization GHCR
       packages: write
-      # Clean key cache step
-      actions: write
 
     outputs:
       docker_tag: ${{ steps.build.outputs.docker_tag }}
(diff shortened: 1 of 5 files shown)


.github/workflows/container.yml
```diff
@@ -146,23 +142,13 @@ jobs:
...
-      - if: steps.cache-container-mounts.outputs.cache-hit == 'true'
-        name: Clean key cache container mounts
-        continue-on-error: true
-        env:
-          GH_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-        run: gh cache delete container-mounts-${{ hashFiles('./container/*.dockerfile') }}
+          # FIXME: we don't have uv.lock yet, so use requirements hash
+          key: "container-uv-${{ matrix.arch }}-${{ hashFiles('./requirements*.txt') }}"
+          restore-keys: "container-uv-${{ matrix.arch }}-"
+          path: "/var/tmp/buildah-cache-1001/uv/"
```
return42: > The cache improvements in #5086 didn't work as expected

By "cache" you mean `actions/cache` .. right?

> because pip doesn't seem to offer reproducible builds, whereas with uv we can do this (in theory). 

:thinking: really? .. without a `.lock` file? .. can't believe .. even if we pin the patch levels, the (sub-sub-) dependencies we have in there won't all pin the patch level.

Isn't the `hashFiles('./requirements*.txt')` here not the same as before this patch? .. or at least I think for this hash a `pyproject.toml` isn't needed.

----

> but we can start playing around with faster container (re)builds.

Yes, lets "play" .. but I still hesitate to add a `pyproject.toml`, read my concerns:

- https://github.com/searxng/searxng/pull/4141#issuecomment-2590167849

inetol: > By "cache" you mean `actions/cache` .. right?

Yes, We are now storing all 3 arch wheels (amd64, arm64, armv7) in `container-mounts-<hash>` GitHub [cache](https://github.com/searxng/searxng/actions/caches). Having this helped in not refetching every wheel while building containers.

> 🤔 really? .. without a `.lock` file? .. can't believe .. even if we pin the patch levels, the (sub-sub-) dependencies we have in there won't all pin the patch level.

*We can't* have reproducible install if we don't use a lockfile, yes. But we could if we migrate pip to use uv native in the future (I updated the PR description)

> Isn't the `hashFiles('./requirements*.txt')` here not the same as before this patch?

No, the prior implementation of the hash was kinda useless as we were always downloading the pip cache, using cached wheels or downloading new versions and forcing a reupload in a serialized way for every arch.

Now we can cache every arch independently and rotate when `requirements*.txt` is modified, as we can purge prebuilt wheels with uv (https://docs.astral.sh/uv/reference/cli/#uv-cache-prune--ci). But I will probably revert this again if I see that child package versions changes frequently and prevents reproducibility (if we can reuse the same venv on various container versions in the same day, will be enough for me)

The FIXME was misleading and erroneous, I removed that.

> Yes, lets "play" .. but I still hesitate to add a `pyproject.toml`, read my concerns:
> 
> * [[enh] add pyproject.toml listing build requirements #4141 (comment)](https://github.com/searxng/searxng/pull/4141#issuecomment-2590167849)

I discovered that uv has a *[pip like compat command](https://docs.astral.sh/uv/pip/)* that doesn't require the creation of  `pyproject.toml` so we can delay this issue for now.

return42 (approved):
LGTM

PR #5204, opened 2025-09-14
[mod] addition of various type hints / engine processors

return42:
Continuation of #5147 .. typification of the engine processors.

BTW:

- removed obsolete engine property https_support
- fixed & improved currency_convert
- engine instances can now implement a `engine.setup()` method

### DIFF
--- a/searx/engines/currency_convert.py
+++ b/searx/engines/currency_convert.py
@@ -1,53 +1,58 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Currency convert (DuckDuckGo)
-"""
+"""Currency convert (DuckDuckGo)"""
 
+import typing as t
 import json
 from searx.result_types import EngineResults
 
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineCurrenciesParams
+    from searx.extended_types import SXNG_Response
+
 # about
 about = {
-    "website": 'https://duckduckgo.com/',
-    "wikidata_id": 'Q12805',
-    "official_api_documentation": 'https://duckduckgo.com/api',
+    "website": "https://duckduckgo.com/",
+    "wikidata_id": "Q12805",
+    "official_api_documentation": "https://duckduckgo.com/api",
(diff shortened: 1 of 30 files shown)


searx/engines/currency_convert.py
```diff
@@ -1,53 +1,58 @@
...
+    from searx.extended_types import SXNG_Response
+
 # about
 about = {
-    "website": 'https://duckduckgo.com/',
-    "wikidata_id": 'Q12805',
-    "official_api_documentation": 'https://duckduckgo.com/api',
+    "website": "https://duckduckgo.com/",
+    "wikidata_id": "Q12805",
+    "official_api_documentation": "https://duckduckgo.com/api",
```
inetol, line 16: Perhaps we should remove the `--skip-string-normalization` flag in Black

https://github.com/searxng/searxng/blob/23257bddce864cfc44d64324dee36b32b1cf5248/manage#L53

inetol (approved):

searx/engines/currency_convert.py
return42: Do we want to do this now? .. I would like to do that someday.

I haven't had the courage to do it yet... it will lead to a lot of diffs and merge conflicts. As part of my typing changes, I'm touching a lot of code and have to rewrite some of it. As part of these changes, I've been performing string normalization manually.

inetol: I'm only acknowledging this as I don't know the reason it was excluded in the first place

return42: >  I don't know the reason it was excluded in the first place

Unfortunately, we have many quirks that can only be explained by history ...

The implementations for SearXNG (formerly searX) began in 2014. Back then, many of the modern code formatter (and checkers) didn't exist. At some point, we introduced pycodestyle and then switched to black. However, we wanted to avoid simultaneously modify all source code with the introduction or the change of the code formatter. --> https://github.com/searxng/searxng/pull/619#issuecomment-999502236

PR #5220, opened 2025-09-19
[mod] py: remove uvloop

inetol:
We get some good stuff without uvloop, 13MB~ less of dependencies, 3 minutes of container build time for armv7 saved, and we are one step closer to NT compatibility. Although it's true that theoretically the raw performance have worsened on network side (we only used uvloop for that), the latest cpython versions have been improving on asyncio performance.

---

This change is motivated by cpython 3.14 release preparations


return42:
``uvloop`` was introduced with httpx [1], presumably to "optimize" the performance of outgoing requests. We have several such "optimizations" in the code whose effectiveness is rarely verified in the context of real applications.

Typically, we only have a few (perhaps 10) outgoing requests to the engines, and these usually have a timeout of 3 seconds and an average response time of perhaps 1 second. Furthermore, SearXNG instances are usually quite "small" and may have 1-5 incoming requests per second. Under these conditions, performance optimizations in the millisecond range are at least questionable.

The image proxy could possibly benefit from uvloop... but until this is confirmed, we can only speculate here.... Therefore, I would also agree that we should remove uvloop for the time being. If, as expected, our users experience significant performance losses again, we can reconsider "optimizations" in this area.

[1] https://github.com/searx/searx/commit/eaa694fb7

return42 (approved):
FYI: to test image proxy I had to rebase on current master

Issue #5223, opened 2025-09-20
Slow start on containers


inetol:
Update to [`2025.9.28-2f0d96b8b`](https://github.com/searxng/searxng/pkgs/container/searxng/528414927?tag=2025.9.28-2f0d96b8b) or later.

---

Since commit https://github.com/searxng/searxng/commit/a0d2ecf43466083925291e749ab9c2a451f6963d , startup times are considerably slower (dependent on hardware)

At first, I thought it might be the py bytecode that was being invalidated, and it could have been the case (as there were changes related to the timestamps), but after PR https://github.com/searxng/searxng/pull/5225, nothing seems to have changed.

Profiling the startup show the following: https://github.com/searxng/searxng/issues/5223#issuecomment-3323083411

return42:
> With a little more digging, it turns out that the `/tmp/` cache databases are to blame (?)

Yes, a reboot (when `/tmp` is dropped) will drop all caches (the SQLite DBs) we have stored in `/tmp`.

I implemented a lazy loading of the `data` caches .. 

https://github.com/searxng/searxng/blob/b7ecc1c240f92252f061745d287accc886f3a742/searx/data/__init__.py#L85-L92

However, I cannot say when which data is loaded into the DB tables .. may some expensive tables are filled very early, when the runtime is build up.

> only the first start would be considerable slower than the next ones.

Are the delays acceptable or should we take another look at what can be done to improve the lazy load of the DBs?

donbowman:
prevents coming online in kubernetes since the readiness probe is only 10seconds/3 failures.

inetol:
> Are the delays acceptable or should we take another look at what can be done to improve the lazy load of the DBs?

I only saw that it got stuck for 3 seconds before starting the worker, but from what they said at https://github.com/searxng/searxng/issues/5218#issuecomment-3314857672, I now have my doubts.

Debug doesn't reveal anything either. It looks a bytecode invalidation issue, but when every module loads, no new pyc were generated (as expected)

inetol:
@return42 Might be of interest, I have attached the pyinstrument profiler in [`init`](https://github.com/searxng/searxng/blob/a57b29b009ca65b2ed49e09fb000fc1a6b58153e/searx/webapp.py#L1372) func inside searx.webapp:

<details>
<summary>First (cold) start</summary>

```console
Profile at /usr/local/searxng/searx/webapp.py:1421

6.068 <module>  searx/webapp.py:1
[...]

PR #5225, opened 2025-09-20
[enh] container: compact venv

inetol:
We can leverage the immutable nature of containers to add additional optimizations.

No debugging or tinkering inside containers, so stripping all unused symbols inside `venv` should be fine. We are also going to compile the bytecode ourselves to modify some parameters related to reproducibility and optimizations.

With these small changes, we have reduced the `venv` layer size by 10MB~

### DIFF
--- a/container/builder.dockerfile
+++ b/container/builder.dockerfile
@@ -1,17 +1,23 @@
 FROM ghcr.io/searxng/base:searxng-builder AS builder
 
-ARG TIMESTAMP_VENV="0"
-
 COPY ./requirements.txt ./requirements-server.txt ./
+
+ENV UV_NO_MANAGED_PYTHON="true"
+ENV UV_NATIVE_TLS="true"
+
+ARG TIMESTAMP_VENV="0"
 
 RUN --mount=type=cache,id=uv,target=/root/.cache/uv set -eux -o pipefail; \
     export SOURCE_DATE_EPOCH="$TIMESTAMP_VENV"; \
     uv venv; \
-    uv pip install --no-managed-python --compile-bytecode --requirements ./requirements.txt --requirements ./requirements-server.txt; \
+    uv pip install --requirements ./requirements.txt --requirements ./requirements-server.txt; \
     uv cache prune --ci; \
+    find ./.venv/lib/ -type f -exec strip --strip-unneeded {} + || true; \
+    find ./.venv/lib/ -type d -name "__pycache__" -exec rm -rf {} +; \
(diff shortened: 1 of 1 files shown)


container/builder.dockerfile
```diff
@@ -1,25 +1,31 @@
...
 RUN --mount=type=cache,id=uv,target=/root/.cache/uv set -eux -o pipefail; \
     export SOURCE_DATE_EPOCH="$TIMESTAMP_VENV"; \
     uv venv; \
-    uv pip install --no-managed-python --compile-bytecode --requirements ./requirements.txt --requirements ./requirements-server.txt; \
+    uv pip install --requirements ./requirements.txt --requirements ./requirements-server.txt; \
     uv cache prune --ci; \
+    find ./.venv/lib/ -type f -exec strip --strip-unneeded {} + || true; \
+    find ./.venv/lib/ -type d -name "__pycache__" -exec rm -rf {} +; \
+    find ./.venv/lib/ -type f -name "*.pyc" -delete; \
+    python -m compileall -q -f -o 2 -j 0 --invalidation-mode=unchecked-hash ./.venv/lib/; \
```
return42: Not sure [`-o 2`](https://docs.python.org/3/using/cmdline.html#cmdoption-OO) and the `PYTHONOPTIMIZE=2` in the entrypoint are a really good idea ..
 
> With these small changes, we have reduced the venv layer size by 10MB~

Yes, but all doc-strings and asserts had been removed .. We don't really know what effects this has on all the (sub- , sub-) dependencies ..

You can try it out in Docker, but usually such optimizations only cause more problems.

container/entrypoint.sh
```diff
@@ -127,4 +127,4 @@ volume_handler "$DATA_PATH"
 # Check for files
 config_handler "$SEARXNG_SETTINGS_PATH" "/usr/local/searxng/searx/settings.yml"
 
-exec /usr/local/searxng/.venv/bin/granian searx.webapp:app
+exec env PYTHONOPTIMIZE=2 /usr/local/searxng/.venv/bin/granian searx.webapp:app
```
return42: Is this PR related to the "Slow start on containers"? (https://github.com/searxng/searxng/issues/5223#issuecomment-3314995089)

inetol: Nope

container/builder.dockerfile
inetol: I'll revert this better. When I tested it, I didn't see any issues, but if some package updates it might break something

PR #5235, opened 2025-09-23
[feat] engine: add Azure resources engine

gioleppe:
## What does this PR do?
This PR adds an engine to search cloud resources on Azure.
It mimics the behavior of the search bar in Azure (but just for resources and resource groups)
I've added the engine and some explanations in the .py file.

A lot of enterprise users have to deal with Azure Public Cloud. 
This helps them easily search for cloud resources without logging in to the Portal first 
(also, it helps me as well :D)

As described, 

1. You should create an App Registration on Azure Entra Id with Reader access on the resources you want to search for.
2. You should create a Secret for the App Registration
3. After that, you should set up appropriate tenant_id, client_id, client_secret values in the settings.yml file
4. Search & enjoy :)

Please let me know if you have comments or tips on my PR, and thanks for your efforts on this great piece of software!

### DIFF
--- a/searx/engines/azure.py
+++ b/searx/engines/azure.py
@@ -1,0 +1,190 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine for Azure resources.  This engine mimics the standard search bar in Azure
+Portal (for resources and resource groups).
+
+Configuration
+=============
+
+You must `register an application in Microsoft Entra ID`_ and assign it the
+'Reader' role in your subscription.
+
+To use this engine, add an entry similar to the following to your engine list in
+``settings.yml``:
+
+.. code:: yaml
+
+   - name: azure
+     engine: azure
+     ...
+     azure_tenant_id: "your_tenant_id"
+     azure_client_id: "your_client_id"
+     azure_client_secret: "your_client_secret"
+     azure_token_expiration_seconds: 5000
+
+.. _register an application in Microsoft Entra ID:
+    https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app
+
+"""
+import typing as t
+
+from searx.enginelib import EngineCache
+from searx.network import post as http_post
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+engine_type = "online"
+categories = ["it", "cloud"]
+
+# Default values, should be overridden in settings.yml
+azure_tenant_id = ""
+azure_client_id = ""
+azure_client_secret = ""
+azure_token_expiration_seconds = 5000
+"""Time for which an auth token is valid (sec.)"""
+azure_batch_endpoint = "https://management.azure.com/batch?api-version=2020-06-01"
+
+about = {
+    "website": "https://www.portal.azure.com",
+    "wikidata_id": "Q725967",
+    "official_api_documentation": "https://learn.microsoft.com/en-us/\
+    rest/api/azure-resourcegraph/?view=rest-azureresourcegraph-resourcegraph-2024-04-01",
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": "JSON",
+    "language": "en",
+}
+
+CACHE: EngineCache
+"""Persistent (SQLite) key/value cache that deletes its values after ``expire``
+seconds."""
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    """Initialization of the engine.
+
+    - Instantiate a cache for this engine (:py:obj:`CACHE`).
+    - Checks whether the tenant_id, client_id and client_secret are set,
+      otherwise the engine is inactive.
+
+    """
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+
+    missing_opts: list[str] = []
+    for opt in ("azure_tenant_id", "azure_client_id", "azure_client_secret"):
+        if not engine_settings.get(opt, ""):
+            missing_opts.append(opt)
+    if missing_opts:
+        logger.error("missing values for options: %s", ", ".join(missing_opts))
+        return False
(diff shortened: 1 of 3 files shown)


searx/engines/azure.py
```diff
@@ -0,0 +1,134 @@
...
+                    res.types.MainResult(
+                        url="https://portal.azure.com/#@/resource"
+                        + f"/subscriptions/{data['subscriptionId']}/resourceGroups/{data['name']}/overview",
+                        title=data["name"],
+                        content=f"Resource Group in Subscription: {data['subscriptionId']}",
+                    )
+                )
+        elif result["name"] == "resources":
+            for data in result["content"]["data"]:
+                print(data)
```
Bnyro: Please remove these `print` statements

searx/engines/azure.py
```diff
@@ -0,0 +1,134 @@
...
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    res = EngineResults()
+    json_data = resp.json()
+    # print(json_data)
+
+    for result in json_data["responses"]:
+        if result["name"] == "resourceGroups":
+            for data in result["content"]["data"]:
+                print(data)
```
Bnyro: Please remove these `print` statements

searx/settings.yml
```diff
@@ -496,6 +496,14 @@ engines:
...
     disabled: true
 
+  - name: azure
+    engine: azure
+    shortcut: az
+    tenant_id: 'your_tenant_id'
+    client_id: 'your_client_id'
+    client_secret: 'your_client_secret'
+    disabled: true
+    categories: [it, cloud]
```
Bnyro: nit: please add a newline here
```suggestion
    categories: [it, cloud]

```

searx/engines/azure.py
```diff
@@ -0,0 +1,134 @@
...
+    url = f"https://login.microsoftonline.com/{t_id}/oauth2/v2.0/token"
+    body = {
+        "client_id": c_id,
+        "client_secret": c_secret,
+        "grant_type": "client_credentials",
+        "scope": "https://management.azure.com/.default",
+    }
+
+    resp: SXNG_Response = http_post(url, body)
+    return resp.json()["access_token"]
```
Bnyro, line 101: Perhaps we should check whether the auth token is valid here, perhaps Azure returns an error response then?

So probably we can just check if the response code is 200 and throw an error otherwise to let the user know that the instance is misconfigured.

searx/engines/azure.py
```diff
@@ -0,0 +1,134 @@
...
+client_id = ""
+client_secret = ""
+
+
+batch_endpoint = "https://management.azure.com/batch?api-version=2020-06-01"
+
+about = {
+    "website": "https://www.portal.azure.com",
+    "wikidata_id": "Q725967",
+    "official_api_documentation": True,
```
Bnyro: You should add a link to the api documentation here instead of setting this to `True`, maybe https://learn.microsoft.com/en-us/rest/api/azure/ is the correct one?

searx/settings.yml
Bnyro: Please comment it out completely because it's not working by default as an API key is required.

gioleppe:
@Bnyro thanks for your comments, I've implemented all the changes you suggested 😄 
@return42 let me know if you have comments as well!

Bnyro (approved):

return42 (commented):
I squashed the commits into one and made a few [“editorial changes”](https://github.com/searxng/searxng/compare/dcceaa6d5f82318b9e636121b3a2435b1c50c2c0..75bba50afc5f9ebeee84febd3f251307221237f6), but nothing should have changed functionally.

Additional I added a commit on top of your branch, in this commit I added a cache for the token .. 

I don't know when an auth token expires, but I've set the cache expire to 10 minutes for now, which will make some login processes obsolete.

I don't have an Azure account, which is why all my changes are untested / please test them all / what do you think about the cache, do see any drawbacks?

searx/engines/azure.py
```diff
@@ -0,0 +1,190 @@
...
+    from searx.search.processors import OnlineParams
+
+engine_type = "online"
+categories = ["it", "cloud"]
+
+# Default values, should be overridden in settings.yml
+azure_tenant_id = ""
+azure_client_id = ""
+azure_client_secret = ""
+azure_batch_endpoint = "https://management.azure.com/batch?api-version=2020-06-01"
```
return42, line 47: To prevent naming conflicts, I have prefixed these settings with `azure_`.

gioleppe:
> I squashed the commits into one and made a few [“editorial changes”](https://github.com/searxng/searxng/compare/dcceaa6d5f82318b9e636121b3a2435b1c50c2c0..75bba50afc5f9ebeee84febd3f251307221237f6), but nothing should have changed functionally.
> 
> Additional I added a commit on top of your branch, in this commit I added a cache for the token ..
> 
> I don't know when an auth token expires, but I've set the cache expire to 10 minutes for now, which will make some login processes obsolete.
> 
> I don't have an Azure account, which is why all my changes are untested / please test them all / what do you think about the cache, do see any drawbacks?

Thanks for the changes @return42  noted for further reference :)
For what concerns the cache, as far as I know the lifetime of the access token should vary between 60 and 90 minutes, as described here https://docs.azure.cn/en-us/entra/identity-platform/access-tokens#token-lifetime so I'd say it's safe to set it at the lower end, maybe just a little bit less just in case. I'd opt around ~50 mins.
There is also a way to define some custom expiration values (albeit not that popular as far as I know) using the Azure CLI, so I'd say we could make it configurable from the user side. I'm pushing a commit in a few mins to address this, let me know if it makes sense to you

gioleppe:
@return42 done, I also changed the token expiration time to be a variable coming from settings.yml, 
please let me know if everything seems okay to you 😄

return42 (approved):
> please let me know if everything seems okay to you

perfect 🚀  .. I just squashed again the commits into one and updated the commit message.

// thanks for coding :-)

PR #5256, opened 2025-09-28
[mod] drop searxng_extra/standalone_searx.py

return42:
This is an old and grumpy hack / SearXNG is a Flask application with client/server structure, which can't be turned into a command line tool the way it was done here.

Maintaining this hack is becoming increasingly complex the more we try to remodel the core code, which is why we should now remove the hack from SearXNG.


inetol (approved):
I don't think anyone uses this (I hope)

PR #5280, opened 2025-10-04
[mod] brand - partial migration of settings to msgspec.Struct

return42:
The settings are currently an untyped key/value structure, whose types are dynamically built at runtime.  The construction process of this structure is *hand-crafted*.

In the long term, we want a static typing of this structure, based on a standard tool. The ``msgspec.Struct`` structures are suitable as a standard tool.

This patch makes a first step towards static typing and implements the "brand" section using ``msgspec.Struct`` structures.

BTW: searx/settings_defaults.py - ``git_url`` and ``git_branch`` had been removed in aee613d256, this is a leftover.

### DIFF
--- a/searx/brand.py
+++ b/searx/brand.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Implementations needed for a branding of SearXNG."""
+# pylint: disable=too-few-public-methods
+
+# Struct fields aren't discovered in Python 3.14
+# - https://github.com/searxng/searxng/issues/5284
+from __future__ import annotations
+
+__all__ = ["SettingsBrand"]
+
+import msgspec
+
+
+class BrandCustom(msgspec.Struct, kw_only=True, forbid_unknown_fields=True):
+    """Custom settings in the brand section."""
+
+    links: dict[str, str] = {}
+    """Custom entries in the footer of the WEB page: ``[title]: [link]``"""
+
+
+class SettingsBrand(msgspec.Struct, kw_only=True, forbid_unknown_fields=True):
+    """Options for configuring brand properties.
+
+    .. code:: yaml
+
+       brand:
+         issue_url: https://github.com/searxng/searxng/issues
+         docs_url: https://docs.searxng.org
+         public_instances: https://searx.space
+         wiki_url: https://github.com/searxng/searxng/wiki
+
+         custom:
+           links:
+             Uptime: https://uptime.searxng.org/history/example-org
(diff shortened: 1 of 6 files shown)


searx/settings_defaults.py
```diff
@@ -138,19 +141,32 @@ def __call__(self, value: t.Any) -> t.Any:
...
+        if isinstance(value, type) and issubclass(value, msgspec.Struct):
+            try:
+                # Type Validation at runtime:
+                # https://jcristharif.com/msgspec/structs.html#type-validation
+                cfg_dict = settings.get(key)
+                cfg_json = msgspec.json.encode(cfg_dict)
+                settings[key] = msgspec.json.decode(cfg_json, type=value)
+            except msgspec.ValidationError as e:
+                msg = str(e)
+                msg = msg.replace("`$.", "`" + ".".join([*path_list, key]) + ".")
```
Bnyro, line 159: That's not really self-explaining, maybe putting an example or some clarification about the msgspec error format here could help?

I.e. `Expected `str`, got `array` - at '$.new_issue_url'` gets converted to `Expected `str`, got `array` - at 'brand.new_issue_url'` by this code.

Bnyro (approved):
I really like the direction these changes are going into!

searx/settings_defaults.py
return42: > That's not really self-explaining,

Yeah, that is true .. [I added a comment to explain](https://github.com/searxng/searxng/compare/162748493037ba2d7565f225a1f9b7dce63777ec..ca49d37447afa227fb8dad9e229e5185bcec3c53)

Bnyro (approved):

searx/brand.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    issue_url: str = "https://github.com/searxng/searxng/issues"
+    """If you host your own issue tracker change this URL."""
+
+    docs_url: str = "https://docs.searxng.org"
+    """If you host your own documentation change this URL."""
+
+    public_instances: str = "https://searx.space"
+    """If you host your own https://searx.space change this URL."""
```
inetol, line 45: I previously had a `false` primitive to hide the link, now I'm using an empty string. I don't know if the "empty string" method is documented somewhere

return42: May we improve the wording in ["...then unset this URL."](https://docs.searxng.org/admin/settings/settings_brand.html#searx.brand.SettingsBrand.new_issue_url) to something like ".. then unset this URL by an empty string"?

PR #5288, opened 2025-10-09
Remove unsupported `--replace` flag from Docker run command

cubukcum:
## What does this PR do?

Removes the unsupported --replace flag from the Docker command in the documentation.

This flag is not recognized by the standard Docker CLI and causes the command to fail, and result:
`unknown flag: --replace
`

### DIFF
--- a/docs/admin/installation-docker.rst
+++ b/docs/admin/installation-docker.rst
@@ -100,7 +100,7 @@ Basic container instancing example:
    $ cd ./searxng/
 
    # Run the container
-   $ docker run --name searxng --replace -d \
+   $ docker run --name searxng -d \
        -p 8888:8080 \
        -v "./config/:/etc/searxng/" \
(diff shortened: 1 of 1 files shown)


inetol:
Damn, I thought `replace` was a thing in Docker.

https://docs.podman.io/en/v5.6.2/markdown/podman-run.1.html#replace

Bnyro (commented):
Perhaps we could use
```--rm        Automatically remove the container when it exits```
?

inetol:
> Perhaps we could use `--rm Automatically remove the container when it exits` ?

But the container is always deleted when the engine/host is stopped, which may not be desirable. It's fair enough just to show a basic manual instantiation example because "advanced method of running containers", at that point you should already know about *pull/rm/start* commands to fetch the latest image.

inetol (approved):

PR #5292, opened 2025-10-09
[feat] preferences hash: show applied settings in pref page when searching with 'search url of the currently saved preferences'

### LINKED ISSUE #5227: The preferences in `http://localhost:8888/preferences` are not updated after `http://localhost:8888/?preferences=[preferences_hash]&q=%s` has been visited in the web browser
### Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG
2025.9.14+23257bd

Installed manually using the official wiki instructions.  

### What happened?
The preferences in http://localhost:8888/preferencesare not updated after http://localhost:8888/?preferences=[preferences_hash]&q=%s has been visited in the web browser.

### Expected behavior
All settings in http://localhost:8888/ should reflect the setting after http://localhost:8888/?preferences=[preferences_hash]&q=%s has been visited in the web browser. For example:
* Preferences -> Engines

Bnyro:
Previously, when using a search url copied from the cookies tab, clicking at the settings icon at the top right would show the browser preferences and not the preferences that were set and used with the search url. Please see https://github.com/searxng/searxng/issues/5227 for more information.

To test:
- change some preferences
- copy the preferences search url in the settings' cookies tab
- reset the preferences or clear cookies
- paste the copied search url into the search bar to search for something
- press the settings icon
- you can now see/preview the actual settings that were used for the search
- by pressing 'save', you can keep these preferences

closes #5227


return42:
> and not the preferences that were set and used with the search url. 

This one? --> grafik

> paste the copied search url into the search bar to search for something

:thinking: this URL has `&q=%s` at the end .. the `%s` is a placeholder .. and I thought this URL is intended to add it as a search-engine to the WEB browser (for browsers that do not support [OpenSearch](https://github.com/searxng/searxng/issues/431)) 

> the "Preference" link is not changed from http://localhost:8888/preferences to http://localhost:8888/preferences?preferences=[preferences_hash] which is required to overview or modify the settings.
..
> - you can now see/preview the actual settings that were used for the search
> - by pressing 'save', you can keep these preferences

To transfer preferences, there's another URL ..

grafik

.. this can also be saved as a bookmark, making it possible to use the bookmarks as "search profiles." For example, I have two bookmarks for my research (in the languages DE or EN).

@davidhedlund, @Bnyro please excuse me if I am slow on the uptake, but for me, the "Search URL" is intended for browsers (only) .. and the "URL to restore" is intended for use cases where a persistence of the preferences (*save cookie*) is needed.

Each of these two URLs has a specific task. If we now use the search URL to transfer the preferences, we mix up their tasks, which can lead to confusion. Here is an example:

In the preferences I set the search language to DE and copy the **search-URL**, which I enter in my web browser as a search engine. I use this search engine to perform a search and then go back to the preferences to activate autocomplete with DuckDuckGo, which I save in the **cookie**.

  The original **search-URL** and the last **cookie** saved no longer match !!

We now have two pref URLS, one is `http://127.0.0.1:8888/preferences?preferences=eJx1WMuu7Ko...&preferences_preview_only=true` where we see the pref-settings from the **search-URL**:

grafik

and the other is `http://127.0.0.1:8888/preferences` endpoint without arguments .. which shows us the settings from the **cookie**:

grafik

Bnyro:
> @davidhedlund, @Bnyro please excuse me if I am slow on the uptake, but for me, the "Search URL" is intended for browsers (only) .. and the "URL to restore" is intended for use cases where a persistence of the preferences (save cookie) is needed.
>
I see your point. What this issue and PR are willing to address is the following:
- If you only have the search URL and search with it, then you would expect the settings to also show the preferences used by this search query by clicking on the settings symbol.
- We shouldn't expect users to copy the `search?preferences=...&q=%s` string, modify it to `preferences?preferences=...`, however that's currently the only way for users if they didn't also backup the other link and want to change the settings.

Alternative solution that comes to my mind
- display two settings icons: browser settings and search query (`preference_hash`) settings
- add a message to the preferences page that says something like `This is a preview of the settings used by your search URL. Press save to apply them to your browser.`

This would likely avoid the confusion :)

return42 (approved):
> then you would expect the settings to also show the preferences used by this search query by clicking on the settings symbol.

OK, I understand, its a valid argument .. from my side we can merge this PR as-is.

Bnyro:
> add a message to the preferences page that says something like This is a preview of the settings used by your search URL. Press save to apply them to your browser.
>
I've added the warning message I mentioned, it should now be self-explaining. Preview:

Screenshot from 2025-10-10 17-57-30


Good to merge from my side.

Bnyro:
@return42 Do you want to have a final look/review or are you okay with the added dialog?

Issue #5304, opened 2025-10-11
!weather crashes with internal server error


cy8aer:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

2025.10.11+ced08e12a (docker-compose)

**How did you install SearXNG?**

Official docker compose

**What happened?**

Searching any weather app crashes searxng with an internal error

**How To Reproduce**

`!weather london'

**Expected behavior**

The weather entries occur

**Screenshots & Logs**

Internal server error.

```
searxng    | /usr/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=15) is multi-threaded, use of fork() may lead to deadlocks in the child.
searxng    |   self.pid = os.fork()
searxng    | 2025-10-11 19:20:44,799 ERROR:searx.webapp: Exception on /search [POST]
searxng    | Traceback (most recent call last):
searxng    |   File "/usr/local/searxng/.venv/lib/python3.12/site-packages/flask/app.py", line 1511, in wsgi_app
searxng    |     response = self.full_dispatch_request()
searxng    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng    |   File "/usr/local/searxng/.venv/lib/python3.12/site-packages/flask/app.py", line 919, in full_dispatch_request
searxng    |     rv = self.handle_user_exception(e)
searxng    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng    |   File "/usr/local/searxng/.venv/lib/python3.12/site-packages/flask/app.py", line 917, in full_dispatch_request
searxng    |     rv = self.dispatch_request()
searxng    |          ^^^^^^^^^^^^^^^^^^^^^^^
searxng    |   File "/usr/local/searxng/.venv/lib/python3.12/site-packages/flask/app.py", line 902, in dispatch_request
searxng    |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
searxng    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
searxng    |   File "/usr/local/searxng/searx/webapp.py", line 751, in search
searxng    |     return render(
searxng    |            ^^^^^^^
searxng    |   File "/usr/local/searxng/searx/webapp.py", line 448, in render
[...]

PR #5309, opened 2025-10-12
[fix] !weather crashes - cls.TURN 'member_descriptor' isn't a float

return42:
The class method ``Compass.point`` is converted into an instance method to circumvent the problem described in [1] (without understanding the cause).

- [1] https://github.com/searxng/searxng/issues/5304#issuecomment-3394140820

----

This PR is a HOTFIX; it would be better if we had an explanation for the problem described in [1].

### DIFF
--- a/searx/weather.py
+++ b/searx/weather.py
@@ -506,15 +506,14 @@ class Compass(msgspec.Struct):
             return self.val
         raise ValueError(f"unknown unit: {unit}")
 
-    @classmethod
-    def point(cls, azimuth: float | int) -> CompassPoint:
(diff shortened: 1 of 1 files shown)


inetol (approved):
Maybe it's an issue with msgspec itself?

Bnyro (approved):

return42:
> Maybe it's an issue with msgspec itself?

@inetol my fail, I haven't RTF manual :-o .. found and fixed the issue, see:

- https://github.com/searxng/searxng/pull/5398

PR #5310, opened 2025-10-12
[mod] container: move `base` to own repository

inetol:
The base images will be now built in [another repository](https://github.com/searxng/base).


return42 (approved):
LGTM

PR #5315, opened 2025-10-12
Naming consistency and no spaces for mullvadleta Google and Brave

geckolinux:
It was confusing to name the Google version as just "mullvadleta", and also confusing for the Brave version to be called "mullvadleta brave" with a space, because there is a separate engine also named "brave". For example:

image

### DIFF
--- a/searx/data/engine_descriptions.json
+++ b/searx/data/engine_descriptions.json
@@ -251,8 +251,8 @@
    "mojeek:ar",
    "ref"
   ],
-  "mullvadleta":"مولفاد هو مزود شبكة خاصة افتراضية ومقره السويد. إنه يلتزم بسياسة صارمة لعدم الاحتفاظ بالسجلات وسياسة خصوصية شاملة للغاية. لا تجمع مولفاد أي معلومات منك عند التسجيل. بدلاً من ذلك ، يتم إنشاء رقم عشوائي واستخدامه كمعرف خاص بك. تدعم اوبن في بي إن و وايرجارد. ويدعم أيضًا قفزات خادم متعددة مع ميزة التجسير. تقدم مولفاد خدمة جيدة للغاية مع تركيز قوي على الخصوصية.",
-  "mullvadleta brave":[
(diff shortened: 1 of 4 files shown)


return42:
I would prefer a dotted notation... as we used for the names of the `brave` engines:

https://github.com/searxng/searxng/blob/8baefcc21e289a7ce932d8e7bad4c1fed377c7b3/searx/settings.yml#L2637-L2656

----

> It was confusing to name the Google version as just "mullvadleta", 

IMO this is a design issue .. not a naming issue.  --> We should better add separators to the names of the engines in the UI.

geckolinux:
Hi there @return42 , thanks for looking at this.

> I would prefer a dotted notation... as we used for the names of the `brave` engines:

That sounds like a good idea too.
So In that case, it would be good to have a dotted extension for both providers that Mullvad Leta can use (i.e. `mullvadleta.google` and `mullvadleta.brave`) since it doesn't have it's own "mullvadleta" search index.
Would you like me to try to rework my PR in that way? I'm not sure why the automated checks are finding this error:

```
      File "/home/runner/work/searxng/searxng/searx/webapp.py", line 1392, in init
        searx.search.initialize(enable_checker=True, check_network=True, enable_metrics=metrics)
        ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/searxng/searxng/searx/search/__init__.py", line 42, in initialize
        initialize_network(settings_engines, settings['outgoing'])
        ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/searxng/searxng/searx/network/network.py", line 412, in initialize
        NETWORKS[engine_name] = NETWORKS[network]
                                ~~~~~~~~^^^^^^^^^
    KeyError: 'mullvadleta'
```

> We should better add separators to the names of the engines in the UI.

Good point, maybe something as simple as just comma separators would do the trick.

return42:
For future naming, we can agree on dotted notation.

However, we should avoid changing the usual names if possible, as there are certainly users who use these names in their search syntax... and these users would then have to get used to the new names... even the cookies for which preferences have been set in the active engines would be defective... 

What we need is a PR that separates the names in the UI.

Issue #5198, opened 2025-09-10
The concept of rendering (less/css) the results depending on two variables does not work


return42:
The rendering of a result is currently determined by two variables (CSS classes), one is the template (the result type) and the other is a category ..

by template (result type):

https://github.com/searxng/searxng/blob/a9b088d832021271c872a29675abd2887dbdcf65/client/simple/src/less/style.less#L312-L317

by category:

https://github.com/searxng/searxng/blob/a9b088d832021271c872a29675abd2887dbdcf65/client/simple/src/less/style.less#L372-L381

For me, the question is: what constitutes the category? 

The code currently arbitrarily uses the engine's first category .. see `category-{{ result['category']}}` in the template ..

https://github.com/searxng/searxng/blob/a9b088d832021271c872a29675abd2887dbdcf65/searx/templates/simple/macros.html#L20-L22

and `res.category` in the processing of the results (`engine.categories[0]` is the first):

https://github.com/searxng/searxng/blob/a9b088d832021271c872a29675abd2887dbdcf65/searx/results.py#L216-L220

This leads to considerable confusion. If, for example, an engine is configured in a different category during customization, its results are suddenly rendered differently.

Here's a (safely constructed) example that clearly shows the impact.

First category is `videos`:

```yaml
  - name: google videos
    categories: ['videos', 'web']
```

.. in the hit list there is a *border* around each result:

<img width="706" height="652" alt="Image" src="https://github.com/user-attachments/assets/eceef11a-c1ba-4c34-995b-2d2010d141a4" />

And now with `web` as first category .. 

```yaml
  - name: google videos
    categories: ['web', 'videos']
```

.. in the result list we do no longer see the *border*:

<img width="706" height="652" alt="Image" src="https://github.com/user-attachments/assets/7bed6111-bf32-4152-9be3-82223b7f06ac" />

----

In my opinion, no category can be determined: engines have multiple categories, and even in the `categories_as_tab`, multiple categories can be selected.

To make matters worse, the category names are explicitly stated in LESS. However, the concept of categories is fundamentally freely configurable; the operator of an instance can freely choose which category names they want to have as tabs and which engine they want to assign to which categories.

----

Long story short: category names are freely configurable, and there can't be a CSS for arbitrary class names.

The categories in CSS must be abandoned and the stylesheet should be determined solely by the result class (the template).

PR #5320, opened 2025-10-13
[mod] typification of SearXNG: add new result type `File`

return42:
This PR adds a new result type: `File`

    Python class: searx/result_types/file.py
    Jinja template: searx/templates/simple/result_templates/file.html
    CSS (less) client/simple/src/less/result_types/file.less

Class `File` (singular) replaces template `files.html` (plural).  The renaming
was carried out because there is only one file (singular) in a result. Not to be
confused with the category 'files' where in multiple results can exist.

As mentioned in issue [1], the class `.category-files` was removed from the CSS
and the stylesheet was adopted in `result_types/file.less` (there based on the
templates and no longer based on the category).

[1] https://github.com/searxng/searxng/issues/5198


Bnyro:
Perhaps it'd make sense to split the changes to mediawiki and reroll into separate commits for better readability?

searx/engines/recoll.py
```diff
@@ -13,131 +13,144 @@
...
+        thumbnail = embedded = ""
+        if mtype in ["audio", "video"]:
+            embedded_url = '<{ttype} controls height="166px" ' + 'src="{url}" type="{mtype}"></{ttype}>'
+            embedded = embedded_url.format(ttype=mtype, url=quote(url.encode("utf8"), "/:"), mtype=result["mtype"])
+        if mtype in ["image"] and subtype in ["bmp", "gif", "jpeg", "png"]:
+            thumbnail = url
+
+        res.add(
+            res.types.File(
+                title=_str("label"),
```
Bnyro: To be honest I'm not sure if that really makes things more readable, in my opinion it makes more sense to just use `result.get("...", "")` directly here as it's not much additional code.

searx/result_types/file.py
```diff
@@ -0,0 +1,86 @@
...
+
+    author: str = ""
+    """Author of the file."""
+
+    embedded: str = ""
+    """URL of an embedded media type (audio or video) / is collapsible."""
+
+    mtype: str = ""
+    """Used for displaying :py:obj:`File.embedded`, it value automatically
+    populated from the base type of :py:obj:`File.mimetype`, and can be
```
Bnyro:
```suggestion
    """Used for displaying :py:obj:`File.embedded`. Its value is automatically
    populated from the base type of :py:obj:`File.mimetype`, and can be
```

searx/result_types/file.py
```diff
@@ -0,0 +1,86 @@
...
+    explicitly set to enforce e.g. ``audio`` or ``video`` when mimetype is
+    something like "application/ogg" but its know the content is for example a
+    video."""
+
+    subtype: str = ""
+    """Used for displaying :py:obj:`File.embedded`, it value automatically
+    populated from the subtype of :py:obj:`File.mimetype`, and can be explicitly
```
Bnyro: same as above

searx/result_types/file.py
```diff
@@ -0,0 +1,86 @@
...
+    set to enforce a subtype for the :py:obj:`File.embedded` element.
+    """
+
+    def __post_init__(self):
+        super().__post_init__()
+
```
Bnyro, line 74: Is the `mimetype` attribute mandatory or optional?

I think we could also try to guess it based on the file name using [`mimetypes.guess_mimetype`](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type) so that it also works if it's not possible to get the mimetype from the engine's API.

client/simple/src/less/result_types/file.less
Bnyro, line 1: Good idea to put the result types into separate `less` files, the `style.less` already contains way too much code imo

searx/result_types/file.py
return42: > Is the `mimetype` attribute mandatory or optional?

optional

> I think we could also try to guess it based on the file name 

Good point! .. I implemented `guess_mimetype` and fixed the issues you reported above --> [diff](https://github.com/searxng/searxng/compare/dff395be6327d6ce1b8a6fae03904e0ae9d9523d..0b395f78a69ef7083a1f2d85b15c100b7b6ede8a)

Bnyro: Perhaps we could also try to guess it from `self.embedded` if `self.filename` doesn't exist?

searx/engines/wikicommons.py
```diff
@@ -1,102 +1,206 @@
...
+                duration = datetime.timedelta(seconds=int(seconds))
+            except OverflowError:
+                pass
+
+        if wc_search_type == "file":
+            res.add(
+                res.types.File(
+                    title=title,
+                    url=url,
+                    content=content,
```
Bnyro, line 156:
```suggestion
                    content=content,
                    filename=title,
```
I don't know if that makes sense because it'd be displayed twice then, but perhaps it's worth setting the `filename` too.

Bnyro:
Just two nitpicks, apart from that lgtm :+1: 

Thanks for looking into the `mimetypes` thing!

searx/engines/wikicommons.py
return42: The `title` is not suitable because the file extension has already been removed.  The filename in `media_url` is more suitable --> [diff](https://github.com/searxng/searxng/compare/46e96496c2ef7a83c77e2f0453b1e66aecc85e58..f57f91f31e8c90baefc9831c5504ae3763d8a9b6)

searx/result_types/file.py
return42: Sure --> [diff](https://github.com/searxng/searxng/compare/46e96496c2ef7a83c77e2f0453b1e66aecc85e58..f57f91f31e8c90baefc9831c5504ae3763d8a9b6)

return42:
> Just two nitpicks,

.. fixed ..

> apart from that lgtm 👍

Thanks for review and additions .. I'm going to merge ..

PR #5325, opened 2025-10-14
[chore] Mention the black theme style in settings.yml

benpiano800:
## What does this PR do?

This PR mentions the "Black" style for the Simple theme in the comment in settings.yml that lists the other styles.

The "Black" style was not previously mentioned, so instance maintainers may not know about it.

Replace line 141 in your settings.yml with the changed line.


return42 (approved):
FYI `theme_args.simple_style` also documented in the [ui settings](https://docs.searxng.org/admin/settings/settings_ui.html).

PR #5337, opened 2025-10-17
[fix] engine mullvadleta - ignore HTTP 403 & 429 response

### LINKED ISSUE #5328: Mullvad Leta: Cooldown on blocked requests is too long
https://github.com/searxng/searxng/commit/576d30ffcdf23c2bbff6b1c624d1a7ad13e29abd

searxng-docker

**What happened?**
The engine is almost always blocked on a publicly accessible SearXNG instance.

**How To Reproduce**
Use advanced search (e.g. site:), search a phone number, etc.

**Expected behavior**
The logic should handle the blocked requests (403, 429), but not put the engine on a 1 day cooldown (too long). From what I can see, Mullvad only blocks the search request and doesn't prevent you from doing more searches.

**Screenshots & Logs**
N/A

**Additional context**
The same restrictions apply to the Brave/Google engines on Mullvad Leta. It doesn't matter if you're using Mullvad's VPN and a proper browser, you'll still get blocked for specific searches with a 403 or 429 HTTP status code.

return42:
It doesn't matter if you're using Mullvad's VPN and a proper browser, you'll still get blocked for specific searches [1] with a 403 or 429 HTTP status code. Mullvad only blocks the search request and doesn't prevent you from doing more searches.

The logic should handle the blocked requests (403, 429), but not put the engine on a cooldown.

[1] https://leta.mullvad.net/search?q=site%3Afoo+bar&engine=brave

Closes: https://github.com/searxng/searxng/issues/5328

### DIFF
--- a/searx/engines/mullvad_leta.py
+++ b/searx/engines/mullvad_leta.py
@@ -43,9 +43,12 @@ import babel
 from httpx import Response
 from lxml import html
 from searx.enginelib.traits import EngineTraits
+from searx.extended_types import SXNG_Response
+
 from searx.locales import get_official_locales, language_tag, region_tag
 from searx.utils import eval_xpath_list
 from searx.result_types import EngineResults, MainResult
+from searx.network import raise_for_httperror
 
 search_url = "https://leta.mullvad.net"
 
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Tested, can confirm this prevents our internal rate limit from getting triggered so that Mullvad continues to work :+1:

Issue #5299, opened 2025-10-10
all search results for forums.servethehome.com point to main forum page


foolab:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG** docker 2025.10.8+898fbe7bc

**How did you install SearXNG?** via docker compose

**What happened?**
Whenever I search for something that has results from servethehome.com forums, the links are broken and point to the main forum page instead of the relevant thread.
 
**How To Reproduce**
search for something that will lead to results from servethehome.com forums. For example:
"servethehome forum ryzen" or "mj11-ec1"

**Expected behavior**
Link to a proper forum thread.

**Screenshots & Logs**

foolab:
This is a screenshot showing the issue. Hovering over the link with the mouse shows the actual URL in the statusbar. It points to the main forums page and clicking it will lead to the main page.

<img width="1280" height="736" alt="Image" src="https://github.com/user-attachments/assets/69a0ffd5-1768-4b55-b053-53c0e38615ff" />

inetol:
Looks like URI query is getting stripped out:

<img width="710" height="300" alt="Image" src="https://github.com/user-attachments/assets/8262bb36-25c4-4f32-9c98-b51674a9d197" />

(image shows DuckDuckGo search "servethehome forum ryzen")

inetol:
@foolab Can you try disabling `Tracker URL remover` and see if the issue is resolved?

<img width="1169" height="307" alt="Image" src="https://github.com/user-attachments/assets/a1d64881-5091-4bc9-8949-cf8987512cc4" />

foolab:
Disabling tracker URL remover did not help unfortunately

return42:
> Disabling tracker URL remover did not help unfortunately

Sure?  When I run a `make run` instance I can see ..

```
DEBUG   searx   : filter_urls: modify field url URL https://forums.servethehome.com/index.php?threads%2Fnas-build-xeon-vs-epyc-3000-vs-ryzen.31141%2F= -> https://forums.servethehome.com/index.php
DEBUG   searx   : filter_urls: modify field url URL https://forums.servethehome.com/index.php?threads/nas-build-xeon-vs-epyc-3000-vs-ryzen.31141/page-2= -> https://forums.servethehome.com/index.php
[...]

PR #5339, opened 2025-10-18
[fix] TrackerPatternsDB.clean_url: don't delete query argument from new_url

return42:
The query argument for URLs like:

- 'http://example.org?q='       --> query_str is 'q='
- 'http://example.org?/foo/bar' --> query_str is 'foo/bar'

is a *simple string* and not a key/value dict.  This string may only be removed from the URL if one of the patterns matches.

BTW get_pretty_url(): keep such a *simple string* in the path element.

Closes: 
- https://github.com/searxng/searxng/issues/5299

### DIFF
--- a/searx/data/tracker_patterns.py
+++ b/searx/data/tracker_patterns.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Simple implementation to store TrackerPatterns data in a SQL database."""
+# pylint: disable=too-many-branches
 
 import typing as t
 
@@ -119,6 +120,12 @@ class TrackerPatternsDB:
 
         for rule in self.rules():
 
+            query_str: str = parsed_new_url.query
+            if not query_str:
+                # There are no more query arguments in the parsed_new_url on
+                # which rules can be applied, stop iterating over the rules.
+                break
+
             if not re.match(rule[self.Fields.url_regexp], new_url):
                 # no match / ignore pattern
                 continue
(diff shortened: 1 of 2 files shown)


searx/webapp.py
```diff
@@ -356,6 +356,12 @@ def get_pretty_url(parsed_url: urllib.parse.ParseResult):
     path = parsed_url.path
     path = path[:-1] if len(path) > 0 and path[-1] == '/' else path
     path = unquote(path.replace("/", " › "))
+
+    # Keep the query argument for URLs like:
+    # - 'http://example.org?/foo/bar' --> parsed_url.query is 'foo/bar'
+    query_args: list[tuple[str, str]] = list(urllib.parse.parse_qsl(parsed_url.query))
+    if not query_args and parsed_url.query:
+        path += " › .." + parsed_url.query[-24:]
```
Bnyro: We probably only need to display the `".."` if the query length is longer than 24?

searx/data/tracker_patterns.py
```diff
@@ -136,18 +137,42 @@ def clean_url(self, url: str) -> bool | str:
                 #    overlapping urlPattern like ".*"
                 continue
 
-            # remove tracker arguments from the url-query part
+            query_str: str = parsed_new_url.query
```
Bnyro: Unrelated to this PR: Perhaps we could also check if `query_str == ""` in the beginning of the loop? Then we don't need to iterate over all the rules and can just break out to save time / performance, because the query is already empty.

Bnyro (commented):
I can confirm this works as expected, good catch :+1:

searx/data/tracker_patterns.py
return42: That's a really good suggestion. I've now structured the loop a little differently and placed the termination condition at the beginning of the iteration --> [diff](https://github.com/searxng/searxng/compare/16967596f1c41a3804a419e60f7bb419fb818ada..780ffff1d56e8ff719026f05be5158367f751ccb)

PR #5366, opened 2025-10-24
[feat] plugins: Add "define <term>" Plugin to Show Definitions in Searches

macarooni-man:
## What does this PR do?

This PR adds a new plugin: **`define`**, which provides inline dictionary definitions to search results, powered by Wiktionary. I use this feature a lot with search engines, and added this feature to my own instance for that reason. I also thought it would be beneficial to share with the community!

image

<br><br>

When a user searches for a query beginning with

```
define <term>
```

the plugin fetches definitions via the official Wiktionary REST API and renders them as a answer card using a modified `Translations` type.

The plugin:

* Displays definitions grouped by part of speech (noun, verb, adjective, etc.)
* Includes a few example usages from Wiktionary
* Rewrites internal `/wiki/...` links to equivalent local searches (e.g. `/search?q=define+term`)
* Returns results only on the first page of a search
* Adds a button in the botton right corner to view the Wiktionary entry directly
* Falls back gracefully with a short "No definitions were found." message

Implementation notes:

* Fully type-annotated and follows the required code style
* At least from what I can tell, safe HTML rewriting and de-duplication
* Uses SearXNG’s native HTTP client (`searx.network.get`) for requests
* An extensible provider framework to add new sources in the future with `DefinitionProvider`
* Adds a Jinja template for rendering embedded HTML from the Wiktionary REST API `answer/define.html`

---

This plugin gives users quick, inline dictionary definitions without opening another tab or engine result. This is similar to what major search engines provide.
It enhances the SearXNG user experience by:

* Adding native dictionary-style answers directly at the top of search results when triggered
* Leveraging Wiktionary, a free and open data source
* Integrating seamlessly with SearXNG’s existing answerer and translations infrastructure
* Recursive searches with hyperlinks to browse multiple definitions easily
* Providing multilingual results when available (unfortunately it's not in my wheelhouse and still needs work)

This makes "define" queries faster, privacy-friendly, and self-contained in the UI.

---

1. Make sure the plugin is enabled in your SearXNG configuration (`settings.yml`):
  > Note: I've enabled it by default in this PR

   ```yaml
   plugins:
     - searx.plugins.define.SXNGPlugin
   ```

2. Rebuild or reload your dev environment:

   ```bash
   make run
   ```

3. Open SearXNG in your browser and verify that "Define" is enabled in Special Queries > Plugins:

image

4. From the home page, try queries such as:

   * `define effervescence`
   * `define network`
   * `define chicken tender`

5. You should see a compact definition card rendered at the top of the results:

image

<br><br>

To run automated tests:

```bash
make test
```

Expected:
  - `pyright`: no errors
  - `pylint`: 10.00/10
  - `pytest`: all tests pass
> Note that I've tested this in GitHub Codespaces, and on my personal server from a manual installation in Ubuntu Server (minimal) 24.02
<details>
  <summary>view my results</summary>

   ```
   (.venv) @macarooni-man ➜ /workspaces/searxng-define (master) $ make test
   TEST      [yamllint] .github/ISSUE_TEMPLATE/config.yml
   TEST      [black] --exclude
   All done! ✨ 🍰 ✨
   366 files would be left unchanged.
   TEST      [basedpyright] static type check of local modified files
   0 errors, 0 warnings, 0 notes
   TEST      [pylint] ./searx/engines

   --------------------------------------------------------------------
   Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

   TEST      [pylint] ./searx ./searxng_extra ./tests

   --------------------------------------------------------------------
   Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

   TEST      tests/unit
   ...................................................................................................................................................................................................................................................................................................
   Ran 291 tests in 50.368s

   OK
   TEST      robot
   INSTALL   gecko.driver
   INSTALL   geckodriver already installed
   2025-10-24 04:16:49,725 WARNING:searx.botdetection.config: missing config file: /workspaces/searxng-define/tests/robot/limiter.toml

   Running 7 tests
   /workspaces/searxng-define/local/py3/bin/geckodriver
   /usr/bin/firefox
   TEST      [reST markup] README.rst
   TEST      test.shell OK
   TEST      [shfmt] ./manage ./container ./utils
   INSTALL   [pkg.go.dev] ./go.mod: developer and CI tools
   go: downloading mvdan.cc/sh/v3 v3.12.0
   go: downloading github.com/google/renameio/v2 v2.0.0
   go: downloading github.com/rogpeppe/go-internal v1.14.1
   go: downloading golang.org/x/term v0.32.0
   go: downloading mvdan.cc/editorconfig v0.3.0
   go: downloading golang.org/x/sys v0.33.0
   go: downloading github.com/go-quicktest/qt v1.101.0
   go: downloading github.com/google/go-cmp v0.7.0
   go: downloading golang.org/x/tools v0.31.0
   go: downloading github.com/kr/pretty v0.3.1
   go: downloading github.com/kr/text v0.2.0
   ```

</details>

---

## Author's checklist

* [x] Fully typed, tests are clean
* [x] `pylint` score: 10.00/10
* [x] All unit tests pass
* [x] Verified rendering of definitions and examples
* [x] Confirmed proper escaping and link rewriting

Future ideas:

* Add support for multiple dictionary providers
* Add support for more locales
* Add local result caching

I'm open to any feedback, and I also can't wait to see how else this project evolves. I've been really happy with this tool in my network!

### DIFF
--- a/searx/plugins/define.py
+++ b/searx/plugins/define.py
@@ -47,7 +47,7 @@ if typing.TYPE_CHECKING:
 
 # ---------------------------------------  Definition Provider Functionality  ------------------------------------------
 
-
+# Internal dataclass for provider results storage
 @dataclass(eq=True, frozen=False)
 class Definition:
     """
@@ -324,7 +324,7 @@ class DefineHandler:
 
             items.append(
                 Translations.Item(
-                    text=word,  # now just "liquid"
+                    text=word,
                     transliteration=f"({pos})" if pos != "unknown" else "",
                     definitions=definitions_text,
                     examples=examples,
--- a/searx/plugins/define.py
+++ b/searx/plugins/define.py
@@ -124,7 +124,7 @@ class Definition:
 class DefinitionProvider:
     """
     Root class for handling web IO and exposing a common interface.
-    Subclasses should implement `.definitions(word, lang="en") -> list[Definition]`.
+    Subclasses should implement `.lookup(word, lang="en") -> list[Definition]`.
     """
 
     # Override this
--- a/searx/plugins/define.py
+++ b/searx/plugins/define.py
@@ -1,0 +1,370 @@
+from searx.result_types import Answer, Translations
+from searx.answerers import Answerer, AnswererInfo
+from searx.result_types.answer import BaseAnswer
+from searx.extended_types import SXNG_Response
+from flask_babel import gettext as _
+from flask_babel import get_locale
+from searx.network import get
+
+from searx.result_types import EngineResults
+from . import Plugin, PluginInfo
+
+
+from dataclasses import dataclass, field
+from typing import List, Optional
+from html import unescape
+from urllib import parse
+import typing
+import re
+
+if typing.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Request
+    from searx.search import SearchWithPlugins
+    from . import PluginCfg
+
+
+
+# ---------------------------------------  Definition Provider Functionality  ------------------------------------------
+
+@dataclass(eq=True, frozen=False)
+class Definition:
+    """
+    Root data model for a single definition entry
+    """
+
+    word:           str
+    part_of_speech: Optional[str]
+    definition:     str
+    examples:       List[str] = field(default_factory=list)
+    lang:           str = "unknown"
+    provider:       str = "unknown"
+    source_url:     Optional[str] = None
+
+
+    # Part-of-speech normalization
+    pos_normalized = {
+        "noun":         "noun",
+        "verb":         "verb",
+        "adjective":    "adjective",
+        "adverb":       "adverb",
+        "pronoun":      "pronoun",
+        "preposition":  "preposition",
+        "conjunction":  "conjunction",
+        "interjection": "interjection",
+        "proper noun":  "proper noun",
+        "proper_noun":  "proper noun",
+        "determiner":   "determiner",
+        "article":      "article",
+    }
+
+    # Normalize fields & dedupe examples when created
+    def __post_init__(self):
+        self.word       = self._clean_text(self.word)
+        self.definition = self._clean_text(self.definition)
+        self.examples   = self._dedupe(self.examples)
+
+        pos = (self.part_of_speech or "").strip().lower()
+        self.part_of_speech = self.pos_normalized.get(pos, pos or "unknown")
+
+    def _clean_text(self, s: str) -> str:
+        s = unescape(s or "")
+        return " ".join(s.split())
+
+    def _dedupe(self, seq: List[str]) -> List[str]:
+        seen, out = set(), []
+        for x in seq:
+            k = self._clean_text(x)
+            if k and k not in seen:
+                seen.add(k)
+                out.append(k)
+        return out
+
+    @property
+    def normalized_pos(self) -> str:
+        return self.part_of_speech or "unknown"
+
+    @property
+    def title(self) -> str:
+        return f"{self.word} ({self.normalized_pos})" if self.normalized_pos != "unknown" else self.word
+
+    @property
+    def dedupe_key(self) -> tuple:
+        return (self.normalized_pos, self.definition)
+
+    def short_example(self, limit: int = 140) -> Optional[str]:
+        if not self.examples:
+            return None
+
+        ex = self.examples[0]
+        return (ex if len(ex) <= limit else ex[: max(0, limit - 1)] + "…")
+
+
+class DefinitionProvider():
+    """
+    Root class for handling web IO and exposing a common interface.
+    Subclasses should implement `.definitions(word, lang="en") -> List[Definition]`.
+    """
+
+    # Override this
+    name: str
+
+
+    def _get(self, url: str) -> SXNG_Response:
+        return get(url)
+
+    def lookup(self, term: str, lang: str = "en") -> List[Definition]:
+        """
+        Return a list of Definition objects for `word`
+        Must be implemented by subclasses
+        """
+        raise NotImplementedError
+
+
+class WiktionaryProvider(DefinitionProvider):
+    """
+    Wiktionary implementation using the official REST API:
+    https://en.wiktionary.org/api/rest_v1/
+    """
+
+    name:          str = 'wiktionary'
+    api_template:  str = "https://en.wiktionary.org/api/rest_v1/page/definition/{title}"
+    page_template: str = "https://en.wiktionary.org/wiki/{title}"
+
+    # Matches href="/wiki/term#POS"
+    _wiki_href_re = re.compile(r'href="(?P<href>/wiki/[^"]+)"')
+
+    def _rewrite_wiki_links(self, html: str) -> str:
+        """
+        Rewrites <a href="/wiki/term[#Section][:Index]"> -> /search?q=define+term
+        - strips #Part-of-speech fragment
+        - strips namespace/index after first colon
+        """
+
+        if not html:
+            return html
+
+        keyword = DefineHandler.keyword
+
+        def _normalize(href: str) -> str:
+            path = href.split('#', 1)[0]
+            if path.startswith('/wiki/'):
+                path = path[len('/wiki/'):]
+            path = path.split(':', 1)[0]
+
+            term = parse.unquote(path).replace('_', ' ').strip()
+            return term
+
+        def _repl(m: re.Match) -> str:
+            term = _normalize(m.group('href'))
+            if not term:
+                return 'href="#"'
+            q = parse.quote_plus(f'{keyword} {term}').lower()
+            return f'href="/search?q={q}"'
+
+        return self._wiki_href_re.sub(_repl, html)
+
+
+    def lookup(self, term: str, lang: str = "en") -> List[Definition]:
+        url  = self.api_template.format(title=term)
+        resp = None
+
+        try:
+            resp = self._get(url)
+            success = resp.status_code == 200
+        except:
+            success = False
+
+        if not (success and resp):
+            return []
+
+        payload = resp.json()
+        entries = payload.get(lang, [])
+        results: List[Definition] = []
+
+        for entry in entries:
+            pos = entry.get("partOfSpeech")
+            for d in entry.get("definitions", []):
+                def_html = self._rewrite_wiki_links(d.get("definition", ""))
+                ex_list  = [self._rewrite_wiki_links(x) for x in d.get("examples", [])]
+                results.append(
+
+                    Definition(
+                        word           = term,
+                        part_of_speech = pos,
+                        definition     = def_html,
+                        examples       = ex_list,
+                        lang           = lang,
+                        provider       = "wiktionary",
+                        source_url     = self.page_template.format(title=term),
+                    )
+
+                )
+
+        return results
+
+
+
+# ----------------------------------------------  SearXNG Integration  -------------------------------------------------
+
+class DefineHandler():
+    """
+    Links a definition provider and provides a public API to the answerer
+    """
+
+    keyword:  str = "define"
+    provider: DefinitionProvider = WiktionaryProvider()
+
+    default_answer: Answer = Answer(answer=_("No definitions were found."))
+    max_definitions:   int = 5
+    max_examples:      int = 3
+
+
+    def _get_locale(self) -> str:
+        """
+        Prefer the user's selected search language if available
+        otherwise fall back to the UI locale, finally 'en'
+        Returns a ISO code like 'en', 'fr', 'de'
+        """
+
+        loc = get_locale()
+        if loc:
+            return str(loc).split('_')[0].lower()
+
+        return 'en'
+
+
+    # Expects a full query, including 'self.keyword'
+    def _extract_term(self, query: str) -> str | None:
+        extracted = None
+        term_key  = self.keyword + ' '
+
+        if query.startswith(term_key):
+            extracted = query[len(term_key):].strip()
+
+        return extracted
+
+
+    def _make_definitions_answer(self, definitions: list[Translations.Item], url: str | None) -> Translations:
+        if not self.provider.name:
+            raise NotImplementedError('no real provider was specified')
+
+        return Translations(
+            translations = definitions,
+            url          = url,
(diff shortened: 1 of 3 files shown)


searx/plugins/define.py
```diff
@@ -0,0 +1,395 @@
...
+    from searx.search import SearchWithPlugins
+    from . import PluginCfg
+
+
+# ---------------------------------------  Definition Provider Functionality  ------------------------------------------
+
+
+@dataclass(eq=True, frozen=False)
+class Definition:
+    """
```
return42, line 54: Unfortunately, the concept of [result types](https://docs.searxng.org/dev/result_types/index.html) has not yet been fully implemented and is therefore somewhat difficult to recognize --> In the model of result types, all *answers* are of the type [BaseAnswer](https://docs.searxng.org/dev/result_types/answer.html#searx.result_types.answer.BaseAnswer).

In terms of data structure, this class here is very similar to the type `Translations` ..

https://github.com/searxng/searxng/blob/edfa71cdea2cfd1660bdb4285842e4a9f19ca2ec/searx/result_types/answer.py#L98-L99

which has a list of *examples* ..

https://github.com/searxng/searxng/blob/edfa71cdea2cfd1660bdb4285842e4a9f19ca2ec/searx/result_types/answer.py#L124-L125

macarooni-man: Hi @return42, thanks for taking a peek!

I wanted to clarify as it may not have been clear through my implementation & PR description.

The `Definition` class here isn’t an answer type, it's only an internal data model used to normalize and store entries fetched from a provider (currently Wiktionary). The intended purpose is API abstraction to let new providers easily plug in later by returning the same structure.

The `BaseAnswer` emitted by the plugin is a standard `Translations` card, and the `Definition` is converted here in `DefineHandler` with the custom template:

https://github.com/searxng/searxng/blob/77967b2d8b8d6c21b93b0ac59e2a5574c66b792e/searx/plugins/define.py#L272-L335

Bnyro (commented):
Personally I feel like this is so similar to normal engine implementations, that this should be a normal engine instead of a plugin.

Apart from that we already have implemented something very similar by using `!define tree`, you can get definitions from all definition search engines we already have.

So in my opinion it would make sense to only take the implemented wiktionary API from this PR to create a wiktionary engine for the `define` category returning definitions similar to https://github.com/searxng/searxng/blob/master/searx/engines/wordnik.py.

macarooni-man:
@Bnyro hello, thanks for the follow-up!

I didn't know that `!define` was a feature, I just tried it and that's awesome!

I would prefer to integrate with an existing feature rather than creating a plugin after learning about what you suggested, however, I do believe there is more for your project to benefit from my design than simply a provider.

The existing implementation has potential to improve from a usability & UX standpoint:
- If you're explicitly searching for a definition, it's extra work to expand all the drop-downs to view the content, which is a bit anti-user
- There is no point-of-speech delineation, which can be confusing to map definitions to their semantic context
- There are no examples or linked words which I find to be less accessible, but I also understand the design choice of filtering them for safety
- An entry limit is helpful for visibility as some of the definitions are redundant, which without it requires too much scrolling to view other semantics
- This is a personal preference, but I like the ability to have the card and search results at the same time with a dictionary, and that wasn't something I was able to get working from an answerer alone. I also understand the design choice to limit overhead of concurrent requests

- Maybe it's just a "me" issue, but I find the `!define` implementation more challenging to read and decipher. Line spacing and separation of categories for establishing a clear visual hierarchy a goes a long way with that:
image

image

Bnyro:
I agree, your layout definitely has some aesthetic advantages over the current one.

Perhaps it'd make sense to adjust the existing translations template with your suggestions, so that we don't have two maintain two separate templates? This way translations could also benefit from the better UI design :)

macarooni-man:
> I agree, your layout definitely has some aesthetic advantages over the current one.
> 
> Perhaps it'd make sense to adjust the existing translations template with your suggestions, so that we don't have two maintain two separate templates? This way translations could also benefit from the better UI design :)

Absolutely! I also completely agree that it would be much easier to converge from a maintainability, and both a backend and UX design perspective. I think it would be great to improve more than just the definitions.

If I'm being honest, I learned about SearXNG a few days ago and set up a server for personal use, but I started looking through the codebase yesterday. I don't have enough experience with it to be comfortable modifying a core feature without understanding the implications. It's for that reason specifically I chose the plugin approach, as it doesn't break anything 😊

With that being said, I'd be happy to help however I can within my "green" understanding of the project, and you’re more than welcome to use any part of my code in whatever way fits best. I just wanted to share my implementation for personal use with the hope that it could be beneficial in some way.

Bnyro:
To change the style of the existing translations answer template, you can simply:
1. Create a new file `translations.less` here: https://github.com/searxng/searxng/tree/master/client/simple/src/less/result_types and insert your CSS code
2. Add an import to import your style here: https://github.com/searxng/searxng/blob/2cdbbb249a74f7c03a3f5e5fd29d2ae7a82a57fd/client/simple/src/less/style.less#L1166-L1169, similar to the existing ones
3. Commit your changes and run `make static.build.commit` afterwards

To convert your wiktionary code into an engine, you can just create a new file here: https://github.com/searxng/searxng/tree/master/searx/engines and read through the developer documentation: https://docs.searxng.org/dev/index.html. Perhaps this should be fairly simple given the fact that you already implemented this plugin.

If you have specific questions, please don't hesitate to ask. :)

macarooni-man:
Great @Bnyro, thanks for the suggestions!

I've been digging further into the source and noticed there are already a few dictionary-style providers. I honestly hadn't realized that earlier, very interesting! Just to confirm, are these currently the only engine modules using the `Translations` framework? I want to be certain of the proverbial blast radius with core changes before considering a design:
 * wordnik
 * libretranslate
 * dictzone
 * translated

<br>

Additionally, from an architectural standpoint, I'd like to understand your stance on generated HTML. Since you mentioned improving the translation/definition framework as a whole, I'd prefer that any refactor toward an engine should also include the opportunity to bring the existing modules in line with a consistent structure. For example, adopting some of the normalization and layout ideas from my plugin.

To expand a bit on the HTML standardization, my current implementation preserves formatting like **bold**, *italics*, and [internal links](https://www.youtube.com/watch?v=dQw4w9WgXcQ) (rewritten as local SearXNG searches). Since most of the dictionary engines already simplify structured HTML, this could be implemented fairly cleanly. If I restrict the rendered HTML in the template for only a handful of tags, does that approach align with your expectations around sanitization?

Finally, I'd like to clarify configuration possibilities. Ideally, it would be great to expose a few small options:
 * a toggle to load searches concurrently with any cards (if feasible within the answerer framework, and disabled by default)
 * a configurable limit on definition/example entries, defaulting to 5/3 respectively seems comfortable based on my testing

<br>

I know scope creep is a common concern and I want to be respectful of that, but at the same time, there are some features like the HTML parsing which would require modifying how the template is rendered, and also, how data is passed into it. So I just want to be clear about where the line is for you on that.

I mainly want to make sure these goals fit within your long-term direction before I start moving pieces around. Let me know your thoughts!

return42:
> For example, adopting some of the normalization and layout ideas from my plugin.

I have to repeat, the concept behind are the [Result Types](https://docs.searxng.org/dev/result_types/index.html#result-types) ..

https://github.com/searxng/searxng/blob/2cdbbb249a74f7c03a3f5e5fd29d2ae7a82a57fd/searx/result_types/answer.py#L120-L122 

.. in this concept, a result type like `Translations` is bound to a html template in the  *simple* theme `simple/answer/translations.html` and its not intended to bound `Translations` results to a template `define.html`

```python
Translations(
    ...
    template="answer/define.html"
) 
```

On Client side, we have the CSS (aka LESS) https://github.com/searxng/searxng/tree/master/client/simple/src/less 

See https://github.com/searxng/searxng/pull/5366#issuecomment-3443907555

For example, the `WeatherAnswer` type ..

https://github.com/searxng/searxng/blob/2cdbbb249a74f7c03a3f5e5fd29d2ae7a82a57fd/searx/result_types/answer.py#L156-L161

HTML template is in

- https://github.com/searxng/searxng/blob/master/searx/templates/simple/answer/weather.html

and the CSS layout of the client is in 

- https://github.com/searxng/searxng/blob/master/client/simple/src/less/weather.less


> To expand a bit on the HTML standardization, my current implementation preserves formatting like **bold**, _italics_,

Don't do it, the [engines, plugins and *answerers*](https://docs.searxng.org/dev/result_types/index.html#result-types) .. the code that produce the `BaseAnswer`s does not decide about rendering, this code returns a result type .. noting more .. you can re-use existing  `Translations` types if the rendering for your results do not need any special UI .. 

If you want a UI different from `Translations` you have to implement a result (answer) type *Definitions* which is ..

> In terms of data structure, this class here is very similar to the type `Translations` 

see my comment above: https://github.com/searxng/searxng/pull/5366#discussion_r2459955930

I hope my explanations have been able to show “where” we want to go... As already mentioned above, this concept of result-types is currently still under development and therefore somewhat difficult to understand .. If I had more time, I would rebuild your PR... However, that would take longer, as I currently have many other unfinished tasks to deal with.

Perhaps you would like to try to restructure the code in this PR yourself so that it fits better into our concept of result types? On the other hand, if you have some time and patience with me, we can also do this together.

macarooni-man:
@return42 as Bnyro and I were discussing before, we were considering a restructure of `Translations` as a whole to align with the design motif of my plugin, not to use a custom template. That way, all the `Translations` type engines could support those features and improve the entire system.

However, your "don’t do it" framing and the way my questions were brushed aside comes across as fairly dismissive, and if your intent is to collaborate as you suggested, I don't feel like my perspective is being heard. My goal was to clarify architectural intent and find a collaborative path forward, and it seems as though you're firmly against the idea of restructuring the `Translations` template, rendering, or functionality.

I respect that, and that's why I asked about where your scope lies. If that is in fact your stance, I don't have an interest in continuing to work on this. Though, you're more than welcome to use my code.

return42:
> However, your "don’t do it" framing and the way my questions were brushed aside comes across as fairly dismissive, and if your intent is to collaborate as you suggested, 

My english isn't the best .. in the core I offered you my support.

> we were considering a restructure of `Translations` as a whole to align with the design motif of my plugin, not to use a custom template. That way, all the `Translations` type engines could support those features and improve the entire system.

If you prefer to restructure existing code .. then yes, please *consider a restructure of `Translations`* :+1:

macarooni-man:
> If you prefer to restructure existing code .. then yes, please _consider a restructure of `Translations`_ 👍

Okay, great. With this perspective, could you both take another look through my [previous comment](https://github.com/searxng/searxng/pull/5366#issuecomment-3444855179)? I want to ensure that we're on the same page, and that I understand the scope of these changes before proceeding with a design.

Bnyro:
> Additionally, from an architectural standpoint, I'd like to understand your stance on generated HTML. Since you mentioned improving the translation/definition framework as a whole, I'd prefer that any refactor toward an engine should also include the opportunity to bring the existing modules in line with a consistent structure. For example, adopting some of the normalization and layout ideas from my plugin.
>
> To expand a bit on the HTML standardization, my current implementation preserves formatting like bold, italics, and [internal links](https://www.youtube.com/watch?v=dQw4w9WgXcQ) (rewritten as local SearXNG searches). Since most of the dictionary engines already simplify structured HTML, this could be implemented fairly cleanly. If I restrict the rendered HTML in the template for only a handful of tags, does that approach align with your expectations around sanitization?
>
Sure, please do so :+1: 

> Finally, I'd like to clarify configuration possibilities. Ideally, it would be great to expose a few small options:
>
>    a toggle to load searches concurrently with any cards (if feasible within the answerer framework, and disabled by default)
>    a configurable limit on definition/example entries, defaulting to 5/3 respectively seems comfortable based on my testing
>
I think you can do that in a different follow-up PR.

return42:
>> If I restrict the rendered HTML in the template for only a handful of tags, does that approach align with your expectations around sanitization?

> Sure, please do so 👍

I don't have a final meaning about the HTML markup.  In order to separate data from its representation, we have recently made considerable efforts to remove all HTML tags from the result items of the engines and have begun to build up the result types.

The requirement for an *inline* markup for paragraph-like types has been on my mind for quite some time:

We could (as suggested above) use a subset of HTML tags for this purpose, but then we would have to agree on which tags are included in this set, how they can be combined, and much more. In my opinion, this approach is wrong because it would mean we would have to start developing our own markup. Another disadvantage would be the presence of a few HTML tags in the strings of the JSON output.

What we need are basic inline markups such as those found here in the [reST markup](https://docs.searxng.org/dev/reST.html#basic-inline-markup).

For this, we can implement a class `CommonMark` for `str` objects with some basic inline markups, similar to the [DateTime](https://github.com/searxng/searxng/blob/ea4a55fa57a4682f48265d50a9d816801ead0b88/searx/weather.py#L209) type to wrap `datetime.datetime` objects.

Something like (untested) ...

```python
import functools
import msgspec
from markdown_it import MarkdownIt

class CommonMark(msgspec.Struct):
    """String type with some `basic inline markup`_ (CommonMark_)

    .. _basic inline markup: https://docs.searxng.org/dev/reST.html#basic-inline-markup
    .. _CommonMark: https://commonmark.org/
    """

    p: str

    def __post_init__(self):
        # normalize whitespaces to one space
        self.p = " ".join(self.p.split())

    def __str__(self):
        return self.p

    @functools.cached_property
    def html(self):
        """Generates a string with HTML markup from CommonMark (uses markdown-it-py_).

        .. _markdown-it-py: https://github.com/executablebooks/markdown-it-py
        """
        return (
            MarkdownIt("commonmark", {"typographer": True}).enable(["replacements", "smartquotes"]).render(self.p)
        )
```

In line 137 we can replace the `str` by the `CommonMark` from above.

https://github.com/searxng/searxng/blob/ea4a55fa57a4682f48265d50a9d816801ead0b88/searx/result_types/answer.py#L131-L138

And in the template we replace `item.text` by `item.text.html`:

https://github.com/searxng/searxng/blob/ea4a55fa57a4682f48265d50a9d816801ead0b88/searx/templates/simple/answer/translations.html#L5

return42:
Closed due to inactivity.

PR #5382, opened 2025-10-26
[fix] qwant engine: setting accept-language to bypass bot detection

Aadniz:
## What does this PR do?

This PR sets accept language for the Qwant engine. During testing, it seems like setting the accept language gives more success for bypassing bot detection. Tested with a few ~20 searches.

It doesn't seem like Qwant cares what the accept language is, just that it is set to something. Since we are already getting the locale for the query string, I felt it was convenient to just put that as the accept language.

Qwant does not seem to work on any SearXNG instance right now, and this is a fix for this issue.

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -82,6 +82,9 @@ max_page = 5
 """5 pages maximum (``&p=5``): Trying to do more just results in an improper
 redirect"""
 
+# Otherwise Qwant will return 403 if not set
+send_accept_language_header = True
+
 qwant_categ = None
 """One of ``web-lite`` (or ``web``), ``news``, ``images`` or ``videos``"""
 


return42 (approved):
> Qwant does not seem to work on any SearXNG instance right now, and this is a fix for this issue.

Thanks for digging into that issue and giving us back the Qwant engine 🚀 

Engines do have a [`send_accept_language_header`](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.send_accept_language_header) option / I modified your patch, see my [diff](https://github.com/searxng/searxng/compare/efae8b0664b20b082b3ad9972feb4862c74e73ec..cf5af8d746f20ccf928bcbd70429f47803c873fc)

Aadniz:
> Engines do have a [send_accept_language_header](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.send_accept_language_header) option / I modified your patch, see my [diff](https://github.com/searxng/searxng/compare/efae8b0664b20b082b3ad9972feb4862c74e73ec..cf5af8d746f20ccf928bcbd70429f47803c873fc)

Oh ok, I didn't know that. Thanks.

PR #5383, opened 2025-10-26
[fix] deviantart engine: does not return any results

Aadniz:
## What does this PR do?

Deviantart does not return any results. It seems like the `xpath` match has changed.

Maybe there is a more robust way of getting the results, but this is good enough for now.

Currently broken on all instances


return42 (approved):
> Currently broken on all instances

Your fix works like a charm :-) / thanks a lot 👍

Issue #5286, opened 2025-10-07
Bug: google engine


Jeto143:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

Repository: https://github.com/searxng/searxng
Branch: master
Version: 2025.10.4+7bf65d68c

**How did you install SearXNG?**

Docker setup (via Coolify).

**What happened?**

Google no longer yields any results.

**How To Reproduce**

Select Google as your only search engine and try searching something (or select it among other engines and notice that no "google" results appear).

**Screenshots & Logs**

No specific errors appear in the logs. I see this one below on startup though, so hopefully Google didn't ban my server's IP because I didn't address it before or something:
> 2025-10-07 20:23:29,156 ERROR:searx.botdetection: X-Forwarded-For nor X-Real-IP header is set!

**Additional context**

Was working fine an hour or so ago (9 PM CET).

rmaake1:
+1, 4get's google engine isn't working either, i think google may have patched the approach both searxng and 4get have been using to grab results without javascript being required :\ they've been on a warpath blocking easy SERP access

lolcatw:
CSE works. API works. I'm going to scrape both

return42:
For a search query like `!go foo`, the outgoing request is:

    https://www.google.com/search?q=foo&hl=en-US&lr=lang_en&cr=&ie=utf8&oe=utf8&filter=0&start=0&asearch=arc&async=arc_id%3Asrp_-JOX0lWm09eUzwILiJukthn_100%2Cuse_ac%3Atrue%2C_fmt%3Aprog

And in the response (see below) I can see:

*Please check your microphone .. https://support.google.com/chrome/?p\x3dui_voice_search\ .. to start voice search .. voice search not available .. Please check your microphone and audio levels .. Voice search has been turned off..*

I am not sure if these messages are new and if they have any relevance, but I have the impression that the request seems to expect a voice input

----

Here is one example for the response:
```
)]}'
[...]

PR #5387, opened 2025-10-28
[mod] Engines: Temporarily disable Google

benpiano800:
## What does this PR do?

This pull request disables the Google and Google News engines for now, which do not return any results. They should be re-enabled once the Google issue is fixed. It also enables Mullvad Leta by default, which uses Google as a backend and works right now.

Google is the most popular search engine and is enabled by default in SearXNG, but there is an issue with it right now that causes it to not return any results. Replacing it with Mullvad Leta until the engine is resolved should improve result quality.

Disable Google and Google News in your instance and enable Mullvad Leta.

## Related issues
#5286

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1027,6 +1027,7 @@ engines:
   - name: google
     engine: google
     shortcut: go
+    disabled: true
     # additional_tests:
     #   android: *test_android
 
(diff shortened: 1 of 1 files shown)


inetol:
Thanks for the PR. We are not going to default to mullvadleta engine as it has a significantly more sensitive ratelimit than Google engine. Also, Google engine does not explicitly return a search error, so I don't consider it necessary to modify the default `settings.yml` template just to revert those changes in the future, you can always disable Google engine on preferences or on your own instance config.

PR #5396, opened 2025-10-29
[feat] engine: add grokipedia

### LINKED ISSUE #5393: Add new engine: Grokipedia
Working URL to the engine
https://grokipedia.com/search?q=%s

Why do you want to add this engine?
Grokipedia is a newly launched (October 28, 2025) crowdsourced online encyclopedia by Elon Musk's xAI, positioned as an unbiased, AI-enhanced alternative to Wikipedia. It emphasizes "maximum truth" with reduced ideological bias, making it a valuable addition for users seeking diverse, AI-augmented knowledge sources. While not explicitly open source, it's tied to xAI's mission of advancing scientific understanding of the universe, and its content is generated using Grok AI models, potentially allowing for innovative, real-time fact-checking integrations in metasearch tools like SearxNG. 

Features of this engine
Grokipedia offers AI-powered article generation and summarization, focusing on bias reduction and truth-seeking through Grok-4 models. It claims over 885,000 articles (though fewer than Wikipedia's 7 million in English), with minimalist design, no apparent user tracking in its basic interface, and potential for fast, on-demand content creation. Articles include citations, though sourcing can be lighter than Wikipedia's, and it supports searches via a simple query bar for topics like history, science, or current events. 

Applicable category of this engine
general, science

Austin-Olacsi:
## What does this PR do?

adds grokipedia engine

Grokipedia is a Wikipedia alternative that was largely created using AI. Was requested.

make run
!gp test

## Related issues

closes https://github.com/searxng/searxng/issues/5393

### DIFF
--- a/searx/engines/grokipedia.py
+++ b/searx/engines/grokipedia.py
@@ -1,0 +1,52 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Grokipedia (general)"""
+
+from urllib.parse import urlencode
+from searx.utils import html_to_text
+from searx.result_types import EngineResults
+
+about = {
+    "website": 'https://grokipedia.com',
+    "wikidata_id": "Q136410803",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://grokipedia.com/api/full-text-search"
+categories = ['general']
+paging = True
+results_per_page = 10
+
+
+def request(query, params):
+
+    start_index = (params["pageno"] - 1) * results_per_page
+
+    query_params = {
+        "query": query,
+        "limit": results_per_page,
+        "offset": start_index,
+    }
+
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+
+    return params
+
+
+def response(resp) -> EngineResults:
+    results = EngineResults()
+    search_res = resp.json()
+
+    for item in search_res["results"]:
+
+        results.add(
+            results.types.MainResult(
+                url='https://grokipedia.com/page/' + item["slug"],
+                title=item["title"],
+                content=html_to_text(item["snippet"]),
+            )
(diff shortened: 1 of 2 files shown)


inetol (dismissed):

searx/engines/grokipedia.py
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+
+    return params
+
+
+def response(resp):
+    search_res = resp.json()
+    results = []
```
Bnyro: Please use the new `EngineResults`, see https://github.com/searxng/searxng/blob/52ffc4c7f43a0d8e257278d40efd6ab4b58e459f/searx/engines/reuters.py#L77-L93 for example

searx/engines/grokipedia.py
```diff
@@ -0,0 +1,51 @@
...
+
+    for item in search_res.get("results", []):
```
Bnyro:
```suggestion
    for item in search_res["results"]:
```
If there are no results, the API returns an empty array, so that fallback is never used.

searx/settings.yml
```diff
@@ -1072,6 +1072,11 @@ engines:
     play_categ: movies
     disabled: true
 
+  - name: grokipedia
+    engine: grokipedia
+    shortcut: gp
+    disabled: true
```
return42, line 1078: The engine is controversial, see

- https://github.com/searxng/searxng/issues/5393#issuecomment-3560844654

which is why I would like to suggest to set this engine *inactive* by default:


```yaml
    inactive: true
```

Austin-Olacsi: Most Grokopedia articles aren't about political topics.

I personally disagree with setting the engine to inactive, but doing so would be an acceptable compromise to me, so i'll change it to inactive anyway.

Bnyro (approved):

PR #5406, opened 2025-11-01
[mod] client/simple: client plugins

inetol:
This PR defines a new interface for client side *"plugins"* that coexist with server side plugin system. Each plugin (e.g., `InfiniteScroll`) extends the base `Plugin`. Client side plugins are independent and lazy‑loaded via `router.ts` when their `load()` conditions are met. On each navigation request, all applicable plugins are instanced.

```mermaid
classDiagram
direction BT
class Calculator {
   constructor() 
   MathJsInstance math
   run() Promise~string | undefined~
   post(result: string) Promise~void~
}
class InfiniteScroll {
   constructor() 
   run() Promise~void~
   post() Promise~void~
}
class MapView {
   constructor(map: HTMLElement) 
   HTMLElement map
   run() Promise~void~
   post() Promise~void~
}
class Plugin {
   constructor(id: string) 
   string id
   invoke() Promise~void~
   run() Promise~unknown~
   post(result: NonNullable~Awaited~ReturnType~this["run"]~~~) Promise~void~
}

Calculator  -->  Plugin 
InfiniteScroll  -->  Plugin 
MapView  -->  Plugin 
```

Since these are client side plugins, we can only invoke them once DOM is fully loaded. E.g. `Calculator` will not render a new `answer` block until fully loaded and executed.

For some plugins, we might want to handle its availability in `settings.yml` and toggle in UI, like we do for server side plugins. In that case, we extend `py Plugin` instancing only the information and then checking client side if [`settings.plugins`](https://github.com/inetol/searxng/blob/1ad832b1dc33f3f388da361ff2459b05dc86a164/client/simple/src/js/toolkit.ts#L134) array has the plugin id.

Due to those changes, Vite bundling also needed some rework to properly lazy-load.

---

I expected to refactor `search.ts` as a plugin too, but the PR is already somewhat large, so I left it halfway done. No logic changes were made.

I am not entirely convinced about having to [recreate the `answers` div from scratch](https://github.com/inetol/searxng/blob/b3a35a5a5573f5388164ed2adad31915280ed5cf/client/simple/src/js/util/appendAnswerElement.ts) on client side.

Maybe we can add a condition in the [answers.html template](https://github.com/inetol/searxng/blob/b3a35a5a5573f5388164ed2adad31915280ed5cf/searx/templates/simple/elements/answers.html) to add the style [`display:none`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display#display_none) if no `answer` blocks are appended server side?


return42:
Thank you very much! I've taken just a first look at it .. here are a few unsorted questions/suggestions:

> E.g. QueryPlugin (only serves as a "showcase"), which will retrieve the query and append a `answer` block inside the `answers` container.

:thinking: is `QueryPlugin` or `InfiniteScrollPlugin` really a "showcase" that implements all our needs?  From what I understand (not sure I have a correct picture in mind), we have various types of plugins that would require client implementation

1. Pure JS client implementation, no (async) request/response communication from JS client to SearXNG's server, none implementations on server site.  The `QueryPlugin` is an example for this use-case .. I would like to move the calculator plugin, whose [active code is currently running on the server](https://github.com/searxng/searxng/blob/master/searx/plugins/calculator.py), to the client in this way.

2. The active code is on the client; on the server side, there is only a configuration. The `InfiniteScrollPlugin` is an example of this.  In this example, pure HTML content is retrieved from a URL and integrated into the DOM, no process on the received content, no templates on client site.

3. The content is generated by the plugin on the server side, the client makes an async HTTP request and receives a JSON response from the server.  The JSON is processed on the client and content is inserted into the HTML body (HTML templates required on the client side?). 

> as I am not entirely convinced about having to recreate the `answers` div from scratch on JS side if none is present on DOM.

I don't have a blanket answer to this question. I think we first need to consider what the implementations on the client and server side would look like for the third case.

For example, will it be possible for a server-side plugin to register a URL endpoint through which the client (async / what about bot blocking?) can request content from the plugin? / Is there a routing mechanism to pass requests from the client to the plugins and generate the response directly from the plugin? 

-----

I am still in the early stages of forming my thoughts on this topic. I am not even sure whether my above classifications 1, 2, and 3 are appropriate.

Perhaps you already have more concrete ideas?

inetol:
I still have to establish and document the Plugin class, so take this as a pinch of salt.


> 🤔 is `QueryPlugin` or `InfiniteScrollPlugin` really a "showcase" that implements all our needs? From what I understand (not sure I have a correct picture in mind), we have various types of plugins that would require client implementation

ATM, I only added the "answers" as some plugins uses this, but we can easily handle other "DOM modifiers" depending on the conditions we are invoking the plugin.

`InfiniteScrollPlugin` is the converted module of https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/client/simple/src/js/main/infinite_scroll.ts btw.


> 1. Pure JS client implementation, no (async) request/response communication from JS client to SearXNG's server, none implementations on server site.  The `QueryPlugin` is an example for this use-case .. I would like to move the calculator plugin, whose [active code is currently running on the server](https://github.com/searxng/searxng/blob/master/searx/plugins/calculator.py), to the client in this way.
> 2. The active code is on the client; on the server side, there is only a configuration. The `InfiniteScrollPlugin` is an example of this.  In this example, pure HTML content is retrieved from a URL and integrated into the DOM, no process on the received content, no templates on client site.

This is the best use for client side plugins, and easier to implement.

The server may enforce or defaults if "x" plugin is enabled by passing a boolean like we do for `infinite_scroll`

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/webapp.py#L368-L386


> 3. The content is generated by the plugin on the server side, the client makes an async HTTP request and receives a JSON response from the server.  The JSON is processed on the client and content is inserted into the HTML body (HTML templates required on the client side?).

In this case, server side plugins are a better fit.

E.g., the `Tracker URL remover`, not only the execution of the script will delay as needs to fetch the rules from a remote server (if the user is quick to click a result before the plugin finishes its execution, the URL won't be stripped), but having every client download the entire rules file, overloading the source server and making (possible) cross site requests that could breach privacy.


> For example, will it be possible for a server-side plugin to register a URL endpoint through which the client (async / what about bot blocking?) can request content from the plugin? / Is there a routing mechanism to pass requests from the client to the plugins and generate the response directly from the plugin? 

Yes it's possible, but maybe somewhat difficult to handle properly. The client side plugin would fetch from an endpoint created by a server side plugin (if both plugins enabled), client deserializes and saves the data in a property on plugin invoke to be used on `#do()` or whatever methods we implement.

inetol:
I stabilized the Plugin class and sorted out some minor things, also replaced the prototype `QueryPlugin` with the new "migrated" `Calculator` plugin.

image

I still need to figure out how to properly manage client plugins on the server, I probably made a mess in https://github.com/searxng/searxng/pull/5406/commits/69747d7368db1052ec632e9a4a1dcb60183ac318, any ideas are welcome

return42:
> I still need to figure out how to properly manage client plugins on the server, I probably made a mess in [69747d7](https://github.com/searxng/searxng/commit/69747d7368db1052ec632e9a4a1dcb60183ac318), any ideas are welcome

Yes, I see :-) .. I try to give an rough overview about, what I have in mind ..

First I think we should not remove the Plugins on server side .. this implementation is needed to have the plugin in the preferences, where the user can activate/inactivate the plugin .. and these are also the plugins we can manage .. currently, in your patch, the calculator plugin has been removed from the management of the plugins and a new group of settings has been added to cover *client-plugins*.

For the calculator (for example) we need a plugin on server that has at minimum:

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/plugins/calculator.py#L25-L40

The other methods that were required to perform the calculation are no longer needed (code that has been moved to the client) and can be deleted from the plugin on the server side.

We manage the plugins in the `PluginStorage` class:

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/plugins/_core.py#L192-L193

The `PluginStorage` class needs a new method (or `@property` method) `PluginStorage.client_plugins`.  This method iterates over the [registered plugins](https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/plugins/_core.py#L201-L202) and collects the `PluginCfg` objects from each Plugin.

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/plugins/_core.py#L178-L179

```python
class PluginStorage:
    ...
    @property
    def client_plugins(self):
        return [p.client_plugin_cfg for p in self.plugin_list]
```

In the `Plugin(abc.ABC)` class ..

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/plugins/_core.py#L68-L70

we implement a new method (or `@property` method) `Plugin.client_plugin_cfg` that build and returns a `PluginCfg` object.

```python
class Plugin(abc.ABC):
    ...
    @property
    def client_plugin_cfg(self):
        return PluginCfg(active=self.active)
```

To summarize: we have `PluginStorage.client_plugins` from where we get a list of `PluginCfg`.

In order to later generate a JSON string (we can pass to the JS client) from the `PluginCfg` object in a performant and simple manner, `PluginCfg` must be changed from `dataclass` to `msgspec.Struct`.

A `msgspec.Struct` object can be converted [`to_builtins` (python types)](https://jcristharif.com/msgspec/api.html#msgspec.to_builtins) and we can pass this list of objects to the client:

```python

def get_client_settings():
    req_pref = sxng_request.preferences
    return {
        'plugins': msgspec.to_builtins(searx.plugins.STORAGE.client_plugins)
```

-----

I could implement that for you, but I currently have other, more urgent tasks (Google engine, etc.) that I want to work on first. Either you wait until I implement the above, or you try it yourself .. if you think it's taking too long with me :-)

inetol:
Thanks for the guide, a lot cleaner than my hackish implementation. I thought `py Plugin` did more stuff behind the scenes, I should have checked it out earlier.

> To summarize: we have `PluginStorage.client_plugins` from where we get a list of `PluginCfg`.

We really need to differentiate between client or server plugin that much? Both types are plugins, but client ones doesn't implement any server side logic, only instances ~metadata~ info:

https://github.com/inetol/searxng/blob/83507ca671ce3c204ae105b222c77bb3fadc6d49/searx/plugins/calculator.py

For the client, its sufficient to know if the plugin id is inside the array:

https://github.com/inetol/searxng/blob/83507ca671ce3c204ae105b222c77bb3fadc6d49/client/simple/src/js/router.ts#L10-L23

...returned by the list of enabled plugins:

https://github.com/inetol/searxng/blob/83507ca671ce3c204ae105b222c77bb3fadc6d49/searx/webapp.py#L371

I can implement a new storage method if you're not convinced, but I think it's not necessary.


> I could implement that for you, but I currently have other, more urgent tasks (Google engine, etc.) that I want to work on first. Either you wait until I implement the above, or you try it yourself .. if you think it's taking too long with me :-)

Got it :)

return42 (approved):
LGTM: I can't evaluate the TS implementations, but I really like the structure .. this pull request not only brings client plugins, it also brings more structure back to our codebase. .. 🚀 

-----

> For the client, its sufficient to know if the plugin id is inside the array:

:+1: .. for now, its good enough .. if there comes the time, we need to ship more info from server to client, we can consider to serialize `PluginInfo` objects.

-----

@Bnyro this PR turns the *server-side* calculator plugin into a *client-side* plugin.

client/simple/src/js/Plugin.ts
```diff
@@ -0,0 +1,66 @@
...
+  protected readonly id: string;
+
+  /**
+   * @remarks
+   * Don't hold references of this instance outside the class.
+   */
+  protected constructor(id: string) {
+    this.id = id;
+
+    void this.invoke();
```
Bnyro, line 26: I assume we use `void` here to throw away the `Promise` we get in return? 

If we used `this.invoke()` without `void`, this should be the same though, because we're not returning `void this.invoke()`, so it's thrown away anyways?

client/simple/src/js/Plugin.ts
```diff
@@ -0,0 +1,66 @@
...
+  protected abstract run(): Promise<unknown>;
+
+  /**
+   * Post-execution hook.
+   *
+   * @remarks
+   * The hook is only executed if `#run()` returns a truthy value.
+   */
+  // @ts-expect-error
+  protected abstract post(result: NonNullable<Awaited<ReturnType<this["run"]>>>): Promise<void>;
```
Bnyro, line 65: I feel like having two different methods here is a bit unnecessary, since they're always both called sequentially anyways in case of no errors.

Also, the `post` method name is not really self-explaining, i.e. what should be done in `post` and what should be done in `run`? Is `post` supposed to mean something like `manipulateDom` here, so that the data loading and UI updating logic are separated maybe?

client/simple/src/js/loader.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    case "endpoint": {
+      if (!options.where.includes(endpoint)) {
+        // not on the expected endpoint
+        return false;
+      }
+
+      return true;
+    }
+    default: {
```
Bnyro: Maybe we could debug a warning about misconfiguration here here to save time debugging typos when using the wrong value for `options.on`, as I don't think the `default` case should ever be encountered if the plugin is set up properly?

client/simple/src/js/plugin/InfiniteScroll.ts
```diff
@@ -0,0 +1,112 @@
...
+      assertElement(paginationElement);
+
+      paginationElement.replaceChildren(spinnerElement);
+
+      try {
+        const res = await http("POST", action, { body: new FormData(form) });
+        const nextPage = await res.text();
+        if (!nextPage) return;
+
+        const nextPageDoc = new DOMParser().parseFromString(nextPage, "text/html");
```
Bnyro, line 47: Maybe we could also add a new Flask endpoint here in the future that only returns the rendered results (without the search bar etc.) so that we don't have to parse the DOM here? (not related to this PR though)

That would save some CPU and network time

client/simple/src/js/plugin/InfiniteScroll.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (!urlsElement) {
+          throw new Error("URLs element not found");
+        }
+
+        if (articleList.length > 0 && !onlyImages) {
+          // do not add <hr> element when there are only images
+          urlsElement.appendChild(document.createElement("hr"));
+        }
+
+        urlsElement.append(...Array.from(articleList));
```
Bnyro:
```suggestion
        urlsElement.append(...articleList);
```
Creating a temporary array is not necessary here, `articleList` should already be of type `NodeList<HTMLElement` afaik

client/simple/src/js/plugin/InfiniteScroll.ts
```diff
@@ -0,0 +1,112 @@
...
+        observer.unobserve(paginationEntry.target);
+
+        void loadNextPage(() => {
+          const nextObservedElement = document.querySelector<HTMLElement>(observedSelector);
+          if (nextObservedElement) {
+            observer.observe(nextObservedElement);
+          }
+        }).then(() => {
+          // wait until promise is resolved
+        });
```
Bnyro: That's just a `noop` since `then` is running async and not blocking anything?

client/simple/src/js/router.ts
```diff
@@ -0,0 +1,69 @@
...
+      where: [Endpoints.results]
+    });
+  }
+
+  if (settings.plugins?.includes("calculator")) {
+    load(() => import("./plugin/Calculator.ts").then(({ default: Plugin }) => new Plugin()), {
+      on: "endpoint",
+      where: [Endpoints.results]
+    });
+  }
```
Bnyro, line 36: That's an okay workaround for now, but in the future we might want to find a way to not re-declaring the plugin here and only declare it in the Python plugin file. I don't have a real idea how to do that at the moment though.

Bnyro:
I'm a bit sad to see the Python calculator plugin go because I had a lot of fun developing it, but I agree with the overall idea of the changes here.

This might also allow me to finally implement the client-side hostname replace plugin I started in https://github.com/searxng/searxng/pull/2946 in a clean way :)

Great stuff @inetol :)

Bnyro:
> I am not entirely convinced about having to [recreate the answers div from scratch](https://github.com/inetol/searxng/blob/b3a35a5a5573f5388164ed2adad31915280ed5cf/client/simple/src/js/util/appendAnswerElement.ts) on client side.
>
> Maybe we can add a condition in the [answers.html template](https://github.com/inetol/searxng/blob/b3a35a5a5573f5388164ed2adad31915280ed5cf/searx/templates/simple/elements/answers.html) to add the style [display:none](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/display#display_none) if no answer blocks are appended server side?
>
I'd find it very cool to be able to use the Jinja templates on client side too, but I only found some no longer maintained JavaScript libraries to parse Jinja files with JavaScript, e.g. https://www.npmjs.com/package/jinja-to-js.

Bnyro:
Oh, I forgot:

Once merged, we should also do an announcement about the breaking change that infinite scroll is a plugin yet again, so instance admins must change their configuration.

client/simple/src/js/Plugin.ts
inetol: ~Vite~ oxc stripes all `void` keyword on dist, so yes *it's useless* (I don't know about transpilation hinting, maybe it also helps to generate better JS). Only serves as a "best practice" rule enforced by Biome to explicitly handle promises.

https://biomejs.dev/linter/rules/no-floating-promises/

client/simple/src/js/Plugin.ts
inetol: My idea was to introduce further logic in `#invoke()` later on to handle more return types from `#run()`, as well as to differentiate the plugin own runtime logic and it's output transformations.

client/simple/src/js/loader.ts
inetol: `options.on` is typed, comes from [`Options`](https://github.com/searxng/searxng/pull/5406/files/c2377e4623cb8a3e6b4e6b105c211793f65dbc72#diff-02c71d9a4a1bcb2e18ee081f885f7ed918b90414b6435d53f0941cf3be1b886bR6-R13) type object union.

I agree that the `default` case here is completely useless, but this is another "best practice" rule from Biome, only useful for unknown expressions.

https://biomejs.dev/linter/rules/use-default-switch-clause/

client/simple/src/js/plugin/InfiniteScroll.ts
inetol: I think I wrote this because the [`noMisusedPromises`](https://biomejs.dev/linter/rules/no-misused-promises/) rule threw warnings if I set the IntersectionObserver callback async, I'll check if it's fixed.

inetol: 66e887b51619a8841fc265f95924305b8897237c

client/simple/src/js/plugin/InfiniteScroll.ts
inetol: 66e887b51619a8841fc265f95924305b8897237c

inetol:
> I'd find it very cool to be able to use the Jinja templates on client side too, but I only found some no longer maintained JavaScript libraries to parse Jinja files with JavaScript, e.g. https://www.npmjs.com/package/jinja-to-js.

That would be impossible as the theme JavaScript runs in the client and Jinja templates needs to render on the server.

https://www.npmjs.com/package/jinja-to-js would work if the backend is JavaScript based, but then it wouldn't make sense to use Jinja as there are better template options such as JSX... (long etcetera...)

Bnyro:
> https://www.npmjs.com/package/jinja-to-js would work if the backend is JavaScript based, but then it wouldn't make sense to use Jinja as there are better template options such as JSX... (long etcetera...)
>
Oh, I perhaps misread the package details: I thought it would statically compile the Jinja template into a JS file that can be called on runtime from the client, i.e. that `make static.build.commit` would generate boilerplate JS that renders a specific template when called at runtime.

It doesn't make sense to add any additional rendering engines, the idea was to re-use the existing templates to avoid duplicating the same logic with Jinja files for server side and pure JavaScript for client side.

client/simple/src/js/loader.ts
Bnyro: Then you should suppress the linting rule here, because as you said, it can never happen.

See https://biomejs.dev/analyzer/suppressions/.

inetol: https://github.com/searxng/searxng/compare/bb2bc000fa424b28b17b82385797936d6ff195be..be7466eac5f9368b7672a8cb5fedebee6c4dbd6d

Bnyro (approved):

inetol:
Now checking, I noticed that plugins don't render if disabled (except if listed in "Special Queries" category), making it impossible to leave a plugin disabled but available in the UI for the user to reactivate it. As I understand, the plugin should only disappear from the UI if it's not listed in [`plugins:`](https://docs.searxng.org/admin/settings/settings_plugins.html), regardless of its category?

Right now [`infinite_scroll` is disabled by default](https://github.com/searxng/searxng/pull/5406/files#diff-e95e3f454925bc2344d3cf538950b11a1255935b47a609c9589e6ac33d74802fR223-R225), so it should appear in the "User Interface" section as "disabled", but instead it doesn't render at all. Is this expected?

Bnyro:
> As I understand, the plugin should only disappear from the UI if it's not listed in [plugins:](https://docs.searxng.org/admin/settings/settings_plugins.html), regardless of its category?
>
Yes, exactly. So what you are describing is not the expected behavior, it should still be visible in the settings.

return42:
@inetol about visibility of plugins in the UI: if you want, I can have a look at and push my suggestion on top of your branch .. but I don't know when I have time for (may at the end of the week I have time for).

inetol:
It seems that I had some cached responses between branch changes and couldn't see the section rendered. Running `make clean` and starting a new session in the browser fixed the issue 😓

bohanubis:
I know I am using git version of searxng
but is this pr is breaking the interface for clients
commit: ab8224c9394236d2cbcf6ec7d9bf0d7c602ca6ac
image
commit: fb089ae297b27f51777318e3a28bca8b172a4165
image


I tested in both firefox and brave just in case

vojkovic:
@bohanubis Try clearing your cache and trying again - And if the issue still persists please could you also open an issue?

PR #5407, opened 2025-11-01
[enh] py: drop deps

inetol:
The difference between decompression with brotli or gzip in HTML files is negligible for 3 MB of compiled binary package.

`unixthreadname` looks like a weird workaround to me, this wasn't handled on uWSGI/Granian side?

Introduced in https://github.com/searxng/searxng/commit/eaa694fb7d0e47b943bc6d6edb6cb6a40ab2d85e

Closes https://github.com/searxng/searxng/security/code-scanning/276
Closes https://github.com/searxng/searxng/security/dependabot/37


requirements.txt
```diff
@@ -9,8 +9,6 @@ python-dateutil==2.9.0.post0
 pyyaml==6.0.3
 httpx[http2]==0.28.1
 httpx-socks[asyncio]==0.10.0
-Brotli==1.1.0
```
return42, line 12: As far I remember this Brotli requirement was added in first place to the httpx stack, to process incoming HTTP responses:

- [*"Support HTTP/2 and brotli.*"](https://github.com/searx/searx/pull/1856)

However, I don't know if there is any engine that absolutely requires Brotli. Usually, (http incomming) compression can be controlled via the [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding) header.

The default is:

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/webapp.py#L1043

If we remove `Brotli` here (what is OK for me), we should also have a look at the `Accept-Encoding` header in our outgoing requests and remove the `br`

https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/engines/tineye.py#L76

Other places can be found by the following search query:

- https://github.com/search?q=repo%3Asearxng%2Fsearxng+Accept-Encoding&type=code

inetol: There's not a single site that doesn't support uncompressed responses, so sites not serving br content they will fall back on something else or uncompressed, but I don't understand why `Accept-Encoding` header is overridden on some engines. httpx will set the supported encodings automatically, as every HTTP client does.

This looks like a *valid* exception:
https://github.com/searxng/searxng/blob/0245327fc5b5d3b953c7169f58bad989eb54e552/searx/engines/brave.py#L201-L202

...The other matches are not.

return42: > why Accept-Encoding header is overridden on some engines.
> ..
> httpx will set the supported encodings automatically, as every HTTP client does.

I guess it was my fail .. when reverse engineering these engines, I simply copied the header from my browser without considering that SearXNG (or httpx) already sets a default value for it.

> the supported encodings automatically, as every HTTP client does.

Can you please also correct the unnecessary and partially incorrect overwriting of the headers in this branch? Thanks :+1: 

----

update: you already cleaned up the header while I had written the lines above :-)

return42 (approved):
Sorry, totally forgot this PR .. thanks for cleaning the HTTP header.

PR #5408, opened 2025-11-03
Add comprehensive macOS installation guides

krishna-medapati:
This PR adds a comprehensive macOS-specific installation guide to improve documentation for macOS users.

## What's Included
- Complete installation guide for macOS (both Docker and native methods)
- Specific instructions for Apple Silicon (M1/M2/M3) and Intel Macs
- macOS-specific troubleshooting section
- Prerequisites and dependency installation using Homebrew
- Common issues and solutions (port conflicts, permissions, Docker setup)
- File system and volume mounting guidance specific to macOS

## Motivation
Based on issue #5286 and general community feedback, macOS users often encounter platform-specific challenges when installing SearXNG. This documentation aims to reduce support requests and improve the onboarding experience for macOS users.

## Testing Environment
- **OS**: macOS Sequoia 15.7.1 (Build 24G231)
- **Architecture**: Apple Silicon
- **Python**: 3.13.5
- **Git**: 2.50.1

## Type of Change
- [x] Documentation improvement
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist
- [x] Documentation follows reStructuredText (.rst) format
- [x] Content is clear and helpful for macOS users
- [x] Tested commands on actual macOS environment
- [x] No breaking changes to existing documentation
- [x] Ready for review

## Additional Notes
This is my first contribution to SearXNG. Feedback and suggestions for improvement are welcome!

Related to issue #5286

### DIFF
--- a/docs/admin/installation-macos.rst
+++ b/docs/admin/installation-macos.rst
@@ -1,0 +1,281 @@
+.. _installation macos:
+
+=========================
+Installation on macOS
+=========================
+
+This guide covers SearXNG installation specifically for macOS systems, including both Apple Silicon (M1/M2/M3) and Intel-based Macs.
+
+.. contents::
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+Prerequisites
+=============
+
+System Requirements
+-------------------
+
+* macOS 11 (Big Sur) or later
+* macOS 15.7.1 (Sequoia) tested and confirmed working
+* At least 4GB of RAM
+* 2GB of free disk space
+
+Required Software
+-----------------
+
+Homebrew Package Manager
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you don't have Homebrew installed::
+
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+After installation, follow the instructions to add Homebrew to your PATH.
+
+Docker Desktop for Mac
+~~~~~~~~~~~~~~~~~~~~~~
+
+Install Docker Desktop::
+
+    brew install --cask docker
+
+Or download directly from: https://www.docker.com/products/docker-desktop
+
+**Important for Apple Silicon users**: Ensure you download the Apple Silicon version of Docker Desktop.
+
+After installation:
+
+1. Launch Docker Desktop from Applications
+2. Wait for Docker to start (the whale icon in the menu bar should be steady)
+3. Verify Docker is running::
+
+    docker --version
+    docker compose version
+
+Installation Methods
+====================
+
+Method 1: Docker Installation (Recommended)
+--------------------------------------------
+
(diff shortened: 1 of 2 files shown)


docs/admin/installation-macos.rst
```diff
@@ -0,0 +1,281 @@
...
+   :local:
+   :backlinks: entry
+
+Prerequisites
+=============
+
+System Requirements
+-------------------
+
+* macOS 11 (Big Sur) or later
```
Bnyro, line 20: Any specific reason why it doesn't run on older macOS versions?

docs/admin/installation-macos.rst
```diff
@@ -0,0 +1,281 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+::
+
+    docker compose up -d
+
+Step 4: Access Your Instance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Open your browser and go to: http://localhost:8080
```
Bnyro, line 93: This is already documented at https://github.com/searxng/searxng-docker, perhaps you should add a reference to the documentation there instead of re-documenting all this here (because the process is exactly the same).

Bnyro:
There's a lot of overlap with the already existing installation documentation, so it would make more sense to add additional hints there what the alternative way to run that command on macOS would be (e.g. "If you're on macOS, instead of running `apt install docker`, run `brew install --cask docker`.

The "Slow Performance on Apple Silicon" can perhaps be added to the end of the existing installation documentation, but apart from that the steps mentioned here are almost identical to what we already have in our documentation.

Opinion @return42 ?

return42:
I sorted out my last Mac OS systems many years ago, but I suspect:

- Apple shipped the 2007 version of Bash v3 and cannot upgrade to Bash 4 due to GPL3 restrictions.
- Same with services, Apple uses launchd (systemd was written for linux)

TBH, the BSD in macOS was adapted (messed up) for macOS and is now hopelessly outdated. As a result, a whole range of tools and methods have been developed to try to compensate for this shortcoming (e.g. to get up-to-date bash its recommended to install brew).

Native installation:

Basically, it's like Windows: our development environments and “native” deployment are designed for (modern) Linux systems. We don't have the technical capabilities or human resources to address the quirks of different operating systems.

Docker-installation:

I'll leave it up to @inetol to decide what he wants to include in the Docker documentation.  Personally, however, I don't think it's necessary to specifically mention the correct architecture when choosing a container... or is it?

inetol:
> ... he wants to include in the Docker documentation

We already have the docker-compose repo docs. *In the future we will move everything here, but I still have to arrange a couple of things.*

> ... I don't think it's necessary to specifically mention the correct architecture when choosing a container... or is it?

We target for amd64 and arm64 images, so shouldn't be any issues on any system if you can properly install Docker engine on it.

PR #5410, opened 2025-11-03
[fix] qwant engine: order query parameters to prevent 403 forbidden

Aadniz:
## What does this PR do?

This PR fixes an issue where Qwant engine will return 403 forbidden if the query parameters is in an *unusual* order. Since the engine suddenly returned 403 again (after https://github.com/searxng/searxng/pull/5382 fix), I found that it had better success being in the same order as seen requested from www.qwant.com.

I have a feeling we are playing whack-a-mole with Qwant a bit. If this continues, we might want to consider creating a more robust solution that more resemble the actual request.

Qwant does not seem to work on any SearXNG instance right now, and this is a fix for this issue.

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -134,17 +134,17 @@ def request(query, params):
 
     elif qwant_categ == 'images':
 
+        args['count'] = 50
         args['locale'] = q_locale
         args['safesearch'] = params['safesearch']
-        args['count'] = 50
         args['tgp'] = 3
         args['offset'] = (params['pageno'] - 1) * args['count']
 
     else:  # web, news, videos
 
+        args['count'] = 10
         args['locale'] = q_locale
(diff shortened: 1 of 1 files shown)


inetol (approved):

return42:
> This PR fixes an issue where Qwant engine will return 403 forbidden if the query parameters is in an _unusual_ order.

The order does not have a decisive effect. It may be that some requests work in the tests when the order is changed, but this is due to other reasons and does not address the problem. I have tested it myself, and even after the CAPTCHA has been activated, the SearXNG instance keeps running “access denied.”

Qwant has a [DataDome](https://datadome.co/products/datadome-captcha/) protection.

For more details read:

- https://github.com/searxng/searxng/issues/3929#issuecomment-2831487414

Aadniz:
Alright I see, thanks for getting me up to speed with this.

PR #5419, opened 2025-11-06
[mod] yandex engine: add supported languages

Aadniz:
## What does this PR do?

This PR adds support for Yandex's supported languages; Russian, English, Belarusian, French, German, Indonesian, Kazakh, Tatar, Turkish and Ukrainian.

Yandex allows to select multiple languages at once. I'm not sure if we should choose nothing or everything if `all` is chosen  in SearXNG preferences. The results are a tiny bit different selecting all languages compared to selecting no language but mostly the same.

It would be nice to have matching language when searching in German, French, Russian etc..

### DIFF
--- a/searx/engines/yandex.py
+++ b/searx/engines/yandex.py
@@ -28,6 +28,20 @@ search_type = ""
 base_url_web = 'https://yandex.com/search/site/'
 base_url_images = 'https://yandex.com/images/search'
 
+# Supported languages
+yandex_supported_langs = [
+    "ru",  # Russian
+    "en",  # English
+    "be",  # Belarusian
+    "fr",  # French
+    "de",  # German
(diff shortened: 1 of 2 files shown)


return42 (approved):
FYI: I placed d1842ec21 on top of your branch.

> Yandex allows to select multiple languages at once.

SearXNG has a very complex handling mechanism when it comes to language and region selection; SearXNG always attempts to find a suitable engine language for the user's selection. Since not all engines support all languages ​​and regions, approximation methods are used for this purpose.

This concept assumes that only one language or one region can be selected. If more languages ​​were available, a completely new language handling concept would be needed, and its implementation would affect all components of SearXNG (both server and client).

Short answer: We do not plan to select multiple languages at once

PR #5436, opened 2025-11-12
[feat] engines: add devicons engine

Bnyro:
- official website: https://devicon.dev/
- the engine contains a lot of icons of popular software frameworks (e.g. pytest),
so they could for example be useful for visualizing a diagram of the tech stack used in an app

### DIFF
--- a/searx/engines/devicons.py
+++ b/searx/engines/devicons.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Devicons (icons)"""
+
+import typing as t
+
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from extended_types import SXNG_Response
+    from search.processors.online import OnlineParams
+
+
+about = {
+    "website": "https://devicon.dev/",
+    "wikidata_id": None,
+    "official_api_documentation": None,
+    "use_official_api": True,
+    "results": "JSON",
+}
+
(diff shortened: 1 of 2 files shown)


searx/engines/devicons.py
```diff
@@ -0,0 +1,63 @@
...
+                    return True
+
+        return False
+
+    filtered_results = filter(is_result_match, resp.json())
+    for result in filtered_results:
+        for image_type in result["versions"]["svg"]:
+            img_src = f"{cdn_base_url}/icons/{result['name']}/{result['name']}-{image_type}.svg"
+            res.append(
+                res.types.LegacyResult(
```
return42: `EngineResults` is a inheritance of list and has therefore a `append` method, but we should use the dedicated method `res.add` .. 

https://github.com/searxng/searxng/blob/7a1b959646c45a81d3495148b1fa6c2da585eb59/searx/result_types/__init__.py#L65-L67

Bnyro: Ah, I see. Wouldn't it make sense to add some deprecation warning to `EngineResults.append` then? I.e. 

```python
 def add(self, result: Result | LegacyResult): 
     """Add a :py:`Result` item to the result list.""" 
     super().append(result) 

 def append(self, value: Any):
    logger.debug("DeprecationWarning: Calling EngineResults.append directly is deprecated, please use EngineResults.add")
    self.add(result)
```

PR #4974, opened 2025-07-04
[feat] cache: implement redis cache as alternative to sqlite

Bnyro:

PR #5441, opened 2025-11-13
[fix] ahmia engine: requires rotating tokens to work

Aadniz:
## What does this PR do?

Ahmia recently implemented a 60 minute rotating token system when searching. If these tokens are not provided, it will redirect to the homepage, and SearXNG returning a parsing error.
The tokens update every minute, but Ahmia allow up to 1 hour old tokens to be used.

This implementation gets the keys from the hidden input using xpaths, and uses the `EngineCache` to prevent having to do 2 requests on every search. It will however search 2 times for the first initial request towards Ahmia, and also if tokens are older than 1 hour.

I took inspiration from [semantic scholar](/searxng/searxng/blob/7a1b959646c45a81d3495148b1fa6c2da585eb59/searx/engines/semantic_scholar.py) engine to set up the cache.

As of now, SearXNG will return a parsing error when trying to search Ahmia because it is returning a redirect to the homepage.

There is also very few tor-specific search engines on SearXNG, and feel it is important that this works.

1. Clone this PR
2. Set up tor socks proxy (usually it's on port 9050)
3. Use the following settings in settings.yml:
```yaml
categories_as_tabs:
  general:
  images:
  videos:
  news:
  map:
  music:
  it:
  science:
  files:
  social media:
  onions:
  
engines:
  - name: ahmia
    using_tor_proxy: true
    proxies:
      all://:
        - socks5h://127.0.0.1:9050
```
4. Start SearXNG
5. Search for `!ah test` for example

### DIFF
--- a/searx/engines/ahmia.py
+++ b/searx/engines/ahmia.py
@@ -3,9 +3,14 @@
 Ahmia (Onions)
 """
 
+import typing as t
+
 from urllib.parse import urlencode, urlparse, parse_qs
 from lxml.html import fromstring
+from searx.utils import gen_useragent, ElementType
 from searx.engines.xpath import extract_url, extract_text, eval_xpath_list, eval_xpath
+from searx.network import get
+from searx.enginelib import EngineCache
 
 # about
 about = {
@@ -23,6 +28,7 @@ paging = True
 page_size = 10
 
 # search url
+base_url = 'http://juhanurmihxlp77nkq76byazcldy2hlmovfu2epvl5ankdibsot4csyd.onion'
 search_url = 'http://juhanurmihxlp77nkq76byazcldy2hlmovfu2epvl5ankdibsot4csyd.onion/search/?{query}'
 time_range_support = True
 time_range_dict = {'day': 1, 'week': 7, 'month': 30}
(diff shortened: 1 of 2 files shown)


searx/engines/ahmia.py
```diff
@@ -34,10 +40,44 @@
 content_xpath = './/p[1]'
 correction_xpath = '//*[@id="didYouMean"]//a'
 number_of_results_xpath = '//*[@id="totalResults"]'
+name_token_xpath = '//form[@id="searchForm"]/input[@type="hidden"]/@name'
+value_token_xpath = '//form[@id="searchForm"]/input[@type="hidden"]/@value'
+
+CACHE: EngineCache
+"""Persistent (SQLite) key/value cache that deletes its values after ``expire``
```
Bnyro: I was attempting to offer the option to use Redis instead (https://github.com/searxng/searxng/pull/4974), so for the engine's matter it should only know that it's persistent. Apart from that, the `EngineCache` docstring already explains that comment, so it's redundant.
```suggestion
```

Bnyro (commented):
I'm always running into a `ReplyError` of the proxy after the timeout period of SearXNG is finished (even set it to 10 seconds), so probably something wrong on my configuration. (I simply ran `tor` and used the address it told me to use :shrug:)

PS: Not related to your changes, but in case of a proxy error it crashes with
```
  File "/home/username/Projects/searxng/searx/network/client.py", line 111, in handle_async_request
    raise httpx.ProxyError("ProxyError: " + e.args[0], request=request) from e
                           ~~~~~~~~~~~~~~~^~~~~~~~~~~
TypeError: can only concatenate str (not "ReplyError") to str
```
instead of actually displaying the proxy error

Aadniz:
I want to test a bit more before coming back to this. It seems to be very slow, often just times out and it sometimes gives the proxy error as you mentioned that can be hard to reproduce sometimes.

Aadniz:
> I'm always running into a `ReplyError` of the proxy after the timeout period of SearXNG is finished (even set it to 10 seconds), so probably something wrong on my configuration. (I simply ran `tor` and used the address it told me to use 🤷)

@Bnyro It could be that it differs from distro to distro. You can test if the tor proxy is working by running `curl --proxy socks5h://127.0.0.1:9050 ipaddr.me`, or confirm it here by reading through the HTML output: `curl --proxy socks5h://127.0.0.1:9050 https://check.torproject.org/`. On Debian, the tor socks config is located at `/etc/tor/torsocks.conf`

Aadniz:
Although it is incredible slow, I cannot think of a better solution. The nature of the tor network is *just* slow afterall, and having to spend two requests sometimes does not make it better.

I believe those who are interested in running tor-related engines on their SearXNG instances, are willing to accept how slow it is. I recommend setting up the timeout for Ahmia to 20 seconds.

searx/engines/ahmia.py
Aadniz: Accidentally removed 1 line too little. Now, everything should be working.

Bnyro (approved):
Thank you! 

PS: Sorry for the delay, I have been a bit busy recently.

PR #5449, opened 2025-11-18
[mod] docs: add acknowledgements section

inetol:
Moves all images into a `assets/` folder.

---

All companies listed under "Sponsors" provides non-economic support to the organization, such as infrastructure and tooling.


return42:
> All companies listed under "Sponsors" provides non-economic support to the organization, such as infrastructure and tooling.

:thinking: are these "sponsors"?  .. Yes, we use their infrastructure and services, but then we would also have to list others such as GitHub or our providers, hosts, etc. .. I'm asking myself "where do we draw the line?"

Sponsors, in my opinion, are only those actors who actively offer us their support without expecting anything in return. I expect a sponsor to share the values of the community and actively promote them. 

The providers listed here make their services available to the general public free of charge; they do not share the goals of these projects.  And when we use their services, we usually pay with our data .. for example, at GitHub, we are just slaves who train their AI models :-)

inetol:
Let me rephrase the description of the PR. The "Sponsors" section includes companies with which we have explicitly treated to access their paid plans at no cost.

The three companies listed have provided us with services that we currently use and would otherwise have to pay for. For example, Docker through the OSS program, granted us with the Docker Team plan adding several exclusive features.

docs/index.rst
```diff
@@ -56,4 +56,33 @@ If you don't trust anyone, you can set up your own, see :ref:`installation`.
    utils/index
    src/index
 
+
+--------
+Sponsors
+--------
+
+The following companies are supporting the maintenance of SearXNG:
```
Bnyro:
```suggestion
--------
Credits
--------

For SearXNG's development, we are allowed to use the following services for free we would otherwise need to pay for :
```

docs/index.rst
```diff
@@ -56,4 +56,33 @@ If you don't trust anyone, you can set up your own, see :ref:`installation`.
...
+.. flat-table::
+   :widths: 1 1
+
+   * - .. image:: /assets/sponsors/docker.svg
+          :target: https://docker.com
+          :alt: Docker
+          :align: center
+          :height: 100 px
+
+     - .. image:: /assets/sponsors/tuta.svg
```
Bnyro, line 76: What's the matter with Tuta, don't they provide free emails anyways? Do we have a special plan there?

Bnyro:
In that manner, it might also make sense to mention Weblate - but in the end the list could be endless.

I agree with Markus that `Sponsors` is probably the wrong word though.

docs/index.rst
inetol: Only for individuals

inetol:
> In that manner, it might also make sense to mention Weblate - but in the end the list could be endless.

We have no relationship with Weblate, nor with Codeberg which is who actually selfhosts `translate.codeberg.org` free for everyone.

Bnyro (approved):

Issue #5438, opened 2025-11-13
Bug: presearch engine 403 forbidden


Aadniz:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

7a1b959646c45a81d3495148b1fa6c2da585eb59

**How did you install SearXNG?**

Following step-by-step installation guide: https://docs.searxng.org/admin/installation-searxng.html

**What happened?**

With the current version, Presearch is currently returning 403 forbidden with a cloudflare CAPTCHA.
This applies to presearch, presearch images, presearch videos and presearch news.

**How To Reproduce**

Start up a SearXNG instance, and search for `!ps test` for example.

**Screenshots & Logs**

Flask log:
```
/home/chiya/Dev/python/searxng/.venv/bin/python searx/webapp.py 
INFO    searx                         : max_request_timeout=None
INFO    searx                         : version: 2025.11.10+7a1b95964
DEBUG   asyncio                       : Using selector: EpollSelector
DEBUG   searx.data                    : init searx.data.ENGINE_DESCRIPTIONS
DEBUG   searx.data                    : init searx.data.USER_AGENTS
DEBUG   searx.data                    : init searx.data.WIKIDATA_UNITS
DEBUG   searx.data                    : init searx.data.EXTERNAL_URLS
DEBUG   searx.data                    : init searx.data.ENGINE_TRAITS
DEBUG   searx.data                    : init searx.data.EXTERNAL_BANGS
DEBUG   searx.webapp                  : static directory is /home/chiya/Dev/python/searxng/searx/static
DEBUG   searx.webapp                  : templates directory is /home/chiya/Dev/python/searxng/searx/templates
DEBUG   searx.data                    : init searx.data.LOCALES
/home/chiya/Dev/python/searxng/searx/valkeydb.py:41: DeprecationWarning: setting redis.url is deprecated, use valkey.url
  warnings.warn("setting redis.url is deprecated, use valkey.url", DeprecationWarning)
INFO    searx.valkeydb                : connecting to Valkey db=0 path='/run/redis/redis-server.sock'
ERROR   searx.valkeydb                : [chiya (1000)] can't connect valkey DB ...
Traceback (most recent call last):
  File "/home/chiya/Dev/python/searxng/.venv/lib/python3.13/site-packages/valkey/connection.py", line 319, in connect
    sock = self.retry.call_with_retry(
        lambda: self._connect(), lambda error: self.disconnect(error)
    )
[...]

PR #5460, opened 2025-11-20
[fix] engines - set hard timouts in *sub-request*

return42:
The requests changed here all run outside of the network context timeout, thereby preventing the engine's timeout from being applied (the engine's timeout can become longer than it was configured).

### DIFF
--- a/searx/engines/archlinux.py
+++ b/searx/engines/archlinux.py
@@ -120,7 +120,7 @@ def fetch_traits(engine_traits: EngineTraits):
         'zh': 'Special:搜索',
     }
 
-    resp = get('https://wiki.archlinux.org/')
+    resp = get('https://wiki.archlinux.org/', timeout=3)
     if not resp.ok:  # type: ignore
         print("ERROR: response from wiki.archlinux.org is not OK.")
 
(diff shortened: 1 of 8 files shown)


Bnyro (approved):
Maybe it would be nice to make the timeout here depend on the engines `timeout` value, e.g. `timeout/2` because that way instance admins can also fix timeout issues by changing the timeout in `settings.yml`, but I also see that it might not be the best idea.

return42:
> Maybe it would be nice to make the timeout here depend on the engines `timeout` value, e.g. ``

TBH I just wanted to solve a problem... actually, we would need to examine the network stack more closely .. [but I don't have the time for that at the moment](https://github.com/searxng/searxng/issues/5438#issuecomment-3559024562)

----

I am currently working on a POC that is supposed to make it possible for our engines to be developed better. In the course of my tests on this POC, I come across various small errors that I am trying to fix in this context (and with these PRs), without having to rewrite a component in SearXNG.

- https://github.com/searxng/searxng/pull/5459
- https://github.com/searxng/searxng/pull/5458
- https://github.com/searxng/searxng/pull/5460
- ..  and more coming soon.

PR #5468, opened 2025-11-22
[POC] feat: enhance browser fingerprinting with curl_cffi_proxy

DoiiarX:
## What does this PR do?

This PR implements a curl_cffi-based proxy endpoint to enhance browser fingerprinting and bypass anti-crawler detection mechanisms. The implementation follows a non-intrusive design pattern similar to the existing image proxy, allowing seamless integration without breaking existing functionality.

Key changes include:
- **New proxy module**: Added `searx/proxy/curl_cffi_proxy.py` that provides an HTTP proxy endpoint using curl_cffi library for advanced browser impersonation
- **Automatic URL conversion**: Modified `searx/search/processors/online.py` to automatically convert engine URLs to use the curl_cffi_proxy when enabled
- **Network client enhancements**: Updated `searx/network/client.py` to allow HTTP requests for local proxy endpoints while maintaining security
- **Cross-platform improvements**: Fixed Windows compatibility issues in multiprocessing (calculator plugin) and improved path handling using `pathlib.Path.as_posix()`
- **Dependency management**: Modernized the project with `uv` package manager, adding `pyproject.toml` and `uv.lock` for better dependency tracking
- **Code quality**: Translated all Chinese comments to English for better international collaboration

## Why is this change important?

1. **Anti-crawler bypass**: Many search engines and websites implement sophisticated anti-crawler detection. The curl_cffi library provides advanced browser fingerprinting capabilities that can bypass these restrictions more effectively than standard HTTP clients.

2. **Non-intrusive design**: By following the image proxy pattern, this implementation allows gradual adoption without requiring changes to existing engine code. Engines can opt-in to use the proxy when needed.

3. **Windows compatibility**: The fixes for Windows multiprocessing and path handling ensure the project works correctly across all major operating systems.

4. **Modern tooling**: Using `uv` for dependency management provides faster installation, better lock file management, and improved reproducibility.

5. **Maintainability**: Translating comments to English improves code accessibility for the international open-source community.

## How to test this PR locally?

1. **Install dependencies**:
   ```bash
   uv sync
   ```

2. **Enable curl_cffi_proxy** in `searx/settings.yml`:

   > To enable the curl_cffi Proxy, find or create the following section in your `searx/settings.yml` file under the `outgoing:` block.  
   > 
   > - You can adjust the `proxy` address to your local or remote proxy endpoint if needed.  
   > - If websites you are accessing are slow to respond, consider increasing the `extra_proxy_timeout` value to add extra seconds to request timeouts.

   ```yaml
[...]

PR #5476, opened 2025-11-25
[POC] network: add impersonate parameter

return42:
## [enh] network: add impersonate parameter (#4801)

This patch requires httpx-curl-cffi [1] and curl-cffi [2] and adds a new network
parameter named: ``impersonate``.  See ``BrowserTypeLiteral`` in [3] for valid
browser types.

This patch is based on PR [PR-4801] and replaces [PR-5468]

[1] https://github.com/vgavro/httpx-curl-cffi
[2] https://github.com/lexiforest/curl_cffi
[3] https://github.com/lexiforest/curl_cffi/blob/main/curl_cffi/requests/impersonate.py
[PR-4801] https://github.com/searxng/searxng/pull/4801
[PR-5468] https://github.com/searxng/searxng/pull/5468

## [POC] enable impersonate 'chrome' for the mojeek engines

HINT: So far, no evidence has been provided that the impersonate mode represents an improvement for the Mojeeks engines!

If an impersonate mode is chosen, then curl_cffi sets the headers that the corresponding browser would also set.

Sometimes, however, these headers don't match; for example, the "Accept" header is set by curl_cffi if it hasn't already been set by the application.  Here at the Mojeek engine, it is noticeable that this header has not been set so far. When it is set, however, it must be set to `Accept: */*` when searching for images.

Conclusion:

1. Whether the impersonate mode should be used must be evaluated individually for each engine.

2. Before the impersonate mode is activated, proof should be provided that it represents an improvement for the respective engine.

------

There are a lot of open questions / things we have to test:

* [x] check [HTTP-307 Temporary Redirect](https://developer.mozilla.org/docs/Web/HTTP/Reference/Status/307)
* [x] check `enable_http2: true` works 
* [x] check `enable_http2: false` works 
* [x] check `verify: false` works
* [ ] check `verify: "/path/to/ca.pem"` works
* [ ] check `local_addresses` works
* [ ] check HTTP proxy
* [ ] check HTTPS proxy
* [ ] check SOCKS proxy

---
List of engines that require this patch:
- Searchzee
- possibly Exa (#6409)
- possibly Brave (unconfirmed)
- Deutsche Bahn (https://www.bahn.de/) if we should ever want to add that
- all engines that currently have `enable_http2` set to `False`
- all engines that use Cloudflare bot protection


jzakirov:
Any chance to merge that? Looks pretty solid improvement

return42:
> Any chance to merge that? Looks pretty solid improvement

@jzakirov *thanks for the flowers* :-) ... but we're still a long way from that -> in the intro I added a few tests we still need to do. Besides the technical issues we still need to test, we also have no experience with *impersonate* (which is in my opinion the main showstopper for now) ... 

`curl-impersonate` sets various HTTP headers; I already mentioned the `Accept: */*` header above, just to give one example. There are engines in which we explicitly set different headers / conversely, in all engines we do not set headers that curl-impersonate sets -> We have no experience with how this will affect the engines.

Furthermore, we have to assume that the SearXNG operators will try to achieve better results using `network.impersonate`. If someone reports an error, we won't know if `network.impersonate` was set (to what value?). Setting this option will fundamentally changes the communication, and it will be difficult to perform an error analysis without this information.

So we're not just dealing with technical problems; "impersonate" also raises fundamental questions ...

To summarize, we're not just dealing with technical problems; "impersonate" also raises fundamental questions about a reliable bug tracking and development process.

The latter is one of the main reasons why I have hesitated to merge this patch to this day.

searx/network/client.py
```diff
@@ -192,12 +224,26 @@ def new_client(
...
-    transport = get_transport(verify, enable_http2, local_address, None, limit, retries)
+    if impersonate:
+        logger.debug("transport layer for this client is impersonate: %s", impersonate)
+        transport = httpx_curl_cffi.AsyncCurlTransport(
+            impersonate=impersonate,
+            default_headers=True,
+            debug=get_setting("general.debug"),
+            curl_options=curl_options,
+            http_version=httpx_curl_cffi.CurlHttpVersion.V3 if enable_http2 else httpx_curl_cffi.CurlHttpVersion.V1_1,
+            local_address=local_address,
```
Bnyro, line 235:
```suggestion
            local_address=local_address,
            verify=verify,
```
I tested that and it works. (also needs to be added above).

https://badssl.com is a good site for testing that.

Bnyro:
`enable_http2` and `verify` work.

> check verify: "/path/to/cert.pem" works
>
I can't get that one working.

Bnyro:
Enabling/disabling redirects also works.

Bnyro:
The `searchzee` engine now also requires the impersonate mode:
- copying the curl request from the browser devtools doesn't work (gives "too many requests")
- `./curl_firefox133 "https://searchzee.com/app.js"` works

So it'd be great if we could continue work on this - in my opinion we could even make impersonate the default. Yes, few engines might break, but there are many more engines that would benefit from it by having lower rate limits or starting to work at all.

shameez-struggles-to-commit:
Production data point for this POC, in case it helps the evaluation:

We've been running the #4801-style backport (`impersonate: chrome` via httpx-curl-cffi) on a small self-hosted instance since 2026-08-13, applied to the `duckduckgo`, `duckduckgo web`, `brave`, and `startpage` engines (default network for everything else).

Before: DDG/Brave/Qwant would land in `SearxEngineCaptcha`/429 within minutes of concurrent use, and the default `suspended_times` then parked them for hours.
After: nine days of steady use with no fingerprint-based CAPTCHAs on the impersonated engines; the suspensions we still see are ordinary rate limiting that clears with the standard backoff. Non-impersonated engines are unaffected.

One more observation that matches @Bnyro's searchzee note: Brave now appears to be in the same "browser TLS fingerprint required" camp — plain httpx gets challenged quickly even with a browser UA, while the impersonated fingerprint is left alone.

Happy to share the engine→network mapping or suspension stats if useful.

return42:
Superseded by:

- https://github.com/searxng/searxng/pull/6620

PR #5503, opened 2025-12-01
[feat] engines: add lucide icons

Bnyro:
According to https://scribe.rip/design-bootcamp/8-free-icon-libraries-that-are-not-fontawesome-or-material-275cccb3752c, Lucide is one of the largest sources for icons that can be used for developing apps (similar to flaticon).


return42 (approved):
LGTM 👍

Issue #5029, opened 2025-07-19
AI picture filter


cy8aer:
**Is your feature request related to a problem? Please describe.**

Today's images searches are not able to find or filter AI created images. As described in https://techcrunch.com/2025/07/18/duckduckgo-now-lets-you-hide-ai-generated-images-in-search-results/ Duckduckgo now has a filter for AI generated images.

**Describe the solution you'd like**

A button in image searches for showing/not showing AI generated images. Duckduckgo has a button like this now too.

**Describe alternatives you've considered**

Alternatives would be _not_ having a filter of course. This is an additional feature.

**Additional context**

As I understand the filter itself is buiit on the uBlock Huge AI blocklist. So the origin of an image is tested against this filter list: https://github.com/laylavish/uBlockOrigin-HUGE-AI-Blocklist and others

ghost:
Might be a duplicate of: https://github.com/searxng/searxng/issues/1587

PR #5515, opened 2025-12-04
[feat] pixiv engine: add filter for AI generated images

Austin-Olacsi:
## What does this PR do?

Adds an option to filter AI generated pictures from pixiv search results.

Some people may prefer to have AI generated pictures removed.
I might try to add similar filters to other image search engines when possible.

Search with and without the filter. Some images will be removed.

## Related issues

related: https://github.com/searxng/searxng/issues/5029

### DIFF
--- a/searx/engines/pixiv.py
+++ b/searx/engines/pixiv.py
@@ -17,6 +17,7 @@ about = {
 # Engine configuration
 paging = True
 categories = ['images']
+remove_ai_images = False
 
 # Search URL
 base_url = "https://www.pixiv.net/ajax/search/illustrations"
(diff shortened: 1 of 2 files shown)


searx/engines/pixiv.py
```diff
@@ -17,6 +17,7 @@
 # Engine configuration
 paging = True
 categories = ['images']
+filter_ai_images = ""
```
Bnyro:
```suggestion
filter_ai_images = False
```

Bnyro (approved):
Please add some documentation to the `filter_ai_images` attribute: 
- Does setting it to `True` only return AI-generated images or does it remove all AI-generated results (and thus only shows non-AI results)?

Austin-Olacsi:
Setting to true will remove the AI images, thus showing only non-AI results.

To make that more clear, i changed the name to `remove_ai_images`.

Bnyro (approved):

Bnyro:
Thank you 👍

PR #5534, opened 2025-12-05
[mod] themes: calculator unit support

inetol:
*Unit converter* plugin no longer exists, it's now part of *Calculator*.

Screenshot

You can also perform operations with units:

Screenshot

---

Unknown plugins in `settings.yml` will softfail and not register.


return42:
I'm not sure if it's a good idea to move the unit converter to the JS client. I see two disadvantages:

1. No-JS clients will lose the unit converter function
2. Compared to the (hard coded) unit system of mathjs, SearxNG's wikidata units are generic and [updated regularly from wikidata](https://github.com/searxng/searxng/commits/master/searx/data/wikidata_units.json) / and thanks to wikidata, SearXNG has a more unit definitions.


----


Definition of Units (and the SI and non SI conversion factors) in MathsJS:

https://github.com/josdejong/mathjs/blame/4bbe862bb553bd765685b46d977e8ee226e4abf0/src/type/unit/Unit.js#L1649-L1657

Definitions of units (and the SI conversion factors) in SearXNG / generic generated from wikidata units .. 

https://github.com/searxng/searxng/blob/ddc6d68114011c3cb7cc2cd778f48ac89a0a37b2/searx/data/wikidata_units.json#L1-L6

inetol:
> 1. No-JS clients will lose the unit converter function

Unit conversion isn't a main feature on SearXNG, we can delegate this to be a client only thing.

> 2. Compared to the (hard coded) unit system of mathjs, SearxNG's wikidata units are generic and [updated regularly from wikidata](https://github.com/searxng/searxng/commits/master/searx/data/wikidata_units.json) / and thanks to wikidata, SearXNG has a more unit definitions.

It's broken because I can't convert anything other than distance units. Even then, now we can do expressions using these units (e.g., *"1.3GiB * 53min"*), and if needed, [add new units based on other expressions](https://github.com/searxng/searxng/pull/5534/files#diff-c876fe6a3d03efa28b27a296b0e735783d0589d0c65152b8dfdbf4642537e8daR88-R92).

return42:
> It's broken because I can't convert anything other than distance units.
> Even then, now we can do expressions using these units (e.g., _"1.3GiB * 53min"_),

Sorry for being slow in uptake, not sure I fully understand you .. TBH I'm confused:

How is unit conversion related to a multiplication (or division) of units, like _"1.3GiB * 53min"_  = "GiB min"? .. not related, but personally I know ``Bytes/time`` (GiB/s) what counts a performance value .. what isn't a request to convert units(!)

>  and if needed, [add new units based on other expressions](https://github.com/searxng/searxng/pull/5534/files#diff-c876fe6a3d03efa28b27a296b0e735783d0589d0c65152b8dfdbf4642537e8daR88-R92).

The `Unit` implementation of mathjs allows artihmetic operations on `Unit` objects (complex numbers?) by any other unit .. for mathematical operation on units, mathjs has implemented its own system of (physical and non physical) units such as meters or, for example, mass ..

- https://github.com/josdejong/mathjs/blame/4bbe862bb553bd765685b46d977e8ee226e4abf0/src/type/unit/Unit.js#L1558

- https://mathjs.org/docs/datatypes/units.html

On the oposite, the unit conversion in SearXNG is based on the [International System of Units (aka SI)](https://en.wikipedia.org/wiki/International_System_of_Units).

In MathJS we have (for example) a unit (type) `BASE_UNITS.SURFACE` for `m²` .. in the SI System this is *derived* unit.

The SI knows about (the) [seven base units](https://en.wikipedia.org/wiki/International_System_of_Units#SI_base_units) .. apart from these seven base quantities, there are no other known physical quantities, which is why these are sufficient to perform conversions between the different unit systems (e.g. 20 °C to °F = 68 °F).

> Unit conversion isn't a main feature on SearXNG, 

Yeah, thats true .. SearXNG main feature are search results .. and the unit-converter is just a plugin, but ..

> we can delegate this to be a client only thing.

.. isn't true , at least not for our no-JS clients.

As a graduate physicist, I prefer the SI system because it allows all physical unit of measure to be converted (with the help of Wikidata).

MathJS could be a plus if we want to handle other units of measure (non physical) like `byte` .. but that would be an addition to the existing solution (the server plugin based on SI).

.. or do I miss something?

inetol:
> How is unit conversion related to a multiplication (or division) of units, like _"1.3GiB * 53min"_ = "GiB min"? .. not related, but personally I know `Bytes/time` (GiB/s) what counts a performance value .. what isn't a request to convert units(!)

I wrote wrongly. I meant to say, for example *"(300 Mb/s + 980 kilobit/sec) to GB/hour"*, Previously, it was not possible to perform math and conversions of this type.

> As a graduate physicist, I prefer the SI system because it allows all physical unit of measure to be converted (with the help of Wikidata).
> 
> MathJS could be a plus if we want to handle other units of measure (non physical) like `byte` .. but that would be an addition to the existing solution (the server plugin based on SI).
> 
> .. or do I miss something?

I can't tell much as I don't study this, but comparing units such as "acres" between the [Wikidata db](https://github.com/searxng/searxng/blob/ddc6d68114011c3cb7cc2cd778f48ac89a0a37b2/searx/data/wikidata_units.json#L7262) and the math.js [`UNITS`](https://github.com/josdejong/mathjs/blame/4bbe862bb553bd765685b46d977e8ee226e4abf0/src/type/unit/Unit.js#L1836) object, they use the same factor (`value` property is the factor in math.js) with usually less precision in the latter.

return42:
> they use the same factor

If they used other factors, the results would be wrong ;-) . It is inherent in the nature of the task that the two solutions are similar. We generate our unit database from Wikidata, which receives corrections and additions every week. The database from MathJS was initially created 10 years ago and since then has been little corrected or supplemented. Accordingly, the Wikidata database is much more comprehensive / it contains far more symbols / respectively designations for physical quantities -> more conversion factors.

Here is just one example of a traditional unit `ha` converted to its (derived) SI unit `m²`

    10ha to m²

-----

>  `(300 Mb/s + 980 kilobit/sec) to GB/hour`

This is far more than a conversion .. To be precise, it isn't a unit conversion (its a operation with complex numbers / `unit` implementation seems to me similar to complex numbers): `to GB/hour` only converts the prefix from `s` to `h` and the prefix `M` to `G`.  Explanation: compared to my example from above, we can take this example:

    1 km² to m²

In this example the unit is `m` (SI unit for *lenght*) the request just wants to replace the prefix `k` --> its not a unit conversion, like shown above in the `ha` example.

But of course you're right, MathJS `Unit`s provide a useful **additional** value if we can perform such complex operations on the client / MathJS can provide us with enhancements, but it *can't replace* the existing (server-side Wikidata) solution / from my point, the arguments have not been refuted so far:

> 1. No-JS clients will lose the unit converter function
> 2. Compared to the (hard coded) unit system of mathjs, SearxNG's wikidata units are generic and [updated regularly from wikidata](https://github.com/searxng/searxng/commits/master/searx/data/wikidata_units.json) / and thanks to wikidata, SearXNG has a more unit definitions.

PR #5540, opened 2025-12-06
[fix] themes: clear search input

### LINKED ISSUE #5539: Delete search history by clicking the "x"
The search history cannot be deleted by clicking the "x" after a search. This worked up to version 2025.12.1-ab8224c93.

Image

inetol:
Closes https://github.com/searxng/searxng/issues/5539


Bnyro (approved):
Tested, can confirm this fix works.

Btw, didn't we have `cursor: pointer` set on the `x` (clear button) previously? It no longer seems to show the right cursor that indicates that the `x` is clickable.

inetol:
> Btw, didn't we have `cursor: pointer` set on the `x` (clear button) previously? It no longer seems to show the right cursor that indicates that the `x` is clickable.

Never had one I think

PR #5542, opened 2025-12-08
Update Marginalia Search integration to use the new API

vlofgren:
## What does this PR do?

The code changes the marginalia search integration to use the newest version of the API,
which has better best practices such as not encoding the query as a path element.

The old API is deprecated.  It will continue to be available, but not receive new features.

I am the maintainer and operator of Marginalia Search and would like to see API consumers use the v2 API instead of the deprecated V1 API.  

During testing I also noticed the old API integration incorrectly escapes the queries using quote_plus, which was not understood by the marginalia backend.  A %20-coding would have worked, but migrating to the new API accomplishes the same.

I'm attaching some screenshots showing this behavior, which is mitigated with the PR (though %20-encoding the query would have also fixed this issue as seen in the 2nd screenshot, which I included mostly to show what the expected search results are for the query).

badquery

goodquery

Edit settings.yml and add an API key to the marginalia block,
enable the search engine and verify results are returned.

## Author's checklist

I have ran the code locally and verified it works.

### DIFF
--- a/searx/engines/marginalia.py
+++ b/searx/engines/marginalia.py
@@ -28,7 +28,7 @@ Implementations
 """
 
 import typing as t
-from urllib.parse import urlencode, quote_plus
+from urllib.parse import urlencode
 from searx.utils import searxng_useragent
 from searx.result_types import EngineResults
 from searx.extended_types import SXNG_Response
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thank you for upstreaming this to SearXNG!

I have to admit that the new Marginalia website looks really awesome, great job on the website + API rework :)

Also thank you for working on and providing an open source search index, this really looks like a great project will allow more people to ditch the big players (Google, Bing, ...) and instead get some more qualitative results instead of the current large amount of AI results most engines present to you.

PR #5559, opened 2025-12-12
[fix] searx.network: fix string concatenation of proxy error message

Aadniz:
## What does this PR do?

This PR fixes an issue where it will fail to concatenate a string when trying to raise a `ReplyError`, resulting in the engine(s) getting a "unexpected crash", instead of the correct "proxy error".

It is an important change, because it otherwise hides useful debug information we would like to see. If a person experiences an issue that is hard to reproduce, the current debug logs will only slow us down.

## Author's checklist

@Bnyro - If you would like to try this fix.

## Related issues

During testing of Ahmia onion engine fix (https://github.com/searxng/searxng/pull/5441), this issue was sometimes experienced. It is a bit hard to reproduce.


Bnyro (approved):
Thank you!

Issue #5562, opened 2025-12-13
Custom "content path" for Mediawiki engine


Seres67:
**Is your feature request related to a problem? Please describe.**

Mediawiki instances not using the default `/wiki/` path don't work in the current state.
Examples:
https://stardewvalleywiki.com/Sprinkler
https://wiki.warframe.com/w/Nova

**Describe the solution you'd like**

Add support for a custom "content path" (what i call it)  

I have implemented this on my fork but I don't know the codebase enough to reflect this across the docs and other places. I also don't know Python very much.
https://github.com/searxng/searxng/compare/master...Seres67:searxng:master

gioleppe:
@Seres67 if you open a pull request, we can check it and work on it together! 
I don't think much documentation is needed except for maybe a comment in the settings.yml file, and some clarification in the mediawiki engine docs page!

return42:
Close due to inactivity

----

> https://stardewvalleywiki.com/Sprinkler
> https://wiki.warframe.com/w/Nova

AFAIK these wikis use an (very) old wikimedia implementation which does not support the [Action_API](https://www.mediawiki.org/wiki/API:Action_API) / for more details read https://docs.searxng.org/dev/engines/mediawiki.html#mediawiki-engine

PR #5573, opened 2025-12-17
Custom "content path" for MediaWiki engine 

Seres67:
## What does this PR do?

This PR allows for custom content paths in URLs in the MediaWiki engine. Some wikis do not use the default and searxng generates incorrect links.

I'm running my fork locally and it seems to be working fine, though more tests should be done.
I am also far from proficient in Python and did the first thing that came to my mind.

Related to #5562

### DIFF
--- a/searx/engines/mediawiki.py
+++ b/searx/engines/mediawiki.py
@@ -157,7 +157,7 @@ def response(resp):
         timestamp = result.get('timestamp')
 
         url = (
-            f"{base_url.rstrip('/')}".format(language=resp.search_params['language']) + content_path.rstrip('/') + '/' + quote(title.replace(' ', '_').encode())
+            base_url.format(language=resp.search_params['language']) + content_path.strip('/') + '/' + quote(title.replace(' ', '_').encode())
         )
         if sectiontitle:
             # in case of sectiontitle create a link to the section in the wiki page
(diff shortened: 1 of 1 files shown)


searx/engines/mediawiki.py
```diff
@@ -152,7 +157,10 @@ def response(resp):
         timestamp = result.get('timestamp')
 
         url = (
-            base_url.format(language=resp.search_params['language']) + 'wiki/' + quote(title.replace(' ', '_').encode())
+            base_url.format(language=resp.search_params['language'])
+            + content_path.rstrip('/')
+            + ('/' if content_path.rstrip('/') else '')
+            + quote(title.replace(' ', '_').encode())
```
Bnyro, line 164: Why are you first removing the ending `/` and afterwards adding it again? Probably some typo? In case this is meant to change `wiki////` to `wiki/`, we shouldn't handle it this way because either that's just a configuration error or it could be intended and is the actual path.

Seres67: I don't know python so idk how you'd do this but it is to accept for example `wiki` and `wiki/`

Bnyro: Then you should do + `/` without the if condition

Seres67: I thought about that while typing my comment :D I'm trying that and pushing if it works

Seres67: Ah right I remember why I did that! You end up with https://stardewvalleywiki.com//Sprinkler instead of https://stardewvalleywiki.com/Sprinkler.

return42:
related https://github.com/searxng/searxng/issues/5562#issuecomment-4235418171

Issue #5571, opened 2025-12-16
Search API broken


gsantner:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
For example 2025.12.13+d720fe41c  for https://kantan.cat/

**How did you install SearXNG?**
I'm a user of SearXNG, not a administrator.

**What happened?**
The documentation tells about a search API: <https://docs.searxng.org/dev/search_api.html>. 
According to documentation `<SERVICE>/search?q=hello&format=json` is supposed to work, correct me if I'm wrong.
But it doesn't.
I tried to use it now on many hosts listed at <https://searx.space/>, but it works at none.

Is the usage wrong, or the API broken?

**How To Reproduce**
Try various hosts from <https://searx.space/>, and see that the API nowhere works.

**Expected behavior**
That the API returns search results and in the given format, as documented.

gioleppe:
@gsantner if you see a message such as this one

<img width="1264" height="212" alt="Image" src="https://github.com/user-attachments/assets/f810b5ee-d3c0-4173-a830-ce108d719842" />

It's probably just that they did not enable the json format.
You can easily do so in a self hosted instance! check out the format section here: https://docs.searxng.org/admin/settings/settings_search.html#settings-search 

This is an example of what I get by running searxng locally and by adding the json format in settings.yaml
<img width="839" height="864" alt="Image" src="https://github.com/user-attachments/assets/9347b570-b023-4820-9845-cce46e0d8f84" />

@return42  @Bnyro if @gsantner confirms he's seeing a forbidden message I'd say it's safe to close this issue

gsantner:
Hello,

as every instance has it disabled, I assume the "Search API feature" is disabled by default.
I would suggest to note that in documentation, so users don't expect it to be available.

gioleppe:
Thanks for your feedback @gsantner, I agree that the documentation is somewhat obscure on this.
I opened a PR aiming to solve this, please let me know if you think the phrasing is better now!
If you find other unclear points in the docs I encourage you to open a PR yourself so that everyone benefits from it.

@return42 let me know if you think the new version of the docs is better 😄

PR #5574, opened 2025-12-17
Search API Docs: usage and format support explanation

gioleppe:
## What does this PR do?

This PR aims to explain better how to configure and use the Search HTTP API.

Some users complained about the docs being a little bit obscure, especially with respect to supported response formats.
I agree that the search api docs page could be better phrased, so I decided to rewrite it a little to make it more explicit.

make docs.live and check the Search API docs page

## Related issues

Inspired by #5571

### DIFF
--- a/docs/dev/search_api.rst
+++ b/docs/dev/search_api.rst
@@ -4,14 +4,33 @@
 Search API
 ==========
 
-The search supports both ``GET`` and ``POST``.
+SearXNG supports querying via a simple HTTP API.
+Two endpoints, ``/`` and ``/search``, are supported for both GET and POST methods.
+The GET method expects parameters as URL query parameters, while the POST method expects parameters as form data.
 
-Furthermore, two endpoints ``/`` and ``/search`` are available for querying.
+If you want to consume the results as JSON, CSV, or RSS, you need to set the
+``format`` parameter accordingly. Supported formats are defined in settings.yml, under the ``search`` section.
+Requesting an unset format will return a 403 Forbidden error. Be aware that many public instances have these formats disabled. 
+You can still get HTML-formatted results and parse them yourself.
 
+
+Endpoints:
 
 ``GET /``
-
 ``GET /search``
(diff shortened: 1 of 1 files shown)


docs/dev/search_api.rst
```diff
@@ -4,15 +4,34 @@
...
 ==========
 
-The search supports both ``GET`` and ``POST``.
+SearXNG supports querying via a simple HTTP API.
+Two endpoints, ``/`` and ``/search``, are supported for both GET and POST methods.
+The GET method expects parameters as URL query parameters, while the POST method expects parameters as form data.
 
-Furthermore, two endpoints ``/`` and ``/search`` are available for querying.
+If you want to consume the results as JSON, CSV, or RSS, you need to set the
+``format`` parameter accordingly. Supported formats are defined in settings.yml, under the ``search`` section.
```
Bnyro:
```suggestion
``format`` parameter accordingly. Supported formats are defined in ``settings.yml``, under the ``search`` section.
```
nit

docs/dev/search_api.rst
```diff
@@ -4,15 +4,34 @@
...
+Requesting an unset format will return a 403 Forbidden error. Be aware that many public instances have these formats disabled. 
+You can still get HTML-formatted results and parse them yourself.
```
Bnyro: I wouldn't advertise scraping SearXNG instances in our documentation.

If instance admins don't want to enable the API, they probably also don't want other to scrape results from their instances.

gioleppe: Yeah I agree. Let's say that if you are reading SearXNG's docs about the search API  you're probably smart enough to figure this out by yourself, but I'll cut the last part nonetheless :smile:

gioleppe:
@Bnyro added the backticks, removed the "problematic" phrase in the docs as you suggested :wink:

Bnyro (approved):
Thanks!

PR #5575, opened 2025-12-18
[mod] replace '&bull;' in #suggestions list with a CSS styled one

Elvyria:
## What does this PR do?

This commit removes the "dot" prefix value from the `#suggestions` list's entries and adds it back via CSS instead.  
Was doing some styling, and the chosen approach was a bit crazy.

It makes styling easier and less hacky for those who are interested, removing the dot with just styles is currently not as easy as I would like it to be.

Also the old dot is vertically not centered, maybe a font thing.

## Author's checklist

Before / After

image
image


inetol:
There are a lot of weird things about how HTML/CSS is laid out. The *bullet* is a literal character, so depending on the font the browser chooses to render it, you might see it one way or another (or any at all).

The proper solution would be to build the suggestions wrapper as an unordered list instead of a *list* of forms.

Elvyria:
What would be the advantage of doing it with `<ul>` instead of a 6 line fix?
I can do that, but I don't think messing out the existing layout for the sake of a dot styling is worth it.

Bnyro:
> What would be the advantage of doing it with `<ul>` instead of a 6 line fix? I can do that, but I don't think messing out the existing layout for the sake of a dot styling is worth it.
>
Probably better accessibility because screen readers know better what's going on there.

Your current solution is more of a CSS hack that wouldn't be needed if we were using the correct HTML instead :)

Elvyria:
Does this one look more suitable for this project ?

client/simple/src/less/style.less
```diff
@@ -571,22 +571,16 @@ article[data-vim-selected].category-social {
...
-    flex-flow: column;
-    justify-content: flex-end;
-
-    form {
-      display: inline-block;
-      flex: 1 1 50%;
-    }
+    padding-left: 1em;
+    margin: 0;
   }
```
Bnyro, line 585: I'd also make the bullets blue, i.e.
```css
.wrapper {
  ...
  li::marker {
    color: var(--color-result-link-font);
  }
}
```

Apart from that lgtm :+1:

Bnyro: Oh, and we should also remove the left padding from the list with
```css
.wrapper {
  padding-left: 0;
  list-style-position: inside;
}
```

inetol (approved):

PR #5601, opened 2025-12-26
feat(favicon): add favicon_url in json format

archever:
## What does this PR do?
add favicon_url field in json format

get favicon_url info from json outout

set favicon resolver and call search api with json format

## Related issues

### DIFF
--- a/searx/webutils.py
+++ b/searx/webutils.py
@@ -22,6 +22,7 @@ from flask_babel import gettext, format_date  # type: ignore
 from searx import logger, get_setting
 
 from searx.engines import DEFAULT_CATEGORY
+from searx import favicons
 
 if TYPE_CHECKING:
     from searx.enginelib import Engine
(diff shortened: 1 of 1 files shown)


searx/webutils.py
```diff
@@ -164,13 +165,18 @@ def get_json_response(sq: "SearchQuery", rc: "ResultContainer") -> str:
...
         'corrections': list(rc.corrections),
         'infoboxes': rc.infoboxes,
         'suggestions': list(rc.suggestions),
         'unresponsive_engines': get_translated_errors(rc.unresponsive_engines),
     }
+    for res in rc.get_ordered_results():
+        r = res.as_dict()
+        if res.parsed_url:
+            r['favicon_url'] = favicons.favicon_url(res.parsed_url.netloc)
+        data['results'].append(r)
```
Bnyro: Please put this before `data` is created instead of modifying it afterwards, i.e.

```py
results = rc.get_ordered_results()
for res in results:
    res = res.as_dict()
    if res.parsed_url:
        r['favicon_url'] = favicons.favicon_url(res.parsed_url.netloc)

...
data = {
    ...,
    "results": results,
    ...
}
```

archever: done

return42:
FYI favicons-proxy had been implemented for the HTTP/HTML client:To reduce the number of (HTTP) roundtrips a favicon URL most often is a [data-URL](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data).

https://github.com/searxng/searxng/blob/a5c946a321a6b5edeacc789b673b04456e45530b/searx/favicons/proxy.py#L208-L212

https://github.com/searxng/searxng/blob/a5c946a321a6b5edeacc789b673b04456e45530b/searx/favicons/proxy.py#L232

PR #5615, opened 2025-12-29
[fix] yep: fix 403 forbidden errors

Bnyro:
Apparently, yep has been broken for a while. Measures to fix it:
- only use HTTP/1.1, because our HTTP2 client gets fingerprinted and blocked
- send the `Origin` HTTP header


inetol (approved):
*I haven't seen any issues with not including Origin*

Bnyro:
> _I haven't seen any issues with not including Origin_
>
Strange, I thought it was required when I tested it. Anyways it probably doesn't hurt to keep it.

Thanks for the review :+1:

Issue #2295, opened 2023-03-30
Google: add a way to disable the thumbnails for the results


unixfox:
**Is your feature request related to a problem? Please describe.**
Google engine was recently changed, and it does now show thumbnails for some results:
![image](https://user-images.githubusercontent.com/4016501/228871310-abcd5a53-b215-4988-8b05-19944309d6e6.png)

I preferred when there were no thumbnails taking "useless" space.

**Describe the solution you'd like**
An ability to disable the thumbnails in the results given by Google engine.

**Describe alternatives you've considered**
Does this ability to disable thumbnails already exist somewhere? I haven't found it.

**Additional context**

return42:
We can add an option to disable .. should not be a big task.

But then thumbnails will be disabled for all google results / see my  https://github.com/searxng/searxng/issues/2287#issuecomment-1486882220 --> google has two kinds of results

One for topics like celebrity ... (which has a thumb for every result item)

![grafik](https://user-images.githubusercontent.com/554536/228877516-2949fe01-f9b1-4f25-bb1c-6b9c9cbcbd63.png)

One for common search terms .. (which has most often non thumbs in the result list / only a favicon is shown) ..

![grafik](https://user-images.githubusercontent.com/554536/228878133-035f1212-1050-4aa3-a1c4-70eb82418a43.png)

------

OT:  ...

![grafik](https://user-images.githubusercontent.com/554536/228879247-9f1cacd8-3f0a-4b74-baaa-e63c0641b50f.png)


don't use "all" language --> most SearXNG engines will fallback to "en" or do not give a language setting in the request and the origin search engine will choose a locale based on the location of the server IP.

I always recommend to use a language (fr) or a locale (fe-BE).  Alternatively use `:auto`.

dalf:
> An ability to disable the thumbnails in the results given by Google engine.

What if there are displayed on the right side ?

dalf:
What if thumbnails are displayed on right side (for LTR languages):

![image](https://user-images.githubusercontent.com/1594191/229342703-0efc59b0-a126-4ef4-8851-63624788eb04.png)

unixfox:
Would prefer to have no thumbnails at all, for having as many results as possible on the screen.

return42:
> Would prefer to have no thumbnails at all, for having as many results as possible on the screen.

Should we change the title:

> Google: add a way to disable the thumbnails for the results

[...]

PR #5617, opened 2025-12-30
[feat] add plugin which allows admins to hide thumbnails for search results

artooro:
## What does this PR do?

This PR adds a new official plugin that allows searxng instance administrators to easily disable thumbnails in search results.
It was done as a plugin instead of a theme option for simplicity.
The way it works is setting the `thumbnail` result attribute to None for all results.

I have a scenario where there is a desire for a search engine with no images. This is possible with stock searxng except for the thumbnails in the search results.
I've also noticed others requesting this as a feature, so instead of being an external plugin, it seems appropriate as an official plugin

1) Set the value of active to true in settings.yml (it is set to false by default of course)
2) Search for something likely to return a thumbnail, such as the word "chocolate".

```
  searx.plugins.hide_thumbnails.SXNGPlugin:
    active: true
```

## Related issues

Issue https://github.com/searxng/searxng/issues/2295 is closed as "not planned" but I'm hoping this contribution will be welcome.

### DIFF
--- a/searx/plugins/hide_thumbnails.py
+++ b/searx/plugins/hide_thumbnails.py
@@ -1,0 +1,39 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+"""
+This plugin removes the 'thumbnail' property from search results if the 'hide_thumbnails' setting is enabled.
+"""
+
+import typing
+from flask_babel import gettext
+
+from searx.plugins import Plugin, PluginInfo
+from searx.result_types import Result
+
+if typing.TYPE_CHECKING:
+    from searx.search import SearchWithPlugins
+    from searx.extended_types import SXNG_Request
+    from searx.result_types import Result
+    from searx.plugins import PluginCfg
+
+
+class SXNGPlugin(Plugin):
+    id: str = "hide_thumbnails"
+
+    def __init__(self, plg_cfg: "PluginCfg") -> None:
+        super().__init__(plg_cfg)
+        self.info = PluginInfo(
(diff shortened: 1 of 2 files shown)


return42:
> It was done as a plugin instead of a theme option for simplicity.

When implemented as a plugin, the SearXNG server continues to load the favicons from the network and consumes resources.  Only at the end, when all results are compiled, are the favicons removed from the results.

If no favicons are needed, none should be obtained from the network.

> Issue https://github.com/searxng/searxng/issues/2295 is closed as "not planned"

"not planed", since this feature request wants to disable favicons for a single engine (google) 

> but I'm hoping this contribution will be welcome.

Yes, contrib is welcome: but why are you implement a Plugin when you can simply switch off the feature [search.favicon_resolver](https://docs.searxng.org/admin/settings/settings_search.html)

artooro:
@return42 I just looked at the `favicon_resolver` setting and it's off by default (I never turned it on) and also doesn't appear to be related to thumbnails. I'm pretty sure it has nothing to do with this PR.

The thumbnail code doesn't seem to involve any fetching, just some parsing, for example the brave engine uses this:

```
thumbnail: str = eval_xpath_getindex(result, ".//a[contains(@class, 'thumbnail')]//img/@src", 0, default="")
```

If you would prefer that I work on a setting to make the skipping of thumbnails happen for each engine instead of being stripped later in the process, it would be more work but I'm willing to do it. My thought process was that it would be preferred to keep the engines simple and strip it as a post processing step.

PR #5644, opened 2026-01-08
[fix] engine: switch to using GSA for iPhone useragent

### LINKED ISSUE #5286: Bug: google engine

themg95:
## What does this PR do?

Switched to using the GSA for iPhone user agent and changed the parsing accordingly

fixes google not working 

query the google engine

## Related issues

Closes #5286


themg95:
To explain more, I was experimenting with the NID cookie with random user agents. Then, I noticed that sometimes the search request would return results, even without the NID cookie. After looking into it more and running some tests, I realized that this was only happening with the Google Search App on an iPhone. I think google might've forgotten to add the NID check there.

Also, I've tested the rate limits of this method, and after 3k identical requests, I only encountered a captcha once, but the subsequent requests went through, so I assume that there's no rate limit, but very, very rarely, there might be a random captcha. Unfortunately though, in my testing, sometimes (not frequently enough to be a big issue but enough to be a minor annoyance) the request times out.

inetol:
This will not be a permanent solution, and I doubt it will take them long to patch it now that it has been disclosed, but it will give us more time to build a proper way to handle that requirement.

themg95:
Yeah, that's what I thought as well but this is the best we got for the time being.

searx/data/__init__.py
```diff
@@ -105,3 +105,16 @@ def ahmia_blacklist_loader() -> list[str]:
...
+def gsa_useragents_loader() -> list[str]:
+    """Load data from `gsa_useragents.txt` and return a list of user agents
+    suitable for Google.  The user agents are fetched by::
+
+      searxng_extra/update/update_gsa_useragents.py
+
+    This function is used by :py:mod:`searx.engines.google`.
+
+    """
+    with open(data_dir / 'gsa_useragents.txt', encoding='utf-8') as f:
```
Bnyro, line 126: nit: Instead of loading this data every time when `gsa_useragents_loader` is called (i.e. each time a user makes a request), please use the `lazy_globals` and `data_json_files` as above. 

This way, the user agents are loaded into memory the first time `gsa_useragents_loader` is called and afterwards the memory-cached value will be returned instead of loading them from the disk file again (which might in some cases be significantly slower if there are many concurrent requests).

searx/engines/google.py
```diff
@@ -262,6 +265,7 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     # HTTP headers
 
     ret_val['headers']['Accept'] = '*/*'
+    ret_val['headers']['User-Agent'] = random.choice(useragents)
```
Bnyro:
```suggestion
    ret_val['headers']['User-Agent'] = utils.gen_gsa_useragent()
```
Please move the user agent selection to `searx/utils.py`, as we already have all our other user-agent related methods there: 
https://github.com/searxng/searxng/blob/master/searx/utils.py#L84

searx/data/__init__.py
themg95: hmm ok, i just copied the `ahmia_blacklist` code which doesn't do that. 

anyway, i tried making it fetch the file and assign it as a global variable in `google.py`, so i assume it'll only run once? or do engines get re-imported on every request?

Bnyro:
Thanks a lot for the research and the PR @themg95 !

searx/data/__init__.py
Bnyro: The answer doesn't seem to be relevant anymore because you already found a solution yourself as it seems :)

I'll answer it anyways:
- Each engine is initialized once per thread (e.g. if you run SearXNG with 4 threads, each engine is created 4 times)
- So in this case, loading it in the `init` method of the engine would also have been fine. But the now-implemented solution seems cleaner anyways.

Bnyro (approved):

tiekoetter (approved):
Tested on my public instance

searx/engines/google.py
```diff
@@ -262,6 +262,7 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     # HTTP headers
 
     ret_val['headers']['Accept'] = '*/*'
+    ret_val['headers']['User-Agent'] = gen_gsa_useragent()
```
themg95, line 265: That's the whole point of this PR, though? To change the user agent to bypass the no-js restrictions

return42: Yeah .. my test was set up incorrectly .. sorry for the post!

yeyuchen198:
It seems not working today

return42:
The google "GSA for iPhone User-Agent" don't work any longer .. can you please have a look / thanks a lot.

PR #5674, opened 2026-01-19
[fix] google: decode urls before passing them onto results

### LINKED ISSUE #5673: Google search result URLs are always percent-encoded, messing up URL parameters
2026.1.16+2d9f213ca (4eb373b55)

**How did you ~~install~~ use SearXNG?**
Public instances at https://searx.tiekoetter.com/ and https://search.rhscz.eu/

**What happened?**

Search result URLs are percent-encoded when they should not be. This seems to only occur with the Google engine. This is mainly an issue with URLs that contain URL parameters (such as `?id=12345&foo=bar`, which gets encoded as `%3Fid%3D12345%26foo%3Dbar`)

**How To Reproduce**

1. Do one of the following searches:
    * https://searx.tiekoetter.com/search?q=!go+hacker+news+searxng
    * https://search.rhscz.eu/search?q=!go+hacker+news+searxng
2. Click on any result, e.g. the one titled “SearXNG is a free internet metasearch engine”.
3. Observe a 404 error with “Unknown.” displayed, because the URL is now https://news.ycombinator.com/item%3Fid%3D39948044 when it should be https://news.ycombinator.com/item?id=39948044.

**Expected behavior**

URLs aren’t percent-encoded when they shouldn’t be. The correct behavior can be seen with a different engine, e.g. https://search.rhscz.eu/search?q=!szn+news.ycombinator.com+searxng (news.ycombinator.com is third in my result list) (Bing, Yahoo and DDG all don’t work as testing engines for this.)

**Screenshots & Logs**
*None*

**Additional context**
Seems to be browser-independent, as it happens on both Firefox and Chromium.

themg95:
## What does this PR do?

Decodes the URLs from google

The previous commit that fixed google also broke some URLs that have query parameters, this PR fixes that.

Run the test queries in #5673

## Related issues
Fixes #5673

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -17,7 +17,7 @@ import re
 import random
 import string
 import time
-from urllib.parse import urlencode
+from urllib.parse import urlencode, unquote
 from lxml import html
 import babel
 import babel.core
@@ -373,7 +373,7 @@ def response(resp: "SXNG_Response"):
             if raw_url is None:
                 logger.debug('ignoring item from the result_xpath list: missing url of title "%s"', title)
                 continue
-            url = raw_url[7:].split('&sa=U')[0]  # remove the google redirector
+            url = unquote(raw_url[7:].split('&sa=U')[0])  # remove the google redirector
 
             content_nodes = eval_xpath(result, './/div[contains(@data-sncf, "1")]')
             for item in content_nodes:


inetol (approved):

Openstreetmapler:
~~**Regression: This might have re-triggered issue #5286!**~~

EDIT: Based on the actual commit, it doesn’t seem plausible anymore that this broke something.

PR #5675, opened 2026-01-20
[fix] iqiyi: update album_info key

hubutui:
subscriptionContent -> subscriptContent

## What does this PR do?

update album_info key for iqiyi

`subscriptionContent -> subscriptContent`, this is a minor change.


Bnyro (approved):
Thanks for the fix :+1:

PR #5689, opened 2026-01-25
[feat] add artstation engine

Austin-Olacsi:
## What does this PR do?

adds new engine for artstation

decent image results.

make run
!as bmw

### DIFF
--- a/searx/engines/artstation.py
+++ b/searx/engines/artstation.py
@@ -1,0 +1,105 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Artstation (images)"""
+
+import re
+import typing as t
+from json import dumps
+
+from searx.result_types import EngineResults
+from searx.network import post
+from searx.enginelib import EngineCache
+
+# Engine metadata
+about = {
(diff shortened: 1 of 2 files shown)


searx/engines/artstation.py
```diff
@@ -0,0 +1,80 @@
...
+# Engine configuration
+paging = True
+categories = ['images']
+results_per_page = 20
+
+# Search URL
+base_url = "https://www.artstation.com/api/v2/search/projects.json"
+
+
+def fetch_csrf_tokens():
```
Bnyro, line 44: Subsequent requests on Artstation can (and should) use the same CSRF token. That makes it harder to detect us as a bot, as it might be conspicuous if we request so many CSRF tokens from the same IP.

Hence, we should cache the token inside an `EngineCache`. See e.g. https://github.com/searxng/searxng/blob/eea189286388dc270e2bcd7a3b17cc4cd76d761e/searx/engines/360search.py#L40

Bnyro:
Thanks for implementing this engine, looks very interesting :+1:

Bnyro (approved):
Thank you!

PR #5713, opened 2026-01-31
[feat] add nist.gov vulnerability search

Austin-Olacsi:
## What does this PR do?

Adds nist.gov vulnerability search to the IT category.

It's a massive, constantly updated database of security vulnerabilities. 

make run
!nvd cross site scripting

### DIFF
--- a/searx/engines/nvd.py
+++ b/searx/engines/nvd.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""National Vulnerability Database (it)"""
+
+from urllib.parse import urlencode
+from datetime import datetime
+from searx.result_types import EngineResults
+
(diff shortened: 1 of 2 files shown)


searx/engines/nvd.py
```diff
@@ -0,0 +1,69 @@
...
+
+    query_params = {
+        "resultType": "records",
+        "keyword": query,
+        "rowCount": results_per_page,
+        "offset": start_index,
+    }
+
+    params["url"] = f"{base_url}?{urlencode(query_params)}"
+    params['headers']['content-type'] = "application/json"
```
Bnyro: This is not needed - it should only be set for POST/PATCH/... requests when the request body is JSON, but we use a GET request here.

Tested and it works fine without it.

PR #5728, opened 2026-02-09
feat: Implement Brave Api Search

csicar:
- use proper Brave API with api_key for search

## What does this PR do?

This PR add a new search engine called `braveapi`, which uses the brave api instead of the crawler

Brave runs into ratelimits a lot. This solves that problem.

Setup a `braveapi` engine locally with an `api_key`

Result from an example request:
image

### DIFF
--- a/searx/engines/braveapi.py
+++ b/searx/engines/braveapi.py
@@ -1,0 +1,126 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine to search using the Brave (WEB) Search API.
+
+.. _Brave Search API: https://api-dashboard.search.brave.com/documentation
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`api_key`
+
+Optional settings are:
+
+- :py:obj:`results_per_page`
+
+.. code:: yaml
+
+  - name: braveapi
+    engine: braveapi
+    api_key: 'YOUR-API-KEY'  # required
+    results_per_page: 20     # optional
+
+The API supports paging and time filters.
+"""
+
+import typing as t
+
+from urllib.parse import urlencode
+from dateutil import parser
+
+from searx.exceptions import SearxEngineAPIException
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://api.search.brave.com/",
+    "wikidata_id": None,
+    "official_api_documentation": "https://api-dashboard.search.brave.com/documentation",
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": "JSON",
+}
+
+api_key: str = ""
+"""API key for Brave Search API (required)."""
+
+categories = ["general", "web"]
+paging = True
+safesearch = True
+time_range_support = True
+
+results_per_page: int = 20
+"""Maximum number of results per page (default 20)."""
+
+base_url = "https://api.search.brave.com/res/v1/web/search"
+"""Base URL for the Brave Search API."""
+
+time_range_map = {"day": "past_day", "week": "past_week", "month": "past_month", "year": "past_year"}
+"""Mapping of SearXNG time ranges to Brave API time ranges."""
+
+
+def init(_):
+    """Initialize the engine."""
+    if not api_key:
+        raise SearxEngineAPIException("No API key provided")
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    """Create the API request."""
+    search_args: dict[str, str | int | None] = {
+        "q": query,
+        "count": results_per_page,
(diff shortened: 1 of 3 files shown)


searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+        return None
+
+def response(resp):
+    """Processes the API response"""
+    results = []
+    
+    if not resp.ok:
+        raise SearxEngineAPIException(f'HTTP error {resp.status_code}')
+        
+    data = json.loads(resp.text)
```
Bnyro:
```suggestion
    data = resp.json()
```

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+    
+    if api_key == "":
+        raise SearxEngineAPIException('No API key provided')
+    
+    search_args = {
+        'q': query,
+        'count': 20,
+    }
+    
+    # Seitennavigation
```
Bnyro: In english please.

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+
+def request(query, params):
+    """Creates the API request"""
+    
+    if api_key == "":
+        raise SearxEngineAPIException('No API key provided')
+    
+    search_args = {
+        'q': query,
+        'count': 20,
```
Bnyro: Please make this configurable via a global variable, e.g. `results_per_page = 20`, so that it can be configured via YAML.

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+    'week': 'past_week',
+    'month': 'past_month',
+    'year': 'past_year'
+}
+
+def request(query, params):
+    """Creates the API request"""
+    
+    if api_key == "":
+        raise SearxEngineAPIException('No API key provided')
```
Bnyro: Please put this check into the `init` method, see e.g. https://github.com/searxng/searxng/blob/b5bb27f231e5f24b3985cd7cbd3f371486c21a11/searx/engines/gitea.py#L80

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+    "website": "https://api.search.brave.com/",
+    "wikidata_id": "Q22906900",
+    "official_api_documentation": "https://api-dashboard.search.brave.com/documentation",
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": "JSON",
+}
+
+api_key = ''
+categories = ['general', 'news', 'videos', 'images']
```
Bnyro:
```suggestion
categories = ['general', 'web']
```
Your implementation currently doesn't support news, videos and images as far as I can tell?

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+paging = True
+safe_search = True
```
Bnyro:
```suggestion
safesearch = True
```

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+    if api_key == "":
+        raise SearxEngineAPIException('No API key provided')
+    
+    search_args = {
+        'q': query,
+        'count': 20,
+    }
+    
+    # Seitennavigation
+    if params.get('pageno'):
```
Bnyro: pageno is always set, so this if condition is always `true`. Please set pageno directly in `search_args` (line 66-69)

searx/engines/braveapi.py
```diff
@@ -0,0 +1,116 @@
...
+        return parser.parse(published_date_raw)
+    except parser.ParserError:
+        return None
+
+def response(resp):
+    """Processes the API response"""
+    results = []
+    
+    if not resp.ok:
+        raise SearxEngineAPIException(f'HTTP error {resp.status_code}')
```
Bnyro: This will never happen, SearXNG only calls the `response` method if the response is `ok`.

searx/engines/braveapi.py
Bnyro: This should not be an API exception, it should be a `ValueError`.

Bnyro:
Please format your code with `make format.python`, see https://docs.searxng.org/dev/makefile.html#make-format.

csicar:
Hi, @Bnyro!
Thanks for the early review! I went through your proposals and some other changes. Should be ready for review now :)

return42 (approved):
I force pushed the nits I had (manly type hints) / but I don't have an API key to test .. @csicar can you please make a functional test of my modifications / thanks a lot 👍

csicar:
Will do! 
Regarding testing: I think you might be able to get a free api key from brave for testing. I, at least, would think it would be worth it for them ;)

csicar:
Ok, I tested on 1dadb051f89455e8a494ab1a0ddbdcb78cca17dc. `braveapi` search works👍

Bnyro (approved):

PR #5759, opened 2026-02-20
 [feat] image results page: also display request statistics (i.e. engine response times)

Bnyro:
Apparently the engine statistics have been hidden from the images page when
infinite scroll was added as a preference in https://github.com/searxng/searxng/commit/56e34947a6368e6154064c52fa23d21ecda7ab4c

However, now that we also display engine statistics in the results page,
I don't see any reason why this should still be hidden from the images tab.

For example, this is very useful for seeing which engines work and which ones are broken.


return42:
TL;DR; I don't think there is a trivial solution as long the client UI is not capable to layout what the server defines (except @inetol knows more).

I tried to document what areas a client should have / see [result types](https://docs.searxng.org/dev/result_types/index.html)

grafik

----

> For example, this is very useful for seeing which engines work and which ones are broken.

Yeah, thats what I miss since a long time.

My *workaround* is to mix other (non image) results into the result-list, for example use:

    !ddg !images cookies

to get WEB-Results and Image-results in one result list.

> I don't see any reason why this should still be hidden from the images tab.

TBH this is not a "real" image-tab .. its just a category named "Images" .. whatever the selected category is, the SearXNG result-list can contain any of the defined [result types](https://docs.searxng.org/dev/result_types/index.html).

This circumstance was not taken into account in the previous development of the UI, which is why the UI looks a bit broken when we mix image results with other results (like from the query above / try it out and move/scroll/select to see how broken it is).

grafik

------

With the patch from this PR and a *pure* image-result-list the info bar is displayed, but at the top instead of on the side.

grafik

Bnyro:
Thanks for the detailed write-up :+1: 

> With the patch from this PR and a _pure_ image-result-list the info bar is displayed, but at the top instead of on the side.
> grafik
>
Yeah, it's definitely not perfect - but in my opinion it's better to display it at the top than not displaying it at all.

I'm also not sure if it would be that useful to display the response statistics at the side, because that would also reduce the width the images can take by a lot. I.e. if we showed it as a real sidebar, the images only would have about 80% of the screen width and to the right of it is a lot of blank space (except for at the top where the sidebar is, but the sidebar is very small, so most of the space below it remains unused...)

inetol (approved):
No functional issues, LGTM

return42 (approved):

PR #5760, opened 2026-02-20
[mod] engines: remove seekr engine

Bnyro:
Looks like https://www.seekr.com/ was sold to an other company
that does something with enterprise AI and no longer provides any
web search engine.


return42 (approved):
Don't forget to remove https://github.com/searxng/searxng/blob/master/docs/dev/engines/online/seekr.rst

Bnyro:
Good catch!

PR #5764, opened 2026-02-21
[feat] engines: add pexels engine

Bnyro:
## What does this PR do?
- this PR adds support for the Pexels image engine

Important:
This PR adds [cloudscraper](https://github.com/VeNoMouS/cloudscraper) as a new dependency. I'm not sure if it's worth adding it as dependency, but it seems to be the only way to bypass the Cloudflare page at <https://pexels.com>. 

I think there are actually some more engines that would benefit from this library.

The dependency probably won't be needed anymore if we switch to a browser impersonation library like `curl-impersonate`, but I'm not sure about the progress of https://github.com/searxng/searxng/pull/5476.

### DIFF
--- a/searx/engines/pexels.py
+++ b/searx/engines/pexels.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Pexels (images)"""
+
+import re
+
+from urllib.parse import urlencode
+from lxml import html
+
+import cloudscraper
+
+from searx.result_types import EngineResults
+from searx.utils import eval_xpath_list
+from searx.enginelib import EngineCache
+from searx.exceptions import SearxEngineAPIException
+from searx.network import get
+
+
+# about
+about = {
+    "website": 'https://www.pexels.com',
+    "wikidata_id": 'Q101240504',
+    "official_api_documentation": 'https://www.pexels.com/api/',
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": 'JSON',
+}
+
+base_url = 'https://www.pexels.com'
+categories = ['images']
+results_per_page = 20
+
+paging = True
+time_range_support = True
+time_range_map = {'day': 'last_24_hours', 'week': 'last_week', 'month': 'last_month', 'year': 'last_year'}
+
+SECRET_KEY_RE = re.compile('"secret-key":\b*"(.*?)"')
+SECRET_KEY_DB_KEY = "secret-key"
+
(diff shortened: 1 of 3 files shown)


searx/engines/pexels.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def init(engine_settings):
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+
+
+def _get_secret_key():
+    scraper = cloudscraper.create_scraper()
+    resp = scraper.get(base_url)
```
return42, line 51: > This PR adds [cloudscraper](https://github.com/VeNoMouS/cloudscraper) as a new dependency.

Didn't know before .. seems the right tool for us 🚀 .. however we should aware (have an eye on) that the content from such a HTTP (GET) request is executed in a JS interpreter on the SearXNG server ... AFAIK https://pypi.org/project/Js2Py/ is the default for executing JS code in the (cloud-)`scraper` instance.

return42:
> The dependency probably won't be needed anymore if we switch to a browser impersonation library like curl-impersonate, but I'm not sure about the progress of https://github.com/searxng/searxng/pull/5476.

Its not clear to me, why do you think curl-impersonate can help to solve a JS challenge?

Bnyro:
> Its not clear to me, why do you think curl-impersonate can help to solve a JS challenge?
>
I've done the same request to `https://pexels.com` with curl - and didn't get a cloudflare challenge. So they likely only show the challenge if the client request looks suspicious (probably TLS fingerprinting)

searx/engines/pexels.py
Bnyro: On a side note, we should avoid updating to v3 of this library:

See https://github.com/gentoo/gentoo/pull/43733#pullrequestreview-3217599441

return42 (approved):
LGTM 👍

searx/engines/pexels.py
return42: @Bnyro is cloudscraper really needed? .. in https://github.com/searxng/searxng/pull/5953 a Referer header was added to avoid CF's challenge.

Bnyro: Seems like Cloudscraper never really worked - it just set the right HTTP version or request headers to bypass botdetection, but didn't actually solve the challenge.

I've sent https://github.com/searxng/searxng/pull/5954 to remove Cloudscraper again.

PR #5766, opened 2026-02-22
Make a password generator (something similar to duckduckgo's password generator)

Anjo2807:
## What does this PR do?
Add a password generator answerer module

Convenience purposes

1. Run the dev container
2. type in randompass or randompassword (you can add a number argument to set the password length... maximum is 128)
3. you will see the answer
## Author's checklist
- [x] Get the password gen working

## Related issues
Its an addition

### DIFF
--- a/searx/answerers/randompass.py
+++ b/searx/answerers/randompass.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-module-docstring
+
+import random
+import string
+
+from flask_babel import gettext
+
+from searx.result_types import Answer
+from searx.result_types.answer import BaseAnswer
+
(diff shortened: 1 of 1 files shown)


inetol (changes requested):
This must be implemented client side.

You can extend https://github.com/searxng/searxng/blob/master/client/simple/src/js/Plugin.ts , see https://github.com/searxng/searxng/blob/master/client/simple/src/js/plugin/Calculator.ts as example.

Anjo2807:
> This must be implemented client side.
> 
> You can extend https://github.com/searxng/searxng/blob/master/client/simple/src/js/Plugin.ts , see https://github.com/searxng/searxng/blob/master/client/simple/src/js/plugin/Calculator.ts as example.

thanks ill add this

Issue #3407, opened 2024-04-19
Checker on Friday runs without a gain in knowledge


return42:
Checker has been added in ..

- https://github.com/searxng/searxng/pull/3262

but without any further work / without an evaluation this run is of no use ... 

> I remember there are some GH action that annotated the source code (?), we can use that too (in addition to the JSON output).

origin: https://github.com/searxng/searxng/pull/3262#discussion_r1510231406

> I see trillions of tracebacks in my tty / may we do not show the full traceback, just one line should be enough .. or we completely suppress these messages that are displayed when loading the images:

origin: https://github.com/searxng/searxng/pull/3262#pullrequestreview-1912919964

> The purpose is rather to have reminder of what thing that might be broken, then we can fix them. If it becomes also forgotten, then we can be remove the embedded checker... If it gets traction, then it might be a good thing for the QA.

origin: https://github.com/searxng/searxng/pull/3262#issuecomment-1975111916

---- 

In its current state the checker ends (every Friday) with an error, [trillions of traceback](https://github.com/searxng/searxng/actions/runs/8748199981/job/24007855291) but without a note what engines has been failed.  --> a report is needed.

dalf:
Would mind to help on https://github.com/searxng/searxng/pull/3312 ?

return42:
superseded by:

- https://github.com/searxng/searxng/pull/5767

PR #5767, opened 2026-02-22
[mod] drop SearXNG's checker

return42:
Motivation: There is no analysis for the checker that could be evaluated in any meaningful way.

- https://github.com/searxng/searxng/issues/3407
- https://github.com/searxng/searxng/pull/3312

The checker would need to be completely redesigned, but even then, its usefulness and the maintenance required for it would be disproportionate.

TBH: In its current form, it is useless and only consumes resources and causes the engines to be blocked, because these tests (query terms) come from *hundreds* of instances and could be interpreted as bot attacks.

Related issues: [search.checker](https://github.com/searxng/searxng/issues?q=label%3A%22search.checker%22)


Bnyro (approved):
I definitely agree with this decision.

Until I read the checker issue you linked, I didn't even know that the checker runs automatically in the background on every instance?! That causes quite a lot of unnecessary network traffic nonody benefits from, neither the instances nor the engines that get flooded with this traffic.

We already have searx.space to generate reliability stats, so that should suffice.

return42:
>  I didn't even know that the checker runs automatically in the background on every instance?!

I have to correct myself, scheduling was disabled by default in 73c86f9bf ..  5 years ago ..  I'm to long in this project :-)

PR #5769, opened 2026-02-24
fix: fix sogou engine results fetch with direct url

### LINKED ISSUE #4811: sogou engine sometimes does not return real URL
The URL returned by sogou sometimes starts with `https://www.sogou.com/link?url=`. eg: 
```bash
https://www.sogou.com/link?url=hedJjaC291Ok-E9WTygIKqZt6DizBOEjAwiXrfyunF09WCxkIl4-35JEUH4NNc7H
```
and the real url is:
```text
https://www.sohu.com/a/897047241_100106801
```
The url parameter seems to be encrypted, and it seems that there is no way to directly extract the real URL from it.

if you send a HTTP GET request the sogou url, you get this:
```html
<meta content="always" name="referrer"><script>window.location.replace("https://www.sohu.com/a/897047241_100106801")</script><noscript><META http-equiv="refresh" content="0;URL='https://www.sohu.com/a/897047241_100106801'"></noscript>
```

hezhijie0327:
## What does this PR do?

1. Refactor Sogou engine result fetch, use direct url instead of jump link `https://www.sogou.com/link?url=`
2. Support get publish date from results

|Before|After|
|-|-|
|image|image|

`!sogou test`

## Related issues

Closes #4811
Related #4827

### DIFF
--- a/searx/engines/sogou.py
+++ b/searx/engines/sogou.py
@@ -1,9 +1,12 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Sogou search engine for searxng"""
 
+import re
+from datetime import datetime
 from urllib.parse import urlencode
 from lxml import html
 
+from searx.exceptions import SearxEngineCaptchaException
 from searx.utils import extract_text
 
 # Metadata
@@ -21,7 +24,12 @@ categories = ["general"]
 paging = True
 time_range_support = True
 
-time_range_dict = {'day': 'inttime_day', 'week': 'inttime_week', 'month': 'inttime_month', 'year': 'inttime_year'}
+time_range_dict = {
+    "day": "inttime_day",
+    "week": "inttime_week",
+    "month": "inttime_month",
+    "year": "inttime_year",
+}
 
 # Base URL
 base_url = "https://www.sogou.com"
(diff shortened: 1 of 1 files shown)


searx/engines/sogou.py
```diff
@@ -33,35 +42,80 @@ def request(query, params):
...
+                    url = match.group(1)
+                else:
+                    url = f"{base_url}{url}"
+            cite_date = extract_text(item.xpath('.//span[@class="cite-date"]'))
+            if cite_date:
+                date_str = cite_date.strip().lstrip("-").strip()
+                try:
+                    publishedDate = datetime.strptime(date_str, "%Y-%m-%d")
+                except (ValueError, TypeError):
+                    publishedDate = None
```
Bnyro: Please refactor this into its own method instead of doing the almost the exact same stuff in the `if` and in the `else` branch.

(it probably also makes sense to move the creation of the result dictionary, i.e.
```py
              {
                    "title": title,
                    "url": url,
                    "content": content,
                    "publishedDate": publishedDate,
                }
            
```
to the new method as well

searx/engines/sogou.py
```diff
@@ -33,35 +42,80 @@ def request(query, params):
...
-            content = extract_text(item.xpath('.//div[contains(@class, "fz-mid space-txt")]'))
+    for item in dom.xpath(
+        '//div[contains(@class, "rb")] | //div[contains(@class, "vrwrap") and not(.//div[contains(@class, "special-wrap")])]'
+    ):
+        title = None
+        url = None
+        content = None
+        publishedDate = None
+
+        item_html = html.tostring(item, encoding="unicode")
```
Bnyro, line 68: Isn't unicode the default anyways?

searx/engines/sogou.py
```diff
@@ -33,35 +42,80 @@ def request(query, params):
...
     params["url"] = f"{base_url}/web?{urlencode(query_params)}"
     return params
 
 
 def response(resp):
+    if (
+        resp.status_code == 302
+        and resp.next_request is not None
+        and str(resp.next_request.url).startswith("http://www.sogou.com/antispider")
+    ):
```
Bnyro, line 58: I think you also have to set `params["follow_redirects"] = False`, otherwise this can never be true because SearXNG automatically follows the redirect instead of calling `response` with the redirect resp.

searx/engines/sogou.py
hezhijie0327: Without `encoding="unicode"`, `html.tostring()` returns bytes, but re.search() expects a string.

image

searx/engines/sogou.py
hezhijie0327: Done.

Bnyro:
Thanks 👍 

Please also run `make format` to fix the CI

hezhijie0327:
Done

Bnyro (approved):

PR #5772, opened 2026-02-26
[feat] fynd: add safe search support

mgcsysinfcat:
## What does this PR do?

Add safe search parameter to fynd search URL

safe search support

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -234,8 +234,9 @@ def request(query, params):
         time_range = time_range_url.format(time_range_val=time_range_val)
 
     safe_search = ''
-    if params['safesearch']:
-        safe_search = safe_search_map[params['safesearch']]
+    safe_search_val = params.get('safesearch')
+    if safe_search_val is not None:
+        safe_search = safe_search_map[safe_search_val]
 
     fargs = {
         'query': urlencode({'q': query})[2:],
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -901,7 +901,12 @@ engines:
 
   - name: fynd
     engine: xpath
-    search_url: https://fynd.bot/?search={query}&offset={pageno}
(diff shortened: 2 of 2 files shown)


Bnyro (changes requested):
Thanks for the PR :+1: 

However, this doesn't work yet, you have to change it to something like
```yml
    search_url: https://fynd.bot/?search={query}&offset={pageno}{safe_search}
    safesearch: true
    safe_search_map:
      0: '&safe=0'
      1: '&safe=1'
      2: '&safe=1'
```

See https://github.com/searxng/searxng/blob/master/searx/engines/xpath.py#L101

mgcsysinfcat:
> Thanks for the PR 👍
> 
> However, this doesn't work yet, you have to change it to something like
> 
> ```yaml
>     search_url: https://fynd.bot/?search={query}&offset={pageno}{safe_search}
>     safesearch: true
>     safe_search_map:
>       0: '&safe=0'
>       1: '&safe=1'
>       2: '&safe=1'
> ```
> 
> See https://github.com/searxng/searxng/blob/master/searx/engines/xpath.py#L101

Hi, Bnyro

I found a issue that prevents it from taking effect:
https://github.com/searxng/searxng/blob/c3e3d2d85de1d82ecb6bbde469c1961306c268f7/searx/engines/xpath.py#L237

When params['safesearch'] is 0, the condition evaluates to False, so the mapping is skipped.
If 0 is intended to be a valid safesearch value, this behavior is incorrect.

Bnyro:
You're right, this should be `params["safesearch"] is not None`.

Please just go ahead and also fix it in this PR here (it's only a very small fix, so it's fine to not create a new PR for it)

Bnyro:
Using `dicto = {...}` instead of `dicto = defaultdict(dict)` doesn't work here because when using `dicto["unset-key"]`, this throws a key error on normal dicts while it return a `{}` if using `defaultdict`.

Bnyro (approved):
Thanks 👍

PR #5781, opened 2026-02-28
[feat] botdetection: Add AbuseIPDB API integration for IP blocking

Serpensin:
## What does this PR do?

This PR adds a new bot detection module `abuseipdb` that checks incoming IP addresses against the AbuseIPDB API to detect and block malicious IPs.

### Features implemented:

1. **IP Checking**: Queries the AbuseIPDB `/check` API endpoint to get the abuse confidence score for each incoming IP
2. **IP Blocking**: Blocks IPs with `abuseConfidenceScore` >= configured threshold (default: 75)
3. **Caching**: Results are cached in Valkey with configurable TTL (default: 24 hours)
4. **Tor Support**: Optional skip for Tor exit nodes
5. **Privacy**: IP addresses are hashed (SHA-256, first 16 chars) before storing in cache and logging
6. **Fail-open**: If the API is unavailable (429, 500, connection errors), requests are allowed through
7. **Reporting**: Includes a `report_ip()` function to report malicious IPs to AbuseIPDB

### Files added/modified:

- `searx/botdetection/abuseipdb.py` - New module with IP checking logic
- `searx/compat.py` - Added environment variable support for all settings
- `searx/limiter.py` - Integrated abuseipdb into the filter chain
- `searx/limiter.toml` - Added configuration section
- `tests/unit/test_abuseipdb.py` - Comprehensive unit tests
- `docs/src/searx.botdetection.rst` - Documentation

The AbuseIPDB integration provides an additional layer of protection against malicious users:

1. **Proactive blocking**: Blocks known bad IPs before they can abuse the search engine
2. **Reduces load**: Prevents scrapers, attackers, and other malicious actors from consuming resources
3. **Community benefit**: Optionally reports malicious IPs back to AbuseIPDB to help the community
4. **GDPR compliant**: IP addresses are hashed before storage to protect user privacy

### Prerequisites:
- Running Valkey/Redis instance
- AbuseIPDB API key (free tier available at https://www.abuseipdb.com/account/api)

### Configuration:
Add to your `limiter.toml`:

```toml
[botdetection.abuseipdb]
enabled = true
api_key = "your_api_key_here"
confidence_threshold = 75
skip_tor = false
cache_time = 86400
```

Or use environment variables:
```bash
SEARXNG_ABUSEIPDB_ENABLED=true
SEARXNG_ABUSEIPDB_API_KEY=your_api_key
SEARXNG_ABUSEIPDB_CONFIDENCE_THRESHOLD=75
SEARXNG_ABUSEIPDB_SKIP_TOR=false
SEARXNG_ABUSEIPDB_CACHE_TIME=86400
```

Since tests are included, you can simply do a `make test`, to test everything together. But since these are only mock tests, you can either connect to a VPN, which will most likely have a high abuse score, connect to tor with skip_tor set to false, or set the confidence threshold to 0, which will block everything.

## Author's checklist

- [x] All tests pass (`make test`)
- [x] Code follows PEP 8 and project conventions
- [x] Documentation updated
- [x] Unit tests added for all new functions
- [x] Privacy: IP addresses are hashed before storage/logging
- [x] Fail-open behavior: API errors allow requests through
- [x] Configuration supports both TOML and environment variables

I added the ability to report IPs, so that a someone can add the ability. Or one could also include fail2ban, which supports reporting jail entries to AbuseIPDB automatically.


return42:
Thanks a lot for your PR and your work on this task, nevertheless I have some points:

- Send a request to a abuse DB for every single/new IP is not performant and will slow down the SearXNG server massively. 

- AbuseIPDB is just one source, we should implement a concept with the ability to add IPs from other DBs also.

- The concept should block IP-ranges .. block a subnet, not just a single IPs .. (please have IPv6 in mind).

----

A long time ago I implemented a *valkey-RangeTable* for fast lookup of IP ranges ..

- https://github.com/searxng/searxng/pull/2501

> The ZRANGEBYSCORE query scales with O(log(N)). To make it more clear: a query in a table with 1 million ranges takes just twice as long as the query in a table with 1000 ranges.
>
> A RangeTable is excellent for IP ranges and ultra-fast queries to see if an IP (which is nothing but an integer) is contained in one of the ranges of the table.

When we rebase and merge the PR-2501, we “only” need sources from which we obtain the IPs or subnets to be blocked and load them into this RangeTable.

When I developed this back then, there was no further interest in such a concept, which is why it fell into oblivion.

If there is interest in such a concept today, I can rebase the PR and we can start to implement adapters for AbusIPDB or any other other block lists.

However, this requires that the entire IP database (DB of subnets know for abuse) be available. I am uncertain whether it is possible to download the entire database with AbuseIPDB.

return42:
Please allow me to make the following comments:

The code in this PR gives me the impression that it was developed in large part by AI -> there are code passages that lie around unused, there are indirectives that are superfluous .. and concepts for fast lookup are missed.

It is perfectly acceptable to use AI as an assistant during development work. However, it becomes problematic when developers rely too heavily on AI and believe that AI can implement better solutions than they can themselves.

AI is a wonderful assistant, but developers still need to contribute the concepts .. and solutions for the specific challenges that arise during runtime.

Serpensin:
Yes, opencode was used pretty heavily. But I only use AI to write code, that I can actually read and understand in what it does, which is why I even decided to implement this at all.

But if this isn't that good, I get it. I have no idea, what actually goes into creating something that big. I mainly do some small discord bots here and there. The main reason, that I wrote this implementation, is because I was bored and wanted something to do. XD
And I must say, that it was pretty fun, working on it.

inetol:
Applies https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

Issue #4195, opened 2025-01-16
Auto-detecting Bulgarian as Russian


thenack0:
**How did you install SearXNG?**
I installed it using the TrueNAS SCALE catalog.

**What happened?**
Hello,

When the search language is set to "Auto-detect [auto]", about 70% of the time SearXNG incorrectly detects Bulgarian as Russian. This might be because the two languages share a lot of words in common.

If that can't easily be fixed, can we exclude Russian from the language detection list or deactivate it somehow?

Thanks in advance.

**How To Reproduce**
1. Set the search language to "Auto-detect [auto]".
2. Perform searches in Bulgarian.
3. Notice that the detected language is Russian approximately 70% of the time.

**Expected behavior**
The language detection should identify Bulgarian correctly when performing searches in Bulgarian.

**Screenshots & Logs**

**Additional context**
Is there a way to exclude Russian from the language detection list or deactivate it?

return42:
If you are a maintainer of a SearXNG instance you can set `default_lang` and `languages` / see https://github.com/searxng/searxng/issues/3604#issuecomment-2186484253

If you are a user, you can set the preferred language:

- in your WEB browser --> https://github.com/searxng/searxng/issues/3604#issuecomment-2186484253
- in the preferences in the cookies 

![Image](https://github.com/user-attachments/assets/c9088907-ac12-487d-9765-095833b8b67a)

- or using the search syntax `:bg foo` --> https://docs.searxng.org/user/search-syntax.html#select-language

----

Related: https://docs.searxng.org/src/searx.utils.html#searx.utils.detect_language

unixfox:
@return42 it's not solving the original problem though, which is that SearXNG categorize Bulgarian as Russian.

Think like SearXNG would categorize German as Dutch and give you Dutch results when you search with words written in German, how that would be annoying :)

return42:
> which is that SearXNG categorize Bulgarian as Russian.

Sure, the language model of fasttext is to small .. we can consider to use a more elaborated language model .. but this small model already eats [~56MB per worker](https://github.com/searxng/searxng/pull/1969#issuecomment-1345366676) .. 

I think we better use a (third-party language) service than loading large language models into worker's memory .. 

I reopened the issue and marked as "feature request".

thenack0:
> [@return42](https://github.com/return42) it's not solving the original problem though, which is that SearXNG categorize Bulgarian as Russian.
> 
[...]

PR #5795, opened 2026-03-02
[mod] drop fasttext-predict

### LINKED ISSUE #5790: lets drop  fasttext-predict
Motivation:

- I never had luck fasttext-predict (`detect_language`) returns the intended language .. the LLM is to small and query terms often contain proper names that make a correct language detection impossible.

- LLM of fasttext-predict consumes a lot of RAM without offering the user any added value.

----

https://github.com/searxng/searxng/blob/8fc549574fb30bba47ea6011e1475ceffbbc977b/searx/utils.py#L686-L687

- https://github.com/searxng/fasttext-predict / fasttext with wheels and no external dependency

vojkovic:
## What does this PR do?

Removes the `fasttext-predict` dependency and the language detection code.

If a user now selects `auto` for the search language, the detected language now
falls back directly to the `Accept-Language` header sent by the browser (which was already the fallback when fasttext returned no result).

- fasttext's [language detection is unreliable](https://github.com/searxng/searxng/issues/4195) for some languages, especially short search queries, and in particular for queries containing proper names which is a common case.
- `fasttext-predict` consumes [significant memory](https://github.com/searxng/searxng/pull/1969#issuecomment-1345366676) without offering users much real value.
- the upstream fasttext project was archived by Meta in 2024
- users already have two better alternatives: the `Accept-Language` header and the search-syntax language prefix (e.g. `:fr` or `:de`).

`make run`

Verify that the language dropdown 'auto' option still works correctly and solely uses the Accept-Language header.

## Related issues

Closes https://github.com/searxng/searxng/issues/5790

https://github.com/searxng/searxng/issues/4195

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,6 @@ httpx-socks[asyncio]==0.10.0
 sniffio==1.3.1
 valkey==6.1.1
 markdown-it-py==3.0.0
-fasttext-predict==0.9.2.4
 tomli==2.4.0; python_version < '3.11'
 msgspec==0.20.0
 typer==0.24.1
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -25,15 +25,10 @@ from lxml.etree import XPath, XPathError, XPathSyntaxError
 from lxml.etree import ElementBase, _Element  # pyright: ignore[reportPrivateUsage]
 
 from searx import settings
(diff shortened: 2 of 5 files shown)


Bnyro (approved):

inetol (approved):

return42 (approved):
Thanks a lot 👍

PR #5799, opened 2026-03-02
[mod] engines: remove searchcode engine

inetol:
Microslop fault https://web.archive.org/web/20260111060432/https://boyter.org/posts/searchcode-is-being-rebooted/

Related:

- https://github.com/searxng/searxng/pull/5170


return42 (approved):
searchcode.com has not been rebooted .. so sad :-(

PR #5800, opened 2026-03-02
[mod] engines: remove stract engine

inetol:
Engine probably dead, developer doesn't have time to maintain it anymore?


return42:
@mikkeldenker @Bnyro do you know the state of [stract.com](https://stract.com)?

Related:

- https://github.com/searxng/searxng/pull/3099

Bnyro (commented):
Please also remove https://github.com/searxng/searxng/blob/master/searx/autocomplete.py#L271, apart from that lgtm.

I don't know what's up with Stract, but as it seems to be unmaintained it makes sense to drop it.

return42 (approved):

Issue #5056, opened 2025-07-28
Unify developer tools


inetol:
Related https://github.com/searxng/searxng/pull/5018#discussion_r2235175331

Right now, we have diferent ways to install and manage tools..:

* ~asdf (https://github.com/searxng/searxng/blob/943c8fb85bf1b3b2374517dd59da1a44046f99ff/.tool-versions)~
* nvm (https://github.com/searxng/searxng/blob/943c8fb85bf1b3b2374517dd59da1a44046f99ff/.nvmrc and npm)
* Custom Go handling in https://github.com/searxng/searxng/pull/5018
* Use the system wide binaries.

We could try unifying everything under one tool like https://mise.jdx.dev to avoid versions mismatch, headaches while trying to setup a proper dev environment or ease future development.

From what I've seen, [mise](https://mise.jdx.dev/tasks/) *could* also replace `make`, taking a *small* step forward in allowing development under NT.

knguyen1:
Any interest in moving to `uv` and `ruff`?  A [PEP621](https://peps.python.org/pep-0621/)-compliant `pyproject.toml` file will need to be added.

Why?  Maintaining separate `requirements.txt` files is icky.  And `pip install` (and dependency resolutions) even with cache is maddeningly slow.

I helped a team did it recently and went fine.  We're fully on ruff, uv, pre-commit, etc. now.  https://github.com/metabrainz/picard/pull/2690

return42:
> Any interest in moving to `uv` and `ruff`?

nope .. `uv` does not solve one issue and `ruff` is really a worse code checker .. we use basedpyright, see

- https://github.com/searxng/searxng/pull/5075

FYI: the opener does not ask for python only: python is just one component of many in the developer environment of SearXNG needed to implement the server part. GH-Actions, Node, go, make, bash and ... are other tools we need for code quality, deployment and other tasks.

knguyen1:
Yeah, I'm looking at the repo and at some point, all the tooling got away from you.  There's lots of duplication of logic and inconsistent configs/drifts.  I asked GPT5 for an opinion and here's what it came up with.

Thoughts?  I'll take a stab at this if there's buy in of the plan.

---

[removed since user would like to come up with their own migration plan]

return42:
>  I asked GPT5 for an opinion and here's what it came up with.

Sorry, this is the point where I leave the discussion.

[...]

PR #5801, opened 2026-03-02
[mod] tool: replace `black` with `ruff`

inetol:
`ruff` as py repo formatter.

In addition to Black features...
Handles [Docstring](https://docs.astral.sh/ruff/formatter/#docstring-formatting) and removes extra br.

---

 Damn large diff

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,9 +54,7 @@ jinja_contexts = {
     'searx': {
         'engines': searx.engines.engines,
         'plugins': searx.plugins.STORAGE,
-        'version': {
-            'node': os.getenv('NODE_MINIMUM_VERSION')
-        },
+        'version': {'node': os.getenv('NODE_MINIMUM_VERSION')},
         'enabled_engine_count': sum(not x.disabled for x in searx.engines.engines.values()),
         'categories': searx.engines.categories,
         'categories_as_tabs': {c: searx.engines.categories[c] for c in searx.settings['categories_as_tabs']},
@@ -65,6 +63,7 @@ jinja_contexts = {
 jinja_filters = {
     'group_engines_in_tab': searx.webutils.group_engines_in_tab,
 }
+
 
 # Let the Jinja template in configured_engines.rst access documented_modules
 # to automatically link documentation for modules if it exists.
@@ -85,6 +84,7 @@ def setup(app):
     app.connect('env-before-read-docs', before_read_docs)
     app.connect('source-read', source_read)
 
+
 # usage::   lorem :patch:`f373169` ipsum
 extlinks = {}
 
@@ -98,16 +98,12 @@ extlinks['origin'] = (GIT_URL + '/blob/' + GIT_BRANCH + '/%s', 'git://%s')
 extlinks['patch'] = (GIT_URL + '/commit/%s', '#%s')
 extlinks['docs'] = (DOCS_URL + '/%s', 'docs: %s')
 extlinks['man'] = ('https://manpages.debian.org/jump?q=%s', '%s')
-#extlinks['role'] = (
+# extlinks['role'] = (
 #    'https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-%s', '')
-extlinks['duref'] = (
-    'https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#%s', '%s')
-extlinks['durole'] = (
-    'https://docutils.sourceforge.io/docs/ref/rst/roles.html#%s', '%s')
-extlinks['dudir'] =  (
-    'https://docutils.sourceforge.io/docs/ref/rst/directives.html#%s', '%s')
-extlinks['ctan'] =  (
-    'https://ctan.org/pkg/%s', 'CTAN: %s')
+extlinks['duref'] = ('https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#%s', '%s')
+extlinks['durole'] = ('https://docutils.sourceforge.io/docs/ref/rst/roles.html#%s', '%s')
+extlinks['dudir'] = ('https://docutils.sourceforge.io/docs/ref/rst/directives.html#%s', '%s')
+extlinks['ctan'] = ('https://ctan.org/pkg/%s', 'CTAN: %s')
 
 extensions = [
     'sphinx.ext.imgmath',
@@ -115,13 +111,13 @@ extensions = [
     'sphinx.ext.viewcode',
     "sphinx.ext.autodoc",
     "sphinx.ext.intersphinx",
-    "sphinx_issues", # https://github.com/sloria/sphinx-issues/blob/master/README.rst
+    "sphinx_issues",  # https://github.com/sloria/sphinx-issues/blob/master/README.rst
     "sphinx_jinja",  # https://github.com/tardyp/sphinx-jinja
     "sphinxcontrib.programoutput",  # https://github.com/NextThought/sphinxcontrib-programoutput
     'linuxdoc.kernel_include',  # Implementation of the 'kernel-include' reST-directive.
-    'linuxdoc.rstFlatTable',    # Implementation of the 'flat-table' reST-directive.
-    'linuxdoc.kfigure',         # Sphinx extension which implements scalable image handling.
-    "sphinx_tabs.tabs", # https://github.com/djungelorm/sphinx-tabs
+    'linuxdoc.rstFlatTable',  # Implementation of the 'flat-table' reST-directive.
+    'linuxdoc.kfigure',  # Sphinx extension which implements scalable image handling.
+    "sphinx_tabs.tabs",  # https://github.com/djungelorm/sphinx-tabs
     'myst_parser',  # https://www.sphinx-doc.org/en/master/usage/markdown.html
     'notfound.extension',  # https://github.com/readthedocs/sphinx-notfound-page
 ]
@@ -131,21 +127,19 @@ autodoc_default_options = {
     'member-order': 'bysource',
 }
 
-myst_enable_extensions = [
-  "replacements", "smartquotes"
-]
+myst_enable_extensions = ["replacements", "smartquotes"]
 
 suppress_warnings = ['myst.domains']
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
-    "babel" : ("https://babel.readthedocs.io/en/latest/", None),
+    "babel": ("https://babel.readthedocs.io/en/latest/", None),
     "flask": ("https://flask.palletsprojects.com/en/stable/", None),
     "flask_babel": ("https://python-babel.github.io/flask-babel/", None),
     "werkzeug": ("https://werkzeug.palletsprojects.com/en/stable/", None),
     "jinja": ("https://jinja.palletsprojects.com/en/stable/", None),
-    "linuxdoc" : ("https://return42.github.io/linuxdoc/", None),
-    "sphinx" : ("https://www.sphinx-doc.org/en/master/", None),
+    "linuxdoc": ("https://return42.github.io/linuxdoc/", None),
+    "sphinx": ("https://www.sphinx-doc.org/en/master/", None),
     "valkey": ('https://valkey-py.readthedocs.io/en/stable/', None),
     "pygments": ("https://pygments.org/", None),
     "lxml": ('https://lxml.de/apidoc', None),
@@ -161,7 +155,6 @@ html_theme_options = {
     "source_repository": "https://github.com/searxng/searxng/",
     "source_branch": "master",
     "source_directory": "docs/",
-
     # Show link to repository & searx.space in footer
     "footer_icons": [
         {
@@ -212,6 +205,4 @@ html_copy_source = True
 
 # LaTeX ----------------------------------------------------------------
 
-latex_documents = [
-    (master_doc, "searxng-{}.tex".format(VERSION_STRING), html_title, author, "manual")
-]
+latex_documents = [(master_doc, "searxng-{}.tex".format(VERSION_STRING), html_title, author, "manual")]
--- a/searx/botdetection/config.py
+++ b/searx/botdetection/config.py
@@ -7,11 +7,10 @@ structured dictionaries.  The configuration schema is defined in a dictionary
 structure and the configuration data is given in a dictionary structure.
 """
 
-import typing
-
 import copy
 import logging
 import pathlib
+import typing
 
 import tomllib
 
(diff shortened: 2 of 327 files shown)


return42:
Please read:

- https://github.com/searxng/searxng/pull/5075#issuecomment-3167435426

inetol:
I only want ruff as a formatter, not the linter part https://github.com/searxng/searxng/pull/5075#issuecomment-3171883154

Bnyro (approved):
I've used Ruff for some other Python projects recently and have been quite happy with it, didn't notice any issues when using the formatter.

I also think that the formatting changes here are reasonable, so from my side a big 👍 .

PS: I've also switched to using Ruff for formatting myself (also for SearXNG) :)

searx/botdetection/config.py
```diff
@@ -235,10 +235,7 @@ def validate(
     ``schema_dict``.  Argument deprecated is a dictionary that maps deprecated
     configuration names to a messages::
 
-        deprecated = {
-            "foo.bar" : "config 'foo.bar' is deprecated, use 'bar.foo'",
-            "..."     : "..."
-        }
+        deprecated = {"foo.bar": "config 'foo.bar' is deprecated, use 'bar.foo'", "...": "..."}
```
Bnyro, line 236: is it possible to disable formatting in some places? e.g. here it doesn't make much sense to change it this way

searx/engines/wikidata.py
```diff
@@ -562,9 +559,7 @@ def get_where(self):
         return """OPTIONAL { ?article{language} schema:about ?item ;
              schema:inLanguage "{language}" ;
              schema:isPartOf <https://{language}.wikipedia.org/> ;
-             schema:name ?articleName{language} . }""".replace(
-            '{language}', self.language
-        )
```
Bnyro, line 213: the changes in this file here also don't seem right really

searx/favicons/cache.py
```diff
@@ -292,10 +292,7 @@ class FaviconCacheSQLite(sqlitedb.SQLiteAppl, FaviconCache):  # pyright: ignore[
     """Delete blobs.sha256 (BLOBs) no longer in blob_map.sha256."""
 
     SQL_ITER_BLOBS_SHA256_BYTES_C = (
-        "SELECT b.sha256, b.bytes_c FROM blobs b"
-        "  JOIN blob_map bm "
-        "    ON b.sha256 = bm.sha256"
-        " ORDER BY bm.m_time ASC"
+        "SELECT b.sha256, b.bytes_c FROM blobs b  JOIN blob_map bm     ON b.sha256 = bm.sha256 ORDER BY bm.m_time ASC"
```
Bnyro: this makes it very hard to read the SQL statement

searx/preferences.py
```diff
@@ -369,9 +367,15 @@ def from_http_request(cls, http_request: SXNG_Request):
...
-            # fmt: on
+            lang, qvalue = [
+                _.strip()
+                for _ in (
+                    l.split(';')
+                    + [
+                        'q=1',
+                    ]
+                )[:2]
+            ]
```
Bnyro: not sure what happened here, but this is also looks weird

docs/conf.py
```diff
@@ -190,13 +184,7 @@ def source_read(app, docname, source):
...
 
 html_sidebars = {
-    "**": [
-        "globaltoc.html",
-        "project.html",
-        "relations.html",
-        "searchbox.html",
-        "sourcelink.html"
-    ],
+    "**": ["globaltoc.html", "project.html", "relations.html", "searchbox.html", "sourcelink.html"],
```
return42: It was intentional to not format `./docs/conf.py` and `./setup.py`  .. if we want to format theses files we can add them to ..

https://github.com/searxng/searxng/blob/f6e360a1348a4ddb0ac87a836a22f9e18e1ef98d/manage#L54

searx/botdetection/config.py
```diff
@@ -301,7 +297,7 @@ def _validate(
             issue_list.append(
                 SchemaIssue(
                     'invalid',
-                    (f"data_dict: type mismatch '{name}':" f" expected {type(schema_value)}, is: {type(data_value)}"),
+                    (f"data_dict: type mismatch '{name}': expected {type(schema_value)}, is: {type(data_value)}"),
```
return42, line 297: That speaks for ruff 👍

searx/botdetection/config.py
```diff
@@ -346,7 +342,6 @@ def dict_deepupdate(base_dict: dict[str, typing.Any], upd_dict: dict[str, typing
         # For each upd_key & upd_val pair in upd_dict:
 
         if isinstance(upd_val, dict):
-
```
return42, line 349: There was a long discussion in 2023, before black stopped to "remove newline after code block open"  

[If the author is unable to use paragraphs to add emphasis, the code becomes unreadable and the goal of code formatting (readability) is lost.](https://github.com/searxng/searxng/pull/2159#issuecomment-1425723977)

searx/engines/doku.py
```diff
@@ -26,12 +26,8 @@
...
 # Doku is OpenSearch compatible
 base_url = 'http://localhost:8090'
-search_url = (
-    # fmt: off
-    '/?do=search'
-    '&{query}'
-    # fmt: on
-)
+search_url = '/?do=search&{query}'
+
```
return42, line 40: Formatting "one line for one argument" was intended.

searx/engines/wikidata.py
```diff
@@ -526,9 +526,7 @@ def get_select(self):
     def get_where(self):
         return """  OPTIONAL { ?item p:{name} ?{name}Node .
     ?{name}Node rdf:type wikibase:BestRank ; ps:{name} ?{name} .
-    OPTIONAL { ?{name}Node psv:{name}/wikibase:quantityUnit ?{name}Unit. } }""".replace(
-            '{name}', self.name
-        )
+    OPTIONAL { ?{name}Node psv:{name}/wikibase:quantityUnit ?{name}Unit. } }""".replace('{name}', self.name)
```
return42: Its not critical here, but this is a query string ([SPARQL](https://en.wikipedia.org/wiki/SPARQL)) and should not modified by the code formatter.

searx/engines/yahoo_news.py
```diff
@@ -38,12 +38,7 @@
 categories = ['news']
 
 # search-url
-search_url = (
-    # fmt: off
-    'https://news.search.yahoo.com/search'
-    '?{query}&b={offset}'
-    # fmt: on
-)
+search_url = 'https://news.search.yahoo.com/search?{query}&b={offset}'
```
return42: Formatting "one line for one argument" was intended.

searx/search/processors/online_currency.py
```diff
@@ -91,11 +90,13 @@ def get_params(self, search_query: "SearchQuery", engine_category: str) -> Onlin
...
-        )  # pyright: ignore[reportAssignmentType]
+            from_iso4217,
+            ui_locale.language,  # pyright: ignore[reportAssignmentType]
+        )
         to_name: str = CURRENCIES.iso4217_to_name(
-            to_iso4217, ui_locale.language
-        )  # pyright: ignore[reportAssignmentType]
+            to_iso4217,
+            ui_locale.language,  # pyright: ignore[reportAssignmentType]
+        )
```
return42, line 97: Was it handcrafted? .. looks more readable 👍 

In black we can archive the same format just by adding a comma to the end of the last command: [A pre-existing trailing comma informs Black to always explode contents of the current bracket pair into one item per line. ](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas)

Related:

- [Pragmatism](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#pragmatism)

searx/external_bang.py
```diff
@@ -61,9 +61,7 @@ def resolve_bang_definition(bang_definition: str, query: str) -> tuple[str, int]
     return (url, rank)
 
 
-def get_bang_definition_and_autocomplete(
-    bang: str, external_bangs_db: dict[str, t.Any] | None = None
-):  # pylint: disable=invalid-name
+def get_bang_definition_and_autocomplete(bang: str, external_bangs_db: dict[str, t.Any] | None = None):  # pylint: disable=invalid-name
```
return42, line 64: That is a bit wired; in the `ruff.toml` the config  is `line-length = 120` but here we have 135 characters in this line.

----

This is just one example, there are more lines with a length above 120 chars.

utils/searxng_check.py
```diff
@@ -21,17 +20,16 @@
...
 if os.path.isfile(OLD_SETTING):
-    msg = (
-        '%s is no longer valid, move setting to %s' % (
-            OLD_SETTING,
-            os.environ.get('SEARXNG_SETTINGS_PATH', '/etc/searxng/settings.yml')
-        ))
+    msg = '%s is no longer valid, move setting to %s' % (
+        OLD_SETTING,
+        os.environ.get('SEARXNG_SETTINGS_PATH', '/etc/searxng/settings.yml'),
+    )
```
return42, line 23: handcrafted? .. looks more readable than before 👍

return42 (commented):
I haven't comment every item, but below you will find a few of my assessments of the code changes (formatting) that have been made.

----

After reviewing the PR, I have come to the conclusion that ruff formats f-strings better than black... however, some strings are also destroyed -> in the doc-strings, but ruff also made unacceptable changes to the preformatted (SQL) statements.  IMO: These and similar strings should not be formatted.

Since black [v24.4.1](https://github.com/psf/black/blob/main/CHANGES.md#2441), f-string syntax is also supported... I'm not sure if we can optimize/configure black to process f-strings better.

searx/engines/yahoo_news.py
Bnyro: But anyways this doesn't really seem to be important here imo, even though it was intended. It's also reasonably good to read without splitting it into two lines

searx/engines/wikidata.py
inetol: Query wasn't modified, only replace function inlining

searx/engines/wikidata.py
inetol: Query wasn't modified, only replace function inlining

searx/external_bang.py
inetol: It probably don't detect these kinds of comments. Weird enough, if you format by changing the length to 80 and then revert back to 120, all those cases seem fixed.

The reason I didn't switch to 80 in the first place is that diff is enormous.

searx/engines/wikidata.py
return42: That is what I said by "Its not critical here" .. but the point is, that every hand formatted string will be collapsed or reformatted in a way that the string isn't  readable anymore .. take `SQL_ITER_BLOBS_SHA256_BYTES_C` for a better example.

searx/botdetection/config.py
```diff
@@ -235,10 +235,12 @@ def validate(
     ``schema_dict``.  Argument deprecated is a dictionary that maps deprecated
     configuration names to a messages::
 
+        ```
         deprecated = {
             "foo.bar" : "config 'foo.bar' is deprecated, use 'bar.foo'",
             "..."     : "..."
         }
+        ```
```
return42, line 237: Why did you add markdown's quote here? .. This is a [literal block](https://docs.searxng.org/dev/reST.html#literal-blocks) in reST markup, see `::`.

--- 

The funny thing: ruff fixes the issue in the second commit :-)

searx/external_bang.py
return42: About 120 and 80 character line length: ["code formatter accepts length of 120, but editor should prefer 80"](https://github.com/searxng/searxng/pull/4892)

searx/botdetection/config.py
inetol: This code snipp is stuffed into a single line on format..:

```py
deprecated = {"foo.bar": "config 'foo.bar' is deprecated, use 'bar.foo'", "...": "..."}
```

If it's okay, I'll leave it like this

return42:
Should we really continue pursuing this PR? ... Ruff is actually already out of the running due to the aforementioned shortcomings and limitations.

inetol:
I'd really like to use "ruff" instead of "black". Maybe we can do incremental updates by cherrypicking commits into separate smaller PR.

searx/botdetection/config.py
return42: no, its not OK .. this is a doc-string with a plain text block (code block) .. 

Ruff has no business altering this plain-text markup in doc-strings. This intrusiveness on Ruff's part is merely a symptom of its flawed implementation. To reiterate: I consider this code checker to be very poor, and I do not wish to use it.

We invested many years in Black until it reached a state where it could be used effectively; now that everything is in order, this PR proposes introducing yet another code checker into our toolchain — I do not think this is a good idea.

searx/engines/wikidata.py
return42: It is unacceptable that Ruff performs code formatting here that ultimately only makes it harder for developers to read the code. The purpose of a code formatter is to format code for improved readability; unfortunately, this is not possible with Ruff when it starts messing around inside strings and unilaterally altering them.

inetol:
Here's what I'm going to do... For now I'll push the formatted changes in separate PRs. In the end, we'll see whether it's worth switching to "ruff".

return42:
> I'd really like to use "ruff" instead of "black". 

Why, that is my question .. / I can't see any good reasons for.

inetol:
> ... I can't see any good reasons for.

Formats Docstring, handles edge cases like code wrapping or superfluous spaces, and is faster to check the codebase.

return42:
I can only repeat:

> Formats Docstring, 

Its not its job and it does this job very ugly .. me and Bnyro named some reasons here in this PR already.

> handles edge cases like code wrapping 

I don't want imports sorted by the ruff, its not its job, and it does this job very ugly .. if we want sort the import order it is recommended to use https://pypi.org/project/isort/

> or superfluous spaces,
 
These spaces (empty lines) are added for readability I don't want this *harass mode* from ruff .. we had a long discussion about this topic in the black project and I want to avoid going this cycle again.

> faster to check the codebase.

"We're the fastest" we can read that every other line at Astral. That’s no surprise, considering they implement such shoddy tools that completely disregard every rule. The guys at Astral genuinely believe that no matter what they touch, they’ll inevitably make it "faster". This advertising from Astral really gets on my nerves; in my book, they’re a total dump of a company.

inetol:
OK, I agree. As for the import sorting, if it comes from the "fmtide" commit, it's from the IDE normalizer.

All the PRs with the formatting changes are already open, although 80% of the diff consists of whitespace removal and import sorting, so it might not be worth polluting the hist for this 🤷

Closing this.

return42:
> if it comes from the "fmtide" commit, it's from the IDE normalizer.

What do you mean by "fmtide" commit? .. and what is the "IDE normalizer"?

-----

> All the PRs with the formatting changes are already open, although 80% of the diff consists of whitespace removal and import sorting, so it might not be worth polluting the hist for this 🤷

I do not wish to unnecessarily pollute the version history with this PRs, as I do not agree with the removal of the well intended placed line breaks, given that the PRs impair readability.

Regarding import order, we could consider using `isort`, provided that is deemed desirable.

inetol:
> What do you mean by "fmtide" commit? .. and what is the "IDE normalizer"?

The default LSP Python server in the IDE

> Regarding import order, we could consider using `isort`, provided that is deemed desirable.

Maybe later, when https://github.com/searxng/searxng/issues/5056 is ready

PR #5803, opened 2026-03-03
[feat] autocomplete: add bing autocompleter

Bnyro:
- go to the settings
- set the autocompleter to bing
- search something

## Author's checklist
- Bing returns some weird Unicode characters in its response, e.g. when you load ` https://www.bing.com/AS/Suggestions?pt=page.serp&qry=donald&csr=1&cvid=ABF762C5D15D4BAFBA24B0C3F9BC6712`  suggestions look like `donald glover`
- therefore we manually need to filter out these unicode characters

Apart from that the implementation is pretty straight forward.

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """This module implements functions needed for the autocompleter."""
+
 # pylint: disable=use-dict-literal
+import string
+import random
 
 import json
 import typing as t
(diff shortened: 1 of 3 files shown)


searx/autocomplete.py
```diff
@@ -53,6 +56,26 @@ def baidu(query: str, _sxng_locale: str) -> list[str]:
     return results
 
 
+def bing(query: str, _sxng_locale: str) -> list[str]:
+    # baidu search autocompleter
```
inetol: baidu?

inetol (approved):
LGTM

searx/autocomplete.py
Bnyro: Thanks for the catch, copied and pasted from the baidu autocompleter above

PR #5804, opened 2026-03-03
[feat] plugins: add blocklists plugin with support for easylist blocklists

Bnyro:
This plugin allows to load filter rules from Easylist filter rule files and blocks all results that match the filter.

These changes are motiviated by https://github.com/searxng/searxng/issues/5029: There also are filter lists that can block AI images, but I haven't done much research about what lists exist there. The lists suggested in the docs are only about filtering AI-generated articles frome the general search results.

Implementation notes:
- the filter rule files are loaded on each startup, so there's room for future improvement
- the performance is not that good if the filter lists are large - so lists should be <2000 entries in total

Relevant resources:
- https://help.eyeo.com/en/adblockplus/how-to-write-filters
- https://github.com/alvi-se/ai-ublock-blacklist

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -238,6 +238,9 @@ plugins:
   searx.plugins.hostnames.SXNGPlugin:
     active: true
 
+  searx.plugins.blocklists.SXNGPlugin:
+    active: true
+
   searx.plugins.time_zone.SXNGPlugin:
     active: true
 
@@ -276,6 +279,13 @@ plugins:
 # '(.*\.)?youtube\.com$': 'yt.example.com'
 # '(.*\.)?youtu\.be$': 'yt.example.com'
 #
+
+
+# Configuration of the "Blocklists plugin":
+#
+# blocklists:
+#   - "https://raw.githubusercontent.com/alvi-se/ai-ublock-blacklist/refs/heads/master/list.txt"
+#   - "https://raw.githubusercontent.com/iam-py-test/my_filters_001/refs/heads/main/antimalware.txt"
 
 
 categories_as_tabs:
--- a/docs/dev/easylist.rst
+++ b/docs/dev/easylist.rst
@@ -1,0 +1,13 @@
+.. _easylist parser:
+
+===============
+Easylist parser
+===============
+
+.. contents:: Contents
+   :depth: 2
+   :local:
+   :backlinks: entry
+
+.. automodule:: searx.easylist
+  :members:
--- a/docs/dev/plugins/blocklists.rst
+++ b/docs/dev/plugins/blocklists.rst
@@ -1,0 +1,9 @@
+.. _blocklists plugin:
+
+==========
+Blocklists
+==========
+
+.. automodule:: searx.plugins.blocklists
+   :members:
+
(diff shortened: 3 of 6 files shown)


searx/settings.yml
```diff
@@ -278,6 +281,13 @@ plugins:
 #
 
 
+# Configuration of the "Blocklists plugin":
+#
+# blocklists:
+#   - "https://raw.githubusercontent.com/alvi-se/ai-ublock-blacklist/refs/heads/master/list.txt"
+#   - "https://raw.githubusercontent.com/iam-py-test/my_filters_001/refs/heads/main/antimalware.txt"
```
return42, line 288: If I enable this lists, I get:

```
DEBUG   searx.network                 : HTTP Request: GET https://raw.githubusercontent.com/alvi-se/ai-ublock-blacklist/refs/heads/master/list.txt "HTTP/1.1 200 OK" (text/plain; charset=utf-8)
DEBUG   searx.network                 : HTTP Request: GET https://raw.githubusercontent.com/iam-py-test/my_filters_001/refs/heads/main/antimalware.txt "HTTP/1.1 200 OK" (text/plain; charset=utf-8)
Process granian-worker:
  File "/home/markus/.local/share/mise/installs/python/3.10.18/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/home/markus/.local/share/mise/installs/python/3.10.18/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/share/searxng/local/py3/lib/python3.10/site-packages/granian/server/mp.py", line 90, in wrapped
    callback = callback_loader()
  File "/share/searxng/local/py3/lib/python3.10/site-packages/granian/_internal.py", line 65, in load_target
    module = load_module(path)
  File "/share/searxng/local/py3/lib/python3.10/site-packages/granian/_internal.py", line 47, in load_module
    __import__(module_name)
  File "/share/searxng/searx/webapp.py", line 1387, in <module>
    init()
  File "/share/searxng/searx/webapp.py", line 1355, in init
    searx.plugins.initialize(app)
  File "/share/searxng/searx/plugins/__init__.py", line 109, in initialize
    STORAGE.init(app)
  File "/share/searxng/searx/plugins/_core.py", line 250, in init
    if not plg.init(app):
  File "/share/searxng/searx/plugins/blocklists.py", line 86, in init
    rule = searx.easylist.parse(line)
  File "/share/searxng/searx/easylist.py", line 197, in parse
    raise EasylistFilterParserException("rule is missing a pattern")
searx.easylist.EasylistFilterParserException: rule is missing a pattern
[ERROR] Unexpected exit from worker-1
```

Bnyro: Sorry, I only tested the first blocklist (https://github.com/alvi-se/ai-ublock-blacklist).

It seems like the second list contains rules like `$all,domain=example.com|example.org` which don't have any URL glob to match, but I don't think that's worth supporting for now because 99% of the rules don't use it and it would take a lot of work to support all the possible `extra_options`.

Bnyro: Fixed and rebased with the latest commit.

return42:
Copied from a internal discussion:

>> The lists and the syntax are very complex .. they have to be, because they selectively prevent individual parts of a page from being reloaded in a web browser and also consider the context in which something is being loaded .. we rarely, if ever, have such parts (results) in our result lists (at least that's my assumption) .. would it perhaps make sense if we only removed entries from the lists that block an entire domain .. for example, only entries like .. `||get-express-vpn.online^$all` .. ?

> `all`, `doc` and `document` would make sense, I think, yes.

PR #5807, opened 2026-03-05
[upd] engines: sync "aa" mirrors

inetol:


Bnyro (approved):
Their mirrors change more frequent than the weather forecast...

return42:
From where did you got the new URLs?

Bnyro:
@return42 They're listed on the bottom of their website, e.g. on https://annas-archive.gl

inetol:
Also in https://en.wikipedia.org/wiki/Anna%27s_Archive or their official subreddit

return42:
> Their mirrors change more frequent than the weather forecast...
>> Also in https://en.wikipedia.org/wiki/Anna%27s_Archive or their official subreddit

May we can fetch (and cache) them in the init method of the engine? .. like we do in the radio-browser engine:

https://github.com/searxng/searxng/blob/c1b211aeead88d06e65565aaef61c0d79e65a8df/searx/engines/radio_browser.py#L59-L62

https://github.com/searxng/searxng/blob/c1b211aeead88d06e65565aaef61c0d79e65a8df/searx/engines/radio_browser.py#L65-L70

PR #5827, opened 2026-03-07
[fix] engines: remove google arc/async params

vojkovic:
## What does this PR do?

Removes the `asearch` and `async` command from searches using the google engine. 

Deployed on my instance to see if it can work in production - https://github.com/privau/searxng/commit/acada95f9fc33a9a8f23cf7e3a349950a3690924

PR is a draft currently as the thumbnails aren't working. (They weren't working before as well to be fair.) I do believe they are being returned in the HTML though, so they just need to be parsed correctly.

I found that this fixes the current problem with the google engine not returning any results.

`make run`

Confirm google engine is working.

## Author's checklist

 - [ ] Thumbnails to return with results

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -287,9 +287,7 @@ def request(query: str, params: "OnlineParams") -> None:
     """Google search request"""
     # pylint: disable=line-too-long
     start = (params["pageno"] - 1) * 10
-    str_async = ui_async(start)
     google_info = get_google_info(params, traits)
-    logger.debug("ARC_ID: %s", str_async)
 
     # https://www.google.de/search?q=corona&hl=de&lr=lang_de&start=0&tbs=qdr%3Ad&safe=medium
     query_url = (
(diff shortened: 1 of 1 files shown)


tiekoetter (approved):

Bnyro (approved):
Works fine for me too! 👍

vojkovic:
Didn't have the time to do the thumbnails today. Considering the fixing of the Google engine is the biggest priority I will merge this now and work on the thumbnails later.

unixfox:
Just sharing this here.

Removing the `asearch` and `async` parameters might not be the best idea for the public instances because those were introduced to bypass the rate limiting on Google: https://github.com/searxng/searxng/issues/159

This is why there were a lot of discussion about how to keep this functionality working, like this one: https://github.com/searxng/searxng/issues/5286#issuecomment-3398566747 and this one: https://github.com/searxng/searxng/issues/5286#issuecomment-3423496189

If in the future there would only be a few public instances working with google engine, the removal of `asearch` and `async` parameters in this PR would likely be the culprit.

PR #5841, opened 2026-03-10
[feat] add Advanced theme with grid view and multi-fetch engine support

SeriousConcept1134:
### What does this PR do?

This PR introduces the **Advanced** theme to SearXNG, forked from the default "Simple" theme, with several enhancements focused on media discovery and advanced layout control. Key changes include:
- **Advanced Media Layouts**: A new resizable grid view for video results, including a UI toggle and a range slider for thumbnail size adjustments.
- **Results per Page (Video)**: Implementation of a "Results per Page" selector (10-50) for video searches.
- **Engine Multi-fetch Capability**: The `google_videos` engine has been refactored to perform multiple asynchronous fetches to aggregate higher result counts as requested by the user.
- **Dynamic Theme Management**: A header toggle allowing users to switch between Light, Dark, and Black (OLED) modes dynamically.
- **Persistent Settings in Pagination**: Enhanced pagination forms to ensure the `results_per_page` setting persists across searches and page navigation.
- **Adaptive UI**: The "Back to Top" button now adjusts its position dynamically to accommodate different layouts (list vs. grid) in video results.
- **Backend Integration**: Core updates to `search.models`, `preferences`, and the `abstract` engine processor to natively handle the `results_per_page` parameter.

### Why is this change important?

The "Advanced" theme provides a more powerful and customizable interface for media discovery. By enabling native grid resizing and high-volume result fetching for videos, SearXNG becomes a more versatile tool for power users without compromising privacy or the clean aesthetic of the original design.

1. Build the new theme:
   ```bash
   ./manage themes.all
   ```
2. Start the SearXNG instance:
   ```bash
   ./manage webapp.run
   ```
3. Navigate to Preferences and select the **Advanced** theme.
4. Test the **Theme Toggle** and its persistence via cookies.
5. Perform a video search:
   - Use the **Results per Page** dropdown to fetch 50 results.
   - Use the **Grid View Toggle** and **Resizer** to adjust the layout.
   - Verify that **Pagination** maintains the results per page setting.
   - Check that the **Back to Top** button remains accessible in both list and grid views.

### Author's checklist

- [x] All code formatted with `black`.
- [x] Frontend linting and formatting (`biome`, `stylelint`) passed in `client/advanced`.
- [x] Multi-fetch engine logic verified.
- [x] Setting persistence verified across pagination.

### Related issues

None.


return42:
TBH: this PR is hard to review .. I see hundreds of copied files and topics not related to a design change on client side:

> * **Engine Multi-fetch Capability**: The `google_videos` engine has been refactored to perform multiple asynchronous fetches to aggregate higher result counts as requested by the user.

Without going into the details .. I made a small user test and as a user (sorry for my ignorance) I can't see any "advanced" feature, that could not be implemented in a simple theme(?).

@inetol what do you think?

SeriousConcept1134:
> TBH: this PR is hard to review .. I see hundreds of copied files and topics not related to a design change on client side:
> 
> > * **Engine Multi-fetch Capability**: The `google_videos` engine has been refactored to perform multiple asynchronous fetches to aggregate higher result counts as requested by the user.
> 
> Without going into the details .. I made a small user test and as a user (sorry for my ignorance) I can't see any "advanced" feature, that could not be implemented in a simple theme(?).
> 
> @inetol what do you think?

I'll try to explain a bit better:
The Google Videos engine only returns 10 results per page. This is a hard limitation. In order to facilitate the "results per page" functionality, I had to come up with a way to display more than 10 results per page if the user selects, for example 50 results, and uses the Google Videos engine exclusively (via `!gov <query>`).
This is where the multi-fetch capability comes into play. If I you select 50 results per page on the video search, using Google Videos, it will fetch 5x10 via a loop, and combine the results together to fit on a single page. I was concerned at first this approach could trigger bot detection, but after some testing it seems to work well.

Regarding my choice to create an "Advanced" theme, it is just a play on the word "Simple" :) But you are correct that all of these improvements can be applied and merged into the default "Simple" theme (as "Advanced" is just a modified copy of "Simple"). The idea behind this approach was having the ability to toggle these extra features on/off by simply switching up the theme selection. I assumed because these changes are novel that you would want the default theme preserved as-is just in case. If you notice most of the new files are just copies of their counterparts from the "Simple" theme.

With regards to easing the review process, would you like me to add more code commentary to give context to all the changes to make the process go more smoothly?

Let me know if you prefer a different approach. I can instead just modify the "Simple" theme and forego the need for a secondary theme selection.

return42:
> The Google Videos engine only returns 10 results per page. This is a hard limitation. In order to facilitate the "results per page" functionality,
> .. This is where the multi-fetch capability comes into play. If I you select 50 results per page on the video search, using Google Videos, it will fetch 5x10 via a loop, and combine the results together to fit on a single page.

Will not work / is not intended / does not fit to how SearXNG works .. there is a outer timeout limit (including runtime of functions `request` & `response`) .. if there is no option "results per page" in the first request that goes to google we can't offer  "results per page".

However this discussion does not fit to the scope of this PR.

> you notice most of the new files are just copies of their counterparts from the "Simple" theme.

Doubling the code base is not a strategy for serious development.

> Let me know if you prefer a different approach. 

Let's hear what @inetol has to say. Personally, I think features like the day/night switch can be incorporated into the existing code. We could also add a new theme, but I'm against simply copying code and burdening the maintainers with the task of maintaining these copies.

SeriousConcept1134:
> Will not work / is not intended / does not fit to how SearXNG works .. there is a outer timeout limit (including runtime of functions `request` & `response`) .. if there is no option "results per page" in the first request that goes to google we can't offer "results per page".

It's okay, I can easily drop this approach, as it was always intended as a workaround. The "results per page" drop-down menu in the videos section otherwise works exactly as intended. The only downside will be that if you want to get results from Google Videos exclusively (via `!gov <query>`) then you will only get 10 results per page, regardless of the number you selected in the "results per page" drop-down menu. I still think this is a very worthwhile feature to have, even without adapting the Google Videos engine for it.

> Doubling the code base is not a strategy for serious development.

That's a fair point, and I did not consider that. As I said, most/all of these changes can be simply integrated into the default "Simple" theme, rather than having a separate theme for them. Although to be fair, any addition of a new theme to the stock SearXNG will involve expanding the code base to facilitate it. I guess that's one of the reasons there's only ever been one single theme (which kind of begs the question of the usefulness of the theme selection feature?).

> Let's hear what @inetol has to say. Personally, I think features like the day/night switch can be incorporated into the existing code. We could also add a new theme, but I'm against simply copying code and burdening the maintainers with the task of maintaining these copies.

I do acknowledge that having a single PR for all these changes may be a little too ambitious. I suggest breaking down the changes and making a separate PR for each of them will make them easier to digest and review. Also I can drop the "Advanced" theme altogether, and instead just focus on modifying the existing "Simple" theme.

I propose to create separate PRs for the following features:
- Adding the dynamic theme style (Light/Dark) switch.
- Adding a toggle that allows showing the search results in grid view, in the videos section (+grid view dynamic resizer).
- Adding the drop-down menu to set "X results per page" in the videos search section (with options being: 10, 20, 30, 40, 50).

With regards to the Google Videos multi-fetch capability, I will simply drop that implementation entirely if you say it goes against the project's philosophy/guidelines.

Let me know what you think, thanks.

inetol:
Bad timing for new features to themes, especially if they are this big.

I don't like our current architecture for themes at all, and this PR reflects that perfectly. There was already the [Oscar theme issue](https://github.com/searxng/searxng/issues/1169) back then, and this is pretty much the same thing.

The only way to make a frontend more independent and facilitate future development is to not rely on direct templating and instead use the JSON endpoint as a common data source. This is what I am working on behind the scenes atm.

Bnyro:
Due to the fact that the changes seem like they were not properly reviewed by a human and do a lot of different things that shouldn't be done in a single PR, we're closing this.

PR #5858, opened 2026-03-13
[fix] Youtube video - "Error 153" send HTTP Referrer (referrerpolicy)

### LINKED ISSUE #5844: Youtube video - "Error 153"
Hello

Sorry if this question has already been asked, but it seems that YouTube videos can no longer be played in the browser.

Is this a bug in my instance or is this a general problem?
Thanks

return42:
For videos from ``www.youtube.com`` this patch adds ``referrerpolicy`` and other by YT required permission policies for the iframe.

API Clients that use the YouTube embedded player (including the YouTube IFrame Player API) must provide identification through the HTTP ``Referer`` request header[1].

If the error still occurs, check if server headers (e.g., in .htaccess or Nginx) are overriding the referrer policy[2].

[1] https://developers.google.com/youtube/terms/required-minimum-functionality#embedded-player-api-client-identity
[2] hint comes from AI: https://chat.mistral.ai
[3] https://developers.google.com/youtube/player_parameters

Close: https://github.com/searxng/searxng/issues/5844

### DIFF
--- a/searx/templates/simple/result_templates/videos.html
+++ b/searx/templates/simple/result_templates/videos.html
@@ -18,7 +18,11 @@
 {{- result_sub_footer(result) -}}
 {% if result.iframe_src -%}
 <div id="result-video-{{ index }}" class="embedded-video invisible">
-  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen></iframe>
+  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen
+    {% if result.parsed_url.hostname in ("www.youtube.com", ) -%}
+    allow="picture-in-picture" referrerpolicy="origin">
+    {%- endif -%}
+  </iframe>
 </div>
 {%- endif %}
 {{ result_footer(result) }}


searx/templates/simple/result_templates/videos.html
```diff
@@ -18,7 +18,11 @@
 {{- result_sub_footer(result) -}}
 {% if result.iframe_src -%}
 <div id="result-video-{{ index }}" class="embedded-video invisible">
-  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen></iframe>
+  <iframe data-src="{{result.iframe_src}}" frameborder="0" allowfullscreen
+    {% if result.parsed_url.hostname in ("www.youtube.com", ) -%}
+    allow="encrypted-media; picture-in-picture" referrerpolicy="origin">
```
Bnyro: Where did you find that `encrypted-media` is needed? For me, all embeds I tested also play fine without it

Bnyro (approved):

searx/templates/simple/result_templates/videos.html
return42: TBH: I thought `encrypted-media` could not be wrong. AFAIK its needed for DRM protected content.

> For me, all embeds I tested also play fine without it

Not sure YT has any DRM content .. what do you think, should we better remove `encrypted-media`?

------

Update: now I see [allow.encrypted-media](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe) is not supported by Safari .. to answer my question myself: YES, remove :-)

PR #5870, opened 2026-03-19
[feat] client/simple: show/hide autocomplete list on focus/blur

Aadniz:
## What does this PR do?

This PR improves the ability to close the autocomplete suggestion dropdown when searching.

The autocomplete suggestion list is currently a bit hard to close if it has been spawned once. Sometimes you may write something in the input field, change your mind and then want to click the first result on the page. But since the dropdown cannot be closed, the dropdown is in the way to be able to click on the first result.

In the screenshot below is an example of the entire suggestion dropdown covering the first result.

image

My implementation uses focus and blur events on the input field, and shows or hides the dropdown accordingly.

The "Tab" shortcut to close the dropdown has been removed because the "blur" event essentially does the same thing.

I also added a shortcut for pressing Escape to close the dropdown. I used the "keydown" event instead of "keyup" so that it would ensure the code executes before any browser defaults can intervene. I know preventDefault could also be used, but I think it's best to not try to guess too much on how users navigate around.

I personally find it a bit annoying that the dropdown is often in the way of the first result, and I struggle to close it.

Search for something that will bring up the suggestion menu, and click out of the input, or tab out of the input to have it being closed. Then you can click the input field again to show the dropdown again.

## AI Disclosure


Aadniz:
I committed all the static files, let me know if I should do that or not.

return42:
> I committed all the static files, let me know if I should do that or not.

Nope :-) .. To separate the modified source code from the build products use `make static.build.commit` / see [Development Quickstart](https://docs.searxng.org/dev/quickstart.html#development-quickstart)

Aadniz:
Yep alright, I just saw that when I had a look. Check the commits now

Bnyro (approved):
Looks and works well, good to merge from my side.

Thanks!

Bnyro:
No objections from other maintainers, hence merging.

Issue #4653, opened 2025-04-23
Bug: brave engine `SearxEngineTooManyRequestsException`


PuuGoo:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2025.3.30+7b4612e86

**How did you install SearXNG?**

**What happened?**

**How To Reproduce**

**Expected behavior**

**Screenshots & Logs**

**Additional context**

**Technical report**

Error
  * Error: searx.exceptions.SearxEngineTooManyRequestsException
  * Percentage: 20
  * Parameters: `('Too many request',)`
  * File name: `searx/search/processors/online.py:116`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

return42:
> Error: searx.exceptions.SearxEngineTooManyRequestsException
> Percentage: 20

On my instance the brave engine have a fail rate at 90% .. @unixfox @allendema @Austin-Olacsi you are familiar with the brave engine, could you please take a look at this error / Thanks!

unixfox:
I guess there is nothing that can really be done. Closing

return42:
I haven't tested it, so I don't know if it works at all, but it might be worth taking a look at this project:

- https://github.com/neon-mmd/websurfx/blob/rolling/src/engines/brave.rs

... maybe it will help us improve the reliability a bit(?) / on my instance, reliability has dropped to 40%.

update: dropped to 0% :-(

omentic:
FYI, I would really appreciate any improvements it's possible to make -- Brave has its own index, so mixing its results into the output helps with relevant ranking quite a bit. For now, I'm trying out the Mullvad Leta Brave backend. 

Seems to work decently if anyone comes here looking for an alternative.

ahmed-raza-siddqui:
Hi all, I'd like to share another suggestion. Maybe allow specifying API key, since that might increase the number of searches & get around rate-limiting aspect

vojkovic:
Since migrating to https://github.com/searxng/searxng/commit/be836e614a2f2e7512e2e1915a0bc37d51ee865c, I have had better reliability with brave. Can someone test if they notice the same?

return42:
Confirmed, I had some parsing errors but non SearxEngineTooManyRequestsException anymore 👍  

[...]

PR #5871, opened 2026-03-19
[feat] engines: add karmasearch engine (brave mirror)

Bnyro:
## What does this PR do?
- Implement https://karmasearch.org, which uses the Brave index

- Brave has very strict rate-limits, so having a second way for loading Brave results is definitely worth it
- see e.g. https://github.com/searxng/searxng/issues/4653

- !kai test
- !ka test
- !kav test
- !kan test

## Author notes
- ~Languages/regions are not supported yet, I couldn't figure out how to import the existing traits from Brave here. @return42, could you please help here?~ Done

## AI Disclosure


return42:
FYI: last force push was just a rebase before I start my review.

searx/engines/karmasearch.py
```diff
@@ -0,0 +1,152 @@
...
+from datetime import datetime
+from urllib.parse import urlencode
+import typing as t
+
+from dateutil import parser
+
+from searx import locales
+
+# we use the same language/locale traits as Brave
+from searx.engines.brave import fetch_traits  # pylint: disable=unused-import
```
return42: That is a wrong assumption. The regions of Brave are single-digit values like "de" and the regions of Karmasearch are the locale tags like "de-DE", for example.

From Brave:

Image

From Karmasearch:

Image

I will push a WIP commit with the fix of mine.

return42 (approved):
I pushed a WIP commit with my suggestion, @Bnyro can you please have a look .. if patch is OK for you, squash the WIP patch into the initial commit .. 

Thanks a lot for this engine 👍

Bnyro:
> I pushed a WIP commit with my suggestion, @Bnyro can you please have a look .. if patch is OK for you, squash the WIP patch into the initial commit ..
> 
> Thanks a lot for this engine 👍
>
Thank you for the fixup, works very reliably for the languages I tested!

I've only removed one debug statement that you probably forgot, and squashed it.

Thanks a lot for the review :+1:

Bnyro:
Apparently they now use Cloudflare :/

I tried using Cloudscraper, but it looks like Cloudscraper is no longer working either...

return42:
> Apparently they now use Cloudflare :/

@Bnyro this issue is fixed in:

- https://github.com/searxng/searxng/pull/5953

> I tried using Cloudscraper, but it looks like Cloudscraper is no longer working either...

the solution seems very simple:

> This PR adds a "Referer" header when searching towards Karmasearch so that the Cloudflare bot protection doesn't show up.

PR #5881, opened 2026-03-20
[feat] engines: add boardreader engine

Bnyro:
## What does this PR do?
- add support for boardreader.com

- Boardreader is a nice search engine to search through forums

- !boa :de test
- !boa :en hello

## AI Disclosure


return42:
I fixed some minor type issue [diff](https://github.com/searxng/searxng/compare/08912778c85266b9a20b2067f162af562f03552b..a074da1aff13aedd56b78a6c4fbae2b9893ade5d)

Often I see ` [Keyword]<search term>[/Keyword]` in the title and in the content field, e.g. when I search for "BMW":

grafik

And we have HTML numeric character reference (NCR) in title and content. e.g. `&#039;` when I search for "!boa don't do this at home guys" ..

grafik

searx/engines/boardreader.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def init(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_name=engine_settings["name"])
+    return True
+
+
+def _get_session_id() -> str:
+    cached: str = CACHE.get(CACHE_SESSION_ID_KEY)
```
Bnyro:
```suggestion
    cached: str | None = CACHE.get(CACHE_SESSION_ID_KEY)
```
I think that `get` returns `None` if nothing is found?

Bnyro: Fixed that now.

Bnyro:
Thanks for the review :+1: 

> Often I see  [Keyword]<search term>[/Keyword] in the title and in the content field, e.g. when I search for "BMW":
>
Very good point, I've added a Regex to just remove these "[Keyword]" strings now. This seems to be reliable enough for our case.

> And we have HTML numeric character reference (NCR) in title and content. e.g. &#039; when I search for "!boa don't do this at home guys" ..
>
Good catch, switched to using `html_to_text`.

return42 (approved):

Issue #4964, opened 2025-07-02
bing: results are often unrelevant to the search query


Aadniz:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**

39c50dc013944a0a27b4354c23f406956ac45971

**How did you install SearXNG?**

Manually executing `searx/webapp.py`

**What happened?**

Bing often returns unrelevant results. Sometimes matching 1 word, sometimes matching 2 words, and sometimes showing random results completely unrelevant to the search query.

The bug also consists of the same top-level domain appearing on nearly all results. This is also inconsistent.

**How To Reproduce**

The best method to trigger completely unrelevant results is having many words in the search query, like "how to install docker debian".

**Screenshots & Logs**

The screenshot below shows what looks like "michael" only matching in this search query with results only from michaels.com

![Image](https://github.com/user-attachments/assets/2190d1ed-e001-45e0-8131-6a88a5ff9a56)

The screenshot below doesn't seem to resemble anything in the search query at all.

![Image](https://github.com/user-attachments/assets/4ec5f7b4-7e9c-4cdd-a0f9-676c12f516ca)

Same search on another instance yuck.

![Image](https://github.com/user-attachments/assets/7794223e-a740-4417-be38-f6715803888c)

return42:
What strikes me is that in your queries the request for search results in English was enabled, but the result lists are in other languages, which reminds me very much of an old problem:

["It seems to me that bing only changes the language on some requests but not on all requests .. but what is striking is the fact that quitting and restarting the uwsgi allows an immediate switch of the language"](https://github.com/searxng/searxng/issues/2977)

The Bing engine needs an overhaul; we have many open bugs that are already labeld by --> [`bing-engine`](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A%22bing-engine%22&type=issues&p=2).

lkyu-ly:
same problem
sometimes it looks like only the first word was matched; and sometimes the domain name matched;
most importantly, the search results are almost totally different from what the original bing webpage returns.

LtqxWYEG:
Same here and I've also seen this happening with other search engines, such as Yahoo sometimes, but never as much as with Bing _lately_.

<img width="1113" height="944" alt="Image" src="https://github.com/user-attachments/assets/f402dbef-3892-4c01-a6ef-51d9adc85721" />

[...]

PR #5882, opened 2026-03-20
[feat] engines: add support for aol.com (bing index)

Bnyro:
## What does this PR do?
- add support for aol.com, which uses Bing

- The bing engine is very unreliable: https://github.com/searxng/searxng/issues/4964

- !aol test
- !aoli test
- !aolv test

## AI Disclosure

### DIFF
--- a/searx/engines/aol.py
+++ b/searx/engines/aol.py
@@ -1,0 +1,208 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""AOL supports WEB, image, and video search.  Internally, it uses the Bing
+index.
+
+AOL doesn't seem to support setting the language via request parameters, instead
+the results are based on the URL.  For example, there is
+
+- `search.aol.com <https://search.aol.com>`_ for English results
+- `suche.aol.de <https://suche.aol.de>`_ for German results
+
+However, AOL offers its services only in a few regions:
+
+- en-US: search.aol.com
+- de-DE: suche.aol.de
+- fr-FR: recherche.aol.fr
+- en-GB: search.aol.co.uk
+- en-CA: search.aol.ca
+
+In order to still offer sufficient support for language and region, the `search
+keywords`_ known from Bing, ``language`` and ``loc`` (region), are added to the
+search term (AOL is basically just a proxy for Bing).
+
+.. _search keywords:
+    https://support.microsoft.com/en-us/topic/advanced-search-keywords-ea595928-5d63-4a0b-9c6b-0b769865e78a
+
+"""
+
+from urllib.parse import urlencode, unquote_plus
+import typing as t
+
+from lxml import html
+from dateutil import parser
+
+from searx.result_types import EngineResults
+from searx.utils import eval_xpath_list, eval_xpath, extract_text
+
+if t.TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


return42:
I added language and region support (see doc-string for "how") .. further I fixed some minor (type) issues and added AOL to the online docs .. can you please have a deeper look / thanks!

return42:
Sorry, forgot to add `aol.rst` (online docs) in my last force push .. its now in [diff](https://github.com/searxng/searxng/compare/e84ddca2c35a06d802ee0a2c50dcabfc98919279..eb3b26ac671424950bee7d4172dd7c27ffa7e242)

searx/engines/aol.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def init(_):
+    if search_type not in ("search", "image", "video"):
+        raise ValueError(f"unsupported search type {search_type}")
+
+
+def request(query: str, params: "OnlineParams") -> None:
+
+    language, region = (params["searxng_locale"].split("-") + [None, None])[:2]
```
Bnyro:
```suggestion
    language, region = (params["searxng_locale"].split("-") + [None])[:2]
```
This looks a bit hackish, but anyways:
The second `None` is never needed, because `split` returns always at least one element:
```bash
$ python3
Python 3.14.3 (main, Feb 10 2026, 19:44:45) [GCC 14.2.1 20250405] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "".split("-")
['']
```

Bnyro:
> I added language and region support (see doc-string for "how") .. further I fixed some minor (type) issues and added AOL to the online docs .. can you please have a deeper look / thanks!
>
Very cool, thanks! I didn't know that Bing supports such search syntax actions like setting the language, that makes AOL much more useful.

From my testing it seems to reliably always return results that we're actually searching for (i.e. it respects the query, not like the Bing engine...), so this could mean that we finally can get proper Bing results for all languages again.

searx/engines/aol.py
return42: > This looks a bit hackish, 

Yes its a bit hackish .. I use this often when I need a list with minimal n-items in. Its a one-liner where if-len-conditions are to verbose.

> The second `None` is never needed

Good catch .. [fixed](https://github.com/searxng/searxng/compare/eb3b26ac671424950bee7d4172dd7c27ffa7e242..1531697c18aba57e61db42ef1c6bce6eb3c3ee03)

return42:
Sorry forgot to say: image search does not work on my side (this issue isn't related to my patches, even 3b9a8f15fef did not work) .. can you please have a deeper look / thanks!

Bnyro:
> Sorry forgot to say: image search does not work on my side (this issue isn't related to my patches, even [3b9a8f1](https://github.com/searxng/searxng/commit/3b9a8f15fef724786516ef6f4983c79faf38d904) did not work) .. can you please have a deeper look / thanks!
>
Until now, it just used the same code as for videos (which was by no means intended). Not sure what happened there, I probably forgot to commit some changes.

I could swear that it already worked previously...

Anyways, images are fixed now.

return42 (approved):
>  images are fixed now.

perfect :-)

Issue #5889, opened 2026-03-21
[feat] Generic engine for fetching pages through an external rendering service


klosowsk:
## Generic engine for fetching pages through an external rendering service

**Is your feature request related to a problem? Please describe.**

The xpath and json_engine generic engines work great for search providers that return results over plain HTTP. But a growing number of providers require JavaScript rendering to return meaningful results .
Currently, the only way to handle this is to write a full custom engine module per provider.

**Describe the solution you'd like**
A new generic engine (e.g. xpath_proxy) that works like xpath but delegates page fetching to a configurable external HTTP rendering service instead of fetching directly.
The engine would:
1. Build the search URL using SearXNG's existing template system (query, paging, language, time range, safe search - all handled the same as xpath)
2. POST that URL to a configurable rendering service endpoint
3. Parse the response - either raw HTML (using XPath config, same as xpath.py) or structured JSON if the rendering service returns pre-extracted results

My initial though was to be able to do something like this:
```
- name: example (rendered)
  engine: xpath_proxy
  proxy_url: http://my-rendering-service:8080
  search_url: https://example.com/search?q={query}&page={pageno}
  results_xpath: //div[@class="result"]
  url_xpath: .//a/@href
  title_xpath: .//h3
  content_xpath: .//p
  paging: true
  categories: [general]
```
The rendering service API contract would be minimal:
```
POST {proxy_url}/render
Request:  {"url": "https://...", "timeout": 15000}
Response: {"html": "...", "url": "..."}
````

The engine stays completely vendor-agnostic -  it doesn't care what the rendering service is (Splash, Browserless, any self-hosted solution).

**Describe alternatives you've considered**
- Custom engine per provider on my forks
- HTTP-level proxying: cannot render javascript
- Plugin system: cannot search other souces

**Additional context**
- **I have a working implementation and I'm happy to open a PR if there's interest.**

In discussion #5651, return42 noted that the ongoing problems with major web search providers raise doubts about whether SearXNG can continue avoiding JS-enabled solutions, and invited PRs with real browser integration. In that same thread, Material-Scientist demonstrated the pattern (Playwright-backed custom engines for Google, Imgur, etc.) - but each one required copy-pasting an entire engine and hardcoding the proxy calls, resulting in duplicated boilerplate.
[...]

PR #5890, opened 2026-03-21
[feat] engines: add xpath_proxy generic engine for external rendering services

klosowsk:
## What does this PR do?
- Add a new generic engine `xpath_proxy` that fetches pages through a configurable external HTTP rendering service
- Works like the `xpath` engine but delegates page fetching to an external service instead of making direct HTTP requests
- Supports two response modes: structured JSON (pre-extracted results) or raw HTML (parsed locally with XPath)
- Many search providers now require JavaScript rendering to return results - the existing `xpath` and `json_engine` can't handle this since they fetch pages directly via HTTP
- Without this, anyone needing JS-rendered results has to write a full custom engine per provider, duplicating the same proxy/fetch boilerplate
- Related to the discussion in #5651 where browser integration approaches were explored
- Ref: #5889
- Set up a rendering service that accepts `POST {"url": "..."}` and returns `{"html": "..."}` or `{"results": [...]}` - for example [camoufox-scraper](https://github.com/klosowsk/camoufox-scraper)
- Add an engine entry in `settings.yml`:
```yaml
- name: example (rendered)
  engine: xpath_proxy
  proxy_url: http://localhost:8080
  search_url: https://example.com/search?q={query}
  results_xpath: //div[@class="result"]
  url_xpath: .//a/@href
  title_xpath: .//h3
  content_xpath: .//p
  categories: [general]
  ```
  
- Search using !example test

Author notes
- ~400 lines, closely modeled on xpath.py
- No new dependencies
- I have this running in production with a self-hosted rendering service
- Happy to iterate on naming, config structure, or anything else based on feedback

AI Disclosure

### DIFF
--- a/searx/engines/xpath_proxy.py
+++ b/searx/engines/xpath_proxy.py
@@ -1,0 +1,404 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""The XPath Proxy engine is a *generic* engine that extends :ref:`xpath_engine`
+by fetching pages through a configurable external rendering service instead of
+making direct HTTP requests.
+
+This is useful for search providers that require JavaScript rendering to return
+meaningful results.  The rendering service is a black box — it receives a URL,
+renders it (e.g. using a headless browser), and returns either:
+
+- **Structured JSON**: pre-extracted results, suggestions, and CAPTCHA status.
+- **Raw HTML**: the engine parses it locally using XPath selectors (same as
+  :ref:`xpath_engine`).
+
+.. _XPath selector: https://quickref.me/xpath.html#xpath-selectors
+
+Configuration
+=============
+
+Proxy:
+
+- :py:obj:`proxy_url`
+- :py:obj:`proxy_endpoint`
+- :py:obj:`proxy_profile`
+- :py:obj:`render_timeout`
+
+Request (inherited from :ref:`xpath_engine`):
+
+- :py:obj:`search_url`
+- :py:obj:`lang_all`
+- :py:obj:`cookies`
+- :py:obj:`headers`
+
+Paging (inherited from :ref:`xpath_engine`):
+
+- :py:obj:`paging`
+- :py:obj:`page_size`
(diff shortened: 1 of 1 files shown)


Bnyro (commented):
> ~400 lines, closely modeled on xpath.py
> No new dependencies
>
Are you sure that you haven't used AI? It's not "closely modeled", it's copy and paste.

Anyways, if we decide to accept this, the xpath proxy should be integrated into the existing `xpath.py` instead of copying the whole file.

klosowsk:
@Bnyro 
Creating a separate file was intentional to avoid touching xpath.py during the initial implementation, but I agree that integrating into the existing engine is the right approach and my bad...

What do you think...  one plan would be to add an optional rendering proxy mode to xpath.py - a few new config variables (something like: `proxy_url`, `proxy_timeout`) that, when set, route the request through an external rendering service instead of fetching directly. When not set, xpath.py behaves exactly as it does today.

The proxy can a raw HTML (parsed by the existing XPath logic)
But - this is also cool - also a structured JSON (mapped directly to results, that also supports proxy)? but this could be a future engine with a separated proxy... (probably something to do later) 

Does this direction align with what you had in mind, or would you prefer a different integration approach?

return42:
> and claude helped on the implementation/validation...

and in the opening messages you accepted ..

> AI Disclosure
> 
> * [x]  I hereby confirm that I have not used any AI tools for creating this PR.

-----

Intelligence is based on the ability to think; machines are incapable of this .. ideally, humans should be able to think, but since the advertising for "machine learning" has appropriated the term "intelligence," this ability seems to have been completely lost in humans.

-----

> Does this direction align with what you had in mind, or would you prefer a different integration approach?

Familiarize yourself with the source code and take note of the rules of this forum.

**And above all, start thinking for yourself**: ["xpath"](https://github.com/searxng/searxng/blob/master/searx/engines/xpath.py) vs. ["JSON"](https://github.com/searxng/searxng/blob/master/searx/engines/json_engine.py)

PR #5892, opened 2026-03-23
[fix] google: switch to using "Google App" for Android useragent

### LINKED ISSUE #5867: Bug: google engine - HTTP 403
Branch: master
Version: 2026.3.17+2bb8ac17c

    make run

**What happened?**

Google appears to have changed its bot protection today; all requests to Google are being rejected with an HTTP 403 error. See also the "Google response" column on [searx.space](https://searx.space/)

**How To Reproduce**

    !go foo

**Technical report**

Error
  * Percentage: 100
  * Parameters: `('HTTP error 403 (suspended_time=180)',)`
  * File name: `searx/search/processors/online.py:205`
  * Code: `response = req(params["url"], **request_args)  # pyright: ignore[reportArgumentType]`

themg95:
## What does this PR do?

Switches the user agent (again) to bypass Google's shit

Fixes Google

Just search!

## Related issues

Fixes #5867

## AI Disclosure

- [x] I have used AI tools for working on the changes in this pull request and will attach a list of all AI tools I used and how I used them. I hereby confirm that I haven't used any other tools than the ones I mention below.

I just used ChatGPT Basic to fix a very small error in my regex

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -69,7 +69,7 @@ filter_mapping = {0: "off", 1: "medium", 2: "high"}
 
 # Suggestions are links placed in a *card-section*, we extract only the text
 # from the links not the links itself.
-suggestion_xpath = '//div[contains(@class, "ouy7Mc")]//a'
+suggestion_xpath = '//div[contains(@class, "gGQDvd iIWm4b")]//a'
 
 
 _arcid_range = string.ascii_letters + string.digits + "_-"
@@ -269,6 +269,15 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     ret_val["headers"]["Accept"] = "*/*"
     ret_val["headers"]["User-Agent"] = gen_gsa_useragent()
 
+    # Hardcoded default ENID Header required alongside the Android Google App
+    # User Agent
+    ret_val["headers"]["__Secure-ENID"] = (
+        "28.SE=II9FMkz92GewodDwKRBFsMISph7GsQs8JYLdXmAlprl6UcC02O2p7kfQlAWuwT"
+        "oygcrqHpmwQSH57b0c2kXfRfo35J8aV5FYSeUzYB67hqZQ2tZB7-o0hlTKwb5qMjn8Cf"
+        "w_AZ2s_6KIFMAl2goXGcXHSfgu4jwZOqShlHCcag0ppy_NnxJYWxpLkaeuGCICwWoIFJ"
+        "HP6Gy4BOkIEsl1N_k6F6jMF_OklE9qIubiyKkNaA"
+    )
+
     # Cookies
 
     # - https://github.com/searxng/searxng/pull/1679#issuecomment-1235432746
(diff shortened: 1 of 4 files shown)


themg95:
While I was digging through the requests that the Google App for Android sends, I noticed that, when searching from the home screen widget, the first request that gets sent is the search request. This made me think that this meant that the hardcoded starting security code (`__Secure-ENID`) is accepted server-side for this purpose before being exchanged for the JS challenge. And I was right!

return42 (approved):

searx/engines/google.py
```diff
@@ -269,6 +269,11 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     ret_val["headers"]["Accept"] = "*/*"
     ret_val["headers"]["User-Agent"] = gen_gsa_useragent()
 
+    # Hardcoded default ENID Header required alongside the Android Google App User Agent
+    ret_val["headers"][
+        "__Secure-ENID"
+    ] = "28.SE=II9FMkz92GewodDwKRBFsMISph7GsQs8JYLdXmAlprl6UcC02O2p7kfQlAWuwToygcrqHpmwQSH57b0c2kXfRfo35J8aV5FYSeUzYB67hqZQ2tZB7-o0hlTKwb5qMjn8Cfw_AZ2s_6KIFMAl2goXGcXHSfgu4jwZOqShlHCcag0ppy_NnxJYWxpLkaeuGCICwWoIFJHP6Gy4BOkIEsl1N_k6F6jMF_OklE9qIubiyKkNaA"
+
```
return42: > This made me think that this meant that the hardcoded starting security code (__Secure-ENID) is accepted server-side for this purpose before being exchanged for the JS challenge. And I was right!

not sure how long this ENID will work .. however, lets have a try / thanks for your work on this 👍

return42:
FYI I force pushed [diff](https://github.com/searxng/searxng/compare/e3b68ef667ea22e06dea66e09ff1b399aa38dec6..c9247964af7fc4a0b55b04b9e55a40f3cf36c563) to pass CI test (line to long iisue).

searx/engines/google.py
themg95: I think it'll work for quite a while because I think the ENID is hardcoded in the APK. The reason I think this is that I've tried this on two separate Android VMs and both sent the exact same ENID. Furthermore, while testing this, I got rate limited on 3-4 IP addresses and the ENID still works. So, I think if Google wanted to revoke this ENID they'd have to push an update to the Google App.

Though, I still think that a public instance owner should pull this commit and try it for a few days.

return42: I'm not familiar with the Google-App / I don't have a Google Play Store account ;-) .. What's new to me is that the ENID is sent in the HTTP header (!!!) .. I only know this value from cookies; when you use the Google website, the browser is prompted to set this cookie ..

grafik

vojkovic:
Works for me 👍 Images also work - nice work!

sharanjitdev:
hmm interesting i was also investigating mobile requests but not got so much time.

themg95:
> it seems only deepends on the UA:

I don't remember if I actually tested it without the ENID, lol. Oh well...

searx/engines/google.py
return42: >>> I think it'll work for quite a while because I think the ENID is hardcoded in the APK. 

>> it seems only deepends on the UA

> I don't remember if I actually tested it without the ENID, lol. 

@themg95 Please clarify this discrepancy. The code review was based on the information that the ENID is hard-coded in the APK and passed in the HTTP header, not in the cookie.

If this isn't the case, then all my other assumptions about it were incorrect.

- https://github.com/searxng/searxng/pull/5894

themg95: I still stand by what I said about the ENID, I think it's hardcoded but I just didn't realize that the server didn't check it...

PR #5894, opened 2026-03-24
[fix] google engine - don't set __Secure-ENID in the HTTP header

return42:
PR #5892 only deepends on the UA and the __Secure-ENID is not needed [1].

- [1] https://github.com/searxng/searxng/pull/5892#issuecomment-4114835195

Suggested-by: @yeyuchen198


Bnyro (approved):
Works without __Secure-ENID for me, lgtm 👍

PR #5896, opened 2026-03-24
[mod] remove the "Submit a new issue on Github" form

### LINKED ISSUE #5820: Lets remove the "Submit a new issue on Github" form
Submitting an issue on github isn't a end user task .. most issue reports comming from this form are not filled out, since end users are not familar with the needs of a developer community.

Image

I have the impression that there are a lot of end users who don't know what they're doing... who even create a GH account just to post a blank form... some post several (blank) issue reports in a row .. even if it's all done with the best of intentions, it's not only unnecessary, it's also annoying to have to clean up the issue tracker from these junk entries every day

return42:
Submitting an issue on GitHub isn't a end user task .. most issue reports coming from this form are not filled out, since end users are not familiar with the needs of a developer community.

- Closes: https://github.com/searxng/searxng/issues/5820


Bnyro (commented):
https://github.com/searxng/searxng/blob/2c1ce3bd37a2548d7e8a96127d19d7b6164d746b/client/simple/src/less/new_issue.less also needs to be removed I think? (and the line where it's imported: https://github.com/searxng/searxng/blob/2c1ce3bd37a2548d7e8a96127d19d7b6164d746b/client/simple/src/less/style.less#L14)

Btw, are the translations automatically removed, or do we need to do that as well? (https://github.com/searxng/searxng/blob/2c1ce3bd37a2548d7e8a96127d19d7b6164d746b/searx/translations/pa/LC_MESSAGES/messages.po#L447)

return42:
> also needs to be removed I think? (and the line where it's imported: 

Sure, thanks for pointing out :+1: .. [fixed](https://github.com/searxng/searxng/compare/a65bc140ac3478fc3cf423322a8656419639f14e..6c6e2efb467800ac855047f214d374e79414448f)

> Btw, are the translations automatically removed, or do we need to do that as well?

Should be automatically removed by ...

https://github.com/searxng/searxng/blob/c4f51aa4ac3a232bd19fbc828cac7b3c5282440d/utils/lib_sxng_weblate.sh#L126-L129

...

https://github.com/searxng/searxng/blob/c4f51aa4ac3a232bd19fbc828cac7b3c5282440d/utils/lib_sxng_weblate.sh#L152-L156

Bnyro (approved):

vojkovic (approved):

PR #4779, opened 2025-05-13
[fix] engine archlinux: avoid Anubis challenge by User-Agent "SearXNG"

return42:
Of the archlinux wikis only wiki.archlinux.org has a has Anubis challenge.

About Anubis[1]:

> Anubis decides to present a challenge using this logic:
>
> - User-Agent contains "Mozilla"
> ...
> This should ensure that git clients, RSS readers, and other low-harm clients
> can get through without issue ..

[1] https://github.com/TecharoHQ/anubis/blob/6c0ff3f4d5e7a66ea85e5873e6ee4c28692091d7/docs/docs/design/how-anubis-works.mdx#challenge-presentation

Suggested-by: @unixfox https://github.com/searxng/searxng/issues/4646#issuecomment-2855322406
Closes: https://github.com/searxng/searxng/issues/4646

Issue #4646, opened 2025-04-21
Bug: arch linux wiki engine


BobBoba:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2025.4.18+5ae3b3f17

**How did you install SearXNG?**
installed SearXNG using using searxng-docker

**How To Reproduce**
any search using arch linux wiki engine returns http timeout error below

**Technical report**

Error
  * Error: httpx.TimeoutException
  * Percentage: 35
  * Parameters: `(None, None, None)`
  * File name: `searx/search/processors/online.py:116`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

Error
  * Error: httpx.HTTPStatusError
  * Percentage: 35
  * Parameters: `('502', 'Bad Gateway', 'wiki.archlinux.org')`
  * File name: `searx/search/processors/online.py:116`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

return42:
Seems EN ``!arch_linux_wiki pacman :en`` has timeouts while DE ``!arch_linux_wiki pacman :de`` not. 

---- 

update: found the cause --> https://wiki.archlinux.org has a bot blocker

> ![Image](https://github.com/user-attachments/assets/a7527b6a-985f-43b2-9538-728d5fc97afe)
>
> You are seeing this because the administrator of this website has set up [Anubis](https://github.com/TecharoHQ/anubis) to protect the server against the scourge of [AI companies aggressively scraping websites](https://thelibre.news/foss-infrastructure-is-under-attack-by-ai-companies/). This can and does cause downtime for the websites, which makes their resources inaccessible for everyone.
>
> Anubis is a compromise. Anubis uses a [Proof-of-Work](https://anubis.techaro.lol/docs/design/why-proof-of-work) scheme in the vein of [Hashcash](https://en.wikipedia.org/wiki/Hashcash), a proposed proof-of-work scheme for reducing email spam. The idea is that at individual scales the additional load is ignorable, but at mass scraper levels it adds up and makes scraping much more expensive.
>
> Ultimately, this is a hack whose real purpose is to give a "good enough" placeholder solution so that more time can be spent on fingerprinting and identifying headless browsers (EG: via how they do font rendering) so that the challenge proof of work page doesn't need to be presented to users that are much more likely to be legitimate.
>
[...]

PR #5907, opened 2026-03-25
[fix] unsplash: fix engine due to anubis bot blocking

Bnyro:
Unsplash started using [Anubis](https://anubis.techaro.lol/)
for blocking crawlers. Therefore, requests using common
user agents (e.g. Firefox, Chrome) must pass a JavaScript
challenge.

However, other user agents seem unaffected for now, hence
settings the UA to something different does still work.

### DIFF
--- a/searx/engines/unsplash.py
+++ b/searx/engines/unsplash.py
@@ -4,6 +4,8 @@
 from urllib.parse import urlencode, urlparse, urlunparse, parse_qsl
 from json import loads
 
+from searx.utils import searxng_useragent
+
 # about
 about = {
     "website": 'https://unsplash.com',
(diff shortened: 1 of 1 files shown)


return42 (approved):
> However, other user agents seem unaffected for now, hence
> settings the UA to something different does still work.

TBH I have never understood how the Anubis challenge is supposed to provide protection --> [Anubis decides to present a challenge using this logic: User-Agent contains "Mozilla"](https://github.com/TecharoHQ/anubis/blob/dbd64e0f4f23c37476b40b332781c676ffa413ef/docs/docs/design/how-anubis-works.mdx#challenge-presentation)

-----

Related:

- https://github.com/searxng/searxng/pull/4779
- https://github.com/searxng/searxng/issues/4646#issuecomment-2855322406

Bnyro:
Yeah, the default configuration is quite weird - I'm using [`go-away`](https://git.gammaspectra.live/git/go-away/) personally for some services and require challenges for all user agents, the Mozilla thing doesn't make much sense to me either.

Bnyro:
I tried to use the same patch for an Android app I maintain (https://github.com/you-apps/WallYou), but Unsplash blocks me no matter which user agent I set (even if I keep it blank) :/

Doing the same with `xh` always gives me a proper response.

I hate these techniques...

return42:
> Doing the same with `xh` always gives me a proper response.

xh sends `Accept-Encoding` header by default / may this header is missed in your app's requests?

    GET / HTTP/1.1
    Accept: */*
    Accept-Encoding: gzip, deflate, br, zstd
    Connection: keep-alive
    User-Agent: xh/0.24.0

Bnyro:
Good idea, that's worth a try, thanks!

Bnyro:
Yep, it's been `Accept: */*`. Thanks :)

PR #5908, opened 2026-03-25
[enh] py: add default HTTP headers

inetol:
Set default HTTP headers from the [searxng-docker repo](https://github.com/searxng/searxng-docker/blob/master/Caddyfile) on both dynamic and static assets.

## AI Disclosure

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -113,10 +113,7 @@ server:
   # Is overwritten by ${SEARXNG_METHOD}
   method: "POST"
   default_http_headers:
-    X-Content-Type-Options: nosniff
-    X-Download-Options: noopen
-    X-Robots-Tag: noindex, nofollow
-    Referrer-Policy: no-referrer
+    Server: SearXNG
 
 valkey:
   # URL to connect valkey database. Is overwritten by ${SEARXNG_VALKEY_URL}.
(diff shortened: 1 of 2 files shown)


return42:
Here are a few thoughts that spontaneously come to mind on the topic ..

------

Related to:

- https://github.com/searxng/searxng/issues/140

> This issue suggests to define the default headers directly a place that is keep up to date: the code, so the configuration is not spread into different locations.

-------

Which headers need to be set and how they need to be set can depend, among other things, on which features are used in SearXNG; for example [CSP](https://github.com/search?q=repo%3Asearxng%2Fsearxng%20label%3A%22CSP%22&type=issues), `img-src` can be more restricted if the image proxy is enabled.

inetol:
The thing is I don't know how to handle this without making it bad, so I'm just copypasting it.

Besides, we've already had some issues with CSP in the past, I'm not too keen on having to reapply restrictions.

return42:
We don't have to solve all in one .. but I'm just think a little:

If I'm not wrong; with this PR we have identical headers implemented in 

Whitenoise `add_headers_function=static_headers` and in 

```python
@app.after_request
def add_default_headers(response: flask.Response):
```

May its better to have (to start with) a new python module with a single point of definition?

PR #5910, opened 2026-03-26
Fix Chinese character encoding in search engines (bing, baidu)

15802221580:
## Problem

When searching with Chinese characters directly in URL (not URL-encoded), SearXNG returns irrelevant or garbled results.

**Example:**
- Query `黄仁勋` returns French content instead of Chinese results about Jensen Huang
- Query `人工智能` returns garbled content

## Root Cause

HTTP request bytes containing UTF-8 encoded Chinese characters are incorrectly decoded as Latin-1:

```
Original: 黄 → UTF-8: \xe9\xbb\x84 → Wrongly decoded: é» → Re-encoded: %C3%A9%C2%BB%C2%84 (wrong)
Expected: 黄 → UTF-8: \xe9\xbb\x84 → Correct encoding: %E9%BB%84
```

### bing.py
Detect double-encoding in `request()` and convert back to proper UTF-8:
```python
if any(127 < ord(c) < 256 for c in query):
    query = query.encode("latin-1").decode("utf-8")
```

### baidu.py
Extract correctly encoded query from response URL and add optional curl_cffi for browser impersonation.

## Testing

| Query | Before | After |
|-------|--------|-------|
| 黄仁勋 | French content ❌ | Correct Chinese results ✅ |
| NVIDIA | Correct ✅ | Correct ✅ |
| 人工智能 | Garbled ❌ | Correct Chinese results ✅ |

## Files Changed

- `searx/engines/bing.py` - Add encoding fix in request()
- `searx/engines/baidu.py` - Add encoding fix and curl_cffi support

### DIFF
--- a/searx/engines/baidu.py
+++ b/searx/engines/baidu.py
@@ -1,13 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Baidu_
+"""Baidu"""
 
-.. _Baidu: https://www.baidu.com
-"""
-
-# There exits a https://github.com/ohblue/baidu-serp-api/
-# but we don't use it here (may we can learn from).
-
-from urllib.parse import urlencode
+from urllib.parse import urlencode, unquote
 from datetime import datetime
 from html import unescape
 import time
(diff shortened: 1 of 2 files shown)


return42:
Please ..

- pay attention to the rules regarding [CONTRIBUTING](https://github.com/searxng/searxng/blob/master/CONTRIBUTING.rst).
- Fill out the template for PRs carefully and
- Accept our AI policies.
- Read our ['Development Quickstart'](https://docs.searxng.org/dev/quickstart.html) and
- clean up your code and the commit history.

We are temporarily closing the PR to give you the opportunity for improvement.

BTW: project's language is English .. this includes even more the comments in the code!

PR #5920, opened 2026-03-28
[campaign] Keep Android Open - https://keepandroidopen.org/

return42:
## What does this PR do?

Support the https://keepandroidopen.org/ campaign by adding the [countdown banner](https://keepandroidopen.org/banner) to our online docs

SearXNG stands for user rights

    make docs.live

grafik

## AI Disclosure

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,0 +1,6 @@
+{% extends "!layout.html" %}
+
+{% block extrahead %}
+  {{ super() }}
+  <script defer src="https://keepandroidopen.org/banner.js?size=minimal&animation=off"></script>
+{% endblock %}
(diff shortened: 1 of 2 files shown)


inetol (commented):
Sketchy script. Anyway, I think the organization shouldn't involve itself and its users in matters that don't concern it.

vojkovic (commented):
I totally agree with the message and it is a noble idea but I have to side with Ivan on this. Probably isn't the place for these matters given SearXNG is not an android app.

Bnyro:
We (LibreTube) wanted to sign their letter and asked the KeepAndroidOpen people to do so, but they rejected our PR (https://github.com/keepandroidopen/keepandroidopen.github.io/pull/157) and ghosted our e-mails.

So I agree with the others - the idea behind the initiative is right, but the initiative itself is very sketchy and SearXNG is not really part of the Android ecosystem.

Anyways, I think it's important to spread the word about Android's current path towards becoming a closed platform, thank you for doing so @return42 !

PR #5923, opened 2026-03-28
Handle any exception when git is not available

koush:
## What does this PR do?

git not being available (downloaded as tarball) causes searxng to not install. the version code failing shouldn't prevent install. git unavailable causes a FileNotFoundException and not whatever is being expected here.

pip install https://github.com/searxng/searxng/archive/refs/heads/master.zip without git available

## AI Disclosure

- [X] I have used AI tools for working on the changes in this pull request and will attach a list of all AI tools I used and how I used them. I hereby confirm that I haven't used any other tools than the ones I mention below.

I used opencode to track the issue down before fixing it.

### DIFF
--- a/searx/version.py
+++ b/searx/version.py
@@ -95,12 +95,12 @@ def get_information() -> tuple[str, str, str, str, str]:
 
     try:
         version_string, version_tag, docker_tag = get_git_version()
-    except subprocess.CalledProcessError as ex:
-        logger.error("Error while getting the version: %s", ex.stderr)
+    except Exception as ex:
+        logger.error("Error while getting the version: %s", ex)
     try:
         git_url, git_branch = get_git_url_and_branch()
-    except subprocess.CalledProcessError as ex:
-        logger.error("Error while getting the git URL & branch: %s", ex.stderr)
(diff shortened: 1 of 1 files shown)


return42:
> pip install https://github.com/searxng/searxng/archive/refs/heads/master.zip without git available

This installation method is not supported, please read the docs: https://docs.searxng.org/admin/installation.html

koush:
I know, I am fixing the issues that prevent searxng from being used as a package via pip and as a dependency import in other projects, rather than requiring a standalone server. That’s working and this was the one of a few changes I needed to make.

return42:
> I know, I am fixing the issues that prevent searxng from being used as a package via pip and as a dependency import in other projects, rather than requiring a standalone server.

SearXNG is a set of services not a python package (otherwise it would be listed on pypi, but it isn't) ... you should NEVER install the searxng code into your user's python environment, it was not made for and will damage your python environment. 

NEVER use the searxng code as a dependency in your python projects!!!

koush:
Obviously using a virtualenv but yeah agreed.

PR #5758, opened 2026-02-20
[fix] complete overhaul of the DuckDuckGo engines

return42:
DDG has reimplemented its bot protection and the DDG engines "images", "news" and "videos" no longer work in SearXNG and DDG-Web access often ends with a CAPTCHA.

Related:

- issue #4824
- https://github.com/ggfevans/searxng/blob/mod-sidecar-harvester/docs/ddg-bot-detection-research.md

-------

I test this PR in my public instance ...

State of this draft:

- DDG-Web, -News, -Videos seem to work so far (with some hick-ups)
- DDG-Images often ends in a CAPTCHA (when requesting page numbers >1) 

------

Please read the doc-string (and commit messages) to understand the details of DDG's bot detection and how we try to come around.

PR #5931, opened 2026-03-29
[fix] Bug: duckduckgo engine, Error: SearxEngineCaptchaException

nicolasdato:
Adding the Accept and changing the Referer.
Making a random choice of User-Agent, Accept and Accept-Language.

Looks like one can bypass the block by changing those headers.

## What does this PR do?

This PR makes a random choice of UA, Accept, and Accept-Language for each duckduckgo search. Maybe this can be improved with more options for each header.

I don't know if this makes sense, but it looks like it works. I still get blocked if I make several searches at the same time, but I feel that it's not as frequent as before, and that I'm unblocked automatically reasonably fast without the need to use the duckduckgo website.

I'm running searxng locally. From time to time, duckduckgo responded with the captcha error. The only way to unblock me was to use their website. However, I discovered that making 1 simple search with curl was enough. So, I used searxng until I got the CAPTCHA error for duckduckgo, then I run one curl command, and that solved the CAPTCHA error.

So I thought, there shouldn't be anything special with the website, if using curl solved the issue this could be also be solved easy in searxng.

What I think is that duckduckgo was blocking me because searxng was using the same combination of UA, Accept, Accept-Languages, and don't know what else. So by making the curl call I was changing the some of those headers and duckduckgo would unblock me.

I got the curl query from the browser and removed headers and I ended up using:

```curl 'https://html.duckduckgo.com/html/'      --compressed    -X POST         -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:148.0) Gecko/20100101 Firefox/148.0'      -H 'Accept: text/html,application/xhtml+xml,*/*;q=0.8'  -H 'Accept-Language: en,en-US;q=0.9'    -H 'Accept-Encoding: gzip, deflate'     -H 'Referer: https://html.duckduckgo.com/html/'      -H 'Content-Type: application/x-www-form-urlencoded'    -H 'Connection: keep-alive'     -H 'DNT: 1'     --data-raw 'q=asd&b='```

Duckduckgo was constantly blocking me, and that's annoying.

## Related issues

Improvement for #4824

## AI Disclosure

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -173,6 +173,8 @@ import typing as t
 import babel
 import lxml.html
 
+from random import choice
+
 from searx import locales
 from searx.enginelib import EngineCache
 from searx.enginelib.traits import EngineTraits
@@ -215,8 +217,6 @@ time_range_dict: dict[str, str] = {"day": "d", "week": "w", "month": "m", "year"
 _CACHE: EngineCache = None  # pyright: ignore[reportAssignmentType]
 """Persistent (SQLite) key/value cache that deletes its values after ``expire``
 seconds."""
-
-_HTTP_User_Agent: str = gen_useragent()
 
 
 def get_cache() -> EngineCache:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: searx/engines/duckduckgo.py [hunk 1]
```diff
@@ -378,12 +378,26 @@ def request(query: str, params: "OnlineParams") -> None:
 
     # The vqd value is generated from the query and the UA header. To be able to
     # reuse the vqd value, the UA header must be static.
-    headers["User-Agent"] = _HTTP_User_Agent
-    headers["Referer"] = "https://html.duckduckgo.com/"
+    headers["User-Agent"] = gen_useragent()
```
return42, line 381: > This PR makes a random choice of UA

Are you sure the UA should be changed? ... In #5758, I set the UA statically because (as you can see in line 379) it’s factored into the calculation of the vqd value ... and if the UA header and the vqd value don’t match, then the SearXNG IP will definitely be classified as a bot.

It’s possible that a curl command could resolve this block... your observation is very interesting... I didn’t know that, for example... maybe we should look further into this direction to see how we can resolve these blocks.

nicolasdato:
Well, unfortunately I think I posted too early. I payed more attention, and tried changing the UA and also with the same UA (as per @return42 comment). Now I feel I'm beeing blocked as always.

What does help is using the curl command. With debug logs, after being blocked I run curl using the same last headers as searxng and it worked and unblocked me. So maybe changing headers isn't helping.

PR #5936, opened 2026-03-31
[fix] tubearchivist engine: add trailing slash to API URL

AgentScrubbles:
## What does this PR do?

The base URL for tubearchivist's search now requires a trailing backslash (`/`). This is required because TubeArchivist rewrote the backend to Django. Django will return a 404 without the trailing slash.

The TubeArchivist engine is broken actively without this trailing backslash since the TubeArchivist update.

Have an existing TubeArchivist instance.
Use the existing instance, note that it breaks, in the UI it reports a 404 error.
Adding the slash then resolves this issue.

## AI Disclosure

### DIFF
--- a/searx/engines/tubearchivist.py
+++ b/searx/engines/tubearchivist.py
@@ -127,7 +127,7 @@ def request(query, params):
         return False
 
     args = {'query': query}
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
You're right https://docs.tubearchivist.com/api/docs/, also lists `/api/search/` as correct route.

Thank you!

PR #5943, opened 2026-04-02
DDG: Restoring Sec-Fetch-* headers

nicolasdato:
## What does this PR do?

Looks like those headers are needed to bypass the bot detection.

Those were removed in 2e6eeb1 saying that "Sec-Fetch-* headers are already set by the OnlineProcessor". But in 029b74e those headers were removed from OnlineProcessor.

DDG is blocking requests pretty fast, just two consecutive searches without a few seconds is enought to get blocked

Open several tabs (like 4) to search simultaneously.

## Related issues

Should fix or improve #4824 

## AI Disclosure

### DIFF
--- a/searx/engines/duckduckgo.py
+++ b/searx/engines/duckduckgo.py
@@ -379,6 +379,12 @@ def request(query: str, params: "OnlineParams") -> None:
     # The vqd value is generated from the query and the UA header. To be able to
     # reuse the vqd value, the UA header must be static.
     headers["User-Agent"] = _HTTP_User_Agent
+
+    headers["Sec-Fetch-Dest"] = "document"
+    headers["Sec-Fetch-Mode"] = "navigate"
+    headers["Sec-Fetch-Site"] = "same-origin"
+    headers["Sec-Fetch-User"] = "?1"
+
     headers["Referer"] = "https://html.duckduckgo.com/"
 
     ui_lang = params["searxng_locale"]


Bnyro (commented):
If these headers are set, pagination doesn't work (at least that's been the case two months ago). See https://github.com/searxng/searxng/pull/5758#pullrequestreview-3834221131.

Though if you find a way to set these headers and make pagination work, we should definitely go for this.

nicolasdato:
Hi @Bnyro, thanks for the heads up.

I've just tried using just the ddg engine and going to the second, third, and fourth page, and it looks like it's working. I've also tried with images (!ddi), but I didn't add the headers to duckduckgo_extra.py, and it looks like it's working too.

I can't be sure if searxng isn't skipping something or not, it looks like I get similar results as using the html.duckduckgo.com website and going to the next page there, although some items are in a different order, or missing. But even repeating search in the website doesn't give me the same results, so maybe some items are missing because for that particular response to searxng was different from the response to the website.

You said:
> but this still got me rate-limited starting at page 2.

What does it mean? How can I try to replicate the pagination issue? Maybe I'm not seeing something.

What I've found was that (apparently) just using `Sec-Fetch-Mode: navigation` is enough to allow me to search several times in a row. However, I guess that technically the four Sec-Fetch-* headers always go together. Maybe if just Sec-Fetch-Mode works for you and pagination works, maybe we should use that to improve #4824 .

Bnyro (approved):
> I've just tried using just the ddg engine and going to the second, third, and fourth page, and it looks like it's working. I've also tried with images (!ddi), but I didn't add the headers to duckduckgo_extra.py, and it looks like it's working too.
>
I've tested it as well now. You're right, search works normally without rate-limits, my bad!

The rate-limits only affected `duckduckgo_extra`, I didn't notice that you only changed the `duckduckgo` engine here.

Thank you!

PR #5954, opened 2026-04-05
[fix] pexels: circumvent botdetection by passing referer header

Bnyro:
As a side effect, Cloudscraper is no longer needed.
It probably only ever worked by setting the correct request headers,
so we don't really need it since we can just set the right request
headers and ciphersuites ourselves.

### DIFF
--- a/searx/engines/pexels.py
+++ b/searx/engines/pexels.py
@@ -25,6 +25,11 @@ about = {
 
 base_url = 'https://www.pexels.com'
 categories = ['images']
+
+api_key = "H2jk9uKnhRmL6WPwh89zBezWvr"
+"""
+Fallback API key to use when SearXNG fails to automatically extract one from the website.
+"""
 results_per_page = 20
 
 paging = True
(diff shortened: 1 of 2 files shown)


vojkovic:
Looks like a good change, one less dependency would be nice.

Unfortunately, I tested this PR on both on a Github Codespace and at home on a residential ip and got a 403 error / access denied with the pexel engine both times.

This is different to what is currently being experienced (before this pr) which is a `SearxEngineAPIException`.

Bnyro:
Hmm, it's no longer working for me either, I'll look into it again.

Bnyro:
@vojkovic can you please test it again?

I found an interesting repository about Cloudflare: https://github.com/luminati-io/bypass-cloudflare

vojkovic:
> @vojkovic can you please test it again?

Unfortunately still getting 403s - does it all work fine for you?

Will read through the rest of that repo tomorrow. I'm not sure if it's still possible to scrape Cloudflare reliably without a browser anymore.

Bnyro:
Okay, next try

> does it all work fine for you?
>
Yes, it did. Though I just noticed that we cache the server key, so most of my tests have been useless because I didn't disable the cache ...

I'm confident it works more reliable now - it seems like I previously removed some headers that I thought were useless, but were actually needed :)

Bnyro:
If it still doesn't work, I'm out.

vojkovic:
Still receiving 403s. Tried again at home and on a fresh codespace.

inetol:
Same here

return42 (approved):
When I tested this PR a few days ago, I had the same problems as @vojkovic and @inetol (HTTP 403). 

Today I tested the PR again and am now getting the correct results.

searx/engines/pexels.py
```diff
@@ -27,6 +25,11 @@
 
 base_url = 'https://www.pexels.com'
 categories = ['images']
+
+api_key = "H2jk9uKnhRmL6WPwh89zBezWvr"
```
Bnyro, line 29: I've additionally added a fallback API key (taken from https://git.lolcat.ca/lolcat/4get/src/branch/master/scraper/pexels.php#L105, idk where they got it from).

Not sure how long that will last, but apparently this key has been working for a while now.

PR #5960, opened 2026-04-09
[mod] result templates: move iframe to macro and fix page rendering for non-YouTube links

Bnyro:
## What does this PR do?
- This PR moves the `iframe` logic into a macro, so that `videos.html` and `general.html` both can benefit from the workaround to fix YouTube results by @return42 in https://github.com/searxng/searxng/pull/5858
- It also fixes that only YouTube videos contained the closing `>` after `<iframe border="0" ...`, the regression has been caused by https://github.com/searxng/searxng/pull/5858

Currently, the page breaks if there's any non-YouTube Iframe:
Screenshot from 2026-04-09 22-09-48

Here, the page ends in the middle of the results and the footer and page number selector are not visible.

- search for videos with iframes that are not from YouTube
- see that the page doesn't break

## AI Disclosure

### DIFF
--- a/searx/templates/simple/macros.html
+++ b/searx/templates/simple/macros.html
@@ -67,3 +67,13 @@
          class="checkbox-onoff"{{- ' ' -}}
          {%- if checked -%} checked{%- endif -%}/>
 {%- endmacro -%}
+
+<!-- iframe that additionally sets some extra feature attrs for videos -->
+{%- macro iframe(iframe_src) -%}
+  <iframe data-src="{{iframe_src}}" frameborder="0" allowfullscreen
+    {% if result.parsed_url.hostname in ("www.youtube.com", ) -%}
+    allow="picture-in-picture" referrerpolicy="origin"
+    {%- endif -%}
+    >
+  </iframe>
+{%- endmacro -%}
--- a/searx/templates/simple/result_templates/default.html
+++ b/searx/templates/simple/result_templates/default.html
@@ -1,4 +1,4 @@
-{% from 'simple/macros.html' import result_header, result_sub_header, result_sub_footer, result_footer with context %}
+{% from 'simple/macros.html' import iframe, result_header, result_sub_header, result_sub_footer, result_footer with context %}
 
 {{ result_header(result, favicons, image_proxify) -}}
 {{- result_sub_header(result) -}}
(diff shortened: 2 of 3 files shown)


return42:
https://github.com/searxng/searxng/blob/08ef7a63d7ffd278dddd68cbb643f30d969a4329/searx/result_types/_base.py#L357-L358

As far as I know, we only use iframe_src for embedded videos... the name iframe_src has therefore always been somewhat misleading (because it's too general).

TBH I'm not a big fan of this (big) "macro" file.

I while ago I started to implement https://github.com/searxng/searxng/tree/master/searx/templates/simple/elements

May be it is a good time to implement a `embedded-video.html`? .. and we might also want to rename `iframe_src` to `video_src` (or something similar) at this time?

-----

The iframe_src was hacked into the `default.html` template at some point (mea culpa)... I'm no longer happy with it... I also don't know why there should be "default" results that include a video... shouldn't they be video results anyway? That would also simplify the layout considerably.

Bnyro:
> The iframe_src was hacked into the default.html template at some point (mea culpa)... I'm no longer happy with it... I also don't know why there should be "default" results that include a video... shouldn't they be video results anyway? That would also simplify the layout considerably.
>
In that case, we could just remove the `iframe_src` from the general template and keep the embedded video logic inside `videos.html` because it's not used anywhere else.

I'm not sure if that's a good idea though, I've just done a quick search to see where `iframe_src` is used in `default.html` and looked for other edge cases - it seems like IFrame is also used for other media than videos:
- Audio files (Deezer, Spotify, ...): e.g. https://github.com/searxng/searxng/blob/67af4894d4449d2343b1f690bcaa9832ff031335/searx/engines/deezer.py#L57
- Uses general template, although this should probably just use `videos.html`: https://github.com/searxng/searxng/blob/67af4894d4449d2343b1f690bcaa9832ff031335/searx/engines/acfun.py#L105
- Tagesschau uses the default template, also if the result contains a video. I think that is actually reasonable, because there often still is a news article, so the result is not only a video and I therefore wouldn't put it into `videos.html` - https://github.com/searxng/searxng/blob/67af4894d4449d2343b1f690bcaa9832ff031335/searx/engines/tagesschau.py
- many more probably

return42 (approved):
> I'm not sure if that's a good idea though, I've just done a quick search to see where `iframe_src` is used in `default.html` and looked for other edge cases - it seems like IFrame is also used for other media than videos:

Thanks for explanation .. very helpful .. after thinking twice I agree with your analysis / conclusion:

> * I think that is actually reasonable, because there often still is a news article, so the result is not only a video and I therefore wouldn't put it into `videos.html` -

-----

FYI: last force push was just a rebase on current master.

PR #5848, opened 2026-03-12
docs: add AI policy

Bnyro:
## What does this PR do?
- This adds an AI policy, based on the ones from other projects (linked below).
- Basically, we allow AI usage to a certain extent, but the human definitely is fully responsible for the code and must understand everything.
- I'm personally in favor of banning LLMs completely from the project, but as discussed internally, we couldn't find a general agreement there.

## Why is this change important?
- We're seeing an increasing amount of low effort AI-generated PRs, that don't match our code structure at all, touch unrelated code parts or do other bad things

Related policies at other projects:
- https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md
- https://github.com/rnk/llvm-project/blob/tool-policy/llvm/docs/AIToolPolicy.md
- https://github.com/matplotlib/matplotlib/blob/main/doc/devel/contribute.rst#L189

Overview of general attitude towards AI in FOSS projects:
- https://oss-ai-policies.netlify.app/

Examples of low-effort PRs:
- https://github.com/searxng/searxng/pull/5841
- https://github.com/searxng/searxng/pull/5824
- https://github.com/searxng/searxng/pull/5840
- https://github.com/searxng/searxng/pull/5602

PR #5973, opened 2026-04-12
[mod] improves the readability of PR and issue templates 

return42:
### What does this PR do?

Adjustments to the PR & issue templates / Blank issues are no longer allowed (`blank_issues_enabled: false`)

Improves the readability of PR and issue templates to make filling them out a little easier (at least visually in Markdown).

C&P the markdown into a GH issue / PR and click on the "Preview" tab.

### Related issues

- https://github.com/searxng/searxng/pull/5848

### Code of Conduct

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,41 +1,50 @@
 ---
-name: Bug report
-about: Report a bug in SearXNG
-title: ''
-labels: bug
-assignees: ''
-
+name: "Bug report"
+about: Report a bug in SearXNG"
+labels: ["bug"]
+type: "bug"
 ---
-<!-- PLEASE FILL THESE FIELDS, IT REALLY HELPS THE MAINTAINERS OF SearXNG -->
 
-**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
-<!-- If you are running on master branch using git execute this command
-in order to fetch the latest commit ID:
-```
-git log -1
-```
-If you are using searxng-docker then look at the bottom of the SearXNG page
-and check for the version after "Powered by SearXNG"
+_Replace this placeholder with a meaningful and precise description of the bug._
 
-Please also stipulate if you are using a forked version of SearXNG and
-include a link to the fork source code.
+<!-- FILL IN THESE FIELDS .. and delete the comments after reading.
+
+     Use Markdown for formatting ->  https://www.markdowntools.io/cheat-sheet
 -->
-**How did you install SearXNG?**
-<!-- Did you install SearXNG using the official wiki or using searxng-docker
-or manually by executing the searx/webapp.py file? -->
-**What happened?**
-<!-- A clear and concise description of what the bug is. -->
 
-**How To Reproduce**
-<!-- How can we reproduce this issue? (as minimally and as precisely as possible) -->
+### How To Reproduce?
 
-**Expected behavior**
+<!-- How can we reproduce this issue? (as minimally and as precisely as
+     possible) -->
+
+### Expected behavior
+
 <!-- A clear and concise description of what you expected to happen. -->
 
-**Screenshots & Logs**
+### Screenshots & Logs
+
 <!-- If applicable, add screenshots, logs to help explain your problem. -->
 
-**Additional context**
+### Version of SearXNG
+
+<!-- Commit number if you are using on master branch and stipulate if you forked
+     SearXNG -->
+
+<!-- Look at the bottom of the SearXNG page and check for the version after
+     "Powered by SearXNG" If you are using a forked version of SearXNG include a
+     link to the fork source code. -->
+
+### How did you install SearXNG?
+
+<!-- Did you install SearXNG using the official documentation or using
+     searxng-docker? -->
+
+### Additional context
+
 <!-- Add any other context about the problem here. -->
 
-- [ ] I read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and hereby confirm that this issue conforms with the policy.
+### Code of Conduct
+
+[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst
+
+- [ ] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,41 +1,49 @@
...
+
 <!-- Add any other context about the problem here. -->
 
-- [ ] I read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and hereby confirm that this issue conforms with the policy.
+### Code of Conduct
+
+[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst
+[basic formatting syntax]: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax
+
+- [ ] I have mastered [basic formatting syntax].
```
Bnyro: I'm not sure if we really need that, most people probably only write plain text and don't need to know about all this formatting. That also makes it more complicated for new people with little GitHub knowledge to open issues.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,41 +1,49 @@
...
-name: Bug report
-about: Report a bug in SearXNG
-title: ''
-labels: bug
-assignees: ''
-
+name: "Bug report"
+about: Report a bug in SearXNG"
+title: "[bug]"
+labels: ["bug"]
```
Bnyro, line 4:
```suggestion
labels: ["bug"]
type: bug
```
We could also use the new issue types from GitHub here, looking at https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms, it seems like we can automatically assign them this way.

PULL_REQUEST_TEMPLATE.md
Bnyro, line 1: I feel like this template is still too complicated. In my case, I usually only need
- What does this PR do?
- How to test this PR locally?
- Related issues

"Why is this change important?"  has a big overlap with "what does this PR do" and is often quite unnecessary, and "Author's checklist" has usually a big overlap with "How to test this PR locally" (or I don't really understand what Author's checklist is for)

Bnyro (commented):
Looks like really good changes :+1:  When I created issues in the last time, I also noticed that the templates are possibly a bit too verbose / complicated

.github/ISSUE_TEMPLATE/bug-report.md
return42: A checkbox is probably overkill, but I did want to have a note about Markdown in the template, which is why I added a link to a cheat sheet at the top of the comment [diff](https://github.com/searxng/searxng/compare/638e5e4c518f971488f6858b2ca63a79bb1b96cc..76ab985e0e86cf7b212d4c3be4dbfc89f87a954e)

PULL_REQUEST_TEMPLATE.md
return42: Yes, I agree with you, I have now only left the most necessary sections in the template (and in the top I added a comment with the link to the Markdown cheat-sheet).

.github/ISSUE_TEMPLATE/bug-report.md
return42: Good catch, I removed the `title` (prefix) and added a value for the `type` [diff](https://github.com/searxng/searxng/compare/992304286c5dad9b133df4889ba9d54ead736439..c301c14ba11d293a00d933c07e226549a30cf748)

---

> We could also use the new issue types from GitHub here,

TBH: contrary to popular opinion, in my opinion YAML is unsuitable for process (or form) control .. it's hard to read, complicated to test .. to name just two points of criticism.  If someone would like such a form control, then they can send another PR for it.

_We are digging deeper and deeper into GH addictions. In the long term, I would personally like to move away from GH._

Bnyro (approved):

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -1,41 +1,50 @@
 ---
-name: Bug report
-about: Report a bug in SearXNG
-title: ''
-labels: bug
-assignees: ''
-
+name: "Bug report"
+about: Report a bug in SearXNG"
+labels: ["bug"]
```
inetol, line 4: Can we remove `labels` as we have `type` now?

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -1,23 +1,32 @@
 ---
-name: Feature request
-about: Request a new feature in SearXNG
-title: ''
-labels: enhancement
-assignees: ''
-
+name: "Feature request"
+about: "Request a new feature in SearXNG"
+labels: ["new feature"]
```
inetol, line 4: Same as https://github.com/searxng/searxng/pull/5973/changes#r3093288696

.github/ISSUE_TEMPLATE/engine-request.md
```diff
@@ -1,33 +1,46 @@
...
-name: Engine request
-about: Request a new engine in SearXNG
-title: ''
-labels: enhancement, engine request
-assignees: ''
-
+name: Engine request"
+about: Request a new engine in SearXNG"
+labels: ["engine request"]
+type: "feature"
```
inetol, line 5: *This is a good example of issue types*

inetol (commented):
LGTM

.github/ISSUE_TEMPLATE/bug-report.md
Bnyro: Not sure about that, because we still have many issues without issue type.

E.g., if you want to search for all bugs, you can just filter by the bug label. If we remove that label, you will have the filter for the bug issue type - although we still have many issues without a type assigned to it.


EDIT: I see you already labeled most of them by hand, so my point is invalid.

return42: Nope .. I agree with what @vojkovic said in a internal discussion

> My understanding is that types aren't supposed to replace labels but exist alongside them ... So, I would say it makes sense to use both: use types for triaging issues but keep labels like 'new feature' in tandem so you can still make the same searches ...

.github/ISSUE_TEMPLATE/engine-request.md
return42: Is exactly why I requested your review :-)

Issue #5972, opened 2026-04-12
AOL images return HTTP500 Service Not Available


Bnyro:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
latest commit

**What happened?**

Searching `!aoli` returns HTTP500.

**Additional context**

- Copying the browser request of the image page into CURL still gives you the same error, but when using [Curl Impersonate](https://github.com/lexiforest/curl-impersonate) it works.
- Strangely, other AOL categories, namely AOL general and AOL news are not affected by this.

Related:

- https://github.com/searxng/searxng/pull/5975

**AI Disclosure**
- [X] I read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and hereby confirm that this issue conforms with the policy.

PR #5975, opened 2026-04-13
[mod] aol engines: disable by default

return42:
### What does this PR do?

Disable all AOL engines by default.

The AOL engines deliver no results or too many incorrect results; in #5972 it is reported that the AOL-images do not work, and on AOL-web the language selection seems to depend more on the IP than on the selected language / Locally, the language selection works for me, on the public server only English results are displayed, which significantly worsens the result list.

Do some queries .. there should be none AOL results in :-)

### Related issues

- #5972 

### Code of Conduct

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -426,18 +426,21 @@ engines:
     search_type: search
     categories: [general]
     shortcut: aol
+    disabled: true
 
   - name: aol images
     engine: aol
     search_type: image
     categories: [images]
     shortcut: aoli
+    disabled: true
 
   - name: aol videos
     engine: aol
     search_type: video
     categories: [videos]
     shortcut: aolv
+    disabled: true
 
   - name: arch linux wiki
     engine: archlinux


Bnyro (commented):
I personally would only disable AOL images because it doesn't work - general and news often return usable results as far as I can see (although not always)?

return42:
> general and news often return usable results

There is no AOL engine for news .. and `general` / AOL-web returns (at least on my IP) irrelevant results ..

grafik

Bnyro:
Sorry, I meant videos.

Yeah, I see, when testing things on my instance, results are not that good either. It only worked fine when using SearXNG locally when I implemented the engine.

Bnyro (approved):

Aadniz:
The irrelevant results looks very similar to what we see with Bing engine (https://github.com/searxng/searxng/issues/4964)

philipwilk:
dont mean to necro but holy cow the results from aol web are awful

Bnyro:
I still have AOL enabled on my instance and the results are usually very good - idk why it only returns good results for some IPs, but it definitely can be decent enough to be used.

return42:
Very strange — I am now getting good results for Web & Video as well:

- `!aol !aolv bmw :de-CH`

However, the Images search is still broken:

- `!aoli`
- https://github.com/searxng/searxng/issues/5972

PR #5980, opened 2026-04-15
fix(startpage): handle string date field in news results

### LINKED ISSUE #5979: Startpage engine fails when `date` field is string instead of integer (TypeError)
 
- SearXNG - 2026.4.13+ee66b070a (official Docker image, no fork)

- Using the official searxng/searxng Docker image, deployed via Dockhand on Ubuntu 24.04 LTS. No custom modifications to the image

**What happened?**

- The Startpage engine (news) raises a TypeError during result parsing:

- `TypeError: unsupported operand type(s) for /: 'str' and 'int'`

- Likely occurring  in searx/engines/startpage.py when processing the date field. Seems like it's looking for an integer timestamp (milliseconds), but received a string.

**How To Reproduce**

1. Enable the Startpage engine (news category)
2. Perform a search that returns Startpage news results
3. Observe logs during execution

**Expected behavior**

- The engine should handle unexpected data types gracefully:
  - Validate that result["date"] is numeric before performing arithmetic
  - Parsing safely if the value is a string or fallback to skipping the result
  - Most importantly; avoid raising an unhandled exception that breaks the engine response due to a date type mismatch.

**Screenshots & Logs**

- Relevant log excerpt:

```
ERROR:searx.engines.startpage news: exception : unsupported operand type(s) for /: 'str' and 'int'
File "/usr/local/searxng/searx/engines/startpage.py", line 348, in _get_news_result
    publishedDate = datetime.fromtimestamp(result["date"] / 1000)
TypeError: unsupported operand type(s) for /: 'str' and 'int'
```

**Additional context**

- This appears to be caused by a change or inconsistency in Startpage’s response format, where the date field could be returned as either a string or integer timestamps at any time.
- Adding a type check or safe conversion would help to make the engine more robust against upstream changes.

- [x ] I read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and hereby confirm that this issue conforms with the policy.
  - ChatGPT used to help analyze logs and validate issue.

xyaz1313:
## Problem

The Startpage news engine crashes with a `TypeError` when the `date` field in the API response is returned as a string instead of an integer.

**Error:**
```
TypeError: unsupported operand type(s) for /: 'str' and 'int'
File "searx/engines/startpage.py", line 348, in _get_news_result
    publishedDate = datetime.fromtimestamp(result["date"] / 1000)
```

## Root Cause

Line 348 in `searx/engines/startpage.py` performs arithmetic division `result["date"] / 1000` directly on the API response value. Startpage's API inconsistently returns the timestamp as either an integer or a string — when it's a string (e.g., `"1713244800000"`), Python raises `TypeError` because you can't divide a string by an int.

## Fix

Wrapped the conversion in `float()` with a `try/except` guard, following the same defensive pattern already used in other engines (e.g., `piratebay.py` uses `float(result["added"])` inside a bare except). The fix:

1. Uses `float()` instead of raw division — handles both `int` and `str` inputs
2. Catches `ValueError` (unconvertible strings) and `TypeError` (None/non-numeric)
3. Falls back to `publishedDate = None` on failure, so the result still returns

**Changed lines: 347-350 in `searx/engines/startpage.py`**

## Testing

- Verified the fix handles: integer timestamps, string timestamps (e.g., `"1713244800000"`), `None`, empty string, and non-numeric strings
- Syntax check passes (auto-verified by patch tool)
- No test file exists for startpage engine; the fix is minimal and follows established project patterns

## Impact

- **Scope:** Startpage news category only
- **Risk:** Very low — `float()` is a superset of the original behavior for numeric inputs; try/except is purely additive
- **Backward compatibility:** Fully maintained — existing integer timestamps work identically

Fixes #5979

### DIFF
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -345,7 +345,10 @@ def _get_news_result(result):
 
     publishedDate = None
     if result.get("date"):
-        publishedDate = datetime.fromtimestamp(result["date"] / 1000)
+        try:
+            publishedDate = datetime.fromtimestamp(float(result["date"]) / 1000)
+        except (ValueError, TypeError):
+            pass
 
     thumbnailUrl = None
     if result.get("thumbnailUrl"):


Bnyro:
Please read our [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and state where and which LLMs you used.

The PR description is definitely AI generated.

searx/engines/startpage.py
```diff
@@ -345,7 +345,10 @@ def _get_news_result(result):
 
     publishedDate = None
     if result.get("date"):
-        publishedDate = datetime.fromtimestamp(result["date"] / 1000)
+        try:
+            publishedDate = datetime.fromtimestamp(float(result["date"]) / 1000)
```
Bnyro, line 349:
```suggestion
            publishedDate = datetime.fromtimestamp(int(result["date"]) / 1000)
```
Timestamps are always int.

return42:
closed by https://github.com/searxng/searxng/pull/6053

PR #5981, opened 2026-04-16
Add MCP service plugin for SearXNG

norci:
## What does this PR do?

Add MCP (Model Context Protocol) server plugin for SearXNG. Exposes search engines via MCP tools:
- `get_engines()` - List available search engines by category
- `get_engine_info(name)` - Get details for a specific engine
- `search(query, ...)` - Execute web searches via MCP
- `engines://list` - Resource endpoint for engine list

MCP allows external AI applications (Claude Desktop, Cursor, etc.) to leverage SearXNG as a search backend. This enables:
- AI agents to perform web searches using SearXNG's 180+ engines
- Privacy-preserving search via self-hosted instance
- Unified search interface across multiple AI tools

1. Configure `mcp_port` in settings.yml
2. Run SearXNG with the MCP plugin enabled
3. Connect via MCP client:
   ```python
   from mcp import Client
   async with Client("http://localhost:PORT/mcp") as client:
       result = await client.call_tool("search", {"query": "test"})
   ```

## Author's checklist

- [x] MCP server starts on configured port
- [x] Tool calls return properly formatted results
- [x] Error handling for invalid engines/queries

## AI Disclosure
- [x] I have used AI tools for working on the changes in this pull request and will attach a list of all AI tools I used and how I used them. I hereby confirm that I haven't used any other tools than the ones I mention below.

AI Tools used:
- Claude Code ( Sonnet 4 ) - Primary development tool, architecture and implementation
- GitHub CLI - PR creation
- OpenCode (Sisyphus) - Orchestration and code management

### DIFF
--- a/container/.env.example
+++ b/container/.env.example
@@ -13,3 +13,6 @@
 
 # Listen to a specific port.
 #SEARXNG_PORT=8080
+
+# MCP server port (optional, requires mcp_port in settings.yml)
+#SEARXNG_MCP_PORT=8765
--- a/container/docker-compose.yml
+++ b/container/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     restart: always
     ports:
       - ${SEARXNG_HOST:+${SEARXNG_HOST}:}${SEARXNG_PORT:-8080}:${SEARXNG_PORT:-8080}
+      - ${SEARXNG_MCP_PORT:-8765}:${SEARXNG_MCP_PORT:-8765}
     env_file: ./.env
     volumes:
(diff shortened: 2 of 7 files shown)


tiekoetter:
My most important criticism is that something like this shouldn't be enabled by default. It also looks like this was fully AI generated without the human in the process, if there even is one, which I think violates our AI policy (https://github.com/searxng/searxng/blob/master/AI_POLICY.rst).

inetol:
There are already several MCP servers built on top of sxng as standalone projects (which is how it should be)

PR #5984, opened 2026-04-17
[fix] engine: google-news - Google pushed a frontend update

### LINKED ISSUE #5852: [Bug] Google News Engine returns 0 results - HTML Structure & Protobuf URL Encoding Changes
### The Situation & Our Setup

To be completely transparent, we are not exactly sure how to patch this codebase, but we have gathered substantial evidence isolating a total failure in the `google_news.py` parser.

We initially deployed SearXNG simply to get cleaner, more flexible Google results for low-volume, manual browsing. There was zero heavy scraping, high-velocity API usage, or automated pulling involved. It worked flawlessly for about a day, then completely died.

Our environment runs SearXNG via Docker, hosted inside an LXD container, and routed out through a Tailscale mesh IP. Because of the sudden breakage, we aggressively A/B tested our network variables to rule out IP bans:

* We tested with and without outbound rotating proxies.
* We destroyed and rebuilt the `latest` Docker image from scratch.
* We verified that querying Google News *directly* from the exact same host machine works perfectly and returns a full HTML payload.

The issue is definitively the SearXNG code failing to parse the new Google response.

### The Evidence

Around March 9–10, 2026, Google pushed a frontend update to Google News that completely changed the HTML structure of search results. The parser relies on specific xpath selectors and a base64 URL decoding method, both of which are now broken.

**1. The HTML Structure is Gone**
The parser searches for results using the xpath selector `//div[@class="xrnccd"]` which matched wrapper divs containing `<article>` elements. Google removed this entire structure. The class "xrnccd" returns 0 matches in the current response. Everything is now in flat `<a>` tags.

**2. The URL Encoding Changed**
The old href format contained a base64-encoded payload where the real article URL was embedded as plaintext bytes. The new href format (`./read/CBMi...`) uses a protobuf-wrapped, encrypted payload. Base64 decoding produces binary data with no `http` URL anywhere inside. The old decoding method returns None for every single result.

### Diagnostic Commands to Reproduce

We wrote these diagnostic scripts to prove the selectors and decoders are failing on the host machine. Run these on the SearXNG host to verify:

**Confirm Google returns valid HTML (Not IP blocked):**

```bash
curl -s -L -H "User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36" \
"https://news.google.com/search?q=bradesco&hl=pt-419&gl=BR&ceid=BR:pt-419" \
> /tmp/gnews.html

```

*(Expected: Large HTML file. If you get a small file, your IP is blocked)*

**Confirm old xpath selector is dead:**

```bash
grep -c "xrnccd" /tmp/gnews.html

```

*(Expected: 0)*

**Prove old URL decode is broken:**

```python
import base64
from lxml import html
with open('/tmp/gnews.html') as f:
    doc = html.fromstring(f.read())
    href = doc.xpath('//a[@data-n-tid]')[0].get('href','')
    href = href.split('?')[0].split('/')[-1]
    decoded = base64.urlsafe_b64decode(href + '====')
    print('http found:', b'http' in decoded)

```

*(Expected: http found: False)*

### Blockers for a Fix

Updating the xpath selectors to `//a[@data-n-tid]` is straightforward. However, fixing the URL decoding requires either returning a pass-through URL directly or integrating a resolver API.

Since the SearXNG Docker image is Void Linux-based and lacks `pip`, adding new dependencies to decrypt the protobuf links natively inside the container isn't straightforward. We are logging this in hopes maintainers can make a design decision on how to handle the new protobuf links.

return42:
### What does this PR do?

Around March 9–10, 2026, Google pushed a frontend update to Google News that completely changed the HTML structure of search results.

This is a complete overhaul of the Google News engine.

- The real URL is encoded in the "jslog" attribute.
  Suggested-by @SeriousConcept1134: the attribute is a base64 encoded JSON
- CEID list is updated
- The typification was pushed forward
   
Closes: https://github.com/searxng/searxng/issues/5852

- `make run`
- `!gon time`

HINT: to get correct thumbnails, `image_proxy` needs to be activated:

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index bd5422908..31e680fd7 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -104,7 +104,7 @@ server:
 
   secret_key: "ultrasecretkey"  # Is overwritten by ${SEARXNG_SECRET}
   # Proxy image results through SearXNG. Is overwritten by ${SEARXNG_IMAGE_PROXY}
-  image_proxy: false
+  image_proxy: true
   # 1.0 and 1.1 are supported
   http_protocol_version: "1.0"
   # POST queries are "more secure!" but are also the source of hard-to-locate
```

### Related comments

- https://github.com/searxng/searxng/issues/5852#issuecomment-4254438184
- https://github.com/searxng/searxng/issues/5852#issuecomment-4265598833

### Code of Conduct

  If I have not used AI tools for working on the changes in this PR.


Bnyro (commented):
I'm unfortunately only getting the HTTP302 as well :/

Doesn't happen on `master` branch.

return42:
@SeriousConcept1134 @Bnyro I've overhauled the engine now; it should be working properly (at least from my local IP) .. could you please test it again? Thanks!

searx/engines/google_news.py
```diff
@@ -64,231 +72,250 @@
...
-        href = href[href.index(b'http') :].split(b'\xd2')[0]
-        href = href.decode()
-
-        title = extract_text(eval_xpath(result, './article/h3[1]'))
+        href: str = eval_xpath_getindex(result, "./a[@target='_blank']/@href", 0, default=0)
+        if not href:
+            continue
+        url = href
+        if href.startswith("./"):
+            url = base_url + href[1:]
```
Bnyro:
```suggestion
        url: str = eval_xpath_getindex(result, "./a[@target='_blank']/@href", 0, default=0)
        if not url
            continue
        if url.startswith("./"):
            url = base_url + url[1:]
```
I don't think the intermediate "href" makes things more readable here

searx/engines/google_news.py
```diff
@@ -64,231 +72,250 @@
...
+        # Try to extract the real URL from jslog
+        real_url: str | None = None
+        if jslog:
+            # jslog format is usually: "95014; 5:<base64>; track:click,vis".  We
+            # want the second part (index 1) after splitting by ";"
+            parts: list[str] = jslog.split(";")
+            if len(parts) > 1:
+                b64_data: str = parts[1].split(":")[-1].strip()
+                # Pad base64 if necessary
+                b64_data += "=" * (-len(b64_data) % 4)
```
Bnyro, line 161: Is this padding really needed? From my tests, it also worked without it.

Also, I don't really understand how it works - the idea is that the amount of characters in the base64 string is always divisible by 4?

searx/engines/google_news.py
```diff
@@ -64,231 +72,250 @@
...
+                decoded_data: list[str | None] = json.loads(base64.b64decode(b64_data).decode("utf-8"))
+                # The URL is typically the last element in the decoded array
+                if isinstance(decoded_data, list) and len(decoded_data) > 0 and isinstance(decoded_data[-1], str):
+                    if decoded_data[-1].startswith("http"):
+                        real_url = decoded_data[-1]
```
Bnyro: Possibly we could also iterate over all array elements and select the last one that starts with `http`.

From my tests, it seems like all other fields are `None` anyways, e.g. `[None, None, None, None, None, None, None, None, 'https://example.com'']`

Bnyro (commented):
Works well for me, thank you for working on this!

searx/engines/google_news.py
return42: Yes, padding is needed:

    >>> base64.b64decode("bGlnaHQgdw==", )
    b'light w'
    
    >>> base64.b64decode("bGlnaHQgdw=", )
    ...
    binascii.Error: Incorrect padding

> Also, I don't really understand how it works 

Padding is a part of the base64 notation; [The encoding process represents 24-bit groups of input bits as output strings of 4 encoded characters. ](https://www.rfc-editor.org/rfc/rfc4648#section-4)

> From my tests, it also worked without it.

Because google-news does it well .. but for safety, we should assert the padding

searx/engines/google_news.py
return42: That's true .. code simplified, see [diff](https://github.com/searxng/searxng/compare/5c6f266ccab02c7b08149ea8b4f48479ef75bcd8..eda5da8a365e273f58d362da84c566595ed4c631)

searx/engines/google_news.py
return42: True :+1:, (no need to iterate and) the two `if` conditions can be simplified into one condition, see [diff](https://github.com/searxng/searxng/compare/5c6f266ccab02c7b08149ea8b4f48479ef75bcd8..eda5da8a365e273f58d362da84c566595ed4c631)

PR #6019, opened 2026-04-26
Fix Result.defaults_from() inverted logic

FaFre:
### What does this PR do?

Fixes one bugs that prevented cross-engine result deduplication from working correctly ~+ added an improvement for better result matching.~

1. **`Result.defaults_from()` had inverted logic**: because of this issue always only one search engine was shown in the engine set.

2. ~**URL hash was too strict for cross-engine dedup**~: ~Two engines returning the same page with minor URL differences (e.g. `www.example.com/page` vs `example.com/page`.  or `www.example.com/page#anchor1` vs `www.example.com/page#anchor2`) want to be count as the same result to improve scoring (which is more important).
The has is still quite unique and should avoid false positives:~

```python
...
url = self.parsed_url
        netloc = url.netloc.removeprefix("www.")
        return hash(
            f"{self.template}"
            + f"|{netloc}|{url.path}|{url.params}|{url.query}"
            + f"|{self.img_src}"
        )
```

   - Configure at least two general-search engines
   - Search for a query likely to return overlapping results
   - ~Before fix: each result shows only one engine name~
   - After fix: results found by multiple engines should list multiple engine names

### Code of Conduct

  AI tools used:
  - **OpenCode (glm-5.1)**: Used to understand the issue and validate the implementation.

### DIFF
--- a/searx/result_types/_base.py
+++ b/searx/result_types/_base.py
@@ -32,7 +32,7 @@ import msgspec
 from searx import logger as log
 
 WHITESPACE_REGEX = re.compile('( |\t|\n)+', re.M | re.U)
-UNKNOWN = object()
+UNSET = object()
 
 
 def _normalize_url_fields(result: "Result | LegacyResult"):
@@ -326,12 +326,13 @@ class Result(msgspec.Struct, kw_only=True):
 
     def defaults_from(self, other: "Result"):
         """Fields not set in *self* will be updated from the field values of the
-        *other*.
+        *other*.  If a field is set (exists) but contains an empty string
+        or the value ``None``, it is also considered *not set*.
         """
         for field_name in self.__struct_fields__:
-            self_val = getattr(self, field_name, False)
-            other_val = getattr(other, field_name, False)
-            if self_val:
+            self_val = getattr(self, field_name, UNSET)
+            other_val = getattr(other, field_name, UNSET)
+            if self_val is UNSET and other_val not in (UNSET, "", None):
                 setattr(self, field_name, other_val)
 
 
@@ -440,8 +441,6 @@ class LegacyResult(dict[str, t.Any]):
        Do not use this class in your own implementations!
     """
 
-    UNSET: object = object()
-
     # emulate field types from type class Result
     url: str | None
     template: str
@@ -512,7 +511,7 @@ class LegacyResult(dict[str, t.Any]):
             )
 
     def __getattr__(self, name: str, default: t.Any = UNSET) -> t.Any:
-        if default == self.UNSET and name not in self:
+        if default == UNSET and name not in self:
             raise AttributeError(f"LegacyResult object has no field named: {name}")
         return self[name]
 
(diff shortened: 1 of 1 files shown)


return42:
>  Before fix: each result shows only one engine name

Sure? I can't reproduce this issue .. here is the example I tested with current master branch (without this PR):

grafik

return42:
> 2\. **URL hash was too strict for cross-engine dedup**: Two engines returning the same page with minor URL differences .. want to be count as the same result to improve scoring (which is more important).

Your assumption are not correct for any cases ..

> `www.example.com/page` vs `example.com/page`

are often the same, but not in ever case.

> 2\. `www.example.com/page#anchor1` vs `www.example.com/page#anchor2`)

are pointing to the same page, but to different content .. and in the result list the user wants to see both links .. otherwise he can't jump to the content of one or the other.

FaFre:
> Sure? I can't reproduce this issue .. 

Mhm that is weird, are you using the latest master? I need this fix, just confirmed it again.

FaFre:
> > 2. **URL hash was too strict for cross-engine dedup**: Two engines returning the same page with minor URL differences .. want to be count as the same result to improve scoring (which is more important).
> 
> Your assumption are not correct for any cases ..
> 
> > `www.example.com/page` vs `example.com/page`
> 
> are often the same, but not in ever case.
> 
> > 2. `www.example.com/page#anchor1` vs `www.example.com/page#anchor2`)
> 
> are pointing to the same page, but to different content .. and in the result list the user wants to see both links .. otherwise he can't jump to the content of one or the other.

Got it, I will keep this in my fork then. I personally prefer having a higher ranking of specific sites through multiple engine hits than detailed anchors. Let me know if fix 1 can be reproduced so I will revert this.

return42:
> Mhm that is weird, are you using the latest master? I need this fix, just confirmed it again.

Please give me your example or show me more details .. otherwise I can't reproduce / analyze whats going wrong.

FaFre:
Unfortunately I use custom engines. Although this should not influence the merging. However because of that it will be not possible to reproduce. Will close this for now then, I guess when someone else is raising another issue the fix is ready here.

searx/result_types/_base.py
```diff
@@ -331,7 +331,7 @@ def defaults_from(self, other: "Result"):
         for field_name in self.__struct_fields__:
             self_val = getattr(self, field_name, False)
             other_val = getattr(other, field_name, False)
-            if self_val:
+            if not self_val:
```
return42: > I guess when someone else is raising another issue the fix is ready here.

can we reopen the PR? .. beside what I have said about the other points in this PR, I guess this patch here is a real fix.

FaFre: yes sure

FaFre:
@return42 only `not` is now applied, i reverted the hash changes

return42:
@FaFre thanks for reopening .. I squashed the commits and forced pushed on your branch.

As you can see in the new commit, the old implementation had other flaws as well; for example, an integer with the value 0 was treated as "not set." I've now fixed these issues.

Could you please take a look at and test my changes? Thanks!

FaFre:
your fix is working fine for me!

PR #6024, opened 2026-04-27
[fix] yacy: add time_range_support with client-side date filtering

songproducer:
### What does this PR do?

YaCy does not enforce datestart/dateend in global (P2P) mode, so results are now filtered client-side based on publishedDate. Adds time_range_support = True so SearXNG does not skip the engine when a time filter is active.

1. Run a local YaCy instance on the host machine (port 8090)                                         
2. Configure SearXNG with YaCy as an engine (base_url pointing to your instance, enable_http: true)  
3. Perform a search using only the YaCy engine with no time filter, confirm results appear          
4. Repeat the search with a time filter (e.g. "Last year"), confirm results are filtered to that date range and results without a pubDate are excluded                                        
5. Confirm no results are returned for a time range where YaCy has no indexed content

  I used Claude Code and tested that the fix indeed works (at least for date crawled)

### DIFF
--- a/searx/engines/yacy.py
+++ b/searx/engines/yacy.py
@@ -53,14 +53,18 @@ Implementations
 # pylint: disable=fixme
 
 
+import logging
 import random
+from datetime import datetime, timedelta, timezone
 from json import loads
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse, parse_qs
 from dateutil import parser
 
 from httpx import DigestAuth
 
 from searx.utils import html_to_text
+
+logger = logging.getLogger('searx.engines.yacy')
 
 # about
 about = {
@@ -75,6 +79,7 @@ about = {
 # engine dependent config
 categories = ['general']
 paging = True
+time_range_support = True
 number_of_results = 10
 http_digest_auth_user = ""
 """HTTP digest user for the local YACY instance"""
@@ -100,6 +105,13 @@ base_url: list[str] | str = []
 """The value is an URL or a list of URLs.  In the latter case instance will be
 selected randomly.
 """
+
+_TIME_RANGE_OFFSETS = {
+    'day': timedelta(days=1),
+    'week': timedelta(weeks=1),
+    'month': timedelta(days=31),
+    'year': timedelta(days=365),
+}
 
 
 def init(_):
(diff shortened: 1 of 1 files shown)


searx/engines/yacy.py
```diff
@@ -53,15 +53,19 @@
...
 from json import loads
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse, parse_qs
 from dateutil import parser
 
 from httpx import DigestAuth
 
 from searx.utils import html_to_text
 
+logger = logging.getLogger('searx.engines.yacy')
```
return42: logger is already defined by internal *monkey* patching.

https://github.com/searxng/searxng/blob/a7ac696b4a146af905397b8b9a07b18f4cd42bb5/searx/engines/__builtins__.pyi#L17

searx/engines/yacy.py
```diff
@@ -138,6 +150,14 @@ def request(query, params):
     if params['language'] != 'all':
         args['lr'] = 'lang_' + params['language'].split('-')[0]
 
+    # add date range if specified
+    time_range = params.get('time_range')
+    if time_range and time_range in _TIME_RANGE_OFFSETS:
+        now = datetime.now(timezone.utc)
+        start = now - _TIME_RANGE_OFFSETS[time_range]
+        args['datestart'] = start.strftime('%Y/%m/%d')
+        args['dateend'] = now.strftime('%Y/%m/%d')
```
return42: Sure there exists a `datestart` and `dateend` argument in the API? Please show, from where you have this information.

searx/engines/yacy.py
```diff
@@ -186,7 +213,18 @@ def response(resp):
...
+                except Exception:  # pylint: disable=broad-except
+                    pass
+
+            # skip results outside the requested date range
+            if date_start and date_end:
+                if publishedDate is None:
+                    continue
+                pub = publishedDate if publishedDate.tzinfo else publishedDate.replace(tzinfo=timezone.utc)
+                if not (date_start <= pub <= date_end):
+                    continue
```
return42, line 221: Makes no sense to sort out by pubDate afterwards.

return42 (changes requested):
> I used Claude Code and tested that the fix indeed works (at least for date crawled)

Sorry to say, but I have my doubts ["The human-in-the-loop fully understand all code."](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst)

See my comments below.

searx/engines/yacy.py
songproducer: In my testing, the time filter was returning the same results regardless of which option was selected. Filtering on pubDate client-side is what actually made it work.

searx/engines/yacy.py
songproducer: Agreed, removed them from the API args entirely. Date bounds are now stored in search_params and used for client-side filtering only.

searx/engines/yacy.py
songproducer: Good catch, removed. I wasn't aware it was injected by _builtins_.py

searx/engines/yacy.py
return42: You miss understood me .. it makes no sense to filter out on client site .. when the search engine does not support date ranges, it is pointless to do it on client side.

For example: the users selects date rang "last day" and the result list from page "one" does not contain a result with a pub-date that matches this range, but on page two there are results that do match .. so what is the sense? look at page one to see there a no results, go to page 2, or page n and look there are any matches?

-----

> In my testing, the time filter was returning the same results regardless of which option was selected

That is because you set `time_range_support = True` but the yacy doesn't (you removed it from the request).

PR #6046, opened 2026-05-05
[fix] engine 'mwmbl': handle empty 'extract' field gracefully

### LINKED ISSUE #6034: Bug: mwmbl engine crashes with IndexError when result extract is empty
The mwmbl engine crashes with an `IndexError: list index out of range` when the
API returns a result where the `extract` field is an empty list.

### How To Reproduce?

1. Enable the mwmbl engine
2. Search for a term that triggers results with empty `extract` fields from the mwmbl API
3. The engine crashes instead of returning partial results or skipping the entry

### Expected behavior

The engine should gracefully handle results where `extract` is an empty list --
either by skipping the result or setting `content` to an empty string.

### Screenshots & Logs

```
ERROR:searx.engines.mwmbl: exception : list index out of range
  File "searx/search/processors/online.py", line ..., in search
  File "searx/search/processors/online.py", line ..., in _search_basic
  File "searx/engines/mwmbl.py", line 46, in response
    'content': result['extract'][0]['value'],
               ~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

The problematic code in `searx/engines/mwmbl.py`, line 46:

```python
'content': result['extract'][0]['value'],
```

No check is performed to verify that `result['extract']` is non-empty before
indexing into it.

**Suggested fix:**

```python
'content': ''.join(part['value'] for part in result.get('extract', [])),
```

This mirrors the `title` handling pattern which concatenates all parts via list comprehension.

### Version of SearXNG

Branch: master
Version: 2026.4.24+a7ac696b4

Docker (searxng-docker)

### Additional context

- The `title` field in the same function uses a list comprehension that would
  produce an empty list (and thus empty string) if `result['title']` were empty,
  but `extract` does not have the same defensive handling.
- Related (but different bug): #3030 was an HTTP 500 from the upstream mwmbl API.

### Code of Conduct

SAY-5:
Closes #6034

The mwmbl engine raises `IndexError: list index out of range` when the API returns a result whose `extract` field is an empty list. Iterate over the extract parts and join them, mirroring the existing title handling.

**AI disclosure (per AI_POLICY.rst):** Claude Code was used to draft the patch and PR text; the diagnosis matches the issue reporter's analysis and I confirmed the fix locally against the three relevant inputs (empty extract list, missing extract key, normal multi-part extract).

### DIFF
--- a/searx/engines/mwmbl.py
+++ b/searx/engines/mwmbl.py
@@ -39,11 +39,12 @@ def response(resp):
 
     for result in json_results:
         title_parts = [title['value'] for title in result['title']]
+        extract_parts = [extract['value'] for extract in result.get('extract', [])]
         results.append(
             {
                 'url': result['url'],
                 'title': ''.join(title_parts),
-                'content': result['extract'][0]['value'],
+                'content': ''.join(extract_parts),
             }
         )
 


Bnyro (approved):
Thanks 👍

PR #6049, opened 2026-05-06
 [fix] wikidata: crashes when querying due to missing escaping of quotation marks

FaFre:
### What does this PR do?

Fix outer quotes of select that would otherwise crash the engine.

Run any wiki data request.

### Code of Conduct

### DIFF
--- a/searx/engines/wikidata.py
+++ b/searx/engines/wikidata.py
@@ -146,7 +146,7 @@ class WDAttribute:
         self.name: str = name
 
     def get_select(self):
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thank you!

Fyi, I changed it so that the `"` is used outside of the string and `'` inside, instead of `'` outside and `"` inside. That's just because we want our formatter to enforce the use of double quotes instead of single quotes for strings everywhere in the future.

PR #6076, opened 2026-05-11
[feat] Makefile: allow to selectively update the traits of only a single engine instead of all

Bnyro:
### What does this PR do?
- Previously, `make data.traits` would fetch traits for all engines, which is very slow and by side-effect touches engine data that are unrelated to the engine you're currently working one
- To be faster with developing `fetch_traits` support for engines, this is very helpful (at least for me). I'm not sure if there's already an other way to only run `fetch_traits` of a specific engine, but I couldn't find any

- `make "ENGINES=brave" data.traits`
- `make "ENGINES=brave google" data.traits`

### Code of Conduct


searxng_extra/update/update_engine_traits.py
```diff
@@ -77,21 +83,36 @@
...
+            engines_cfg.append(eng_data)
 
     load_engines(engines_cfg)
     # traits_map = EngineTraitsMap.from_data()
-    traits_map = fetch_traits_map()
-    sxng_tag_list = filter_locales(traits_map)
-    write_languages_file(sxng_tag_list)
+    traits_map = fetch_traits_map(engine_names_filter is not None)
+
+    if engine_names_filter is None:
```
Bnyro, line 110: This is not optimal - we're completely disabling the update to `sxng_locales.py` here.

The problem is that if we only load e.g. brave's locales, and then run `write_languages_file`, `sxng_locales.py` will only contain Brave's locales, so it probably misses a lot of languages that are supported by other engines.

We should probably merge the newly fetched data with the existing locales in `sxng_locales.py`, but I wanted to hear other opinions first.

return42:
You don't know how long I missed this feature :-)

However, I think we can implement this somewhat more simply if we use the development environment and invoke the script directly from within it -- rather than constructing the parameters from the `make` environment variables. Moreover, the latter approach has the disadvantage that we cannot properly pass engine names containing spaces.

For this reason, I have created an alternative PR.

- https://github.com/searxng/searxng/pull/6077

PR #6077, opened 2026-05-12
update_engine_traits.py: add option to update engines selective

return42:
### What does this PR do?

Previously, `update_engine_traits.py` would fetch traits for all engines, which is very slow and by side-effect touches engine data that are unrelated to the engine you're currently working on.

To be faster with developing `update_engine_traits.py` supports now engine arguments.

To test, jump into the developer environment and run the script::

    $ ./manage dev.env
    (dev.env)$ ./searxng_extra/update/update_engine_traits.py --help

### Related issues

- https://github.com/searxng/searxng/pull/6076

### Code of Conduct

- [x] **I hereby confirm that this PR conforms with the [AI Policy].** (no AI used)

### DIFF
--- a/searxng_extra/update/update_engine_traits.py
+++ b/searxng_extra/update/update_engine_traits.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Update :py:obj:`searx.enginelib.traits.EngineTraitsMap` and :origin:`searx/languages.py`
+"""Update :py:obj:`searx.enginelib.traits.EngineTraitsMap` and :origin:`searx/sxng_locales.py`
 
 :py:obj:`searx.enginelib.traits.EngineTraitsMap.ENGINE_TRAITS_FILE`:
   Persistence of engines traits, fetched from the engines.
 
-:origin:`searx/languages.py`
+:origin:`searx/sxng_locales.py`
   Is generated  from intersecting each engine's supported traits.
 
 The script :origin:`searxng_extra/update/update_engine_traits.py` is called in
@@ -28,8 +28,8 @@ from searx.engines import load_engines
 from searx.enginelib.traits import EngineTraitsMap
 
 # Output files.
-languages_file = Path(searx_dir) / 'sxng_locales.py'
-languages_file_header = """\
+sxng_locales_file = Path(searx_dir) / 'sxng_locales.py'
+sxng_locales_file_header = """\
 # SPDX-License-Identifier: AGPL-3.0-or-later
 '''List of SearXNG's locale codes used for the search language/region.
 
@@ -42,7 +42,7 @@ languages_file_header = """\
 
 sxng_locales = (
 """
-languages_file_footer = """,
+sxng_locales_file_footer = """,
 )
 '''
 A list of five-digit tuples:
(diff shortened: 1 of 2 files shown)


Bnyro (commented):
Could you please also add a note about it to the developer documentation? I'm pretty sure I'll forget the correct command a few times, and it's more fun to look it up in our new Furo docs instead of searching your comment on GitHub :)

searxng_extra/update/update_engine_traits.py
```diff
@@ -75,43 +78,63 @@
...
 }
 
+app = typer.Typer()
 
-def main():
 
-    engines_cfg = []
+@app.command()
+def cli(engines: t.Annotated[list[str] | None, typer.Argument()] = None):
+    """Update ``data/engine_traits.json`` and ``languages.py``.
```
Bnyro:
```suggestion
    """Update ``data/engine_traits.json`` and ``searx/sxng_locales.py``.
```
Or is there a `languages.py` I don't know?

searxng_extra/update/update_engine_traits.py
```diff
@@ -75,43 +78,63 @@
...
 
+    Optionally, if arguments are provided via the command line, these are
+    interpreted as the names of the engines that should be updated.  All other
+    engines will be left untouched.
+    """
+
+    all_eng_names: list[str] = [_["name"] for _ in settings["engines"]]
```
Bnyro:
```suggestion
    all_eng_names: list[str] = [engine["name"] for engine in settings["engines"]]
```
I think we should generally only use `_` as variable name if we don't use the variable at all? At least that's common in other programming languages.

Not sure if there's any official resource about that, but e.g. https://alpopkes.com/posts/python/magical_universe/day_7_underscore_patterns/ also explains it this way.

searxng_extra/update/update_engine_traits.py
```diff
@@ -75,43 +78,63 @@
...
+    if engines:
+        unknown: list[str] = [_ for _ in engines if _ not in all_eng_names]
```
Bnyro: same as above (`_` naming)

Bnyro:
I'd personally still like to have a `make data.traits` shorthand for it (e.g. `make ENGINES=... data.traits` because I think it's nicer if this can be run with just one single command instead of having to enter the `venv` before.

searxng_extra/update/update_engine_traits.py
return42: I agree with you, the underscore `_` name is a critical since it is used for multiple purposes e.g. its often used as an alias for [gettext](https://docs.python.org/3/library/gettext.html#gettext.gettext) and name collisions are to be expected.

searxng_extra/update/update_engine_traits.py
return42: Good catch .. I forgot it was renamed in 2023 in c9cd376186d .. I'm to long in this project :-)  .. I renamed all occurrences, see [diff](https://github.com/searxng/searxng/compare/95ca55e7c43ed5ba5e9baa54f4a009f8f770eef6..f1c272c9e9281efbf5e81abfc647ae7f956b9f40).

return42:
> Could you please also add a note about it to the developer documentation? 
..
> I'd personally still like to have a `make data.traits` shorthand for it

If you'd like, I can add that to the Makefile... but I'd prefer that developers get used to the development environment... it's better suited for development than the Makefile targets. Here's an excerpt from [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) ..

grafik

return42:
@Bnyro for better documentation I added 9bc5f1925 to the "Developer-Quickstart" ...

grafik

Bnyro (approved):

PR #6082, opened 2026-05-16
[chore] settings.yml: set broken engines yahoo and karmasearch to inactive

Bnyro:
Both engines stopped working a while ago, and hence both are currently unusable. So it doesn't make sense to keep them enabled.

### Code of Conduct

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1203,6 +1203,7 @@ engines:
     categories: [general, web]
     search_type: web
     shortcut: ka
+    inactive: true
 
   - name: karmasearch images
     engine: karmasearch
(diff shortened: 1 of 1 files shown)


return42:
From my IP I get good results from yahoo *Web* (`!yh bmw :fr-CA`) .. *News* ends up with HTTP-500 (`!yhn bmw`)

Bnyro:
> From my IP I get good results from yahoo _Web_ (`!yh bmw :fr-CA`) .. _News_ ends up with HTTP-500 (`!yhn bmw`)
>
Interesting, I get a HTTP Protocol Error for Yahoo web (but not if using curl impersonate, then it was HTTP500 I think)

PR #6090, opened 2026-05-17
[fix] preferences: opening preferences page is very slow (multiple seconds on bad hardware)

Bnyro:
I've been profiling the `/preferences` endpoint using werkzeug's `ProfilerMiddleware` (i.e. just do `app.wsgi_app = ProfilerMiddleware(app.wsgi_app)`) and look at the outputs in the terminal when doing `make run`.

It turns out that 95%+ of the time spent were inside babel's Locale parsing (> 700ms on my machine). That's because, when opening the settings, we loaded the full engine traits of each engine and checked if it matches the user-defined search language. As we have 250+ engines, and babel is very slow when parsing Locale's, this took a very long time.

By removing this feature that shows whether the selected search language is supported by the engine, the load time went down from 800ms to 50ms on my machine (which is still very slow, but well, that's future work on optimizing).

I'm pretty sure that somebody will complain about losing this feature, but I don't think it's any important to see whether an engine supports your current search language, I don't think anyone really goes to the engines settings and enables all engines that have the "supports search language" checkmark checked... In my eyes, the performance is much more important. (at least I would be very frustrated if I try out a project and it takes multiple seconds to load the settings page...)

- The difference can be seen by just opening the `/preferences` before and after the change, but it's also possible to statistically see that:
Apply the following patch:
```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index f5e8e5cf3..d2c506ae7 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -3,6 +3,7 @@
 """WebApp"""
 # pylint: disable=use-dict-literal
 
+from werkzeug.middleware.profiler import ProfilerMiddleware
 import json
 import os
 import sys
@@ -1382,6 +1383,7 @@ def static_headers(headers: Headers, _path: str, _url: str) -> None:
         headers[header] = str(value)
 
 
+app.wsgi_app = ProfilerMiddleware(app.wsgi_app)
 app.wsgi_app = ProxyFix(app.wsgi_app)
 app.wsgi_app = WhiteNoise(
     app.wsgi_app,
```
Observe the outputted `cumtime` in the first row of the printed profiling logs before and after the changes in this PR.

### DIFF
--- a/searx/enginelib/traits.py
+++ b/searx/enginelib/traits.py
@@ -116,19 +116,6 @@ class EngineTraits:
             return self.all_locale
         return locales.get_engine_locale(searxng_locale, self.regions, default=default)
 
-    def is_locale_supported(self, searxng_locale: str) -> bool:
-        """A *locale* (SearXNG's internal representation) is considered to be
-        supported by the engine if the *region* or the *language* is supported
-        by the engine.
-
-        For verification the functions :py:func:`EngineTraits.get_region` and
-        :py:func:`EngineTraits.get_language` are used.
-        """
-        if self.data_type == "traits_v1":
-            return bool(self.get_region(searxng_locale) or self.get_language(searxng_locale))
-
-        raise TypeError("engine traits of type %s is unknown" % self.data_type)
-
     def copy(self):
         """Create a copy of the dataclass object."""
         return EngineTraits(**dataclasses.asdict(self))
--- a/searx/templates/simple/preferences/engines.html
+++ b/searx/templates/simple/preferences/engines.html
@@ -23,7 +23,6 @@
         <th class="checkbox-col">{{- _("Allow") -}}</th>{{- '' -}}
         <th class="name">{{- _("Engine name") -}}</th>{{- '' -}}
         <th class="shortcut">{{ _("!bang") -}}</th>{{- '' -}}
-        <th>{{- _("Supports selected language") -}}</th>{{- '' -}}
         <th>{{- _("SafeSearch") -}}</th>{{- '' -}}
         <th>{{- _("Time range") -}}</th>{{- '' -}}
         <th>{{- _("Weight") }}</th>
@@ -69,9 +68,6 @@
               </th>{{- '' -}}
               <td class="shortcut">{{- '' -}}
                 <span class="bang">{{ '!' + shortcuts[search_engine.name] }}</span>{{- '' -}}
-              </td>{{- '' -}}
-              <td>
-                {{- checkbox(None, supports[search_engine.name]['supports_selected_language'], true) -}}
               </td>{{- '' -}}
               <td>
                 {{- checkbox(None, supports[search_engine.name]['safesearch'], true) -}}
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -918,9 +918,6 @@ def preferences():
             'rate80': rate80,
             'rate95': rate95,
             'warn_timeout': e.timeout > settings['outgoing']['request_timeout'],
-            'supports_selected_language': e.traits.is_locale_supported(
-                str(sxng_request.preferences.get_value('language') or 'all')
-            ),
             'result_count': result_count,
         }
     # end of stats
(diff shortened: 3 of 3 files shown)


return42:
> I'm pretty sure that somebody will complain about losing this feature, but I don't think it's any important to see whether an engine supports your current search language,

It is the only way for the user to determine whether it makes sense for them to activate *this* engine.

----

I can't notice any difference on my laptop (response time is < 1sec) .. is it really noticeable on your hardware?

> Observe the outputted `cumtime` in the first row of the printed profiling logs before and after the changes in this PR.

If I insert `ProfilerMiddleware`, the response time jumps from <1 sec to >15 sec. With a measurement like this -- where the profiler itself consumes fifteen times the execution time -- I cannot determine whether this optimization is relevant for the application as a whole.

For this reason, I performed my own measurements to determine the elapsed time; here is my patch:

```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index b0820c753..44068ced9 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -860,6 +860,8 @@ def autocompleter():
 @app.route('/preferences', methods=['GET', 'POST'])
 def preferences():
     """Render preferences page && save user preferences"""
+    import time
+    total_start = time.perf_counter()
 
     # pylint: disable=too-many-locals, too-many-return-statements, too-many-branches
     # pylint: disable=too-many-statements
@@ -955,6 +957,7 @@ def preferences():
 
     # supports
     supports = {}
+    start = time.perf_counter()
     for _, e in filtered_engines.items():
         supports_selected_language = e.traits.is_locale_supported(
             str(sxng_request.preferences.get_value('language') or 'all')
@@ -966,8 +969,10 @@ def preferences():
             'safesearch': safesearch,
             'time_range_support': time_range_support,
         }
+    end = time.perf_counter()
+    print(f"XXXX supports_selected_language - elapsed time {(end -start)}")
 
-    return render(
+    ret_val = render(
         # fmt: off
         'preferences.html',
         preferences = True,
@@ -994,6 +999,10 @@ def preferences():
         doi_resolvers = get_setting("doi_resolvers", {}),
         # fmt: on
     )
+    total_end = time.perf_counter()
+    print(f"XXXX preferences total - elapsed time {(total_end - total_start)}")
+    return ret_val
```

and here are my results:
 
    XXXX supports_selected_language - elapsed time 0.17942957699415274
    XXXX preferences total - elapsed time 0.43878034299996216   

> It turns out that 95%+ of the time spent were inside babel's Locale parsing

If my measurement method and my assumptions are correct, (only) 40% of the runtime is spent in the loop.

So, maximum optimization would never yield more than 40% .. if that really is too slow on your hardware, then it can surely be optimized somehow .. but simply removing the entire feature strikes me as a bit too drastic.

Could you please run my performance measurements on the hardware where the preferences are too slow? Thanks.

Bnyro:
For reference, here's my instance: https://s.chatoyer.de/

It takes >2 seconds to load the settings page. I don't currently have time for it, but I'll deploy my patch on the weekend for testing purposes.

> If my measurement method and my assumptions are correct, (only) 40% of the runtime is spent in the loop.
>
Even if it only takes 40% of the time, that's still way too much for the fact that probably nobody ever uses this feature.

I see your argument, but let's be honest: most people either use the defaults or they enable all engines that they know - I'm 100% sure that almost nobody will
1. Select a fixed search language and then
2. Open the settings to look into the table which engines are supported

I didn't even know that feature exists before doing the performance profiling.

Anyways, if you're not okay with the feature, we should think about optimizing this using caching. Imo it's very counter-beneficial if even a simple setting page load takes 2 seconds. SearXNG already has the reputation of being quite bloated, so we should really consider if this feature is any useful given my argumentation above and in the PR description.

return42:
> SearXNG already has the reputation of being quite bloated,

:-)

In the past, we implemented too many niche features and spent too little time on rewrites (who actually enjoys doing that?). In that sense, I agree with you: it really would be better if we adopted a somewhat more lightweight approach.

TBH, I’ve never actually used the preferences myself; I find it all a bit overwhelming, and I get along perfectly well with just the defaults and the search syntax.

~I’ll take another look at the code to see if I can optimize anything further without overcomplicating things~ .. otherwise, the feature will have to go. If we do remove it, however, we should also make sure to strip out the corresponding code from the traits; otherwise, we’ll just end up with unused code sitting around forever.

return42 (approved):
FYI: the `is_locale_supported` method is no longer needed .. I removed the method and [force pushed](https://github.com/searxng/searxng/compare/0e15f16dce71ff3ebdedabe4fa4128f06f7bdf60..9d31619316964b841da856859c94a568fe9d5631).

Bnyro:
Very nice, thank you for testing this and merging ❤️

PR #6091, opened 2026-05-17
[feat] engines: add 500px.com engine

Bnyro:
- !500px tree

### Code of Conduct

### DIFF
--- a/searx/engines/500px.py
+++ b/searx/engines/500px.py
@@ -1,0 +1,172 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=invalid-name
+"""500px_ is a online network for photographers with millions of members
+worldwide. Photographers come to 500px to discover and share incredible photos,
+gain meaningful exposure, compete in photo contests, and license their photos
+through our exclusive distribution partners.
+
+.. _500px: https://500px.com
+
+"""
+
+import typing as t
+
+import random
+import string
+
+from searx.result_types import EngineResults
+from searx.enginelib import EngineCache
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+# about
+about = {
+    "website": "https://500px.com",
+    "wikidata_id": "Q354894",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://500px.com"
+api_url = "https://api.500px.com"
+
+categories = ["images"]
+
+paging = True
+results_per_page = 30
+"""Number of results to return in the request.
+
+The default was taken from the WEB UI, where the GraphQL query sets the value to
+*static*: ``first: 30``.
+"""
+
+
+def page_hash(pageno: int, query: str):
+    return f"<pageno:{pageno} ({results_per_page})>" + CACHE.secret_hash(query)
+
+
+CACHE: EngineCache
+"""Persistent (SQLite) key/value cache that deletes its values after ``expire``
+seconds.
+
+For introspection (in the developer environment) use::
+
+    $ ./manage dev.env
+    (dev.env)$ python -m searx.enginelib cache status
+    ...
+    [eng_500px] 2026-05-18 18:52:38 <pageno:2 (40)>6da7...76a3f7 --> (str:8) cG9zLTM5
+    [eng_500px] 2026-05-18 18:52:43 <pageno:3 (40)>6da7...76a3f7 --> (str:8) cG9zLTc5
+
+In the output from the example above, we see cached *cursor* for follow up
+pages, the query term is a hash value and the date shows the expire date and
+time."""
+
+
+SXNG_query = """query PhotoSearchPaginationContainerQuery(
+    $first: Int, $cursor: String, $search: String!, $sort: PhotoSort, $filters: [PhotoSearchFilter!], $nlp: Boolean
+) {
+  ...SXNG_query
+}
+
+fragment SXNG_query on Query {
+  photoSearch(sort: $sort, first: $first, after: $cursor, search: $search, filters: $filters, nlp: $nlp) {
+    edges {
+      node {
+        id
+        canonicalPath
+        name
+        description
+        width
+        height
+        photographer: uploader {
+          displayName
+        }
+        images(sizes: [35, 33]) {
+          size
+          url
+          jpegUrl
+          webpUrl
+          id
+        }
+      }
+      cursor
+    }
+    pageInfo {
+      endCursor
+      hasNextPage
+    }
+  }
+}
+"""
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE, SXNG_query  # pylint: disable=global-statement
+    CACHE = EngineCache(str(engine_settings.get("name")))
+    rand_str: str = "".join(random.choice(string.ascii_letters) for _ in range(5))
+    SXNG_query = SXNG_query.replace("SXNG_query", "PhotoSearchPaginationContainer_query_1" + rand_str)
+    return True
+
+
+def request(query: str, params: "OnlineParams") -> None:
+
+    cursor: str | None = None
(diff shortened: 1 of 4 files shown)


searx/settings.yml
```diff
@@ -310,6 +310,12 @@ engines:
     shortcut: 360sov
     disabled: true
 
+  - name: 500px
+    engine: 500px
+    shortcut: "500"
+    disabled: true
+    timeout: 5
```
return42, line 317: Every time a completely new search term is used, this timeout is exceeded... especially with search terms that are very general in nature, such as "star" or "house".

return42 (approved):
I played with the GraphQL query (reduced the result table) .. and added pagination .. the "cursor" is stored in the `EngineCache` .. read doc-strings for more details and see [diff](https://github.com/searxng/searxng/compare/49c21649a359ed880bda125f1207a3454e90fcc0..b0f1aa0e7b53c81f14de7b7018e2eccf2bd3e763).

searx/engines/500px.py
```diff
@@ -0,0 +1,164 @@
...
+    "results": "JSON",
+}
+
+base_url = "https://500px.com"
+api_url = "https://api.500px.com"
+
+categories = ["images"]
+
+paging = True
+nb_per_page = 40
```
Bnyro:
```suggestion
results_per_page = 40
```
We typically call this `results_per_page`, so probably better to stick to that naming scheme

searx/engines/500px.py
```diff
@@ -0,0 +1,164 @@
...
+
+In the output from the example above, we see cached *cursor* for follow up
+pages, the query term is a hash value and the date shows the expire date and
+time."""
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    # HINT: cache name is a table and SQL table names must start with a letter
+    CACHE = EngineCache(engine_settings.get("name", "500px"))  # pyright: ignore[reportAny]
```
Bnyro:
```suggestion
    CACHE = EngineCache(engine_settings["name"])  # pyright: ignore[reportAny]
```
I think engines always have a name? Otherwise that's an ugly misconfiguration

searx/engines/500px.py
```diff
@@ -0,0 +1,164 @@
...
+    return True
+
+
+SXNG_query = """query PhotoSearchPaginationContainerQuery(
+    $first: Int, $cursor: String, $search: String!, $sort: PhotoSort, $filters: [PhotoSearchFilter!], $nlp: Boolean
+) {
+  ...SXNG_query
+}
+
+fragment SXNG_query on Query {
```
Bnyro, line 55: if we always send the query named as `SXNG_Query`, doesn't this make it very easy for them to identify and block us?

I agree that `PhotoSearchPaginationContainer_query_1JAfuE` is also not perfect, but possibly we can just do `.PhotoSearchPaginationContainer_query_<6 random letters>`

searx/settings.yml
Bnyro: Good catch. In my tests it sometimes took up to 7 seconds when requesting the 2nd, 3rd or 4th page, so I think even 5 seconds are not enough

searx/engines/500px.py
Bnyro: Couldn't possibly lowering the amount of results improve the response time? Do you still know that 500PX's website is using? (too lazy to test :) ) If we use the same amount of results per page as their website, maybe they still have it cached and we can get our results faster

Bnyro (commented):
Using the engine cache is a very good idea, thank you for looking into pagination and cleaning up the query 👍 

I tried to implement it using `params["engine_data"]` like in https://github.com/searxng/searxng/blob/master/searx/engines/piped.py, but that doesn't seem to be persisted for the next query. Probably the pagination of Piped never worked either? idk

searx/engines/500px.py
```diff
@@ -0,0 +1,164 @@
...
+          jpegUrl
+          webpUrl
+          id
+        }
+      }
+      cursor
+    }
+    pageInfo {
+      endCursor
+      hasNextPage
```
return42: FYI: The value of `endCursor` is base64 encoded, for example `cG9zLTM5` which is decoded `pos-39`.

Bnyro: Oh wow, how did you figure that out?

return42: The higher values ​​for `endCursor` feature contain the typical padding characters `==` .. and the values ​​were always identical for a specific page number, regardless of the query.

return42:
> I tried to implement it using params["engine_data"] like in https://github.com/searxng/searxng/blob/master/searx/engines/piped.py, 

its using *hidden fields* in the HTML form send to the user .. described here: https://docs.searxng.org/dev/templates.html#engine-data-form

> but that doesn't seem to be persisted for the next query. Probably the pagination of Piped never worked either? idk

Can't speak for piped, but youtube make use of it and does work:

grafik

----

OT: did you ever heard of CCC's Blinkenlights-Projekt :-)

Bnyro:
> OT: did you ever heard of CCC's Blinkenlights-Projekt :-)
>
Nope, but looks very cool. I didn't expect them to be able to put lights onto the Elbphilharmonie :)

Bnyro:
Bdcause it might be useful to you: There is this tool called CyberChef (https://gchq.github.io/CyberChef/) that can automatically detect the encoding of a text. If you select the "Magic" tool there, you can just paste in `cG9zLTM5` and it tells you that it's base64 for `pos-39`. Also works with JWTs and other kind of reversable encodings. This helped me a lot when working on the SwissCows engine :)

Bnyro:
> its using hidden fields in the HTML form send to the user .. described here: https://docs.searxng.org/dev/templates.html#engine-data-form
>
Hmm, I'll have to try it again soon then, maybe I just used it wrong.

searx/engines/500px.py
return42: > Do you still know that 500PX's website is using? 

Web-UI's GraphQL query sets the value to *static*: ``first: 30``.

searx/settings.yml
return42: > If we use the same amount of results per page as their website, maybe they still have it cached and we can get our results faster

:thinking: .. Today, the initial query -- no matter what I try -- consistently responds in under 5 seconds, and subsequent queries or pages are delivered in under 3 seconds -- regardless of the value of `results_per_page`. (in my tests two days ago the initial response toked up to 30sec and more / it's possible it was a temporary glitch on their end.

--> We should leave the timeout as is for now :+1:

searx/engines/500px.py
return42: As far I can say, the `<6 random letters>` changes once a day (tried with different WEB-browsers from the same IP) .. I don't have a clue what the meaning of this value is / assuming it has any meaning at all. The prefix `query_1` seems to be a constant and `JAfuE` might be a random string (?).

I modified the `def setup(...)` function, see [diff](https://github.com/searxng/searxng/compare/b0f1aa0e7b53c81f14de7b7018e2eccf2bd3e763..b4caaaf257de107cac93877e86c91eeac61a57d0)

return42:
@Bnyro I toked your remarks into account and fixed the issue / may you have time to give it another try / thanks :+1:

Bnyro:
> @Bnyro I toked your remarks into account and fixed the issue / may you have time to give it another try / thanks 👍
>
Looks very good :+1: 

I noticed that we still use the `EngineCache` though, although as you figured out, we don't actually need it, right? I've pushed a commit that automatically calculates the page cursor based on your findings and removes the engine cache.

searx/settings.yml
Bnyro: From my recent use it's always been less than 2 seconds, so agreed :+1:

return42:
> I've pushed a commit that automatically calculates the page cursor

Much better, thanks a lot .. I just modified the GraphQL query in my last force push and removed the `pageInfo` block we no longer need --> [diff](https://github.com/searxng/searxng/compare/7c5684bf34c9da4fbf2012414cbe1522b0aa39ab..abc7c364fc15d9cc74a04ca36555f68c15b48660)

PR #6097, opened 2026-05-19
[feat] engines: add swisscows general, images, videos and news engine

Bnyro:
This is probably the most complicated engine implementation I did so far, took more than 5 hours to implement this.

Caveats:
- locatization is not implement yet, although I'm not sure if it's that useful to have it
- news only supports very few languages (only German ones I think, didn't look it up yet. definitely no english results) and doesn't use any protection mechanism like the others -> therefore I split it into a different engine

I've attached the reverse engineered (manually deobfuscated) JavaScript in the first commit message.

- !sw test
- !swi test
- !swn test
- !swv test

### Code of Conduct

- I used ChatGPT to pretty-print and understand the obfuscated JavaScript code from Swisscows, but not for writing any code.

### DIFF
--- a/searx/engines/swisscows.py
+++ b/searx/engines/swisscows.py
@@ -266,7 +266,7 @@ def response(resp: "SXNG_Response") -> EngineResults:
                     thumbnail=result.get("thumbnail", {}).get("url"),
                 )
             )
-        elif result["type"] == "VideoCollection":
+        elif swisscows_category == "videos" and result["type"] == "VideoCollection":
             for video in result["hasPart"]:
                 res.add(_video_result(video))
         elif result["type"] == "ImageObject":
--- a/searx/engines/swisscows.py
+++ b/searx/engines/swisscows.py
@@ -2,6 +2,8 @@
 # pylint: disable=invalid-name
 """Swisscows (general, images, videos)"""
 
+import typing as t
+
 import base64
 import codecs
 import hashlib
@@ -11,9 +13,9 @@ import random
 from datetime import datetime
 from urllib.parse import urlencode
 
-import typing as t
+from babel.core import get_global
 
-from searx.result_types import EngineResults, LegacyResult
+from searx.result_types import EngineResults, LegacyResult  # pyright: ignore[reportPrivateLocalImportUsage]
 from searx.utils import humanize_number, html_to_text
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


Austin-Olacsi:
It only seems to work when the queries do not contain any spaces.

Austin-Olacsi:
Searched for "dog" got a KeyError:

dog

Bnyro:
Fixed both.

For better debugging, please share the crash log from `make run`  in the future, that makes it easier to debug (in this case it was trivial, but in many other cases, it can be very helpful to have the full stacktrace) :)

return42 (approved):
> locatization is not implement yet, although I'm not sure if it's that useful to have it

why not useful? .. I pushed a commit on top to add region support

>  news only supports very few languages .. 

yes, or with other words: swisscows support regions where lang 'de' spoken

> -> therefore I split it into a different engine

👍 

------

I've pushed a second commit on top .. where video results from the web search are ignored and not added to results list .. is that okay? ..

Image

Bnyro:
Thanks for the locale support, very cool :+1: 

I've done some small code quality and typo changes (https://github.com/searxng/searxng/compare/e24d2a86c15e8ddf57f548d5a0fb7c519d48e5f5..5b586f1382553180650b36001cc94f545e0b4409), but nothing big.

Unfortunately Swisscows News stopped working now - even on their website, the "news" category just disappeared :shrug: Anyways, maybe it'll be back soon, so I'd still keep it for now.

Additionally, I set all of them to inactive (as discussed on Matrix) - I don't want bots to destroy the efforts I invested in reverse engineering within a few hours.

searx/engines/swisscows.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def generate_nonce(length: int = 32) -> str:
+    """
+    Generate a random char sequence with the given length.
+    """
+    return "".join([random.choice(NONCE_ALPHABET) for _ in range(length)])
+
+
+def caesar_shift_with_switch_case(s: str, offset: int = 13) -> str:
```
Bnyro, line 98: I just learned that this is called `ROT13` and there's a full Wikipedia article about it (https://en.wikipedia.org/wiki/ROT13) and how people use this as a joke. I'd suspect that the Swisscows devs are quite funny computer scientists :)

PR #6103, opened 2026-05-21
[mod] engine fyyd: typing added, no functional change

return42:
no functional change .. just some *normalization* and typification.

  No AI here :-)

### DIFF
--- a/searx/engines/fyyd.py
+++ b/searx/engines/fyyd.py
@@ -1,15 +1,23 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Fyyd (podcasts)"""
 
+import typing as t
+
 from datetime import datetime
 from urllib.parse import urlencode
 
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
 about = {
-    'website': 'https://fyyd.de',
-    'official_api_documentation': 'https://github.com/eazyliving/fyyd-api',
-    'use_official_api': True,
-    'require_api_key': False,
-    'results': 'JSON',
+    "website": "https://fyyd.de",
+    "official_api_documentation": "https://github.com/eazyliving/fyyd-api",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
 }
 categories = []
 paging = True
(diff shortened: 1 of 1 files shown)


searx/engines/fyyd.py
```diff
@@ -18,31 +26,30 @@
...
+    params["url"] = f"{base_url}/0.2/search/podcast?{urlencode(args)}"
 
 
-def response(resp):
-    results = []
+def response(resp: "SXNG_Response"):
+    res = EngineResults()
 
-    json_results = resp.json()['data']
+    json_results: list[dict[str, str]] = resp.json()["data"]  # pyright: ignore[reportAny]
```
Bnyro, line 41:
```suggestion
    json_results: list[dict[str, t.Any]] = resp.json()["data"]  # pyright: ignore[reportAny]
```
e.g. `result["categories"]` is an integer list, `result["rank"]` is int

Bnyro (approved):

searx/engines/fyyd.py
return42: To be correct we have to define a [TypedDict](https://typing.python.org/en/latest/spec/typeddict.html) for the JSON object .. but that would go too far; in my opinion / the effort would not be justified.

So I opted for a (simple but incorrect) generalization; a *more* correct generalization would be

    dict[str, t.Any]

or even better:

    JSONType: t.TypeAlias = dict[str, "JSONType"] | list["JSONType"] | str | int | float | bool | None

However, this has the disadvantage that the type checker would now require an explicit typecast for each value.

grafik

Therefore, I opted for the *simple but incorrect* type, especially since we're only passing string values ​​to the `MainResult` constructor anyway.

If that's okay with you, I'd like to leave it as is (`dict[str, str]`).

Bnyro: That kind of outplays type checks then, so in this case they're not really useful anyways?

Nevertheless, fine with me

return42: When dealing with data structures from external sources, type checking doesn't really make sense .. at least not unless you know the data structure 100% precisely.

We develop our engines to be "robust," leveraging the advantages of Python's dynamic typing to our benefit; however, in development we cannot disable the type checker for the engines, as we wish to take full advantage of its benefits for our internal APIs (such as `SXNG_Response`, `OnlineParams`, etc.).

Long story told short: 

#### *Defining types in a dynamically typed language always entails a certain contradiction and requires compromises in order to leverage the advantages of both worlds.*

PR #6105, opened 2026-05-21
[fix] ci: documentation workflow dependencies

### LINKED ISSUE #6099: [docs] uWSGI architecture diagram is broken
### How To Reproduce?
- open https://docs.searxng.org/admin/architecture.html#architecture-uwsgi

### Expected behavior
- see a graph rendered via `dot`

### Screenshots & Logs

Image

The issue is probably caused by #3408

inetol:
### What does this PR do?

Adds missing host dependencies:
```console
2026-05-21T10:07:29.2781657Z kfigure: check installed tools ...
2026-05-21T10:07:29.2784450Z �[91mWARNING: dot(1) not found, for better output quality install graphviz from http://www.graphviz.org�[39;49;00m
2026-05-21T10:07:29.2785295Z �[91mWARNING: convert(1) not found, for SVG to PDF conversion install ImageMagick (https://www.imagemagick.org)�[39;49;00m
```

You can't

### Related issues

Closes https://github.com/searxng/searxng/issues/6099

### Code of Conduct

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -50,11 +50,14 @@ jobs:
             python-${{ env.PYTHON_VERSION }}-${{ runner.arch }}-
           path: "./local/"
 
+      - name: Setup dependencies
+        run: sudo ./utils/searxng.sh install buildhost
+
       - name: Setup venv
         run: make V=1 install
 
       - name: Build documentation
-        run: make V=1 docs.clean docs.html
+        run: make V=1 docs.html
 
       - if: github.ref_name == 'master'
         name: Release


inetol:
OK

```console
kfigure: check installed tools ...
use dot(1) from: /usr/bin/dot
use convert(1) from: /usr/bin/convert
```

.github/workflows/documentation.yml
```diff
@@ -31,6 +31,11 @@ jobs:
       contents: write
 
     steps:
+      - name: Setup dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y graphviz imagemagick texlive-latex-base texlive-latex-extra
```
return42: From https://docs.searxng.org/admin/buildhosts.html

Image

I assume `fonts-dejavu` is no longer needed (since furo) .. and `shellcheck` is not needed to build the online docs .. but what about the others:

https://github.com/searxng/searxng/blob/d3deacc6d4667a1fcff0b1eb9636c7b5f494b0d0/utils/searxng.sh#L62-L65

if these are also not needed anymore, we should also fix it in the code fragment above.

inetol: They are dependencies (probably). `librsvg2-bin` and `latexmk` idk, `texlive-latex-recommended` is covered by `texlive-latex-extra`, `texlive-extra-utils` not needed here.

> ... we should also fix it in the code fragment above.

Not in this PR

return42: > Not in this PR

In the past, we had a "Single Source of Truth" by creating a script that we also invoked within our GitHub workflows.

     run: sudo ./utils/searxng.sh install buildhost

TBH, I would prefer it if we went back to that approach.

Not only does it provide a "Single Source of Truth", but it also offers the advantage that we can test every step of the build offline .. for instance, I always test changes to the build processes within a local container.

It goes even further: the reST source code for the screenshot in the documentation—which I included in my comment above is generated directly from the `SEARXNG_BUILD_PACKAGES_debian` in the script (also shown above, below the screenshot).

If it bothers you that `install buildhost` installs more than is actually needed for the dedicated GH workflow (created in #4733), we could certainly adjust that in the script as well .. though, in my opinion, that wouldn't really belong in this PR.

The primary goal of this PR is simply to fix what was unintentionally broken in https://github.com/searxng/searxng/pull/4733#pullrequestreview-4322106854

inetol: I agree. https://github.com/searxng/searxng/issues/5056 also deals this issue.

inetol: > If it bothers you that `install buildhost` installs more than is actually needed ...

Not much of a difference at install time, I squashed the changes to use the `install buildhost` script instead.

return42: FYI: I created https://github.com/searxng/searxng/issues/6239

return42 (approved):

PR #6117, opened 2026-05-22
[mod] hardening of the `Result.filter_urls()` method

return42:
### What does this PR do?

Hardening of the `Result.filter_urls()` method: exceptions in the execution of the callback must be caught / ignored and logged on the ERROR log.

Apply this patch to provoke a ValueError exception::

```diff
diff --git a/searx/data/tracker_patterns.py b/searx/data/tracker_patterns.py
index ed4415bce..695ed05d2 100644
--- a/searx/data/tracker_patterns.py
+++ b/searx/data/tracker_patterns.py
@@ -114,6 +114,7 @@ class TrackerPatternsDB:
         Returns bool ``True`` to use URL unchanged (``False`` to ignore URL).
         If URL should be modified, the returned string is the new URL to use.
         """
+        raise ValueError("test callback exceptions")

         new_url = url
         parsed_new_url = urlparse(url=new_url)
```

Start a `make run` instance and query for example `amazon` .. have a look at the
ERROR log:

    ERROR   searx.result_types: filter_urls (field 'url'): \
        ignore ValueError('test callback exceptions') from \
        callback searx/data/tracker_patterns.py:117

### Code of Conduct

  no AI here :-)

### DIFF
--- a/searx/result_types/_base.py
+++ b/searx/result_types/_base.py
@@ -19,6 +19,7 @@
 __all__ = ["Result"]
 
 import typing as t
+import types
 
 import re
 import urllib.parse
@@ -29,7 +30,9 @@ from collections.abc import Callable
 
 import msgspec
 
-from searx import logger as log
+from searx import logger
+
+log = logger.getChild("result_types")
 
 WHITESPACE_REGEX = re.compile('( |\t|\n)+', re.M | re.U)
 UNSET = object()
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Good fix, definitely doesn't make sense to pollute the log with full stacktraces when the filter func fails.

searx/result_types/_base.py
```diff
@@ -125,8 +128,20 @@ def _filter_urls(
...
+            new_url = filter_func(result, field_name, url_src)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            # pylint: disable=no-member
+            _tb: types.TracebackType = exc.__traceback__.tb_next.tb_next  # type: ignore
+            log.error(
+                "filter_urls (field '%s'): ignore %s from callback %s:%s",
+                field_name,
+                repr(exc),
+                _tb.tb_frame.f_code.co_filename,
+                _tb.tb_lineno,
```
Bnyro, line 141: Is there really no simpler way to print a summary of the exception? I couldn't find anything with a quick research, but that still looks a bit too complicated for simply printing an error without the full stack trace

return42: AFAIK, not in this case: we have to inspect the exception's traceback object to find the line of source code within the callback that is responsible for the exception. 

But I agree with you - it looks very ugly, and I felt like it took me an eternity to figure out how to work with that traceback object :-)

PR #6122, opened 2026-05-23
[feat] engines: add flaticon icons engine

Bnyro:
### What does this PR do?
- Unfortunately getting the SVG files requires a paid subscription. Tools like https://addons.mozilla.org/en-US/firefox/addon/flaticon-downloader-bymapree/ can bypass the subscription and download them if you have an account, but without an account, there's no way to get the SVG as far as I can tell.

- !fli bird

### Code of Conduct

### DIFF
--- a/searx/engines/flaticon.py
+++ b/searx/engines/flaticon.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Flaticon_ is a database for icons.
+
+.. _Flaticon: https://www.flaticon.com
+"""
+
+from urllib.parse import urlencode
+
+import typing as t
+
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+about = {
+    "website": "https://www.flaticon.com",
+    "wikidata_id": "Q105283791",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://www.flaticon.com"
+
+categories = ["images", "icons"]
+paging = True
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    args = {
+        "word": query,
+    }
+    params["headers"].update(
+        {
+            # important: query term is not URL encoded in the referer string
+            "Referer": f"{base_url}/search?word={query}",
+            "X-Requested-With": "XMLHttpRequest",
+        }
+    )
+    params["url"] = f"{base_url}/ajax/search/{params['pageno']}?{urlencode(args)}"
+
+
+def _fix_url(url: str) -> str:
+    return url.replace(r"\/", "/")
+
+
+def response(resp: "SXNG_Response"):
+    res = EngineResults()
+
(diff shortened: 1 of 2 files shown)


Bnyro:
Idk why the CI is broken.

return42:
> Idk why the CI is broken.

don't worry .. I just restarted the jobs .. and now they are running successfully :-)

return42 (approved):
I modified some nits: [diff](https://github.com/searxng/searxng/compare/418742a8962efbfbb785a5ebbdeaf683d4a5c3cf..40e0be735c614d89b8aa3919d66e64814843746b)

> Unfortunately getting the SVG files requires a paid subscription.

I also sometimes wonder whether it makes sense to promote such sales platform .. after all, it’s basically no different than if we were providing and maintaining an engine for Amazon or Temu.

If we were to incorporate more such sales platforms, we would need to consider how to differentiate them. It would certainly be difficult to establish a clear distinction, which is why I believe the simplest approach is to mark these platforms as inactive .. and, if appropriate, include a comment within the settings.

Like I have done here: [diff](https://github.com/searxng/searxng/compare/40e0be735c614d89b8aa3919d66e64814843746b..8fb933e0e2974c6e07892f6b3fad212be80953a0) .. if it is OK for you.


----


**I find it difficult to make a decision in this regard, but fundamentally, I am reluctant to perform bug tracking for implementations that extend the reach of commercial providers .. displacing potentially free alternatives**

Bnyro:
> I also sometimes wonder whether it makes sense to promote such sales platform .. after all, it’s basically no different than if we were providing and maintaining an engine for Amazon or Temu.
>
> If we were to incorporate more such sales platforms, we would need to consider how to differentiate them. It would certainly be difficult to establish a clear distinction, which is why I believe the simplest approach is to mark these platforms as inactive .. and, if appropriate, include a comment within the settings.
>
Flaticon is probably the standard "search engine" if you need an icon for something (e.g. a presentation or a website). We can get and use the 512px version for free, which is sufficient for most use cases.

I agree that we shouldn't support their business model, e.g. I would be fine with removing the link to Flaticon in the search result (i.e. the `url`) and replace it with the Url to the 512px image. Not sure if that's that user friendly though.


In my eyes it's already much better for privacy if you can search and find icons from them via SearXNG, so that you don't have to use their awful website. So I wouldn't really call this what we are doing "promoting", it's rather "(ab)using for our needs".

Bnyro:
But yeah, `inactive: true` is fine for me.

Bnyro:
The same would probably also apply for https://github.com/searxng/searxng/pull/6123, with the difference that we can bypass their paywall, so you can get all the icons without subscription (well, the license thing might be conplicated, but fair use is definitely possible).

return42:
> We can get and use the 512px version for free, which is sufficient for most use cases.
> ...
> In my eyes it's already much better for privacy if you can search and find icons from them via SearXNG

As I suspected, it's incredibly difficult to establish universally applicable criteria for this.  From the perspective you mentioned, my argument is no longer valid .. I've removed the `inactive`.

> removing the link to Flaticon in the search result .. Not sure if that's that user friendly though.

No, it isn't user friendly .. given that the non-vector-based representations are free, a reference to the source is also legitimate/desirable.

PR #6123, opened 2026-05-23
[feat] engines: add icons8 image engine

Bnyro:
### What does this PR do?
- implement https://icons8.com
- there's an option `fetch_svg` which loads the source SVG files - it has the caveat that it requires one request per image though (this bypasses the icons8 subscription plan btw, so it's rather a grayzone :) )

- !i8 bird

### Notes
- using multi-request code was inspired by dalf's work in #4827 - without it the engine would be much slower
- multi-request: https://github.com/searxng/searxng/blob/7420706a5074803a78f3c795cf79427746e504e0/searx/network/__init__.py#L101

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/icons8.py
+++ b/searx/engines/icons8.py
@@ -1,0 +1,126 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Icons8_ is a database for icons.
+
+.. _Icons8: https://icons8.com
+"""
+
+from urllib.parse import urlencode
+
+import typing as t
+
+from searx.result_types import EngineResults, ImageRef
+from searx.network import multi_requests, Request
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+about = {
+    "website": "https://icons8.com",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://icons8.com"
+api_url = "https://search-app.icons8.com"
+cdn_url = "https://img.icons8.com"
+
+categories = ["images", "icons"]
+paging = True
+
+icons8_fetch_svg = False
+"""Whether to additionally load the full resolution SVG icon.
+
+.. attention::
+
+   This does one request per icon (!), so be aware that this might be very slow
+   and the engine timeout should be increased.  Multi-requests trigger a flood
+   of requests to the origin; typically, which leads to the SearXNG IP being
+   blocked very quickly.
+"""
+
+results_per_page = 20
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    args = {
+        "amount": results_per_page,
+        "offset": (params["pageno"] - 1) * results_per_page,
+        "term": query,
+    }
+    params["url"] = f"{api_url}/api/iconsets/v7/search?{urlencode(args)}"
+
+
+def _fetch_svgs(ids: list[str]) -> dict[str, str]:
+    responses = multi_requests(
+        [
+            Request.get(
+                f"https://api-icons.icons8.com/siteApi/icons/icon?id={id}&svg=true",
+            )
+            for id in ids
+        ]
+    )
+
+    svgs: dict[str, str] = {}
+    for i, resp in enumerate(responses):
+        if isinstance(resp, Exception):
+            continue
+
+        # the image is a base64 encoded SVG, so we can display it
+        # as data:image
+        b64_svg: str = resp.json()["icon"]["svg"]
+        svgs[ids[i]] = f"data:image/svg+xml;base64,{b64_svg}"
+
+    return svgs
+
+
+# any very big resolution will give us the max-res image
+def _get_image_url(result: dict[str, t.Any], img_format: str, resolution: int = 10000) -> str:
+    return f"{cdn_url}/{result['platform']}/{resolution}/{result['commonName']}.{img_format}"
+
+
+def response(resp: "SXNG_Response"):
+    res = EngineResults()
+    icons = resp.json()["icons"]
+
+    svgs = {}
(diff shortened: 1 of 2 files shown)


return42:
> * using multi-request code

definitely not in line with our *unblood campaign* :) .. the multi-request code was added in a1e8af079 .. currently, there is no use for this, and in my view, this code should have been removed long ago.

Multi-requests trigger a flood of requests to the origin; typically, which leads to the SearXNG IP being blocked very quickly.

In my view, it would be significantly better if we extended the images.html template to allow for the display of links to alternative formats as well .. where `https://api-icons.icons8.com/siteApi/icons/icon?id={id}&svg=true` is such a alternative format.

I am currently in the process of preparing something like this, though I still need a little more time for it.

Bnyro:
> In my view, it would be significantly better if we extended the images.html template to allow for the display of links to alternative formats as well .. where https://api-icons.icons8.com/siteApi/icons/icon?id={id}&svg=true is such a alternative format.
>
That's indeed a good idea - ~that could also be useful for engines that require some specific headers to be set to load images (I think pixiv for example needs specific headers? not sure if it was that engine)~

return42:
> I am currently in the process of preparing something like this

FYI: first step is done:

- https://github.com/searxng/searxng/pull/6142

in a follow up PR I will add the *alternative* formats

Bnyro:
Now that #6153 is merged, I added support for alternative formats: [diff](https://github.com/searxng/searxng/compare/a438470d1291c258c8737607de3b4f745bf789d4..be90d9ce393f8f5280d4b67c5fa431ed31bce738).

searx/engines/icons8.py
```diff
@@ -0,0 +1,119 @@
...
+                url=_get_image_url(result, "png"),
+                subtype="png",
+            ),
+            ImageRef(
+                url=_get_image_url(result, "jpg"),
+                subtype="jpeg",
+            ),
+        ]
+        if result["id"] in svgs:
+            img_url = thumbnail_url = svgs[result["id"]]
```
return42, line 107: `img_url` (aka `Image.img_src`) is a link and can't handle `data:image/svg+xml;base64` 

https://github.com/searxng/searxng/blob/b3e08f2a44640ffc45d5955f5ff43b4a4d555578/searx/templates/simple/result_templates/images.html#L17

https://github.com/searxng/searxng/blob/b3e08f2a44640ffc45d5955f5ff43b4a4d555578/searx/templates/simple/result_templates/images.html#L23

Can we please drop the *fetch_svg* feature: [.. definitely not in line with our unblood campaign](https://github.com/searxng/searxng/pull/6123#issuecomment-4535933388)?

Bnyro: I personally need this option because I often search for SVG icons (due to smaller size and editable). If it's not in SearXNG, I would need to write my own small web app for scraping the `SVG`.

Can't we just tell `image_proxify` to ignore `data:` links? I don't think this will make you a happy, but I'd really like to keep this feature :)

searx/engines/icons8.py
```diff
@@ -0,0 +1,119 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Icons8_ is a database for icons."""
```
Bnyro: Oh, I forgot to put a link here apparently.

searx/engines/icons8.py
return42: > Can't we just tell `image_proxify` to ignore `data:` links?

Seems it already does:

https://github.com/searxng/searxng/blob/b3e08f2a44640ffc45d5955f5ff43b4a4d555578/searx/webapp.py#L307-L315

> I personally need this option because I often search for SVG icons (due to smaller size and editable)

OK, I understand .. however, how can you use (copy?) this SVG if it is a base64 encoded [data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme)?

I tested this *fetch_svg* feature and I was puzzled, because I only see links there that don't work, and I couldn't access the original SVG anywhere.

Bnyro: > OK, I understand .. however, how can you use (copy?) this SVG if it is a base64 encoded [data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme)?
>
You can right-click the image and press "save image as" in Firefox. That just gives you the underlying decoded SVG without the Base64.

image

> I tested this fetch_svg feature and I was puzzled, because I only see links there that don't work, and I couldn't access the original SVG anywhere.
>
Really? I just tested it and it still works. E.g. I managed to get the SVG of the bird in the screenshot above

Bnyro: Note that you have to uncomment all the 3 lines, otherwise it won't load the SVGs or timeout:
```yaml
    fetch_svg: true
    results_per_page: 10
    timeout: 10
```

return42: FYI: I fixed some nits I had: [diff](https://github.com/searxng/searxng/compare/d6215b0ab989f0d944ccd2a47f99e4875fac2d0a..cc35770ad3a7cd0f2f8f5ba386095c7060cc90f1)

Thanks for pointing out "Save Image As .." .. however, the UX is not good with all these *dead links* (links with a data image URL) .. further when I enable the image proxy no more images are shown ..

grafik

Long story told short: Until now, images were always data and URLs were always real addresses on the web. If URLs can now also be images, then this no longer fits with the URL handling we have implemented (python code and html templates).

I already tried out a few things, but then it got out of hand for me.

If we really want to support image data URLs, then we have to put more work into it, maybe you can still come up with something .. I currently don't have any simple solutions for it.

Bnyro: > If we really want to support image data URLs, then we have to put more work into it, maybe you can still come up with something .. I currently don't have any simple solutions for it.
>
I'll look into this when I have time, let's just keep this PR open in idle state until I have a solution for that in a different PR :+1:

PR #6125, opened 2026-05-23
[mod] enable image proxy for public instances

### LINKED ISSUE #4697: Option `server:public_instance` should activate `server:image_proxy`
For better privacy of the instances listed on https://searx.space we should consider to activate image proxy when `server:public_instance` is `true`.

Related:

- https://docs.searxng.org/admin/settings/settings_server.html
- searxng-instances [add-instance template](https://github.com/searxng/searx-instances/blob/54135f280e613540604e11ed577c09d31ff71b20/.github/ISSUE_TEMPLATE/add-instance.yaml?plain=1#L65-L71)
- https://github.com/searxng/searx-instances/discussions/417

vojkovic:
### What does this PR do?

When `public_instance` is set to true, this will also lock the image proxy as enabled. This will provide better privacy for users of the instances on searx.space and also help to prevent a class of CSP errors when onboarding a new instance.

Frankly, I am in favour of removing the toggle for the user to disable this feature from the preferences entirely and leaving it up the instance maintainer's discretion but I have left that out of this PR.

Set `public_instance` to true, and keep `image_proxy` false.

Run `make run`

Verify that the image proxy is still working and the option in preferences is gone.

Before:
Screenshot

After:
Screenshot

With `public_instance` false the behaviour should be the same.

### Related issues

Closes https://github.com/searxng/searxng/issues/4697

### Code of Conduct

Cursor used to search for any required changes to the docs.


Bnyro (approved):

Bnyro:
> Frankly, I am in favour of removing the toggle for the user to disable this feature from the preferences entirely and leaving it up the instance maintainer's discretion but I have left that out of this PR.
>
I agree on that, I don't think any user would change that setting from `on` -> `off`, and if we force public instances to have the setting enabled, users don't manually need to toggle `off` -> `on`. So imo it's fine to remove it in this PR, but let's wait for @return42 .

return42:
> I agree on that, I don't think any user would change that setting from `on` -> `off`, and if we force public instances to have the setting enabled, users don't manually need to toggle `off` -> `on`. So imo it's fine to remove it in this PR, but let's wait for @return42 .

I fully agree. In the past, there were efforts to make as many features as possible optionally available to the user via the preference menu; however, this case demonstrates just how nonsensical that approach can be. The settings menu should be drastically streamlined - and features that concern the administrator (particularly their resources) should not be included there at all.

I have actually already discussed this with @Bnyro: SearXNG needs to become "leaner"  - both in terms of its codebase and its user interface. We do not need to jump into every single niche, no matter how small, and it is perfectly fine if we gradually phase out some of these niche solutions.

vojkovic:
> The settings menu should be drastically streamlined

Absolutely agree. It looks like just a lot of scope creep over the years. A lot of these choices are defensible in isolation but it's probably time the menu had a clean up :)

> features that concern the administrator (particularly their resources) should not be included there at all.

Definitely, what do you think of making the HTTP method (GET/POST) be instance policy too? Same with Query in Title?

DOI / hotkeys / URL format / center alignment can be tucked under an advanced block

Also while I was looking around I found [advanced_search](https://github.com/searxng/searxng/blob/46071a011a386a3b64e94c48ae97ffd9198f8abc/searx/preferences.py#L471-L473)? Do you know what it's for - looks to be unused? Leftover from the oscar theme maybe?

I can start any of this here or in a follow up PR. What do you prefer @return42?

return42:
> Definitely, what do you think of making the HTTP method (GET/POST) be instance policy too? Same with Query in Title?

Sure .. BTW I locked HTTP method on my instance 

https://github.com/searxng/searxng/blob/28ef4f7447debd6f988963c80b3ad15046c65908/searx/settings.yml#L171-L172

> Leftover from the oscar theme maybe?

This really does seem to be from the Oscar theme .. 16a864164 .. a leftover we've been dragging around for 10 years .. definitely a candidate for our *unblood-campaign* :)

lets remove https://github.com/search?q=repo%3Asearxng%2Fsearxng%20advanced_search&type=code

> I can start any of this here or in a follow up PR.

As you prefer .. may we implement topics from above here in this PR and when this is merged we can implement PR for the *advanced block*? ..

> DOI / hotkeys / URL format / center alignment can be tucked under an advanced block

----

Long story told short: I agree with your suggestions .. Just go for :) .. the implementation is partial complex / a mess .. If you stuck at some point, we can sort it out together in the review --> thank you so much for taking on these topics :+1:

vojkovic:
Fantastic, actually - I'm going to hold off on making any more changes here for now - will wait until both this and https://github.com/searxng/searxng/pull/6126 are merged first and then sort out the rest. 👍

return42:
I'm in review with this PR .. In the meantime, I have posted the following text in the 'SearXNG public instance owner' chat ...

*Heads up, if you haven't enabled the image proxy on your public instance for some reason .. when [PR-6125](https://github.com/searxng/searxng/pull/6125) gets merged, the image proxy is enabled (mandatory) for all public instances by default.*

*If the image proxy has not been activated for you so far, it will significantly increase the load of requests and the data volume.*

return42 (approved):
While I was in review with this PR I implemented a `msgspec.Struct` for the preference configuration  .. and added a commit on top of your branch:

no functional change / except the missing online doc which is now available::

    $ make docs.live
    $ xdg-open "http://127.0.0.1:8000/admin/settings/settings_preferences.html"

if you think my commit is better placed in a separated PR, I can send a follow up PR .. or if you wish to squash, we can squash this patch into yours .. whatever you prefer. 

----

> Also while I was looking around I found [advanced_search](https://github.com/searxng/searxng/blob/46071a011a386a3b64e94c48ae97ffd9198f8abc/searx/preferences.py#L471-L473)? Do you know what it's for - looks to be unused? Leftover from the oscar theme maybe

The unused `advanced_search` configuration was not adopted during the typification and does no longer exists.

searx/settings.yml
```diff
@@ -154,14 +154,13 @@ ui:
...
-# preferences:
-#   lock:
+preferences:
+  # Lock arbitrary settings on the preferences page.
+  lock: []
 #     - categories
 #     - language
 #     - autocomplete
-#     - favicon
+#     - favicon_resolver
```
vojkovic, line 163: Also needs an update here :) I think this has been a bug for a while 🤦‍♂️ I missed it in my review
    https://github.com/searxng/searxng/blob/01159b82febe7f9af12fe74dd3a38f614d67257c/searx/templates/simple/preferences.html#L183-L185

searx/settings_defaults.py
```diff
@@ -236,16 +237,13 @@ def apply_schema(settings: dict[str, t.Any], schema: dict[str, t.Any], path_list
         },
         'center_alignment': SettingsValue(bool, False),
         'results_on_new_tab': SettingsValue(bool, False),
-        'advanced_search': SettingsValue(bool, False),
```
vojkovic, line 239: Thanks for removing i also noticed it was here when poking around:

https://github.com/searxng/searxng/blob/01159b82febe7f9af12fe74dd3a38f614d67257c/client/simple/src/js/toolkit.ts#L8

searx/settings.yml
```diff
@@ -154,14 +154,13 @@ ui:
   # URL formatting: pretty, full or host
   url_formatting: pretty
 
-# Lock arbitrary settings on the preferences page.
-#
-# preferences:
-#   lock:
+preferences:
+  # Lock arbitrary settings on the preferences page.
+  lock: []
```
vojkovic, line 159: I was also looking at locking (and hiding) `ui.default_theme` now the oscar theme has been removed. It is a useless dropdown atm.

Would be better instead of removing the feature entirely if someone wants to add new themes in the future.

Maybe best to go in another PR because it doesn't look straightforward?

return42: > Maybe best to go in another PR because it doesn't look straightforward?

yes .. The current situation is somewhat convoluted.  We used to have two themes (Oscar and Simple) but no design selection; then Oscar was deleted, and the Simple theme was expanded to include various designs. Consequently, the underlying implementation regarding themes is currently somewhat ambiguous and is in need of an overhaul .. 

In any case, from an implementation standpoint, we should keep our options open so that we can develop a new theme at any time .. what I have in mind: A new theme can also serve as an exit strategy, should we one day feel that the current "simple theme" is no longer adequate to meet the requirements as they have evolved.

searx/settings.yml
return42: I will fix it in my commit

vojkovic:
Great work. LGTM :)

return42 (approved):
I fixed the issues related to your last comments [[diff]](https://github.com/searxng/searxng/compare/1aed6e8e26479162dbb7d2e0c9758e68c4aec5b3..ead22bb368a1518df3ccfe83dba0485627a562c5) .. From my side, this PR is ready. If you feel the waiting period for the public owners has been sufficient, go ahead and merge the PR.

vojkovic:
I did a couple of measurements and the bandwidth usage ended up being only 1-5GB / day on a popular instance so I don't expect the resource usage to surprise anyone. Didn't measure the difference for cpu and memory but my servers hover at 5-25% utilisation so it can't be that crazy.

image

Issue #4931, opened 2025-06-23
Rework of preferences page


vojkovic:
Leon and I were discussing this last evening, but I think some things could be reworked in the preferences page. One of the big complaints I've heard with searx/ng is that the preferences are too complicated, so I think a rework is necessary.

![Image](https://github.com/user-attachments/assets/76e5b3f7-fcd3-47f4-a213-3c98c2846d8d)

There are also some options on other pages like the center alignment that could be reworked. I also would like to move the description of each preference under the preference name itself for better readability  or have it only appear if the user hovers over it.

The mobile view of the preferences could probably also do with a fix:

![Image](https://github.com/user-attachments/assets/5db14f55-88c7-4256-a67f-e8a3c96addb3)

Happy to make any changes once we can agree what should be modified.

Bnyro:
My two cents: 
- For the default categories bar: a dropdown probably wouldn't work there, because users may want to select multiple categories
- For the unit converter and calculator: These are plugins and by design, SearXNG allows its users to disable plugins individually. I agree it could be moved to the special queries page though.

I very much like the other ideas though:
- Hiding the engine tokens seems like a good idea, it's useless if there are no engines with tokens set up anyways.
- For the DOI rewrite, I still don't know what it does 100% because I never used it. I'm not even sure if it's worth keeping that plugin in SearXNG tbh.

virtadpt:
I think it would be a good idea to move the plugins onto a separate page.  It's counter-intuitive to look at "special queries" when the docs say "plugin."  Took me a few days to figure that out when I first heard about it, other users on my side of things had a similar response.

return42:
About "Default Categories"

yes, drop down would be better / but please have in mind it is an multi selection. 

About "Engine tokens"

leave it as it is for now... I would like to limit the first PR to the UI... background: the preferences are poorly/complicatedly implemented in the core, which affects the entire core of SearXNG --> I am currently working on the core implementation for the preferences... once the core implementation has been revised, we can think again about how we can improve this field in the UI.


About "Plugins & tabs" ..

>  It's counter-intuitive to look at "special queries" when the docs say "plugin." 

Then we have to improve the docs ..

About plugins see [`PluginInfo.preference_section` .. *The value `query` is reserved for plugins that are activated via a keyword as part of a search query*](https://docs.searxng.org/dev/plugins/development.html#searx.plugins.PluginInfo.preference_section): 

[...]

PR #6126, opened 2026-05-23
[enh] simplify and clean up preferences

vojkovic:
### What does this PR do?

This PR is a start to begin to clean up the preferences page. I have started with just two small changes:

- The 'engine tokens' field is only shown now if there are any engines with tokens configured. Offering to drop if @return42 is still working on the [preferences implementation](https://github.com/searxng/searxng/issues/4931#issuecomment-2998947896) as it's only a two line change.

- I have made many changes to the preferences of the DOI Resolver. Primarily, I have combined the previous two settings (one dropdown and one toggle) into just a single dropdown similar to the favicon resolver. I am not sure if this goes against the 'spirit' of the plugins but thought it would be better to discuss it here and maybe we will come up with a better alternative.

Screenshot

`make run`

Go to /preferences under General:

Search when DOI resolver is disabled when off ('-') and enabled when a resolver is selected.

The tokens block should be hidden unless an engine has tokens in settings.

### Related issues

Related: https://github.com/searxng/searxng/issues/4931

### Code of Conduct

Cursor for occasional suggestions and to ask it about edge cases.

### DIFF
--- a/searx/plugins/oa_doi_rewrite.py
+++ b/searx/plugins/oa_doi_rewrite.py
@@ -19,9 +19,6 @@ if typing.TYPE_CHECKING:
     from searx.plugins import PluginCfg
 
 
-ahmia_blacklist: list = []
-
-
 def filter_url_field(result: "Result|LegacyResult", field_name: str, url_src: str) -> bool | str:
     """Returns bool ``True`` to use URL unchanged (``False`` to ignore URL).
     If URL should be modified, the returned string is the new URL to use."""
@@ -29,11 +26,15 @@ def filter_url_field(result: "Result|LegacyResult", field_name: str, url_src: st
     if field_name != "url":
         return True  # use it unchanged
 
+    resolver_url = get_doi_resolver()
+    if not resolver_url:
+        return True  # use it unchanged
+
     doi = extract_doi(result.parsed_url)
     if doi and len(doi) < 50:
         for suffix in ("/", ".pdf", ".xml", "/full", "/meta", "/abstract"):
             doi = doi.removesuffix(suffix)
-        new_url = get_doi_resolver() + doi
+        new_url = resolver_url + doi
         if "doi" not in result:
             result["doi"] = doi
         log.debug("oa_doi_rewrite: [URL field: %s] %s -> %s", field_name, url_src, new_url)
@@ -53,7 +54,7 @@ class SXNGPlugin(Plugin):
             id=self.id,
             name=gettext("Open Access DOI rewrite"),
             description=gettext("Avoid paywalls by redirecting to open-access versions of publications when available"),
-            preference_section="general",
+            preference_section=None,
         )
 
     def on_result(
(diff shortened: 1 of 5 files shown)


searx/plugins/oa_doi_rewrite.py
```diff
@@ -83,7 +87,7 @@ def extract_doi(url):
 
 def get_doi_resolver() -> str:
     doi_resolvers = get_setting("doi_resolvers")
-    selected_resolver = sxng_request.preferences.get_value('doi_resolver')[0]
-    if selected_resolver not in doi_resolvers:
-        selected_resolver = get_setting("default_doi_resolver")
+    selected_resolver = sxng_request.preferences.get_value('doi_resolver')
+    if not selected_resolver or selected_resolver not in doi_resolvers:
+        return ''
```
Bnyro:
```suggestion
        return ""
```
nit: we're currently on our way to replace all single quotes with double quotes when we refactor the files, so please help with this as well here :)

Bnyro: Imo, `' '` should be used for single characters, whereas `"abc"` is for full strings (as seen in many other programming languages)

searx/templates/simple/preferences/doi_resolver.html
```diff
@@ -1,20 +1,17 @@
...
-          {{- doi_resolver_name }} - {{ doi_resolver_url -}}
+                {%- if doi_resolver_name == current_doi_resolver %} selected="selected" {%- endif -%}>
+          {{- doi_resolver_name -}}
         </option>
       {%- endfor -%}
     </select>{{- '' -}}
   </div>{{- '' -}}
   <div class="description">
-    {{- _('Select service used by DOI rewrite') -}}
+    {{- _('Rewrite DOI links in results with open access versions') -}}
```
Bnyro: For many people it's probably not even clear what a "DOI" is, so imo we could very well hide this setting in an advanced section and also quickly explain here that DOI is about "digital media (e.g. books)"

searx/preferences.py
Bnyro, line 1: please double quote :)

Bnyro (commented):
I think `default_doi_resolver` is now unused?

It only seems to be used in
```py
@app.route('/config')
def config():
return {
   ...
   'default_doi_resolver': settings['default_doi_resolver'],
}
```
and some tests. It probably makes sense to still keep a way for the instance admin to set a default DOI resolver

searx/templates/simple/preferences/doi_resolver.html
vojkovic: Yes I agree, unfortunately there's not really a better place yet to put this setting in the time being especially given it's a dropdown.

I changed the description to the one here: https://docs.searxng.org/admin/plugins.html

I think it describes the purpose of the plugin a little better.

Bnyro (approved):

return42:
> * Offering to drop if @return42 is still working on the [preferences implementation](https://github.com/searxng/searxng/issues/4931#issuecomment-2998947896) as it's only a two line change.

I actually worked on a rewrite of the Preferences once, but that was a very long time ago, and at some point, I lost track of it... please don't let that hold you back.

If you're okay with it, could you please introduce the content-related changes via your PRs .. then — provided it seems appropriate to me — I would like to carry out my technical modernizations during the review process, just as I did the typification of the preferences in https://github.com/searxng/searxng/pull/6125.

return42:
FYI: when #6125 has been merged, I will rebase this PR and start my review ..

return42:
FYI: last force push was just a rebase to solve conflict with current master / I will continue my review later.

PR #6127, opened 2026-05-23
[mod] qwant: remove web lite and improve request spoofing

Bnyro:
### What does this PR do?
- <https://lite.qwant.com> seems to be dead.
- The request parameters were changed to match the ones from the Qwant website.
- Qwant is now set to inactive by default due to its strict rate-limits

- !qw test

### Code of Conduct

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -6,7 +6,6 @@ engineered by reading the network log of https://www.qwant.com/ queries.
 For Qwant's *web-search* two alternatives are implemented:
 
 - ``web``: uses the :py:obj:`api_url` which returns a JSON structure
-- ``web-lite``: uses the :py:obj:`web_lite_url` which returns a HTML page
 
 
 Configuration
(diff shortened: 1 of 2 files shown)


return42 (approved):
I've made a few adjustments to calm down the static type checker and to standardize the string notation -> [diff](https://github.com/searxng/searxng/compare/2b8f691f147664716121e09d491c2159ca60075b..28f6937755071d1492f7b0bc4fa86060cc9ec01f)

> Qwant is now set to inactive by default due to its strict rate-limits

The qwant engines are working very well on my local IP (for now), I would like to test them in the field first before we decide to disable/inactivate them; latter seems a bit too drastic to me.

Bnyro:
> The qwant engines are working very well on my local IP (for now), I would like to test them in the field first before we decide to disable/inactivate them; latter seems a bit too drastic to me.
>
I didn't see them working on my own or any public instance, so I assumed the rate-limits are just very strict. But good to know it works for you :+1:

return42:
>> I would like to test them in the field first before we decide to disable

> I didn't see them working on my own or any public instance, so I assumed the rate-limits are just very strict.

I have to agree with you, Qwant isn't working on my public IP either :-o

- https://github.com/searxng/searxng/pull/6148

Issue #6129, opened 2026-05-24
Search results report 0 results even though there are results passed in the json payload.


kylearoberts:
Search results report 0 results even though there are results passed in the json payload.

### How To Reproduce?

http://localhost:8080/search?q=query&format=json

### Expected behavior

"number_of_results": number of results

### Screenshots & Logs

```json
[...]

Issue #5034, opened 2025-07-22
number_of_results in JSON always response in 0


n3r0-b1n4ry:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
searxng-docker - 2025.7.19-b95a3e9

**How did you install SearXNG?**
`docker run --name searxng -d -p 8888:8080 -v "./config/:/etc/searxng/" -v "./data/:/var/cache/searxng/" searxng/searxng:latest`
 My **settings.yml** is default

**What happened?**

The field **number_of_results** in JSON format results is always responses in 0 
**How To Reproduce**

Please check how field **number_of_results** return

**Expected behavior**

all setting is default, query input from keyboard to WebUI

**Screenshots & Logs**

My uri: search?q=vietnam&category_news=1&category_it=1&category_general=1&category_science=1&pageno=1&language=auto&time_range=&safesearch=0&format=json
<img width="1401" height="278" alt="Image" src="https://github.com/user-attachments/assets/23c97d6a-1ee6-4bb7-b132-fefe7407fcd5" />

**Additional context**

Issue #2987, opened 2023-11-03
JSON `number_of_results` is wrong


miranyx:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
023.11.1+b5a8ddfec, no fork
**How did you install SearXNG?**
I am using searxng-docker inside a docker-compose file.
**What happened?**
On json the field `number_of_results` is 0 even though there are results.

**How To Reproduce**
Search a term. Press on the json results.

**Expected behavior**
The number is equal to the number of returned results.

**Screenshots & Logs**
![20231103_13h42m22s_grim](https://github.com/searxng/searxng/assets/64905406/4b1bf1ee-aeeb-47c0-92fa-ceb8a033a376)

unixfox:
I don't think it's a bug.

This item "number_of_results" is a what the number of results the engine think there are for a specific keyword. For example, if you type something on Google, you will get "About XXX results":

![image](https://github.com/searxng/searxng/assets/4016501/137b6e3c-9a02-47e0-921a-785b6e55a65b)

So we return this number to the user. It's not the number of results available through the API or in the search page of searxng.

------

Related:

- https://github.com/searxng/searxng/issues/2987
- https://github.com/searxng/searxng/issues/2457

miranyx:
So when should it increase since when I search for e. g: `ok` it is also still 0?

unixfox:
Some engine may not give us the number of actual results, so in this case we return "0" because we do not know the amount.

For Google, it's very specific because the parsing for the number of results is broken, that's why you are getting "0".

miranyx:
Ah okay, so not a bug will close it then

Issue #2457, opened 2023-05-24
number_of_results always set to 0 in API results


chprasad:
I am seeing `number_of_results` always set to 0 in results returned by API calls to endpoints described at https://docs.searxng.org/dev/search_api.html. 

Of course, the actual number of results returned is non-zero. 

Version: 2023.4.3+464e6966.

unixfox:
which engine did you use?

chprasad:
I did not limit the search to a specific engine. I did constrain the category to 'science'.

unixfox:
please send the instance url used if it's a public instance

gorootde:
Same for me. Even if I set the engine only to DuckDuckGo it returns zero.

return42:
The `number_of_results` indicate the number of hits in the INdex, they do not indicate how many results are in the answer.

In the past, search engines such as google or ddg had an indication on the first page of a search term of how many hits there were for this term in total in their index.

This information was added up in SearXNG and delivered under ``number_of_results`` ... nowadays the search engines no longer indicate how many hits there are in the index and so this field in SearXNG is also superfluous.

--> This field must be removed from SearXNG.

return42:
- https://github.com/searxng/searxng/pull/6130

PR #6130, opened 2026-05-25
[unbload] drop meaningless field ``number_of_results_xpath`` from results

return42:
In the result-list, the ``number_of_results`` indicate the number of hits in the Index, they do not indicate how many results are in the answer.

In the past, search engines such as google or ddg had an indication on the first page of a search term of how many hits there were for this term in total in their index.

This info was added up in SearXNG and delivered under ``number_of_results``. Nowadays the search engines no longer indicate how many hits there are in the index and so this field in SearXNG is also superfluous.

`make run` and test modified engines

### Related issues

- https://github.com/searxng/searxng/issues/2457#issuecomment-2566181574
- https://github.com/searxng/searxng/issues/2987
- https://github.com/searxng/searxng/issues/5034
- https://github.com/searxng/searxng/issues/6129

### Code of Conduct

no AI used

### DIFF
--- a/searx/engines/ahmia.py
+++ b/searx/engines/ahmia.py
@@ -39,7 +39,6 @@ url_xpath = './h4/a/@href'
 title_xpath = './h4/a[1]'
 content_xpath = './/p[1]'
 correction_xpath = '//*[@id="didYouMean"]//a'
-number_of_results_xpath = '//*[@id="totalResults"]'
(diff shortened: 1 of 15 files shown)


Bnyro (approved):
Agreed, very good decision imo:
- that number doesn't give any value to the user
- it takes additional efforts to implement it per engine
- almost no engine supports it anyways

PR #6132, opened 2026-05-25
[mod] engine: MyMemory Translated - typification and html to text

return42:
The implementation is normalized, type annotations are applied, and the results are freed from the HTML markup (which is partially present).

    !tl en-de typification

### Code of Conduct

  no AI used


Bnyro (approved):
Lgtm and works fine 👍

PR #6133, opened 2026-05-25
[fix] bing: Allow redirects for bing_images and bing_videos

hubutui:
In some regions where geoblocking is employed (e.g. China), www.bing.com redirects to the regional version of Bing. Apply the same workaround as bing.py so image and video search work in those regions.

### What does this PR do?

set `params['allow_redirects'] = True` for bing images and bing videos to make them work in China mainland.

### Code of Conduct

- [ ] **I hereby confirm that this PR conforms with the [AI Policy].**

  If I have not used AI tools for working on the changes in this PR, I will

This is a minor change.

### DIFF
--- a/searx/engines/bing_images.py
+++ b/searx/engines/bing_images.py
@@ -66,6 +66,10 @@ def request(query, params):
 
     params["url"] = base_url + "?" + urlencode(query_params)
 
+    # in some regions where geoblocking is employed (e.g. China),
+    # www.bing.com redirects to the regional version of Bing
+    params['allow_redirects'] = True
+
     return params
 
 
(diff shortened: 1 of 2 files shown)


return42:
To which domain is the redirection taking place?

.. you can overwrite the `base_url` in your local setting.

hubutui:
it will redirect to cn.bing.com. what do mean overwrite the `base_url` in the setting?

return42:
In your local settings.yml file:

```yaml
  - name: bing
    base_url: https://cn.bing.com/search
```

hubutui:
cool, it works. maybe we should also update this: https://github.com/searxng/searxng/blob/b0d8af96bf0eeeb87ab900a6259b65554c0c71ac/searx/engines/bing.py#L114-L116

return42:
@hubutui thanks for pointing out .. please follow #6134

Issue #5243, opened 2025-09-25
Add Chinese Bing Search Engine


ajdgsu:
**Working URL to the engine**
https://cn.bing.com/ (Must use Chinese ip to access)

**Why do you want to add this engine?**
Although we have added several Chinese search engines, but these engines have a lot of limitations. The Bing search engine doesn't have any noticeable limitations, and you can get much more Chinese results than the original Bing, plus access speed is faster than the original Bing with Chinese ip.

**Features of this engine**
Same as original Bing except much more Chinese results

| www.bing.com | cn.bing.com |
|-------|-------|
| <img src="https://github.com/user-attachments/assets/5963e377-bd6b-41a9-be0c-1bb8d14ec9bd" height="400"> | <img src="https://github.com/user-attachments/assets/38fd9b0a-008c-4287-a594-fb856dc2c17a" height="400"> |

**How can SearXNG fetch the information from this engine?**
Please note that **cn.bing.com** must be accessed with a Chinese ip, if you access it with an ip from another country, Bing will automatically redirect to www.bing.com. Similarly, if you access www.bing.com with a Chinese ip, Bing will also automatically redirect to **cn.bing.com**

Same as original Bing Engine method, only need to change www.bing.com to **cn.bing.com**

**Applicable category of this engine**
To be honest, I'm not sure this amount of modification is worth becoming a new search engine, perhaps we should add a switch to the original engine for choosing whether to use **cn.bing.com** for searches.

**Additional context**
In fact, if you deploy SearxNG to a Chinese ip server and try to search with Bing, you'll get a empty return result due to the 302 redirect issue I mentioned before. Since SearxNG uses www.bing.com for search, and accessing that URL with a Chinese ip will be asked to redirect to **cn.bing.com** for searching. But SearxNG doesn't follow the request to redirect to **cn.bing.com** for searching, instead it directly returns the 302 redirect as a search result (obviously this is an empty result)

Here's an example:
```
curl -v https://www.bing.com/search?q=ubuntu+24.04&pq=ubuntu+24.04
```
```
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
[...]

PR #6134, opened 2026-05-25
[fix] bing engines - geoblocking in China

return42:
In regions like China, the domain must be adjusted to avoid a redirect.

Suggested / tested by @hubutui in https://github.com/searxng/searxng/pull/6133#issuecomment-4534637069

### Related issues

- https://github.com/searxng/searxng/issues/5243
- https://github.com/searxng/searxng/pull/5324
- https://github.com/searxng/searxng/pull/6133

### Code of Conduct

  no AI used

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -531,18 +531,22 @@ engines:
     engine: bing
     shortcut: bi
     disabled: true
+    # base_url: https://cn.bing.com  # for instances hosted in China
 
   - name: bing images
     engine: bing_images
     shortcut: bii
+    # base_url: https://cn.bing.com  # for instances hosted in China
 
   - name: bing news
     engine: bing_news
     shortcut: bin
+    # base_url: https://cn.bing.com  # for instances hosted in China
 
   - name: bing videos
     engine: bing_videos
     shortcut: biv
+    # base_url: https://cn.bing.com  # for instances hosted in China
 
   - name: bitchute
     engine: bitchute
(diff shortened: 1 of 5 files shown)


searx/settings.yml
```diff
@@ -531,18 +531,22 @@ engines:
     engine: bing
     shortcut: bi
     disabled: true
+    # base_url: https://cn.bing.com/search
```
Bnyro: Maybe we should also add a comment that this is only for china? otherwise people e.g. from Europe might try to uncomment it and don't know what's going on

searx/settings.yml
```diff
@@ -531,18 +531,22 @@ engines:
...
 
   - name: bing images
     engine: bing_images
     shortcut: bii
+    # base_url: https://cn.bing.com/images/async
```
Bnyro: In my opinion the path `/images/async` should be hidden in the engine implementation, so that `base_url` is really only the host part of the URL without the path. That also makes it harder for admins to provide a wrong value by accident.

At least according to your changes, the paths are the same for cn.bing.com and bing.com (and probably also other subdomains of Bing).

return42:
@Bnyro thanks for your suggestions .. modified accordingly [diff](https://github.com/searxng/searxng/compare/0d46189f4694999389ba2db6ffe0a87162517276..ed547e667f19cd5ee8799476baf776816f42fd6b)

Bnyro (approved):

PR #6142, opened 2026-05-25
[mod] typification of SearXNG: add new result type Image

return42:
### What does this PR do?

- implement result type class `searx.result_types.image.Image`
- move CSS for the image details to `src/less/result_types/image.less` 

...

- Python class:   searx/result_types/image.py
- Jinja template: searx/templates/simple/result_templates/images.html
- CSS (less)      client/simple/src/less/result_types/image.less

    !images foo

### Code of Conduct

  no AI used


searx/result_types/image.py
```diff
@@ -0,0 +1,44 @@
...
+    resolution: str = ""
+    """The resolution of the image (e.g. ``1920 x 1080`` pixel)"""
+
+    img_format: str = ""
+    """The format of the image (e.g. ``png``)."""
+
+    source: str = ""
+    """Source of the image."""
+
+    filesize: str = ""
```
Bnyro, line 42: I think the `author` field is missing?

return42: This and other fields are inherited from class [MainResult](https://docs.searxng.org/dev/result_types/main/mainresult.html#searx.result_types._base.MainResult.author).

Bnyro: I didn't know author is a MainResult field, "schon wieder was gelernt" ;)

Bnyro (approved):

searx/result_types/image.py
return42: > I didn't know author is a MainResult field

To be complete ..

In the past, we had only one type (the dynamic `dict` type) where fields were set "arbitrarily"; which fields actually get rendered was (and currently still is) determined solely by the respective HTML template. Thus, there were fields such as `audio_src`, which is rendered in `default.html` (corresponding to `MainResult`), but not, for instance, in [`code.html`](https://github.com/searxng/searxng/blob/master/searx/templates/simple/result_templates/code.html).

With the inheritance (the static types) we are currently introducing, `MainResult` serves as the base for all other result types to be rendered in the [area-main-results](https://docs.searxng.org/dev/result_types/index.html#area-main-results) .. such as `Code`, for instance.

https://github.com/searxng/searxng/blob/28ef4f7447debd6f988963c80b3ad15046c65908/searx/result_types/code.py#L44

This also makes `audio_src` a property of the `Code` class .. a real drawback (a flaw in the model) that can only be resolved once all engines / all results from the engines have been migrated to the class model.

PR #6144, opened 2026-05-25
[feat] engines: add zapmeta general search engine

Bnyro:
### What does this PR do?
- adds zapmeta, an english general search engine (found at https://www.searchenginemap.com/)
- results are quite good in my opinion, probably uses Bing?

- !zapmeta test

### Code of Conduct

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -174,6 +174,10 @@ number, but an offset.'''
 first_page_num = 1
 '''Number of the first page (usually 0 or 1).'''
 
+send_page_num_on_first_page = True
+'''Whether to include the page number in the request for the first page.
+This can help if an engine blocks request that send a page number for the first page.'''
+
 time_range_support = False
 '''Engine supports search time range.'''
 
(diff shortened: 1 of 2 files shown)


searx/engines/xpath.py
return42, line 24: new option needs to be added to the online docs https://docs.searxng.org/dev/engines/xpath.html

searx/engines/xpath.py
```diff
@@ -174,6 +174,10 @@
 first_page_num = 1
 '''Number of the first page (usually 0 or 1).'''
 
+send_page_num_on_first_page = True
```
return42, line 177: On new (and modified) code, please let us use (as possible and sensible) type hints (`bool`) and strings in double quotation marks.

Can you implement a follow up PR to add the same option to the [`json_engine.py`](https://github.com/searxng/searxng/blob/master/searx/engines/json_engine.py)?

searx/engines/xpath.py
Bnyro: Good catch, thanks!

searx/engines/xpath.py
Bnyro: Sure, done in ..

- https://github.com/searxng/searxng/pull/6200

PR #6146, opened 2026-05-25
[feat] engines: add dogpile (general, news, images, videos) 

Bnyro:
### What does this PR do?
- add support for the Dogpile search engine (found at https://seirdy.one/posts/2021/03/10/search-engines-with-own-indexes/)
- it seems to use the same index as startpage (because results are similar and they share the `qadf` request parameter)

- !dogpile test
- !dogi test
- !dogn test
- !dogv test

### Code of Conduct

### DIFF
--- a/searx/engines/dogpile.py
+++ b/searx/engines/dogpile.py
@@ -1,0 +1,101 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Dogpile is a metasearch engine by the American advertising company `System1`_.
+
+.. _System1: https://system1.com/
+"""
+
+import typing as t
+from datetime import datetime, timezone
+import html
+
+from searx.utils import format_duration, html_to_text, humanize_number
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://www.dogpile.com",
+    "wikidata_id": "Q3595363",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+paging = True
+safesearch = True
+
+categories = ["general"]
+dogpile_categ = "search"
+"""Category to search in. Can be either "search", "images", "videos" or "news"."""
+
+
+base_url = "https://www.dogpile.com"
+safe_search_map = {0: "none", 1: "moderate", 2: "heavy"}
+
+
+def init(_):
+    if dogpile_categ not in ("search", "images", "videos", "news"):
+        raise ValueError("invalid search type: %s" % dogpile_categ)
+
+
+def request(query: str, params: "OnlineParams"):
+    params["url"] = f"{base_url}/api/{dogpile_categ}"
+    params["method"] = "POST"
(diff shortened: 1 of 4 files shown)


return42 (changes requested):
```
  File "/searxng/searx/search/processors/abstract.py", line 155, in init_engine
    init_ok = self.engine.init(eng_setting)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/searxng/searx/engines/dogpile.py", line 40, in init
    raise ValueError("invalid search type: %s" % dogpile_categ)
ValueError: invalid search type: general
```

BTW can you rebase to solve the conflict in the settings.yml / thanks 👍

Bnyro:
Rebased. Sorry for the crash in `settings.yml`, I first tested `general` search and then added the other categories step by step without testing if `general` still works

searx/engines/dogpile.py
```diff
@@ -0,0 +1,100 @@
...
+                    url=result["clickUrl"],
+                    title=html_to_text(result["title"]),
+                    content=html_to_text(result["description"]),
+                )
+            )
+        elif dogpile_categ == "news":
+            res.add(
+                res.types.MainResult(
+                    url=result["clickUrl"],
+                    title=html_to_text(result["title"]),
```
return42: Just to let you know / I will fix it in my PR review .. (the issue is limited to "News")

God knows why, but they’re passing HTML entities in the JSON string through to their website .. But that’s not even the worst part; the worst part is that they also cut off these entities at the end of the string, thereby breaking them 🤦 .. just take a look at the end of the `description` string from the examples below:  .. 

    model have long prized. The &lt;b&gt;BMW&lt ...'
                                               ^---> the `;` is missed 

```
(Pdb) result["title"]
"From Throttle Auctions: Rare 'Clown Shoe' &lt;b&gt;BMW&lt;/b&gt; Surfaces With a One-of-One Claim"
(Pdb) result["description"]
'A 2002 &lt;b&gt;BMW&lt;/b&gt; M Coupe billed as a singular example of its kind has come to market, carrying a configuration that enthusiasts of the model have long prized. The &lt;b&gt;BMW&lt ...'
```

image

searx/engines/dogpile.py
```diff
@@ -0,0 +1,100 @@
...
+                    content=html_to_text(result["description"]),
+                    thumbnail=result["thumbnailUrl"],
+                    publishedDate=datetime.fromtimestamp(result["date"], tz=timezone.utc),
+                )
+            )
+        elif dogpile_categ == "videos":
+            res.add(
+                res.types.MainResult(
+                    template="videos.html",
```
return42: For result types that do not yet have their own type definition -- those where we still have to set the `template` ourselves -- we must use the `LegacyResult` class.

return42 (approved):
I’ve made a few minor [modifications](https://github.com/searxng/searxng/compare/6d008fcd27b3f88c3719c4c67093069fce26c6e7..90d36ca1929a103703c35e0a9d7bf7f620eb30e1) .. apart from that, it’s a pretty usable engine 👍

searx/engines/dogpile.py
Bnyro: Oh, I missed that when testing - I thought doing just `html_to_text` would be fine, but apparently it's not. Funny engine...

Bnyro:
Thanks for the review :+1:

PR #6150, opened 2026-05-26
[feat] engines: add fireball engine (general, news, videos)

Bnyro:
### What does this PR do?
- adss https://fireball.com/, a Germany-based search engine

- !fire germany
- !firen germany
- !firev germany
- test different languages

### Code of Conduct

### DIFF
--- a/searx/engines/fireball.py
+++ b/searx/engines/fireball.py
@@ -1,0 +1,176 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Fireball_ is a Germany-based, privacy-focused search engine.
+
+It likely doesn't have its own index, but it's unclear where its results come from.
+
+.. _Fireball: https://fireball.com
+"""
+
+import typing as t
+
+from datetime import datetime
+from urllib.parse import urlencode
+
+from searx.enginelib import EngineCache
+from searx.exceptions import SearxEngineAPIException
+from searx.extended_types import SXNG_Response
+
+from searx.result_types import EngineResults
+from searx.network import post
+from searx.utils import html_to_text
+
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://fireball.com",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+base_url = "https://fireball.com"
+categories = ["general"]
+fireball_category = "web"  # values: "web", "news", "videos"
+
+paging = False
+safesearch = True
+
+safe_search_map = {0: "off", 1: "moderate", 2: "strict"}
+
+CACHE: EngineCache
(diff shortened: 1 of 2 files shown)


searx/engines/fireball.py
```diff
@@ -0,0 +1,170 @@
...
+                res.types.MainResult(
+                    url=result["url"],
+                    title=html_to_text(result["title"]),
+                    content=html_to_text(result["description"]),
+                    thumbnail=thumbnail or "",
+                    publishedDate=published_date,
+                )
+            )
+        elif fireball_category == "videos":
+            thumbnail = result.get("thumbnail", {}).get("original")
```
return42: The image URLs from the `src` field seem not to work (not in the video category at least) [example URL](https://fireball.com/images/aHR0cHM6Ly9pbWdzLnNlYXJjaC5icmF2ZS5jb20vdzh2a3lTVk5UNTg3Wjh3VEZOckkwR01NdDR3WFlUd1RYV1V5YXJnbnhKMC9yczpmaXQ6MjAwOjIwMDoxOjAvZzpjZS9hSFIwY0hNNkx5OXBMbmwwL2FXMW5MbU52YlM5MmFTOVIvUTNaeFFVczRUR0pOYnk5dC9ZWGh5WlhOa1pXWmhkV3gwL0xtcHdadw==) .. if `original` is set, the image could be loaded from this URL.

return42 (approved):
A really good engine... fun fact: it's difficult to use on their website... only after integrating it into SearXNG can I access their news, videos, and images 😆 

I've made a few minor changes... but beside, the engine is truly perfect... Thanks!

searx/engines/fireball.py
Bnyro: Good point - fixed.

Bnyro: Seems like the engine is slowly dying - it's not even possible to see the videos results on the fireball.de website

Bnyro:
> A really good engine... fun fact: it's difficult to use on their website... only after integrating it into SearXNG can I access their news, videos, and images 😆
> 
> I've made a few minor changes... but beside, the engine is truly perfect... Thanks!
>
Yeah, I was really surprised when finding `videos`, `images` and `news` results in the JSON response :rofl: 

Funnily, you can visit URLs like <https://fireball.com/search/?f=images&q=test> which show a placeholder for images, but don't load the actual images...


This is the first time where we can say without a doubt, that the SearXNG implementation is better than the official website :tada:

searx/engines/fireball.py
return42: > Good point - fixed.

FYI you reverted the modification from mine .. see diff from you last force push .. https://github.com/searxng/searxng/pull/6150#event-26590466009

Bnyro: Oh, sorry. I didn't see that you pushed changes.

Bnyro: Created https://github.com/searxng/searxng/pull/6240 with your changes.

PR #6153, opened 2026-05-27
[mod] image results: add list of alternative formats

return42:

To test alternatives formats apply patch from below, query ``!flaticon bmw`` and
open the detail view for the image.

```diff
diff --git a/searx/engines/flaticon.py b/searx/engines/flaticon.py
index 06b6a8e25..d88388705 100644
--- a/searx/engines/flaticon.py
+++ b/searx/engines/flaticon.py
@@ -8,7 +8,7 @@ from urllib.parse import urlencode

 import typing as t

-from searx.result_types import EngineResults
+from searx.result_types import EngineResults, ImageRef

 if t.TYPE_CHECKING:
     from searx.extended_types import SXNG_Response
@@ -61,6 +61,14 @@ def response(resp: "SXNG_Response"):
                 thumbnail_src=_fix_url(result["png"]),
                 img_src=_fix_url(result["png512"]),
                 author=result["team_name"],
+                formats=[
+                    ImageRef(label="PNG 100x100", url="https://example.org/test.png", subtype="png"),
+                    ImageRef(label="SVG", url="https://example.org/test.svg", subtype="svg+xml"),
+                    ImageRef(url="https://example.org/test.jpg", subtype="jpeg"),
+                    ImageRef(url="https://example.org/test.bmp", subtype="bmp"),
+                    ImageRef(url="https://example.org/test.ico", subtype="x-icon"),
+                    ImageRef(url="https://example.org/test.tif", subtype="tiff"),
+                ],
             )
         )
```

grafik

### Related issues

- https://github.com/searxng/searxng/pull/6123#issuecomment-4535933388

### Code of Conduct

  No AI used.

### DIFF
--- a/searx/result_types/image.py
+++ b/searx/result_types/image.py
@@ -7,14 +7,51 @@ template.
    :members:
    :show-inheritance:
 
+.. autoclass:: ImageRef
+   :members:
+
 """
+# pylint: disable=too-few-public-methods
+__all__ = ["Image", "ImageRef"]
 
-__all__ = ["Image"]
-
+import types
 import typing as t
+from collections.abc import Callable
+
+import msgspec
+
+from ._base import MainResult, Result, log, LegacyResult
+
+MimeSubType = t.Literal["png", "svg+xml", "jpeg", "bmp", "x-icon", "tiff"]
+
+MIMESUB: dict[MimeSubType, str] = {
+    "png": "PNG",
+    "svg+xml": "SVG",
+    "jpeg": "JPG",
+    "bmp": "BMP",
+    "x-icon": "ICO",
+    "tiff": "TIF",
+}
 
 
-from ._base import MainResult
+class ImageRef(msgspec.Struct, kw_only=True):
+    """Reference to an (alternative) image format"""
+
+    url: str
+    """URL of the image reference."""
+
+    subtype: MimeSubType
+    """Subtype (mimetype) of the image format."""
+
+    label: str = ""
+    """Label of the reference, default is build from the uppercase of
+    :py:obj:`Image.ImageRef.subtype`."""
+
+    mtype: t.Literal["image"] = "image"
+
+    def __post_init__(self):
+        if not self.label:
+            self.label = MIMESUB.get(self.subtype, self.subtype.upper())
 
 
 @t.final
@@ -42,3 +79,29 @@ class Image(MainResult, kw_only=True):
     filesize: str = ""
     """Size of bytes in :py:obj:`human readable <searx.humanize_bytes>` notation
     (e.g. ``1MB`` for ``1024*1024`` Bytes filesize)."""
+
+    formats: list[ImageRef] = []
+    """List of links to alternative image formats."""
+
+    def filter_urls(self, filter_func: "Callable[[Result | LegacyResult, str, str], str | bool ]"):
+
+        for _ref in self.formats[:]:
+            _name = f"Image.formats:{_ref.label}"
+            try:
+                _url = filter_func(self, _name, _ref.url)
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                # pylint: disable=no-member
+                _tb: types.TracebackType = exc.__traceback__.tb_next.tb_next  # type: ignore
+                _fn = _tb.tb_frame.f_code.co_filename
+                _lno = _tb.tb_lineno
+                log.error("filter_urls: [%s] ignore %s from callback %s:%s", _name, repr(exc), _fn, _lno)
+                continue
+
+            if isinstance(_url, str):
+                log.debug("filter_urls: [%s] URL %s -> %s", _name, _ref.url, _url)
+                _ref.url = _url
+            elif not _url:
+                log.debug("filter_urls: [%s] drop ref %s", _name, _ref)
+                self.formats.remove(_ref)
+
+        return super().filter_urls(filter_func)
--- a/searx/templates/simple/result_templates/images.html
+++ b/searx/templates/simple/result_templates/images.html
@@ -51,7 +51,13 @@
       <hr>
       <p class="result-author">{{ _label(_("Author"), result.author) }}</p>
       <p class="result-resolution">{{ _label(_("Resolution"), result.resolution) }}</p>
-      <p class="result-format">{{ _label(_("Format"), result.img_format) }}</p>
+      <p class="result-format">
+        <span>{{ _("Image formats") }}:</span>
+        {{- "" -}}<a {{ _target(results_on_new_tab) }} href="{{ result.img_src }}">{{- result.img_format or _("original format") -}}</a>
+        {%- for ref in result.formats -%}
+          &nbsp;| <a {{ _target(results_on_new_tab) }} href="{{ ref.url }}">{{ ref.label }}</a>
+        {%- endfor %}
+      </p>
       <p class="result-filesize">{{ _label(_("Filesize"), result.filesize) }}</p>
       <p class="result-source">{{ _label(_("Source"), result.source) }}</p>
       <p class="result-engine">{{ _label(_("Engine"), result.engine) }}</p>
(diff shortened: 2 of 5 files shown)


searx/result_types/image.py
```diff
@@ -7,14 +7,40 @@
...
-from ._base import MainResult
+    label: str = ""
+    """Label of the reference, default is build from the uppercase of
+    :py:obj:`Image.ImageRef.subtype`."""
+
+    mtype: t.Literal["image"] = "image"
+
+    def __post_init__(self):
+        if not self.label:
+            self.label = self.subtype.upper()
```
Bnyro: `subtype` seems quite useless here, because it's only used as a fallback for `label`, wouldn't it be easier to just let the engine implementor set `label` directly? E.g. they set `label="SVG"`, and done.

searx/result_types/image.py
```diff
@@ -42,3 +68,29 @@ class Image(MainResult, kw_only=True):
...
+        for _ref in self.formats[:]:
+            _name = f"Image.formats:{_ref.label}"
+            try:
+                _url = filter_func(self, _name, _ref.url)
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                # pylint: disable=no-member
+                _tb: types.TracebackType = exc.__traceback__.tb_next.tb_next  # type: ignore
+                _fn = _tb.tb_frame.f_code.co_filename
+                _lno = _tb.tb_lineno
+                log.error("filter_urls: [%s] ignore %s from callback %s:%s", _name, repr(exc), _fn, _lno)
```
Bnyro, line 97: As we start doing this in multiple places, we could almost write a helper function that we could call like `log.print_exception(f"filter_urls: [{_name}]", exc)` or similar

searx/result_types/image.py
Bnyro: For SVG that's a bit unintuitive, because most users probably know what `SVG` is, but not what `SVG+XML` is. So I don't think the `+XML` should be displayed (is there any other way to store SVGs than in XML anyways?).

searx/templates/simple/result_templates/images.html
```diff
@@ -1,28 +1,71 @@
...
+    <div class="result-images-labels">
+      <h4>{{ result.title | striptags }}</h4>
+      <p class="result-content">
+        {%- if result.content %} {{ result.content | striptags }}
+        {%- else %} &nbsp;
+        {%- endif -%}
+      </p>
+      <hr>
+      {%- if result.formats %}
+        <p class="result-formats"><span>{{ _("Alternative formats") }}:</span>
```
Bnyro: This `<p>` is never closed? :)

searx/templates/simple/result_templates/images.html
```diff
@@ -1,28 +1,71 @@
...
+          {%- for ref in result.formats -%}
+            <a {{ _target(results_on_new_tab) }} href="{{ ref.url }}">{{ ref.label }}</a>
+            {%- if not loop.last %} | {% endif -%}
+          {%- endfor %}
+          <hr>
+      {%- endif %}
+      <p class="result-author">{{ _label(_("Author"), result.author) }}</p>
+      <p class="result-resolution">{{ _label(_("Resolution"), result.resolution) }}</p>
+      <p class="result-format">{{ _label(_("Format"), result.img_format) }}</p>
```
Bnyro: I wonder if we shouldn't remove this `format` displayed here and instead display the `img_src` with the `img_format` like it would be an `ImageRef`?

So something like
```jinja
      <p class="result-formats"><span>{{ _("Image formats") }}:</span></p>
       <a {{ _target(results_on_new_tab) }} href="{{ result.img_srcl }}">{{ result.img_src or _("Default") }}</a>
          
         {%- for ref in result.formats -%}
            <a {{ _target(results_on_new_tab) }} href="{{ ref.url }}">{{ ref.label }}</a>
            {%- if not loop.last %} | {% endif -%}
          {%- endfor %}
```

Bnyro: (sorry for the broken formatting)

Bnyro (commented):
Really cool, can't wait to use it. I already have some candidates that have multiple image sources :)

searx/result_types/image.py
return42: Not sure I understood you correctly .. In this data structure, `subtype` is mandatory, the examples from the patch show a minimal use case where `tiff` is the subtype and the name "TIFF" of the link is build from:

```python
ImageRef(url="https://example.org/test.tif", subtype="tiff"),
```

Another use case shows, where it make sense to additional set the label .. assume you have different formats in size:

```python
ImageRef(label="PNG 100x100", url="https://example.org/test.png", subtype="png"),
```
.. or the subtype is not what you want to use for the link name:

```python
ImageRef(label="SVG", url="https://example.org/test.svg", subtype="svg+xml"),
```

searx/result_types/image.py
return42: Another place we inspect the traceback is here:

https://github.com/searxng/searxng/blob/780ee325641c756e220a3820286221820d6c76c5/searx/result_types/_base.py#L131-L143

but `log.print_exception` would be a method of the [logger object](https://docs.python.org/3/library/logging.html#logger-objects) that not exists :-( 

Another drawback in your example: Modern f-strings have the disadvantage compared to the `%s` notation that they are always evaluated, regardless of whether the message is sent to an output or never displayed.

The `%s` notation is indeed inconvenient, but it prevents the passed arguments from always having to be converted into strings, which can positively affect performance in production when it comes to DEBUG messages (strings needed for debugging should only be generated if they actually end up in an output)

searx/result_types/image.py
Bnyro: But SearXNG doesn't use / display the subtype anywhere? So in fact, we could just remove the `subtype` field and replace all its uses with using `label`, and it would still all be the same, right?

searx/result_types/image.py
Bnyro: It's probably possible to do this without f-strings as well, i.e. we can still forward the string template and parameters to the underlying `log.error`.

But yeah, that's complicated if we can't change `log`

return42: > But yeah, that's complicated if we can't change `log`

Personally I think the python's builtin logging solution is complicated crap from last century.. there are much better solutions: [loguru](https://github.com/Delgan/loguru)

Bnyro: Agreed, that looks very nice

searx/result_types/image.py
return42: I think we should always specify the full MIME type in the result; this can be relevant on the client side when the client selects a suitable player for the media .. I added a mapping.

return42: I added a mapping from mime-type to human-readable-type and fixed the issues you reported in your review --> [diff](https://github.com/searxng/searxng/compare/7c8afd66424292243341467117f823cb1ee7892e..2ac9b12df026f2df3d0cfc97662bd0083bb9f8d9)

searx/result_types/image.py
```diff
@@ -7,14 +7,51 @@
...
+MimeSubType = t.Literal["png", "svg+xml", "jpeg", "bmp", "x-icon", "tiff"]
+
+MIMESUB: dict[MimeSubType, str] = {
+    "png": "PNG",
+    "svg+xml": "SVG",
+    "jpeg": "JPG",
+    "bmp": "BMP",
+    "x-icon": "ICO",
+    "tiff": "TIF",
+}
```
Bnyro, line 34: I hoped that [`mimetypes`](https://docs.python.org/3/library/mimetypes.html) has a way to get the human-readable format name of a mimetype - but apparently, it can't, so we have to hardcode them ourselves :/

searx/templates/simple/result_templates/images.html
Bnyro: Why did you mark this here as resolved? I'm fine with not implementing it, I just would like to know why :)

Bnyro (commented):
Please have a quick look at the suggestion in https://github.com/searxng/searxng/pull/6153#discussion_r3326716211, apart from that lgtm 👍 

(I haven't tested it yet though - as I'm currently too lazy to modify an engine to support multiple image formats)

searx/templates/simple/result_templates/images.html
return42: Sorry, mistake from my side .. I fixed it accordingly, see [diff](https://github.com/searxng/searxng/compare/cfbfc549e40053f5b0369ed11447193c887250be..15d423aaa5885d45fab6fd0a5c0fc26c34b02458)

searx/engines/selfhst.py
```diff
@@ -35,7 +35,7 @@ def response(resp):
         img_format = None
         for format_name in ('SVG', 'PNG', 'WebP'):
             if item[format_name] == 'Yes':
-                img_format = format_name.lower()
+                img_format = format_name
```
Bnyro: Why did you change this?

Before: `https://cdn.jsdelivr.net/gh/selfhst/icons/svg/nextcloud.svg`

After: `https://cdn.jsdelivr.net/gh/selfhst/icons/SVG/nextcloud.SVG` -> url is invalid

Bnyro (approved):
Tested, very cool! 👍 

Just the one nitpick in `selfhst.py`, otherwise ready to merge from my side

searx/engines/selfhst.py
return42: Good catch :+1: .. [fixed](https://github.com/searxng/searxng/compare/15d423aaa5885d45fab6fd0a5c0fc26c34b02458..32e88b21d5be5f2c12d1e541dd157d326219f06d)

Bnyro (approved):

PR #6174, opened 2026-05-30
fix(braveapi): correct freshness param name and guard against None injection

### LINKED ISSUE #6173: braveapi engine consistently returns HTTP 422 from Brave Search API through server pipeline
## Bug Description

The `braveapi` engine consistently receives HTTP 422 responses from the Brave Search API when making search requests through the full SearXNG server pipeline. Direct HTTP calls to the same Brave API endpoint with the same parameters succeed.

## Steps to Reproduce

1. Configure a `braveapi` engine with a valid API key in settings.yml
2. Make a search request to the SearXNG instance targeting the brave-api engine
3. Observe the engine always returns an HTTP 422 error

## Expected Behavior

The `braveapi` engine should successfully fetch and return search results from the Brave Search API, as it does when called via direct httpx.

## Actual Behavior

Every request through the SearXNG server pipeline returns `422 Unprocessable Entity`. The error is logged as:

```
searx.exceptions.SearxEngineAccessDeniedException: HTTP error 422 (suspended_time=180)
```

The engine is suspended after each failed attempt and never returns results.

## Investigation Results

The 422 only occurs when requests go through SearXNG's full request pipeline (engine processor → `_send_http_request` → `searx.network.get()`). The following all return HTTP 200 successfully:

1. Direct `httpx.get()` to the same Brave API URL with the same headers
2. `searx.network.get()` called with the same parameters from inside the running container (via `docker exec`)
3. `searx.network.Network.request()` called directly from inside the container

The failing path is specifically the combination of:
- Engine's `request()` → sets `params["url"]` and `params["headers"]`
- Processor's `_send_http_request()` → builds `request_args` from params
- `searx.network.get(url, **request_args)` → makes the HTTP call

This suggests a subtle interaction between how `_send_http_request` constructs the request arguments and how they're processed by the async network layer in a granian worker context.

A workaround that confirms the root cause is in the network pipeline: replacing the braveapi engine with one that calls `searx.network.get()` directly from the `request()` function (bypassing `_send_http_request`) resolves the issue.

## Environment

- SearXNG version: `searxng/searxng:latest` (2026.5.10-df1f24fb7)
- Python version: 3.14.4
- httpx version: 0.28.1
- Deployed via Docker Compose
- Brave Search API endpoint: `https://api.search.brave.com/res/v1/web/search`
- Engine configuration: `engine: braveapi` with valid `api_key`

---

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

magnus919:
## Summary

Fixes two bugs in the braveapi engine time_range parameter handling that can cause HTTP 422 errors from the Brave Search API.

## Root Cause

1. **Wrong parameter name**: The engine sends `time_range` but the Brave Search API expects `freshness` per the [API documentation](https://api-dashboard.search.brave.com/documentation/web-search/query). While Brave currently accepts both, the code should match the documented contract.

2. **None injection through urlencode()**: `time_range_map.get()` returns `None` when the time range value is not in the canonical map (e.g., the string `"all"` used by SearXNG to indicate "no time filter"). Python's `urlencode()` encodes `None` as the literal string `"None"`, producing `freshness=None` in the query string. The Brave API returns HTTP 422 ("Unprocessable Entity") for this invalid parameter value.

## Fix

```python
# Before
if params["time_range"]:
    search_args["time_range"] = time_range_map.get(params["time_range"])

# After
if params["time_range"]:
    mapped = time_range_map.get(params["time_range"])
    if mapped:
        search_args["freshness"] = mapped
```

The `if mapped:` guard prevents `None` from entering `search_args` when the time range value is not one of the four canonical values in `time_range_map`. This also corrects the parameter name to match the API contract.

Closes: #6173

---

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

### DIFF
--- a/searx/engines/braveapi.py
+++ b/searx/engines/braveapi.py
@@ -79,7 +79,9 @@ def request(query: str, params: "OnlineParams") -> None:
 
     # Apply time filter if specified
     if params["time_range"]:
-        search_args["time_range"] = time_range_map.get(params["time_range"])
+        mapped = time_range_map.get(params["time_range"])
+        if mapped:
+            search_args["freshness"] = mapped
 
     # Apply SafeSearch if enabled
     if params["safesearch"]:


return42:
You ignored the PR template, instead you copied a lot of AI nonsense .. and how can this PR close a issue that is already closed as not planed .. sorry, I will block you for a longer time.

PR #6176, opened 2026-05-30
Improve container security by running as non-root user

RinZ27:
### What does this PR do?

This change improves container security by running the SearXNG process as a non-privileged user. I added the USER instruction to the Dockerfile to ensure that, after initial setup in the entrypoint, the container operates under the 'searxng' user instead of root.

Build the container using the modified Dockerfile and verify the running user ID inside the container using 'id' or 'whoami'.

### Related issues

N/A

### Code of Conduct

AI tools used: Gemini CLI (orchestration, analysis, and code modification).


inetol (commented):
Not ready yet

PR #6185, opened 2026-05-31
[fix] no such table during engine init

### LINKED ISSUE #6181: Docker image fails with 'can't register engine processor'
### How To Reproduce?

Start SearXNG using docker-compose

### Expected behavior

The container starts and does not produce errors.

### Screenshots & Logs

```
❯ curl -fsSL \
    -O https://raw.githubusercontent.com/searxng/searxng/master/container/docker-compose.yml \
    -O https://raw.githubusercontent.com/searxng/searxng/master/container/.env.example
❯ mv .env.example .env

❯ docker compose up -d
[+] up 19/19
 ✔ Image docker.io/valkey/valkey:9-alpine Pulled                                                                                                                                                                   3.3s
 ✔ Image docker.io/searxng/searxng:latest Pulled                                                                                                                                                                   4.9s
 ✔ Network searxng_default                Created                                                                                                                                                                  0.0s
 ✔ Volume searxng_valkey-data             Created                                                                                                                                                                  0.0s
 ✔ Volume searxng_core-data               Created                                                                                                                                                                  0.0s
 ✔ Container searxng-valkey               Started                                                                                                                                                                  0.4s
 ✔ Container searxng-core                 Started                                                                                                                                                                  0.4s
❯ docker logs searxng-core
SearXNG 2026.5.30-bd863f16b
...
... INFORMATION
... "/etc/searxng/settings.yml" does not exist, creating from template...
...
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
[INFO] Starting granian (main PID: 1)
[INFO] Listening at: http://:::8080
[INFO] Spawning worker-1 with PID: 933
2026-05-30 16:28:09,697 ERROR:searx.engines: loading engine ahmia failed: set engine to inactive!
2026-05-30 16:28:09,774 ERROR:searx.engines: loading engine torch failed: set engine to inactive!
2026-05-30 16:28:09,798 WARNING:searx.botdetection.config: missing config file: /etc/searxng/limiter.toml
[INFO] Started worker-1
2026-05-30 16:28:09,805 ERROR:searx.searx.search.processor: Init method of engine radio browser failed due to an exception.
  File "/usr/local/searxng/searx/search/processors/abstract.py", line 155, in init_engine
    init_ok = self.engine.init(eng_setting)
  File "/usr/local/searxng/searx/engines/radio_browser.py", line 62, in init
    server_list()
    ~~~~~~~~~~~^^
  File "/usr/local/searxng/searx/engines/radio_browser.py", line 67, in server_list
    servers = CACHE.get("servers", [])
  File "/usr/local/searxng/searx/enginelib/__init__.py", line 175, in get
    return ENGINES_CACHE.get(key, default=default, ctx=self.table_name)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/cache.py", line 452, in get
    self.maintenance()
    ~~~~~~~~~~~~~~~~^^
  File "/usr/local/searxng/searx/cache.py", line 266, in maintenance
    if not force and int(time.time()) < self.next_maintenance_time:
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/cache.py", line 340, in next_maintenance_time
    return self.cfg.MAINTENANCE_PERIOD + self.properties.m_time("LAST_MAINTENANCE", int(time.time()))
                                         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/sqlitedb.py", line 452, in m_time
    res = self.DB.execute(self.SQL_M_TIME, (name,))
sqlite3.OperationalError: no such table: properties
2026-05-30 16:28:09,807 ERROR:searx.search.processors: can't register engine processor: radio browser (init failed)
```

### Version of SearXNG

SearXNG 2026.5.30-bd863f16b

Following [the official documentation](https://github.com/searxng/searxng).

### Code of Conduct

vojkovic:
### What does this PR do?

Fixes a race condition when the sqlite cache is made for the first time. Many engines can run init in parallel which triggers the cache maintenance which then reads the table before it's created:

```
sqlite3.OperationalError: no such table: properties
```

`make container`

Restart several times

Confirm no more race conditions

### Related issues

Closes: https://github.com/searxng/searxng/issues/6181

### Code of Conduct

Opus 4.8 to narrow down where the race is happening

### DIFF
--- a/searx/sqlitedb.py
+++ b/searx/sqlitedb.py
@@ -327,7 +327,6 @@ class SQLiteAppl(abc.ABC):
 
         if self._init_done:
             return False
-        self._init_done = True
 
         logger.debug("init DB: %s", self.db_url)
         self.properties.init(conn)
@@ -342,6 +341,7 @@ class SQLiteAppl(abc.ABC):
                 raise sqlite3.DatabaseError("Expected DB schema v%s, DB schema is v%s" % (self.DB_SCHEMA, ver))
             logger.debug("DB_SCHEMA = %s", ver)
 
+        self._init_done = True
         return True
 
     def create_schema(self, conn: sqlite3.Connection):
@@ -368,6 +368,9 @@ class SQLiteProperties(SQLiteAppl):
          PRIMARY KEY (name))
 
     """
+
+    _locks: dict[str, threading.Lock] = {}
+    _locks_lock = threading.Lock()
 
     SQLITE_JOURNAL_MODE: str = "WAL"
 
@@ -406,11 +409,16 @@ CREATE TABLE IF NOT EXISTS properties (
 
         if self._init_done:
             return False
-        self._init_done = True
-        logger.debug("init properties of DB: %s", self.db_url)
-        res = conn.execute(self.SQL_TABLE_EXISTS)
-        if res.fetchone() is None:  # DB schema needs to be be created
-            self.create_schema(conn)
+        with self._locks_lock:
+            db_lock = self._locks.setdefault(self.db_url, threading.Lock())
+        with db_lock:
+            if self._init_done:
+                return False
+            logger.debug("init properties of DB: %s", self.db_url)
+            res = conn.execute(self.SQL_TABLE_EXISTS)
+            if res.fetchone() is None:  # DB schema needs to be be created
+                self.create_schema(conn)
+            self._init_done = True
         return True
 
     def __call__(self, name: str, default: t.Any = None) -> t.Any:


return42:
I am convinced that this is a problem with the transactions: normally, the DDL statements should be completed as part of the initialization.

https://github.com/searxng/searxng/blob/7159b8aed3c37cbc849782abf737ac83253b9fc8/searx/sqlitedb.py#L412-L413

The fact that the race condition has been occurring (`no such table: properties`) since we started loading a very large dataset (`WIKIDATA_PROPERTIES`) into the database, indicates to me that the DB schema has not (yet) been committed ..

----

Important to note: the transaction behavior of the SQLite engine and the Python `sqlite3` module differs. To put it in highly simplified terms:

- SQLite engine: DDL statements are unbound and are always written to the database immediately.
- `sqlite3` module: As long as a DDL statement (e.g., `CREATE TABLE ...`) has not been committed, the table is available only within the current database connection.

----

To implement transaction scopes in Python's sqlite3 module, `with` contexts can be used:

https://github.com/searxng/searxng/blob/7159b8aed3c37cbc849782abf737ac83253b9fc8/searx/sqlitedb.py#L458-L460

.. so much for the theory :-)  ... but as the current issue demonstrates, DDL statements within a `with` context are not committed (which is exactly the opposite of what an SQL engine would do).

Up to this point, this has been a highly simplified explanation; there are other factors that can influence the behavior—such as the Python version, the compile options used to build the SQLite engine into the Python interpreter, and, not least, the transaction options used to establish the database connection.

Related:

- https://docs.python.org/3/library/sqlite3.html#transaction-control

-------

Long story told short: I implemented a DRAFT

- https://github.com/searxng/searxng/pull/6187

Unfortunately, I'm not familiar with Docker and don't know how to reproduce the error. Therefore, I'd like to ask if you could test my draft; further iterations might be necessary if the `conn.commit()` call turns out not to be sufficient.

vojkovic:
@return42 I am pretty sure the race is being caused by this:

1. A enters init and _init_done = True and calls create_schema

2. B enters init and sees _init_done is True and then returns early and it never calls create_schema

3. B accesses the db which triggers the maintenance which queries properties table

4. At this point A is still inside create_schema. The table creation is not executed yet or let alone committed.

The commit doesn't get a chance to matter for B, the window is between steps one and four, not between the commit and the read so I don't think your changes will fix it but are probably worth keeping anyways. :-)

---

> Unfortunately, I'm not familiar with Docker and don't know how to reproduce the error. 

I also am having troubles reproducing it reliably. I'ld actually expect the DDL on a seperate connection idea to be slightly worse in isolation because of the extra work on the init which increases the window for B to slip in and cause the race.

return42:
The description is correct, but...

this hasn't been a problem until now; so, the question for me is: why didn't we encounter this issue sooner?

@inetol's bisect analysis revealed that the race condition you described only manifests now that we are loading the large dataset (700KB) from the Wikidata engine's initialization into the cache .. although the race between A and B has actually existed all along.

The only plausible explanation to me is that the DB schema (CREATE TABLE) and the large dataset (SQL INSERT) are contained within one and the same transaction — and that is poorly modeled.

My hope was that by isolating the schema creation, I could restore the old runtime behavior, but as @inetol's [second test shows](https://github.com/searxng/searxng/pull/6187#pullrequestreview-4401124083), that *__isn't enough__*, and you're right in your assumption ...

>  I'ld actually expect the DDL on a seperate connection idea to be slightly worse in isolation because of the extra work on the init which increases the window for B to slip in and cause the race.

*__isn't enough__*: the revised model (my PR) is necessary from the perspective of competing processes .. from the perspective of competing threads, a lock like the one in your PR is required.

Give me a little more time to think about and try out some other options ... for example, it would make sense to create the base schema (table `properties`) when the application is initialized, rather than waiting until the engines are initialized.

I need to set up a test scenario somehow... but I'm having a bit of trouble with that right now because I don't have the time to spare at the moment...

return42:
> Give me a little more time to think about and try out some other options ...

Sorry for the delay. I'm currently very busy and can't look into this topic in more detail, to have a hot-fix, I'll merge this pull request for now. 

> ... for example, it would make sense to create the base schema (table `properties`) when the application is initialized, rather than waiting until the engines are initialized.

Later, when I have more time, we'll need to address the modeling of transaction spacees again.

- https://github.com/searxng/searxng/pull/6187

return42 (approved):

vojkovic:
Yes, I agree your PR will also be necessary.

Sorry - also meant to look at why didn't we encounter this issue sooner but have been stuck doing other things. Hopefully I can revisit this issue in like a week from now when I am less busy.

vojkovic:
Argh is there a bug in my code? It is working fine locally but as soon as I deploy I'm getting 5xx errors. Need to debug....

return42:
> Argh is there a bug in my code? 

Can't believe .. @inetol tested this code in https://github.com/searxng/searxng/issues/6181#issuecomment-4586267123

update: haven't seen there is already a bug report about ..

- https://github.com/searxng/searxng/issues/6214

martinjgrunwald:
Sorry to necro this, but for me it still doesn't work. I still get the same docker output as in the first post. After disabling the radio browser I still get this output:
```
SearXNG 2026.6.8-f3fab143b
Updating certificates in /etc/ssl/certs...
0 added, 0 removed; done.
Running hooks in /etc/ca-certificates/update.d...
done.
[INFO] Starting granian (main PID: 1)
[INFO] Listening at: http://:::8080
[INFO] Spawning worker-1 with PID: 924
2026-06-08 16:59:26,013 ERROR:searx.engines: loading engine ahmia failed: set engine to inactive!
2026-06-08 16:59:26,135 ERROR:searx.engines: loading engine torch failed: set engine to inactive!
2026-06-08 16:59:26,188 WARNING:searx.botdetection.config: missing config file: /etc/searxng/limiter.toml
[INFO] Started worker-1
```
And my web page looks distorted, the info and settings logos are huge

I am running Searxng in Docker on an Oracle Free Tier VM with an ARM64 CPU

vojkovic:
> Sorry to necro this, but for me it still doesn't work.

Yes, the changes here were reverted in a later commit and are still being worked on.

The logs you sent show searxng is starting correctly. Your symptoms sounds like the CSS isn't loading, does this happen connecting directly into the container or via a reverse proxy like nginx or apache?

Try clearing your browser cache first.

ARM image works fine locally.

return42:
> Yes, the changes here were reverted in a later commit and are still being worked on.

Can someone here please test this PR:

- https://github.com/searxng/searxng/pull/6187

PR #6188, opened 2026-05-31
[fix] duckduckgo: image requests get blocked

Bnyro:
### What does this PR do?
- for some reason all i.js requests sending the accept language header get blocked, so this removes it

Related:

- https://github.com/searxng/searxng/pull/5758

- !ddi test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/search/processors/online.py
+++ b/searx/search/processors/online.py
@@ -152,7 +152,6 @@ class OnlineProcessor(EngineProcessor):
         # add Accept-Language header
         # https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language
 
-        headers["Accept-Language"] = "en,en-US;q=0.7,en;q=0.3"
         if self.engine.send_accept_language_header and search_query.locale:
             _l = search_query.locale.language
             _t = search_query.locale.territory or _l
--- a/searx/engines/duckduckgo_extra.py
+++ b/searx/engines/duckduckgo_extra.py
@@ -41,7 +41,9 @@ safesearch_cookies = {0: "-2", 1: None, 2: "1"}
 safesearch_args = {0: "1", 1: None, 2: "1"}
 
 search_path_map = {"images": "i", "videos": "v", "news": "news"}
+
 _HTTP_User_Agent: str = gen_useragent()
+send_accept_language_header = False
 
 
 def init(engine_settings: dict[str, t.Any]):


searx/search/processors/online.py
```diff
@@ -152,7 +152,6 @@ def get_params(self, search_query: "SearchQuery", engine_category: str) -> Onlin
         # add Accept-Language header
         # https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language
 
-        headers["Accept-Language"] = "en,en-US;q=0.7,en;q=0.3"
```
return42, line 155: We had previously added this in 9be839a12027 -> I am not sure if it is a good idea to remove the default `Accept-Language` header here for *all* engines .. a web browser always sends an `Accept-Language` header; see my example for `i.js` below.

> for some reason all `i.js` requests sending the accept language header get blocked, so this removes it

I think it would be better if we removed the header only for the `i.js` requests within the DDG image search... though, as the request example below suggests, that might not actually be the real issue after all... we might need to investigate this in more detail.

```HTTP
GET /i.js?o=json&q=hallo&l=de-de&vqd=4-201794583779345072107815084038705298825&p=1&ct=DE&jsa=807&jsa_hash=12fdbff20c12b26780d147db67f9fe96&dp=XKkHlSugGwBH4DIgEE7nqvSVuld4txNtMFArB_lns6ARGtTo5lTAxi2wc-yUh6r9J0jMmGDICKe75mMlsURlEGUulRuopF_9q9vxypUyjmWNlcpet2OHz3_59YH6b0QYTb2FfbdzuDz-TKABVqc4dA.EcQE77HVo2w5zD4-G8FUDQ&bpia=1&a=ftsa HTTP/2
Host: duckduckgo.com
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0
Accept: */*
Accept-Language: de,en-US;q=0.7,en;q=0.3
Accept-Encoding: gzip, deflate, br, zstd
Referer: https://duckduckgo.com/
DNT: 1
Sec-GPC: 1
Connection: keep-alive
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: same-origin
Priority: u=4
Pragma: no-cache
Cache-Control: no-cache
TE: trailers
```

Bnyro: The thing is that even if we set `send_accept_language_header` to `False`, it sends a language header. I agree that it should probably be sent by default, but it's very inconvenient if you explicitly disable it and it still gets sent anyways...

My guess that the header just looked a bit different than what normal Firefox browsers do - hence the ratelimit. But idk.

Bnyro: Can confirm this also fixed ddg images on my server - so this definitely works as a temporary workaround.

But yeah, still not that happy with it.

PR #6192, opened 2026-06-01
[feat] engines: add duckduckgo web engine as alternative to html.duckduckgo.com

Bnyro:
## Changes

html.duckduckgo.com captchas all my IPs very fast. I figured out that using duckduckgo.com works even if html.duckduckgo.com is captcha-ed, hence adding support for duckduckgo.com's general web search here.

This implementation fetches the link to the first API page (i.e. ``links.duckduckgo.com/d.js?...``) from duckduckgo.com and uses the ``n`` parameter of the API to fetch all subsequent pages.

This also means that it's not possible to immediately search for the first page - the first and the second page would need to be loaded first.

The reason why we can't just normally use the `vqd` value is that the API URLs require an additional parameter `dp` which seems generated at server-side, so we can't build it ourselves and must scrape it from the HTML pages.

### Notes
- This is a very basic implementation - localization, safesearch and timerange don't work yet. It should be possibly to just manipulate the query parameters of the search URL (e.g. just replace `&l=us-en` with some other language, but I haven't had time to look into this yet.

### Related issues
- https://github.com/searxng/searxng/issues/4824

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/duckduckgo_web.py
+++ b/searx/engines/duckduckgo_web.py
@@ -1,0 +1,154 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""DuckDuckGo Web (general)
+
+This implementation fetches the link to the first API page
+(i.e. ``links.duckduckgo.com/d.js?...``) from duckduckgo.com and uses the ``n``
+parameter of the API to fetch all subsequent pages.
+
+This also means that it's not possible to immediately search for the third
+page - the first and the second page would need to be loaded first.
+
+The reason why we can't just normally use the `vqd` value is that the API URLs
+require an additional parameter `dp` which seems generated at server-side, so we
+can't build it ourselves and must scrape it from the HTML pages.
+"""
+
+import typing as t
+
+from urllib.parse import quote_plus
+from lxml import html
+
+from searx.utils import html_to_text, gen_useragent, extract_text, eval_xpath
+from searx.result_types import EngineResults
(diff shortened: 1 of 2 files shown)


searx/engines/duckduckgo_web.py
```diff
@@ -0,0 +1,149 @@
...
+
+    dom = html.fromstring(resp.text)
+    first_page_link = extract_text(eval_xpath(dom, "//link[@id='deep_preload_link']/@href"))
+
+    if not first_page_link:
+        logger.error("vqd: failed to load first page JS url from ddg response (return empty string)")
+        return ""
+
+    logger.debug("got link to first page from duckduckgo.com request: '%s'", first_page_link)
+    CACHE.set(cache_key, first_page_link)
```
return42: The `cache_key` is build up from `query` .. the default expire time is:

https://github.com/searxng/searxng/blob/f3fab143be3069bbcdcec9169bcf6ee030437a61/searx/cache.py#L48

.. do we really need the `first_page_link` hold for 7 days?  I'm afraid that will really bloat the database... would it perhaps suffice to keep the link in the cache for 2 or 4 hours?

----

I can fix it in my current review.

Bnyro: Agreed, probably a few hours is enough. I just didn't know how long these are valid, so I didn't specify a duration.

return42 (approved):
LGTM

PR #6194, opened 2026-06-02
[fix] tracker_url_remover: cache rule set in process, not per result URL

dmarzzz:
## What

`TrackerPatternsDB.clean_url()` is called once per result URL, and it called `self.rules()` on every call. `rules()` re-runs `init()` and `ExpireCacheSQLite.pairs()`, so for **every single result URL** the Tracker-URL-Remover plugin:

- ran the cache maintenance check (`properties.m_time`),
- re-read the list of cache tables (`properties` scan, `cache.table_names`),
- and re-read + unpickled the **entire** ClearURL rule set from SQLite (`cache.pairs`).

The cost scales with result count, which scales with the number of engines queried.

This PR materializes the rule set into an in-process list once (`_cached_rules`) and reuses it across `clean_url()` calls, invalidating the copy in `load()` / `add()` whenever the underlying cache is written. The rule set only changes when the cache is (re)written, so the in-process copy is always consistent with it.

## Why

While profiling a sustained `N=32` engine fan-out with py-spy, ~82% of the request worker's CPU was attributed to these per-URL SQLite round-trips, all under `searx.plugins.tracker_url_remover`:

| self% | function |
|---:|---|
| 22.0 | `cache.py:table_names` |
| 20.7 | `sqlitedb.py:m_time` |
| 18.6 | `cache.py:pairs` |
| 16.2 | `sqlitedb.py:__call__` |
| 3.3 | `tracker_patterns.py:clean_url` |

The rule matching itself was a rounding error; the SQLite metadata reads + full-table re-read dominated. Most result URLs carry no tracking query args, so the rule loop breaks on the first rule and the per-URL cost is essentially just those repeated SQLite reads.

## Result

Micro-bench, 630-rule set, 200 arg-free result URLs (old code measured directly via `git stash`, not simulated):

- before: **24.6 µs/url**
- after: **2.5 µs/url** (~10×)

A regression test asserts the rule set is read from SQLite **once** across many `clean_url()` calls (previously: once per URL).

## Scope / follow-up

This targets the SQLite per-URL tax only. For URLs that *do* carry query args, the cost shifts to regex-matching the full rule set per URL (the `re` module's compiled-pattern cache thrashes once the rule count exceeds its limit). Precompiling the rule patterns is a sensible follow-up but is intentionally out of scope here to keep the change minimal and easy to review.

## Tests

- New `tests/unit/test_tracker_patterns.py` (4 tests): tracker stripping, untracked-URL passthrough, the read-once regression, and cache-write invalidation.
- Existing `test_plugins` / `test_results` still pass.
- `black` (line-length 120) clean, `pylint` 10.00/10.

### DIFF
--- a/searx/data/tracker_patterns.py
+++ b/searx/data/tracker_patterns.py
@@ -43,6 +43,11 @@ class TrackerPatternsDB:
 
     def __init__(self):
         self.cache = get_cache()
+        # In-process copy of the materialized rule set.  ``None`` means "not yet
+        # built"; it is (re)built lazily from the SQLite cache by
+        # :py:obj:`self._cached_rules` and invalidated whenever the underlying
+        # cache is written (:py:obj:`self.load` / :py:obj:`self.add`).
+        self._rules: "list[RuleType] | None" = None
 
     def init(self):
         if self.cache.properties("tracker_patterns loaded") != "OK":
@@ -66,6 +71,8 @@ class TrackerPatternsDB:
             rows.append((key, value, None))
 
         self.cache.setmany(rows, ctx=self.ctx_name)
+        # the rule set in the cache changed -> drop the in-process copy
+        self._rules = None
 
     def add(self, rule: RuleType):
         key = rule[self.Fields.url_regexp]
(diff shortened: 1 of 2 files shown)


Bnyro:
You clearly didn't read the AI policy.

Issue #338, opened 2021-09-22
Selenium tests: check console output.


dalf:
See #337

The Selenium tests should check the console output to make sure there is no javascript errors.

FaresIbrahim32:
Hello Im interested in solving this issue

PR #6197, opened 2026-06-04
[fix] issue #338 selenium robot tests: fail on browser-side JavaScript errors 

FaresIbrahim32:
### What does this PR do?

Makes the Selenium/Splinter robot tests fail when the browser reports
JavaScript errors, so client-side regressions are caught automatically in CI.

Previously the tests only asserted that certain text/elements were present on
the page; a broken script that still rendered the expected markup would pass
unnoticed. This PR subscribes to the browser console and fails the run if any
JavaScript errors are detected while a test executes.

Changes (all in `tests/robot/__main__.py`):

- Start the Firefox browser with the `webSocketUrl` capability, which enables
  the **WebDriver BiDi** protocol. This is required because Firefox/geckodriver
  does not support the classic `driver.get_log('browser')` API (that is
  Chrome/CDP only). BiDi is the modern, cross-browser, W3C-standard way to read
  the console and is available in the pinned `selenium==4.44.0`.
- Add `_collect_browser_errors(browser)`, which registers two BiDi log handlers
  via `browser.driver.script`:
  - `add_javascript_error_handler` — captures uncaught JavaScript exceptions.
  - `add_console_message_handler` — captures `error`-level console messages.
- Rework `run_robot_tests()` so each test still runs in its own browser, then
  records any errors that were collected. After all tests have run, if any
  errors were captured it raises an `AssertionError` that lists them per test.
  `main()` already converts that into a non-zero exit code, so CI fails.
- Add a short settle (`sleep(1)`) after each test to allow the asynchronous
  BiDi log events to be delivered before the errors are inspected.

```sh
make test.robot
```

This downloads geckodriver (via `gecko.driver`) and runs
`python -m tests.robot`, which starts the SearXNG webapp and drives a headless
Firefox. Requires Firefox to be installed locally.

To confirm the new check actually fails on JavaScript errors, temporarily
introduce a JS error in a template/script served by the test instance (for
example, call an undefined function on page load) and re-run `make test.robot`:
the run should now fail and print the offending console messages grouped by
test name.

Notes / edge cases:

- The check fails on both uncaught JS exceptions **and** `error`-level console
  messages. The latter is slightly broader (e.g. a failed resource load logged
  at error level would fail a test); this can be narrowed to JS exceptions only
  if preferred.
- Requires a geckodriver/Firefox new enough to support WebDriver BiDi.

### Related issues

Closes hopefully : #338

### Code of Conduct

  AI tools used:

  - **Claude Code (Anthropic)** — used to research the existing robot test
    setup, identify that Firefox/geckodriver requires WebDriver BiDi (rather
    than the Chrome-only `get_log` API) for console access,
    in `tests/robot/__main__.py`, and verify it against the repo's `black` and
    `pylint` configuration. All changes were reviewed before submission.

### DIFF
--- a/tests/robot/__main__.py
+++ b/tests/robot/__main__.py
@@ -9,6 +9,8 @@ import traceback
 import pathlib
 import shutil
 
+from time import sleep
+
 from splinter import Browser
 
 import tests as searx_tests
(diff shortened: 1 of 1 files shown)


return42:
> * [x]  **I hereby confirm that this PR conforms with the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst).**

Sorry, you didn't ...

*Every discussion, issue, and pull request is read and reviewed by humans. It is a boundary point at which people interact with each other and the work done. It is rude and disrespectful to approach this boundary with low-effort, unqualified work, since it puts the burden of validation on the maintainer.*

*It takes a lot of maintainer time and energy to review AI-generated contributions! Sending the output of an LLM to open source project maintainers extracts work from them in the form of design and code review, so we call this kind of contribution an "extractive contribution".*

*The golden rule is that a contribution should be worth more to the project than the time it takes to review it, which is usually not the case if large parts of your PR were written by LLMs.*

PR #6207, opened 2026-06-05
chore: improve searxng maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/__init__.py, tests/robot/__main__.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


Bnyro:
This is the most useless change I've seen in a while?

PR #6209, opened 2026-06-06
[fix] use fromisoformat in NVD engine

### LINKED ISSUE #6098: Use `datetime.fromisoformat` instead of `datetime.strptime`
### Problem
- We currently use something like `date = datetime.strptime(date_string, "%Y-%m-%dT%H:%M:%S.%f")`, although `datetime.fromisoformat` can handle that and is more robust

### Describe the solution you'd like
- Use `datetime.fromisoformat` instead.

### Code of Conduct

Anthomos:
## Summary
- replace the NVD engine's manual `strptime` call with `datetime.fromisoformat`
- add a focused regression test for published timestamps with and without fractional seconds

## Testing
- `./local/py3/bin/python -m nose2 -v tests.unit.test_engine_nvd`
- `./local/py3/bin/python -m nose2 -v tests.unit.test_engine_tineye`

Closes #6098.

AI disclosure: I used Codex as a coding aid for this patch and test setup, then reviewed the final change and test scope before opening the PR.

### DIFF
--- a/searx/engines/nvd.py
+++ b/searx/engines/nvd.py
@@ -44,7 +44,7 @@ def response(resp) -> EngineResults:
 
         cve_id = item["cve"]["id"]
         description = item["cve"]["descriptions"][0]["value"]
-        date = datetime.strptime(item["cve"]["published"], "%Y-%m-%dT%H:%M:%S.%f")
+        date = datetime.fromisoformat(item["cve"]["published"])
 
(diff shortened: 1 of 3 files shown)


Bnyro:
> Closes https://github.com/searxng/searxng/issues/6098.
>
It doesn't.

>  I used Codex as a coding aid for this patch and test setup, then reviewed the final change and test scope before opening the PR.
>
These unit tests are just trash, doing this per engine is unmaintainable.

PR #6212, opened 2026-06-06
[feat] engines: add tiger.ch engine

Bnyro:
### What does this PR do?
- add support for https://tiger.ch (general, news)
- it is disabled and inactive by default because it's just a metasearch engine like SearXNG is, so it's mostly useful for bypassing rate-limits on other engines
- (it has its own German index, but it's not that great)
- in theory it supports different locales, but I was too lazy to implement that (I only need German and English results anyways, which are returned by default...)

- !tig test
- !tign test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/tiger.py
+++ b/searx/engines/tiger.py
@@ -1,0 +1,167 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Tiger_ is a Swiss meta search engine.
+
+.. _Tiger: https://tiger.ch
+"""
+
+from json import loads
+import random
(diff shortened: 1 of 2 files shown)


return42 (approved):
> in theory it supports different locales, but I was too lazy to implement that 

The News seems DE only, but for the Web search, the `send_accept_language_header` (which is *on* by default) does a good job .. at least in my tests (`!tig bmw :fr`)

Bnyro:
(fixed some typos in the description of the math challenge)

Issue #5976, opened 2026-04-13
karmasearch engine - returns HTTP error 403


return42:
karmasearch engine always returns HTTP 403 .. the fix from https://github.com/searxng/searxng/pull/5953 does no longer work (@Aadniz).

### How To Reproduce?

    !karmasearch John Doe

### Version of SearXNG

tested from my server's IP and in a `make run` instance.

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [x] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.

Aadniz:
Quite the cat and mouse game on this engine.

It seems a bit flakey right now atleast. I tested connecting (normally) to karmasearch from 2 different ip addresses:

<img width="2560" height="1305" alt="Image" src="https://github.com/user-attachments/assets/f6aed372-1403-47c4-b87f-ffc6c75fa37e" />

Usually my go-to method to try get around CAPTCHA is to do the request in firefox first, copy the request as curl, paste the command into the terminal, and slowly remove headers (and cookies) until you can pin-point what makes a difference.

return42:
Same on my side (IPs) ..

<img width="623" height="377" alt="Image" src="https://github.com/user-attachments/assets/478659f1-012d-48a5-ad8d-b88a989a552b" />

I also tried from different tor channels .. its always the same .. seems karmasearch.org is more or less not usable anymore ("Host Error").  

<img width="679" height="376" alt="Image" src="https://github.com/user-attachments/assets/c1e8aa04-3626-4232-a0bb-0b78574fcc3d" />

Jesterium:
**EDIT**

I found the proper syntax here: https://github.com/searxng/searxng/blob/master/searx/settings.yml


- https://github.com/searxng/searxng/pull/6082

---

How do you disable this engine entirely?

Issue #6136, opened 2026-05-25
Artic images protected behind Cloudflare


Bnyro:
### How To Reproduce?
- `!artic test`
- produces image links such as: https://www.artic.edu/iiif/2//92da7102-0562-f4da-55f9-1d29828b20d7/full/843,/0/default.jpg
- click the link and get cloudflared

### Expected behavior
- see any images, but you only get placeholders, i.e. all images fail to load

### Additional context
- We can fetch a base64 thumbnail by adding the `thumbnail` field to the request, i.e. `https://api.artic.edu/api/v1/artworks/search?q=hello&fields=id,title,artist_display,medium_display,image_id,date_display,dimensions,artist_titles,thumbnail`. The quality is so bad that I'd consider removing the engine instead

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [X] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.

inetol:
I'm not surprised, the `image_proxy` and `favicon_proxy` features can be exploited to download content massively, and the public instances are a common target for such abuse.

Guess which site gets the majority of traffic

<img width="300" height="131" alt="Image" src="https://github.com/user-attachments/assets/ea1b1dd8-fba3-4c51-9f48-6e63d4138ede" />

return42:
> features can be exploited to download content massively, and the public instances are a common target for such abuse.

@inetol are you sure? How could these be misused?

Image-Proxy has a HMAC in form field `h`

https://github.com/searxng/searxng/blob/0037d43d872f44033d7ae3677651362ca19462b3/searx/webapp.py#L318

https://github.com/searxng/searxng/blob/0037d43d872f44033d7ae3677651362ca19462b3/searx/webapp.py#L997-L998

And favicon endpoint .. [online doc](https://docs.searxng.org/src/searx.favicons.html#searx.favicons.proxy.favicon_proxy) ..

https://github.com/searxng/searxng/blob/0037d43d872f44033d7ae3677651362ca19462b3/searx/webapp.py#L986

https://github.com/searxng/searxng/blob/0037d43d872f44033d7ae3677651362ca19462b3/searx/favicons/proxy.py#L127-L139

inetol:
Since I'm not M€ta I can't see the details about the requests. Many of them returned a 200 code and didn't come from residential IPs. Since covering these endpoints with Anubis, I haven't seen any more bandwidth spikes 🤷

[...]

PR #6213, opened 2026-06-06
[del] karmasearch: remove engine (cloudflared)

Bnyro:
The engine is using very aggressive Cloudflare blocking for
a while now, no matter if using a normal browser like Firefox
or not.

Related:
- https://github.com/searxng/searxng/issues/5976

AI Policy: No AI used.


return42:
It's always sad when we lose an engine because it's blocked, but maybe one day we'll find a way to bypass the blocking. Therefore, we shouldn't forget to mark the pull requests and issues accordingly.

- [`engine removal`](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A"engine+removal")
- [`cloudflared`](https://github.com/search?q=repo%3Asearxng%2Fsearxng+label%3A"cloudflared")

.. if you can think of any other PRs or issues that these labels fit .. (?)

Bnyro:
Only https://github.com/searxng/searxng/issues/6136 comes into my mind, but it's not removed yet.

Bnyro:
Looks like we're responsible for heexy.org adding Cloudflare too :/

PR #6217, opened 2026-06-07
[feat] engines: add seek-ninja general engine

Bnyro:
### What does this PR do?
- add support for https://seek.ninja (general)
- it's very slow because the engine uses Server-side events, that incrementally send data in their HTTP response (https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/)
- i.e. we wait for the end of the response (7+ seconds), even though the results data arrives within a few seconds -> it's very slow, because SearXNG wants to get the full response body before it calls the `response(resp)` method
- We could use https://github.com/florimondmanca/httpx-sse, but I'm not sure how to integrate this into SearXNG and if it's worth it

- !sen test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/seekninja.py
+++ b/searx/engines/seekninja.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Seek ninja (general)"""
+
+from json import loads
+from hashlib import sha256
+from urllib.parse import urlencode, quote_plus
+
+import typing as t
+
+from searx.extended_types import SXNG_Response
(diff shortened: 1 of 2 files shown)


return42 (approved):
> We could use httpx-sse [2], but I'm not sure how to integrate this into SearXNG and if it's worth it

I wasn't familiar with that protocol yet .. oh, if only I had more time—then I’d love to take a closer look at that .. but unfortunately, the day only has 24 hours .. and the next  PR is waiting for a review :-)

PR #6222, opened 2026-06-08
[feat] plugins: color picker plugin

### LINKED ISSUE #5704: Feature Request: Color Picker
When users search for terms like "color picker", "rgb picker", or "colorpicker", they are usually looking for a quick way to select or inspect a color, not a list of websites. Currently, they need to open external pages, which is slower and unnecessary for such a simple task.

Detect common color-picker-related keywords and display a simple inline color picker directly in the results page.
This could be used maybe with  `<input type="color">`, showing HEX/RGB values for quick copy.

bryce-hoehn:
### What does this PR do?

Adds a color picker tool as a quick result. Matches terms like "color", "color picker', etc.

image

Publishing as a draft for now to get more feedback. Currently the color picker uses the iro.js library, which is imported via the CDN in the jinja template. My thought process being that as an optional plugin, its probably not worth installing additional client dependencies. On the other hand, as a privacy tool, I'm not sure how people feel about relying on a third party CDN.

Was also hoping to get some feedback on the options presented here. DuckDuckGo and Google color pickers for reference:

image

image

Both DuckDuckGo and Google present the color values in a more human readable form, but personally whenever I'm using a color picker I'm just trying to copy and paste the values.

Future enhancements:
- I was initially planning to add an inline copy to clipboard button like Google's, but it seems to be a PITA with vanilla JS/CSS. Could at least add some style to these buttons though.
- I think it'd be cool to incorporate a palette generator like DuckDuckGo's at some point.

Download this branch, run it as normal, and make a query for "color".

### Related issues

Closes: [#5704](https://github.com/searxng/searxng/issues/5704)

### Code of Conduct

- [X ] **I hereby confirm that this PR conforms with the [AI Policy].**

I have Zoo Code installed and connected to GLM 5.1. AI was used strictly for helping me navigate the codebase. At no point was it given write access to the repo.


return42:
> Publishing as a draft for now to get more feedback. 

Thanks a lot :-)

> On the other hand, as a privacy tool, I'm not sure how people feel about relying on a third party CDN.

Yep, that’s definitely a problem .. but as a code maintainer, I see others ..

> Currently the color picker uses the iro.js library, which is imported via the CDN in the jinja template.

Currently, the plugins are designed solely to provide result items; if a plugin requires its own specific result type, it effectively ceases to be a plugin.

Strictly speaking, `class ColorPicker(BaseAnswer, kw_only=True):` is not an answer type; this apparent type exists here merely as a vehicle to deliver JS code to the client .. or to load additional JS code.

Which brings me to the main point of my criticism: to me, a color picker is a purely client-side application; it doesn't actually require any server-side code.

.. but lets hear what other say ..

inetol:
The problem is that writing HTML in plain JavaScript is pain: https://github.com/searxng/searxng/blob/master/client/simple/src/js/util/appendAnswerElement.ts

What you could do, is move the script so that it loads with the client side instead.

bryce-hoehn:
I agree that as a client side feature it ideally shouldn't touch the server code.

However, as far as I can tell all of the frontend HTML is currently rendered server-side via Jinja templates. 

Long term, are there any plans or discussion about decoupling the frontend from Flask? Decoupling the two would make it much easier to customize the frontend and take the rendering load off the server.

I'm thinking the HTML templates could be converted to [Svelte](https://svelte.dev/docs/svelte/getting-started). This would keep it relatively lightweight, and Svelte's syntax is pretty similar to Jinja's. It could then communicate with the backend via an api instead of direct server side rendering.

I know I'm kind of coming in blind and proposing a significant refactor, but if this is something the community would be interested in I'd be happy to work on it.

also if you got two notifications it's because the github mobile app was being buggy.

inetol:
> Long term, are there any plans or discussion about decoupling the frontend from Flask? Decoupling the two would make it much easier to customize the frontend and take the rendering load off the server.

Yes, we are working on this now

bryce-hoehn:
Sweet, in that case it'd probably be best to hold off on client side plugins until then.

KraXen72:
I think Iro.js is no longer maintained/on hiatus, plus it doesn't support the new color spaces like oklch/oklab/cielab, it's srgb only. I know i ran into this when making my color picker chrome extension, and decided to build something custom, so i'm not sure what would be a good replacement.

return42:
> I think Iro.js is no longer maintained

Yes, seems dead .. beside this, we named two more reasons to close this PR / I will close PR now .. nevertheless; thanks a lot :+1:

PR #6223, opened 2026-06-08
[fix] yep: send Sec-Fetch headers to bypass "access denied"

Bnyro:
### What does this PR do?
- avoids yep's botblocking by sending Sec-Fetch-* headers (as the browser does)

- !yep test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/yep.py
+++ b/searx/engines/yep.py
@@ -16,17 +16,18 @@ if t.TYPE_CHECKING:
     from searx.search.processors import OnlineParams
 
 about = {
-    'website': 'https://yep.com/',
-    'official_api_documentation': 'https://docs.developer.yelp.com',
-    'use_official_api': False,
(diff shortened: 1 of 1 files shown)


return42 (approved):
On my server, Yep works fine even without the Sec-Fetch headers .. though perhaps just because the engine is rarely used. Regardless, just the code refactoring and `EngineResults` make this PR worth to merge :-)

Bnyro:
> On my server, Yep works fine even without the Sec-Fetch headers .. though perhaps just because the engine is rarely used. Regardless, just the code refactoring and `EngineResults` make this PR worth to merge :-)
>
For me, it's hit or miss - About 50% of requests work without Sec-Fetch headers (from server and local IP). With Sec-Fetch, all requests work, so it's definitely more reliable with them.

PR #6227, opened 2026-06-09
[feat] limiter: add a secret bypass key setting and header

chris-paganon:
### What does this PR do?

Add a new limiter_bypass_key setting and corresponding SEARXNG_LIMITER_BYPASS_KEY environment variable. 

If the `X-SearXNG-Limiter-Bypass` header is sent with a matching secret, the limiter is fully bypassed. It's especially useful for the owner of a public instance to be able to make their own API requests unrestricted. 

- Set `server.limiter_bypass_key` to `my-secret-bypass-key`
- Send an API request like `curl -H 'X-SearXNG-Limiter-Bypass: my-secret-bypass-key' http://localhost:8080/search?q=searxng`: the limiter should be skipped.
- Retry without header or with wrong key: you should hit 429 Too Many Requests

### Code of Conduct

- [x] I hereby confirm that I haven't used any other tools than the ones I mention below.

I have used AI tools for working on the changes in this PR: Codex with GPT 5.5. Asked some questions to Cursor's composer 2.5 here and there to help me better understand the codebase.

Most of the code is written with AI, I understand this doesn't really match the "AI should never be the main author of the PR" close. However the changes are not very large. I have thoroughly reviewed, built and tested it. This PR message is fully written by me. I fully understand all the code written. I spent some time manually testing the changes in a docker build.

Feel free to close the PR if that's not enough though.

### DIFF
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -12,6 +12,7 @@
        bind_address: "127.0.0.1"
        secret_key: "ultrasecretkey"           # change this!
        limiter: false
+       limiter_bypass_key: false
        public_instance: false
        image_proxy: false
        method: "POST"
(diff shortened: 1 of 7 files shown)


inetol:
If you need to exclude certain clients from the limiter, you would either disable the limiter (if the instance is private) or whitelist their IP addresses (if trusted_proxies is properly set):

- https://docs.searxng.org/admin/searx.limiter.html#limiter-toml

This PR doesn't address any scenarios that cannot already be handled using the current features.

PR #6229, opened 2026-06-09
[feat] engines: add resulthunter (web, images)

Bnyro:
### What does this PR do?
- add support for https://resulthunter.com
- this is especially interesting because resulthunter.com uses the Brave index, so it could be used to bypass Brave's ratelimits

- !reh test
- !rehi test

### Related issues
- #4653
### Code of Conduct

No AI used.


searx/engines/resulthunter.py
```diff
@@ -0,0 +1,117 @@
...
+import typing as t
+from urllib.parse import urlencode
+
+from lxml import html
+
+from searx.result_types import EngineResults
+from searx.utils import eval_xpath_list, eval_xpath, extract_text
+
+# as it uses brave internally, it has the same locales and timerange/safesearch types
+from searx.engines.brave import safesearch_map, time_range_map, fetch_traits  # pylint: disable=unused-import
```
return42, line 17: When we import `fetch_traits` from `brave` here, the *engine traits* of the Brave engine are adopted exactly as they are ..

searx/engines/resulthunter.py
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    # uses Brave's engine traits
+    engine_language = traits.get_language(params["searxng_locale"], "all")
+    if engine_language and engine_language != "all":
+        args["search_lang"] = engine_language
+
+    engine_region = traits.get_region(params["searxng_locale"], "all")
+    if engine_region and engine_region != "all":
+        args["country"] = engine_region
```
return42, line 64: .. are you sure that Brave and Resulthunter use similar but not identical values ​​(strings) for `country` selection?

For instance compare `args["country"]="de-CH"` from Relhunter (above) with `params["cookies"]["country"]="ch"` (below)


https://github.com/searxng/searxng/blob/e3d4fbe570e2a3ee8aad911e0c96ea78d45b0ac8/searx/engines/brave.py#L226-L227

return42 (approved):
Beside a question from my review, this PR is ready to merge.

searx/engines/resulthunter.py
Bnyro: Yes. The country at relhunter can on be set via search / query arguments (not persisted in the settings), and are passed as args, e.g. `https://resulthunter.com/search?q=test&search_type=web&freshness=&search_lang=en&country=GB`

searx/engines/resulthunter.py
Bnyro: Yes, that's my intention?

searx/engines/resulthunter.py
return42: `&country=GB` is not what your code `args["country"] = engine_region` does .. have a look at my example from brave, to get the `GB` part of the region string, you have to split the string and take the last item -> `engine_region.split("-")[-1].lower()`

searx/engines/resulthunter.py
return42: My question here relates to the *split string* used to retrieve the region value; I assume that if the two engines have identical traits, the values ​​for the country selection will also be identical.

searx/engines/resulthunter.py
Bnyro: > `&country=GB` is not what your code `args["country"] = engine_region` does .. have a look at my example from brave, to get the `GB` part of the region string, you have to split the string and take the last item -> `engine_region.split("-")[-1].lower()`
>
Are you sure? Tested with `en-US` as lang
```py
    engine_language = traits.get_language(params["searxng_locale"], "all")
    engine_region = traits.get_region(params["searxng_locale"], "all")
    logger.debug("lang: %s, region: %s", engine_language, engine_region)
```

```sh
DEBUG   searx.engines.resulthunter    : lang: all, region: us
```

Using `traits.get_language` is definitely wrong because it's stored in `traits.custom["ui_lang"]`, but I think the region part is okay as it is now.

Bnyro: Pushed the fix for the `ui_lang` and fixed a small xpath issue that would also try to parse news results on the general results page, [diff](https://github.com/searxng/searxng/compare/eec00544c45184b43c1aa2fc15882086bafa393d..13988f3493af175ea5b630d25802a24faa09340d)

return42: > Are you sure? Tested with `en-US` as lang

sorry, you are right .. but then the `.split("-")[-1]` in the Brave engine is superfluous: 

https://github.com/searxng/searxng/blob/e3d4fbe570e2a3ee8aad911e0c96ea78d45b0ac8/searx/engines/brave.py#L226-L227

Bnyro: Yeah, true :+1:

PR #6233, opened 2026-06-10
[feat] engines: add reloado engine (general, german)

Bnyro:
### What does this PR do?
- adds support for https://reloado.com (german)
- as it has its own index, the results are hit or miss and mostly German, but still worth integrating imo

- !rel test

### Code of Conduct

  No AI used.


return42 (approved):
LGTM 👍

PR #6230, opened 2026-06-09
[chore] devops: add languages config for helix editor

Bnyro:
The default Helix configuration for Python is different, so the pylint warnings aren't shown and the formatter re-formats files by accident when you edit an existing file.

Therefore, this commit adds `python` language configuration to ease developing SearXNG with Helix Editor [^1].

[^1]: https://helix-editor.com

### Related issues
- #5056 

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [X] **I hereby confirm that this PR conforms with the [AI Policy].**

  No AI.

PR #6241, opened 2026-06-11
 [chore] helix config: enable pylint and use black via pylsp

Bnyro:
I didn't know `black` can be called via `pylsp` and that pylint isn't enabled by default, hence a quick fixup for #6230

### Related issues
- #6230 

### Code of Conduct

No AI used.

### DIFF
--- a/.helix/languages.toml
+++ b/.helix/languages.toml
@@ -1,10 +1,11 @@
 [[language]]
 name = "python"
 language-servers = ["basedpyright", "pylsp"]
-formatter = { command = "black", args = [
-    "--target-version",
-    "py311",
-    "--line-length",
-    "120",
-    "--skip-string-normalization",
-] }
+auto-format = true
+
+[language-server.pylsp.config.pylsp]
+plugins.pylint.enabled = true
+plugins.isort.enabled = true
+plugins.black.enabled = true
+plugins.black.skip_string_normalization = true
+plugins.black.line_length = 120


return42:
FYI: not directly related to this PR, but related its context "development environments" ..  from ..

- https://github.com/searxng/searxng/pull/5075

> For integration into your IDE (editor) use the basedpyright-langserver (LSP) installed by `pipx basedpyright` and read:
> 
> * https://docs.basedpyright.com/latest/installation/ides/
> 
> 
> The `$REPO_ROOT/pyrightconfig.json` uses the virtualenv found in `$REPO_ROOT/local/py3` and can be created by a `make pyenv`

For instance, in my [emacs LSP config (eglot)](https://github.com/return42/sandbox/blob/2ddb9a10c31146dc3b5d82b67268ccdf43029958/src/dot.files/.emacs#L424) I just start the `basedpyright-langserver` in the `$REPO_ROOT` .. this service looks out for a virtual env configured `$REPO_ROOT/pyrightconfig.json`.

https://github.com/searxng/searxng/blob/a1490676e3c04239b3a314cb1fbb3144bc894b14/pyrightconfig.json#L2-L3

This ensures that the project's local virtual environment is always used (the one in which SearXNG runs). I don't know how it works with Helix, but many IDEs use a virtual environment located in the user's $HOME directory. However, this isn't the same environment in which SearXNG (`make run`) runs.

The global virtual environment in $HOME often has different packages installed, or even packages that aren't installed in the SearXNG environment. Only by using the project's local environment can you ensure that the IDE can represent the context that SearXNG will have at runtime.

----

While the project's local environment ensures that my IDE (Emacs) displays exactly the same messages that a `make test` would show me, it also introduces another complication if, like me, you frequently switch between branches:

If the requirements (package dependencies) differ between two branches, a `make run` builds a new local environment and deletes the old one.

LSP server running in the IDE will not survive the delete of the old environment / the LSP server in the IDE must be restarted (I restart Emacs completely to also clear all other open buffers, which I no longer need for the review of the new PR anyway).

Bnyro:
> This ensures that the project's local virtual environment is always used (the one in which SearXNG runs). I don't know how it works with Helix, but many IDEs use a virtual environment located in the user's $HOME directory. However, this isn't the same environment in which SearXNG (make run) runs.
>
> The global virtual environment in $HOME often has different packages installed, or even packages that aren't installed in the SearXNG environment. Only by using the project's local environment can you ensure that the IDE can represent the context that SearXNG will have at runtime.
>
As far as I can tell, Helix seems to properly start `basedpyright` from the same directory. I.e., I can see the documentation for all the dependencies, although some of them are not installed on my root system.

>  it also introduces another complication if, like me, you frequently switch between branches:

> If the requirements (package dependencies) differ between two branches, a make run builds a new local environment and deletes the old one.

> LSP server running in the IDE will not survive the delete of the old environment / the LSP server in the IDE must be restarted (I restart Emacs completely to also clear all other open buffers, which I no longer need for the review of the new PR anyway).
>
That sounds very ugly - I didn't test the new LSP config much yet, but I'm not motivated to run `lsp-restart` in Helix all the time. I'll see if I'll run into that issue too, thanks for the hint :)

PR #6182, opened 2026-05-30
[feat] engines: add kagi engine (general, news, images, videos)

bkgrndnoiz:
### What does this PR do?

Adds engine for Kagi API (v1)

https://www.kagi.com/

### How to test this PR locally?

Enable in settings with valid api_key

>general
- !kg test
>news
- !kgn test
>images
- !kgi test
>videos
- !kgv test

### Related issues

Closes: #2247

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [X] **I hereby confirm that this PR conforms with the [AI Policy].**

  If I have used AI tools for working on the changes in this PR, I will
  attach a list of all AI tools I used and how I used them. I hereby confirm
  that I haven't used any other tools than the ones I mention below.

PR #6244, opened 2026-06-11
[feat] engines: add support for Kagi (requires API key)

Bnyro:
### What does this PR do?
- add support for https://kagi.com
- requires an API key - the one I tested with was provided by @becks0815 (thanks for that!)

- uncomment the kagi part in `settings.yml`
- set the api key in all `api_key` fields
- start searching kagi, e.g. with `!kagi rust` 

### Related issues
- #2247 
- #4220 
- https://github.com/searxng/searxng/pull/6182

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/kagi.py
+++ b/searx/engines/kagi.py
@@ -1,0 +1,190 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Kagi_ is a paid, privacy-focused search engine.
+
+Using it requires an API key. If you have a Kagi account, you can obtain an API
+key in the `API portal`_.
+
+To enable Kagi, add the following to the ``engines`` seciton of
+``settings.yml``:
+
+.. code:: yaml
+
+  - name: kagi
+    engine: kagi
+    categories: [general, web]
+    shortcut: kg
+    api_key: ""
+    kagi_categ: search
+
+  - name: kagi.news
+    engine: kagi
+    categories: [news, web]
+    shortcut: kgn
+    api_key: ""
+    kagi_categ: news
+
+  - name: kagi.images
+    engine: kagi
+    categories: [images, web]
+    shortcut: kgi
+    paging: false
+    api_key: ""
+    kagi_categ: images
+
+  - name: kagi.videos
+    engine: kagi
+    categories: [videos, web]
+    shortcut: kgv
+    api_key: ""
+    kagi_categ: videos
+
+.. _Kagi: https://kagi.com
+.. _Api Portal: https://help.kagi.com/kagi/api/overview.html
+"""
+
+
+from datetime import datetime, timedelta
+
+import typing as t
+import html
+
+from searx.extended_types import SXNG_Response
+from searx.result_types import EngineResults
+from searx.utils import parse_duration_string
+
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineParams
+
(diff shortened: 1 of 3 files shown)


searx/engines/kagi.py
```diff
@@ -0,0 +1,185 @@
...
+            days=time_range_to_days_map[params["time_range"]]  # pyright: ignore[reportArgumentType]
+        )
+        oldest_result_date = datetime.now() - time_period
+        filters["after"] = oldest_result_date.strftime("%Y-%m-%d")
+
+    # there doesn't seem to be a list of languages anywhere,
+    # so we just assume that it supports all languages
+    if params["searxng_locale"] != "all":
+        country = params["searxng_locale"].split("-")[-1]
+        filters["region"] = country.lower()
```
return42: Unfortunately, that isn't possible "as is" (I [fixed](https://github.com/searxng/searxng/compare/d074a8a08fa466dd0a9da67af54a26fbd5d75f8b..b3397223bb5aa53041d427961649e7ce2728d23a) it); for instance, if one were to search for ":en foo", `en` would be treated as a region, causing the API to return a '400 Bad Request' error. The only viable option would be something like ":en-US foo"—but that completely contradicts user expectations, as users generally assume they will find results in a specific language regardless of the associated region. This is particularly true for global languages ​​like Spanish or English, where it is impossible to generate meaningful result lists without a language selection option.

The fact that the Kagi API does not offer a language selection feature is all the more frustrating given that the Kagi index itself provides data on the language relevance of a URL.  The more I look into Kagi, the less I think of it. 

Without a language selection option, an ordinary user cannot really make use of this engine's results; worse still, the engine clutters the result lists with irrelevant links to content in languages ​​the user does not understand.

We have far better alternatives in SearXNG that don't require a registration (API key).

return42 (commented):
Did some mods to fix an issue with the "region" (see below) ..

Implementation of the engine LGTM 👍 

----

*I’m really disillusioned by the Kagi's index and its API. TBH even if they gave me an API key for free, I wouldn’t use Kagi in my SearXNG instance.*

searx/engines/kagi.py
Bnyro: > Unfortunately, that isn't possible "as is" (I [fixed](https://github.com/searxng/searxng/compare/d074a8a08fa466dd0a9da67af54a26fbd5d75f8b..b3397223bb5aa53041d427961649e7ce2728d23a) it); for instance, if one were to search for ":en foo", en would be treated as a region, causing the API to return a '400 Bad Request' error.
>
Good catch! I didn't consider language codes that are not directly correlated to a country (such as `en`). I'm still frustrated that they failed to create a list of regions they support (the link in their docs for it gives a 404), and their documentation also doesn't match the actual API output. Some fields were named wrong (e.g. `videos` instead of `video`) and some fields were just completely missing (e.g. I couldn't find any example for an images response)...

> The fact that the Kagi API does not offer a language selection feature is all the more frustrating given that the Kagi index itself provides data on the language relevance of a URL. 
>
Yes, it's really sad that they lack such a fundamental feature even though they already have all the information needed in their index...

becks0815:
Hi, just wanted to say THX for the implementation, it works without problems. In case any developer needs to look into the kagi implementation, send a note to me please, and I will hand out an api key for this purpose.

PR #6250, opened 2026-06-12
[feat] engines: add German tonline engine (general, news, images, videos)

Bnyro:
T-Online is a German news portal. Also returns English results.

It gets its web results from Google, image results from Flickr and videos results from YouTube.

For images and videos, it additionally returns result from its news catalog. However, for pagination we have to specify the result type (e.g. either videos from YouTube or from T-Online), so we use flickr/youtube there instead of tonline because the tonline results are usually irrelevant.

I got a CAPTCHA once while testing - but I wasn't able to run into an other one after, so I'm not sure how to detect them.

- set the engine to active
- !tol test
- !toli test
- ...

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/tonline.py
+++ b/searx/engines/tonline.py
@@ -1,0 +1,149 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""T-Online_ is a German news portal, which is powered by Ströer, a German
+advertising company, not by Deutsche Telekom (contrary to its name).
+
+It gets its web results from Google, image results from Flickr and videos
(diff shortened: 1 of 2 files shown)


return42 (approved):
Nice engine for our DE audience 👍 

I just modified some nits, mainly to calm done the static type checkers: [diff](https://github.com/searxng/searxng/compare/443154dc48c5cb565e86451c68b0fe034ebeea1e..6cee08ea731c7b47dba779b1a4a51da68426cc88)

PR #6255, opened 2026-06-13
[feat] engines: add searchzee engine (general, news)

Bnyro:
The results seem to be from Brave (i.e. they are exactly the same). But it doesn't have any strict rate-limits, so that's nice.

News support time ranges, but apart from that, unfortunately it doesn't support any advanced features like safesearch or languages.

Possibly this engine is already too complicated for only being a `json_engine`, but since it doesn't require any tricky authentication and only supports general and news, `json_engine` is the quickest way to implement it

- set it to active n `settings.yml`
- !sz test
- !sznw test (szn was already taken)

### Code of Conduct

No AI used.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -2836,6 +2836,49 @@ engines:
     shortcut: rehi
     disabled: true
 
+  - name: searchzee
+    engine: json_engine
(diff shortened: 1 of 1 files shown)


return42 (approved):
LGTM

PR #6258, opened 2026-06-14
[chore] complete and normalize the attributes of engine objects

return42:
Drop outdated engine attributes: supported_languages, language_aliases

Complete, normalize and document the type definitions for the engine-module and engine-class.

For the ``engine.about`` section of the configuration, a type check is performed based on structure ``searx.enginelib.EngineAbout``.

The property ``engine.about.language`` no longer exists; existing values have been migrated to ``engine.language``.

### Code of Conduct

  no AI used

### DIFF
--- a/searx/enginelib/__init__.py
+++ b/searx/enginelib/__init__.py
@@ -41,7 +41,7 @@ if t.TYPE_CHECKING:
     from searx.enginelib.traits import EngineTraits
     from searx.extended_types import SXNG_Response
     from searx.result_types import EngineResults
-    from searx.search.processors import OfflineParamTypes, OnlineParamTypes
+    from searx.search.processors import OfflineParamTypes, OnlineParamTypes, ProcessorType
 
 ENGINES_CACHE: ExpireCacheSQLite = ExpireCacheSQLite.build_cache(
     ExpireCacheCfg(
@@ -180,7 +180,7 @@ class EngineCache:
         return ENGINES_CACHE.secret_hash(name=name)
 
 
-class EngineAbout(msgspec.Struct):
+class EngineAbout(msgspec.Struct, kw_only=True):
     """Additional fields describing the engine.
 
     .. code:: yaml
@@ -213,12 +213,11 @@ class EngineAbout(msgspec.Struct):
     results: str = ""
     """Data format of the source (online-engines: of the response)."""
 
-    language: str = ""
-    """If the engine supports only one language, this language is specified
-    here (``en``, ``de``, ``"no"`` or ..); otherwise, the value remains empty.
+    description: str = ""
+    """Brief description of the engine and where it gets its data from.
 
-    For the YAML configuration: think of the `YAML-Norway problem
-    <https://ruuda.nl/2023/the-yaml-document-from-hell#the-norway-problem>`_
+    This value should only be set as long as no description of the data source
+    is available via a :py:obj:`EngineAbout.wikidata_id`.
     """
 
 
@@ -227,6 +226,8 @@ class Engine(abc.ABC):  # pylint: disable=too-few-public-methods
 
     Further documentation see :ref:`general engine configuration`.
 
+    The defaults are taken from :py:obj:`searx.engines.ENGINE_DEFAULT_ARGS`.
+
     .. hint::
 
        This class is currently never initialized and only used for type hinting.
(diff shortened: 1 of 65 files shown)


searx/enginelib/__init__.py
```diff
@@ -213,12 +213,11 @@ class EngineAbout(msgspec.Struct):
...
 
-    language: str = ""
-    """If the engine supports only one language, this language is specified
-    here (``en``, ``de``, ``"no"`` or ..); otherwise, the value remains empty.
+    description: str = ""
+    """Brief description of the engine and where it gets its data from.
 
-    For the YAML configuration: think of the `YAML-Norway problem
-    <https://ruuda.nl/2023/the-yaml-document-from-hell#the-norway-problem>`_
+    This value should only be set as long as no description of the data source
```
Bnyro, line 219: I already thought about adding such a field too to document `xpath` and `json_engine` engines, good idea 👍

searx/enginelib/__init__.py
```diff
@@ -309,41 +322,41 @@ class Engine(abc.ABC):  # pylint: disable=too-few-public-methods
...
     """Weighting of the results of this engine (:ref:`weight <settings engines>`)."""
 
+    proxies: dict[str, dict[str, str]]
+    """Set proxies for a specific engine (YAML):
+
+    .. code:: yaml
+
+       proxies :
+         http:  socks5://proxy:port
+         https: socks5://proxy:port
```
Bnyro, line 357: Couldn't we just set `all://: socks5://proxy:port` here?

searx/enginelib/traits.py
```diff
@@ -159,22 +159,25 @@ def _set_traits_v1(self, engine: "Engine | types.ModuleType") -> None:
...
-        if hasattr(engine, "language"):
-            if engine.language not in languages:
-                raise ValueError(_msg % (engine.name, "language", engine.language))
-            traits.languages = {engine.language: languages[engine.language]}
-
-        regions = traits.regions
-        if hasattr(engine, "region"):
-            if engine.region not in regions:
+        if engine.language:
+            if engine.language_support:
```
Bnyro, line 163: Why do we need this distinction here with `language_support`? Can't we just check if there are traits, and if 
- yes: we limit the traits to only this one language
- no: we do nothing

Bnyro: Also, I've never set `language_support = True`, but languages worked nevertheless. So I still don't really get what `language_support` exists for?

searx/enginelib/__init__.py
return42: This field exists since a long time, https://github.com/searxng/searxng/pull/811 .. has, however, fallen into oblivion due to the lack of documentation. With this PR, we now have documentation for this field in a central place.

*My motivation for this PR was precisely such reasons, to either eliminate many lost, by now partly redundant switches or to firmly fix their meaning.*

searx/enginelib/__init__.py
return42: We would presumably do that in the `outgoing.proxy` settings?

https://github.com/searxng/searxng/blob/b3e08f2a44640ffc45d5955f5ff43b4a4d555578/searx/settings.yml#L202-L205

These are the individual proxy settings for a specific engine; do you perhaps have a better example that we could include in the documentation?

searx/enginelib/traits.py
return42: > Why do we need this distinction here with `language_support`? Can't we just check if there are traits ..

Nope, there are various engines that support languages ​​without requiring traits; for some online engines, the `Accept-Language` header sent by default is enough.

----

The interpretation of this `language_support` became somewhat unclear over time.

In the past, this option served a purely informational purpose; currently, we only evaluate it in the online documentation .. 

https://github.com/searxng/searxng/blob/b3e08f2a44640ffc45d5955f5ff43b4a4d555578/docs/user/configured_engines.rst?plain=1#L73-L75

I believe somewhere in the past we used to have this option in the HTML templates for the (engine) preferences page as well.

-----

We need to pin down the interpretation of this option (even if it is purely informational): I am of the opinion that only the engine developer can assess whether an engine (in general) is able to support languages; therefore, they must set this option in the engine's source code.

return42: > Also, I've never set `language_support = True`, but languages worked nevertheless.

Same on my side ;-) .. thanks for reminding .. to fix, I placed a commit on top:

- [``[fix] set language_support for engines with languages in traits``](https://github.com/searxng/searxng/pull/6258/commits)

searx/enginelib/__init__.py
Bnyro: No, I was just wondering why we need two lines, one for `http` and one for `https` instead of setting both at the same time `all://`

Bnyro (approved):

PR #6259, opened 2026-06-14
[feat] engines: add Sina engine (News)

hezhijie0327:
### What does this PR do?

Add support for Sina news (China)

https://search.sina.com.cn/

`!sina test`

image

### Code of Conduct

No AI used. It's a JSON format response, easy to impl.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -3110,6 +3110,36 @@ engines:
     base_url: https://info.searchtoday.site
     disabled: true
 
+  - name: sina
+    engine: json_engine
+    shortcut: sina
+    categories: news
+    paging: true
+    first_page_num: 1
+    search_url: https://search.sina.com.cn/api/search?q={query}&tp=mix&sort=0&page={pageno}&size=10&from=search_result
+    results_query: data/list
+    url_query: url
+    title_query: title
+    content_query: intro
+    thumbnail_query: thumb
+    title_html_to_text: true
+    time_range_map:
+      day: d
+      week: w
+      month: m
+      year: y
+    time_range_support: true
+    time_range_url: "&from=advanced_search&time={time_range_val}"
+    disabled: true
+    inactive: true
+    language: zh
+    about:
+      website: https://search.sina.com.cn
+      wikidata_id: Q938668
+      use_official_api: false
+      require_api_key: false
+      results: JSON
+
   # - name: webcrawler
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Works fine for me, thanks 👍 

The only thing I noticed is that if there are no results, the `json_engine` crashes:
```
ERROR   searx.engines.sina            : exception : 'NoneType' object is not iterable
  File "/home/bnyro/Projects/searxng/searx/search/processors/online.py", line 253, in search
  File "/home/bnyro/Projects/searxng/searx/search/processors/online.py", line 239, in _search_basic
  File "/home/bnyro/Projects/searxng/searx/engines/json_engine.py", line 416, in response
    for result in rs:
                  ^^
TypeError: 'NoneType' object is not iterable
```

I'm not sure if we can handle that - the JSON engine currently assumes that the `results_query` exists in the JSON result, even if there are no results. But if we changed that to ignore all errors when `results_query` doesn't exist, it's possible that some `json_engine`s just always return 0 results and we don't notice that the `results_query` is false. Opinion @return42 ?

searx/settings.yml
```diff
@@ -3110,6 +3110,34 @@ engines:
...
+      year: y
+    time_range_support: true
+    time_range_url: "&from=advanced_search&time={time_range_val}"
+    disabled: true
+    inactive: true
+    about:
+      website: https://search.sina.com.cn
+      use_official_api: false
+      require_api_key: false
+      results: JSON
```
Bnyro, line 3141:
```suggestion
      results: JSON
      language: cn
```

searx/settings.yml
```diff
@@ -3110,6 +3110,34 @@ engines:
...
+      day: d
+      week: w
+      month: m
+      year: y
+    time_range_support: true
+    time_range_url: "&from=advanced_search&time={time_range_val}"
+    disabled: true
+    inactive: true
+    about:
+      website: https://search.sina.com.cn
```
Bnyro, line 3137:
```suggestion
      website: https://search.sina.com.cn
      wikidata_id: Q938668
```

hubutui:
Nice job! Maybe you could also add other engines from sina. Currently, you set `tp=mix` (综合), could you also add:

```text
tp=news (新闻)
tp=slide (图集)
tp=video (视频)
```

PR #6267, opened 2026-06-16
[feat] engines: add tusksearch (web, news, videos, images)

Bnyro:
### What does this PR do?
- add support for tusksearch.com
- the engine supports setting a political direction like "left", "right" or "center" - i decided to set it to center and not provide an option for instance operators to bias the results into the left or right direction (although from my testing, this setting doesn't really move the results into a specific direction, their filter doesn't seem to work that well tbh)
- more info in the commit message

- !tu test
- !tui test
- ...

### Code of Conduct

I've used Claude code with a throw-away account to help reverse engineering the engine (after I spent countless hours failing to figure out where the headers are set due to code obfuscation).


return42 (approved):
LGTM

PR #6272, opened 2026-06-16
[feat] engines: add magnific images engine

Bnyro:
### What does this PR do?
- add support for https://magnific.com
- run by the same people as flaticon (#6122)
- unfortunately you need to login to get the full resolution image, but the 600px thumbnail images are not too bad for many use cases

- !mag tree

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/magnific.py
+++ b/searx/engines/magnific.py
@@ -1,0 +1,62 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Magnific_ is a database for images.
+
+.. _Magnific: https://www.magnific.com
+"""
+
+from urllib.parse import urlencode
+
+import typing as t
+
+from searx.result_types import EngineResults
+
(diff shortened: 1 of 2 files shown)


searx/engines/magnific.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    args = {"term": query, "filters[ai-generated][excluded]": 1, "page": params["pageno"], "locale": "en"}
+    if free_images_only:
+        args["filters[license]"] = "free"
+
+    params["headers"].update(
+        {
+            "Referer": f"{base_url}/search?{urlencode(args)}",
```
return42: Is it really needed to add the `args` to the `Referer`?

Bnyro: You're right, it's not. Removed the args.

PR #6275, opened 2026-06-18
[feat] engines: add support for search.ch/web (Swiss)

Bnyro:
### What does this PR do?
- add support for https://search.ch/web

- !sch test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/xpath.py
+++ b/searx/engines/xpath.py
@@ -292,7 +292,7 @@ def response(resp) -> EngineResults:  # pylint: disable=too-many-branches
 
     if results_xpath:
         for result in eval_xpath_list(dom, results_xpath):
-            url = extract_url(eval_xpath_list(result, url_xpath, min_len=1), search_url)
+            url = extract_url(eval_xpath(result, url_xpath), search_url)
             title = extract_text(eval_xpath_list(result, title_xpath, min_len=1))
             content = extract_text(eval_xpath_list(result, content_xpath))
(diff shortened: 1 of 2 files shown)


searx/engines/xpath.py
```diff
@@ -292,7 +292,7 @@ def response(resp) -> EngineResults:  # pylint: disable=too-many-branches
 
     if results_xpath:
         for result in eval_xpath_list(dom, results_xpath):
-            url = extract_url(eval_xpath_list(result, url_xpath, min_len=1), search_url)
+            url = extract_url(eval_xpath(result, url_xpath), search_url)
```
Bnyro, line 295: Unfortunately I had to change this because the `url_xpath` of the engine doesn't return a list of xpath elements, but a string instead.

return42: > of the engine doesn't return a list of xpath elements,

do you expect any side effects for the other engines / did you tested some of the other engines?

return42 (approved):
LGTM / except the question from my review ..

searx/engines/xpath.py
Bnyro: The only difference this makes is that we don't enforce a minimum length of 1 found xpath item.

In practice, this means that now `extract_url` raises an error `Empty url ersultset`, while previously `eval_xpath_list` has thrown that error.

So engines that were working before still work - and broken engines are still broken :)

PR #6276, opened 2026-06-18
fix: use correct iar parameter for DuckDuckGo news search

### LINKED ISSUE #6257: Duckduckgo news returns general results instead of news articles.
_Duckduckgo News: returns general results instead of news articles._

### How To Reproduce?

Enable `duckduckgo news` engine. Search news of any keywords.

### Expected behavior

Results should be news articles like on duckduckgo.com "News" tab.

### Screenshots & Logs

Image

Image

### Version of SearXNG

2026.6.13+d14fa1f6e

docker; searxng/searxng:latest

### Additional context

Using the devtools I confirmed that duckduckgo "News" Tab uses `iar=news` and returns a **different** vqd value than the
same query with `iar=images`.

### Code of Conduct

decembercomposer697-hue:
Fixes #6257: DuckDuckGo news VQD token was fetched with iar=images
instead of iar=news, causing news searches to return general results.

### DIFF
--- a/searx/engines/duckduckgo_extra.py
+++ b/searx/engines/duckduckgo_extra.py
@@ -60,7 +60,7 @@ def fetch_vqd(
 
     logger.debug("fetch_vqd: request value from from duckduckgo.com")
     resp = get(
-        url=f"https://duckduckgo.com/?q={quote_plus(query)}&iar=images&t=h_",
+        url=f"https://duckduckgo.com/?q={quote_plus(query)}&iar={iar_param}&t=h_",
         headers=params["headers"],
         timeout=2,
     )


searx/engines/duckduckgo_extra.py
```diff
@@ -60,7 +60,7 @@ def fetch_vqd(
 
     logger.debug("fetch_vqd: request value from from duckduckgo.com")
     resp = get(
-        url=f"https://duckduckgo.com/?q={quote_plus(query)}&iar=images&t=h_",
+        url=f"https://duckduckgo.com/?q={quote_plus(query)}&iar={iar_param}&t=h_",
```
return42, line 63: Where is this `iar_param` variable defined?

    ************* Module searx.engines.duckduckgo_extra
    searx/engines/duckduckgo_extra.py:63:65: E0602: Undefined variable 'iar_param' (undefined-variable)

Bnyro (commented):
I think that also doesn't solve the issue 100% because the `vqd` tokens are cashed and shared among different duckduckgo engines (news, videos, images, ...). This means, you would also need to update the cashing logic so that it additionally remembers the `iar_param`.

E.g. if you do the following:
- Search `test` in the images category -> vqd will be cached
- Search `test` in the news category -> reuses cached vqd

Then, DuckDuckGo news will still use the wrong vqd value from the images. I.e., https://github.com/searxng/searxng/blob/master/searx/engines/duckduckgo.py#L230-L233 would need to be changed as well.

return42:
I'm going to close this PR .. next time, please read [Development Quickstart](https://docs.searxng.org/dev/quickstart.html), run test locally and fill out the PR template.

PR #6286, opened 2026-06-19
fix(kagi): safe access unpromised result properties

### LINKED ISSUE #6252: Kagi Engine Parse Errors
Followup issue from

- https://github.com/searxng/searxng/issues/2247#issuecomment-4693877163

I added the new engine to my instance and configured it, thanks for working on it! I did get some results from the engine, but did start running into a parse issue.
Image

example query: searx metrics

I did some quick testing and found that the snippet attribute isn't guaranteed in the response, such as this one:
Image

[Their API documentation](https://kagi.redocly.app/api/docs/openapi/search/search#search/search/t=response&c=200&path=data/search) seems to indicate the same, only the url and title properties are required in the search result object.

### Expected behavior

Successful query response parsing.

### Screenshots & Logs

Image

```
2026-06-12 15:49:02,306 WARNING:searx.engines.kagi: ErrorContext('searx/engines/kagi.py', 152, 'content=html.unescape(result["snippet"]),', 'KeyError', None, ()) False
2026-06-12 15:49:02,306 ERROR:searx.engines.kagi: exception : 'snippet'
  File "/usr/local/searxng/searx/search/processors/online.py", line 253, in search
  File "/usr/local/searxng/searx/search/processors/online.py", line 239, in _search_basic
  File "/usr/local/searxng/searx/engines/kagi.py", line 152, in response
    content=html.unescape(result["snippet"]),
                          ~~~~~~^^^^^^^^^^^
KeyError: 'snippet'
```

Image

### Version of SearXNG

commit hash de8a3de15adbb5b72dd106444e3d0a70bfd196b5

issue occurs on docker and local build (`make run`)

### Code of Conduct

holysoles:
kagi pr

### What does this PR do?

I was testing the new Kagi engine and found for some queries I was getting a `KeyError` exception from the result parsing.

[Their API documentation](https://kagi.redocly.app/api/docs/openapi/search/search#search/search/t=response&c=200&path=data/search) clarifies that only the `url` and `title` properties are required/guaranteed in the search result object, for example:

Image

This PR ensures we only assumes the `url`  key exists, and we use `.get()` to retrieve values for keys that may not be present.

- checkout this branch
- uncomment the kagi part in settings.yml
- set the api key in all api_key fields (let me know if you need to borrow one)
- start the server with `make run`
- start searching kagi, e.g. with `!kagi searx metrics`, `!kagi tabby cat`

### Related issues

- Closes #6252
- #2247 ([comment](https://github.com/searxng/searxng/issues/2247#issuecomment-4692976877))

### Code of Conduct

No AI used

### DIFF
--- a/searx/engines/kagi.py
+++ b/searx/engines/kagi.py
@@ -139,9 +139,13 @@ def response(resp: "SXNG_Response") -> EngineResults:
 
     if kagi_categ in ("images", "videos"):
         # the JSON key is "image" for "images" and "video" for "videos"
-        json_results = json_data["data"][kagi_categ[:-1]]
+        json_results = json_data["data"].get(kagi_categ[:-1])
     else:
-        json_results = json_data["data"][kagi_categ]
+        json_results = json_data["data"].get(kagi_categ)
+
+    # if no results were found, the response doesn't contain the results field
+    if not json_results:
+        return res
 
     for result in json_results:
         published_date: datetime | None = None
--- a/searx/engines/kagi.py
+++ b/searx/engines/kagi.py
@@ -42,10 +42,10 @@ To enable Kagi, add the following to the ``engines`` seciton of
 .. _Api Portal: https://help.kagi.com/kagi/api/overview.html
 """
 
-
 from datetime import datetime, timedelta
 
 import typing as t
+
 
 from searx.extended_types import SXNG_Response
 from searx.result_types import EngineResults
@@ -76,7 +76,12 @@ kagi_categ: t.Literal["search", "images", "news", "videos"] = "search"
 base_url = "https://kagi.com"
 
 safe_search_map = {0: False, 1: True, 2: True}
-time_range_to_days_map: dict[TimeRangeType, int] = {"day": 1, "week": 7, "month": 30, "year": 365}
+time_range_to_days_map: dict[TimeRangeType, int] = {
+    "day": 1,
+    "week": 7,
+    "month": 30,
+    "year": 365,
+}
 
 api_key = ""
 """Kagi API key. Required for using this engine."""
@@ -147,8 +152,8 @@ def response(resp: "SXNG_Response") -> EngineResults:
             res.add(
                 res.types.MainResult(
                     url=result["url"],
-                    title=html_to_text(result["title"]),
-                    content=html_to_text(result["snippet"]),
+                    title=html_to_text(result.get("title", "no title available")),
+                    content=html_to_text(result.get("snippet", "")),
                     thumbnail=result.get("image", {}).get("url") or "",
                     publishedDate=published_date,
                 )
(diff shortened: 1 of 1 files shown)


return42:
Please read [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) and fix issue reported by the GitHub-CI .. 

```
--- /home/runner/work/searxng/searxng/searx/engines/kagi.py	2026-06-19 06:00:02.016059+00:00
+++ /home/runner/work/searxng/searxng/searx/engines/kagi.py	2026-06-19 06:00:09.798055+00:00
@@ -159,13 +159,11 @@
                 res.types.Image(
                     url=result["url"],
                     title=html.unescape(result.get("title", "")),
                     img_src=result.get("image", {}).get("url"),
                     resolution=f"{result.get('image', {}).get('width')}x{result.get('image', {}).get('height')}",
-                    thumbnail_src=result.get("props", {})
-                    .get("thumbnail", {})
-                    .get("url"),
+                    thumbnail_src=result.get("props", {}).get("thumbnail", {}).get("url"),
would reformat /home/runner/work/searxng/searxng/searx/engines/kagi.py
                 )
             )
         elif kagi_categ == "videos":
             length: timedelta | None = None
             if result.get("props", {}).get("duration"):

Oh no! 💥 💔 💥
```

searx/engines/kagi.py
```diff
@@ -148,8 +150,8 @@ def response(resp: "SXNG_Response") -> EngineResults:
             res.add(
                 res.types.MainResult(
                     url=result["url"],
-                    title=html.unescape(result["title"]),
-                    content=html.unescape(result["snippet"]),
+                    title=html.unescape(result.get("title", gettext("no title available"))),
```
return42: SearXNG requires a title string, otherwise the result will not shown in the result list.

return42 (approved):
> Please read [Development Quickstart](https://docs.searxng.org/dev/quickstart.html) and fix issue reported by the GitHub-CI ..

I fixed the issue reported by the CI, I added a translated string to the title if there is no title and I modified the commit message.

I force pushed my modifications on your branch, can you please have a final look before we merge / thanks 👍

searx/engines/kagi.py
holysoles: ah got it, thank you for the catch. I ran the checks before, but not after updating title to be optional, apologies.

Looks great!

Bnyro: I don't think it's worth adding a translatable string for that - 99% of results will have a title and as this is not used on other engines, it's not worth letting 50+ translators translate this into their own language. I'd just do
```suggestion
                    title=html.unescape(result.get("title", "no title available")),
```

Bnyro: Also, @holysoles, are you really sure that results can have no `title` field on Kagi? Or did you just make everything here optional to be on the safe side? I would personally guess that the `title` is always present.

holysoles: @Bnyro I had asked Kagi support about the `title` missing from specifically image results, and they replied with the following on Thursday.

> The image results "title' should be marked as optional, as many images likely don't have titles - as you've noticed.
>The only required field there should be the "url" field.
>
>We'll update the documentation to reflect this, thanks for pointing it out!

So I made the update to all `title` access to be on the safe side.

Their docs are still not fixed though, so it is  unclear if they mean all results or just images. Happy to wait until they do so for further confirmation, whatever you prefer.

Bnyro: We don't need to wait for them to answer, thanks for clarifying :+1:

holysoles: Sorry for the delay here @Bnyro , I've now addressed the feedback for removing the translatable string

Bnyro (approved):
When testing, I noticed that Kagi currently crashes when there are no results found, so I fixed that [in a follow-up commit](https://github.com/searxng/searxng/pull/6286/changes/634800c85669b322d5b63a1fdec743493d55fa4f).

I also fixed the CI, so good to merge now.

Thank you!

PR #6294, opened 2026-06-19
[fix] autocomplete suggestrelevance scores for chromium

vojkovic:
### What does this PR do?

If SearXNG is set as the default search engine in a Chromium based browser, the url bar only shows 3 autocomplete suggestions max even if there are more.

This PR fixes that by adding `google:suggestrelevance` scores to the autocomplete response when using the browser url bar. This is a seemingly undocumented feature in Chromium. There's also more I found like `google:suggestdetail` which allows for images in the autocomplete, which I am currently working on.

`make run`, select an autocomplete method in preferences, add `https://example.com/autocompleter?q=%s` to your default search engine's suggestions url in chromium.

image

Verify more than three results are returned

### Code of Conduct

AI was used to help find the undocumented feature

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -845,7 +845,10 @@ def autocompleter():
         mimetype = 'application/json'
     else:
         # the suggestion request comes from browser's URL bar
-        suggestions = json.dumps([sug_prefix, results])
+        relevances = {
+            'google:suggestrelevance': [600 - i for i in range(len(results))]
+        }  # chromium only shows 3 suggestions unless we attach relevances
+        suggestions = json.dumps([sug_prefix, results, [], [], relevances])
         mimetype = 'application/x-suggestions+json'
 
     suggestions = escape(suggestions, False)


searx/webapp.py
```diff
@@ -845,7 +845,10 @@ def autocompleter():
         mimetype = 'application/json'
     else:
         # the suggestion request comes from browser's URL bar
-        suggestions = json.dumps([sug_prefix, results])
+        relevances = {
+            'google:suggestrelevance': [600 - i for i in range(len(results))]
```
Bnyro, line 849: Where does the 600 come from? 600 = best result score? maybe add a comment

Bnyro:
You can just use Firefox, then you can see all suggestions without any weird hacks ;)

searx/webapp.py
vojkovic: Just any number works, it's not magic, it just needs to be high enough to decrement and not go negative.

Google's autocomplete were mostly around 600 i found so I just went with that

vojkovic:
FYI just rebased on latest master

return42 (approved):
LGTM .. thanks 👍 

@Bnyro do you have any open questions?

Bnyro (approved):
Lgtm, feel free to merge @vojkovic  👍

PR #6297, opened 2026-06-20
[feat] engines: add giphy (images/videos)

Bnyro:
### What does this PR do?
- add support for giphy.com
- not sure if's well suited for the images category because GIFs are in fact videos, but I still think that GIFs are rather images than videos

- !gif test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/giphy.py
+++ b/searx/engines/giphy.py
@@ -1,0 +1,127 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Giphy (images)"""
+
+import random
+from urllib.parse import urlencode
+import re
+
+import typing as t
+
+from lxml import html
+
+from searx.enginelib import EngineCache
+from searx.exceptions import SearxEngineAPIException
+from searx.network import get
+from searx.result_types import EngineResults
+from searx.result_types.image import ImageRef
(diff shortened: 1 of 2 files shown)


searx/engines/giphy.py
```diff
@@ -0,0 +1,126 @@
...
+}
+
+base_url = "https://giphy.com"
+api_url = "https://api.giphy.com"
+
+categories = ["images"]
+paging = True
+page_size = 15
+
+giphy_categ: t.Literal["gifs", "stickers", "clips"] = "gifs"
```
return42:
```python
>>> GiphyCateg = t.Literal["gifs", "stickers", "clips"]
>>> GIPHY_CATEGORIES = t.get_args(GiphyCateg)
>>> giphy_categ: GiphyCateg = "gifs"
>>> giphy_categ in GIPHY_CATEGORIES
True
```

return42 (approved):
LGTM

searx/engines/giphy.py
Bnyro: If I only knew that before duplicating the type information for 20+ engines for checking if the category is right. :)

Thanks!

PR #6313, opened 2026-06-26
[feat] bilibili engine: add time range support

hubutui:
### What does this PR do?

Adds time range filtering (`day`, `week`, `month`, `year`) to the Bilibili video engine by passing `pubtime_begin_s` and `pubtime_end_s` to the Bilibili search API.

1. Enable the Bilibili engine in `searx/settings.yml`.
2. Start SearXNG and search in the web UI with `!bil <keyword>`.
3. Select a time range filter and confirm results are returned.

### Code of Conduct

  - **Cursor**: assisted with implementing the code changes in `searx/engines/bilibili.py`.

### DIFF
--- a/searx/engines/bilibili.py
+++ b/searx/engines/bilibili.py
@@ -8,6 +8,7 @@ import random
 import string
 from urllib.parse import urlencode
 from datetime import datetime, timedelta
+from zoneinfo import ZoneInfo
 
 from searx import utils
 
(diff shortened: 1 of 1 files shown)


searx/engines/bilibili.py
```diff
@@ -50,6 +54,15 @@ def request(query, params):
         "search_type": "video",
     }
 
+    if params.get("time_range") in time_range_dict:
+        cn_tz = timezone(timedelta(hours=8))
+        now = datetime.now(cn_tz)
+        pubtime_end_s = int(now.replace(hour=23, minute=59, second=59, microsecond=0).timestamp())
+        begin_day = now - timedelta(days=time_range_dict[params["time_range"]])
+        pubtime_begin_s = int(begin_day.replace(hour=0, minute=0, second=0, microsecond=0).timestamp())
```
Bnyro: Please move this into a helper method and add a docstring comment to explain how this works

searx/engines/bilibili.py
```diff
@@ -50,6 +54,15 @@ def request(query, params):
         "search_type": "video",
     }
 
+    if params.get("time_range") in time_range_dict:
+        cn_tz = timezone(timedelta(hours=8))
+        now = datetime.now(cn_tz)
+        pubtime_end_s = int(now.replace(hour=23, minute=59, second=59, microsecond=0).timestamp())
```
Bnyro: Why do you set this to the end of the day? Do you expect Bilibili to serve videos that are from the future?

searx/engines/bilibili.py
Bnyro: Why do you set hour, minute, ... to 0? Does Bilibili not work otherwise? Searching within the last 7 days means searching withing the last 7 * 24 hours, so you shouldn't start to search from the beginning of the day.

searx/engines/bilibili.py
```diff
@@ -50,6 +54,15 @@ def request(query, params):
         "search_type": "video",
     }
 
+    if params.get("time_range") in time_range_dict:
+        cn_tz = timezone(timedelta(hours=8))
```
Bnyro: Please use [ZoneInfo](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfo), e.g. `ZoneInfo("Asia/Shanghai")`

searx/engines/bilibili.py
hubutui: I'm replicating the official bilibili web behavior. When you select 'recent day' on their web search, the request actually uses 00:00:00 to 23:59:59 of the current day (Beijing time) as the time range. This is not arbitrary — bilibili's UI treats 'recent day' as a calendar day (today), not a rolling 24-hour window.

I'd prefer to keep it as-is to stay consistent with the web client's default parameters. This also makes the behavior predictable: users get the same result set for the same date, which aligns with how the official site works. A rolling window would be more logically correct in absolute terms, but it would deviate from the official implementation.

Bnyro: I checked it on their website, you're right. Let's keep it this way then.

Bnyro: Anyways, please add a comment into the code because it's not self-explaining why we're doing this.

Bnyro (approved):
Thank you!

Bnyro:
Sorry for the delay, I could have sworn I already pressed "merge" 3 days ago :/

PR #6314, opened 2026-06-26
[fix] engine baidu: fix image search blocked by antiFlag

hubutui:
### What does this PR do?

Fixes Baidu image search failing with `antiFlag: 1` when the `acjson` API is called without session cookies from `image.baidu.com`.

The engine now warms up cookies with a GET to the image homepage, caches them for one hour via `EngineCache`, and attaches them to image search requests. If Baidu still returns `antiFlag`, the engine raises `SearxEngineAccessDeniedException` instead of returning empty or broken results.

1. Enable the **baidu images** engine in `searx/settings.yml` (set `disabled: false` for the `baidu images` entry).
2. Start SearXNG and search in the web UI with `!bdi <keyword>`.
3. Confirm image results are returned instead of an access-denied error.

### Code of Conduct

  - **Cursor**: assisted with implementing the code changes in `searx/engines/baidu.py`.

### DIFF
--- a/searx/engines/baidu.py
+++ b/searx/engines/baidu.py
@@ -7,13 +7,17 @@
 # There exits a https://github.com/ohblue/baidu-serp-api/
 # but we don't use it here (may we can learn from).
 
+import typing as t
+
 from urllib.parse import urlencode
 from datetime import datetime
 from html import unescape
 import time
 import json
 
-from searx.exceptions import SearxEngineAPIException, SearxEngineCaptchaException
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
> Enable the baidu images engine in searx/settings.yml (set disabled: false for the baidu images entry).
>
You don't even need to do that, `!bdi <query>` works without it. It only doesn't work if you set `inactive: true`.

Looks good to me, although I also don't get ratelimited without these cookies. But if it helps for you, we should merge it.

hubutui:
yes, I have rebase to the latest commit, and there are no conflicts now. These cookies helps me a lot.

PR #6322, opened 2026-06-26
[mod] images.html: display all engines that found the result, not only the first one

Bnyro:
### What does this PR do?
- display all image engines that returned the image instead of only the first one
- that makes it behave similar to normal results

- search for images, e.g. `!images libretube`
- look at the image details, some should now have multiple engines listed

### Code of Conduct

No AI used.

### DIFF
--- a/searx/templates/simple/result_templates/images.html
+++ b/searx/templates/simple/result_templates/images.html
@@ -60,7 +60,7 @@
       </p>
       <p class="result-filesize">{{ _label(_("Filesize"), result.filesize) }}</p>
       <p class="result-source">{{ _label(_("Source"), result.source) }}</p>
-      <p class="result-engine">{{ _label(_("Engine"), result.engine) }}</p>
+      <p class="result-engines">{{ _label(_("Engines"), ", ".join(result.engines)) }}</p>
       <p class="result-url"><span>{{ _("View source") }}:</span>{{- "" -}}
         <a {{ _target(results_on_new_tab) }} href="{{ result.url }}">{{ result.url }}</a>
       </p>


searx/templates/simple/result_templates/images.html
```diff
@@ -60,7 +60,7 @@ <h4>{{ result.title | striptags }}</h4>
       </p>
       <p class="result-filesize">{{ _label(_("Filesize"), result.filesize) }}</p>
       <p class="result-source">{{ _label(_("Source"), result.source) }}</p>
-      <p class="result-engine">{{ _label(_("Engine"), result.engine) }}</p>
+      <p class="result-engines">{{ _label(_("Engines"), ", ".join(result.engines)) }}</p>
```
Bnyro, line 63: For better HTML semantics, we could make this an `ul` and put all engines into `li`s, but that would require much more changes I don't think it's that useful

return42: For me it is OK .. here is how other templates answer this question:

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/templates/simple/macros.html#L49-L52

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/templates/simple/result_templates/keyvalue.html#L19

return42 (approved):
LGTM 👍

searx/templates/simple/result_templates/images.html
Bnyro: Thanks for the pointer - I quickly tried if that also works here, but that's a bit more complex because the `_label` macro we are using here doesn't accept HTML input and we would need to duplicate the label logic for the results here.

So we'll go for my initial approach for now.

PR #6328, opened 2026-06-28
fix(google engine): raise diagnostic error when page has 0 organic results

### LINKED ISSUE #6171: [Google engine] Successful responses can produce no parsed results without diagnostics
Google engine can return successful responses with no parsed results and no visible engine error.

Some Google queries return no visible results in SearXNG even though the request appears to complete successfully. This does not seem to be limited to one instance, one proxy setup, or one query. Multiple tested instances are affected to varying degrees, @return42 and @vojkovic reproduced the same general behavior on their own instances with different queries.

At least one captured example does not look like a CAPTCHA, consent page, or Google block page. Instead, Google returned a valid-looking localized search engine results page containing a direct answer widget, but no normal organic result blocks.

### How To Reproduce?

Test query:

```
!google :de-DE time
```

The attached files are raw upstream Google HTML responses captured from SearXNG on my infrastructure for the test query.
- [no_results.html](https://github.com/user-attachments/files/28395066/no_results.html)
- [results.html](https://github.com/user-attachments/files/28395067/results.html)

Jefsky:
## Problem

Google can return a 200 OK response that contains only non-organic page elements (direct answer widgets, knowledge panels, weather widgets, "no results found" messages, etc.) without any organic result blocks. The engine currently silently returns an empty result set without any diagnostic, leaving users confused — they see an empty search page with no indication of what happened.

Reported in #6171 with a specific example: `!google :de-DE time` — Google returns a direct answer widget (current time in Berlin) but no organic result blocks.

## Root Cause

The `response()` function parses organic results using XPath `//a[@data-ved and not(@class)]`. When Google returns a page without organic results (e.g. direct answer, knowledge panel, weather widget), this XPath yields nothing and the function returns an empty `EngineResults()` without any diagnostic.

## Fix

After the parsing loop, if zero results were produced, the engine now:

1. Inspects the DOM for known indicators of non-organic page types (direct answer widgets, knowledge panels, top stories carousels, weather widgets, calculators, "no results found" messages, spelling corrections)
2. Logs a **warning** with the page size, title, classified indicators, and a text snippet — so operators can diagnose what Google returned
3. Raises `SearxEngineResponseException` with a descriptive message — so the user sees an engine error instead of a blank result page

## Example diagnostic log output

```
WARNING:searx.engines.google:Google returned 5284 byte page with 0 parsed organic results. Title: "time - Google Search". Page indicators: direct_answer_widget. Snippet: "Zeit in Berlin, Deutschland Jetzt ..."
```

## Testing

1. Run a query that triggers a direct answer widget, e.g. `!google :de-DE time`
2. Before the fix: silent empty results, no error shown
3. After the fix: engine error "Google returned a page without organic search results. Indicators: direct_answer_widget. Check the SearXNG logs for details."
4. Other queries that normally return organic results should be unaffected

Closes #6171

### DIFF
--- a/searx/engines/google.py
+++ b/searx/engines/google.py
@@ -24,7 +24,7 @@ import babel.languages
 from lxml import html
 
 from searx.enginelib.traits import EngineTraits
-from searx.exceptions import SearxEngineCaptchaException
+from searx.exceptions import SearxEngineCaptchaException, SearxEngineResponseException
 from searx.locales import get_official_locales, language_tag, region_tag
 from searx.result_types import EngineResults
 from searx.utils import (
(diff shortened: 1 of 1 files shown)


Bnyro:
Closing because of low quality / slop, please read our [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst).

searx/engines/google.py
```diff
@@ -425,6 +425,80 @@ def response(resp: "SXNG_Response"):
...
+
+        # Look for Google's "no results found" message
+        _no_result_signals = [
+            'did not match any documents',
+            'keine Ergebnisse',
+            'no results',
+            'aucun r\u00e9sultat',
+            'nessun risultato',
+            'geen resultaten',
+            '\u6ca1\u6709\u7b26\u5408\u6761\u4ef6\u7684\u641c\u7d22\u7ed3\u679c',
```
Bnyro, line 450: What is this? Tf?

searx/engines/google.py
```diff
@@ -425,6 +425,80 @@ def response(resp: "SXNG_Response"):
...
+
+        _body_pg_len = len(resp.text)
+        _diag_indicators = ", ".join(_diag_parts) if _diag_parts else "none"
+        _diag_msg = (
+            f'Google returned {_body_pg_len} byte page with 0 parsed organic results. '
+            f'Title: "{_title_text}". '
+            f'Page indicators: {_diag_indicators}. '
+            f'Snippet: "{_page_text[:300].strip()}"'
+        )
+        logger.warning(_diag_msg)
```
Bnyro, line 491: This is way too error prone. Chances are very high that we raise an error here although there are some relevant results.

return42:
Please familiarize yourself with the rules:

- [Development Quickstart Guide](https://docs.searxng.org/dev/quickstart.html)
- [How-to contribute](https://github.com/searxng/searxng/blob/master/CONTRIBUTING.rst#how-to-contribute)
- [Git Commits & Change Management](https://docs.searxng.org/dev/commits.html)
- [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst)

PR #6337, opened 2026-06-29
fix: also remove dependent engines when a network engine is removed (#6336)

### LINKED ISSUE #6336: `use_default_settings.engines.remove` will crash down the service if only removing the base `network` engine but remaining the related engines
> But this is not an issue; `brave` network (created for the `brave` engine) is used by other engines e.g. brave.images:
> 
> However, this is an implicit dependency that wasn't mentioned in the documentation until I found this issue. So, if I want to remove Brave, my configuration should look like this:
> ```
> use_default_settings:
>   engines:
>     remove:
>       - brave
>       - brave.images
>       - brave.videos
>       - brave.news
> ```
> Maybe is it a good idea to add more details about it in the documentation?
> 
> Similar engines like: brave, lemmy communities, piped, presearch, qwant, yandex, yacy 

 _Originally posted by @reply2future in [#5356](https://github.com/searxng/searxng/issues/5356#issuecomment-4828637463)_

Image [settings.yaml documentation](https://docs.searxng.org/admin/settings/settings_engines.html)

The `network` is the inheritance, but the refer to the engine object. Maybe building the network object before removing engine?

muhamedfazalps:
Fixes #6336

When use_default_settings.engines.remove removes a network engine (e.g. brave), engines that reference that network (e.g. brave.images, brave.videos, brave.news) were left behind and would crash at runtime attempting to use the removed network.

The fix: after removing engines by name, also remove any engines whose network attribute matches a removed engine name.


tiekoetter (commented):
????

PR #6339, opened 2026-06-29
[fix] searchzee: circumvent botblocking

Bnyro:
### What does this PR do?
- fixes the engine as it now requires an API key ("search token")

- enable engine
- !sz test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/searchzee.py
+++ b/searx/engines/searchzee.py
@@ -1,0 +1,96 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""SearchZee is a small, indie project, web and news results pulled from
+independent search infrastructure."""
+
+import typing as t
+from urllib.parse import urlencode
+
+from searx.exceptions import SearxEngineAPIException
+from searx.extended_types import SXNG_Response
+from searx.network import get
+from searx.result_types import EngineResults
+from searx.utils import extr, html_to_text
+from searx.enginelib import EngineCache
+
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://searchzee.com",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+    "description": (
+        "SearchZee is a small, indie project, the web and news results"
+        " are pulled from an independent search infrastructure."
+    ),
+}
+categories: list[str] = None  # type: ignore[reportAssignmentType]
+
+paging = True
+
+SearchzeeCategType = t.Literal["web", "news"]
+searchzee_categ: SearchzeeCategType = None  # type: ignore[reportAssignmentType]
+
+
+CACHE: EngineCache
+"""Cache for storing the scraped API Token."""
+
+base_url = "https://searchzee.com"
+
+# only supports for news
+time_range_map = {"day": "pd", "week": "pw", "month": "pm", "year": "py"}
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    if searchzee_categ not in t.get_args(SearchzeeCategType):
+        raise ValueError("invalid category: %s" % searchzee_categ)
+
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])  # type: ignore[reportAny]
+    return True
+
+
+def _obtain_api_token() -> str:
+    token: str | None = CACHE.get("token")  # type: ignore[reportAny]
+    if token:
+        return token
+
+    token_resp = get(
+        f"{base_url}/app.js",
+    )
+    if not token_resp.ok:
+        raise SearxEngineAPIException("failed to obtain api key")
+
+    token = extr(token_resp.text, "const SEARCHZEE_API_TOKEN = \"", "\";")
+    CACHE.set("token", token, expire=3600)
+
+    return token
+
+
+def request(query: str, params: "OnlineParams"):
+    params["headers"]["X-SearchZee-Token"] = _obtain_api_token()
+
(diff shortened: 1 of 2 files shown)


searx/engines/searchzee.py
```diff
@@ -0,0 +1,78 @@
...
+
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://searchzee.com",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
```
return42, line 23:
```python
    "description": (
        "SearchZee is a small, indie project, web and news results"
        " pulled from independent search infrastructure."
        )
```

searx/engines/searchzee.py
```diff
@@ -0,0 +1,78 @@
...
+base_url = "https://searchzee.com"
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+    return True
+
+
+def _obtain_api_token() -> tuple[str, str, str]:
```
return42: The returned value is a `str` .. isn't it shown in your IDE (LSP)?

image

image

searx/engines/searchzee.py
```diff
@@ -0,0 +1,78 @@
...
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://searchzee.com",
+    "official_api_documentation": None,
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+categories = ["general"]
```
return42: Makes no sense to not migrate news also .. the risk "news" query will block the IP is given.

https://github.com/searxng/searxng/blob/bfeaad6c376c9dc1d402456699e03139bbac4514/searx/settings.yml#L3011-L3032

searx/engines/searchzee.py
```diff
@@ -0,0 +1,78 @@
...
+    args = {"q": query, "type": "web", "offset": params["pageno"] - 1}
+    params["url"] = f"{base_url}/api/search?{urlencode(args)}"
+    params["headers"]["X-SearchZee-Token"] = _obtain_api_token()
+
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    res = EngineResults()
+
+    results = resp.json()["results"]
+
```
return42, line 85: to narrow the type, you can add `results: list[dict[str,str]]` above the loop.

return42 (changes requested):
See my comments below ..

searx/engines/searchzee.py
Bnyro: Whoops, forget to do that.

searx/engines/searchzee.py
Bnyro: Somehow my Pyright setup works only sometimes - I don't know. Copied this from Chatnoir, thanks for the catch!

searx/engines/searchzee.py
Bnyro: Shouldn't this go into the docstring?

return42: See https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.EngineAbout.description

Will be shown in the UI ..

image

Bnyro: But then all of the engines should use this instead of the docstring? :)

Maybe it would be handy if SearXNG would automatically parse the docstring as a description.

return42: The descriptions are build by:

https://github.com/searxng/searxng/blob/d115c61a7049ce798b73713446dbcfc77bb911b7/searxng_extra/update/update_engine_descriptions.py#L219-L220

.. and  the `about.description` [should only be set as long as no description of the data source is available via a `wikidata_id`.](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.EngineAbout.description)

> But then all of the engines should use this instead of the docstring? :)

Docstrings are often unsuitable, as they contain content intended for admins and developers / not for the end user.

searx/engines/searchzee.py
```diff
@@ -0,0 +1,96 @@
...
+        "SearchZee is a small, indie project, the web and news results"
+        " are pulled from an independent search infrastructure."
+    ),
+}
+categories: list[str] = None  # type: ignore[reportAssignmentType]
+
+paging = True
+
+SearchzeeCategType = t.Literal["web", "news"]
+searchzee_categ: SearchzeeCategType = None  # type: ignore[reportAssignmentType]
```
return42, line 34: [If `None` is assigned to an option in the engine file, it has to be redefined in the settings, otherwise SearXNG will not start with that engine](https://docs.searxng.org/dev/engines/engine_overview.html#overrides)

searx/engines/searchzee.py
```diff
@@ -0,0 +1,96 @@
...
+time_range_map = {"day": "pd", "week": "pw", "month": "pm", "year": "py"}
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    if searchzee_categ not in t.get_args(SearchzeeCategType):
+        raise ValueError("invalid category: %s" % searchzee_categ)
+
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])  # type: ignore[reportAny]
+    return True
```
return42, line 52: FYI To return a bool should no longer be needed, when we rebase this branch.

return42 (approved):
I force pushed the nits I had, see [diff](https://github.com/searxng/searxng/compare/4c579dc5c837ba2a21e7a5e717c57dd068bc9d61..f3f821cb445b7e971a223d5ca63994705e1ec126) and the comments below.

PR #6340, opened 2026-06-29
[fix] heexy: circumvent botblocking

Bnyro:
### What does this PR do?
- fixes the engine as it now requires a request token for using the API

- enable engine
- !he test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/heexy.py
+++ b/searx/engines/heexy.py
@@ -13,9 +13,13 @@ It seems to use Bing internally, as the image thumbnails are loaded from Bing.
 from urllib.parse import urlencode
 
 import typing as t
+from lxml import html
 
-from searx.exceptions import SearxEngineAccessDeniedException
+from searx.enginelib import EngineCache
+from searx.network import get
+from searx.exceptions import SearxEngineAPIException, SearxEngineAccessDeniedException
 from searx.result_types import EngineResults
+from searx.utils import eval_xpath, extract_text, gen_useragent
 
 if t.TYPE_CHECKING:
     from searx.extended_types import SXNG_Response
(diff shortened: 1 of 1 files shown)


searx/engines/heexy.py
```diff
@@ -38,14 +42,44 @@
...
 
-def init(_):
+
+def init(engine_settings: dict[str, t.Any]) -> bool:
     if heexy_categ not in ("web", "image"):
         raise ValueError("invalid search category: %s" % heexy_categ)
 
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+    return True
```
return42: Since the code executed here in the engine initialization isn't time consuming, you can use the `setup` method (saves to create a thread on start up) .. related:

- #6343

searx/engines/heexy.py
```diff
@@ -38,14 +42,44 @@
...
+CACHE: EngineCache
+"""Cache for storing the ``X-Data-Cacheft`` token (acts like an API key)."""
 
-def init(_):
+
+def init(engine_settings: dict[str, t.Any]) -> bool:
     if heexy_categ not in ("web", "image"):
         raise ValueError("invalid search category: %s" % heexy_categ)
 
+    global CACHE  # pylint: disable=global-statement
```
return42: a global should be declared on top of the function.

return42 (changes requested):
See my comments below.

searx/engines/heexy.py
Bnyro: I read in one of the docstrings that slow things like database initialization should be done in `init`, probably misunderstood that, as `EngineCache` doesn't actually initialize any database (it re-uses the existing SQLite db)

PR #6341, opened 2026-06-29
[fix] qwant: access denied due to mismatching query order

Bnyro:
### What does this PR do?
- fixes the qwant engine by passing the `tgp` param at the correct position
-  tgp seems to be short for "test group" (look at the XHR requests the browser sends to "https://www.qwant.com/action/ui") - its actual value doesn't matter, as long as it's sent and at the correct position in the query params and doesn't change too frequently

- !qw test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -38,6 +38,7 @@ Implementations
 
 """
 
+import random
 import typing as t
 
 from datetime import (
@@ -89,6 +90,11 @@ qwant_categ: str = None  # pyright: ignore[reportAssignmentType]
 
 safesearch = True
 
+# tgp seems to be short for "test group" - its actual value doesn't matter, as
+# long as it's sent and at the correct position in the query params and doesn't
+# change too frequently
+test_group_value = random.randint(1, 3)
+
 # fmt: off
 qwant_news_locales = [
     "ca_ad", "ca_es", "ca_fr", "co_fr", "de_at", "de_ch", "de_de", "en_au",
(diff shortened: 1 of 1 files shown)


searx/engines/qwant.py
```diff
@@ -114,14 +120,17 @@ def request(query: str, params: "OnlineParams") -> None:
     results_per_page = 10
     if qwant_categ == "images":
         results_per_page = 50
+
+    # NOTE: the order here has to be exactly the same as on the website - in
+    # case the engine breaks, this should be the first thing to test
```
return42: I can't confirm that .. as soon as I search `!qwant foo` multiple times, I run into a block pretty quickly .. it seems to be limited to the FFox UA: in a real FFox session, the request then also gets blocked and redirected to a CAPTCHA challenge .. if I start a fresh Chrome session, no CAPTCHA appears and I get the results.

Even if I change the order or use different values, the situation doesn't get any better

return42 (commented):
If it is really an improvement for, than lets merge .. on my local IP it doesn't help / see comment below.

searx/engines/qwant.py
Bnyro: I always get access denied without this change, both from my server and local IP.

I haven't really tested how it affects ratelimiting, but it's definitely better to run into rate limits after some requests instead of getting blocked immediately.

I decided to shuffle the order of the query arguments now, because that makes it harder to fingerprint SearXNG requests.

PR #6343, opened 2026-06-30
[fix] clarify the mess of Engine.setup and Engine.init

return42:
Clarify the mess of Engine.setup() and Engine.init()

### Code of Conduct

  none AI used

### DIFF
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -269,21 +269,27 @@ def is_engine_active(engine: "Engine | types.ModuleType"):
 
 
 def call_engine_setup(engine: "Engine | types.ModuleType", engine_data: dict[str, t.Any]) -> bool:
-    setup_ok = False
+
+    setup_ok: bool | None = False
     setup_func = getattr(engine, "setup", None)
 
     if setup_func is None:
         setup_ok = True
     elif not callable(setup_func):
-        logger.error("engine's setup method isn't a callable (is of type: %s)", type(setup_func))
+        logger.error(f"engine's setup method isn't a callable (is of type: {type(setup_func)})")
     else:
         try:
             setup_ok = engine.setup(engine_data)
         except Exception as e:  # pylint: disable=broad-except
-            logger.exception('exception : {0}'.format(e))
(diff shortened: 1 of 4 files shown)


Bnyro (approved):

Bnyro:
I don't see why `init` and `setup` return boolean values? In my opinion, the method either runs successfully or throws an error (that's what currently all engines are doing), I haven't yet seen any engine that does `return False` in case it fails to initialize?

---

On a side note, I saw that a lot of engines use the `init` method for checking if the provided configuration (e.g. search type) is right and initializing the engine cache.

See https://github.com/search?q=repo%3Asearxng%2Fsearxng%20def%20init&type=code.

We should probably open a tracking issue for refactoring that? That also lowers the chances that somebody uses the wrong method due to copy and paste in the future.

return42:
> I don't see why `init` and `setup` return boolean values? In my opinion, the method either runs successfully or throws an error (that's what currently all engines are doing)

An exception is thrown when a state cannot be handled; however, it must also be possible for the engine to activate or deactivate itself for some reasons.

We don't have a robust example of this in our engines, but imagine an engine that is supposed to be active by default when a specific context exists .. and inactive when that context is absent .. without logging an error when the context is missing, since the lack of context does not necessarily constitute an exceptional situation.

> On a side note, I saw that a lot of engines use the `init` method for checking if the provided configuration (e.g. search type) is right and initializing the engine cache .. We should probably open a tracking issue for refactoring that? 

Yes, we should.

searx/engines/__init__.py
```diff
@@ -269,21 +269,27 @@ def is_engine_active(engine: "Engine | types.ModuleType"):
...
         try:
             setup_ok = engine.setup(engine_data)
         except Exception as e:  # pylint: disable=broad-except
-            logger.exception('exception : {0}'.format(e))
+            logger.exception(f"(PID {os.getpid()}) {engine.name}: engine SETUP failed, exception: {e}")
+            setup_ok = False
+
+    # The evaluation of the return value is analogous to Engine.init
+    if setup_ok is None:
+        setup_ok = True
```
return42, line 289: This is the (only) functional change in this PR; it aligns `init()` and `setup()` methods, and a `return True` is no longer required (in either methods).

PR #6349, opened 2026-07-02
[feat] engine: add paid yandex.com search api

ArsenBilov:
### What does this PR do?

Adds paid Yandex Search API engine.

Yandex Search API is paid and it needs API key to use it. But you could get free credits if you register new Yandex cloud account. Also I could give API Key for test purposes with expiration configured.

I tested my changes two ways: 
- locally via Docker and sent few search requests
- deployed into Kubernetes and sent another search requests
In both ways I got responses, that I expected for, and they are very similar to results that user gets when uses yandex.com in browser.

### Related issues

No opened issues related to Yandex Search API.

### Code of Conduct

Most of code was written by hand. After that I used Claude Opus 4.8 for code-reviewing based on official Yandex documentation and for translating commentaries, because I'm not so good in English. Opus gave me some good ideas to improve naming and suggested to refactor some code. I accepted it.


return42:
> Most of code was written by hand.

Nice work :+1: .. add a `yandex_api.rst` file to https://github.com/searxng/searxng/tree/master/docs/dev/engines/online and your doc-strings are in the online docs under https://docs.searxng.org/dev/engines/index.html#online-engines

You can test build of the docs locally: [`make docs.live`](https://docs.searxng.org/dev/contribution_guide.html#documentation)

ArsenBilov:
> > Most of code was written by hand.
> 
> Nice work 👍 .. add a `yandex_api.rst` file to https://github.com/searxng/searxng/tree/master/docs/dev/engines/online and your doc-strings are in the online docs under https://docs.searxng.org/dev/engines/index.html#online-engines
> 
> You can test build of the docs locally: [`make docs.live`](https://docs.searxng.org/dev/contribution_guide.html#documentation)

I added one more commit with rst-file and some logging in debug-mode. I tested docs on local PC, thanks for hint.

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+from searx.result_types import EngineResults
+from searx.utils import extract_text
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+about = {
+    "website": "https://yandex.cloud/en/services/search-api",
```
return42, line 51: this URL is redirected to a aistudio.yandex.ru domain .. can you clarify whats wrong with this link .. and all the other links given in this file?

ArsenBilov: I think domain "yandex.cloud" it's some kind of aggregation website, that redirects users to proper pages of Yandex Cloud Services sites with documentation, pricing and other information. But it's only my assumption. 
What is better: insert links to concrete pages, where info is described, or delete links, and SearxNG user should find information themself?

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+    req_l10n = l10n
+    lang = params["searxng_locale"].split("-")[0].lower()
+    if lang in language_map:
+        req_search_type, req_l10n = language_map[lang]
+
+    body: dict[str, t.Any] = {
+        "query": {
+            "searchType": req_search_type,
+            "queryText": query,
+            "familyMode": safesearch_map.get(params["safesearch"], "FAMILY_MODE_MODERATE"),
```
Bnyro:
```suggestion
            "familyMode": safesearch_map[params["safesearch"]],
```
The only possible values here are 0, 1 and 2, which you already cover in `safesearch_map`, so `FAMILY_MODE_MODERATE` is never used here.

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+            "page": str(params["pageno"] - 1),
+        },
+        "groupSpec": {
+            "groupMode": "GROUP_MODE_FLAT",
+            "groupsOnPage": str(page_size),
+            "docsInGroup": "1",
+        },
+        "l10n": req_l10n,
+        "folderId": folder_id,
+        "responseFormat": "FORMAT_XML",
```
Bnyro, line 155: Are there any other formats supported or is XML the only one? JSON would be preferable.

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+search_type: str = "SEARCH_TYPE_COM"
+"""Search domain / type.  One of ``SEARCH_TYPE_RU`` (yandex.ru),
+``SEARCH_TYPE_TR`` (yandex.com.tr), ``SEARCH_TYPE_KK`` (yandex.kz),
+``SEARCH_TYPE_BE`` (yandex.by), ``SEARCH_TYPE_UZ`` (yandex.uz) or
+``SEARCH_TYPE_COM`` (yandex.com, international)."""
+
+l10n: str = "LOCALIZATION_EN"
+"""Language of the search-result notifications.  One of ``LOCALIZATION_RU``,
+``LOCALIZATION_BE``, ``LOCALIZATION_KK``, ``LOCALIZATION_UK``,
+``LOCALIZATION_TR`` or ``LOCALIZATION_EN``."""
```
Bnyro: Couldn't these just be put into a single configuration option like
```py
yandex_default_language = "en"
```
You already defined which ones belong together, in `language_map`, so you could just do
```py
search_type, lang = language_map[yandex_default_language]
```

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+}
+
+
+def setup(_) -> bool:
+    """Validate credentials and paging limits when the engine is loaded."""
+    if not api_key or not folder_id:
+        raise SearxEngineAPIException("missing 'api_key' and/or 'folder_id' in engine settings")
+    if not 1 <= page_size <= 100:
+        raise SearxEngineAPIException("'page_size' must be in the range 1..100 (Yandex 'groupsOnPage')")
+
```
Bnyro, line 121: Please also add a check that the configured language is actually supported, i.e. is in `language_map`

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+    global max_page  # pylint: disable=global-statement
+    # ceiling division: the last (partial) page of the 250-result cap is valid
+    max_page = -(-250 // page_size)
```
Bnyro: Why `-(-...)`? Can't you just do `max_page = 250 // page_size`?

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+    return True
+
+
+def request(query: str, params: "OnlineParams"):
+    logger.debug("query: %r", query)  # pylint: disable=undefined-variable
```
Bnyro: please remove this :)

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+    data: dict[str, t.Any] = resp.json()
+
+    error = data.get("error")
+    if error:
+        message = error.get("message", error) if isinstance(error, dict) else error
+        raise SearxEngineAPIException(f"Yandex Search API error: {message}")
+
+    raw_data = data.get("rawData")
+    if raw_data is None:
+        raw_data = data.get("response", {}).get("rawData")
```
Bnyro: Please show examples or documentation where the JSON layout differs like that so that we need this case distinction here, seems strange.

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,239 @@
...
+# A query returns at most 250 results, so the number of reachable pages depends
+# on ``page_size`` (25 for the default of 10).  The exact value -- rounding the
+# last, partial page up -- is (re)computed in setup().
+max_page = 25
+
+# Credentials, overwritten via settings.yml
+api_key: str = ""
+"""Yandex Cloud API key, passed as ``Authorization: Api-Key <api_key>``."""
+
+folder_id: str = ""
```
Bnyro:
```suggestion
yandex_folder_id: str = ""
```

searx/engines/yandex_api.py
ArsenBilov: Regarding official documentation https://aistudio.yandex.ru/docs/en/search-api/api-ref/WebSearch/search.html , there is only 2 available formats: FORMAT_XML and FORMAT_HTML. I suggest using FORMAT_XML because it's easier to parse.

searx/engines/yandex_api.py
ArsenBilov: You are right. I occasionally mixed sync and async realisation of Yandex Websearch API. So my code was some kind of dead code. I rewrite this and commit new.

searx/engines/yandex_api.py
ArsenBilov: I agree, this was overengineered . We need to round up search results and I will use math.ceil instead.

ArsenBilov:
@return42 @Bnyro , could you please review my commits and approve auto-tests? I appreciate any feedback and looking forward to merge PR into master branch to use it in corporate environment.

searx/engines/yandex_api.py
Bnyro: Use a link that actually works please and points to seatch details, e.g. the one you shared in the other thread about the XML/HTML format.

ArsenBilov:
Hello, @Bnyro @return42 . Do you have any proposals to my PR which should be implemented before merging? Is there anything else I could improve in this PR to increase the chances of it being merged? I'd really appreciate any feedback or suggestions.

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,224 @@
...
+    categories: [general, web]
+    inactive: false
+    api_key: ""           # Yandex Cloud API key (``Api-Key``)
+    yandex_folder_id: ""  # Yandex Cloud folder id
+    # optional, see below:
+    yandex_default_language: en
+
+Implementations
+===============
+
```
Bnyro: This section is empty?
```suggestion
```

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,224 @@
...
+        raise SearxEngineAPIException("missing 'api_key' and/or 'yandex_folder_id' in engine settings")
+    if not 1 <= page_size <= 100:
+        raise SearxEngineAPIException("'page_size' must be in the range 1..100 (Yandex 'groupsOnPage')")
+    if yandex_default_language not in language_map:
+        raise SearxEngineAPIException(f"'yandex_default_language' must be one of {sorted(language_map)}")
+
+    global max_page  # pylint: disable=global-statement
+    # round up: the last (partial) page of the 250-result cap is still reachable
+    max_page = math.ceil(250 / page_size)
+    return True
```
Bnyro:
```suggestion
def setup(_):
    """Validate credentials and paging limits when the engine is loaded."""
    if not api_key or not yandex_folder_id:
        raise SearxEngineAPIException("missing 'api_key' and/or 'yandex_folder_id' in engine settings")
    if not 1 <= page_size <= 100:
        raise SearxEngineAPIException("'page_size' must be in the range 1..100 (Yandex 'groupsOnPage')")
    if yandex_default_language not in language_map:
        raise SearxEngineAPIException(f"'yandex_default_language' must be one of {sorted(language_map)}")

    global max_page  # pylint: disable=global-statement
    # round up: the last (partial) page of the 250-result cap is still reachable
    max_page = math.ceil(250 / page_size)
```

`setup` doesn't need to return a value (https://github.com/searxng/searxng/blob/1c3bb1e88f8781bd725c57a9a1c609ce0bb09926/searx/enginelib/__init__.py#L374), returning a value here is legacy / only used by old engines

searx/engines/yandex_api.py
```diff
@@ -0,0 +1,224 @@
...
+
+# Engine configuration
+categories = ["general", "web"]
+paging = True
+safesearch = True
+
+# A query returns at most 250 results, so the number of reachable pages depends
+# on ``page_size`` (25 for the default of 10).  The exact value -- rounding the
+# last, partial page up -- is (re)computed in setup().
+max_page = 25
```
Bnyro: Making `max_page` a global variable gives the impression that it can be configured via `settings.yml`.

Engine-internal variables should usually be named smth like `_MAX_PAGE` to indicate it's private and behaves like a constant (assuming it was initialized), 

In this case, it's better to, just move the computation of `max_page` into the `request` method.
 `max_page = math.ceil(250 / page_size)` is only 2 mathematical operations and an assignment, the performance benefit from only computing `max_page` once at startup is much less than the additional code complexity this introduces.

searx/settings.yml
```diff
@@ -2575,6 +2575,17 @@ engines:
...
+  - name: yandex api
+    engine: yandex_api
+    categories: [general, web]
+    shortcut: yda
+    inactive: true
+    # api_key: ""
+    # folder_id: ""
+    # search_type: SEARCH_TYPE_COM
+    # l10n: LOCALIZATION_EN
+    # timeout: 10.0
```
Bnyro: Is the engine really so slow that it takes 10 seconds? In that case, maybe we should mention that in the engine's docstring think? Or otherwise remove the timeout line here.

Bnyro (commented):
Sorry for the delay, seems like we forgot about the PR.

I've left some final suggestions/questions, after that we can merge it.

Bnyro (approved):

PR #6360, opened 2026-07-04
[fix] strip engine HTML + unescape entities before escape() in result snippets

wileykestner:
## Summary

Engines (Brave, Kagi, etc.) return snippets with `<strong>` keyword emphasis tags (Brave's API does this via `text_decorations=true`, the default). The current ordering runs `escape()` **first**, turning `<strong>` → `&lt;strong&gt;`. `highlight_content()`'s guard `if content.find("<") != -1: return content` is intended to leave already-HTML alone, but `escape()` has already consumed every literal `<`, so the guard is dead code on this path. The escaped tags then pass through, the template renders `title`/`content` with `|safe` (required so `highlight_content`'s trusted `<span class="highlight">` renders), and `&lt;strong&gt;` shows as visible literal text in results.

## Fix

Strip tags **before** `escape()` so engine markup doesn't leak as `&lt;strong&gt;` literal text, letting SearXNG's own highlight spans handle keyword emphasis cleanly. `html.unescape()` runs before stripping so that any entity-encoded markup in a snippet is normalized to real tags first and caught by the same strip.

## Test

Adds `test_search_html_strips_engine_markup_in_snippets` covering the fix: engine-supplied `<strong>`/`&lt;b&gt;` markup in title/content is stripped before `escape()` so it doesn't leak as literal `&amp;lt;strong&amp;gt;` text, while SearXNG's own highlight spans still apply to the query.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -12,6 +12,8 @@ from timeit import default_timer
 from html import escape
 from io import StringIO
 import typing
+import html
+import re
 
 import urllib
 import urllib.parse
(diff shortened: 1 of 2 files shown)


vojkovic:
You did not read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst):

> Issues and PR descriptions must be fully human-written. Do not post output from Large Language Models or similar generative AI as comments on any of our discussion forums (e.g. GitHub Issues, Matrix, ...), as such comments tend to be formulaic and low content. If you're a not a native English speaker, using AI for translating self-written issue texts to English is okay, but please keep the wording as close as possible to the original wording.

PR #6361, opened 2026-07-04
fix(kagi): guard against missing fields in partial API responses

wileykestner:
## What does this PR do?

Replaces unguarded `result["..."]` accesses in `kagi.response()` with `.get(key, default)` so a single partial result row no longer raises `KeyError` and flags the engine as a parsing error.

- search/news: `result["title"]` / `result["snippet"]` → `.get(..., "")`
- images: chained `result["image"]["width"]`/`["height"]` → guard the `image` dict and degrade `resolution` to `""` when it or the dims are missing; `img_src=image.get("url", "")` (a bare `image.get("url")` returns `None` when `image` is absent, which then `TypeError`s in `Image.__post_init__` → `mimetypes.guess_type(None)`)
- videos: `result["title"]` / `result["snippet"]` → `.get(..., "")` and `result["props"]` → `result.get("props", {})`

`url` remains unguarded — a result without a URL is unusable regardless, and Kagi always includes it.

Kagi's API can omit `snippet`, `title`, `image`, or `props` on some results. With unguarded accesses, a single partial row raised `KeyError` and the engine was reported as `unresponsive_engines: [['kagi', 'parsing error']]` — results collected before the failing row survived (a query still returned ~20 results), but the error fired every time a partial row appeared. The observed production hit was `KeyError: 'snippet'` at `kagi.py:152`.

Behavior is identical for well-formed responses; the only change is graceful handling of partial responses.

\`\`\`bash
.venv/bin/python -m nose2 -v tests.unit.test_engine_kagi
\`\`\`

Verified fail-before / pass-after:

- **upstream master** + the new test file dropped in: **8 errors** (`KeyError: 'snippet'` / `'title'` / `'image'`, `TypeError: NoneType` from `mimetypes.guess_type`, missing `props`)
- **this branch**: **11/11 ok**

## Author's checklist

- [x] Defensive parsing covers search/news/images/videos branches
- [x] `img_src` default prevents the `None`→`TypeError` path in `Image.__post_init__`
- [x] `url` intentionally left unguarded (truly required field)
- [x] Tests added where no kagi engine test existed upstream
- [x] Well-formed response behavior unchanged

### DIFF
--- a/searx/engines/kagi.py
+++ b/searx/engines/kagi.py
@@ -148,37 +148,42 @@ def response(resp: "SXNG_Response") -> EngineResults:
             res.add(
                 res.types.MainResult(
                     url=result["url"],
-                    title=html.unescape(result["title"]),
-                    content=html.unescape(result["snippet"]),
+                    title=html.unescape(result.get("title", "")),
+                    content=html.unescape(result.get("snippet", "")),
                     thumbnail=result.get("image", {}).get("url") or "",
                     publishedDate=published_date,
                 )
             )
         elif kagi_categ == "images":
+            image = result.get("image", {})
+            width = image.get("width")
+            height = image.get("height")
+            resolution = f"{width}x{height}" if width and height else ""
             res.add(
                 res.types.Image(
                     url=result["url"],
-                    title=html.unescape(result.get("title")),
-                    img_src=result.get("image", {}).get("url"),
-                    resolution=f"{result['image']['width']}x{result['image']['height']}",
+                    title=html.unescape(result.get("title", "")),
+                    img_src=image.get("url", ""),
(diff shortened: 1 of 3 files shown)


vojkovic:
You did not read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst):

> Issues and PR descriptions must be fully human-written. Do not post output from Large Language Models or similar generative AI as comments on any of our discussion forums (e.g. GitHub Issues, Matrix, ...), as such comments tend to be formulaic and low content. If you're a not a native English speaker, using AI for translating self-written issue texts to English is okay, but please keep the wording as close as possible to the original wording.

Issue #6359, opened 2026-07-03
GSA for iPhone useragent do no longer work


emirhansrchggl-cmd:
Google released a new update on July 1st. Now, when I perform a search using any keyword while logged out, the request returns an 'enable JS' error. Before July 1st, we could bypass this issue by adding 'NSTNWV' to the User-Agent string, but that method no longer works. Is there anyone who can offer a solution for this?

-----

Goolge's GSA for iPhone useragent do no longer work ..

- https://github.com/searxng/searxng/pull/5644

Bnyro:
If we don't manage to fix the HTML version of Google, we could implement a Google CSE engine and take the API tokens from different search engines that connect to Google CSE client-side. There are plenty of them, e.g. https://gibiru.com/, https://alohafind.com or https://www.blackle.com. Example URL: `https://cse.google.com/cse/element/v1?rsz=filtered_cse&num=10&hl=en&source=gcsc&cselibv=&cx=partner-pub-8993703457585266%3A4862972284&q=foobar&safe=off&cse_tok=AHy-FZNaFQvsl484LEnd3eeELjaY%3A1783088090130&exp=&callback=google.search.cse.api3588&rurl=`.

I'm not sure how long these CSE tokens are valid, so we'd need to see if we have to automatically scrape new ones or can just hardcode them.

This wouldn't last forever though, I think I've read somewhere that Google will be shutting down CSE some day.

~PS: I'll implement a CSE engine for SearXNG in the evening today or tomorrow, not sure when I will find the time.~ I'll let vojkovic do that :)

inetol:
Is it really worth the effort at this point? Following Bing, Google has also announced that it will be shutting down its csearch endpoint API, replaced by some *Vertex AI Search* bs https://developers.google.com/custom-search/v1/overview .

vojkovic:
Today I have been experimenting with different ways to get the HTML Google engine working. Surprisingly, the engine works like before just with the addition of a cookie from a signed in account.

I wouldn't be surprised if it's just a matter of getting the correct UA to get it working again.

> This wouldn't last forever though, I think I've read somewhere that Google will be shutting down CSE some day.

From what I've seen on-line, the CSE engines will work until the start of next year - so it should be an acceptable stopgap.

> There are plenty of them, e.g. https://gibiru.com/, https://alohafind.com/ or https://www.blackle.com/

Thanks for mentioning those sites - I used these CSE cx tokens:
```
partner-pub-8993703457585266:4862972284
[...]

PR #6364, opened 2026-07-04
[feat] engines: add google cse engine

vojkovic:
### What does this PR do?

Adds a new google cse engine as an alternative to the currently broken html google engine. It works by generating a token which is cached for 1 hour and it only uses the public CSE id from blackle.com. We do have two others but I think just hardcoding one will be enough.

Will be broken in 2027 due to google deprecating it but I it will be good while it lasts.

I found I can get up to 20 results per page. Setting safesearch and language works and I just copied the traits from the google engine. 

Also set old google engine as inactive as it is completely broken for now.

`make run` and do `!goc test`

Also live on my public instance

### Related issues

- https://github.com/searxng/searxng/issues/6359

### Code of Conduct

Only cursor for some parts and for suggestions here and there


searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+    return True
+
+
+def _cse_token() -> dict[str, str]:
+    cx = "partner-pub-8993703457585266:4862972284"  # blackle.com
```
Bnyro: I don't like hardcoding it this way, we should make a global
```py
cse_tokens: list[str] = ["partner-pub-8993703457585266:4862972284"]
```
and possibly also add the other providers we can find, so that we can just always pick a random one per request or at engine startup.

Additionally, that also allows instance admins to set different tokens, e.g.
```yaml
engines:
  - name: google cse
    cse_tokens:
      - "custom_token_123"

File: searx/engines/google_cse.py [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+    if token:
+        return token
+
+    resp = get(f"https://www.google.com/cse/cse.js?cx={cx}", timeout=10)
+    if not resp.ok:
+        raise SearxEngineAPIException("failed to obtain cse token")
+
+    end = resp.text.rfind("});")
+    start = resp.text.rfind("({")
+    opts = loads(resp.text[start + 1 : end + 1])
```
Bnyro: We have a helper method for that in `searx.utils`
```suggestion
    raw_json_content = extr(resp.text, "({", "});")
    opts = loads("{" + raw_json_content + "}")
```

searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+
+    cse_tok = opts.get("cse_token")
+    if not cse_tok:
+        raise SearxEngineAPIException("failed to obtain cse token")
+
+    exp = opts.get("exp") or []
```
Bnyro:
```suggestion
    exp = opts.get("exp")
```
This `or` is useless because doing `",".join([])` evaluates to `""`, which is covered by the else case in l66 anyways :)

searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+    token = {
+        "cse_tok": cse_tok,
+        "cselibv": opts.get("cselibVersion", ""),
+        "exp": ",".join(exp) if isinstance(exp, list) else "",
```
Bnyro: Out of interest, did you have any cases where `exp` has been something other than a `list` or `None`?

I'm just wondering why we don't do `",".join(exp) if exp else ""` here

searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+    google_info = get_google_info(params, traits)
+    info = google_info["params"]
+
+    hl = info["hl"]
+    if params.get("searxng_locale", "all") == "all" or "ZZ" in hl:
+        hl = "en"
+
+    args = {
+        "rsz": "filtered_cse",
+        "num": "20",
```
Bnyro: Please move this into a global variable `page_size = 20` 
```suggestion
        "num": str(page_size),
```
Then you can also use that in the page number calculation in l103

searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+        args["start"] = str(start)
+
+    params["url"] = "https://cse.google.com/cse/element/v1?" + urlencode(args)
+    params["cookies"] = google_info["cookies"]
+    params["headers"].update(google_info["headers"])
+    params["headers"]["Referer"] = "https://cse.google.com/"
+
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    raise_for_httperror(resp)
```
Bnyro: That's a No-Op, because by default `params["raise_for_httperror"]` is `True`, which automatically lets the engine crash on a non-200 http code before ever calling `response(resp)`

searx/engines/google_cse.py
```diff
@@ -0,0 +1,137 @@
...
+
+    json_resp = resp.text[resp.text.find("{") : resp.text.rfind("}") + 1]
+    data = loads(json_resp)
+
+    if error := data.get("error"):
+        message = error.get("message", "unknown error")
+        if error.get("code") == 429:
+            raise SearxEngineTooManyRequestsException(message=f"google cse: {message}")
+        raise SearxEngineAPIException(f"google cse: {message}")
```
Bnyro, line 143: This will never get called unless you set `params["raise_for_httperror"] = False` in the `request` method.

Bnyro (commented):
Looks very cool, thanks for looking into it!

I left some comments above.

searx/engines/google_cse.py
vojkovic: > we should make a global

Sure

> also add the other providers we can find, so that we can just always pick a random one per request or at engine startup.

Respectfully I disagree - at first I tried that, but then I noticed one of them had some weird bias toward reddit and other sites which I think can be configured. Plus I am not noticing any rate limiting based on the token.

Re. instance admins selecting their own tokens I am not sure I see where the benefit of that would be, if you could kindly explain. :)

searx/engines/google_cse.py
vojkovic: I was swapping things in and out of there earlier - will fix :)

searx/engines/google_cse.py
vojkovic: 👍

Bnyro:
We can also add time range support relatively easy by e.g. setting `args["sort"] = "date:r:20110101:20131231"` (source: https://stackoverflow.com/questions/47665573/specifying-a-date-range-in-google-custom-search-api).

So we could do something like
```diff
diff --git a/searx/engines/google_cse.py b/searx/engines/google_cse.py
index d97b4fe66..d7549e582 100644
--- a/searx/engines/google_cse.py
+++ b/searx/engines/google_cse.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Google Custom Search Engine"""
 
+import datetime
 import typing as t
 from json import loads
 from urllib.parse import urlencode
@@ -29,6 +30,7 @@ about = {
 categories = ["general", "web"]
 paging = True
 max_page = 5
+time_range_support = True
 language_support = True
 safesearch = True
 
@@ -69,6 +71,15 @@ def _cse_token() -> dict[str, str]:
     return token
 
 
+def _get_start_and_end_date_str(time_range: str) -> tuple[str, str]:
+    time_range_map = {"day": 1, "week": 7, "month": 30, "year": 365}
+
+    end_date = datetime.datetime.now()
+    start_date = end_date - datetime.timedelta(days=time_range_map[time_range])
+
+    return start_date.strftime("%Y%m%d"), end_date.strftime("%Y%m%d")
+
+
 def request(query: str, params: "OnlineParams") -> None:
     token = _cse_token()
 
@@ -91,6 +102,10 @@ def request(query: str, params: "OnlineParams") -> None:
         "callback": "_",
         "rurl": "",
     }
+    if params["time_range"]:
+        start_date, end_date = _get_start_and_end_date_str(params["time_range"])
+        args["sort"] = f"date:r:{start_date}:{end_date}"
+
     if info.get("lr"):
         args["lr"] = info["lr"]
     if info.get("cr"):

```

searx/engines/google_cse.py
vojkovic: Yes you are right, thats a leftover from when I had it set to false. 👍 will fix

searx/engines/google_cse.py
vojkovic: Same here 👍

searx/engines/google_cse.py
Bnyro: First, we definitely make this a global variable instead of hardcoding the same partner token twice :)

> Re. instance admins selecting their own tokens I am not sure I see where the benefit of that would be, if you could kindly explain. :)
>
Second, I think this would help in case of
- a ratelimit when many people use it at the same time (we haven't tested that yet), so instance admins can act quickly
- gentleness: I don't know if the people hosting e.g. blackle.com are charged per request, but if so, it's more fair to distribute it among different Google CSE engines.

I agree that we shouldn't include tokens that have high biases, but I can probably find another dozen Google CSE engines within an hour of research (just clicking through https://stolenhistory.net/threads/alternative-search-engines-video-sites-to-google-youtube.3771/)

searx/sxng_locales.py
```diff
@@ -15,8 +15,10 @@
     ('bg', 'Български', '', 'Bulgarian', '\U0001f310'),
     ('bg-BG', 'Български', 'България', 'Bulgarian', '\U0001f1e7\U0001f1ec'),
     ('ca', 'Català', '', 'Catalan', '\U0001f310'),
+    ('ca-ES', 'Català', 'Espanya', 'Catalan', '\U0001f1ea\U0001f1f8'),
     ('cs', 'Čeština', '', 'Czech', '\U0001f310'),
     ('cs-CZ', 'Čeština', 'Česko', 'Czech', '\U0001f1e8\U0001f1ff'),
+    ('cy', 'Cymraeg', '', 'Welsh', '\U0001f310'),
```
return42: To avoid new entries, that comes with a new engine, you can increment the min values here:

https://github.com/searxng/searxng/blob/747cec4c2304fc2e06ceb4a46dbe25bfb769db1e/searxng_extra/update/update_engine_traits.py#L134-L135

searx/engines/google_cse.py
vojkovic: cse.js has 3 {{ blocks so this won't work. The actual config we want is the last one in the sequence.

searx/engines/google_cse.py
vojkovic: Nah I didn't - will just go simple with yours

searx/engines/google_cse.py
Bnyro: Oh, sorry.

In this case we could either add a `from_end: bool` parameter to `searx.utils.extr` or add a new method `searx.utils.rextr` or something

searx/engines/google_cse.py
vojkovic: > First, we definitely make this a global variable instead of hardcoding the same partner token twice :)

Yes 100% - done

> a ratelimit when many people use it at the same time

FYI 4get uses just one token too and I have it running and already pushed 50,000 request through it, no issues.

> I don't know if the people hosting e.g. blackle.com are charged per request, but if so, it's more fair to distribute it among different Google CSE engines.

I looked into this and basically the cx id is public and is only to tell google what search engine configuration to use, it doesn't tie into any billing or anything thankfully this is a non issue. (Although I would completely agree if this wasn't the case)

That website is a gold mine btw.

Bnyro: Yeah, that's where I got the inspiration for most of the recently added search engines from :)

searx/engines/google_cse.py
vojkovic: Yes we could, but is there an issue using rfind? 😄

Bnyro: Yes, it's evil :shaking_face: 

To be honest, it's just less error-prone if you have a pre-built `extr` method that you can use, because you don't have to implement the "complicated" index logic yourself.

However, in this case, the current approach is probably fine, since there are only few other engines that also require to search from the end instead of beginning.

Bnyro (approved):

Bnyro:
You should probably squash time range support into the first commit, otherwise lgtm :+1:

searx/sxng_locales.py
vojkovic: Ach so :)

Okay I think I did that right, but how do you do the engine_descriptions.json?

`data.all` is crashing for me :(

myshadowchat:
Is it possible to see new SERP results that were indexed 1 or 2 hours ago in google.com search using Google CSE?

return42:
> You should probably squash time range support into the first commit, otherwise lgtm 👍

I will squash the commits in my review .. I'm working on atm ..

return42:
> Is it possible to see new SERP results that were indexed 1 or 2 hours ago in google.com search using Google CSE?

TBH, online marketing, SEO and all kind of this shitty click rate optimizing is not in our focus.

Here we are reviewing PRs, testing new stuff .. you are welcome to contribute your experience with this PR .. but please don't annoy us with such SEO shit!

searx/engines/google_cse.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    token = _cse_token()
+
+    google_info = get_google_info(params, traits)
+    info = google_info["params"]
+
+    hl = info["hl"]
+    if params.get("searxng_locale", "all") == "all" or "ZZ" in hl:
```
return42, line 93: Oops, this is an issue in the `get_google_info(..)` .. I will fix the function in my review.

searx/engines/google.py
```diff
@@ -197,7 +197,7 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     #   https://developers.google.com/custom-search/docs/xml_results_appendices#interfaceLanguages
 
     # https://github.com/searxng/searxng/issues/2515#issuecomment-1607150817
-    ret_val["params"]["hl"] = f"{lang_code}-{country}"
+    ret_val["params"]["hl"] = f"{lang_code}"
```
return42, line 200: I fixed the `ZZ` issue here / BTW: the default lang_code is already `en`:

https://github.com/searxng/searxng/blob/747cec4c2304fc2e06ceb4a46dbe25bfb769db1e/searx/engines/google.py#L173-L174

searx/engines/google.py
```diff
@@ -268,7 +268,11 @@ def get_google_info(params: "OnlineParams", eng_traits: EngineTraits) -> dict[st
     # HTTP headers
 
     ret_val["headers"]["Accept"] = "*/*"
-    ret_val["headers"]["User-Agent"] = gen_gsa_useragent()
+    # GSA for iPhone useragent had been added in
+    # - https://github.com/searxng/searxng/pull/5644
+    # but doe no longer work
+    # - https://github.com/searxng/searxng/issues/6359
+    # ret_val["headers"]["User-Agent"] = gen_gsa_useragent()
```
return42, line 275: We should remove the GSA code #5644 in a follow up PR.

----

PR is on the way ..

- https://github.com/searxng/searxng/pull/6366

searx/engines/google_cse.py
```diff
@@ -0,0 +1,158 @@
...
+    for item in data.get("results", []):
+        url = item.get("unescapedUrl")
+        if not url:
+            continue
+        results.add(
+            results.types.MainResult(
+                url=url,
+                title=item.get("titleNoFormatting", ""),
+                content=item.get("contentNoFormatting", ""),
+                thumbnail=item.get("richSnippet", {}).get("cseThumbnail", {}).get("src", ""),  # type: ignore
```
return42, line 155: I added the thumbnail .. is it OK?

searxng_extra/update/update_engine_traits.py
```diff
@@ -131,8 +131,8 @@ def log(msg: str):
 def filter_locales(traits_map: EngineTraitsMap) -> set[str]:
     """Filter language & region tags by a threshold."""
 
-    min_eng_per_region = 18
-    min_eng_per_lang = 23
+    min_eng_per_region = 19
+    min_eng_per_lang = 24
```
return42, line 135: To avoid new entries I increased the threshold.

return42 (approved):
Perfect work .. and a perfect engine, the best google engine we ever had :-)

I just added some nits [[diff]](https://github.com/searxng/searxng/compare/1f24e4cb7f628f167f8097a279dc0b3ff1a87f31..96613c512f87987b313cab6f21c3e397d90a9d4e), mainly type hints to calm down my LSP.

Have a look at my comments below, if you are fine with, lets merge 👍

searx/engines/google_cse.py
```diff
@@ -0,0 +1,158 @@
...
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://www.google.com",
+    "wikidata_id": "Q2233943",
+    "official_api_documentation": "https://developers.google.com/custom-search/docs/element",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSONP",
+    "description": "Platform for creating custom search engines based on Google Search.",
```
return42, line 28: Since there are no translations available [Q2233943](https://www.wikidata.org/wiki/Q2233943) I decided to hard code it here.

While you asked, to fetch description in a development cycle:

```
$ ./manage dev.env
(dev.env)$ python ./searxng_extra/update/update_engine_descriptions.py ^C
```

But this is nothing we do in a development cycle: this script will fire *thousands* of requests to fetch description in all languages from various resources the wikidata (ID) is just one resource .. the script takes a very long time (~20min) and will modify the entire `searx/data/engine_descriptions.json` file .. not really what we want to have in a PR.

We have a CI action for this, the last run was in:

- https://github.com/searxng/searxng/pull/6335

However, the quality depends heavily on how quickly the script is blocked... I've already tried to improve the script a bit:

- https://github.com/searxng/searxng/pull/6309

But the quality keeps getting worse because more and more resources are starting to block this script (which is why I didn't merge #6335).

Right now, even Wikidata requests are being blocked after the first 20 or 30 requests... that would be an area where we could make improvements centrally... but unfortunately, I don't have time for that :-o

vojkovic: Righto, hardcoding it will be fine for now although I am pretty sure there are translations available --> 
image

Hopefully it's improved when we move to self hosted runners - just an idea but we can try rotate v6 outbound addresses to scrape wikidata and to get around the blocking.

searxng_extra/update/update_engine_traits.py
vojkovic: Thanks, I see what you meant now

searx/engines/google_cse.py
vojkovic: Yep it's okay, works with image proxy and no errors.

Great addition 👍

searx/engines/google.py
vojkovic: 👍

searx/settings.yml
```diff
@@ -1199,6 +1199,11 @@ engines:
   - name: google
     engine: google
     shortcut: go
+    inactive: true
```
vojkovic, line 1202: @Bnyro @return42 

Please be aware I set the google engine to inactive.

I am not sure if it will cause issues. I did notice a substantial increase in 500 errors after swapping which I believe can be attributed to this.

Image

Can you please advise on if you think this is worth changing?

return42: Can you see the 500 responses in the log?  Do you see a traceback?

I tried, but can't reproduce a 500.

Bnyro: I personally would be fine with bots spamming `!google <query>` and getting no results.

But if we set the engine to inactive, it wouldn't be possible to use it all anymore, so the 500 errors can't come from that? If you do e.g. `!google foo` and the `google` engine is inactive, it just searches the full query with all available search engines, e.g. searches `!google foo` with duckduckgo, startpage, ...

vojkovic: Oops, I found the issue with the 500x error was just a small issue with my instance.

But separately, searx.space is reporting an error when google is inactive, I'll send in a PR to change it to the cse engine temporarily...

Screenshot

Anyhow everything else is very good; thanks for your work - LGTM

return42:
FYI: last force push was just a rewording of the commit messages .. I added PR number to the first line of the commit message ..

grafik

searx/engines/google_cse.py
```diff
@@ -0,0 +1,158 @@
...
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    json_resp = resp.text[resp.text.find("{") : resp.text.rfind("}") + 1]
+    data = loads(json_resp)
+
+    # not the real types, but a sufficient approximation
+    item: dict[str, str]
+    error: dict[str, str | int]
+
+    if error := data.get("error"):
```
return42, line 139: BTW: thanks for teaching me .. `:=` is an operator I missed since a long time .. haven't realized it was added in Python 3.8 :-)

Bnyro: Same, I forgot about the operator immediately after I read the release highlights...

PR #6367, opened 2026-07-05
[fix] results: display internationalized domain names human-friendly

Bnyro:
### What does this PR do?

Some search engines encode domains that contain special characters in the IDN format, i.e. by using Punycode (https://en.wikipedia.org/wiki/Punycode).

This means that domains formatted like that are not human readable and thus very unintuive.

Also, as only some engines use Punycode, this often causes a result to appear multiple times, once with a Punycode domain (e.g. xn--allestrungen-9ib.de), and once with a normal domain (e.g. allestörungen.de).

Always formatting the domains nicely has the caveat that the official sites are harder to distinguish from malicious clones that just swapped out some characters by using Punycode, e.g. replaced `a` with `á`. I don't think that will be much of an issue though - I don't think Punycode results previously stopped users from clicking the link, and I also think that most search engines filter out such bad results or don't even have them indexed.

## How to test the changes?
- Enable a bunch of engines and search for `alle störungen`

Before this PR, some results are duplicated
image
image

After this PR, they're merged into a single result:
image

### Code of Conduct

No AI used.

### DIFF
--- a/searx/result_types/_base.py
+++ b/searx/result_types/_base.py
@@ -52,6 +52,12 @@ def _normalize_url_fields(result: "Result | LegacyResult"):
             result.parsed_url = urllib.parse.urlparse(result.url)
 
     if result.parsed_url:
+        # properly format special characters (e.g. "ä", "ö") in IDN domains
+        # e.g. "xn--strung-xxa.de" becomes "störung.de"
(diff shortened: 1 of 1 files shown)


return42 (approved):
LGTM .. I wasn't aware of *Punycode* before .. thanks a lot 👍

PR #6368, opened 2026-07-05
[mod] show img_load_error.svg if thumbnail fails to load

return42:
Brave and a few other engines occasionally provide dead URLs for the thumbnails, which looks pretty ugly in the UI ..

grafik

As already with the image search, in case of an error `img_load_error.svg` should be displayed here.

Look out for bad thumbnail URLs or patch like shown here:

```diff
diff --git a/searx/engines/google_cse.py b/searx/engines/google_cse.py
index a0261e324..b823a1f2e 100644
--- a/searx/engines/google_cse.py
+++ b/searx/engines/google_cse.py
@@ -152,7 +152,7 @@ def response(resp: "SXNG_Response") -> EngineResults:
                 url=url,
                 title=item.get("titleNoFormatting", ""),
                 content=item.get("contentNoFormatting", ""),
-                thumbnail=item.get("richSnippet", {}).get("cseThumbnail", {}).get("src", ""),  # type: ignore
+                thumbnail="https://not-exist.io/foo.png"
             )
         )
     return results
```

.. and query `!goc foo`

grafik

### Code of Conduct

  no AI used


vojkovic:
Wouldn't it be better to just not show any image at all? Or will it cause a layout shift?

Bnyro (commented):
Wouldn't it be better to hide the thumbnail instead of showing a placeholder?

I don't really see much value in the placeholder in this case, because there are many other general results that also don't have any thumbnail and therefore just don't show any image.

return42:
> Wouldn't it be better to just not show any image at all? Or will it cause a layout shift?

If we remove, we should remove the parent tag, then we should not have a layout shift, however there will be a *jitter* in the layout when the tag is removed.

One other reason not to remove; if the XPath selector for a thumbnail is no longer valid (needs an update), the thumbnail will silently disappear .. and no one will report this bug.

The problem is actually caused by unreliable engines like Brave... perhaps we should remove the thumbnail there.

These unreliable URLs also keep reappearing when the results from the engines are merged... and then Brave can overwrite the actually correct thumbnail from, for example, Google.

inetol (approved):
Pretty fix, LGTM

PR #6369, opened 2026-07-05
[feat] google cse engine: add support for images

Bnyro:
### What does this PR do?
- add support for Google CSE images
- the normal google images engine doesn't work for me anymore, so this is a good workaround for now

- !goci test

### Related issues
- #6364 

### Code of Conduct

No AI used.


vojkovic (approved):

return42 (approved):
To avoid indentation in if branches, I introduced helper functions and I also added another commit fixing a leftover on top.

Since the old engines are dead, should we use their shortcuts for the new google engines?

Bnyro:
I don't have any preference on that.

But that should probably be done in an other PR anyways.

Thanks for the reviews 👍

PR #6370, opened 2026-07-05
[feat] engines: add avalw.org (general)

Bnyro:
### What does this PR do?
Add avalv.org as a json engine.

The index is very small, so it often doesn't find anything.

It also has news support, but the results are unusable.

- !av rust

### Code of Conduct

No AI used.


return42 (approved):
LGTM

PR #6394, opened 2026-07-10
[refactor] engines: use datetime.fromisoformat instead of strptime

### LINKED ISSUE #6098: Use `datetime.fromisoformat` instead of `datetime.strptime`

OneVth:
### What does this PR do?

Replace `strptime` calls that parse ISO 8601 dates with `fromisoformat`.

Changes:

* Engines using RFC 2822 or other formats that are not compatible with ISO 8601 (e.g. bt4g, google_scholar) are left unchanged.
* In most files this is a direct replacement of `strptime(text, "format")` with `fromisoformat(text)`.
* For code that treats `Z` as a literal (e.g. `"%Y-%m-%dT%H:%M:%S.%fZ"` in huggingface.py), `rstrip("Z")` is added to keep the output naive and preserve the existing behavior.
* In sogou.py the date is extracted with a regular expression before `strptime`, so strings like `"2026-7-11"` can also be converted to a datetime. `fromisoformat` does not accept this format, so the `datetime` is built directly instead.

Run the test suite:

* `make test.unit` , all 339 tests pass
* `make format.python`
* `make test.pyright` , no new errors compared to master

After the tests, I also ran the instance locally with `make run` and manually confirmed that the date rendering of several affected engines is identical before and after this PR. For example, `!hf llama` and `!arx machine learning`:

|Before this PR|After this PR|
|---|---|
|image|image|
|image|image|

As shown above, the date rendering does not change before and after this PR.

### Related issues

Closes: #6098

### Code of Conduct

I used Claude Code to understand the issue, analyze the codebase, and figure out which files needed to be changed. All other code changes were made by me.

Since English is not my native language, I used Claude Desktop to help with the commit message and this PR. I wrote the commit message and PR in my native language (Korean), and asked Claude Desktop to translate them into English and suggest suitable wording for the context.

### DIFF
--- a/searx/engines/sogou.py
+++ b/searx/engines/sogou.py
@@ -95,7 +95,8 @@ def _parse_date(text):
         date_match = re.search(r"(\d{4}-\d{1,2}-\d{1,2})", text)
         if date_match:
             try:
-                return datetime.strptime(date_match.group(1), "%Y-%m-%d")
+                y, m, d = date_match.group(1).split("-")
+                return datetime.fromisoformat(f"{y}-{int(m):02d}-{int(d):02d}")
             except (ValueError, TypeError):
                 pass
     return None
--- a/searx/engines/sogou.py
+++ b/searx/engines/sogou.py
@@ -96,7 +96,7 @@ def _parse_date(text):
         if date_match:
             try:
                 y, m, d = date_match.group(1).split("-")
-                return datetime.fromisoformat(f"{y}-{int(m):02d}-{int(d):02d}")
+                return datetime(year=int(y), month=int(m), day=int(d))
             except (ValueError, TypeError):
                 pass
     return None
--- a/searx/engines/acfun.py
+++ b/searx/engines/acfun.py
@@ -83,7 +83,7 @@ def extract_video_data(video_block):
         published_date = None
         if create_time:
             try:
-                published_date = datetime.strptime(create_time.strip(), "%Y-%m-%d")
+                published_date = datetime.fromisoformat(create_time.strip())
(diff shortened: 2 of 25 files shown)


searx/engines/sogou.py
```diff
@@ -95,7 +95,8 @@ def _parse_date(text):
         date_match = re.search(r"(\d{4}-\d{1,2}-\d{1,2})", text)
         if date_match:
             try:
-                return datetime.strptime(date_match.group(1), "%Y-%m-%d")
+                y, m, d = date_match.group(1).split("-")
+                return datetime.fromisoformat(f"{y}-{int(m):02d}-{int(d):02d}")
```
Bnyro:
```suggestion
                return datetime.datetime(year=y, month=m, day=d)
```
It doesn't make any sense to go the extra step from `date (y,m,d) -> string -> datetime`, we can immediately build a `datetime` from `y, m, d` :)

Bnyro (commented):
Looks good overall at first glance, thanks for looking into this 👍 

Left one comment above.

searx/engines/sogou.py
OneVth: Thanks! I changed it to build the datetime directly as you suggested.

I added `int()` since the date parts from the regex are strings.

Bnyro (approved):
Thanks!

PR #6405, opened 2026-07-13
[fix] qwant: engine blocked with captcha

Bnyro:
Qwant now requires a `datadome` cookie that it returns in the first search response as a `Set-Cookie`.

This cookie has to be sent for all requests, otherwise they will be blocked.

This means that now, the first search request is blocked (results in CAPTCHA), and only the subsequent searches work (same happens on the Qwant website for me).

However, I don't think it's worth repeating the same search request multiple times very quickly because that also makes us more suspicious.

- search twice: `!qw test`, `!qw test`

### Related issues
- #3929 

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/qwant.py
+++ b/searx/engines/qwant.py
@@ -51,6 +51,7 @@ from urllib.parse import urlencode
 import babel
 from flask_babel import gettext  # pyright: ignore[reportUnknownVariableType]
 
+from searx.enginelib import EngineCache
 from searx.enginelib.traits import EngineTraits
 from searx.exceptions import (
     SearxEngineAccessDeniedException,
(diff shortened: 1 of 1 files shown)


searx/engines/qwant.py
```diff
@@ -163,8 +179,8 @@ def response(resp: "SXNG_Response") -> EngineResults:
         error_code = data.get("error_code")
```
Bnyro, line 179: I don't know if this `error_code` parameter still exists, never got it when testing.

searx/engines/qwant.py
```diff
@@ -163,8 +179,8 @@ def response(resp: "SXNG_Response") -> EngineResults:
...
         if error_code == 24:
             raise SearxEngineTooManyRequestsException()
-        if search_results.get("data", {}).get("error_data", {}).get("captchaUrl") is not None:
-            raise SearxEngineCaptchaException()
+        if search_results.get("url") is not None:
+            raise SearxEngineCaptchaException(suspended_time=0)
```
Bnyro, line 183: I've added `suspended_time=0` here on CAPTCHAs so that the second request (that now has the required cookie) can be done immediately.

return42 (approved):
LGTM .. thanks a lot 👍

PR #6409, opened 2026-07-14
[feat] engines: add exa (general)

Bnyro:
### What does this PR do?
- add support for https://exa.ai - its results are intended for AIs, but they're good enough to be used as normal results

- !exa test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/exa.py
+++ b/searx/engines/exa.py
@@ -1,0 +1,96 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Exa is a web search built for AI agents."""
+
+import typing as t
+
+from searx.exceptions import SearxEngineAPIException
+from searx.extended_types import SXNG_Response
+from searx.network import post
+from searx.result_types import EngineResults
+from searx.enginelib import EngineCache
+
+if t.TYPE_CHECKING:
+    from searx.search.processors import OnlineParams
+
+about = {
+    "website": "https://exa.ai",
+    "official_api_documentation": "https://exa.ai/docs/reference/search-api-guide",
+    "use_official_api": False,
+    "require_api_key": False,
+    "results": "JSON",
+}
+categories = ["general"]
+
+page_size = 10
+"""Page size must be less or equal than 10."""
+
+exa_mode: t.Literal["fast", "auto", "instant", "deep", "deep-lite", "deep-reasoning"] = "auto"
+"""Search type. See the `documentation <https://exa.ai/docs/reference/search#body-type-one-of-0>`_."""
+
+CACHE: EngineCache
+"""Cache for storing the scraped API Token."""
+
+base_url = "https://exa.ai"
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])  # type: ignore[reportAny]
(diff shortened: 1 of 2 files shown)


searx/engines/exa.py
```diff
@@ -0,0 +1,96 @@
...
+        raise SearxEngineAPIException("failed to obtain api key")
+
+    resp_json = token_resp.json()
+    token = resp_json["token"]
+    if not token:
+        raise SearxEngineAPIException("failed to obtain api key")
+
+    # subtract a bit from the expiry time to make sure there's no state
+    # where we use the old, invalid token although it's no longer valid
+    expire = resp_json["expiresIn"] - 50
```
Bnyro: I came across the issue that the `token` wasn't automatically deleted from cache in time, so I got "token invalid" responses.

Not sure if this was just because the token expired before `expiresIn`, or if the expiry cache has a bug.

return42: Is the `expire` time less than the `MAINTENANCE_PERIOD` (2h) of the cache? .. 

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/enginelib/__init__.py#L46-L56

Bnyro: It's usually 5 minutes.

That would explain it. Maybe we could add a check to the `CACHE.get` method that returns `None` if the entry is expired?

return42: > check to the `CACHE.get` method that returns `None` if the entry is expired?

Yes, that is a very good suggestion :+1: .. check `expire` column in `row[0]` -->

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/cache.py#L462-L466

if expire time is reached, we should return `default` .. but we should not delete the item in the DB .. the DELETE statement is to expensive .. and in most cases, a `None` will be immediately followed by a `cache.set` that resets the value of this `key` anyway (bulk deletions as part of maintenance interval is far more cost-effective) ..

FYI: here is the table schema ..

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/cache.py#L305-L309

- `strftime('%s', 'now')` is an SQLite expression and  
- `%s` [seconds since 1970-01-01](https://sqlite.org/lang_datefunc.html)

On the python side *expire* is ..

https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/cache.py#L279

Bnyro: > if expire time is reached, we should return default .. but we should not delete the item in the DB .. the DELETE statement is to expensive .. and in most cases, a None will be immediately followed by a cache.set that resets the value of this key anyway (bulk deletions as part of maintenance interval is far more cost-effective) ..
>
Very good point, didn't think about that!

Thanks for the insights on the cache implementation - I'll see if I can build a PR using that knowledge this evening or tomorrow :)

Bnyro: Sent https://github.com/searxng/searxng/pull/6414 (tested using the tiger engine by by modifying the expire time).

Bnyro:
Probably some LLM slop engines already picked up this PR and managed to get Exa to improve its bot-blocking, so this PR unfortunately no longer works :/

PR #6411, opened 2026-07-15
[fix] public domain image archive: crashes upon response

Bnyro:
The code used here has always been "bad" because `about` shouldn't be used as data source, but the engine probably broke when type checks / dataclasses for the about parameter in engines has been added with <https://github.com/searxng/searxng/pull/6258>.

Error log:
```
WARNING searx.engines.public domain im: ErrorContext('searx/engines/public_domain_image_archive.py', 143, '\'url\': _clean_url(f"{about[\'website\']}/images/{result[\'objectID\']}"),', 'TypeError', None, ("'EngineAbout' object is not subscriptable",)) False
ERROR   searx.engines.public domain im: exception : 'EngineAbout' object is not subscriptable
  File "/home/bnyro/Projects/searxng/searx/search/processors/online.py", line 253, in search
  File "/home/bnyro/Projects/searxng/searx/search/processors/online.py", line 239, in _search_basic
  File "/home/bnyro/Projects/searxng/searx/engines/public_domain_image_archive.py", line 143, in response
    'url': _clean_url(f"{about['website']}/images/{result['objectID']}"),
                         ~~~~~^^^^^^^^^^^
TypeError: 'EngineAbout' object is not subscriptable
```

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/public_domain_image_archive.py
+++ b/searx/engines/public_domain_image_archive.py
@@ -140,7 +140,7 @@ def response(resp):
         results.append(
             {
                 'template': 'images.html',
-                'url': _clean_url(f"{about['website']}/images/{result['objectID']}"),
+                'url': _clean_url(f"{pdia_base_url}/images/{result['objectID']}"),
                 'img_src': _clean_url(base_image_url),
                 'thumbnail_src': _clean_url(base_image_url + THUMBNAIL_SUFFIX),
                 'title': f"{result['title'].strip()} by {result['artist']} {result.get('displayYear', '')}",


return42:
`about['website']` can replaced by `about.website` ... however as you mentioned, its better to use *base_url* .. same here:

https://github.com/searxng/searxng/blob/f2432e33d6e3778cb0e976d602bb259d60081d64/searx/engines/qwant.py#L321

Bnyro:
> about['website'] can replaced by about.website ... however as you mentioned, its better to use base_url .. same here:
>
Yes, I thought so, but it's still a very ugly hack to rely on the `about` config for engine-specific logic such as base URLs.

return42:
Sorry if my English was unclear, I completely agree with you... I just wanted to point out that there's a second case where `about["website"]` was used instead of base_url... could you please fix that one too? Thanks!

PR #6413, opened 2026-07-15
[fix] tiger: remember auth cookie for 2 months, not 1 day

Bnyro:
### What does this PR do?
- fix the calculation for the cache expiry in for the `tiger` engine, the previous one assumed that `expire` is set in `minutes`, but it's actually set in `seconds`, so it has been off by factor 60.

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/tiger.py
+++ b/searx/engines/tiger.py
@@ -117,7 +117,7 @@ def _obtain_session_code() -> str:
     if not code:
         raise SearxEngineAPIException("failed to obtain session code")
 
-    CACHE.set("session", code, expire=60 * 24 * 60)  # cookie is valid for two months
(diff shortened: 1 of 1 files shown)


searx/engines/tiger.py
```diff
@@ -117,7 +117,7 @@ def _obtain_session_code() -> str:
     if not code:
         raise SearxEngineAPIException("failed to obtain session code")
 
-    CACHE.set("session", code, expire=60 * 24 * 60)  # cookie is valid for two months
+    CACHE.set("session", code, expire=60 * 24 * 60 * 60)  # cookie is valid for two months
```
return42, line 120: 🤔 2 month, 1 day .. 1 hour?

Bnyro: Sorry, what? 🤨

PR #6414, opened 2026-07-15
[fix] engine cache: return default value when cache entry expired

Bnyro:
It's possible that a cache entry is expired but has not yet been automatically deleted by the periodic maintenance because that, by default, runs only every 2 hours.

We now explicitly check for expiry, and if expired, return the default value.

Related:
- see https://github.com/searxng/searxng/pull/6409#discussion_r3584712601

### Code of Conduct

No AI used.

### DIFF
--- a/searx/cache.py
+++ b/searx/cache.py
@@ -458,12 +458,22 @@ class ExpireCacheSQLite(sqlitedb.SQLiteAppl, ExpireCache):
         # Before values are taken from the table, a maintenance interval may
         # need to be carried out.
         self.maintenance()
-        sql = f"SELECT value FROM {table} WHERE key = ?"
+        sql = f"SELECT value, expire FROM {table} WHERE key = ?"
         row = self.DB.execute(sql, (key,)).fetchone()
         if row is None:
             return default
 
-        return self.deserialize(row[0])
+        # Check if value is expired. It's possible that it's expired but has not
+        # yet been automatically deleted by the periodic maintenance
+        (value, expire) = row
+        now = time.time()
+        if expire < now:
+            # The record is deleted during the maintenance interval. Deleting
+            # the record at this point offers no advantage, as a SELECT
+            # statement must be executed for every cache.get request anyways.
+            return default
+
+        return self.deserialize(value)
 
     def pairs(self, ctx: str) -> Iterator[tuple[str, typing.Any]]:
         """Iterate over key/value pairs from table given by argument ``ctx``.


searx/cache.py
```diff
@@ -458,12 +458,22 @@ def get(self, key: str, default: typing.Any = None, ctx: str | None = None) -> t
...
-        return self.deserialize(row[0])
+        # Check if value is expired. It's possible that it's expired but has not
+        # yet been automatically deleted by the periodic maintenance
+        (value, expire) = row
+        now = time.time()
+        if expire < now:
+            # we don't delete the cache value here because that's unnecessary performance
+            # overhead, and engine implementations usually re-fetch the data and call `set`
+            # again when the cached value doesn't exist
+            return default
```
return42, line 474: I think variable `now` is not needed, `if expire < time.time()`.

I would phrase the comment slightly differently (assuming we need it at all):

*The record is deleted during the maintenance interval. Deleting the record at this point offers no advantage, as a SELECT statement must be executed for every `cahce.get` request anyway to determine that no record (or an outdated one) exists.*

return42 (approved):
LGTM .. I left a quick comment below.

Big improvement .. No idea why I didn't think of that myself right at the start.

searx/cache.py
Bnyro: > I think variable now is not needed, if expire < time.time().
>
In my opinion this extra step helps reading the code because I never used `time.time()` before, so I assumed it would be similar like a `datetime` object and just create an unitialized `time` object, not that it is a unix timestamp for the current time.

> I would phrase the comment slightly differently (assuming we need it at all):
>
I took over your suggestion, thanks.

I think we need a comment here to explain why
- it's possible that expired entries are still stored
- we don't delete the item here

Otherwise, if somebody looks at this again in a year, it might be unclear what we do here.

searx/cache.py
Bnyro, line 51: On a different note: Shouldn't this either be `2 * 60 * 60` or the comment should say `1h`?

Because `60 * 60 seconds` is definitely not `2h`? :)

Bnyro: Same thing here: https://github.com/searxng/searxng/blob/58e02a01ae1d3a422f0d4aead2a30804e5b59e1b/searx/enginelib/__init__.py#L50

return42: Oops .. I think "once per hour" is a good rule (is what we have and I don't see a reason to lift up to 2h) .. lets fix the comments ..

> the comment should say `1h`?

:+1:

searx/cache.py
return42: > so I assumed it would be similar like a `datetime`

:-) `time` and some other modules are very old and most often not much more than a wrapper around the C-standard libs (glib, musl / POSIX stuff) ..  time -> time.h / random, sys and os are based on stdlib.h etc.

searx/cache.py
virtadpt: What about making it configurable, but defaulting to 3600 seconds?

Bnyro: But why is such an option useful for the instance admin? Obviously you don't want a very long maintenance period because then your storage goes full, and you don't want to  have a very short one because of the performance overhead.

So 1h should be reasonably good for everyone

virtadpt: Some instances are in situations where storage is relatively plentiful (locally speaking) but network bandwidth isn't.  Mine is in such a network.

Bnyro: But this setting doesn't affect the amount of network requests at all.

The only thing that setting does is removing cache entries that are no longer valid. Cache entries that are still valid (have an expiry time in the future) remain in the cache.

If cache entries are expired but not removed yet because there's been no maintenance in the meantime, this doesn't mean that the cached value magically begins to be valid again. Instead, (after this PR), we tell the engine that there's no cache entry found in that case (because when storing the cache value, the engine specified that the value is going to expire in X seconds).

If we would forward the expired value to the engine, its HTTP requests would very likely get blocked because the token (e.g. scraped authentication token) is no longer valid anyways, so that doesn't improve the engine situation.

PR #6432, opened 2026-07-19
[feat] engines: add Exa Search API engine

kesku:
### What does this PR do?

- Add an Exa Search API engine `exaapi` (uses official API)

1. Get a key from https://dashboard.exa.ai/api-keys
2. Set `api_key` and `inactive: false`
3. Restart SearXNG and search with engine enabled

### Related issues

none

### Code of Conduct

AI tools used:
- OpenAI Codex: wrote `exaapi.rst` (tidied by hand)

All other code in this PR was written by me.

### DIFF
--- a/docs/dev/engines/online/exaapi.rst
+++ b/docs/dev/engines/online/exaapi.rst
@@ -1,0 +1,8 @@
+.. _exaapi engine:
+
+==============
+Exa API Engine
+==============
+
+.. automodule:: searx.engines.exaapi
+   :members:
--- a/searx/engines/exaapi.py
+++ b/searx/engines/exaapi.py
@@ -1,0 +1,167 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine to search using the official `Exa Search API`_. Exa is a search engine for AI agents.
+
+.. _Exa Search API: https://exa.ai/docs/reference/search
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`api_key`
+
+You can obtain an API key from the `API Key section <https://dashboard.exa.ai/api-keys>`_ in the Exa dashboard.
+
+Optional settings are:
+
+- :py:obj:`results_per_page`
+- :py:obj:`search_type`
+- :py:obj:`content_mode`
+- :py:obj:`content_max_characters`
+
+.. code:: yaml
+
+  - name: exaapi
+    engine: exaapi
+    api_key: "..."
+    results_per_page: 10
+    search_type: auto
+    content_mode: highlights
+    inactive: false
+
+The API supports SafeSearch and region-aware results.
+"""
+
+import typing as t
+
+from dateutil import parser
+
+from searx.exceptions import SearxEngineAPIException
+from searx.result_types import EngineResults
+from searx.utils import html_to_text
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+SearchType = t.Literal["fast", "auto", "instant", "deep", "deep-lite", "deep-reasoning"]
+ContentMode = t.Literal["highlights", "text"]
+
+about = {
+    "website": "https://exa.ai",
+    "wikidata_id": None,
+    "official_api_documentation": "https://exa.ai/docs/reference/search",
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": "JSON",
+}
+
+api_key: str = ""
+"""API key for Exa Search API (required)."""
+
+categories = ["general", "web"]
+safesearch = True
+
+base_url = "https://api.exa.ai/search"
+results_per_page: int = 10
+"""Maximum number of results per request. Value must be between 1 and 100, default is 10."""
+
+search_type: SearchType = "auto"
+"""Search type. Default is auto, see documentation for more information."""
+
+content_mode: ContentMode = "highlights"
+"""Content to request from the API: ``highlights`` (excerpts) or ``text`` (page text)."""
+
+content_max_characters: int = 500
+"""Maximum characters for the requested content."""
+
+
+def init(_):
+    if not api_key:
+        raise SearxEngineAPIException("No API key provided")
+    if not 1 <= results_per_page <= 100:
+        raise ValueError("results_per_page must be between 1 and 100")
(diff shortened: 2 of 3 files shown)


privacytime101:
So is this #6409 except meant for official API keys?

kesku:
@privacytime101 Yeah, exactly. That lets SearXNG take full advantage of Exa's API features with much better limits, similar to engines like https://github.com/searxng/searxng/blob/master/searx/engines/braveapi.py or https://github.com/searxng/searxng/blob/master/searx/engines/wolframalpha_api.py

Bnyro:
It would make more sense to put this into one single implementation because there's a big overlap in how it works.

kesku:
@Bnyro I did have a look at that PR before opening this. Main reason I kept them separate is it's already the pattern used elsewhere: `brave`/`braveapi`, `youtube_noapi`/`youtube_api`, `wolframalpha_noapi`/`wolframalpha_api`, etc. are all split between the unofficial/consumer flow and the official authenticated API.

These two are similarish at a high level but they're still different. One uses a consumer token flow (`/api/token/issue` + `/api/search`), while the other uses the documented `api.exa.ai/search` endpoint with a key.

I'd rather not make one engine switch between different auth methods + endpoints + schemas depending on whether `api_key` is configured as that feels messy 🙂 but happy to be convinced otherwise.

Bnyro:
If the response schemas are different, we should do different engines (to be honest I haven't checked that)

kesku:
I think they are different enough.

The request bodies differ slightly (`num_results` vs `numResults`, API has `moderation` + `userLocation`), and the consumer endpoint and official API don't return identical result fields.

The official API can itself return two shapes depending on the requested content mode: `contents.text` requests the page text, while `contents.highlights` requests dense excerpts from the full text.

privacytime101:
@kesku Cool, I just have a suggestion, could you maybe include `search_type` in the default settings.yml section? I see it has a default value of "auto" but I think it's an important option to have visible (I'm just looking at the docs, not sure). Also, would the deep outputs be useful for us? Thanks.

kesku:
> could you maybe include search_type in the default settings.yml section

Good idea! Just added `search_type` and `content_mode` @privacytime101 🙂 

I've had really good results with the deep-lite model for most general web searches.

Bnyro:
@kesku please rebase instead of merging the master branch, adding merge commits only makes things harder in the end. Thank you!

docs/dev/engines/online/exaapi.rst
```diff
@@ -0,0 +1,11 @@
...
+
+==============
+Exa API Engine
+==============
+
+.. automodule:: searx.engines.exaapi
+   :members:
+
+The API engine requires an API key from Exa. This can be obtained from the
+`API Dashboard <https://dashboard.exa.ai/api-keys>`_.
```
Bnyro:
```suggestion
```
The documentation automatically contains the full docstring of the `exaapi.py` file, so that extra information is not needed here because it's already in the docstring.

You can test the docs with `make docs.live`

searx/engines/exaapi.py
```diff
@@ -0,0 +1,175 @@
...
+
+SearchType = t.Literal["fast", "auto", "instant", "deep", "deep-lite", "deep-reasoning"]
+SEARCH_TYPES: tuple[SearchType, ...] = (
+    "fast",
+    "auto",
+    "instant",
+    "deep",
+    "deep-lite",
+    "deep-reasoning",
+)
```
Bnyro:
```suggestion
```
Not needed, you can get a list of all search types with `t.get_args(SearchType)`.

searx/engines/exaapi.py
```diff
@@ -0,0 +1,175 @@
...
+
+ContentMode = t.Literal["highlights", "text"]
+CONTENT_MODES: tuple[ContentMode, ...] = ("highlights", "text")
```
Bnyro:
```suggestion
```
see above

searx/engines/exaapi.py
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    # Apply SafeSearch if enabled
+    if params["safesearch"]:
+        body["moderation"] = True
+
+    # Apply region-aware results if specified
+    locale_parts = params["searxng_locale"].split("-")
+    region = locale_parts[-1]
+    if len(locale_parts) > 1 and len(region) == 2 and region.isalpha():
```
Bnyro:
```suggestion
    if len(locale_parts) > 1:
```
Aren't country codes two letters anyways? I don't think we need these checks, all locales in SearXNG satisfy these conditions (except for the ones that don't have a country, so the first check is right)

kesku: True, I copied this from an older Exa integration I wrote a while back when the locale handling wasn't based on ISO standards

searx/engines/exaapi.py
kesku: addressed! good call

Bnyro:
Also, please squash the changes into a single commit to make it easier to review. Thanks.

searx/engines/exaapi.py
```diff
@@ -0,0 +1,165 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Engine to search using the official Exa Search API. Exa is a search engine for AI agents.
```
Bnyro:
```suggestion
"""Engine to search using the official `Exa Search API`_. Exa is a search engine for AI agents.
```
Otherwise the link to the search api you defined below is not shown in the docs.

searx/engines/exaapi.py
```diff
@@ -0,0 +1,165 @@
...
+
+.. _Exa Search API: https://exa.ai/docs/reference/search
+
+Configuration
+=============
+
+The engine has the following mandatory setting:
+
+- :py:obj:`api_key`
```
Bnyro, line 11:
```suggestion
- :py:obj:`api_key`

You can obtain an API key from the `API Key section <https://dashboard.exa.ai/api-keys>`_ in the Exa dashboard.
```

Bnyro (approved):
Thanks!

privacytime101:
@kesku @Bnyro It seems that by default the settings.yml entry causes `ERROR:searx.engines: Engine config error: ambiguous shortcut: -`?

Manually adding a shortcut entry fixes it

Bnyro:
Good catch, fixed in https://github.com/searxng/searxng/pull/6463

PR #6433, opened 2026-07-19
[upd] update to gecko driver v37

kesku:
### What does this PR do?

Bump `GECKODRIVER_VERSION` from `v0.36.0` to `v0.37.0`

GHA runners currently ship Firefox 152. Selenium warns geckodriver 0.36.0 may not be compatible and recommends 0.37.0:

CleanShot 2026-07-19 at 11 59 50@2x

### How to test this PR locally?

```
make test.robot
```

Confirm warning is gone.

### Related issues

- #3141
- #3735

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [x] **I hereby confirm that this PR conforms with the [AI Policy].**

  If I have used AI tools for working on the changes in this PR, I will
  attach a list of all AI tools I used and how I used them. I hereby confirm
  that I haven't used any other tools than the ones I mention below.
  
No AI used.

PR #6434, opened 2026-07-19
[fix] manage: detect correct geckodriver platform

kesku:
### What does this PR do?

`./manage gecko.driver` identifies platforms incorrectly. It uses `platform.architecture()` which is pointer width, NOT cpu arch, and matches strings that python never returns (eg `mac 64bit` instead of `darwin arm64`/`darwin x86_64`), so wrong geckodriver  gets downloaded or the script constructs an invalid download URL.

This switches platform detection to `platform.system()` and `platform.machine()`, maps platform strings to [Mozilla's published release assets](https://github.com/mozilla/geckodriver/releases) (`macos-aarch64`, `linux-aarch64`, `win-aarch64`, etc)

CleanShot 2026-07-19 at 13 32 34@2x

On Windows, it also downloads the correct `.zip` archive and installs `geckodriver.exe`.

Unsupported platforms now fail with a clear error instead of constructing an invalid download URL.

32-bit Linux is dropped (Mozilla discontinued `linux32` builds in geckodriver 0.37.0)

```bash
rm -f local/py3/bin/geckodriver

./manage gecko.driver

local/py3/bin/geckodriver -V
file local/py3/bin/geckodriver
```

Depending on platform, this should download:
- Apple Silicon (M1/M2/M3): `macos-aarch64`
- Intel Mac: `macos`
- Linux x86_64: `linux64`
- Linux ARM64: `linux-aarch64`
- Windows AMD64: `win64`
- Windows ARM64: `win-aarch64`
- Windows x86: `win32`

### Related issues

Follow-up to https://github.com/searxng/searxng/pull/6433#discussion_r3610453038

### Code of Conduct

OpenAI Codex used for looking up geckodriver release assets / platform strings (it just curled for me). Code written by me.

### DIFF
--- a/manage
+++ b/manage
@@ -48,7 +48,7 @@ PATH="${PY_ENV}/bin:${REPO_ROOT}/node_modules/.bin:${GOROOT}/bin:${GOPATH}/bin:$
 
 PYOBJECTS="searx"
 PY_SETUP_EXTRAS='[test]'
-GECKODRIVER_VERSION="v0.36.0"
+GECKODRIVER_VERSION="v0.37.0"
 # SPHINXOPTS=
 BLACK_OPTIONS=("--target-version" "py311" "--line-length" "120" "--skip-string-normalization")
 BLACK_TARGETS=("--exclude" "(searx/static|searx/languages.py)" "--include" 'searxng.msg|\.pyi?$' "searx" "searxng_extra" "tests")
--- a/manage
+++ b/manage
@@ -181,22 +181,48 @@ gecko.driver() {
             build_msg INSTALL "geckodriver already installed"
             return
         fi
-        PLATFORM="$(python -c 'import platform; print(platform.system().lower(), platform.architecture()[0])')"
+        PLATFORM="$(python -c 'import platform; print(platform.system().lower(), platform.machine().lower())')"
+        EXT="tar.gz"
         case "$PLATFORM" in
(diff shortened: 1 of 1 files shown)


kesku:
Depends on #6433. This is opened against master because the base branch only exists on my fork. Please merge #6433 first and I’ll rebase this branch, or close #6433 and merge this PR instead.

Both currently include the geckodriver 0.37.0 update because that release drops 32-bit Linux support, and keeping the version bump with the platform changes makes the history cleaner 🙂

return42:
Since we don't support M$ nor macOS we don't need a `EXT` variable nor do we need to switch to platform.machine

*This is an old bash script, we should not investigate to much.*

kesku:
@return42 we can remove the mac/windows parts if that's the issue. I only touched them because `master` still contains those cases; those should be removed from master! 🙂

The PR also fixes two Linux issues:

1. After #6433 bumped the driver to v37 the `"linux 32bit"` branch builds a URL for a `linux32` asset Mozilla no longer publishes, so that path 404s. It should be removed!

2. On Linux `aarch64`, `platform.architecture()[0]` still returns `"64bit"`, so the current code matches `"linux 64bit"`, sets `ARCH=linux64`, and downloads an x86-64 binary onto an ARM machine, resulting in an `Exec format error`. `platform.architecture()` only reports pointer width. It cannot distinguish `x86_64` from `aarch64`

The mac/windows parts of the code have been in the script since 2017. I did not introduce them.

PR #6439, opened 2026-07-22
[fix] container: ignore invalid SEARXNG_PORT values

### LINKED ISSUE #6437: Kubernetes: `SEARXNG_PORT` conflicts with service environment variable and causes startup failure
Starting the latest SearXNG Docker image in Kubernetes fails because the `SEARXNG_PORT` environment variable conflicts with the legacy service environment variables injected by Kubernetes.

### How To Reproduce?

1. Deploy `searxng/searxng:2026.7.19-6da6eee26` to Kubernetes.
2. Create a Service named `searxng`.
3. Start the Pod.

The container exits with:

```text
Error: Invalid value for '--port': 'tcp://10.x.x.x:80' is not a valid integer.
```

Setting the following environment variable works around the issue:

```yaml
env:
  - name: SEARXNG_PORT
    value: "8080"
```

### Expected behavior

SearXNG should start successfully without requiring `SEARXNG_PORT` to be overridden, even when the Kubernetes Service is named `searxng`.

### Screenshots & Logs

```text
Error: Invalid value for '--port': 'tcp://10.x.x.x:80' is not a valid integer.
```

Relevant environment variables:

```text
PORT=8080
SEARXNG_PORT=tcp://10.x.x.x:80
SEARXNG_SERVICE_HOST=10.x.x.x
SEARXNG_SERVICE_PORT=80
```

### Version of SearXNG

```
2026.7.19-6da6eee26
```

Docker image deployed on Kubernetes (k3s).

### Additional context

The issue appears to be caused by Kubernetes automatically injecting `SEARXNG_PORT=tcp://...` for a Service named `searxng`. The image seems to interpret this value as the listening port instead of an integer.

### Code of Conduct

burakemirsezen:
### What does this PR do?

Ignores non-numeric SEARXNG_PORT values instead of failing to start.

This fixes an issue with Kubernetes, which passes `SEARXNG_PORT` env if the service is named `searxng`. That env is `tcp://...` and it prevents the startup of the app. In such a case the env will be ignored instead of causing the startup issue.

Run the entrypoint.sh while setting SEARXNG_PORT to valid and invalid values.

### Related issues

Closes #6437.

### Code of Conduct

  None

### DIFF
--- a/container/entrypoint.sh
+++ b/container/entrypoint.sh
@@ -112,6 +112,15 @@ if [ "$(id -u)" -eq 0 ]; then
 fi
 
 # ENVs aliases
-export GRANIAN_PORT="${SEARXNG_PORT:-$GRANIAN_PORT}"
+case "${SEARXNG_PORT:-}" in
+    '') ;;
+    *[!0-9]*)
+        echo "WARNING: ignoring non-numeric SEARXNG_PORT='$SEARXNG_PORT'" >&2
+        unset SEARXNG_PORT
+        ;;
+    *)
+        export GRANIAN_PORT="$SEARXNG_PORT"
+        ;;
+esac
 
 exec /usr/local/searxng/.venv/bin/granian searx.webapp:app
--- a/container/entrypoint.sh
+++ b/container/entrypoint.sh
@@ -112,10 +112,10 @@ if [ "$(id -u)" -eq 0 ]; then
 fi
 
 # ENVs aliases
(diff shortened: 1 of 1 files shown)


burakemirsezen:
I saw the previous discussions on this after opening this PR. I think this is still a useful fix. It does not change normal SearXNG Docker usage, but it avoids a confusing startup failure for kube users when it injects a `SEARXNG_PORT` variable.

inetol (approved):

PR #6465, opened 2026-07-28
[mod] docs: sidebar remove verbose features

inetol:
### What does this PR do?

Remove all references to `dev.searxng.org` and get rid of some highlights

`make docs.live`

### Code of Conduct


docs/index.rst
vojkovic, line 82: Should this be removed too?

inetol: We still have a Browserstack acc.

vojkovic: Will we use it though now that dev.searxng.org is gone?

inetol: I usually use it to test older Chromium/Safari versions. I haven't looked into automation, though.

vojkovic (approved):

PR #6470, opened 2026-07-29
 [mod] engines: replace [random.choice(...) for ... in range(...)] with random.choices

Bnyro:
### What does this PR do?
- some misc cleanup regarding random string generation to make it the same everywhere


inetol (approved):

Bnyro:
Thanks for the fixup @inetol !

Issue #6177, opened 2026-05-30
Hardcoded fallback API key in Pexels engine


RinZ27:
### Description

The Pexels engine currently uses a hardcoded fallback API key in 'searx/engines/pexels.py'. While this provides a 'plug-and-play' experience for some users, it presents a potential security risk by exposing a static credential in the codebase. I recommend moving this fallback key to a configuration-based approach (e.g., settings.yml or environment variables) to improve secret management and maintainability.

### How To Reproduce?

Inspect 'searx/engines/pexels.py' at line 29.

### Expected behavior

Credentials should not be hardcoded in the source code.

### Code of Conduct

- [x] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.

AI tools used: Gemini CLI.

vojkovic:
https://github.com/searxng/searxng/pull/5954/changes#r3067801198

PR #6476, opened 2026-07-30
[ci] add ai policy checkbox detection

vojkovic:
### What does this PR do?

Adds CI to automatically close issues and give them the `invalid:slop` tag.

Only affects new users and authors who haven't contributed yet just to reduce the number false positives e.g. if someone forgets by mistake.

The motivation for this is to reduce the burden on the maintainers responding to low quality work and bad AI actors.

Make an issue / pr as a new user.

Examples:
[Checkbox Ticked](https://github.com/vojkovic/searxng/issues/5)
[Checkbox Unticked](https://github.com/vojkovic/searxng/issues/4)

### Code of Conduct

Cursor to teach me github script

### DIFF
--- a/.github/workflows/ai-policy.yml
+++ b/.github/workflows/ai-policy.yml
@@ -1,0 +1,38 @@
+---
+# yamllint disable rule:line-length
+name: AI Policy
+
+# Closes any new issues and PRs from people (or agents) who don't accept the AI Policy
+
+# yamllint disable-line rule:truthy
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  check:
+    name: Check AI Policy
+    # for issues with an author who has not contributed before
+    if: >-
+      github.event.sender.type != 'Bot' &&
+      contains(fromJSON('["NONE","FIRST_TIMER","FIRST_TIME_CONTRIBUTOR"]'),
+               github.event.issue.author_association ||
+               github.event.pull_request.author_association)
+    runs-on: ubuntu-26.04-arm
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: "false"
+
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3  # v9.0.0
+        with:
+          script: |
+            const script = require('./.github/scripts/ai_policy.cjs');
+            await script({ github, context });
--- a/.github/scripts/ai_policy.cjs
+++ b/.github/scripts/ai_policy.cjs
@@ -1,0 +1,39 @@
+// Closes issues and prs whose authors/agents don't accept the ai policy
+// https://github.com/searxng/searxng/blob/master/AI_POLICY.rst
+
+module.exports = async ({ github, context }) => {
+  const item = context.payload.pull_request || context.payload.issue;
+  const body = item.body || '';
+  const kind = context.payload.pull_request ? 'pull request' : 'issue';
+
+  // https://github.com/searxng/searxng/pull/6476#discussion_r3683782481
+  const hasBox = /\[[Xx]\].*AI Policy/.test(body);
+  const hasRef = /\[AI Policy\]:\s*https:\/\/github\.com\/searxng\/searxng\/.*AI_POLICY/.test(body);
+  if (hasBox && hasRef) {
+    return;
+  }
+
+  const { owner, repo } = context.repo;
+  await github.rest.issues.createComment({
+    owner,
+    repo,
+    issue_number: item.number,
+    body:
+      'Hello! Thank you for your contribution.\n\n' +
+      `Unfortunately your ${kind} was closed as the AI Policy has not been accepted.\n\n` +
+      `Please open a new ${kind} after confirming your contribution aligns with our AI Policy.`,
+  });
+  await github.rest.issues.addLabels({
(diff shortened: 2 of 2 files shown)


.github/workflows/ai-policy.yml
```diff
@@ -0,0 +1,110 @@
...
+    if: >-
+      github.event_name == 'issues' &&
+      github.event.sender.type != 'Bot' &&
+      contains(fromJSON('["NONE","FIRST_TIMER","FIRST_TIME_CONTRIBUTOR"]'),
+               github.event.issue.author_association)
+    runs-on: ubuntu-26.04-arm
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3  # v9.0.0
```
Bnyro: Why pin the commit instead of the tag?
```suggestion
      - uses: actions/github-script@v9.0.0
```

.github/workflows/ai-policy.yml
```diff
@@ -0,0 +1,110 @@
...
+        with:
+          script: |
+            const body = context.payload.issue.body || '';
+            const lines = body.split(/\r?\n/).map((l) => l.trim().replace('[X]', '[x]'));
+
+            const ref = '[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst';
+            const box = '- [x] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.';
```
Bnyro:
```suggestion
            const box = '- [x] I read the [AI Policy]';
```
I think it's fine to only check the start of the string, e.g. if somebody by accident removes the "." in the end of the sentence that would also block them. Even better would be to match with a regex like `/\[[Xx]\].*AI Policy/` to make it robust if we change the text in the future. Then you also don't need to do the `replace('[X]', '[x]'))` above

.github/workflows/ai-policy.yml
```diff
@@ -0,0 +1,110 @@
...
+              issue_number: n,
+              labels: ['invalid:slop'],
+            });
+            await github.rest.issues.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: n,
+              state: 'closed',
+              state_reason: 'not_planned',
+            });
```
Bnyro: I think this part is exactly the same as for issues? Then we could move it into a different script file as described in https://github.com/actions/github-script#run-a-separate-file, having the same commenting logic twice is a bit ugly.

The `body` parameter in `createComment` seems to differ, but we can probably just replace this text to be less specific or just pass a boolean whether it's a PR or issue to the script.

.github/workflows/ai-policy.yml
```diff
@@ -0,0 +1,110 @@
...
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number }}
+  cancel-in-progress: true
+
```
Bnyro: Why do we need this? Does that just re-run the workflow if the PR author changes the PR text very quickly after opening the PR (i.e. ticks the box although it wasn't ticked when initially opening the PR)?

I thought that `types: [opened]` already makes sure that it only runs once when the PR is opened?

Bnyro:
Looks very helpful, thanks for working on it so far :+1:

.github/workflows/ai-policy.yml
vojkovic: We pin with the version hash since https://github.com/searxng/searxng/pull/5099

It makes it more reliable

.github/workflows/ai-policy.yml
vojkovic: Yeah I agree tbh... the main motivation for having two seperate ones was that the box label changes from PR / issue; if we do it with the regex method you mentioned, it would be much cleaner.

vojkovic: Yeah let's move it into a seperate file - if nothing else, my IDE's syntax highlighter will appreciate it.

.github/workflows/ai-policy.yml
vojkovic: You are right... I thought it would trigger twice if the author edited their commit but apparently it is a seperate type 'edited'. Will remove

.github/workflows/ai-policy.yml
Bnyro: Didn't see that PR, thanks :+1:

.github/workflows/ai-policy.yml
vojkovic: Will swap to the regex. My motivation for the ref check btw was that some bots seem to also remove the ref:

- https://github.com/searxng/searxng/issues/6177

What do you think? Is it worth keeping?

Bnyro: Good point. I think we should do a second regex to search the link definition then?

vojkovic:
Thank you for your comments, it is always a pleasure reading them :)

I have force pushed https://github.com/searxng/searxng/commit/852f55bcdb28b0d773a418f39785dd17605ca239 to address them.

I can't really test it because I'm the repo owner but maybe you could make two issues with and without the checkbox to see if it works as intended -> https://github.com/vojkovic/searxng/issues

Bnyro (approved):
Works! https://github.com/vojkovic/searxng/issues/9 and https://github.com/vojkovic/searxng/issues/8.

Looks great now, thanks again! Good to merge from my side, idk if we should wait for somebody else's review.

return42 (approved):
LGTM .. thanks a lot 👍

inetol (approved):

PR #6481, opened 2026-08-01
[feat] autocomplete: add kagi autocompleter

vojkovic:
### What does this PR do?

Adds a Kagi Autocompleter backend.

Used https://kagisuggest.com/api/autosuggest from [here](https://help.kagi.com/kagi/getting-started/setting-default.html#custom_use).

I am hoping it is less monitored than the more public endpoint.

Enable kagi autocompleter in preferences

Type something into search bar

Confirm autocompleter results work

### Related issues

none

### Code of Conduct

Cursor found the r arg for the locale

### DIFF
--- a/searx/autocomplete.py
+++ b/searx/autocomplete.py
@@ -157,6 +157,24 @@ def google_complete(query: str, sxng_locale: str) -> list[str]:
     return results
 
 
+def kagi(query: str, sxng_locale: str) -> list[str]:
+    """Autocomplete from Kagi."""
+
+    args: dict[str, str] = {'q': query}
+
+    if '-' in sxng_locale:
+        args['r'] = sxng_locale.split('-')[1].lower()
+
+    resp = get("https://kagisuggest.com/api/autosuggest?" + urlencode(args))
+    results: list[str] = []
+
+    if resp.ok:
+        data = resp.json()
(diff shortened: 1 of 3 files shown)


searx/autocomplete.py
```diff
@@ -157,6 +157,24 @@ def google_complete(query: str, sxng_locale: str) -> list[str]:
     return results
 
 
+def kagi(query: str, sxng_locale: str) -> list[str]:
+    """Autocomplete from Kagi."""
+
+    args: dict[str, str] = {'q': query}
+
+    if '-' in sxng_locale:
+        args['r'] = sxng_locale.split('-')[1].lower()
```
Bnyro, line 166: Where did you find the `r` parameter? From my testing it doesn't seem to change anything (tried with `&r=de`, maybe `de` is just not supported by the autocompleter, idk)

Bnyro (approved):
Works well (except for the region option? see above). Nice finding that it works without signup.

Lgtm 👍

searx/autocomplete.py
vojkovic: Cursor (AI) found it just with a little fuzzing but it definitely does work - I checked it before
Screenshot

return42 (approved):

Issue #6228, opened 2026-06-09
Address bar does not show search query - Search engine is added to a browser the way that a GET/POST method can not be changed from instance afterward


slrslr:
### Is your feature request related to a problem?

When i add https://find.xenorio.xyz to the Firefox 151 while instance has set POST in its Privacy settings page, and then change the setting to GET, it does not respect my choice (to use GET) when searching using browser address bar, because the search results page address bar does not show my search query (GET should show it, POST not).

I have to delete the search engine and re-add it again with GET set in settings. This is confusing and can lead to removal of SearxNG by user, because he may want to have search query indexed in history and copy-able.

Details (incl. images): https://github.com/tiekoetter/searxng/issues/75

### Describe the solution you'd like

Respect change of a setting of a GET/POST.

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [x] I read the [AI Policy] and hereby confirm that this issue conforms with the policy.

inetol:
It can't be fixed. You have to configure the instance first before adding the search in the browser.

https://github.com/dewitt/opensearch/blob/590770d9a1da863fc175159a2e70d9e64e1076d0/opensearch-1-1-draft-6.md

PR #6482, opened 2026-08-01
[fix] autocomplete: drop POST method

vojkovic:
### What does this PR do?

Drops the POST method support entirely from the autocompleter. Firefox cannot do POST requests in the URL bar autosuggestions - for it to work, it needs GET requests. This means if the instance has configured POST method by default, the autocomplete will not work in the URL bar.

This problem is compounded by the fact that Firefox does not send the cookies when fetching the opensearch.xml file, so **even if the user has selected 'GET', it does not matter**.

I have been doing this on my [fork](https://github.com/privau/searxng/commit/a750d1651fc98240429209a668350592d732f7f0) successfully for two years.

The autocompleter is currently the only place outside of /search where the 'HTTP Method' is considered. We don't do this for the favicon proxy nor the image proxy, and I propose we don't do it here either.

Chrome has never supported POST requests for autocomplete, so there are no losses there.

1. Confirm HTTP Method is POST in settings.yml (already set by default) and enable any autocompleter.
2. Type in the search box to confirm that suggestions still appear (and it is using GETs to /autocompleter in devtools).
3. Open /opensearch.xml, the method should be GET for the suggestions URL
4. Add search engine into firefox and test that it works now.

### Related issues

- https://github.com/privau/searxng/issues/39
- https://github.com/searxng/searxng/discussions/4262
- https://github.com/searxng/searxng/issues/711 (Related)

### Code of Conduct

Cursor did code review but no edits... code is mine


Bnyro (approved):
> This means if the instance has configured POST method by default, the autocomplete will not work in the URL bar.
>
Ah, that explains why auto-suggestions never worked for me on Firefox's search bar.

I don't see any benefit in keeping the POST method supported for suggestions then, thanks for looking into this 👍

searx/templates/simple/opensearch.xml
```diff
@@ -13,7 +13,7 @@
   </Url>
   {% endif %}
   {% if autocomplete %}
-    <Url rel="suggestions" type="application/x-suggestions+json" method="{{ opensearch_method }}" template="{{ url_for('autocompleter', _external=True) }}?q={searchTerms}"/>
+    <Url rel="suggestions" type="application/x-suggestions+json" method="GET" template="{{ url_for('autocompleter', _external=True) }}?q={searchTerms}"/>
```
return42, line 16: Not sure, can we remove some more occurrences of `opensearch_method` ..

- https://github.com/search?q=repo%3Asearxng%2Fsearxng%20opensearch_method&type=code

including the if/else conditions?  Do we even need a distinction between `method` and `opensearch_method` anymore?

Not really related, but to mention in that context:

- https://github.com/searxng/searxng/issues/6228
- https://github.com/searxng/searxng/pull/2333#issuecomment-1565392120
- https://github.com/searxng/searxng/issues/1889#issuecomment-1565366772

return42 (approved):
LGTM 👍 .. I added a comment.

searx/templates/simple/opensearch.xml
vojkovic: > Not sure, can we remove some more occurrences of `opensearch_method` ..

Good question. The other occurrences in the opensearch.xml are for for the /search endpoint which is a separate topic... I am not sure we are ready to discuss the POST vs GET debate again 😆 - but, in short my view is the benefits of POST are outweighed by its UX constraints, so I would make GET the default and then also enable some privacy respecting autocomplete backend like brave or ddg by default.

Long term, I would like to eventually implement a richer autocomplete with descriptions and images (look at my autocomplete on priv.au as an example).

image


> Do we even need a distinction between method and opensearch_method anymore

I think it is just used for the template... I'm not fussed either way

https://github.com/searxng/searxng/blob/8892414dc38dd57728b7f62f33152ea80e3b305f/searx/webapp.py#L1201

return42: > Good question. The other occurrences in the opensearch.xml are for for the /search endpoint

Not all .. this one seems wrong since a long time ..

https://github.com/searxng/searxng/blob/8892414dc38dd57728b7f62f33152ea80e3b305f/searx/templates/simple/opensearch.xml#L18

I'm not sure if we even need this line... we don't actually have a “description,” do we (or am I misunderstanding something)?

The endpoint we have is `opensearch.xml` ..

https://github.com/searxng/searxng/blob/8892414dc38dd57728b7f62f33152ea80e3b305f/searx/webapp.py#L1188-L1201

But the longer I look at the code, the less I understand it. The `opensearch.xml` endpoint is only accessible via GET; the autocomplete feature only works via GET... yet the `search` endpoint is the only one that supports both POST and GET... and we use the name `method` everywhere for that ... except here .. that should be cleaned up.

And line 1194-1196 should be removed .. make no sense to me, or I'm wrong?

----

>  I am not sure we are ready to discuss the POST vs GET debate again 😆 -

I fully agree with you, but I've been the one to start them the last few times; now it's someone else's turn ;-)

> I would like to eventually implement a richer autocomplete with descriptions and images 

Wow looks great ... But does it really have to be Helene Fischer? When I hear her music (German "Schlager"), my ears start to bleed. ;-)

vojkovic: > I'm not sure if we even need this line... we don't actually have a “description,” do we (or am I misunderstanding something)?

Ha - it is odd!... Okay I looked up the opensearch spec: [`rel="self"` is the OSD automatic update url](https://developer.mozilla.org/en-US/docs/Web/XML/Guides/OpenSearch#supporting_automatic_updates_for_opensearch_description), it's not a description... and `method` is ignored on every single client I looked at (they just GET regardless) so that part is safe to remove. Done ([diff](https://github.com/searxng/searxng/compare/b29552fcba8488b6405a6a7c3178df14bc37fd8e..e55e0ef7d3442e77fa8a4091ad3b286e6f7e3871))

> But the longer I look at the code, the less I understand it.

I am feeling this very much right now 😄

> And line 1194-1196 should be removed .. make no sense to me, or I'm wrong?

Ugh, they look useful unfortunately.... Chromium famously [does not support POST search](https://issues.chromium.org/issues/40304654) - BTW [the code was added 13 years ago](https://github.com/searxng/searxng/commit/174daf703b341455c3320b61034cdb74bee9b55d#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R109)... 🤯 

So I would keep 1194-1196 and also `opensearch_method` for the time being.

---

> now it's someone else's turn ;-)

😭 I'm sorry!!! I think it is best for me to make a new PR for the GET and setting an autocomplete by default, or we can merge https://github.com/searxng/searxng/pull/2333 although I would go with brave instead of ddg as it supports the rich suggestions I was talking about.

Also, Helene Fischer is my favourite... how could you?? 😢 Lines such as *'die tasse kaffee und auch das glas wein'* represent the pinnacle of German lyricism.

return42: > Ugh, they look useful unfortunately.... Chromium famously [does not support POST search](https://issues.chromium.org/issues/40304654) - BTW [the code was added 13 years ago](https://github.com/searxng/searxng/commit/174daf703b341455c3320b61034cdb74bee9b55d#diff-ab9553a5df12bab91ccac7a4b9af78fbd160e35bffae9a340bf75ed736d50a07R109)... 🤯

Sorry, I didn't look closely enough 🤦 

> So I would keep 1194-1196 and also `opensearch_method` for the time being.

Sure, makes sense.

--> PR is ready to merge :+1: 

> I think it is best for me to make a new PR for the GET and setting an autocomplete by default

Yes, please make a new one, #2333 is from 2023 and does not include the POST vs. GET discussion.

-----

> Also, Helene Fischer is my favourite... how could you?? 😢 Lines such as _'die tasse kaffee und auch das glas wein'_ represent the pinnacle of German lyricism.

OMG, what kind of impression did my fellow countrymen leave on you... 

Let's see if it's still your favorite when you're sitting on a train in Germany that, for some inexplicable reason, has broken down somewhere again, and there's a group in the train car that's celebrating a bachelor party to the music of Helene Fischer .. / OK, torture of this kind falls under the Geneva Conventions .. but who cares these days :-)

vojkovic: 😆 😆 😆 

Thats it... Putting Atemlos durch die Nacht on repeat out of pure spite now

PR #6484, opened 2026-08-03
[mod] defaults: GET method and ddg autocomplete

### LINKED ISSUE #1889: [Feature request] Enable autocomplete by default

vojkovic:
### What does this PR do?

Changes the default of `server.method` from `POST` to `GET` and `search.autocomplete` from disabled to `duckduckgo`. 

I originally wanted to use brave but I ran into some issues getting consistent results when testing -> ddg is more reliable and privacy respecting enough to make it a solid enough default. (Brave also doesn't support languages but ddg does). Perhaps we can use kagi once https://github.com/searxng/searxng/pull/6481 is merged, as that has the ability for rich output and language support.

Autocomplete being off by default also means `/opensearch.xml` is often added without the suggestions URL and Firefox will cache a broken setup unless the user picks an autocompleter backend and then adds the engine back.

The motivation for swapping to GET by default is that while POST provides the very minor benefit of keeping search queries out of the user's history, it causes a LOT of UX pain. Due to this, GET has been recommended to the instance maintainers for a while (#3619  - read through that PR's comments for a more complete picture). In the past, we have been divided on this issue, so I think it's time we discuss it again to get everyone's views. 😛 

`make run` and confirm preferences default to GET and duckduckgo.

### Related issues

- Closes: https://github.com/searxng/searxng/issues/1889
- https://github.com/searxng/searxng/issues/711
- https://github.com/searxng/searxng/issues/3590
- https://github.com/searxng/searxng/pull/3619
- https://github.com/searxng/searxng/pull/2333
- https://github.com/searxng/searxng/pull/6482
- https://github.com/searxng/searxng/discussions/4262

### Code of Conduct

Cursor for testing

### DIFF
--- a/searx/templates/simple/elements/apis.html
+++ b/searx/templates/simple/elements/apis.html
@@ -4,7 +4,7 @@
     <div class="wrapper">
       {%- for output_type in search_formats -%}
         <div class="left">
-          <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
+          <form method="{{ method }}" action="{{ url_for('search') }}">
             <input type="hidden" name="q" value="{{ q|e }}">
             {%- for category in selected_categories -%}
               <input type="hidden" name="category_{{ category }}" value="1">
--- a/docs/admin/settings/settings_search.rst
+++ b/docs/admin/settings/settings_search.rst
@@ -8,7 +8,7 @@
 
    search:
      safe_search: 0
-     autocomplete: ""
+     autocomplete: "duckduckgo"
      favicon_resolver: ""
      default_lang: ""
      ban_time_on_fail: 5
@@ -32,7 +32,7 @@
   - ``2``: Strict
 
 ``autocomplete``:
-  Existing autocomplete backends, leave blank to turn it off.
+  Existing autocomplete backends, set blank to turn it off.
 
   - ``360search``
   - ``baidu``
--- a/docs/admin/settings/settings_server.rst
+++ b/docs/admin/settings/settings_server.rst
@@ -14,7 +14,7 @@
        limiter: false
        public_instance: false
        image_proxy: false
-       method: "POST"
+       method: "GET"
        default_http_headers:
          X-Content-Type-Options : nosniff
          X-Download-Options : noopen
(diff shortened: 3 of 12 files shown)


utils/templates/etc/searxng/settings.yml
```diff
@@ -8,7 +8,6 @@ general:
 
 search:
   safe_search: 2
-  autocomplete: 'duckduckgo'
```
Bnyro, line 11: I don't think it hurts to keep this so that the instance admin can easily change it without consulting the docs?

searx/templates/simple/elements/apis.html
```diff
@@ -4,7 +4,7 @@
     <div class="wrapper">
       {%- for output_type in search_formats -%}
         <div class="left">
-          <form method="{{ method or 'POST' }}" action="{{ url_for('search') }}">
+          <form method="{{ method or 'GET' }}" action="{{ url_for('search') }}">
```
Bnyro: I experimented a bit with it and it seems like `method` is always set, so if I'm not missing smth we can actually drop that `or GET` part everywhere.

vojkovic: Yeah I noticed it too... TBH I just did it for completeness. I cleaned it up in the latest commit.

And if `method` were ever empty, I think HTML forms already default to GET with no `method` set so we're still fine even then.

utils/templates/etc/searxng/settings.yml
vojkovic: Fair enough... I have kept the line

Bnyro (approved):

return42 (approved):

inetol (approved):

mrpaulblack (approved):
Hi,
I have been using POST as the default on my instance. But I agree there is a lot of pain for example when reopening a closed tab on mobile or navigating back. Basically every time the form request invalidates the user gets a scary "do you want to resubmit the form" prompt.

So for me its also fine to switch to GET as the default. The only real benefit for using POST have been the edge cases when the user is behind a company proxy for example or to keep the search queries out of the browser history and it is not worth it anymore compared to the drawbacks of POST.

tiekoetter (approved):

PR #6485, opened 2026-08-03
[fix] autocompleter: suggestions contain HTML escape codes

Bnyro:
### What does this PR do?

The results were previously HTML-escaped for some reason. That doesn't really make much sense because they never got unescaped anywhere.

For example, `A&B` gets escaped to `A&amp;B` and never unescaped, so the autocompletion frontend shows `A&amp;B`, which isn't user-friendly at all.

Also, it doesn't make sense to escape the full JSON instead of only escaping the suggestion texts individually because that makes it even more unclear.

- enable an autocompleter and search for something with a `&` in it
- discovered when testing https://github.com/searxng/searxng/pull/6481

### Code of Conduct

No AI used.

### DIFF
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -851,7 +851,6 @@ def autocompleter():
         suggestions = json.dumps([sug_prefix, results, [], [], relevances])
         mimetype = 'application/x-suggestions+json'
 
-    suggestions = escape(suggestions, False)
     return Response(suggestions, mimetype=mimetype)
 
 


vojkovic (approved):
Also noticed this before, thank you for the patch...

I noticed it was added in https://github.com/searxng/searxng/pull/555 to stop what I'm guessing was a theoretical XSS or html injection? No longer became a concern after you replaced autocomplete-js last year thankfully!

LGTM

Bnyro:
> I noticed it was added in https://github.com/searxng/searxng/pull/555 to stop what I'm guessing was a theoretical XSS or html injection? 
>
I didn't find that PR (only did git blame and found a megacommit where dalf refactored everything).

Anyways it's still very strange that this escaping has been done on the server side and not the client side, I would guess that browser autocompleters don't support HTML escaped content? ...

Thanks for looking into it, merging.

PR #6490, opened 2026-08-05
[feat] engines: add jina (general)

Bnyro:
- obtain an API key from https://jina.ai/api-dashboard/ (doesn't require signup fortunately)
- set `inactive: false` and `api_key: <YOUR_API_KEY>`
- search `!ji searxng`

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/jina.py
+++ b/searx/engines/jina.py
@@ -1,0 +1,89 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Jina is a search AI and part of Elastic, the company behind ElasticSearch.
+
+The engine requires an API key, you can get one from the
+`API dashboard <https://jina.ai/api-dashboard/>`_ without signup.
+
+.. code:: yaml
+
+  - name: jina
+    engine: jina
+    shortcut: ji
+    api_key: "jina_..."
+    jina_engine: reader
+    inactive: false
+
+By default, Jina's own index is used. You can change that by setting a different :py:obj:`jina_engine`.
+"""
+
+import typing as t
+from urllib.parse import urlencode
+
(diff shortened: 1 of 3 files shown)


searx/engines/jina.py
```diff
@@ -0,0 +1,72 @@
...
+    "official_api_documentation": "https://s.jina.ai/docs",
+    "use_official_api": True,
+    "require_api_key": True,
+    "results": "JSON",
+}
+
+categories = ["general"]
+paging = True
+
+jina_engine = "reader"
```
vojkovic, line 42: Hmmm I am not really a big fan of the results from the 'reader' engine.

Image

Vs 'google' engine

Image

searx/engines/jina.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def response(resp: "SXNG_Response"):
+    res = EngineResults()
+
+    json_resp: dict[str, t.Any] = resp.json()
+
+    result: dict[str, str]
+    for result in json_resp["data"]:
+        res.add(res.types.MainResult(url=result["url"], title=result["title"], content=result["description"]))
```
vojkovic: I noticed there is also a `date` field - do you think it worth adding `publishedDate`?
Image

searx/engines/jina.py
Bnyro: Well, if we set it to the google index by default the engine is kind of useless because we already have a working google engine implementation (that doesn't require an API key).

But I agree, the results ain't perfect.

searx/engines/jina.py
Bnyro: Good point, done!

PR #6499, opened 2026-08-07
[feat] engines: add Tavily Search API engine

diogoserrano:
### What does this PR do?
Added a new search engine entry to the existing listing of SearXNG based on [Tavily Search API](https://docs.tavily.com/documentation/api-reference/endpoint/search), using brave.py as the baseline.

```sh
# static checks (as run in CI's integration workflow)
$ ./manage test.yamllint test.black test.pylint test.rst test.unit

# functional test: enable the engine in settings.yml
#   - name: tavily
#     engine: tavily
#     api_key: "tvly-..."
#     inactive: false

$ make run
# then search with: !tvy <query>
```

Without an API key the engine is marked broken at startup (expected).

### Related issues
n/a

### Code of Conduct

No AI used.


Bnyro (commented):
Please undo the changes to `searx/settings.yml` and thr two other files you probably changed by accident.

diogoserrano:
sorry about that, my IDE have some defaults after every commit and changed/removed some files.

Bnyro:
The linter is still failing with `searx/settings.yml:3466:34: [error] no new line character at the end of file (new-line-at-end-of-file)`, adding that newline to the very end of `settings.yml` should fix the CI :)

searx/engines/tavily.py
```diff
@@ -0,0 +1,159 @@
...
+
+- :py:obj:`api_key`
+
+You can obtain an API key from the `Tavily dashboard <https://app.tavily.com/home>`_.
+
+The key is configured via ``settings.yml`` only — this engine does **not** read
+any environment variable (there is no ``TAVILY_API_KEY`` fallback), in line with
+the other API-key engines in SearXNG (e.g. ``braveapi``, ``exaapi``). To keep
+the key out of the config file, mount or template your own ``settings.yml`` at
+deployment time.
```
Bnyro: I think it should be self-explaining that the only way to configure it is `settings.yml`. We never supported env configurations for any engines, and the documentation never mentions it, so this paragraph is not needed.
```suggestion
```

searx/engines/tavily.py
```diff
@@ -0,0 +1,159 @@
...
+api_key: str = ""
+"""API key for Tavily Search API (required)."""
+
+categories = ["general", "web"]
+safesearch = True
+time_range_support = True
+
+base_url = "https://api.tavily.com/search"
+
+results_per_page: int = 10
```
Bnyro:
```suggestion
page_size: int = 10
```
This is rather a nitpick, but we recently renamed `results_per_page` to `page_size` everywhere so that the configuration name is always the same no matter which engine you use.

searx/engines/tavily.py
```diff
@@ -0,0 +1,159 @@
...
+        value: Raw date string from the API
+
+    Returns:
+        Parsed datetime object or None if parsing fails
+    """
+    if not value:
+        return None
+    try:
+        return parser.parse(value)
+    except (parser.ParserError, TypeError, OverflowError):
```
Bnyro: I don't think `TypeError` and `OverflowError` can happen here? I think the other engines only catch `ParserError`

Bnyro (commented):
Very good PR, thank you 👍 

I've suggested some small improvements above. Haven't tested the changes because I was too lazy to sign up.

I read that it's possible to boost results for specific regions via Tavily's API with the `country` parameter (https://docs.tavily.com/documentation/api-reference/endpoint/search?landing=docs.tavily.com%2Fdocumentation%2Fapi-reference%2Fendpoint%2Fsearch#body-country).

In SearXNG, that could be implemented using [engine traits](https://docs.searxng.org/dev/engines/enginelib.html#module-searx.enginelib.traits) by implementing the `fetch_traits` method for the engine (see e.g. the Brave or Bing engine for a reference implementation). You don't have to implement this though, I'm also fine with merging the PR without `country` support.

searx/engines/tavily.py
```diff
@@ -0,0 +1,154 @@
...
+"""Search depth: ``basic``, ``advanced``, ``fast`` or ``ultra-fast``. Default is ``basic``."""
+
+topic: Topic = "general"
+"""Search topic: ``general``, ``news`` or ``finance``. Default is ``general``."""
+
+
+def init(_):
+    if not api_key:
+        raise SearxEngineAPIException("No API key provided")
+    if not 1 <= results_per_page <= 20:
```
Bnyro: Obviously this and the other occurences of `results_per_page` also have to be changed to `page_size`.

Also, please don't forget to change it in the docstring too!

searx/settings.yml
```diff
@@ -3471,3 +3479,4 @@ doi_resolvers:
   sci-hub.ru: 'https://sci-hub.ru/'
 
 default_doi_resolver: 'oadoi.org'
+
```
Bnyro, line 3482: Please remove that line, that should fix the CI.

Bnyro:
To generate the engine traits, run
```sh
./manage dev.env
./searxng_extra/update/update_engine_traits.py tavily
```
That should update `searx/data/engine_traits.json` with an entry for tavily.

Then you should be able to test if it works by doing search queries in different languages.

PR #6511, opened 2026-08-09
[fix] data: clear the "currencies loaded" flag when the load fails

### LINKED ISSUE #6500: Currency conversion returns no results when CurrenciesDB cache is marked loaded but empty

ryanchou1994:
### What does this PR do?

`CurrenciesDB.init()` sets the "currencies loaded" flag before it writes any row.
If `load()` fails, the flag stays set on an empty cache, and since `init()` only
loads while the flag is unset, nothing reloads it — currency conversion returns
"No results" from then on.

This clears the flag again when `load()` raises, so the next `init()` retries.
The parallel-init guard still works.

Note: `tracker_patterns.py` has the same guard but fails differently —
`iter_clear_list()` returns instead of raising, so `load()` finishes with zero
rows and the flag stays `OK`. I see exactly that on my own instance. I did not
touch it, that one needs your call.

`python -m pytest tests/unit/test_data_currencies.py -q`

Revert the `try/except` and `test_failed_load_does_not_mark_cache_as_loaded`
fails while the other one still passes. Full `tests/unit`: 341 passed, `black`
and `pylint` clean.

### Related issues

Closes: #6500

### Code of Conduct

### DIFF
--- a/searx/data/currencies.py
+++ b/searx/data/currencies.py
@@ -29,7 +29,15 @@ class CurrenciesDB:
         if self.cache.properties("currencies loaded") != "OK":
             # To avoid parallel initializations, the property is set first
             self.cache.properties.set("currencies loaded", "OK")
-            self.load()
+            try:
+                self.load()
+            except Exception:  # pylint: disable=broad-except
+                # The property is set before the rows are written, so a failed
+                # load would leave the cache flagged as loaded while holding no
+                # currency data.  Every later init() would then short-circuit
+                # and the data would stay missing for the lifetime of the DB.
+                self.cache.properties.set("currencies loaded", "")
+                raise
         # F I X M E:
         #     do we need a maintenance .. rember: database is stored
         #     in /tmp and will be rebuild during the reboot anyway
--- a/tests/unit/test_data_currencies.py
+++ b/tests/unit/test_data_currencies.py
@@ -1,0 +1,50 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+# pylint: disable=missing-class-docstring,invalid-name
+"""Tests for :py:obj:`searx.data.currencies.CurrenciesDB`."""
+
+import tempfile
+import pathlib
+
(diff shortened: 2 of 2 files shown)


Bnyro (changes requested):
The question that remains unanswered here is why `load` should fail. `load` only reads data from a pre-downloaded file that is hosted in SearXNG's code at https://github.com/searxng/searxng/blob/master/searx/data/currencies.json, so no network requests involved.

I don't want to offend you, but I don't think you understand this change and I also don't believe your changes weren't AI generated.

Issue #6514, opened 2026-08-11
Update breaks socks support?


rich0:
Searxng crashes after launch:
TypeError: AnyioProxy.__init__() got an unexpected keyword argument 'proxy_ssl'
2026-08-11 19:50:27,569 ERROR:searx.network.ipv4: Error

May be related to python-socks v3.0.0 update?

### How To Reproduce?

failing docker image 7a39e9e2610fb6f332393d0f1c69534d5f1fd4e85cb4b204cd1b1768195b8ee9
working image cd22dbf7fac8b25d5e39e65af53a4b47009df5ad3162e270c477fd5b30ce03f2 

config includes:
    outgoing:
      request_timeout: 10.0
      using_tor_proxy: true
      extra_proxy_timeout: 10
      proxies:
        all://:
          - socks5h://tor-proxy.tor-operator.svc.cluster.local:9050

### Expected behavior

Normal launch.  Config has been stable for many updates - I launch almost all of them.

### Screenshots & Logs

More complete log(error is repetitive):
2026-08-11 19:50:27,578 ERROR:searx.network.image_proxy: Error
Traceback (most recent call last):
  File "/usr/local/searxng/searx/network/network.py", line 324, in check
    await network.get_client()
  File "/usr/local/searxng/searx/network/network.py", line 211, in get_client
    if self.using_tor_proxy and not await self.check_tor_proxy(client, proxies):
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/searx/network/network.py", line 184, in check_tor_proxy
    response = await client.get("https://check.torproject.org/api/ip", timeout=60)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/searxng/.venv/lib/python3.14/site-packages/httpx/_client.py", line 1768, in get
    return await self.request(
           ^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
    )
    ^
  File "/usr/local/searxng/.venv/lib/python3.14/site-packages/httpx/_client.py", line 1540, in request
    return await self.send(request, auth=auth, follow_redirects=follow_redirects)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6515, opened 2026-08-11
Revert #6495

inetol:
Reverts searxng/searxng#6495

Related https://github.com/searxng/searxng/issues/6514

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,11 +18,11 @@ myst-parser==5.0.0
 linuxdoc==20260504
 aiounittest==1.5.0
 yamllint==1.38.0
-wlc==2.1.1
+wlc==2.1.0
 coloredlogs==15.0.1
 docutils>=0.21.2;python_version <= "3.11"
 docutils>=0.22.4; python_version > "3.11"
 parameterized==0.9.0
-granian[reload]==2.8.0
+granian[reload]==2.7.9
 basedpyright==1.39.9
 types-lxml==2026.2.16
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -1,2 +1,2 @@
-granian==2.8.0
-granian[pname]==2.8.0
+granian==2.7.9
+granian[pname]==2.7.9
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ sniffio==1.3.1
 valkey==6.1.1
 markdown-it-py==4.2.0
 msgspec==0.21.1
-typer==0.27.1
+typer==0.27.0
 isodate==0.7.2
 whitenoise==6.12.0
 typing-extensions==4.16.0


return42:
IDK why but dependandbot seems to ignore .. [httpx-socks 0.12.0](https://pypi.org/project/httpx-socks/)

https://github.com/searxng/searxng/blob/54613defc7d4cbbc1d8ec3ba269b90717eab0958/requirements.txt#L11

BTW httpx project seems more or less dead .. last release two years ago and the collaboration was shut down on questionable grounds:

- https://github.com/encode/httpx/discussions/3784

----

May we should migrate to httpx2 one day, read note at: https://httpx2.pydantic.dev/

vojkovic:
> BTW httpx project seems more or less dead .. last release two years ago and the collaboration was shut down on questionable grounds:

Wow, super strange...

> Maybe we should migrate to httpx2 one day

Httpx2 looks really good we can hopefully swap to that hopefully without many changes.

The other option I was looking at was [curl_cffi](https://github.com/lexiforest/curl_cffi), which I just saw has been tried:

- https://github.com/searxng/searxng/pull/5476
- https://github.com/searxng/searxng/pull/4801

I've recently noticed google doing some basic TLS fingerprinting so perhaps it will become necessary at some point in the future but for now, I don't know if the juice is worth the squeeze.

return42:
> Httpx2 looks really good we can hopefully swap to that hopefully without many changes.

yeah, is more or less compatible, but from my research there is no httpx-cffi as it used in 

- https://github.com/searxng/searxng/pull/5476

I think we have two options .. stay with httpx and add httpx-cffi .. what an achievable goal would be, without having to re-test proxies and sock

Or we build a new network stack directly on [curl_cffi](https://github.com/lexiforest/curl_cffi) and may look out for modern solutions like [wreq](https://python.wreq.org/en/latest/getting-started/introduction/#behavior) .. however, the proxy setup will likely change, and we will need many iterations to reach a state again that can fully satisfy even complex SearXNG setups.

vojkovic:
> but from my research there is no httpx-cffi

Yeah you are right, but to be honest it doesn't look like the hardest task to implement it ourselves: https://github.com/vgavro/httpx-curl-cffi/blob/main/httpx_curl_cffi/transport.py. I would really like to merge #5476 one day. 

However, I think I would prefer the 2nd option in the long term - wreq also looks really good too from a look through the docs and code, but curl_cffi is the larger more established project and already fits all our needs.

> the proxy setup will likely change, and we will need many iterations to reach a state again that can fully satisfy even complex SearXNG setups.

Yeah that all is true. Perhaps we will need to deprecate some settings, we are a little cursed currently by the number of features and settings we have it makes it hard to do work like this.

rich0:
You might already be aware, but I don't think the reverted commit was actually the cause here.  New builds still contain the error.  I think the CI image rebuild just pulled an updated upstream dependency and was triggered by the original commit, but after reverting it the builds still contain the problem.

rich0:
Since there was some discussion here, I'm going to cross-reference: https://github.com/searxng/searxng/pull/6534

PR #6516, opened 2026-08-12
[fix] dogpile: access denied due to missing origin header

Bnyro:
### Code of Conduct

No AI used.


Bnyro:
Maybe we should also set the engine to `inactive: true` so that it must be manually enabled. It seems like they adapted their bot-blocking to block SearXNG requests, so it could benefit from not being used by all these AI scrapers that use SearXNG.

return42 (approved):
LGTM

PR #4506, opened 2025-03-16
[feat] plugins: Implement Quick Answer search-augmented LLM summary plugin (via OpenRouter)

tempname1024:
## What does this PR do?
Implements search-augmented OpenRouter-based LLM summaries with inline citations to SearXNG engine results; see [Mojeek summaries](https://blog.mojeek.com/2024/04/mojeek-search-summary.html), [Kagi Quick Answer](https://help.kagi.com/kagi/ai/quick-answer.html).

This plugin supports [OpenRouter](https://openrouter.ai/), a unified interface and API platform that provides access to hundreds of LLMs, and allows users to configure their own access tokens, providers, and models:
- The `/quick_answer` route asynchronously proxies queries built in the post_search context between the client and OpenAI-compatible API; no client-side cross-origin requests.
- KaTeX and Marked libraries used to support formatted rendering of markdown and mathematical expressions.

https://github.com/user-attachments/assets/71064233-c3bb-4d27-a3b2-4de8e6fb6e45

## Why is this change important?
Search engines have incorporated large language models as a means to surface information of relevance to the user's query and to reduce the frequency of hallucination by constraining LLM contexts to search results. However, these systems are generally black boxes; Kagi et al. do not disclose the models upon which their services depend, and the context content is not visible to users. This plugin provides a user-centered alternative to otherwise opaque, privacy-compromising services.

## How to test this PR locally?
1. Preferences → Enable Quick Answer and configure your access token, model, and LLM providers.
2. Append an `?` to the end of your search queries to invoke Quick Answer.

## Author's checklist
1. The `results-sidebar-width` is currently pretty narrow; I'd bumped it to `35rem` for the video above. Updating that value is very likely beyond the scope of this PR / should be voted on, but thought I'd mention more space there is probably needed to contain longer-form responses.

## Related issues
Closes #4097

Issue #4097, opened 2024-12-07
[feature request] ai summerizer


celevra:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
it would be great if we could integrate ollama or openai or whatever to summerize the search request with the first 3 search resultes returned by searxNG, like how brave search it does

![image](https://github.com/user-attachments/assets/f6862301-d982-459a-a3b8-5c40239b1e44)

neon-mmd:
I think this issue is related to #2163. What do you say? :sweat_smile:

LyndonArmitage:
If this is implemented it will definitely need to be an optional feature as plenty of people probably won't want to use an LLM due to cost, privacy or ideological reasons.

Implementation wise, from what I recall Searxng doesn't parse the whole HTML of a search result, and just presents a summary from the engine it is using. So would this AI summary be created using that limited information or use the full page contents? Alternatively, it could hand the page URL to the LLM to summarise, assuming it can crawl a web page.

celevra:
i'm using open web ui to interact with ollama. In open web ui it is possible to activate websearch and send requests to searx, here is the wiki:
https://docs.openwebui.com/features/web_search/

that means it is allready possible (or code exists) to feed ollama with the help of the api from searx.
So we just need the way back into the searx frontend

celevra:
if it helps i can provide access to a server running ollama, open-web-ui and searxNG

tempname1024:
I'd written a quick implementation using the OpenAI schema (via OpenRouter), KaTeX, and marked to support Quick Answer functionality and replace my use of Kagi. OpenRouter requests are mediated by SearXNG and routed through a /quick_answer route to be streamed back to the client and rendered with search-supported inline citations and a references section. Users provide and store their own token via preferences. Quick Answer is triggered by ending a query in a "?".

If there's interest and this kind of feature isn't orthogonal to the goals of SearXNG, I'm happy to rework the plugin as needed to upstream the changes.

![2024-12-23_23-55](https://github.com/user-attachments/assets/db47b917-fc7d-4b00-833f-37df39ec1c35)

celevra:
Yes that looks great! Is it possible to create a setting where the question mark isnt needed?

tempname1024:
> Yes that looks great! Is it possible to create a setting where the question mark isnt needed?

Yes! An engine-variety !qa bang and toggle to run on every query would probably be more appropriate than the question mark route.

return42:
[...]

PR #6539, opened 2026-08-16
[feat] plugin: AI summary of search results from a self-hosted LLM

jasonwitty:
### What does this PR do?

A SearXNG plugin similar to Brave search and Google search that summarize the first 5 search results using a **self-hosted LLM**. With the release of Gemma models you can host your own LLM on reasonable hardware and get good results. Integrating with a local LLM fits the project's philosophy of privacy and control. This also is a great homelab project for anyone that wants to get experience running local models and also is running SearXNG.

Like Brave and Google, the AI summary only displays when an infobox does not exist. (wikipedia) The AI summary is loaded asynchronously not to block any search results while displayer a cursor animation and is presented prominently above the results.

Compatible with any LLM server that speaks the OpenAI chat completions API. (Ollama, vLLM, llama.cpp, LM Studio, TGI, liteLLM gateway - with authentication)

1-desktop

2-mobile

### How it works

When a search is performed, an empty placeholder is placed ahead of the results when an infobox does not exist, the browser then posts to /ai_summary, which opens a streaming request to the LLM server and re-emits tokens as NDJSON. Skipped on page 2 and beyond, outside general, on non-HTML formats, and when an infobox or instant answer already exists.

AI summary box initially appears with a cursor animation and is replaced with the AI summary when the request completes. Set to only display at a minimum height to avoid covering search results with a more button to expand. Additional text box appears to allow you to ask a follow up question when more is clicked. (Basically a clone of how this feature works on brave search)

The client code side follows the existing plugin pattern — one file in client/simple/src/js/plugin/, one conditional load in router.ts matching calculator and infinite-scroll, one LESS import. No build-config changes and no additional npm dependencies.

Disabled by default, enabled via configuration. Once enabled, custom setting page will be available to allow users to configure the LLM server URL and other settings. Settings can be locked to avoid SSRF vulnerabilities on public instances of SearXNG.

Tested / Verified - Tested with my own Gemma model running locally over several weeks with both direct ollama and LiteLLM router configurations. Verified on mobile, web, chromium, brave, firefox.

Tests — 43 covering the endpoint, validation, key scoping, retries and preference handling.

Documentation provided including a brief quick start including instructions to host your own LLM on a local server.

Add settings YAML to your instance's `settings.yml` file:

```yaml
ai_summary:
  base_url: "http://127.0.0.1:11434"
  model: "gemma4:e4b"
  grounding: true
```

Follow quick start instructions provided in the AI Summary plugin documentation.

#### Addressing PR #4506

I am aware of the PR 4506 that takes a different approach to resolving #4097. The original request does not state how or what type of LLM server will handle the summarization but it was my intention to host my own LLM on a local server. This obviously requires a little more than a raspberry pi but you would be surprised how good the newer open source models are. Personally I am running the model used by my local SearXNG instance on an A2000 which is only a 75w / 6G GPU and it works great. PR 4506 takes a different approach by integrating with open router, which is fine if thats what you want. I personally think both plugins can live side by side, literally as the placement of boxes will not conflict with each other. If someone wants to use a third party service to summarize results they can use that one and this plugin is specifically designed to work with local LLM servers.

Local LLM Solution to issue:  #4097 - [feature request] ai summerizer

### AI Disclosure

Claude code (cli) / Zed Text Editor
Opus 5 - Coding tasks, testing, assistance with documentation.
Fable 5 - Security and code review. 


return42:
> [x] **I hereby confirm that this PR conforms with the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst).**

Sorry, you didn't .. Unfortunately, you consistently ignore every rule mentioned there.

PR #6541, opened 2026-08-16
[feat] engines: add iconify (icons)

Bnyro:
### What does this PR do?
- add support for <https://icon-sets.iconify.design>
- iconify also searches in `devicons`, `material icons` and `selfh.st/icons`, so in theory we could remove them (it makes no sense to remove them though while they still work, but it might be easier to remove than to fix one of these engines if they break in the future)

### Known quirks
- sometimes, the icon size we set isn't right and cuts off parts of the SVG, see e.g. by searching `!ico rust`
- the engine requires a lot of HTTP requests, but at least from my location, the engine is extremely fast nevertheless

- !ico test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/iconify.py
+++ b/searx/engines/iconify.py
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Iconify aggregates icons from different open source icon sets."""
+
+import codecs
+import typing as t
+from urllib.parse import urlencode
+
+from searx.network import get
+from searx.result_types import EngineResults
+
+if t.TYPE_CHECKING:
+    from extended_types import SXNG_Response
+    from search.processors.online import OnlineParams
+
+
+about = {
+    "website": "https://iconify.design",
+    "wikidata_id": None,
+    "official_api_documentation": None,
+    "use_official_api": True,
+    "results": "JSON",
+}
+
+categories = ["images", "icons"]
+paging = True
+
+base_url = "https://api.iconify.design"
+page_size = 20
+
+
+def request(query: str, params: "OnlineParams"):
+    # actual number of results isn't exaxctly the same as the page size, only approximately
+    args = {"query": query, "start": (params["pageno"] - 1) * page_size, "limit": page_size}
+    params["url"] = f"{base_url}/search?{urlencode(args)}"
+
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    res = EngineResults()
+
+    # we request all icons from the same icon set with one request
+    icon_sources: dict[str, list[str]] = {}
+    for icon in resp.json()["icons"]:
+        icon: str
+        icon_source, icon_name = icon.split(":", 1)
+        icon_sources[icon_source] = icon_sources.get(icon_source, []) + [icon_name]
+
+    for source, names in icon_sources.items():
+        icon_resp = get(f"{base_url}/{source}.json?icons={','.join(names)}").json()
+
+        icon_name: str
+        icon_info: dict[str, str]
+        for icon_name, icon_info in icon_resp["icons"].items():
+            w = icon_resp.get("width", 24)
+            h = icon_resp.get("height", 24)
+            icon_svg = f'<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 {w} {h}">{icon_info["body"]}</svg>'
+            b64_icon = codecs.encode(icon_svg.encode(), encoding="base64").decode("utf-8")
+            res.add(
+                res.types.Image(
+                    url=f"{base_url}/{source}.json?icons={icon_name}",
+                    title=icon_name,
+                    img_src=f"data:image/svg+xml;base64,{b64_icon}",
+                )
+            )
+
+    return res
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1344,6 +1344,11 @@ engines:
       require_api_key: false
       results: JSON
 
+  - name: iconify
+    engine: iconify
+    shortcut: ico
+    disabled: true
+
   - name: iseek
     engine: iseek
     shortcut: isk


searx/engines/iconify.py
```diff
@@ -0,0 +1,65 @@
...
+        icon_name: str
+        icon_info: dict[str, str]
+        for icon_name, icon_info in icon_resp["icons"].items():
+            w = icon_resp.get("width", 24)
+            h = icon_resp.get("height", 24)
+            icon_svg = f'<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 {w} {h}">{icon_info["body"]}</svg>'
+            b64_icon = codecs.encode(icon_svg.encode(), encoding="base64").decode("utf-8")
+            res.add(
+                res.types.Image(
+                    url=f"{base_url}/{source}.json?icons={icon_name}",
```
return42: This "URL" points to a JSON response .. 

Image


I tried ``f"https://icon-sets.iconify.design/?query={icon_name}"`` but it did not work either.

The SVG link isn't a valid href :

```
<a rel="noreferrer" href="data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My...">
SVG
</a>
```

return42 (approved):
Beside my remark about the links in the UI, the engine LGTM 👍 

FYI: I rebased branch and modified the commit message.

---

OT: [Maybe we should think about an alternative icon pool. There are two things that bother me about IonIcons:](https://github.com/searxng/searxng/pull/6759#issuecomment-5751235814) .. 

I took a few random samples from Iconify; the selection and quality seem significantly better to me than what I'm used to from IonIcons... Maybe Iconify is a candidate for a new icon pool?

searx/engines/iconify.py
Bnyro: I finally found the API docs at https://iconify.design/docs/api/queries.html - this allows us to use proper icon links - idk why I didn't find that earlier ... :)

Bnyro:
Thanks for the review!

> I took a few random samples from Iconify; the selection and quality seem significantly better to me than what I'm used to from IonIcons... Maybe Iconify is a candidate for a new icon pool?
>
Iconify sources icons from 200+ different icon sets. So it probably also contains all IonIcons, among others. So if we use it as our icon set, we basically have 200+ different icon styles in SearXNG. Not sure if that's so good :)

return42:
> Iconify sources icons from 200+ different icon sets. So it probably also contains all IonIcons, among others. 

Yes, [contains all IonIcons](https://github.com/iconify/icon-sets/blob/master/collections.md#ionicons) https://icon-sets.iconify.design/ion/ .. but for unknown reasons, the quality is mutch better, for instance the [radio-outline.svg](https://github.com/searxng/searxng/blob/master/client/simple/src/svg/ionicons/radio-outline.svg?short_path=f03a32f) we had to fix manually by setting the ``fill="currentColor"`` for the `<cycle />` element:

	<circle cx="256" cy="256.02" r="32" fill="currentColor" />

Compare: in the origin the fill attribute isn't set

https://github.com/ionic-team/ionicons/blob/main/src/svg/radio-outline.svg?short_path=1d9c031

but is already set in the identical icon from iconify  -> https://icon-sets.iconify.design/ion/radio-outline/

> So if we use it as our icon set

Even though I can understand the argument about the >220 mixed styles, I still see two advantages... and wouldn't want to completely dismiss the idea.

- The quality is better than the original (!)
- There is an API that could potentially be used in the build process (it’s not required, but it might offer some advantages?)
- If an icon in the standard style (IonIcon) is missing, there are alternatives in other styles with appropriate licenses that can be used instead.

> we basically have 200+ different icon styles in SearXNG. Not sure if that's so good :)

If we try to stick to one style (IonIcon) and ensure the alternative icons we need to source from other sets fits, then we should be able to avoid a optical collision  of styles.

Strictly speaking, that's exactly what we're already doing manually today... and iconify could free us from having to manually edit the SVGs

Bnyro:
> Yes, [contains all IonIcons](https://github.com/iconify/icon-sets/blob/master/collections.md#ionicons) https://icon-sets.iconify.design/ion/ .. but for unknown reasons, the quality is mutch better, for instance the [radio-outline.svg](https://github.com/searxng/searxng/blob/master/client/simple/src/svg/ionicons/radio-outline.svg?short_path=f03a32f) we had to fix manually by setting the `fill="currentColor"` for the `<cycle />` element:
> 
> ```
> <circle cx="256" cy="256.02" r="32" fill="currentColor" />
> ```
> 
> Compare: in the origin the fill attribute isn't set
> 
> https://github.com/ionic-team/ionicons/blob/main/src/svg/radio-outline.svg?short_path=1d9c031
> 
> but is already set in the identical icon from iconify -> https://icon-sets.iconify.design/ion/radio-outline/
> 
That's funny, I didn't expect that they fix the icons from other icon sets :)

> Even though I can understand the argument about the >220 mixed styles, I still see two advantages... and wouldn't want to completely dismiss the idea.
> 
> * The quality is better than the original (!)
> * There is an API that could potentially be used in the build process (it’s not required, but it might offer some advantages?)
> 
> * If an icon in the standard style (IonIcon) is missing, there are alternatives in other styles with appropriate licenses that can be used instead.
> 
> If we try to stick to one style (IonIcon) and ensure the alternative icons we need to source from other sets fits, then we should be able to avoid a optical collision of styles.
> 
> Strictly speaking, that's exactly what we're already doing manually today... and iconify could free us from having to manually edit the SVGs
>
Good points. I think they even have a JavaScript library, so we could in theory also use a similar workflow to IonIcons, i.e. import the JavaScript library and take the icons from it. I don't have a real preference on changing the icon library, we should do whatever is the easiest to maintain, and I agree that manually sourcing icons because IonIcons don't work does not really make maintenance easy.

PS: I'm personally always just using the Material Icons icon catalog from Google (because I'm mostly working on Android apps, so it makes sense to use them because then they then have the same style as the other icons used in Android). So I don't have any overview about which icon sets are good and which ain't.

Issue #644, opened 2021-12-22
Disable devianart when SafeSearch is enabled


not-my-profile:
While deviantart  [lets its users tag content as Mature Content](https://www.deviantartsupport.com/en/article/what-is-mature-content-453216) of course many users don't do that. This is problematic because it results in SearXNG users that enable SafeSearch getting shown unsolicited NSFW images in the image search.

This can currently be reproduced with `master` with:

1. `make run`
2. set SafeSearch to strict
3. perform an image search for `test`

The best course of action for us seems to be to categorize devianart as NSFW and automatically disable all search engines categorized as NSFW when a user enables SafeSearch.

dalf:
( There are multiple issues in searx about this subject, one has this idea )

There 3 types of engines:
* engines with a safe search filter and supported by SearXNG
* engines with unsafe content without filter supported by SearXNG (either because the filter doesn't exist, or because SearXNG doesn't support this filter).
* engines with safe content (for example: https://en.wikimini.org/ ).

So the `safesearch` attribute is not enough:
https://github.com/searxng/searxng/blob/fe751b0ee8af955390ed4caf56e2c9b267f5caa3/searx/engines/google.py#L47

Two mutually exclusive solutions I can think of:
* `safesearch` can have three values: filter, unsafe, safe.
* add a new attribute `safe_content`, `False` by default, it has to be explicitly set to `True`. I prefer this solution because the content is explicitly described as safe.

From one of these solutions (or another), if the users make a query with a safe search filter, the unsafe engines can be skipped.

Would it solve the issue ?

not-my-profile:
If there can be three states I think we should have an enum with three variants instead of two booleans. How about:

* `content_safety: safe` the content is always safe
* `content_safety: unsafe` the content may be unsafe (but there is no safesearch support)
* `content_safety: configurable` the content may be unsafe (but there is safesearch support)

So for example: Google would have `configurable`, openstreetmap would have `safe` and devianart would have `unsafe`.

return42:
`content_safety` .. why not using the existing `safesearch`

    safesearch = [filter|save|unsave]

not-my-profile:
`safesearch: unsafe` doesn't make sense to me semantically ... either an engine supports safesearch or it's unsafe ... not both.

return42:
ATM the meaning of `safesearch` attribute is; 

- `True` : engine (aka module) supports a SafeSearch filter
- `False`: engine (aka module) does not support a SafeSearch filter

[...]

PR #6269, opened 2026-06-16
[feat] image results: automatically guess mimetype based on path

Bnyro:
### What does this PR do?
- In most cases the search engine doesn't need to support getting the image type (e.g. `JPEG`), as that information is already contained in the URL path
- i.e., we take the `img_src` or `thumbnail_src` URL path to automatically guess the mimetype of the image using the `mimetypes` stdlib module
- that avoids that we always show "original format" instead of the actual image format

### How to test this PR locally?

Activate for instance the heexy images engine ..

```diff
diff --git a/searx/settings.yml b/searx/settings.yml
index a509d94d9..78195ace4 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -1258,7 +1258,6 @@ engines:
     heexy_categ: image
     shortcut: hei
     disabled: true
-    inactive: true
```

and search for `!hei bmw`

### Related issues
- #6153 

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [X] **I hereby confirm that this PR conforms with the [AI Policy].**

No AI used.

PR #6542, opened 2026-08-17
[fix] engines: deviantart access denied

vojkovic:
### What does this PR do?

Fix access denied errors coming from deviantart - the other fix is a medium severity SSRF:
```
https://searxng.com/search?q=test&pageno=2&engines=deviantart&engine_data-deviantart-nextpage=https://example.com
```

Currently there is no url validation in `nextpage_url`, so I replaced it with just a token in the `cursor` arg instead of passing the whole url.

Results are still poor (see https://github.com/searxng/searxng/issues/644), so the engine has been set to inactive. The blurred result removal is working now but still a lot of results are nsfw because they aren't being tagged correctly 🤷

### Code of Conduct

Cursor found the SSRF in a vuln scan and then assisted in fixing the engine

### DIFF
--- a/searx/engines/deviantart.py
+++ b/searx/engines/deviantart.py
@@ -1,10 +1,17 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Deviantart (Images)"""
 
+import typing as t
+
 import urllib.parse
 from lxml import html
 
+from searx.result_types import EngineResults
 from searx.utils import extract_text, eval_xpath, eval_xpath_list
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
 
 # about
 about = {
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Thank you for looking into this! The engine has been on my list for a while but I kinda forgot about it.

I've pushed a commit on top of yours to additionally
- extract the `author` field
- migrate to `EngineResults` - that has the neat advantage that SearXNG now also displays the format of the image automatically without us extracting it (see https://github.com/searxng/searxng/pull/6269).

Please look into them if you find time. You can squash them into one commit if you want, or just keep them two, as you prefer.

Your changes lgtm 👍

vojkovic:
Thanks for modernising it :)

I have fixed the author xpath, sometimes it got unlucky and added the avatar alt ([diff](https://github.com/searxng/searxng/compare/c3cd3692379c004f8ba152e90fb1405e500065e5..3e0c22c3009fd318df657ef571ef2d1cdc28f029)):

image

Will merge soon

PR #6546, opened 2026-08-18
[fix] google: use Nokia UA

### LINKED ISSUE #3277: Bug: google news
Whatever the instances, the google news is broken.

The request is redirected to https://consent.google.com
As far I remember it requires one cookie value per outgoing IP to fix the issue (if this is the case --> related to #2685 SAME_HTTP_CLIENT )

### LINKED ISSUE #6359: GSA for iPhone useragent do no longer work

vojkovic:
### What does this PR do?

- Uses the nokia user-agents discussed in https://github.com/searxng/searxng/issues/6359#issuecomment-5114886182 to allow google search w/o javascript.
- Removes the `lr` and `cr` params from google engines. My understanding of these params is they only ever reduce the number of results, so they tend to leave only the bad quality results unless the locale is something generic like en.
- Removes the google country specific tlds. [According to google](https://blog.google/products-and-platforms/products/search/country-code-top-level-domains/) this hasn't affected search results in almost a decade and it just redirects to the main .com anyways since 2025.
- Modernising the engine to use typed results.
- Also fixes google news and google image & video search. Google news now uses the search 'news' tab instead of news.google.com.
- Enables all these engines, are we happy with having both CSE and google enabled at the same time?

In the last 12h, my instance has pushed 36k requests at a 100% success rate so this engine is very reliable.

image

Test all engines work:

```
!go searxng
!goi searxng
!gon searxng
!gov searxng
```

### Related issues

- Closes: https://github.com/searxng/searxng/issues/6359
- Closes: https://github.com/searxng/searxng/issues/3277

### Code of Conduct

Cursor helped with the implementation


Bnyro (commented):
Very nice! I don't have time to exhaustively review it yet, so only one quick comment.

> Enables all these engines, are we happy with having both CSE and google enabled at the same time?
>
Having `google cse (images)` and `google (images)` enabled at the same time likely only causes duplicated results, so there's no real benefit in it I think. We should probably pick the one that's more reliable (which is likely Google CSE because they likely won't change their API until it's shut down?).

searx/engines/google.py
```diff
@@ -300,130 +243,96 @@ def detect_google_sorry(resp: "SXNG_Response"):
...
+def wml_dom(resp: "SXNG_Response"):
+    detect_google_sorry(resp)
+    text = resp.text
+    if text.lstrip().startswith("<?xml"):
+        text = text.split("?>", 1)[-1]
+    return html.fromstring(text)
+
+
+# Google search request
+def request(
```
Bnyro: Please move this into a `google_request` helper method.

The `request` method is defined in https://github.com/searxng/searxng/blob/5ffd32ca2f5e3648bcaa61ad18b94f258852edcb/searx/enginelib/__init__.py#L409 and is only supposed to take 2 arguments, anything else is very confusing / hackish.

searx/engines/google.py
```diff
@@ -300,130 +243,96 @@ def detect_google_sorry(resp: "SXNG_Response"):
...
+    if start:
+        args["start"] = start
 
-def parse_url_images(text: str):
-    data_image_map = {}
+    query_url = f"https://www.google.com/wml/search?{urlencode(args)}"
+    if use_time_range and params["time_range"] in time_range_dict:
+        query_url += "&" + urlencode({"tbs": "qdr:" + time_range_dict[params["time_range"]]})
+    if use_safesearch and params["safesearch"]:
+        query_url += "&" + urlencode({"safe": (safesearch_map or filter_mapping)[params["safesearch"]]})
```
Bnyro, line 327: Why do you concatenate this into the query string instead of putting it into `args` before `query_url` is built?

searx/engines/google_images.py
```diff
@@ -1,40 +1,31 @@
...
-    "website": 'https://images.google.com',
-    "wikidata_id": 'Q521550',
-    "official_api_documentation": 'https://developers.google.com/custom-search',
+    "website": "https://images.google.com",
+    "wikidata_id": "Q521550",
+    "official_api_documentation": "https://developers.google.com/custom-search",
     "use_official_api": False,
     "require_api_key": False,
-    "results": 'JSON',
+    "results": "HTML",
```
Bnyro: Shouldn't this be `XML`? Other engines use this too as results type (see https://github.com/search?q=repo%3Asearxng%2Fsearxng+%22results%22%3A+%22xml%22&type=code).

Same for the other google engines, they should also be `XML`.

searx/engines/google_news.py
```diff
@@ -64,261 +32,58 @@
 
 # engine dependent config
 categories = ["news"]
-paging = False
+paging = True
+max_page = 50
+"""`Google max 50 pages`_
+
+.. _Google max 50 pages: https://github.com/searxng/searxng/issues/2982
```
Bnyro:
```suggestion
"""Google supports up to page 50, see the `Google max page discussion`_.

.. _Google max page discussion: https://github.com/searxng/searxng/issues/2982
```

searx/engines/google_images.py
```diff
@@ -1,40 +1,31 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""This is the implementation of the Google Images engine using the internal
-Google API used by the Google Go Android app.
+"""Google Images"""
```
Bnyro: Might be worth to write in all these docstrings that the images/videos/news engine implementation is based on `google.py`.

searx/engines/google.py
Bnyro, line 11:
```suggestion

This implementation uses the Nokia user agents to request an XML layout from Google. The normal web version requires executing JavaScript to load the results and therefore is currently not used here. See `Google discussion`_ for more information on that topic.

_Google discussion: https://github.com/searxng/searxng/issues/6359
```

Bnyro (commented):
Works very well, didn't run into any ratelimits when testing 👍

Very cool, let's hope it will continue to work for a while :)

Left some comments above.

searx/engines/google.py
vojkovic: Ha good question... that's how it was done before but I will fix it to use `args`

vojkovic:
Thanks for your thorough review @Bnyro, I appreciate it:

> Having google cse (images) and google (images) enabled at the same time likely only causes duplicated results, so there's no real benefit in it I think. We should probably pick the one that's more reliable

I disabled google and google images engine in settings.yml, from some testing the results aren't 1:1 but they are broadly quite similar...

Also I tidied the docstrings and removed some superfluous ones.

Bnyro (approved):

return42 (approved):
> * Removes the `lr` and `cr` params from google engines. My understanding of these params is they only ever reduce the number of results, so they tend to leave only the bad quality results unless the locale is something generic like en.

I think these parameters should not be deleted from the `get_google_info` function.

They are needed to ensure the regionality of the search results. They are not needed for image or video search, but users expect regionality in web search.

I have placed a WIP commit on your branch, which includes the new parameter `use_locales` that is set to `False` for images and videos.

To check the regionality of an engine, I, for example, look for the branches of large manufacturers such as BMW in Taiwan.

As a user from TW who is looking for BMW, I would expect that a query like `!go
BMW :zh-TW` would show me the branch in Taiwan at the top of the results.

Here the comparison ..

#### `use_locales=False`

My local IP is located in de-DE, so google gives me results related to my IP (not to my query)

Image

#### `use_locales=True`

Image

If you agree with the changes from WIP, please squash them into your commit

-----

> Removes the google country specific tlds. 

👍 .. I've added another commit on top of your branch to update the traits

    [mod] data: update searx.data - update_engine_traits.py

vojkovic:
> I think these parameters should not be deleted from the get_google_info function.

On closer inspection I think you are correct. I really only tested the subvariants of 'en', i.e. en-AU, en-GB etc... I agree the results in other languages are much better.

> They are not needed for image or video search, but users expect regionality in web search.

In this case, I don't understand why they are not wanted for image and video search. Do users not also want regionality here? 

EDIT: Sort of offtopic but I also wonder if the `uule` param can be of use to us. I will do some tests later when I have a bit of time.

https://docs.brightdata.com/scraping-automation/serp-api/query-parameters/google#uule

return42:
> In this case, I don't understand why they are not wanted for image and video search. Do users not also want regionality here?

TL; DR; .. We can do it however you think best... here are my thoughts on it:

This is merely an assumption on my part: when I search for a video, I’m less interested in its origin than in its content; the same applies to searches for articles, images, software, etc. (without `:zh-TW`).

However, when I search for a service, a repair shop, a store, or other real-world entities, I want regional relevance (with `:zh-TW`).

But of course, you’re right—we could leave it up to the user to decide individually whether or not to opt for a regional search (with or without `:zh-TW`).

That’s where my second assumption comes into play: there will be less tech-savvy users who aren't aware of this distinction but have a region like `zh-TW` set as a default in their cookie settings... and who might then wonder why they’re getting so few results in image searches (to pick up on your argument).

But if someone searches for articles (on the web) with a region preset, they’ll face the same issue of limited results... either way—no matter what I assume—these remain assumptions that are open to debate.

By the way, I’d therefore recommend not setting anything in the preferences; that way, only the language setting applies :-)

vojkovic:
Awesome, thanks for clarifying - the longer I think about it, I think your `use_locales` is the correct approach --> I squashed the WIP into my commit.

PR #6551, opened 2026-08-19
[fix] calculator: XSS from innerHTML

vojkovic:
### What does this PR do?

The calculator plugin is currently using `innerHTML`. math.js will happily parse a quoted string so a query like:

```
/search?q=%22%3Cimg%20src=x%20onerror=alert(1)%3E%22
```

executes as XSS, although this is blocked by the CSP on most public instances.

Also fixes a bug where if you paste any quoted string like "searxng" it will respond with `"searxng" = searxng` in the answer box by checking it is a number.

Verify the XSS no longer exists and calculator still works as before

### Code of Conduct

Cursor found XSS in security audit and assisted with fix


inetol (approved):
Ouch.

*Infinity* must also be considered as a valid output value (`3 !!! 2`). I've applied the change already..: https://github.com/searxng/searxng/compare/840503a07902faab886ac957f9bfbf972e6a8563..9d3481c15290709c1fe833b68454f3e99c7c2ee5

vojkovic:
Fair enough, I am happy if you want to merge now.

I noticed some weird things with Infinity like "log(0)" but it's prob not worth changing to be honest.

Screenshot

client/simple/src/js/util/appendAnswerElement.ts
```diff
@@ -23,7 +23,7 @@ export const appendAnswerElement = (element: HTMLElement | string | number): voi
 
   if (!(element instanceof HTMLElement)) {
     const span = document.createElement("span");
-    span.innerHTML = element.toString();
+    span.textContent = element.toString();
```
Bnyro, line 26: I didn't know that `textContent` exists, I've always used `innerText`. Good to know! (For interested readers: https://stack.chatoyer.de/exchange/stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext)

PR #6554, opened 2026-08-19
[fix] preferences: cap zlib decompress

vojkovic:
### What does this PR do?

Caps the zlib decompression to ~~64KB~~ 16KB. A small blob currently could expand to be like hundreds of MB on one request which would effectively be a DOS attack against an instance. ~~64KB is much more than any real request, honestly it could probably be lower like 16KB but this gives some future room.~~

Also invalid blobs also returned 500 because that path wasn't in a try/except like the cookies are.

Request `/?preferences=foobar`, it should not return a 500 anymore.

### Code of Conduct

Cursor found this during a vulnerability scan

### DIFF
--- a/searx/preferences.py
+++ b/searx/preferences.py
@@ -6,7 +6,7 @@
 import typing as t
 
 from base64 import urlsafe_b64encode, urlsafe_b64decode
-from zlib import compress, decompress
+from zlib import compress, decompressobj
 from urllib.parse import parse_qs, urlencode
 from collections import OrderedDict
 from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Yet another very good security patch 👍 

I just checked it. When modifying all settings, I get `3782B` in size for the query string. So I would be fine with lowering the limit down to even 8 or 16KB. But 64KB is also fine, still small enough to make it hard enough to DoS the server using that route.

Can confirm this fixes the internal server error when visiting `/?preferences=...`, so lgtm!

vojkovic:
TBH I didn't mention it earlier but it was still pretty bad at 64KB with my POC. I lowered it to 16KB... I would go lower but I suppose engine tokens can be any length and I don't want to interfere with that.

It should still give us plenty of room to not have to think about it. 😛

inetol (approved):

PR #6556, opened 2026-08-19
[fix] py: (un)trusted proxies

inetol:
### What does this PR do?

A missing condition allowed an attacker to mask their IP address from the server. This exploit can only be carried under specific client & server environments.

Report details on our security mailbox or Matrix thread (`$St5ADwTOP1wK3reRmmEEAC_KroIXoWxyw_TSUM4zaJM`)

### Code of Conduct

* Ministral 3, Initial report brainstorm and validation helper

### DIFF
--- a/searx/botdetection/trusted_proxies.py
+++ b/searx/botdetection/trusted_proxies.py
@@ -63,6 +63,20 @@ class ProxyFix:
         proxy_list: list[str] = cfg.get("botdetection.trusted_proxies", default=[])
         return [ip_network(net, strict=False) for net in proxy_list]
 
+    def is_trusted_proxy(
+        self,
+        addr: IPv4Address | IPv6Address | None,
+        trusted_proxies: list[IPv4Network | IPv6Network],
+    ) -> bool:
+        if addr is None:
+            return False
+
+        for net in trusted_proxies:
+            if addr.version == net.version and addr in net:
+                return True
+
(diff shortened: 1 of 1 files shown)


searx/botdetection/trusted_proxies.py
```diff
@@ -95,19 +101,21 @@ def __call__(self, environ: "WSGIEnvironment", start_response: "StartResponse")
...
-                    addr = addr.ipv4_mapped
-                orig_remote_addr = addr.compressed
+                orig_remote_ip = ip_address(orig_remote_addr)
+                if orig_remote_ip.version == 6 and orig_remote_ip.ipv4_mapped:
+                    orig_remote_ip = orig_remote_ip.ipv4_mapped
+                orig_remote_addr = orig_remote_ip.compressed
             except ValueError as exc:
                 logger.error("REMOTE_ADDR: %s / discard REMOTE_ADDR from WSGI environment", exc)
                 orig_remote_addr = None
+                orig_remote_ip = None
```
Bnyro, line 118: Is it really needed to set `orig_remote_ip` to `None` here? I think the only statement here that can fail is the call to `ip_address(...)`, so when it crashes at that point, `orig_remote_ip` is still `None` from its declaration. But doesn't hurt to have it here anyways.

Bnyro (approved):

PR #6571, opened 2026-08-23
[mod] network: migrate to httpx2

vojkovic:
### What does this PR do?

Switches the http client from the unmaintained httpx and httpx-socks to [httpx2](https://github.com/pydantic/httpx2)

SOCKS is now built in so we don't need to use httpx-socks / python-socks anymore. However, SOCKS4 is not supported by httpx2, so it won't work anymore (only supports SOCKS5/SOCKS5h).

Also the TLS verification in httpx2 uses the OS trust store instead of certifi, so that's just something to keep in mind.

Test and run. Do some searches, if you have proxies check they still work.

### Related issues

- https://github.com/searxng/searxng/pull/6515
- https://github.com/searxng/searxng/issues/6514

### Code of Conduct

Cursor helped with the migration to check I didn't miss anything


vojkovic:
The other option is that we skip httpx2 and just go straight to curl-cffi. I have started a bit of testing on my `cffi` branch -> see https://github.com/vojkovic/searxng/commit/23eefeeaf7d4d91a5345d1d688dddf014eee00f9

What do you think?

return42:
> What do you think?

:+1: 

Since we would have to test all network options anyway, a direct switch makes more sense in my opinion than upgrading to httpx2; if we decided to switch later, we would have to test everything all over again.

I don't think we gain much with httpx2 (other than becoming updateable again). We might want to consider using curl (cffi) or switching straight to [wreq](https://docs.rs/wreq/latest/wreq/); there is a Python wrapper for it --> https://python.wreq.org/en/latest/getting-started/introduction/

vojkovic:
I will open another PR for curl-cffi when it is ready. :+1:

> We might want to consider using curl (cffi) or switching straight to [wreq](https://docs.rs/wreq/latest/wreq/)

I looked at wreq and it doesn't support http3/quic yet: https://github.com/0x676e67/wreq/tree/http3 
- https://github.com/0x676e67/wreq-python/discussions/306

I think the ecosystem for rust h3 is not ready yet, curl supports everything and has been around forever

return42:
Do we have http3 use cases? AFAIK just google supports http3 (enum 30 in the example below).

```python
>>> from curl_cffi import requests  
... from curl_cffi.const import CurlHttpVersion
... 
... response = requests.get(
...     "https://www.duckduckgo.com", 
...     http_version=CurlHttpVersion.V3
... )  
... print(response.http_version)
... 
3    # is V2_0
>>> CurlHttpVersion.V2_0
<CurlHttpVersion.V2_0: 3>

>>> response = requests.get(
...     "https://www.google.com", 
...     http_version=CurlHttpVersion.V3
... )  
... print(response.http_version)
... 
30  # is V3
>>> CurlHttpVersion.V3
<CurlHttpVersion.V3: 30>

```

And I don't know if the "impersonate" feature is comparable to the one in http(2); since the communication differs, I imagine client characterization might be a bit more sophisticated—and you might ultimately want to revert to standard HTTP anyway (much like I prefer IPv4 to avoid being too easily identifiable).

Long story short, I lack the necessary know-how to make that call. I did notice, however, that wreq updates browser fingerprints more frequently than curl-impersonate (which has a "Pro" version) and that the code is far cleaner than in the curl_cffi project.

However, I would also support `curl_cffi`, since we already have some experience (PoC #5476) with it anyway.

vojkovic:
> Do we have http3 use cases?

It is a bit faster via less round trips:

image

> AFAIK just google supports http3

So does bing and brave and anything behind a cdn

Browsers have supported http3 for years now so if anything I think it should make us less identifiable. I think it is actually a bit simpler to impersonate because it QUIC forces just TLS 1.3 but I could see cases where it's not wanted so probably having settings we can change depending on the engine is best.

return42:
> It is a bit faster via less round trips:

Yes, it is faster, but the primary goal of scraping is not to be detected as a bot ...

My gut feeling tells me we’ll run into more issues if we deploy HTTP/3 (just as I’m convinced that IPv4 is better suited for obfuscation). In any case, `curl_cffi` will give us the chance to gain experience here, and that alone is a strong argument for `curl_cffi` :-)

--- 

My gut feeling: Google and similar companies will always aim to support as many devices as possible (smart TVs, smartphones, corporate networks that only allow specific protocols and ports, etc.). 

Only when requests associated with a profile (however that profile is generated) reach a threshold that is no longer acceptable in terms of their profit-driven objectives will they subject those profiles to closer scrutiny.

> Browsers have supported http3 for years now 

For instance, there are corporate networks that do not route UDP over port 443; consequently, the HTTP/2 fallback remains widespread (regardless of browser implementations), and the fingerprinting issue is well-resolved for HTTP/2, whereas QUIC is uncharted territory.

"Profile"  markers often exist at a different "level" (in sense of semantic); I am currently reading this article, which does a good job of contextualizing some of the requirements for me: 

- https://blog.crawlex.net/blog/detecting-curl-impersonate-utls/

The article examines the topic from the perspective of providers wishing to create bot-profiles; while much is technically feasible, the scope of reach is often of greater importance to the company involved, which is why providers (as I indicated above) will employ profiles only very cautiously.

> I think it is actually a bit simpler to impersonate because it QUIC forces just TLS 1.3 

After reading the article, I have doubts about this assumption. As far as I can tell, the problem is simply shifted to a different level (one where there isn't as much experience as with classic HTTP+TCP).

Please don't get me wrong YES, we can rely on curl_cffi, but NO to the assumption that HTTP/3 would solve any of our problems is incorrect.

----

update: May my assumptions about http/3 are wrong: [*It has also been noticed by many users, that, for a lot of sites, there is less or even none detection when using http/3.*](https://curl-cffi.readthedocs.io/en/latest/impersonate/fingerprint.html)

----

Related: 

- https://impersonate.pro/tools
- https://impersonate.pro/blog/http3-fingerprints

vojkovic:
> Yes, it is faster, but the primary goal of scraping is not to be detected as a bot ...

Wouldn't you like to have your cake and eat it too? :)

> My gut feeling tells me we’ll run into more issues if we deploy HTTP/3

I am venturing to guess your assumption (and it is a fair assumption) will be wrong - my view is that as http3 is newer, bots almost never use it and websites won't have the profiles to block it. I don't know any http clients other than the browsers and curl who have support for it. It is not a niche feature anymore either, for example, Cloudflare reports [30% http3 usage](https://radar.cloudflare.com/adoption-and-usage) for humans and >2% for bots. This will certainly help us to blend in better.

Also to be clear I don't mean only ever do http3, it's just that for those engines which support it, it can only ever be a positive for us. The others will use http2 or http1.1 as usual.

Take a look at my WIP branch - I have changed how it is implemented, basically we can have a `enable_http3` bool set for the instances that support it: https://github.com/searxng/searxng/commit/14800131a5856c95874819d1c8ef35fa6d6da2a8

> The article examines the topic from the perspective of providers wishing to create bot-profiles; while much is technically feasible, the scope of reach is often of greater importance to the company involved, which is why providers (as I indicated above) will employ profiles only very cautiously.

Yes it was an interesting read, I experienced this myself with the duckduckgo engine, it rejected the chrome impersonation but for some reason the safari impersonation was let through: https://github.com/searxng/searxng/commit/c8f65c3dfed9ce19fc39955d78171e7e5086b40c still sure if it was the UA or TLS profile that was blocked but interesting nonetheless.

I have got a lot of engines working now with some patches. I hope I can upstream these as soon as possible:

https://au.priv.au/search?q=dog&category_general=1&language=en&time_range=&safesearch=1&theme=simple

---

> just as I’m convinced that IPv4 is better suited for obfuscation

I also think it is very similar scenario but I feel you may have misjudged this. I'm sort of off topic now but on my instance I use a proxy which rotates through an enormous amount of ipv6 addresses and then falls back to v4 when needed: 
- https://github.com/vojkovic/http-proxy-ipv6-pool-docker
- https://github.com/vojkovic/egresso

This allows for more favourable rate limits, with v4 the options are much slimmer, according to Google and Cloudflare v4 and v6 are used about equally now so choosing either one is unlikely to make your requests stick out more.

Bnyro:
On the HTTP/3 topic: I personally think that it makes a lot of sense to support it. Even if there's fingerprinting, the fingerprinting won't be stricter than with HTTP/2. And in the worst case, we can still disable HTTP/3 on a per-engine basis. It's always better to have more options to try for circumventing bot detection because that increases chances to find a working configuration for the engine.

The shorter handshake of QUIC is definitely nice, but I don't think that this will bring any notable speed improvements. SearXNG isn't really a performance beast anyways, and the loading times of a few extra ms are barely notable.

> sort of off topic now but on my instance I use a proxy which rotates through an enormous amount of ipv6 addresses and then falls back to v4 when needed:
>
>    https://github.com/vojkovic/http-proxy-ipv6-pool-docker
>    https://github.com/vojkovic/egresso
>
> This allows for more favourable rate limits, with v4 the options are much slimmer, according to Google and Cloudflare v4 and v6 are used about equally now so choosing either one is unlikely to make your requests stick out more.
>
(Off-topic) Thanks for sharing, I'll definitely need to try that too and see if it helps me to avoid ratelimiting / CAPTCHAs.

return42:
@vojkovic I have to apologize; I made assumptions :facepalm: and didn't realize you already had a very clear picture .. thanks a lot for your clarifying words. :+1:

> (Off-topic) Thanks for sharing, I'll definitely need to try that too and see if it helps me to avoid ratelimiting / CAPTCHAs.

"me too"  :-)

PR #6586, opened 2026-08-25
[fix] yandex: handle empty anti-bot responses

cerXXXX:
### What does this PR do?

The changes fix empty responses that Yandex's anti-bot protection sometimes
returns. Occasionally, instead of answering, it replies with HTTP 302 carrying
an empty body and no Location header, so no search data can be obtained:
status=302, resp.text=''. The original search URL itself keeps working, and
the block is short-lived - after a short while Yandex returns the requested
data again.

The fix covers the suspension timeout after a detected captcha and re-issuing
the same search request once, dropping host/content-length headers while
keeping the original cookies. A single retry is usually enough to get an
answer; if not, the engine takes a short 5-minute suspension instead of
crashing. `catch_bad_response()` now returns a usable response object: when the body is
empty, it returns the result of the re-request. Callers therefore have to use
the returned value (`resp = catch_bad_response(resp)`) - otherwise the
original empty response would be parsed again and the retry would be wasted.
These changes raise the number of successful Yandex responses and
reduce waiting time when a captcha appears. The result was confirmed by
repeated requests against SearXNG on my own local instance.

Run an instance and search through the yandex engine until the anti-bot
window hits: before this PR the engine shows "parsing error" with a
`Document is empty` traceback in the logs; after this PR the same search
either returns results (transparent retry) or shows a clean suspension that
recovers automatically after ~5 minutes.

### Related issues

#6058 - similar problem with 360search.

### Code of Conduct

  The Ox Alpha model (running in the opencode CLI) was used for writing the
  code, investigating how correct responses depend on request headers, and
  for translating this PR into English. I have reviewed every change myself
  and verified the fix on my local instance.

### DIFF
--- a/searx/engines/yandex.py
+++ b/searx/engines/yandex.py
@@ -4,8 +4,10 @@
 from json import loads
 from urllib.parse import urlencode
 from html import unescape
+from httpx import HTTPError
 from lxml import html
-from searx.exceptions import SearxEngineCaptchaException
+from searx.exceptions import SearxEngineCaptchaException, SearxException
+from searx.network import get
 from searx.utils import humanize_bytes, eval_xpath, eval_xpath_list, extract_text, extr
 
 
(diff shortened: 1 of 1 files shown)


searx/engines/yandex.py
```diff
@@ -105,7 +135,7 @@ def response(resp):
         return results
 
     if search_type == 'images':
-        catch_bad_response(resp)
+        resp = catch_bad_response(resp)
```
Bnyro, line 138: Please move this line outside the `if` condition, it's the same for `search_type == 'web'` and `search_type == 'images'`.

Bnyro (commented):
Can't reproduce the 302's from my side (tried to get rate-limited, but failed).

SearXNG already has a `retries` option which allows to repeat a request on a bad HTTP response status (https://github.com/searxng/searxng/blob/7a01f5830c8ae263a8884e4fde84cc8c100b053f/docs/admin/settings/settings_outgoing.rst?plain=1#L85), I wonder if we couldn't use that instead?

Using it should work by creating a global variable like `retries = 5` in the engine.

searx/engines/yandex.py
```diff
@@ -48,9 +50,37 @@
...
 
 
+def _refetch(resp):
+    """Yandex anti-bot sometimes replies with an empty 302 response (no
+    Location header) during short periods of time.  One immediate re-request
+    over a fresh connection usually succeeds, so try this once before giving
+    up."""
+    headers = {k: v for k, v in resp.request.headers.items() if k.lower() not in ('host', 'content-length')}
+    try:
+        retried = get(str(resp.request.url), headers=headers)
```
Bnyro, line 60: This ignores the cookies.

PR #6602, opened 2026-08-31
Improve result deduplication for tracking URLs

mikee1010:
Improve result deduplication by ignoring common tracking query parameters such as utm_* and fbclid while preserving meaningful query parameters.

### DIFF
--- a/searx/result_types/_base.py
+++ b/searx/result_types/_base.py
@@ -37,6 +37,35 @@ log = logger.getChild("result_types")
 WHITESPACE_REGEX = re.compile('( |\t|\n)+', re.M | re.U)
 UNSET = object()
 
+TRACKING_QUERY_PARAMS = {
+    "fbclid",
+    "gclid",
+    "dclid",
+    "msclkid",
+    "mc_cid",
+    "mc_eid",
+    "igshid",
+    "ref",
+    "ref_src",
+}
+
+
+def _normalize_query_for_dedup(query: str) -> str:
+    """Remove known tracking parameters from a URL query for deduplication."""
+    if not query:
+        return ""
+
+    pairs = urllib.parse.parse_qsl(query, keep_blank_values=True)
+
+    filtered = [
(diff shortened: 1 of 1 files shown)


Bnyro (commented):
This has a somewhat big overlap with what https://github.com/searxng/searxng/blob/master/searx/plugins/tracker_url_remover.py is doing. For reference, the tracker URL remover in SearXNG takes the params it should remove from https://github.com/ClearURLs/Rules, which includes much more rules than just the 9 params you remove with your PR .

It might make sense to move the tracker url remover to SearXNG's core so that it's not a plugin anymore, and then just use the tracker param removing logic from there.

Opinion @return42 ?

return42:
> Opinion @return42 ?

In deed .. I reopened this PR to have this discussion :-)

Moving to core functionality makes sense to me ..

Implementing it as a plugin currently has only disadvantages compared to implementing it as a core function:

https://github.com/searxng/searxng/blob/d226b78bc4c9ab93a84849b8ad128a68c41be17c/searx/plugins/tracker_url_remover.py#L44-L46

.. ultimately results in the callback `self.filter_url_field` being called within the loop for each URL field ..

https://github.com/searxng/searxng/blob/d226b78bc4c9ab93a84849b8ad128a68c41be17c/searx/result_types/_base.py#L128-L138

I can imagine that we could implement this more efficiently as a core function.

-----

We had already discussed whether it makes sense to provide a switch for this, and we concluded that presumably no one would want tracker remnants in their links.

grafik

inetol:
If some param matching is failing, it's also because [the source list is no longer maintained](https://github.com/ClearURLs/Rules/commits/master/).

We should replace the current source with a maintained one https://raw.githubusercontent.com/uBlockOrigin/uAssets/refs/heads/master/filters/privacy-removeparam.txt

... Alas, we need an easylist parser (https://github.com/searxng/searxng/pull/5804)

PR #6623, opened 2026-09-02
[fix] hardening engine: ddg weather - get GEO location from response

return42:
### What does this PR do?

The previous implementation ran into an error if the search term contained words other than just the location (ValueError was raised).

To test engine use search terms like:

    !ddw weather berlin germany

### Code of Conduct

  no AI tools used


Bnyro (approved):
Can confirm the issue and the fix, lgtm 👍

PR #6620, opened 2026-09-02
[mod] network: migrate to curl_cffi

vojkovic:
### What does this PR do?

Migration of the http client from httpx to the more modern [curl_cffi](https://github.com/lexiforest/curl_cffi) to bypass TLS inspection of the ClientHello.

- By default the requests mimic chrome, but this can be set per engine with params["impersonate"].
- Http/3 is also now supported and opt-in per engine 
- Removes pool_maxsize / keepalive_expiry there is no equivalent in curl, only kept pool_connections which is similar to max_clients
- Clean up of docs and fix tests

Engine changes:
- Google engine with impersonate=chrome99_android (https://github.com/searxng/searxng/issues/6570)
- DDG now works again, images also required the impersonate=none for some reason (I think it was the firefox UA mixed with chrome TLS fingerprint)(https://github.com/searxng/searxng/issues/6596)
- Yacy auth swapped to use CurlOpts for the auth
- Enables http/3 on the engines that support it

### How to test this PR locally?

Test engines still work and don't have regressions. I tested all the main ones and couldn't find anything, also check proxies still work.
### Related issues

- Closes: https://github.com/searxng/searxng/issues/6596
- Closes: https://github.com/searxng/searxng/issues/6570
- Related: [#5476](https://github.com/searxng/searxng/pull/5476)
- Related: [#4801](https://github.com/searxng/searxng/pull/4801)
- Related: [#6571](https://github.com/searxng/searxng/pull/6571)

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [x] **I hereby confirm that this PR conforms with the [AI Policy].**

  If I have used AI tools for working on the changes in this PR, I will
  attach a list of all AI tools I used and how I used them. I hereby confirm
  that I haven't used any other tools than the ones I mention below.

Cursor helped with some bits of the migration

PR #6624, opened 2026-09-02
[fix] searchzee: bypass botblocking

Bnyro:
### What does this PR do?
- fix searchzee (only works if using `impersonate`, see https://github.com/searxng/searxng/pull/6620)

- enable and use searchzee after rebasing the PR on https://github.com/searxng/searxng/pull/6620

### Related issues
- requires https://github.com/searxng/searxng/pull/6620

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/searchzee.py
+++ b/searx/engines/searchzee.py
@@ -4,13 +4,11 @@ independent search infrastructure."""
 
 import typing as t
 from urllib.parse import urlencode
+import uuid
 
-from searx.exceptions import SearxEngineAPIException
 from searx.extended_types import SXNG_Response
-from searx.network import get
 from searx.result_types import EngineResults
-from searx.utils import extr, html_to_text
-from searx.enginelib import EngineCache
+from searx.utils import html_to_text
 
 if t.TYPE_CHECKING:
     from searx.search.processors import OnlineParams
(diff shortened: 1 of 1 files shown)


searx/engines/searchzee.py
```diff
@@ -34,43 +32,19 @@
...
-
-    token = extr(token_resp.text, "const SEARCHZEE_API_TOKEN = \"", "\";")
-    CACHE.set("token", token, expire=3600)
-
-    return token
-
 
 def request(query: str, params: "OnlineParams"):
-    params["headers"]["X-SearchZee-Token"] = _obtain_api_token()
+    params["cookies"]["szs"] = uuid.uuid4().hex
```
vojkovic: nit: I found the cookie can be set to anything, but the site uses ``crypto.randomUUID() `` so it would probably be less suspicious to use ``str(uuid.uuid4())`` so it has got the dashes.

e.g. e5ad46f55f884b3992cd8bd56bd1b2cb vs e5ad46f5-5f88-4b39-92cd-8bd56bd1b2cb

vojkovic (approved):
Another great fix, just one single nit from me 👍

searx/engines/searchzee.py
Bnyro: Thanks for the catch!

I thought that `uuid.uuid4().hex` would include the dashes, and I didn't bother with testing my assumption since the engine worked well already :)

PR #6629, opened 2026-09-02
[mod] s1search: update mirrors to the ones that are not cloudflared

Bnyro:
### What does this PR do?
- update the engines that are based on s1search, as s1search added cloudflare captchas to some of the existing mirrors
- also set them to inactive so that they don't get put behind cloudflare immediately as well

### Code of Conduct

No AI used.


Bnyro:
The only other page with the same request layout I found that is not behind Cloudflare is https://search.nation.online, but it requires the `sc` parameter for all requests, so it's not working with the existing `s1search` implementation.

Let's hope the 2 I added above remaining functional for a while.

PR #6637, opened 2026-09-03
 [feat] engines: add europe pmc (science, scientific publications)

rdurnik:
This PR adds Europe PMC as a new database source. The database also provides links to full-text PDFs.

There are still issues that may need to be addressed:
- The Europe PMC API does not support traditional pagination and instead uses cursorMark. To return only the requested page, the current implementation slices the results after retrieval, which requires re-downloading data for each page request. There may be a more efficient way to handle this.
- We still need to address timeout errors.

### DIFF
--- a/searx/engines/europepmc.py
+++ b/searx/engines/europepmc.py
@@ -1,0 +1,150 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""`Europe PMC`_ provides comprehensive access to life sciences literature from
+trusted sources.  With Europe PMC you can search and read millions of
+publications, preprints and other documents enriched with links to supporting
+data, reviews, protocols, and other relevant resources.
+
+.. _Europe PMC: https://europepmc.org/
+
+Configuration
+=============
+
+.. code:: yaml
+
+   - name: europepmc
+     engine: europepmc
+     shortcut: epmc
+
+Implementations
+===============
+
+"""
+
+import typing as t
+
+from datetime import datetime
+from urllib.parse import urlencode
+
+from dateutil.parser import isoparse
+
+from searx.enginelib import EngineCache
+from searx.result_types import EngineResults
+from searx.utils import html_to_text
+
+if t.TYPE_CHECKING:
+    from searx.extended_types import SXNG_Response
+    from searx.search.processors import OnlineParams
+
+
+about = {
+    "website": "https://europepmc.org/",
+    "wikidata_id": "Q5412157",
+    "official_api_documentation": "https://europepmc.org/RestfulWebService",
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["science", "scientific publications"]
+paging = True
+
+# engine dependent config
+search_url = "https://www.ebi.ac.uk/europepmc/webservices/rest/search"
+article_url = "https://europepmc.org/article/"
+
+page_size = 20
+
+CACHE: EngineCache
+"""Cache for storing the pagination cursor."""
+
+
+def setup(engine_settings: dict[str, t.Any]):
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+
+
+def _cache_key(query: str, page: int) -> str:
+    return f"{query}|{page}"
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    args = {
+        "query": query,
+        "format": "json",
+        "resultType": "core",
+        "pageSize": page_size,
+    }
+
+    if params["pageno"] > 1:
+        if cursor := CACHE.get(_cache_key(query, params["pageno"])):
+            args["cursorMark"] = cursor
+        else:
+            # no cached cursor for that page
+            params["url"] = None
+            return
+
+    params["url"] = f"{search_url}?{urlencode(args)}"
+
+
+def response(resp: "SXNG_Response") -> EngineResults:
+    res = EngineResults()
+
+    json_resp = resp.json()
+
+    # store pagination cursor for loading next pages in cache
+    if next_cursor := json_resp.get("nextCursorMark"):
+        next_page = resp.search_params["pageno"] + 1
+        query = resp.search_params["query"]
+        CACHE.set(_cache_key(query, next_page), next_cursor)
+
+    all_results = json_resp.get("resultList", {}).get("result", [])
+
+    for item in all_results:
+        source = item.get("source", "")
+        identifier = item.get("id", "")
+        url = f"{article_url}{source}/{identifier}" if source and identifier else ""
+
+        journal_info: dict[str, t.Any] = item.get("journalInfo", {})
+        journal: dict[str, t.Any] = journal_info.get("journal", {})
+
+        res.add(
+            res.types.Paper(
+                url=url,
+                title=html_to_text(item.get("title", "")),
+                content=html_to_text(item.get("abstractText", "")),
+                journal=journal.get("title", ""),
+                issn=[journal.get("issn", "")],
+                authors=_get_authors(item),
+                doi=item.get("doi", ""),
+                publishedDate=_get_published_date(item),
+                type=", ".join((item.get("pubTypeList", {})).get("pubType", [])),
+                pdf_url=_get_pdf_url(item),
+                html_url=url,
+            )
+        )
+
+    return res
+
+
+def _get_authors(item: dict[str, t.Any]) -> list:
+    """Extract the list of authors from the item."""
+    if authors := item.get("authorString", None):
+        authors = [author.strip().rstrip(".") for author in authors.split(",") if author.strip()]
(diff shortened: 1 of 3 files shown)


Bnyro:
Please use the pull request template next time and tick the AI policy checkbox.

searx/engines/europepmc.py
```diff
@@ -0,0 +1,145 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Europe PMC_ provides comprehensive access to life sciences literature from
```
Bnyro:
```suggestion
"""`Europe PMC`_ provides comprehensive access to life sciences literature from
```
Otherwise the link won't work.

Bnyro: You can test it with `make docs.live`

searx/engines/europepmc.py
```diff
@@ -0,0 +1,145 @@
...
+article_url = "https://europepmc.org/article/"
+
+
+def request(query: str, params: "OnlineParams") -> None:
+    args = urlencode(
+        {
+            "query": query,
+            "format": "json",
+            "resultType": "core",
+            "pageSize": 100,
```
Bnyro:
```suggestion
            "pageSize": page_size,
```
Otherwise the `page_size` variable is useless

searx/engines/europepmc.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _extract_starting_page(resp: "SXNG_Response") -> int:
+    """Extract the starting page number from the response's search parameters.
+
+
+    The Europe PMC API does not support paging (only cursorMark), so we need to slice the results to return only the requested page and redownload with each page."""
+    pageno: int = resp.search_params.get("pageno", 1) or 1
+    starting_page = (pageno - 1) * page_size
+    return starting_page
```
Bnyro: If the engine returns all results with a single request, you should just show all of the results at once instead of adding some pseudo-pagination here. If you already have all results loaded, it doesn't make sense to only show some of them.

If the website doesn't support pagination, the engine implementation shouldn't either.

searx/engines/europepmc.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _get_abstract(item: dict[str, t.Any]) -> str:
+    """Convert the abstract text from HTML to plain text."""
+    html_abstract = item.get("abstractText", "")
+    return html_to_text(html_abstract)
```
Bnyro: That's not enough code to be its own function. Inline to where it's called.

searx/engines/europepmc.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+    return res
+
+
+def _extract_starting_page(resp: "SXNG_Response") -> int:
+    """Extract the starting page number from the response's search parameters.
+
+
+    The Europe PMC API does not support paging (only cursorMark), so we need to slice the results to return only the requested page and redownload with each page."""
```
return42: just a nit; recommended line length is 80 characters, even the CI will accept up to 130.

---

BTW: can you please squash the commits into one / thanks!

And please take not of the developer-docs: https://docs.searxng.org/dev/quickstart.html

grafik

rdurnik:
Thank you @Bnyro and @return42 for the code review!

return42 (approved):
Thanks, cewl engine 👍  .. 

Depending on the search query, the 6-sec timeout might be too short .. I had to set it significantly higher (10 seconds?) for some of my queries.

What do you think?

rdurnik:
> Thanks, cewl engine 👍 ..
> 
> Depending on the search query, the 6-sec timeout might be too short .. I had to set it significantly higher (10 seconds?) for some of my queries.
> 
> What do you think?

Sure

searx/engines/europepmc.py
```diff
@@ -0,0 +1,121 @@
...
+    "use_official_api": True,
+    "require_api_key": False,
+    "results": "JSON",
+}
+
+categories = ["science", "scientific publications"]
+
+# engine dependent config
+paging = False
+page_size = 900
```
Bnyro: The engine is probably so slow due to the large page size? Have you tried lowering it to e.g. 50?

Bnyro:
Pagination seems to actually be possible using the `cursorMark` parameter described in https://europepmc.org/RestfulWebService#!/Europe32PMC32Articles32RESTful32API/search and storing it in an `EngineCache`, e.g. like the `duckduckgo_web.py` engine.

I'll push a commit to implement it later or tomorrow if I have time.

Bnyro (commented):
It looks like you disabled that maintainers can push to your branch, so I can't add changes to your PR? Please enable it next time, see https://docs.github.com/en/pull-requests/how-tos/work-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork.

Anyways, here's the patch for adding pagination support:
```diff
diff --git a/searx/engines/europepmc.py b/searx/engines/europepmc.py
index 188ffcc4f..475a7a22a 100644
--- a/searx/engines/europepmc.py
+++ b/searx/engines/europepmc.py
@@ -27,6 +27,7 @@ from urllib.parse import urlencode
 
 from dateutil.parser import isoparse
 
+from searx.enginelib import EngineCache
 from searx.result_types import EngineResults
 from searx.utils import html_to_text
 
@@ -45,30 +46,58 @@ about = {
 }
 
 categories = ["science", "scientific publications"]
+paging = True
 
 # engine dependent config
-paging = False
-page_size = 900
 search_url = "https://www.ebi.ac.uk/europepmc/webservices/rest/search"
 article_url = "https://europepmc.org/article/"
 
+page_size = 20
+
+CACHE: EngineCache
+"""Cache for storing the pagination cursor."""
+
+
+def setup(engine_settings: dict[str, t.Any]):
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+
+
+def _cache_key(query: str, page: int) -> str:
+    return f"{query}|{page}"
+
 
 def request(query: str, params: "OnlineParams") -> None:
-    args = urlencode(
-        {
-            "query": query,
-            "format": "json",
-            "resultType": "core",
-            "pageSize": page_size,
-        }
-    )
-    params["url"] = f"{search_url}?{args}"
+    args = {
+        "query": query,
+        "format": "json",
+        "resultType": "core",
+        "pageSize": page_size,
+    }
+
+    if params["pageno"] > 1:
+        if cursor := CACHE.get(_cache_key(query, params["pageno"])):
+            args["cursorMark"] = cursor
+        else:
+            # no cached cursor for that page
+            params["url"] = None
+            return
+
+    params["url"] = f"{search_url}?{urlencode(args)}"
 
 
 def response(resp: "SXNG_Response") -> EngineResults:
     res = EngineResults()
 
-    all_results = resp.json().get("resultList", {}).get("result", [])
+    json_resp = resp.json()
+
+    # store pagination cursor for loading next pages in cache
+    if next_cursor := json_resp.get("nextCursorMark"):
+        next_page = resp.search_params["pageno"] + 1
+        query = resp.search_params["query"]
+        CACHE.set(_cache_key(query, next_page), next_cursor)
+
+    all_results = json_resp.get("resultList", {}).get("result", [])
 
     for item in all_results:
         source = item.get("source", "")
@@ -81,7 +110,7 @@ def response(resp: "SXNG_Response") -> EngineResults:
         res.add(
             res.types.Paper(
                 url=url,
-                title=item.get("title", ""),
+                title=html_to_text(item.get("title", "")),
                 content=html_to_text(item.get("abstractText", "")),
                 journal=journal.get("title", ""),
                 issn=[journal.get("issn", "")],
diff --git a/searx/settings.yml b/searx/settings.yml
index 8e9ee4fc4..cf2e754f7 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -807,7 +807,6 @@ engines:
   - name: europepmc
     engine: europepmc
     shortcut: epmc
-    timeout: 6.0
 
   - name: erowid
     engine: xpath
```

Note that you can't just jump from e.g. page 1 to page 8, you must first load page 1 to be able to load page 2, then page 2 to be able to load page 3, etc...

rdurnik:
@Bnyro I've added this patch. Thank you!

return42:
@rdurnik:  You were already asked above to grant us maintainers write access to the branch. Unfortunately, you have not complied with this request—is there a specific reason for this?

Write access is the first requirement for collaborating on a change proposal, so I would like to ask you once again to grant us maintainers this access. Thank you very much!

rdurnik:
Unfortunately, because we pushed this PR as an organization, the option to retrospectively grant write access simply does not appear (even if I have admin rights). Please let me know if we can finish this PR without it. Next time, we will try to find an alternative approach. Sorry for the inconvenience.

To be honest, I pushed this PR by accident, which is why I did not follow the guidelines. We had actually planned to do this later together with some other engine additions. I did not expect you to reopen the PR. 😄

Bnyro (approved):
Thanks!

PR #6638, opened 2026-09-03
[feat] engines: add JS-based s1search engine

Bnyro:
### What does this PR do?
- add a new engine for s1search sites that include the __RAMPJS__ javascript variable
- so that's basically a different response layout for the s1search engine, but with the same results

- enable the engine and search !s1 test

### Related issues
- #6629 

### Code of Conduct

No AI used.

### DIFF
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -3307,13 +3307,6 @@ engines:
       website: https://minecraft.wiki/
       wikidata_id: Q105533483
 
-  # s1search google engines / mirrors
-  - name: searchtoday
-    engine: s1search
-    shortcut: std
-    base_url: https://info.searchtoday.site
-    disabled: true
-
   - name: sina
     engine: json_engine
     shortcut: sina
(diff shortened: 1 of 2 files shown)


searx/settings.yml
```diff
@@ -3307,13 +3307,6 @@ engines:
       website: https://minecraft.wiki/
       wikidata_id: Q105533483
 
-  # s1search google engines / mirrors
```
Bnyro, line 3310: dropped because cloudflared (oversight from #6629 )

return42 (approved):
LGTM 

--- 

I don't know if this applies to this engine, but if an engine, or rather the index it queries, is only available in one language, then it might make sense to set the [language](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.language) option.

For example the t-online engine:

https://github.com/searxng/searxng/blob/a1144dda3e97668c9d445022b7019c224cd4bb1e/searx/engines/tonline.py#L44

Image

Bnyro:
> I don't know if this applies to this engine, but if an engine, or rather the index it queries, is only available in one language, then it might make sense to set the [language](https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.language) option.
> 
Just tested it. It can return results in all languages, depending on the query. Pretty sure it uses the Google index, so it can handle more than just one single language.

PR #6639, opened 2026-09-03
[fix] neosearch: requests blocked due to missing xsrf token

Bnyro:
- !neosearch test

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/neosearch.py
+++ b/searx/engines/neosearch.py
@@ -6,10 +6,14 @@
 
 from json import loads
 import typing as t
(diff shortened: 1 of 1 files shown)


vojkovic (approved):
Great fix, works very well - looks like the engine supports h3 as well if you want to include it. :)

Bnyro:
> Great fix, works very well - looks like the engine supports h3 as well if you want to include it. :)
>
Thanks for the hint, enabled h3 :)

PR #6649, opened 2026-09-04
[feat] engines: add searchrockit.com (general, news, images)

Bnyro:
### What does this PR do?
- the general and news results are very fast (within a second for me)
- the images engine is very slow (I had to increase the timeout to 2 minutes to test it...)

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/searchrockit.py
+++ b/searx/engines/searchrockit.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""SearchRockit is an American search engine. It allegedly has its own index,
+but the results seem to come from Google."""
+
+import typing as t
+from urllib.parse import urlencode
+from lxml import html
+from dateutil import parser
+
+from searx.result_types import EngineResults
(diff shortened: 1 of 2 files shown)


searx/engines/searchrockit.py
```diff
@@ -0,0 +1,87 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""SearchRockit is an Amerian P2P search engine."""
```
vojkovic: typo: 'American' - also I don't think it is P2P, most results look like they come from google

vojkovic:
Engine and results look good. 👍 

It is behind CF so supports h3. Once you rebase you should be able to add `enable_http3 = True` to take advantage of it.

searx/engines/searchrockit.py
Bnyro: Yeah. Their website claims that the web results are from their own index, only the "latest" category would be from Google. But hard to believe that. Changed the wording here.

PR #6650, opened 2026-09-04
[fix] engines: tusksearch referer header

vojkovic:
### What does this PR do?

Adds a referer header to bypass tusksearch's botdetection.

### Code of Conduct

Cursor helped in finding the header


Bnyro (approved):
Works, nice 👍

PR #6651, opened 2026-09-04
[fix] engines: resulthunter search_source arg

vojkovic:
### What does this PR do?

Adds search_source arg to the resulthunter engine to bypass bot detection.

The arg can really be set to anything to work but the site uses three: 'header' 'other' 'extension', all returned the same results so I just picked 'other'.

### Code of Conduct

Cursor helped in finding the arg

### DIFF
--- a/searx/engines/resulthunter.py
+++ b/searx/engines/resulthunter.py
@@ -52,6 +52,7 @@ def request(query: str, params: "OnlineParams") -> None:
         "q": query,
         "search_type": resulthunter_categ,
         "offset": params["pageno"] - 1,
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Works, thanks!

I didn't even notice the engine always returned 0 results, maybe we should look into detecting access denied responses in a different PR (will do that).

vojkovic:
> maybe we should look into detecting access denied responses in a different PR (will do that).

LLMs are really good for that btw since they never get tired and can find issues when there isn't an explicit error being thrown

PR #6652, opened 2026-09-04
[fix] engines: update brave images/videos parser and news xpath

vojkovic:
### What does this PR do?

- Brave changed their format for images/videos engine so there is <script> tag rubbish in front, fixes that so it seeks to the data part
- Updated the news xpath for the new classes 
- When testing, I also found some image results with a literal \ in the title couldn't parse properly which was a bug in js_obj_str_to_json_str treating it as escaped, see first test.

- `!brave.images ai` <- this query gave the parsing error before 
- Check videos and news engines work too

### Code of Conduct

Cursor helped find the parsing error while fuzzing results and found updated xpaths

### DIFF
--- a/searx/engines/brave.py
+++ b/searx/engines/brave.py
@@ -248,13 +248,13 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
     #    node_ids: [0, 19],
     #    data: [{type:"data",data: .... ["q","goggles_id"],route:1,url:1}}]
     #          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-    text = text[text.index("<script") : text.index("</script")]
-    if not text:
-        raise ValueError("can't find JS/JSON data in the given text")
+    #    form: null,
+    #    error: null
+    # });
     start = text.index("data: [{")
-    end = text.rindex("}}]")
-    js_obj_str = text[start:end]
-    js_obj_str = "{" + js_obj_str + "}}]}"
+    newline = text.index("\n", start)
+    end = text.rindex("}}]", start, newline)
+    js_obj_str = "{" + text[start:end] + "}}]}"
     # js_obj_str = js_obj_str.replace("\xa0", "")  # remove ASCII for &nbsp;
     # js_obj_str = js_obj_str.replace(r"\u003C", "<").replace(r"\u003c", "<")  # fix broken HTML tags in strings
     json_str = js_obj_str_to_json_str(js_obj_str)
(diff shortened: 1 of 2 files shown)


inetol (approved):

searx/utils.py
```diff
@@ -735,13 +735,13 @@ def js_obj_str_to_json_str(js_obj_str: str) -> str:
...
-            # * the previous part ends with a black slash
-            # * the current part is a single quote
-            # when detected the blackslash is removed on the previous part
+            # drop a trailing \ that was escaping the quote
+            # leave it if it has been escaped twice as a literal i.e. two \ and ' in a row
             if blackslash_just_before and p[:1] == "'":
-                parts[i - 1] = parts[i - 1][:-1]
+                prev = parts[i - 1]
+                n_bs = len(prev) - len(prev.rstrip("\\"))
+                if n_bs % 2 == 1:
```
Bnyro:
```suggestion
                num_backslashes = len(prev) - len(prev.rstrip("\\"))
                if num_backslashes % 2 == 1:
```
The code is already hard enough to understand, so please don't use mystic abbreviations that make it even harder :)

(I've always been in favor of using a 3rd party library for this, but that also has its downsides).

searx/engines/brave.py
```diff
@@ -248,13 +248,10 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
-    text = text[text.index("<script") : text.index("</script")]
-    if not text:
-        raise ValueError("can't find JS/JSON data in the given text")
     start = text.index("data: [{")
-    end = text.rindex("}}]")
-    js_obj_str = text[start:end]
-    js_obj_str = "{" + js_obj_str + "}}]}"
+    chunk = text[start : text.index("</script", start)]
+    end = chunk.rindex("}}]")
+    js_obj_str = "{" + chunk[:end] + "}}]}"
```
Bnyro: I think that could still be improved a bit, because if there's another }}] in the script, after the data, it would still find that.

Given that the data is always on a single line, I think it would be better to only match }}] before a newline.
I.e.
```diff
diff --git a/searx/engines/brave.py b/searx/engines/brave.py
index d5fc465ff..8d0994227 100644
--- a/searx/engines/brave.py
+++ b/searx/engines/brave.py
@@ -248,10 +248,13 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
     #    node_ids: [0, 19],
     #    data: [{type:"data",data: .... ["q","goggles_id"],route:1,url:1}}]
     #          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    #    form: null,
+    #    error: null
+    # });
     start = text.index("data: [{")
-    chunk = text[start : text.index("</script", start)]
-    end = chunk.rindex("}}]")
-    js_obj_str = "{" + chunk[:end] + "}}]}"
+    newline = text.index("\n", start)
+    end = text.rindex("}}]", start, newline)
+    js_obj_str = "{" + text[start:end] + "}}]}"
     # js_obj_str = js_obj_str.replace("\xa0", "")  # remove ASCII for &nbsp;
     # js_obj_str = js_obj_str.replace(r"\u003C", "<").replace(r"\u003c", "<")  # fix broken HTML tags in strings
     json_str = js_obj_str_to_json_str(js_obj_str)
```

Bnyro (commented):
Works well, thanks 👍 

Just two nits above.

vojkovic:
Very good recommendations thank you! [diff](https://github.com/searxng/searxng/compare/222bbe772c41847fe9affc9e62c52328440f2532..8d1f17486f1e941e0421199737f1ad55394b9646)

Bnyro (approved):

PR #6653, opened 2026-09-04
[fix] engines: dogpile token

vojkovic:
### What does this PR do?

Dogpile returns errors without an api token and a `dp_api_token` cookie. The token is needed for the non-web engines and the cookie is needed for all of them.

This PR requests a token and caches it for 240s so it doesn't need to be requested every time and is also shared across all the engine categories (images/videos/news) and adds the `dp_api_token`.

Enable engines and test:

``!dog``
``!dogi``
``!dogv``
``!dogn``

### Code of Conduct

Cursor helped find the endpoint and cookie

### DIFF
--- a/searx/engines/dogpile.py
+++ b/searx/engines/dogpile.py
@@ -8,6 +8,9 @@ import typing as t
 from datetime import datetime, timezone
 import html
 
+from searx.enginelib import EngineCache
+from searx.exceptions import SearxEngineAPIException
+from searx.network import post
 from searx.utils import format_duration, html_to_text, humanize_number
 from searx.result_types import EngineResults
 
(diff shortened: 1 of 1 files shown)


searx/engines/dogpile.py
```diff
@@ -35,15 +38,35 @@
...
 
 def setup(_: dict[str, t.Any]) -> bool | None:
     if dogpile_categ not in ("search", "images", "videos", "news"):
         raise ValueError("invalid search type: %s" % dogpile_categ)
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache("dogpile")  # one token for search/images/videos/news
+    return True
+
+
+def obtain_token() -> str:
```
Bnyro: nit: please prefix private methods with _, i.e. _obtain_token, so that nobody imports them by accident

Bnyro (approved):
Good finding with the dogpile token endpoint. I only found the token embedded in the HTML pages, the API call is much cleaner.

Works fine for me, nice job 👍

PR #6657, opened 2026-09-05
[fix] resulthunter: detect blocked requests instead of returning 0 results

Bnyro:
### What does this PR do?
- detect if resulthunter blocks a request instead of just returning 0 results
- e.g. detects if something like https://resulthunter.com/search?q=rust&search_type=web&freshness=&search_lang=&country= is shown instead of actual results
- maybe we could also have tried to directly match the "Installation required" page from above, but possibly they sometimes also show other sites or change it in the future, so I think it's better to check for the results layout

- !reh test before #6651 is merged (or by undoing the changes from it)
- request should fail because it's blocked

### Related issues
- see https://github.com/searxng/searxng/pull/6651#pullrequestreview-5120658883

### Code of Conduct

No AI used.

### DIFF
--- a/searx/engines/resulthunter.py
+++ b/searx/engines/resulthunter.py
@@ -10,6 +10,7 @@ from urllib.parse import urlencode
 from lxml import html
 
 from searx import locales
+from searx.exceptions import SearxEngineResponseException
 from searx.result_types import EngineResults
 from searx.utils import eval_xpath_list, eval_xpath, extract_text
 
@@ -110,6 +111,11 @@ def _image_results(doc: "ElementBase") -> EngineResults:
 
 def response(resp: "SXNG_Response") -> EngineResults:
     doc = html.fromstring(resp.text)
+
+    # if the request was wrong (e.g. missing params), the site doesn't contain a result container
+    # and instead shows an "Installation required" page to download the resulthunter browser extension
+    if not eval_xpath(doc, "//div[contains(@class, 'organic-results-container')]"):
+        raise SearxEngineResponseException()
 
     match resulthunter_categ:
         case "web":


searx/engines/resulthunter.py
```diff
@@ -111,6 +112,11 @@ def _image_results(doc: "ElementBase") -> EngineResults:
 def response(resp: "SXNG_Response") -> EngineResults:
     doc = html.fromstring(resp.text)
 
+    # if the request was wrong (e.g. missing params), the site doesn't contain a result container
+    # and instead shows an "Installation required" page to download the resulthunter browser extension
+    if not eval_xpath(doc, "//div[contains(@class, 'organic-results-container')]"):
+        raise SearxEngineResponseException()
```
vojkovic, line 118: nit / question: is SearxEngineResponseException the correct exception to throw? I haven't seen it be used almost ever

I would have use SearxEngineAccessDeniedException

vojkovic (approved):
Another good PR 👍

Just my little nit

searx/engines/resulthunter.py
Bnyro: Yeah, I thought about using that too.

But access denied kind of reads like you were ratelimited/IP-blocked (that's what I also get on some IPs of my server).

So I wanted to use something different to show that it's not an IP related issue, but rather engine implenentation specific.

If you know an other error type that we could use, I'd be happy with it, I just don't want to use AccessDenied.

vojkovic: I think we should use this exception more often now.

I agree a lot of people see AccessDenied and think it's a problem with their specific setup rather than the implementation of the engine.

PR #6658, opened 2026-09-05
[fix] engines: duckduckgo web bypass botdetection

vojkovic:
### What does this PR do?

Duckduckgo web broke after the curl_cffi migration.

Fixes the ddg web engine with two fixes:
- Uses impersonate from an old safari version. I found this worked better than chrome or firefox.
- Very rarely some queries like `site:github.com searxng` are returning js challenges which are easy enough to solve.

Enable duckduckgo_web because its disabled by default and test it works now.

Check `!ddgw searxng site:github.com`

### Code of Conduct

Cursor helped with the implementation and found the ddg challenge while fuzzing results.

### DIFF
--- a/searx/engines/duckduckgo_web.py
+++ b/searx/engines/duckduckgo_web.py
@@ -14,11 +14,12 @@ can't build it ourselves and must scrape it from the HTML pages.
 """
 
 import typing as t
+import re
 
-from urllib.parse import quote_plus
+from urllib.parse import quote_plus, urljoin
 from lxml import html
 
-from searx.utils import html_to_text, gen_useragent, extract_text, eval_xpath
+from searx.utils import html_to_text, extract_text, eval_xpath
 from searx.result_types import EngineResults
 from searx.enginelib import EngineCache
 from searx.network import get
@@ -38,7 +39,6 @@ about = {
 # engine dependent config
 categories = ["general"]
 paging = True
-_HTTP_User_Agent: str = gen_useragent()
 
 base_url = "https://duckduckgo.com"
 
@@ -73,6 +73,8 @@ def _fetch_first_page_link(
     resp = get(
         url=f"{base_url}/?q={quote_plus(query)}&t=h_&ia=web",
         headers=headers,
+        impersonate="firefox",
+        default_headers=False,
         timeout=2,
     )
 
@@ -96,6 +98,43 @@ def _cache_key(query: str, pageno: int) -> str:
     return f"nextpage_url|{query}|{pageno}"
 
 
+def _solve_jsa(resp: "SXNG_Response") -> "SXNG_Response":
+    """Duckduckgo sometimes issues a challenge instead of json."""
+
+    # length that a real browser would report for where the broken snippet is
+    html_len = {
+        "<p><div></p><p></div": 32,
+        "<li><div></li><li></div": 29,
+        "<div><div></div><div></div": 33,
+        "<br><div></br><br></div": 23,
+    }
+
+    js = resp.text or ""
+    jsa_match = re.search(r"let jsa = (\d+);.*?DDG\.deep\.initialize\('([^']+)'", js, re.S)
+    if not jsa_match:
+        return resp
+
+    js_functions = dict(re.findall(r"let (\w+) = function\(num\) \{([^}]*)\};", js))
+    jsa = int(jsa_match.group(1))
+    try:
+        for name in re.findall(r"jsa = (\w+)\(jsa\);", js):
+            body = js_functions[name]
+            mul = re.search(r"num \* (\d+)", body)
+            jsa = jsa * int(mul.group(1)) if mul else jsa + html_len[re.search(r"`([^`]+)`", body).group(1)]
+    except (KeyError, AttributeError):
+        return resp
+
+    params = resp.search_params
+    follow = get(
+        urljoin("https://links.duckduckgo.com", jsa_match.group(2) + str(jsa)),
+        headers=params["headers"],
+        impersonate="firefox",
+        default_headers=False,
+    )
+    follow.search_params = params
+    return follow
+
+
 def request(query: str, params: "OnlineParams") -> None:
 
     if len(query) >= 500:
@@ -103,25 +142,15 @@ def request(query: str, params: "OnlineParams") -> None:
         params["url"] = None
         return
 
-    headers = params["headers"]
-
-    # The vqd value is generated from the query and the UA header.  To be able
-    # to reuse the vqd value, the UA header must be static.
-    headers["User-Agent"] = _HTTP_User_Agent
-    headers["Accept"] = "*/*"
-    headers["Referer"] = f"{base_url}/"
-    headers["Host"] = "duckduckgo.com"
-
-    # Sec-Fetch headers are required to not get blocked when sending a Firefox user agent
-    headers["Sec-Fetch-Dest"] = "script"
-    headers["Sec-Fetch-Mode"] = "no-cors"
-    headers["Sec-Fetch-Site"] = "same-site"
+    # firefox TLS only
+    params["impersonate"] = "firefox"
+    params["default_headers"] = False
 
     api_url = ""
     if params["pageno"] > 1:
         api_url = CACHE.get(_cache_key(query, params["pageno"]))
     else:
-        api_url = _fetch_first_page_link(query, headers)
+        api_url = _fetch_first_page_link(query, params["headers"])
 
     if not api_url:
         params["url"] = None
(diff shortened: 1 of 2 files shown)


searx/engines/duckduckgo_web.py
```diff
@@ -96,6 +99,42 @@ def _cache_key(query: str, pageno: int) -> str:
...
+        "<br><div></br><br></div": 23,
+    }
+
+    js = resp.text or ""
+    m = re.search(r"let jsa = (\d+);.*?DDG\.deep\.initialize\('([^']+)'", js, re.S)
+    if not m:
+        return resp
+
+    fns = dict(re.findall(r"let (\w+) = function\(num\) \{([^}]*)\};", js))
+    jsa = int(m.group(1))
```
Bnyro:
```suggestion
    jsa_match = re.search(r"let jsa = (\d+);.*?DDG\.deep\.initialize\('([^']+)'", js, re.S)
    if not jsa_match:
        return resp
    jsa = int(jsa_match.group(1))

    js_functions = dict(re.findall(r"let (\w+) = function\(num\) \{([^}]*)\};", js))
```
Sorry for always complaining about function naming, but I think that if the code is rather complex/non-self-explaining because it very much depends on the engine itself (i.e. the engine result layout), it's better to use some more descriptive variable names :)

searx/engines/duckduckgo_web.py
```diff
@@ -134,9 +174,18 @@ def request(query: str, params: "OnlineParams") -> None:
...
-    for result in res_json["results"]:
+    # check if ddg returns a challenge
+    # e.g. 'site:github.com searxng'
+    if "let jsa =" in (resp.text or ""):
+        resp = _solve_jsa(resp)
+
+    try:
+        results = resp.json()["results"]
+    except (ValueError, KeyError, TypeError):
+        return res
```
Bnyro:
```suggestion
     results = resp.json()["results"]
```
```suggestion
     results = resp.json().get("results", [])
```
Why do we catch the exceptions here? If the engine doesn't return valid "results", that probably means that the engine is broken again?

Also we can probably avoid try/except by just using an if condition, right?

searx/engines/duckduckgo_web.py
Bnyro: Except for the referer, all other headers can/should be removed because they are set by curl impersonate.

I tested it and even the referer header doesn't seem to be necessary, but it doesn't hurt to keep it.

searx/engines/duckduckgo_web.py
Bnyro: (and the `_HTTP_User_Agent` var can be dropped)

Bnyro (commented):
Can confirm that using the safari impersonate param works, good find 👍 

I couldn't reproduce the jsa "captcha", so I couldn't test it unfortunately.

searx/engines/duckduckgo_web.py
return42: Before using curl-impersonate, we defined a global _HTTP_User_Agent in the ddg engines. 

https://github.com/search?q=repo%3Avojkovic%2Fsearxng+_HTTP_User_Agent&type=code

This was necessary because the vqd value we cached was only valid for the given UA value. Now, with "impersonate", the network stack (curl_cffi) sets the UA header, as it is part of the "impersonate" feature.

All ddg engines that use the vqd value would now need to be revised so that the vqd value is stored in the cache depending on the "impersonate" parameter.

rqi14:
Ran into the same thing and measured the mechanism, in case it helps narrow the fix.

The trigger is `default_headers`, not the impersonation profile as such. `new_client()` in `searx/network/client.py` sets `default_headers=True` whenever impersonation is on, so curl_cffi adds the impersonated browser's full header set — including Chrome's `sec-ch-ua*` client hints. `duckduckgo_web` then overrides `User-Agent` with a Firefox one (`_HTTP_User_Agent = gen_useragent()`), and DDG sees Chrome client hints sitting next to a Firefox UA.

Measured against `searxng/searxng:latest` (2026.9.5-c7f3080aa, curl_cffi 0.16.1), single residential exit, same query, on `links.duckduckgo.com/d.js?o=json`:

| impersonate | default_headers | result |
|---|---|---|
| chrome | True (current default) | `202 {}` |
| chrome | False | `200`, 11 results |
| firefox | True | `202 {}` |
| none | False | `200`, 11 results |

So the impersonation profile isn't what decides it here — `default_headers` is. That might also explain why an older Safari profile "worked better": worth checking whether it still fails with `default_headers=True`.

One caveat for anyone tempted by a global switch: `default_headers=False` on the session leaves SearXNG with no `User-Agent` at all (verified against an echo endpoint — only `Host` and `Accept-Encoding` go out), and google/startpage/wikipedia immediately start returning CAPTCHA/403. It also can't be narrowed to "requests that set their own UA": `google.py` sets a Nokia UA on purpose and still needs the rest of the default headers. A per-request `default_headers=False` on the DDG calls specifically is the smallest thing that worked here.

searx/engines/duckduckgo_web.py
vojkovic: Please don't stop calling me out on my shitty naming :-)

I will get better eventually I promise 😆

searx/engines/duckduckgo_web.py
vojkovic: > Except for the referer, all other headers can/should be removed because they are set by curl impersonate.

I was able to remove all these headers and then also disabled `default_headers` in curl-cffi ( I added a new setting for it ). I can add a bit more docs for it if you're happy with it. This removes those headers set by curl impersonate

With impersonate, I don't know what it is exactly but something in those headers it sends gets blocked. Now this request is absolutely insane looking now, but it seems ddg really doesn't mind:

```
Host: duckduckgo.com
Accept-Encoding: gzip, deflate, br
Accept-Language: en,en-US;q=0.7,en;q=0.3
```

There's no UA 😂... but it works?

return42: > With impersonate, I don't know what it is exactly but something in those headers it sends gets blocked.

Deepening on what kind of request we do, we have to overwrite some of the default headers curl-impersonates generates .. 

```python
from curl_cffi.requests import get, post
for imp in ["chrome", "firefox"]:
    resp =get("https://httpbin.org/headers",impersonate=imp)
    print(f"# GET impersonate: {imp}")
    print(resp.text)
    resp =post("https://httpbin.org/post", impersonate=imp, data= {"K": "V"})
    print(f"# POST impersonate: {imp}")
    print(resp.text)
```

For my example I use the firefox ..

```JSON
# GET impersonate: firefox
{
  "headers": {
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8", 
    "Accept-Encoding": "gzip, deflate, br, zstd", 
    "Accept-Language": "en-US,en;q=0.9", 
    "Host": "httpbin.org", 
    "Priority": "u=0, i", 
    "Sec-Fetch-Dest": "document", 
    "Sec-Fetch-Mode": "navigate", 
    "Sec-Fetch-Site": "none", 
    "Sec-Fetch-User": "?1", 
    "Upgrade-Insecure-Requests": "1", 
    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:147.0) Gecko/20100101 Firefox/147.0", 
    "X-Amzn-Trace-Id": "Root=1-6aa101da-694c2172079fa1176e785f97"
  }
}
```

When we call the `d.js` endpoint from ddg we have to overwrite at least the `Accept` and the `Sec-Fetch-Dest` header.

To compare, here is what headers a real Firefox will send to the `d.js` endpoint ..

grafik

-----

May OT: I would prefer to impersonate `firefox` since Chrome tends to send more headers and IDK if we have to overwrite the browser specific `Sec-Ch-...` headers depending of what kind of request we do ... 

```JSON
# GET impersonate: chrome
{
  "headers": {
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7", 
    "Accept-Encoding": "gzip, deflate, br, zstd", 
    "Accept-Language": "en-US,en;q=0.9", 
    "Host": "httpbin.org", 
    "Priority": "u=0, i", 
    "Sec-Ch-Ua": "\"Not;A=Brand\";v=\"8\", \"Chromium\";v=\"150\", \"Google Chrome\";v=\"150\"", 
    "Sec-Ch-Ua-Mobile": "?0", 
    "Sec-Ch-Ua-Platform": "\"macOS\"", 
    "Sec-Fetch-Dest": "document", 
    "Sec-Fetch-Mode": "navigate", 
    "Sec-Fetch-Site": "none", 
    "Sec-Fetch-User": "?1", 
    "Upgrade-Insecure-Requests": "1", 
    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36", 
    "X-Amzn-Trace-Id": "Root=1-6aa101d9-3973008e37cefa790a066b89"
  }
}
```

vojkovic: Thanks for this, I have done this in the latest force push with these headers. It still works

I also prefer to impersonate firefox. It wasn't a problem in this case but I think the TLS impersonation is slightly worse unfortunately because curl-impersonate is built with BoringSSL (used by Chrome and Safari) but Mozilla uses NSS so maybe just watch out for that IDK if it will be a problem in the future.

return42: > because curl-impersonate is built with BoringSSL (used by Chrome and Safari) but Mozilla uses NSS 

Thanks for the tip /  I hadn't in mind ... yes, we still need to gain some more experience .. it's not easy with a moving target, but we shouldn't lose sight of that point either.

return42 (approved):

Bnyro (approved):

PR #6660, opened 2026-09-05
[feat] Modify brave (api) search to show all the data from the API response.

gbeliako:
### What does this PR do?

Brave API returns a lot of data in the response which can be used by searxng.
This includes: infobox, discussions, faq, locations, news, videos, movie, q&a, book, article, restaurant. 

A valid brave API key is required, with:
api_key: 'key' in settings.yml

I've been using this new version on my instance for a few weeks.

### Future work:
- (current) results order is based on result type, not on brave's prefered [mixed] order.
- summary and rich results fields, which use some callback mode
- sometimes an infobox tag/field is an img (apart from thumbnail) which needs to be parsed/converted into img type.

I used (locally-hosted) qwen to help develop infobox and location result functions.

### DIFF
--- a/searx/engines/braveapi.py
+++ b/searx/engines/braveapi.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Engine to search using the Brave (WEB) Search API.
 
-.. _Brave Search API: https://api-dashboard.search.brave.com/documentation
+.. _Brave Search API: https://api-dashboard.search.brave.com/api-reference/web/search/get
 
 Configuration
 =============
@@ -13,6 +13,7 @@ The engine has the following mandatory setting:
 Optional settings are:
 
 - :py:obj:`results_per_page`
+- :py:obj:`brave_category` (search, videos, images, news, goggles)
 
 .. code:: yaml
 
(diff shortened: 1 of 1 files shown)


return42:
Thank you for your contribution; I appreciate your good intentions, but unfortunately, we cannot make much use of this AI-generated code. If you examine the code, you will see just how much superfluous code it contains.

AI can enormously boost developer productivity and product quality, but it cannot replace developers, testers, or reviewers.

We could generate unreviewed code from AI models by the thousand every minute ourselves; we don't need contributors for that. Instead, we have deliberately chosen a development model based on human collaboration—always with the aim of building expertise and offering our users a stable, maintainable product.

As is unfortunately so often the case with these submissions, our guidelines were completely disregarded here. If a PR confirms that our AI guidelines have been acknowledged, one should at least make the effort to actually read them.

O-ton: _"AI should never be the main author of the PR."_ .. _"There are Humans Here"_.

https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

We also have a quick-start guide for developers, but unfortunately, in this age of AI, even such a small bit of guidance isn't appreciated anymore—everyone just jumps right in, regardless of how helpful their contribution is or isn't.

https://docs.searxng.org/dev/quickstart.html

Please do not take offense at my bluntness; even though your contribution is well-intentioned, open-source projects are drowning in a flood of such submissions. They are a huge hindrance to open-source projects .. if only because maintainers like me are forced, several times a day, to draft messages that come across as unfriendly, just like this one. Yet, we could actually be doing much more valuable work for the community.

PR #6662, opened 2026-09-05
[mod] engines: remove uses of `enable_http2 = False`

Bnyro:
### What does this PR do?
`enable_http2 = False` has only been needed because we were flagged by HTTP2 fingerprinting. Now, since we use curl cffi, we can bypass the fingerpinting, so we can use HTTP2 just fine without getting blocked.

- use the engines

### Related issues
- #6620 

### Code of Conduct

### DIFF
--- a/searx/engines/luxxle.py
+++ b/searx/engines/luxxle.py
@@ -37,16 +37,12 @@ about = {
 }
 
 categories = []
-safeseach = True
+safesearch = True
 
 base_url = "https://luxxle.com"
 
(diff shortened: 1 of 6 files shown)


Bnyro:
The only remaining engine using this is cara, which however no longer works. So I opened https://github.com/searxng/searxng/pull/6663 to remove it instead of changing enable_http2.

vojkovic (approved):
Tested all six and they still work 👍

PR #6666, opened 2026-09-06
[fix] engine: braveapi - set JSON Accept header

### LINKED ISSUE #6665: Every braveapi search fails with HTTP 422 because request() never sets Accept: application/json
I discovered this issue while trying to configure Deepseek Harness to work with my SearXNG instance. DSH originally reported the issue as exceeded quota but the Brave dashboard showed no usage at all. After wasting way too much time trying to figure out an API key/ env var issue, I used my locally hosted Qwen model to test and troubleshoot and ultimately found the issue with its help. I double checked my local Qwen model’s work with Google and Gemini led me way off track looking at the recently closed issue 6545  “braveapi pagination sends an invalid offset.” After validating that fix was already in my image I ran the proposed fix by ChatGPT and it agreed with the fix. I double checked all the work and I think it's a straightforward issue and I'll also submit the PR.

Every braveapi search fails with HTTP 422 even with a valid API key. braveapi.py::request() sets only the X-Subscription-Token header and never sets Accept. The same request succeeds against the same endpoint as soon as Accept: application/json is sent.

### How To Reproduce?

Configure the official braveapi engine with a valid API key.
Enable json under search.formats
Run any query restricted to the engine: 
```
curl 'http://<searxng-instance>/search?q=test&format=json&engines=braveapi'
```

You can also see the issue with a direct curl:
This fails:

```
curl -s -H 'X-Subscription-Token: API_KEY' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' 'https://api.search.brave.com/res/v1/web/search?q=test&count=5'
```

Replace the Accept header with 'Accept: application/json' and it is successful:

```
curl -sS -H 'X-Subscription-Token: API_KEY' -H 'Accept: application/json' 'https://api.search.brave.com/res/v1/web/search?q=test&count=5'
```

Note: Brave free tier is ~1 req/s: space the two calls, or a 429 will masquerade as a second failure mode (this cost me an hour).

### Expected behavior

Search results.

### Screenshots & Logs

Container log excerpt:

```text
2026-09-06 07:03:31,466 WARNING:searx.network.braveapi: HTTP Request failed: GET https://api.search.brave.com/res/v1/web/search?q=kubernetes+sidecar+pattern&count=20&offset=0&text_decorations=False
2026-09-06 07:04:33,625 WARNING:searx.network.braveapi: HTTP Request failed: GET https://api.search.brave.com/res/v1/web/search?q=kubernetes+sidecar+pattern&count=20&offset=0&text_decorations=False
```

Brave response body (AI agent captured):

```json
{
  "type": "ErrorResponse",
  "error": {
    "id": "REDACTED",
    "status": 422,
    "detail": "Unable to validate request parameter(s)",
    "meta": {
      "errors": [
        {
          "type": "enum",
          "loc": ["header", "accept"],
          "msg": "Input should be 'application/json' or '*/*'",
          "input": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
          "ctx": {
            "expected": "'application/json' or '*/*'"
          }
        }
      ]
    },
    "code": "VALIDATION"
  },
  "time": 1788684948
}
```

### Version of SearXNG

not forked
2026.9.5+c7f3080aa

Official documentation using docker.io/searxng/searxng

### Additional context

n/a

### Code of Conduct

- [x] I read the [AI Policy] and hereby confirm that this issue conforms with the policy

Backroads4Me:
### What does this PR do?

This sets `Accept: application/json` explicitly at the end of
`braveapi.request()`, so the engine asks for the media type it parses instead
of inheriting whatever the transport supplies.

The added unit test exercises `request()` with both a missing `Accept` header
and an HTML-style one, so neither state can reintroduce the failure.

Automated validation:

- All 12 engine tests pass.
- Black and Pylint pass.
- Type checking reports zero errors and 17 warnings unchanged from the unmodified engine.
- The regression test fails without the fix and passes with it.

For manual testing, add a valid Brave API key.
Enable `json` under `search.formats`.
Run any query restricted to the engine:

```bash
curl 'http://<searxng-instance>/search?q=test&format=json&engines=braveapi'
```

The search should return results, and the JSON response's `unresponsive_engines` should contain no `braveapi` error.

### Related issues

Closes #6665

### Code of Conduct

I used the following AI tools during this work:

- Local Qwen 3.8 Flash Next: Used for initial debugging, log analysis, and identifying the missing `Accept` header.
- Gemini: Used for background research on historical `braveapi` issues.
- ChatGPT: Used for final validation of the proposed code fix and unit test.

-----

Related: https://api-dashboard.search.brave.com/api-reference/web/search/get

grafik


Bnyro (changes requested):
Please remove the test, it doesn't bring any value and just makes it more complicated to maintain the engine.

If we can't even trust that assigning `params["headers"]["Accept"] = ...` actually assigns that value without writing a test case for it, then we'd be completely lost.

Backroads4Me:
Done. Test has been removed.

return42 (approved):
FYI: In the intro of this PR, I added a link to the relevant API documentation (and a screenshot)

Bnyro (approved):

Issue #4950, opened 2025-06-27
Startpage engine issues (to-do list)


useralias:
update from the maintainers: we have tagged startpage issue, see [Label "startpage"](https://github.com/search?q=repo%3Asearxng%2Fsearxng%20state%3Aopen%20label%3A%22startpage%22&type=issues)

## To-do list for the Startpage engine
Just a list of things to do to potentially fix the issues the Startpage engine has at the moment.
I plan on addressing them at some point, but if you do want to work on them comment here so the work doesn't get done twice.

### Form data
- [ ] Order the query parameters for first page and subsequent pages like a real request.
- [ ] Use a new sc code for every search request (currently cached for an hour for all searches)
- [ ] **lui** and **language** form parameter need to include the full language name (e.g. english instead of en)
- [ ] Add support for "all" region (not included at the moment)

### Cookies
- params['cookies']['preferences']
  - [ ] language, language_ui and search_results_region should always be present
- [ ] Add missing locationPref cookie

### Response parsing
- [x] Add missing CAPTCHA handling in main response function (only handled in get_sc_code)

### Extra notes
- abd and abe stand for "adblock detected" and "adblock enabled", they are both set to 0 if javascript is disabled (or there's no adblocker).
- Startpage has the possibility of making first page searches without an **sc** code (via GET requests, opensearch config and extension have this)
- Depending on if we want to mimic a search query that uses javascript or without, there are some obvious pings on response here:
  - No javascript: a request is sent to https://www.startpage.com/sp/nojst?bdst=[token] 
  - Javascript: a request is sent to https://www.startpage.com/sp/footerImage/cdn/images/startpage-logo-gradient.svg?sc=[code]&cdw=1&rpw=1 (&cdw=1&rpw=1 are included on adblock detection)

return42:
The Startpage search engine needs an overhaul, the reverse engineered implementation of the lang/region support is 3 years old .. can't believe it works any longer 

- https://github.com/searxng/searxng/pull/2269 

@Bnyro Maybe you'd like to take a closer look at this engine, if you have the time?

MayCXC:
Reverse-engineered the current Startpage request/cookie format from a live Firefox session (saved settings, then read `document.cookie`) on 2026-09-16, in case it helps with these items. Mapping to the to-do:

**`preferences` cookie (current, verbatim):**

```
[...]

PR #6669, opened 2026-09-06
[fix] engines: startpage anubis solver

### LINKED ISSUE #6520: Constant CAPTCHA error from Startpage
from Startpage during any search at https://sok.nordh.tech/

get_sc_code: got redirected to https://www.startpage.com/sp/captcha (suspended_time=3600)
searx/engines/startpage.py:209
get_sc_code
raise SearxEngineCaptchaException(
### How To Reproduce?
Just chose Startpage among the search engines

### Expected behavior

Search results from Startpage

### Version of SearXNG

Been like this for the last couple of builds, right now on 2026.8.12 ...

Not using a forked version

Using the official documentation of mid-April this year

### Additional context

no

### Code of Conduct

vojkovic:
### What does this PR do?

Startpage put anubis on their search endpoint. Unfortunately, there is no way to get around this with any impersonate.

The anubis challenge difficulty is currently set to 4 which is high enough so this engine has set `inactive: true` now, I don't think it should be on for everyone anymore by default. The max search is set to be 8 times - this is so it wont loop forever if startpage change something in the future.

Also fixed the language fallbacks to be the correct ones (still needs to be properly redone someday to address https://github.com/searxng/searxng/issues/4950).

Enable startpage
Test all startpage engines work

### Related issues

- Closes: https://github.com/searxng/searxng/issues/6520

### Code of Conduct

Cursor helped with the implementation.

### DIFF
--- a/searx/engines/startpage.py
+++ b/searx/engines/startpage.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
-"""Startpage's language & region selectors are a mess ..
+"""Startpage requires solving an Anubis POW captcha (difficulty 4).
+Solving it requires a lot of CPU, so the engine is set inactive by default.
+
+Startpage's language & region selectors are a mess ..
 
 .. _startpage regions:
 
@@ -84,6 +87,7 @@ Startpage's category (for Web-search, News, Videos, ..) is set by
 """
 # pylint: disable=too-many-statements
 
+import hashlib
 import re
 import typing as t
 from collections import OrderedDict
@@ -99,7 +103,7 @@ from searx.enginelib import EngineCache
 from searx.enginelib.traits import EngineTraits
 from searx.exceptions import SearxEngineCaptchaException
 from searx.locales import region_tag
-from searx.network import get  # see https://github.com/searxng/searxng/issues/762
+from searx.network import get, post  # see https://github.com/searxng/searxng/issues/762
 from searx.utils import (
     eval_xpath,
     extr,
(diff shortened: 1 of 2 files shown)


searx/engines/startpage.py
```diff
@@ -176,6 +177,45 @@ def setup(_: dict[str, t.Any]) -> bool | None:
 sc_code_cache_sec = 3600
 """Time in seconds the sc-code is cached in memory :py:obj:`get_sc_code`."""
 
+# startpage's anubis difficulty is set to 4
+max_difficulty = 4
+
+
+def _solve_anubis(resp) -> str:
```
Bnyro, line 187: Because Anubis is also used by other websites, we could possibly move this to `searx/utils.py` so that it can be used by other engines too?

Though I'm not sure if this solver here works for Anubis in general or if Startpage uses a special version that is a bit different. You probably know that better than me :)

vojkovic: TBH I do not know much about Anubis since I haven't used it in almost two years - but I know they have some weird rule where you can't change the turnstile image without paying the dev for a commercial license - and startpage has a special loading page. Anubis is also MIT licenced so maybe startpage just forked it lol because I don't see any reference to 'botstopper' (the commercial version).

And the cookie is called `spchal-auth` which is unique.

Looks like they are pissing off a lot of users with this change too so it will be interesting to see if they keep it.

- https://www.reddit.com/r/StartpageSearch/comments/1w62jmv/what_happened_to_startpage/
- https://www.reddit.com/r/StartpageSearch/comments/1w9zfoq/what_is_ts/
- https://www.reddit.com/r/StartpageSearch/comments/1w4hfw1/completely_unusable_lately/

TLDR - I think it is better to keep it in startpage.py :-)

return42 (approved):
LGTM 👍

Bnyro (approved):

searx/settings.yml
Bnyro, line 2345: Please also add a comment why it's inactive, i.e. that it solves a complex Anubis captcha that consumes a rather large amount of computer power.

Bnyro: Or add it to the Startpage engine's docstring.

vojkovic: Good point, thank you. Will do!

PR #5793, opened 2026-03-01
[enh] engines: rework bing engine

inetol:
Only Bing-Web has been reworked.

Some features now require JavaScript (paging and time-range results).

Cookies no longer work, parameters such as `cc`, `ui`, ... alter the results.

The engine only appears to use the locale from `Accept-Language` header properly.

The rest of Bing's child engines (Bing-Image, Bing-Video, ...) seem to benefit from using `mkt` param in conjunction with the `Accept-Language` header override, although Bing-Web does not (?)

---

Superseedes https://github.com/searxng/searxng/pull/3416

Closes https://github.com/searxng/searxng/issues/4627
Closes https://github.com/searxng/searxng/issues/4422
Closes https://github.com/searxng/searxng/issues/4386
Closes https://github.com/searxng/searxng/issues/3402
Closes https://github.com/searxng/searxng/issues/3358
Closes https://github.com/searxng/searxng/issues/2641

Issue #2977, opened 2023-10-30
Reuse of http client and SSL context leads to  difficult to assess side effects


return42:
When I start a local `make run` instance, I got what I expected:

- send query `!bing bmw :zh-CN` I got https://www.bmw.com.cn --> perfect / OK
- send query `!bing bmw :en-US` I got https://www.bmwusa.com --> perfect / OK

On my server where I have uWSGI I am observing something very strange:

1. resart my uwsgi instance
2. send query `!bing bmw :zh-CN` I got https://www.bmw.com.cn --> perfect / OK
3. send query `!bing bmw :en-US` I got https://www.bmw.com.cn again -- why ?
4. resart my uwsgi instance
5. send query `!bing bmw :en-US` I got https://www.bmwusa.com --> perfect / OK
6. send query `!bing bmw :zh-CN` I got https://www.bmwusa.com again -- why ?

It seems to me that bing only changes the language on some requests but not on all requests .. but what is striking is the fact that quitting and restarting the uwsgi allows an immediate switch of the language ... a once selected language does not "survive" a restart of the uwsgi ... and that makes me suspect that somewhere deep in the architecture (uwsgi or in the http / tcp) goes wrong.

With the help from @dalf I was able to eliminate the issue when I merged the new network stack and applied a patch that removes all http client & SSL optimization:

- merge: https://github.com/searxng/searxng/pull/2685
- AND patch 6a9bbcb3a to remove http client & SSL optimization

With this I was able to switch bing's language on every request (and without the need to restart uWSGI service).

I have occasionally had the feeling that engines behave differently on my server than in a local `make run` instance, but with no engine was this as noticeable as with bing.  Now I understand why @jazzzooo and I had such different experiences when testing the bing engine.

- https://github.com/searxng/searxng/pull/2822

Simply removing all optimizations was only a method to limit the unusual behavior to one component (the HTTP client and the SSL context). In the long run we have to answer the following questions:

A. What is the benefit of our optimizations? For this we would need a benchmark (with/without optimization).
B. Why does the problem only occur in a uWSGI service but not in a `make run` instance.
C. We need to further narrow down what exactly about an optimization can have negative side effects, for this we will have to descend into the depths of httpx and SSL.

dalf:
For reference, ten years ago:
* https://github.com/searx/searx/issues/100
* https://github.com/searx/searx/pull/192

dalf:
Here a micro benchmark on SSLContext creation time:

<details>

[...]

PR #6671, opened 2026-09-07
[fix] engines: bing first word results

### LINKED ISSUE #4964: bing: results are often unrelevant to the search query

vojkovic:
### What does this PR do?

Fixes the bing web engine, it was just using the first word of the query for the search and return random junk other times. see: https://github.com/vojkovic/searxng/issues/10

Swapped to use bing's setlang and cc params. I found `us`, `cn`, `ru` return complete garbage 100% of the time. I reckon that if you don't have an ip address from there it will just return garbage, so those three are skipped. Also removed accept language override because it didn't change anything anymore.

The other bing engines are unchanged and continue to work.

```
!bing searxng
!bing weather :en
!bing wetter berlin :de-DE
```

### Related issues

- Closes: https://github.com/searxng/searxng/issues/4964
- Related: https://github.com/vojkovic/searxng/issues/10

### Code of Conduct

Cursor helped fuzz bing with these params.

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -72,43 +72,21 @@ def get_locale_params(engine_region: str | None) -> dict[str, str] | None:
     return {"mkt": engine_region}
 
 
-def override_accept_language(params: "OnlineParams", engine_region: str | None) -> None:
-    """Override the ``Accept-Language`` header.
-
-    The default header built by :py:class:`~searx.search.processors.online.OnlineProcessor`
-    appends ``en;q=0.3`` as a fallback language::
-
-        Accept-Language: de,de-DE;q=0.7,en;q=0.3
-
-    Bing seems to better select the results locale based on the
-    ``Accept-Language`` value header.
-
-    This function is shared with :py:mod:`searx.engines.bing_images`,
-    :py:mod:`searx.engines.bing_news`, and :py:mod:`searx.engines.bing_videos`.
-    """
-
-    if not engine_region or engine_region == "clear":
-        return
-
-    lang = engine_region.split("-")[0]
-    params["headers"]["Accept-Language"] = f"{engine_region},{lang};q=0.9"
-
-
 def request(query: str, params: "OnlineParams"):
     """Assemble a Bing-Web request."""
 
     engine_region = traits.get_region(params["searxng_locale"], traits.all_locale)
-
-    override_accept_language(params, engine_region)
 
     query_params: dict[str, str | int] = {
         "q": query,
         "adlt": _safesearch_map.get(params.get("safesearch", 0), "off"),
     }
 
-    locale_params = get_locale_params(engine_region)
-    if locale_params:
-        query_params.update(locale_params)
+    if engine_region and engine_region != "clear":
+        lang, _, cc = engine_region.partition("-")
+        query_params["setlang"] = lang
+        if cc and cc not in ("us", "cn", "ru"):  # bing just sends junk for these
+            query_params["cc"] = cc
 
     params["url"] = f"{base_url}/search?{urlencode(query_params)}"
 
--- a/searx/engines/bing_images.py
+++ b/searx/engines/bing_images.py
@@ -9,7 +9,6 @@ from lxml import html
 from searx.engines.bing import (  # pylint: disable=unused-import
     fetch_traits,
     get_locale_params,
-    override_accept_language,
 )
 
 # about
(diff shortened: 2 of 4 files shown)


tiekoetter (approved):

searx/engines/bing.py
```diff
@@ -99,16 +99,16 @@ def request(query: str, params: "OnlineParams"):
 
     engine_region = traits.get_region(params["searxng_locale"], traits.all_locale)
 
-    override_accept_language(params, engine_region)
-
```
return42, line 103: I could imagine that we could also remove this function for Bing News, Images, and Videos... the Accept-Language header is already being set by the online processor.

https://github.com/searxng/searxng/blob/c7f3080aac5de13b619c4a5ab36590a2c5165e1c/searx/search/processors/online.py#L153-L159 

https://github.com/searxng/searxng/blob/c7f3080aac5de13b619c4a5ab36590a2c5165e1c/searx/engines/__init__.py#L52

The function was added in:

- https://github.com/searxng/searxng/pull/5793

@inetol what do you think?

Bnyro (approved):
Good find, works for me too 👍

searx/engines/bing.py
inetol: Good for me :+1:

inetol (commented):
I don't see any improvements nor regressions in my case.

[We can also remove the override lang function](https://github.com/searxng/searxng/pull/6671/changes#r3947640013) as it does nothing now

vojkovic:
> I don't see any improvements nor regressions in my case.

Do you mean it has always been working for you or that there's issues? @inetol 

---

Bing is a really weird engine... I've been working on it for the best of half a day and now it is going back to the same sort of results:

Screenshot

I'm moving this PR to a draft while I see if there's a proper fix.

Bnyro:
> Bing is a really weird engine... I've been working on it for the best of half a day and now it is going back to the same sort of results:
>
@inetol and I (and probably @return42 too) already spent at least one day each to fix the bing engine. Usually when the engine works for one person, it breaks for the others that test it :/

inetol:
> Do you mean it has always been working for you or that there's issues? @inetol

I haven't had this kind of issue with Bing in the first place, even with long queries. The engine results on both this PR and master are the same.

*btw, when I did the Bing rework, [I found an RSS format endpoint](https://blogs.bing.com/search/January-2005/RSS-Feeds-for-Search-Results), but neither safesearch nor pagination worked there, and for the most part the selected lang didn't either (it was inferred on the client AS location). Could be useful?*

Bnyro:
> _btw, when I did the Bing rework, [I found an RSS format endpoint](https://blogs.bing.com/search/January-2005/RSS-Feeds-for-Search-Results), but neither safesearch nor pagination worked there, and for the most part the selected lang didn't either (it was inferred on the client AS location). Could be useful?_
>
I think that is very cool, certainly worth a new engine. That could also be used as a fallback if the bing engine ever gets blocked.

return42:
> @inetol and I (and probably @return42 too) already spent at least one day each to fix the bing engine. Usually when the engine works for one person, it breaks for the others that test it :/

Sadly true .. summarized over the years, I wasted weeks for the bing engine:

I gave up on Bing a long time ago .. Bing exhibits quirks in many respects: There were (there are?) issues when trying to use Bing via a VPN (even as user in a web browser) .. and once Bing has associated a specific IP address with a region (or marketplace?), it continues to respond to queries with different regional settings by returning results from the previously associated marketplace.  If you’re developing for the Bing engine and don’t switch the language or region during your tests, you’ll eventually get somewhat usable results; but as soon as you try search queries with different languages or regions, the engine starts returning crazy results.

The behavior Bing exhibits is hard to describe because it’s hardly rational and follows an incomprehensible logic. My impression has always been that Microsoft’s product development is based on idiotic assumptions that are far removed from practical reality.

To me, Bing is just as broken as all the other products from the M$ camp — products built on some commercial motives that I’ll probably never understand.

return42:
Another thought just occurred to me: one reason for the strange results could be the clients we're reusing.  It might be worth revisiting my observations from back then.

forget the words about "uWSGI" in there, the relevant section: *".. and applied a patch that removes all http client & SSL optimization .."* .. this patch had removed reusing of HTTP client.

return42 (approved):
>  I found us, cn, ru return complete garbage 100% of the time.

The behavior can also be reproduced/confirmed from my IP address (DE).

Issue #5900, opened 2026-03-24
Often the images of brave's thumbnails can't be loaded.


return42:
In a `make run` instance, try for example `!brave :en bmw` ..

<img width="655" height="878" alt="Image" src="https://github.com/user-attachments/assets/c6da3546-ea48-4b0f-aac4-7c7326828ad2" />

PR #6679, opened 2026-09-08
[mod] theme: improve thumbnails display on results page

vojkovic:
### What does this PR do?

Deletes thumbnails instead of showing an error image and also moves them to the right side. Also gives them a background and makes them into rounded squares for better ui/ux. I took inspiration from brave's web search.

Thumbnails currently sit on the left side and can be all sorts of shapes, leading to some strange layouts. This also means if we delete thumbnails after page loading it causes a very bad layout shift which is reduced by moving them to the right side.

Screenshot

Check thumbnails look more like this:
Screenshot

Check that rtl layout has the opposite

### Related issues

- https://github.com/searxng/searxng/issues/5900 (Doesn't fix it but makes it less jarring)

### Code of Conduct

Cursor assisted with the implementation


return42 (approved):
A major improvement to the UX 👍 


--- 
The rounded corners make it clear that this is just a cropped section of an image. In the past, without these rounded corners, it looked like a full image... and when the aspect ratio didn't match, it always looked al little “off.”

tiekoetter (approved):

client/simple/src/js/router.ts
```diff
@@ -3,6 +3,22 @@
...
+    if (!(img instanceof HTMLImageElement && img.classList.contains("thumbnail"))) return;
+    img.closest("a.thumbnail_link")?.remove();
+  },
+  true
+);
+for (const img of document.querySelectorAll<HTMLImageElement>("img.thumbnail")) {
+  if (img.complete && img.naturalWidth === 0) {
+    img.closest("a.thumbnail_link")?.remove();
+  }
+}
```
Bnyro: I don't think this should be in router.ts, why not put it into results.ts where the other thumbnail error code is as well?

Bnyro (commented):
Nice work 👍 

I'm a bit concerned about making all thumbnails squares. Most thumbnails are (probably?) 16/9, so by cropping them to a square, 43% of the image just gets lost / hidden.

E.g. as seen here, where the thumbnails become kind of useless because the text is cut off.
Image

I tried experimenting with it, and produced

Image

and 

Image

But I'm not really happy with either of them, so idk...

client/simple/src/js/router.ts
vojkovic: True, actually while I was looking at the other thumbnail error code I realised because it's not in an eventlistener, the infinite scroll would be broken so I'll fix that as well. 
Screenshot

vojkovic:
Thanks for looking into other options, but out of these I think I still prefer the square icons.

> Most thumbnails are (probably?) 16/9

I do agree it does look a little off with 16/9 images too, especially videos. I have no idea if we can detect specifically videos but it would be nice if we could do what brave and google do and move them like this but IDK if this is really possible. I think this is possible via .result-videos


Screenshot
Screenshot

EDIT:

I'm also sort of coming around to how [degoog](https://degoog.org/search?q=dog) does it, I think this looks really good, a bit of a compromise I think it is 4:3? And it has a good size, I think the current size I chose is a bit too big.

image

Actually after using [degoog](https://github.com/degoog-org/degoog) a little more, I am reminded by how beautiful the frontend is, the maintainer really did a good job - and [it can use searxng under the hood](https://github.com/degoog-org/degoog/blob/aeaaa4228ed019e2226b9edad6bd3eeb92c6d311/src/server/extensions/compatibility-layer/searx/catalog.ts). Shows what we might be able to do someday :-)

vojkovic:
In my latest commit, I made the desktop icons now into 4:3 rectangles, which I think is a nice compromise - (on mobile, due to space I kept it as squares). I also set them to fit to the box and placed a blurred background behind them. This gives us the best of both worlds and lets us always see the full image without needing to faff around with the size of the box.

Screenshot


And put the video thumbnails back the left. These are 16/9 always:
Screenshot

What do you think now @Bnyro ?

searx/templates/simple/result_templates/packages.html
```diff
@@ -70,6 +70,7 @@
     </div>
   {%- endif -%}
 </div>{{- '' -}}
+</div>
```
Bnyro, line 73: How did this get in here? 👀 

I don't think there's a matching `<div>`?

Bnyro (commented):
Looks nice, I also like that videos are now just 16/9 👍 

I feel like the blur is currently too "eye-catching", I think it should be more subtle so that it is less visible.

The thing that I would change is the `scale` on the background image, in my opinion it looks better without it. I.e. the scale causes everything to not align anymore, e.g. in the first image here (Rust) the orange part in the background is much more on the outside than in the actual image. Or for the Mona Lisa, you can still see a lot of details of her in the blurred background, which makes it look weird.

E.g. with scale:

Image

Image

Without scale:
Image

Image

Or what also looks better is changing the blur to something like `0.8rem`:

Image

Image

searx/templates/simple/result_templates/packages.html
vojkovic: result_header now opens a new div: <div class="result_inner"> so it is there to close it

Bnyro:
Sorry for all the nitpicks, but we will all see the change daily when using SearXNG, so I want it to be close to perfect :face_holding_back_tears:

vojkovic:
Def not nitpicks! Thanks for reviewing... I removed the scale.

> Or what also looks better is changing the blur to something like 0.8rem:

Do you also want me to change the blur? Up to you, both look the same to me :P

Bnyro:
> Do you also want me to change the blur? Up to you, both look the same to me :P
>
I think removing the scale is enough :+1:

Bnyro (approved):
Nice, thank you!

Issue #4307, opened 2025-02-11
Bug: mojeek engine


cameronj86:
**Version of SearXNG, commit number if you are using on master branch and stipulate if you forked SearXNG**
Repository: https://github.com/searxng/searxng
Branch: master
Version: 2025.2.3+4ab7984ed+dirty

**How did you install SearXNG?**
Docker, but it is happening across various public instances as well
**What happened?**
Errors out

**How To Reproduce**
`!mjk test`

**Expected behavior**
Results

**Additional context**
I also tried the following:
* Visiting mojeek directly to make sure my IP wasn't blocked
* Visiting public instances to make sure that it was widespread
* Selecting instances that don't use ipv6 (as per https://github.com/searxng/searxng/discussions/3188) to confirm that the issue is unrelated
* Changed `outgoing:   source_ips: 0.0.0.0` in my `settings.yml` file to make sure ipv6 wasn't the issue on my end.

Error
  * Error: searx.exceptions.SearxEngineAccessDeniedException
  * Percentage: 100
  * Parameters: `('HTTP error 403',)`
  * File name: `searx/search/processors/online.py:116`
  * Function: `_send_http_request`
  * Code: `response = req(params['url'], **request_args)`

jj-code-00:
Having the same issue.

shinodark:
It seems to me that "&fmt=" argument in generated url by searxng triggers an automated tools detection from mojeek

log from searxng with http 403 triggered : 
 WARNING:searx.network.mojeek: HTTP Request failed: GET https://www.mojeek.com/search?q=searxng&safe=0&fmt=&lb=fr&arc=fr&s=0

With manually removed fmt arg it seems ok :
https://www.mojeek.com/search?q=searxng&safe=0&lb=fr&arc=fr&s=0

return42:
@shinodark thanks a lot for pointing out 👍 .. today I merged https://github.com/searxng/searxng/pull/4354 to fix the issue.

elvisisvan:
still happening in version `2025.2.20+28d1240fc` on multiple instances (tested on https://search.ononoki.org and https://opnxng.com)
![Image](https://github.com/user-attachments/assets/af3a861e-17ae-4844-989a-9bacca92a9a6)

cameronj86:
`2025.2.20+caf0dd537` is working as expected w/ my self-hosted version. The public instances on searx.space have moved beyond this commit so I can't test publicly, but perhaps something in the more recent commits are playing a role here.

[...]

PR #6696, opened 2026-09-11
 [feat] mojeek: automatically solve altcha captchas 

Bnyro:
### What does this PR do?
- add a CAPTCHA solver for Mojeeks Altcha CAPTCHAs
- they seem to occur no matter the IP - before these changes, Mojeek always returned 0 results

- enable mojeek and use it, e.g. !mjk foo

### Related issues
- #4307

### Code of Conduct

No AI used.

<details>
<summary>Python script for testing</summary>

```py
from hashlib import pbkdf2_hmac
import json
import codecs
import random
import requests
import typing as t

challenge = requests.get("https://www.mojeek.com/captcha/challenge").json()
print(challenge)


def _array_startswith(arr: bytes, prefix: bytes) -> bool:
    return arr[: len(prefix)] == prefix


def solve_altcha(parameters: dict[str, t.Any], maxCounter: int = 1000) -> tuple[str, int] | None:
    """Solves Altcha CAPTCHAs. It derives keys using PBKDF2 until the derived
    key starts with the ``keyPrefix``` from the challenge. If the solver does not
    finish after ``maxCounter`` iterations, returns ``None``."""

    nonce = bytes.fromhex(parameters["nonce"])
    salt = bytes.fromhex(parameters["salt"])
    keyPrefix = bytes.fromhex(parameters["keyPrefix"])
    cost = parameters["cost"]
    keyLength = parameters["keyLength"]
    # e.g. "PBKDF2/SHA-256" -> "sha256"
    hashAlgorithm = parameters["algorithm"].split("/")[-1].replace("-", "").lower()

    counter = 0
    while counter < maxCounter:
        # Appends the counter to the nonce, i.e. writes the counter as a big-endian 32-bit integer.
        secret = nonce + counter.to_bytes(length=4)

        key = pbkdf2_hmac(hash_name=hashAlgorithm, password=secret, salt=salt, iterations=cost, dklen=keyLength)
        if _array_startswith(key, keyPrefix):
            return key.hex(), counter

        counter += 1

    return None


key, counter = solve_altcha(challenge["parameters"])
solution = {
    "challenge": challenge,
    "solution": {"counter": counter, "derivedKey": key, "time": random.randint(100, 200)},
}
solution_encoded = codecs.encode(json.dumps(solution).encode(), "base64").decode()
resp = requests.post("https://www.mojeek.com/captcha/verify", files={"altcha": (None, solution_encoded)})
print(resp.cookies["chllg"])
````
</details>

### DIFF
--- a/searx/utils.py
+++ b/searx/utils.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 """Utility functions for the engines"""
 
+from hashlib import pbkdf2_hmac
 import time
 
 import re
@@ -811,3 +812,34 @@ def format_duration(duration: str | int) -> str:
     if length.tm_hour:
         return time.strftime("%H:%M:%S", length)
     return time.strftime("%M:%S", length)
+
+
+def _array_startswith(arr: bytes, prefix: bytes) -> bool:
+    return arr[: len(prefix)] == prefix
+
+
+def solve_altcha(parameters: dict[str, t.Any], maxCounter: int = 1000) -> tuple[str, int] | None:
+    """Solves Altcha CAPTCHAs. It derives keys using PBKDF2 until the derived
+    key starts with the ``keyPrefix``` from the challenge. If the solver does not
+    finish after ``maxCounter`` iterations, returns ``None``."""
+
+    nonce = bytes.fromhex(parameters["nonce"])
+    salt = bytes.fromhex(parameters["salt"])
+    keyPrefix = bytes.fromhex(parameters["keyPrefix"])
+    cost = parameters["cost"]
+    keyLength = parameters["keyLength"]
+    # e.g. "PBKDF2/SHA-256" -> "sha256"
+    hashAlgorithm = parameters["algorithm"].split("/")[-1].replace("-", "").lower()
+
+    counter = 0
+    while counter < maxCounter:
+        # Appends the counter to the nonce, i.e. writes the counter as a big-endian 32-bit integer.
+        secret = nonce + counter.to_bytes(length=4)
+
+        key = pbkdf2_hmac(hash_name=hashAlgorithm, password=secret, salt=salt, iterations=cost, dklen=keyLength)
+        if _array_startswith(key, keyPrefix):
+            return key.hex(), counter
+
+        counter += 1
+
+    return None
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -17,7 +17,6 @@ def random_string(length, choices=string.ascii_letters):
 
 
 class TestUtils(SearxTestCase):
-
     def test_gen_useragent(self):
         self.assertIsInstance(utils.gen_useragent(), str)
         self.assertIsNotNone(utils.gen_useragent())
@@ -116,7 +115,6 @@ class TestUtils(SearxTestCase):
 
 
 class TestXPathUtils(SearxTestCase):  # pylint: disable=missing-class-docstring
-
     TEST_DOC = """<ul>
         <li>Text in <b>bold</b> and <i>italic</i> </li>
         <li>Another <b>text</b> <img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs="></li>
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: searx/engines/mojeek.py [hunk 1]
```diff
@@ -51,13 +63,44 @@
...
+    key, counter = solution
 
-def request(query, params):
+    solution = {
+        "challenge": challenge,
+        "solution": {"counter": counter, "derivedKey": key, "time": random.randint(100, 200)},
+    }
+    solution_encoded = codecs.encode(json.dumps(solution).encode(), "base64")
+    mp = curl_cffi.CurlMime()
+    mp.addpart(name="altcha", data=solution_encoded)
```
Bnyro, line 95: I don't really like this - i.e. we directly access curl_cffi here instead for sending multipart data instead of using the network abstraction. But I don't think we have an inbuilt way to send multipart form data?

searx/engines/mojeek.py
```diff
@@ -51,13 +63,44 @@
...
+    if not solution:
+        raise SearxEngineAPIException("failed to solve CAPTCHA")
+    key, counter = solution
 
-def request(query, params):
+    solution = {
+        "challenge": challenge,
+        "solution": {"counter": counter, "derivedKey": key, "time": random.randint(100, 200)},
+    }
+    solution_encoded = codecs.encode(json.dumps(solution).encode(), "base64")
```
vojkovic:
```suggestion
    solution_encoded = base64.b64encode(json.dumps(solution).encode())
```

I've always used base64, is there a difference?

searx/engines/mojeek.py
```diff
@@ -51,13 +63,44 @@
...
+"""Cache for storing the auth cookie after solving the CAPTCHA."""
+
 
-def setup(_: dict[str, t.Any]) -> bool | None:
+def setup(engine_settings: dict[str, t.Any]) -> bool | None:
     if search_type not in ("", "images", "news"):
         raise ValueError(f"Invalid search type {search_type}")
 
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
```
vojkovic:
```suggestion
    CACHE = EngineCache(engine_settings["engine"])
```

We can share this token with all the mojeek engines.

vojkovic (commented):
Another great PR, two comments from me

searx/engines/mojeek.py
Bnyro: I think the codecs package is just a wrapper around all kinds of codecs that exist, so it's a convenience method to access all encodings that Pythons stdlib supports ("ascii", "utf8", "base64", ...).

So yeah, the base64 package is probably the better choice.

searx/engines/mojeek.py
Bnyro: I also put them into the same `network` (not sure how the network stuff works after the curl cffi migration, but I assume that it still does something. It won't hurt to set it ig).

searx/engines/mojeek.py
```diff
@@ -51,13 +63,44 @@
...
+        "challenge": challenge,
+        "solution": {"counter": counter, "derivedKey": key, "time": random.randint(100, 200)},
+    }
+    solution_encoded = base64.b64decode(json.dumps(solution).encode())
+    mp = curl_cffi.CurlMime()
+    mp.addpart(name="altcha", data=solution_encoded)
+    resp = post(f"{base_url}/captcha/verify", multipart=mp)
+
+    token = resp.cookies["chllg"]
+    CACHE.set("chllg", token)
```
return42, line 99: AFAIK the cookie is valid for 8 or 9 hours.  Our default for the engine cache is 7 days:

https://github.com/searxng/searxng/blob/461f174b09fc151f49257aa5206417aca8930efa/searx/enginelib/__init__.py#L46-L56

searx/settings.yml
```diff
@@ -2804,24 +2830,30 @@ engines:
   - name: mojeek
     shortcut: mjk
     engine: mojeek
+    network: mojeek
```
return42: Causes a key error (is it only me?)

```
File "/share/searxng/searx/network/network.py", line 412, in initialize
    NETWORKS[engine_name] = NETWORKS[network]
                            ~~~~~~~~^^^^^^^^^
KeyError: 'mojeek'
```

As far as I recall, the network is named after the engine; if you want to run other engines using the same client, all subsequently configured engines must be added to this automatically created network.

searx/engines/mojeek.py
return42: @Bnyro you modified to `base64.b64decode` .. while @vojkovic suggested `base64.b64encode`

With *decode* you will get:

```
Invalid base64-encoded string: number of data characters (441) cannot be 1 more than a multiple of 4
```

The error probably didn't occur for you during the test because you still had a valid value in the cache... we always have to clear the cache before running tests. `rm /tmp/sxng_cache_ENGINES_CACHE.db`

searx/engines/mojeek.py
Bnyro: Looks like it's valid one whole month:
image

So I'll keep the default of 7 days.

searx/settings.yml
Bnyro: Thanks! I never used the networks (and didn't test it either :) )

searx/engines/mojeek.py
Bnyro: Whoops, sorry!

I didn't test it again after the changes because I didn't expect anything to break from them (I should have tested it...).

Fixed now. And tested :)

return42 (approved):
👍

vojkovic (approved):

PR #6690, opened 2026-09-10
[mod] defaults: disable image engines with poor results

vojkovic:
### What does this PR do?

The image search defaults are a bit of a mix of results atm with random stock photos, some broken thumbnails (artic) and svg icons.

Following user feedback, I have set `disabled: true` for these engines: pinterest, flickr, pexels, unsplash, openverse, lucide, devicons, artic, bing images

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [x] **I hereby confirm that this PR conforms with the [AI Policy].**

  If I have used AI tools for working on the changes in this PR, I will
  attach a list of all AI tools I used and how I used them. I hereby confirm
  that I haven't used any other tools than the ones I mention below.

No ai

PR #6699, opened 2026-09-12
[fix] engine: bing images - remove market place argument mkt

return42:
### commit ``[fix] engine: bing images - remove market place argument mkt``

The ``mkt`` argument does no longer exists and the ``async`` argument has been renamed to ``mmasync``.

For still unknown reasons (IP based?), some users had to observe that the title was missing [1].  I myself was not able to reproduce this error, however the evaluation of the title was additionally expanded by the attribute value of the ``<a title=".."`` element.

Related:

- [1] https://github.com/searxng/searxng/pull/6690#issuecomment-5636811572

### commit ``[mod] engine: bing images - use internal API & type hints``

Switching to internal APIs such as ``EngineResults`` and adding more type hints.  No functional changes.

    !bing_images :en-US tony kroos
    !bing_images :de tony kroos

The language- and region-dependent search works quite well according to my observation. However, if Bing-Images can only deliver few results, for example because the query was restricted in time (last day), then Bing starts to include hits in the result list that are biased by the IP of the request.

Bing Images and the Bing Web API both use the ``cc`` argument; unlike with the Bing Web search, however, I could not detect any problems with the image search when using "us" as the region.

### Code of Conduct

  no AI used

### DIFF
--- a/searx/engines/bing.py
+++ b/searx/engines/bing.py
@@ -62,8 +62,8 @@ def get_locale_params(engine_region: str | None) -> dict[str, str] | None:
 
     The ``mkt`` parameter takes a full ``<language>-<country>`` code.
 
-    This function is shared with :py:mod:`searx.engines.bing_images`,
-    :py:mod:`searx.engines.bing_news`, and :py:mod:`searx.engines.bing_videos`.
+    This function is shared with :py:mod:`searx.engines.bing_news`, and
+    :py:mod:`searx.engines.bing_videos`.
     """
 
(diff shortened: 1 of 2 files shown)


vojkovic (approved):
Works perfectly

Image

What do you think, should we reenable `bing_images` again in https://github.com/searxng/searxng/pull/6690?

return42:
> What do you think, should we reenable `bing_images` again in #6690?

I would vote for .. it was my motivation to give bing another try even I was frustrated of Bing :-)

PR #6700, opened 2026-09-12
[fix] engines: dogpile web needs api token

vojkovic:
### What does this PR do?

Fixes the dogpile web engine. Now needs a token like the other categories.

### Code of Conduct

no ai


return42 (approved):
> ogpile web engine. Now needs a token like the other categories

Confirmed before / after .. 👍

PR #6703, opened 2026-09-12
[mod] add new category 'stock_images' - a specialization of 'images'

return42:
#### [mod] add new category ~'gallery'~ `stock_images` - a specialization of 'images'

Engines like Pexels provide images that one would more likely expect in a category named `stock_images`.

By classifying engines like Pexels more precisely, we enable a more specific handling of them, which also helps to avoid mixing them with the more general image search on the internet. [1]

We leave these engines *enabled* by default, take them out of ``images`` and move them into the specialization ``stock_images``.

This makes it possible, on the one hand, for the admin to configure a tab in the UI, and on the other hand, the user can always select the group directly using the search syntax ``!stock_images ...``.

- [1] https://github.com/searxng/searxng/pull/6690#issuecomment-5646291679

### Code of Conduct

  noAI


return42:
You now have a better overview of the content provided by each engine. Which engines more, if any, could we move from “images” to “gallery”?

vojkovic:
Would it be better named `stock_images` or something similar, probably not `stock` as that might be confused with financial stocks but gallery is a bit vague.

Some engines that will def fit:
- unsplash
- pixabay images
- stocksnap
- picjumbo
- shopify stock
- adobe stock
- 1x
- 500px
- flickr

(Maybe also openverse and the public domain image archive)

return42:
> Would it be better named `stock_images` or something similar,

I often use the search syntax and `!stock_images` are 13 chars long and one is a underline, compared to `!gallery` its a hard to type name.  I don't know how `gallery` comes across to a native speaker; I was hoping it would work as an umbrella term for everything that, in a broader sense, has a *artistic or commercial value*.

> Some engines that will def fit:

- unsplash
- pixabay images
- stocksnap
- picjumbo
- shopify stock
- adobe stock
- 1x
- 500px
- flickr

The engines you list are, in fact, fit into a classification named "stock images". I think we need to define the classification more clearly.  By that, I also mean what expectations we have for a "general image search".  A clear distinction is usually not possible, since even our sources already offer a mix of content.

Where, for example, would you classify an engine that provides paintings by artists?

- general image search
- a category named “Gallery”
- something named “XYZ”

BTW: we already have a image category named ["icons"](https://docs.searxng.org/user/configured_engines.html#group-icons) but we also have a very long list of image engines [without further subgrouping](https://docs.searxng.org/user/configured_engines.html#id3).

vojkovic:
> I often use the search syntax and !stock_images are 13 chars long and one is a underline

I mean we already have `!social_media` which is the same. When I think of gallery, it brings up more thoughts of like arts and paintings.

> Where, for example, would you classify an engine that provides paintings by artists?

Hmmmm tricky good point, how about this? - I think we should split this into two subcategories of images instead of a whole new category:

- `stock_images` : For the stock image engines, like those I mentioned
- `gallery` : For the art engines (artic, artstation, deviantart, etc...)

return42:
> I mean we already have `!social_media` which is the same.

you are right, for the engines you suggested lets use category name `stock_images`.

> I think we should split this into two subcategories of images instead of a whole new category:

What we need to keep in mind:

1. Technically speaking, there are no subcategories in the strict sense .. organizationally, the *subcategories* are derived from the `categories_as_tabs`.

2. From an admin’s perspective, `categories_as_tabs` is the tool for structuring their content.

3. In the context of “enabled/disabled by default”: We should avoid having the admin have to activate many individual engines when they want to structure their content (UI tabs).

So if we leave an engine enabled by default and **move** it from `images` to a new category, the admin only needs to add the new category to `categories_as_tabs` to restructure their content. If the new category is also part of the translations (see `searx/searxng.msg` in this PR), this UI structure would also be available internationally.

Maybe I’m taking my thoughts a bit too far.

return42:
I've updated the patch based on our previous discussions; see 22fddef2a.

Do you have any other suggestions that we might want to consider?

vojkovic:
Apologies, missed the message - nothing else from me. LGTM

PR #6704, opened 2026-09-12
[fix] neocities: very slow due to large result titles

Bnyro:
### What does this PR do?
- limit neocities titles to 200 characters
- e.g. the second result at https://neocities.org/search?q=Firefox+is+already+running%2C+but+is+not+responding has a much too long title

- !neocities Firefox is already running, but is not responding
- observe that before these changes, this loads a veeery long time
- I'm not sure why it takes SearXNG so long to render long titles, but this partially DOSed my instance for some days

### Code of Conduct

No AI Used.


vojkovic:
Would it be better to do this for all engines in the result type? And maybe we could append `...` to indicate the title has been truncated.

Could also be done for the description field

Bnyro:
Thanks @vojkovic, agreed. Done in https://github.com/searxng/searxng/pull/6713. I hope that this doesn't break any engines that have important info hidden in their very long titles / contents, but I think it should be fine.

PR #6705, opened 2026-09-12
Baidu engine: suspend for a short time on CAPTCHA, handle HTML soft-block

kosmo888:
## What this changes

Two changes in `searx/engines/baidu.py`, both about how the engine reacts to a
Baidu CAPTCHA. It does **not** make the Baidu JSON endpoint searchable again —
that needs a separate change — it fixes the failure handling around it.

**1. Suspend for 300 s instead of the global default (86400 s)**

A Baidu CAPTCHA is short-lived. The engine used the global
`search.suspended_times.SearxEngineCaptcha` default of 86400 s (1 day), so a
single CAPTCHA disabled Baidu for a whole day, long after the site was
reachable again. Same pattern as `searx/engines/quark.py`, which already passes
`suspended_time=900` for the analogous Alibaba CAPTCHA.

**2. Catch `JSONDecodeError` and report it as a CAPTCHA**

When Baidu answers HTTP 200 with an HTML page instead of the expected JSON
payload, `json.loads(text, strict=False)` raised a `json.JSONDecodeError`. That
exception is not in the suspend list of
`OnlineProcessor._send_http_request()` (`SearxEngineCaptchaException`,
`SearxEngineTooManyRequestsException`, `SearxEngineAccessDeniedException`), so
it fell through to the broad `except Exception` branch, which calls
`handle_exception(..., suspend=False)`. The engine was therefore *never*
suspended, and every subsequent search request went to Baidu again — which
keeps the trigger alive.

Verified with a unit test using a mocked response: before the change the HTML
body raised `JSONDecodeError` and no suspension; after the change it raises
`SearxEngineCaptchaException` with `suspended_time=300`.

## Background measurements

From outside SearXNG, with `curl_cffi` (`impersonate="chrome"`), on several
egress IPs:

- `https://www.baidu.com/s?wd=<q>` + `tn=json` (what the engine sends) returned
  HTTP 302 to `wappass.baidu.com/static/captcha` on every attempt, including the
  first request from a previously unused IP.
- The same URL with only `wd` returned HTTP 200 with a complete result page.
- After a triggered CAPTCHA, a plain `wd` request worked again within 2-17 s
  (coarse 60 s polling bounded it at 62 s). That is the basis for choosing 300 s
  rather than the 86400 s default.

## Testing

- New unit tests: `tests/unit/engines/test_baidu.py` (4 tests)
- Full unit suite: `Ran 343 tests ... OK`
- `pylint searx/engines/baidu.py tests/unit/engines/test_baidu.py`: 10.00/10

## AI disclosure

Per the AI Policy: the code and this description were prepared with AI
assistance (Hermes Agent). The measurements above were run and reviewed by the
human contributor, who is able to explain and maintain the changes.

- [x] My contribution complies with the [AI Policy].


vojkovic:
> Per the AI Policy: the code and this description were prepared with AI
assistance (Hermes Agent)

You did not read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst)

PR #6712, opened 2026-09-13
[mod] github: convert issue templates to yaml forms

vojkovic:
### What does this PR do?

Converts the issue templates to yaml so people fill in the fields, don't need to delete the comments and also stop people accidentally missing the ai policy checkbox and by having it `required: true` it will lead to much less false positives.

Install method and the engine category are dropdowns too now which is super nice. I removed expected behaviour / additional context / engine features since no one was filling those in anyways.

I made my default branch on this branch so you can go here to test it out - https://github.com/vojkovic/searxng/issues

### Related issues

- https://github.com/searxng/searxng/pull/5973
- https://github.com/searxng/searxng/pull/5848
- https://github.com/searxng/searxng/pull/6476
 
### Code of Conduct

Cursor helped me with the migration


.github/scripts/ai_policy.cjs
```diff
@@ -8,7 +8,7 @@ module.exports = async ({ github, context }) => {
 
   // https://github.com/searxng/searxng/pull/6476#discussion_r3683782481
   const hasBox = /\[[Xx]\].*AI Policy/.test(body);
-  const hasRef = /\[AI Policy\]:\s*https:\/\/github\.com\/searxng\/searxng\/.*AI_POLICY/.test(body);
+  const hasRef = /\[AI Policy\](?::\s*|\()https:\/\/github\.com\/searxng\/searxng\/.*AI_POLICY/.test(body);
```
inetol, line 11: The ai_policy script (also the workflow) makes no sense now as users are forced to check the checkbox before posting an issue. (we can keep the policy as a note)

inetol (changes requested):
Note, you can also "Preview" each file to ""preview"" the template.

\  \  \
 V V V

return42 (approved):
*Well done, young Padawan. The dark side of the Force won't stand a chance anymore.*

.github/scripts/ai_policy.cjs
vojkovic: It should actually make more sense now because if the checkbox isn't ticked, it means the issue wasn't created using the issue template.

e.g. agents using the `gh issue` command to programmatically open up a new ticket.

Also PRs also are still checked with the script.

Bnyro (approved):
Good job 👍

.github/scripts/ai_policy.cjs
inetol: > e.g. agents using the `gh issue` command to programmatically open up a new ticket.

Damn, I see..

inetol (approved):

PR #6713, opened 2026-09-13
[feat] results: limit titles and content to fixed length

Bnyro:
### What does this PR do?
- limits titles to 200 chars and the result content to 1200 chars
- see the discussion in #6704 
- in short: very long titles can make SearXNG become very slow / timeout when serving the results page and are a bad user experience because you have to scroll a lot to get to the next result, and it looks bad.

-  !neocities Firefox is already running, but is not responding

### Related issues
- #6704

### Code of Conduct

No AI used.


searx/result_types/_base.py
```diff
@@ -92,6 +94,27 @@ def _normalize_url_fields(result: "Result | LegacyResult"):
...
+    # don't split at characters, but at full words instead
+    # e.g. _truncate_text("hello world", 9) yields "hello" instead of "hello wor"
+    text = text.rsplit(" ", 1)[0]
+    return f"{text} …"
+
+
+def _normalize_text_field(text: str, field_name: str, max_length: int) -> str:
+    if text and not isinstance(text, str):
+        text = str(text)
+        log.debug(f"result: invalid type of field '{field_name}': %s", str(text))
```
return42: 🤔 `f"%s", str(str(text))`  versus `f"{text}"`

searx/result_types/_base.py
```diff
@@ -92,6 +94,27 @@ def _normalize_url_fields(result: "Result | LegacyResult"):
             ).geturl()
 
 
+def _truncate_text(text: str, max_length: int) -> str:
+    text = text[:max_length]
+    # don't split at characters, but at full words instead
+    # e.g. _truncate_text("hello world", 9) yields "hello" instead of "hello wor"
+    text = text.rsplit(" ", 1)[0]
+    return f"{text} …"
```
return42: untested:

```python
def truncate(text, max_len):
    if len(text) > max_len:
        text = text[:max_len].rsplit(" ", 1)[0] + "…"
    return text
```

searx/result_types/_base.py
```diff
@@ -92,6 +94,27 @@ def _normalize_url_fields(result: "Result | LegacyResult"):
...
+    if text and not isinstance(text, str):
+        text = str(text)
+        log.debug(f"result: invalid type of field '{field_name}': %s", str(text))
+
+    text = WHITESPACE_REGEX.sub(" ", text).strip()
+
+    # truncate too long result titles and contents
+    if text and len(text) > max_length:
+        text = _truncate_text(text, max_length)
+    return text
```
return42, line 109:
```python
return _truncate_text(text, max_length)
```

TBH: a function _truncate_text is not needed for a two liner:

```python
if len(text) > max_len:
    text = text[:max_len].rsplit(" ", 1)[0] + "…"
return text
```

Bnyro: Good catch! I initially didn't have the _normalize_text_field method, so I had to call _truncate_text twice - now after the refactoring, the method is useless.

return42 (approved):

PR #6714, opened 2026-09-13
[fix] public domain image archive: fails to extract API url

Bnyro:
### What does this PR do?
- relax the regex for finding the proxy url as they now put it into backticks instead of quotation marks

### Code of Conduct

No AI used.


return42 (approved):
LGTM 👍

PR #6716, opened 2026-09-13
[fix] 500px: migrate to new search endpoint

Bnyro:
### What does this PR do?
- The old API endpoint stopped working, so we now use the new one the website also uses.

### Code of Conduct

No AI used.


searx/engines/500px.py
```diff
@@ -46,47 +44,77 @@
...
-      cursor
+      __typename
+    }
+    pageInfo {
+      hasNextPage
+      endCursor
+      __typename
     }
+    totalCount
+    __typename
```
return42: AFAIK pageInfo and totalCount are not needed (C&P leftover?)

searx/engines/500px.py
```diff
@@ -46,47 +44,77 @@
...
+            size_1024
+            size_2048
+            size_4k
+            __typename
+          }
+          uploader {
+            displayName
+            __typename
+          }
+          isNsfw
```
return42, line 95: Can we use this value for "safe search" capability?

searx/engines/500px.py
```diff
@@ -95,42 +123,37 @@ def request(query: str, params: "OnlineParams") -> None:
...
-                    "url": base_url + node["canonicalPath"],
-                    "thumbnail_src": thumbnail_src,
-                    "img_src": img_src,
-                    "title": node["name"],
-                    "content": node["description"],
-                    "author": node["photographer"]["displayName"],
-                    "resolution": f"{node['width']}x{node['height']}",
-                }
+            res.types.Image(
+                template="images.html",
```
return42: When using result classes, there is no need to set the template (should not set).

searx/engines/500px.py
Bnyro: Good point, looking into it :+1:

searx/engines/500px.py
Bnyro: True, copy and paste leftover.

searx/engines/500px.py
Bnyro: I still seem to get nsfw results (i.e. results with isNsfw set to true) even if I set the excludeNsfw filter. Either I'm doing something wrong or the filter doesn't work.

return42: Seems 500px does not have such a filter:

grafik

return42 (approved):

PR #6722, opened 2026-09-14
[fix] engines: flaticon encode non ascii words

vojkovic:
### What does this PR do?

Small fix to queries like café / München by using quote() in the Referer

### Code of Conduct

Cursor found this while reviewing engines


Bnyro (approved):
Thanks, good catch!

Bnyro:
Are you still using the engine successfully? I seem to always get blocked with HTTP403 after a few requests if I don't use the cookies it previously set with other API responses.

So looks like we have to cache the `bm_s` cookie in an `EngineCache`. I tried
```diff
diff --git a/searx/engines/flaticon.py b/searx/engines/flaticon.py
index 5816774db..e721d9a84 100644
--- a/searx/engines/flaticon.py
+++ b/searx/engines/flaticon.py
@@ -8,6 +8,7 @@ from urllib.parse import quote, urlencode
 
 import typing as t
 
+from searx.enginelib import EngineCache
 from searx.result_types import EngineResults
 
 if t.TYPE_CHECKING:
@@ -29,17 +30,28 @@ base_url = "https://www.flaticon.com"
 categories = ["images", "icons"]
 paging = True
 
+CACHE: EngineCache
+"""Cache for storing the ``bm_s`` session cookie to bypass ratelimits."""
+
+
+def setup(engine_settings: dict[str, t.Any]):
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+
 
 def request(query: str, params: "OnlineParams") -> None:
     args = {
         "word": query,
     }
-    params["headers"].update(
-        {
-            "Referer": f"{base_url}/search?word={quote(query)}",
-            "X-Requested-With": "XMLHttpRequest",
-        }
-    )
+
+    params["headers"].update({
+        "Referer": f"{base_url}/search?word={quote(query)}",
+        "X-Requested-With": "XMLHttpRequest",
+    })
+
+    if bm_s := CACHE.get("bm_s"):
+        params["cookies"]["bm_s"] = bm_s
+
     params["url"] = f"{base_url}/ajax/search/{params['pageno']}?{urlencode(args)}"
 
 
@@ -50,6 +62,10 @@ def _fix_url(url: str) -> str:
 def response(resp: "SXNG_Response"):
     res = EngineResults()
 
+    # cache `bm_s` cookie to bypass ratelimiting
+    if bm_s := resp.cookies.get("bm_s"):
+        CACHE.set("bm_s", bm_s, expire=30 * 24 * 3600)  # 1 month
+
     result: dict[str, str]  # TBH: dict[str, t.Any]
     for result in resp.json()["items"]:
         tags = [
```
But it looks like the `bm_s` cookies I get as response are not valid, so idk...

vojkovic:
Yeah I also always get 403s now. I will look into it too - it must be a recent change because it was working when I was testing it a few days ago obviously 😆

Bnyro:
Magnific is run by the same company. I opened a PR to solve their CAPTCHA here: https://github.com/searxng/searxng/pull/6757.

Though unfortunately it looks like Flaticon doesn't use the same challenge (e.g. cookies are named differently and I couldn't see the challenge page).

vojkovic:
> Though unfortunately it looks like Flaticon doesn't use the same challenge

It is using Akamai CPR -> https://www.akamai.com/products/content-protector

I don't think there is a way to get around this unfortunately 😕 We may have to remove this engine or make it work via their api, they seem to offer keys for personal use -> https://api.flaticon.com/v3/docs/index.html

PR #6723, opened 2026-09-14
[fix] engines: yahoo modernisation and ybv cookie

### LINKED ISSUE #6721: Yahoo Search returns no results
### Description

Every Yahoo search returns no results; the results page shows yahoo: parsing error.  It doesn't follow Yahoo's redirects.  Identified and tested with Claude Code.

### How to reproduce

Search with !yahoo

### Screenshots & Logs

WARNING:searx.engines.yahoo: ErrorContext('searx/engines/yahoo.py', 219, 'dom = html.fromstring(resp.text)', 'lxml.etree.ParserError', None, ()) False

### Version of SearXNG

2026.9.13+d4ce87c23

Docker latest

### Code of Conduct

vojkovic:
### What does this PR do?

Fixes the yahoo engine since yahoo changed their bot detection. 

Screenshot

There is now a tracking pixel that mints a `YBV` cookie.

Also cleaned up and modernised the engine up a bit (i.e. EngineResults, `searxng_locale`, more
region subdomains). It now works in Australia whereas it didn't work at all before.

Screenshot

### Related issues

- Closes: #6721

### Code of Conduct

Cursor helped with the implementation


searx/engines/yahoo.py
```diff
@@ -111,94 +110,62 @@
...
-
-def parse_url(url_string):
+CACHE: EngineCache
+"""YBV cookie"""
+
+
+def setup(engine_settings: dict[str, t.Any]) -> bool:
+    global CACHE  # pylint: disable=global-statement
+    CACHE = EngineCache(engine_settings["name"])
+    return True
```
Bnyro:
```suggestion
def setup(engine_settings: dict[str, t.Any]):
    global CACHE  # pylint: disable=global-statement
    CACHE = EngineCache(engine_settings["name"])
```

The return value here is only useful if it'd be `False` to disable the engine. Otherwise, not returning anything has the same effect.

searx/engines/yahoo.py
```diff
@@ -212,45 +179,59 @@ def parse_url(url_string):
...
-        title = eval_xpath_getindex(result, title_xpath, 0, default='')
-        title: str = extract_text(title)
-        content = eval_xpath_getindex(result, './/div[contains(@class, "compText")]', 0, default='')
-        content: str = extract_text(content, allow_none=True)
-
-        # append result
-        results.append(
-            {
-                'url': url,
+        title = extract_text(eval_xpath_getindex(result, ".//h3//a/@aria-label", 0, default=None), allow_none=True)
```
Bnyro, line 225:
```suggestion
        title = extract_text(eval_xpath_getindex(result, ".//a/h3", 0, default=None), allow_none=True)
```
For me, there's no a in an h3 - what works for me is the suggestion above, so the result layouts may differ depending on the region. I attached the HTML response I got below. Seems to be reproducible with en-US as Searxng locale.

[yahoo_response.html](https://github.com/user-attachments/files/32203919/foo.html)

searx/engines/yahoo.py
```diff
@@ -111,94 +110,62 @@
...
+def request(query: str, params: "OnlineParams") -> None:
+    parts = params["searxng_locale"].split("-")
+    lang = yahoo_languages.get(parts[0], "any")
+
+    url_params: dict[str, str | int] = {"p": query}
+    if params["time_range"] in time_range_dict:
+        url_params["btf"] = time_range_dict[params["time_range"]]
+    if params["pageno"] == 1:
+        url_params["iscqry"] = ""
+    elif params["pageno"] >= 2:
```
Bnyro:
```suggestion
    else:
```

searx/engines/yahoo.py
```diff
@@ -212,45 +179,59 @@ def parse_url(url_string):
...
-def response(resp):
-    """parse response"""
-
-    results = []
-    dom = html.fromstring(resp.text)
+def _yahoo_html(resp: "SXNG_Response") -> "SXNG_Response":
+    cookies = dict(resp.search_params["cookies"])
+    params = resp.search_params
+    for _ in range(4):
+        header = resp.headers.get("set-cookie")
```
Bnyro: I'm pretty sure that `resp.cookies["YBV"]` works too and is cleaner?

searx/engines/yahoo.py
```diff
@@ -212,45 +179,59 @@ def parse_url(url_string):
...
+            return resp
+        # request ourselves instead of following it
+        resp = get(
+            urljoin(resp.url, loc),
+            cookies=cookies,
+            headers=params["headers"],
+            allow_redirects=False,
+        )
+        resp.search_params = params
+    return resp
```
Bnyro, line 210:
```suggestion
def _yahoo_html(resp: "SXNG_Response") -> "SXNG_Response":
    params = resp.search_params

    for _ in range(4):
        if ybv := resp.cookies.get("YBV"):
            if ybv.startswith("v0.2"):
                CACHE.set("YBV", ybv, expire=86400)

        if resp.status_code == 200:
            break

        loc: str | None = resp.headers.get("location")
        if resp.status_code not in (302, 307) or not loc:
            break

        # request ourselves instead of following it
        resp = get(
            urljoin(resp.url, loc),
            cookies=dict(resp.cookies),
            headers=params["headers"],
            allow_redirects=False,
        )

    resp.search_params = params
    return resp
```

searx/engines/yahoo.py
```diff
@@ -212,45 +179,59 @@ def parse_url(url_string):
...
 
-def response(resp):
-    """parse response"""
-
-    results = []
-    dom = html.fromstring(resp.text)
+def _yahoo_html(resp: "SXNG_Response") -> "SXNG_Response":
+    cookies = dict(resp.search_params["cookies"])
+    params = resp.search_params
+    for _ in range(4):
```
Bnyro: Where is the 4 coming from? Mind moving it to a constant with a doc comment? :)

Bnyro:
Very nice! Left some comments above.

searx/engines/yahoo.py
vojkovic: I had a feeling this was happening: I can't visit search.yahoo.com in aus it keeps redirecting me to the au.search.yahoo.com 😛

searx/engines/yahoo.py
vojkovic: It is because we need to make a few requests. I will add the docstirng

Tracking pixel (solving this gives YBV by only valid for 60s.

->

.gif challenge (this gives the 24h YBV)

->

(optional) geo redirect (i.e. au.search.yahoo.com)

->

200 with results

vojkovic:
Thanks for your comments. I have addressed them in https://github.com/searxng/searxng/commit/36b8ad0f276a274f2a294c481266338632722609 ([diff](https://github.com/searxng/searxng/compare/8ff10a3f43ce5a2fa9f12148d8d8cb54a42b4408..36b8ad0f276a274f2a294c481266338632722609))

I only didn't use `cookies=dict(resp.cookies)` because that would remove the `sB` cookie

Bnyro (approved):

PR #6724, opened 2026-09-14
[mod] simple: apply nojs via stylesheet

vojkovic:
### What does this PR do?

The search bar currently briefly flashes with the no-js text 'search' when loading the page before javascript loads to correct it. This PR makes the icon the default instead. Also, the reason for having the text in nojs (even though nojs can render icons) is that the Tor browser on the 'Safest' mode sets `svg.disabled` true in ff which would hide the search icon.

Instead we can have another stylesheet specifically for nojs within a <noscript> which will allow us to have much better control of the nojs experience in the future as well. For example, this would allow us to eventually remove all other icons from the nojs experience too.

With JS: You should not notice any shifts in the search bar and also the clear button only should show when there is text in the search bar.

Without JS: You should just see the 'search' text. Clear is hidden because it won't work with nojs.

### Code of Conduct

Cursor assisted me for some parts only


inetol (approved):
LGTM

Bnyro (approved):

PR #6728, opened 2026-09-15
[fix] yandex: image result xpath + modernization

### LINKED ISSUE #6710: Yandex Image Search - Intermittent Parsing Failures
When searching with Yandex Images, depending on the query, the results encounter a JSON decoding error.

The issue seems to stem from one of the inner image results having `false}}}`, so the rest of the JSON string gets truncated short.

If you use xpath to pull the json data instead, it returns data successfully (at least on these examples).
```python
html_data = html.fromstring(resp.content)
data_element = html_data.xpath('//div[contains(@class, "page-layout__content-wrapper")]//div[@class="Root"]')
json_data = data_element[0].get('data-state')
json_resp = loads(json_data)
```

### How To Reproduce?

Preferences -> Engines -> Images: Disable all, then enable yandex images

Two examples that currently produce the error are search strings "king kong wallpaper" and "bugs bunny movie"

### Expected behavior

about 30 image results to show up per page

### Screenshots & Logs

Docker Logs
Baresearch
Baresearch2

### Version of SearXNG

ghcr.io/searxng/searxng:latest / 2026.9.12-d4f00d15d

Docker, pulling from ghcr.io/searxng/searxng:latest

### Additional context

I've only tested a couple dozen searches and found two, and confirmed it's the same issue on my local instances and a couple public instances. I don't know if / when the results will change for the provided example searches.

### Code of Conduct

WookieeOnCrack:
### What does this PR do?

- Resolves #6710 with more robust xpath and fallback method
- Updates to `MainResult` / `Image` types
- Added type hinting
- Added "content" image result field

Can test before / after when running search on "king kong poster" or "stairway to china"

### Related issues

Closes: #6710

### Code of Conduct

- Grok - code review


Bnyro:
Please run `make format.python` to fix the CI.

searx/engines/yandex.py
```diff
@@ -48,12 +55,12 @@
...
 
-def catch_bad_response(resp):
-    if resp.headers.get('x-yandex-captcha') == 'captcha':
+def catch_bad_response(resp: "SXNG_Response") -> None:
+    if resp.headers.get("x-yandex-captcha") == "captcha":
         raise SearxEngineCaptchaException()
 
 
-def request(query, params):
+def request(query: str, params: "OnlineParams") -> "OnlineParams":
```
Bnyro, line 63:
```suggestion
def request(query: str, params: "OnlineParams"):
```

searx/engines/yandex.py
```diff
@@ -71,82 +78,95 @@ def request(query, params):
...
-    if search_type == 'web':
-        params['url'] = f"{base_url_web}?{urlencode(query_params_web)}"
-    elif search_type == 'images':
-        params['url'] = f"{base_url_images}?{urlencode(query_params_images)}"
+    if search_type == "web":
+        params["url"] = f"{base_url_web}?{urlencode(query_params_web)}"
+    elif search_type == "images":
+        params["url"] = f"{base_url_images}?{urlencode(query_params_images)}"
 
     return params
```
Bnyro, line 92:
```suggestion
```

searx/engines/yandex.py
```diff
@@ -71,82 +78,93 @@ def request(query, params):
...
-                    'resolution': f'{image_source["w"]} x {image_source["h"]}',
-                }
-            )
-
-        return results
-
-    return []
+    if params["pageno"] > 1:
+        query_params_web |= {"p": params["pageno"] - 1}
+        query_params_images |= {"p": params["pageno"] - 1}
```
Bnyro:
```suggestion
        query_params_web["p"] = params["pageno"] - 1
        query_params_images["p"] = params["pageno"] - 1
```

searx/engines/yandex.py
```diff
@@ -71,82 +78,93 @@ def request(query, params):
...
+
+    params["cookies"] = {"cookie": "yp=1716337604.sp.family%3A0#1685406411.szm.1:1920x1080:1920x999"}
+
+    if search_type == "web":
+        params["url"] = f"{base_url_web}?{urlencode(query_params_web)}"
+    elif search_type == "images":
+        params["url"] = f"{base_url_images}?{urlencode(query_params_images)}"
+
+
+def _parse_json_results(dom: html.HtmlElement) -> dict:
```
Bnyro:
```suggestion
def _parse_json_results(dom: html.HtmlElement) -> dict[str, t.Any]:
```

searx/engines/yandex.py
WookieeOnCrack: Thank you. I didn't know it shouldn't return anything. I was just trying to add a type hint for what it was already doing.

searx/engines/yandex.py
```diff
@@ -71,82 +78,93 @@ def request(query, params):
...
+            try:
+                json_resp = loads(json_data)
+                break
+            except Exception:
+                logger.debug("failed parsing data-state json")
+                continue
+    if not json_resp:
+        # fallback to extr(..., 'advRsyaSearchColumn":null}}')
+        json_resp = _parse_json_results_fallback(dom)
+    return json_resp
```
Bnyro:
```suggestion
    # attempt to parse using xpath - finding element with "data-state" attribute
    data_elements = dom.xpath("//*[@data-state]/@data-state")
    for json_data in data_elements:
        try:
            json_resp = loads(json_data)
            if json_resp.get("location") == "/images/search/":
                return json_resp
        except Exception:
            logger.debug("failed parsing data-state json")

    # fallback to extr(..., 'advRsyaSearchColumn":null}}')
    return _parse_json_results_fallback(dom)
```
The code can be simplified using early returns.

searx/engines/yandex.py
```diff
@@ -71,82 +78,93 @@ def request(query, params):
...
+    )
+    json_data = '{"location":"/images/search/' + content_between_tags + 'advRsyaSearchColumn":null}}'
+
+    if content_between_tags == "fail":
+        logger.debug('Attempting fallback to "serpFooter" key')
+        # try falling back to a another unique json key and trailing colon
+        content_between_tags = extr(html_sample, '{"location":"/images/search/', '"serpFooter":')
+        json_data = '{"location":"/images/search/' + content_between_tags.rstrip(",") + "}}"
+
+    return loads(json_data)
```
Bnyro: I don't think this fallback is ever going to be used, at least it never got called when I tested the engine.

I'm in favor of removing it completely.

Bnyro (commented):
Thanks a lot for the pull request. Left some more comments after a bit of testing, lgtm after that 👍

searx/engines/yandex.py
WookieeOnCrack: The fallback should really only happen if the `data-state` attribute name changes. In normal operation it should never run (which matches what you saw).

I’m fine removing it if we prefer the simpler “fail loudly when the primary path breaks” approach...that way any future markup change shows up immediately in the engine stats instead of being papered over.

Happy to drop it and just let `loads()` raise on failure.

Bnyro: Yeah, if the yandex website changes that way, there are probably also other things that need to be changed in our engine, so we would need to make new bug fixes anyways.

So let's drop it.

Bnyro (approved):
Fixed some other small issues in https://github.com/searxng/searxng/compare/e63254236efc328e523ee7322bdb9e0a0bd331fd..bff3146ae92124feb6c76192bad1327d1910bf74.

Thanks a lot for the PR, merging!

WookieeOnCrack:
Thanks for baring with me. I'm not accustomed to the workflows.

PR #6730, opened 2026-09-15
[fix] image_proxy compressed images

### LINKED ISSUE #6719: Image isn't loaded when image proxy is active
### How To Reproduce?

For example there are issues when requesting images from `!devicons icon`

Without image proxy:

Image

With image proxy enabled:

Image

In this example one URL for the image is:

- https://cdn.jsdelivr.net/gh/devicons/devicon@latest/icons/devicon/devicon-plain-wordmark.svg

With image-proxy enabled the URL is shown here in the curl command:

```
curl 'http://127.0.0.1:8888/image_proxy?url=https%3A%2F%2Fcdn.jsdelivr.net%2Fgh%2Fdevicons%2Fdevicon%40latest%2Ficons%2Fdevicon%2Fdevicon-plain-wordmark.svg&h=fbc2c44b7ca16e0617a28feccf1294f5e03d5b11e99cd5614b00eca84b3d0468'
```

When we look at the stream from the curl command, we see an SVG string that suddenly breaks off in the middle.

Provided my analysis was not erroneous, the stream is not terminated by the image proxy; rather, it seems to me that cdn.jsdelivr.net interrupts the stream for reasons unknown to me.

I wondered if it might be related to establishing the connection, so I tested this little PoC .. but without any effect:

```diff
diff --git a/searx/webapp.py b/searx/webapp.py
index 8be325230..61033dc54 100755
--- a/searx/webapp.py
+++ b/searx/webapp.py
@@ -1003,13 +1003,25 @@ def image_proxy():
     resp = None
     try:
         request_headers = {
-            'User-Agent': gen_useragent(),
-            'Accept': 'image/webp,*/*',
+            'Accept': 'image/avif,image/webp,image/png,image/svg+xml,image/*;q=0.8,*/*;q=0.5',
             'Sec-GPC': '1',
             'DNT': '1',
+            "Sec-Fetch-Dest": "image",
+            "Sec-Fetch-Mode": "no-cors",
+            "Sec-Fetch-Site": "cross-site",
+            "Accept-Language": "de,en;q=0.7,en-US;q=0.3",
+            "Accept-Encoding": "gzip, deflate, br, zstd",
+            "Connection": "keep-alive",
+            "Alt-Used": "cdn.jsdelivr.net",
         }
         set_context_network_name('image_proxy')
-        resp, stream = http_stream(method='GET', url=url, headers=request_headers, allow_redirects=True)
+        resp, stream = http_stream(
+            method='GET',
+            url=url,
+            headers=request_headers,
+            allow_redirects=True,
+            impersonate="firefox",
+        )
         content_length = resp.headers.get('Content-Length')
         if content_length and content_length.isdigit() and int(content_length) > maximum_size:
             return 'Max size', 400
```

### Code of Conduct

vojkovic:
### What does this PR do?

This pr fixes the image proxy decompressing brotli/gzip in transit. This was mostly just an issue with svgs since they can be compressed by brotli/gzip but like pngs cannot.

Docs: https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

Setting accept_encoding to none stops curl from unzipping the gzip / br and instead just proxies like how httpx was doing before.

Enable image proxy and search devicons, confirm it works

### Related issues

- Closes: https://github.com/searxng/searxng/issues/6719

### Code of Conduct

cursor helped track when the regression happened


searx/network/__init__.py
```diff
@@ -201,6 +201,7 @@ def delete(url: str, **kwargs: t.Any) -> SXNG_Response:
 
 async def stream_chunk_to_queue(network, queue, method: str, url: str, **kwargs: t.Any):
     try:
+        kwargs.setdefault('accept_encoding', None)
```
Bnyro, line 205: Please add a short comment here why we need it so that nobody removes it by accident :)

Bnyro (approved):
Thanks for investigating and fixing! This bug was already getting very annoying for me ...

return42 (approved):
Thanks a lot 👍 

-----

one nit I have: Descriptions like the ones we often see in PRs:

```
This pr fixes the image proxy decompressing brotli/gzip in transit.
This was mostly just an issue with svgs since they can be compressed
by brotli/gzip but like pngs cannot.

Docs: https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

Setting accept_encoding to none stops curl from unzipping the gzip / br 
and instead just proxies like how httpx was doing before.
```

should always be included in the commit message. This makes it easier to work with the Git history. Try the following, for example:

    git log -S brotli

With a simple command like this, you can quickly find all relevant commits. Without the description from the PR, your commit wouldn’t be displayed, since the keyword “brotli” isn’t part of the commit and isn’t part of the commit message.  That’s a shame, because the statement “This PR fixes the image proxy decompressing brotli/gzip in transit” is particularly important in the context of this patch.

In case you didn’t know: If the text is already included when writing the commit message, GitHub will also use that commit message when creating the PR .. at least if the PR consists of only one commit :-)

vojkovic:
Thanks for bringing it up :-) @return42 

How's that now?

Also do you prefer rebase and merge or squash and merge when using github?

return42:
> How's that now?

Perfect / thanks a lot!

> Also do you prefer rebase and merge or squash and merge when using github?

It depends on the specific case and how much work we want to put into it; so far, we've come up with the following compromise.

The advantage of GH-squash is that the PR number is included at the top of the commit message title. If you’re looking up a commit, you just need to click on that number to go directly to the PR.

However, we also have PRs where we want to keep the commits separate .. for example, if we have a “build” commit, we don’t want it to be mixed in with the previous PR containing the actual change (*clear diff*).

In such cases - that is, when keeping the commits separate is relevant to the content (for whatever reason) - we use GH-rebase.  However, this means the PR number isn’t included in the commit; that’s the trade-off. 

Sometimes, when I feel it’s very important to include the PR number in the GH-rebase merge, I edit the commit messages again before the merge .. but since I’m a lazy person, I only do that very rarely :-)

Bnyro:
I think the most important thing is that we only have squash and rebase, and not these additional "Merge pull request ..." commits that bloat the Git history :)

return42:
Merge commits really shine in a true versioned release cycles when there are multiple release branches (*semantic versioning with major, minor, and patch levels*). For example, this allows security patches to be deployed quickly and easily between major releases.   But even here, SearXNG is quite unique; we use rolling releases, and the master branch is the only *truth*.

In a release context like this, merge commits offer only a minor advantage; a linear, easy-to-read history is more important to us, which is why we avoid merge commits entirely

PR #6731, opened 2026-09-15
[feat] yandex: simplify request method, add time range support

Bnyro:
### What does this PR do?
We previously set the 'cookie' cookie value with a timestamp from 2024 to `yq=...`. The cookie is not actually used in Yandex, instead we probably wanted to set the `yq` cookie instead. But since the engine works without it, it can be dropped.

Also simplifies the request method a bit to be easier to read.

### Related issues
- will rebase after #6728

### Code of Conduct

- [ X] **I hereby confirm that this PR conforms with the [AI Policy].**


return42 (approved):
Time range support works 👍

PR #6735, opened 2026-09-16
[fix] startpage: send the locationPref cookie its search form sets

MayCXC:
### What

Startpage's search form sets a `locationPref` cookie alongside `preferences`. With no user location selected it carries the "none" preference, base64 of the URL-encoded JSON:

```json
{"type":"none","locPref":{"type":"none","device_location":null,"geo_location":null}}
```

The engine's goal is a request identical to the one the search form builds, so it now sends this cookie too.

Addresses one item from the Startpage to-do list in #4950 (Cookies: "Add missing locationPref cookie").

### Verification

- The constructed value byte-matches a `locationPref` cookie captured from a live browser session after saving Startpage settings.
- Reproduced on current master: Startpage returns full results both with and without the cookie, directly and through a proxy, so this is a fidelity change with no regression.

### AI usage disclosure

Multiple AI models were used for planning and code. There is no separation between the human and AI contribution in either the code or the descriptions.

- [x] I have read the [AI Policy](https://github.com/searxng/searxng/blob/master/AI_POLICY.rst) and disclosed my AI usage above.


vojkovic:
> Startpage's search form sets a locationPref

I can't reproduce the `locationPref` cookie, can you show me where you are seeing that being set? @MayCXC

return42:
@vojkovic He belongs to that group of contributors who thoughtlessly flood open-source projects with a deluge of AI-generated contributions: AI contributions based on the “spray and pray” approach … in the hope that at least one of them will hit the mark.

PR #6738, opened 2026-09-16
[fix] cache: authenticate serialized values before unpickling

DeryFerd:
## Problem / Motivation

`ExpireCache` stores its key/value pairs in a SQLite file, and `ExpireCacheSQLite.deserialize()` handed the stored BLOB straight to `pickle.loads()` with no check on where it came from. The database path is assembled in `ExpireCacheCfg.__post_init__` as `tempfile.gettempdir() + /sxng_cache_<name>.db`, which means the filename is predictable from the cache name alone. On a typical host that file lands in a world writable temp directory and is created with mode `0666`, so any local process or user that can write it can drop a pickle of their choosing into any row.

I reproduced the whole chain before writing any code: build a cache on its default path, overwrite one row's value with a `__reduce__` payload, then call `get()`. The payload executed as the SearXNG process. The cache file has no integrity check at all today, so there is nothing stopping the read path from unpickling attacker-supplied bytes.

## Why it matters

SearXNG is commonly deployed on shared or multi-user hosts, and `/tmp` is world writable by design. On those systems the cache file is not private, and its name is not a secret either. That turns an ordinary local file write into code execution in the search process, which is a much shorter path than the rest of the threat model assumes.

This is not a new concern in the codebase. When the `ExpireCache` abstraction was originally added in PR #4650, dalf pointed out that using pickle here is only acceptable when the database is authenticated, because otherwise it "can lead to execution of unknown code (if the database is not trusted)". The Fernet encryption layer that would have supplied that authentication was dropped later in the same review, so the pickle remained without the property that justified it. This PR restores the missing check.

## What changed (motivation → approach → change)

The fix is to make the cache verify that a value was written by an instance that knows the configured `ExpireCacheCfg.password` before it is unpickled. That password already exists and is already used by `secret_hash()` for the keys, so the values can use the same secret.

`serialize()` now prepends a 32-byte HMAC-SHA256 signature of the pickled bytes, keyed by `ExpireCacheCfg.password`. `deserialize()` splits the signature off, recomputes it, and compares with `hmac.compare_digest()` before calling `pickle.loads()`. A value that does not match raises `ValueError` and is never unpickled.

On the read side, `get()`, `pairs()`, and `state()` route through a small helper that catches that `ValueError`, logs a warning, and treats the row as a cache miss or skips it. A bad value degrades to "not cached" instead of taking the instance down or executing anything.

Two details are worth calling out for review:

- Old values are invalidated once, automatically. The existing `hash_token` already truncates all cache tables whenever the password changes. I folded the value format into that same token, so on the first start after this patch the tables are truncated once and repopulated. Everything in the cache is rebuildable, so this is a cold-cache moment, not data loss.
- The serialized values keep using pickle, on purpose. The value format is unchanged, so tuples and other rich types survive the round trip. Switching to a plain JSON or msgspec encoding would have silently converted tuples to lists and broken consumers like the tracker-patterns loader, which relies on tuple return values.

One behaviour change to be aware of: the signature is part of the BLOB that `MAX_VALUE_LEN` measures, so the effective payload budget shrinks by 32 bytes against the existing 10240 ceiling. I checked the boundary directly. A value whose pickle is 10261 bytes is now rejected at 10293 with the signature included, where before it would have fit. If that matters, `MAX_VALUE_LEN` can be nudged up by `SIGNATURE_LEN`, but I left the limit alone to keep this patch minimal.

The diff is two files:

- `searx/cache.py` — signature on serialize/deserialize, fail-safe reads, format in the hash token
- `tests/unit/test_cache.py` — new regression test

## Tests

The regression test is `tests/unit/test_cache.py`. I wrote it before touching `searx/cache.py` and watched it fail against the unmodified code with `['tampered'] != []`, which is the payload actually executing. After the fix it passes. I then reverted only `searx/cache.py` once more and confirmed the test fails again, so the assertion is proven to catch the vulnerability rather than just happening to be green.

The test covers three things: a legitimate round trip that must preserve the value and its tuple type, a tampered value that must not be deserialized and must come back as a miss, and an expired tampered value that must not be deserialized either.

I also checked the check itself against malformed input separately, since a length-prefixed compare is easy to get wrong. Empty bytes, a truncated blob, a signature with no payload, a valid-length wrong signature, a payload with no signature, and a single flipped byte in the payload are all refused, while freshly written values still deserialize.

Commands and results:

```
python -m nose2 -s tests/unit test_cache
Ran 3 tests in 0.084s
OK

python -m nose2 -s tests/unit
Ran 342 tests
FAILED (errors=1)

python -m pylint --rcfile .pylintrc searx/cache.py tests/unit/test_cache.py
Your code has been rated at 10.00/10

basedpyright --level warning searx/cache.py tests/unit/test_cache.py
0 errors, 0 warnings, 0 notes

black --check (repo options from ./manage)
2 files would be left unchanged
```

The single error in the full run is `test_webapp.ViewsTestCase.test_search_html`, which fails on a `TemplateNotFound: result_templates/default.html`. I confirmed it is unrelated by running the same command on a clean checkout of master with my change absent: it reports `Ran 339 tests` and the identical `FAILED (errors=1)`. My branch adds three tests and leaves that error exactly as it was.

## Manual verification

If you want to see the bug and the fix by hand, this reproduces it without touching the app:

```python
import os, pickle, sqlite3, tempfile
os.environ['SEARXNG_SETTINGS_PATH'] = 'searx/settings.yml'
from searx.cache import ExpireCacheCfg, ExpireCacheSQLite

marker = os.path.join(tempfile.gettempdir(), 'pwned.txt')
db = os.path.join(tempfile.gettempdir(), 'repro_cache.db')
if os.path.exists(db):
    os.unlink(db)
cache = ExpireCacheSQLite.build_cache(ExpireCacheCfg(name='REPRO', db_url=db, password=b'pw'))
cache.set('k', ('legit',), expire=None, ctx='t')
print('legit round trip:', cache.get('k', ctx='t'))

class Evil:
    def __reduce__(self):
        return (os.system, (f'echo PWNED > "{marker}"',))

with sqlite3.connect(db) as conn:
    conn.execute("UPDATE t SET value=? WHERE key='k'", (pickle.dumps(Evil()),))

print('read after tamper:', cache.get('k', default='MISS', ctx='t'))
print('marker created:', os.path.exists(marker))
```

On master this prints the legitimate value, then `0` from the read, and creates the marker file. With the patch applied it prints `discard cache value: signature mismatch`, returns `MISS`, and never creates the marker.

## Related issues

There is no existing issue for the cache integrity problem. The concern itself was raised during review of PR #4650, which is where the authentication requirement and its later removal are recorded.

## Checklist

- [x] Regression test written first and confirmed to fail against the unpatched code
- [x] Existing unit tests still pass (same single pre-existing Windows template error as master)
- [x] `black`, `pylint`, and `basedpyright` clean on the changed files
- [x] No unrelated refactors; two files changed
- [x] No secrets, credentials, or internal links in the diff

### Code of Conduct

If I have used AI tools for working on the changes in this PR, I will attach a list of all AI tools I used and how I used them. I hereby confirm that I haven't used any other tools than the ones I mention below.

AI tools used: ZCode. Extent: I used it to help read the cache code path, discuss the signature approach, and draft the test and this description. I reproduced the vulnerability myself, reviewed every line of the diff, and ran all the checks listed above.


return42:
> SearXNG is commonly deployed on shared or multi-user hosts, and `/tmp` is world writable by design.


There are access permissions at the file level! If the /tmp directory is still considered problematic, then the solution lies in the deployment.

Unfortunately, you haven’t read our AI guidelines and are forcing us to wade through their AI-generated half-truths… As is often the case with posts of this kind, there’s a grain of truth in it; taken as a whole, however, it’s useless because the AI made too many false assumptions.

PR #6740, opened 2026-09-16
[fix] network: stop http redirects

vojkovic:
Report details on our security mailbox or Matrix thread (`$yFihQGXaRAXVXunTldMYQJzsLWr5rSbLqlx5L-eYXNQ`)

### What does this PR do?

Stop curl from redirecting from a HTTPS origin to a HTTP page with these two opts:

- https://curl.se/libcurl/c/CURLOPT_REDIR_PROTOCOLS_STR.html
- https://curl.se/libcurl/c/CURLOPT_PROTOCOLS_STR.html

This also removed the need for `check_url`, so I removed that too. This changes basically nothing as there are no engines that currently redirect from HTTPS -> HTTP so you shouldn't notice any changes.

Check that http sites won't load with enable_http to false and vice versa.

### Code of Conduct

Cursor assisted with finding curl docs


inetol:
I've added the `Reported-by` field for the person who reported this issue. [diff](https://github.com/searxng/searxng/compare/f45d6b2716799c1f062c41f344ce903d7cf253e3..49f270b853067b5e1313bf0501d807a7d03f4082)

return42 (approved):
LGTM 👍

PR #6688, opened 2026-09-10
[feat] engines: add findborg.com (general, news, images, videos)

Bnyro:
### What does this PR do?
- Many engines use the JSON results from Brave (e.g. Brave, BraveAPI, Tusksearch, ...), so the result parsing logic is always the same. This PR deduplicates this code.
- Second, this adds support for findborg.com which also uses the Brave results.

### How to test this PR locally?
- Test the findborg engine and test that brave, tusksearch and braveapi still work.
- !fi !br !tu test
- !fii !bri !tui test
- ...

### Code of Conduct

[AI Policy]: https://github.com/searxng/searxng/blob/master/AI_POLICY.rst

- [X] **I hereby confirm that this PR conforms with the [AI Policy].**

No AI used.

PR #6742, opened 2026-09-17
[mod] engines: brave - modernize response handling

WookieeOnCrack:
### What does this PR do?

Updates the Brave engine. Changes center on streamlined response handling:
 - all search types now use the returned JSON data for parsing results
 - search results updated to `MainResult` or `Image` ("suggestions" still utilizes `LegacyResult`)
 - additional fields added: resolution (images), views (videos)

I disabled other engines so only brave was active. 
 - Tested General, Images, Videos, News
 - A/B comparison against other online instances from searx.space
 - Brave seems to cap general results at 6 pages, so navigating to page 7 should trigger `"noResults" = True` in the JSON data.

### Related issues

N/A - Brave is currently functional. This started as a simple update to add the resolution field for images, but I got carried away with it.

### Code of Conduct

 - Grok (code review only)


Bnyro:
Looks great! I did created similar methods to parse Brave's JSON layout in https://github.com/searxng/searxng/pull/6688 (there are some engines that use results from Brave too and return the same JSON), so I can probably change my PR to use the methods you created here.

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
+def _parse_search_result(result: dict[str, t.Any]) -> MainResult:
+    thumbnail: dict[str, t.Any] = result.get("thumbnail", {})
+    return MainResult(
+        template="default.html",
+        title=result.get("title", ""),
+        content=html_to_text(result.get("description", "")),
+        url=result.get("url", ""),
+        publishedDate=_extract_published_date(result.get("page_age")),
+        pubdate=result.get("age", ""),
+        thumbnail=thumbnail.get("src", "") if thumbnail else "",
```
Bnyro: We previously had issues here because Brave also returns logos as images.

The `logo` field in the response tells us if it's a logo or a normal thumbnail, so we can just filter the logos out. See e.g. the workaround here: https://github.com/searxng/searxng/pull/6688/changes?diff=unified#diff-8f2861fee34ca55c610cbc1aea56028117901f76315d5a26f22360949528ba17R385

```suggestion
        thumbnail=thumbnail.get("src", "") if thumbnail and not thumbnail.get("logo") else "",
```

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
+    videos_resp: dict[str, t.Any] = body_resp.get("videos", {})
+    if videos_resp and "results" in videos_resp:
+        for result in videos_resp.get("results", []):
+            results.add(_parse_video_result(result))
+    # related queries -> suggestion (No MainResult?)
+    query: dict[str, t.Any] = body_resp.get("query", {})
+    if query and "related_queries" in query:
+        for x in query.get("related_queries", []):
+            suggestion = " ".join(val[1] for val in x)
+            results.add(results.types.LegacyResult({"suggestion": suggestion}))
```
Bnyro:
```suggestion
            results.add(results.types.LegacyResult(suggestion=suggestion))
```

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
+def _get_response_data(json_data: dict[str, t.Any], category: str | None = None) -> dict[str, t.Any]:
+    """Navigate the Brave embedded JSON to the category-specific response object."""
+    # Brave’s structure is mostly consistent but has a couple of quirks:
+    # - most categories live under data[1].data.body.response.<category>
+    # - news omits the intermediate "body" key
+    try:
+        data: dict[str, t.Any] = json_data["data"][1]["data"]
 
+        if data.get("noResults"):  # Boolean Value
+            logger.debug("Brave search returned no results.")
```
Bnyro:
```suggestion
```
That's not worth logging a warning, we don't want to bloat the logs too much with unnecessary details

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
-def _parse_videos(json_resp: dict[str, t.Any]) -> EngineResults:
-    res = EngineResults()
+    # map the parse function to a category key
+    category_map = {
+        _parse_search_result: "search",
+        _parse_news_result: "news",
+        _parse_image_result: "images",
+        _parse_video_result: "videos",
+    }
+    category = category_map.get(parse_func, "search")
```
Bnyro:
```suggestion
```
Why don't you just use `brave_category`?

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
-            template="images.html",
-            url=result["url"],
-            title=result["title"],
-            source=result["source"],
-            img_src=result["properties"]["url"],
-            thumbnail_src=result["thumbnail"]["src"],
-        )
-        res.add(item)
+        body_resp = data["body"]["response"]
+        if category in ("search", "goggles"):
```
Bnyro, line 363: The `goggles` case is currently unreachable, because it's not part of `category_map`.

Why don't you just use `brave_category` instead here?

searx/engines/brave.py
```diff
@@ -261,155 +256,153 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
+    if not json_resp:
+        # if _get_response_data returns {} - indicates it was parsed successfully but had "noResults" = True
+        return results
+    json_results: list[dict[str, t.Any]] = json_resp["results"]
+    for result in json_results:
+        results.add(parse_func(result))
 
-        res.add(item)
+    # general search / goggle might have secondary items
+    if parse_func is _parse_search_result:
```
Bnyro:
```suggestion
    if brave_category in ("search", "googles"):
```
Matching the type of a callback is not very clean imo, it's interesting that Python even allows you to do this.

searx/engines/brave.py
WookieeOnCrack: doh. 🤦 I should have just used "web" instead of "search" in the category map - it's not really mapping to the brave_category, but to what json path need to be fetched. The responses for "goggles" and "search" are handled same way.

I'll push an update later tonight, hopefully I can make it less confusing.

Using a parameter keeps it consistent when the web/search/goggles has to re-use the function to parse videos.

searx/engines/brave.py
WookieeOnCrack: I'll refactor it. The statement itself is quite efficient though...and perfectly valid. Python is checking to see if the underlying `parse_func` and `_parse_search_result` objects point at the same memory address, which means it's actually faster / less overhead than comparing two strings.

searx/engines/brave.py
WookieeOnCrack: I'm good with removing it, but it doesn't really fire very often. It was mostly added to clarify that the no-results was because brave was saying "there are no results", and not because of an error. The current brave engine is throwing an error when this occurs.

return42:
For the future: create a (feature) branch / never use your master branch for a PR!

WookieeOnCrack:
> For the future: create a (feature) branch / never use your master branch for a PR!

Will do. Thank you for the tip.

searx/engines/brave.py
Bnyro: Yes, it's the normal behavior that engines crash when result parsing fails - so this warning is indeed useless.

You can still do `!brave <short query>` to test if the engine works.

searx/engines/brave.py
```diff
@@ -261,155 +256,145 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
-    res = EngineResults()
-    dom = resp.html()
+def _parse_secondary_items(json_data: dict[str, t.Any], results: EngineResults):
+    # video results utilize same schema as video search -> re-use _parse_video_result
+    body_resp: dict[str, t.Any] = _get_response_data(json_data)
+    videos_resp: dict[str, t.Any] = body_resp.get("videos", {})
+    if videos_resp and "results" in videos_resp:
+        for result in videos_resp.get("results", []):
+            results.add(_parse_video_result(result))
+    # related queries -> suggestion (No MainResult?)
```
Bnyro:
```suggestion
    # related queries -> suggestion
```

searx/engines/brave.py
```diff
@@ -261,155 +256,145 @@ def extract_json_data(text: str) -> dict[str, t.Any]:
...
+    if query and "related_queries" in query:
+        for x in query.get("related_queries", []):
+            suggestion = " ".join(val[1] for val in x)
+            results.add(results.types.LegacyResult(suggestion=suggestion))
+
+
+def _parse_news_result(result: dict[str, t.Any]) -> MainResult:
+    thumbnail: dict[str, t.Any] = result.get("thumbnail", {})
+    return MainResult(
+        template="default.html",
```
Bnyro, line 306:
```suggestion
```
That's the default

Bnyro:
Please fix the remaining comments, then lgtm to merge :+1:

PR #6743, opened 2026-09-17
[mod] typification of SearXNG: add new result type Video

return42:

    !videos 

### Code of Conduct

  no AI


searx/result_types/video.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-License-Identifier: AGPL-3.0-or-later
+"""Typification of the *video* results.  Results of this type are rendered in
+the :origin:`video.html <searx/templates/simple/result_templates/videos.html>`
```
Bnyro:
```suggestion
the :origin:`videos.html <searx/templates/simple/result_templates/videos.html>`
```
nit: typo

searx/utils.py
```diff
@@ -795,7 +806,7 @@ def parse_duration_string(duration_str: str) -> timedelta | None:
 
     try:
         # prepending ["00"] here inits hours to 0 if they are not provided
-        time_parts = (["00"] + duration_str.split(":"))[:3]
+        time_parts = (["00"] + duration_str.split(":"))[-3:]
```
Bnyro, line 809: good catch!

Bnyro (approved):
Looks great 👍

searx/result_types/video.py
return42: fixed [diff](https://github.com/searxng/searxng/compare/758315b04aa6d24cc7e6690b2532482429a75dff..ebb0773431bc54c9b875b9ada11f135831265666)

vojkovic:
Thanks for taking the time to do this, looks really good. I have a couple review items I am just going through at the minute.

I suppose now is a good time to ask if we should keep the video embeds. Personally, I have never used them but I am open changing my opinion. What do you think?

return42:
That's a valid question; I sometimes use the embedded video as a preview.  For audio streams, I also use it to listen to entire tracks.

But the user experience and performance of the show/hide button aren't particularly good; whether it's a video stream or an audio stream, clicking “hide” doesn't stop the player, and the stream keeps playing .. which unnecessarily eats up bandwidth.

With iframe elements, there’s no other way to handle this (or?), but with HTML audio/video streams, there’s surely a better solution.

Back then, we chose iframes—albeit reluctantly—because it was the easiest to implement.

From a “privacy” standpoint, embedded streaming is questionable, but I just find it very convenient when I want to quickly take a look at a stream.

searx/engines/kagi.py
vojkovic, line 1: Needs to be converted from a dict into kwargs

searx/engines/dogpile.py
```diff
@@ -98,14 +98,13 @@ def response(resp: "SXNG_Response"):
...
-                res.types.LegacyResult(
-                    template="videos.html",
+                res.types.Video(
                     url=result["clickUrl"],
                     title=html_to_text(result["title"]),
                     content=html_to_text(result["description"]),
                     thumbnail=result["thumbnailUrl"],
                     publishedDate=datetime.fromisoformat(result["publishDate"]),
-                    length=format_duration(result["duration"]),
+                    length=timedelta(result["duration"]),
```
vojkovic:
```suggestion
                    length=timedelta(seconds=result["duration"]),
```
Screenshot

searx/engines/odysee.py
```diff
@@ -59,46 +64,39 @@ def request(query, params):
...
-            }
+
+        res.add(
+            res.types.Video(
+                title=item["title"],
+                url=url,
+                content=item["description"] or "",
+                author=item["channel"],
+                publishedDate=formatted_date,
+                length=timedelta(item["duration"]),
```
vojkovic:
```suggestion
                length=timedelta(seconds=item["duration"]),
```
Screenshot

searx/engines/kagi.py
return42: Aargh .. haven't tested / its fixed now, but I can't test since I don't have a kagi API key :-o

return42:
@vojkovic @Bnyro thanks for your reviews! .. the outstanding issues have now been resolved .. I'd like to move the discussion regarding embedded streams to another PR or another discussion.  It concerns both video and audio streams. I suspect we need to handle the streams / specifically their formats and availability .. differently .. perhaps something slightly similar to how we handle images, where we can already handle the MIME types separately. 

https://github.com/searxng/searxng/blob/1f7711ef1f16c80b24c093e0d39c34f22731424e/searx/result_types/image.py#L39-L56

----

One last note: not all video engines have been migrated to the Video class in this PR.  Some of the video engines that haven’t been migrated would need to be re-evaluated anyway... and if they continue to deliver good results, their implementation should be completely revamped.

searx/engines/kagi.py
vojkovic: Me neither... it's probably fine but maybe @Bnyro can test if he still has the key from https://github.com/searxng/searxng/pull/6244

Bnyro: Tested it, unfortunately the API key I used is already expired :/

vojkovic (approved):
I don't see the updated commit with the review fixes but once you push that, LGTM from me

return42:
> I don't see the updated commit with the review fixes but once you push that, LGTM from me

sorry, sure force-push would help :facepalm:

inetol (approved):
LGTM

PR #6754, opened 2026-09-18
[mod] categories: move app search engines to it category instead of files

Bnyro:
### What does this PR do?
- The files category is supposed to be for files that can be downloaded directly (e.g. torrents)
- Apps (e.g. Play store, F-Droid, ...) don't satisfy that. They are very similar to the `packages` category (one could say they're like linux packages, but for mobile devices). So imo they should be placed in the it category, not the files category. It's very confusing to have them in the files category while packages are in the it category.
- I also thought about merging the `apps` (mobile) and `packages` (desktop) categories, but not sure about that

### Code of Conduct

No AI used.


return42 (approved):
> * I also thought about merging the `apps` (mobile) and `packages` (desktop) categories, but not sure about that

IDK .. for me *packages* can be anything in a ecosystem, for instance a lib from pypi or npm (developer) .. or a office suite from distro's package manager (users).  The term *Apps* I would use for a self-contained application that can generally be installed and used directly by a user.

It is difficult to draw a clear distinction because the abstract umbrella term would certainly be *packages* :-)

Bnyro:
> IDK .. for me packages can be anything in a ecosystem, for instance a lib from pypi or npm (developer) .. or a office suite from distro's package manager (users). The term Apps I would use for a self-contained application that can generally be installed and used directly by a user.
>
> It is difficult to draw a clear distinction because the abstract umbrella term would certainly be packages :-)
>
Agreed, then let's keep the category called "apps" :+1: 

I think that I would also call many Linux programs "apps" although they are not self-contained (i.e. depend on other packages), but it's very hard to distinguish between that, yeah.

PR #6755, opened 2026-09-18
[fix] pixabay: crashes when there are no results

Bnyro:
### What does this PR do?
- If there are no results for a query, Pixabay redirects to another page with a similar query
- SearXNG previously tried to parse the JSON anyways, which made the engine crash

- !pixi tove lo

### Code of Conduct

No AI.


return42 (approved):
LGTM 👍

PR #6756, opened 2026-09-18
[del] library of congress: now behind cloudflare

Bnyro:
### Code of Conduct


return42 (approved):
Missing LOC is really too bad .. I disabled the engine a while ago on my instance :-o

PR #6758, opened 2026-09-18
[fix] engines: tusksearch cors request headers

vojkovic:
Added the ``Origin`` and ``Sec-Fetch-Site`` http headers to bypass tusksearch's latest bot detection.

Fixes all tusksearch engines

### Code of Conduct

Cursor to test


return42 (approved):
tested before and after the patch .. works 👍

PR #6759, opened 2026-09-19
[feat] categories: add icons for remaining category tabs

Bnyro:
### What does this PR do?
Add icons to these categories:
Screenshot from 2026-09-19 12-50-15
- otherwise it looks ugly if you enable some of these categories because they all have the same default "browser" icon that general has
- also 

- enable the following in `categories_as_tabs`:
```yaml
categories_as_tabs:
  packages:
  stock images:
  blogs:
  icons:
  movies:
  repos:
  weather:
  translate:
```
- enable at least one engine for each category (otherwise categories may be hidden)
- search something

### Related issues
- https://github.com/searxng/searxng/pull/6703

### Code of Conduct

No AI used.


return42:
@Bnyro I'm getting interrupted right now, so I've pushed my current progress to your branch... but I still need to test it... I'll explain that later too... just give me a little more time :-)

Bnyro:
> @Bnyro I'm getting interrupted right now, so I've pushed my current progress to your branch... but I still need to test it... I'll explain that later too... just give me a little more time :-)
>
Oh, I didn't get that the SVGs are auto-generated. We should properly describe that in the docstring of `icons.html` :)

return42:
> We should properly describe that in the docstring of `icons.html` :)

Good point, I added it .. but I'm still not done .. I'll continue tomorrow ..

return42:
I finished my rework for now; please see commit message for the details.

To test all icons I patched these two files ..

```diff
diff --git a/searx/engines/__init__.py b/searx/engines/__init__.py
index cb5e3a7d2..89600cb59 100644
--- a/searx/engines/__init__.py
+++ b/searx/engines/__init__.py
@@ -263,8 +263,8 @@ def is_engine_active(engine: "Engine | types.ModuleType"):
         return False
 
     # exclude onion engines if not using tor
-    if 'onions' in engine.categories and not using_tor_proxy(engine):
-        return False
+    # if 'onions' in engine.categories and not using_tor_proxy(engine):
+    #    return False
 
     return True
 
diff --git a/searx/settings.yml b/searx/settings.yml
index b587a4efc..3cf8c7d70 100644
--- a/searx/settings.yml
+++ b/searx/settings.yml
@@ -278,17 +278,30 @@ plugins:
 
 
 categories_as_tabs:
+  apps:
+  blogs:
+  dictionaries:
+  files:
   general:
+  icons:
   images:
-  # stock images:
-  videos:
-  news:
+  it:
+  lyrics:
   map:
+  movies:
   music:
-  it:
+  news:
+  onions:
+  packages:
+  radio:
+  repos:
   science:
-  files:
   social media:
+  stock images:
+  translate:
+  TV:
+  videos:
+
 
 engines:
   - name: 360search
@@ -844,6 +857,32 @@ engines:
   - name: duckduckgo
     engine: duckduckgo
     shortcut: ddg
+    categories: [
+      apps,
+      blogs,
+      dictionaries,
+      files,
+      general,
+      icons,
+      images,
+      it,
+      lyrics,
+      map,
+      movies,
+      music,
+      news,
+      onions,
+      packages,
+      radio,
+      repos,
+      science,
+      social media,
+      stock images,
+      translate,
+      TV,
+      videos,
+      ]
+
 
   - name: duckduckgo web
     engine: duckduckgo_web
```

Started a `make run` instance and I opened the preferences:

image

As you can see we have some icons used twice (e.g. apps & packages) .. and the new svg icons for onions and lyrics are *"improvable"*.

Bnyro:
> I finished my rework for now; please see commit message for the details.
> 
> Started a `make run` instance and I opened the preferences:
> image
> 
> As you can see we have some icons used twice (e.g. apps & packages) .. and the new svg icons for onions and lyrics are _"improvable"_.
>
Thanks a lot for the rework, now I also get how our icons work :) Thanks for the docs!

I didn't really understand the difference between `CATEGORY_NAMES` and `CATEGORY_GROUPS`, good to have them merged now :)

I noticed that the Onions / Tor icon is too small (as per your screenshot) and the lyrics icon has too thin strokes so that it doesn't match the style of the other icons. So I fixed the scaling of the Tor icon and replaced the Lyrics icon with the one from Google's Material icons: https://fonts.gstatic.com/s/i/short-term/release/materialsymbolsoutlined/lyrics/default/24px.svg.

image

Tested, and everything seems good now :+1:

vojkovic:
If I may just quickly chime in, from the screenshot the icon for 'Icons' has some spots coloured in black:

image

return42:
> the icon for 'Icons' has some spots coloured in black:

Yeah, thanks for pointing out .. the same issue is in the radio ico, the central dot is "black" and not "currentColor".

grafik

grafik

.. I will look for a fix.

return42:
I fixed the icons .. and added a new one for packages:

grafik

copied from the commit message:

```
[feat] categories: add icons for remaining category tabs (#6759)

Icons are provided for the main categories known to us.

The distinction between CATEGORY_GROUPS and CATEGORY_NAMES is abandoned in favor
of the latter.

HINT: we need single-color icons; transparent background and the color must be
`currentColor` so that the icons can be used in light/dark themes.

The following SVGs from the ionicons had to be revised (use ``currentColor``).

- client/simple/src/svg/ionicons/logo-rss.svg
- client/simple/src/svg/ionicons/radio-outline.svg
- client/simple/src/svg/ionicons/color-palette-outline.svg

For the following categories, I did not find any usable icons in Ionicons; new
SVGs were created for these categories.

- onions: client/simple/src/svg/ionicons/tor-browser.svg
- lyrics: client/simple/src/svg/ionicons/song-lyric-music.svg [1]
- packages: client/simple/src/svg/ionicons/package_2.svg [1]

[1] Based on Google's Material icons: https://fonts.google.com/icons
```

return42:
> replaced the Lyrics icon with the one from Google's Material icons

Maybe we should think about an alternative *icon pool*.  There are two things that bother me about IonIcons:

1. The selection is very limited; common icons like “tor” or “lyrics” are missing.
2. The icons aren't implemented properly; for light/dark mode, we have to rework some of them (such as logo-rss.svg, radio-outline.svg, and color-palette-outline.svg in this PR).

return42 (approved):

vojkovic (commented):
I have pushed a small [WIP](https://github.com/searxng/searxng/pull/6759/commits/3dc11e88872fbf9caef9ca6f8bee9c56db6ea3b1) commit to your branch with these changes:

- Fixed the icon catalog docstring
- Removed the xml declarations for the lyrics and tor svgs so they are like the other icons.
- Made TV category lowercase everywhere because it was a mix, now it is like the IT category. I think no engines are even in this category - can it just be removed?

return42:
@vojkovic .. thanks for the fix!

> * Made TV category lowercase everywhere because it was a mix, now it is like the IT category. I think no engines are even in this category - can it just be removed?


:thinking: No, please not... To me, TV refers to broadcast channels with live programming, so it's something completely different from “movies.”  Once again, the distinction isn't easy to make, since TV channels often have media libraries where ‘movies’ can be found.  Still, as a user, I would expect a TV channel to be excluded from the term “movies.”

Note on the capitalization of “IT” and “TV” (at least in German-speaking countries):

These are abbreviations where I would expect capital letters; I would rather convert “it” to “IT” than ‘TV’ to “tv.”


As far as I recall

- capital letters are also used for abbreviations in English-speaking countries ..
- we used lowercase letters for that because the !bangs (search syntax) still distinguished between uppercase and lowercase letters (?)

vojkovic:
Thanks for the reply... maybe I do not know how people are using the categories. I was just commenting that we don't have any engines in this category at all. You can ignore that change if you would like.

---

Now that I think about it a little more, my 2c is that we should try actually try to avoid acronyms altogether: ``IT`` could become ``Tech`` and ``TV`` could become ``Television``. But you can ignore this if you want.

Also FYI, the capitalisation is being done via CSS:
music category

return42:
> Now that I think about it a little more, my 2c is that we should try actually try to avoid acronyms altogether:

Really? .. :thinking: .. I don't know if I'm the only one who does this, but I often use the search syntax and like to type as little as possible (`!TV` <-> `!Television`). Acronyms also have the advantage that they are the same in so many languages.

> Also FYI, the capitalisation is being done via CSS:

If we avoid acronyms (upper letters) capitalize is OK, but not when we use lower case acronyms, for instance look at `it` -> `It` ..

grafik

Here the example from the acronym TV (upper letters) ..

grafik


----

> I was just commenting that we don't have any engines in this (TV) category at all. 

Sure, its an valid argument .. I've been using this category for years in my `categories_as_tabs`... please cut me some slack if I'm being a little self-centered here and would like to keep this category .. otherwise, I'd lose the icon from the UI that my users have already gotten used to ..  *thanks* :-)

Bnyro:
> Really? .. 🤔 .. I don't know if I'm the only one who does this, but I often use the search syntax and like to type as little as possible (!TV <-> !Television). Acronyms also have the advantage that they are the same in so many languages.
>
Abbreviations are easier to type than the full category names, so I agree with @return42 on that. Obviously that only makes sense if the abbreviations are widely known and everybody knows what they mean.

The `It` category should probably be called `IT` in the UI, otherwise it's not really clear that it's an abbreviation, we should change that.

Generally, I don't think it hurts to have more categories than we actually need/use. It doesn't take any maintenance efforts to just keep more icons and category names in place, and gives admin such as @return42 more freedom in which categories they want to display to the user.

vojkovic:
Fair enough. I just didn't really know how people were using these it seems :-)

Thanks for sharing @return42 - More than happy for you remove my change

return42:
> The `It` category should probably be called `IT` in the UI, otherwise it's not really clear that it's an abbreviation, we should change that.

`It` is an old issue .. and we need to find a smart way where 

- we can use `it` (lowercase) in the config and the engine modules 
- we have a `IT` in the GUI elements, in the UI translations and
- where we have `!it` in the search syntax

I'm not sure if we need to make any further modifications to implement this, so I'd like to move this issue to a separate PR.

----

> More than happy for you remove my change

I merged your 3dc11e88872fbf9caef9c, except the *TV* / *tv* patch, I [reverted](https://github.com/searxng/searxng/compare/3dc11e88872fbf9caef9ca6f8bee9c56db6ea3b1..42d9261e8b5ce09ebca83f28508c9dfaa24c7e97)

----

From my side this PR is ready to merge.

vojkovic (approved):

vojkovic:
> `It` is an old issue

Yes, it should be for another PR. TBH I have never been a fan of the name which is why I offered to change it to 'Tech' which is a lot clearer IMO, still short, and also wont be confused for :it (Italiano). But that's for another discussion.

Bnyro:
Thanks a lot for your help on this @return42 and @vojkovic, I feel like I'm the one that opened the PR but you're the ones that did the work :laughing:  

> Yes, it should be for another PR. TBH I have never been a fan of the name which is why I offered to change it to 'Tech' which is a lot clearer IMO, still short, and also wont be confused for :it (Italiano). But that's for another discussion.
>
Agreed, `Tech` is more self-explaining than `IT`, and the Italiano distinction is a valid point.

return42:
> Maybe we should think about an alternative _icon pool_. There are two things that bother me about IonIcons:
> 
> 1. The selection is very limited; common icons like “tor” or “lyrics” are missing.
> 
> 2. The icons aren't implemented properly; for light/dark mode, we have to rework some of them (such as logo-rss.svg, radio-outline.svg, and color-palette-outline.svg in this PR).

Just wanted to ref the discussion at:

- https://github.com/searxng/searxng/pull/6541#issuecomment-5792605820

PR #6763, opened 2026-09-21
[fix] engine: braveapi - forward the SearXNG locale as country and search_lang

geertvanzoest:
## What does this PR do?

The Brave Search API accepts `country` (ISO 3166 alpha-2) and `search_lang` (ISO 639) query arguments, but the `braveapi` engine never sent them, so Brave always fell back to its default region (US) regardless of the locale selected in SearXNG. With `nl-NL` a query for `aldi` returned aldi.us, Instacart and Google Play (US) before aldi.nl.

This derives both arguments from `params['searxng_locale']`:

| searxng_locale | sent to Brave |
|---|---|
| `nl-NL` | `country=NL&search_lang=nl` |
| `nl` | `search_lang=nl` |
| `all` / unset | nothing (Brave defaults) |

and sets `language_support = True` so the locale selector applies to the engine. The scraping engine `brave.py` already does the equivalent via the `country` and `ui_lang` cookies.

Region/language selection is a basic expectation for a general engine; without it the API engine is unusable as the main general engine on a non-US instance.

Configure `braveapi` with an API key, search with `:nl-NL aldi` and `:en-US aldi` — the request URL (debug log) now carries `country`/`search_lang` and the results differ per region. `:all` sends neither.

## Related issues

Related: #5728 (engine introduction)


return42:
> The Brave Search API accepts `country` (ISO 3166 alpha-2) and `search_lang` (ISO 639) query arguments, but the `braveapi` engine never sent them,

`ui_lang` and `country` are send within the cookies:

https://github.com/searxng/searxng/blob/49064747a2df982ac963a8dee922febf07ae6a2d/searx/engines/brave.py#L221-L225

If this is no longer valid, it should fixed in the lines from above.

----

BTW: next time please use the PR template and accept our project rules.

return42:
Sorry, I mixed brave.py with braveapi.py, but the solution will be similar (engine traits can be build from the API description):

- [country](https://api-dashboard.search.brave.com/api-reference/web/search/get#parameter-country)
- [search_lang](https://api-dashboard.search.brave.com/api-reference/web/search/get#parameter-search_lang)
- [ui_lang](https://api-dashboard.search.brave.com/api-reference/web/search/get#parameter-ui_lang)

grafik

PR #6381, opened 2026-07-08
Fix Brave API HTML tags and favicon thumbnail issues (#6375)

jayant733:
### What does this PR do?

This PR fixes the Brave API issues reported in #6375.

Changes:
- Remove HTML tags from result titles and descriptions before displaying them.
- Skip thumbnails that are site logos/favicons.
- Handle cases where the API returns `"web": null` without raising an exception.

### Testing

To verify the changes, I tested the Brave API engine locally.

1. Created a Brave Search API key.
2. Added the API key to `settings.yml`:

```yaml
engines:
  - name: braveapi
    engine: braveapi
    api_key: "YOUR-BRAVE-API-KEY"
    inactive: false
```

3. Restarted the SearXNG server:

```bash
env PYTHONPATH=. local/py3/bin/python -m searx.webapp
```

4. Searched for queries such as **"python programming"** and verified that:
   - Titles and descriptions are displayed without HTML tags.
   - Site logos/favicons are not shown as thumbnails.
   - The engine continues to work correctly even when the API returns an unexpected response.

### Related issue

Fixes #6375

### Code of Conduct

- [x] I hereby confirm that this PR conforms with the AI Policy.

### AI Usage

Tool used: Antigravity.

I used Antigravity to help inspect the codebase and understand the reported issue. I reviewed the suggested changes, implemented the fix, tested everything locally with a Brave API key, and verified the final behavior before opening this PR.
Also to write the pr.

PR #6764, opened 2026-09-21
[mod] engine: braveapi - add 'thumbnails' setting to drop web-result thumbnails

geertvanzoest:
## What does this PR do?

Adds an engine setting `thumbnails` (default `true`, current behaviour unchanged) to the `braveapi` engine. With `thumbnails: false` web results are rendered without the thumbnail Brave attaches, like most other general engines.

Brave serves result thumbnails through its own image CDN (`imgs.search.brave.com`), which re-fetches the origin image on demand. For origins with expiring signed URLs (Facebook CDN, YouTube channel avatars, ...) that fetch fails with HTTP 403/404, so the result list shows the theme's image-load-error placeholder on seemingly random results. Measured on a live instance: 2 of 14 thumbnails for a single query failed this way. Instances that prefer a clean general result list can now opt out.

```yaml
- name: braveapi
  engine: braveapi
  api_key: '...'
  thumbnails: false
```

Search for `aldi`: no thumbnails on braveapi results. Without the setting the behaviour is unchanged.

## Related issues

Related: #6381 (favicon-thumbnail filter)

## Code of Conduct

- [ ] **I hereby confirm that this PR conforms with the [AI Policy].**


return42:
Please accept *Code of Conduct* and we can reopen.

PR #6766, opened 2026-09-21
[fix] video thumbnails on left side again

vojkovic:
Since #6743, the videos stylesheet isn't working:

- Imports video.less now to apply the stylesheet.
Screenshot

- Ammends thumbnail_bg to dim the background and blur it more to make the thumbnail easier to see.

Screenshot

Prior (not the best example):
Screenshot

### Related issues

- https://github.com/searxng/searxng/pull/6679#issuecomment-5749128329

### Code of Conduct

No ai


client/simple/src/less/style.less
```diff
@@ -1161,3 +1161,4 @@ pre code {
 @import "result_types/paper.less";
 @import "result_types/file.less";
 @import "result_types/image.less";
+@import "result_types/video.less";
```
return42, line 1164: > Since https://github.com/searxng/searxng/pull/6743, the videos stylesheet isn't working:
> Imports video.less now to apply the stylesheet.

Aargh, I'm sorry .. in the initial commit [a618f61](https://github.com/searxng/searxng/commit/a618f61576f4128978dc44ab35f1e1032d1c4b7d#diff-fde5a1148569ae250fe0ba91c27dfb0df16fb03932220c0c878ba28e32c82723R1132) I added this line, but I lost the line somewhere in one of the rebase cycles .. thanks a lot for the fix 👍

---

upd: I added `(#6766)` suffix to the commit messages.

return42 (approved):

client/simple/src/less/style.less
vojkovic: > upd: I added `(#6766)` suffix to the commit messages.

Ahhh thank you, let me know if I have this right: the suffix is just because we can't squash and merge because of the 'build' commit?

return42: > the suffix is just because we can't squash and merge because of the 'build' commit?

right .. [In such cases - that is, when keeping the commits separate is relevant to the content (for whatever reason) - we use GH-rebase. However, this means the PR number isn’t included in the commit; that’s the trade-off ... Sometimes, when I feel it’s very important to include the PR number in the GH-rebase merge, I edit the commit messages again before the merge .. but since I’m a lazy person, I only do that very rarely :-)](https://github.com/searxng/searxng/pull/6730#issuecomment-5696550597)

vojkovic: Thanks, do you guess the PR number by the one before it or do you usually just add it after you've created the PR by going back and adding it?

Also sorry if that is a stupid question :-)

return42: If, I add the PR number after the PR is created .. and sometimes I'm asleep and make a typo .. https://github.com/searxng/searxng/pull/6766/commits

grafik
 
As I've already mentioned: it's not that important - if you have the time and feel like it, you can do it, but I wouldn't want to force these rules on everyone.

vojkovic: Thanks very much! I don't really have a style of my own so I will try to make an effort :)

PR #6768, opened 2026-09-22
[fix] duckduckgo_web: make preload link timeout configurable

### LINKED ISSUE #6762: duckduckgo_web hardcodes timeout and ignores global settings
### Description

The duckduckgo_web engine passes `timeout=2` explicitly to `searx.network.get`, which overrides the thread timeout set from `outgoing.request_timeout`. There is no way to tune it via settings. With added latency such as a challenge solver via MITM proxy, this regularly fails and suspends the engine (ban_time_on_fail). 

This setting should follow the global timeout or at least be extended to a bare minimum of 5-10 seconds.

### How to reproduce

Use duckduckgo_web engine and proxy it via a solver such as Trawl. Issue a search request and inspect searxng log: 
```
WARNING:searx.engines.duckduckgo web: ErrorContext('searx/engines/duckduckgo_web.py', 72, 'resp = get(', 'curl_cffi.requests.exceptions.Timeout', None, (None, None, 'duckduckgo.com')) False
```

### Screenshots & Logs

_No response_

### Version of SearXNG

2026.9.19-e831fc2a1

Docker Compose

### Code of Conduct

quackerd:
### What does this PR do?

This PR actually fixes #6762 by adding a `duckduckgo_web` exclusive parameter `preload_link_timeout` that allows specifying the timeout for fetching the `deep_preload_link`. Previously the timeout was hardcoded to 2 which is too short and breaks challenge solver MITM proxies, as the issue discussed. The default value of `5.0` was chosen to be consistent with the default timeout value in `duckduckgo.py`.

The previously rejected PR #6767 removes the timeout and falls back to the default timeout, which is incorrect because `deep_preload_link` is not the actual request but a prereq request that constructs the final request, hence it shouldn't inherit global settings.

A. build a local searxng container.
B. configure searxng with ddg_web (config/settings.yml):
```yaml
use_default_settings:
  engines:
    keep_only: [duckduckgo_web]

search:
  formats: [html, json]

server:
  secret_key: "test"
  limiter: false
  public_instance: false

outgoing:
  request_timeout: 30
  verify: false
  proxies:
    all://:
      - http://trawl:8192

engines:
  - name: ddg
    engine: duckduckgo_web
    preload_link_timeout: 10
    disabled: false
```
C. Wire up searxng with trawl using the sample docker compose:
```yaml

name: ddgtest

services:
  trawl:
    image: ghcr.io/germondai/trawl:latest
    shm_size: 1gb
    environment:
      MITM_ENABLED: "true"
      MITM_HOST: "0.0.0.0"
      MITM_PORT: "8192"
      MITM_DEBUG: "true"
    networks: 
      - ddgtest

  searxng:
    image: localhost/searxng/searxng:latest
    environment:
      SEARXNG_PORT: "8080"
      FORCE_OWNERSHIP: false
    volumes:
      - ./config:/etc/searxng
    ports:
      - "127.0.0.1:8080:8080"
    depends_on:
      - trawl
    networks: 
      - ddgtest

networks:
  ddgtest:

```

D. `docker compose up -d`
E. Run a query `curl 'http://localhost:8080/search?q=searxng&format=json'` and it should succeed.
F. Change `preload_link_timeout` from 10 to 0.1 in `settings.yml`.
G. `docker compose up -d --force-recreate`
H. Run the same query now the query should fail and searxng logs will reproduce the exact log in #6762 

The above workflow was already verified locally. `./manage test.unit` was also successful but it is less relevant to this specific PR.

### Related issues

Closes: #6762 

### Code of Conduct

  that I haven't used any other tools than the ones I mention below:

No AI was used in this PR.


Bnyro (commented):
Thanks for looking into this :+1: 

Wouldn't it be better to just not set a timeout at all so that the request uses the engine-configured `timeout` https://docs.searxng.org/dev/engines/enginelib.html#searx.enginelib.Engine.timeout, i.e. the `timeout` variable from the YAML config?

I don't see why we should add a separate parameter for that, I think that's kind of confusing.

quackerd:
I think semantically the globally configured timeout is referring to the final request that actually fetches the search query or at least the total timeout of all requests needed for one search query. However, the deep_link fetch request is a prerequisite request used to construct the final request. The same quest chain and standalone timeout pattern also appears in `DuckDuckGo.py` and that was how I rationalize the choice.

Removing the timeout makes sense if the `request_timeout` parameter means every individual request including prerequisite requests. Just that potentially each query can take a multiple of `request_timeout`.

With that said if the maintainer thinks the 2nd semantic makes more sense I can simply remove the timeout.

Bnyro:
> Removing the timeout makes sense if the request_timeout parameter means every individual request including prerequisite requests. Just that potentially each query can take a multiple of request_timeout.
> 
To be honest I'm not 100% sure anymore if the `get` method uses the engine's timeout by default, I don't think it does. So removing the timeout argument probably won't do it. Anyways, we could set `get(..., timeout=timeout)` to use the value from the configuration (it's set to 3.0 by default if it's not overriden anywhere).
> With that said if the maintainer thinks the 2nd semantic makes more sense I can simply remove the timeout.
>
Yeah, I think that would be cleaner.

quackerd:
Per our discussion, I made the preload link request timeout use the *remaining budget* of global configuration `outgoing.request_timeout` by simply removing the `timeout = 2`. The `network.get` function reads the `timeout` from thread local storage if not explicitly passed, which was populated by `search.initialize()` -> `OnlineProcessor.init_engine()` -> `init_network_in_thread(..., timeout_limit = self.engine.timeout)`. `self.engine.timeout` is populated by `outgoing.request_timeout`

To test the new workflow, similarly use the above setup and change the `outgoing.request_timeout` to 10 and 0.1, then inspect the searxng debug log. The 10s timeout request should succeed whereas the 1s timeout request should timeout right at the preload link fetch request.

